PR #39835, opened 2025-05-20
Only reuse SMTP authentication data for testing endpoint when the same auth, host, port and user are passed

### LINKED ISSUE #39486: Email server credentials can be harvested through host/port manipulation
### Before reporting an issue

### Area

admin/api

### Describe the bug

Attack scenarios:
1. Using test connection:

### Version

26.1.5

### Regression

### Expected behavior

### Actual behavior

### How to Reproduce?

Method 1 (Test Connection):

Method 2 (Email Workflow):
3. Save the configuration

### Anything else?

  - Forgot Password flow
  - Email verification

rmartinc:
Closes #39486

Adding more or less the same checks that were added to the test ldap endpoint. The authentication data is only added from the config if the passed connection details are the same (authenticator, host, port and user).


mposolda (approved):
@rmartinc Thanks!

PR #39845, opened 2025-05-20
Revert changes to exception handling in RealmsAdminResource#importRealm

### LINKED ISSUE #39753: POST realm API returns 400 on conflict instead of 409 in version 26.2.4
### Before reporting an issue

### Area

admin/api

### Describe the bug

In the latest release of Keycloak, 26.2.4, the POST /admin/realms API now returns a 400 status code if the realm already exists instead of the 409 status code returned previously.

This is the full response:
```
400: b'{"errorMessage":"Realm <name> already exists"}'
```

This API changes breaks a number of keycloak clients that expect a 409 status code to be returned when a realm already exists.

### Version

26.2.4

### Regression

### Expected behavior

If the realm already exists, the API should return a 409 status code.

### Actual behavior

The API returns a 400 status code when the realm exists.

### How to Reproduce?

Call the POST /admin/realms API with the name of a realm that already exists, and the response status code will be 400 instead of 409.

### Anything else?

_No response_

Example in postman: 

Image

sguilhen:
- ModelDuplicateException and ModelIllegalException were wrongfully handled as ModelException, returning wrong status code

Signed-off-by: Stefan Guilhen <sguilhen@redhat.com>

Closes #39753

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/RealmsAdminResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/RealmsAdminResource.java
@@ -176,6 +176,12 @@ public class RealmsAdminResource {
         } catch (PasswordPolicyNotMetException e) {
             logger.error("Password policy not met for user " + e.getUsername(), e);
             throw ErrorResponse.error("Password policy not met. See logs for details", Response.Status.BAD_REQUEST);
+        } catch (ModelIllegalStateException mise) {
+            logger.error(mise.getMessage(), mise);
+            throw ErrorResponse.error(mise.getMessage(), Response.Status.INTERNAL_SERVER_ERROR);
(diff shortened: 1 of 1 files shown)


sguilhen:
Simply catching/rethrowing the exceptions to prevent them from being handled as ModelException so they can be handled by KeycloakErrorHandler instead doesn't work well because the error handler doesn't seem to process the error messages correctly - in ModelDuplicateException case, the error message ends up being set as "unknown_error" with no description due to how the exception is handled.

I think we need to revisit the error handler so it can handle the model exceptions more adequately, but this should probably be handled in a major release as the exception handling cleanup in multiple endpoints might result in different error codes being returned in some cases.

pedroigor (approved):

PR #39882, opened 2025-05-21
Picture of the token-exchange flow in the documentation

### LINKED ISSUE #39881: Picture of the token-exchange flow in the documentation
### Description

Can we have some simple picture of the token-exchange flow in the documentation? Maybe it is nice for the illustration as the token-exchange flow is relatively non-trivial.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mposolda:
closes #39881

The picture for illustration of the token-exchange flow. It uses same notation (client names etc) as the token-exchange flow example from above.


graziang (approved):
@mposolda Thanks! Sequence diagrams are always helpful to better understand the flow

andymunro (approved):
Just back from PTO. Looks good, @mposolda

mposolda:
> @mposolda - is there a source code to this image so future contributors can update it? IMHO it would be beneficial to add the source next to the file in the same place plus maybe a README or TXT file stating which software is used to modify the source file.

@ahus1 That is good point and I am sorry for the late response :-) I've originally used some first tool I found for sequential diagram, but it was not good (and was not open source). I rewrote that with the use of draw.io - the same stuff, which you guys use for the diagrams in the HA guide. What do you think now?

ahus1 (approved):
Thank you for this change! 

There is the convention to name the Draw.io / diagrams.net files ".dio.svg" to show that they are editable. While it is not a technical requirement, it shows people that it is editable (at least when they know about the convention and the tool). So I renamed the file, and this is now ready to be merged.

PR #39903, opened 2025-05-22
LDAP group mapper skips configured filter and imports all groups with…

### LINKED ISSUE #37537: LDAP group mapper skips configured filter and imports all groups with memberOf strategy when fetching the user's groups
### Before reporting an issue

### Area

ldap

### Describe the bug

When a group mapper is configured with a filter and the `GET_GROUPS_FROM_USER_MEMBEROF_ATTRIBUTE`  strategy, the filter is respected when syncing the LDAP groups into Keycloak. However, if we select an LDAP user in the `Users` screen, and go to `Groups`, the `getGroupsStream()` call ends up fetching all groups it finds under the `memberOf` attribute in the user LDAP entry and imports them, ignoring the configured filter.

### Version

26.1.2

### Regression

### Expected behavior

Only the groups that match the ldap group mapper filter should be synced

### Actual behavior

All groups found under the `memberOf` attribute in the user entry are synced

### How to Reproduce?

- Start the embedded server (`mvn -f testsuite/utils/pom.xml exec:java -PLdap`)
- Using a tool like Apache Directory Studio, edit the users to reference the groups they belong to using an attribute (instead of the `member` attribute in the groups). For simplicity, add group CNs using a regular attribute like `postalCode` to avoid having to configure the member-of overlay as that attribute is not availble out of the box).
- Create an LDAP provider with a group mapper using the `GET_GROUPS_FROM_USER_MEMBEROF_ATTRIBUTE` strategy  and `Preserve Group Inheritance` set to `Off`. Use the `postalCode` as the memberOf attribute in the config.
- Add a group filter - something like `(cn=ldap-admin)`
- Sync the groups and see that only the `ldap-admin` group was imported
- Go to Users then select a user that belongs to more groups, and then click on Groups
- Check that all groups the user is a member of have been imported, ignoring the filter

### Anything else?

_No response_

vramik:
… memberOf strategy when fetching the user's groups

Closes #37537

### DIFF
--- a/federation/ldap/src/main/java/org/keycloak/storage/ldap/mappers/membership/UserRolesRetrieveStrategy.java
+++ b/federation/ldap/src/main/java/org/keycloak/storage/ldap/mappers/membership/UserRolesRetrieveStrategy.java
@@ -31,9 +31,7 @@ import org.keycloak.storage.ldap.idm.query.internal.LDAPQueryConditionsBuilder;
 import org.keycloak.utils.StreamsUtil;
 
 import java.util.Collections;
-import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


sguilhen (approved):
LGTM, thanks @vramik !

pedroigor (approved):

PR #39907, opened 2025-05-22
Email server credentials can be harvested through host/port manipulation

### LINKED ISSUE #39486: Email server credentials can be harvested through host/port manipulation

### LINKED ISSUE #39931: Changing SMTP host without resetting credentials
### Before reporting an issue

### Area

admin/api

### Describe the bug

Despite partial mitigation in PR #39835, the email server configuration in Keycloak still contains a security vulnerability that allows credential harvesting. 

This issue is particularly critical in production environments, where the SMTP server is not an isolated component, but is often integrated with other internal systems. If an attacker can extract SMTP credentials, they may be able to reuse them to gain unauthorised access to other internal resources.

### Version

nightly

### Regression

### Expected behavior

From a security standpoint, the password field should be cleared if the host is changed.

### Actual behavior

An attacker with `manage-realm` permissions can modify the SMTP server's host and port without having to re-enter the stored credentials. This enables credential theft via a passback attack, whereby Keycloak is directed to connect to an attacker-controlled server.  

### How to Reproduce?

Look at the related issue: #39486 

### Anything else?

_No response_

Foorcee:
Closes #39931

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
@@ -123,6 +123,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


ahus1:
@Foorcee - there has already been PR #39835 which closed the issue you referenced. So is this PR still necessary?

Foorcee:
@ahus1 PR #39835 partially fixed the issue, it did not resolve all the cases described in issue #39486. The aim of this PR is to close those remaining gaps.

The SMTP server is a critical resource within the company, and from a security perspective, it's important that this is handled appropriately at this point.

ahus1:
@Foorcee: As the parent issue has already been closed and backported, please create a new issue so this PR can be tracked independently (except if you have agreed something different with @mposolda and @rmartinc)

Foorcee:
@ahus1 Thanks! I've reopened the issue as a bug, even though it's technically a vulnerability. I don't know how much public discussion is intended here, but since the original issue was already public, I have followed that precedent.

rmartinc:
@Foorcee The second scenario is not only happening for the SMTP server, it's the same for the LDAP configuration (user providers in general?), all Identity Providers or any thing that uses a password for a server that can be faked (more or less complicated depending the protocol nature of the server). So I agree with @ahus1 and it's better to file a new issue for the whole problem. Keycloak should think in the scenario as a whole and if a solution for that is really worthy (because as I commented using the vault key, the admin can always know the password setting a fake server).

Closing the issue, file a new one commenting the second scenario is more general and not only for the SMTP configuration.

Issue #20128, opened 2023-05-03
Add support for `reencrypt` OpenShift Routes to the Operator


vmuzikar:
The Operator currently creates `passthrough` Routes by default (if TLS is not configured, unlike Nginx Ingress that uses uses a reencrypt strategy. Passthrough has (among others) performance implications on Keycloak.

We should either make `reencrypt` the default strategy, or make it configurable (affecting both Routes and Nginx Ingress).

**Considerations**
* This is potentially a breaking change.
* `reencrypt` requires configuring the Route to trust the Keycloak's certificate. This might be a bit tricky doing manually and not something we want to force the users to do. Keycloak should be automatically configured to use TLS key/cert that is generated by OpenShift CA (and therefore implicitly trusted by the Routes).

**Workaround**
Disable the default Ingress and create a custom Route.

vmuzikar:
The routing strategy should probably honor our `proxy` settings, see https://github.com/keycloak/keycloak/issues/14400

ckupe:
This currently breaks deployments of Keycloak on Openshift, in the use case where we are trying to use the cluster-deployed instance of keycloak as an OIDC IdP for the Openshift cluster Oauth client.

Redirection in the middle of auth chain with the default `passthrough` encryption results in "Page not Found" sending the user back to the keycloak pod.

With wildcard certs and openshift-ingress, this is a poor UX. Hard refreshing the browser session has varying success depending on the browser.

gilvansfilho:
@shawkins as we talked about...

To provide first class support to Openshift reencrypt routes I think we need below scenarios.

| Scenario | http.tlsSecret | ingress.tlsSecret | ingress.mode | service annotation                                                        | pod certificate      | route certificate             |
|----------|----------------|-------------------|--------------|---------------------------------------------------------------------------|----------------------|-------------------------------|
| 1        | (not defined)  | (not defined)     | reencrypt    | service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret | generated-tls-secret | (cluster default certificate) |
| 2        | pod-tls-secret | (not defined)     | reencrypt    |                                                                           | pod-tls-secret       | (cluster default certificate) |
| 3        | (not defined)  | custom-route-tls  | reencrypt    | service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret | generated-tls-secret | custom-route-tls              |
[...]

Issue #15395, opened 2022-11-08
Add support for custom labels for all Operator managed resource


vmuzikar:
Add a field to Keycloak CR for setting custom additional labels on all K8s resources that are managed by the Operator.

vmuzikar:
CC @jsorah

vmuzikar:
Possibly blocked by #13453.

csviri:
this could be also a feature to [dependent resources](https://javaoperatorsdk.io/docs/dependent-resources), we already support label selector but the labels are not explicitly added. cc @metacosm @shawkins

Jamstah:
When looking at custom labels, it might be sensible to also look at custom annotations. We've found in the past with our customers that after providing labels, a request for annotations comes closely after :)

shawkins:
Managed kafka simply made the assumption that all labels from the CR would be propogated to the Kafka resource.  However it did not make that assumption for annotations probably because it wanted to be more intentional about what was passed through.

A similar case could be made to set the cr labels on all the managed resources - whatever filtering you want to do for the main cr you probably want to do for the managed as well.

Jamstah:
I would say k8s goes the other way, labels set on a deployment are not automatically set on it's pods.

shawkins:
> I would say k8s goes the other way, labels set on a deployment are not automatically set on it's pods.

I'm not sure that's a good example because of the explicit PodTemplate or are you saying that the labels for managed resources for keycloak should be coming from the unsupported PodTemplate?

Jamstah:
I was really just making the counter point to what managed kafka is doing and trying to think about examples as close to base kubernetes as I could.

I don't have a strong opinion on what the right thing for keyboard to do is.

csviri:
Just thinking on JOSDK level:

> Managed kafka simply made the assumption that all labels from the CR would be propogated to the Kafka resource.

This seems to me quite a strong assumption to be applied in general. But could be a feature flag for Dependent Resource to copy the labels and/or annotations from the primary resources.

shawkins:
Services and Ingress are now covered. That primarily leaves the Keycloak statefulset/pods via a mechanism that is not the unsupported podTemplate. The Jobs (import and update) could also be targeted individually. Also NetworkPolicies.

shawkins:
[...]

PR #39925, opened 2025-05-23
add labels and annotations to service (httpSpec)

### LINKED ISSUE #23283: Allow Keycloak operator to parameterize the Service annotations and labels
### Description

The use case is to deploy Keycloak on OpenShift, and leverate the OpenShift Service Certificates for the Keycloak TLS certificate, see https://docs.openshift.com/container-platform/4.12/security/certificates/service-serving-certificate.html.

On OCP, there is a way to request OCP to auto-generate the Certificate and put it into the Secret, for any Service which has annotation like `service.beta.openshift.io/serving-cert-secret-name=<secret_name>`.

However, currently Keycloak operator does not allow for any parameterization of the created Service, see https://github.com/keycloak/keycloak/blob/main/operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceDependentResource.java#L79 - meaning there is no way to pass any annotation or label via Keycloak CR for the Keycloak opeator to apply on the created Service resource.

The ask here is to introduce a parameterization via Keycloak CR of the created Service resource, so that one can pass arbitrary annotation or label into the Service resource.

Conceptually it's similar to the `.spec.unsupported.podTemplate`, where one can override the definition of Keycloak Pod, https://github.com/keycloak/keycloak/blob/main/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeployment.java#L182C23-L182C23)

Now, it's true that OCP creates the certificate/secret AFTER the Service is created, which means that Keycloak CR would point to the to-be-created TLS Secret and must tolerate the situation that the Secret does not yet exist and try to create the Service anyway, which is similar to https://github.com/keycloak/keycloak/issues/23087, but requires more thoughts in terms of coordinating asynchronous operations - which most probably would require separate work item.

### Discussion

_No response_

### Motivation

Don't need to externally supply the certificate for Keycloak, leverage the built-in OpenShift mechanism of service certificates.

### Details

_No response_

gilvansfilho:
closes #23283

### DIFF
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -473,4 +473,25 @@ They need to access {project_name} to scrape the available metrics.
 
 Check the https://kubernetes.io/docs/concepts/services-networking/network-policies/[Kubernetes Network Policies documentation] for more information about NetworkPolicies.
 
+=== Parameterizing service labels and annotations
+
+If you need to set custom labels or annotations to keycloak service you can do that through `spec.http.labels` and `spec.http.annotations`
+
+.Custom service labels and annotations
+[source,yaml]
+----
+apiVersion: k8s.keycloak.org/v2alpha1
+kind: Keycloak
+metadata:
+  name: example-kc
+spec:
+  http:
+    labels:
+      label1: label-value1
+      label2: label-value2
+    annotations:
+      annotation1: annotation-value1
+      annotation2: annotation-value2
+----
+
 </@tmpl.guide>
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceDependentResource.java
@@ -17,6 +17,9 @@
 package org.keycloak.operator.controllers;
 
 import java.util.Optional;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Service;
@@ -76,11 +79,19 @@ public class KeycloakServiceDependentResource extends CRUDKubernetesDependentRes
 
     @Override
     protected Service desired(Keycloak primary, Context<Keycloak> context) {
+
+        Map<String,String> labels = Utils.allInstanceLabels(primary);
+        var optionalSpec = Optional.ofNullable(primary.getSpec().getHttpSpec());
+        optionalSpec.map(HttpSpec::getLabels).ifPresent(labels::putAll);
+
+        Map<String,String> annotations = optionalSpec.map(HttpSpec::getAnnotations).orElse(new HashMap<>());
+
         Service service = new ServiceBuilder()
                 .withNewMetadata()
                 .withName(getServiceName(primary))
                 .withNamespace(primary.getMetadata().getNamespace())
-                .addToLabels(Utils.allInstanceLabels(primary))
+                .addToLabels(labels)
+                .addToAnnotations(annotations)
                 .endMetadata()
                 .withSpec(getServiceSpec(primary))
                 .build();
--- a/operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/spec/HttpSpec.java
+++ b/operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/spec/HttpSpec.java
@@ -18,11 +18,14 @@
 package org.keycloak.operator.crds.v2alpha1.deployment.spec;
 
 import java.util.Optional;
+import java.util.Map;
+
+import org.keycloak.operator.Constants;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+
 import io.sundr.builder.annotations.Buildable;
-import org.keycloak.operator.Constants;
 import org.keycloak.operator.crds.v2alpha1.CRDUtils;
 import org.keycloak.operator.crds.v2alpha1.deployment.Keycloak;
 import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakSpec;
@@ -44,6 +47,12 @@ public class HttpSpec {
 
     @JsonPropertyDescription("The used HTTPS port.")
     private Integer httpsPort = Constants.KEYCLOAK_HTTPS_PORT;
+
+    @JsonPropertyDescription("Annotations to be appended to the Service object")
+    Map<String, String> annotations;
+
+    @JsonPropertyDescription("Labels to be appended to the Service object")
+    Map<String, String> labels;
 
     public String getTlsSecret() {
         return tlsSecret;
@@ -95,4 +104,20 @@ public class HttpSpec {
                 .map(KeycloakSpec::getHttpSpec);
     }
 
+    public Map<String, String> getAnnotations() {
+        return annotations;
+    }
+
+    public void setAnnotations(Map<String, String> annotations) {
+        this.annotations = annotations;
+    }
+
+    public Map<String, String> getLabels() {
+        return labels;
+    }
+
+    public void setLabels(Map<String, String> labels) {
+        this.labels = labels;
+    }
+    
 }
--- a/operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakServicesTest.java
+++ b/operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakServicesTest.java
@@ -31,12 +31,15 @@ import java.time.Duration;
 import java.util.Map;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 
 @QuarkusTest
 public class KeycloakServicesTest extends BaseOperatorTest {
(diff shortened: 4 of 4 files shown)


vmuzikar (commented):
@gilvansfilho Thank you for the contribution!

gilvansfilho:
> Another thought is whether it really makes sense to add labels and annotations specifically to Services. Whether not have it covered in a generic manner for all Operator managed resources (see #15395). @shawkins WDYT?
> 
> @gilvansfilho Can you share the use case you are implementing this for?

The main motivation is to be possible to add `service.alpha.openshift.io/serving-cert-secret-name` annotation allowing Openshift to gerente a cluster certificate for keycloak's pods. Today to use reencrypt routes in openshift we need to mimic the default service created by Operator and add this annotation. Then we need to create a new route with reencrypt termination using the custom service.

With this PR we reduce the steps needs to have `reencrypt` routes with keycloak. Just the route object will be craft by hand.

shawkins:
> If the only motivation for having dedicated labels and annotations setting for services is the reencrypt, I'd rather go down the path of a first class support for reencrypt. 

The exisitng issue for this is https://github.com/keycloak/keycloak/issues/20128

Getting this supported end-to-end is not trivial, and of course initially openshift specific.

There are two cases - reencrypt using a serving certificate, and reencrypt using the http key / cert. We'll also need to have the key / cert for the route - there isn't currently a spot for this on the CR, unless you want to somehow make the http key / cert apply which is also discussed on #34777

These cases would look like:

Re-encrypt implied:

```
http:
  tlsSecret: secret
ingress:
  tlsSecret: other-secret
```

Re-encrypt using serving certificate would need some way to indciate the annotation needs added. That would either be a mode for the ingress, or something on the http spec to indicate a key / cert should be generated:

```
http:
  ...
ingress:
  mode: reencrypt
  tlsSecret: secret
```

gilvansfilho:
At Openshift to use `reencrypt` one is not mandatory to provide a route certificate. A `reencrypt` route without an explicit certificate will use the cluster default  certificate.

vmuzikar:
> A `reencrypt` route without an explicit certificate will use the cluster default certificate.

I'd vote for rather using the cert generated for the service via the annotation.

> unless you want to somehow make the http key / cert apply

That's a good question. Would there be a real life use case where you'd need different cert and key for Keycloak and the Route in re-encrypt case? Maybe when you don't trust Keycloak so much and don't want to give it the TLS private key that will be also used externally? :D

Maybe we could have something like:
```yaml
spec:
  http:
    tlsSecret: keycloak-tls
  ingress:
    tls:
      secret: my-other-secret
      generated: true|false # when set to true, secret must be left blank; currently works only on OpenShift
```

gilvansfilho:
> I'd vote for rather using the cert generated for the service via the annotation.

I too. But keep in mind that cert generated via the annotation is for the pod (service layer), not for the route. For the route openshift allows to provide a custom cert or use cluster default cert.

For `reencrypt` two certs are involved:
- the service / pod cert (spec.http.tlsSecret)
- the route / ingress cert (spec.ingress.tlsSecret new property)

So the `spec.ingress.generated` do not make sense for me (talking about openshift)

shawkins:
> But keep in mind that cert generated via the annotation is for the pod (service layer), not for the route. For the route openshift allows to provide a custom cert or use cluster default cert.

Ok, so we have either an implicit reencrypt where both tlsSecrets are specified, or an explicit reencrypt mode where both tlsSecrets are optional.

> So the spec.ingress.generated do not make sense for me (talking about openshift)

Are you good with an ingress mode or similar flag?

Also I vote to proceed with the additional metadata support proposed in this pr, and tackle the openshift specific reencrypt support under the other issue.

gilvansfilho:
>Are you good with an ingress mode or similar flag?

Sure

---

To provide first class support to Openshift reencrypt routes I think we need below scenarios...

| Scenario | http.tlsSecret | ingress.tlsSecret | ingress.mode | service annotation                                                        | pod certificate      | route certificate             |
|----------|----------------|-------------------|--------------|---------------------------------------------------------------------------|----------------------|-------------------------------|
| 1        | (not defined)  | (not defined)     | reencrypt    | service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret | generated-tls-secret | (cluster default certificate) |
| 2        | pod-tls-secret | (not defined)     | reencrypt    |                                                                           | pod-tls-secret       | (cluster default certificate) |
| 3        | (not defined)  | custom-route-tls  | reencrypt    | service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret | generated-tls-secret | custom-route-tls              |
| 4        | pod-tls-secret | custom-route-tls  | reencrypt    |                                                                           | pod-tls-secret       | custom-route-tls              |


# Scenario 1: Reencrypt with cluster domain and serving certificate for pod

```
spec:
  http:
    ...
  ingress:
    mode: reencrypt
```

Operator will have to:
- Create a `reencrypt` route with no cert (so Openshift will use cluster default certificate)
- Create a service adding the annotation `service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret`
- Under the hoods `spec.http.tlsSecret` must be set to `generated-tls-secret`

With that we have:
- A `reencrypt` openshift route using cluster default certificate (keycloak exposed using default openshift cluster domain). 
- Pod using certificate generated by Openshift (serving certificate)

# Scenario 2: Reencrypt with cluster domain and custom certificate for pod

```
spec:
  http:    
    tlsSecret: pod-tls-secret
  ingress:
    mode: reencrypt
```

User must provide: 
- `pod-tls-secret` to be used by keycloak pods

Operator will have to:
- Create a `reencrypt` route with no cert (so Openshift will use cluster default certificate)
- Create a service in same way is created today

With that we have:
- A `reencrypt` openshift route using cluster default certificate (keycloak exposed using default openshift cluster domain). 
- Pod using certificate provided through `spec.http.tlsSecret`

# Scenario 3: Reencrypt with custom domain certificate and serving certificate for service

```
spec:
  http:
    ...
  ingress:
    mode: reencrypt
    tlsSecret: custom-route-tls
```

User must provide: 
- `custom-route-tls` to be used by router

Operator will have to:
- Create a `reencrypt` route with the custom certificate 
- Create a service adding the annotation `service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret`
- Under the hoods `spec.http.tlsSecret` must be set to `generated-tls-secret`

With that we have:
- A `reencrypt` openshift route using custom certificate (keycloak exposed using domain of the certificate provided through `spec.ingress.tlsSecret` )
- Pod using certificate generated by Openshift (serving certificate)

# Scenario 4: Reencrypt with custom domain certificate and custom certificate for pod

```
spec:
  http:
    tlsSecret: pod-tls-secret
  ingress:
    mode: reencrypt
    tlsSecret: custom-route-tls
```

User must provide: 
- `pod-tls-secret` to be used by keycloak pods
- `custom-route-tls` to be used by router

Operator will have to:
- Create a `reencrypt` route with the custom certificate 
- Create a service in same way is created today

User must provide `pod-tls-secret` to be used by keycloak pods

With that we have:
- A `reencrypt` openshift route using custom certificate (keycloak exposed using domain of the certificate provided through `spec.ingress.tlsSecret` )
- Pod using certificate provided through `spec.http.tlsSecret`


-----

I guess all exposed above will require a lot of work.

> Also I vote to proceed with the additional metadata support proposed in this pr, and tackle the openshift specific reencrypt support under the other issue.

So, I also vote to proceed with this PR, allowing to inject custom labels and annotatios to services, So one who is trying to use `reencrypt` routes before we have first class support for it will have a easier way to it, like bellow:

```
kind: Keycloak
metadata:
  name: example-kc
spec:
  http:
    tls-secret: generated-tls-secret
    annotations:
      service.alpha.openshift.io/serving-cert-secret-name: generated-tls-secret
  ingress:
    enabled: false
```

Then will need to manually create a `reencrypt` route

shawkins:
@gilvansfilho our understanding is the same. Scenario 4 is what I was thinking doesn't need to explicitly specify the mode as reencrypt is implied. Can you move / add the previous comment to #20128?

gilvansfilho:
@shawkins 
> Scenario 4 is what I was thinking doesn't need to explicitly specify the mode as reencrypt is implied

Totally agree

> Can you move / add the previous comment to https://github.com/keycloak/keycloak/issues/20128?

Done

vmuzikar:
@gilvansfilho Thank you for providing the analysis!

> Also I vote to proceed with the additional metadata support proposed in this pr, and tackle the openshift specific reencrypt support under the other issue.

Thinking about it more considering the scenarios, I'm gonna agree with @shawkins here. I don't even think we need OpenShift specific support. We could just expose the TLS secret field in the ingress spec and document some use cases (the scenarios above) that will use the service annotations and/or the TLS secret fields. WDYT?

gilvansfilho:
Agree :)

gilvansfilho:
Team, How will we go ahead? 
To merge this one allowing other user cases we don't have in mind? 
To close this one? Adding explicit reencrypt support through https://github.com/keycloak/keycloak/issues/20128 ?

FYI @vmuzikar @shawkins

shawkins:
> To merge this one allowing other user cases we don't have in mind?

Yes, I think we are good to merge this one and let https://github.com/keycloak/keycloak/issues/20128 and https://github.com/keycloak/keycloak/issues/34777 be used for any additional work.

shawkins (approved):
LGTM, thank you @gilvansfilho

vmuzikar (approved):
LGTM, nice work @gilvansfilho.

Let's address the reencrypt scenario by adding TLS secret to Ingress in #20128.

PR #39930, opened 2025-05-23
Implemented OIDC Prefixed User Attribute Mapper

pszabo891:
The new OIDC mapper adds the possibility to prefix the values of the claims when imported into user attributes.

image

Closes #39929

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/mappers/PrefixedUserAttributeMapper.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/mappers/PrefixedUserAttributeMapper.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.broker.oidc.mappers;
+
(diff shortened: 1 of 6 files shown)


pedroigor (commented):
@pszabo891 The new mapper makes sense for me, but don't we want to extend the `UserAttributeMapper` and add the prefix logic on top of it?

pszabo891:
@pedroigor I have implemented the mapper originally as a plugin, so creating a new mapper was the most straightforward way to bring the functionality into the codebase without changing existing code.

I'm happy to rework it and implement as part of the `UserAttributeMapper` if there is interest in merging it. I will raise a new PR and tag you along.

pedroigor:
I think we can have it as a separate mapper, but with the prefixing logic applied to values. The admin would usually choose between your mapper or the regular user attribute mapper, but never both, right? 

Another concern I have is how to solve conflicts with the root attributes, such as email, first, and last names. Now they will be prefixed and, considering they are single-valued, the last update will win?

pszabo891:
@pedroigor If we keep it as a separate mapper, you are correct, for each claim only one mapper will be needed - either the `UserAttributeMapper` or the `PrefixedUserAttributeMapper` - it's up to the Administrator to decide.

Those root attributes are only impacted by the prefixing mapper if the Administrator explicitly configures the mapper for them and then it behaves as expected, although there might not be a real use case for prefixing those attributes. Or perhaps I missed something?

If prefixing the root attributes is a concern, we can also implement it so, that we skip them and apply the prefix only to the rest of the attributes.

pedroigor:
From a UX perspective, keeping a single mapper is better so that we can add the `prefix` configuration property to the `UserAttributeMapper` and handle the values accordingly if a `prefix` is set. Otherwise, I'm afraid that people will be confused about when to use one or another, or even have both in their broker configuration.

Regarding the root attributes, I don't think it makes much sense to use a prefix on them. I cannot think about a use case  mainly because the values would be shown everywhere. Imagine the account console showing as your first name `<prefix>Joe`. But as you said, it is a matter of configuration, and if the admin sets a prefix for any of the root attributes, they will be prefixed too.

pszabo891:
@pedroigor I have created a new PR #42002 which adds the prefixing functionality in the existing mapper.

pedroigor:
Thanks. Closing in favor of https://github.com/keycloak/keycloak/pull/42002.

PR #39935, opened 2025-05-24
Update login-update-password.ftl

nexdatum:
**Minor change.**
An unnecessary slash character was found in line 68 (<button/></button> corrected to <button></button>).


mposolda:
@nexdatum Thanks for the PR, but I think this is done on purpose. HTML elements like `input` and `button` are supposed to have `/` at the end to signal that there is no other related "ending" tag for them (EG. `<input />` VS `<input></input` ).

Also this is used in same way in much more files, so if we want to change this, we should do it consistently in all the files and not just on single place.

Finally, there is no GH issue.

If you think this is "real" issue, then please create GH issue and describe the reasons for this and then send the other PR with the update in all the files to make this change consistent (not just in single file). See "contributors guide" for more details. But TBH I don't think that this change is valid and necessary.

PR #39936, opened 2025-05-24
Update acr-to-loa-mapping.adoc

moathshareef:


mposolda:
@moathshareef Thanks for the PR, but I don't think it is correct as the name of the claim in the token is `acr` . Also there is not GH issue linked to this (Please see the contributions guide for the details and re-open the PR if you are still interested in it).

Issue #33436, opened 2024-09-16
Need more l18n dialogs


ssilvert:
### Description

We now have localization dialogs to help admins do localization for User Profile attributes and User Profile attribute groups.

However, there are several other parts of the admin console where this dialog could be helpful.

This includes, but is probably not limited to:

- Client Details --> Client Name
- Client Details --> Description
- User Profile --> Attribute --> Annotations --> Helper text (above and under)
- User Profile --> Attribute --> Annotations --> Input placeholder

Not sure if the following are currently localized?  If so, they could benefit from the dialog as well
- Organization Name
- Organization Alias
- Organization Description. 
- Realm Name
- Realm Display Name
- Realm HTML Display Name
- Realm --> Email --> From display name
- Realm --> Email --> Reply to display name
- Identity Provider Display Name
- User Federation Display Name

ssilvert:
For client details, Erik implemented this but it looks like it will turn out to be problematic and bug-prone.  So we won't do it for that.

PR #39955, opened 2025-05-26
added translation component to annotations

edewit:
related: #33436


ssilvert (changes requested):
@edewit I tried this out and it doesn't work.  

When you try to type a translation into the dialog, the dialog will go away.

It will let me type in a translation if I save the attribute first and then try to add the inputHelperTextBefore annotation.  But in that case, the save button never activates and I can't save the translations.

PR #39957, opened 2025-05-26
Removed dead local stores

### LINKED ISSUE #39698: Dead local stores
### Description

Greetings! I detected some occurences of dead local store using a static analyzer. 

### Discussion

_No response_

### Motivation

Dead stores refer to assignments made to local variables that are subsequently never used or immediately overwritten. Such assignments are unnecessary and don’t contribute to the functionality or clarity of the code. They may even negatively impact performance. Removing them enhances code cleanliness and readability. Even if the unnecessary operations do not do any harm in terms of the program’s correctness, they are - at best - a waste of computing resources.

### Details

1. The value `xmlEvent = StaxParserUtil.getNextEndElement(xmlEventReader)` is assigned but never used at the following places: 

https://github.com/keycloak/keycloak/blob/ebe3e15a9e0f99a235b4b85829b898b39a8b74bc/saml-core/src/main/java/org/keycloak/saml/processing/core/parsers/util/SAML11ParserUtil.java#L604-L613

https://github.com/keycloak/keycloak/blob/ebe3e15a9e0f99a235b4b85829b898b39a8b74bc/saml-core/src/main/java/org/keycloak/saml/processing/core/parsers/util/SAML11ParserUtil.java#L543-L552

2. The same applies to the variable `ignored` here:

https://github.com/keycloak/keycloak/blob/ebe3e15a9e0f99a235b4b85829b898b39a8b74bc/server-spi-private/src/main/java/org/keycloak/validate/validators/UriValidator.java#L165-L167

To resolve the issue in these cases we should remove unnecessary assignments.

---
3. The `sig` array is initialized to an empty array:

https://github.com/keycloak/keycloak/blob/ebe3e15a9e0f99a235b4b85829b898b39a8b74bc/saml-core/src/main/java/org/keycloak/saml/BaseSAML2BindingBuilder.java#L429-L436

However, it is important to note that this array is subsequently overwritten in the `try` block when the signature is generated. It might be better to consider initializing the `sig` to `null`.

---

I can try to submit a PR if you're interested.

Anchels:
Closes #39698

### DIFF
--- a/saml-core/src/main/java/org/keycloak/saml/BaseSAML2BindingBuilder.java
+++ b/saml-core/src/main/java/org/keycloak/saml/BaseSAML2BindingBuilder.java
@@ -426,7 +426,7 @@ public class BaseSAML2BindingBuilder<T extends BaseSAML2BindingBuilder> {
                 rawQuery = rawQuery.substring(pos + 1);
             }
             Signature signature = signatureAlgorithm.createSignature();
-            byte[] sig = new byte[0];
+            byte[] sig = null;
             try {
                 signature.initSign(signingKeyPair.getPrivate());
                 signature.update(rawQuery.getBytes(GeneralConstants.SAML_CHARSET));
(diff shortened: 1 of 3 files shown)


rmartinc (approved):
Thanks @Anchels! yep, better to have the code clean. So thanks for this.

@mposolda Please re-run the failed CI job because it seems to be not related.

pskopek (approved):
LGTM

mposolda (approved):
@rmartinc @pskopek Thanks!

PR #39977, opened 2025-05-26
Resolve home organization when requesting the scope for all organizations

### LINKED ISSUE #39864: IdP redirect fails when user belongs to multiple organizations with organization:* scope
### Before reporting an issue

### Area

organizations

### Describe the bug

In our web application, when requesting the organization:* scope, the authentication flow fails to redirect users to their Identity Provider (IdP) if they belong to multiple organizations.

### Version

26.2.4

### Regression

### Expected behavior

If a user belongs to a single organization with an associated IdP, the authentication flow correctly redirects to the IdP.

If a user belongs to multiple organizations, the authentication flow should still redirect to the appropriate IdP.

### Actual behavior

When a user belongs to multiple organizations and the organization:* scope is requested, the authentication flow does not redirect to the IdP and instead proceeds with the username/password form.

If the organization:* scope is not requested, the user is correctly redirected to the IdP, regardless of whether they belong to one or multiple organizations.

### How to Reproduce?

1. Configure a user to belong to multiple organizations, each with an associated IdP.
2. Initiate authentication in the web application with the organization:* scope.
3. Observe that the flow does not redirect to the IdP and instead presents the username/password form.
4. Repeat the process without the organization:* scope and confirm the correct IdP redirect.

### Anything else?

_No response_

pedroigor:
Closes #39864

### DIFF
--- a/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
+++ b/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
@@ -215,19 +215,18 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
     }
 
     private boolean shouldUserSelectOrganization(AuthenticationFlowContext context, UserModel user) {
-        OrganizationProvider provider = getOrganizationProvider();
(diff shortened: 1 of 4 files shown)


vramik (approved):

ahus1 (approved):
Approving based on @vramik's review

PR #39989, opened 2025-05-27
Move V1 specific methods from AbstractTokenExchangeProvider

### LINKED ISSUE #39988: Move V1 specific methods from AbstractTokenExchangeProvider
### Description

Current `AbstractTokenExchangeProvider` contains lots of the methods, which are specific to token-exchange:V1 implementation and are overriden in the `StandardTokenExchangeProvider`. It is very likely that when we introduce more V2 implementations of token-exchange (external-internal, internal-external etc), those will be more aligned with internal-internal V2 implementation rather than with V1 implementation. For example:

- The FGAP:v1 likely will not be used for other V2 token exchange providers
- The refresh-token will not be the default `requested_token_type` in V2 providers like it is in V1 provider
- etc

Having the V1 methods on the `AbstractTokenExchangeProvider` is maybe not ideal as there is a chance that those will be accidentally called from V2 providers (which may not be safe) . So those would likely need to be overriden. So perhaps makes sense to rather add V1 specific methods to the `V1TokenExchangeProvider` directly?

mposolda:
closes #39988


douglaspalmer (approved):
LGTM

graziang (approved):
@mposolda thanks!

mposolda:
@douglaspalmer @graziang Thanks!

PR #40005, opened 2025-05-27
fix: adding a -- separator for spi options

### LINKED ISSUE #39063: Optimized startup fails from `kc.spi-connections-http-client-default-expect-continue-enabled` passed at runtime
### Before reporting an issue

### Area

core

### Describe the bug

Upgrading from Keycloak 24 to 26.0.10 (Red Hat build), I notice that it no longer starts successfully with optimized startup, due to passing the `expect-continue-enabled` option.

### Version

26.0.10

### Regression

### Expected behavior

Keycloak starts up successfully, and the `DefaultHttpClientFactory` determines the desired value for `expect-continue-enabled` at runtime.

### Actual behavior

Keycloak starts up with an error:

```
The following build time options have values that differ from what is persisted - the new values will NOT be used until another build is run: kc.spi-connections-http-client-default-expect-continue-enabled
```

### How to Reproduce?

1. Create an optimized Keycloak image
2. start the server in production mode, passing both `--optimized` and `--spi-connections-http-client-default-expect-continue-enabled=true`

### Anything else?

I'm not familiar with how Keycloak determines build-time options to persist, but it seems strange to me that this is the only connections SPI option that is flagged, since there are others like `socket-timeout-millis` that I pass. I did notice that the option is absent in the factory's `getConfigMetadata()` method:

https://github.com/keycloak/keycloak/blob/4ae7d60784a46fe1a771acf90923f81452bf4c5e/services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java#L261-L344

shawkins:
closes: #39063


vmuzikar (commented):
> are you ok with this being a breaking change from the perspective of users who are relying on auto-builds with -provider, -provider-default, -enabled? If so I can add a note for that.

Yes, please add a note to the upgrading guide about that. Maybe with some reasoning about it.

> And / or do you want a warning for now for the possibly ambiguous cases?

You mean to the docs? Definitely yes. I know you've done some updates to the docs but IMHO we need to be much more explicit. Basically to explain the difference between the two forms and the auto build behaviour. It might very easily get confusing to the users.

Additionally, I think we should update all docs occurences to the `--` format.

shawkins:
Updated with a warning when changed the docs to use the new format.

There's still a reference to spi-authentication-sessions-map-auth-sessions-limit (now spi-authentication-sessions--map--auth-sessions-limit) that should be removed correct?

shawkins:
Also updated to internally map to the new spi form - this also highlighted that the PropertyMapper logic needs to maintain an additional mapping to the legacy form.

core/src/main/java/org/keycloak/Config.java
```diff
@@ -235,6 +235,10 @@ public static interface Scope {
 
         Scope scope(String... scope);
 
+        /**
+         * @deprecated to be removed
```
mabartos: Shouldn't we explain more why it's been deprecated? I'd say yes.

core/src/main/java/org/keycloak/Config.java
```diff
@@ -235,6 +235,10 @@ public static interface Scope {
...
+         */
+        @Deprecated
```
mabartos, line 244:
```suggestion
        @Deprecated(forRemoval = true, since = "26.3.0")
```

We might start providing more info when we deprecate things for better traceability and management of deprecations. WDYT?

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -80,6 +84,15 @@ the user account will be updated to (un)mark the email as verified.
 In the future, we might evaluate changing this specific configuration to avoid automatic updates on the email verification
 status on federated user accounts depending on the use cases and the demand from the community.
 
+=== Buildtime SPI Options
+
+SPI options ending in -enabled, -provider-default, or -provider were previously treated as buildtime. However in some instances this was not correct as a provider could have a configuration property ending in one of those suffixes as well.
```
mabartos:
```suggestion
SPI options ending in `-enabled`, `-provider-default`, or `-provider` were previously treated as buildtime. However in some instances this was not correct as a provider could have a configuration property ending in one of those suffixes as well.
```

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -80,6 +84,15 @@ the user account will be updated to (un)mark the email as verified.
...
+
+To resolve this ambiguity, and any potential ambiguity involving SPI and provider names, a new SPI option format was introduced where the scopes and suffix are separated by -- instead of -.
```
mabartos:
```suggestion
To resolve this ambiguity, and any potential ambiguity involving SPI and provider names, a new SPI option format was introduced where the scopes and suffix are separated by `--`(double dash) instead of `-`(dash). 
```

Maybe be more explicit here with the description?

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -80,6 +84,15 @@ the user account will be updated to (un)mark the email as verified.
 In the future, we might evaluate changing this specific configuration to avoid automatic updates on the email verification
 status on federated user accounts depending on the use cases and the demand from the community.
 
+=== Buildtime SPI Options
+
+SPI options ending in -enabled, -provider-default, or -provider were previously treated as buildtime. However in some instances this was not correct as a provider could have a configuration property ending in one of those suffixes as well.
+
+To resolve this ambiguity, and any potential ambiguity involving SPI and provider names, a new SPI option format was introduced where the scopes and suffix are separated by -- instead of -. 
+
+An SPI property ending in -enabled, -provider-default, or -provider will only be treated as buildtime if it is using the new format, e.g. `spi-<spi-name>--<provider-name>--enabled` will be recognized as a buildtime option.
```
mabartos:
```suggestion
An SPI property ending in `-enabled`, `-provider-default`, or `-provider` will only be treated as buildtime if it is using the new format, e.g. `spi-<spi-name>--<provider-name>--enabled` will be recognized as a buildtime option.

For instance, the correct way to reference your custom email template is: `--spi-email-template--mycustomprovider--enabled` (not `--spi-email-template-mycustomprovider-enabled`).
```

I think it'd be good to have some example here. WDYT?

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -80,6 +84,15 @@ the user account will be updated to (un)mark the email as verified.
...
 status on federated user accounts depending on the use cases and the demand from the community.
 
+=== Buildtime SPI Options
+
+SPI options ending in -enabled, -provider-default, or -provider were previously treated as buildtime. However in some instances this was not correct as a provider could have a configuration property ending in one of those suffixes as well.
+
+To resolve this ambiguity, and any potential ambiguity involving SPI and provider names, a new SPI option format was introduced where the scopes and suffix are separated by -- instead of -. 
+
+An SPI property ending in -enabled, -provider-default, or -provider will only be treated as buildtime if it is using the new format, e.g. `spi-<spi-name>--<provider-name>--enabled` will be recognized as a buildtime option. 
+Options using the legacy format and ending in ending in -enabled, -provider-default, or -provider will instead be treated as runtime and a warning will be emitted.
```
mabartos:
```suggestion
Options using the legacy format and ending in `-enabled`, `-provider-default`, or `-provider` will instead be treated as runtime and a warning will be emitted.   
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
```diff
@@ -399,7 +403,15 @@ private static void sanitizeMappers(Map<OptionCategory, List<PropertyMapper<?>>>
         private static void handleMapper(PropertyMapper<?> mapper, BiConsumer<String, PropertyMapper<?>> operation) {
             operation.accept(mapper.getFrom(), mapper);
             if (!mapper.getFrom().equals(mapper.getTo())) {
-                operation.accept(mapper.getTo(), mapper);
+                String to = mapper.getTo();
+                operation.accept(to, mapper);
+                if (to.startsWith(KC_SPI_PREFIX)) {
+                    if (!mapper.getTo().contains("--")) {
+                        throw new IllegalStateException("Mapper should use the new form of the spi option: " + to);
```
mabartos:
```suggestion
                        throw new IllegalStateException("Mapper should use the new form of the SPI option with the `--` separator: " + to);
```

mabartos (commented):
@shawkins Nice work! Using `--` separator makes these SPI separations more visible and I like it more than before. 

Just added some small comments that might improve things a little bit more IMO.

core/src/main/java/org/keycloak/Config.java
shawkins: Updated, and also we can't use the annotation properties in core because we're forcing Java 8 compatibility.

shawkins:
> @shawkins Nice work! Using `--` separator makes these SPI separations more visible and I like it more than before.
> 
> Just added some small comments that might improve things a little bit more IMO.

Thank you for the review. Applied the suggestions.

core/src/main/java/org/keycloak/Config.java
mabartos: Ahh, I see. That's a shame we still need to have it there.

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -80,6 +84,17 @@ the user account will be updated to (un)mark the email as verified.
...
+=== Buildtime SPI Options
+
+SPI options ending in `-enabled`, `-provider-default`, or `-provider` were previously treated as buildtime. However in some instances this was not correct as a provider could have a configuration property ending in one of those suffixes as well.
+
+To resolve this ambiguity, and any potential ambiguity involving SPI and provider names, a new SPI option format was introduced where the scopes and suffix are separated by `--`(double dash) instead of `-`(dash). 
+
+An SPI property ending in `-enabled`, `-provider-default`, or `-provider` will only be treated as buildtime if it is using the new format, e.g. `spi-<spi-name>--<provider-name>--enabled` will be recognized as a buildtime option.
+
+For instance, the correct way to reference your custom email template is: `--spi-email-template--mycustomprovider--enabled` (not `--spi-email-template-mycustomprovider-enabled`).
+Options using the legacy format and ending in `-enabled`, `-provider-default`, or `-provider` will instead be treated as runtime and a warning will be emitted.
```
mabartos:
```suggestion
For instance, the correct way to reference your custom email template is: `--spi-email-template--mycustomprovider--enabled` (not `--spi-email-template-mycustomprovider-enabled`).

Options using the legacy format and ending in `-enabled`, `-provider-default`, or `-provider` will instead be treated as runtime and a warning will be emitted.   
```

nit: Would be good to separate it if you do some other changes.

mabartos (commented):
@shawkins Looks good, but one last thing - we should move these upgrading notes you provided to the "Notable changes" section and probably not to the "Breaking changes". Or is there anything I missed?

core/src/main/java/org/keycloak/Config.java
```diff
@@ -235,6 +235,13 @@ public static interface Scope {
 
         Scope scope(String... scope);
 
+        /**
+         * @deprecated since 26.2.3, to be removed
```
mabartos:
```suggestion
         * @deprecated since 26.2.6, to be removed
```

nit: I believe it'll fit into the 26.2.6

shawkins:
>  Or is there anything I missed?

It can be viewed as a breaking change unfortunately.

core/src/main/java/org/keycloak/Config.java
shawkins: Corrected - should have just been 26.3

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
shawkins: added the spacing

mabartos (dismissed):
@shawkins Ok, LGTM, thanks!

shawkins:
@mabartos had to correct the generated options output as well, along with a few more places where -- was missing or being rendered incorrectly.

mabartos (dismissed):
LGTM

vmuzikar:
I plan to give this a quick look later today.

vmuzikar (commented):
@shawkins Sorry for the late review and thank you for the changes!

Overall LGTM but might be good to add some tests for checking the selected SPI options are correctly treated as buildtime.
Additionally, I wonder if we should have a test for the env var form, if `__` works as expected.

vmuzikar:
Can you also please rebase?

shawkins:
> Overall LGTM but might be good to add some tests for checking the selected SPI options are correctly treated as buildtime.

We already had a test or two along those lines (see StartCommandDistTest.errorSpiBuildtimeAtRuntime for example), and the arquilian logic relies on implicit rebuilds due to provider option changes.

> Additionally, I wonder if we should have a test for the env var form, if __ works as expected.

I can certainly add something. The replacement logic in KcEnvConfigSource is now quite straight-forward all _ in a non-wildcard key will become -, so there's nothing special about this case.

vmuzikar:
> We already had a test or two along those lines (see StartCommandDistTest.errorSpiBuildtimeAtRuntime for example), and the arquilian logic relies on implicit rebuilds due to provider option changes.

But we don't cover all the SPI build time options (i.e. `--provider`, `--enabled` and `--provider-default`), do we?

> I can certainly add something. The replacement logic in KcEnvConfigSource is now quite straight-forward all _ in a non-wildcard key will become -, so there's nothing special about this case.

Yeah, it'd be good to have something to make sure that e.g. Quarkus doesn't messes that up somewhere.

shawkins:
> But we don't cover all the SPI build time options (i.e. --provider, --enabled and --provider-default), do we?

I'll add something after lunch.

> Yeah, it'd be good to have something to make sure that e.g. Quarkus doesn't messes that up somewhere.

Quarkus? These entries are only handled by Keycloak logic.

vmuzikar:
> Quarkus? These entries are only handled by Keycloak logic.

Yeah, but I meant some too smart interceptors maybe. Still I'd feel safer with some tests. :)

shawkins:
> Yeah, but I meant some too smart interceptors maybe. Still I'd feel safer with some tests. :)

See if the the additional changes are sufficient.

vmuzikar (dismissed):
LGTM, thank you.

shawkins:
@mabartos @vmuzikar the rebase was trivial, just wanted to check if either of you want to review again.

mabartos (approved):
LGTM

vmuzikar (approved):

pl-me:
@shawkins Hi, I noticed you have changed MicroProfileScope.getPropertyNames() to filter keys using double-dash-separator only. That means it is a breaking change if you rely on getPropertyNames() and still using single-dash options. Am I right?

As getPropertyNames() is now deprecated I guess I should move away from using that function anyway. I just wanted to make sure your change was intentional.

shawkins:
> As getPropertyNames() is now deprecated I guess I should move away from using that function anyway. I just wanted to make sure your change was intentional.

The change was intentional, as was marking it as deprecated. Yes you should not rely on this function.

PR #40007, opened 2025-05-27
Generate a UUID to be the JTI instead of reusing the nonce.

MichaelCordingley:
Fixes https://github.com/keycloak/keycloak/issues/40160


MichaelCordingley:
Done

mposolda (approved):
Approving. I think PR is correct behavior as long as tests are OK

PR #40011, opened 2025-05-28
#12682 [CORS] Allow Access-Control-Allow-Headers customization

### LINKED ISSUE #12682: Allow CORS Access-Control-Allow-Headers customization
### Description

Allow customization for CORS Access-Control-Allow-Headers

### Discussion

https://keycloak.discourse.group/t/customizing-access-control-allow-headers/7672

### Motivation

Keycloak has to support custom headers.
The lacking of this functionality makes keycloak incompatible, for example, with distributed tracing systems (opentelemetry, elastic-agent).

These are some of the headers that are commonly sent : 

traceparent
uber-trace-id
x-amzn-trace-id
x-b3-sampled
x-b3-spanid
x-b3-traceid
elastic-apm-traceparent

### Details

Access-Control-Allow-Headers are stubbed in

https://github.com/keycloak/keycloak/blob/36da2d20e97a10821c7305881d3ffa343b5ad001/services/src/main/java/org/keycloak/services/resources/Cors.java

Please allow customization for Access-Control-Allow-Headers via GUI (in clients configurations) or via environment variables.

dteleguin:
This PR adds the ability to programmatically customize the `Access-Control-Allow-Headers` CORS header returned by Keycloak.

Highlights:
* `HeaderSet` is a `TreeSet<String>` wrapper featuring fast clone operation (`HeaderSet.copyOf()`) and cached string representation. It is currently used for `DefaultCors#allowedHeaders` and `Cors#DEFAULT_ALLOW_HEADERS` only, but could be potentially used for other fields in the future;
* the `DefaultCors` constructor is made public so that the extensions could reuse it.

Closes: #12682

### DIFF
--- a/services/src/main/java/org/keycloak/services/cors/DefaultCors.java
+++ b/services/src/main/java/org/keycloak/services/cors/DefaultCors.java
@@ -47,12 +47,13 @@ public class DefaultCors implements Cors {
     private ResponseBuilder builder;
     private Set<String> allowedOrigins;
     private Set<String> allowedMethods;
+    private final HeaderSet allowedHeaders = HeaderSet.copyOf(DEFAULT_ALLOW_HEADERS);
     private Set<String> exposedHeaders;
 
     private boolean preflight;
     private boolean auth;
 
-    DefaultCors(KeycloakSession session) {
+    public DefaultCors(KeycloakSession session) {
         this.session = session;
         this.request = session.getContext().getHttpRequest();
         this.response = session.getContext().getHttpResponse();
@@ -113,6 +114,12 @@ public class DefaultCors implements Cors {
     }
 
     @Override
+    public Cors allowedHeaders(String... allowedHeaders) {
+        this.allowedHeaders.addAll(Arrays.asList(allowedHeaders));
+        return this;
+    }
+
+    @Override
     public Cors exposedHeaders(String... exposedHeaders) {
         if (this.exposedHeaders == null) {
             this.exposedHeaders = new HashSet<>();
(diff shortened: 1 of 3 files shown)


JonasDev17:
Any update on this?

services/src/main/java/org/keycloak/services/cors/DefaultCors.java
```diff
@@ -165,13 +172,9 @@ public void add() {
 
         if (preflight) {
             if (auth) {
-                response.setHeader(ACCESS_CONTROL_ALLOW_HEADERS, String.format("%s, %s", DEFAULT_ALLOW_HEADERS, AUTHORIZATION_HEADER));
-            } else {
-                response.setHeader(ACCESS_CONTROL_ALLOW_HEADERS, DEFAULT_ALLOW_HEADERS);
+                allowedHeaders(AUTHORIZATION_HEADER);
```
mposolda, line 175: The `DefaultCors` is per-request object. So adding `authorization` header this way means that headers needs to be re-computed from the `HeaderSet` in every request.

Could we instead do something that:
- Extensions have a way to configure headers at the KeycloakSessionFactory level (not at the `KeycloakSession` level, which is effectively single HTTP request)
- The server could still cache the allowed headers as a string (not as a set)
- When new header is supposed to be added by some extension, the String could be converted to Set and then re-computed the whole String. However since this would be done only at factory level (typically at server startup), we do not need to have huge performance optimizations for this. This means we can also get rid of introducing custom set implementation `HeaderSet` and we will be able to live just with the default set impls provided by JDK.
- When adding `authorization` header, it could be still added as a string to the "default allow headers" as it is today in Keycloak main. No need of additional overhead of parsing the string set and then re-compute the string at every request.

WDYT?

mposolda (commented):
@dteleguin Thanks for the PR! Added some review comment.

stianst:
This only allows headers to be customized by extensions, and not by server config as what the issue was originally about. I also think it's quite a complex approach for it.

I've sent a separate PR for the server config option here https://github.com/keycloak/keycloak/pull/43767

mposolda:
Closing as the related issue is already resolved now and there was no PR to my last comment anyway.
@dteleguin Feel free to create dedicate issue/discussion if you see a need to add the custom headers by extensions, but hopefully the server-config is sufficient for most of the use-cases.

PR #40017, opened 2025-05-28
Revise Client Policies Codes - SecureRedirectUrisEnforcerExecutorTest

tnorimat:
closes #40016


tnorimat:
@mposolda Hello, if possible, could you check the PR?

mposolda:
@tnorimat TBH it seems to me that having `SecureRedirectUrisEnforcerExecutorTest` as separate test is better than having everything to the `ClientPoliciesExecutorTest` ? The `ClientPoliciesExecutorTest` is already quite big. Also the `SecureRedirectUrisEnforcerExecutor`  is relatively complex executor with bigger amount of various config options and it can be good to have ability to easily able to run the tests related to this executor in case of some changes needed in that executor.

Or any specific reason for moving this to single class?

tnorimat:
@mposolda Thank you for the review. I agree  with you. I keep `SecureRedirectUrisEnforcerExecutorTest` individual class. I will only revise `SecureRedirectUrisEnforcerExecutorTest`.

tnorimat:
@mposolda Hello, if possible, could you re-check the PR?

mposolda (approved):
@tnorimat Thanks!

PR #40040, opened 2025-05-28
Use more concise PKCE method label in OIDC Client config

### LINKED ISSUE #40039: Abbreviate text in PKCE method configuration label in OIDC Client configuration
### Description

Use the brief `PKCE Method` text as a label and use the full wording in the description.

Suggested description: "Choose which Proof Key for Code Exchange (PKCE) method is used. If not specified, Keycloak does not apply PKCE to a client unless the client sends an authorization request with an appropriate code challenge and code exchange method."

### Discussion

_No response_

### Motivation

Image

Image

### Details

_No response_

thomasdarimont:
Fixes #40039


mposolda (approved):
@thomasdarimont +1 that new label and tooltip is better. Thanks for this idea

Issue #39506, opened 2025-05-06
Expand health check support


shawkins:
### Description

There should be a supported way to perform a health check on the main interface, and there always be a way to support HTTP based health checks - see https://github.com/keycloak/keycloak/issues/38126 and https://github.com/keycloak/keycloak/discussions/36537

### Discussion

https://github.com/keycloak/keycloak/discussions/39376 https://github.com/keycloak/keycloak/discussions/37287

### Motivation

To allow for health checks under a broader set of scenarios than having encrypted management interface access allows for. 

Having the ability to support an HTTP healthcheck means that the check can be performed in BASH, such that a custom image is note needed.

### Details

https://github.com/keycloak/keycloak/issues/40945 allowed for the management interface to run on http, rather than https. 

https://github.com/keycloak/keycloak/issues/41303 allows for health checks on the main interface for users in environments where the port of the health check must be the same as the main application.

Using the first mechanism or the second mechanism with http-enabled true will allow for an HTTP health check in most cases. https://github.com/keycloak/keycloak/issues/38126 documents how to utilize that.

https://github.com/quarkusio/quarkus/issues/47531 would be nice to have for those wishing to better secure http access.

shawkins:
All subissues have been addressed as of KC 26.4.

PR #40044, opened 2025-05-28
fix: adding health check on the main interface

### LINKED ISSUE #41303: Allow for health check on main interface
### Description

We have two options:

### Discussion

_No response_

### Details

_No response_

shawkins:
@mabartos @vmuzikar this implements two of the stategies for providing an http only health check as mentioned on #39506

If a user has enabled https and health checks they can either:

- use the http-management-scheme option to keep the management interface as http. If they want to enforce the access as local, then they would set the management host to localhost. If they want external access, then an edge proxy should be utilized in front.

- also enable http and use the health/ready endpoint on our main interface. This is similar to enabling the legacy management interface, but we can be selective about what routes we add and restirct access to only localhost (I mentioned restricting access to http on the issue, but localhost should be better). https://github.com/quarkusio/quarkus/issues/47531 could be seen as a prerequisite for this if we want to limit the exposure of the http port.

I'm not making any accomodation for the case where a user has not enabled https and not enabled health checks - it is up to the user what url, such as a welcome page redirect, or welcome page resource, to serve as the health check.

Along with this change we would document or even add an http health check script.

WDYT?

closes: #41303

Also, not directly related to this change, but I don't think we can validate the https options on the management interface like we're currently doing. The propertymapping interceptor is turned off during validation, so we aren't getting the parent values to correclty determine if https is in use. Resolving to the parent is also cumbersome using context proceed. 

I removed that validation logic from the ManagementPropertyMappers in what was committed, but we could also do the config validation differently - that is turn the PropertyMappingInterceptor back on for the PropertyMapper.validate call. EDIT: moved this to #40103


quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -254,6 +255,21 @@ void configureRedirectForRootPath(BuildProducer<RouteBuildItem> routes,
                 );
     }
 
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    @Consume(ConfigBuildItem.class)
+    void configureLocalHealthCheck(BuildProducer<RouteBuildItem> routes,
+                                      HttpRootPathBuildItem httpRootPathBuildItem,
+                                      KeycloakRecorder recorder) {
+        if (isHealthEnabled() && !Configuration.isTrue(ManagementOptions.LEGACY_OBSERVABILITY_INTERFACE)) {
```
shawkins: This could also be made contingent on whether the management interface is http, but that of course is a different port.

mabartos: @shawkins IMO, if we'd like to expose health endpoint on the main server as well, we'd need to have a separate config option to allow it if users want to have such a behavior. If the health is enabled, we should have it exposed only on the management interface and if users want to have otherwise, they need to explicitly enable it. 

Or what do you think?

shawkins: > IMO, if we'd like to expose health endpoint on the main server as well, we'd need to have a separate config option to allow it if users want to have such a behavior

My thinking is that this not need to be opt-in because access is effectively restricted to only local clients - we would document this as a side-effect of health-enabled=true. Or do you have concerns about potential abuses of the logic in KeycloakLocalReadinessHandler? The only concern I can think of is having a potential clash with user contributed end-points. 

In any case if we can avoid more configuration options, I would like that.

The logic in this pr is specifically for users who:
- still need https access for the management interface
- can set http-enabled=true 
  - If they otherwise would want http-enabled=false (probably because there's no proxy involved), then the quarkus enhancement about specifying the http host separately from https host would also be needed. Or we'd have to add a global filter preventing http access to anything other than the health endpoint(s). 

If this seems too nuanced, we can just leave this scenario unaddressed.

EDIT: forgot about #37287 - users are indicating that they still need the legacy observability behavior. That is the health check needs to be on the same port as the main application.

shawkins:
Updated this pr to point to #41303 instead.

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
mabartos: @shawkins Thanks for the additional input. After reading these resources and discussions, I agree to address at least the scenario for allowing HTTP health checks from local served on the main HTTP interface. 

> In any case if we can avoid more configuration options, I would like that.

+1

Added some comment to the parent issue.

PR #40065, opened 2025-05-29
Use previous password in SMTP when the the authType defaults to basic

### LINKED ISSUE #39781: SMTP password overwritten with asterisks
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When using SMTP password authentication, saving realm settings in the UI without re-entering the password updates the password in the database to be all asterisks. 

### Version

26.2.4

### Regression

### Expected behavior

The SMTP password should be unchanged.

### Actual behavior

The password is saved as all asterisks. 

### How to Reproduce?

Configure a realm to use SMTP password authentication. Reopen the realm settings UI and press save. 

### Anything else?

_No response_

rmartinc:
Closes #39781

The password was not recovered from the previous configured one when the `authType` was not sent (setting it to asterisks). I have also cleaned all the authentication data when the `auth` is not enabled (I prefer to not maintain that data if not used, and this was already done for the `authType` so doing the same now). The little test in `RealmTest` was modified to check the password is correct after the update with no `authType`.

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
@@ -891,8 +891,16 @@ public class DefaultExportImportManager implements ExportImportManager {
 
             Map<String, String> config = new HashMap<>(rep.getSmtpServer());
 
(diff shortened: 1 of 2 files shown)


mposolda (approved):
@rmartinc Thanks!

isaacjensen:
Will there be a minor version released with this fix in? What is the cadence for minor versions?

PR #40078, opened 2025-05-29
Add warning about using exec in custom entrypoint scripts

somln:
Closes #39817

### DIFF
--- a/docs/guides/server/containers.adoc
+++ b/docs/guides/server/containers.adoc
@@ -152,9 +152,25 @@ Health check endpoints are available at `https://localhost:9000/health`, `https:
 
 Opening up `https://localhost:9000/metrics` leads to a page containing operational metrics that could be used by your monitoring solution.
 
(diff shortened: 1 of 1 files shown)


ahus1:
@somln - thank you for the docs PR. I updated it a little, as I didn't want the whole section to be a warning. 
I also updated the level of the section to be on the same as the other sections below "creating a customized and optimized container image". I also moved it before "Building the container image" I would assume people would update the entrypoint before building. 

Please let me know if you agree with these changes, or if I introduce new problems or parts that are difficult to understand. 

@andymunro - can you please review the changes if they are aligned with the other docs and the style guide? Thanks!

andymunro (approved):

somln:
@ahus1  Looks great to me — thanks for refining the structure and tone! I agree with the placement and heading level changes.

ahus1 (approved):
Thank you for this PR @somln and for picking up this help-wanted task!

PR #40089, opened 2025-05-30
Throw an exception if transport mTLS keystore or Truststore does not exist

ryanemerson:
Closes #40024


mhajas (approved):
Thanks @ryanemerson! LGTM!

PR #40092, opened 2025-05-30
feat: More flexible TOTP QR codes (#40091)

sonOfRa:
A possible implementation of more flexible QR code generation, as described in #40091


js/apps/admin-ui/src/authentication/policies/OtpPolicy.tsx
```diff
@@ -238,6 +241,18 @@ export const OtpPolicy = ({ realm, realmUpdated }: OtpPolicyProps) => {
               labelOff={t("off")}
             />
           )}
+          <SwitchControl
+            name={"otpPolicyUseEmail"}
+            label={t("otpPolicyUseEmail")}
+            labelIcon={t("otpPolicyUseEmailHelp")}
+            labelOn={t("on")}
+            labelOff={t("off")}
+          />
```
edewit:
```suggestion
          <DefaultSwitchControl
            name={"otpPolicyUseEmail"}
            label={t("otpPolicyUseEmail")}
            labelIcon={t("otpPolicyUseEmailHelp")}
          />
```

sonOfRa: Didn't know that existed, I'll apply it :). Since there's an existing SwitchControl directly above it, should I adjust it there as well, or do I not touch the existing code because it's not the scope of this PR?

sonOfRa: I went ahead and turned the SwitchControl above also into DefaultSwitchControl.

sonOfRa:
Something I just realized from the failing test: Should I even introduce it as a new attribute AND realm level field, or should I just add it as an attribute? It looks like the other OTP configs are both realm fields and attributes, but a lot of other things added more recently aren't. Is there a "right way" to do it currently?

js/apps/admin-ui/src/authentication/policies/OtpPolicy.tsx
edewit: :+1:

edewit (dismissed):

ahus1:
@sonOfRa - some tests fail, and even a rebase didn't fix them. Can you please have a look at the failing tests?

sonOfRa:
> @sonOfRa - some tests fail, and even a rebase didn't fix them. Can you please have a look at the failing tests?

Yeah, it's a test that validates whether realm attributes aren't additionally also present as fields: [RealmTest.excludesFieldsFromAttributes](https://github.com/keycloak/keycloak/blob/669cc2533ce8849c0f8ea48dd813db874e586269/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/admin/realm/RealmTest.java#L315-L324)

So now I'm wondering if it is correctly implemented the way I did: The two new fields are each are *both* realm attributes as well as fields on the realm representation. I did it that way because the `OTPPolicy.REALM_REUSABLE_CODE_ATTRIBUTE` is also implemented that way. However it seems like this may kind of a legacy way to do things, and not what I should be doing? See my comment above.

sonOfRa:
Hi @ahus1, 

I'm coming back to this as we also want to get this feature going in our setup: What do you think the correct way to implement this is? Both as a realm attribute AND a field on the realm (like the other OTP configs, but this seems to be deprecated?) or just one of them? I'd love to get this one done :)

stianst:
I don't like introducing these two new options on the OTP policy; for the issuer part maybe we should consider adding a short display name or something like that to the realm.

For the username part we already have the option to decide if user login with username and/or email. Could we not display email if login with username is disabled?

sonOfRa:
That sounds fine with me! So basically:

1. New realm attribute like `shortDisplayName`, and then the issuer name precedence would be `shortDisplayName` -> `displayName` -> `name`.
2. For accountName `accountName = realm.isRegistrationEmailAsUsername() && user.email != null ? user.email : user.username`

Simpler than my original approach and achieves the same goal. Do you think we should also add some explicit handling for realms that are `loginWithEmailAllowed`, but not `registrationEmailAsUsername`, or just implicitly use the username in those cases?

sonOfRa:
I've added the attribute `displayNameShort` and adjusted the issuer generation in both places (`TotpUtils` and `OTPPolicy`) to use the user's email address instead of the username in case the realm is `registrationEmailAsUsername` is set. Does this look more like what you have in mind, @stianst?

stianst (commented):
Sorry for the late reply, this looks good to me, but we need some tests to verify it works as expected

server-spi/src/main/java/org/keycloak/models/RealmModel.java
```diff
@@ -85,6 +85,10 @@ interface RealmAttributeUpdateEvent extends ProviderEvent {
 
     void setDisplayNameHtml(String displayNameHtml);
 
+    String getDisplayNameShort();
+
+    void setDisplayNameShort(String displayNameShort);
```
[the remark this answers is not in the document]
sonOfRa: Do we really want to address this one? This feels wrong to me, as no other methods on RealmModel are defaulted like this?

server-spi-private/src/main/java/org/keycloak/models/utils/ModelToRepresentation.java
```diff
@@ -493,6 +494,7 @@ public static RealmRepresentation toRepresentation(KeycloakSession session, Real
         rep.setRealm(realm.getName());
         rep.setDisplayName(realm.getDisplayName());
         rep.setDisplayNameHtml(realm.getDisplayNameHtml());
+        rep.setDisplayNameShort(realm.getDisplayNameShort());
```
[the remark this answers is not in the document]
sonOfRa: Not sure if we want to actually add this to the brief representation? It does become less brief the more things we add :)

sonOfRa:
@stianst I've added a test class covering the QR cases I could think of. Do you think we need any additional tests?

Copilot had a few suggestions and I applied the ones that looked reasonable to me. The other two (making the methods default methods, as well as adding the new field to `#toBriefRepresentation`) I skipped so far because they didn't seem as obvious as the other ones. Thoughts?

PR #40094, opened 2025-05-30
Improving query efficience when querying user group membership and role mappings

### LINKED ISSUE #39831: Optimise Composite Index Order in USER_GROUP_MEMBERSHIP Table to Improve Query Efficiency
### Before reporting an issue

### Area

core

### Describe the bug

We did load testing on Oracle and while doing so we found that a particular query on USER_GROUP_MEMBERSHIP table is using too much CPU.

We ran  explain plan for the query used ' SELECT GROUP_ID FROM USER_GROUP_MEMBERSHIP WHERE USER_ID=?" and we get 'INDEX SKIP SCAN' .

The USER_GROUP_MEMBERSHIP table is having combined index CONSTRAINT_USER_GROUP on (group_id, user_id) .

There is another index on the column (user_id), but somehow oracle is not picking up.

This skip index is costing too much resources and stalling our system leading to poor efficiency in load tests.

### Version

24.0.5

### Regression

### Expected behavior

Anyways in the composite index, the group_id column has least cardinality and is used as leading column.

From the above, it seems clear that instead of having this (group_id, user_id) we shall change index to (user_id, group_id).

Also change the select queries which has 'group_id' first when combined with 'user_id', to vice-versa where user_id is selected first to align with cardinality.

This should improve the performance

### Actual behavior

This skip index is costing too much resources and stalling our system leading to poor efficiency in load tests.

We tested in 24.0.5, but it should be the same for latest versions as well

### How to Reproduce?

Create users and add them to group and assign them any role. 
Do load test for logins

### Anything else?

We haven't checked other tables, seems there might be few indices lurking around with leading least cardinality column

pedroigor:
Closes #39831

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
@@ -132,3 +132,13 @@ a new parameter `subGroupsCount` was introduced to the following endpoints:
 
 With this parameter, clients can decide whether the count should be returned to each individual group returned. To not break existing deployments,
 this parameter defaults to `true` so that the count is returned if the parameter is not set.
+
+=== A new index was added to the `USER_GROUP_MEMBERSHIP` table
+
+In this release, a new index `constraint_user_group_by_user` was added to the `USER_GROUP_MEMBERSHIP` table to improve
+the query efficiency when fetching user group memberships.
+
+The expectation is that databases will favor this index and improve query efficiency by reducing the overall plan and execution times,
+as well as the cost.
+
+For larger databases, this index should be manually created. Please, check the server logs when upgrading to the new server version.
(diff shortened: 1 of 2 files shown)


pedroigor:
@sschu what do you think about this index update? Does it work well in MSSQL?

sschu:
@pedroigor I am not sure this improves anything. On the USER_GROUP_MEMBERSHIP_TABLE the primary key is GROUP_ID, USER_ID. So finding all users of a given group is already fast. There is also an index IDX_USER_GROUP_MAPPING containing USER_IDs. So looking up the group memberships of a user is also already fast.

The setup on USER_ROLE_MAPPING is similar. So I think this is not needed. If the indexes weren't working, we would probably have run into troubles already...

pedroigor:
@sschu Looks like I'm a lot confused about this. Thanks.

imashwin-v:
The problem is the Indexes aren't working on Oracle. 
It is efficient but not as efficient as full index scan
image

image

sschu:
@imashwin-v Whether an index is used or not is decided by the database and depends on table statistics for example. If your statistics are not up-to-date or the table doesnt contain a lot of entries, doing a skip scan might be faster.

PR #40098, opened 2025-05-30
Make UPDATE_TIME unique for MIGRATION_MODEL table

### LINKED ISSUE #40088: Make UPDATE_TIME unique for MIGRATION_MODEL table
### Description

Based on the conversation here: https://github.com/keycloak/keycloak/pull/39994#discussion_r2109606990, we should add a unique constraint to the UPDATE_TIME column in MIGRATION_MODEL table to ensure there are no duplications. When duplicates are present, the behavior might be broken for deployments as different resource versions would be returned which might lead to not properly loading resources (like account console, login screens, etc.)

A cleanup script should be also added. 

### Discussion

https://github.com/keycloak/keycloak/pull/39994#discussion_r2109606990

### Motivation

_No response_

### Details

_No response_

mabartos:
- Closes #40088

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/JpaUpdate26_2_6_RemoveDuplicateMigrationModelTime.java
+++ b/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/JpaUpdate26_2_6_RemoveDuplicateMigrationModelTime.java
@@ -1,0 +1,86 @@
+package org.keycloak.connections.jpa.updater.liquibase.custom;
+
+import liquibase.exception.CustomChangeException;
+import liquibase.statement.core.DeleteStatement;
+import liquibase.structure.core.Catalog;
+import liquibase.structure.core.Column;
+import liquibase.structure.core.Schema;
+import org.keycloak.migration.ModelVersion;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+/**
+ * Cleanup script for removing duplicated migration model update time in the MIGRATION_MODEL table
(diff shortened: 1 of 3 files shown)


mabartos:
@ahus1 Here's a PR for making the MIGRATION_MODEL.UPDATE_TIME unique with the cleanup script and creating the unique constraint. 

EDIT: I tried to have a cluster of 5 pods on the older KC version and then change the image with built KC 26.3.0 and everything should be fine. 

Not sure if we need the batch approach due to possible limited count of records, but at least we're sure it's working even in the other places and may be beneficial for some special environments.

mabartos:
@ahus1 Hi, sorry to bother you with this PR, but would you find some time to review it? As it's tied to a specific version of the patch release, it'd be great to have it merged soon rather than making other changes later when the release is out - I need to also prepare a backport.

Thank you!

ahus1:
@mabartos - I did a review, and pushed two commits: 

* One is to fix the functionality, as it had problems with the schema that was added twice.
* The other commit rewrites the test, as it would have been unstable as you cannot guarantee that the two entries are created in the same second: One might have been created at the "end" of the old second, and the other one at the "beginning" of the new second, and then the test would fail. 

Please let me know if you are ok with this changes, and then we can merge it.

mabartos:
@ahus1 Thanks for the additional changes! LGTM

>  it had problems with the schema that was added twice.

I wonder why I haven't faced the issue with my testing approach, but I agree with these changes.

> The other commit rewrites the test, as it would have been unstable as you cannot guarantee that the two entries are created in the same second: One might have been created at the "end" of the old second, and the other one at the "beginning" of the new second, and then the test would fail.

Yep, better.

ahus1:
> I wonder why I haven't faced the issue with my testing approach, but I agree with these changes.

I've been using a PostgreSQL for some manual testing, and it complained that there is no table "public.public.MIGRATION_MODEL`

ahus1 (approved):

mabartos:
> I've been using a PostgreSQL for some manual testing, and it complained that there is no table "public.public.MIGRATION_MODEL`

@ahus1 Interesting. Thanks for the review, anyway!

PR #40103, opened 2025-05-30
fix: enabling property mapping during validation

### LINKED ISSUE #40095: PropertyMapper validation should run with PropertyMappingInterceptor enabled
### Description

Having the PropertyMappingInterceptor disabled during validation makes it hard to reason about values dependent upon other configuration values.

shawkins:
closes: #40095

I was enabling property mapping to refine the validation of tls related management properties as noticed on #40044

However after reviewing again, it seems like we should just remove this validation. The problems with the logic:
- http and https can be enabled simultaneously, but the validation is effectively disabled if http is enabled
- it doesn't match what is being done for the main interface, where we just rely upon the TlsUtil failure to produce a message. If we remove the validation let TlsUtil handling determine if the config is valid.
- the message doesn't distingish that it's a problem related to the management https. The default handling also doesn't do a good job at distinguishing that it's the management interface - but that should be ideally be improved in quarkus, and we can add logic to report a more specific keycloak message.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
@@ -522,7 +522,12 @@ public class Picocli {
             return;
         }
 
-        mapper.validate(configValue);
+        PropertyMappingInterceptor.enable();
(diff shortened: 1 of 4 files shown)


mabartos (approved):
LGTM

PR #40105, opened 2025-05-30
Fix NPE during external-internal token exchange in case that user exists

### LINKED ISSUE #40104: NPE during external-internal token exchange in case that user exists
### Before reporting an issue

### Area

token-exchange

### Describe the bug

During external-internal token exchange, there is NPE. It happens just when user already exists in the `consumer` realm and is linked with the IDP, which issued the exchanged token.

We do not test this yet in our testsuite as existing tests for external-internal token exchange tests the scenario when user does not yet exists in the consumer realm and is created by external-internal token exchange request.

This issue looks as a regression of https://github.com/keycloak/keycloak/commit/7cc055f8a6e4574ff0b0057c570c 

Whole stacktrace:
```
java.lang.NullPointerException: Cannot invoke "org.keycloak.sessions.AuthenticationSessionModel.getAuthNote(String)" because "authSession" is null
	at org.keycloak.broker.provider.AbstractIdentityProvider.updateEmail(AbstractIdentityProvider.java:179)
	at org.keycloak.broker.provider.AbstractIdentityProvider.updateBrokeredUser(AbstractIdentityProvider.java:174)
	at org.keycloak.protocol.oidc.tokenexchange.AbstractTokenExchangeProvider.importUserFromExternalIdentity(AbstractTokenExchangeProvider.java:413)
	at org.keycloak.protocol.oidc.tokenexchange.AbstractTokenExchangeProvider.exchangeExternalToken(AbstractTokenExchangeProvider.java:297)
	at org.keycloak.protocol.oidc.tokenexchange.V1TokenExchangeProvider.tokenExchange(V1TokenExchangeProvider.java:103)
	at org.keycloak.protocol.oidc.tokenexchange.AbstractTokenExchangeProvider.exchange(AbstractTokenExchangeProvider.java:114)
	at org.keycloak.protocol.oidc.grants.TokenExchangeGrantType.process(TokenExchangeGrantType.java:92)
```

### Version

Keycloak main from 2025-05-30 (after commit 7cc055f8a6e4574ff0b0057c570c )

### Regression

### Expected behavior

External-internal token-exchange works for existing users

### Actual behavior

External-internal token-exchange does not work for existing users and throws NPE

### How to Reproduce?

- Make sure there is Keycloak user linked to some IDP
- Send external-token exchange request with the `subject_token` as the token from the IDP

### Anything else?

_No response_

mposolda:
closes #40104

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/broker/provider/AbstractIdentityProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/broker/provider/AbstractIdentityProvider.java
@@ -175,6 +175,9 @@ public abstract class AbstractIdentityProvider<C extends IdentityProviderModel> 
 
     protected void updateEmail(UserModel user, BrokeredIdentityContext context) {
         AuthenticationSessionModel authSession = context.getAuthenticationSession();
+        // Could be the case during external-internal token exchange
+        if (authSession == null) return;
+
         boolean isNewUser = Boolean.parseBoolean(authSession.getAuthNote(BROKER_REGISTERED_NEW_USER));
 
         if (isNewUser || IdentityProviderSyncMode.FORCE.equals(getConfig().getSyncMode())) {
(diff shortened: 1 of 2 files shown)


pedroigor (dismissed):

pedroigor:
LGTM.

But I'm wondering if we really want to update the email when doing token exchanges? Perhaps it makes more sense to skip it ...

pedroigor (approved):
Thanks. I think this was the behavior before my last change to email/verification in brokering.

PR #40107, opened 2025-05-30
Fix KcOidcBrokerTokenExchangeTest.testExternalInternalTokenExchangeUs…

### LINKED ISSUE #40106: Two same tests in KcOidcBrokerTokenExchangeTest
### Before reporting an issue

### Area

token-exchange

### Describe the bug

Since https://github.com/keycloak/keycloak/commit/3aac1eb4 there are two exactly same test methods in the `KcOidcBrokerTokenExchangeTest` . Methods are:
- `testExternalInternalTokenExchange()`
- `testExternalInternalTokenExchangeUsingIssuer()`

Was it `testExternalInternalTokenExchangeUsingIssuer()` updated by accident?
It looks that either the `testExternalInternalTokenExchangeUsingIssuer()` should be updated as it was before or test should be removed if it i problematic. To me, it would makes sense to update it as it was before (at least it seems that it works when I've updated it back, but I may be missing something....)

Using `issuer` instead of IDP-alias is something, which we support in the external-internal token exchange and hence makes sense to have it tested (See `AbstractTokenExchangeProvider.locateExchangeExternalTokenByAlias` )

### Version

nightly (from 2025-05-30)

### Regression

### Expected behavior

Test for the scenario with external-internal token exchange when parameter `subject_issuer` references IDP issuer instead of IDP alias

### Actual behavior

Test `testExternalInternalTokenExchangeUsingIssuer()` is testing alias and tests exactly the same as `testExternalInternalTokenExchange()`

### How to Reproduce?

See above

### Anything else?

_No response_

mposolda:
…ingIssuer() as it was before

closes #40106

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/broker/KcOidcBrokerTokenExchangeTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/broker/KcOidcBrokerTokenExchangeTest.java
@@ -116,7 +116,9 @@ public class KcOidcBrokerTokenExchangeTest extends AbstractInitializedBaseBroker
 
     @Test
     public void testExternalInternalTokenExchangeUsingIssuer() throws Exception {
-        assertExternalToInternalExchange(bc.getIDPAlias(), true, false);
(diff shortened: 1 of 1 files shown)


martin-kanis (approved):

mposolda:
@martin-kanis Thanks!

PR #40110, opened 2025-05-30
refinements related to service methods, serdes

shawkins:
Took the liberty of rebasing to main since I don't believe there is other pending work.

Re-opening the prior pull request, but this time it won't be a discussion about defaults.

Added a rough in of ClientService method refinements. In an ideal world this would all be pretty similar across types, possibly even allow for some generic base interfaces / classes.

**Jackson notes**:

I'd advocate for using NON_ABSENT (or NON_NULL) at a top level to allow for "empty" values such an empty string. However this has the side effect that any possibly fully empty child objects or collections will need to be specifically marked as NON_EMPTY (fabric8 does this).

**Nullability**:

Using not null / non absent heavily implies we will avoid using null as a value. If we want to use the existing representation classes for our own tests for example you can't do something like:

var client = new ClientRepresentation();
client.setSomeField(null);

Then call the endpoint and expect the serialized form to contain

"someField":null

This is how things are with the fabric8 kubernetes client built-in classes as well - and I don't think we've come across a case yet for simple put or post operations where this is a problem. 

Kubernetes explains the relationship between their metadata, defaults, and nullability here: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable - basically if a field is not explicitly nullable, then a null value or omitting the field are the same thing.

If we do need allow for some fields having explicit null values in get / put operations - and want to use our representation classes for that (which will be handy for our own tests), we'll need to be able to differentiate between null and unset. That could done internally with a wrapper, or a map. For example if wanted appUrl to be explicitly nullable:

```
public class ClientRepresentation {
    protected final Map<String, Object> state = new LinkedHashMap<>();

    @JsonPropertyDescription("URL to the application's homepage that is represented by this client")
    public String getAppUrl() {
        return state.get("appUrl");
    }

    public void setAppUrl(String appUrl) {
        state.put("appUrl", state);
    }

    public void unsetAppUrl() {
        state.remove("appUrl"); 
    }

    @JsonAnyGetter
    Map<String,String> getMap() {
        return Collections.unmodifiableMap(state);
    }

    ...

}
```

We may end up wanting a map like this in general if we want to capture additionalProperties like we do in fabric8.

To be useful for creating patches, you have to do something like the above for all field, so I'll probably just show how we can approach patching (json patch and json merge) in the next pr by just avoiding using the representation class to deserialize / create json merge patches.

**Defaults**:

Added just one default, which is in ClientRepresentation / Jackson annotation, which is the protocol defaulting to OIDC - if that's not appropriate, I'll remove it.

@keycloak/cloud-native WDYT?

### DIFF
--- a/server-spi/src/main/java/org/keycloak/services/client/ClientService.java
+++ b/server-spi/src/main/java/org/keycloak/services/client/ClientService.java
@@ -5,18 +5,51 @@ import org.keycloak.representations.admin.v2.ClientRepresentation;
 import org.keycloak.services.Service;
 import org.keycloak.services.ServiceException;
 
+import jakarta.ws.rs.core.Response.Status;
+
 import java.util.Optional;
 import java.util.stream.Stream;
 
 public interface ClientService extends Service {
 
-    Optional<ClientRepresentation> getClient(RealmModel realm, String clientId);
+    public static class ClientSearchOptions {
+        // TODO
+    }
 
-    Optional<ClientRepresentation> getClient(RealmModel realm, String clientId, Boolean fullRepresentation);
+    public static class ClientProjectionOptions {
+        // TODO
+    }
 
-    Stream<ClientRepresentation> getClients(RealmModel realm);
+    public static class ClientSortAndSliceOptions {
+        // order by
+        // offset
+        // limit
+        // NOTE: this is not always the most desirable way to do pagination
+    }
 
-    ClientRepresentation createOrUpdateClient(RealmModel realm, ClientRepresentation client) throws ServiceException;
+    Optional<ClientRepresentation> getClient(RealmModel realm, String clientId, ClientProjectionOptions projectionOptions);
+
+    Stream<ClientRepresentation> getClients(RealmModel realm, ClientProjectionOptions projectionOptions, ClientSearchOptions searchOptions, ClientSortAndSliceOptions sortAndSliceOptions);
+
+    ClientRepresentation deleteClient(RealmModel realm, String clientId);
+
+    Stream<ClientRepresentation> deleteClients(RealmModel realm, ClientSearchOptions searchOptions);
+
+    default ClientRepresentation createOrUpdateClient(RealmModel realm, ClientRepresentation client) throws ServiceException {
+        // If we can paramertize the services, it makes sense to define default handling like this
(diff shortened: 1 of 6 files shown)


vmuzikar:
+100 for NON_ABSENT (or at least NON_NULL). I'd really hate to have ambiguity between null and not set.

shawkins:
@vmuzikar @mabartos and others, I believe this is ready. Other than using a JsonProperty annotation to describe the protocol default, there is nothing else that is in the initial scope wrt defaults.

We have talked about a pipeline of defaulting, validation, and canonicalization that should be performed prior to storage. In reviewing the old logic the defaulting and validation seem to be handled in RepresentationToModel, adjunct logic like the LoginProtocolFactory, and in the service layer. Since we are mapping directly to the model layer, we will end up needing to replicate all that logic as it's dependent upon the legacy representation logic. It might be worth capturing why the approach of mapping directly to the model layer is preferable initially to an approach where we layer on top of the v1 representation.

server-spi/src/main/java/org/keycloak/services/client/ClientService.java
```diff
@@ -5,18 +5,51 @@
...
+            return createClient(realm, client);
+        } catch (ServiceException e) {
+            if (!e.getSuggestedResponseStatus().filter(Status.CONFLICT::equals).isPresent()) {
+                throw e;
+            }
+        }
+        return updateClient(realm, client);
+    }
+
+    ClientRepresentation updateClient(RealmModel realm, ClientRepresentation client) throws ServiceException;
```
vmuzikar, line 51: Will we need some switch/flag for partial/PATCH updates?

vmuzikar (approved):
@shawkins Sorry for the late review and thanks for the PR. LGTM.

We'll need to address the PATCHing (how do we remove a field? can we set it to `null` to remove it? etc.) but let's handle that as a separate PR.

server-spi/src/main/java/org/keycloak/services/client/ClientService.java
shawkins: > Will we need some switch/flag for partial/PATCH updates?

That will be a separate method.

PR #40126, opened 2025-06-02
added DCL pattern implementation

### LINKED ISSUE #40030: Potential thread safety Issue with lazy init of transformerFactory at TransformerUtil. SAST
### Description

This method contains an unsynchronized lazy initialization of a static field. After the field is set, the object stored into that location is further updated or accessed. The setting of the field is visible to other threads as soon as it is set. If the further accesses in the method that set the field serve to initialize the object, then you have a very serious multithreading bug, unless something else prevents any other thread from accessing the stored object until it is fully initialized.

Even if you feel confident that the method is never called by multiple threads, it might be better to not set the static field until the value you are setting it to is fully populated/initialized.

See [CWE-543: Use of Singleton Pattern Without Synchronization in a Multithreaded Context](https://cwe.mitre.org/data/definitions/543.html)

### Discussion

_No response_

### Motivation

See Description

### Details

Incorrect lazy initialization and update of static field `transformerFactory` in method `getTransformerFactory()` :

https://github.com/keycloak/keycloak/blob/f473f8df3e80c1c336aa253b3529900a55f57ad0/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java#L70
https://github.com/keycloak/keycloak/blob/f473f8df3e80c1c336aa253b3529900a55f57ad0/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java#L95-L136

Problem: The current implementation is not thread-safe. If multiple threads call `getTransformerFactory()` simultaneously, they may end up creating multiple instances of `transformerFactory`. 

Proposed solution: Consider using "Double-Checked-Locking" pattern for thread-safe lazy initialization.

Anchels:
Closes #40030

### DIFF
--- a/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java
+++ b/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java
@@ -37,7 +37,6 @@ import javax.xml.stream.XMLEventReader;
 import javax.xml.stream.events.Attribute;
 import javax.xml.stream.events.Characters;
 import javax.xml.stream.events.Comment;
-import javax.xml.stream.events.EndElement;
 import javax.xml.stream.events.Namespace;
 import javax.xml.stream.events.StartElement;
 import javax.xml.stream.events.XMLEvent;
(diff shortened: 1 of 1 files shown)


ahus1:
Thank you for the PR, I'm having a look at it.

ahus1:
@Anchels - I had a look at the PR and it looks valid. 

One thing I noticed is that the volatile variable should only be assigned after it is fully initialized. Otherwise there is a risk of a race condition, and other threads might access a partially initialized instance. 

Please let me know if you agree with the changes.

Anchels:
@ahus1 - I share your perspective. Thank you for your constructive feedback.

ahus1 (approved):

PR #40127, opened 2025-06-02
Add server type cluster to test-framework

### LINKED ISSUE #39962: Create a POC of running 2 containers in the new testsuite
### Description

Investigate options how to run two container (with two different images) within one test case in the new test framework.

### Discussion

_No response_

### Motivation

We need to test this for supporting rolling updates for patch releases.

### Details

_No response_

mhajas:
Closes #39962
This PR is adding the following:

- Creates a new `ContainerKeycloakCluster` implementation of `KeycloakServer` that can start multiple containers of Keycloak with configurable images
- Updates `DockerKeycloakDistribution` so it can be configured with a different Keycloak image
- Updates `InjectKeycloakUrls` annotation with `nodeIndex` field so we can specify index of the node we want the URL for


stianst:
To better separate things you should move the clustering distribution to an extension (`test-framework/clustering`) rather than have it in core (we want that to be small and cover only the stuff most folks use).

The clustering specific tests should also be moved to `tests/clustering`. We want to break up more specialized tests into separate modules. You can then also specify the default server mode for the clustering tests by adding `tests/clustering/src/test/resources/keycloak-test.properties` with `kc.test.server=cluster`.

test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractContainerTestDatabase.java
```diff
@@ -80,7 +82,12 @@ public String getPassword() {
     }
 
     public String getJdbcUrl() {
-        return container.getJdbcUrl();
+        var url = container.getJdbcUrl();
+        if (internal) {
```
stianst, line 86: `internal` doesn't make much sense to me here. Could you elaborate a bit on what this stuff does?

test-framework/core/src/main/java/org/keycloak/testframework/server/ContainerKeycloakClusterSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+public class ContainerKeycloakClusterSupplier extends AbstractKeycloakServerSupplier {
+
+    private static final Logger LOGGER = Logger.getLogger(ContainerKeycloakClusterSupplier.class);
+
+    @ConfigProperty(name = "debug", defaultValue = "false")
+    boolean debug = false;
```
stianst, line 11: `debug` doesn't seem to do anything in `DockerKeycloakDistribution`

Around logging it may be nice to replicate what was done for databases, that allows including the logging for managed containers, and configuring that using the logging config for the test framework.

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakUrls.java
```diff
@@ -67,4 +67,8 @@ private KeycloakUriBuilder toBuilder(String url) {
     public String getToken(String realm) {
         return baseUrl + "/realms/" + realm + "/protocol/" + OIDCLoginProtocol.LOGIN_PROTOCOL + "/token";
     }
+
+    public boolean isEnabled() {
```
stianst, line 71: What does this do?

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakUrlsSupplier.java
```diff
@@ -9,12 +9,28 @@ public class KeycloakUrlsSupplier implements Supplier<KeycloakUrls, InjectKeyclo
...
     @Override
     public boolean compatible(InstanceContext<KeycloakUrls, InjectKeycloakUrls> a, RequestedInstance<KeycloakUrls, InjectKeycloakUrls> b) {
-        return true;
+        return a.getAnnotation().nodeIndex() == b.getAnnotation().nodeIndex();
+    }
+
+    @Override
+    public String getRef(InjectKeycloakUrls annotation) {
+        //TODO: ref to identify the instances, is this correct?
+        return "node-" + annotation.nodeIndex();
```
stianst, line 34: Ref should be supplied by the user; if they want to inject two different KeycloakURLs they can give it a ref.

Take a look at managed realm for instance, and how you would inject two different realms into a test case.

stianst (commented):
I'm okay with this more or less how it is, but would put it into separate places.

One challenge here is none of the other stuff is aware about the multiple nodes/URLs (managed realms, admin clients, oauth client, etc.).

Think for clustering tests in general it would be useful to have a reverse proxy in the mix (something we can add later though), so that there is a single URL to reach the cluster; then have some mechanism to specify the node directly.

Then maybe something to allow switching how all the clients contact the cluster, so would be more like:

```
mytest() {
   urls.reverserProxy();
   oauth.doLogin();
   urls.node(2);
   oauth.doLogin();
}
```

For now though I would probably just not change `KeycloakUrls` and instead just add a `ManagedCluster` or something where you add a method `ManagedCluster.getUrl(int node)` or something like that.

test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractContainerTestDatabase.java
mhajas: We can't use localhost to connect to database because Keycloaks running in containers do not have access to it. What this flag is doing is that it configures Keycloak with IP address of the database container that is accessible from another container. Internal means it is within Podman/Docker network. I don't like it much too but I didn't find any better way. Maybe we can remove the configuraiton option and use this only when `cluster` server type is used. WDYT?

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakUrls.java
mhajas: It is basically a shortcut to check we didn't end up in this branch: https://github.com/keycloak/keycloak/pull/40127/files#diff-7ad204a4e57825d8a74d7fd8cc4df19b56ea7bdda01d7667bf3d0ced1209c6c5R23 when looking for a node with given index.

mhajas:
I am closing this in favor of https://github.com/keycloak/keycloak/pull/40283 since I am not able to push to this PR as it is opened from Pedro's branch.

PR #40131, opened 2025-06-02
make property name unique

### LINKED ISSUE #40128: Unable to set LoA field in auth-flow-enforcer
### Before reporting an issue

### Area

admin/ui

### Describe the bug

This is about configuring an "**auth-flow-enforcer**" in a client policy profile.
The two fields "Auth Flow **Alias**" and "Auth Flow **Loa**" are always having the same value when editing any of them.
Somehow the UI always mirrors the entered value directly into the other field.
This bug prevents setting a reasonable LoA for an auth flow alias. 

### Version

26.2.5

### Regression

### Expected behavior

Entering text into field "Auth Flow **Alias**" must not alter the value of field "Auth Flow **Loa**" and vice versa.

### Actual behavior

Entering text into field "Auth Flow **Alias**" also puts the same text into field "Auth Flow **Loa**".

### How to Reproduce?

- go to _Realm settings_ > _Client profiles_ 
- create a new profile
- in the new profile add an executor of type "auth-flow-enforcer"
- enter text into field "Auth Flow **Alias**"

### Anything else?

_No response_

edewit:
fixes: #40128

### DIFF
--- a/services/src/main/java/org/keycloak/services/clientpolicy/executor/AuthenticationFlowSelectorExecutorFactory.java
+++ b/services/src/main/java/org/keycloak/services/clientpolicy/executor/AuthenticationFlowSelectorExecutorFactory.java
@@ -34,13 +34,14 @@ public class AuthenticationFlowSelectorExecutorFactory implements ClientPolicyEx
     public static final String PROVIDER_ID = "auth-flow-enforcer";
 
     public static final String AUTH_FLOW_ALIAS = "auth-flow-alias";
+    public static final String AUTH_FLOW_LOA = "auth-flow-loa";
 
     private static final ProviderConfigProperty AUTH_FLOW_ALIAS_PROPERTY = new ProviderConfigProperty(
             AUTH_FLOW_ALIAS, "Auth Flow Alias", "Insert the alias of the authentication flow",
             ProviderConfigProperty.STRING_TYPE, null);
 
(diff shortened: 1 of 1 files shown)


ahus1:
@graziang - can you please review this PR? 

I assume there is also no good way to migrate / fix any existing configurations ... so we would need to add something to the release notes that they need to update the LOA again manually after upgrading? 

cc: @mposolda

graziang (commented):
This executor was added in 26.2 to select an auth flow from client policies and optionally set the level of authentication to be applied once the selected flow is completed.
LOA must be an integer, and due to this bug, both the LOA and flow alias end up having the same value. So, for the "Auth Flow Loa" value to work as intended, the auth flow would need to use a workaround by naming the flow with an integer. This makes me think that no one has actually used this executor with "Auth Flow Loa" so far, but only with "Auth Flow Alias" to select the flow.
So IMO, a migration is not necessary, and simply renaming the property as in this PR is fine for me, maybe adding a line in the release notes to mention that the "Auth Flow Loa" needs to be reset in the executor if it was used.

ahus1 (approved):

ahus1:
@graziang - thank you for the clarification. As I understand that it wasn't working before, and as the issue appears in the release notes in upstream and downstream anyway, this is then good to be merged as is.

PR #40134, opened 2025-06-02
Introduce ExternalToInternalTokenExchangeProvider. Make it working wi…

mposolda:
…th Google IDP using token-info endpoint instead of user-info endpoint

closes #40146
closes #40133

- PR introduces provider `ExternalToInternalTokenExchangeProvider` and the feature `TOKEN_EXCHANGE_EXTERNAL_INTERNAL_V2` . See https://github.com/keycloak/keycloak/issues/40146 for some reasoning for using "external internal" only for now and make it separated from "internal external" .

- Similarly like we did for "standard token exchange", we may want to develop supported "external internal" and at the same time make the token-exchange:v1 works the same as before. This means also that Identity providers may need to behave differently for V1 and V2. Due to this, I've added the method `getVersion` on the `TokenExchangeProvider` and methods like `exchangeExternalTokenV1Impl` and `exchangeExternalTokenV2Impl` on the Identity provider.

- The `exchangeExternalTokenV1Impl` works the same as before for V1 token exchange.

- The `exchangeExternalTokenV2Impl` is implemented only for Google provider now. So Google is the only provider, which works with `ExternalToInternalTokenExchangeProvider` right now. Instead of using only user-info endpoint for the verification (like V1 is doing), The V2 uses token-verification, which works with Google token-info endpoint and checks that token has expected audience. See some details in the GH issue https://github.com/keycloak/keycloak/issues/40133 and in the [google doc](https://docs.google.com/document/d/1hmUpMfvAwyRBvUhCD01IEGNjx1yIh9a8FpGCQlmrOno/edit?tab=t.0#heading=h.ptu7z1gumjz) regarding this.

Other providers (OIDC, other social providers) can hopefully be done in a follow-up PRs. Created some subtasks of https://github.com/keycloak/keycloak/issues/40132 for this.

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -75,6 +75,7 @@ public class Profile {
 
         TOKEN_EXCHANGE("Token Exchange Service", Type.PREVIEW, 1),
         TOKEN_EXCHANGE_STANDARD_V2("Standard Token Exchange version 2", Type.DEFAULT, 2),
+        TOKEN_EXCHANGE_EXTERNAL_INTERNAL_V2("External to Internal Token Exchange version 2", Type.EXPERIMENTAL, 2),
 
         WEB_AUTHN("W3C Web Authentication (WebAuthn)", Type.DEFAULT),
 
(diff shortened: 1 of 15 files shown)


graziang (approved):
@mposolda Thanks, good start for external to internal! I create the task to add the tests

mposolda:
@graziang Thanks for the review

PR #40150, opened 2025-06-02
Adjusted null checks

### LINKED ISSUE #40061: Redundant null-checks. SAST
### Description

Redundant check of a known non-null value against the constant null.

### Discussion

_No response_

### Motivation

Code clarity and maintainability

### Details

1. The `old` variable is being compared to null twice:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/saml-core-api/src/main/java/org/keycloak/saml/common/constants/JBossSAMLConstants.java#L253-L258

2. The method `String.split()` does not return null:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java#L74-L77

3. The `qual` variable is being compared to null:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java#L389-L391
although it has already been dereferenced before:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java#L383-L384

4. Same for `attributeValue` variable here:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/saml-core/src/main/java/org/keycloak/saml/processing/core/saml/v2/writers/BaseWriter.java#L233-L239

---

Anchels:
Closes #40061

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
+++ b/server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
@@ -73,7 +73,7 @@ public class ComponentModelScope implements Scope {
     @Override
     public String[] getArray(String key) {
         final String[] res = get(prefix + key, "").split("\\s*,\\s*");
-        return (res == null) ? origScope.getArray(key) : res;
+        return (res.length == 0) ? origScope.getArray(key) : res;
     }
 
     @Override
--- a/server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
+++ b/server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
@@ -73,7 +73,7 @@ public class ComponentModelScope implements Scope {
     @Override
     public String[] getArray(String key) {
         final String[] res = get(prefix + key, "").split("\\s*,\\s*");
-        return (res.length == 0) ? origScope.getArray(key) : res;
+        return (res == null) ? origScope.getArray(key) : res;
     }
 
     @Override
--- a/saml-core-api/src/main/java/org/keycloak/saml/common/constants/JBossSAMLConstants.java
+++ b/saml-core-api/src/main/java/org/keycloak/saml/common/constants/JBossSAMLConstants.java
@@ -252,7 +252,7 @@ public enum JBossSAMLConstants {
                 String key = c.get();
                 if ((old = m.put(key, c)) != null) {
 //                    System.out.println("WARNING: " + old);
-                    if (old != null && c.getAsQName().equals(old.getAsQName())) {
+                    if (c.getAsQName().equals(old.getAsQName())) {
                         throw new IllegalStateException("Same name " + key + " used for two distinct constants: " + c + ", " + old);
                     }
                     m.put(key, null);
--- a/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java
+++ b/saml-core/src/main/java/org/keycloak/saml/common/util/TransformerUtil.java
@@ -380,7 +380,7 @@ public class TransformerUtil {
                 if (prefix != null && ! prefix.isEmpty())
                     qual = (localPart != null && ! localPart.isEmpty()) ? prefix + ":" + localPart : prefix;
 
-                if (qual.equals("xmlns"))
+                if ("xmlns".equals(qual))
                     continue;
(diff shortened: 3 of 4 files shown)


shawkins (dismissed):
LGTM

server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
```diff
@@ -73,7 +73,7 @@ public String get(String key, String defaultValue) {
     @Override
     public String[] getArray(String key) {
         final String[] res = get(prefix + key, "").split("\\s*,\\s*");
-        return (res == null) ? origScope.getArray(key) : res;
+        return (res.length == 0) ? origScope.getArray(key) : res;
```
ahus1: When the `get().split()` never returns null, the new code changes the runtime behavior. Now an empty value will fall back to the original scope.

If we want to have a logic where we can override the value with an empty string to create an empty list, we should rather change the code to the following: 

```
    public String[] getArray(String key) {
        String val = get(prefix + key, null);
        if (val == null) {
            return origScope.getArray(key);
        }
        return val.split("\\s*,\\s*");
    }
```

Again, this would be a runtime change. For better documentation in the changelog and the associated discussion, I would take this one into a separate issue and PR.

ahus1 (changes requested):
Except for the last change all changes seems to be good to me. See below for a suggested change.

server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
shawkins: > I would take this one into a separate issue and PR.

That seems good. It's clear that the original code was not correct, but you also correct that we should make sure that any new behavior is well-defined.

Original behavior:
- all calls to getArray would return res, the origScope value is not used.

In this pr:
- calls to getArray where an empty string or no config value exists will return the origScope value

@ahus1 proposed behavior:
- calls to getArray where no config value exists will return the origScope value

I'm inclined to agree with @ahus1, but would also point out that the treatment of null vs empty is not consistent. See SystemPropertiesConfigProvider and JsonConfigProvider both perform isEmpty checks - but importantly MicroProfileConfigProvider does not.

Anchels: @ahus1, @shawkins - Thanks! Should I revert this change?

ahus1: I have reverted that change and will approve the rest of the changes in a minute.  

Please move this discussion to a new bug issue so we can decide which direction we want to go. Thanks!

ahus1 (approved):

server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java
shawkins: To add a bit more for whatever issue is created - the empty to null conversion is in the MicroProfileConfigProvider logic. It's using config.getOptionalValue, which applies a converter that treats empty as null. For example if you put `spi...--some-prop=` in the keycloak.conf, then use scope.get("some-prop", "default") - it will return "default", not "".

That makes the original proposal consistent with the logic in MicroProfileConfigProvider, SystemPropertiesConfigProvider, and somewhat JsonConfigProvider (although its array logic is not consistent either), but inconsistent with the rest of the methods in ComponentModelScope. 

It would be good to hear from a core team if null or empty string values are even expected in component model configs. 

And in general we may also need to document / refine our handling of empty value with our other config logic - unfortunately the smallrye getConfigValue and getOptionalConfig methods have differring behavior wrt to empty strings and I don't belive we have considered that. cc @mabartos @vmuzikar

Anchels: > Please move this discussion to a new bug issue so we can decide which direction we want to go. Thanks!

Moved the discussion to https://github.com/keycloak/keycloak/issues/42000

PR #40168, opened 2025-06-02
Added new searchByAttributes function to UsersResource with the exact parameter

### LINKED ISSUE #39609: Users searchAttributes broken for empty value
### Before reporting an issue

### Area

admin/client-java

### Describe the bug

When i use the keycloak admin client i retrieve a `org.keycloak.admin.client.resource.UsersResource`.
Now i use the searchByAttributes function: 
```
List<UserRepresentation> searchByAttributes(@QueryParam("first") Integer var1, @QueryParam("max") Integer var2, @QueryParam("enabled") Boolean var3, @QueryParam("briefRepresentation") Boolean var4, @QueryParam("q") String var5);
```
When the query String is something like `myAttribute:myValue` it still works fine. But a query like `myAttribute:` now returns an empty list. Previously it would return all Users that have that attribute.
This changed in 26.0.8 ([https://github.com/keycloak/keycloak/commit/ad679b872978eafda5d47f83cecbaa0873c4a3ff](url))

I also built the server locally and changed the default to false then it worked again.
An easy fix would be to allow setting the `exact` parameter for `searchByAttributes`

### Version

26.2.4

### Regression

### Expected behavior

All Users that have this attribute should be returned.

### Actual behavior

An empty List of Users is returned.

### How to Reproduce?

1. Create a user
2. Add a custom attribute like `myAttribute` with the value `myValue`
3. Use the admin client `searchByAttributes` with the following query: `myAttribute :`
4. The result will be an empty list

### Anything else?

_No response_

Michael-AT-Corporation:
Closes #39609

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -1,0 +1,36 @@
+== Breaking changes
+
+Breaking changes are identified as requiring changes from existing users to their configurations.
+In minor or patch releases we will only do breaking changes to fix bugs.
+
+=== <TODO>
+
+== Notable changes
+
+Notable changes where an internal behavior changed to prevent common misconfigurations, fix bugs or simplify running {project_name}.
+
+=== Usage of the `exact` request parameter when searching users by attributes
+
+If you are querying users by attributes through the User API where you want to fetch users that match a specific attribute key (regardless the value),
+you should consider setting the `exact` request parameter to `false` when invoking the `/admin/realms/{realm}/users` using
+the `GET` method.
+
+For instance, searching for all users with the attribute `myattribute` set should be done as follows:
+
+```
+GET /admin/realms/{realm}/users?exact=false&q=myattribute:
+```
+
+The {project_name} Admin Client is also updated with a new method to search users by attribute using the `exact` request parameter.
+
+== Deprecated features
+
+The following sections provide details on deprecated features.
+
+=== <TODO>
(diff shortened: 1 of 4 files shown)


pedroigor:
Hi @Michael-AT-Corporation, sorry for the delay ....

I've updated your PR to include a note in the upgrade guide as well as small updates to the test.

Thanks again for your contribution.

pedroigor (dismissed):

ahus1 (changes requested):
The KC 26.3 release branch has already been created, so this will end up in 26.4. This needs to be reflected by moving the upgrading guide notes to a new file.

pedroigor:
@ahus1 Can I send a PR to 26.3 release branch as well?

ahus1:
@pedroigor - even if you send a PR to the 26.3 release branch, it will IMHO not be part of the 26.3.0 release, but 26.3.1 - and then it will again need its own upgrading guide. You'll need to check with @pskopek as AFAIK the release of 26.3.0 is in progress already.

pedroigor:
@ahus1 can you please re-review?

ahus1 (dismissed):
Upgrading guide now looks good.

ahus1:
> Upgradng guide now looks good

... one could argue a bit more if you plan to backport it to earlier releases. Still I won't start that.

pedroigor (dismissed):

pedroigor (approved):

Michael-AT-Corporation:
@pedroigor Thanks for your help 🙂

PR #40175, opened 2025-06-03
Updated documentation to handle the conf folder on upgrades

ahus1:
Closes #40046

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
@@ -132,3 +132,11 @@ a new parameter `subGroupsCount` was introduced to the following endpoints:
 
 With this parameter, clients can decide whether the count should be returned to each individual group returned. To not break existing deployments,
 this parameter defaults to `true` so that the count is returned if the parameter is not set.
+
+=== Upgrade procedure changed for the distribution
+
+If you are upgrading {project_name} by downloading the distribution, the upgrade procedure has been changed. Previously it recommended copying over the contents from the `conf/` folder from the old to the new installation.
+The new procedure recommends to re-apply any changes to files in the `conf/` in case you modified any files in that directory, and leave them otherwise unchanged.
+
+This prevents accidentally downgrading functionality, for example, by using an old `cache-ispn.xml` file from a previous version.
+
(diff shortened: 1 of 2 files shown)


ahus1:
@mhajas  / @ryanemerson / @vmuzikar / @shawkins - The copying over of the `conf` folder proved problematic as it would perpetuate old ISPN cache configurations. 

This is now a reworded version - please let me know your thoughts. A follow-up idea would be not to have a `cache-ispn.xml` file in the folder by default, but that would be beyond this PR IMHO.

I'm also considering to backport at lest the docs changes to 26.2.

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -132,3 +132,11 @@ a new parameter `subGroupsCount` was introduced to the following endpoints:
 
 With this parameter, clients can decide whether the count should be returned to each individual group returned. To not break existing deployments,
 this parameter defaults to `true` so that the count is returned if the parameter is not set.
+
+=== Upgrade procedure changed for the distribution
+
+If you are upgrading {project_name} by downloading the distribution, the upgrade procedure has been changed. Previously it recommended copying over the contents from the `conf/` folder from the old to the new installation.
+The new procedure recommends to re-apply any changes to files in the `conf/` in case you modified any files in that directory, and leave them otherwise unchanged.
```
mhajas: Should we limit this only to `cache-ispn.conf` file? We don't have any defaults in `keycloak.conf` so copying should not be breaking.

mhajas (commented):
Thanks for this improvement. I think this makes sense for `cache-ispn`, however, I am not sure about other files. I added a comment into code.

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
ahus1: Added another commit, please let me know if this addresses your comment.

ryanemerson (approved):
+1 to this change, the docs are very clear now.

"Re-apply your changes based on the `cache-ispn.xml`..." doesn't strike me as particularly user-friendly though, so I think we should make a habit of trying to minimise these changes in patch releases as much as possible*.

\* I understand in this user case their issue wasn't caused by an actual change to the file, I'm just making the wider point.

ahus1:
@ryanemerson: 

> Re-apply your changes based on the cache-ispn.xml..." doesn't strike me as particularly user-friendly 

In the related case, they were still having a KC 26.0 or earlier `cache-ispn.xml` which was copied over to 26.1 and then to 26.2. The only fix to not having to re-applying changes to cache-ispn.xml is to not have one, and move more to CLI parameters.

ryanemerson:
> The only fix to not having to re-applying changes to cache-ispn.xml is to not have one, and move more to CLI parameters.

I agree and I think that's what we should be aiming for longer term.

mhajas (approved):
Thanks @ahus1, LGTM

PR #40178, opened 2025-06-03
Removed the keyName field in KeySelectorUtilizingKeyNameHint

### LINKED ISSUE #40067: Always null field in KeySelectorUtilizingKeyNameHint. SAST
### Description

All writes to this field are of the constant value null, and thus all reads of the field will return null. 

### Discussion

_No response_

### Motivation

Code clarity

### Details

The private field `keyName` is only ever set to null following changes in a commit cab7e504:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/saml-core/src/main/java/org/keycloak/saml/processing/core/util/XMLSignatureUtil.java#L132-L163

It is utilized in just one location for logging purposes:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/saml-core/src/main/java/org/keycloak/saml/processing/core/util/XMLSignatureUtil.java#L522-L532

Anchels:
Closes #40067

### DIFF
--- a/saml-core/src/main/java/org/keycloak/saml/processing/core/util/XMLSignatureUtil.java
+++ b/saml-core/src/main/java/org/keycloak/saml/processing/core/util/XMLSignatureUtil.java
@@ -135,8 +135,6 @@ public class XMLSignatureUtil {
 
         private boolean keyLocated = false;
 
-        private String keyName = null;
-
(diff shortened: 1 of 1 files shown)


shawkins (approved):
LGTM

ahus1 (approved):

PR #40182, opened 2025-06-03
Added override equals methods to child classes

### LINKED ISSUE #40177: Subsidiary class doesn`t override "equals" method
### Description

Hello. guys!
This time, Minor defects were found with overriding the `equals` method in the child class.

This class extends a class that defines an `equals` method and adds fields, but doesn't define an `equals` method itself.
Even if you don't need to override the `equals` method, consider overriding it anyway to document the fact that the `equals` method for the subclass just return the result of invoking `super.equals(o)`.

https://github.com/keycloak/keycloak/blob/d656aa747f54845abd976e85cd644d7590432e0f/server-spi-private/src/main/java/org/keycloak/models/utils/ReadOnlyUserModelDelegate.java#L31
and here
https://github.com/keycloak/keycloak/blob/d656aa747f54845abd976e85cd644d7590432e0f/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultAttributes.java#L66

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

---

LM4O322:
Added override equals methods to child classes

Closes #40177

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/ReadOnlyUserModelDelegate.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/ReadOnlyUserModelDelegate.java
@@ -155,6 +155,11 @@ public class ReadOnlyUserModelDelegate extends UserModelDelegate {
         throw readOnlyException("role mapping for role " + role.getName());
     }
 
+    @Override
+    public boolean equals(Object o) {
+        return super.equals(o);
+    }
+
     private RuntimeException readOnlyException(String detail) {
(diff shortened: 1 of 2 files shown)


shawkins:
Needs the conflict resolved and additional comments to clarify why these methods exist.

LM4O322:
I will be glad to help you, but a little later
At the moment I do not have the ability to correct the inclusion request

shawkins:
Closing due to lack of updates, and the issue was already closed.

Issue #40195, opened 2025-06-03
Documentation of Argon2 hash-length configuration property is incorrect.


RutgerLubbers:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

docs

### Describe the bug

The argon2 password SPI has a configuration for the hash-length. This configuration's documentation is wrong.
The hash-length is documented to be set via the `spi-password-hashing-argon2-type` property, but it should be `spi-password-hashing-argon2-hash-length`.

The documentation comes from the `getConfigMetaData()` method, from the `Argon2PasswordHashProviderFactory` class. 

What needs to change is:
```java

        builder.property()
                .name(TYPE_KEY)
                .type("int")
                .helpText("Hash length")
                .defaultValue(Argon2Parameters.DEFAULT_HASH_LENGTH)
                .add();
```
into
```java

        builder.property()
                .name(HASH_LENGTH_KEY)
                .type("int")
                .helpText("Hash length")
                .defaultValue(Argon2Parameters.DEFAULT_HASH_LENGTH)
                .add();
```

I can create a PR for this.

### Version

26.2.5

### Regression

- [ ] The issue is a regression

### Expected behavior

The documentation should be correct.

### Actual behavior

The documentation shows `spi-password-hashing-argon2-type` instead of `spi-password-hashing-argon2-hash-length`.

### How to Reproduce?

See the argon2 documentation at `https://www.keycloak.org/server/all-provider-config#_argon2`

ahus1:
Thank you for reporting this. As a follow-up, once this is merged, maybe you could create an issue and a pull request to make `ProviderConfigurationBuilder` safe to prevent adding the same name twice as means of defensive programming to prevent future situations like this.

RutgerLubbers:
> Thank you for reporting this. As a follow-up, once this is merged, maybe you could create an issue and a pull request to make `ProviderConfigurationBuilder` safe to prevent adding the same name twice as means of defensive programming to prevent future situations like this.

I've created https://github.com/keycloak/keycloak/issues/40233 for this, now I can work on a PR.

PR #40196, opened 2025-06-03
Update documentation for Argon2 hash-key length to use the correct property.

RutgerLubbers:
Update the Argon2 documentation to use the correct property for the hash-length. 
See https://github.com/keycloak/keycloak/issues/40195


RutgerLubbers:
Use other PR with proper signed commit.

PR #40201, opened 2025-06-03
Change discovery in Kubernetes to jdbc-ping

### LINKED ISSUE #39544: Change discovery in Kubernetes to `jdbc-ping`
### Description

With 26.1, we switched to the cache-stack `jdbc-ping` as the default in all environments except on Kubernetes. 

While `kubernetes` as a stack is stable, it add another configuration option that people need to understand, and a different option that needs testing. 

So the idea is to make `jdbc-ping` the default for the KC Operator. 

Along with that, deprecate the `kubernetes` stack.

### Discussion

https://github.com/keycloak/keycloak/issues/39454#issuecomment-2858416954

### Motivation

Reduce hidden errors due to complexity. Reduce the amount of documentation. Avoid having people adding JGroups parameter for DNS discovery. 

### Details

The headless service in Kubernetes is still required by the StatefulSet.

ryanemerson:
Closes #39544

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
@@ -100,6 +100,12 @@ Consider one of the following cases and recommended migration steps:
 * With the feature `persistent-user-session` feature enabled, the remote store configuration is ignored and {project_name} will print a warning.
 ====
 
+=== Kubernetes cache stack has been deprecated
+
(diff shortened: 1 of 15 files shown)


ahus1 (approved):
Thank you for the change. 

cc: @keycloak/cloud-native  - The headless service that was initially added to do the discovery will still be in place as I hear this is a best practice for any StatefulSet that we keep. 

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

> StatefulSets currently require a [Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) to be responsible for the network identity of the Pods. You are responsible for creating this Service.

Issue #40099, opened 2025-05-30
[Keycloak Operator CI] - Test OLM Installation


ikhomyn:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

ci

### Describe the bug

`error parsing HTTP 499 response body: invalid character '<' looking for beginning of value: "<!DOCTYPE html>\n<html>\n  <head>\n    <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n    <meta charset=\"utf-8\">\n    <title>Application Error</title>\n    <style media=\"screen\">\n      html,body,iframe {\n        margin: 0;\n        padding: 0;\n      }\n\n      html,body {\n        height: 100%;\n        overflow: hidden;\n      }\n\n      iframe {\n        width: 100%;\n        height: 100%;\n        border: 0;\n      }\n    </style>\n  </head>\n  <body>\n    <iframe src=\"[https://www.herokucdn.com/error-pages/application-error.html\](https://www.herokucdn.com/error-pages/application-error.html/)"></iframe>\n  </body>\n</html>"`
Error: Process completed with exit code 1.

https://github.com/keycloak/keycloak/actions/runs/15336283564/job/43154257333#step:9:5913
https://github.com/keycloak/keycloak/actions/runs/15324971197/job/43135161773#step:9:769
https://github.com/keycloak/keycloak/actions/runs/15322156793/job/43108326992#step:9:762
https://github.com/keycloak/keycloak/actions/runs/15307248702/job/43063268179#step:9:773
https://github.com/keycloak/keycloak/actions/runs/15306958721/job/43062250593#step:7:41

### Version

latest

### Regression

- [ ] The issue is a regression

### Expected behavior

-

### Actual behavior

-

### How to Reproduce?

-

shawkins:
Marking as important given the frequency with which this occurring. Not sure yet on level of effort to wokaround.

ahus1:
This is also frequently failing in the 26.0 branch, see https://github.com/keycloak/keycloak/actions/runs/15921210362/attempts/1?pr=40750

It would be good to backport it to 26.0 as well. Thanks!

shawkins:
> This is also frequently failing in the 26.0 branch, see https://github.com/keycloak/keycloak/actions/runs/15921210362/attempts/1?pr=40750
> 
> It would be good to backport it to 26.0 as well. Thanks!

[...]

PR #40202, opened 2025-06-03
JGroups: Switch to "per-destination" bundler for jdbc-ping

ryanemerson:
Closes #39545

I have manually tested that the per-destination bundler gets applied via the remote debugger. Do we have any existing tests that allows us to extract the Infinispan `DefaultCacheManager` and verify the applied settings are correct for a given user config?

I looked in `model/infinispan`, but all I could find were outdated tests which I propose we remove: https://github.com/keycloak/keycloak/issues/40203

I'm also aware of the `src/test/java/org/keycloak/it/cli/dist` tests, however AFAIK they only allow verification of logs unless a `TestProvider` implementation is specified.


ryanemerson:
OLM failure not related https://github.com/keycloak/keycloak/issues/40099

ahus1 (approved):

PR #40215, opened 2025-06-03
Add credential_type to event in OTPFormAuthenticator and UsernamePasswordForm

### LINKED ISSUE #40118: credential_type is not set for password logins in LOGIN events
### Before reporting an issue

### Area

authentication

### Describe the bug

Keycloak does not populate `credential_type` in the LOGIN event for any password-based login – whether the account is local or federated via LDAP – and, furthermore, offers no detail to distinguish between an interactive password entry and an automatic login due to an existing session (cookie authenticator). By contrast, WebAuthnPasswordless logins correctly set `credential_type` to `webauthn-passwordless`, and it is also clear from the event details that those are fresh authentications. This lack of information makes it impossible to tell, when consuming user event streams (e.g., for audit logging or downstream processing), whether a user actively entered a password or simply continued an existing session.

### Version

2.6.5

### Regression

### Expected behavior

`credential_type` should be always populated in the `LOGIN` event.

### Actual behavior

.

### How to Reproduce?

.

### Anything else?

_No response_

jirutka:
Closes #40118

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/DefaultAuthenticationFlow.java
+++ b/services/src/main/java/org/keycloak/authentication/DefaultAuthenticationFlow.java
@@ -20,6 +20,7 @@ package org.keycloak.authentication;
 import org.jboss.logging.Logger;
 import org.keycloak.authentication.authenticators.conditional.ConditionalAuthenticator;
 import org.keycloak.authentication.authenticators.util.AuthenticatorUtils;
+import org.keycloak.events.Details;
 import org.keycloak.models.AuthenticationExecutionModel;
 import org.keycloak.models.AuthenticationFlowModel;
 import org.keycloak.models.Constants;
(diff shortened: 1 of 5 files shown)


ahus1:
Can you please update a test case to check that the details are present for the event? 

Maybe the LoginTest#loginSuccess can be amended. [Some of the tests](https://github.com/keycloak/keycloak/blob/0b30c872c8a5b6c224f75ef6e4e1c8be4005d6e3/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/LoginTest.java#L446) in that file already check for events. 

This test might be one where you can add a check for an event around OTP - or maybe even both, and then you don't need to update the first one. 

https://github.com/keycloak/keycloak/blob/c22f76867f595f0f7c327376a66a08dca01949e8/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/federation/storage/UserStorageOTPTest.java#L129

Thanks!

ahus1:
Added some ideas on how to make this more generic to pick up all credentials used for successfully logging in.  This needs more work, especially tests.

ahus1:
Closing for now as there was no response since June.

PR #40224, opened 2025-06-04
Sequential transactions instead of nested transactions

### LINKED ISSUE #40171: SQL error when logging in for first time (per user) after Keycloak upgrade
### Before reporting an issue

### Area

core

### Describe the bug

After upgrading from an older version to Keycloak 26.2.5, there is a step where the password is upgraded from the past format to argon2. 

When the user tries to login they get an "Internal server error" message and in the Keycloak logs I see:
Caused by: org.hibernate.StaleStateException: Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1; statement executed: update CREDENTIAL set CREATED_DATE=?,CREDENTIAL_DATA=?,PRIORITY=?,SALT=?,SECRET_DATA=?,TYPE=?,USER_ID=?,USER_LABEL=?,VERSION=? where ID=? and VERSION=?

If they retry the login again, they're logged in and the password is upgraded.

### Version

26.2.5

### Regression

### Expected behavior

I would expect the user to be logged in and the password upgrade to happen seamlessly. 

### Actual behavior

The first login attempt fails, the second login attempt succeeds. 

### How to Reproduce?

Upgrade from a much older version of Keycloak (using stepped upgrades between major changes like the move from Wildfly to Quarkus and then first Quarkus to latest Quarkus), and try to login. 

### Anything else?

_No response_

ahus1:
Closes #40171

### DIFF
--- a/services/src/main/java/org/keycloak/credential/PasswordCredentialProvider.java
+++ b/services/src/main/java/org/keycloak/credential/PasswordCredentialProvider.java
@@ -22,6 +22,7 @@ import io.micrometer.core.instrument.Tag;
 import org.jboss.logging.Logger;
 import org.keycloak.common.util.Time;
 import org.keycloak.credential.hash.PasswordHashProvider;
+import org.keycloak.models.AbstractKeycloakTransaction;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.ModelException;
 import org.keycloak.models.PasswordPolicy;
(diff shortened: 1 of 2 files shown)


sschu:
@ahus1 I wonder if this would be worth a backport to KC26? Could be really annoying for people going to that version.

ahus1:
@sschu - I assume you mean backporting to 26.2.x? That would be a yes. Older versions are not affected.

sschu:
@ahus1 Ah, ok. I thought they might be affected since Argon2 was introduced earlier. Then 26.2 would be great.

rmartinc (approved):
Thanks @ahus1 for your work on this! LGTM!

ahus1:
@mposolda - can you please approve and merge this one based on @rmartinc's review?

abstractj (approved):

PR #40234, opened 2025-06-04
Make the checkbox "Sign out from other devices" unchecked by default

rmartinc:
Closes #39975

Just changing the default to not checked. Tests should be tweaked to adapt to the new default. Adding a note in the upgrading guide together with the Alexander's previous note because they are related to the same `Sign out from other devices` checkbox.


mposolda (approved):
@rmartinc Thanks!

PR #40236, opened 2025-06-04
Feed data in export format to the created user

### LINKED ISSUE #40235: PasswordHashingTest#testPasswordRehashedWhenCredentialImportedWithDifferentKeySize fails to successfully log in
### Before reporting an issue

### Area

core

### Describe the bug

When working on #40171 I found that for this login the test fails.

### Version

main

### Regression

### Expected behavior

The login should succeed

### Actual behavior

It fails with a wrong password, but there is no check in the test

### How to Reproduce?

Add an appPage check

```
        loginPage.open();
        loginPage.login(username, password);
        appPage.assertCurrent();
```

### Anything else?

I'll prepare a PR. The problem is that not a plain text password should be passed when creating the user, but a full credential including the hashed password and the salt etc.

ahus1:
Closes #40235


ahus1:
@rmartinc - can you please review this one as IMHO it is related to the core-iam team?

rmartinc (approved):
Thanks @ahus1! I think this is for us the @keycloak/core-clients team. So it is OK.

LGTM! Previously the password set was really the encoded value in 256 and re-encoded with 512. So it was of the correct 512 size even the login was KO. Thanks for fixing this.

ahus1:
@mhajas - can you please approve based on @rmartinc's approval?

abstractj (approved):

Issue #40232, opened 2025-06-04
Setting of `type` of `Argon2PasswordHashProviderFactory` is incorrect,


RutgerLubbers:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

authentication

### Describe the bug

The `type` (`i`, `d` or `id`) is read from the configuration by the version key. This should be the type key.
In code:
```java
@Override
public void init(Config.Scope config) {
    version = config.get(VERSION_KEY, Argon2Parameters.DEFAULT_VERSION);
    type = config.get(VERSION_KEY, Argon2Parameters.DEFAULT_TYPE);
    ...
```
this should be 
```java
@Override
public void init(Config.Scope config) {
    version = config.get(VERSION_KEY, Argon2Parameters.DEFAULT_VERSION);
    type = config.get(TYPE_KEY, Argon2Parameters.DEFAULT_TYPE);
    ...
```

Further there is a typo in `cpuCoreSempahore`, this should be `cpuCoreSemaphore`. 

I can make a PR for both issues (well, one issue, one typo).

### Version

26.2 / main

### Regression

- [ ] The issue is a regression

### Expected behavior

The type should be read from the correct environment / configuration variable.

### Actual behavior

It is read from the wrong variable.

### How to Reproduce?

Try to configure the Argon2 type using the documented property, this will not work.

### Anything else?

I can make a PR for this.

ahus1:
@RutgerLubbers - as before, if you want this in the next 26.2.x patch release, please create a backport PR as usual.

PR #40239, opened 2025-06-04
Configure Argon2's `type` correctly in `Argon2PasswordHashProviderFactory`

RutgerLubbers:
Fix for https://github.com/keycloak/keycloak/issues/40232


mposolda (approved):
@RutgerLubbers Thanks for catching and fixing this!

PR #40256, opened 2025-06-04
Invalidate user cache entries when email or username are different from storage

### LINKED ISSUE #40085: Federated user IDs are not correctly evicted from cache
### Before reporting an issue

### Area

storage

### Describe the bug

we recently discovered a bug in Keycloak’s cache handling when caching users that are loaded from a user federation with a MAX_LIFESPAN setting. We've already briefly discussed this on the security mailing list and decided not to treat it as security related, and therefore we're opening it here.
 
We identified the problem as the following:
 
When a user is looked up in Keycloak via its username or email, Keycloak not only caches the user object, but also caches the ID of the user that matches the given username or email.
While the user object in the cache is correctly removed from the cache after the user federation’s MAX_LIFESPAN setting, the cached ID in the cached query is not evicted.

### Version

26.2.4

### Regression

### Expected behavior

1. The user ‘john.doe’ logs in to any client, causing the user and its ID to be cached
2. In an external system, the user’s username is changed to ‘jane.doe’
3. After the MAX_LIFESPAN period has expired, the user logs in again to a client using their old email ‘john.doe’
4. This login attempt should not succeed

### Actual behavior

1. The user ‘john.doe’ logs in to any client, causing the user and its ID to be cached
2. In an external system, the user’s username is changed to ‘jane.doe’
3. After the MAX_LIFESPAN period has expired, the user logs in again to a client using their old email ‘john.doe’
4. This login attempt *succeeds*. Any tokens (or userinfo request) produced from that new login session will however contain the *new* username ‘jane.doe’, despite the user logging in successfully as ‘john.doe’
5. Generally, the cached query will eventually disappear in a high load system by being displaced by other entries, however in a low load system, we have observed these cache entries lingering for hours or even days
6. This behaviour is the same, regardless of the “email as username” or other login settings in the keycloak realm

### How to Reproduce?

I've build a small reproducer for the problem which describes the bug in detail and allows easy reproduction
https://github.com/sonOfRa/keycloak-cache-invalidation

### Anything else?

In our installation, we're currently running the following patch, which fixes the problem. However we're not sure if this is the desired solution for upstream, and would like to discuss possible alternatives.

```patch
diff --git a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserCacheSession.java b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserCacheSession.java
index 86475fa81a..6b673a1c7f 100755
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserCacheSession.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserCacheSession.java
@@ -297,8 +297,14 @@ public class UserCacheSession implements UserCache, OnCreateComponent, OnUpdateC
                 return getDelegate().getUserByUsername(realm, username);
 
             }
-            logger.trace("return getUserById");
-            return getUserById(realm, userId);
+
+            UserModel cachedUser = getUserById(realm, userId);
+            if (cachedUser.getUsername().equals(username)) {
+                logger.trace("return getUserById");
+                return cachedUser;
+            }
+            cache.invalidateObject(cacheKey);
+            return getDelegate().getUserByUsername(realm, username);
         }
     }
 
@@ -435,7 +441,12 @@ public class UserCacheSession implements UserCache, OnCreateComponent, OnUpdateC
                 return getDelegate().getUserByEmail(realm, email);
 
             }
-            return getUserById(realm, userId);
+            UserModel cachedUser = getUserById(realm, userId);
+            if (cachedUser.getEmail().equals(email)) {
+                return cachedUser;
+            }
+            cache.invalidateObject(cacheKey);
+            return getDelegate().getUserByEmail(realm, email);
         }
     }
 
```

What this patch does is simply check if the email or username for the cached ID still matches the user fetched from the federation, and if it does not, invalidate the cache entry and re-fetch it by username or email.
Note that we haven't tried this patch against 26.2.4, we're running it against 25.0.6 in our environment, so it might not apply cleanly on newer versions, but the concept could still be applied.

pedroigor:
Closes #40085

* Ensures that entries returned from the cache when querying by username or email continue to reference the up-to-date value from the underlying storage. This is the case when the user storage provider has a cache policy to expire entries eventually, or that disables caching of their federated users.
* Simplifies how cache policies that rely on a `MAX_LIFESPAN` by leveraging the existing mechanisms on Infinispan, therefore removing redundant code on the Keycloak side.

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
@@ -422,7 +422,11 @@ public class UserStorageManager extends AbstractStorageManager<UserStorageProvid
     public UserModel getUserByUsername(RealmModel realm, String username) {
         UserModel user = localStorage().getUserByUsername(realm, username);
         if (user != null) {
-            return importValidation(realm, user);
+            user = importValidation(realm, user);
+            // Case when username was changed directly in the userStorage and doesn't correspond anymore to the username from local DB
+            if (user != null && username.equalsIgnoreCase(user.getUsername())) {
+                return user;
+            }
         }
 
         return mapEnabledStorageProvidersWithTimeout(realm, UserLookupProvider.class,
--- a/model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
@@ -29,6 +29,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -113,7 +115,14 @@ public class UserStorageManager extends AbstractStorageManager<UserStorageProvid
      * @param user
      * @return
      */
-    protected UserModel importValidation(RealmModel realm, UserModel user) {
+    protected UserModel validateUser(RealmModel realm, UserModel user) {
+        if (user == null) {
+            return null;
+        }
+
+        if (user.isFederated()) {
+            user = validateFederatedUser(realm, user);
+        }
 
         if (isReadOnlyOrganizationMember(user)) {
             if (user instanceof CachedUserModel cachedUserModel) {
@@ -122,14 +131,17 @@ public class UserStorageManager extends AbstractStorageManager<UserStorageProvid
             return new ReadOnlyUserModelDelegate(user, false);
         }
 
-        if (user == null || !user.isFederated()) return user;
+        return user;
+    }
 
-        UserStorageProviderModel model = getStorageProviderModel(realm, user.getFederationLink());
+    private UserModel validateFederatedUser(RealmModel realm, UserModel user) {
+        if (!user.isFederated()) {
+            return user;
+        }
+
+        UserStorageProviderModel model = getUserStorageProviderModel(realm, user);
+
         if (model == null) {
-            // remove linked user with unknown storage provider.
-            logger.debugf("Removed user with federation link of unknown storage provider '%s'", user.getUsername());
-            deleteInvalidUserCache(realm, user);
-            deleteInvalidUser(realm, user);
             return null;
         }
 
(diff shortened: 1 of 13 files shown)


ahus1:
I pushed some changes that should provide proper eviction of the different user related information that honors the caching defined in the provider. 

I tested this manually with LDAP switch to non-importing mode.

Also, I think the checking when something was cached is obsolete and can be removed, as either the ISPN behavior was fixed, or just misunderstood when this was written. IMHO there is no need check the `cacheTimestamp` again.

pedroigor:
> Also, I think the checking when something was cached is obsolete and can be removed, as either the ISPN behavior was fixed, or just misunderstood when this was written. IMHO there is no need check the `cacheTimestamp` again.

I also stumbled upon the assumption that ISPN could not guarantee eviction of entries, and it would be great if it were redundant. I'll look at your changes and the test failures.

ahus1:
@pedroigor - maybe removing the redundant code could be deferred to 26.3, while fixing the current problems at hand could be backported to 26.2. Let's discuss more of this tomorrow in the meeting I scheduled.

pedroigor:
> @pedroigor - maybe removing the redundant code could be deferred to 26.3, while fixing the current problems at hand could be backported to 26.2. Let's discuss more of this tomorrow in the meeting I scheduled.

Works for me.

Looks like we can not rely on the entry lifespan on ISPN, and the assumption was correct, at least regarding testing it. There are test failures that are related to skipping the code in `validateCache` based on the lifespan set to the user storage provider cache policy.

I think the failure is due to the fact that calling `setTimeOffset` during tests won't change the time in ISPN, and the eviction scheduler/logic won't remove the entry.

pedroigor:
After talking with @pruivo, I updated the tests to enable/disable the ISPN time service to make sure entries are evicted in ISPN as expected when changing the time offset during tests.

There are failures in tests with a weird assumption that realm resources instances (e.g: `UserModel`) from a `KeycloakSession` will eventually return the adapter instance from the user storage during searches. AFAIK, it should always expect an adapter from the cache layer or null?

ahus1:
> There are failures in tests with a weird assumption that realm resources instances (e.g: `UserModel`) from a `KeycloakSession` will eventually return the adapter instance from the user storage during searches. AFAIK, it should always expect an adapter from the cache layer or null?

I've seen some cases where caching was disabled completely, then there wouldn't be an adapter returned. Let's have a look later today to see if this is the case in your setup.

ahus1:
@pedroigor - I fixed the tests, and then restored the original logic as much as possible to avoid breaking any existing functionality.

When testing this, I found problems with `dailyTimeout()` as it would return the daily time was before the current time, which is IMHO a algorithm, and not in line with `dailyEvictionBoundary()`. 

Once the tests are green, I would suggest to backport the PR ti 26.2.x. 

For 26.3, I would like to deprecate `CacheableStoreageProviderModel#shouldInvalidate` and also deprecate the methods `RealmCacheSession#validateCache` and `UserCacheSession#validateCache`. Possible together with `CachedModel#getCacheTimestamp` and `CachedObject#getCacheTimestamp`.

ahus1:
@pedroigor - I think I fixed the one test that remained failing, please review my latest commit.

model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
```diff
@@ -422,7 +422,11 @@ public UserModel getUserById(RealmModel realm, String id) {
     public UserModel getUserByUsername(RealmModel realm, String username) {
         UserModel user = localStorage().getUserByUsername(realm, username);
         if (user != null) {
-            return importValidation(realm, user);
+            user = importValidation(realm, user);
+            // Case when username was changed directly in the userStorage and doesn't correspond anymore to the username from local DB
+            if (user != null && username.equalsIgnoreCase(user.getUsername())) {
```
pedroigor: Looks like we assume this check is done by https://www.keycloak.org/docs/26.2.5/server_development/#importeduservalidation-interface. 

The `importValidation` will return the same model instance if your provider is not implementing it. In this case, you will eventually return a stale model, potentially different than the username from the provider.

But I guess our recommendation is to implement the `ImportedUserValidation` interface for user storage providers if the underlying store is writable and you want consistency?

testsuite/model/src/test/java/org/keycloak/testsuite/model/user/UserSyncTest.java
```diff
@@ -270,17 +271,15 @@ public void testInvalidUsersAreDeleted() {
...
+            return user1.getId();
         });
 
-        // validate imported user
+        // validate imported user, user will be deleted and re-created
         withRealm(realmId, (session, realm) -> {
-            assertThat(session.users().getUserByUsername(realm, "user1"), is(nullValue()));;
-            UserModel deletedUser = UserStoragePrivateUtil.userLocalStorage(session).getUserByUsername(realm, "user1");
-            assertThat(deletedUser, is(nullValue()));
+            assertThat(session.users().getUserByUsername(realm, "user1"), not(equalTo(oldUserId)));
```
pedroigor:
```suggestion
            assertThat(session.users().getUserByUsername(realm, "user1").getId(), not(equalTo(oldUserId)));
```

This assertion should fail if the user storage provider is not implementing the `ImportedUserValidation`.

model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
ahus1: The change your are annotating is a change you did :D ... I'm good with this change as it reads to me as follows: Validate the given user. If the user doesn't exist any more (null returned), or if the username that is part of the search has changed (second part of the if), then do not return directly, but try all the other stores to find a (new) possible match. 

UPDATE: I think this is better than before as returning null from an invalidation and not trying to re-do the lookup is worse.

testsuite/model/src/test/java/org/keycloak/testsuite/model/user/UserSyncTest.java
ahus1: Yes, but for this test we know it implements it, and we know that it will invalidate the existing user (by changing the ID), and we want to see that it is re-created.

model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
pedroigor: I'm not sure about this change now :)

The federated user is linked with a specific provider, and in this case, the `importValidation` returning null indicates the provider no longer recognizes the user, and the account will be deleted from the local database. There is no need to run the lookup. The account will be eventually imported from a different storage, if the case.

If the provider is a `ImportedUserValidation` and the username does not match the value of the search, it is perhaps safer to also remove the user from the local database? Or we just remove this check and rely on the provider to run this check as suggested by the documentation?

I'll review this code.

ahus1 (approved):
Thank you for this extensive change, and all the added tests. I reviewed the tests, and also did a manual test locally where the username of a user changed several times. Appoving.

PR #40258, opened 2025-06-04
case insensitive match on organization identity provider domain

### LINKED ISSUE #40253: Case sensitive Organization/IDP linking on domain
### Before reporting an issue

### Area

organizations

### Describe the bug

If you link an Identity Provider to an organization, a user is unable to login if they do not enter the domain in all lowercase. 

When configuring domains for an organization, even when entered as `TEST.COM`, it saves as `test.com`

Image

If a user tries to sign in with domain `TEST.COM` they get this error:

Image

but if you enter the domain as lowercase, the user login flow is handled correctly and they are redirected to the IDP.

### Version

26.2.5

### Regression

### Expected behavior

User should be able to enter their username/email address when signing in a case insensitive manor.

### Actual behavior

User my enter domain as strictly lowercase

### How to Reproduce?

Setup an organization and link to a particular Identity Provider. Try logging in with the domain lowercase and uppercase.

### Anything else?

_No response_

ryan-morris:
Closes #40253 

Currently when matching on Identity Providers for an Organization, if a user enters the domain part of their email when trying to login as anything but lowercase, the specified IDP flow is unavailable.

### DIFF
--- a/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
+++ b/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
@@ -259,10 +259,10 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
             return false;
         }
 
-        // first look for an IDP that matches exactly the specified domain
+        // first look for an IDP that matches exactly the specified domain (case-insensitive)
         IdentityProviderModel idp = organization.getIdentityProviders()
                 .filter(broker -> IdentityProviderRedirectMode.EMAIL_MATCH.isSet(broker) &&
-                    domain.equals(broker.getConfig().get(OrganizationModel.ORGANIZATION_DOMAIN_ATTRIBUTE))).findFirst().orElse(null);
+                    domain.equalsIgnoreCase(broker.getConfig().get(OrganizationModel.ORGANIZATION_DOMAIN_ATTRIBUTE))).findFirst().orElse(null);
 
         if (idp != null) {
             // redirect the user using the broker that matches the specified domain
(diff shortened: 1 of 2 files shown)


ahus1:
@ryan-morris - this PR looks good at first sight. Can you please also add a test case? I assume that https://github.com/keycloak/keycloak/blob/a39a64e4d19258b7c43fb2c65011d719c01958dc/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/authentication/OrganizationAuthenticationTest.java#L49 might be the right file to add another test. 

cc: @sguilhen - maybe you want to take over this one for reviewing?

ahus1:
Please don't worry about the OLM test for now. Unfortunately it is currently unstable, see #40099

martin-kanis (approved):
@ryan-morris Thanks for the fix. Looks good to me.

If you want, you can squash the commits into the single one. Alternatively, we can do it during merging.

ahus1:
> Alternatively, we can do it during merging.

I'll take care of the squash, no need to run the build again for that.

ahus1 (approved):

sapirnat:
Hey great fix. Thanks.

One more thing.. The issue is fixed for a case where a specific domain is chosen but need to add the same fix for an ANY scenario.

Need to add this to this line as well: 

https://github.com/keycloak/keycloak/blob/26.4.0/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java#L304

` 
idp = organization.getIdentityProviders().filter(IdentityProviderRedirectMode.EMAIL_MATCH::isSet)
                .filter(broker -> ANY_DOMAIN.equals(broker.getConfig().get(OrganizationModel.ORGANIZATION_DOMAIN_ATTRIBUTE)))
                .filter(broker -> organization.getDomains().map(OrganizationDomainModel::getName).anyMatch(domain::equals))
                .findFirst().orElse(null);
`

sguilhen:
@sapirn2 Can you raise a separate issue for that?

sapirnat:
Thanks! Opened #43613

PR #40265, opened 2025-06-05
Fix: Inconsistent Results between /users and /users/count Endpoints

Barathwaja:
Below are the list of changes done:
1. `/users` endpoint:
-  Added support for filtering by emailVerified.

2. `/users/count` endpoint:
- Updated Javadoc comments for clarity.
- Added support for filtering by enabled, emailVerified, idpAlias, idpUserId, and exact to align with the /users endpoint.

Closes: [38556](https://github.com/keycloak/keycloak/issues/38556), [29295](https://github.com/keycloak/keycloak/issues/29295)

### DIFF
--- a/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/UsersResource.java
+++ b/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/UsersResource.java
@@ -250,6 +250,29 @@ public interface UsersResource {
       @QueryParam("first") Integer firstResult,
       @QueryParam("max") Integer maxResults);
 
+    /**
+     * Returns the users that can be viewed and match the given filters.
+     *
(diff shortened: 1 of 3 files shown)


pedroigor:
Thanks, @Barathwaja. Added some tests and signed the commit for you.

There is still a small overlap between the search and count methods depending on the code path, but we can sort it out in a separate PR.

pedroigor (approved):

Barathwaja:
> Thanks, @Barathwaja. Added some tests and signed the commit for you.
> 
> There is still a small overlap between the search and count methods depending on the code path, but we can sort it out in a separate PR.

@pedroigor - Sounds good. Share your observations when you get a chance, and I’ll take a look.

PR #39805, opened 2025-05-19
Move UserTest.java to the new testsuite

lhanusov:
Part of: #34494

PR #40269, opened 2025-06-05
Migrate FineGrainAdminUnitTest.java to the new testsuite

lhanusov:
Part of: #34494


tests/base/src/test/java/org/keycloak/tests/admin/finegrainedadminv1/FineGrainedAdminRestTest.java
```diff
@@ -0,0 +1,323 @@
...
+                Assertions.assertEquals(1, queryUsers.size());
+
+                UserRepresentation newGroupMemberWithoutGroup = UserConfigBuilder.create()
+                        .username("new-group-member").email("new-group-member@keycloak.org").name("New", "Member").build();
+                try {
+                    ApiUtil.createUserWithAdminClient(realmClient.realm(REALM_NAME), newGroupMemberWithoutGroup);
+                    Assertions.fail("should fail with HTTP response code 403 Forbidden");
+                } catch (WebApplicationException e) {
+                    Assertions.assertEquals(403, e.getResponse().getStatus());
+                }
```
vaceksimon: This and the following three statements are similar to [this situation in the UserTest PR](https://github.com/keycloak/keycloak/pull/39805/#discussion_r2164178313).

We can manually create a user via the admin API and check the response for the error code instead of [changing the ApiUtil](https://github.com/keycloak/keycloak/pull/40269/files#diff-3b69de162f6ecd7e32dc5de4722838304ab43d8e475e76cc8d71c8eab86ddef6)

vaceksimon (changes requested):

tests/base/src/test/java/org/keycloak/tests/admin/finegrainedadminv1/FineGrainedAdminRestTest.java
lhanusov: @vaceksimon thanks, updated :-)

tests/base/src/test/java/org/keycloak/tests/admin/finegrainedadminv1/FineGrainedAdminRestTest.java
```diff
@@ -0,0 +1,336 @@
...
+                        .username("new-group-member").email("new-group-member@keycloak.org").name("New", "Member").build();
+                try {
+                    Response response = realmClient.realm(REALM_NAME).users().create(newGroupMemberWithoutGroup);
+                    if (response.getStatus() == 403) {
+                        throw new WebApplicationException(response);
+                    }
+                    Assertions.fail("should fail with HTTP response code 403 Forbidden");
+                } catch (WebApplicationException e) {
+                    Assertions.assertEquals(403, e.getResponse().getStatus());
+                }
```
vaceksimon: I don't think it should be necessary to throw an exception here
```suggestion
                Response response = realmClient.realm(REALM_NAME).users().create(newGroupMemberWithoutGroup);
                Assertions.assertEquals(403, response.getStatus());
                response.close();
```

vaceksimon (changes requested):

tests/base/src/test/java/org/keycloak/tests/admin/finegrainedadminv1/FineGrainedAdminRestTest.java
lhanusov: @vaceksimon these changes are necessary since the REST create method returns response and doesn't throw anything

vaceksimon (approved):

tests/base/src/test/java/org/keycloak/tests/admin/finegrainedadminv1/FineGrainedAdminRestTest.java
lhanusov: thanks, updated, based on discussion -> checking response status without throwing anything

stianst (approved):

PR #40272, opened 2025-06-05
Add webauthn and recovery codes to the default browser flow as disabled

### LINKED ISSUE #39999: Make changes for the browser flow

rmartinc:
Closes #39999

Adding webauthn and recovery codes in the conditional OTP flow as disabled. PR adds the normal webauthn (not passwordless), as I suppose the passwordless is going to be integrated using the passkeys feature and passwordless is more a primary than a 2FA (let me know if I'm wrong). I have not detected important issues except that now to force OTP as required you need to modify two things (flow from conditional to required, otp authenticator from alternative to required). I maintained the `requiredCredentials` differentiation although I think this is not used anymore.

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/DefaultAuthenticationFlows.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/DefaultAuthenticationFlows.java
@@ -377,9 +377,30 @@ public class DefaultAuthenticationFlows {
         // otp processing
         execution = new AuthenticationExecutionModel();
         execution.setParentFlow(conditionalOTP.getId());
-        execution.setRequirement(AuthenticationExecutionModel.Requirement.REQUIRED);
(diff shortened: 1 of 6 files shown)


rmartinc:
One more thing, I doubted about changing the step name `Browser - Conditional OTP` to `Browser - Conditional 2FA` but finally I didn't do it because maybe it means a lot of changes. But we can give it a try if you think it is necessary.

mposolda:
@rmartinc Thanks! I think it will be definitely good to rename the flow to `Browser - Conditional 2FA` as now that subflow is not only OTP. I've created separate issue https://github.com/keycloak/keycloak/issues/40281 , so we can continue with that as a separate issue.

mposolda (approved):

Eng-Fouad:
I think this PR caused this regression: #40954

PR #40279, opened 2025-06-05
Move PartialImportTest.java to the new testsuite

lhanusov:
Part of: #34494


tests/base/src/test/java/org/keycloak/tests/admin/partialimport/AbstractPartialImportTest.java
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    @BeforeEach
+    public void init() {
+        piRep = new PartialImportRepresentation();
+    }
+
+    protected ClientResource clientRolesClient() {
+        return ApiUtil.findClientResourceByName(managedRealm.admin(), CLIENT_ROLES_CLIENT);
+    }
```
vaceksimon: This method can be replaced with `rolesClient.admin()`

tests/base/src/test/java/org/keycloak/tests/admin/partialimport/PartialImportClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+        piRep.setClients(Arrays.asList(client, client2));
+
+        PartialImportResults result = doImport();
+        Assertions.assertEquals(0, result.getOverwritten());
+    }
+
+    @Test
+    public void testOverwriteExistingClientWithServiceAccount() {
+        setOverwrite();
+        piRep.setClients(Collections.singletonList(managedRealm.admin().clients().findByClientId(CLIENT_SERVICE_ACCOUNT).get(0)));
```
vaceksimon:
```suggestion
        piRep.setClients(Collections.singletonList(serviceClient.admin().toRepresentation()));
```

tests/base/src/test/java/org/keycloak/tests/admin/partialimport/PartialImportClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+        Assertions.assertEquals(1, doImport().getOverwritten());
+
+        ClientRepresentation client = managedRealm.admin().clients().findByClientId(CLIENT_SERVICE_ACCOUNT).get(0);
+        managedRealm.admin().clients().get(client.getId()).getServiceAccountUser();
```
vaceksimon:
```suggestion
        serviceClient.admin().getServiceAccountUser();
```

vaceksimon: Looking at this, I realized how little we use `assertDoesNotThrow`.  Or actually only twice in all the tests. Why is it not used? Wouldn't it make it clearer what should and shouldn't happen?

vaceksimon (changes requested):

tests/base/src/test/java/org/keycloak/tests/admin/partialimport/PartialImportClientTest.java
lhanusov: I've used some of your suggestions: assertDoesNotThrow and line 181, line 186 cannot be used, because it's a corner case, where client has a new id and doesn't correspond to the InjectClient one anymore.

vaceksimon (approved):

stianst (dismissed):

stianst (approved):

PR #40283, opened 2025-06-05
Add clustering tests to new test framework

### LINKED ISSUE #39962: Create a POC of running 2 containers in the new testsuite

mhajas:
Closes #39962

This is a replacement PR for https://github.com/keycloak/keycloak/pull/40127

I am sorry for confusion, I didn't realize I won't be able to push to that PR if I open it from Pedro's fork.

### DIFF
--- a/quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
+++ b/quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
@@ -2,6 +2,7 @@ package org.keycloak.it.utils;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.exception.NotFoundException;
+import io.quarkus.bootstrap.utils.BuildToolHelper;
 import io.restassured.RestAssured;
 import org.jboss.logging.Logger;
 import org.keycloak.common.Version;
@@ -16,8 +17,10 @@ import org.testcontainers.images.RemoteDockerImage;
 import org.testcontainers.images.builder.ImageFromDockerfile;
 import org.testcontainers.utility.DockerImageName;
 import org.testcontainers.utility.LazyFuture;
+import org.testcontainers.utility.MountableFile;
 
-import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.List;
@@ -33,9 +36,16 @@ public final class DockerKeycloakDistribution implements KeycloakDistribution {
 
         final ToStringConsumer stdOut = new ToStringConsumer();
         final ToStringConsumer stdErr = new ToStringConsumer();
+        final Consumer<OutputFrame> customLogConsumer;
+        public BackupConsumer(Consumer<OutputFrame> customLogConsumer) {
+            this.customLogConsumer = customLogConsumer;
+        }
 
         @Override
         public void accept(OutputFrame t) {
+            if (customLogConsumer != null) {
+                customLogConsumer.accept(t);
+            }
             if (t.getType() == OutputType.STDERR) {
                 stdErr.accept(t);
             } else if (t.getType() == OutputType.STDOUT) {
@@ -55,20 +65,27 @@ public final class DockerKeycloakDistribution implements KeycloakDistribution {
 
     private String stdout = "";
     private String stderr = "";
-    private BackupConsumer backupConsumer = new BackupConsumer();
-    private final File dockerScriptFile = new File("../../container/ubi-null.sh");
-
+    private BackupConsumer backupConsumer;
+    private Consumer<OutputFrame> customLogConsumer;
     private GenericContainer<?> keycloakContainer = null;
     private String containerId = null;
 
     private final Executor parallelReaperExecutor = Executors.newSingleThreadExecutor();
     private final Map<String, String> envVars = new HashMap<>();
+    private final LazyFuture<String> image;
+
+    private final Map<MountableFile, String> copyToContainer = new HashMap<>();
 
     public DockerKeycloakDistribution(boolean debug, boolean manualStop, int requestPort, int[] exposedPorts) {
+        this(debug, manualStop, requestPort, exposedPorts, null);
+    }
+
+    public DockerKeycloakDistribution(boolean debug, boolean manualStop, int requestPort, int[] exposedPorts, LazyFuture<String> image) {
         this.debug = debug;
         this.manualStop = manualStop;
         this.requestPort = requestPort;
         this.exposedPorts = IntStream.of(exposedPorts).boxed().toArray(Integer[]::new);
+        this.image = image == null ? createImage(false) : image;
     }
 
     @Override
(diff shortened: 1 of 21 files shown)


mhajas:
This Pr addresses all comments from @stianst in https://github.com/keycloak/keycloak/pull/40127

tests/base/pom.xml
```diff
@@ -131,4 +136,16 @@
...
+            <id>mixed-cluster</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <kc.test.server>cluster</kc.test.server>
+            </properties>
+        </profile>
+    </profiles>
+
```
stianst: Guessing this is a left-over from when the tests where in `tests/base` and can be removed now?

```suggestion
```

If it's intentional and you want to be able to run the regular tests with a mixed cluster that would probably require the clustering to support a reverse-proxy, and you'd just run it with:

```
KC_TEST_SERVER=cluster mvn test
```

One of the main aims of the new testsuite is to don't have a ton of Maven profiles for things, but just use env variables instead.

tests/base/pom.xml
```diff
@@ -88,6 +88,11 @@
             <groupId>org.keycloak.testframework</groupId>
             <artifactId>keycloak-test-framework-remote</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.keycloak.testframework</groupId>
+            <artifactId>keycloak-test-framework-clustering</artifactId>
+            <version>999.0.0-SNAPSHOT</version>
+        </dependency>
```
stianst: Did you intend to add this here?

test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractContainerTestDatabase.java
```diff
@@ -12,11 +12,13 @@
 public abstract class AbstractContainerTestDatabase implements TestDatabase {
 
     protected boolean reuse;
+    protected boolean internal;
 
     protected JdbcDatabaseContainer<?> container;
 
     public AbstractContainerTestDatabase() {
         reuse = Config.getValueTypeConfig(TestDatabase.class, "reuse", false, Boolean.class);
+        internal = Config.getValueTypeConfig(TestDatabase.class, "internal", false, Boolean.class);
```
stianst: I'd do this automatically somehow rather than require configuring it. Maybe add `getJdbcUrl(boolean remote)` or something then update the KeycloakServer to fetch the right one depending on running in container or not.

test-framework/clustering/src/main/java/org/keycloak/testframework/clustering/LoadBalancerSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public LoadBalancer getValue(InstanceContext<LoadBalancer, InjectLoadBalancer> instanceContext) {
+        KeycloakServer server = instanceContext.getDependency(KeycloakServer.class);
+
+        if (!(server instanceof ClusteredKeycloakServer)) {
+            throw new IllegalStateException("Load balancer can only be used with ClusteredKeycloakServer");
+        }
+
+        return new LoadBalancer((ClusteredKeycloakServer) server);
```
stianst: Nitpick, but I like this approach (disclaimer: suggestion written in GitHub UI, so may not compile):
```suggestion
        if (server instanceof ClusteredKeycloakServer clusteredServer) {
            return new LoadBalancer(clusteredServer);
        } else {
             throw new IllegalStateException("Load balancer can only be used with ClusteredKeycloakServer");        
        }
```

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
stianst, line 1: Haven't reviewed changes to this file, perhaps ping someone from the @keycloak/cloud-native team to take a look?

tests/base/pom.xml
mhajas: Sorry, I completely missed this file during cleanup

tests/base/pom.xml
mhajas: This was not intentional. Pedro added it during POC and I missed this was there

test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractContainerTestDatabase.java
mhajas: Done. I used `Config.getSelectedSupplier(KeycloakServer.class)` and then used `internal` it when `cluster` is used. See https://github.com/keycloak/keycloak/pull/40283/files#diff-b2bf46ba37f5aa78f7d9a5943737f8dc80fda02931bf7714aa9846145d2920daR42-R52 


I was thinking about adding it as a method on `AbstractKeycloakServerSupplier` so each KeycloakServer supplier can configure whether it runs in container or not. But I needed to add a new method to Registry that can provide Supplier instances for given class. It then looked something like this:
```
        Supplier<KeycloakServer, ?> kcSupplier = instanceContext.getRegistry().findSupplierByType(KeycloakServer.class);
        TestDatabase database = instanceContext.getValue();

        // If both KeycloakServer and TestDatabase run in container, we need to configure Keycloak with internal
        // url that is accessible within docker network
        if (database instanceof AbstractContainerTestDatabase containerDatabase
            && kcSupplier instanceof AbstractKeycloakServerSupplier abstractKcSupplier
            && abstractKcSupplier.runsInContainer()
                ) {
            return serverConfig.options(containerDatabase.serverConfig(true));
        }

        return serverConfig.options(database.serverConfig());
```

Let me know what you prefer. To me it seems this second approach is complicating it too much.

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
mhajas: Yes I will make sure we have review from them before merging.

Pepo48: I'll take a look.

mhajas: @Pepo48, there are some failures in CI related to changes in `DockerKeycloakDistribution` we will look into it.

mhajas:
@Pepo48 I believe this PR is ready for review. The failed windows tests should not be related to this PR. 

I would like to squash commits and make it ready for review. Do you prefer to review the changes now with commits history or I can do it now?

Pepo48:
@mhajas feel free to squash it first. Thanks!

mhajas:
Thanks @Pepo48! I rebased, squashed and CI is now green.

Just to make sure, we don't need a review for the test-framework changes from you. We need a cloud-native eye on the `DockerKeycloakDistribution` changes

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
```diff
@@ -292,4 +341,8 @@ public void clearEnv() {
         this.envVars.clear();
     }
 
+    public int getMappedPort(int port) {
+        return keycloakContainer.getMappedPort(port);
```
Pepo48, line 349: Does it make sense to add a null check? I can imagine a scenario, where KeycloakContainer could be null and subsequently getMappedPort could be called.

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
```diff
@@ -158,6 +194,19 @@ public void setRequestPort(int port) {
         }
     }
 
+    public void copyProvider(String groupId, String artifactId) {
```
Pepo48, line 197: and possibly even here, although is probably less likely to happen:
```
if (providerPath == null) {
        throw new RuntimeException("Failed to resolve provider artifact: " + groupId + ":" + artifactId);
    }
```

Pepo48 (dismissed):
thanks @mhajas for the improvements. I found nothing that would block the PR in it's current stated.

I suggested somewhat cosmetic NPE checks to add, but I'll leave it up to your consideration.

During the review I spotted two unrelated issues to the changes in this PR (and indeed cosmetic) in the class:
* on the line no. 226 I guess we can use logical OR instead of bitwise OR
* on the line no. 276 there is a typo in the word "schecdule"

We can either fix it right away or address it separately in a follow up.

One way or another I'm approving the PR. Thanks again!

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/DockerKeycloakDistribution.java
mhajas: Looking at `resolveArtifact` method it is not possible the result will be `null` (famous last words :D :D) Anyway, I would skip the suggestion. I would say the main point here is to detect if something goes wrong so we don't falsely think the provider was correctly copied while it was not. Given checks in `resolveArtifact` and calling `Files.isRegularFile` we should end up with an exception if the file was not found so I believe this is good enough. WDYT?

mhajas:
@Pepo48 Thank you for the review. I addressed your comments except the one in `copyProvider` method. Please approve if you agree with my reasoning.

Pepo48 (approved):

ahus1 (approved):

PR #40286, opened 2025-06-05
Change to useRealm all the policies tabs

rmartinc:
Closes #40284

Currently the policy tabs are using fetch/update over the realm. But it seems that other parts are using another `useRealm` call that saves the realm. So the realm is not updated in the `useRealm` and another part that uses it is going to revert the changes to the previous ones. I didn't think that the policy was reverted, so I spent long time to realize it was not an error in my new code.


edewit (approved):
If we solve it like this we could even just use the the useRealm hook in these components instead of passing them in.

rmartinc:
@ssilvert If you prefer that I remove the `realm` from the file and add the `useRealm` in the children components just let me know. But I think that it's OK like it is now. :smile:

ssilvert (approved):
This is ok.

PR #40287, opened 2025-06-05
Access Token IDs have less than 128 bits of entropy

douglaspalmer:
Closes #38663

### DIFF
--- a/adapters/saml/core/src/main/java/org/keycloak/adapters/saml/rotation/SamlDescriptorPublicKeyLocator.java
+++ b/adapters/saml/core/src/main/java/org/keycloak/adapters/saml/rotation/SamlDescriptorPublicKeyLocator.java
@@ -24,7 +24,6 @@ import java.security.cert.X509Certificate;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.security.auth.x500.X500Principal;
 import javax.xml.crypto.dsig.keyinfo.KeyInfo;
@@ -34,6 +33,7 @@ import org.jboss.logging.Logger;
 import org.keycloak.adapters.cloned.HttpAdapterUtils;
 import org.keycloak.adapters.cloned.HttpClientAdapterException;
 import org.keycloak.common.util.MultivaluedHashMap;
+import org.keycloak.common.util.SecretGenerator;
 import org.keycloak.common.util.Time;
 import org.keycloak.dom.saml.v2.metadata.KeyTypes;
 import org.keycloak.rotation.KeyLocator;
@@ -179,7 +179,7 @@ public class SamlDescriptorPublicKeyLocator implements KeyLocator {
                 this.publicKeyCacheByKey.put(new KeyHash(x509certificate.getPublicKey()), x509certificate.getPublicKey());
             } else {
                 final X500Principal principal = x509certificate.getSubjectX500Principal();
-                String name = (principal == null ? "unnamed" : principal.getName()) + "@" + x509certificate.getSerialNumber() + "$" + UUID.randomUUID();
+                String name = (principal == null ? "unnamed" : principal.getName()) + "@" + x509certificate.getSerialNumber() + "$" + SecretGenerator.getInstance().generateSecureID();
                 this.publicKeyCacheByName.put(name, x509certificate.getPublicKey());
                 this.publicKeyCacheByKey.put(new KeyHash(x509certificate.getPublicKey()), x509certificate.getPublicKey());
                 LOG.tracef("Adding certificate %s without a specific key name: %s", name, x509certificate);
--- a/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
+++ b/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
@@ -30,6 +30,15 @@ public class SecretGenerator {
     public static SecretGenerator getInstance() {
         return instance;
     }
+    
+    public String generateSecureID() {
+        StringBuilder builder = new StringBuilder(instance.randomBytesHex(16));
+        builder.insert(8, '-');
+        builder.insert(13, '-');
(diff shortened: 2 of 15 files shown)


adapters/saml/core/src/main/java/org/keycloak/adapters/saml/rotation/SamlDescriptorPublicKeyLocator.java
```diff
@@ -179,7 +180,10 @@ private synchronized <T> Key refreshCertificateCacheAndGet(T key, Map<T, Key> ca
                 this.publicKeyCacheByKey.put(new KeyHash(x509certificate.getPublicKey()), x509certificate.getPublicKey());
             } else {
                 final X500Principal principal = x509certificate.getSubjectX500Principal();
-                String name = (principal == null ? "unnamed" : principal.getName()) + "@" + x509certificate.getSerialNumber() + "$" + UUID.randomUUID();
+                SecureRandom random = new SecureRandom();
+                byte[] bytes = new byte[16];
+                random.nextBytes(bytes);
```
rmartinc: Would not be better to use `SecretGenerator.randomBytes(16)`? It's in the common package so I think you can use in any other package.

adapters/saml/core/src/main/java/org/keycloak/adapters/saml/rotation/SamlDescriptorPublicKeyLocator.java
```diff
@@ -179,7 +180,10 @@ private synchronized <T> Key refreshCertificateCacheAndGet(T key, Map<T, Key> ca
...
+                String name = (principal == null ? "unnamed" : principal.getName()) + "@" + x509certificate.getSerialNumber() + "$" + Arrays.toString(bytes);
```
rmartinc: Using `Arrays.toString` would create something like `[56, -120,...]` which is a horrible id IMHO. Why don't you create a new utility method in the `SecretGenerator` class to generate the random array and return it in hexa?

```java
    public static String randomBytesHexa(int length) {
        final StringBuilder sb = new StringBuilder();
        for (byte b : randomBytes(length) {
            sb.append(Character.forDigit((b >> 4) & 0xF, 16));
            sb.append(Character.forDigit((b & 0xF), 16));
        }
        return sb.toString();
    }
```

If you prefer to make it more similar to the UUID format you can also add the hyphen `-` after 4 or 8 bytes.

rmartinc (changes requested):
Thanks @douglaspalmer! WFYT about using the `SecretGenerator` if possible? And I think using `Arrays.toString` is not OK, we can use hexa for example.

rmartinc (approved):
Thganks @douglaspalmer! LGTM! Seeing the `SecretGenerator` now I would have preferred to not make the method static and use the `getInstance` but it's a minor detail. It's OK to me.

douglaspalmer:
> Thganks @douglaspalmer! LGTM! Seeing the `SecretGenerator` now I would have preferred to not make the method static and use the `getInstance` but it's a minor detail. It's OK to me.

Done, can you please review again?

rmartinc (approved):
Thanks @douglaspalmer! LGTM!

rmartinc:
@douglaspalmer Rebase this to see if CI runs OK.
@mposolda For me this is OK, now the IDs are just the hexa with no hyphens (so previously `58e0a7d7-eebc-11d8-9669-0800200c9a66` and now just the hexa `4e4d4046ab29803d1f38273d0938cc30`). But I think the ID is OK.

mposolda:
@douglaspalmer Few minor points:

- The issue https://github.com/keycloak/keycloak/issues/38663 mentions access-token IDs (and few other things in the description). But it seems this PR handles various secrets, but does not handle access-token IDs particularly, which are mentioned as main part in the issue? As access-token IDs (and similarly IDTokens or refreshToken Ids) are generated with `KeycloakModelUtils.generateId()` , which itself likely cannot be changed as it is used for many other things (EG. generating IDs for DB entities). See for example `TokenManager.initToken()` where is generated ID of the access-token.  There are few other methods on `TokenManager` for ID tokens and refresh tokens). It can be good to update those places as well.

- How about adding some method on `SecretGenerator` like:

```
public String generateSecureID() {
    SecretGenerator.getInstance().randomBytesHex(16);
}
```
It means that in 5 years or so, when there are new "security best practices", we will not need to change tons of places in the codebase and increase 16 to 32 for example, but we might be able to change just one place :-) WDYT?

mposolda:
@douglaspalmer Looks good. However there are test failures as we have some tests checking for token IDs being UUID. I think that `AssertEvents.isAccessTokenId` needs some update and it will fix big amount of test failures (maybe all).

ahus1:
@douglaspalmer - IMHO changing the IDs of the tokens could justify an entry in the notable changes in `changes-26_3_0.adoc`. While it is an internal behavior, maybe some people are depending on it in some kind of way. 

If you want this PR to be merged rather sooner than later, the updated docs can be handled in a follow-up PR. Leaving that up to you,  @mposolda  and @rmartinc  to decide.

douglaspalmer:
> @douglaspalmer - IMHO changing the IDs of the tokens could justify an entry in the notable changes in `changes-26_3_0.adoc`. While it is an internal behavior, maybe some people are depending on it in some kind of way.
> 
> If you want this PR to be merged rather sooner than later, the updated docs can be handled in a follow-up PR. Leaving that up to you, @mposolda and @rmartinc to decide.

I changed the PR yesterday and added dashes to the IDs. So the change should be invisible now.

mposolda (approved):
@douglaspalmer Sounds good, Thanks. Considering that the entropy is bigger, however the length of the ID and also the format is quite the same, I do not think that this worth to be documented.
@ahus1 If you disagree, feel free to create follow-up issue for this and we can eventually document for 26.3.

ahus1:
> I changed the PR yesterday and added dashes to the IDs. So the change should be invisible now.

Nice, works for me.

PR #40307, opened 2025-06-06
Internal protected field BaseSAML2BindingBuilder builder of nested class

### LINKED ISSUE #40072: The unused this.builder in BaseSoapBindingBuilder. SAST
### Description

`BaseSoapBindingBuilder` `builder`, passed to the `BaseSoapBindingBuilder` constructor initializes the this `this.builder`, which is not used anywhere else. The flags are checked and set for the builder variable that was passed to the constructor, not for the builder field of the class instance itself.

### Discussion

No response

### Motivation

Code clarity and maintainability

### Details

While reviewing the `BaseSoapBindingBuilder` class, I noticed that the constructor stores the builder parameter in an instance field `this.builder`, but the field is never used after being assigned.
All interactions (checking flags like `signAssertions`, calling methods like `signAssertion(document)`) are done directly on the constructor parameter `builder`, not the class field.

https://github.com/keycloak/keycloak/blob/7e9f76a6e57d0108da571f3a016517853f1edaba/saml-core/src/main/java/org/keycloak/saml/BaseSAML2BindingBuilder.java#L233-L253

LM4O322:
Closes #40072

### DIFF
--- a/saml-core/src/main/java/org/keycloak/saml/BaseSAML2BindingBuilder.java
+++ b/saml-core/src/main/java/org/keycloak/saml/BaseSAML2BindingBuilder.java
@@ -232,10 +232,8 @@ public class BaseSAML2BindingBuilder<T extends BaseSAML2BindingBuilder> {
 
     public static class BaseSoapBindingBuilder {
         protected Document document;
-        protected BaseSAML2BindingBuilder builder;
 
         public BaseSoapBindingBuilder(BaseSAML2BindingBuilder builder, Document document) throws ProcessingException {
-            this.builder = builder;
             this.document = document;
             if (builder.signAssertions) {
                 builder.signAssertion(document);


ahus1:
The field you are about to remove is part of the API that could be used by people who extend this. 

So we might first need to deprecate this field. 

Asking @keycloak/core-clients for input.

shawkins (approved):
LGTM - should add to @ahus1's concern (we have both saml-core - with this classes, and saml-core-public, but I have no knowledge of what api level guarentees are being made) this could just be simplied to use non-static inner classes, or since statefulness of the BaseSoapBindingBuilder isn't really used there could just be additional methods on the parent builders - which would simply further extensions, but of course doesn't match the pattern used for everything else.

ahus1 (changes requested):
@LM4O322 - there hasn't been a comment from the core team. If you want to proceed with this PR, please change it to just deprecating the field, and then I'll help adding a note to the upgrading guide that this field is now deprecated.

rmartinc:
@ahus1 @shawkins @LM4O322 I think that the saml-core is a public package so it can be extended in theory. I don't expect anybody doing it, but just deprecate it for the moment.

PR #39123, opened 2025-04-22
adding the ability to get the root configuration

shawkins:

PR #40309, opened 2025-06-06
Make bootstrap admins temporary conditionally

### LINKED ISSUE #40305: Introduce a switch by avoid printing information about temporary bootstrap admin
### Description

The goal of this ticket is to introduce a new switch, `--bootstrap-admin-temporary=true|false` that prevents printing our the information about the temporary bootstrap admin.

### Discussion

https://cloud-native.slack.com/archives/C056XU905S6/p1747682630432699

### Motivation

In some environments, like [UDS](https://uds.defenseunicorns.com/), the initial admin accounts are created by the admin users. This is the desired and intended workflow and information banner is misleading in this case.

### Details

Relates to:

* https://github.com/keycloak/keycloak/issues/30009
* https://github.com/defenseunicorns/uds-identity-config/issues/485
* Also to a degree to https://github.com/keycloak/keycloak/issues/30008

slaskawi:
This Pull Request introduces a flag to make the Bootstrap Admin temporary or permanent. 

Closes #40305

### DIFF
--- a/docs/guides/server/bootstrap-admin-recovery.adoc
+++ b/docs/guides/server/bootstrap-admin-recovery.adoc
@@ -95,6 +95,13 @@ The `--no-prompt` parameter can be useful if the username or client ID should be
 
 This creates a temporary admin user with the default username without prompting for confirmation. For more information, see the <<Default values>> section above.
 
+== Treating the temporary admin account as a permanent one
+
+Both `bootstrap-admin user` and `bootstrap-admin service` commands support the `--is-temporary` parameter. This parameter allows you to treat the created admin account as a permanent one, meaning it will not be marked as temporary and will not be subject to the temporary account restrictions.
+
+A temporary admin account can also be created with the Welcome Screen that is displayed at the first start of {project_name}. In this case, the temporary account behavior is controlled by the `--bootstrap-admin-temporary` parameter. If the variable or setting is not defined, the account will be created as a temporary one.
+
+At the moment, a temporary account is reflected only in a form of visual warnings when using it, it does not actually expire the account. This might change with a future release.
 == Environment variables
 
 For the `bootstrap-admin user` command, both username and password can be optionally set as environment variables:
--- a/quarkus/config-api/src/main/java/org/keycloak/config/BootstrapAdminOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/BootstrapAdminOptions.java
@@ -5,17 +5,25 @@ public class BootstrapAdminOptions {
     public static final String DEFAULT_TEMP_ADMIN_USERNAME = "temp-admin";
     public static final String DEFAULT_TEMP_ADMIN_SERVICE = DEFAULT_TEMP_ADMIN_USERNAME;
     public static final int DEFAULT_TEMP_ADMIN_EXPIRATION = 120;
+    public static final boolean DEFAULT_TEMP_ADMIN_IS_TEMPORARY = true;
     private static final String USED_ONLY_WHEN = " Used only when the master realm is created.";
     private static final String NON_CLI = " Use a non-CLI configuration option for this option if possible.";
 
+    // Since the Configuration object is not available in the WelcomeResource, we'll just pass it through a static
+    // handler. It's definitely a suboptimal approach but the alternatives are to either put it in the Profile
+    // (which doesn't feel right) or RealmModel (which seems like too much).
+    // There have been some other attempts to handle this better (https://github.com/keycloak/keycloak/pull/39123)
+    // but they didn't get too much traction.
+    private static boolean isTemporaryRuntimeAdmin = true;
+
     public static final Option<String> PASSWORD = new OptionBuilder<>("bootstrap-admin-password", String.class)
             .category(OptionCategory.BOOTSTRAP_ADMIN)
-            .description("Temporary bootstrap admin password." + USED_ONLY_WHEN + NON_CLI)
+            .description("Bootstrap admin password." + USED_ONLY_WHEN + NON_CLI)
             .build();
 
     public static final Option<String> USERNAME = new OptionBuilder<>("bootstrap-admin-username", String.class)
             .category(OptionCategory.BOOTSTRAP_ADMIN)
-            .description("Temporary bootstrap admin username." + USED_ONLY_WHEN)
+            .description("Bootstrap admin username." + USED_ONLY_WHEN)
             .defaultValue(DEFAULT_TEMP_ADMIN_USERNAME)
             .build();
 
@@ -27,13 +35,28 @@ public class BootstrapAdminOptions {
 
     public static final Option<String> CLIENT_ID = new OptionBuilder<>("bootstrap-admin-client-id", String.class)
             .category(OptionCategory.BOOTSTRAP_ADMIN)
-            .description("Client id for the temporary bootstrap admin service account." + USED_ONLY_WHEN)
+            .description("Client id for the bootstrap admin service account." + USED_ONLY_WHEN)
             .defaultValue(DEFAULT_TEMP_ADMIN_SERVICE)
             .build();
 
     public static final Option<String> CLIENT_SECRET = new OptionBuilder<>("bootstrap-admin-client-secret", String.class)
             .category(OptionCategory.BOOTSTRAP_ADMIN)
-            .description("Client secret for the temporary bootstrap admin service account." + USED_ONLY_WHEN + NON_CLI)
+            .description("Client secret for the bootstrap admin service account." + USED_ONLY_WHEN + NON_CLI)
             .build();
+
+    public static final Option<Boolean> IS_TEMPORARY = new OptionBuilder<>("bootstrap-admin-temporary", Boolean.class)
+            .category(OptionCategory.BOOTSTRAP_ADMIN)
+            .description("Indicates whether the admin user or service account is temporary." + USED_ONLY_WHEN)
+            .defaultValue(DEFAULT_TEMP_ADMIN_IS_TEMPORARY)
+            .build();
+
+    public static synchronized void setTemporaryRuntimeAdmin(boolean temporary) {
+        isTemporaryRuntimeAdmin = temporary;
+    }
+
+    public static synchronized boolean isTemporaryRuntimeAdmin() {
+        return isTemporaryRuntimeAdmin;
+    }
+
 
 }
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminService.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminService.java
@@ -54,6 +54,9 @@ public class BootstrapAdminService extends AbstractNonServerCommand {
     @Option(paramLabel = "SECRET", names = { "--client-secret:env" }, description = "Environment variable name for the client secret")
     String clientSecretEnv;
 
+    @Option(names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
+    boolean isTemporary;
+
     String clientSecret;
     String clientId;
 
@@ -105,7 +108,7 @@ public class BootstrapAdminService extends AbstractNonServerCommand {
         //BootstrapAdmin bootstrap = spec.commandLine().getParent().getCommand();
         KeycloakSessionFactory sessionFactory = KeycloakApplication.getSessionFactory();
         KeycloakModelUtils.runJobInTransaction(sessionFactory, session -> application
-                .createTemporaryMasterRealmAdminService(clientId, clientSecret, /* bootstrap.expiration, */ session));
+                .createTemporaryMasterRealmAdminService(clientId, clientSecret, isTemporary, /* bootstrap.expiration, */ session));
     }
 
     @Override
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminUser.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminUser.java
@@ -54,6 +54,9 @@ public class BootstrapAdminUser extends AbstractNonServerCommand {
     @Option(paramLabel = "PASSWORD", names = { "--password:env" }, description = "Environment variable name for the admin user password")
     String passwordEnv;
 
+    @Option(names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
+    boolean isTemporary;
+
     String password;
     String username;
 
@@ -105,7 +108,7 @@ public class BootstrapAdminUser extends AbstractNonServerCommand {
         //BootstrapAdmin bootstrap = spec.commandLine().getParent().getCommand();
         KeycloakSessionFactory sessionFactory = KeycloakApplication.getSessionFactory();
         KeycloakModelUtils.runJobInTransaction(sessionFactory, session -> application
-                .createTemporaryMasterRealmAdminUser(username, password, /* bootstrap.expiration, */ session));
+                .createTemporaryMasterRealmAdminUser(username, password, isTemporary, /* bootstrap.expiration, */ session));
     }
 
     @Override
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/BootstrapAdminPropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/BootstrapAdminPropertyMappers.java
@@ -53,6 +53,8 @@ public final class BootstrapAdminPropertyMappers {
                         .paramLabel("client secret")
                         .isMasked(true)
                         .build(),
+                fromOption(BootstrapAdminOptions.IS_TEMPORARY)
+                        .build()
         };
     }
 
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/jaxrs/QuarkusKeycloakApplication.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/jaxrs/QuarkusKeycloakApplication.java
@@ -63,7 +63,8 @@ public class QuarkusKeycloakApplication extends KeycloakApplication {
 
     @Override
     protected void loadConfig() {
-        // no need to load config provider because we force quarkus impl
+        var isTemporary = Boolean.parseBoolean(Configuration.getOptionalKcValue(BootstrapAdminOptions.IS_TEMPORARY.getKey()).orElse("true"));
+        BootstrapAdminOptions.setTemporaryRuntimeAdmin(isTemporary);
     }
 
     @Override
@@ -76,10 +77,10 @@ public class QuarkusKeycloakApplication extends KeycloakApplication {
 
         try {
             //Integer expiration = Configuration.getOptionalKcValue(BootstrapAdminOptions.EXPIRATION.getKey()).map(Integer::valueOf).orElse(null);
-            if (StringUtil.isNotBlank(adminPassword) && !createTemporaryMasterRealmAdminUser(adminUsername, adminPassword, /*expiration,*/ session)) {
+            if (StringUtil.isNotBlank(adminPassword) && !createTemporaryMasterRealmAdminUser(adminUsername, adminPassword, BootstrapAdminOptions.isTemporaryRuntimeAdmin(), /*expiration,*/ session)) {
                 throw new RuntimeException("Aborting startup and the creation of the master realm, because the temporary admin user account could not be created.");
             }
-            if (StringUtil.isNotBlank(clientSecret) && !createTemporaryMasterRealmAdminService(clientId, clientSecret, /*expiration,*/ session)) {
+            if (StringUtil.isNotBlank(clientSecret) && !createTemporaryMasterRealmAdminService(clientId, clientSecret, BootstrapAdminOptions.isTemporaryRuntimeAdmin(), /*expiration,*/ session)) {
                 throw new RuntimeException("Aborting startup and the creation of the master realm, because the temporary admin service account could not be created.");
             }
         } catch (NumberFormatException e) {
@@ -102,12 +103,12 @@ public class QuarkusKeycloakApplication extends KeycloakApplication {
         }
     }
 
-    public boolean createTemporaryMasterRealmAdminUser(String adminUserName, String adminPassword, /*Integer adminExpiration,*/ KeycloakSession session) {
-        return new ApplianceBootstrap(session).createTemporaryMasterRealmAdminUser(adminUserName, adminPassword /*, adminExpiration*/, false);
+    public boolean createTemporaryMasterRealmAdminUser(String adminUserName, String adminPassword, boolean isTemporary, /*Integer adminExpiration,*/ KeycloakSession session) {
+        return new ApplianceBootstrap(session).createMasterRealmAdminUser(adminUserName, adminPassword, isTemporary /*, adminExpiration*/, false);
     }
 
-    public boolean createTemporaryMasterRealmAdminService(String clientId, String clientSecret, /*Integer adminExpiration,*/ KeycloakSession session) {
-        return new ApplianceBootstrap(session).createTemporaryMasterRealmAdminService(clientId, clientSecret /*, adminExpiration*/);
+    public boolean createTemporaryMasterRealmAdminService(String clientId, String clientSecret, boolean isTemporary, /*Integer adminExpiration,*/ KeycloakSession session) {
+        return new ApplianceBootstrap(session).createMasterRealmAdminService(clientId, clientSecret, isTemporary /*, adminExpiration*/);
     }
 
 }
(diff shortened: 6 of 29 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminUser.java
```diff
@@ -54,6 +54,12 @@ static class UsernameOptions {
     @Option(paramLabel = "PASSWORD", names = { "--password:env" }, description = "Environment variable name for the admin user password")
     String passwordEnv;
 
+    @Option(paramLabel = "IS_TEMPORARY", names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
```
slaskawi: A question for reviewers: Shall we document this somewhere? If yes, where? The Admin Account Recovery page doesn't seem like a good fit and I didn't find any other place mentioning these settings.

services/src/main/java/org/keycloak/services/managers/ApplianceBootstrap.java
```diff
@@ -121,7 +121,7 @@ public boolean createMasterRealm() {
      * @param initialUser if true only create the user if no other users exist
      * @return false if the user could not be created
      */
-    public boolean createTemporaryMasterRealmAdminUser(String username, String password, /*Integer expriationMinutes,*/ boolean initialUser) {
+    public boolean createTemporaryMasterRealmAdminUser(String username, String password, boolean isTemporary, /*Integer expriationMinutes,*/ boolean initialUser) {
```
shawkins: This method name should change to createMasterRealmAdminUser, with a similar change to the service account method.

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelp.approved.txt
```diff
@@ -261,6 +261,10 @@ Bootstrap Admin:
 --bootstrap-admin-password <password>
                      Temporary bootstrap admin password. Used only when the master realm is
```
shawkins: The exisiting option messages likely need to remove the term temporary, or clarify that temporary is optional.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminUser.java
```diff
@@ -54,6 +54,12 @@ static class UsernameOptions {
...
+    boolean isTemporary;
+
+    @Option(paramLabel = "is_temporary", names = { "--is-temporary:env" }, description = "Environment variable name for whether the admin account is temporary")
```
shawkins: We shouldn't need the env form of the is-temporary flag - there isn't a security reason why it can't appear in the cli command, and the user can just do --is-temporary=\${TEMP_VAR}

shawkins (changes requested):
Thanks @slaskawi, just a couple of things that could be addressed.

services/src/main/java/org/keycloak/services/managers/ApplianceBootstrap.java
slaskawi: Fixed!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminUser.java
slaskawi: Removed!

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelp.approved.txt
slaskawi: That's a good catch! I updated the descriptions and removed the term "temporary". Should read better now.

slaskawi:
@shawkins Comments addressed and this PR is ready for the next round. 

Highlight of the changes:

- Removed the "temporary" word from the the bootstrap admin help messages
- Introduced a different logging message indicating the created admin is not a temporary account
- Fixed missing assertions in tests that I found when working on this PR
- Renamed methods as suggested in the previous review

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
```diff
@@ -302,4 +305,14 @@ private void csrfCheck(final MultivaluedMap<String, String> formData) {
         }
     }
 
+    private boolean createsTemporaryAdminUser() {
```
shawkins: This seems like it needs more refinement. The screen shown to the user says temporary in several places - do you plan on updating the template as well?

Also behavior doesn't match the behavior of the other properties. If I set the admin username env variable, that does not become the default in the form.

slaskawi: > Also behavior doesn't match the behavior of the other properties. If I set the admin username env variable, that does not become the default in the form.

At the moment, this works on two levels - it can be set via an Environment Property (`KC_BOOTSTRAP_ADMIN_IS_TEMPORARY`) or in the `theme.properties` file. The latter gives a chance of override this in a custom theme. At the moment, the theme override takes the priority over the environment variable to give custom theme authors more flexibility.

I'm open for a discussion if you'd like me to change the mechanism how this works.

slaskawi: I adjusted the theme for these changes.

slaskawi:
@shawkins The comments have been addressed and the change is ready for the next round.

The major changes are in Welcome Page theme. It now doesn't print information about temporary user if the `kcBootstrapAdminIsTemporary` setting is set to `false`.

I'm still considering writing a unit test for it but this PR is also fine without it (this is a bit niche feature).

docs/guides/server/bootstrap-admin-recovery.adoc
```diff
@@ -105,4 +111,6 @@ For the `bootstrap-admin service` command, the client ID is optional and default
 
 <@kc.bootstrapadmin parameters="service --client-id:env <YourClientIdEnv> --client-secret:env <YourSecretEnv>"/>
 
+Both commands also support using the `--is-temporary:env` parameter and enables using a different environment variable to control whether the created account is temporary or permanent.
```
shawkins: This can be removed.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/jaxrs/QuarkusKeycloakApplication.java
```diff
@@ -77,16 +78,17 @@ protected void loadConfig() {
     protected void createTemporaryAdmin(KeycloakSession session) {
         var adminUsername = getOption(BootstrapAdminOptions.USERNAME.getKey(), KEYCLOAK_ADMIN_ENV_VAR);
         var adminPassword = getOption(BootstrapAdminOptions.PASSWORD.getKey(), KEYCLOAK_ADMIN_PASSWORD_ENV_VAR);
+        var isTemporary = Boolean.parseBoolean(Configuration.getOptionalKcValue(BootstrapAdminOptions.EXPIRATION.getKey()).orElse("true"));
```
shawkins: This should not reference IS_TEMPORARY correct? 

The EXPIRATION option was never implemented - it still might at some point, which will cause some additional validation with the IS_TEMPORARY flag.

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
```diff
@@ -302,4 +310,14 @@ private void csrfCheck(final MultivaluedMap<String, String> formData) {
         }
     }
 
+    protected boolean createsTemporaryAdminUser() {
+        try {
+            String themeProperty = "kcBootstrapAdminIsTemporary";
```
shawkins: The method name should be createAdminUser, right?

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
shawkins: I see that it works on two levels, but I'm not sure that either is needed - in particular the usage of the environment variable since it's not consistent behavior.

The creation of the initial admin via the localhost welcome page also seems like it won't come into play with for most production usage.

Why not just have a checkbox in the form for the initial admin to choose? Are you thinking that that anyone who may has themed this page will want to control the one time operation without leaving it up to the initial admin and/or won't want to update their template to add a checkbox?

cc @vmuzikar

docs/guides/server/bootstrap-admin-recovery.adoc
slaskawi: Done.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/jaxrs/QuarkusKeycloakApplication.java
slaskawi: You're eagle-eyed @shawkins ! Thanks for spotting this!

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
slaskawi: I renamed it to `isAdminUserTemporary` to make it more descriptive. In general, this returns `true` if the Welcome Page should create a temporary user.

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
slaskawi: From the Defense Unicorns UDS perspective, we would mostly need the Welcome Page approach. The platform is wired in such a way, the initial Platform Administrators use Port Forward (via `zarf connect` command) and create the initial Keycloak Administrator account this way. We treat this account as a final (and not temporary) one. 

Looking from this perspective, having a checkbox whether the Administrator account is temporary or not may appear weird to our users. Therefore, we'd still need to apply a custom theme and pretend this checkbox is always "checked" for us. Not the best option for us but we can deal with it. 

Since Keycloak has many entry points to create the initial Admin Account (command line, environment variables, initial Realm import and Welcome Page), I tried to deliver a solution for each of them. This way the code is "symmetric". I'm slightly concerns that otherwise this feature would look like "half-baked". But I'll leave the final call to both of you @vmuzikar and @shawkins - how would you like me to approach this?

slaskawi:
Comments addressed and this PR is ready for the next round.

Here's the list of changes:

* Added `WelcomeResourceTest` unit test cover the most important part of the `WelcomeResource`
* Renamed the `createsTemporaryAdminUser` method to `isAdminUserTemporary`
* Fixed the docs

quarkus/config-api/src/main/java/org/keycloak/config/BootstrapAdminOptions.java
```diff
@@ -27,13 +28,19 @@ public class BootstrapAdminOptions {
...
 
     public static final Option<String> CLIENT_SECRET = new OptionBuilder<>("bootstrap-admin-client-secret", String.class)
             .category(OptionCategory.BOOTSTRAP_ADMIN)
-            .description("Client secret for the temporary bootstrap admin service account." + USED_ONLY_WHEN + NON_CLI)
+            .description("Client secret for the bootstrap admin service account." + USED_ONLY_WHEN + NON_CLI)
+            .build();
+
+    public static final Option<Boolean> IS_TEMPORARY = new OptionBuilder<>("bootstrap-admin-temporary", Boolean.class)
+            .category(OptionCategory.BOOTSTRAP_ADMIN)
+            .description("Indicates whether this admin account is temporary." + USED_ONLY_WHEN + NON_CLI)
```
vmuzikar:
```suggestion
            .description("Indicates whether the admin user or service account is temporary." + USED_ONLY_WHEN + NON_CLI)
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/BootstrapAdminPropertyMappers.java
```diff
@@ -53,6 +53,9 @@ public static PropertyMapper<?>[] getMappers() {
                         .paramLabel("client secret")
                         .isMasked(true)
                         .build(),
+                fromOption(BootstrapAdminOptions.IS_TEMPORARY)
+                        .paramLabel("is temporary")
```
vmuzikar: No need for a label on boolean props.
```suggestion
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminUser.java
```diff
@@ -54,6 +54,9 @@ static class UsernameOptions {
     @Option(paramLabel = "PASSWORD", names = { "--password:env" }, description = "Environment variable name for the admin user password")
     String passwordEnv;
 
+    @Option(paramLabel = "IS_TEMPORARY", names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
```
vmuzikar: No need for a label, IMHO.
```suggestion
    @Option(names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/BootstrapAdminService.java
```diff
@@ -54,6 +54,9 @@ static class ClientIdOptions {
     @Option(paramLabel = "SECRET", names = { "--client-secret:env" }, description = "Environment variable name for the client secret")
     String clientSecretEnv;
 
+    @Option(paramLabel = "IS_TEMPORARY", names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
```
vmuzikar: No need for a label, IMHO.
```suggestion
    @Option(names = { "--is-temporary" }, description = "Indicates whether this admin account is temporary, defaults to true", defaultValue = "true")
```

docs/guides/server/bootstrap-admin-recovery.adoc
```diff
@@ -95,6 +95,12 @@ The `--no-prompt` parameter can be useful if the username or client ID should be
 
 This creates a temporary admin user with the default username without prompting for confirmation. For more information, see the <<Default values>> section above.
 
+== Treating the temporary admin account as a permanent one
+
+Both `bootstrap-admin user` and `bootstrap-admin service` commands support the `--is-temporary` parameter. This parameter allows you to treat the created admin account as a permanent one, meaning it will not be marked as temporary and will not be subject to the temporary account restrictions.
+
+A temporary admin account can also be created with the Welcome Screen that is displayed at the first start of {project_name}. In this case, the temporary account behavior is controlled either by the `KC_BOOTSTRAP_ADMIN_IS_TEMPORARY` environment variable or by the `kcBootstrapAdminIsTemporary` setting in `theme.properties`. If the variable or setting is not defined, the account will be created as a temporary one.
+
```
vmuzikar, line 103: We should probably add some clarification on what we mean by "temporary". Perhaps something along these lines:
```suggestion

At the moment, a temporary account is reflected only in a form of visual warnings when using it, it does not actually expire the account. This might change with a future release.
```

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
vmuzikar: > in particular the usage of the environment variable since it's not consistent behavior

Exactly.

Additionally, I do not believe a theme prop is something that should affect this, it's not its responsibility, IMHO.

I'm ok with having the ability to set the temporary flag via the Welcome Screen, though. However, we should use a systematic approach: fully leverage the `bootstrap-admin-temporary` option, i.e. it would accept the env var, the CLI option, etc.

vmuzikar (changes requested):
@slaskawi Thank you for the contribution!

slaskawi:
@vmuzikar @shawkins Thanks for the comments. I've pushed another update, here are the highlights:

- Removed the environment variable handling in `WelcomeResource` as suggested
- Unified approach with bootstrap parameters
- Updated the documentation

quarkus/config-api/src/main/java/org/keycloak/config/BootstrapAdminOptions.java
```diff
@@ -5,17 +5,23 @@ public class BootstrapAdminOptions {
...
     public static final String DEFAULT_TEMP_ADMIN_SERVICE = DEFAULT_TEMP_ADMIN_USERNAME;
     public static final int DEFAULT_TEMP_ADMIN_EXPIRATION = 120;
+    public static final boolean DEFAULT_TEMP_ADMIN_IS_TEMPORARY = true;
     private static final String USED_ONLY_WHEN = " Used only when the master realm is created.";
     private static final String NON_CLI = " Use a non-CLI configuration option for this option if possible.";
 
+    // Since the Configuration object is not available in the WelcomeResource, we'll just pass it through a static
+    // handler. It's definitely a suboptimal approach but the alternatives are to either put it in the Profile
+    // (which doesn't feel right) or RealmModel (which seems like too much).
+    private static boolean isTemporaryRuntimeAdmin = true;
```
slaskawi, line 17: @shawkins @vmuzikar Unfortunately I didn't find any better way to read the `--is-temporary` bootstrap parameter value from the `WelcomeResource`. This approach is not pretty but seems to work correctly. I'm open for suggestions on how to improve it (or we can proceed as is if it's OK).

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
slaskawi: Adjusted

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
```diff
@@ -302,4 +310,16 @@ private void csrfCheck(final MultivaluedMap<String, String> formData) {
         }
     }
 
+    protected String getThemeProperty(String propertyName, String defaultValue) {
```
shawkins: This is a minor behavioral change to the existing logic - catching rather than throwing an IOException. Is this method for subclasses? And / or does it be long as a utility method on Theme?

quarkus/config-api/src/main/java/org/keycloak/config/BootstrapAdminOptions.java
shawkins: Is it truly necessary for the is-temporary option to influnce the WelcomeResource?

The alternative to this static variable to map the is-temporary option to an spi key and use Config.scope

We mused about providing the root configuration to Config via something like https://github.com/keycloak/keycloak/pull/39123 but it didn't move forward.

vmuzikar: > The alternative to this static variable to map the is-temporary option to an spi key and use Config.scope

Yes, that would be the correct approach.

docs/guides/server/bootstrap-admin-recovery.adoc
```diff
@@ -95,6 +95,13 @@ The `--no-prompt` parameter can be useful if the username or client ID should be
 
 This creates a temporary admin user with the default username without prompting for confirmation. For more information, see the <<Default values>> section above.
 
+== Treating the temporary admin account as a permanent one
+
+Both `bootstrap-admin user` and `bootstrap-admin service` commands support the `--is-temporary` parameter. This parameter allows you to treat the created admin account as a permanent one, meaning it will not be marked as temporary and will not be subject to the temporary account restrictions.
+
+A temporary admin account can also be created with the Welcome Screen that is displayed at the first start of {project_name}. In this case, the temporary account behavior is controlled by the `--is-temporary` parameter. If the variable or setting is not defined, the account will be created as a temporary one.
```
vmuzikar: This is not quite correct, is it? For the start command, the option is named `--bootstrap-admin-temporary`. We should mention it more in general that the temporary flag is controlled not only by `--is-temporary`.

quarkus/config-api/src/main/java/org/keycloak/config/BootstrapAdminOptions.java
```diff
@@ -27,13 +33,28 @@ public class BootstrapAdminOptions {
...
 
     public static final Option<String> CLIENT_SECRET = new OptionBuilder<>("bootstrap-admin-client-secret", String.class)
             .category(OptionCategory.BOOTSTRAP_ADMIN)
-            .description("Client secret for the temporary bootstrap admin service account." + USED_ONLY_WHEN + NON_CLI)
+            .description("Client secret for the bootstrap admin service account." + USED_ONLY_WHEN + NON_CLI)
+            .build();
+
+    public static final Option<Boolean> IS_TEMPORARY = new OptionBuilder<>("bootstrap-admin-temporary", Boolean.class)
+            .category(OptionCategory.BOOTSTRAP_ADMIN)
+            .description("Indicates whether the admin user or service account is temporary." + USED_ONLY_WHEN + NON_CLI)
```
vmuzikar: Why would we suggest the CLI, it's not a secret.
```suggestion
            .description("Indicates whether the admin user or service account is temporary." + USED_ONLY_WHEN)
```

vmuzikar (changes requested):
@slaskawi Thank you for the updates.

services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
slaskawi: That's a good point. I'll revert it.

slaskawi:
> Is it truly necessary for the is-temporary option to influnce the WelcomeResource?
> The alternative to this static variable to map the is-temporary option to an spi key and use Config.scope
> We mused about providing the root configuration to Config via something like https://github.com/keycloak/keycloak/pull/39123 but it didn't move forward.

For the UDS Platform, having this functionality in WelcomeResource is vital. Putting this into Config seems a bit artificial as this is not really an SPI and there's only 1 implementation that probably won't be overridden. Maybe I'll just add a link to https://github.com/keycloak/keycloak/pull/39123 in the comment so that we can retrofit it later once we have better options.

docs/guides/server/bootstrap-admin-recovery.adoc
slaskawi: You're absolutely right! Corrected.

slaskawi:
@vmuzikar @shawkins I've addressed your comments and this PR should be ready for the next round of reviews.

docs/guides/server/bootstrap-admin-recovery.adoc
```diff
@@ -95,6 +95,13 @@ The `--no-prompt` parameter can be useful if the username or client ID should be
 
 This creates a temporary admin user with the default username without prompting for confirmation. For more information, see the <<Default values>> section above.
 
+== Treating the temporary admin account as a permanent one
+
+Both `bootstrap-admin user` and `bootstrap-admin service` commands support the `--is-temporary` parameter. This parameter allows you to treat the created admin account as a permanent one, meaning it will not be marked as temporary and will not be subject to the temporary account restrictions.
+
+A temporary admin account can also be created with the Welcome Screen that is displayed at the first start of {project_name}. In this case, the temporary account behavior is controlled by the `--bootstrap-admin-temporary` parameter. If the variable or setting is not defined, the account will be created as a temporary one.
```
vmuzikar, line 102: The `--bootstrap-admin-temporary` also work with the `--bootstrap-admin-*` options used for creating admin user/svc account at startup, no? Not just with the Welcome Screen.

vmuzikar:
> Putting this into Config seems a bit artificial as this is not really an SPI and there's only 1 implementation that probably won't be overridden.

It might feel artificial but it's currently the way of doing that. We don't want to create a direct dependency of core KC codebase on Quarkus dist config, hence the abstraction in form of SPI config layer.

But it's not a real blocker for me.

docs/guides/server/bootstrap-admin-recovery.adoc
shawkins: @vmuzikar that wouldn't be consistent with the bootstrap admin commands existing handling. The commands do not consult the bootstrap-admin=* options. Someone could be setting bootstrap-admin-* options in keycloak.conf for example, and we don't want that influencing the bootstrap admin command unintentionally.

So --bootstrap-admin-temporary should just affect the initial admin creation / Welcome Screen.

vmuzikar: > So --bootstrap-admin-temporary should just affect the initial admin creation / Welcome Screen.

Yes, that's what I meant. Sorry, I was probably not clear before.

I meant that `--bootstrap-admin-temporary` should also affect when you create an admin user/service account at startup, not just the Welcome Screen. And that this behaviour should be documented.

stianst:
I disagree with this approach; we don't have a mechanism to auto-delete the admin account at this point, so this option is complete unnecessary. If/when we do some form of auto removal of the admin account it should include a consideration of the scenarios where someone doesn't want to delete the account, but rather configure it properly.

I propose a completely different approach to the temporary admin account (never been a big fan of this in the first place). See https://github.com/keycloak/keycloak/discussions/41382

stianst (changes requested):
I'm not in favour of adding this option, see my comment

slaskawi:
Closing this one in favor of working on https://github.com/keycloak/keycloak/issues/41388

PR #40313, opened 2025-06-06
Rename "Browser - Conditional OTP" to "Browser - Conditional 2FA" in default browser flow

rmartinc:
Closes #40281

Just renaming the name of the conditional flow. Fixed the tests that are failing after that. This change needs to be then referenced in the documentation because the name appears several times is docs too. Just a reminder for the documentation issue.


mposolda (approved):
@rmartinc Thanks!

PR #40334, opened 2025-06-08
fix: switching to the registry addon for olm testing

### LINKED ISSUE #40099: [Keycloak Operator CI] - Test OLM Installation

shawkins:
Just relying upon the docker daemon is not possible - while you can workaround several opm / olm issues, you can't install the CatalogSource with an imagepullpolicy of IfNotPresent unless you have an image digest - and those are only created by a registry.

So the approach here is to use the registry addon. However I do see that it doesn't always work as expected with minikube (1.32.0 seems fine, but locally minikube 1.35.0 won't start with registry addon enabled). Hopefully we can stick to minikube / kubernetes combinations where all is good, but if not we'll then have to deploy or run an external registry, which has a couple of extra steps.

closes: #40099

### DIFF
--- a/.github/workflows/operator-ci.yml
+++ b/.github/workflows/operator-ci.yml
@@ -143,7 +143,7 @@ jobs:
           kubernetes version: ${{ env.KUBERNETES_VERSION }}
           github token: ${{ secrets.GITHUB_TOKEN }}
           driver: docker
-          start args: --memory=${{ env.MINIKUBE_MEMORY }}
+          start args: --memory=${{ env.MINIKUBE_MEMORY }} --addons=registry --insecure-registry=192.168.49.2:5000
 
       - name: Install OPM
         uses: redhat-actions/openshift-tools-installer@144527c7d98999f2652264c048c7a9bd103f8a82 # v1.13.1
@@ -169,7 +169,7 @@ jobs:
         working-directory: operator
         run: |
           eval $(minikube -p minikube docker-env)
-          ./scripts/olm-testing.sh ${GITHUB_SHA::6}
+          REGISTRY=192.168.49.2:5000 ./scripts/olm-testing.sh ${GITHUB_SHA::6}
 
       - name: Deploy an example Keycloak and wait for it to be ready
         working-directory: operator
--- a/operator/scripts/create-olm-test-catalog.sh
+++ b/operator/scripts/create-olm-test-catalog.sh
@@ -19,7 +19,7 @@ mkdir -p $SCRIPT_DIR/../olm/catalog/test-catalog
     --output yaml > test-catalog/operator.yaml
 
   opm render $BUNDLE_IMAGE:$VERSION \
-    --output=yaml >> test-catalog/operator.yaml
+    --output=yaml --skip-tls >> test-catalog/operator.yaml
 
   cat << EOF >> test-catalog/operator.yaml
 ---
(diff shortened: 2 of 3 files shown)


ahus1 (approved):
LGTM. Not merging yet in case someone else from CND team wants to review it.

.github/workflows/operator-ci.yml
```diff
@@ -169,7 +169,7 @@ jobs:
         working-directory: operator
         run: |
           eval $(minikube -p minikube docker-env)
-          ./scripts/olm-testing.sh ${GITHUB_SHA::6}
+          REGISTRY=192.168.49.2:5000 ./scripts/olm-testing.sh ${GITHUB_SHA::6}
```
vmuzikar, line 172: Is it safe to assume the IP address? Shouldn't we use something like `$(minikube ip)` instead?

shawkins: It already has to be hard-coded in the minikube start --insecure-registries, but yes you could change this spot if you want.

vmuzikar: Do we need to provide the `--insecure-registry` option? Problem is at that point we can't get the minikube's IP as it's not started yet.

shawkins: > Do we need to provide the --insecure-registry option?

Yes - https://minikube.sigs.k8s.io/docs/handbook/pushing/#4-pushing-to-an-in-cluster-using-registry-addon

> Problem is at that point we can't get the minikube's IP as it's not started yet.

Correct - if we change to a different driver, the ip may change. See that the docs refer you to this code https://github.com/kubernetes/minikube/blob/dfd9b6b83d0ca2eeab55588a16032688bc26c348/pkg/minikube/cluster/cluster.go#L408

vmuzikar: Ok, let's keep as is and we'll reconsider if it breaks.

vmuzikar (approved):
LGTM, thanks.

vmuzikar:
I think we could also backport to 26.2.

shawkins:
> I think we could also backport to 26.2.

https://github.com/keycloak/keycloak/pull/40349

PR #40344, opened 2025-06-09
Keycloak fails to start on MySQL Cluster due to missing primary key i…

### LINKED ISSUE #38602: Keycloak fails to start on MySQL Cluster due to missing primary key in databasechangelog
### Before reporting an issue

### Area

storage

### Describe the bug

When starting  Keycloak connected to a MySQL cluster that has replication enabled, the process fails with the following error:

```
2025-03-26 14:16:32,759 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) ERROR: liquibase.exception.LiquibaseException: liquibase.exception.DatabaseException: liquibase.exception.DatabaseException: The table does not comply with the requirements by an external plugin. [Failed SQL: (3098) INSERT INTO keycloak.DATABASECHANGELOG (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, `DESCRIPTION`, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('1.0.0.Final-KEYCLOAK-5461', 'sthorger@redhat.com', 'META-INF/jpa-changelog-1.0.0.Final.xml', NOW(), 1, '9:6f1016664e21e16d26517a4418f5e3df', 'createTable tableName=APPLICATION_DEFAULT_ROLES; createTable tableName=CLIENT; createTable tableName=CLIENT_SESSION; createTable tableName=CLIENT_SESSION_ROLE; createTable tableName=COMPOSITE_ROLE; createTable tableName=CREDENTIAL; createTable tab...', '', 'EXECUTED', NULL, NULL, 'DEV', '2998578512')]
...
 2025-03-26 14:16:32,760 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) ERROR: The table does not comply with the requirements by an external plugin.
```

On MySQL cluster side, we have:

```
8d57f27ed7f9 2025-03-26T14:23:25.062245Z 2387 [ERROR] [MY-011542] [Repl] Plugin group_replication reported: 'Table DATABASECHANGELOG does not have any PRIMARY KEY. This is not compatible with Group Replication.'
```

Disabling the primary key restriction in MySQL does not affect the results, error persists.

The [Liquibase DATABASECHANGELOG](https://docs.liquibase.com/concepts/tracking-tables/databasechangelog-table.html) documentation explicitly says there is no primary key on the DATABASECHANGELOG, apparently to avoid any database-specific restrictions on key lengths. Also, it says that `The composite of id, author, and filename is unique across all rows of the table`.

[MySQL documentation](https://dev.mysql.com/doc/refman/8.4/en/group-replication-requirements.html) also tells that tables need to have primary keys.

### Version

26.1.4

### Regression

### Expected behavior

Keycloak should successfully start with a MySQL cluster.

### Actual behavior

Keycloak fails to start due to `DATABASECHANGELOG` missing a primary key.

### How to Reproduce?

Start keycloak with a MySQL cluster (connection url should list all nodes like `db-url=jdbc:mysql://192.168.10.108:3306,192.168.10.108:3307,192.168.10.108:3308/keycloak?autoReconnect=true&failOverReadOnly=false&maxReconnects=10`)

### Anything else?

_No response_

vramik:
…n databasechangelog

Closes #38602

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/CustomChangeLogHistoryService.java
+++ b/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/CustomChangeLogHistoryService.java
@@ -31,13 +31,16 @@ import java.util.Map;
 import liquibase.ContextExpression;
 import liquibase.Labels;
 import liquibase.change.CheckSum;
+import liquibase.change.ColumnConfig;
 import liquibase.changelog.ChangeSet;
 import liquibase.changelog.RanChangeSet;
 import liquibase.changelog.StandardChangeLogHistoryService;
 import liquibase.database.Database;
 import liquibase.database.core.MySQLDatabase;
 import liquibase.exception.DatabaseException;
+import liquibase.executor.jvm.ChangelogJdbcMdcListener;
 import liquibase.logging.LogFactory;
+import liquibase.statement.core.AddPrimaryKeyStatement;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/CustomChangeLogHistoryService.java
```diff
@@ -47,18 +50,29 @@ public class CustomChangeLogHistoryService extends StandardChangeLogHistoryServi
...
+    }
+
+    @Override
+    public void init() throws DatabaseException {
+        super.init();
+
+        AddPrimaryKeyStatement pkStatement = new AddPrimaryKeyStatement(getLiquibaseCatalogName(), getLiquibaseSchemaName(), getDatabaseChangeLogTableName(), 
+                ColumnConfig.arrayFromNames("ID, AUTHOR, FILENAME"), "PK_DATABASECHANGELOG");
+        ChangelogJdbcMdcListener.execute(getDatabase(), ex -> ex.execute(pkStatement));
+        getDatabase().commit();
```
sguilhen, line 68: I'm wondering if we should check if the database changelog table exists before adding the primary key, like we do in the `getRanChangeSets()` method bellow (the `if (hasDatabaseChangelogTable()) `.

vramik: I believe it might be unnecessary, please correct me if I am wrong. 

It seems that after calling `StandardChangeLogHistoryService`'s init() method (`super.init();`) the table either already exists or is created within it or the exception is thrown. 

The `StandardChangeLogHistoryService` is guarded by `serviceInitialized` boolean flag to ensure the second call of it does nothing and I thing that is something what is missing in the PR. I'll update it.

sguilhen (approved):

ahus1 (approved):
Approving based on @sguilhen's review

PR #40349, opened 2025-06-09
fix: switching to the registry addon for olm testing (#40334)

shawkins:
closes: #40099

Signed-off-by: Steve Hawkins <shawkins@redhat.com>
(cherry picked from commit eb96b4a59f25b0cf0715c21f0b0d8325b4d4c5b9)

### DIFF
--- a/.github/workflows/operator-ci.yml
+++ b/.github/workflows/operator-ci.yml
@@ -180,7 +180,7 @@ jobs:
           kubernetes version: ${{ env.KUBERNETES_VERSION }}
           github token: ${{ secrets.GITHUB_TOKEN }}
           driver: docker
-          start args: --memory=${{ env.MINIKUBE_MEMORY }}
+          start args: --memory=${{ env.MINIKUBE_MEMORY }} --addons=registry --insecure-registry=192.168.49.0/24
 
       - name: Install OPM
         uses: redhat-actions/openshift-tools-installer@v1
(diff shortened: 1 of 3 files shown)


vmuzikar (approved):

vmuzikar:
Seems there is a related error:
```
Get "http://192.168.49.2:5000/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

shawkins:
> Seems there is a related error:

My only conclusion is that every single part of Docker is awful...

I've rerun the job to help clarify if this is error is deterministic.

The only reference for something like this is in https://minikube.sigs.k8s.io/docs/handbook/vpn_and_proxy/ which mentions a proxy blocking access. 

I can also:
- simply add no_proxy, but I don't believe that should be coming into play here
- do an earlier print out of the minikube ip. However you can see from the other runs on 26.2 that the dockerHost is reported as 192.168.49.2
- something else, such as getting the registry pod status / log to try to determine why it's failing
- try doing our own deployment / external start of the registry - but I wouldn't want to bet on whether that will work either

vmuzikar:
What if we used podman as the minikube driver instead? :D

shawkins:
> What if we used podman as the minikube driver instead? :D

It's still considered experimental of course - and we'd still need to run a registry...

Since the second run passed we can rule out a proxy issue. I also see no reason why the ip would change given what I've seen locally and with every other test run, but just in case I've changed it to use a subnet and minikube ip - along with a print out of the ip - that thrid time ran as well, and confirmed the usage of 192.168.49.2.

I've set it run repeatedly to see how often we might encounter this problem.

shawkins:
@vmuzikar the olm tests run 53 more times without a problem, so I'm thinking we are safe to merge even if we've only reduced, but not fully eliminated the flakiness.

vmuzikar:
@shawkins Agree, let's merge it.

Could you please prepare a PR to align the changes in `main` too?

shawkins:
> Could you please prepare a PR to align the changes in main too?

I can, but there was no indication from the additional runs that a different ip is being used - they were done still with a fixed ip as the registry. It also won't prevent an issue if we switch to a different driver. We'd have to instead look at all the possible default ip assignments ranges and proactively add those to the insecure-registry settings.

vmuzikar:
@shawkins Yeah, seems like the IP is actually stable. But I'd still prefer to have the branches aligned with the same workflow for the minikube.

PR #40352, opened 2025-06-09
Remove authentication session during logout if no valid user session is found

### LINKED ISSUE #39923: ModelDuplicateException on next login after deleting an account
### Before reporting an issue

### Area

authentication

### Describe the bug

I can see a ModelDuplicateException caused by a unique constraint violation ([stacktrace.txt](https://github.com/user-attachments/files/20408451/stacktrace.txt)) after a user has been deleted.

### Version

26.2.4

### Regression

### Expected behavior

- next login with an unrelated user is successful

### Actual behavior

- next login fails with status code 500

### How to Reproduce?

* A user with an active session deletes themselves and then logs out (back channel logout)
* The browser is redirected to the login page again
* Another unrelated user attempts to log in through this login page and receives an error with status code 500 and said ModelDuplicateException

### Anything else?

* We use a derived login theme and the double-submit prevention change (https://github.com/keycloak/keycloak/commit/cc4691a3fc926699c892b53838e6bb1398e0cdd5) has already been implemented in our login form
* We use an external user storage implementation; cache is disabled there
* Might be related #38671

pedroigor:
Closes #39923

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/endpoints/LogoutEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/endpoints/LogoutEndpoint.java
@@ -39,6 +39,7 @@ import org.keycloak.locale.LocaleSelectorProvider;
 import org.keycloak.models.AuthenticatedClientSessionModel;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.Constants;
+import org.keycloak.models.KeycloakContext;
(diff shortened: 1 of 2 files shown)


rmartinc (approved):
Thanks @pedroigor! LGTM! Re-run the failed CI job that seems totally unrelated.

ahus1 (approved):

PR #40371, opened 2025-06-10
Integrate passkeys with separate username and password forms

### LINKED ISSUE #40021: Passkeys conditional UI: integration with independent username and password form
### Description

Same idea than in #29596 but in the independent username and the password form (the authenticators that are separated,  the former just presents the username with no password field, and the latter asks for the password as the next step). Probably we will need to do two things:

1. Include the webauthn and the conditional UI code as done in the username password variant. Try to do it re-usable (currently the ftl is using a macro but maybe we need more improvements here).
2. The password form should detect that the previous username already authenticated the user and do not ask for the password as it was already authenticated via passkey. Probably we need to add something to the authn session to check.

This issue should continue with the idea that is finally taken in #29596.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #40021

PR to add passkeys to the separate username and password form authenticators. This PR follows the same idea used in the previous issue but it needs two more things:

1. The flow context adds a new method `success(String)` to mark the authenticator succeeded and it was used the passed credetial (otp, webauthn, cert, password,...). That type is added to the auth session in an attribute `last.authn.credential`.
2. The password authenticator directly successes without waiting for input the password if the passkeys feature is enabled and the previous property is set to passwordless webauthn. That means the username form was authenticated using passkeys and the password step can be skipped.

Tests added in a similar way of the previous PRs.

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/authentication/AbstractAuthenticationFlowContext.java
+++ b/server-spi-private/src/main/java/org/keycloak/authentication/AbstractAuthenticationFlowContext.java
@@ -150,6 +150,15 @@ public interface AbstractAuthenticationFlowContext {
     void success();
 
     /**
+     * Mark the current execution as successful and the auth session sets the
+     * credential type in the authentication session as the last credential used
+     * to authenticate the user.
+     *
+     * @param credentialType The credential used to authenticate the user
+     */
+    void success(String credentialType);
+
+    /**
      * Aborts the current flow
      *
      * @param error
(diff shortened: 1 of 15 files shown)


mposolda (approved):

mposolda:
@rmartinc Approving, Thanks! There are probably some corner-cases when the behaviour might not be the desired one (For example when usernameForm and passwordForm are not at the same level, but in some different subflows of the authentication flow). But IMO great for now.

Romain7495:
Hi @mposolda for the pending change that I proposed 2 weeks ago, the best way is to create a new issue ? 

https://github.com/keycloak/keycloak/pull/40371#pullrequestreview-2914064154

rmartinc:
@Romain7495 What change are you referring to?

Romain7495:
image

image

Maybe you are not seeing it because I am not contributor

rmartinc:
@Romain7495 Yep, create a new issue, I didn't think or test that scenario.

rmartinc:
@Romain7495 I have created https://github.com/keycloak/keycloak/issues/40717, I'll send a PR asap.

Romain7495:
Sorry I saw your message too late @rmartinc 

Thanks!

PR #40373, opened 2025-06-10
Pass Identity Provider config values to Keycloak themes

laureat-natzka:
Pass IDP custom config values to Keycloak themes in `social` variable:

- Allow IDP config values with `kcTheme-` prefix to be accessible in Keycloak themes
- `kcTheme-` prefix is removed eventually, and not needed when accessing the config value from themes
- Example usage in `ftl` template for a config: `kcTheme-someConfig`:
```
<#list social.providers as provider>
  <p>${provider.themeConfig["someConfig"]}</p>
</#list>
```

closes: #40377


laureat-natzka:
@sschu

pedroigor (commented):
If you can provide a single test to make sure the config keys are not removed when updating identity providers, we should be good with it.

Looks like a good change in order to provide more flexibility for custom brokers.

Out of curiosity, how do you manage such attributes?

laureat-natzka:
> If you can provide a single test to make sure the config keys are not removed when updating identity providers, we should be good with it.
> 
> Looks like a good change in order to provide more flexibility for custom brokers.
> 
> Out of curiosity, how do you manage such attributes?

Was thinking about adding a test but wasn't sure how to approach it. Can you expand a bit with what you mean by "updating identity providers" in this context?

I have a nodejs application that manages Keycloak resources. I have few attributes that I'd like to leverage in the themes so this would be quite helpful

pedroigor:
> Can you expand a bit with what you mean by "updating identity providers" in this context?

I mean, assertions to make sure config attributes with `kcTheme-` are not removed during updates. Perhaps not needed, given that during updates, the config map will be updated accordingly with what is sent in the request.

The theme tests can be similar to org.keycloak.testsuite.organization.admin.OrganizationThemeTest. See also https://github.com/keycloak/keycloak/tree/main/testsuite/integration-arquillian/servers/auth-server/services/testsuite-providers/src/main/resources/theme/organization.

pedroigor:
@laureat-natzka Can you please give me the rights to force push and write to your branch? I'm adding documentation as well as squashing commits.

pedroigor:
@laureat-natzka If you can reset your branch to https://github.com/pedroigor/keycloak/tree/idp-config-in-themes, perhaps it will be easier. The test is also updated.

laureat-natzka:
@pedroigor Thanks for the support! Should be reset now

pedroigor:
@laureat-natzka Thank you. Are you OK with the documentation? Anything to improve there?

pedroigor:
One thing I was thinking though is if we can do this:

* Instead of exposing the `kcTheme-` prefix and managing them through the `IdentityProviderRepresentation#config` map, we add a `IdentityProviderRepresentation.themeConfig` map where you just set the key/values you need
* Update the Identity Broker Admin API to check for any key/value defined in the `themeConfig` map and add them to the `config` map with the expected prefix.
* Do not expose `kcTheme-` key/values from the `config` map

My main point is to avoid exposing the `kcTheme-` prefix so that we have more flexibility in changing how theme-specific config options are managed.

From an API perspective, perhaps this approach provides a better UX.

laureat-natzka:
@pedroigor Docs lgtm!

> My main point is to avoid exposing the `kcTheme-` prefix so that we have more flexibility in changing how theme-specific config options are managed.
> 
> From an API perspective, perhaps this approach provides a better UX.

For my use case, the existing approach would be sufficient. If you see it as an issue/inconvenience for long term, let me know 

> Out of curiosity, how do you manage such attributes?

I also manage IdentityProvider resources using `terraform`.
If we'd go with `IdentityProviderRepresentation.themeConfig` (as you explained), I'd need to update the terraform provider accordingly also

pedroigor (approved):

pedroigor:
Thanks, @laureat-natzka. Waiting for CI.

laureat-natzka:
@pedroigor some storage test is failing in CI. Doesn't seem to be related to this. Flakiness?

PR #40384, opened 2025-06-10
Automatically connect to a writer instance of PostgreSQL

### LINKED ISSUE #40383: KC should connect to a writer instance of PostgreSQL automatically
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

As part of #24493 the idea was to automatically connect to a write instance of PostgreSQL using `targetServerType=primary`

Unfortunately the change didn't work for a production deployment, but only for tests.

### Version

main

### Regression

### Expected behavior

KC should by default connect to the write instance

### Actual behavior

There is nothing in the docs, and when running a production deployment there is also no defaults for that

### How to Reproduce?

Look at the DB URL that is created in the Quarkus JPA setups

### Anything else?

I'll prepare a PR

ahus1:
Closes #40383

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -1,3 +1,4 @@
+// ------------------------ Breaking changes ------------------------  //
 == Breaking changes
 
 Breaking changes are identified as requiring changes from existing users to their configurations.
@@ -5,6 +6,7 @@ In minor or patch releases we will only do breaking changes to fix bugs.
 
 === <TODO>
 
+// ------------------------ Notable changes ------------------------ //
 == Notable changes
 
 Notable changes where an internal behavior changed to prevent common misconfigurations, fix bugs or simplify running {project_name}.
@@ -23,14 +25,25 @@ GET /admin/realms/{realm}/users?exact=false&q=myattribute:
 
 The {project_name} Admin Client is also updated with a new method to search users by attribute using the `exact` request parameter.
 
+=== Automatic database connection properties for the PostgreSQL driver
+
+When running PostgreSQL reader and writer instances, {project_name} needs to always connect to the writer instance to do its work.
+
+Starting with this release, and when using the original PostgreSQL driver, {project_name} sets the `targetServerType` property of the PostgreSQL JDBC driver to `primary` to ensure that it always connects to a writable primary instance and never connects to a secondary reader instance in failover or switchover scenarios.
+
+You can override this behavior by setting your own value for `targetServerType` in the DB URL or additional properties.
+
+// ------------------------ Deprecated features ------------------------ //
 == Deprecated features
 
 The following sections provide details on deprecated features.
 
 === <TODO>
 
+// ------------------------ Removed features ------------------------ //
 == Removed features
 
 The following features have been removed from this release.
 
 === <TODO>
+
--- a/docs/guides/server/db.adoc
+++ b/docs/guides/server/db.adoc
@@ -271,6 +271,18 @@ show server_encoding;
 create database keycloak with encoding 'UTF8';
 ----
 
+== Preparing for PostgreSQL
+
+When running PostgreSQL reader and writer instances, {project_name} needs to always connect to the writer instance to do its work.
+When using the original PostgreSQL driver, {project_name} sets the `targetServerType` property of the PostgreSQL JDBC driver to `primary` to ensure that it always connects to a writable primary instance and never connects to a secondary reader instance in failover or switchover scenarios.
+
+You can override this behavior by setting your own value for `targetServerType` in the DB URL or additional properties.
+
+[NOTE]
+====
+The `targetServerType` is only applied automatically to the primary datasource, as requirements might be different for additional datasources.
+====
+
 [[preparing-keycloak-for-amazon-aurora-postgresql]]
 == Preparing for Amazon Aurora PostgreSQL
 
@@ -295,6 +307,8 @@ See the <@links.server id="containers" /> {section} for details on how to build 
 . Configure {project_name} to run with the following parameters:
 `db-url`:: Insert `aws-wrapper` to the regular PostgreSQL JDBC URL resulting in a URL like `+jdbc:aws-wrapper:postgresql://...+`.
 `db-driver`:: Set to `software.amazon.jdbc.Driver` to use the AWS JDBC wrapper.
+
+NOTE: When overriding the `wrapperPlugins` option of the AWS JDBC Driver, always include the `failover` or `failover2` plugin to ensure that {project_name} always connects to the writer instance even in failover or switchover scenarios.
 
 == Preparing for MySQL server
 
--- a/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
@@ -106,6 +106,11 @@ public class DatabaseOptions {
             .description("Deactivate specific named datasource <datasource>.")
             .build();
 
+    public static final Option<String> DB_POSTGRESQL_TARGET_SERVER_TYPE = new OptionBuilder<>("db-postgres-target-server-type", String.class)
+            .category(OptionCategory.DATABASE)
+            .hidden()
+            .build();
+
     /**
      * Options that have their sibling for a named datasource
      * Example: for `db-dialect`, `db-dialect-<datasource>` is created
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
@@ -15,10 +15,12 @@ import org.keycloak.utils.StringUtil;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import static org.keycloak.config.DatabaseOptions.DB;
 import static org.keycloak.config.DatabaseOptions.OPTIONS_DATASOURCES;
 import static org.keycloak.config.DatabaseOptions.getDatasourceOption;
 import static org.keycloak.config.DatabaseOptions.getKeyForDatasource;
@@ -50,6 +52,11 @@ final class DatabasePropertyMappers {
                         .to("quarkus.datasource.jdbc.url")
                         .mapFrom(DatabaseOptions.DB, DatabasePropertyMappers::getDatabaseUrl)
                         .paramLabel("jdbc-url")
+                        .build(),
+                fromOption(DatabaseOptions.DB_POSTGRESQL_TARGET_SERVER_TYPE)
+                        .to("quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType")
+                        .mapFrom(DatabaseOptions.DB, DatabasePropertyMappers::getPostgresqlTargetServerType)
+                        .isEnabled(() -> getPostgresqlTargetServerType(Configuration.getConfigValue(DB).getValue(), null) != null)
                         .build(),
                 fromOption(DatabaseOptions.DB_URL_HOST)
                         .paramLabel("hostname")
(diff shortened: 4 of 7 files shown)


ahus1:
@vmuzikar / @shawkins - I am trying to set a DB property by default similar to 

https://github.com/keycloak/keycloak/blob/24910d9e1c76a1e953936c558ab02a27eb91820e/model/jpa/src/main/java/org/keycloak/connections/jpa/DefaultJpaConnectionProviderFactory.java#L451-L461

but I didn't find a similar spot in Quarkus. When someone passes in a JDBC URL, there is no way good way I found intercept this to add a suffix. 

This also keep me wondering how much automation / magic should be done here. Still IMHO this is a very common reason why people struggle with the DB connection pools, as we've seen in support tickets. 

Is there a different way for this?

cc: @sguilhen

mabartos:
@ahus1 It should be possible to use the transformer for the DB_URL_PROPERTIES property mapper. It'd always append the `targetServerType` to the URL props. However, there's an issue because users would not be able to override it, which is not the right way IMO. Maybe the best approach would be warn them (as you do) if they use the DB URL props with Postgres and the prop is missing. 

@vmuzikar @shawkins Or any other opinion?

shawkins:
> When someone passes in a JDBC URL, there is no way good way I found intercept this to add a suffix.

@mabartos @ahus1 there is general quarkus mechanism via: quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType=primary - but setting that only for pg would require some new logic, or less ideal it would have a side effect (setting a system property) of the db-kind propertymapper.

vmuzikar:
Using `quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType` seems like the right way to do. We'd have a hidden option for that in order to hook up a transformer that will look at the db-kind option.

ahus1:
I have a Quarkus Unit test KeycloakPathConfigurationTest#testAuthEndpointAvailable failing with 

> SRCFG00014: The config property quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType is required but it could not be found in any config source

I did some debugging but without success. I would be happy for a pointer from the @keycloak/cloud-native team.

shawkins:
> I did some debugging but without success. I would be happy for a pointer from the @keycloak/cloud-native team.

This is due to it being a quarkus test - the test logic is performing a quarkus build, not our build command.

Then because the option has a default - and it is currently forced to by this code https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/quarkus/config-api/src/main/java/org/keycloak/config/OptionBuilder.java#L149 - the logic in PropertyMappingInterceptor will advertise the property name to the name iterator. This is needed so that quarkus knows about it when doing it's own property mapping.

From there it does not have a value because it will map to a null value.

Not sure if there's a great way to inhibit the quarkus build, or not contribute our interceptors / configsources which are coming in from classpath mechanisms.

ahus1:
@shawkins, thank you for the pointer. I changed the property to a String without a default, and then that one unit test passed. Let's wait for the full CI results.

ahus1:
Build is now green, looking for a @keycloak/cloud-native team member to review when you have the time. Thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
```diff
@@ -109,6 +128,29 @@ public static PropertyMapper<?>[] getDatabasePropertyMappers() {
         ));
     }
 
+    private static String getPostgresqlTargetServerType(String db, String dbDriver, String dbUrl, String dbUrlProperties, ConfigSourceInterceptorContext c) {
```
mabartos:
```suggestion
    private static String getPostgresqlTargetServerType(String db, ConfigSourceInterceptorContext c) {
```

It'd be better to fetch these additional props inside the method - directly in the if statement for Postgres.

quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
```diff
@@ -106,6 +106,11 @@ public class DatabaseOptions {
             .description("Deactivate specific named datasource <datasource>.")
             .build();
 
+    public static final Option<String> DB_POSTGRESQL_TARGET_SERVER_TYPE = new OptionBuilder<>("db-jdbc-default-options", String.class)
```
vmuzikar: Should we make this specific to the `targetServerType`?
```suggestion
    public static final Option<String> DB_POSTGRESQL_TARGET_SERVER_TYPE = new OptionBuilder<>("db-postgres-target-server-type", String.class)
```

quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
vmuzikar, line 1: I believe we need to reflect this also in the [additional datasources](https://github.com/keycloak/keycloak/blob/ac7e1ffba000a80b021458cb6a83df9348c0385f/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java#L118-L133). CC @mabartos

vmuzikar (changes requested):
@ahus1 Thanks for the PR.

quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
ahus1: @vmuzikar - I was giving it a thought, and then decided not to. I also added this to the docs: 

> The `taretServerType` is only applied automatically to the primary datasource, as requirements might be different for additional datasources.

An additional datasource might point to a read-only replica of a database, or might just not care about what type of DB it is. So therefore I wouldn't apply this to the additional datasources.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigurationTest.java
```diff
@@ -355,6 +354,26 @@ public void testDatabaseProperties() {
...
         assertEquals("test-schema", config.getConfigValue("kc.db-schema").getValue());
         assertEquals("test-schema", config.getConfigValue("kc.db-schema").getValue());
+
+        ConfigArgsConfigSource.setCliArgs("--db=postgres");
+        config = createConfig();
+        assertEquals("primary", config.getConfigValue("quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType").getValue());
+
+        ConfigArgsConfigSource.setCliArgs("--db=postgres");
+        config = createConfig();
+        assertEquals("primary", config.getConfigValue("quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType").getValue());
```
mabartos:
```suggestion
```

Seems duplicated

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigurationTest.java
```diff
@@ -355,6 +354,26 @@ public void testDatabaseProperties() {
...
+
+        ConfigArgsConfigSource.setCliArgs("--db=postgres", "--db-url-properties=\"targetServerType=any\"");
+        config = createConfig();
+        assertNull(config.getConfigValue("quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType").getValue());
```
mabartos:
```suggestion
        ConfigArgsConfigSource.setCliArgs("--db=postgres", "--db-url-properties=?targetServerType=any");
        config = createConfig();
        assertNull(config.getConfigValue("quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType").getValue());
        assertEquals("jdbc:postgresql://localhost:5432/keycloak?targetServerType=any", config.getConfigValue("quarkus.datasource.jdbc.url").getValue());
```

It would be good if you could check even the full URL in your checks.

mabartos (commented):
LGTM, just a few things to improve it.

docs/guides/server/db.adoc
```diff
@@ -271,6 +271,18 @@ show server_encoding;
...
+== Preparing for PostgreSQL
+
+When running PostgreSQL reader and writer instances, {project_name} needs to connect always to the writer instance to do its work.
+When using the original PostgreSQL driver, {project_name} sets the `targetServerType` property of the PostgreSQL JDBC driver to `primary` to ensure that it always connects to a writable primary instance and never connects to a secondary reader instance in failover or switchover scenarios.
+
+You can override this behavior by setting your own value for `targetServerType` in the DB URL or additional properties.
+
+[NOTE]
+====
+The `taretServerType` is only applied automatically to the primary datasource, as requirements might be different for additional datasources.
```
mabartos:
```suggestion
The `targetServerType` is only applied automatically to the primary datasource, as requirements might be different for additional datasources.
```

quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
mabartos: Not sure about this one. 

On one hand, it'd probably be good if additional datasources would try to connect to the writer instance when pg is used by default.

On the other hand, as pointed out, the additional datasources might have different requirements, and it'd be probably better to NOT doing that for these datasources.

If users want it for datasources, they can enable it via the `db-url-properties`. But I don't have a strong preference in this matter.

vmuzikar: > An additional datasource might point to a read-only replica of a database

Why would that be the case? Aren't additional datasources typically used by extensions, e.g. a user provider? That's the main reason I believe `targetServerType` should be also reflected by the additional datasource.

Though, it's not a blocker for me, we could do that as a follow-up.

ahus1: Additional datasources can be used in any kind of extension. Even if they are used for a user provider, we wouldn't know if they are just reading data, or if they also update data.

ahus1:
@mabartos - your review comments should now be addressed. Please re-review when you have the time.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
```diff
@@ -51,6 +53,17 @@ public static PropertyMapper<?>[] getDatabasePropertyMappers() {
...
+                fromOption(DatabaseOptions.DB_POSTGRESQL_TARGET_SERVER_TYPE)
+                        .to("quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType")
+                        .mapFrom(DatabaseOptions.DB, (db, c) ->
+                                getPostgresqlTargetServerType(
+                                        db,
+                                        c))
+                        .isEnabled(() ->
+                                getPostgresqlTargetServerType(
+                                        Configuration.getConfigValue(DB).getValue(),
+                                        null) != null)
```
mabartos:
```suggestion
                        .mapFrom(DatabaseOptions.DB, DatabasePropertyMappers::getPostgresqlTargetServerType)
                        .isEnabled(() -> getPostgresqlTargetServerType(Configuration.getConfigValue(DB).getValue(), null) != null)
```

Would be nicer to have it like this if you don't mind.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
```diff
@@ -114,6 +127,33 @@ public static PropertyMapper<?>[] getDatabasePropertyMappers() {
...
+            }
+            if (dbUrl != null && dbUrl.contains("targetServerType")) {
+                // targetServerType already set to same or different value in db-url, ignore
+                return null;
+            }
+            log.debug("setting targetServerType for PostgreSQL to 'primary'");
+            return "primary";
+        } else {
+            return null;
+        }
```
mabartos:
```suggestion
        Database.Vendor vendor = Database.getVendor(db).orElse(null);
        if (vendor != Database.Vendor.POSTGRES) {
            return null;
        }

        String dbDriver = Configuration.getConfigValue(DatabaseOptions.DB_DRIVER).getValue();
        String dbUrl = Configuration.getConfigValue(DatabaseOptions.DB_URL).getValue();
        String dbUrlProperties = Configuration.getConfigValue(DatabaseOptions.DB_URL_PROPERTIES).getValue();

        if (!Objects.equals(Database.getDriver(db, true).orElse(null), dbDriver) &&
                !Objects.equals(Database.getDriver(db, false).orElse(null), dbDriver)) {
            // Custom JDBC-Driver, for example, AWS JDBC Wrapper.
            return null;
        }
        if (dbUrlProperties != null && dbUrl != null && dbUrl.contains("${kc.db-url-properties:}") && dbUrlProperties.contains("targetServerType")) {
            // targetServerType already set to same or different value in db-url-properties, ignore
            return null;
        }
        if (dbUrl != null && dbUrl.contains("targetServerType")) {
            // targetServerType already set to same or different value in db-url, ignore
            return null;
        }
        log.debug("setting targetServerType for PostgreSQL to 'primary'");
        return "primary";
```

Would be nicer to have it like this to avoid any unnecessary config fetching for other vendors.

docs/guides/server/db.adoc
```diff
@@ -271,6 +271,18 @@ show server_encoding;
 create database keycloak with encoding 'UTF8';
 ----
 
+== Preparing for PostgreSQL
+
+When running PostgreSQL reader and writer instances, {project_name} needs to connect always to the writer instance to do its work.
```
mabartos:
```suggestion
When running PostgreSQL reader and writer instances, {project_name} needs to always connect to the writer instance to do its work.
```

nitpick

docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
```diff
@@ -0,0 +1,35 @@
...
+=== <TODO>
+
+// ------------------------ Notable changes ------------------------ //
+== Notable changes
+
+Notable changes where an internal behavior changed to prevent common misconfigurations, fix bugs or simplify running {project_name}.
+
+=== Automatic database connection properties for the PostgreSQL driver
+
+When running PostgreSQL reader and writer instances, {project_name} needs to connect always to the writer instance to do its work.
```
mabartos:
```suggestion
When running PostgreSQL reader and writer instances, {project_name} needs to always connect to the writer instance to do its work.
```
nitpick

mabartos (commented):
@ahus1 LGTM in general, just a few "polishing" comments to have it fine :))

quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
mabartos: Looks good to me when the note is present:

```
The `targetServerType` is only applied automatically to the primary datasource, as requirements might be different for additional datasources.
```

https://github.com/keycloak/keycloak/pull/40384/files#diff-9fb621f78ce6eace8d47a2d2b588f74b756e786199aecede45b5f3dcca37b64fR281-R284

vmuzikar: Ok, fair enough, let's keep as is. Probably doesn't make sense to assume the `targetServerType` in additional datasources.

ahus1:
@mabartos - your comments have been addressed. Please re-review. Thanks!

mabartos (approved):
@ahus1 Nice, thanks! LGTM

mabartos:
@vmuzikar Do you want to do a final look?

vmuzikar (approved):
LGTM, thank you @ahus1!

PR #40401, opened 2025-06-11
Fix invalid OpenAPI specifications

maelick:
* Admin REST resources: changes boolean and integer arguments with default values to basic java in order to fix openapi errors
* fix getClientRoleComposites endpoint path param client-uuid by renaming it to child-client-uuid

Closes #40400

Note: I am unsure what is the reason behind using object types rather than primitive Java types for boolean and integer arguments. If the change has no impact on the Keycloak behavior, maybe it would be better to change them all to use the primitive types?

### DIFF
--- a/services/src/main/java/org/keycloak/organization/admin/resource/OrganizationMemberResource.java
+++ b/services/src/main/java/org/keycloak/organization/admin/resource/OrganizationMemberResource.java
@@ -164,8 +164,8 @@ public class OrganizationMemberResource {
     public Stream<MemberRepresentation> search(
             @Parameter(description = "A String representing either a member's username, e-mail, first name, or last name.") @QueryParam("search") String search,
             @Parameter(description = "Boolean which defines whether the param 'search' must match exactly or not") @QueryParam("exact") Boolean exact,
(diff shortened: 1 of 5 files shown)


shawkins:
Unfortunately the original issue was already closed due to lack of updates, but the concern remains valid.

@maelick and @RiRa12621 please open new / more targeted issues. Note there has been a subsequent PR opened to address just the client-uuid issue #46016 

The same considerations will be present for admin api v2 - and we are actively looking at alternative client implementation there https://github.com/keycloak/keycloak/issues/43167 including the go client so ideally we shouldn't hit these problems moving foward.

> Note: I am unsure what is the reason behind using object types rather than primitive Java types for boolean and integer arguments. If the change has no impact on the Keycloak behavior, maybe it would be better to change them all to use the primitive types?

I believe the usage of object types is to be consistent with allowing for optional parameters. It should also allow for passing an explicit null.

But as you say that doesn't mesh well with using a Default. Let's see if anyone else weighs in on whether updating the Default or changing to a primitive makes sense here.

cc @keycloak/cloud-native

Issue #39283, opened 2025-04-28
[OID4VCI] Update credential response (even if deferred) to always return an array when not returning a `transaction_id`


IngridPuppet:
### Description

The new behavior of the Credential Endpoint is to think batch issuance first. For that, the structure of credential responses was updated to return an array of credentials.

See https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-credential-response

Related changelog entries:
>    * credential response always returns an array when not returning a transaction_id with the option for additional meta-data
>    * deferred credential response always returns an array (same as credential response)

PR #40409, opened 2025-06-11
[OID4VCI] Always Return Array for Credential Responses

forkimenjeckayang:
This PR updates the Keycloak OID4VCI implementation to comply with the latest [OID4VCI specification](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-credential-response), ensuring that the `credentials` field in the credential response is always an array (even for a single credential).

- **CredentialResponse Model:**  
  - Updated to ensure `credentials` is always a list of credential objects.
  - Removed/avoided any usage of single credential fields.

- **Issuer Endpoint:**  
  - Ensured that immediate issuance always returns an array of credentials.

### Closses Ticket

- #39283


Captain-P-Goldfish (approved):
looks good to me

francis-pouatcha (approved):

mposolda (approved):
@forkimenjeckayang @Captain-P-Goldfish @francis-pouatcha Thanks for the PR and the review!

PR #40414, opened 2025-06-11
Create a transient client session if the RPT is originating from a client session that does not exist in the user session

pedroigor:
Closes #39599


rmartinc (approved):
Thanks @pedroigor! Yes, now you have to create transient client session too (as we are doing in other parts). Thanks!

ahus1 (approved):

PR #40418, opened 2025-06-11
Avoid calling the underlying user storage multiple times when lazily loading model attributes

### LINKED ISSUE #40213: `UserStorageManager.getUserById` called multiple times on POST /realms/{realm}/protocol/{protocol}/token
### Before reporting an issue

### Area

storage

### Describe the bug

We are using user federation with a custom implementation via Keycloak User Storage SPI, this SPI calls an external REST API, these are the Keycloak interfaces the SPI is using:

- `org.keycloak.storage.UserStorageProvider`
- `org.keycloak.storage.user.UserLookupProvider`
- `org.keycloak.storage.user.UserQueryProvider`
- `org.keycloak.credential.CredentialInputValidator`
- `org.keycloak.credential.CredentialInputUpdater`
- `org.keycloak.storage.user.UserRegistrationProvider`

We noticed requests to `POST /realms/{realm}/protocol/{protocol}/token` result in multiple calls to `org.keycloak.storage.UserStorageManager.getUserById`, which seemed strange since we have the following cache configured:

```xml
        <local-cache name="users" statistics="true">
            <encoding>
                <key media-type="application/x-java-object"/>
                <value media-type="application/x-java-object"/>
            </encoding>
            <memory storage="HEAP" max-count="10000" />
            <expiration lifespan="1800000"/>
        </local-cache>
```

After some digging, we found what we think is causing this behavior, a change introduced in this commit https://github.com/keycloak/keycloak/commit/b4b3527df7656bfaaee351ec414e56c683c134c2. To be more precise, the changes in this file https://github.com/keycloak/keycloak/commit/b4b3527df7656bfaaee351ec414e56c683c134c2#diff-2df3739f3d8839fe9431e4147b70a2e5b65d0da63810b090dd30205d11715cc1, which explicitly fetch the user from the external storage, instead of using the cached value, multiple times.

This defeats the purpose of the `users` cache and is resulting in increased response times to get a token, this also happens when refreshing a token or when hitting userinfo.

### Version

25.0.4

### Regression

### Expected behavior

`org.keycloak.storage.UserStorageManager.getUserById` should be called once at most, per request to `POST /realms/{realm}/protocol/{protocol}/token`, the `users` cache should be used.

### Actual behavior

`org.keycloak.storage.UserStorageManager.getUserById` is being called 9 times in our current implementation, the number of times can vary depending on the specific scopes/attributes/roles/groups being included in the token.

### How to Reproduce?

1. Create a custom federated store via https://www.keycloak.org/docs/latest/server_development/index.html#_user-storage-spi.
2. Call `POST /realms/{realm}/protocol/{protocol}/token`.

### Anything else?

_No response_

pedroigor:
Closes #40213

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/GroupAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/GroupAdapter.java
@@ -51,7 +51,7 @@ public class GroupAdapter implements GroupModel {
         this.cacheSession = cacheSession;
         this.keycloakSession = keycloakSession;
         this.realm = realm;
-        modelSupplier = this::getGroupModel;
+        modelSupplier = new LazyModel<>(this::getGroupModel);
     }
 
     protected void getDelegateForUpdate() {
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/LazyModel.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/LazyModel.java
@@ -1,0 +1,21 @@
+package org.keycloak.models.cache.infinispan;
+
+import java.util.function.Supplier;
+
+public class LazyModel<M> implements Supplier<M> {
+
+    private final Supplier<M> supplier;
+    private M model;
+
+    public LazyModel(Supplier<M> supplier) {
+        this.supplier = supplier;
+    }
+
+    @Override
+    public M get() {
+        if (model == null) {
+            model = supplier.get();
+        }
+        return model;
+    }
+}
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
@@ -78,7 +78,7 @@ public class RealmAdapter implements CachedRealmModel {
         this.cached = cached;
         this.cacheSession = cacheSession;
         this.session = session;
-        this.modelSupplier = this::getRealm;
+        this.modelSupplier = new LazyModel<>(this::getRealm);
     }
 
     @Override
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RoleAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RoleAdapter.java
@@ -53,7 +53,7 @@ public class RoleAdapter implements RoleModel {
         this.cacheSession = cacheSession;
         this.session = cacheSession.session;
         this.realm = realm;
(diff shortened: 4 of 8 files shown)


model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/GroupAdapter.java
```diff
@@ -51,7 +51,7 @@ public GroupAdapter(CachedGroup cached, RealmCacheSession cacheSession, Keycloak
         this.cacheSession = cacheSession;
         this.keycloakSession = keycloakSession;
         this.realm = realm;
-        modelSupplier = this::getGroupModel;
+        modelSupplier = new LazyModel<>(this::getGroupModel);
```
mabartos, line 54:
```suggestion
        modelSupplier = Suppliers.memoize(this::getGroupModel);
```

@pedroigor AFAIK, we might leverage the Guava Memoization with all the necessary checks, thread-safety, etc. as we already have the dependency there. WDYT?

We might consider using it also in other places with suppliers.

```
import com.google.common.base.Suppliers;
```

pedroigor: Makes sense for me. I also found a similar implementation from Infinispan. However, the place where this code is used is thread-safe, isn't it?

Could not think about when creating a cache entry will run in multiple threads ...

mabartos: @pedroigor I'd rather go the way with Guava as it's widely spread and maintained.

I haven't checked the code if the thread-safety is necessary, but IMO it makes sense to use a verified existing solution with a proper test coverage rather than creating another custom one. 

But if you think otherwise and have other concerns, discard the suggestion :)

pedroigor: I only think that using a thread-safe component in a single-thread context (if the case) brings wrong assumptions when reading the code.

ahus1 (approved):
This code looks good to me. Unfortunately I found that DefaultLazyLoader is documented to not be thread safe, but is called from multiple threads. I opened issue #40531

pedroigor:
@ahus1 IIRC, not having the  `DefaultLazyLoader` with thread-safety controls was to avoid (unnecessary) blocking threads, as a trade-off. Not sure what is the best (prevent concurrent db loads vs concurrent db loads) as we did not run any benchmark.

ahus1:
We did the analysis for the other cache stampede and found that spiking DB connections are a problem, and that this pattern is working well.

So I suggest to go with the locking. One DB call is always faster than multiple DB calls. And it will make the first requests after the startup of a new node a lot smoother.

PR #40421, opened 2025-06-11
Update links to OAuth 2.1 draft spec and change link from BCP to RFC9700

### LINKED ISSUE #40419: Update links specs in OIDC guide
### Description

In the `Securing applications and services with OpenID Connect` guide, there are links to the OAuth 2.1 draft spec which is outdated and OAuth 2.0 Best Current Security Practice guide, which is currently an RFC (9700, since January). These links should be updated to reflect the latest versions of the draft and rfc.

Additionally, according to RFC 9700, the usage of the ROPC grant type `MUST NOT` be used, so this guide should also reflect this formulation and not state "should not be used".

### Discussion

_No response_

### Motivation

see description

### Details

_No response_

dasniko:
closes #40419

### DIFF
--- a/docs/guides/securing-apps/partials/oidc/oauth21-support.adoc
+++ b/docs/guides/securing-apps/partials/oidc/oauth21-support.adoc
@@ -3,7 +3,7 @@
 
 {project_name} makes it easier for administrators to make sure that their clients are compliant with these specifications:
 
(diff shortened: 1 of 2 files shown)


ahus1 (approved):
Thank you for this documentation update!

cc: @keycloak/core-clients, @thomasdarimont, @tnorimat

tnorimat:
@dasniko Thanks!

PR #40424, opened 2025-06-11
Re-sort the release notes and upgrading guide

ahus1:
Closes #40422

### DIFF
--- a/docs/documentation/release_notes/topics/26_3_0.adoc
+++ b/docs/documentation/release_notes/topics/26_3_0.adoc
@@ -1,59 +1,43 @@
-= Usage of remote stores for embedded caches is restricted
-
-Previously, usage of remote stores was discouraged and the experimental feature `cache-embedded-remote-store` needed to be enabled to run in this setup.
-In this release, the experimental feature was removed and usage of remote stores is restricted.
-If you previously used the `cache-embedded-remote-store` feature, see the link:{upgradingguide_link}[{upgradingguide_name}] for migration steps.
+// Release notes should contain only headline-worthy new features,
+// assuming that people who migrate will read the upgrading guide anyway.
 
 = Asynchronous logging
 
 All available log handlers now support asynchronous logging capabilities.
-Asynchronous logging might be useful for deployments requiring **high throughput** and **low latency** to move log operations in a separate thread.
+Asynchronous logging might be useful for deployments requiring high throughput and low latency to move log operations in a separate thread.
 
 For more details, see the https://www.keycloak.org/server/logging[Logging guide].
 
 = Client initiated linking of the user account to identity provider is based on AIA
 
-In this release, we added Application initiated action (AIA) implementation for linking a user account to the identity provider. The custom protocol, which was previously
+We added Application initiated action (AIA) implementation for linking a user account to the identity provider. The custom protocol, which was previously
 used for client initiated account linking is deprecated now. For more information, see the link:{upgradingguide_link}[{upgradingguide_name}].
-
-= Deprecated for removal the Instagram Identity Broker
-
-It has been a while since discussions started about any activity around the Instagram Identity Broker
-and any objection from the community about deprecating it for removal. For more details, see
-https://github.com/keycloak/keycloak/issues/37967[Deprecate for removal the Instagram social broker].
-
-In this release, the Instagram Identity Broker is deprecated for removal and is not enabled by default.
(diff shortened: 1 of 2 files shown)


docs/documentation/release_notes/topics/26_3_0.adoc
```diff
@@ -1,59 +1,44 @@
...
-
-In this release, the Instagram Identity Broker is deprecated for removal and is not enabled by default.
-If you are using this broker, it is recommended to use the Facebook Identity Broker instead. For more
-details, see link:{upgradingguide_link}[{upgradingguide_name}].
-
 = Recovery Codes supported
 
-In this release, the *Recovery Codes* two-factor authentication is promoted from preview to supported feature. For more information about the 2FA method, see the link:{adminguide_link}#_recovery-codes[Recovery Codes] chapter in the {adminguide_name}.
+The Recovery Codes two-factor authentication is promoted from preview to supported feature. For more information about the 2FA method, see the link:{adminguide_link}#_recovery-codes[Recovery Codes] chapter in the {adminguide_name}.
+For newly created realms, the browser flow now includes the Recovery Authentication Code Form as _Disabled__, and it can be switched to _Alternative_ by admins if they want to use this feature.
```
mhajas:
```suggestion
For newly created realms, the browser flow now includes the Recovery Authentication Code Form as _Disabled_, and it can be switched to _Alternative_ by admins if they want to use this feature.
```

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -152,4 +81,93 @@ This prevents accidentally downgrading functionality, for example, by using an o
...
+
+```
+--features=instagram-broker
+```
+
+It has been a while since discussions started about any activity around the Instagram Identity Broker
+and any objection from the community about deprecating it for removal. For more details, see
+https://github.com/keycloak/keycloak/issues/37967[Deprecate for removal the Instagram social broker].
+
+=== local admin deprecated for removal
```
mhajas:
```suggestion
=== Local admin deprecated for removal
```

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -152,4 +81,93 @@ This prevents accidentally downgrading functionality, for example, by using an o
...
+Use `spi-cache-embedded-default-site-name` instead in setups when running with embedded distributed caches.
+See the https://www.keycloak.org/server/all-provider-config[All provider configuration] for more details on these options.
+
+=== Deprecated proprietary protocol for client initiated linking to the identity provider account
+
+When you want the user, who is authenticated to your client application, to link his or her account to a specific identity provider, consider using the Application initiated action (AIA) based
+mechanism with the action `idp_link`. The proprietary custom protocol for client initiated account linking is deprecated now and might be removed in the future versions. For more information, see the
+Client initiated account link section of the link:{developerguide_link}[{developerguide_name}].
+
+=== Deprecated for removal the Instagram Identity Broker
```
mhajas, line 111: This section sounds like a breaking change to me. Not deprecation. Should we enable the feature by default to not be breaking?

mhajas (commented):
Thanks for the PR, I added a few comments

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
ahus1: I agree, I'm discussing this with @pedroigor at the moment. AFAIK we don't have an option deprecated-and-enabled. This will be handled in a follow-up PR. 

See https://github.com/keycloak/keycloak/discussions/37968 for the link that he shared. If we consider it "broken beyond repair", this could be one possible justification to remove it.

mhajas: Ok, sounds good to me. Thanks for looking into this. If we decide to keep it this way, I think we should alter move this part to the breaking changes section and add an explanation why we need to do this in breaking way.

mhajas (approved):
Thank you @ahus1!

mabartos:
@ahus1 @mhajas Would it be possible to have some sort of template for upgrading guide and potentially release notes in our codebase? As there are more fine-grained parts like 'Breaking changes, Notable changes,...', we could avoid some other formatting issues and avoid wrong copy-paste-delete :) 

Having sth like `changes-template.adoc` ? Just a small improvement for the future us.

PR #40426, opened 2025-06-11
V2 patch

shawkins:
also shows basic field validation

@vmuzikar @mabartos @Pepo48 this shows roughly what json merge patch support looks like - there is very little we need to do as Jackson has support already baked in. There are several TODOs - moving field validation to a common spot, using a common objectmapper, agreeing on the response codes, etc.

The server side method to allow for other patch types has the payload typed as JsonNode (if that doesn't work, then we'd just use String). 

From a rest client perspective you could have a typed method. Tthe user will do something like:

```
ClientRepresentation rep = new ClientRepresentation();
rep.setEnabled(false);
rep.getAdditionalField().put("description", null); // remove description

clients.getClient("id").patchClient(rep, FieldValidation.Strict);
```

### DIFF
--- a/core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
@@ -1,0 +1,31 @@
+package org.keycloak.representations.admin.v2;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+public class BaseRepresentation {
+
+    @JsonIgnore
+    protected Map<String, Object> additionalFields = new LinkedHashMap<String, Object>();
+
+    @JsonAnyGetter
+    public Map<String, Object> getAdditionalFields() {
+        return additionalFields;
+    }
+
+    @JsonAnySetter
+    public void setAdditionalFields(String name, Object value) {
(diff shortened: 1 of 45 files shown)


core/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    public static void main(String[] args) throws JsonProcessingException {
+        ClientRepresentation rep = new ClientRepresentation();
+        rep.setAuth(new Auth());
+        ObjectMapper mapper = new ObjectMapper();
+        var reader = mapper.readerForUpdating(rep);
+        ClientRepresentation patch = new ClientRepresentation();
+        patch.getAdditionalFields().put("auth", null);
```
mabartos, line 241:
```suggestion
        patch.removeAdditionalField("auth");
```

Would it be possible to have sth like this?

core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+public class BaseRepresentation {
```
mabartos, line 12: Agree to have the base representation as these additional fields will probably be necessary for all of them. 

Is there any intention to store metadata in the additional fields?

mabartos (commented):
@shawkins Thanks for this and nice work!

I just briefly went over these changes and added a few higher-level comments. 

Could you please send this PR to the `feature/admin-api-v2` branch for better review and discussion? Thanks!

PR #40428, opened 2025-06-11
Key generation for client authentication is always RSA 2048 with a 10-year validity, regardless of the selected algorithm

### LINKED ISSUE #38620: Key generation for client authentication is always RSA 2048 with a 10-year validity, regardless of the selected algorithm
### Before reporting an issue

### Area

authentication

### Describe the bug

When configuring a client to use the "Signed Jwt" authenticator with a signature algorithm other than RS256, the key generation from the "Keys" tab still produces an RSA key with a 2048-bit modulus and a validity period of 10 years. According to [NIST SP 800-78-5](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-78-5.pdf), 2048-bit RSA keys are acceptable for use through 2030, but beyond that, 3072-bit keys are recommended to maintain at least 128 bits of security strength. Likewise, ANSSI’s TLS recommendations tolerate 2048-bit RSA only for short-term needs and advise the use of 3072-bit or higher for long-term protection of sensitive data.

In addition to key size, shorter lifespans help limit exposure in case of key compromise. While NIST does not prescribe exact lifetimes for authentication keys, limiting them to three years or less is considered best practice. Enable admins to use 3072-bit RSA keys with a three-year validity period would align with both NIST and ANSSI guidance.

### Version

26.1.4

### Regression

### Expected behavior

Keycloak should generate a private key and certificate pair that corresponds to the selected signature algorithm with a validity period of no more than 3 years.

### Actual behavior

Keycloak generates an RSA key pair with a 2048-bit modulus and a validity period of 10 years.

### How to Reproduce?

1.  Go to the Keycloak admin console.
2.  Create a new client or select an existing one.
3.  Navigate to the "Credentials" tab.
4.  Set "Client Authenticator" to "Signed Jwt".
5.  Choose any "Signature algorithm" other than RS256 (e.g., ES256, HS256).
6.  Go to the "Keys" tab.
7.  Click on "Generate new keys".
8.  Export the generated key pair as a PKCS#12 container.
9.  Extract the certificate and examine its validity period.

### Anything else?

_No response_

douglaspalmer:
Closes #38620


rmartinc:
Just my opinion but I think we need UI changes for this, adding the bits and the validity as parameters of the `generate` and `generate-and-download` endpoints (`ClientAttributeCertificateResource`). In the UI the bits can be a select (2048, 3072 and 4098 for example) and the time element we have for example in session timeouts. The default values for me are 3072 and 3 years as commented in the issue. I would just restrict the change for that endpoint (client certificate generation, OIDC and SAML).

mposolda (approved):

mposolda:
@keycloak/ui-maintainers Anyone from the UI team available to review this?

ssilvert (approved):
UI part looks good.

mposolda:
@ssilvert Thanks!

PR #40437, opened 2025-06-12
Add tests for webauthn configuration AvoidSameAuthenticatorRegister

### LINKED ISSUE #27945: Passkey "Avoid same authenticator registration" doesn't work
### Before reporting an issue

### Area

authentication/webauthn

### Describe the bug

The "Avoid same authenticator registration" setting (under Authentication > Webauthn Passwordless Policy) doesn't seem to work when registering new passkey credentials.

### Version

24.0.1 (also observed in 23.0.6)

### Regression

### Expected behavior

Enabling the setting should prevent the WebAuthn Authenticator from creating a new credential and also avoid saving a duplicate credential in the WebAuthn Relying Party (i.e. Keycloak).

### Actual behavior

Enabling the setting does not prevent the WebAuthn Authenticator from creating a new credential and also saves a duplicate credential in the WebAuthn Relying Party (i.e. Keycloak).

### How to Reproduce?

1. Enable the "Avoid same authenticator registration" setting under Authentication > Webauthn Passwordless Policy.
2. Set up a passkey for a new/existing user (either from account console or via a required action). **Prerequisite:** You must have the "WebAuthn Register Passwordless" required action enabled for the realm in Authentication tab.
3. Set up another passkey for the user (either from account console or via a required action).
4. Notice a duplicate passkey is created (in both the authenticator and in Keycloak).

### Anything else?

The expectation is based on the description of [excludeCredentials](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/create#excludecredentials) (see also https://web.dev/articles/webauthn-exclude-credentials) which this setting seems to configure.

**Browser:** macOS Safari Version 17.4

**Authenticators:**
- Mac running macOS 14.4 (23E214)
- iPhone running iOS 17.3.1 acting as a cross-device authenticator for the above Mac.

Did not check other authenticators.

All other settings in Webauthn Passwordless Policy are set to their default values.

rmartinc:
Closes #27945

Testing issue #27945 I detected that we didn't have tests for that. So I'm adding two little tests in the WebAuthn testsuite. Tests are working as expected so the issue is not reproduced or it was fixed previously.

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/webauthn/account/AbstractWebAuthnAccountTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/webauthn/account/AbstractWebAuthnAccountTest.java
@@ -42,6 +42,7 @@ import org.keycloak.testsuite.webauthn.AbstractWebAuthnVirtualTest;
 import org.keycloak.testsuite.webauthn.authenticators.DefaultVirtualAuthOptions;
 import org.keycloak.testsuite.webauthn.authenticators.UseVirtualAuthenticators;
 import org.keycloak.testsuite.webauthn.authenticators.VirtualAuthenticatorManager;
+import org.keycloak.testsuite.webauthn.pages.WebAuthnErrorPage;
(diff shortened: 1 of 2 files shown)


mabartos (approved):
LGTM

PR #40442, opened 2025-06-12
Add templates for release notes and migration guide

mabartos:
Closes #40441

@ahus1 @mhajas As stated in the comment https://github.com/keycloak/keycloak/pull/40424#issuecomment-2965645999, WDYT?


ahus1 (approved):
Thank you for this change. I'd say it might evolve with 1-2 more commits, and with discussion in the team.

PR #40450, opened 2025-06-12
shows what patching logic should look like

shawkins:
also shows basic field validation

Reopening against the right branch...

@vmuzikar @mabartos @Pepo48 this shows roughly what json merge patch support looks like - there is very little we need to do as Jackson has support already baked in. There are several TODOs - moving field validation to a common spot, using a common objectmapper, agreeing on the response codes, etc.

The server side method to allow for other patch types has the payload typed as JsonNode (if that doesn't work, then we'd just use String).

From a rest client perspective you could have a typed method. The user will do something like:

```
ClientRepresentation rep = new ClientRepresentation();
rep.setEnabled(false);
rep.setAdditionalField("description", null); // remove description

clients.getClient("id").patchClient(rep, FieldValidation.Strict);
```

@mabartos suggested adding a convienence method instead of using a direct put - we can certainly do that. In general we'll need to make sure the additional field stuff is well documented as to when / why you would do that. For all the known fields you can even regular methods like "rep.unsetDescription()" as was shown in the defaults pr. However we still do need to expose the general map / set method for additional scenarios.

From the initial / incorrect pr:

> Is there any intention to store metadata in the additional fields?

No. At this point it is strictly for cross-version compatibility and for having a typed object model for creating json merge patches that can remove values. If we want a general metadata facility, like ObjectMeta in Kuberentes, that would need to be it's own construct and properly stored.

### DIFF
--- a/rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
+++ b/rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
@@ -1,14 +1,31 @@
 package org.keycloak.admin.api.client;
 
+import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.PUT;
+import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
-import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
+
+import org.keycloak.admin.api.FieldValidation;
 import org.keycloak.representations.admin.v2.ClientRepresentation;
+
+import com.fasterxml.jackson.databind.JsonNode;
 
 public interface ClientApi {
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     ClientRepresentation getClient();
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation createOrUpdateClient(ClientRepresentation client, @PathParam("fieldValidation") FieldValidation fieldValidation);
+
+    @PUT
+    @Consumes("application/merge-patch+json")
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation patchClient(JsonNode patch, @PathParam("fieldValidation") FieldValidation fieldValidation);
+
 }
--- a/rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
+++ b/rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
@@ -2,6 +2,7 @@ package org.keycloak.admin.api.client;
 
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.PATCH;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
@@ -23,7 +24,7 @@ public interface ClientApi {
     @Produces(MediaType.APPLICATION_JSON)
     ClientRepresentation createOrUpdateClient(ClientRepresentation client, @PathParam("fieldValidation") FieldValidation fieldValidation);
 
-    @PUT
+    @PATCH
     @Consumes("application/merge-patch+json")
     @Produces(MediaType.APPLICATION_JSON)
     ClientRepresentation patchClient(JsonNode patch, @PathParam("fieldValidation") FieldValidation fieldValidation);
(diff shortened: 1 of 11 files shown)


rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
```diff
@@ -1,14 +1,31 @@
...
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     ClientRepresentation getClient();
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation createOrUpdateClient(ClientRepresentation client, @PathParam("fieldValidation") FieldValidation fieldValidation);
+
+    @PUT
```
vmuzikar: Shouldn't this rather be `@PATCH`?

rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
```diff
@@ -1,14 +1,31 @@
...
+    @Consumes("application/merge-patch+json")
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation patchClient(JsonNode patch, @PathParam("fieldValidation") FieldValidation fieldValidation);
```
vmuzikar, line 30: Field validation might need a better name not to be confused with regular validations (from semantic perspective etc.). Possibly something like `unknownFieldsPolicy` but not sure.

Also should we really give the power to set this policy to the user? Shouldn't it rather be a global server option? In other words, do we really want to set this per request?

services/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -36,14 +36,28 @@ public Stream<ClientRepresentation> getClients(RealmModel realm, ClientProjectio
...
-            throw new ServiceException("Client already exists", Response.Status.CONFLICT);
+    public CreateOrUpdateResult createOrUpdate(RealmModel realm, ClientRepresentation client, boolean allowUpdate)
+            throws ServiceException {
+        boolean created = false;
+        ClientModel model = realm.getClientByClientId(client.getClientId());
+        if (model != null) {
+            if (!allowUpdate) {
+                throw new ServiceException("Client already exists", Response.Status.CONFLICT);
+            }
+            // TODO: make sure that removal is the best way to do this
```
vmuzikar: It's probably not. Just from the top of my head, we might accidentally delete some entities with foreign keys. But there might be more caveats.

rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
shawkins: > Field validation might need a better name not to be confused with regular validations (from semantic perspective etc.)

It is possible to tie this to "optional" semantic validation - for example in kubernetes the handling of the StatefulSet serviceName - it is not required unless fieldValidation=Strict.

> Possibly something like unknownFieldsPolicy but not sure.

Yes the naming can be whatever we want.

> Also should we really give the power to set this policy to the user? Shouldn't it rather be a global server option? In other words, do we really want to set this per request?

You would probably want control per request yes. You can certainly make a case for session or server level control as well.

services/src/main/java/org/keycloak/services/client/DefaultClientService.java
shawkins: As long as the overlay of the model state is complete, we won't need to use a delete - based upon the current mapper prototype there will need to be some changes to support that.

rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
shawkins: Yes, I'll make that change.

shawkins:
@vmuzikar made some initial updates based upon the reivew. I'll see about starting some kind of unit / integration tests as well.

rest/admin-api/src/main/java/org/keycloak/admin/api/client/ClientApi.java
vmuzikar: > Yes the naming can be whatever we want.

Thinking about that more, maybe `fieldNameValidation` would be better? WDYT?

> You would probably want control per request yes. You can certainly make a case for session or server level control as well.

Ok. We can add the server level one as a follow-up.

services/src/main/java/org/keycloak/services/client/DefaultClientService.java
vmuzikar: Right. But we can tackle that as a follow-up.

vmuzikar:
> made some initial updates based upon the reivew.

Thanks.

> I'll see about starting some kind of unit / integration tests as well.

Do you want to address it in this PR, or in a follow-up. If the former, we can merge from my PoV.

vmuzikar (approved):

shawkins:
> Do you want to address it in this PR, or in a follow-up. If the former, we can merge from my PoV.

Sorry, I haven't gotten to this yet. Let's go ahead an merge so that we don't have more conflicts later.

PR #40458, opened 2025-06-12
Additional Quarkus devtools dependencies in distribution

### LINKED ISSUE #39227: Quarkus devtools dependencies in 26.2.x
### Before reporting an issue

### Area

dependencies

### Describe the bug

Hello,

Comparing the contents of `lib/lib/deployment` between Keycloak 26.1.5 and Keycloak 26.2.0, I see a few more JAR files:
```diff
--- deployment-list-26.1.5.txt  2025-04-25 14:57:26.998465900 +0000
+++ deployment-list-26.2.0.txt  2025-04-25 14:57:18.444701000 +0000
@@ -5,6 +5,7 @@
 lib/lib/deployment/io.quarkus.http.quarkus-http-http-core
 lib/lib/deployment/io.quarkus.http.quarkus-http-servlet
 lib/lib/deployment/io.quarkus.quarkus-agroal-deployment
+lib/lib/deployment/io.quarkus.quarkus-agroal-dev
 lib/lib/deployment/io.quarkus.quarkus-agroal-spi
 lib/lib/deployment/io.quarkus.quarkus-arc-deployment
 lib/lib/deployment/io.quarkus.quarkus-arc-test-supplement
@@ -17,6 +18,10 @@
 lib/lib/deployment/io.quarkus.quarkus-credentials-deployment
 lib/lib/deployment/io.quarkus.quarkus-datasource-deployment
 lib/lib/deployment/io.quarkus.quarkus-datasource-deployment-spi
+lib/lib/deployment/io.quarkus.quarkus-devtools-base-codestarts
+lib/lib/deployment/io.quarkus.quarkus-devtools-common
+lib/lib/deployment/io.quarkus.quarkus-devtools-message-writer
+lib/lib/deployment/io.quarkus.quarkus-devtools-registry-client
 lib/lib/deployment/io.quarkus.quarkus-devtools-utilities
 lib/lib/deployment/io.quarkus.quarkus-grpc-common-deployment
 lib/lib/deployment/io.quarkus.quarkus-hibernate-orm-deployment
@@ -70,13 +75,17 @@
 lib/lib/deployment/io.quarkus.qute.qute-core
 lib/lib/deployment/io.quarkus.resteasy.reactive.resteasy-reactive-common-processor
 lib/lib/deployment/io.quarkus.resteasy.reactive.resteasy-reactive-processor
+lib/lib/deployment/io.smallrye.common.smallrye-common-version
 lib/lib/deployment/io.smallrye.jandex
 lib/lib/deployment/io.smallrye.smallrye-health-ui
 lib/lib/deployment/io.smallrye.smallrye-open-api-core
+lib/lib/deployment/io.smallrye.smallrye-open-api-model
 lib/lib/deployment/jakarta.servlet.jakarta.servlet-api
 lib/lib/deployment/jakarta.validation.jakarta.validation-api
 lib/lib/deployment/org.aesh.aesh
 lib/lib/deployment/org.aesh.readline-2.6.jar
+lib/lib/deployment/org.apache.commons.commons-compress
+lib/lib/deployment/org.codejive.java-properties
 lib/lib/deployment/org.eclipse.microprofile.openapi.microprofile-openapi-api
 lib/lib/deployment/org.fusesource.jansi.jansi
 lib/lib/deployment/org.graalvm.sdk.nativeimage
```

Most of them have `devtools` or `dev` in their names, so this looks like a mistake.
It seems #38665 tried to remove some dependencies, but maybe it did not go far enough.

Is it expected to have all these new dependencies in the Keycloak release?

Thanks in advance

### Version

26.2.0

### Regression

### Expected behavior

No extra dev dependencies compared to previous releases

### Actual behavior

See description

### How to Reproduce?

Look at the contents of `lib/lib/deployment` in the Keycloak release tarballs.

### Anything else?

Reporting this only because our SBOM analyser is now reporting gradle-wrapper, which we don't usually expect in production packages.
It was found inside one of these new JAR files:
`io.quarkus.quarkus-devtools-base-codestarts-3.20.0.jar:codestarts/quarkus/tooling/gradle-wrapper/base/gradle/wrapper/gradle-wrapper.jar`
So looking for confirmation that this is all as expected from Keycloak maintainers point of view.

mabartos:
Closes #39227

- As we are not currently able to exclude dev dependencies completely, we need to be more selective here: https://github.com/quarkusio/quarkus/issues/47210
- Some basic approaches on how to exclude these dev dependencies will be part of Quarkus 3.23.x, and some other changes later: https://github.com/quarkusio/quarkus/pull/45053
- The only `io.quarkus:quarkus-devtools-common` cannot be excluded from the distribution, but its transitive dependencies can
- Create script that checks the distribution dependency tree and warn us if there are some more - which might be useful even for Quarkus upgrades
Screenshot From 2025-06-12 17-25-30
- I'll create a follow-up issue to track progress on Quarkus side and then removing all these additional excludes

@shawkins @vmuzikar Could you please check it? Thanks!

### DIFF
--- a/quarkus/deployment/pom.xml
+++ b/quarkus/deployment/pom.xml
@@ -34,6 +34,10 @@
                     <groupId>org.eclipse.sisu</groupId>
                     <artifactId>org.eclipse.sisu.inject</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-devtools-utilities</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
@@ -76,6 +80,36 @@
                     <groupId>io.quarkus</groupId>
                     <artifactId>quarkus-devtools-codestarts</artifactId>
                 </exclusion>
+                <!-- Exclude transitive dependencies for 'io.quarkus:quarkus-devtools-common' BEGIN -->
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-devtools-registry-client</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-devtools-utilities</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-devtools-base-codestarts</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.codejive</groupId>
+                    <artifactId>java-properties</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-compress</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.smallrye.common</groupId>
+                    <artifactId>smallrye-common-version</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.dataformat</groupId>
+                    <artifactId>jackson-dataformat-yaml</artifactId>
+                </exclusion>
+                <!-- Exclude transitive dependencies for 'io.quarkus:quarkus-devtools-common' END -->
             </exclusions>
         </dependency>
         <dependency>
@@ -97,6 +131,10 @@
                 <exclusion>
                     <groupId>io.quarkus</groupId>
                     <artifactId>quarkus-devservices-deployment</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-agroal-dev</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
@@ -195,6 +233,12 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit5-internal</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-devtools-utilities</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
--- a/quarkus/dist/check-dev-dependencies.sh
+++ b/quarkus/dist/check-dev-dependencies.sh
@@ -1,0 +1,5 @@
+#!/bin/bash -e
+
+if mvn dependency:tree -Dverbose | grep -E '(-dev:|-devtools-)'; then
+  echo "[WARNING] Detected development dependencies in the build tree."
+fi
--- a/quarkus/dist/pom.xml
+++ b/quarkus/dist/pom.xml
@@ -111,4 +111,35 @@
         </plugins>
     </build>
 
+    <profiles>
+        <profile>
+            <id>check-dev-dependencies</id>
+            <activation>
+                <os>
+                    <family>unix</family>
+                </os>
+            </activation>
+            <build>
+                <plugins>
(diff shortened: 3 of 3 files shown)


mabartos:
Based on the info in https://github.com/keycloak/keycloak/issues/39227

Before:
```
> ll lib/lib/deployment| grep -E '(agroal-dev|base-codestarts|devtools-common|message-writer|registry-client|smallrye-common-version|smallrye-open-api-model|commons-compress|java-properties)'

-rw-r--r--. 1 mabartos mabartos  20K May 13 11:59 io.quarkus.quarkus-agroal-dev-3.20.1.jar
-rw-r--r--. 1 mabartos mabartos 395K May 13 11:51 io.quarkus.quarkus-devtools-base-codestarts-3.20.1.jar
-rw-r--r--. 1 mabartos mabartos 412K May 13 12:08 io.quarkus.quarkus-devtools-common-3.20.1.jar
-rw-r--r--. 1 mabartos mabartos  13K May 13 11:42 io.quarkus.quarkus-devtools-message-writer-3.20.1.jar
-rw-r--r--. 1 mabartos mabartos 223K May 13 11:46 io.quarkus.quarkus-devtools-registry-client-3.20.1.jar
-rw-r--r--. 1 mabartos mabartos  34K Apr 11 16:04 io.smallrye.common.smallrye-common-version-2.12.0.jar
-rw-r--r--. 1 mabartos mabartos  33K Apr 17 16:55 io.smallrye.smallrye-open-api-model-4.0.10.jar
-rw-r--r--. 1 mabartos mabartos 1.1M Aug 17  2024 org.apache.commons.commons-compress-1.27.1.jar
-rw-r--r--. 1 mabartos mabartos  27K Jul 22  2024 org.codejive.java-properties-0.0.7.jar

```

After:
```
> ll lib/lib/deployment| grep -E '(agroal-dev|base-codestarts|devtools-common|message-writer|registry-client|smallrye-common-version|smallrye-open-api-model|commons-compress|java-properties)'

-rw-r--r--. 1 mabartos mabartos 412K May 13 12:08 io.quarkus.quarkus-devtools-common-3.20.1.jar
-rw-r--r--. 1 mabartos mabartos  33K Apr 17 16:55 io.smallrye.smallrye-open-api-model-4.0.10.jar
```

shawkins:
> Could you please check it? Thanks!

This all makes sense. Just wondering about the warning - who will be looking at that, and / or should it be an error instead?

mabartos:
@shawkins Thanks for the review!

> who will be looking at that?

At least we in our team who quite often build the Quarkus distribution as we might clearly see if the list is wider than before (IMHO better than not checking it at all). 

> and / or should it be an error instead?

As we cannot remove all these dev dependencies for now, I would not vote for hard error and build failure. It's even better to only warn about the usage of dev dependencies for the project https://github.com/mabartos/keycloak-quarkus-extensions, where people can add some additional dev dependencies intentionally.

But if we decide to not add any checker, I'm ok with that, but we'd need some other approach to check it to prevent these issues again.

mabartos:
@shawkins @vmuzikar WDYT related to the script? Is it ok to keep it, or you don't see the value here? We would have at least something.

Are you ok to exclude these dependencies? We mitigate the issue at least.

shawkins (approved):
LGTM, seems fine to just keep it as is - especially if it looks like we'll be able to get a proper fix from quarkus later.

vmuzikar:
> WDYT related to the script? Is it ok to keep it, or you don't see the value here? We would have at least something.

As discussed offline, the main problem with the dev dependencies is prod related. And that problem is not specific to the dev dependencies only – there are others which this script won't catch. So it somewhat helps but it's not the ultimate solution for that. We'll still need to consult our internal prod reports.

That said, it was not a blocker for me and it's ok from my perspective to keep the script, albeit it adds some time when building. No need to revert that as it's merged now. :)

PR #40484, opened 2025-06-13
Disabling persistent sessions affects also offline session cache

### LINKED ISSUE #40483: Missing adjustment about offline session caches for volatile sessions
### Before reporting an issue

### Area

infinispan

### Describe the bug

Disabling persistent sessions also triggers the old code for offline sessions, which is not safe to run with a limited amount of sessions.

This is a follow-up to #40472

### Version

main

### Regression

### Expected behavior

Offline session handling should work as before

### Actual behavior

Offline sessions loaded from the database are evicted and might fail the validation

### How to Reproduce?

There has been a user report about this.

### Anything else?

_No response_

ahus1:
Closes #40483

### DIFF
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -132,7 +132,6 @@ The `actionTokens` distributed cache is used to track metadata about action toke
 By default, regular user sessions are stored in the database and loaded on-demand to the cache.
 It is possible to configure {project_name} to store regular user sessions in the cache only and minimize calls to the database.
 
-
 Since all the sessions in this setup are stored in-memory, there are two side effects related to this:
 
 * Losing sessions when all {project_name} nodes restart.
@@ -147,6 +146,7 @@ Similarly, it's crucial that we replicate each entry to two nodes to prevent dat
 When using volatile user sessions, the `max-count` setting for the offline session is ignored.
 
 It is not recommended to use volatile user sessions when using offline sessions extensively due to potentially high memory usage.
+For volatile sessions, the time offline sessions are cached in memory can be limited with the SPI options `spi-user-sessions--infinispan--offline-client-session-cache-entry-lifespan-override` and `spi-user-sessions--infinispan--offline-session-cache-entry-lifespan-override`.
 ====
 
 Follow these steps to enable this setup:
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -144,9 +144,9 @@ Similarly, it's crucial that we replicate each entry to two nodes to prevent dat
 
 [WARNING]
 ====
-When using volatile user sessions, you must not use a `max-count` setting for the offline session caches as this leads to a loss of data due to the current implementation even though the session data is saved in the database. Therefore {project_name} ignores this setting if volatile user sessions are enabled, and prints a warning.
+When using volatile user sessions, the `max-count` setting for the offline session is ignored.
 
-It is not recommended to use volatile user sessions when using offline sessions extensively due the high memory usage.
+It is not recommended to use volatile user sessions when using offline sessions extensively due to potentially high memory usage.
 ====
 
 Follow these steps to enable this setup:
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -138,10 +138,15 @@ Since all the sessions in this setup are stored in-memory, there are two side ef
 * Losing sessions when all {project_name} nodes restart.
 * Increased memory consumption.
 
-Using volatile user sessions also disables an improved handling of offline sessions, where only a limited number of offline sessions needs to be loaded into memory.
-Due to this, it is not recommended to use volatile user sessions when using offline sessions extensively. apply the settings below not only to the regular session caches but also the offline session caches.
+When using volatile user sessions, the cache is the source of truth for user and client sessions.
+Therefore, it is necessary to not define an upper limit on the total number of entries that can be stored via the setting `max-count`.
+Similarly, it's crucial that we replicate each entry to two nodes to prevent data loss on node failures and rolling restarts.
 
-When using volatile user sessions, the cache is the source of truth for user and client sessions. Therefore, it is necessary to not define an upper limit on the total number of entries that can be stored. Similarly, it's crucial that we replicate each entry to at least two nodes to prevent data loss on node failures and rolling restarts.
+[WARNING]
+====
+When using volatile user sessions, you must not use a `max-count` setting for the offline session caches as this leads to a loss of data due to the current implementation even though the session data is saved in the database.
+It is not recommended to use volatile user sessions when using offline sessions extensively due the high memory usage.
+====
 
 Follow these steps to enable this setup:
 
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -144,7 +144,8 @@ Similarly, it's crucial that we replicate each entry to two nodes to prevent dat
 
 [WARNING]
 ====
-When using volatile user sessions, you must not use a `max-count` setting for the offline session caches as this leads to a loss of data due to the current implementation even though the session data is saved in the database.
+When using volatile user sessions, you must not use a `max-count` setting for the offline session caches as this leads to a loss of data due to the current implementation even though the session data is saved in the database. Therefore {project_name} ignores this setting if volatile user sessions are enabled, and prints a warning.
+
 It is not recommended to use volatile user sessions when using offline sessions extensively due the high memory usage.
 ====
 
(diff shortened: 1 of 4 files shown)


docs/guides/server/caching.adoc
```diff
@@ -184,8 +188,9 @@ Each distributed cache, that is a primary source of truth of the data (`authenti
 Non-owner nodes query the owners of a specific cache to obtain data.
 When both owner nodes are offline, all data is lost.
 
-The default number of owners is enough to survive 1 node (owner) failure in a cluster setup with at least three nodes. You are free
-to change the number of owners accordingly to better fit into your availability requirements. To change the number of owners, open `conf/cache-ispn.xml` and change the value for `owners=<value>` for the distributed caches to your desired value.
+The default number of two owners is the minimum number is necessary to survive one node (owner) failure or a rolling restart in a cluster setup with at least two nodes. When one of the owners becomes unavailable, the data is restored from the remaining owner and rebalanced across the remaining nodes.
+A higher number could increase the availability of the data, but would lead to slower writes as more nodes need to be updated.
```
ryanemerson: "A higher number increases the availability of the data, but at the expense of slower writes as more nodes need to be updated."

docs/guides/server/caching.adoc
```diff
@@ -108,21 +108,42 @@ CPU, memory, and network utilization.
 
 These in-memory caches for user sessions and client sessions are limited to, by default, 10000 entries per node which reduces the overall memory usage of {project_name} for larger installations.
 The internal caches will run with only a single owner for each cache entry.
-Consider trade-off between memory consumption and the database utilization and set different sizes for the caches, edit {project_name}'s cache config file (`conf/cache-ispn.xml`) to set a `+<memory max-count="..."/>+` for those caches.
 
-.Volatile user sessions
+.Offline user sessions
+As an OpenID Connect Provider, the server is also capable of authenticating users and issuing offline tokens. Similarly to regular user and client sessions,
+when an offline token is issued by the server upon successful authentication, the server also creates an offline user session and an offline client session.
```
ryanemerson: I think this is clearer if we remove the comparison to !offline sessions and simplify this to:

"As an OpenID Connect Provider, the server is capable of authenticating users and issuing offline tokens. When issuing an offline token after successful authentication, the server creates an offline user session and offline client session."

docs/guides/server/caching.adoc
```diff
@@ -108,21 +108,42 @@ CPU, memory, and network utilization.
...
 
-By default, user sessions are stored in the database and loaded on-demand to the cache.
-It is possible to configure {project_name} to store user sessions in the cache only and minimize the database utilization.
+The following caches are used to store offline sessions:
+
+* offlineSessions
+* offlineClientSessions
+
+Similarly to regular user and client sessions caches, also the offline user and client session caches are limited to 10000 entries per node by default. Items which are evicted from the memory will be loaded on-demand from the database when needed.
```
ryanemerson: "Similar to the user and client session caches, the offline user ..."

or

"Like the user and client  session caches, the offline user ..."

docs/guides/server/caching.adoc
```diff
@@ -108,21 +108,42 @@ CPU, memory, and network utilization.
...
+This cache is needed for the Brute Force Protection feature to work in a multi-node {project_name} setup.
+
+.Action tokens
+Action tokens are used for scenarios when a user needs to confirm an action asynchronously, for example in the emails sent by the forgot password flow.
+The `actionTokens` distributed cache is used to track metadata about action tokens.
+
+=== Volatile user sessions
+
+By default, regular user sessions are stored in the database and loaded on-demand to the cache.
+It is possible to configure {project_name} to store regular user sessions in the cache only and minimize the database utilization.
```
ryanemerson: "minimise calls to the database"?

docs/guides/server/caching.adoc
```diff
@@ -108,21 +108,42 @@ CPU, memory, and network utilization.
...
 
 Since all the sessions in this setup are stored in-memory, there are two side effects related to this:
 
-* Losing sessions on when all {project_name} nodes restart.
+* Losing sessions when all {project_name} nodes restart.
 * Increased memory consumption.
 
+Using volatile user sessions also disables an improved handling of offline sessions, where only a limited number of offline sessions needs to be loaded into memory.
+Due to this, apply the settings below not only to the regular session caches but also the offline session caches.
```
ryanemerson: We're saying the same thing multiple times here. I think we can make this section clearer by stating that the cache cannot have an upper bound here and then in the steps just say what needs to be changed. Something like:

"When using volatile user sessions, the cache is the source of truth for user and client sessions, therefore it is necessary to not define an upper limit on the total number of entries that can be stored. Similarly, it's crucial that we replicate each entry to at least two nodes to prevent data loss on node failures and rolling restarts.

Follow these steps to enable this setup:

1. Edit {project_name}'s cache config file (`conf/cache-ispn.xml`) for caches `sessions`, `clientSessions`, `offlineSessions` and `offlineClientSessions` with the following update:

"

ryanemerson (changes requested):
Thanks @ahus1, +1 to restricting the configuration in this way. Just a few comments regarding the docs.

docs/guides/server/caching.adoc
ahus1: There is the unfortunate interconnection that people need to change the regular and offline cache configuration. So I would like to keep that one paragraph, but moved it a bit higher. Please have a look.

ahus1:
@ryanemerson - I've address your comments in a second commit, still the last comment might need more discussion. Please re-review.

docs/guides/server/caching.adoc
```diff
@@ -108,21 +108,44 @@ CPU, memory, and network utilization.
...
-It is possible to configure {project_name} to store user sessions in the cache only and minimize the database utilization.
 
 Since all the sessions in this setup are stored in-memory, there are two side effects related to this:
 
-* Losing sessions on when all {project_name} nodes restart.
+* Losing sessions when all {project_name} nodes restart.
 * Increased memory consumption.
 
+Using volatile user sessions also disables an improved handling of offline sessions, where only a limited number of offline sessions needs to be loaded into memory.
+Due to this, it is not recommended to use volatile user sessions when using offline sessions extensively. apply the settings below not only to the regular session caches but also the offline session caches.
```
ryanemerson: I think the first sentence here is redundant. In the context of volatile user sessions, we're not disabling an improvement, just ensuring that the semantics are correct for safe usage. I also think the order of paragraphs is now confusing.

How about:

```
* Increased memory consumption.

When using volatile user sessions, the cache is the source of truth for user and client sessions. Therefore, it is necessary to not define an upper limit on the total number of entries that can be stored. Similarly, it's crucial that we replicate each entry to at least two nodes to prevent data loss on node failures and rolling restarts.


[NOTE]
====
As all sessions are stored in-memory, it is not recommend to use volatile user sessions when using offline sessions extensively.
====

Apply the below settings to both the regular and offline session caches.
...
```

ahus1: I pushed another commit. With volatile sessions enabled, offline sessions are stored in the database, but the current implementation leads to a data loss. 

https://github.com/keycloak/keycloak/blob/4b47697c838b9c7373610d826f68d0c0ed8fb635/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java#L297-L299

ryanemerson (approved):

docs/guides/server/caching.adoc
```diff
@@ -108,21 +108,50 @@ CPU, memory, and network utilization.
...
 
+When using volatile user sessions, the cache is the source of truth for user and client sessions.
+Therefore, it is necessary to not define an upper limit on the total number of entries that can be stored via the setting `max-count`.
+Similarly, it's crucial that we replicate each entry to two nodes to prevent data loss on node failures and rolling restarts.
+
+[WARNING]
+====
+When using volatile user sessions, the `max-count` setting for the offline session is ignored.
+
+It is not recommended to use volatile user sessions when using offline sessions extensively due to potentially high memory usage.
```
mhajas, line 148: Should we document here the offline session lifespan override option?

https://github.com/keycloak/keycloak/blob/160ec6ba0d9171bb0dc4f882126f55128eb486e4/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java#L260-L294

mhajas (commented):
Thanks for the PR, I added one comment.

ahus1:
@mhajas - I've addressed your review comments in another set of commits. Please review and approve when you have the time. Thanks!

mhajas (approved):
Thanks for the addition @ahus1! LGTM

PR #40490, opened 2025-06-13
Preserve query parameters when redirecting requests

### LINKED ISSUE #40489: When redirecting old resource versions, keep query parameters
### Description

The redirect should contain all elements, including query parameters

### Discussion

_No response_

### Motivation

While we not use them, some might use query parameters for cache busting or other items. Also, we should have a test to ensure that URL encoding works as expected.

### Details

I'll create a PR

ahus1:
Closes #40489

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/ThemeResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/ThemeResource.java
@@ -23,7 +23,6 @@ import jakarta.ws.rs.Path;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.QueryParam;
-import jakarta.ws.rs.core.CacheControl;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.UriBuilder;
@@ -58,6 +57,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
@@ -72,6 +72,7 @@ import static java.util.stream.Collectors.toSet;
 public class ThemeResource {
 
     private static final Logger log = Logger.getLogger(ThemeResource.class);
+    private static final Pattern RESOURCE_TAG_PATTERN = Pattern.compile("[0-9a-z]{5}");
 
     @Context
     private KeycloakSession session;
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/keycloak/services/resources/ThemeResource.java
```diff
@@ -100,19 +105,43 @@ public Response getResource(@PathParam("version") String version, @PathParam("th
...
 
             if (Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
+                // A simpler way to check for encoded URL characters would be to retrieve the raw values.
+                // Unfortunately, RESTEasy doesn't support this, and UrlInfo will throw an IllegalArgumentException.
+
+                String base = uriInfo.getBaseUri().getPath();
+                base = base.substring(0, base.length() - 1);
+                if (!uriInfo.getRequestUri().toURL().getPath().startsWith(base + UriBuilder.fromResource(ThemeResource.class)
+                        .path("/{version}/").build(version).getPath())) {
+                    log.debugf("No URL encoding should be necessary for the version, returning a 404: %s", uriInfo.getRequestUri().getPath());
```
ryanemerson: My understanding is that RestEasy will always return the `version` in decoded form. I'm struggling to understand scenarios where we need this check, can you provide an example?

ahus1: I am trying to prevent situations where someone passes in strange versions in, which then lead to open an open or half-open redirect.

Imagine a version like `myversion/../../auth/realm/xxx` which is then url-encoded as `myversion%2F..%2F..%2Fauth%2Frealm%2Fxxx` and passed in as the `version` URL fragment. 

So I'd say everything that is encoded is possibly someone trying to exploit it. [And as a version, we only use characters and numbers at the moment.](https://github.com/keycloak/keycloak/blob/7cff8572387e6bbfca160aa60e9a8191d22e5c85/model/jpa/src/main/java/org/keycloak/models/jpa/MigrationModelAdapter.java#L87-L91) I could add that as a regex check in addition, or as a replacement. 

Still, IMHO we need the encoding check for the path.

ryanemerson: Thanks for explaining.

Do we need this first check still given that we only redirect `if (!version.equals(Version.RESOURCES_VERSION) && !hasContentHash)`, won't L123 protect against the scenario you describe?

ahus1: Yes and no: For that branch, it is. But then there is also the part in line 148, and the fallthrough where there is a content hash and the resource is returned to the caller starting in line 156. For that item, I also don't want to accept just any version.

ryanemerson: In L148 we're not checking against the version provided in the url though, we're only comparing the `etag` and `Version.RESOURCES_VERSION`. So I don't *think* an encoded path version can be exploited in this case as a 304 is sent.

Either way, happy to keep the check if you prefer. I do think a regex validating the version would be more explicit though. Adding some additional comments explaining why we're performing these checks would be nice for other contributors.

services/src/main/java/org/keycloak/services/resources/ThemeResource.java
```diff
@@ -100,19 +105,43 @@ public Response getResource(@PathParam("version") String version, @PathParam("th
...
                     // The redirect will lead the browser to a resource that it then (when retrieved successfully) can cache again.
                     // This assumes that it is better to try to some content even if it is outdated or too new, instead of returning a 404.
                     // This should usually work for images, CSS or (simple) JavaScript referenced in the login theme that needs to be
                     // loaded while the rolling restart is progressing.
                     return Response.temporaryRedirect(
-                            UriBuilder.fromResource(ThemeResource.class)
-                                    .path("/{version}/{themeType}/{themeName}/{path}")
-                                    // The 'path' can contain slashes, so encoding of slashes is set to false
-                                    .build(new Object[]{Version.RESOURCES_VERSION, themeType, themeName, path}, false)
+                            redirectUri
```
ryanemerson: Nitpick, but we can fix the formatting here to be a one-liner or:

```java
                    return Response
                          .temporaryRedirect(redirectUri)
                          .build();
```

ahus1:
Thank you. Your review comments should now be addressed.

ahus1:
I see the CI is failing ... looking at it now.

ryanemerson (approved):

ahus1:
@ryanemerson - ok, test is no fixed. Sorry for the noise. Please re-review. Thanks!

ryanemerson (approved):

ahus1:
@mhajas - can you please approve based on Ryan's review? Thanks!

mhajas (approved):
Approving based on the review from @ryanemerson

PR #40491, opened 2025-06-13
Update flows.adoc

minisu:
Closes #40514

Update authentication flows documentation to match new GUI, which uses drop-down menus:

Screenshot 2025-06-14 at 09 55 06


ahus1 (approved):
Thank you for taking the time to update the docs!

PR #40493, opened 2025-06-14
Fix for "Account UI ignores identity provider display order"

Ekamdeep3044:
Fix identity provider UI order bug

Ensures identity providers are displayed in the correct order
by GUI priority fetched in API when rendering the Account UI list.

Closes #40461

### DIFF
--- a/core/src/main/java/org/keycloak/representations/account/LinkedAccountRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/account/LinkedAccountRepresentation.java
@@ -16,7 +16,6 @@
  */
 package org.keycloak.representations.account;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 
 /**
  *
@@ -29,8 +28,6 @@ public class LinkedAccountRepresentation implements Comparable<LinkedAccountRepr
     private String providerName;
     private String displayName;
     private String linkedUsername;
-
-    @JsonIgnore
     private String guiOrder;
 
     public String getLinkedUsername() {
--- a/js/apps/account-ui/src/account-security/LinkedAccounts.tsx
+++ b/js/apps/account-ui/src/account-security/LinkedAccounts.tsx
@@ -86,7 +86,9 @@ export const LinkedAccounts = () => {
           />
           <DataList id="linked-idps" aria-label={t("linkedLoginProviders")}>
             {linkedAccounts.length > 0 ? (
-              linkedAccounts.map(
+              linkedAccounts.sort(
+                  (firstIdp, secondIdp) => Number(firstIdp.guiOrder ?? Number.MAX_SAFE_INTEGER) - Number(secondIdp.guiOrder ?? Number.MAX_SAFE_INTEGER)
+                  ).map(
                 (account, index) =>
                   index !== paramsLinked.max - 1 && (
                     <AccountRow
@@ -140,7 +142,9 @@ export const LinkedAccounts = () => {
           />
           <DataList id="unlinked-idps" aria-label={t("unlinkedLoginProviders")}>
             {unlinkedAccounts.length > 0 ? (
-              unlinkedAccounts.map(
+              unlinkedAccounts.sort(
+                  (firstIdp, secondIdp) => Number(firstIdp.guiOrder ?? Number.MAX_SAFE_INTEGER) - Number(secondIdp.guiOrder ?? Number.MAX_SAFE_INTEGER)
+              ).map(
                 (account, index) =>
                   index !== paramsUnlinked.max - 1 && (
                     <AccountRow
--- a/js/apps/account-ui/src/api/representations.ts
+++ b/js/apps/account-ui/src/api/representations.ts
@@ -66,6 +66,7 @@ export interface LinkedAccountRepresentation {
   displayName: string;
   linkedUsername: string;
   social: boolean;
+  guiOrder: string;
 }
 
 export interface SessionRepresentation {


ahus1 (changes requested):
Thank you for submitting this PR. IMHO this change is incomplete, as it will only sort the current page of results returned from the backend. 

As long as there are only 10 IDPs, this is fine. Once there are more, this will lead to wrong results. Let's discuss in the issue possible alternatives.

ahus1:
IMHO the needs to be solved in the backend before the pagination is done. This would require `JpaIdentityProviderStorageProvider` to return a list already sorted via the GUI order. 

Doing this change, the GUI order would be used also to list the IDPs in the Admin UI. 

Unfortunately the GUI order is not a table column, but a value in the attribute. I wonder what the Keycloak IAM team thinks about this, if the GUI order should be converted to a regular column and the existing data migrated. Having an attribute with is CLOB is difficult to handle in JPQL for sorting. 

@keycloak/core-iam - can you please comment?

Ekamdeep3044:
@ahus1 Thanks for your comment but can you please provide more details on your below statement:

"As long as there are only 10 IDPs, this is fine. Once there are more, this will lead to wrong results" 

Since I tested it with more than 10 IDP's and it is working fine. I Even tested rendering 5-5 IDP's on page even when API returns more than 5 and they were rendered with correct GUI order only.

ahus1:
I see that the pagination is by default 5, so my comment should have been "As long as there are only 5 (unlinked) IDPs, this is fine."

@Ekamdeep3044 - looking at the following code that retrieves the unlinked IdPs, you see that it retrieves the first page of data, which will be 6 items given the default pagination of 5 items per page: 

https://github.com/keycloak/keycloak/blob/025b2ba4421547cb65c8c727957fdba075c7f5df/services/src/main/java/org/keycloak/services/resources/account/LinkedAccountsResource.java#L147-L158

When analyzing this, you see that the items are returned ordered by the IDP alias, as specified in 

https://github.com/ahus1/keycloak/blob/900c496ffe19b7413fab07130926bfc4465a1322/model/jpa/src/main/java/org/keycloak/models/jpa/JpaIdentityProviderStorageProvider.java#L294

Assuming the pagination of 5 items per page, and that your IDPs with aliases a,b,c,d,e have the display order 10,11,12,13,14,15, and the IDP aliases f,g,h,i,j have the display order 1,2,3,4,5. 

The expectation would the to see on the first page f,g,h,i,j and on the second a,b,c,d,e

The current behavior is a bit more complicated as the page is retrieving the pagination size + 1 item. With your code change code change it will sort those 6 items, which will put the 6th item first as that one had the order 1, so you would see on the first page f,a,b,c,d.

Looking at the backend code you will also see that the linked accounts are already sorted in the backend (although a bit inefficiently as it will retrieve all items from the backend before sorting them), while the unlinked accounts are not sorted. 

This is my analysis given the current code. Please give an example of both IDP aliases and display orders that you tested with.

Ekamdeep3044:
@ahus1 Got it. Thanks..

Lets see what @keycloak/core-iam team says about having guiOrder as a column in IDP. That would be more convenient.

sguilhen:
I generally agree with @ahus1 on this one. Not long ago we've made some improvements to IDPs and we've essentially given them their own provider to improve scalability along with some other tweaks. The initial plan was to keep the schema changes to the minimum so we didn't rework the guiOrder to promote it to a first-class attribute of the IDP (it currently exists as a config in a generic name-value table). Also, IIRC the login screens don't rely on pagination - they fetch every available IDP and sort them using the guiOrder in the UI, so we decided to leave that as is.

At that time the account console was still being reworked, so it probably missed the sorting in the UI. However, I do agree that we should probably make this sorted at the backend, and add a param to the API when fetching IDPs to control how they should be returned (by alias or by guiOrder). So I would personally vote to get this done in the backend.

@pedroigor Perhaps something for KC 27 as it will involve DB migration to move the guiOrder into a column in the IDP table?

ahus1:
> However, I do agree that we should probably make this sorted at the backend

I think we have two options here for the account console: 

* Do the proper JPA change and pagination, or 
* fetching all IDPs that can be shown in the account console, sort them, and then return the relevant page to the user. 

We never show IdPs that are linked to an organization, so would it be safe (enough) to assume that there are never that many IdPs in the account console that it would matter much? The pagination might be more relevant to the admin UI? 

So we can do the sorting now, and the refactoring with the new column later. 

BTW, we can do those JPA changes in any minor release (still) except when it leads to Java API changes (which might be the case, or or maybe not, I don't know yet).

pedroigor (commented):
Sorting the brokers linked to a user (federated identity) in the backend should be fine because we won't expect many.

However, for brokers where the user does have a federated identity it will basically introduce back the scalability issue when a realm has a lot of (non-org) brokers available.

@Ekamdeep3044 Your contribution is very welcome and appreciated, but we cannot do it properly now without proper migration. 

@sguilhen Let us put it on hold and wait for more people to chime in so that we can priortize it accordingly. Wdyt?

I would also say that this is related to improvements in the login page because there we are listing all brokers (depending on the hide on login page flag) and the UX is much worse than in the account console.

pedroigor:
@Ekamdeep3044 Changed the PR state to draft so that we can discuss the changes more.

The changes should involve introducing a Liquibase change set to add a column to the `IDENTITY_PROVIDER` entity, a Liquibase migration to move data from the `IDENTITY_PROVIDER_CONFIG` table to the new column related to the GUI order, and updating the `org.keycloak.models.jpa.JpaIdentityProviderStorageProvider#getAllStream` to sort by the GUI order.

Issue #30180, opened 2024-06-05
Keycloak Organizations


pedroigor:
### Description

This top-level epic is about the planned work to deliver the first release of the Keycloak Organization feature.

Keycloak Organization is a feature that leverages and builds on top of the Identity and Access Management (IAM) capabilities of Keycloak to address Customer Identity and Access Management (CIAM) with a focus on Business-to-Business (B2B) use cases.

                                        +-----------+                  
                                        |           |                  
                            +-----------+  Customer +-----------+      
                            |           |           |           |      
    +----------+      +-----+----+      |           |      +----+-----+
    |          |      |          |      +-----------+      |          |
    | End-User +------+  Realm   |                         | End-User |
    |          |      |          |                         |          |
    |          |      |          |                         |          |
    +----------+      +-----+----+                         +----+-----+
                            |           +-----------+           |      
                            |           |           |           |      
                            +-----------+  Partner  +-----------+      
                                        |           |                  
                                        |           |                  
                                        +-----------+                       

In summary, this feature will allow a realm to integrate with third-party entities like customers and business partners so that their identities and how they access protected resources at the realm level are segregated and managed within the scope of the organization they belong to. 

As a result, a realm will be able to provide a different experience when onboarding customer identities or identities from a business partner and mechanisms to secure and manage how they interact with a realm and how they access protected resources from the clients available from a realm. Although not the main focus, it should also be possible to leverage this feature if you just want a better segregation between users, clients, and how they are managed and authenticate to a realm.

The main set of capabilities provided by Keycloak Organizations are:

* * [x] **Manage Organizations**
* * [ ] **Manage Organization Members**
* * * [x] Organization Member Self-Registration based on Identity Brokering
* * * [x] Organization Member Registration based on Invitation Links
* * * [ ] Joining Multiple Organizations 
* * [x] **Mapping organization metadata in tokens** 
* * [ ] **Manage Organization Roles**
* * [ ] **Manage Organization Groups**
* * [ ] **Manage Authentication within the Scope of an Organization**
* * * [ ] Authentication based on the Organization Authentication Policies**
* * [ ] **Manage Clients and Service Accounts**
* * [ ] **Administrative Authorization**
* * * [ ] **RBAC**
* * * [ ] **Fine-Grained** 
* * [ ] **Organization Self-Service**

The planned release dates for Keycloak Organization are:

| Release  | State |
| ------------- | ------------- |
| 25.0.0  | Preview  |
[...]

PR #40508, opened 2025-06-16
Make organization `domains` optional

### LINKED ISSUE #31285: Make domains for organisations optional
### Description

When creating an organisation, a domain **must** be specified.

My suggestion would be to make this **optional**.

The domain is currently used to determine which user wants to log in (into which organisation), for example. Without a domain, it could possibly be adapted as follows:

- if there is only one user with the e-mail globally, it is unambiguous
- if there are several, the request must contain the organisation ID or the name, which apparently must be unique; an input field could be added in the ui where the user can enter the name of their organisation

### Discussion

_No response_

### Motivation

Even if it is rare and becoming increasingly rare, not every organisation (often a "customer" in the SaaS context) has its own domain.

### Details

_No response_

SferaDev:
Closes #31285

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
@@ -42,6 +42,7 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.From;
 import jakarta.persistence.criteria.Join;
+import jakarta.persistence.criteria.JoinType;
 import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
 import org.keycloak.connections.jpa.JpaConnectionProvider;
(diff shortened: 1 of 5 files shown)


ahus1:
As with other pull requests: Please do not rebase a PR unless a reviewer asks you to rebase the PR, or GitHub reports conflicts. Every rebase will trigger unecessary CI runs and notifications to maintainers. The PR will be rebased on merge anyway, so please ignore any GitHub notifications that the branch is outdated.

SferaDev:
@ahus1 How can I get a maintainer review?

ahus1:
Pinging @pedroigor and his team to get a review on this PR.

cvivieca:
Any plans to add this?

SferaDev:
It's been very difficult to get any of my contributions to organizations feature being reviewed 😔

siepkes:
> It's been very difficult to get any of my contributions to organizations feature being reviewed 😔

That would be consistent with what I've heard (in talks, mailinglists, etc.) from the [keycloak-orgs](https://github.com/p2-inc/keycloak-orgs) extension people. It feels like there is a slight miss match with what a part of the community is looking for in an "organizations" feature and what the maintainers want it to be. By which I mean that part of the community sees the "organizations" feature as a way to support SaaS. Like this PR which removes the domain requirement, a requirement which doesn't really work for the SaaS use-case. The [keycloak-orgs](https://github.com/p2-inc/keycloak-orgs) extension also primarily seems to be geared towards SaaS. Now obviously I can't speak for anyone but it feels to me the maintainers seem to view the "organizations" functionality more from a B2B / CIAM perspective (as outlined in https://github.com/keycloak/keycloak/issues/30180). The B2B/CIAM use-case overlaps for 99% with the SaaS use-case but not entirely and I think in the view of the maintainers this is such a case where it does not overlap. 

Ironically, I need this domain requirement removal for a B2B / CIAM scenario. Because I need to deal with totally separate country organizations part of the same company (which share a single domain) but have nothing to do with each other.

gfl-akovalchuk (approved):

doucsag:
Could you maybe take a look at this @pedroigor ? I think this would be beneficial to quite a number of people.

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
```diff
@@ -302,12 +303,11 @@ private TypedQuery<Long> buildCountQuery(CriteriaBuilder builder, CriteriaQuery<
 
     private Predicate buildStringSearchPredicate(CriteriaBuilder builder, CriteriaQuery<?> query, Root<OrganizationEntity> org, String search,
                                                  Boolean exact) {
-        Root<OrganizationDomainEntity> domain = query.from(OrganizationDomainEntity.class);
+        Join<OrganizationEntity, OrganizationDomainEntity> domain = org.join("domains", JoinType.LEFT);
```
pedroigor: We don't need this join, do we?

model/jpa/src/main/java/org/keycloak/organization/jpa/OrganizationAdapter.java
```diff
@@ -175,8 +175,8 @@ public Stream<OrganizationDomainModel> getDomains() {
 
     @Override
     public void setDomains(Set<OrganizationDomainModel> domains) {
-        if (domains == null || domains.isEmpty()) {
-            throw new ModelValidationException("You must provide at least one domain");
+        if (domains == null) {
+            domains = Set.of();
```
pedroigor:
```suggestion
            return;
```

pedroigor (changes requested):

pedroigor:
Thanks, @SferaDev. Can you please check if my comments make sense?

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
SferaDev: The LEFT join ensures organizations without domains are included while the domain predicate is applied. We already have a `IDX_ORG_DOMAIN_ORG_ID` that optimizes this JOIN performance wise.

pedroigor: My point is that queries will only operate on the organization table, as we don't have, AFAIK, any filter or anything else that will end up in an inner join on the domain table. Therefore, we don't need a left join at all?

pedroigor: I've updated the code to skip the left join when there is no search (by name or domain).

pedroigor (approved):

pedroigor:
Thanks, @SferaDev !

Issue #25575, opened 2023-12-14
Allow configuring permissions on a per-context basis


pedroigor:

asergios:
Following!

michelkluger:
good feature, looking forward for this one!

mr-m0nst3r:
any progress, guys?

Splitframe:
As also described here https://github.com/keycloak/keycloak/issues/26359

We also have the use case that a user should be able to set attributes like the Name and Birthdate during the registration, but not be able to change them afterwards. 

Edit: 
This is actually a huge fraud prevention problem (law) for us. 
Anyone know a workaround? Even if we disable the self management console the user still has the right
to change the fields, so they could do it with postman for example.
So we need to have it disabled that the user can change it at all.

ThoreKr:
> We also have the use case that a user should be able to set attributes like the Name and Birthdate during the registration, but not be able to change them afterwards.
> 
> Edit: This is actually a huge fraud prevention problem (law) for us. Anyone know a workaround?

We have a workaround, which is more a side effect how processes interact with each other.
There is a separate age validation process, which stores the birthday in a separate admin-writable and non-user-visible attribute. This attribute is then used whenever the birthday is considered.

You could create an event-handler that takes registration events and copies the attributes to something the user isn't allowed to see or edit.
I agree that this sounds insane.

pedroigor:
I'm open to moving this forward if someone is willing to contribute to the changes.

I was thinking about how we could enhance the UP configuration to support context for individual attribute settings, and I would like to know your thoughts on this proposal.

Today, we have the following fields that could be configured on a per-context basis:

* `Enabled when`
* `Required field`
* `Permission`

Not sure if it makes sense to have `Validations` configured on a per-context basis. But one might arguee that `annotations` might make sense because, depending on the context, you might have the attribute being rendered differently.

Considering the three settings above, a possible solution would be.

For the `Required field`, we could replace `Required for` with a selection of contexts. The reasoning is that, selecting `Both users and admins`, `Only users`, or `Only admins`, is basically about choosing the context where that attribute is required. If both users and admins then all contexts, if only users, only user-facing contexts such as update profile, account, etc. If admin only, then the admin api context.

The behavior of this setting will be to allow setting a white-list of contexts, where if none is provided, it applies to all contexts.

The same idea could be applied to permissions as well. Instead of selecting `User` or `Admin`, you would select the contexts where the read/write permission applies.

[...]

PR #40520, opened 2025-06-16
Disable email verification when email manually changed by idp review.

### LINKED ISSUE #40446: Review Profile makes users prone to phishing attacks
### Description

While setting up a new SaaS-like product offering at Defense Unicorns and testing Social Logins, @bburky found an interesting attack vector that might expose users for phishing attacks. 

> This is very close to an account takeover, but a (weird looking) verification email goes to the victim account (not the attacker). This probably could be successful with some phishing along the side too:
> 
>     1. Assume [victim@defenseunicorns.com](mailto:victim@defenseunicorns.com) is an existing registered (admin or whatever) account
> 
>     2. Attacker self-registers a [attacker@example.com](mailto:attacker@example.com) user with user/pass (not social) and completes email verification. Attacker then logs out of Keycloak
> 
>     3. Attacker does a new social login (I tested with Google) for [attacker@example.com](mailto:attacker@example.com)
> 
>     4. This triggers "first-broker-login" in Keycloak. The user is prompted to confirm merge accounts....
>        a. The attacker clicks the "review profile" button. This page shows all the info provided by the Social login provider... and it's editable?!?! The attacker modifies their email to be [victim@defenseunicorns.com](mailto:victim@defenseunicorns.com)
>        b. The attacker is prompted to merge their account with [victim@defenseunicorns.com](mailto:victim@defenseunicorns.com), they click "add to existing account"
>        c. Attacker is shown "you need to verify your email address to link your account with Google Social Login"
>        i. This email is sent to [victim@defenseunicorns.com](mailto:victim@defenseunicorns.com) (NOT [attacker@example.com](mailto:attacker@example.com)), this prevents this from being zero-interaction.
>        ii. The email sent has a link expiring in 5 minutes, and does include the [attacker@example.com](mailto:attacker@example.com) name in the text
>        iii. It's not unreasonable to assume some phishing (and a more phishy attacker email address) could convince a user to click this link
>        iv. A very similar Keycloak web page is shown to the victim with a "click to proceed" link
> 
>     5. If the link is clicked, the attacker gains access to the victim account

As point 4.iii highlights, the victim receives an email that doesn't contain the attacker email address. It is very easy to get caught with this phishing attack.

### Discussion

_No response_

### Motivation

The primary motivation for adding an extra layer of protection is to defend users from accidentally getting caught with this attack. 

### Details

After reviewing Keycloak code, the Review Profile Action is performed in the default first broker login.

According to the Keycloak manual ([here](https://www.keycloak.org/docs/latest/server_admin/index.html#_detect_existing_user_first_login_flow)), there are a few ways we can set it up:

> When ON, users are presented with the profile page requesting additional information to federate the user’s identities.
> When missing, users are presented with the profile page if the identity provider does not provide mandatory information, such as email, first name, or last name.
> When OFF, the profile page does not display unless the user clicks in a later phase on the Review profile info link in the page displayed by the Confirm Link Existing Account authenticator.

There’s 4th option to disable this step entirely, here’s the description of this option directly from the help popup:

> WARN: In case that user clicks ‘Review profile info’ on link duplications page, the update page will be always displayed. You would need to disable this authenticator to never display the page.

In my opinion, there should be one more option in the Review Profile section that presents the screen to the user but doesn't allow him to modify the "required" attributes. With this option, this attack vector would be neutralized as an attacker wouldn't be able to change his email (and other fields such as First Name, Last Name etc). At the same time, it's absolutely find to add other non-required fields (e.g. Rank, Organization etc).

rmartinc:
Add option to hide the idp review button.
Closes #40446

Draft PR that just adds the ideas presented in the issue comment.

1. The `Verify existing account by Email` steps is not executed when the `username` or the `email` are modified in the `Review Profile` step. This way sending emails to random users is disabled. The PR sets a configuration option to revert to previous behavior (if you think that is not needed just let me know). When disabled only the re-authentication execution is available to link the account.
2. The PR also adds an option to remove the review profile button in `Confirm link existing account`. I managed the option to just check if the `Review Profile` step was executed before. But I preferred to add the option just in case a custom review execution is used or similar. (The same, if you think that is better not to use an option and just check if the review step has been executed just let me know.)

I have chatted a bit with @pedroigor and he thinks we will need to improve email texts too. It's a good idea. We will need an upgrading note too. But I'll do both points if the general idea in this PR is OK.

Tests added. @keycloak/core-iam take a look.


sguilhen:
This gets a +1 from me. I am personally not sure about the option to revert to the current behavior in point 1 above. I would rather always have the username AND email change validations always in place.

rmartinc:
Thanks @sguilhen! Yes, I'm not sure either about adding the previous behavior. I just did it because maybe somebody is using only email to link users, and this way they can revert to previous behavior. Let's wait for @pedroigor and we have three votes.

pedroigor:
It makes more sense to remove both new options. As @rmartinc mentioned, in a scenario where a legitimate user is trying to link an account from an IdP where the email/username is different, it is still possible to finish the linking by re-authenticating. And this mechanism is much better/secure than allowing sending arbitrary e-mails and eventually taking over accounts.

Perhaps the last thing to add is a release and upgrade guide note about this new behavior.

FTR, this is also related to https://github.com/keycloak/keycloak/issues/25575. Ideally, we should control the behavior of user attributes through user profile. Preventing users from changing emails when being federated from an IdP is a valid use case and could also help as an extra security layer to prevent such issues. 

But regardless, we need to prevent sending verification emails when username/email changes in favor of using the user credentials.

rmartinc:
OK, I have done the changes we have been talking about:

* There is no new options for the authenticators. The email validator never sends emails if the email/username is changed during review (so previous wrong behavior is not allowed anymore). This step is only available when the username/email received from IdP is used.
* The review button is not displayed if the idp review authenticator was not successfully executed. This avoids the option too.
* The emails and the confirming page of the email verification have been changed a bit. IMHO the email was quite good and  just added a sentence saying `if you didn't initiate this process...` The confirming page was modified more to differentiate title and body, and adding a message similar to the email for the body.
* A upgrading note was added as a notable change commenting the changes.

Let's see the tests but they worked OK for me in another branch.

pedroigor (approved):

mposolda:
@rmartinc @pedroigor @sguilhen Thanks for this PR and for reviewing it!

mposolda (approved):

slaskawi:
Fantastic news! Thanks a million for creating the PR and reviewing it folks!

PR #40522, opened 2025-06-16
Upgrade webauth4j to current release 0.29.3

rmartinc:
Closes #40023

Upgrading the webauthn4j to current release 0.29.3. It needs some tweaks (mainly `validator` to `verifier` in several places) to accommodate the new version. I also restored the tests in the attestation conveyance that needs to disable the trust-store (the tests uses chrome and it seems that the certificate used in the attestation is a self-signed generated every time; so they cannot be added to the trust-store in the tests).

@pskopek Take a look when you have time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -212,7 +212,7 @@
         </surefire.system.args>
 
         <!-- webauthn support -->
-        <webauthn4j.version>0.21.5.RELEASE</webauthn4j.version>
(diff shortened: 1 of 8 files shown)


pskopek (approved):
Looks Good!

ahus1 (approved):
Approving based on previous reviews. Thanks!

PR #40532, opened 2025-06-16
Making DefaultLazyLoader thread safe

### LINKED ISSUE #40531: DefaultLazyLoader is not thread safe, but is used in a shared instance of CachedRealm
### Before reporting an issue

### Area

infinispan

### Describe the bug

Instances of CachedRealm are stored in the local cache of each Keycloak instance and then accessed concurrently. 

Those instances then hold instances of DefaultLazyLoader which needs to be thread safe as it is accessed from multiple threads concurrently

### Version

main

### Regression

### Expected behavior

Classes accessed from multiple threads should be thread safe. 

This class explicitly states that it is not. This is possibly not doing harm, still it leads to concurrently loading data in parallel threads, which can and should be avoided to prevent a stampede. 

### Actual behavior

Code states it is not thread safe, but is accessed from multiple threads in parallel

### How to Reproduce?

Code review of CachedRealm and DefaultLazyLoader

### Anything else?

I'll create a PR

ahus1:
Closes #40531

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/DefaultLazyLoader.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/DefaultLazyLoader.java
@@ -24,8 +24,9 @@ import java.util.function.Supplier;
 import org.keycloak.models.KeycloakSession;
 
 /**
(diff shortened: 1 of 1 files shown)


ahus1:
@pedroigor - please review when you have the time. Thanks!

pedroigor (approved):

PR #40538, opened 2025-06-16
Issue with Handling Negative Values in Certain Fields of Brute Force Detection

### LINKED ISSUE #40353: Issue with Handling Negative Values in Certain Fields of Brute Force Detection
### Before reporting an issue

### Area

authentication

### Describe the bug

When configuring and enabling **"Brute Force Detection"** within the authentication settings of any realm, it is possible to enter negative values for fields such as Max Login Failures and Max Wait. These negative values are then stored as-is in the Keycloak database. However, from a logical standpoint, such negative values are invalid and should not be allowed. 

The proper solution would be to validate these values within the realm adapter’s setters before they are assigned, ensuring that invalid (negative) inputs are rejected early on.

### Version

26.2.5

### Regression

### Expected behavior

I expect that providing a negative value should trigger an exception, indicating that the value is invalid and not acceptable.

### Actual behavior

Currently, negative values entered in the mentioned fields are stored, but they behave the same way as if the value were set to 1.

### How to Reproduce?

Navigate to your desired realm, go to the Authentication section, then select Security Defenses and choose Brute Force Detection.

### Anything else?

_No response_

douglaspalmer:
Closes #40353

### DIFF
--- a/js/apps/admin-ui/src/realm-settings/security-defences/BruteForceDetection.tsx
+++ b/js/apps/admin-ui/src/realm-settings/security-defences/BruteForceDetection.tsx
@@ -140,7 +140,7 @@ export const BruteForceDetection = ({
               labelIcon={t("failureFactorHelp")}
               controller={{
                 defaultValue: 0,
-                rules: { required: t("required") },
+                rules: { required: t("required"), min: 0 },
               }}
             />
             {bruteForceMode ===
@@ -151,6 +151,7 @@ export const BruteForceDetection = ({
                 labelIcon={t("maxTemporaryLockoutsHelp")}
                 controller={{
                   defaultValue: 0,
+                  rules: { min: 0 },
                 }}
               />
             )}
@@ -170,9 +171,9 @@ export const BruteForceDetection = ({
                     value: t(`bruteForceStrategy.${key}`),
                   }))}
                 />
-                <Time name="waitIncrementSeconds" />
-                <Time name="maxFailureWaitSeconds" />
-                <Time name="maxDeltaTimeSeconds" />
+                <Time name="waitIncrementSeconds" min={0} />
+                <Time name="maxFailureWaitSeconds" min={0} />
+                <Time name="maxDeltaTimeSeconds" min={0} />
               </>
             )}
             <NumberControl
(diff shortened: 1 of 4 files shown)


model/jpa/src/main/java/org/keycloak/models/jpa/RealmAdapter.java
```diff
@@ -326,7 +334,9 @@ public int getMinimumQuickLoginWaitSeconds() {
 
     @Override
     public void setMinimumQuickLoginWaitSeconds(int val) {
-        setAttribute("minimumQuickLoginWaitSeconds", val);
+        if(val >= 0) {
+            setAttribute("minimumQuickLoginWaitSeconds", val);
```
rmartinc: Would not be better to do this in the DefaultExportImportManager (updateRealm and importRealm methods). And I would prefer to throw an exception (ModelException or similar) that will return a 400 bad request for the endpoint. The same for all the properties.

rmartinc (changes requested):
@douglaspalmer I think that it would better to check this when moving from rep to model in the `DefaultExportImportManager`. And I always prefer to return an error than doing nothing. WDYT?

And try with this commit https://github.com/rmartinc/keycloak/commit/2204992575ab93b019f485e3a75d0a5fd70482bb in the UI part. It's not perfect but much better.

model/jpa/src/main/java/org/keycloak/models/jpa/RealmAdapter.java
douglaspalmer: Done

rmartinc (changes requested):
Don't you like the changes in the UI https://github.com/rmartinc/keycloak/commit/2204992575ab93b019f485e3a75d0a5fd70482bb ?

rmartinc (approved):
@douglaspalmer For me it's OK. But maybe adding a method to all the calls would more readable:

```java
public static int checkNonNegativeNumber(int value, String name) {
    if(value < 0) {
        throw new ModelException(name + " may not be a negative value");
    }
    return value;
}
```

And just call it something line:

```java
if (rep.getMaxTemporaryLockouts() != null) {
    newRealm.setMaxTemporaryLockouts(checkNonNegativeNumber(rep.getMaxTemporaryLockouts(), "Maximum temporary lockouts"));
}
```

Besides this way we execute the CI again to remove the failure. But the idea is OK, it's just an improvement.

rmartinc (approved):
Thanks @douglaspalmer! LGTM!

ssilvert (approved):
Approved the UI part.

edewit (approved):

abstractj (approved):

PR #40544, opened 2025-06-17
Use computer language for createTimeout integer value

### LINKED ISSUE #40474: WebAuthn Passwordless Policy Timeout Field Causes Syntax Error When Value Exceeds 1000 Seconds Due to Locale-Specific Number Formatting in FTL Generated JavaScript
### Before reporting an issue

### Area

adapter/javascript

### Describe the bug

When configuring the "Timeout" field within the Keycloak Admin Console's Configure -> Authentication -> Policies -> Webauthn Passwordless Policy section, entering a value greater than 1000 seconds results in a JavaScript Syntax Error.
This appears to be caused by FreeMarker Template Language (FTL) applying locale-specific number formatting, which inserts a comma (,) as a thousand separator into the generated JavaScript code.
This prevents the value from being parsed as a valid Number type by the JavaScript engine.

### Version

26.2.5 (tested with the latest official Docker image as per reproduction steps), 23.0.7 (our currently utilized version, exhibiting the same issue)

### Regression

### Expected behavior

The "Timeout" field should correctly handle and display values greater than 1000 seconds without introducing formatting characters like commas.
The input value should be rendered as a pure numerical value in the generated JavaScript, ensuring it is properly interpreted as a Number type and does not cause a Syntax Error.
The user should be able to proceed with WebAuthn authentication as expected.

### Actual behavior

When a value exceeding 1000 seconds is entered (e.g., 3600), the generated JavaScript code contains a comma (e.g., 3,600 instead of 3600).
This invalidates the number literal in JavaScript, leading to a Syntax Error during script execution, preventing the WebAuthn flow from proceeding correctly.

### How to Reproduce?

1. Start a Keycloak container without any modifications (e.g., using the command from the official getting started guide):
2. Access the Keycloak Admin Console (e.g., http://localhost:8080).
3. Log in with the admin credentials (e.g., admin/admin).
4. Navigate to Manage realms (top left dropdown) and create a new test realm, e.g., test-realm.
5. Switch to the test-realm.
6. Under Manage -> Users, create a new test user, e.g., test.
7. Set a password for the test user in the Credentials tab.
8. For the test user, in the Required user actions input field, set Webauthn Register Passwordless.
9. Navigate to Configure -> Authentication -> Policies -> Webauthn Passwordless Policy.
10. Locate the "Timeout" field and set its value to something greater than 1000 seconds (e.g., 3600 for 1 hour). Click "Save".
11. Attempt to log in to the test-realm using the test user through a Keycloak client application or directly via the account console (e.g., http://localhost:8080/realms/test-realm/account).
12. Open your browser's developer tools (e.g., F12) and inspect the Console tab for JavaScript errors.

Example of the generated JavaScript snippet causing the error:
```javascript
const input = {
  // ... code before ...
  requireResidentKey : 'not specified',
  userVerificationRequirement : 'not specified',
  createTimeout : 3,600, // causing SyntaxError
  excludeCredentialIds : '',
  initLabel : "Passkey (Default Label)",
  // ... code after ...
}
```

### Anything else?

_No response_

rmartinc:
Closes #40474

Just adding the `?c` to write the `createTimeout` in computer syntax to the JavaScript. We did the same some months ago for the algorithms.

### DIFF
--- a/themes/src/main/resources/theme/base/login/login-passkeys-conditional-authenticate.ftl
+++ b/themes/src/main/resources/theme/base/login/login-passkeys-conditional-authenticate.ftl
@@ -109,7 +109,7 @@
                 challenge : '${challenge}',
                 userVerification : '${userVerification}',
                 rpId : '${rpId}',
-                createTimeout : ${createTimeout},
+                createTimeout : ${createTimeout?c},
                 errmsg : "${msg("webauthn-unsupported-browser-text")?no_esc}"
             };
             authButton.addEventListener("click", () => {
@@ -121,7 +121,7 @@
                 challenge : '${challenge}',
                 userVerification : '${userVerification}',
                 rpId : '${rpId}',
-                createTimeout : ${createTimeout},
+                createTimeout : ${createTimeout?c},
                 errmsg : "${msg("passkey-unsupported-browser-text")?no_esc}"
             };
 
--- a/themes/src/main/resources/theme/base/login/passkeys.ftl
+++ b/themes/src/main/resources/theme/base/login/passkeys.ftl
@@ -17,7 +17,7 @@
                challenge : '${challenge}',
                userVerification : '${userVerification}',
                rpId : '${rpId}',
-               createTimeout : ${createTimeout}
+               createTimeout : ${createTimeout?c}
            };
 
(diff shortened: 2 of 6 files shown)


graziang (approved):
@rmartinc thanks!

ahus1 (approved):

PR #40580, opened 2025-06-18
chore(i18n): Update Polish lang in email and account theme

kwiateusz:
Closes #40579


ahus1 (changes requested):
Thank you for this PR. Please update the description from `refs #40579` to `Closes #40579` as this is the syntax that GitHub understands. Also please add the signoff to your commit. Ideally your commit should also contain the `Closes #40579`, as otherwise a maintainer will need to add it manually.

PR #40588, opened 2025-06-18
Removing the Windows optimization as it is currently unstable

ahus1:
Closes #40583


mhajas (approved):
Thanks for the backport

PR #40595, opened 2025-06-18
Copy restored maven repo to home folder in Windows

rmartinc:
Closes #40593

Copying the repository from the download folder to the correct user maven folder. I used copy to avoid any issue if the repo is already created. I checked that it works with and without cache.


rmartinc:
I added a second commit because I saw that, when the cache is not there, the FIPS can fail. I just saw it once but just in case.

ahus1 (approved):
Thank you for looking into this!

PR #40601, opened 2025-06-18
set auto-mount service account token to false in keycloak pods

KyriosGN0:
Signed-off-by: AvivGuiser <aviv.guiser@placer.ai>

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
@@ -292,6 +292,9 @@ public class KeycloakDeploymentDependentResource extends CRUDKubernetesDependent
 
         var specBuilder = baseDeploymentBuilder.editSpec().editTemplate().editOrNewSpec();
 
+        if (!specBuilder.hasAutomountServiceAccountToken()) {
+            specBuilder.withAutomountServiceAccountToken(false);
+        }
         if (!specBuilder.hasRestartPolicy()) {
             specBuilder.withRestartPolicy("Always");
         }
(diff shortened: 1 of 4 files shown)


operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
```diff
@@ -149,6 +149,10 @@ public class KeycloakSpec {
     @JsonPropertyDescription("Configuration for startup probe, by default it is 1 for periodSeconds and 600 for failureThreshold")
     private ProbeSpec startupProbeSpec;
 
+    @JsonProperty("automountServiceAccountToken")
+    @JsonPropertyDescription("Set this to automount the default ServiceAccount Token and Service CA")
+    private boolean automountServiceAccountToken;
```
shawkins: Please use Boolean here instead.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
```diff
@@ -151,7 +151,7 @@ public class KeycloakSpec {
 
     @JsonProperty("automountServiceAccountToken")
     @JsonPropertyDescription("Set this to automount the default ServiceAccount Token and Service CA")
-    private boolean automountServiceAccountToken;
+    private Boolean automountServiceAccountToken = true;
```
shawkins, line 154: Don't default it to true, just treat null as true.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -102,8 +102,6 @@ public class KeycloakDeploymentDependentResource extends CRUDKubernetesDependent
 
     public static final String OPTIMIZED_ARG = "--optimized";
 
-    private boolean useServiceCaCrt;
```
shawkins, line 105: The old logic will need to remain.

KyriosGN0:
@shawkins had a few git issues, i will re open the PR

PR #40605, opened 2025-06-18
set auto-mount service account token to false in keycloak pods

### LINKED ISSUE #38843: Set `automountServiceAccountToken: false` on Keycloak pods
### Description

Keycloak pods (unlike Operator pods) do not need access to Kube API, hence no service account is needed.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

KyriosGN0:
fixes #38843

### DIFF
--- a/docs/guides/high-availability/examples/generated/keycloak.yaml
+++ b/docs/guides/high-availability/examples/generated/keycloak.yaml
@@ -330,6 +330,7 @@ spec:
     limits:
       cpu: "6"
       memory: "2250M"
+  automountServiceAccountToken: true
   db:
     vendor: postgres
     url: jdbc:aws-wrapper:postgresql://<AWS_AURORA_URL_HERE>:5432/keycloak?sslmode=verify-full&sslrootcert=/opt/keycloak/conf/truststores/configmap-keycloak-aurora-rootcert/aurora.pem # <1>
--- a/docs/guides/high-availability/examples/generated/keycloak.yaml
+++ b/docs/guides/high-availability/examples/generated/keycloak.yaml
@@ -330,7 +330,6 @@ spec:
     limits:
       cpu: "6"
       memory: "2250M"
-  automountServiceAccountToken: true
   db:
     vendor: postgres
     url: jdbc:aws-wrapper:postgresql://<AWS_AURORA_URL_HERE>:5432/keycloak?sslmode=verify-full&sslrootcert=/opt/keycloak/conf/truststores/configmap-keycloak-aurora-rootcert/aurora.pem # <1>
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -377,6 +377,7 @@ stringData:
 
 When running on a Kubernetes or OpenShift environment well-known locations of trusted certificates are included automatically.
 This includes `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` and the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` when present.
+In order to not include `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` in Keycloak pods, set the `automountServiceAccountToken` field in the spec to `false`
 
 === Admin Bootstrapping
 
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -378,6 +378,7 @@ stringData:
 When running on a Kubernetes or OpenShift environment well-known locations of trusted certificates are included automatically.
 This includes `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` and the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` when present.
 In order to not include `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` in Keycloak pods, set the `automountServiceAccountToken` field in the spec to `false`
+This is useful if some security policies require that the service account token is not mounted in the pod, but it cannot be `false` if you plan to use an external infinispan cluster or if plan to use the Kubernetes service accounts identity provider or if you have some custom provider logic which requires use of the Kubernetes API.
 
 === Admin Bootstrapping
 
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -378,7 +378,7 @@ stringData:
 When running on a Kubernetes or OpenShift environment well-known locations of trusted certificates are included automatically.
 This includes `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` and the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` when present.
 In order to not include `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` in Keycloak pods, set the `automountServiceAccountToken` field in the spec to `false`
-This is useful if some security policies require that the service account token is not mounted in the pod, but it cannot be `false` if you plan to use an external infinispan cluster or if plan to use the Kubernetes service accounts identity provider or if you have some custom provider logic which requires use of the Kubernetes API.
+This is useful if some security policies require that the service account token is not mounted in the pod, but it cannot be `false` if you plan to use an external infinispan cluster, or if plan to use the Kubernetes service accounts identity provider, or if you have some custom provider logic which expects to implicitly use the Kubernetes API.
 
 === Admin Bootstrapping
 
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
@@ -465,14 +465,7 @@ public class KeycloakDeploymentDependentResource extends CRUDKubernetesDependent
         LinkedHashMap<String, EnvVar> varMap = Stream.concat(Stream.concat(unsupportedEnv.stream(), firstClasssEnvVars.stream()), Stream.concat(additionalEnvVars.stream(), env))
                 .collect(Collectors.toMap(EnvVar::getName, Function.identity(), (e1, e2) -> e1, LinkedHashMap::new));
 
-        String truststores = SERVICE_ACCOUNT_DIR;
-        if (baseDeployment.getSpec().getTemplate().getSpec().getAutomountServiceAccountToken()) {
-            truststores += "ca.crt";
-        }
 
-        if (useServiceCaCrt) {
-            truststores += "," + SERVICE_CA_CRT;
-        }
         if (!Boolean.FALSE.equals(keycloakCR.getSpec().getAutomountServiceAccountToken())) {
             String truststores = SERVICE_ACCOUNT_DIR + "ca.crt";
 
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
@@ -473,7 +473,13 @@ public class KeycloakDeploymentDependentResource extends CRUDKubernetesDependent
         if (useServiceCaCrt) {
             truststores += "," + SERVICE_CA_CRT;
         }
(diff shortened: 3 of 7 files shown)


KyriosGN0:
pretty sure the `testNoAutoMountServiceAccount` passed locally, just to be extra sure i will re-run it locally tomorrow

ahus1:
For the Keycloak Multi-Site setup, IMHO we assume that the CA used in OpenShift is imported automatically so we can communicate with Infinispan securely. 

Looping in @ryanemerson / @pruivo to ensure that the docs are updated for that.

pruivo:
For mult-site HA, Infinispan requires `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` 
[Documentation](https://infinispan.org/docs/infinispan-operator/main/operator.html#hotrod-configuration-builder_clients)

ahus1:
@pruivo - in addition to that, when Keycloak talks to Infinispan in that setup, it uses TLS. And to verify the Infinispan certificate, it needs to know about the CA. I remember there was some kind of mechanism. 

Maybe @shawkins can refresh my memory here how Keycloak picks up that CA?

pruivo:
@ahus1 The operator automatically adds the certificate to the truststore: 
https://github.com/keycloak/keycloak/blob/7736ca20e9d55bb8d58d79cd8fc08c5c8bb01899/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java#L448-L449

ahus1:
@pruivo, thank you, I was looking for that code snippet but didn't find it. So this option would be required in the Keycloak HA setup as described in https://www.keycloak.org/high-availability/deploy-keycloak-kubernetes

So if this is merged for the 26.3 release and we decide to update the docs in a follow-up issue, that follow-up issue should be a blocker for the 26.3 release.

pruivo:
@ahus1 How so? This PR is disabling the service account tokens (`/var/run/secrets/kubernetes.io/serviceaccount/token
`), and I don't find any relation to the service certificates in the Kubernetes docs.

ahus1:
@pruivo - OK, then I was just worried without a reason. Then proceed here, and our tests in the cluster will show if it breaks. Sorry for the noise!

pruivo:
This is the only doc I found: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting
If something breaks after this PR is merged, we will know the culprit :rofl:

vmuzikar:
> I don't find any relation to the service certificates in the Kubernetes docs

Based on https://github.com/keycloak/keycloak/issues/38843#issuecomment-2983882995 it would appear that the CA cert is also not mounted when `automountServiceAccountToken: false`.

KyriosGN0:
> @ahus1 How so? This PR is disabling the service account tokens (`/var/run/secrets/kubernetes.io/serviceaccount/token `), and I don't find any relation to the service certificates in the Kubernetes docs.

the `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (and that entire directory) is controlled by that flag, should i add the a clarification in some doc page that this option has to be true for multi-site setup in k8s?

vmuzikar:
> should i add the a clarification in some doc page that this option has to be true for multi-site setup in k8s?

+1

KyriosGN0:
added the `automountServiceAccountToken: true` to the example keycloak used in this [doc](https://www.keycloak.org/high-availability/deploy-keycloak-kubernetes)

KyriosGN0:
Hey @vmuzikar would it be possible to get someone to take a look here? 

Thanks in advance!

vmuzikar:
@KyriosGN0 Sorry, will take a look this week.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -301,6 +301,12 @@ private StatefulSet createBaseDeployment(Keycloak keycloakCR, Context<Keycloak>
         if (!specBuilder.hasDnsPolicy()) {
             specBuilder.withDnsPolicy("ClusterFirst");
         }
+        if (Boolean.TRUE.equals(keycloakCR.getSpec().getAutomountServiceAccountToken()) || keycloakCR.getSpec().getAutomountServiceAccountToken() == null) {
+            specBuilder.withAutomountServiceAccountToken(true);
+
+        } else if (Boolean.FALSE.equals(keycloakCR.getSpec().getAutomountServiceAccountToken())) {
+            specBuilder.withAutomountServiceAccountToken(false);
+        }
```
shawkins:
```suggestion
        boolean automount = !Boolean.FALSE.equals(keycloakCR.getSpec().getAutomountServiceAccountToken());
        specBuilder.withAutomountServiceAccountToken(automount);
```

The assumption moving forward is that there will be no null case - we'll set the spec automountServiceAccountToken to true or false.

shawkins: Allowing null to be the default handling would probably be best.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -439,14 +445,18 @@ private void addEnvVars(StatefulSet baseDeployment, Keycloak keycloakCR, TreeSet
...
         if (useServiceCaCrt) {
             truststores += "," + SERVICE_CA_CRT;
         }
-
-        // include the kube CA if the user is not controlling KC_TRUSTSTORE_PATHS via the unsupported or the additional
-        varMap.putIfAbsent(KC_TRUSTSTORE_PATHS, new EnvVarBuilder().withName(KC_TRUSTSTORE_PATHS).withValue(truststores).build());
+        if (Boolean.TRUE.equals(baseDeployment.getSpec().getTemplate().getSpec().getAutomountServiceAccountToken()) || useServiceCaCrt) {
+            // include the kube CA if the user is not controlling KC_TRUSTSTORE_PATHS via the unsupported or the additional
+            varMap.putIfAbsent(KC_TRUSTSTORE_PATHS, new EnvVarBuilder().withName(KC_TRUSTSTORE_PATHS).withValue(truststores).build());
+        }
```
shawkins:
```suggestion
        if (!Boolean.FALSE.equals(keycloakCR.getSpec().getAutomountServiceAccountToken())) {
            String truststores = SERVICE_ACCOUNT_DIR + "ca.crt";

            if (useServiceCaCrt) {
                truststores += "," + SERVICE_CA_CRT;
            }
        
            // include the kube CA if the user is not controlling KC_TRUSTSTORE_PATHS via the unsupported or the additional
            varMap.putIfAbsent(KC_TRUSTSTORE_PATHS, new EnvVarBuilder().withName(KC_TRUSTSTORE_PATHS).withValue(truststores).build());
        }
```

Switching to using the entire service account directory if automount is disabled doesn't seem like the right thing to do. How about updating the truststores only if automount is enabled.

KyriosGN0:
@shawkins thanks for the review, i agree with your suggestion and have committed them

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
```diff
@@ -149,6 +149,10 @@ public class KeycloakSpec {
     @JsonPropertyDescription("Configuration for startup probe, by default it is 1 for periodSeconds and 600 for failureThreshold")
     private ProbeSpec startupProbeSpec;
 
+    @JsonProperty("automountServiceAccountToken")
+    @JsonPropertyDescription("Set this to automount the default ServiceAccount Token and Service CA")
```
vmuzikar:
```suggestion
    @JsonPropertyDescription("Set this to to false to disable automounting the default ServiceAccount Token and Service CA. This is enabled by default.")
```

docs/guides/high-availability/examples/generated/keycloak.yaml
```diff
@@ -424,6 +424,7 @@ spec:
     limits:
       cpu: "6"
       memory: "2250M"
+  automountServiceAccountToken: true
```
vmuzikar: We should not be probably modifying this file. Moreover, `true` is the default so no need to specify that like this.

We should rather have somewhere specified in the HA guide that it must not be set to `false`. Maybe @ahus1 or @pruivo can provide some pointers what would be the best place for the affected scenarios.

vmuzikar (changes requested):

docs/guides/high-availability/examples/generated/keycloak.yaml
KyriosGN0: @ahus1 @pruivo if you could point me to the correct doc page, I would really appreciate it!

pruivo: Probably a warning box before this new sub-section:

https://github.com/keycloak/keycloak/blob/74027bfbfe1fc5598e3f01dd9a694fada8e753f8/docs/guides/high-availability/deploy-keycloak-kubernetes.adoc?plain=1#L48

KyriosGN0: Done!

KyriosGN0:
@vmuzikar would it be possible for you take another look? 
Thanks!

stianst (dismissed):
We use this here https://github.com/keycloak/keycloak/pull/42816

KyriosGN0:
> We use this here #42816

Ok, by default it does remain (since its true by default), do you want me to add a note in the docs about this feature ? @stianst

KyriosGN0:
hey @vmuzikar @stianst, is it possible to get a review here?

KyriosGN0:
hey @vmuzikar @stianst, is there something i can do to progress with this PR ?

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -377,6 +377,7 @@ stringData:
 
 When running on a Kubernetes or OpenShift environment well-known locations of trusted certificates are included automatically.
 This includes `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` and the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` when present.
+In order to not include `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` in Keycloak pods, set the `automountServiceAccountToken` field in the spec to `false`
```
shawkins, line 380: Here, in the CRD property description, or both it should be clearer why you would want to set it to false and when that isn't possible.

If this seems like it's getting too nuanced, then I'm still in favor of using a networkpolicy that denies access from the keycloak pods to the api server.

vmuzikar: @KyriosGN0 Sorry for the late reply.

> If this seems like it's getting too nuanced

IMHO it's not.

@KyriosGN0 In order to move this forward, we would need to document that the service account is required by the Kube service accounts identity provider.

> why you would want to set it to false

+1

shawkins: And infinispan.

KyriosGN0: @shawkins do you mean using external infinispan cluster ?

shawkins: Yes - https://github.com/keycloak/keycloak/pull/40605#issuecomment-2990155758

It can also be needed if there is custom provider logic that expects to connect to the api server.

KyriosGN0: @shawkins, @vmuzikar Done!

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -377,6 +377,8 @@ stringData:
...
+This is useful if some security policies require that the service account token is not mounted in the pod, but it cannot be `false` if you plan to use an external infinispan cluster or if plan to use the Kubernetes service accounts identity provider or if you have some custom provider logic which requires use of the Kubernetes API.
```
shawkins:
```suggestion
This is useful if some security policies require that the service account token is not mounted in the pod, but it cannot be `false` if you plan to use an external infinispan cluster, or if plan to use the Kubernetes service accounts identity provider, or if you have some custom provider logic which expects to implicitly use the Kubernetes API.
```

shawkins (dismissed):
Thanks @KyriosGN0 this looks good. Just a couple of minor edits might be good.

vmuzikar (approved):
Thank you for the contribution, LGTM!

PR #40618, opened 2025-06-19
Allow editing the protocol mapper name

ahus1:
Closes #40617

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/ClientAdapter.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/ClientAdapter.java
@@ -448,23 +448,23 @@ public class ClientAdapter implements ClientModel, JpaModel<ClientEntity> {
 
     @Override
     public void updateProtocolMapper(ProtocolMapperModel mapping) {
-        if (mapping.getName() != null && !mapping.getName().equals(entity.getName())) {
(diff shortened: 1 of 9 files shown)


ahus1:
@martin-kanis - I think we might have talked about this a while ago? Please review this PR when you find the time, or let me know if I should look for a different reviewer. Thanks!

martin-kanis (approved):
@ahus1 Thanks! Looks good!

model/jpa/src/main/java/org/keycloak/models/jpa/ClientAdapter.java
```diff
@@ -446,13 +448,23 @@ public void removeProtocolMapper(ProtocolMapperModel mapping) {
...
+            logger.warn("Deprecated: Always pass in the protocol mapper name when updating (since Keycloak 26.4)");
+        }
+        pmEntity.setProtocolMapper(mapping.getProtocolMapper());
+        if (pmEntity.getConfig() == null) {
+            pmEntity.setConfig(mapping.getConfig());
         } else {
-            entity.getConfig().clear();
-            entity.getConfig().putAll(mapping.getConfig());
+            pmEntity.getConfig().clear();
+            pmEntity.getConfig().putAll(mapping.getConfig());
```
pedroigor, line 467: Can we avoid duplicating this code by moving it to some utility class like `ProtocolMapperUtils`?

pedroigor (changes requested):

PR #40621, opened 2025-06-19
Missing Quarkus flag for Syslog counting framing

### LINKED ISSUE #39893: Missing Quarkus flag for syslog logging
### Before reporting an issue

### Area

core

### Describe the bug

When enabling syslog logging with configuration (either command line or config file), and setting `log-syslog-output` to `tcp` or `ssl-tcp`, logs fail to be received, in particular by syslog-ng. This seems to be because of the syslog framing, which is required to do byte counting when TCP or SSL/TLS is used (see RFC5424).

Setting the Quarkus property [quarkus.log.syslog.use-counting-framing](https://quarkus.io/guides/logging#quarkus-core_quarkus-log-syslog-use-counting-framing) allows syslog events to be captured.

I would suggest one of two approaches to resolving this

1. Add an extra config option to mirror the  `quarkus.log.syslog.use-counting-framing` option
2. When setting the `log-syslog-output` config option to `tcp` or `ssl-tcp`, then automatically set the `quarkus.log.syslog.use-counting-framing` option correctly.

Given approach 2, it might even be an upstream change to Quarkus to do this. I so, I can raise on the Quarkus issues.

### Version

26.2.4

### Regression

### Expected behavior

When enabling syslog logging for TCP or SSL-TCP, then byte count framing should be enabled.

### Actual behavior

Without the flag, UDP syslog format framing is used, and syslog messages fail to be received.

### How to Reproduce?

Create a syslog collector (syslog-ng or similar. Enable syslog in Keycloak with the following options.

* --log=console,syslog
* --log-level=INFO
* --log-syslog-app-name=keycloak
* --log-syslog-endpoint=host:port
* --log-syslog-level=debug
* --log-syslog-protocol=ssl-tcp
* --log-syslog-type=rfc5424
* --log-syslog-output=json

Log evens will not be received and syslog-ng in debug mode will repot the framing error. After setting `quarkus.log.syslog.use-counting-framing` in the quarkus.properties, then syslog logging is successful.

### Anything else?

_No response_

mabartos:
- Closes #39893

The property is added in this PR, which contains smarter decisions based on the used protocol same as Quarkus will do once we upgrade to the newer version (not backported to 3.20.x): https://github.com/quarkusio/quarkus/pull/48479

Once we upgrade Quarkus, we can remove the handling of the `protocol-dependent` strategy.

@keycloak/cloud-native-maintainers Could you please check it? Thanks!

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/LoggingPropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/LoggingPropertyMappers.java
@@ -14,7 +14,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 import java.util.function.BiFunction;
-import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.stream.Stream;
 
@@ -219,7 +218,6 @@ public final class LoggingPropertyMappers {
                         .build(),
                 fromOption(LoggingOptions.LOG_SYSLOG_COUNTING_FRAMING)
                         .isEnabled(LoggingPropertyMappers::isSyslogEnabled, SYSLOG_ENABLED_MSG)
-                        .mapFrom(LoggingOptions.LOG_SYSLOG_PROTOCOL, LoggingPropertyMappers::resolveSyslogCountingFramingMapFrom)
                         .transformer(LoggingPropertyMappers::resolveSyslogCountingFraming)
                         .to("quarkus.log.syslog.use-counting-framing")
                         .paramLabel("strategy")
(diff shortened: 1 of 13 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/LoggingPropertyMappers.java
```diff
@@ -411,4 +420,35 @@ private static void validateSyslogMaxLength(String value) {
...
+    private static String resolveSyslogCountingFraming(String value, ConfigSourceInterceptorContext context) {
+        if (SYSLOG_COUNTING_FRAMING_PROTOCOL_DEPENDENT.equals(value)) {
+            return Configuration.getOptionalKcValue(LoggingOptions.LOG_SYSLOG_PROTOCOL)
+                    .map(LoggingPropertyMappers::handleSyslogCountingFramingBasedOnProtocol)
+                    .orElse(Boolean.FALSE.toString());
+        }
+        return value;
+    }
+
+    private static String resolveSyslogCountingFramingMapFrom(String protocol, ConfigSourceInterceptorContext context) {
```
shawkins: Is this needed? Since there is a default value it will be used and transformed without a need to do a mapFrom.

shawkins (commented):
@mabartos looks good, just one thought on the usage of mapFrom in this case.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/LoggingPropertyMappers.java
mabartos: You're right, thanks :)

mabartos:
@shawkins Ready for another review.

cc: @vmuzikar @Pepo48

shawkins (dismissed):
LGTM, thanks @mabartos

mabartos:
@shawkins Added the notable change for 26.3.0

shawkins (approved):
LGTM

mabartos:
@vmuzikar Could you please review/approve this PR once you have a time? Thanks!

vmuzikar (approved):
LGTM, thank you.

PR #40624, opened 2025-06-19
Use offline time calculations when transient created from offline

### LINKED ISSUE #40611: Negative expiration for token exchange using an offline session
### Before reporting an issue

### Area

token-exchange

### Describe the bug

Token exchange can return negative expiration values (in the past) when using an offline session. The main reason is that keycloak is using a transient session, that session is not detected as offline, and therefore the timeouts are calculated using online values.

Example returned response in a demo realm:

```json
{
  "access_token":"ey..",
  "expires_in":-103,
  "refresh_expires_in":0,
  "token_type":"Bearer",
  "not-before-policy":0,
  "session_state":"af927eef-cb86-4c91-8fa0-32a7f65f625a",
  "scope":"profile email",
  "issued_token_type":"urn:ietf:params:oauth:token-type:access_token"
}
```

And the access token is also in the past:

```json
{
  "exp": 1750325886,
  "iat": 1750325989,
  "auth_time": 1750325570,
  "jti": "ftrtte:bebe8121-31e9-457c-b57b-63f4dd64500e",
  "iss": "http://localhost:8080/realms/test-session",
  ...
}

### Version

26.2.5

### Regression

### Expected behavior

The response in token exchange should be OK and with correct expiration values.

### Actual behavior

It returns expiration times in the past-

### How to Reproduce?

1. Modify the realm to have a short SSO Session Max (5 mins for example).
2. Create an offline session using code to token.
3. Wait more than the 5 mins that would have expired an online session.
4. Refresh the offline token to obtain a new access token.
5. With the new access token perform a token exchange request.
6. The values returned are in the past (calculation is done like it was an online session which would have been expired).

### Anything else?

_No response_

rmartinc:
Closes #40611

PR:             https://github.com/keycloak/keycloak/pull/40615
Commit:         https://github.com/keycloak/keycloak/commit/1350da4332718dcfbe0cadb5949ea7023b179a51
PR branch:      backport-40615-26.2
Target branch:  https://github.com/keycloak/keycloak/tree/release/26.2

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
@@ -1039,8 +1039,11 @@ public class TokenManager {
             expiration = Time.currentTimeMillis() + TimeUnit.SECONDS.toMillis(tokenLifespan);
         }
 
+        final boolean offline = userSession.isOffline() || offlineTokenRequested ||
+                (userSession.getPersistenceState() == UserSessionModel.SessionPersistenceState.TRANSIENT &&
+                Constants.CREATED_FROM_PERSISTENT_OFFLINE.equals(userSession.getNote(Constants.CREATED_FROM_PERSISTENT)));
         long sessionExpires = SessionExpirationUtils.calculateClientSessionMaxLifespanTimestamp(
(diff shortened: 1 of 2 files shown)


ahus1 (approved):

PR #40653, opened 2025-06-22
Fix: Removed conditional check preventing use of custom attribute names

### LINKED ISSUE #40497: Creating a user profile attribute "displayName" does not work as expected.
### Before reporting an issue

### Area

user-profile

### Describe the bug

I wanted to make a custom attribute called `displayName` visible and editable to my users. However, when typing the `e` in `displayName`, the localization tag seems to just fail to update.

https://github.com/user-attachments/assets/424389ae-36a2-497c-b3b3-f1db6df379ee

This causes the localization to fail, since in the localization tab it is listed properly but in the user profile it is searching for the wrong tag. (`profile.attributes.displayNam` instead of `profile.attributes.displayName`)

Image

Image

### Version

26.2.5

### Regression

### Expected behavior

I expected the `e` in `displayName` to be filled in like everything else.

### Actual behavior

The `e` in `displayName` does not get filled in appropriately.

### How to Reproduce?

As can be seen in the video:
1. Create an attribute called `displayName`.
2. Observe it not work correctly.

### Anything else?

I was able to work around this by using the browser tools to make the Display name tag for localization of the User Attribute editable, and then manually editing it. Saving the attribute registered the change appropriately.

Barathwaja:
Removed conditional check preventing use of custom attribute names identical to built-in Keycloak attributes. Please review and provide your feedback.

Closes: [40497](https://github.com/keycloak/keycloak/issues/40497)

Thanks!

### DIFF
--- a/js/apps/admin-ui/src/realm-settings/user-profile/attribute/TranslatableField.tsx
+++ b/js/apps/admin-ui/src/realm-settings/user-profile/attribute/TranslatableField.tsx
@@ -98,9 +98,7 @@ export const TranslatableField = ({
   const requiredTranslationName = `${translationPrefix}.0.value`;
 
   useEffect(() => {
-    if (isTranslationRequired(value, t, realm)) {
-      setValue(fieldName, `\${${prefix}.${value}}`);
-    }
+    setValue(fieldName, `\${${prefix}.${value}}`);
   }, [value]);
 
   return (


pedroigor (approved):

pedroigor:
Thanks @Barathwaja. I've updated the commit in your branch to pass the DCO check and rebased.

Barathwaja:
Thanks @pedroigor.

PR #40655, opened 2025-06-22
added a non-null check to check whether the to be exported realm exists

### LINKED ISSUE #39122: Export fails with an unexpected error if the realm does not exist
### Before reporting an issue

### Area

core

### Describe the bug

When exporting a realm and setting a realm that does not exist, you will get an unexpected error as follows:

```
2025-04-22 10:34:11,115 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) Error details:: java.lang.NullPointerException: Cannot invoke "org.keycloak
.models.RealmModel.getId()" because "realm" is null          
        at org.keycloak.models.utils.ModelToRepresentation.toRepresentation(ModelToRepresentation.java:369)
        at org.keycloak.exportimport.util.ExportUtils.exportRealm(ExportUtils.java:82) 
        at org.keycloak.exportimport.util.ExportUtils.exportRealm(ExportUtils.java:78) 
        at org.keycloak.exportimport.util.MultipleStepsExportProvider$1.runExportImportTask(MultipleStepsExportProvider.java:99)
        at org.keycloak.exportimport.util.ExportImportSessionTask.run(ExportImportSessionTask.java:36)
        at org.keycloak.models.utils.KeycloakModelUtils.lambda$runJobInTransaction$1(KeycloakModelUtils.java:341)
        at org.keycloak.models.utils.KeycloakModelUtils.runJobInTransactionWithResult(KeycloakModelUtils.java:460)

```

Instead, exporting a realm that does not exist should give a proper and better error message to make it easier to identify the problem.

### Version

26.2

### Regression

### Expected behavior

Better error message when exporting a realm that does not exist

### Actual behavior

Confusing message when exporting a realm that does not exist

### How to Reproduce?

Run:

```
./kc.sh -v export --dir=/tmp/realm-export --realm=non-existent-realm
```

Where the `realm` option is set with an invalid realm name.

### Anything else?

_No response_

boehmalex:
In the `SingleFileExportProvider.exportRealm(...)` (used when exporting to a single file) we are checking if the realm exists. In case it does not, we throw a proper error message:

> 2025-06-19 21:41:40,483 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) Error details:: java.lang.NullPointerException: realm not found by realm name 'non-existent-realm'
at java.base/java.util.Objects.requireNonNull(Objects.java:235)
at org.keycloak.exportimport.singlefile.SingleFileExportProvider.exportRealm(SingleFileExportProvider.java:88)

I copied this non-null check to the `MultipleStepsExportProvider.runExportImportTask(...)` (used when exporting multiple files to a directory) to have the same check and proper error message if the realm does not exist.

Closes #39122

### DIFF
--- a/model/storage-services/src/main/java/org/keycloak/exportimport/util/MultipleStepsExportProvider.java
+++ b/model/storage-services/src/main/java/org/keycloak/exportimport/util/MultipleStepsExportProvider.java
@@ -35,6 +35,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -102,6 +103,7 @@ public abstract class MultipleStepsExportProvider<T extends MultipleStepsExportP
             @Override
             protected void runExportImportTask(KeycloakSession session) throws IOException {
                 RealmModel realm = session.realms().getRealmByName(realmName);
+                Objects.requireNonNull(realm, "realm not found by realm name '" + realmName + "'");
                 session.getContext().setRealm(realm);
                 RealmRepresentation rep = ExportUtils.exportRealm(session, realm, exportUsersIntoRealmFile, true);
                 writeRealm(realmName + "-realm.json", rep);
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/ExportDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/ExportDistTest.java
@@ -71,4 +71,13 @@ public class ExportDistTest {
         cliResult.assertMessage("Export of realm 'fgap' requested.");
         cliResult.assertMessage("Export finished successfully");
     }
+
+    @Test
+    void testExportNonExistent(KeycloakDistribution dist) {
+        CLIResult cliResult = dist.run("build");
+
+        cliResult = dist.run("export", "--realm=non-existent-realm", "--dir=.");
+        cliResult.assertMessage("realm not found by realm name 'non-existent-realm'");
+    }
+
 }
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/exportimport/ExportImportTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/exportimport/ExportImportTest.java
@@ -22,20 +22,15 @@ import org.hamcrest.Matchers;
 import org.jboss.arquillian.container.spi.client.container.LifecycleException;
 import org.junit.After;
 import org.junit.Test;
-import org.keycloak.Config;
 import org.keycloak.admin.client.resource.RealmResource;
 import org.keycloak.authentication.requiredactions.WebAuthnRegisterFactory;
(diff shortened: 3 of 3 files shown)


shawkins:
@boehmalex thank you for the fix. Can you see if it's easy to add a test for this case in ExportImportTest?

boehmalex:
@shawkins sorry for the late response. I tried to add a test when creating the fix, but with the usage of `testRealmExportImport` i couldn't manage to check for the expected exception, because the exception is converted to a standard `UncaughtServerError` in `KeycloakErrorHandler.getResponse(...)`. And in this exception there is no cause/reason or something else left to get to the root cause.

But i did a little more testing and this could be a test (its actually working):

```java
public void checkForRealmNotFoundExceptionWhenDirExportingNonExistentRealm() throws IOException {
        String realmName = "non-existent-realm";
        String targetDirPath = testingClient.testing().exportImport().getExportImportTestDirectory() + File.separator + "dirRealmExport";
        testingClient.testing().exportImport().setDir(targetDirPath);
        testingClient.testing().exportImport().setRealmName(realmName);
        Config.Scope config = new Config.SystemPropertiesScope("");

        DirExportProviderFactory dirExportProviderFactory = new DirExportProviderFactory();
        dirExportProviderFactory.init(config);

        KeycloakSessionFactory factory = KeycloakApplication.getSessionFactory();
        KeycloakSession keycloakSession = factory.create();
        ExportProvider exportProvider = dirExportProviderFactory.create(keycloakSession);

        try {
            exportProvider.exportModel();
        } catch (NullPointerException e) {
            assertEquals(String.format("realm not found by realm name '%s'", realmName), e.getMessage());
        }
    }
```

But i don't know if this is considered as an OK test, because its a little puzzled together and not using the `testingClient.testing().exportImport().runExport();`

The same test base could be used to test the Exception in `SingleFileExportProvider`.

shawkins:
Thank you for looking more into this @boehmalex

> And in this exception there is no cause/reason or something else left to get to the root cause.

Right, you'd have to inspect the server log instead.

> But i don't know if this is considered as an OK test, because its a little puzzled together and not using the testingClient.testing().exportImport().runExport();

I'm fine with this more direct style of test. It would be clearer though to just by-pass using testingClient.testing().exportImport() methods and instead use ExportImportConfig methods.

boehmalex:
@shawkins Thanks for your input - valid point to use `ExportImportConfig` directly! I will update this PR and provide the tests at the end of next week at the earliest.

boehmalex:
Rebased because of merge conflicts and added tests - see pr conversation history with @shawkins

shawkins:
> Rebased because of merge conflicts and added tests - see pr conversation history with @shawkins

Thank you @boehmalex I made a change to the test logic based upon what was done in some of the other logic. If this builds cleanly, then we should be good to merge this.

shawkins:
I'm not sure why the new tests were causing the load-balancer port conflicts seen in the store it tests. I'll just move it over to a full distribution test instead.

boehmalex:
@shawkins Thanks for your help and efforts finishing this pr!

shawkins (approved):
LGTM, @boehmalex sorry for the difficulty in getting a passing test.

abstractj (approved):

PR #40667, opened 2025-06-23
Fixed the IDs comparison logic

### LINKED ISSUE #40666: Incorrect types of collection arguments in CustomLockService. SAST
### Description

The `Arrays.asList(DBLockProvider.Namespace.values())` returns a `List` of `Namespace` type objects, not their IDs. The `containsAll(...)` method, in turn, checks whether these objects are contained in `currentIds`, which will always return false, since `currentIds` is a `Set` of `Integer` objects:

https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/lock/CustomLockService.java#L129-L139

---

Anchels:
Closes #40666


model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/lock/CustomLockService.java
```diff
@@ -128,7 +128,12 @@ public void init() throws DatabaseException {
 
         try {
             Set<Integer> currentIds = currentIdsInDatabaseChangeLogLockTable();
-            if (!currentIds.containsAll(Arrays.asList(DBLockProvider.Namespace.values()))) {
+
+            if (!currentIds.containsAll(
+		    Arrays.stream(DBLockProvider.Namespace.values())
+		    .map(DBLockProvider.Namespace::getId)
+		    .collect(Collectors.toSet())
+	    )) {
```
shawkins:
```suggestion

           if (!Arrays.stream(DBLockProvider.Namespace.values())
                    .map(DBLockProvider.Namespace::getId).allMatch(currentIds::contains)) {
```

shawkins:
@Anchels thank you for the PR, are you able to add a test for this case as well?

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/lock/CustomLockService.java
Anchels: @shawkins Thanks!

Anchels:
@shawkins sorry, I don't have experience with writing tests

shawkins:
> @shawkins sorry, I don't have experience with writing tests

No problem. 

@rmartinc @sguilhen @ahus1 can you review and indicate whether you want a test case?

sguilhen:
Not sure if this fix is enough - see https://github.com/keycloak/keycloak/blob/main/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/lock/CustomLockService.java#L130-L139

The `CustomInitializeDatabaseChangeLogLockTableStatement` in line 139  should probably be using the ids from the Namespace enum. As it stands now, it is updating the table with the same values it has read in line 130. So, to me, besides fixing the wrong type comparison we should also fix the initialization to set the ids from the enum

Anchels:
@sguilhen, could you please see if the latest changes will help fix the problem? Or should we combine `currentIds` with enum IDs for the initializatoin?

sguilhen:
To be on the safe side I would combine them.

shawkins:
Resolved the conflict. Sorry @Anchels, I did not realize this was waiting on me.

shawkins (approved):
~~LGTM~~ - actually store model tests are not passing with this change. @sguilhen can you reivew your expectations from https://github.com/keycloak/keycloak/pull/40667#issuecomment-3045701535

Closed this issue this was referencing because it was already addressed. @sguilhen @mabartos can you review this pr to see if a new issue is needed, or if this pr can simply be closed at this point?

sguilhen:
@shawkins this can be closed - the original issue was properly fixed and my expectations around the current ids was not correct as I misunderstood what the SQL generator was doing with the current ids passed in the CustomInitializeDatabaseChangeLogLockTableStatement constructor.

PR #40669, opened 2025-06-23
[26.2.x] Missing Quarkus flag for Syslog counting framing

### LINKED ISSUE #39893: Missing Quarkus flag for syslog logging

mabartos:
- Closes #39893
- Changed the default value for `log-syslog-counting-framing` to `false`, instead of `protocol-dependent` used for main
- It means there will be no breaking changes as we only expose the knob for the counting framing

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
@@ -247,4 +247,13 @@ public class LoggingOptions {
             .description("Set the Syslog output to JSON or default (plain) unstructured logging.")
             .build();
 
+    // we can use SyslogConfig.CountingFraming type once https://github.com/quarkusio/quarkus/pull/48479 is present
+    public static final String SYSLOG_COUNTING_FRAMING_PROTOCOL_DEPENDENT = "protocol-dependent";
+    public static final Option<String> LOG_SYSLOG_COUNTING_FRAMING = new OptionBuilder<>("log-syslog-counting-framing", String.class)
+            .category(OptionCategory.LOGGING)
+            .expectedValues(Boolean.TRUE.toString(), Boolean.FALSE.toString(), SYSLOG_COUNTING_FRAMING_PROTOCOL_DEPENDENT)
+            .defaultValue(Boolean.FALSE.toString())
+            .description("If 'true', the message being sent is prefixed with the size of the message. If '%s', the default value is 'true' when '%s' is 'tcp' or 'ssl-tcp', otherwise 'false'."
+                    .formatted(SYSLOG_COUNTING_FRAMING_PROTOCOL_DEPENDENT, LOG_SYSLOG_PROTOCOL.getKey()))
+            .build();
 }
--- a/docs/guides/server/logging.adoc
+++ b/docs/guides/server/logging.adoc
@@ -338,6 +338,25 @@ To use UDP instead of TCP, add the `--log-syslog-protocol` option as follows:
 
 The available protocols are: `tpc`, `udp`, and `ssl-tcp`.
 
+=== Configuring the Syslog counting framing
+
+By default, Syslog messages sent over TCP or SSL-TCP are prefixed with the message size, as required by certain Syslog receivers.
+This behavior is controlled by the `--log-syslog-counting-framing` option.
+
+To explicitly enable or disable this feature, use the following command:
+
+<@kc.start parameters="--log-syslog-counting-framing=true"/>
+
+You can set the value to one of the following:
+
+* `protocol-dependent` – Enable counting framing only when the `log-syslog-protocol` is `tcp` or `ssl-tcp`.
+* `true` – Always enable counting framing by prefixing messages with their size.
+* `false` (default) – Never use counting framing.
+
+Note that using `protocol-dependent` ensures compatibility with most Syslog servers by enabling the prefix only when required by the protocol.
+
+NOTE: In this {project_name} version, the counting framing is turned off by default to prevent breaking changes. It is recommended to use `protocol-dependent`.
+
 === Configuring the Syslog log format
 To set the logging format for a logged line, perform these steps:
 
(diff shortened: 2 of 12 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
```diff
@@ -247,4 +247,13 @@ public String toString() {
             .description("Set the Syslog output to JSON or default (plain) unstructured logging.")
             .build();
 
+    // we can use SyslogConfig.CountingFraming type once https://github.com/quarkusio/quarkus/pull/48479 is present
+    public static final String SYSLOG_COUNTING_FRAMING_PROTOCOL_DEPENDENT = "protocol-dependent";
+    public static final Option<String> LOG_SYSLOG_COUNTING_FRAMING = new OptionBuilder<>("log-syslog-counting-framing", String.class)
+            .category(OptionCategory.LOGGING)
+            .expectedValues(Boolean.TRUE.toString(), Boolean.FALSE.toString(), SYSLOG_COUNTING_FRAMING_PROTOCOL_DEPENDENT)
+            .defaultValue(Boolean.FALSE.toString())
```
mabartos, line 255: This differs from the PR done for main: https://github.com/keycloak/keycloak/pull/40621. Default value changed to 'false' to have the same behavior as before.

mabartos:
@shawkins Could you please also check this backport? Please look at the description. Thanks!

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
shawkins: Does that mean there will need to be a 26.3 release note describing the change in the defaulting?

shawkins (approved):
LGTM, just wondering if we need an upgrade note in the pr.

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
mabartos: @shawkins I was thinking about it as well, but didn't provide it as it depends on the "used" point of view. Before, we didn't have any control mechanism for the Syslog counting framing as it was considered as Quarkus internals. With the PR to 26.3, we introduced the mechanism and changed defaults, which will be the same in the next Quarkus versions. So it might be perceived as some defaulting changed on the Quarkus "side", which we don't really describe. 

But I agree to write some note to the "Notable changes" to have it properly addressed. I'll add it there.

mabartos: Added: https://github.com/keycloak/keycloak/pull/40621/commits/a8174f29f3a49e0ad41c5995589ec278ed07d7ec

vmuzikar:
This is basically a new feature. I do not think we should be adding that to 26.2. I understand it solves an issue for a particular syslog receiver, still I wouldn't consider it a bug that justifies backporting.

mabartos:
> I do not think we should be adding that to 26.2. I understand it solves an issue for a particular syslog receiver, still I wouldn't consider it a bug that justifies backporting.

@vmuzikar I'm on top of the fence  - considering 26.2 will live for some time, inability to use majority of syslog receivers might be an issue and providing quarkus.properties to change only this simple property seems to me like overkill - mainly for containers and operator. That's why I thought it might be good to provide the option itself as well and change the default value to `false` to not introduce any breaking change.

On the other hand, I understand your point, and both options work for me, even with the fact some users/customers might face difficulties to configure Keycloak with Syslog - so I'm more inclined to include it. 

Please, let me know what you think about this consideration. Thanks!

vmuzikar:
@mabartos Thank you for elaborating.

Let me loop in also @ahus1 as he's more involved in the Observability area.

I appreciate it might be a blocker for some users. Though it's not a regression, or even a bug per se. Keycloak 26.2 will soon reach EOL in the community, and AFAIK we haven't heard any RFE from the product side for this. Hence -1 from my side for backporting this.

mabartos:
> I appreciate it might be a blocker for some users. Though it's not a regression, or even a bug per se. Keycloak 26.2 will soon reach EOL in the community, and AFAIK we haven't heard any RFE from the product side for this. Hence -1 from my side for backporting this.

Ok, works for me not to backport it and close this as 26.3 is behind a door for community. We have a workaround described in the issue, so we can potentially react ad-hoc :) I think it's not necessary to involve more people as we have a conclusion.

PR #40676, opened 2025-06-24
Adjusted the comparisons

### LINKED ISSUE #40663: Potential copy-paste issue in PersistentClientSessionEntity.java
### Description

I have identified a potential issue in the `equals()` method in the following code:
https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java#L196-L197
It appears that the instance's fields should be compared to the corresponding fields of passed object, rather than comparing them to `key.clientId`

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

Anchels:
Closes #40663


shawkins:
@ahus1 @rmartinc can you review and / or indicate if test cases are missing

ahus1 (approved):
Thank you for this fix!

Issue #37356, opened 2025-02-14
Enhanced migration story


vmuzikar:
### Description

**This is at the moment mostly a placeholder issue to gather ideas for enhancement in the migration area.**

### Discussion

-

### Issues

- https://github.com/keycloak/keycloak/issues/26889

### Motivation

Current migration story is not ideal, especially with the Operator deployments.
* Migration can be interrupted by the Startup probe, see https://github.com/keycloak/keycloak/issues/35261.
* Migration can be interrupted by a transaction timeout, see https://github.com/keycloak/keycloak/issues/35261.
* Migration is not atomic. Interrupting it might lead to a broken state, see https://github.com/keycloak/keycloak/issues/35261.
* Operator by default deploys a matching Server image. Updating the Operator consequently also updates the Server. This might be confusing and problematic, especially when we do not provide clear instructions for a rollback. See https://github.com/keycloak/keycloak/pull/37295#issuecomment-2659299682.

stianst:
We should also consider expensive operations like creating indexes for big tables; for those we may want to be able to do it async or in the background. Maybe initiated by the admin

PR #40690, opened 2025-06-24
Provide CLI Parameters for jgroups.* options

### LINKED ISSUE #40481: Provide CLI Parameters for jgroups.* options
### Description

We updated the docs to add Java options to append for the JGroups JDBC configuration. This is now difficult to read and we don't validate these parameters. 

### Discussion

https://github.com/keycloak/keycloak/issues/39614

### Motivation

This is difficult to validate, and hard to pass compared to other parameters. The configuration should be consistent with other configuration options, and document

### Details

See https://www.keycloak.org/server/caching for the docs. 

The following items should become first class options: 

* jgroups.bind.address
* jgroups.bind.port
* jgroups.external_port (need to be system properties as they are not in the XML of ISPN)
* jgroups.external_addr (need to be system properties as they are not in the XML of ISPN)

Not exposed on CLI - we can still expose it as a SPI option:
* jgroups.fd.port-offset

The reason the FD port is in the docs is due to the fact that it opens a network port. We don't expect users to change it. 
With ISPN 16, there is a plan to remove FDSOCK from the TCP setup, see https://github.com/infinispan/infinispan/pull/15038

### Implementation idea

Add CLI options that then set the system properties so they are available for all caches.

```
--cache-embedded-network-bind-address
--cache-embedded-network-bind-port
--cache-embedded-network-external-address
--cache-embedded-network-external-port
```

SPI option:

```
--spi-cache-embedded--default--network-fd-port-offset
```

ryanemerson:
Closes #40481

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
+++ b/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
@@ -17,13 +17,23 @@
 
 package org.keycloak.spi.infinispan.impl.embedded;
 
+import static org.infinispan.configuration.global.TransportConfiguration.STACK;
+import static org.keycloak.config.CachingOptions.CACHE_EMBEDDED_PREFIX;
+
 import java.lang.invoke.MethodHandles;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLServerSocket;
+import javax.net.ssl.TrustManager;
 
 import org.infinispan.commons.configuration.attributes.Attribute;
 import org.infinispan.configuration.global.TransportConfigurationBuilder;
@@ -31,6 +41,7 @@ import org.infinispan.configuration.parsing.ConfigurationBuilderHolder;
 import org.infinispan.remoting.transport.jgroups.EmbeddedJGroupsChannelConfigurator;
 import org.infinispan.remoting.transport.jgroups.JGroupsTransport;
 import org.jboss.logging.Logger;
+import org.jgroups.Global;
 import org.jgroups.conf.ClassConfigurator;
 import org.jgroups.conf.ProtocolConfiguration;
 import org.jgroups.protocols.TCP;
@@ -40,6 +51,8 @@ import org.jgroups.stack.Protocol;
 import org.jgroups.util.DefaultSocketFactory;
 import org.jgroups.util.SocketFactory;
 import org.keycloak.Config;
+import org.keycloak.config.CachingOptions;
+import org.keycloak.config.Option;
 import org.keycloak.connections.infinispan.InfinispanConnectionProvider;
 import org.keycloak.connections.infinispan.InfinispanConnectionSpi;
 import org.keycloak.connections.jpa.JpaConnectionProvider;
@@ -48,15 +61,10 @@ import org.keycloak.connections.jpa.util.JpaUtils;
 import org.keycloak.infinispan.util.InfinispanUtils;
 import org.keycloak.jgroups.protocol.KEYCLOAK_JDBC_PING2;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.provider.ProviderConfigurationBuilder;
 import org.keycloak.spi.infinispan.JGroupsCertificateProvider;
-
-import javax.net.ssl.KeyManager;
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLParameters;
-import javax.net.ssl.SSLServerSocket;
-import javax.net.ssl.TrustManager;
-
-import static org.infinispan.configuration.global.TransportConfiguration.STACK;
+import org.keycloak.spi.infinispan.impl.Util;
 
 /**
  * Utility class to configure JGroups based on the Keycloak configuration.
@@ -88,6 +96,7 @@ public final class JGroupsConfigurator {
         if (stack != null) {
             transportOf(holder).stack(stack);
         }
+        configureTransport(config);
         configureDiscovery(holder, session);
         configureTls(holder, session);
         warnDeprecatedStack(holder);
@@ -124,6 +133,17 @@ public final class JGroupsConfigurator {
         if (siteName != null) {
             transport.siteId(siteName);
         }
+    }
+
+    static void createJGroupsProperties(ProviderConfigurationBuilder builder) {
+        Util.copyFromOption(builder, SystemProperties.BIND_ADDRESS.configKey, "address", ProviderConfigProperty.STRING_TYPE, CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS, false);
+        Util.copyFromOption(builder, SystemProperties.BIND_PORT.configKey, "port", ProviderConfigProperty.INTEGER_TYPE, CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_PORT, false);
+        Util.copyFromOption(builder, SystemProperties.EXTERNAL_ADDRESS.configKey, "address", ProviderConfigProperty.STRING_TYPE, CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_ADDRESS, false);
+        Util.copyFromOption(builder, SystemProperties.EXTERNAL_PORT.configKey, "port", ProviderConfigProperty.INTEGER_TYPE, CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_PORT, false);
+    }
+
+    private static void configureTransport(Config.Scope config) {
+        Arrays.stream(SystemProperties.values()).forEach(p -> p.set(config));
     }
 
     private static void configureTls(ConfigurationBuilderHolder holder, KeycloakSession session) {
@@ -267,6 +287,71 @@ public final class JGroupsConfigurator {
             if (protocol instanceof KEYCLOAK_JDBC_PING2 kcPing) {
                 kcPing.setJpaConnectionProviderFactory(factory);
             }
+        }
+    }
+
+    private enum SystemProperties {
+        BIND_ADDRESS(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS, Global.BIND_ADDR, "jgroups.bind.address"),
+        BIND_PORT(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_PORT, Global.BIND_PORT, "jgroups.bind.port"),
+        EXTERNAL_ADDRESS(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_ADDRESS, Global.EXTERNAL_ADDR),
+        EXTERNAL_PORT(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_PORT, Global.EXTERNAL_PORT);
+
+        final Option<?> option;
+        final String property;
+        final String altProperty;
+        final String configKey;
+
+        SystemProperties(Option<?> option, String property) {
+            this(option, property, null);
+        }
+
+        SystemProperties(Option<?> option, String property, String altProperty) {
+            this.option = option;
+            this.property = property;
+            this.altProperty = altProperty;
+            this.configKey = configKey();
+        }
+
+        void set(Config.Scope config) {
+            String userConfig = fromConfig(config);
+            if (userConfig == null) {
+                // User property is either already set or missing, so do nothing
+                return;
+            }
+            checkPropertyAlreadySet(userConfig, property);
+            if (altProperty != null)
+                checkPropertyAlreadySet(userConfig, altProperty);
+            System.setProperty(property, userConfig);
+        }
+
+        void checkPropertyAlreadySet(String userValue, String property) {
+            String userProp = System.getProperty(property);
+            if (userProp != null) {
+                logger.warnf("Conflicting system property '%s' and CLI arg '%s' set, utilising CLI value '%s'",
+                      property, option.getKey(), userValue);
+                System.clearProperty(property);
+            }
+        }
+
+        String fromConfig(Config.Scope config) {
+            if (option.getType() == Integer.class) {
+                Integer val = config.getInt(configKey);
+                return val == null ? null : val.toString();
+            }
+            return config.get(configKey);
+        }
+
+        String configKey() {
+            // Strip the scope from the key and convert to camelCase
+            String key = option.getKey().substring(CACHE_EMBEDDED_PREFIX.length() + 1);
+            StringBuilder sb = new StringBuilder(key);
+            for (int i = 0; i < sb.length(); i++) {
+                if (sb.charAt(i) == '-') {
+                    sb.deleteCharAt(i);
+                    sb.replace(i, i+1, String.valueOf(Character.toUpperCase(sb.charAt(i))));
+                }
+            }
+            return sb.toString();
         }
     }
 }
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
@@ -1,5 +1,8 @@
 package org.keycloak.quarkus.runtime.configuration.mappers;
 
+import static org.keycloak.quarkus.runtime.configuration.Configuration.getOptionalKcValue;
+import static org.keycloak.quarkus.runtime.configuration.mappers.PropertyMapper.fromOption;
+
 import java.io.File;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -8,8 +11,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.function.BooleanSupplier;
 
-import com.google.common.base.CaseFormat;
-import io.smallrye.config.ConfigSourceInterceptorContext;
 import org.keycloak.common.Profile;
 import org.keycloak.config.CachingOptions;
 import org.keycloak.config.Option;
@@ -19,8 +20,9 @@ import org.keycloak.quarkus.runtime.cli.PropertyException;
 import org.keycloak.quarkus.runtime.configuration.Configuration;
 import org.keycloak.utils.StringUtil;
 
-import static org.keycloak.quarkus.runtime.configuration.Configuration.getOptionalKcValue;
-import static org.keycloak.quarkus.runtime.configuration.mappers.PropertyMapper.fromOption;
+import com.google.common.base.CaseFormat;
+
+import io.smallrye.config.ConfigSourceInterceptorContext;
 
 final class CachingPropertyMappers {
 
@@ -98,6 +100,26 @@ final class CachingPropertyMappers {
                         .isEnabled(() -> Configuration.isTrue(CachingOptions.CACHE_EMBEDDED_MTLS_ENABLED), "property '%s' is enabled".formatted(CachingOptions.CACHE_EMBEDDED_MTLS_ENABLED.getKey()))
                         .validator(CachingPropertyMappers::validateCertificateRotationIsPositive)
                         .build(),
+                fromOption(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS)
+                        .paramLabel("address")
+                        .to("kc.spi-cache-embedded--default--network-bind-address")
+                        .isEnabled(CachingPropertyMappers::cacheSetToInfinispan, "Infinispan clustered embedded is enabled")
+                        .build(),
+                fromOption(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_PORT)
+                       .paramLabel("port")
+                       .to("kc.spi-cache-embedded--default--network-bind-port")
+                       .isEnabled(CachingPropertyMappers::cacheSetToInfinispan, "Infinispan clustered embedded is enabled")
+                       .build(),
+                fromOption(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_ADDRESS)
+                       .paramLabel("address")
+                       .to("kc.spi-cache-embedded--default--network-external-address")
+                       .isEnabled(CachingPropertyMappers::cacheSetToInfinispan, "Infinispan clustered embedded is enabled")
+                       .build(),
+                fromOption(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_PORT)
+                       .paramLabel("port")
+                       .isEnabled(CachingPropertyMappers::cacheSetToInfinispan, "Infinispan clustered embedded is enabled")
+                       .to("kc.spi-cache-embedded--default--network-external-port")
+                       .build(),
                 fromOption(CachingOptions.CACHE_REMOTE_HOST)
                         .paramLabel("hostname")
                         .to("kc.spi-cache-remote--default--hostname")
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -33,6 +33,14 @@ Starting with this release, and when using the original PostgreSQL driver, {proj
 
 You can override this behavior by setting your own value for `targetServerType` in the DB URL or additional properties.
 
+=== JGroups system properties replaced with CLI options
+
+Until now it was necessary to configure JGroups network addresses and ports using the `+jgroups.bind.*+` and `+jgroups.external_*+`
+system properties. In this release we have introduced the following CLI options to allow these addresses and ports to be
+configured directly via {project_name}: `cache-embedded-network-bind-address`, `cache-embedded-network-bind-port`,
+`cache-embedded-network-external-address`, `cache-embedded-network-external-port`. Configuring ports using the old
+properties will still function as before, but we recommend to change to the CLI options as this may change in the future.
+
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
 
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -321,22 +321,25 @@ To ensure a healthy {project_name} clustering, some network ports need to be ope
 The table below shows the TCP ports that need to be open for the `jdbc-ping` stack, and a description of the traffic that goes through it.
 
 |===
-|Port |Property | Description
+|Port |Option| Property | Description
 
 m|7800
+m|cache-embedded-network-bind-port
 m|jgroups.bind.port
 |Unicast data transmission.
 
 m|57800
+m|
 m|jgroups.fd.port-offset
 |Failure detection by protocol `FD_SOCK2`.
 It listens to the abrupt closing of a socket to suspect a {project_name} server failure.
-The `jgroups.fd.port-offset` property defines the offset from the `jgroups.bind.port`.
+The `jgroups.fd.port-offset` property defines the offset from the `cache-embedded-network-bind-port` option or `jgroups.bind.port` property.
 By default, the offset is set to 50000, making the failure detection port 57800.
 
 |===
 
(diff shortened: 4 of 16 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
```diff
@@ -98,6 +103,24 @@ public static PropertyMapper<?>[] getClusteringPropertyMappers() {
                         .isEnabled(() -> Configuration.isTrue(CachingOptions.CACHE_EMBEDDED_MTLS_ENABLED), "property '%s' is enabled".formatted(CachingOptions.CACHE_EMBEDDED_MTLS_ENABLED.getKey()))
                         .validator(CachingPropertyMappers::validateCertificateRotationIsPositive)
                         .build(),
+                fromOption(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS)
+                        .paramLabel("address")
+                        .to("kc.spi-cache--embedded--default--network-bind-address")
```
shawkins:
```suggestion
                        .to("kc.spi-cache-embedded--default--network-bind-port")
```

The spi is cache-embedded right?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -322,8 +322,15 @@ private StatefulSet createBaseDeployment(Keycloak keycloakCR, Context<Keycloak>
                         && (customImage.isPresent() || operatorConfig.keycloak().startOptimized())) {
             containerBuilder.addToArgs(OPTIMIZED_ARG);
         }
-        // Set bind address as this is required for JGroups to form a cluster in IPv6 envionments
-        containerBuilder.addToArgs(0, "-Djgroups.bind.address=$(%s)".formatted(POD_IP));
+
+        boolean bindAddressRequired = Optional.ofNullable(containerBuilder.getArgs())
+              .orElse(List.of())
+              .stream()
+              .noneMatch(a -> a.contains("--cache-embedded-network-bind-address"));
```
shawkins: This should only match if someone used the unsupported podTemplate to update the main container's args. Do you mean be checking specifically for this, or do you want to check for additionalOptions usage of cache-embedded-network-bind-address?

shawkins (commented):
Looks good. Just a couple of things.

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -44,6 +44,14 @@ If a user tries to create a credential with an already existing name, there is a
 
 To enhance security, only users with the `admin` role in the `master` realm (server admins) can assign admin roles. This ensures that critical permissions cannot be delegated by realm-level administrators.
 
+=== JGroups system properties replaced with CLI options
+
+Until now it was necessary to configure JGroups network addresses and ports using the `jgroups.bind.*` and `jgroups.external_*`
```
ahus1: Nitpick: Escape bits that should be shown literally by using passthrough with plus signs to prevent AsciiDoc from picking up special characters for formatting.

```suggestion
Until now it was necessary to configure JGroups network addresses and ports using the `+jgroups.bind.*+` and `+jgroups.external_*+`
```

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -126,6 +136,42 @@ public static void configureTopology(Config.Scope config, ConfigurationBuilderHo
...
+
+    private static void setJGroupsNetworkProperty(String key, Supplier<String> config) {
+        String userConfig = config.get();
+        if (userConfig == null) {
+            // User property is either already set or missing, so do nothing
+            return;
+        }
+
+        String[] split = key.split("(?=[A-Z])", 3);
+        String property = "jgroups.%s.%s".formatted(split[1].toLowerCase(), split[2].toLowerCase());
```
ahus1: I assume this will work, but is hard to understand for me, and might fail with future changes. Could we have a switch statement instead which those values that we support? If you want to make it fancy, try using enums to make it safe.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
ahus1: Yes, the SPI is `cacheEmbedded` - but note that this should still be `...network-bind-address` not `...-port`.

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testStartDevHelp.approved.txt
```diff
@@ -36,6 +36,14 @@ Cache:
                      The maximum number of entries that can be stored in-memory by the crl cache.
 --cache-embedded-keys-max-count <max-count>
                      The maximum number of entries that can be stored in-memory by the keys cache.
+--cache-embedded-network-bind-address <address>
+                     IP address used by clustering transport.
```
ahus1: Please extends the docs:

```suggestion
--cache-embedded-network-bind-address <address>
                     IP address to bind to for the clustering transport. By default, ...
```

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testStartDevHelp.approved.txt
```diff
@@ -36,6 +36,14 @@ Cache:
...
+--cache-embedded-network-bind-port <port>
+                     Port used by clustering transport.
```
ahus1: Please extend the docs. People should understand that this is usually optional (same for `--cache-embedded-network-bind-address`)

```suggestion
--cache-embedded-network-bind-port <port>
                     Port to bind to for the clustering transport. By default, ...
```

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testStartDevHelp.approved.txt
```diff
@@ -36,6 +36,14 @@ Cache:
...
+--cache-embedded-network-external-address <address>
+                     IP address that other instances in the cluster should use to contact this node.
```
ahus1:
```suggestion
--cache-embedded-network-external-address <address>
                     IP address that other instances in the cluster should use to contact this node. Set only it is different from `cache-embedded-network-bind-address`, for example when this instance is behind a firewall.
```

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testStartDevHelp.approved.txt
```diff
@@ -36,6 +36,14 @@ Cache:
...
+--cache-embedded-network-external-port <port>
+                     Port that other instances in the cluster should use to contact this node.
```
ahus1:
```suggestion
--cache-embedded-network-external-port <port>
                     Port that other instances in the cluster should use to contact this node. Set only if ...
```

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ryanemerson: Good point, looking at this again I realise there's actually a bug as the external address properties are in the format `jgroups.%s_%s` not `jgroups.%s.%s`.

I have reworked to utilise an Enum.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
ryanemerson: Fixed to use `cache-embedded` for the various options.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
ryanemerson: Thanks, I had forgot about `additionalOptions`. I have updated the logic so that we only set the bind-address explicitly if it's not already specified via `additionalOptions` or a custom command via unsupported podTemplate.

I have added test cases to `operator/src/test/java/org/keycloak/operator/testsuite/unit/PodTemplateTest.java`.

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testStartDevHelp.approved.txt
ryanemerson: I have pushed a commit updating the various descriptions.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
```diff
@@ -98,6 +103,24 @@ public static PropertyMapper<?>[] getClusteringPropertyMappers() {
                         .isEnabled(() -> Configuration.isTrue(CachingOptions.CACHE_EMBEDDED_MTLS_ENABLED), "property '%s' is enabled".formatted(CachingOptions.CACHE_EMBEDDED_MTLS_ENABLED.getKey()))
                         .validator(CachingPropertyMappers::validateCertificateRotationIsPositive)
                         .build(),
+                fromOption(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS)
+                        .paramLabel("address")
+                        .to("kc.spi-cache-embedded--default--network-bind-address")
+                        .validator(CachingPropertyMappers::validateBindAddress)
```
pruivo: Add `isEnabled` to all the mapping
`.isEnabled(CachingPropertyMappers::cacheSetToInfinispan, "Infinispan clustered embedded is enabled")`

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
```diff
@@ -256,4 +279,24 @@ private static void validateCertificateRotationIsPositive(String value) {
...
+        String matchType = address.split(":")[0];
+        if (Global.MATCH_ADDR.equals(matchType) || Global.MATCH_HOST.equals(matchType) || Global.MATCH_INTF.equals(matchType))
+            return;
+
+        throw new PropertyException("Option '%s'. Invalid address: '%s'".formatted(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS.getKey(), address));
+    }
+
+    private static void validateExternalAddress(String address) {
+        if (!InetAddressUtils.isIPv4Address(address) && !InetAddressUtils.isIPv6Address(address))
+            throw new PropertyException("Option '%s'. Invalid address: '%s'".formatted(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_ADDRESS.getKey(), address));
```
pruivo: I wouldn't do any validation in Keycloak as it restricts what can be set. For example, you cannot put the network interface, but `-Djgroups.bind.address=eth0` or  `-Djgroups.bind.address=localhost` are both valid.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -269,4 +288,56 @@ public void afterCreation(Protocol protocol) {
             }
         }
     }
+
+    private enum SystemProperties {
+        BIND_ADDRESS(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_ADDRESS, "jgroups.bind.address"),
+        BIND_PORT(CachingOptions.CACHE_EMBEDDED_NETWORK_BIND_PORT, "jgroups.bind.port"),
+        EXTERNAL_ADDRESS(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_ADDRESS, "jgroups.external_addr"),
+        EXTERNAL_PORT(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_PORT, "jgroups.external_port");
```
pruivo:
```suggestion
        EXTERNAL_ADDRESS(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_ADDRESS, Global.EXTERNAL_ADDR),
        EXTERNAL_PORT(CachingOptions.CACHE_EMBEDDED_NETWORK_EXTERNAL_PORT, Global.EXTERNAL_PORT);
```

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -269,4 +288,56 @@ public void afterCreation(Protocol protocol) {
...
+
+        final Option<?> option;
+        final String property;
```
pruivo, line 300: This needs to be a list. `jgroups.bind.address` is an Infinispan property, but `public static final String BIND_ADDR="jgroups.bind_addr";` needs to be set too. Same for the bind port. 
Using a custom JGroups configuration file won't be able to read the Infinispan properties.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -322,8 +322,10 @@ private StatefulSet createBaseDeployment(Keycloak keycloakCR, Context<Keycloak>
                         && (customImage.isPresent() || operatorConfig.keycloak().startOptimized())) {
             containerBuilder.addToArgs(OPTIMIZED_ARG);
         }
-        // Set bind address as this is required for JGroups to form a cluster in IPv6 envionments
-        containerBuilder.addToArgs(0, "-Djgroups.bind.address=$(%s)".formatted(POD_IP));
+        if (isBindAddressRequired(containerBuilder.getArgs(), keycloakCR.getSpec().getAdditionalOptions())) {
+            // Set bind address as this is required for JGroups to form a cluster in IPv6 environments
+            containerBuilder.addToArgs("--cache-embedded-network-bind-address=$(%s)".formatted(POD_IP));
+        }
```
pruivo: This is confusing to me... why would anyone set the bind address in the Keycloak CR!?

pruivo: I would remove this line and change the environment variable name in line:
https://github.com/keycloak/keycloak/blob/7736ca20e9d55bb8d58d79cd8fc08c5c8bb01899/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java#L537
i.e. replace `POD_ID` with `KC_CACHE_EMBEDDED_NETWORK_BIND_ADDRESS`

pruivo (changes requested):

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ryanemerson: Me and Pedro discussed this and the `jgroups.bind_addr` system property always takes precedence over the Infinispan property `jgroups.bind.address` and any values configured in the JGroup's stack, so it's acceptable to just set the `jgroups.bind_addr` property.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
ryanemerson: A user specified bind address doesn't make sense in the context of the Operator, as it will result in the same address being used by all pods, therefore I have reverted ^ in favour of Pedro's suggestion: https://github.com/keycloak/keycloak/pull/40690#discussion_r2169018805

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ryanemerson: I have also made it so that we warn if either `jgroups.bind_addr` or `jgroups.bind.address` is set with the CLI option.

pruivo (approved):

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -534,7 +532,7 @@ private List<EnvVar> getDefaultAndAdditionalEnvVars(Keycloak keycloakCR) {
             }
         }
 
-        envVars.add(new EnvVarBuilder().withName(POD_IP).withNewValueFrom().withNewFieldRef()
```
shawkins: There is a small chance this could be seen as a breaking change - if users assume that the POD_ID environment variable is always available.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -83,7 +83,7 @@
 )
 public class KeycloakDeploymentDependentResource extends CRUDKubernetesDependentResource<StatefulSet, Keycloak> {
 
-    public static final String POD_IP = "POD_IP";
+    public static final String BIND_ADDRESS = "KC_CACHE_EMBEDDED_NETWORK_BIND_ADDRESS";
```
shawkins: What is the expectation for users who are specifying an image - won't there be an incompatibility with existing images with the new operator as this option won't yet exist? 

I know we don't go into detail about this scenario in the general docs, but for the most part changes in the operator tend to be backwards compatible at least to last supported major version, or dependent upon specific enablement.

If we don't want version dependent logic in the operator, we'll at least need instructions to update the images first, before upgrading the operator.

cc @vmuzikar @ahus1

ryanemerson: > What is the expectation for users who are specifying an image - won't there be an incompatibility with existing images with the new operator as this option won't yet exist?

I feel like we're in very murky water here, given that the Operator doesn't officially support multiple Keycloak versions but it seems we don't want to break things when using older server versions. It would be great if this was documented somewhere :slightly_smiling_face:

> If we don't want version dependent logic in the operator, we'll at least need instructions to update the images first, before upgrading the operator.

By update the images first, I assume you mean to a version with `--cache-embedded-network-bind-address`?

I don't think the user should ever be baking things related to the bind address into the image. Unless the user is providing a special JGroups String, such as "SITE_LOCAL", this value will typically be different for every container, hence the use of POD_IP.

We were previously relying on `-Djgroups.bind.address` being set as the first container arg, even when an unsupported podTemplateSpec was provided. AFAIU this means the only way a user could override this property would be to:

1. Explicitly provide `-Djgroups.bind.address` as part of `JAVA_OPTS_APPEND`
2. Configure `-Djgroups.bind_addr` as this takes precedence over `-Djgroups.bind.address` at the JGroups level



> If we don't want version dependent logic in the operator, we'll at least need instructions to update the images first, before upgrading the operator.

An alternative is that we fallback to using `-Djgroups.bind.address` in the Operator and make the transition to the CLI args as part of KC 27.0.

The `-Djgroups.bind.address` property will still work with my changes. The user will get a warning if they also configure `--cache-embedded-network-bind-address`, but this should never be necessary with the Operator anyway.

shawkins: > By update the images first, I assume you mean to a version with --cache-embedded-network-bind-address?

Yes, rebuilding against Keycloak 26.4 prior to upgrading the operator to 26.4.

Otherwise KC_CACHE_EMBEDDED_NETWORK_BIND_ADDRESS won't be recognized / used.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
ahus1: I would argue this is a notable change, as it changes an internal behavior. We have never stated that the Operator will put that to the environment variables. So as long as we document it as a notable change, we should be fine.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
ahus1: As it is not officially supported to run different minor releases in KC and the Operator, I then recommend to add it to notable changes. As this was added for people running in IPv6 only environments, this restricts the audience even further. 

Is there a planned docs on how to upgrade Keycloak with custom images and the Operator?

shawkins: > As it is not officially supported to run different minor releases in KC and the Operator

The only mention we have of this is a note on https://www.keycloak.org/operator/customizing-keycloak "To ensure full compatibility of Operator and Operand, make sure that the version of Keycloak release used in the custom image is aligned with the version of the operator." - which could even imply that you need to update your custom images before doing a micro update of the operator.

> Is there a planned docs on how to upgrade Keycloak with custom images and the Operator?

The OLM instructions were updated to warn the user to prevent automatic upgrades. As far as I know we haven't elaborated, and don't have an issue captured, on all the steps necessary to actually upgrade the operator. Database backup and restore is especially hand-wavy. Some of this should be covered under https://github.com/keycloak/keycloak/issues/37356

ahus1: The [upgrading guide](https://www.keycloak.org/docs/latest/upgrading/index.html#_prep_migration) states as the first step: 

> Shutdown Keycloak.

I would say this is true for any upgrade of a minor version as of today, and we are about to fix this for patch releases. 

So applying this together with the rule of align the version of KC and the KC Operator (at least with the same minor/major version):

* Scale deployment to 0 
* Upgrade the operator
* Update the KC image in the CR
* Scale up the number of replicas via the CR

I wouldn't push for that, and we can defer that operator change to 26.5 release to allow for a more rolling update way at least when people don't skip minor releases. 

Let me know your thoughts.

shawkins: For minor and micro releases I would want the sequence to be:

- update your CRs to new images - I know it's not considered supported, but there are few circumstances where this won't work, so we should push for this instead.
  - for micro, and even some validated minor releases, the expectation is that this should be rolled in. For everything else a recreate would happen.
  - the admin is in control of the blast radius of this change. It does not require all Keycloaks have a coordinated outage window
- Upgrade the operator
  - there is still a potential for a wide blast radius as all managed Keycloaks will be updated simultaneously, but for those with a custom image hopefully we've limited that to just rolling restarts

cc @mabartos @vmuzikar

ahus1: I suppose what you suggest here will work with the change @ryanemerson is proposing here: 

* The old config to pass the POD_IP via the Java System Property is understood by the new Keycloak version - I hope @ryanemerson can confirm that. 

AFAIK the KC Operator is installed per namespace, and I would imagine people would usually not run multiple Keycloaks in a namespace. Even if they do, given the coupling between the operator image and the Keycloak image, they should be all the same version. 

I give my +1 to have the procedure @shawkins described above documented in a follow-up issue in the CND team.

pruivo: JGroups will always check the system properties and use its value if it is set; it has priority over every other configuration path. The changes to the operator code can be reverted without a problem.

ryanemerson: +1 to @shawkins suggestion. The only issue I can see is if the server dropped some functionality that the Operator relied upon, e.g. when upgrading from 26.x to 27.x, however I think that's very much an edge case that we can take measures to prevent.

> * The old config to pass the POD_IP via the Java System Property is understood by the new Keycloak version - I hope @ryanemerson can confirm that.

Yes, that is the case. Just with the caveats I describe in https://github.com/keycloak/keycloak/pull/40690#discussion_r2180266346


> AFAIK the KC Operator is installed per namespace, and I would imagine people would usually not run multiple Keycloaks in a namespace. Even if they do, given the coupling between the operator image and the Keycloak image, they should be all the same version.

The problem with the Operator framework is that it relies on CRDs which are installed in the global namespace of K8s, so mixing and matching Operator versions doesn't work well in practice and is not recommended. These limitations were a big motivator for some of the changes in [OLM V1](https://www.redhat.com/en/blog/announcing-olm-v1-next-generation-operator-lifecycle-management).

ryanemerson: I have pushed a commit reverting the Operator changes and I have created https://github.com/keycloak/keycloak/issues/40884 to track modifying the Operator in the future.

vmuzikar: Just wanted to share my 2 cents here.

Officially, we do not support using older KC image with newer Operator (or vice versa). That said, we do not have a good documented process how to upgrade when using custom images. So we're trying to avoid doing breaking changes in the Operator when not needed. And the current breaking changes policy in the server helps us with that: we're not doing any in minor releases. That enables us to use newer server version with older operator. Hence the approach @shawkins makes sense to me to document.

That said, this only works for minor releases. I think we need to make it clear that for major upgrades (where we will not even consider supporting zero-downtime upgrade), the user must scale down their deployment before upgrading the Operator.

This might change in the future if we make the operator aware of the server version to support using different server versions with different operator versions. But for now, clear docs is IMHO the way to go.

ahus1:
@vmuzikar - with the Operator changes removed, I think the open questions are now resolved. Can you please confirm that adding the CLI options is ok for the CND team?

```
--cache-embedded-network-bind-address <address>
                     IP address used by clustering transport. By default, SITE_LOCAL is used.
                       Available only when Infinispan clustered embedded is enabled.
--cache-embedded-network-bind-port <port>
                     The Port the clustering transport will bind to. By default, port 7800 is used.
                       Available only when Infinispan clustered embedded is enabled.
--cache-embedded-network-external-address <address>
                     IP address that other instances in the cluster should use to contact this
                       node. Set only if it is different to cache-embedded-network-bind-address,
                       for example when this instance is behind a firewall. Available only when
                       Infinispan clustered embedded is enabled.
--cache-embedded-network-external-port <port>
                     Port that other instances in the cluster should use to contact this node. Set
                       only if it is different to cache-embedded-network-bind-port, for example
                       when this instance is behind a firewall Available only when Infinispan
                       clustered embedded is enabled.
```

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
```diff
@@ -44,6 +44,14 @@ If a user tries to create a credential with an already existing name, there is a
...
 To enhance security, only users with the `admin` role in the `master` realm (server admins) can assign admin roles. This ensures that critical permissions cannot be delegated by realm-level administrators.
 
+=== JGroups system properties replaced with CLI options
+
+Until now it was necessary to configure JGroups network addresses and ports using the `+jgroups.bind.*+` and `+jgroups.external_*+`
+system properties. In this release we have introduced the following CLI options to allow these addresses and ports to be
+configured directly via {project_name}: `cache-embedded-network-bind-address`, `cache-embedded-network-bind-port`,
+`cache-embedded-network-external-address`, `cache-embedded-network-external-port`. Configuring ports using the old
+properties will still function as before, but we recommend to change to the CLI options as this may change in the future.
+
```
ahus1: This would need to go to `changes-26_4_0.adoc` as 26.3 has been released by now. Please update the PR

ahus1 (changes requested):
Thank you for the PR, looks good to me.

vmuzikar (dismissed):
@ahus1 Yes, the options work for me. I haven't deeply reviewed other changes, though.

docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
ryanemerson: I have rebased and squashed on the latest `main`.

ahus1 (approved):

PR #40623, opened 2025-06-19
Create clustering test cases for OIDC flows

ryanemerson:
Closes #39965

PR #40691, opened 2025-06-24
Fix wrong dependent registering in Registry

mhajas:
Closes #40756

While looking into this:
https://github.com/keycloak/keycloak/pull/40623#discussion_r2161345556

I noticed there seems to be an inconsistency in Registry between:
https://github.com/keycloak/keycloak/blob/c01736a9cdf95ddc0c020abcd7b6550c3a94ed46/test-framework/core/src/main/java/org/keycloak/testframework/injection/Registry.java#L87

and
https://github.com/keycloak/keycloak/blob/c01736a9cdf95ddc0c020abcd7b6550c3a94ed46/test-framework/core/src/main/java/org/keycloak/testframework/injection/Registry.java#L188

The direction of registering dependency is different between these two calls. This was causing the issue in the discussion above, where with `BEFORE_REALM` it worked while with `BEFORE_KEYCLOAK_SERVER` it did not.

With these changes the cleaning is done in the correct order.

This PR is adding:
- switching the direction of registered dependency to be consistent
- changing the name from dependency to dependent as it seems to be more correct term
- adding a bunch of additional debug logging that helped me investigate this

Please let me know whether this is a relevant issue that we want to fix and if yes I will create a GHI

### DIFF
--- a/test-framework/core/src/main/java/org/keycloak/testframework/injection/AbstractInterceptorHelper.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/injection/AbstractInterceptorHelper.java
@@ -22,9 +22,9 @@ public abstract class AbstractInterceptorHelper<I, V> {
             value = intercept(value, interception.supplier, interception.existingInstance);
             registry.getLogger().logIntercepted(value, interception.supplier);
             if (interception.existingInstance != null) {
-                interception.existingInstance.registerDependency(instanceContext);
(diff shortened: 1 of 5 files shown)


stianst:
Closing this as I'm going to replace it with a different approach. The renaming makes sense, but the lifecycle management needs to be a bit different so updating to require that suppliers declare their dependencies up front. This allows compatibility checking and also detecting cycling dependencies.

PR #40698, opened 2025-06-24
Resolve resources with same URI if the permission request is based on URI matching

pedroigor:
Closes #40695


sguilhen (approved):

ahus1 (approved):

PR #40721, opened 2025-06-25
Do not show update email link if the email attribute is not writable

### LINKED ISSUE #39669: Hide update email link in account console when email is read-only in user profile
### Before reporting an issue

### Area

user-profile

### Describe the bug

Currently, there is a link to update email in account console when email is read-only for the user in user profile. When user tries to update the email, new form is opened but email can't be updated anyway. It would be better to hide the link. 

### Version

main

### Regression

### Expected behavior

Hide the link

### Actual behavior

see description

### How to Reproduce?

see description

### Anything else?

_No response_

pedroigor:
Closes #39669

* Certain attributes, like the `email`, might only be updated through an AIA. In this case, the UP configuration will allow setting a "annotation decorator" to dynamically resolve annotations based on contextual data so that UIs can check whether or not a link to initiate the action should be rendered.
* The reason why this issue is resolved using an annotation is because the attribute is indeed readonly when at the `ACCOUNT` context, and we can not rely on the `org.keycloak.representations.idm.UserProfileAttributeMetadata#readOnly` field to decide to show or hide the link. In addition to that, annotations serve as a hint to UIs about how to render attributes.

### DIFF
--- a/js/apps/account-ui/src/personal-info/PersonalInfo.tsx
+++ b/js/apps/account-ui/src/personal-info/PersonalInfo.tsx
@@ -119,11 +119,15 @@ export const PersonalInfo = () => {
             ((key: unknown, params) =>
               t(key as TFuncKey, params as any)) as TFunction
           }
-          renderer={(attribute) =>
-            attribute.name === "email" &&
-            updateEmailFeatureEnabled &&
-            updateEmailActionEnabled &&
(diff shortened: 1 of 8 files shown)


martin-kanis (approved):
LGTM

ahus1 (approved):

PR #40723, opened 2025-06-25
fix: refinements for ipv6 only testing

### LINKED ISSUE #21277: Support IPv6 only environments
### Description

As a follow-up to https://github.com/keycloak/keycloak/issues/15003 we should also support IPv6 environments where Keycloak is deployed.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
With these changes and running maven with -Djava.net.preferIPv6Addresses, the tests should run successfully. 

As part of this PR, or a seperate task we can see about creating a smaller ipv6 test run like we discussed initially.

closes: #21277


vmuzikar (approved):
LGTM, thanks.

PR #40729, opened 2025-06-26
Allow passkeys login when user has no password credential

### LINKED ISSUE #40717: Allow passkeys login when user has no password credential
### Before reporting an issue

### Area

authentication/webauthn

### Describe the bug

Continuation of #40021. Currently when a user that only has the passwordless webauthn credential tries to login with the separated username and password form, the following error happens: `Cannot login, credential setup required`.

This is because the `configuredFor` of the password form just returns true when the password is setup ([here](https://github.com/keycloak/keycloak/blob/26.2.5/services/src/main/java/org/keycloak/authentication/authenticators/browser/PasswordForm.java#L47)). We have to add the case when the user authenticated using passkeys.

### Version

999-SNAPSHOT

### Regression

### Expected behavior

The user should login OK.

### Actual behavior

Error: `Cannot login, credential setup required`

### How to Reproduce?

1. Configure username and password using separated forms.
2. Create a user and register a passkey.
3. Remove the password using the admin UI.
4. Try to login using the passkey.

### Anything else?

_No response_

Romain7495:
closes: #40717

Hi, @rmartinc, this is a first proposal to allow login when user has no password 

An alternative should be to compare directly `lastCredential` with `webauthn-passwordless`

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/browser/PasswordForm.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/browser/PasswordForm.java
@@ -53,8 +53,15 @@ public class PasswordForm extends UsernamePasswordForm implements CredentialVali
 
     @Override
     public boolean configuredFor(KeycloakSession session, RealmModel realm, UserModel user) {
-        return user.credentialManager().isConfiguredFor(getCredentialProvider(session).getType())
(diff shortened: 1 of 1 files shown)


rmartinc:
@Romain7495 I have already fixed this in https://github.com/keycloak/keycloak/pull/40719. :smile: Give it a try!

Romain7495:
@rmartinc just saw with the diff of the PR 😂 

Thanks, it works perfectly

PR #40735, opened 2025-06-26
Move section in the upgrading guide

### LINKED ISSUE #39063: Optimized startup fails from `kc.spi-connections-http-client-default-expect-continue-enabled` passed at runtime

ahus1:
Closes #39063

This moves the section as it is IMHO not a notable change, but a deprecation. As we would eventually remove support for the old format, this is IMHO the right section.

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
@@ -19,18 +19,6 @@ user profile configuration where too much information was returned in the past.
 
 Notable changes where an internal behavior changed to prevent common misconfigurations, fix bugs or simplify running {project_name}.
 
-=== SPI Options naming
-
-SPI options ending in `-enabled`, `-provider-default`, or `-provider` are treated as buildtime options. However, in some instances, this was not correct as a provider could have a configuration property ending in one of those suffixes as well.
-
-To resolve this ambiguity, and any potential ambiguity involving SPI and provider names, a new SPI option format was introduced where the scopes and suffix are separated by `--`(double dash) instead of `-`(dash).
-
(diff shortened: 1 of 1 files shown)


vmuzikar (approved):
@ahus1 Thank you for the PR.

I'm not sure I'd say this is a deprecation note. We've introduced a new format while effectively deprecating the old. So it's kinda a new feature and a deprecation. A notable change sounds more descriptive to me, though I don't have a strong opinion about it, so approving regardless.

ahus1:
@vmuzikar - I hear you. AFAIK there isn't yet a warning for all SPI options in the old format, and we also don't provide a note that all SPI options (regardless if their suffix is for "enabled" or one of the other special ones). With putting this section into the "deprecated" area, it is now official that we're eventually stop supporting it.

PR #40740, opened 2025-06-26
Rework floating promises to avoid concurrency side effects

### LINKED ISSUE #40739: Avoid floating promises in UI code
### Description

In #40494 there was an issue with concurrency due to a floating promise. I found that there is an eslint check than can help prevent those. 

### Discussion

#40494

### Motivation

Looking at places where it raises its concerns, this is seems that some of the tests could benefit from an occasional `await`, which might actually make them more stable, as they assume some non-concurrent actions? 

### Details

I'll create a PR.

ahus1:
Closes #40739

### DIFF
--- a/js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
+++ b/js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
@@ -51,6 +51,7 @@ export const Oid4Vci = () => {
 
   useEffect(() => {
     if (initialSelected !== selected && credentialsIssuer !== undefined) {
+      // eslint-disable-next-line
       requestVCOffer(context, selectOptions[selected], credentialsIssuer).then(
         (blob) => {
           const reader = new FileReader();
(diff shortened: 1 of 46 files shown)


ahus1:
@keycloak/ui - I would be happy to receive a feedback on this approach. Thanks!

js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
```diff
@@ -51,6 +51,7 @@ export const Oid4Vci = () => {
 
   useEffect(() => {
     if (initialSelected !== selected && credentialsIssuer !== undefined) {
+      // eslint-disable-next-line
```
jonkoops: Instead of disabling all rules for the next line, it would be better to disable the specific rule.
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
```

jonkoops (dismissed):
One comment, but LGTM

ahus1:
@jonkoops - thanks, please re-review and approve when you have the time.

jonkoops (approved):
LGTM, thanks @ahus1, this has been a pet peeve of mine for a while.

Issue #39279, opened 2025-04-28
[OID4VCI] Return `credential_identifiers` in the Token Response as per the spec and enable later use in the Credential Request


IngridPuppet:
### Description

See https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-identifying-credentials-bei

Mainly:
- Return `credential_identifiers` in the Token Response
- Update Credential Endpoint to work when presented `credential_identifiers`

Associated changelog entry:
>  add an option to return credential_identifiers in the Token Response and use them in the Credential Request, when scopes are used in the Authorization Request.

Also, related changelog entry:
> make `credential_identifiers` mandatory for `authorization_details` flow

Awambeng:
I'm currently working on this

Ogenbertrand:
This ticket: https://github.com/keycloak/keycloak/issues/39278 is already returning the Credential Identifier in the Token Response, IDT we need this anymore.

Awambeng:
I was working on this ticket until after the discussion with @forkimenjeckayang, who mentioned that he was already implementing a similar solution. I agreed for him to proceed, so I suggest we keep this ticket open until his PR is merged, after which we can close it.

Issue #39278, opened 2025-04-28
[OID4VCI] Add support for parsing and understanding `authorization_details` at the Token Endpoint


IngridPuppet:
### Description

See https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-request-credential-issuance:
> The Wallet can use `authorization_details` in the Token Request to request a specific Credential Configuration in both the Authorization Code Flow and the Pre-Authorized Code Flow. The value of the `authorization_details` parameter is defined in (#(authorization-details)).

forkimenjeckayang:
WIll start working on this

PR #40751, opened 2025-06-27
[OID4VCI]: Add support for parsing and understanding authorization_details at the Token Endpoint

forkimenjeckayang:
## Summary
This PR implements spec-compliant support for parsing and understanding the `authorization_details` parameter at the Token Endpoint for OpenID4VCI (OpenID for Verifiable Credential Issuance).

### 1. Authorization Details Processing
- Full support for parsing the `authorization_details` parameter in token requests.
- Supports both `credential_configuration_id` and `format`-based authorization details, as per the latest OID4VCI spec.
- Validates the structure of authorization details and prevents invalid combinations.

### 2. Credential Identifier Generation & Management
- Generates unique credential identifiers (UUID) for each credential issuance request.
- Credential identifiers are managed ephemerally within the client session context, which stores the entire `authorization_details` parameter (including credential identifiers).
- Ensures identifiers are reused only within the same client session context, and not across sessions.

### 3. Grant Type Support
- Fully supports both `PreAuthorizedCode` and `AuthorizationCode` grant types for credential issuance.
- Ensures correct handling and validation of `authorization_details` for both flows.

### 4. Response Enhancement
- Returns the processed `authorization_details` (including generated credential identifiers) in token responses.
- Maintains backward compatibility with the existing token response format.
- Includes comprehensive validation and error handling for all edge cases.

### 5. Test Suite Improvements
- Refactored and expanded the test suite to cover both JWT and SD-JWT credential flows.
- Added robust positive and negative tests for both pre-authorized and authorization code flows.
- Ensured all tests are spec-compliant and validate both the request and response handling of `authorization_details`.

## Closses
Issue https://github.com/keycloak/keycloak/issues/39278
Issue https://github.com/keycloak/keycloak/issues/39279


services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetailResponse.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+package org.keycloak.protocol.oid4vc.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+/**
+ * Represents an authorization_details object in the Token Response as per OID4VCI.
+ */
```
wistefan: Not sure if I misunderstand something her, but where in the spec are ```format``` and ```vct``` defined? As far as I see, [6.2](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-successful-token-response) defines ```credential_identifiers``` and ```credential_configurations_supported``` and everything from [5.1.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#authorization-details)(e.g. ```type```, ```credential_configuration_id``` and  ```claims```)

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -207,9 +221,80 @@ public Response process(Context context) {
...
+            } else {
+                throw re;
+            }
+        }
+
+        // Add authorization_details to the response
+        List<AuthorizationDetailResponse> authDetailsResponse = clientSessionCtx.getAttribute(AUTHORIZATION_DETAILS_RESPONSE_KEY, List.class);
+        if (authDetailsResponse != null) {
+            try {
+                Map<String, Object> responseMap = objectMapper.convertValue(res, new TypeReference<>() {
```
wistefan: Why do you convert this to a map and add, instead of using the "otherClaims" property of the AccessTokenResponse?

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -207,9 +221,80 @@ public Response process(Context context) {
...
+        clientSessionCtx.setAttribute(AUTHORIZATION_DETAILS_RESPONSE_KEY, authorizationDetailsResponse);
+
         return createTokenResponse(user, userSession, clientSessionCtx, scopeParam, true, s -> {return new TokenResponseContext(formParams, parseResult, clientSessionCtx, s);});
     }
 
+    // Override createTokenResponse to include authorization_details
+    @Override
+    protected Response createTokenResponse(UserModel user, UserSessionModel userSession, ClientSessionContext clientSessionCtx,
+                                           String scopeParam, boolean code, Function<TokenManager.AccessTokenResponseBuilder, ClientPolicyContext> clientPolicyContextGenerator) {
+        clientSessionCtx.setAttribute(Constants.GRANT_TYPE, context.getGrantType());
```
wistefan: I think its not a good solution to copy so much code from the superclass. The duplication makes it harder to maintain and is error prone in case of changes. I would either add an abstract method, handling authorization details to the OAuth2GrantTypeBase and implement just the stuff you added their or (if it is applicable to all implementations) implement it directly in the superclass

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+            AuthorizationDetailResponse responseDetail = buildAuthorizationDetailResponse(detail, userSession, supportedCredentials, supportedFormats);
+            authDetailsResponse.add(responseDetail);
+        }
+
+        return authDetailsResponse;
+    }
+
+    private List<AuthorizationDetail> parseAuthorizationDetails(String authorizationDetailsParam) {
+        try {
+            return objectMapper.readValue(authorizationDetailsParam, new TypeReference<List<AuthorizationDetail>>() {
```
wistefan: You can use JsonSerialization.readValue(...) from org.keycloak.util instead of instantiating a new ObjectMapper(which is expensive) for this class

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+            });
+        } catch (Exception e) {
+            logger.warnf(e, "Invalid authorization_details format: %s", authorizationDetailsParam);
+            throwInvalidRequest("Invalid authorization_details format");
+            throw new IllegalStateException("Unreachable");
```
wistefan: I think its more readable to have the "throwInvalidRequest" method just returning the exception here(e.g. ```throw getInvalidRequestException(...)) instead of the throw method and the unreachable throw in the line below

mposolda:
@forkimenjeckayang Could you please rebase on top of latest main? There was a bigger PR merged (by @Captain-P-Goldfish ) and it can be likely good to rebase on top of latest main (where that PR is included) to make sure tests are still OK.

I hope to review the PR later next week.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+    }
+
+    private AuthorizationDetailResponse buildAuthorizationDetailResponse(AuthorizationDetail detail, UserSessionModel userSession, Map<String, SupportedCredentialConfiguration> supportedCredentials, List<String> supportedFormats) {
+        String credentialConfigurationId = detail.getCredentialConfigurationId();
+        String format = detail.getFormat();
+        String vct = detail.getVct();
+
+        if (credentialConfigurationId != null) {
+            // Check for existing credential identifier
+            String noteKey = "credential_identifier_" + credentialConfigurationId;
```
wistefan: Since this is used in 2 places(e.g. when the note is created and here), it would be better maintainable to define the string template as a constant somewhere and use it in both places.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+            List<String> credentialIdentifiers = new ArrayList<>();
+            synchronized (userSession) {
```
wistefan: Why is it required to synchronize here?

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -91,18 +96,55 @@ public Response process(Context context) {
...
+                            "can not get encryption KEK", Response.Status.BAD_REQUEST);
+                } else {
+                    throw re;
+                }
+            }
+
+            // If authorization_details is present, serialize the response and add it
+            if (authorizationDetailsResponse != null) {
+                try {
+                    Map<String, Object> responseMap = objectMapper.convertValue(tokenResponse, new TypeReference<Map<String, Object>>() {
```
wistefan: Same as in the AuthorizationCodeGrantType, just use the "otherClaims"

wistefan (commented):
I added some comments. I'm on vacation for the next 3 weeks, so someone else might review the answers.

wistefan:
It might be good if @dteleguin can also have a look here.

services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetailResponse.java
forkimenjeckayang: You're right on this. `format` and `vct` shouldn't be part of the `authorization_details` parameter in the token resposne

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
forkimenjeckayang: right

services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetailResponse.java
IngridPuppet: Please keep in mind we're implementing Draft 15, not the latest version. Draft 15 indeed includes format and format-specific fields. https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#authorization-details

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
forkimenjeckayang: The synchronization is used here to prevent a potential race condition when multiple concurrent requests for the same user session attempt to read and/or set the credential identifier note. Without synchronization, it’s possible that two threads could both see that the note is missing, both generate a new identifier, and both set it—resulting in inconsistent or duplicate identifiers for the same session and credential configuration.

services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetailResponse.java
forkimenjeckayang: I got confused for a sec, You're right @IngridPuppet . In the case where the `credential_configuration_id` is not used , `format` will be used and subsequently we'll have `vct` set.

services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetail.java
```diff
@@ -0,0 +1,81 @@
...
+    private String type;
+
+    @JsonProperty("credential_configuration_id")
+    private String credentialConfigurationId;
+
+    @JsonProperty("format")
+    private String format;
+
+    @JsonProperty("vct")
+    private String vct;
```
IngridPuppet: The `vct` field is one specific to the SD-JWT VC format. You would want to refactor the implementation so format-specific fields could be supported for other formats in the future. For a broad overview, see "Authorization Details" subsections under https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-credential-format-profiles.

services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetailResponse.java
```diff
@@ -0,0 +1,82 @@
...
+public class AuthorizationDetailResponse {
+
+    @JsonProperty("type")
+    private String type;
+
+    @JsonProperty("credential_configuration_id")
+    private String credentialConfigurationId;
```
IngridPuppet: Considering the spec reads:

> In addition to the parameters defined in [Section 5.1.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#authorization-details), this specification defines the following parameter to be used with the authorization details type openid_credential in the Token Response: `credential_identifiers`.

Isn't it appropriate to make this response model simply extend `AuthorizationDetail`?

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -207,9 +221,80 @@ public Response process(Context context) {
...
+        AccessToken token = tokenManager.createClientAccessToken(session, realm, client, user, userSession, clientSessionCtx);
+
```
IngridPuppet: In addition to the point raised by @wistefan regarding code duplication here, please take into account that with some refactoring, it should be possible to call the base class implementation when supercharging it. We'd agree it is really important not to be invasive in this class.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -69,6 +79,13 @@ public abstract class OAuth2GrantTypeBase implements OAuth2GrantType {
 
     private static final Logger logger = Logger.getLogger(OAuth2GrantTypeBase.class);
 
+    protected static final ObjectMapper objectMapper = new ObjectMapper();
```
IngridPuppet: Maybe use `JsonSerialization.mapper`.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+        String vct = detail.getVct();
+
+        // If authorization_servers is present, locations must be set to issuer identifier
+        if (authorizationServers != null && !authorizationServers.isEmpty() && OPENID_CREDENTIAL_TYPE.equals(type)) {
+            List<String> locations = detail.getLocations();
+            if (locations == null || locations.size() != 1 || !issuerIdentifier.equals(locations.get(0))) {
+                logger.warnf("Invalid locations field in authorization_details: %s, expected: %s", locations, issuerIdentifier);
+                throwInvalidRequest("Invalid authorization_details");
+            }
+        }
```
IngridPuppet: Maybe add to the error message that the issue is with the locations field.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+
+        // Validate type
+        if (!OPENID_CREDENTIAL_TYPE.equals(type)) {
+            logger.warnf("Invalid authorization_details type: %s", type);
+            throwInvalidRequest("Invalid authorization_details");
+        }
```
IngridPuppet: Maybe add to the error message the value expected.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+
+        // Ensure exactly one of credential_configuration_id or format is present
+        if ((credentialConfigurationId == null && format == null) || (credentialConfigurationId != null && format != null)) {
+            logger.warnf("Exactly one of credential_configuration_id or format must be present. credentialConfigurationId: %s, format: %s", credentialConfigurationId, format);
+            throwInvalidRequest("Invalid authorization_details");
+        }
```
IngridPuppet: Same. add more context?

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+            List<String> credentialIdentifiers = new ArrayList<>();
+            synchronized (userSession) {
+                String existingIdentifier = userSession.getNote(noteKey);
+                if (existingIdentifier != null) {
+                    credentialIdentifiers.add(existingIdentifier);
+                } else {
+                    String newIdentifier = UUID.randomUUID().toString();
+                    credentialIdentifiers.add(newIdentifier);
+                    userSession.setNote(noteKey, newIdentifier);
+                }
```
IngridPuppet: As per my understanding, you are already storing the whole authorization details response in the session. I don't quite get the added value of storing these identifiers separately.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -242,6 +259,150 @@ protected void checkClient() {
...
+            }
+            // Prepare response details with validated format
+            AuthorizationDetailResponse responseDetail = new AuthorizationDetailResponse();
+            responseDetail.setType(OPENID_CREDENTIAL_TYPE);
+            responseDetail.setFormat(format);
+            responseDetail.setVct(vct);
```
IngridPuppet: I believe it is more spec-aligning to match the credential configuration ID and return it instead. WDYT?

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
IngridPuppet, line 1: Regarding the changes made to this file, is there a way you could create a separate class and only have calls to one or more of its methods here? I find it invasive to have that much OID4VCI-specific code here.

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -91,18 +96,55 @@ public Response process(Context context) {
...
+            // Process authorization_details
+            List<AuthorizationDetailResponse> authorizationDetailsResponse = processAuthorizationDetails(clientSession.getUserSession());
+
+            AccessTokenResponse tokenResponse;
+            try {
+                tokenResponse = responseBuilder.build();
+            } catch (RuntimeException re) {
+                if ("can not get encryption KEK".equals(re.getMessage())) {
+                    throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_REQUEST,
+                            "can not get encryption KEK", Response.Status.BAD_REQUEST);
```
IngridPuppet: Maybe...
```suggestion
                if ("cannot get encryption KEK".equals(re.getMessage())) {
                    throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_REQUEST,
                            "cannot get encryption KEK", Response.Status.BAD_REQUEST);
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJWTIssuerEndpointTest.java
IngridPuppet, line 1: Did you forget to push some code in this file? Most of the introduced imports aren't used and it does not look like other edits are that necessary.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.keycloak.util.JsonSerialization;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.UUID;
+
+import static org.junit.Assert.*;
+import static org.keycloak.protocol.oid4vc.model.Format.SD_JWT_VC;
```
IngridPuppet: Star imports are so far avoided.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsTest.java
```diff
@@ -0,0 +1,180 @@
...
+        HttpPost postPreAuthorizedCode = new HttpPost(openidConfig.getTokenEndpoint());
+        List<NameValuePair> parameters = new LinkedList<>();
+        parameters.add(new BasicNameValuePair(OAuth2Constants.GRANT_TYPE, PreAuthorizedCodeGrantTypeFactory.GRANT_TYPE));
+        parameters.add(new BasicNameValuePair(PreAuthorizedCodeGrantTypeFactory.CODE_REQUEST_PARAM, credentialsOffer.getGrants().getPreAuthorizedCode().getPreAuthorizedCode()));
+        parameters.add(new BasicNameValuePair("authorization_details", authDetailsJson));
+        UrlEncodedFormEntity formEntity = new UrlEncodedFormEntity(parameters, StandardCharsets.UTF_8);
+        postPreAuthorizedCode.setEntity(formEntity);
+        try (CloseableHttpResponse tokenResponse = httpClient.execute(postPreAuthorizedCode)) {
+            assertEquals(HttpStatus.SC_BAD_REQUEST, tokenResponse.getStatusLine().getStatusCode());
+        }
```
IngridPuppet: Assert the error message?

IngridPuppet: Could you please add blank lines between meaningful code blocks for improved readability?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsTest.java
```diff
@@ -0,0 +1,180 @@
...
+        List<AuthorizationDetail> authDetails = List.of(authDetail);
+        String authDetailsJson = JsonSerialization.writeValueAsString(authDetails);
+        HttpPost postPreAuthorizedCode = new HttpPost(openidConfig.getTokenEndpoint());
+        List<NameValuePair> parameters = new LinkedList<>();
+        parameters.add(new BasicNameValuePair(OAuth2Constants.GRANT_TYPE, PreAuthorizedCodeGrantTypeFactory.GRANT_TYPE));
+        parameters.add(new BasicNameValuePair(PreAuthorizedCodeGrantTypeFactory.CODE_REQUEST_PARAM, credentialsOffer.getGrants().getPreAuthorizedCode().getPreAuthorizedCode()));
+        parameters.add(new BasicNameValuePair("authorization_details", authDetailsJson));
+        UrlEncodedFormEntity formEntity = new UrlEncodedFormEntity(parameters, StandardCharsets.UTF_8);
+        postPreAuthorizedCode.setEntity(formEntity);
```
IngridPuppet: Is there a way to reuse code? Most of the steps are similar with the other method.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsTest.java
```diff
@@ -0,0 +1,180 @@
...
+        HttpGet getOpenidConfiguration = new HttpGet(credentialIssuer.getAuthorizationServers().get(0) + "/.well-known/openid-configuration");
+        OIDCConfigurationRepresentation openidConfig;
+        try (CloseableHttpResponse response = httpClient.execute(getOpenidConfiguration)) {
+            assertEquals(HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+            String s = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
+            openidConfig = JsonSerialization.readValue(s, OIDCConfigurationRepresentation.class);
+        }
+        AuthorizationDetail authDetail = new AuthorizationDetail();
+        authDetail.setType(OPENID_CREDENTIAL_TYPE);
+        authDetail.setFormat(SD_JWT_VC);
```
IngridPuppet: It appears the pr̂esent request goes through without a `vct`.... but should it?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtIssuingEndpointTest.java
IngridPuppet, line 1: Same. Are the changes here that necessary? Introduced imports are left unused.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsAndCredentialIdPersistenceTest.java
```diff
@@ -0,0 +1,796 @@
...
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.UUID;
+
+import static org.junit.Assert.*;
+import static org.keycloak.protocol.oid4vc.model.Format.JWT_VC;
+import static org.keycloak.protocol.oid4vc.model.Format.SD_JWT_VC;
+import static org.keycloak.protocol.oidc.grants.OAuth2GrantTypeBase.OPENID_CREDENTIAL_TYPE;
+
```
IngridPuppet: Star imports are so far avoided, AFAIK. There is also an unused import here.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsAndCredentialIdPersistenceTest.java
```diff
@@ -0,0 +1,796 @@
...
+            openidConfig = JsonSerialization.readValue(s, OIDCConfigurationRepresentation.class);
+        }
+
+        // 5. Prepare authorization_details with credential_configuration_id
+        AuthorizationDetail authDetail = new AuthorizationDetail();
+        authDetail.setType(OPENID_CREDENTIAL_TYPE);
+        authDetail.setCredentialConfigurationId("test-credential");
+        if (credentialIssuer.getAuthorizationServers() != null && !credentialIssuer.getAuthorizationServers().isEmpty()) {
+            authDetail.setLocations(Collections.singletonList(credentialIssuer.getCredentialIssuer()));
+        }
```
IngridPuppet: The setup is known. Why use conditional branching rather than simply setting the locations if that's needed?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsAndCredentialIdPersistenceTest.java
```diff
@@ -0,0 +1,796 @@
...
+
+        // 4. Get the openid-configuration
+        HttpGet getOpenidConfiguration = new HttpGet(credentialIssuer.getAuthorizationServers().get(0) + "/.well-known/openid-configuration");
+        OIDCConfigurationRepresentation openidConfig;
+        try (CloseableHttpResponse response = httpClient.execute(getOpenidConfiguration)) {
+            assertEquals(HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+            String s = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
+            openidConfig = JsonSerialization.readValue(s, OIDCConfigurationRepresentation.class);
+        }
```
IngridPuppet: These steps look quite identical. Could you reuse code for easier consumption?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsAndCredentialIdPersistenceTest.java
```diff
@@ -0,0 +1,796 @@
...
+        List<NameValuePair> parameters = new LinkedList<>();
+        parameters.add(new BasicNameValuePair(OAuth2Constants.GRANT_TYPE, PreAuthorizedCodeGrantTypeFactory.GRANT_TYPE));
+        parameters.add(new BasicNameValuePair(PreAuthorizedCodeGrantTypeFactory.CODE_REQUEST_PARAM, credentialsOffer.getGrants().getPreAuthorizedCode().getPreAuthorizedCode()));
+        parameters.add(new BasicNameValuePair("authorization_details", authDetailsJson));
+        UrlEncodedFormEntity formEntity = new UrlEncodedFormEntity(parameters, StandardCharsets.UTF_8);
+        postPreAuthorizedCode.setEntity(formEntity);
+        try (CloseableHttpResponse tokenResponse = httpClient.execute(postPreAuthorizedCode)) {
+            assertEquals(HttpStatus.SC_BAD_REQUEST, tokenResponse.getStatusLine().getStatusCode());
+        }
+    }
```
IngridPuppet: Assert the error message?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsAndCredentialIdPersistenceTest.java
```diff
@@ -0,0 +1,796 @@
...
+            assertFalse("Different sessions should generate different identifiers for the same credential_configuration_id", firstIdentifier.equals(thirdIdentifier));
+        }
+    }
+
+    @Test
+    public void testCredentialIdentifierDifferentSessions() throws Exception {
+        // Test that different sessions generate different identifiers for the same credential_configuration_id
+
+        // 1. First session - get a token and make a request
+        String token1 = getBearerToken(oauth);
```
IngridPuppet: I wonder how necessary is this test. Even though the implementation works that way, are we violating anything if it didn't?

IngridPuppet (changes requested):
This PR updates the logic at the token endpoint to process OID4VCI authorization details. I checked the implementation and have concerns related to its "invasiveness" and others, some overlapping with comments raised by @wistefan. Hello @forkimenjeckayang - Please could you check?

services/src/main/java/org/keycloak/protocol/oid4vc/model/AuthorizationDetail.java
thomasdarimont: Perhaps move those into a generic map?
```java
@JsonIgnore
  protected Map<String, Object> additionalFields = new HashMap<String, Object>();

  @JsonAnyGetter
  public Map<String, Object> getAdditionalFields() {
      return additionalFields;
  }

  @JsonAnySetter
  public void setAdditionalField(String name, Object value) {
      this.additionalFields.put(name, value);
  }

```

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
forkimenjeckayang: You're right. We're stroing the whole authorization details response in the client session context. Maybe going further to store the identifiers seperately doesn't really add value.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
forkimenjeckayang: right

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsTest.java
forkimenjeckayang: i think the vct field is optional ?
Looking at this: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-using-authorization-details
or i'm mistaken

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsAndCredentialIdPersistenceTest.java
forkimenjeckayang: Not really , we're not voilating anything. Just wanted to make sure of identifier persistence in session.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsTest.java
IngridPuppet: It is the combination of the SD-JWT VC format and a vct that enables recovering a credential configuration ID, given that there might be multiple credentials of type SD-JWT VC offered for issuance. Notice that format-specific attributes are defined in Appendix A. You should refer to this instead: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-authorization-details-5

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.UUID;
+
+public class OID4VCAuthorizationDetailsProcessor {
+    private static final Logger logger = Logger.getLogger(OID4VCAuthorizationDetailsProcessor.class);
+    private final KeycloakSession session;
+    private final EventBuilder event;
+    private final MultivaluedMap<String, String> formParams;
+    private final Cors cors;
+
+    public static final String AUTHORIZATION_DETAILS_PARAM = "authorization_details";
```
thomasdarimont: Note that this constant is also already defined here:
org.keycloak.protocol.oidc.OIDCLoginProtocol#AUTHORIZATION_DETAILS_PARAM

Perhaps the official parameters and response attributes should be added to the org.keycloak.OAuth2Constants.

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -188,12 +192,18 @@ public Response process(Context context) {
             throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_GRANT, cpe.getErrorDetail(), Response.Status.BAD_REQUEST);
         }
 
+        String scopeParam = codeData.getScope();
+        ClientSessionContext clientSessionCtx = DefaultClientSessionContext.fromClientSessionAndScopeParameter(clientSession, scopeParam, session);
+
+        // OID4VCI: Process authorization_details using the processor
+        OID4VCAuthorizationDetailsProcessor oid4vciProcessor = new OID4VCAuthorizationDetailsProcessor(session, event, formParams, cors);
```
thomasdarimont: Since this effectively introduces support to process RAR `authorization_details` I wonder if it makes sense to have a more generic implementation here. 

Perhaps we can have an interface like AuthorizationDetailsProcessor, with a DefaultAuthorizationDetailsProcessor implementation and a special OID4VCIAuthorizationDetailsProcessor implementation. 

Then we could then associate the AuthorizationDetailsProcessor implementation to the OIDCLoginProtocol and query it from there - if we detect that OpenID4VCI is used, we can re turn the OID4VCIAuthorizationDetailsProcessor.

thomasdarimont (commented):
Implementation looks good, but I have a few remarks.

forkimenjeckayang:
> Implementation looks good, but I have a few remarks.

Hello @thomasdarimont , I've addressed the changes you suggested.
Could you please check again if everyhting is okay..

core/src/main/java/org/keycloak/OAuth2Constants.java
```diff
@@ -155,6 +155,9 @@ public interface OAuth2Constants {
     String AUTHENTICATOR_METHOD_REFERENCE = "amr";
 
     String CNF = "cnf";
+
+    String AUTHORIZATION_DETAILS_PARAM = "authorization_details";
+    String AUTHORIZATION_DETAILS_RESPONSE_KEY = "authorization_details_response";
 }
```
IngridPuppet: Is this a normative OAuth2 constant? Thought that was a custom attribute for keeping things in the session. If I'm right, then I don't think it should be added here.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/AuthorizationDetailsProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.keycloak.models.UserSessionModel;
+import org.keycloak.models.ClientSessionContext;
+import org.keycloak.protocol.oid4vc.model.AuthorizationDetailResponse;
+
+import java.util.List;
+
+public interface AuthorizationDetailsProcessor {
+    List<AuthorizationDetailResponse> process(UserSessionModel userSession, ClientSessionContext clientSessionCtx);
+}
```
IngridPuppet: Missing authorship and comments?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/DefaultAuthorizationDetailsProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+package org.keycloak.protocol.oid4vc.issuance;
+
+import org.keycloak.models.UserSessionModel;
+import org.keycloak.models.ClientSessionContext;
+import org.keycloak.protocol.oid4vc.model.AuthorizationDetailResponse;
+
+import java.util.List;
+import java.util.Collections;
+
+public class DefaultAuthorizationDetailsProcessor implements AuthorizationDetailsProcessor {
```
IngridPuppet: Missing authorship?

server-spi-private/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantType.java
```diff
@@ -83,6 +84,7 @@ public static class Context {
         protected Cors cors;
         protected Object tokenManager;
         protected String grantType;
+        public LoginProtocol protocol;
```
IngridPuppet: Follow the trend and make it protected, adding a setter?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        List<AuthorizationDetail> authDetails = parseAuthorizationDetails(authorizationDetailsParam);
+        List<String> supportedFormats = new ArrayList<>(SUPPORTED_FORMATS);
+        Map<String, SupportedCredentialConfiguration> supportedCredentials = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session);
+        List<AuthorizationDetailResponse> authDetailsResponse = new ArrayList<>();
+
+        // Retrieve issuer metadata and identifier for locations check
+        CredentialIssuer issuerMetadata = (CredentialIssuer) new OID4VCIssuerWellKnownProvider(session).getConfig();
+        List<String> authorizationServers = issuerMetadata.getAuthorizationServers();
```
IngridPuppet: I suggest you refactor the `OID4VCIssuerWellKnownProvider` class to call `OID4VCIssuerWellKnownProvider.getAuthorizationServers()` and avoid loading the whole config tree. WDYT?

services/src/main/java/org/keycloak/protocol/oidc/OIDCLoginProtocol.java
```diff
@@ -80,7 +89,7 @@ public class OIDCLoginProtocol implements LoginProtocol {
     public static final String LOGIN_PROTOCOL = "openid-connect";
     public static final String STATE_PARAM = "state";
     public static final String SCOPE_PARAM = "scope";
-    public static final String AUTHORIZATION_DETAILS_PARAM = "authorization_details";
+    public static final String AUTHORIZATION_DETAILS_PARAM = OAuth2Constants.AUTHORIZATION_DETAILS_PARAM ;
```
IngridPuppet: You could just "import static" it...

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -200,14 +194,21 @@ public Response process(Context context) {
...
-        List<AuthorizationDetailResponse> authorizationDetailsResponse = processAuthorizationDetails(userSession);
+        String scopeParam = codeData.getScope();
+        ClientSessionContext clientSessionCtx = DefaultClientSessionContext.fromClientSessionAndScopeParameter(clientSession, scopeParam, session);
+
+        // OID4VCI: Process authorization_details using the processor
+        AuthorizationDetailsProcessor processor = ((OIDCLoginProtocol) context.protocol).getAuthorizationDetailsProcessor(session, event, formParams, cors);
+        List<AuthorizationDetailResponse> authorizationDetailsResponse = processor.process(userSession, clientSessionCtx);
+        if (authorizationDetailsResponse != null && !authorizationDetailsResponse.isEmpty()) {
+            clientSessionCtx.setAttribute(AUTHORIZATION_DETAILS_RESPONSE_KEY, authorizationDetailsResponse);
+        }
```
IngridPuppet: Shouldn't this be conditional to the type of the protocol instance in the context? I wonder how safe is the cast to OIDCLoginProtocol.

IngridPuppet (commented):
Hello @forkimenjeckayang - Thank you for addressing my concerns and considering my suggestions. Given this and other recent changes, I have a few additional minor comments. Please could you check?

IngridPuppet (approved):
LGTM

server-spi-private/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantType.java
```diff
@@ -118,6 +120,10 @@ public void setClientAuthAttributes(Map<String, String> clientAuthAttributes) {
             this.clientAuthAttributes = clientAuthAttributes;
         }
 
+        public void setProtocol(LoginProtocol protocol) {
```
mposolda: I suppose it might be possible to do something like this in the constructor of the `OAuth2GrantType.Context` : 
```
this.protocol = session.getProvider(LoginProtocol.class, client.getProtocol());
```
Then we don't need method `setProtocol` and also we don't need any changes in the `TokenEndpoint` class in this PR. Is it ok to do that?

mposolda (commented):
Thanks for the PR and sorry for the late response.

It can be great if we can avoid references to `oid4vci` from the internal classes like `AuthorizationCodeGrantType` and `OIDCLoginProtocol` to avoid inter-dependencies between protocols .

Also it seems to me that `authorization_details` is not OID4VCI specific, but it is RAR construct and the format of this document is not oid4vci specific. For now, I have this suggestion:

- The `AuthorizationDetailsProcessor` would be dedicated provider interface and will have it's own `ProviderFactory` and `Spi` class. I can imagine something like this:
```
public interface AuthorizationDetailsProcessor extends Provider {
    /**
     * @return authorizationDetailResponse if this processor is able to process the given "authorization_details" parameter. Returns null if authorization_details parameter is incompatible with this processor. In that case, the other AuthorizationDetailsProcessor may follow and try to parse the response
     */
    List<AuthorizationDetailsResponse> process(UserSessionModel userSession, ClientSessionContext clientSessionCtx, String authorizationDetailsParameter);
} 
```
This interface might be in the dedicated package as it is not `oid4vci` specific. Maybe `org.keycloak.protocol.oidc.rar` inside the `server-spi-private` module.

- The class `AuthorizationDetailsResponse` might be quite generic and also not specific to `oid4vci` as it may serve the different RAR objects from different areas. Maybe just something like:
```
public class AuthorizationDetailsResponse {
    
    @JsonIgnore
    private Map<String, Object> otherClaims = new HashMap<>();

    @JsonAnyGetter
    public Map<String, Object> getOtherClaims() {
        return otherClaims;
    }

    @JsonAnySetter
    public void setOtherClaims(String name, Object value) {
        this.otherClaims.put(name, value);
    }
}
```

- The concrete implementation of the `AuthorizationDetailsProcessor` might be in the `oid4vci` codebase and can be enabled just if the `oid4vci` feature is enabled (like many other providers related to `oid4vci`). And it can possibly return subclass of `AuthorizationDetailsResponse` with the claims specific to `oid4vci` .

- The code in the `AuthorizationCodeGrantType` can be possibly updated to something like this:

```
        String authorizationDetailsParam = formParams.getFirst(AUTHORIZATION_DETAILS_PARAM);
        if (authorizationDetailsParam != null) {
 	       AuthorizationDetailsResponse authzDetailsResponse = session.getKeycloakSessionFactory()
        	        .getProviderFactoriesStream(AuthorizationDetailsProcessor.class)
        	        .sorted((f1, f2) -> f2.order() - f1.order())
        	        .map(f -> session.getProvider(AuthorizationDetailsProcessor.class, f.getId()))
        	        .map(authzDetailsProcessor -> authzDetailsProcessor.process(userSession, clientSessionCtx, authorizationDetailsParam))
        	        .filter(authzDetailsResponse -> authzDetailsResponse != null)
                	.findFirst()
                	.ifPresent((authzDetailsResponse) -> clientSessionCtx.setAttribute(AUTHORIZATION_DETAILS_RESPONSE_KEY, authorizationDetailsResponse);                	
       }
```

This also means no need for `DefaultAuthorizationDetailsProcessor` and constant `AUTHORIZATION_DETAILS_RESPONSE_KEY` can be likely moved to some place outside of `oid4vci` . Maybe to the `Constants` class? And there will not be a need for the `OIDCLoginPRotocol.getAuthorizationDetailsProcessor` .

Could this work? Also please see my inline comment.

forkimenjeckayang:
> Thanks for the PR and sorry for the late response.
> 
> It can be great if we can avoid references to `oid4vci` from the internal classes like `AuthorizationCodeGrantType` and `OIDCLoginProtocol` to avoid inter-dependencies between protocols .
> 
> Also it seems to me that `authorization_details` is not OID4VCI specific, but it is RAR construct and the format of this document is not oid4vci specific. For now, I have this suggestion:
> 
> * The `AuthorizationDetailsProcessor` would be dedicated provider interface and will have it's own `ProviderFactory` and `Spi` class. I can imagine something like this:
> 
> ```
> public interface AuthorizationDetailsProcessor extends Provider {
>     /**
>      * @return authorizationDetailResponse if this processor is able to process the given "authorization_details" parameter. Returns null if authorization_details parameter is incompatible with this processor. In that case, the other AuthorizationDetailsProcessor may follow and try to parse the response
>      */
>     List<AuthorizationDetailsResponse> process(UserSessionModel userSession, ClientSessionContext clientSessionCtx, String authorizationDetailsParameter);
> } 
> ```
> 
> This interface might be in the dedicated package as it is not `oid4vci` specific. Maybe `org.keycloak.protocol.oidc.rar` inside the `server-spi-private` module.
> 
> * The class `AuthorizationDetailsResponse` might be quite generic and also not specific to `oid4vci` as it may serve the different RAR objects from different areas. Maybe just something like:
> 
> ```
> public class AuthorizationDetailsResponse {
>     
>     @JsonIgnore
>     private Map<String, Object> otherClaims = new HashMap<>();
> 
>     @JsonAnyGetter
>     public Map<String, Object> getOtherClaims() {
>         return otherClaims;
>     }
> 
>     @JsonAnySetter
>     public void setOtherClaims(String name, Object value) {
>         this.otherClaims.put(name, value);
>     }
> }
> ```
> 
> * The concrete implementation of the `AuthorizationDetailsProcessor` might be in the `oid4vci` codebase and can be enabled just if the `oid4vci` feature is enabled (like many other providers related to `oid4vci`). And it can possibly return subclass of `AuthorizationDetailsResponse` with the claims specific to `oid4vci` .
> * The code in the `AuthorizationCodeGrantType` can be possibly updated to something like this:
> 
> ```
>         String authorizationDetailsParam = formParams.getFirst(AUTHORIZATION_DETAILS_PARAM);
>         if (authorizationDetailsParam != null) {
>  	       AuthorizationDetailsResponse authzDetailsResponse = session.getKeycloakSessionFactory()
>         	        .getProviderFactoriesStream(AuthorizationDetailsProcessor.class)
>         	        .sorted((f1, f2) -> f2.order() - f1.order())
>         	        .map(f -> session.getProvider(AuthorizationDetailsProcessor.class, f.getId()))
>         	        .map(authzDetailsProcessor -> authzDetailsProcessor.process(userSession, clientSessionCtx, authorizationDetailsParam))
>         	        .filter(authzDetailsResponse -> authzDetailsResponse != null)
>                 	.findFirst()
>                 	.ifPresent((authzDetailsResponse) -> clientSessionCtx.setAttribute(AUTHORIZATION_DETAILS_RESPONSE_KEY, authorizationDetailsResponse);                	
>        }
> ```
> 
> This also means no need for `DefaultAuthorizationDetailsProcessor` and constant `AUTHORIZATION_DETAILS_RESPONSE_KEY` can be likely moved to some place outside of `oid4vci` . Maybe to the `Constants` class? And there will not be a need for the `OIDCLoginPRotocol.getAuthorizationDetailsProcessor` .
> 
> Could this work? Also please see my inline comment.

Thank you @mposolda  for this review.
I applied suggested changes where necessary.
Could you please look into this again ?

server-spi-private/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantType.java
```diff
@@ -108,6 +113,9 @@ public void setFormParams(MultivaluedHashMap<String, String> formParams) {
 
         public void setClient(ClientModel client) {
             this.client = client;
+            if (this.protocol == null && client != null) {
```
mposolda:
```suggestion
            if (client != null) {
```
Maybe a nitpick, but I can suggest to update to something like this? In case that new client has different protocol, then it can be good to update the `this.protocol` as well to the new protocol though?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import org.keycloak.protocol.oidc.rar.AuthorizationDetailsProcessor;
+import org.keycloak.protocol.oidc.rar.AuthorizationDetailsProcessorFactory;
+
+/**
+ * Factory for creating OID4VCI-specific authorization details processors.
+ * This factory is only enabled when the OID4VCI feature is available.
+ *
+ * @author <a href="mailto:Forkim.Akwichek@adorsys.com">Forkim Akwichek</a>
+ */
+public class OID4VCAuthorizationDetailsProcessorFactory implements AuthorizationDetailsProcessorFactory {
```
mposolda: I think this should implement also `OID4VCEnvironmentProviderFactory` to make sure that the provider is available just when `oid4vci` feature is available? See other provider factories implementing this interface for the inspiration.

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -188,12 +196,36 @@ public Response process(Context context) {
...
+        if (authorizationDetailsParam != null) {
+            try {
+                session.getKeycloakSessionFactory()
+                        .getProviderFactoriesStream(AuthorizationDetailsProcessor.class)
+                        .sorted((f1, f2) -> f2.order() - f1.order())
+                        .map(f -> session.getProvider(AuthorizationDetailsProcessor.class, f.getId()))
+                        .map(authzDetailsProcessor -> authzDetailsProcessor.process(userSession, clientSessionCtx, authorizationDetailsParam))
+                        .filter(authzDetailsResponse -> authzDetailsResponse != null)
+                        .findFirst()
+                        .ifPresent(authzDetailsResponse -> clientSessionCtx.setAttribute("authorization_details_response", authzDetailsResponse));
```
mposolda:
```suggestion
                        .ifPresentOrElse(authzDetailsResponse -> clientSessionCtx.setAttribute(AUTHORIZATION_DETAILS_RESPONSE, authzDetailsResponse),
                        () -> logger.debugf("No available AuthorizationDetailsProcessor being able to process authorization_details '%s', authorizationDetailsParam));                        
```
I suppose it can be useful to know also when there is not available AuthorizationDetailsProcessor, which is being able to process the request. Sorry for not mention in the original PR.

Also will be good to have the constant for `authorization_details_response` and add it somewhere (maybe to class `Constants`) to be able to easily search for the usages. This `Constant` might be used in various other places (EG. in method `addCustomTokenResponseClaims` as well).

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -188,12 +196,36 @@ public Response process(Context context) {
...
+            } catch (RuntimeException e) {
+                if (e.getMessage() != null && e.getMessage().contains("Invalid authorization_details")) {
+                    event.error(Errors.INVALID_REQUEST);
+                    throw new CorsErrorResponseException(cors, "invalid_request", e.getMessage(), Response.Status.BAD_REQUEST);
```
mposolda: Unfortunately, we are doing it already, but for the future, we are trying to avoid adding "random" exception details into the error responses to make sure to avoid some issues like XSS for example (in case that error-response is being shown in the UI of the client application and client-application does not properly handle it). Is it maybe possible to just add something to error details like "Error when processing authorization_details" and instead log the exception? For example with something like `logger.warnf("Error when processing authorization_details", e);`

mposolda (commented):
@forkimenjeckayang Thanks! PR looks much better now. I've added few more minor comments.

Also there are test failures, which might not be directly related to your PR, but maybe it helps to rebase on top of latest main. Unfortunately, there is right now the opened issue which the keycloak build is broken https://github.com/keycloak/keycloak/issues/42496 , so you may need to wait a bit before rebase until that is fixed in main (which should be hopefully later today).

forkimenjeckayang:
> @forkimenjeckayang Thanks! PR looks much better now. I've added few more minor comments.
> 
> Also there are test failures, which might not be directly related to your PR, but maybe it helps to rebase on top of latest main. Unfortunately, there is right now the opened issue which the keycloak build is broken #42496 , so you may need to wait a bit before rebase until that is fixed in main (which should be hopefully later today).

Thanks again for the update @mposolda 
I have applied the remaining suggestion. Hope the recent changes address all your concerns.
i just realized this https://github.com/keycloak/keycloak/issues/42496 has been closed. Will make necessary changes then to be up to date with main.

forkimenjeckayang:
@mposolda , some CIs still fail and I do not beleive they are related to the changes I made.

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
mposolda: Thanks for the updates. The class `OAuth2Constants` from `core` module is dedicated to the standard parameters mentioned in the specification. But per my understanding, this `AUTHORIZATION_DETAILS_RESPONSE` is not a standard parameter, but just an internal note added into client session. That is why I've rather suggested to use the class `Constants` from `server-spi-private` module (sorry that I made the typo in the name of the class in my comment above).

mposolda:
@forkimenjeckayang Thanks for the latest update!

I've added one more nitpick comment regarding the constant `AUTHORIZATION_DETAILS_RESPONSE` .

However the bigger issues are failing tests. It also seems to me that those are not related to your changes, but at the same time, they are not failing in any other Keycloak PRs. Also it is strange that the same test for fine-grained admin permissions failed in all the DB builds in your PR and not in the single build. Which may indicate that this might have some relation to your changes.

Could you please try to rebase again? Eventually it may be good if you have opportunity to try to run the tests locally on your laptop (the `UserSessionLimitsTest` and the FineGrained admin permissions test) and doublecheck if they pass for you locally.

Please let me know if tests pass locally for you after doing this. We may figure next steps based on that...

forkimenjeckayang:
@m

> @forkimenjeckayang Thanks for the latest update!
> 
> I've added one more nitpick comment regarding the constant `AUTHORIZATION_DETAILS_RESPONSE` .
> 
> However the bigger issues are failing tests. It also seems to me that those are not related to your changes, but at the same time, they are not failing in any other Keycloak PRs. Also it is strange that the same test for fine-grained admin permissions failed in all the DB builds in your PR and not in the single build. Which may indicate that this might have some relation to your changes.
> 
> Could you please try to rebase again? Eventually it may be good if you have opportunity to try to run the tests locally on your laptop (the `UserSessionLimitsTest` and the FineGrained admin permissions test) and doublecheck if they pass for you locally.
> 
> Please let me know if tests pass locally for you after doing this. We may figure next steps based on that...

Hello @mposolda 
I've adjusted the constant placement.
Regarding the tests, these pass locally:
- FineGrainedAdminWithTokenExchangeTest
- FineGrainedAdminWithTokenExchangeDisabledTest
- AbstractFineGrainedAdminTest 
But this stil fails
- UserSessionLimitsTest.testRealmSessionCountAndClientSessionCountExceededAndOldestClientSessionShouldBePrioritized

core/src/main/java/org/keycloak/OAuth2Constants.java
```diff
@@ -160,6 +160,8 @@ public interface OAuth2Constants {
 
     String CNF = "cnf";
 
+    String AUTHORIZATION_DETAILS_PARAM = "authorization_details";
```
tnorimat, line 164: How about adding the following comment?
```
// RAR - https://datatracker.ietf.org/doc/html/rfc9396
```

server-spi-private/src/main/java/org/keycloak/protocol/oidc/rar/AuthorizationDetailsProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+package org.keycloak.protocol.oidc.rar;
+
+import org.keycloak.models.UserSessionModel;
+import org.keycloak.models.ClientSessionContext;
+import org.keycloak.provider.Provider;
+
+import java.util.List;
+
+/**
+ * Provider interface for processing authorization_details parameter in OAuth2/OIDC token requests.
```
Captain-P-Goldfish: documentation: The `authorization_details` parameter can also be used in AuthorizationRequests: openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-5.1.1

server-spi-private/src/main/java/org/keycloak/protocol/oidc/rar/AuthorizationDetailsProcessorSpi.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.keycloak.protocol.oidc.rar;
+
+import org.keycloak.provider.Provider;
+import org.keycloak.provider.ProviderFactory;
+import org.keycloak.provider.Spi;
+
+/**
+ * SPI for authorization details processors that handle the authorization_details parameter
+ * in OAuth2/OIDC token requests as per RAR (Rich Authorization Requests) specification.
```
Captain-P-Goldfish: documentation again: should also be usable for AuthorizationRequests: openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-5.1.1

server-spi-private/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantType.java
```diff
@@ -173,6 +181,10 @@ public Object getTokenManager() {
         public String getGrantType() {
             return grantType;
         }
+
+        public LoginProtocol getProtocol() {
```
tnorimat: It seems that this method is not used. Why would you add `protected LoginProtocol protocol;` ?

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -188,12 +196,38 @@ public Response process(Context context) {
...
+                if (e.getMessage() != null && e.getMessage().contains("Invalid authorization_details")) {
+                    logger.warnf(e, "Error when processing authorization_details");
+                    event.error(Errors.INVALID_REQUEST);
+                    throw new CorsErrorResponseException(cors, "invalid_request", "Error when processing authorization_details", Response.Status.BAD_REQUEST);
+                } else {
+                    throw e;
+                }
+            }
+        }
+
```
Captain-P-Goldfish: I think it would be better to extract that part into a protected method. This would also allow to override this behaviour if necessary

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsFlowTest.java
```diff
@@ -0,0 +1,313 @@
...
+
+    @Test
+    public void testPreAuthorizedCodeWithInvalidAuthorizationDetails() throws Exception {
+        String token = getBearerToken(oauth, client, sdJwtTypeCredentialClientScope.getName());
+        Oid4vcTestContext ctx = prepareOid4vcTestContext(token);
+
+        AuthorizationDetail authDetail = new AuthorizationDetail();
+        authDetail.setType(OPENID_CREDENTIAL_TYPE);
+        authDetail.setCredentialConfigurationId(sdJwtTypeCredentialClientScope.getAttributes().get(CredentialScopeModel.CONFIGURATION_ID));
+        authDetail.setFormat(SD_JWT_VC); // Invalid: format should not be combined with credential_configuration_id
```
Captain-P-Goldfish, line 154: Why is SD-JWT an invalid format? 
Draft 17: Appendix A.3.3 claims different:
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html?utm_source=chatgpt.com#name-authorization-details-5
```json
[
  {
    "type": "openid_credential",
    "credential_configuration_id": "PID_SD_JWT_VC",
    "claims": [
      {"path": ["given_name"]},
      {"path": ["family_name"]}
    ]
  }
]
```

Captain-P-Goldfish (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -0,0 +1,211 @@
...
+        for (AuthorizationDetail detail : authDetails) {
+            validateAuthorizationDetail(detail, supportedFormats, supportedCredentials, authorizationServers, issuerIdentifier);
+            AuthorizationDetailsResponse responseDetail = buildAuthorizationDetailResponse(detail, userSession, supportedCredentials, supportedFormats, clientSessionCtx);
+            authDetailsResponse.add(responseDetail);
+        }
+
+        return authDetailsResponse;
+    }
+
+    private boolean canProcess(String authorizationDetailsParameter) {
```
tnorimat: Line 123 also checks `OPENID_CREDENTIAL_TYPE.equals(detail.getType()` and throws an exception, which means that the method `validateAuthorizationDetail` successfully pass if all of `AuthorizationDetails` have a type `openid_credential`.

Considering this point, the method `canProcess` seems not to be needed. What do you think?

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
```diff
@@ -188,12 +196,38 @@ public Response process(Context context) {
             throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_GRANT, cpe.getErrorDetail(), Response.Status.BAD_REQUEST);
         }
 
+        String scopeParam = codeData.getScope();
+        ClientSessionContext clientSessionCtx = DefaultClientSessionContext.fromClientSessionAndScopeParameter(clientSession, scopeParam, session);
+
+        // Process authorization_details using provider discovery
```
tnorimat: How about moving the codes of Line199-20 after 205?
By doing so, keycloak does not run these codes if keycloak receives an autorization request without `authorization_details` parameter, which is a dominant case so far.

tnorimat (commented):
@forkimenjeckayang Hello, I added some review comments. I appreciate it if you would check them.

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
mposolda: +1 to this suggestion. As it seems that same code is needed by `AuthorizationCodeGrantType` and also `PreAuthorizedGrantType` and in the future, maybe there are more grants wanting to use RAR. Maybe makes sense to add to `OAuth2GrantTypeBase` ? Maybe protected method will need to take some parameters, which are available just in the parent classes .

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsFlowTest.java
forkimenjeckayang: Concerning this @Captain-P-Goldfish , the PR was created some time ago and it's draft-15 compliant. According to the draft-15 spec ,we cannot have both `format` and `credential_configuration_id` in the `authorization_details` parameter. There is follow up PR which is draft-16 compliant and addresses the new structure of authorization details . Once we have this merged , i'll send forth the PR with the authorization details update.
For now , we are pushing foward with draft-16. Draft 17 will come in later.

services/src/main/java/org/keycloak/protocol/oidc/grants/AuthorizationCodeGrantType.java
forkimenjeckayang: Yes, it makes sence to add the method to `OAuth2GrantTypeBase`

server-spi-private/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantType.java
forkimenjeckayang: @tnorimat I added the `protocol` field and `getProtocol()` method to address previous review comments from @mposolda. However, you're right that the `getProtocol()` method is currently unused. I've removed the unused `getProtocol()` method while keeping the `protocol` field and its initialization logic, as the field may be needed for future functionality or other grant types.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
forkimenjeckayang: You're right @tnorimat

forkimenjeckayang:
@mposolda , @tnorimat  and @Captain-P-Goldfish 
Please could you have another look?

forkimenjeckayang:
Also @mposolda , i still cannot figgure out why the CIs fail. With the recent changes on main included here, i tried to run locally  those tests that fail in the CI and they fail still. OIDVCI related tests pass. So i really do not know..

tnorimat (approved):
LGTM

mposolda:
@forkimenjeckayang I've sent you the PR to your branch to fix the tests https://github.com/adorsys/keycloak-oid4vc/pull/144 . If you merge that to your branch, it should automatically update this PR and it should fix the tests inside this PR, so we would be able to finally merge it to Keycloak main.

forkimenjeckayang:
> @forkimenjeckayang I've sent you the PR to your branch to fix the tests [adorsys#144](https://github.com/adorsys/keycloak-oid4vc/pull/144) . If you merge that to your branch, it should automatically update this PR and it should fix the tests inside this PR, so we would be able to finally merge it to Keycloak main.

Thank you @mposolda . I just included the changes.

mposolda (approved):
@forkimenjeckayang @IngridPuppet @thomasdarimont @wistefan @Captain-P-Goldfish @tnorimat @VinodAnandan Thanks everyone for the collaboration on this PR!

Issue #32262, opened 2024-08-20
SAML Frontchannel Logout missing via Redirect or Post Binding is missing signature if login happened via artifact binding


nxpthx:
### Before reporting an issue

- [X] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

saml

### Describe the bug

Given you are using a SAML Client using the artifact binding for login and a configured frontchannel-logout.
The front-channel logout URLS of the SAML client have a configuration value in either POST or Redirect logout URL.

The request generation in the SAML-Protocol, takes the Session Note of the login Binding into account to consider, whether the logout request needs to be signed, even though the binding of the logout is configured differently.

In that case the logout request fails due to a missing signature.

### Version

25.0.4

### Regression

- [ ] The issue is a regression

### Expected behavior

Once the logout process determined which Binding to use for logout, the logout dedicated information should be used.
Therefore the logout process will be signed, in case it is not an artifact binding, and document signing is configured for the client.

### Actual behavior

The signing of the logout request is skipped, when the login request happened using the artifact binding.

### How to Reproduce?

You will need a SAML service provider capable of doing logins using the artifact binding and front-channel logouts using the POST or Redirect binding. The SAML- SP must require logout requests to be signed and the Login of the SP needs to use the artifact binding

Given that software you configure a client with the following configuration
- The client must not have POST or Artifact binding enforced.
- Sign documents needs to be activated 
- Front Channel Logout is activated 
- A "Logout Service Redirect Binding URL" or "Logout Service POST Binding URL" is configured
- Artifact Binding URL and Artifact Resolution Service URLs are configured.

Steps to produce the issue:

1. Login (successfully) to the SP via the IP, so there is a session not as Artifact Binding
2. Logout with that user, so a front channel logout is send to the SP.
3. The logout will fail, as the SP complains about a missing request signature.

### Anything else?

The line in question to be found here: 
https://github.com/keycloak/keycloak/blob/95e649f457a83c3ce61d8aad0626754d8027a8bd/services/src/main/java/org/keycloak/protocol/saml/SamlProtocol.java#L709

PR #40757, opened 2025-06-27
Use POST binding for logout when REDIRECT url is not set and forced POST

rmartinc:
Closes #40637

A little regression created in 26.2 for SAML by #32262. Previously if just the admin URL (master SAML processing URL) was set, POST was used for frontchannel logout (see [this line in 26.1.5](https://github.com/keycloak/keycloak/blob/26.1.5/services/src/main/java/org/keycloak/protocol/saml/SamlProtocol.java#L389)). After that issue the default is REDIRECT for that configuration, even if force POST is configured to ON in the client. I also changed that method to do a short-circuit evaluation (reversing the order and returning immediately). The meaningful change is `if (postUrlSet && !redirectUrlSet)` to `if (!redirectUrlSet && (postUrlSet || samlClient.forcePostBinding()))`. Created one integration test and modified the unitary test too.

Note that I didn't revert the functionality completely because IMHO it makes more sense like it is in the PR. The issue is only when there is only master saml processing URL, in that case previously it was always POST, currently is always REDIRECT, and with the PR its POST or REDIRECT depending the configuration of the `Force POST binding` switch. Normally the logout URLs are filled if you use a SP descriptor to create the client, and the `Force POST binding`  switch is by default set to true.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/saml/SamlProtocol.java
+++ b/services/src/main/java/org/keycloak/protocol/saml/SamlProtocol.java
@@ -375,33 +375,35 @@ public class SamlProtocol implements LoginProtocol {
         boolean redirectUrlSet = !(logoutRedirectUrl == null || logoutRedirectUrl.trim().isEmpty());
         boolean artifactUrlSet = !(logoutArtifactUrl == null || logoutArtifactUrl.trim().isEmpty());
 
-        // Default to Redirect
-        String bindingType = SAML_REDIRECT_BINDING;
-
-        // fall back to post binding if no redirect URL is set
-        if (postUrlSet && !redirectUrlSet) {
-            bindingType = SAML_POST_BINDING;
-        }
-
-        // evaluate how the user logged in
-        if (clientSession != null) {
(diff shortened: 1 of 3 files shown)


mposolda (approved):
@rmartinc Thanks!

ahus1:
Pinging @nxpthx as this fix was related to the PR created for #32262.

I'm sorry that I've contributed to this issue by approving the PR.

nxpthx:
@rmartinc While I agree, the the "default fallback" changed, which is fixed by your update, I am not sure if we still behave like users would expect.
So yes it makes sense to use the "force Options" and afterwards the same Binding type as used for Login IF there is only the master-saml-url configured.
If a user explicitly configured a URL for logging out via a specific binding, he might expect, that this URL is taken for logout.

rmartinc:
@nxpthx Yes, the only important change is that if, and it is just executed in the last case when there is no redirect logout Url (only adminUrl). So I think we are good. It's a bit of a corner case when there is no specific logout URL configured.

Issue #40387, opened 2025-06-10
Improve management of additional datasources


mabartos:
### Description

At this moment, in Keycloak, there is support for additional datasources, but many items missing. The additional datasources might be useful in situations when users can specify where certain Keycloak entities are stored. It means like users are able to specify that they want to have users in different DB than clients, or persistent sessions. It might be good for persistent sessions as possibly different type of store can be used for this purposes.

In order to improve the additional datasources story, we need to add/improve:

- [x] Keycloak properties (CLI, conf/properties files,...)
- [x] Keycloak ENV handling (https://github.com/keycloak/keycloak/pull/40760)
- [ ] Documentation and guides
- [x] Quickstarts

vmuzikar:
Resolving. #39250 will be handled as a follow-up.

PR #40760, opened 2025-06-27
test: Provide test cases for datasources ENV vars handling

mabartos:
Related-to: https://github.com/keycloak/keycloak/issues/40387

We should provide some simple test cases for ENV var handling for additional datasources, as we're lacking them. 

@shawkins Could you please check it? Thanks

### DIFF
--- a/quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/DatasourcesConfigurationTest.java
+++ b/quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/DatasourcesConfigurationTest.java
@@ -387,4 +387,52 @@ public class DatasourcesConfigurationTest extends AbstractConfigurationTest {
                 "quarkus.datasource.\"users\".jdbc.max-size", "115"
         ));
     }
+
+    @Test
+    public void envVarsHandling() {
+        putEnvVars(Map.of(
+                "KC_DB_KIND_USER_STORE", "postgres",
+                "KC_DB_URL_FULL_USER_STORE", "jdbc:postgresql://localhost/KEYCLOAK",
+                "KC_DB_USERNAME_USER_STORE", "my-username",
+                "KC_DB_KIND_MY_STORE", "mariadb"
+        ));
+        initConfig();
+
+        assertConfig(Map.of(
+                "db-kind-user-store", "postgres",
+                "db-url-full-user-store", "jdbc:postgresql://localhost/KEYCLOAK",
+                "db-username-user-store", "my-username",
+                "db-kind-my-store", "mariadb",
+                "db-kind-my.store", "mariadb"
+        ));
+
+        assertExternalConfig(Map.of(
+                "quarkus.datasource.\"user-store\".db-kind", "postgresql",
+                "quarkus.datasource.\"user-store\".jdbc.url", "jdbc:postgresql://localhost/KEYCLOAK",
+                "quarkus.datasource.\"user-store\".username", "my-username",
+                "quarkus.datasource.\"my-store\".db-kind", "mariadb",
+                "quarkus.datasource.\"my.store\".db-kind", "mariadb"
+        ));
+    }
+
+    @Test
+    public void envVarsSpecialChars() {
+        putEnvVars(Map.of(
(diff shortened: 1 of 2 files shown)


quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/DatasourcesConfigurationTest.java
```diff
@@ -387,4 +387,52 @@ public void poolSizeInherit() {
...
+                "db-kind-my-store", "mariadb",
+                "db-kind-my.store", "mariadb"
+        ));
+
+        assertExternalConfig(Map.of(
+                "quarkus.datasource.\"user-store\".db-kind", "postgresql",
+                "quarkus.datasource.\"user-store\".jdbc.url", "jdbc:postgresql://localhost/KEYCLOAK",
+                "quarkus.datasource.\"user-store\".username", "my-username",
+                "quarkus.datasource.\"my-store\".db-kind", "mariadb",
+                "quarkus.datasource.\"my.store\".db-kind", "mariadb"
```
mabartos, line 414: Even though we're able to obtain the configuration for both variants (`my-store` and `my.store`), the named datasource in Quarkus is `my.store`. See https://github.com/keycloak/keycloak/pull/40760/files#diff-161196ef0d1fee87940b57c35aa7d52ccc6da78c93eed92064e51483e047e101R36

It would be probably better to have the `-` (minus) char the default for datasources, but probably not a big deal when we document it. 

cc: @shawkins

shawkins: > Even though we're able to obtain the configuration for both variants (my-store and my.store)

That is a side effect of the quarkus env config source. Anything that could map to the env form can be used. Only the my.store form is the canonical name that is advertised to quarkus (my!store). It would be fine to remove the assertions looking for the my-store form.

> It would be probably better to have the - (minus) char the default for datasources

Yes, it would be best to do that for everything that is not logging. It doesn't make sense to default to the quarkus convention when we use '-'. I was going to add this for the spi work, but that didn't move forward, so it can be done here if you want.

mabartos: > Yes, it would be best to do that for everything that is not logging. It doesn't make sense to default to the quarkus convention when we use '-'. I was going to add this for the spi work, but that didn't move forward, so it can be done here if you want.

@shawkins Agree. So, could you please provide such changes?

vmuzikar: > it would be best to do that for everything that is not logging. It doesn't make sense to default to the quarkus convention when we use '-'

+1
Though, it would be considered a breaking change. We're already releasing the initial additional datasources support in 26.3

vmuzikar (approved):
I'd still proceed with the PR as is. We should address the `.` vs `-` in a separate PR.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/DatasourcesConfigurationTest.java
mabartos: > Though, it would be considered a breaking change. We're already releasing the initial additional datasources support in 26.3

@vmuzikar Depends on how you look at it. Understand that we already have these options for it, but we don't state anywhere that it's supported, and we don't even have the documentation for it yet. So maybe it should not be a big deal.

mabartos:
> I'd still proceed with the PR as is. We should address the . vs - in a separate PR.

@vmuzikar Agree.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/DatasourcesConfigurationTest.java
vmuzikar: @mabartos Good point that there's no docs. Though it's included in `--help` and NOT marked as a preview or something. We should've probably done that, I didn't realize it. :/

PR #40762, opened 2025-06-27
Change naming for disabling additional datasource

### LINKED ISSUE #40761: Change naming for disabling additional datasource
### Description

As mentioned in the comment https://github.com/keycloak/keycloak/pull/37640#discussion_r2093052242, we decided to use different naming for disabling/deactivating a particular datasource. 

Change from `db-active-<datasource>` to `db-enabled-<datasource>` to be aligned with other KC properties.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mabartos:
Closes #40761

@vmuzikar Could you please check it? Thanks!

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
@@ -100,10 +100,10 @@ public class DatabaseOptions {
             .defaultValue(10000)
             .build();
 
-    public static final Option<Boolean> DB_ACTIVE_DATASOURCE = new OptionBuilder<>("db-active-<datasource>", Boolean.class)
+    public static final Option<Boolean> DB_ENABLED_DATASOURCE = new OptionBuilder<>("db-enabled-<datasource>", Boolean.class)
             .category(OptionCategory.DATABASE_DATASOURCES)
             .defaultValue(true)
-            .description("Deactivate specific named datasource <datasource>.")
+            .description("If specific named datasource <datasource> should be enabled.")
             .build();
 
     /**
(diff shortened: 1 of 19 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
```diff
@@ -100,10 +100,10 @@ public class DatabaseOptions {
             .defaultValue(10000)
             .build();
 
-    public static final Option<Boolean> DB_ACTIVE_DATASOURCE = new OptionBuilder<>("db-active-<datasource>", Boolean.class)
+    public static final Option<Boolean> DB_ENABLED_DATASOURCE = new OptionBuilder<>("db-enabled-<datasource>", Boolean.class)
             .category(OptionCategory.DATABASE_DATASOURCES)
             .defaultValue(true)
-            .description("Deactivate specific named datasource <datasource>.")
+            .description("Enable specific named datasource <datasource>.")
```
vmuzikar: Given the default value is `true`, I'd suggest using slightly different desc, maybe something like:
```suggestion
            .description("Disable specific named datasource <datasource>.")
```
WDYT?

mabartos: @vmuzikar Both work for me as we don't have any convention on how to name these boolean prop mappers. 

It depends on the PoV again; it might also be awkward to see "Disable something. Default: true". It might evoke the "disablification" is true, so it is disabled :smile: So, I rather used the "Enable something. Default: true" -> enabled by default. 

Updated a little bit differently - as we usually see for our props.

vmuzikar (approved):
LGTM, thanks!

mabartos:
Closing this PR as the options for additional datasources might need some more refinements and possibly the best approach for that would be to enable the datasource at build time, but having the possibility to deactivate in runtime. I'll provide some more details later.

PR #40766, opened 2025-06-27
make abstract class AbstractUserRoleMappingMapper public

### LINKED ISSUE #40765: Make abstract class AbstractUserRoleMappingMapper public
### Description

Currently, the abstract class `AbstractUserRoleMappingMapper` is package private only. It's the only abstract class in package `org.keycloak.protocol.oidc.mappers` which is package-private. I don't see a reason why.

I need a custom role mapper with very special logic, but would like to use also the`AbstractUserRoleMappingMapper` in my implementation. Unfortunately I can't use it atm.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

dasniko:
(was package-private)

closes #40765


dasniko:
I don't think that this flaky test is related to this PR

ahus1:
I'm ok to change it, but as usual, please note that you use APIs in the service package at your own risk, and they might change in patch or minor releases.

ahus1 (approved):

PR #40787, opened 2025-06-30
Fix typo in consent scope

### LINKED ISSUE #40786: Typo in Consent Scope Representation
### Before reporting an issue

### Area

account/api

### Describe the bug

There's a typo in the response of the `/account/applications` endpoint

Image

### Version

26.2.5

### Regression

### Expected behavior

It should return `displayText`

### Actual behavior

It returns `displayTest`

### How to Reproduce?

Do a request to the `/account/applications` endpoint

### Anything else?

_No response_

SferaDev:
Closes #40786

### DIFF
--- a/core/src/main/java/org/keycloak/representations/account/ConsentScopeRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/account/ConsentScopeRepresentation.java
@@ -23,15 +23,15 @@ public class ConsentScopeRepresentation {
 
     private String name;
 
-    private String displayTest;
+    private String displayText;
 
     public ConsentScopeRepresentation() {
     }
 
-    public ConsentScopeRepresentation(String id, String name, String displayTest) {
+    public ConsentScopeRepresentation(String id, String name, String displayText) {
         this.id = id;
         this.name = name;
-        this.displayTest = displayTest;
+        this.displayText = displayText;
     }
 
     public String getId() {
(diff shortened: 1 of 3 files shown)


ahus1:
Adding a "hold" label as this would be a breaking change to the current users of the API

core/src/main/java/org/keycloak/representations/account/ConsentScopeRepresentation.java
```diff
@@ -23,15 +23,22 @@ public class ConsentScopeRepresentation {
 
     private String name;
 
+    /**
+     * @deprecated Use {@link #displayText} instead. This field will be removed in KC 27.0.
+     */
+    @Deprecated
     private String displayTest;
```
ahus1: Could you please remove this deprecated field, and only keep the new one? Then the old getter/setter would only point to that new field. IMHO this should be enough to create the right JSON, and parse the JSON correctly, assuming people will send one or the other.

ahus1 (changes requested):
Thank you for the PR. See below for one suggested change.

core/src/main/java/org/keycloak/representations/account/ConsentScopeRepresentation.java
SferaDev: You're right, I've double checked and the getter and setter in the representation are enough. Thanks!

image

ahus1 (approved):

PR #40789, opened 2025-06-30
 Fallback to display text consent scopes

SferaDev:
Closes #40788

image

### DIFF
--- a/js/apps/account-ui/src/applications/Applications.tsx
+++ b/js/apps/account-ui/src/applications/Applications.tsx
@@ -216,7 +216,8 @@ export const Applications = () => {
                       </DescriptionListTerm>
                       {application.consent.grantedScopes.map((scope) => (
                         <DescriptionListDescription key={`scope${scope.id}`}>
-                          <CheckIcon /> {t(scope.name as TFuncKey)}
+                          <CheckIcon />{" "}
+                          {t(scope.name as TFuncKey, scope.displayText)}
                         </DescriptionListDescription>
                       ))}
                     </DescriptionListGroup>


js/apps/account-ui/src/applications/Applications.tsx
```diff
@@ -216,7 +216,7 @@ export const Applications = () => {
                       </DescriptionListTerm>
                       {application.consent.grantedScopes.map((scope) => (
                         <DescriptionListDescription key={`scope${scope.id}`}>
-                          <CheckIcon /> {t(scope.name as TFuncKey)}
+                          <CheckIcon /> {t(scope.displayText as TFuncKey)}
```
SebastEnn: Lets take the profile scope as an example.
This change results in displaying _profile_ instead of _${profileScopeConsentText}_. 
What we want instead is to display the variable of profileScopeConsentText in the users language (in English _User profile_).

This isn't solving the issue.
`{t("myLanguageVariable")}` should translate the text, but in this case it doesn't.

SebastEnn: I've looked into #40787.

Changing this to `scope.displayTest` until the mentioned typo issue is fixed might be a temporary solution. 
I think the typo will be fixed at the next major release [27.0](https://github.com/keycloak/keycloak/milestone/59) in about 9 months.

SferaDev: Yeah, sorry this was in draft and not ready for review. Now I have finished it 😉

js/apps/account-ui/src/applications/Applications.tsx
```diff
@@ -216,7 +216,8 @@ export const Applications = () => {
                       </DescriptionListTerm>
                       {application.consent.grantedScopes.map((scope) => (
                         <DescriptionListDescription key={`scope${scope.id}`}>
-                          <CheckIcon /> {t(scope.name as TFuncKey)}
+                          <CheckIcon />{" "}
+                          {t(scope.name as TFuncKey, scope.displayTest)}
```
SebastEnn: You've added `scope.displayTest` (typo intended) as a second argument. The string behind it will be displayed if no translation for the first argument has been found.
Can you explain what the idea behind this is? Maybe with example cases?

SferaDev: Sure, the case is when defining a custom scope that is not built-in in Keycloak and has not been added as a translation in the realm settings.

You can find a picture in the PR description with one case

ssilvert: I don't have a problem with this, but you do need to fix the typo.

ssilvert (changes requested):

ssilvert (approved):

Issue #32724, opened 2024-09-07
Refine property mapping logic


shawkins:
### Description

To make the property mapping logic more straight-forward I propose to:
- remove the use of Optional<String> - it's debatable whether using an Optional argument is a good practice and we are already checking for null when it's used as a return value.
- separate parent mapping / transformation from the regular transformation. Several transformers read oddly because the incoming value could be from the parent / mapFrom value or the actual option. It will clarify and help future proof things to have that functionality separate
- the getConfigValue logic should be made more regular. There are currently 3 separate logic paths that could be better aligned
- add support for unsetting a value - to support #23771

PR #40790, opened 2025-06-30
Unable to configure TLS reloading in Keycloak version 26.2.0 or later

mabartos:
Closes #40713

The `-1` was propagated to the `quarkus.management.ssl.certificate.reload-period` property, which was not correct. The reload period for HTTP was not set, which is the correct way. This PR manages both properties behave the same.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
@@ -93,8 +93,7 @@ public final class HttpPropertyMappers {
                         .build(),
                 fromOption(HttpOptions.HTTPS_CERTIFICATES_RELOAD_PERIOD)
                         .to("quarkus.http.ssl.certificate.reload-period")
-                        // -1 means no reload
-                        .transformer((value, context) -> "-1".equals(value) ? null : value)
+                        .transformer(HttpPropertyMappers::transformNegativeReloadPeriod)
                         .paramLabel("reload period")
                         .build(),
                 fromOption(HttpOptions.HTTPS_CERTIFICATE_FILE)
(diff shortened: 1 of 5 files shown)


shawkins:
For additional context, this was due to https://github.com/keycloak/keycloak/issues/32724 - for most mapFrom operations we were using a distinct input value, such that the general transformer did not apply. So the change created a separate mapFrom transformer. It also seemed more complicated to chain the transformations - apply mapFrom first, then the general one, so they were left separate. 

An alternative to this change is to automatically use the parent transformer when it is present as the mapFrom transformer when it is not specified.

shawkins:
@mabartos there are two options to prevent the redundency in this situation - it will come up again if we have any other base property with a tranformation.

1. Change https://github.com/keycloak/keycloak/blob/2b44c5676f7ebe340da55a94766a8a11f3cf0eb1/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java#L261 to `var theMapper = parentValue && this.parentMapper != null ? this.parentMapper : this.mapper;` - that is if we don't specify a parentMapper, then fall through to the tranformer.

2. Do cascading application of transforms - apply the parentMapper first, then apply the regular transform. While this is more logical, it wasn't desirable for specifying the cache config file - but ultimately won't currently make a difference.

mabartos:
> An alternative to this change is to automatically use the parent transformer when it is present as the mapFrom transformer when it is not specified.

@shawkins Thanks. Make sense to me to use the parent mapper if the mapFrom mapper is not specified. It seems to me it's not so invasive change, and we can backport it without risking the mapping will break apart. 

Btw. we would need to have a diagram on how the transformation works for prop mappers soon :smile:

shawkins:
> Btw. we would need to have a diagram on how the transformation works for prop mappers soon 😄

Right, it's not always clear how the map from, transformer, and default play together.

shawkins (approved):
LGTM, thanks @mabartos

vmuzikar (approved):
The gist of the changes LGTM but relying on @shawkins' deeper review.

PR #40794, opened 2025-06-30
Updating upgrading guide about changes in user-profile-commons.ftl template

### LINKED ISSUE #39562: Breaking template change: Unknown `locale` input field added to user-profile registration page
### Before reporting an issue

### Area

user-profile

### Describe the bug

Since Keycloak 26.2.0 a new input field named `locale` appeared on the registration page when using a custom user-profile-commons theme with the User Profile feature.
This seems to be a regression introduced in #38549.

The template was changed to add `type=hidden` to the locale field, but older themes just display this field as a regular input.

<image height="200" src="https://github.com/user-attachments/assets/8f31d928-3fa6-4323-b7f7-4b834382373a">

## References

- https://github.com/keycloak/keycloak/pull/38549
- https://github.com/keycloakify/keycloakify/issues/837

### Version

>= 26.2.0

### Regression

### Expected behavior

An existing theme for user-profile-commons continue to work as before without showing the additional `locale` input field.

### Actual behavior

Existing themes for user-profile-commons show an additional `locale` input field.

### How to Reproduce?

- Run Keycloak >= 26.2.0
- Configure a custom user-profile-commons.ftl theme with the previous template from https://github.com/keycloak/keycloak/blob/9d472355036b26cf29eb2b1a93f2dd70e8b70b32/themes/src/main/resources/theme/keycloak.v2/login/user-profile-commons.ftl
- Enable User Profile
- Open the user registration page

### Anything else?

It should be discussed if either:

- This regression should be noted in the migration guide for Keycloak 26.2.0
- The user-profile-commons template should be changed to be backwards compatible

pedroigor:
Closes #39562

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_2_6.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_2_6.adoc
@@ -37,3 +37,9 @@ In case the provider needs to modify the information sent by the identity provid
 If the data received from the identity provider is mandatory and cannot be modified, then the *Review Profile* step in the *First Login Flow* can be disabled to avoid any user intervention.
 
 For more information, see link:{adminguide_link}#_identity_broker_first_login[the First login flow section of the {adminguide_name}].
+
+=== Updates to the `user-profile-commons.ftl` theme template
+
+The `user-profile-commons.ftl` changed to improve support for localization. See https://github.com/keycloak/keycloak/issues/38029.
+As a result, and if you are extending this template, pages might start displaying a `locale` field. To avoid that, update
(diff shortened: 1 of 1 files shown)


ahus1 (approved):

ahus1:
@pedroigor - I clicked approve, but that was actually wrong: 

* The problem was introduced to a ticket merged to 26.2, and so the upgrading guide for 26.2 should contain that paragraph. And it needs to be added to the 26.2 branch and also the 26.3 and the main branch, otherwise people will never see it in upstream.

PR #40795, opened 2025-06-30
Improve realm search by name (#40793)

### LINKED ISSUE #40793: Realm search in manage realm view should also include display-name
### Description

The search function in the manage-realm view should also include the display name.

### Discussion

_No response_

### Motivation

In scenarios where realm names are technical (UUIDs or internal names) the realm display name is used as a human identifier.
The ability so search for a realm by display name is more user friendly.

### Details

Current realm search in manage-realm:

Image

Realm with display name including demo is not considered.
Image

thomasdarimont:
Default realm search now also considers realm displayName by name. To explicitly search by name or displayName we now support the prefixes "name:" and "displayName:".

We now also consider the realm display name if set in realm searches by name.

Fixes #40793

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java
@@ -93,6 +93,8 @@ import org.keycloak.protocol.oidc.OIDCLoginProtocol;
  */
 public class JpaRealmProvider implements RealmProvider, ClientProvider, ClientScopeProvider, GroupProvider, RoleProvider, DeploymentStateProvider {
     protected static final Logger logger = Logger.getLogger(JpaRealmProvider.class);
+    public static final String REALM_NAME_SEARCH_PREFIX = "name:";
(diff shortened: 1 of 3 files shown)


edewit:
Display name can be translated, so it doesn't take that into account

thomasdarimont:
@edewit good point, yes indeed if realm admins choose to translate the realm display name, then the search will not work for those scenarios.
However, the majority of installations I have seen use a fixed realm display name in local deployments, especially in SaaS deployments with a realm per tenant model, where the realm name is often generated. 
Think of a long list of realms with uuids as names but redable display names. Admins who manage the platform now needs to maintain a separate searchable list that map (technical) realm-name to realm display names. since the display name is the only way to have a human friendly "name" in that case.

ssilvert (changes requested):
@thomasdarimont I think this is a nice improvement.  But the user needs to know how to do the different queries from the admin UI.

We have a few places in the UI where you can search based on different criteria.  See the user search for an example.

thomasdarimont:
Accidentially pushed this.

PR #40804, opened 2025-06-30
fix: reducing memory footprint

### LINKED ISSUE #35932: Importing a realm takes more than 1 minute when multiple others exist.
### Description

I have a keycloak installation with a 1000 realms that look like the one I am trying to import. (with few clients,scopes...)
[jhipster-realm.json](https://github.com/user-attachments/files/18144114/jhipster-realm.json)
Each time I try to import a similar realm it takes more than a minute.

### Discussion

_No response_

### Motivation

Make keycloak perform well with many realms.

### Details

For example getClientScopes takes 4 seconds when I have 10_000 client scopes, but when I have few it takes few miliseconds. This doesn't make much sense to me since scopes and there mappings are indexed in the DB.
Same is true for getClientRole.

shawkins:
closes: #35932

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/CachedCount.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/CachedCount.java
@@ -22,18 +22,18 @@ import org.keycloak.models.cache.infinispan.entities.InRealm;
 
 public class CachedCount extends AbstractRevisioned implements InRealm {
 
-    private final RealmModel realm;
+    private final String realm;
     private final long count;
 
     public CachedCount(Long revision, RealmModel realm, String cacheKey, long count) {
         super(revision, cacheKey);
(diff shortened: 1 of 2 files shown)


ahus1:
@shawkins - this looks like a valid fix. Unfortunately also CachedMembership seems to have the same problematic pattern. Would that be a follow-up issue?

shawkins:
> @shawkins - this looks like a valid fix. Unfortunately also CachedMembership seems to have the same problematic pattern. Would that be a follow-up issue?

Seems fine to address them together even if CachedMembership isn't used by the issue reproducer.

ahus1 (approved):
Approving, thank you for the fix. Not merging yet as @vramik is added as a second reviewer.

Adding backport tags to the parent issue as it is IMHO worth backporting.

vramik (approved):
Thanks @shawkins for the PR

PR #40820, opened 2025-07-01
Changing default passwordless webauthn policy to follow recommended values in the documentation

rmartinc:
Closes #40792

The PR creates two `WebAuthnPolicy` with the default values for each version (two-factor and passwordless). This way we can change defaults more easily the next time. Tests modified to work with the new values (for example a new selenium authenticator is needed now for all passwordless tests). We need another round of documentation changes, but we can do that when we remove the paragraph saying that passkeys are tech preview.


graziang (approved):
@rmartinc Thanks!

mposolda (approved):
@rmartinc Nice improvement, Thanks!
@graziang Thanks for the review!

PR #40827, opened 2025-07-01
Load declarative component instead of any user profile component

### LINKED ISSUE #40817: DeclarativeUserProfileProvider#setConfiguration picks wrong component when multiple UserProfileProvider implementations exist
### Before reporting an issue

### Area

user-profile

### Describe the bug

I am attempting to implement a second UserProfileProvider SPI called TestUserProfileProvider that extends DeclarativeUserProfileProvider. My goal is to store and manage an additional user profile configuration in the DB and parse users received over a custom endpoint, leveraging the existing DeclarativeUserProfileProvider.

I created the provider and corresponding factory. The factory’s getId() returns test-user-profile. I also customized the configuration keys as follows:
```
public static final String UP_COMPONENT_CONFIG_KEY = "aup.test-user.profile.config";
protected static final String PARSED_UP_CONFIG_COMPONENT_KEY = "aup.parsed.test-up.config";
```
However, when [DeclarativeUserProfileProvider#setConfiguration](https://github.com/keycloak/keycloak/blob/c9f38d36a97764dc161b6126eb8eff24b3c9a1c1/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProvider.java#L216) is invoked, it executes:
```
Optional<ComponentModel> optionalComponent = realm.getComponentsStream(realm.getId(), UserProfileProvider.class.getName()).findAny();
```

This findAny() may return any component of type UserProfileProvider, including my TestUserProfileProvider, rather than the DeclarativeUserProfileProvider. This causes the provider_id column to incorrectly have the value test-user-profile instead of declarative-user-profile.

### Version

main

### Regression

### Expected behavior

DeclarativeUserProfileProvider#setConfiguration should select the correct component for its own configuration—specifically, the component whose provider_id matches declarative-user-profile. This ensures that it loads and stores configuration as intended without interference from other UserProfileProvider implementations.

### Actual behavior

When multiple UserProfileProvider implementations exist, findAny() may return a different provider (e.g., test-user-profile). This results in:
	•	Configuration being associated with the wrong provider (provider_id = test-user-profile instead of declarative-user-profile)
	•	Potential misconfiguration and runtime errors

### How to Reproduce?

1.	Implement a custom UserProfileProvider (e.g., TestUserProfileProvider) that extends DeclarativeUserProfileProvider.
2.	Ensure its factory getId() returns test-user-profile.
3.	Register both DeclarativeUserProfileProvider and TestUserProfileProvider in Keycloak.
4.	Trigger code that invokes DeclarativeUserProfileProvider#setConfiguration (E.g. update a user profile attribute in realm settings).
5.	Observe that realm.getComponentsStream(...).findAny() may return the wrong component.
6.	Check DB and note the provider_id is test-user-profile where declarative-user-profile is expected.

### Anything else?

_No response_

statist32:
The component corresponding to the UserProfileProvider should be loaded to correctly load and store the profile config in respect to the provider_id and provider_type.

Closes #40817

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -204,6 +204,15 @@ public class OID4VCIssuerEndpoint {
     }
 
     /**
+     * Generates a unique notification ID for use in CredentialResponse.
+     *
+     * @return a unique string identifier
+     */
+    private String generateNotificationId() {
+        return SecretGenerator.getInstance().randomString();
+    }
+
+    /**
      * the OpenId4VCI nonce-endpoint
      *
      * @return a short-lived c_nonce value that must be presented in key-bound proofs at the credential endpoint.
@@ -408,7 +417,9 @@ public class OID4VCIssuerEndpoint {
 
         Object theCredential = getCredential(authResult, supportedCredentialConfiguration, credentialRequestVO);
         if (SUPPORTED_FORMATS.contains(requestedFormat)) {
-            responseVO.addCredential(theCredential);
+            responseVO
+                    .addCredential(theCredential)
+                    .setNotificationId(generateNotificationId());
         } else {
             throw new BadRequestException(getErrorResponse(ErrorType.UNSUPPORTED_CREDENTIAL_TYPE));
         }
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialResponse.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialResponse.java
@@ -50,11 +50,12 @@ public class CredentialResponse {
         return this;
     }
 
-    public void addCredential(Object credential) {
+    public CredentialResponse addCredential(Object credential) {
         if (this.credentials == null) {
             this.credentials = new ArrayList<>();
         }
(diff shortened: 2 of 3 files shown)


services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProvider.java
```diff
@@ -238,7 +240,7 @@ public void setConfiguration(UPConfig configuration) {
 
     private Optional<ComponentModel> getComponentModel() {
         RealmModel realm = session.getContext().getRealm();
-        return realm.getComponentsStream(realm.getId(), UserProfileProvider.class.getName()).findAny();
+        return realm.getComponentsStream(realm.getId(), UserProfileProvider.class.getName()).filter(componentModel -> componentModel.getProviderId().equals(DeclarativeUserProfileProviderFactory.ID)).findFirst();
```
pedroigor: The provider ID is already available from `org.keycloak.userprofile.DeclarativeUserProfileProvider#providerId`. It should be better than having it hard-coded.

In your case, it will also help to resolve the proper component using your provider ID given that you are extending this class.

pedroigor (changes requested):
@statist32 Thanks, LGTM. A single comment about how to resolve components by the provider ID.

pedroigor:
@statist32 Closed in favor of https://github.com/keycloak/keycloak/pull/41012. Thanks.

PR #40831, opened 2025-07-01
UserSession Offline removed from DB if not in cache

pruivo:
Fixes #40754

The method `org.keycloak.models.sessions.infinispan.InfinispanUserSessionProvider#importUserSession(org.keycloak.models.RealmModel, org.keycloak.models.UserSessionModel)` was rewritten. The main differences are that I'm using `putIfAbsent` to import the session, and I'm importing all the client sessions in parallel.

I also took the opportunity to fix the code warnings (unused code, parameters with constant values, missing types, etc).

### DIFF
--- a/docs/documentation/server_admin/topics/sessions/offline.adoc
+++ b/docs/documentation/server_admin/topics/sessions/offline.adoc
@@ -24,20 +24,4 @@ Clients can request an offline token by adding the parameter `scope=offline_acce
 
 {project_name} will limit its internal cache for offline user and offline client sessions to 10000 entries by default, which will reduce the overall memory usage for offline sessions.
 Items which are evicted from memory will be loaded on-demand from the database when needed.
-To set different sizes for the caches, edit {project_name}'s cache config file to set a `+<memory max-count="..."/>+` for those caches.
-
-If you disabled feature `persistent-user-sessions`, it is possible to reduce memory requirements using a configuration option that shortens lifespan for imported offline sessions. Such sessions will be evicted from the Infinispan caches after the specified lifespan, but still available in the database. This will lower memory consumption, especially for deployments with a large number of offline sessions.
-
-To specify the lifespan override for offline user sessions, start {project_name} server with the following parameter:
-
-[source,bash]
-----
---spi-user-sessions--infinispan--offline-session-cache-entry-lifespan-override=<lifespan-in-seconds>
-----
-
-Similarly for offline client sessions:
-
-[source,bash]
-----
---spi-user-sessions--infinispan--offline-client-session-cache-entry-lifespan-override=<lifespan-in-seconds>
-----
+See the server configuration guide to change this default.
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -42,6 +42,12 @@ configured directly via {project_name}: `cache-embedded-network-bind-address`, `
 `cache-embedded-network-external-address`, `cache-embedded-network-external-port`. Configuring ports using the old
 properties will still function as before, but we recommend to change to the CLI options as this may change in the future.
 
+=== Volatile user sessions affecting offline session memory requirements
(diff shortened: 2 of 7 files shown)


ahus1:
Thank you for the PR. I assume you want to enable users to have a size-limited offline user session and client session cache. 

If you want to go down that route, there are IMHO some additional changes necessary: 

* We recently added safeguards that disallow setting a cache limit. If we now support a cache limit, having a cache limit of 10_000 for the offline cache would be a sensible new default, and we should allow that CLI parameter then.  
* When using a size-limited cache, we found in our testing that this should be used with a single owner, as the eviction could lead to a situation where the entry is present at the backup owner, while the primary owner already forgot about it, and then the backup owner will read it from the local cache still. That should be part of the safeguard/automatic configuration
* As an unbounded cache for the offline sessions is a bad idea, we would only support size-limited offline caches, so no need to keep the old behavior. 

Note that when we upgrade the non-persistent user sessions like this, we would need to maintain this code side-by-side with the persistent user sessions. Some of the logic (for example batching the writes to the DB) are quite different in the two implementations, and will require understanding both in the team.

pruivo:
> I assume you want to enable users to have a size-limited offline user session and client session cache.

What I want is `InfinispanUserSessionProvider` to be an "infinispan only " provider with no database access, even for offline sessions... but I will piss off a lot of people if I do that :rofl: 

Yes, I want the users to have options: small deployments to use unbounded caches and large deployment with bounded caches.

> If you want to go down that route, there are IMHO some additional changes necessary:

Do you want me to do them in this PR or another one? 

> When using a size-limited cache, we found in our testing that this should be used with a single owner, as the eviction could lead to a situation where the entry is present at the backup owner, while the primary owner already forgot about it, and then the backup owner will read it from the local cache still. That should be part of the safeguard/automatic configuration

Read-only is fine; the problem is when updates are involved. Because we use `computeIfPresent` to update the cache, some owners may miss the update.

> Some of the logic (for example batching the writes to the DB) are quite different in the two implementations, and will require understanding both in the team.

Honestly, the only database accesses I see on `InfinispanUserSessionProvider` are for create, import, and remove. If you look at the user and client session adapters, the setter methods never interact with the database (except for `org.keycloak.models.sessions.infinispan.UserSessionAdapter#setLastSessionRefresh`) :shrug: 
For example:

https://github.com/keycloak/keycloak/blob/4dc4de7c1218bd129b4ae6853a74b7a2954b6eec/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/AuthenticatedClientSessionAdapter.java#L175-L191

ahus1:
> Do you want me to do them in this PR or another one?

Due to the latest hardcoded safeguards here, I think we need to do additional changes, otherwise users won't be able to use it with a size-limited cache, as we would simply ignore those settings:

https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java#L220-L233

Please schedule a 1:1 if you want to discuss it in more details, that might be quicker.

pruivo:
@ahus1 I updated/removed the safeguards.

ahus1:
I had a look, updated some docs and pushed it in a more opinionated and IMHO safe-to-use direction. Please let me know your thoughts. Happy to jump on a call during the next days.

pruivo:
@ahus1 I'm ok with the changes.

ahus1 (approved):
I gave it another test locally and approve it. Thank you for the PR.

ahus1:
@pruivo - it would be nice if you could add the enhancement of parallel loading of client sessions also to PersistentUserSessionProvider#importSessionsWithExpiration. Thanks!

pruivo:
> @pruivo - it would be nice if you could add the enhancement of parallel loading of client sessions also to PersistentUserSessionProvider#importSessionsWithExpiration. Thanks!

Issue to track this idea:  https://github.com/keycloak/keycloak/issues/41074

PR #40834, opened 2025-07-01
fix: map just logging env wildcards to .

### LINKED ISSUE #40833: Change database wildcards to default to - instead of .
We explicitly document the expectation for logging, but have left the handling for other wildcards unspecified. The intent in other changes for 26.3 was to make everything else use -, but that did not end up making it in.

_Originally posted by @mabartos in https://github.com/keycloak/keycloak/pull/40760#discussion_r2177623822_

shawkins:
closes: #40833

### DIFF
--- a/docs/guides/server/configuration.adoc
+++ b/docs/guides/server/configuration.adoc
@@ -167,10 +167,13 @@ PowerShell handles quotes differently. It interprets the quoted string before pa
 
 === Formats for environment variable keys with special characters
 
-Some configuration keys, such as those for logging, may contain characters such as `_` or `$` - e.g. `kc.log-level-package.class_name`. Non-alphanumeric characters in your configuration key must be converted to `\_` in the corresponding environment variable key.
+Non-alphanumeric characters in your configuration key must be converted to `_` in the corresponding environment variable key.
 
-The automatic handling of the environment variable key may not preserve the intended key. For example `KC_LOG_LEVEL_PACKAGE_CLASS_NAME` will become `kc.log-level-package.class.name` because logging properties default to replacing `_` in the "wildcard"
-part of the key with `.` as that is what is most commonly needed in a class / package name. However this does not match the intent of `kc.log-level-package.class_name`.
+Env variables are converted back to normal option keys by lower-casing the name and replacing `\_` with `-`. Logging wildcards are the exception as `_` in the category is replaced with period `.` instead. Logging categories are commonly class / package names, which are more likely to use `.` rather than `-`.
+
+WARNING: automatic mapping of the environment variable key to option key may not preserve the intended key 
+
+For example `kc.log-level-package.class_name` will become the environment variable key `KC_LOG_LEVEL_PACKAGE_CLASS_NAME`. That will automatically be mapped to `kc.log-level-package.class.name` because `_` in the logging category will be replaced by `.`. Unfortunately this does not match the intent of `kc.log-level-package.class_name`.
 
 You have a couple of options in this case:
 
--- a/docs/guides/server/configuration.adoc
+++ b/docs/guides/server/configuration.adoc
@@ -169,9 +169,9 @@ PowerShell handles quotes differently. It interprets the quoted string before pa
 
 Non-alphanumeric characters in your configuration key must be converted to `_` in the corresponding environment variable key.
 
-Env variables are converted back to normal option keys by lower-casing the name and replacing `\_` with `-`. Logging wildcards are the exception as `_` in the category is replaced with period `.` instead. Logging categories are commonly class / package names, which are more likely to use `.` rather than `-`.
+Environment variables are converted back to normal option keys by lower-casing the name and replacing `\_` with `-`. Logging wildcards are the exception as `_` in the category is replaced with `.` instead. Logging categories are commonly class / package names, which are more likely to use `.` rather than `-`.
 
-WARNING: automatic mapping of the environment variable key to option key may not preserve the intended key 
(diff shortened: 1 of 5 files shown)


docs/guides/server/configuration.adoc
```diff
@@ -167,10 +167,13 @@ PowerShell handles quotes differently. It interprets the quoted string before pa
 
 === Formats for environment variable keys with special characters
 
-Some configuration keys, such as those for logging, may contain characters such as `_` or `$` - e.g. `kc.log-level-package.class_name`. Non-alphanumeric characters in your configuration key must be converted to `\_` in the corresponding environment variable key.
+Non-alphanumeric characters in your configuration key must be converted to `_` in the corresponding environment variable key.
 
-The automatic handling of the environment variable key may not preserve the intended key. For example `KC_LOG_LEVEL_PACKAGE_CLASS_NAME` will become `kc.log-level-package.class.name` because logging properties default to replacing `_` in the "wildcard"
-part of the key with `.` as that is what is most commonly needed in a class / package name. However this does not match the intent of `kc.log-level-package.class_name`.
+Env variables are converted back to normal option keys by lower-casing the name and replacing `\_` with `-`. Logging wildcards are the exception as `_` in the category is replaced with period `.` instead. Logging categories are commonly class / package names, which are more likely to use `.` rather than `-`.
```
mabartos:
```suggestion
Environment variables are converted back to normal option keys by lower-casing the name and replacing `\_` with `-`. Logging wildcards are the exception as `_` in the category is replaced with period instead. Logging categories are commonly class / package names, which are more likely to use `.` rather than `-`.
```

docs/guides/server/configuration.adoc
```diff
@@ -167,10 +167,13 @@ PowerShell handles quotes differently. It interprets the quoted string before pa
...
+
+WARNING: automatic mapping of the environment variable key to option key may not preserve the intended key
```
mabartos:
```suggestion
WARNING: Automatic mapping of the environment variable key to option key may not preserve the intended key. 
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
```diff
@@ -119,7 +123,10 @@ public boolean matchesWildcardOptionName(String name) {
 
     public Optional<String> getKcKeyForEnvKey(String envKey, String transformedKey) {
         if (transformedKey.startsWith(fromPrefix)) {
-            return Optional.ofNullable(getFrom(envKey.substring(fromPrefix.length()).toLowerCase().replace("_", ".")));
+            if (replacementChar != null) {
+                return Optional.ofNullable(getFrom(envKey.substring(fromPrefix.length()).toLowerCase().replace('_', replacementChar)));
+            }
+            return Optional.of(transformedKey);
```
mabartos, line 129: ATM, there's no way the replacementChar might be null, right? 

I'd rather throw an exception if the replacementChar is null and not returning the transformedKey.

Or what's the intention here?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
```diff
@@ -42,6 +42,10 @@ public WildcardPropertyMapper(Option<T> option, String to, BooleanSupplier enabl
             throw new IllegalArgumentException("Invalid wildcard from format. Wildcard must be at the end of the option.");
         }
 
+        if (option == LoggingOptions.LOG_LEVEL_CATEGORY) {
+            replacementChar = '.';
```
mabartos, line 46: jfyi - It seems fine to have it hardcoded for now, and if we have some other requirements for replacement char, we can adjust it. So, +1 to have it like this.

mabartos (commented):
@shawkins Overall LGTM and agree with defaulting to `-` instead of `.`.

Just added a few small suggestions.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
shawkins: It was meant to be null, will update.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
shawkins: That was the thought - this is the only exception we're likely to have.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
mabartos: I see the `transformedKey` contains the replaced chars done in the `KcEnvConfiSource`, so returning the `transformedKey` LGTM.

Thanks!

mabartos (approved):
@shawkins LGTM, thanks!

mabartos:
@shawkins Could you please create a backport for 26.3 ?

shawkins:
> @shawkins Could you please create a backport for 26.3 ?

https://github.com/keycloak/keycloak/pull/40865

PR #40835, opened 2025-07-01
Use MgmtPermissionsV2 by default

### LINKED ISSUE #40192: REST Admin API - ClientsResource response with 200 OK even needed roles are missing
### Before reporting an issue

### Area

admin/api

### Describe the bug

The clients resource API endpoint responses with 200 OK even when I'm not allowed to use the API.

The API should behave in a consistent way. This is unexpected behaviour.

### Version

26.2.5

### Regression

### Expected behavior

I expect a 402 Forbidden as response when the token has no admin-role or the query-clients role.

### Actual behavior

When I call with query-users role the REST API on https://SERVER/admin/realms/REALM/clients (GET) I get a status 200 OK with an empty JSON-array "[]" as response.

### How to Reproduce?

* create a user with the role "query-users"
* get an access token for the user
* call `
curl --location 'https://SERVER/admin/realms/REALM/clients' --header 'Authorization: Bearer ACCESS-TOKEN'
`

### Anything else?

_No response_

vramik:
Closes https://github.com/keycloak/keycloak/issues/40192


pedroigor:
@vramik LGTM. Can you solve the conflicts?

pedroigor (approved):

PR #40856, opened 2025-07-02
Verification of external OIDC token by introspection-endpoint. Adding…

### LINKED ISSUE #40167: Verification of external OIDC token by introspection-endpoint
### Description

This issue is a follow-up of https://github.com/keycloak/keycloak/pull/40134 . It is about support for exchange external-internal tokens when external token is the token issued by `OIDCIdentityProvider` .

Possible approach is to:
- Add support for "introspection endpoint" into the OIDC Identity provider configuration and probably also to the response from "Discovery endpoint" (Well-known endpoint).

- Introspection response could be validated if it contains either:
  - `azp` claim with client ID of the client configured as the client in the IDP
  - `aud` claim with client ID of the client configured as the client in the IDP
  - `aud` claim with issuer or token-endpoint of the KEycloak server.

Not sure if same approach could be used for `KeycloakIdentityProvider` ?

More details in the [google doc](https://docs.google.com/document/d/1hmUpMfvAwyRBvUhCD01IEGNjx1yIh9a8FpGCQlmrOno/edit?tab=t.0#heading=h.q65tmwm7g605) .

mposolda:
… ExternalInternalTokenExchangeV2Test

closes #40167
closes #40198

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
@@ -56,19 +56,25 @@ import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.models.UserSessionModel;
+import org.keycloak.protocol.oidc.AccessTokenIntrospectionProviderFactory;
 import org.keycloak.protocol.oidc.OIDCLoginProtocol;
+import org.keycloak.protocol.oidc.OIDCLoginProtocolService;
 import org.keycloak.protocol.oidc.TokenExchangeContext;
 import org.keycloak.protocol.oidc.TokenExchangeProvider;
 import org.keycloak.protocol.oidc.endpoints.AuthorizationEndpoint;
+import org.keycloak.protocol.oidc.endpoints.TokenIntrospectionEndpoint;
 import org.keycloak.protocol.oidc.utils.PkceUtils;
 import org.keycloak.representations.AccessTokenResponse;
 import org.keycloak.representations.JsonWebToken;
+import org.keycloak.representations.oidc.TokenMetadataRepresentation;
 import org.keycloak.services.ErrorPage;
 import org.keycloak.services.ErrorResponseException;
 import org.keycloak.services.Urls;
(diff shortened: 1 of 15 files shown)


graziang (approved):
@mposolda thanks!

rmartinc (approved):
@mposolda LGTM!

jonkoops (approved):

services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
```diff
@@ -754,6 +761,111 @@ protected BrokeredIdentityContext exchangeExternalTokenV2Impl(TokenExchangeConte
...
+        if (tokenClientId != null && tokenClientId.equals(getConfig().getClientId())) {
+            // Consider external token valid if issued to same client, which was configured as the client on IDP side
+            clientValid = true;
+        } else if (tokenMetadata.getAudience() != null && tokenMetadata.getAudience().length > 0) {
+            tokenAudiences = Arrays.stream(tokenMetadata.getAudience()).toList();
+            if (tokenAudiences.contains(getConfig().getClientId())) {
+                // Consider external token valid if client configured as the IDP client included in token audience
+                clientValid = true;
+            } else {
+                // Consider valid introspection also if token contains audience where URL is Keycloak server (either as issuer or as token-endpoint URL).
```
pedroigor, line 789: Can you elaborate why we need this validation? It seems KC specific?

mposolda: I am not sure if this is KC specific as it seems some servers might be adding the URL of the Keycloak server (which means either `issuer` or `tokenEndpoint URL` ) and use it as an `aud` of the token . This means that Keycloak server is an intended audience of the external token and hence could be considered valid for external-internal token exchange.

The similar requirement is mentioned in the specification for jwt-bearer grant type (which is to some extent related to the token exchange specification). See here https://datatracker.ietf.org/doc/html/rfc7523#section-3 and the point 3.

I've added some more details into the Google documentation https://docs.google.com/document/d/1hmUpMfvAwyRBvUhCD01IEGNjx1yIh9a8FpGCQlmrOno/edit?tab=t.0#heading=h.hcjkgh4519vu .

pedroigor (approved):

PR #40873, opened 2025-07-02
task: better document property mapping

shawkins:
closes: #40872

@mabartos this isn't quite a diagram, but it might be helpful. Additionally, more javadocs could be added to the builder methods.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
@@ -108,32 +108,20 @@ public class PropertyMapper<T> {
      * <p>
      * In preference order we are looking for:
      * <pre>
-     *  [ {@link #from} ] ---> [ {@link #mapFrom} ] ---> [ default ] ---> [try other external]
-     * (explicit)     (derived)       (fallback)
+     *  [ {@link #from} ] ---> [ {@link #mapFrom} ] ---> [ {@link #getDefaultValue()} ] ---> [ directly retrieve `to` value if applicable ]
+     * (explicit)      (derived)        (fallback)
      * </pre>
-     * <ul>
-     *   <li>the {@link #from} value       (the property is directly specified)
-     *   <li>the {@link #mapFrom} value    (the property is inferred from another value)
-     *   <li>the default
-     *   <li>If we did not find a value, we simply move on in the interceptor chain.
-     * </ul>
      * <p>
      *
      * <b>2. Transform found value</b>
      * <p>
      * If we found a value for the attribute name, it needs to be transformed via {@link #transformValue} method. How to transform it?
-     * <pre>
-     * [ name = {@link #from} ? ] ---> [ name = {@link #to} ? ] ---> [ value from parent ? ]
-     *        | y                    | y                     | y
-     *        |                      |                       |
-     *      value            transform(value)      parent.transform(value)
-     * </pre>
      * <ul>
-     *   <li>If the name matches {@link #to}, we use the {@link #mapper} from the {@link PropertyMapper.Builder#transformer} method.
-     *   <li>If we found the value on the parent side ({@link #mapFrom}), we use the {@link #parentMapper} specified either via {@link PropertyMapper.Builder#mapFrom(Option, ValueMapper)} method or parent's {@link PropertyMapper.Builder#transformer} method.
+     *   <li>If the name matches {@link #from} or we using the {@link #mapFrom} value, then apply the {@link PropertyMapper.Builder#transformer} or the {@link PropertyMapper.Builder#mapFrom(Option, ValueMapper)}
+     *   <li>Check if the value may contain an expression and expand it using SmallRye logic
+     *   <li>Finally the returned {@link ConfigValue} is made to match what was requested - with the name, value, rawValue, and ordinal set appropriately.
      * </ul>
      * <p>
-     * Finally the returned {@link ConfigValue} is made to match what was requested - with the name, value, rawValue, and ordinal set appropriately.
      */
     ConfigValue getConfigValue(String name, ConfigSourceInterceptorContext context) {
         String from = getFrom();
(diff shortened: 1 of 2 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -99,6 +99,23 @@ public class PropertyMapper<T> {
...
+     *
+     * <p>
+     * If we did not find a value, we simply move on in the interceptor chain.
+     * <p>
+     * If we found a value, it needs to be transformed via {@link #transformValue(String, ConfigValue, ConfigSourceInterceptorContext, boolean)}. If the mapFrom value is being used or the name matches "to", the
+     * either the parentMapper (from {@link PropertyMapper.Builder} mapFrom methods), or the mapper (from {@link PropertyMapper.Builder} transformer methods) is applied.
+     * <p>
+     * Finally the returned {@link ConfigValue} is made to match what was requested - with the name, value, rawValue, and ordinal set appropriately.
+     * <p>
+     */
```
mabartos: @shawkins Thanks for creating this PR!

```suggestion
    /**
     * This is the heart of the property mapping logic. In the first step, we need to find the value of the property and then transform it into a form of our needs.
     * <p>
     *
     * <b>1. Find value</b>
     * <p>
     * In preference order we are looking for:
     * <pre>
     *  [ {@link #from} ] ---> [ {@link #mapFrom} ] ---> [ default ] ---> [try other external]
     * (explicit)     (derived)       (fallback)
     * </pre>
     * <ul>
     *   <li>the {@link #from} value       (the property is directly specified)
     *   <li>the {@link #mapFrom} value    (the property is inferred from another value)
     *   <li>the default
     *   <li>If we did not find a value, we simply move on in the interceptor chain.
     * </ul>
     * <p>
     *
     * <b>2. Transform found value</b>
     * <p>
     * If we found a value for the attribute name, it needs to be transformed via {@link #transformValue} method. How to transform it?
     * <pre>
     * [ name = {@link #from} ? ] ---> [ name = {@link #to} ? ] ---> [ value from parent ? ]
     *        | y                    | y                     | y
     *        |                      |                       |
     *      value            transform(value)      parent.transform(value)
     * </pre>
     * <ul>
     *   <li>If the name matches {@link #to}, we use the {@link #mapper} from the {@link PropertyMapper.Builder#transformer} method.
     *   <li>If we found the value on the parent side ({@link #mapFrom}), we use the {@link #parentMapper} specified either via {@link PropertyMapper.Builder#mapFrom(Option, ValueMapper)} method or parent's {@link PropertyMapper.Builder#transformer} method.
     * </ul>
     * <p>
     * Finally the returned {@link ConfigValue} is made to match what was requested - with the name, value, rawValue, and ordinal set appropriately.
     */
```

I took it a little bit further to provide at least some illustration of the process to make it more understandable.

Here how it looks like in my IDE with all these formatting:

image

@shawkins WDYT? And please check if the info is valid.

shawkins: Looks good, but I'd say that the transformValue logic would be better described as: 

`if name == from || value from parent` then apply applicable mapper

Alternatively it might be better to just refer to that method here and add some docs there.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -100,6 +100,29 @@ public class PropertyMapper<T> {
...
+    /**
+     * This is the heart of the property mapping logic. In the first step, we need to find the value of the property and then transform it into a form of our needs.
+     * <p>
+     *
+     * <b>1. Find value</b>
+     * <p>
+     * In preference order we are looking for:
+     * <pre>
+     *  [ {@link #from} ] ---> [ {@link #mapFrom} ] ---> [ {@link #getDefaultValue()} ] ---> [ directly retrieve `to` value if applicable ]
+     * (explicit)      (derived)        (fallback)
```
mabartos:
```suggestion
     *  [ {@link #from} ] ---> [ {@link #mapFrom} ] ---> [ {@link #getDefaultValue()} ] ---> [ {@link #to} ]
     * (explicit)     (derived)           (fallback)         (fallback)
```

Might be better to only use the link to `to` as even default value is not returned if it's not specified. 
That might even look better. @shawkins WDYT?
image

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -100,6 +100,29 @@ public class PropertyMapper<T> {
...
+     * </pre>
+     * <p>
+     *
+     * <b>2. Transform found value</b>
+     * <p>
+     * If we found a value for the attribute name, it needs to be transformed via {@link #transformValue} method. How to transform it?
+     * <ul>
+     *   <li>If the name matches {@link #from} or we using the {@link #mapFrom} value, then apply the {@link PropertyMapper.Builder#transformer} or the {@link PropertyMapper.Builder#mapFrom(Option, ValueMapper)}
+     *   <li>Check if the value may contain an expression and expand it using SmallRye logic
```
mabartos:
```suggestion
     *   <li>If the value contains an expression, expand it using SmallRye logic
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -100,6 +100,29 @@ public class PropertyMapper<T> {
...
+     *   <li>Finally the returned {@link ConfigValue} is made to match what was requested - with the name, value, rawValue, and ordinal set appropriately.
+     * </ul>
+     * <p>
```
mabartos:
```suggestion
```

mabartos (commented):
@shawkins Just some small adjustments. Let me know what you think. Thanks!

shawkins:
Looks good, applied the changes.

mabartos (approved):
@shawkins Thanks! LGTM

PR #40904, opened 2025-07-03
implementing json patch

shawkins:
@keycloak/cloud-native this is roughly what json patch support looks like. We would also need to put the fabric8 zjsonpatch dependency in the lib. 

If we don't want the additional dependency, the alternative is to support jakarta.json / parsson (coming from Quarkus) - that seems slightly less efficient as we'll need to convert back and forth from the   JsonValue representation.

Let me know what you think.

### DIFF
--- a/quarkus/runtime/pom.xml
+++ b/quarkus/runtime/pom.xml
@@ -408,6 +408,10 @@
 
         <!-- Keycloak Dependencies-->
         <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>zjsonpatch</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.logging</groupId>
             <artifactId>commons-logging-jboss-logging</artifactId>
         </dependency>
--- a/rest/admin-api/pom.xml
+++ b/rest/admin-api/pom.xml
@@ -36,5 +36,9 @@
             <groupId>org.keycloak</groupId>
             <artifactId>keycloak-services</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>zjsonpatch</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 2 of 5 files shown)


vmuzikar (commented):
The changes look good, straightforward enough. My only concern would be the [`zjsonpatch` dependency](https://github.com/fabric8io/kubernetes-client/tree/main/zjsonpatch). It seems more like an internal utility lib for Fabric8 rather than something truly standalone (e.g. there doesn't seem to be any docs etc.). I would be worried if we could truly rely on it (in terms of breaking changes, or if it simply doesn't disappear with the next Fabric8 release). Or it's not that case?

shawkins:
> It seems more like an internal utility lib for Fabric8 rather than something truly standalone (e.g. there doesn't seem to be any docs etc.). I would be worried if we could truly rely on it (in terms of breaking changes, or if it simply doesn't disappear with the next Fabric8 release). Or it's not that case?

It's just a fork of the original zjsonpatch to strip it of additional dependencies and to always be aligned with our version of jackson https://github.com/fabric8io/kubernetes-client/pull/6276 

It has not been promoted for external usage, but it's highly unlikely to disappear.

The biggest downside to the usage of parsson is that it requires an additional serialization / deserialization from string. That is: object -> string (or char stream) -> apply patch -> string (or char stream) -> result object

The initial conversion from object to string will be handled by an ObjectMapper, which we can ultimately tweak if needed. However the JsonValue -> string conversion will be handled by a parsson JsonWriter which is not based upon Jackson and it's not clear how we would implement customizations - at the very least any Jackson annotations would not be used for this conversion step.

shawkins:
@vmuzikar @mabartos should be working now - our prototype doesn't persist the data yet, but it does return the modified entity, so the tests are just checking for that.

mabartos (approved):
@shawkins LGTM

shawkins:
@vmuzikar are you good with merging?

vmuzikar (approved):
@shawkins Sorry, sure, let's merge it.

PR #40905, opened 2025-07-03
fix: correcting to use the X-Forwarded-Proto header

### LINKED ISSUE #40903: Proxy detection needs tweaked for insecure context warning
The code is looking for the "X-Forwarded" header - it should look instead for any of the X-Forwarded-* headers.

_Originally posted by @shawkins in https://github.com/keycloak/keycloak/discussions/36094#discussioncomment-13617030_

shawkins:
It should suffice to just check for X-Forwarded-Proto

close: #40903


sschu:
@shawkins Why are you using the `X-Forwarded-Proto` header here to detect if a forwarded header is in place? I would have said `X-Forwarded-For` is the primary header and `X-Forwarded-Proto` might not even be set. Or am I missing something here?

shawkins:
Since this is the secure context check, either the Forwarded or X-Forwarded-Proto header will need to be set to change the protocol from http to https. This message is hopefully just a pointer to what could be wrong. hostname-debug or the startup log should contain a more detailed description.

sschu:
@shawkins Ah, I havent seen the context of this, sorry for the noise!

vmuzikar (approved):
LGTM, thanks.

PR #40906, opened 2025-07-03
fix: further refining logic and docs around import/export

### LINKED ISSUE #40874: Update code and documentation for import of a new realm
### Description

We can update the docs and the operator to indicate that importing new realms does not require the cluster to be shutdown, or for the cluster to be restarted after import has completed.

shawkins:
I believe this better reflects the current state of things.

As we've disucssed offline and on other issues, we're not directly mentioning memory issues either due to too many statements in a single transaction (https://github.com/agroal/agroal/pull/129) or https://github.com/keycloak/keycloak/issues/40875

closes: #40874

### DIFF
--- a/docs/guides/server/importExport.adoc
+++ b/docs/guides/server/importExport.adoc
@@ -19,7 +19,7 @@ The `import` and `export` commands are essentially server launches that exit bef
 
 It is recommended that all {project_name} nodes are stopped prior to using the `kc.[sh|bat] export` command. This ensures that the results will have no consistency issues with user or realm modifications during the export.
 
-It is required that all {project_name} nodes are stopped prior to performing an `kc.[sh|bat] import` command with the overwrite option. 
+It is required that all {project_name} nodes are stopped prior to performing an `kc.[sh|bat] import` command with the override option. 
 The command does not attach to the cache cluster, so realm deletion will not be handled correctly. You may also consider using the Admin API to delete realms that need to be overwritten prior to running the import.
 ====
 
--- a/docs/guides/server/importExport.adoc
+++ b/docs/guides/server/importExport.adoc
@@ -20,7 +20,7 @@ The `import` and `export` commands are essentially server launches that exit bef
 It is recommended that all {project_name} nodes are stopped prior to using the `kc.[sh|bat] export` command. This ensures that the results will have no consistency issues with user or realm modifications during the export.
 
 It is required that all {project_name} nodes are stopped prior to performing an `kc.[sh|bat] import` command with the override option. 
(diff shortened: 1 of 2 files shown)


docs/guides/server/importExport.adoc
```diff
@@ -8,17 +8,22 @@
...
-The default count of users per file and per transaction is fifty, but you may use a much larger value if desired.
+The default count of users per file is fifty, but you may use a much larger value if desired.
 
 The `import` and `export` commands are essentially server launches that exit before bringing up the full server. They are not currently designed to be run from the same machine as a running server instance, which may result in port or other conflicts.
 
 It is recommended that all {project_name} nodes are stopped prior to using the `kc.[sh|bat] export` command. This ensures that the results will have no consistency issues with user or realm modifications during the export.
+
+It is required that all {project_name} nodes are stopped prior to performing an `kc.[sh|bat] import` command with the override option. 
+The command does not attach to the cache cluster, so realm deletion will not be handled correctly. You may also consider using the Admin API to delete realms that need to be overwritten prior to running the import.
```
ahus1:
```suggestion
The command does not attach to the cache cluster, so overwriting a realm will lead to inconsistent caches in the cluster, which then would show and use inconsistent or outdated information. Instead of overwriting a realm with the import command, consider using the Admin API to delete realms that need to be overwritten prior to running the import.
```

operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportController.java
```diff
@@ -129,8 +129,7 @@ public void updateStatus(KeycloakRealmImportStatusBuilder status, KeycloakRealmI
                 status.addStartedMessage("Import Job started");
             } else if (oldStatus.getSucceeded() != null && oldStatus.getSucceeded() > 0) {
                 if (!lastReportedStatus.isDone()) {
-                    Log.info("Job finished performing a rolling restart of the deployment");
-                    rollingRestart(realmCR, client); // could be based upon a hash annotation on the deployment instead
+                    Log.info("Job finished");
```
ahus1, line 133:
```suggestion
                    // no need to restart Keycloak as we're only importing new realms and are not overwriting existing realms
                    Log.info("Job finished");
```

ahus1 (changes requested):
Thank you for this PR. See below for an attempt to be more specific on how overwriting a realm with a currently running cluster can lead to problems.

shawkins:
@ahus1 thank you for the review. Applied the changes.

ahus1 (approved):
Approving. As there is another team member from CND added, not merging yet. Let me know if you want to to be merged now.

shawkins:
> Let me know if you want to to be merged now.

Should be good based upon your review.

vmuzikar (approved):
LGTM, thanks.

PR #40922, opened 2025-07-04
Add Selective Consent for clients

oculos:
This PR introduces the concept of Selective Consent.

The idea is simple:

As of today, Consent requirement is on or off on clients. Either all users must consent to a client or all users won't have that requirement.

However, often Keycloak has different classes of users. A common pattern is the devision of internal and external users. The introduction of Organizations makes this pattern even more obvious.

The consequence of handling different types of users is that the grounds for processing personal data might be different depending on what type of user, or, rather, what type of data subject the user is. If the user is an employee, the grounds of processing personal data might not be based on consent. However, guest users must certainly consent for transferring of personal data to an application.

This PR aims to address the need of requiring consent for transferring of personal data only for some users: 

Screenshot 2025-07-04 at 09 28 33

The main idea is that it will be possible to require consent only for users with a given attribute and a given value.

When I first started implementing this, I wanted to add an own tab to handle this, and offering the possibility to configure multiple attributes. I also wanted to add the possibility of requiring consent based on organization. The problem I faced with that approach was the need of creating Set<String> for organizations and Map<String,String> for attributes. Actually, it was done, but working with them on the React code was painful. Besides, it was hard to find out the user membership to an organization based on the user object.

So I dropped that more complete approach in favor of something much simpler: since one can easily add attributes to an organization, just adding a simple attribute per client will probably be sufficient for most use cases.

If, however, there's the need of implementing this with more fine tuned possibilities, like adding organizations and "any or all" multiple attributes, we can discuss the viability of such implementation.


thomasdarimont:
Hello @oculos! That's an interesting idea! I have seen this issue myself quite regularly in practice when you combine 1st- and 3rd-party user bases. 

However, instead of implementing this right away, I recommend that you create a discussion first to elaborate on the problem and explore other potential solutions.

The attribute-based approach might work for some scenarios. However, I can imagine that other aspects should be taken into account to decide the need for consent handling:
- having a specific role (explicitly or implicitly assigned)
- being a member of a particular group/hierarchy
- having a specific IdP link 
- etc.

There are also other, more advanced use cases, which are not covered by this. 
EntraID, for example, allows for the configuration of consent handling per app to be done by the first user of a tenant (think "Organization") on behalf of all other users of that tenant. 

I think consent handling is a complex topic that might warrant a dedicated "consent policy" SPI to handle those scenarios accordingly.

PR #40927, opened 2025-07-04
More secure call of Facebook debug token

mposolda:
closes #40926

Unfortunately POST method does not work, but at least it is possible to send the client credentials in the HTTP header, which is a bit more secure than using GET parameter.


graziang (approved):
@mposolda, Good catch! Thanks!

PR #40936, opened 2025-07-04
Fix NPE when client is not set in context during token encoding

### LINKED ISSUE #40935: NPE thrown when encoding a token without having a client set in the session
### Before reporting an issue

### Area

core

### Describe the bug

Encoding a token can result in an NPE in case no client is set to the session.
In the DefaultTokenManager some functions check for that fact but not the type(TokenCategory) function. As a result, when calling encode(Token) a NPE is thrown.

### Version

26.2.5

### Regression

### Expected behavior

I would like to have the token encoded without having an NPE thrown.

### Actual behavior

The NPE is thrown.

### How to Reproduce?

```java
        AccessToken newToken = new AccessToken();
        newToken.id(KeycloakModelUtils.generateId());
        newToken.type(TokenUtil.TOKEN_TYPE_BEARER);
        ...
        ...
        newTokenString = session.tokens().encode(newToken);
``` 

### Anything else?

I will provide a PR to fix the issue.

mkrueger92:
This commit fixes an issue throwing an NPE when trying to encode a token without having a client set in the session context. In other places in this class (like getSignatureAlgorithm(String)) this is checked. But in the type(TokenCategory) it was forgotten to check.

Fixes #40935

### DIFF
--- a/services/src/main/java/org/keycloak/jose/jws/DefaultTokenManager.java
+++ b/services/src/main/java/org/keycloak/jose/jws/DefaultTokenManager.java
@@ -244,7 +244,7 @@ public class DefaultTokenManager implements TokenManager {
         switch (category) {
             case ACCESS:
                 ClientModel client = session.getContext().getClient();
-                return OIDCAdvancedConfigWrapper.fromClientModel(client).isUseRfc9068AccessTokenHeaderType()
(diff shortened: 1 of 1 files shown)


mposolda (approved):
@mkrueger92 Thanks

PR #40937, opened 2025-07-04
Allow configure encryption details for SAML clients

rmartinc:
Closes #40933

Draft PR to add configuration details to SAML assertion encryption. The main idea is that all the details (all the algorithms) can be configured. missing points:

* IMHO we should use AES_GCM_256 instead of AES_128 at least. Change other defaults? What happens to upgrades? The best approach would be configure very secure by default and allow to lower the configuration. But the problem is that right now we have to deal with existing clients. We can create an upgrading task to configure all clients like now (insert select over the attributes for example).
* Add a upgrading/release note accordingly (upgrading or just release if we don't change defaults).

The current draft does not change any default. It just allows to configured the options individually per client.


rmartinc:
One more thing, now I see that there is a `Signature and Encryption` section in the `General Settings` tab. I can move the UI to that part. They will appear only if the `Encrypt assertions` is enabled in the `Keys` tab.

jonkoops:
Sounds like a better place to move it to, go for it :)

rmartinc:
I have modified the default encryption values. Now by default the encryption uses aes_256_gcm, rsa-oaep, sha-256 and mgf1sha256 when nothing is configured. It's the migration task the one in charge to set the previous old configuration to existing clients to not use the new default values and maintain backwards compatibility. I have added a note in the migration guide but maybe it's better to add the note at the release notes now. Moving out of draft for the review.

mposolda:
@rmartinc Nice, Thanks for this PR!

Question: I've looked also at the backwards compatibility regarding the SAML Identity provider and SAML adapter as those also uses stuff from the saml-core. Could you confirm if those behave in the backwards compatible and that after upgrade to 26.4, people will not need to do any manual changes in their configuration to stay backwards compatible? It seems to me that this is fine already in the current PR (as SAML IDP and SAML Adapter are decrypting assertions and I do not see any updates in the decryption logic). Could you please confirm?

There is a conflict in `messages_en.properties`, but I guess you are waiting for the UI review from @jonkoops before fixing that?

rmartinc:
> Question: I've looked also at the backwards compatibility regarding the SAML Identity provider and SAML adapter as those also uses stuff from the saml-core. Could you confirm if those behave in the backwards compatible and that after upgrade to 26.4, people will not need to do any manual changes in their configuration to stay backwards compatible? It seems to me that this is fine already in the current PR (as SAML IDP and SAML Adapter are decrypting assertions and I do not see any updates in the decryption logic). Could you please confirm?

Yes, decryption is OK. When keycloak is used in the adapter or the identity provider acts as SP (client) and there there is no encryption (only signing). The only thing that is encrypted is the assertion and the only one that performs that encryption is the server.  So those parts never encrypt anything. They just decrypt and that part is done just reading the XML sent, so the decryption is OK because it just uses the XML returned and decrypts using it. The only thing you can configure in the decryption is if you force rsa1_5 or rsa-oaep as the encryption type, and that is not change with the PR (we are still using rsa-oaep as before).

I will update @jonkoops to rebase the changes. Those are just the new labels.

@mposolda @jonkoops Where do you prefer the note for this? Release notes or upgrading? I added to upgrade but I can move it to release notes if necessary.

mposolda:
> > Question: I've looked also at the backwards compatibility regarding the SAML Identity provider and SAML adapter as those also uses stuff from the saml-core. Could you confirm if those behave in the backwards compatible and that after upgrade to 26.4, people will not need to do any manual changes in their configuration to stay backwards compatible? It seems to me that this is fine already in the current PR (as SAML IDP and SAML Adapter are decrypting assertions and I do not see any updates in the decryption logic). Could you please confirm?
> 
> Yes, decryption is OK. When keycloak is used in the adapter or the identity provider acts as SP (client) and there there is no encryption (only signing). The only thing that is encrypted is the assertion and the only one that performs that encryption is the server. So those parts never encrypt anything. They just decrypt and that part is done just reading the XML sent, so the decryption is OK because it just uses the XML returned and decrypts using it. The only thing you can configure in the decryption is if you force rsa1_5 or rsa-oaep as the encryption type, and that is not change with the PR (we are still using rsa-oaep as before).
> 
That is what I though, Thanks!

> I will update @jonkoops to rebase the changes. Those are just the new labels.
> 
> @mposolda @jonkoops Where do you prefer the note for this? Release notes or upgrading? I added to upgrade but I can move it to release notes if necessary.

Not sure, maybe it is fine to have both? Upgrading guide makes sense as you did as it is slightly related to the upgrading, however maybe some small note in release notes is also ok? But whatever you prefer.

js/apps/admin-ui/src/clients/add/SamlEncryption.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const { watch, setValue } = useFormContext();
+  const keyEncryptionAlgorithm = watch(
+    convertAttributeNameToForm<FormFields>(
+      "attributes.saml.encryption.keyAlgorithm",
+    ),
+    "",
+  );
+
+  // remove optional fields if not displayed
+  if (keyEncryptionAlgorithm !== "http://www.w3.org/2009/xmlenc11#rsa-oaep") {
```
jonkoops: It looks like these kinds of URLs are repeated quite a lot, perhaps we can add them all as constants on top of the file and use those?

For example:
```ts

const ALGORITHM_RSA_OAEP = "http://www.w3.org/2009/xmlenc11#rsa-oaep";
const ALGORITHM_RSA_OAEP_MGF1P = "http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p";
```

js/apps/admin-ui/src/clients/add/SamlEncryption.tsx
```diff
@@ -0,0 +1,166 @@
...
+    setValue(
```
jonkoops: This sets the value on every render of the component, if you need to handle side effects they should be handled inside a [`useEffect()`](https://react.dev/reference/react/useEffect) that specifies its dependencies, so it will only be triggered if the selected algorithm changes.

For example:

```ts
useEffect(() => {
  if (keyEncryptionAlgorithm !== ALGORITHM_RSA_OAEP) {
    setValue(
      convertAttributeNameToForm<FormFields>("attributes.saml.encryption.maskGenerationFunction"),
     ""
    );
  }
}, [keyEncryptionAlgorithm]);
```

js/apps/admin-ui/src/clients/add/SamlSignature.tsx
```diff
@@ -96,6 +101,7 @@ export const SamlSignature = () => {
               value: name,
             }))}
           />
+          {samlEncryption?.toString() === "true" && <SamlEncryption />}
```
jonkoops: There should be no need to do optional chaining here, the default value is provided in the `watch()` call, so if the value is not set it should return `"false"`, there is also no way the value can be anything but a string, as that is the only data type supported by attributes (aside from sets of strings). You can likely get away with the following:

```suggestion
          {samlEncryption === "true" && <SamlEncryption />}
```

jonkoops (changes requested):
It would perhaps also be good to add some integration tests to the admin console to verify the form works as expected.

js/apps/admin-ui/src/clients/add/SamlEncryption.tsx
rmartinc: Done!

js/apps/admin-ui/src/clients/add/SamlEncryption.tsx
rmartinc: Done!

js/apps/admin-ui/src/clients/add/SamlSignature.tsx
rmartinc: Done!

rmartinc:
@jonkoops I have enhanced the test for encryption to check the values are set correctly when you enable / disable the encryption. Besides I have fixed another issue in `should check SAML capabilities`, it failed sometomes in my tests (it's just the order, it seems the select hides the switch if you execute in the previous order). Let's see the tests.

rmartinc:
Re-run the operator CI when you have time!

mposolda (dismissed):

jonkoops (dismissed):
LGTM!

jonkoops (approved):

mposolda (approved):

PR #40939, opened 2025-07-04
Change error to 400 for unknown user

### LINKED ISSUE #39079: AuthenticationFlowException when a user tries a password grant using a service account
### Before reporting an issue

### Area

authentication

### Describe the bug

When a user decides to try and use an existing service account's username (`service-account-client_id`) in a password grant against another client, it triggers an exception like:
```
2025-04-18 13:08:07,132 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-16) Uncaught server error: org.keycloak.authentication.AuthenticationFlowException
	at org.keycloak.authentication.AuthenticationProcessor.validateUser(AuthenticationProcessor.java:1200)
	at org.keycloak.authentication.AuthenticationProcessor.setAutheticatedUser(AuthenticationProcessor.java:294)
	at org.keycloak.authentication.AuthenticationProcessor$Result.setUser(AuthenticationProcessor.java:473)
	at org.keycloak.authentication.authenticators.directgrant.ValidateUsername.authenticate(ValidateUsername.java:99)
	at org.keycloak.authentication.DefaultAuthenticationFlow.processSingleFlowExecutionModel(DefaultAuthenticationFlow.java:465)
	at org.keycloak.authentication.DefaultAuthenticationFlow.processFlow(DefaultAuthenticationFlow.java:269)
	at org.keycloak.authentication.AuthenticationProcessor.authenticateOnly(AuthenticationProcessor.java:1095)
	at org.keycloak.protocol.oidc.grants.ResourceOwnerPasswordCredentialsGrantType.process(ResourceOwnerPasswordCredentialsGrantType.java:114)
	at org.keycloak.protocol.oidc.endpoints.TokenEndpoint$quarkusrestinvoker$processGrantRequest_6408e15340992839b66447750c221d9aaa837bd7.invoke(Unknown Source)
	at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:141)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:638)
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2675)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2654)
	at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1627)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1594)
```

### Version

25.0.6, also works on nightly

### Regression

### Expected behavior

The user gets some kind of HTTP 4xx with invalid_user_credentials or similar as an error message.

### Actual behavior

The user gets HTTP/1.1 500 Internal Server Error.

### How to Reproduce?

- Spin up a nightly instance of Keycloak.
- Create a `test` client and enable Service account roles.
- Send this HTTP call or similar:
```bash
curl -X POST "http://0.0.0.0:8080/realms/master/protocol/openid-connect/token" \                                             
-d "client_id=admin-cli" \
-d "client_secret=admin-cli" \
-d "grant_type=password" \
-d "username=service-account-test" \
-d "password=anything"  -v
```
- Get a 500 error code

### Anything else?

The main issue with this is that it's an easy way for someone who only knows a valid client_id with service accounts enabled, and any other client_id, a way to cause exceptions on demand. This can potentially trigger service availability alerts and whatnot.

keshavprashantdeshpande:
Closes #39079


keshavprashantdeshpande:
- It seems to me that authentication session is not removed. See other occurences below in the same method, which clears authentication session for some error scenarios. :  DONE

- It also seems to me that error event is not thrown. See all other errors handled inside ResourceOwnerPasswordCredentialsGrantType and how are they creating an error event. Will be also good the event in the test method you added. : DONE

- I think that for service-account user, the status, error and error_description returned should be exactly the same like for unknown user. This is to avoid the possibility for "user enumeration" . However there might be the difference in the error event as the error event is supposed to be consumed by administrator, so it can be useful if administrator knows that someone tried to authenticate service account : I am not sure if I follow completely what you said maybe can you elaborate a little bit more

@mposolda

mposolda:
@keshavprashantdeshpande Thanks for the updated PR!

For the last paragraph: It will be good to avoid "user enumeration" . This is kind of attack, which allows potential attacker to figure what are the valid usernames on the server or valid service-accounts on the server. In case that there are different error messages for those cases:
- Username not found
- Username found, but password is incorrect
- Username found, but it is a service account

Then attacker would know that he successfully guessed some valid username (or valid service-account username). Hence it is good to make sure that there are same error messages for those cases.

When looking at this, it seems to me that it can be better to avoid doing any changes in the `ResourceOwnerPasswordCredentialsGrant` class, but maybe handle this in the layer of `ValidateUsername` class? Perhaps for example somewhere here https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/directgrant/ValidateUsername.java#L81 , it might be possible to add something like this:
```
        if (user.getServiceAccountClientLink() != null) {
            AuthenticatorUtils.dummyHash(context);
           context.getEvent().detail(Details.REASON, "User is a service account");
            context.getEvent().error(Errors.INVALID_USER); (maybe this constant will need to be introduced to the `Errors` class)
            Response challengeResponse = errorResponse(Response.Status.UNAUTHORIZED.getStatusCode(), "invalid_grant", "Invalid user credentials");
            context.failure(AuthenticationFlowError.INVALID_USER, challengeResponse);
            return;
        }
```

Will this work for you?

stianst (changes requested):
Should have same response as if an invalid user or password is used:
```
status=401, error=invalid_grant, description='Invalid user credentials'
```

keshavprashantdeshpande:
@stianst @mposolda Changes done

mposolda (approved):
@keshavprashantdeshpande Thanks!

Issue #40890, opened 2025-07-03
Keycloak Operator 26.3.0 fails to update to 26.3.0


bo0ts:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

operator

### Describe the bug

When upgrading the Keycloak Operator to 26.3.0 on an existing Keycloak instance the Keycloak StatefulSet is not updated to the new version. The operator logs the following error:

```
2025-07-03 11:16:25,467 ERROR [io.jav.ope.pro.eve.EventProcessor] (ReconcilerExecutor-keycloakcontroller-96) Error during event processing ExecutionScope{ resource id: ResourceID{name='keycloak', namespace='keycloak'}, version: 1543606041}: io.javaoperatorsdk.operator.AggregatedOperatorException: Exception(s) during workflow execution. Details:
[...]

PR #40894, opened 2025-07-03
fix: detecting when we can set the serviceName

shawkins:
closes: #40890

PR #40943, opened 2025-07-05
fix: detecting when we can set the serviceName (#40894)

shawkins:
closes: #40890

Signed-off-by: Steve Hawkins <shawkins@redhat.com>
(cherry picked from commit 919838089ffefd5924ab4cbcf6bc17a2f2eb7ed7)


ahus1 (approved):
Approving based on parent PR's reviews.

PR #40964, opened 2025-07-07
Do not add steps if feature disabled in default flows

rmartinc:
Allow login if a step is disabled even the authenticator is not enabled by profile
Closes #40954

When webauthn and recovery-codes were added to the default forms we didn't think that the features could be disabled. The PR just adds two things:

1. The steps are just added if the features are enabled.
2. The `DefaultAuthenticationFlow.isConditionalAuthenticator` checks the `isEnabled` before loading the authenticator. This way even having the steps disabled the login is allowed. Test added for this point.


mposolda (approved):
@rmartinc Thanks

PR #40989, opened 2025-07-08
Remove workaround for handling Syslog counting framing

mabartos:
- Closes https://github.com/keycloak/keycloak/issues/40683

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/OptionBuilder.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/OptionBuilder.java
@@ -23,6 +23,7 @@ public class OptionBuilder<T> {
     private String description;
     private Optional<T> defaultValue;
     private List<String> expectedValues;
+    private boolean transformEnumValues;
     // Denotes whether a custom value can be provided among the expected values
     private boolean strictExpectedValues;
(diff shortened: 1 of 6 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/OptionBuilder.java
```diff
@@ -93,13 +94,24 @@ public OptionBuilder<T> expectedValues(List<String> expected) {
     }
 
     public OptionBuilder<T> expectedValues(Class<? extends Enum> expected) {
-        return expectedValues(Stream.of(expected.getEnumConstants()).map(Object::toString).collect(Collectors.toList()));
+        return expectedValues(Stream.of(expected.getEnumConstants())
+                .map(Object::toString)
+                .map(value -> transformEnumValues ? Option.transformEnumValue(value) : value)
```
shawkins: Is there a way to do this where there isn't a inter-dependency between expectedValues and transformEnumValues? If you set transformEnumValues after calling expectedValues for example you get different behavior.

mabartos: Nice point, updated.

shawkins (approved):
LGTM, thanks @mabartos

PR #40994, opened 2025-07-08
Improve handling of enum types for Keycloak options

### LINKED ISSUE #40993: Improve handling of enum types for Keycloak options
### Description

ATM, using the external Enum type as Keycloak option type is hard due to limitations of possible mapping of enum values to Strings. 

SmallRye Config uses a mapping mechanism to normalize these enum values, so f.e. enum value PROTOCOL_DEPENDENT is mapped to 'protocol-dependent' String. We should do the same for our codebase as it improves handling of our enum values as no toString() overrides are not necessary and we'll have a consistent way on how to deal with them. 

Moreover, we can leverage Quarkus-specific enum types and does not have to maintain them on our own. If Quarkus adds some other value to an enum, Keycloak users might use it and we don't need to provide the next enum value to our own enum type - decreasing maintenance of these values as it's usually only a 1:1 mapping of these values. During upgrades of Quarkus and changing some of these enum values, we would see failed test `HelpCommandDistTest` that shows us what has changed and we can react more quickly. 

### Discussion

_No response_

### Motivation

Improve maintanance
Decrease types duplications

### Details

_No response_

mabartos:
Closes #40993

WiP - needs some refinements and needs to decide on how it could work.

### DIFF
--- a/docs/maven-plugin/src/main/java/org/keycloak/guides/maven/Options.java
+++ b/docs/maven-plugin/src/main/java/org/keycloak/guides/maven/Options.java
@@ -46,7 +46,7 @@ public class Options {
                         m.isBuildTime(),
                         null,
                         m.getDescription(),
-                        m.getDefaultValue().map(Object::toString).orElse(null),
+                        m.getDefaultValueString().orElse(null),
                         m.getExpectedValues(),
                         m.isStrictExpectedValues(),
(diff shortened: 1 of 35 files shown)


mabartos:
It will probably be better to explicitly set if these enum values should be transformed. 

See https://github.com/keycloak/keycloak/pull/40989

PR #40999, opened 2025-07-08
Default jdbc-ping cluster setup for distributed caches fails in Oracle

### LINKED ISSUE #40784: Default jdbc-ping cluster setup for distributed caches fails in Oracle
### Before reporting an issue

### Area

infinispan

### Describe the bug

Running Keycloak using default jdbc-ping cluster setup for distributed caches fails in Oracle.

### Version

26.2.5

### Regression

### Expected behavior

Default jdbc-ping cluster setup works. No errors shown on logs.

### Actual behavior

Logs keep printing error reading table JDBC_PING.
`ERROR [org.keycloak.quarkus.runtime.storage.infinispan.jgroups.impl.KEYCLOAK_JDBC_PING2] (main) 0f387c618b43-48282: failed reading from the DB: java.sql.SQLSyntaxErrorException: ORA-00942: table or view does not exist`

### How to Reproduce?

Download drivers:
```
mkdir drivers
curl https://repo1.maven.org/maven2/com/oracle/database/jdbc/ojdbc17/23.6.0.24.10/ojdbc17-23.6.0.24.10.jar --output drivers/ojdbc17.jar
curl https://repo1.maven.org/maven2/com/oracle/database/nls/orai18n/23.6.0.24.10/orai18n-23.6.0.24.10.jar --output $SCRIPTPATH/../drivers/orai18n.jar
```
Create Dockerfile:
```
FROM quay.io/keycloak/keycloak:26.2.5
ENV KC_DB=oracle
COPY --chown=keycloak:keycloak --chmod=644 drivers/*.jar /opt/keycloak/providers/
WORKDIR /opt/keycloak
RUN keytool -genkeypair -storepass password -storetype PKCS12 -keyalg RSA -keysize 2048 -dname "CN=server" -alias server -ext "SAN:c=DNS:localhost,IP:127.0.0.1" -keystore conf/server.keystore
ENTRYPOINT ["/opt/keycloak/bin/kc.sh", "start"]
```
Build image:
```
docker build . --tag my-keycloak
```
Run Keycloak
```
docker run -it --rm -p 8080:8080 -p 8443:8443 \
-e KC_BOOTSTRAP_ADMIN_USERNAME=admin  \
-e KC_BOOTSTRAP_ADMIN_PASSWORD=123 \
-e KC_DB_URL_DATABASE=<db-name> \
-e KC_DB_SCHEMA=<db-schema> \
-e KC_DB_USERNAME=<db-username> \
-e KC_DB_PASSWORD=<db-password> \
-e KC_HOSTNAME_STRICT=false \
-e KC_DB_URL_HOST=<db-host> \
--name MyKeycloak \
my-keycloak
```

### Anything else?

Oracle server: AWS RDS Engine version 19.0.0.0.ru-2024-01.rur-2024-01.r1

ryanemerson:
Closes #40784
Closes #41105

In general we rely upon Hibernate/Liquibase and the DB drivers to handle Schema case sensitivity for us, for example: https://github.com/keycloak/keycloak/blob/87fbe2327047bc8b29cee7e500c917c4846828f5/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/storage/database/jpa/QuarkusJpaConnectionProviderFactory.java#L112

However, when creating the SQL query for JDBC_PING and `JpaRealmProvider#preRemove` we are explicitly quoting the name of the default schema making it case-sensitive.

When querying the default schema via Hibernate, the schema name is returned case insensitively in lower-case. Therefore, in the case of DBs that don't default to lower-case, e.g. Oracle, applying quotations around the name in the SQL will lead to a mismatch and the schema will not be found.


ahus1 (changes requested):
The old code made it case sensitive by adding the quote all the time, and now it is never adding them so it will break for schemata that contain special characters like `-`.

Can you figure how hibernate handles it that it adds the quotes conditionally?  

We don't test schemata well today, so it might be good to add one.

ryanemerson:
Good points @ahus1, my  initial attempt was overly simplistic. I have reworked `JpaUtils#getTableNameForNativeQuery` so that instead of querying hibernate for the configured default schema, we rely upon `JpaConnectionProviderFactory#getSchema` and apply quotes to the schema only when `GlobalConfiguration.PRESERVE_SCHEMA_CASE` has been set to true.

I have also created #41105 to also track the issue with `JpaRealmProvider#preRemove` so that any users who run into this will be able to see there's a fix.

For testing I have introduced two new tests:

- `CustomSchemaTest` - Which ensures that #41105 is resolved for all tested DBs
- `JdbcPingCustomSchemaTest` - Which ensures that JDBC_PING is correctly configured with Postgres and a cluster forms as expected. I have also manually executed this with `KC_TEST_DATABSE=oracle` to make sure it works as expected.

In order to be able to create a Postgres schema before Keycloak startup, I have added the ability for initScripts to be executed by the Database. If/when you're happy with the proposed changes, we should ask the testing team to review this part of the work.

model/jpa/src/main/java/org/keycloak/connections/jpa/util/JpaUtils.java
```diff
@@ -56,9 +58,10 @@ public class JpaUtils {
     private static final Logger logger = Logger.getLogger(JpaUtils.class);
 
     public static String getTableNameForNativeQuery(String tableName, EntityManager em) {
-        String schema = (String) em.getEntityManagerFactory().getProperties().get(HIBERNATE_DEFAULT_SCHEMA);
         final Dialect dialect = em.getEntityManagerFactory().unwrap(SessionFactoryImpl.class).getJdbcServices().getDialect();
-        return (schema==null) ? tableName : dialect.openQuote() + schema + dialect.closeQuote() + "." + tableName;
+        IdentifierHelper identifierHelper = em.getEntityManagerFactory().unwrap(SessionFactoryImpl.class).getJdbcServices().getJdbcEnvironment().getIdentifierHelper();
+        String schema = em.getEntityManagerFactory().unwrap(SessionFactoryImpl.class).getSessionFactoryOptions().getDefaultSchema();
+        return (schema==null) ? tableName : identifierHelper.toIdentifier(schema).render(dialect) + "." + tableName;
```
ahus1, line 64: The initial suggestion used the Liquibase setting, but I think this is mixing JPA and Liquibase items up more than they should be mixed up. And when quoting the schema, we should apply the same logic as Hibernate applies - only quote it when there are special characters. 

I suggest to keep the rest of the code untouched. I added a separate commit, but then I found that there were some conflicts in the intermediate commit, so I squashed your and my changes. 

Please have a look and let me know.

ahus1 (commented):
I would like to keep the original logic mostly unchanged and not mix JPA logic with Liquibase options. See below for a suggestion. 

Please review your test case if it still makes sense ... maybe we should should force a schema name that enforces quotes? Feel free to join a 1:1 on discussing the tests.

ryanemerson:
> Please review your test case if it still makes sense ... maybe we should should force a schema name that enforces quotes?

I have pushed a new commit that refactors the existing test I added, as well as adding a new one that specifically tests schemas with `-` in the name to make sure that we test with/without Liquibase `PRESERVE_SCHEMA_CASE` set.

model/jpa/src/main/java/org/keycloak/connections/jpa/util/JpaUtils.java
ryanemerson: Thanks for pushing the change. I agree that your approach is cleaner and it makes sense to utilise JPA/Hibernate here. I wasn't aware of `IdentifierHelper`, good to know :+1:

ahus1: I wasn't aware of IdentifierHelper before this PR, it was something that turned up in my research. I started searching for something I considered good enough when seeing the potential discrepancy between Hibernate handling and Liquibase properties.

ryanemerson:
The use of `KC_TEST_DATABASE_REUSE` is causing problems with subsequent steps in the store tests as we now have multiple DBs being kept alive and the old tests are sometimes connecting to the DB provisioned with a different schema name etc. I'm working on a fix, but I'll change the PR back to draft until this is resolved.

ahus1 (approved):
Thank you for this change!

ahus1:
@vaceksimon - FYI, Ryan added an option to initialize the database. Please have a look as a follow-up to see if additional cleanup/refactoring would be necessary as a follow-up. Thanks!

ahus1:
@ryanemerson - once this is merged, please provide a backport for 26.2 + 26.3. I suggest to limit the backport to the changes in `JpaUtils` only, and skip the changes to the testsuite.

vaceksimon:
Thanks for letting me know @ahus1 I would like to do a little cleanup to unify the database config style with the rest of the framework. See #41319

PR #41001, opened 2025-07-08
[OID4VCI] Add support for credential_response_encryption in credential request

### LINKED ISSUE #39310: [OID4VCI] Add support for credential_response_encryption at credential endpoint
### Description

The credential request has an optional `credential_response_encryption` parameter as described in https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-request. This ticket is about adding support for it

### LINKED ISSUE #41031: [OID4VCI] Determine algorithms for credential_response_encryption
### Description

Right now we are determining the algorithms for credential_response_encrpytion by directly adding these values into the realm-attributes.
I would prefer to get these algorithms by checking the available keys of the realm-configuration that do support the specific algorithms instead of doing it like this:

```java
@Override
    public void configureTestRealm(RealmRepresentation testRealm) {
        Map<String, String> attributes = Optional.ofNullable(testRealm.getAttributes()).orElseGet(HashMap::new);
        attributes.put("credential_response_encryption.alg_values_supported", "[\"RSA-OAEP\"]");
        attributes.put("credential_response_encryption.enc_values_supported", "[\"A256GCM\"]");
        attributes.put("credential_response_encryption.encryption_required", "true");
        attributes.put("batch_credential_issuance.batch_size", "10");
        attributes.put("signed_metadata", "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmb28iOiJiYXIifQ.XYZ123abc");
        testRealm.setAttributes(attributes);
        super.configureTestRealm(testRealm);
    }
```

The required value makes sense of course but the algorithms should be determined by the available keys.

Ogenbertrand:
This PR aims at supporting credential_response_encryption during the issuing of credentials request. If the Client requested an encrypted response by including the credential_response_encryption object in the request, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the parameters from the credential_response_encryption object. If the Credential Response is encrypted, the media type of the response MUST be set to application/jwt. If encryption was requested in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.

See: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-request

Closes: https://github.com/keycloak/keycloak/issues/39310
Closes: https://github.com/keycloak/keycloak/issues/41031


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -423,9 +458,137 @@ public Response requestCredential(
...
+            LOGGER.errorf("Serialization failed: %s", e.getMessage());
+            throw new WebApplicationException(
+                    Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+                            .entity(new ErrorResponse()
+                                    .setErrorDescription("Failed to serialize response"))
+                            .type(MediaType.APPLICATION_JSON)
+                            .build());
+        } catch (JWEException e) {
+            LOGGER.errorf("Encryption operation failed: %s", e.getMessage());
+            throw new WebApplicationException(
```
wistefan, line 522: Is this necessarily a 500? Couldn't it happen if the JWK contains a Key that is incompatible with its defined algorithm? I rather think a 400 is more likely here than a 500

wistefan (approved):
Minor comment, LGTM

mposolda:
@Ogenbertrand Could you please rebase on top of latest main? There was a bigger PR merged (by @Captain-P-Goldfish ) and it can be likely good to rebase on top of latest main (where that PR is included) to make sure tests are still OK

mposolda:
@Ogenbertrand Could you please check the failing test? It looks like it is caused by the changes in this PR.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -125,9 +108,66 @@ private String getSignedMetadata(KeycloakSession session) {
...
+                .map(KeyWrapper::getAlgorithm)
+                .filter(algorithm -> algorithm != null && !algorithm.isEmpty())
+                .filter(algorithm -> Arrays.asList(JWEConstants.RSA_OAEP, JWEConstants.RSA_OAEP_256).contains(algorithm))
+                .distinct()
+                .collect(Collectors.toList());
+
+        if (supportedEncryptionAlgorithms.isEmpty()) {
+            supportedEncryptionAlgorithms.add(JWEConstants.RSA_OAEP);
+            supportedEncryptionAlgorithms.add(JWEConstants.RSA_OAEP_256);
+        }
```
IngridPuppet: I feel this eventually defeats the purpose of the method. Shouldn't we at least check we have an RSA key before defaulting to these algorithms, even though it is very likely?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCIssuerWellKnownProviderTest.java
```diff
@@ -139,70 +152,128 @@ public void testMinimalJwtCredentialHardcodedTest()
...
+                            issuer.getSignedMetadata());
+                });
+    }
+
+    private static CredentialIssuer getCredentialIssuer(KeycloakSession session) {
+        RealmModel realm = session.getContext().getRealm();
+
+        realm.setAttribute("oid4vci.encryption.required", "true");
+        realm.setAttribute("batch_credential_issuance.batch_size", "10");
+        realm.setAttribute("signed_metadata", "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmb28iOiJiYXIifQ.XYZ123abc");
```
IngridPuppet: Reuse corresponding constants? I think at least one exists for "oid4vci.encryption.required": `ATTR_ENCRYPTION_REQUIRED`.

IngridPuppet (commented):
I added some minor comments. Could you please check them after addressing the issue with the pipeline?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCIssuerEndpointTest.java
```diff
@@ -374,7 +374,10 @@ public static Pair<JWK, PrivateKey> generateRsaJwkWithPrivateKey() throws NoSuch
         jwk.setModulus(modulus);
         jwk.setPublicExponent(exponent);
 
-        return Pair.of(jwk, privateKey);
+        Map<String, Object> result = new HashMap<>();
+        result.put("jwk", jwk);
+        result.put("privateKey", privateKey);
+        return result;
```
IngridPuppet, line 380: I wonder why you needed to change the signature of this method. Proper typing is now lost and, IMO, it is not idiomatically better.

IngridPuppet (approved):
LGTM

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -408,13 +448,141 @@ public Response requestCredential(
...
+     * @return true if the JWK is valid for encryption, false otherwise
+     */
+    private boolean isValidJwkForEncryption(JWK jwk, String expectedAlg) {
+        if (jwk == null) {
+            return false;
+        }
+        if (expectedAlg != null && !expectedAlg.equals(jwk.getAlgorithm())) {
+            return false;
+        }
+        return "enc".equals(jwk.getPublicKeyUse());
```
babisRoutis: A minor comment

To my understanding, `isValidJwkForEncryption` function requires that the given JWK has a `use` equal to `enc`.

Given that 
- According to RFC 7517 the `use` is optional, and
- `credential_response_encryption` contains a single `jwk` 

Perhaps, the above condition could be relaxed, to require that `use` is either `enc` or omitted.

babisRoutis: Dear @Ogenbertrand  
Thank you for considering my comment

services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialResponseEncryptionMetadata.java
```diff
@@ -0,0 +1,66 @@
...
+package org.keycloak.protocol.oid4vc.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+/**
+ * Represents the credential_response_encryption metadata for an OID4VCI Credential Issuer.
+ * {@see https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-11.2.2}
```
babisRoutis: The link points to draft 16, currently.
If the PR aims for this draft, then there is a new attribute `zip_values_supported`, that must be added. Otherwise, please neglect this comment.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -408,13 +448,143 @@ public Response requestCredential(
...
+        }
+        String publicKeyUse = jwk.getPublicKeyUse();
+        return publicKeyUse == null || "enc".equals(publicKeyUse);
+    }
+
+    private boolean isSupportedEncryption(CredentialResponseEncryptionMetadata metadata, String alg, String enc) {
+        if (metadata == null) {
+            return false;
+        }
+        return (metadata.getAlgValuesSupported() == null || metadata.getAlgValuesSupported().contains(alg)) &&
```
babisRoutis: To my understanding, the `CredentialResponseEncryptionMetadata` is either `null` or it contains a non-empty list of supported algorithms and encryption methods.

If I am not mistaken the method can return `true` in case the above lists are both `null`.
I have the impression, that this is not correct. 

VCI requires those two arrays to be present and if somehow any of those arrays is `null` (or empty) the outcome should be `false`, I guess.

babisRoutis: PS: In draft 15,16 specification is not explicit about non-empty arrays for `enc_values_supported`. 

But this is explicitly stated to the current editor's [draft](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-issuer-metadata-p)

babisRoutis (approved):
Dear @Ogenbertrand 
Thank you for considering my comments. 
From what I can tell, PR LGTM

mposolda (approved):
@Ogenbertrand @babisRoutis @wistefan @IngridPuppet Thanks everyone for the work and review of this PR!

I am approving as well based on your reviews. This could be merged from my point of view once CI is good.

PR #41005, opened 2025-07-08
fix: prevent quarkus from persisting logging runtime defaults

### LINKED ISSUE #40977: Loglevel recorded from build phase
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

When I set a log level during the build phase, and don't set a log level in the run phase later, the log level setting from the build phase is remembered. 

This is a surprising behavior as log level is supposed to be a runtime option with the default of INFO

### Version

main

### Regression

### Expected behavior

Always use INFO independent from the log level in the build phase

### Actual behavior

The log level from an (auto) build is remembered

### How to Reproduce?

* Start KC with auto-build enabled and set a log level like `--log-level="INFO,org.hibernate.SQL:debug"`
* Start KC again with no log level set -> SQL log messages will still appear

### Anything else?

_No response_

shawkins:
closes: #40977

@mabartos this may address the datasource issue as well.

There are two things happening here:
1. when there's a value to transform we assume that the default won't apply - this is not always true, especially with wildcards.
2. when the transform returns null, do we mean "unset", or use default? The http reload issue fixed the meaning as "unset", but that's also not generally true.

EDIT: none of the above is correct. The change here is to just prevent quarkus from persisting our runtime values.

### DIFF
--- a/quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
+++ b/quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
@@ -471,6 +471,9 @@ public class PicocliTest extends AbstractConfigurationTest {
         var value = nonRunningPicocli.config.getConfigValue("quarkus.log.category.\"org.keycloak\".level");
         assertEquals("quarkus.log.category.\"org.keycloak\".level", value.getName());
         assertEquals("WARN", value.getValue());
+        value = nonRunningPicocli.config.getConfigValue("quarkus.log.category.\"org.keycloak1\".level");
+        assertEquals("quarkus.log.category.\"org.keycloak1\".level", value.getName());
+        assertNull(value.getValue());
     }
 
     @Test
@@ -480,6 +483,9 @@ public class PicocliTest extends AbstractConfigurationTest {
         var value = nonRunningPicocli.config.getConfigValue("quarkus.log.category.\"org.keycloak\".level");
         assertEquals("quarkus.log.category.\"org.keycloak\".level", value.getName());
         assertEquals("WARN", value.getValue());
+        value = nonRunningPicocli.config.getConfigValue("quarkus.log.category.\"org.keycloak1\".level");
+        assertEquals("quarkus.log.category.\"org.keycloak1\".level", value.getName());
+        assertNull(value.getValue());
     }
 
     @Test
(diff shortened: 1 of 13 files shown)


shawkins:
I was over thinking this - we really just need to return the default for resolveCategoryLogLevelFromParentLogLevelOption

shawkins:
Updated yet again to just not allow quarkus to persist our runtime values.

shawkins:
Converting back to draft - there are several places in our code we are looking at raw values. I need to validate why we are doing that.

shawkins:
Summary of changes:
- tweak iterateNames - we only need to expose the name and the to when we see the from property.
- tweak propertymapper to not set the raw value if we're doing a build
  - remove our use of the rawValue - that is not needed because in all of these cases we want the resolved value
- refine the logging mapping logic to use a map rather than a list and to allow for multiple root entries (just in case the default slips through).

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
```diff
@@ -471,6 +478,9 @@ public void wildcardLevelForCategory() {
         var value = nonRunningPicocli.config.getConfigValue("quarkus.log.category.\"org.keycloak\".level");
         assertEquals("quarkus.log.category.\"org.keycloak\".level", value.getName());
         assertEquals("WARN", value.getValue());
+        value = nonRunningPicocli.config.getConfigValue("quarkus.log.category.\"org.keycloak1\".level");
+        assertEquals("quarkus.log.category.\"org.keycloak1\".level", value.getName());
+        assertNull(value.getValue());
```
mabartos, line 483: What exactly does this case check? It should pass(be true) even without these changes in this PR, right?

mabartos (approved):
@shawkins LGTM, just one question.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
shawkins: It would, but some of my intermediate attempts at addressing this relied upon proving the default value when not set. We don't want to do that because that overrides any quarkus level setting. I just wanted to make sure there was a unit test covering that.

mabartos:
@shawkins Maybe it'd be good to consider it for 26.3.x backport. WDYT?

shawkins:
> @shawkins Maybe it'd be good to consider it for 26.3.x backport. WDYT?

https://github.com/keycloak/keycloak/pull/41166

PR #41013, opened 2025-07-08
Adding a config to the UPDATE_EMAIL action to force users to verify email

### LINKED ISSUE #32569: Verify email when using UPDATE_EMAIL action without depending on realm wide setting
### Description

We would like to start using the UPDATE_EMAIL action to prevent users from updating email to an invalid email. 
However, we do not want to enable the realm wide **Verify Email** because that forces ALL existing users to verify their email.

### Discussion

_No response_

### Motivation

In an ideal scenario, we would like to:
 - Allow existing users that do not have a verified email to continue logging in with their existing username and password (this won't be allowed if we enable the Verify Email realm setting)
 - When a user wants to update its email, allow to do so ONLY after verifying the new address (this won't work unless we enablet the Verify Email realm setting).

We believe that both settings shouldn't be coupled together.

### Details

An naive implementation could be perhaps to evaluate the need to verify the new email before updating in a "per update request basis" using a conditional flag in the action token?

That way, you would have:
 - If you send UPDATE_EMAIL action and the realm has the realm.verifyEmail = true, then verification will be required (same as now)
 - If you send UPDATE_EMAIL action with an additional action parameter "verifyBeforeUpdate = true" , even if  the realm has the realm.verifyEmail = false, should ask for confirmation of the new address before updating.
 - If you send UPDATE_EMAIL action and the realm has the realm.verifyEmail = false and the token doens't have this conditional flag, it will update the email without confirmation.

pedroigor:
Closes #32569 

* Adds a `Force email verification` setting to the `UPDATE_EMAIL` required action to decide whether or not users should be asked to confirm their emails despite of the value set to the `Verify email` setting at the realm level.
* Make sure email verification is enforced through the `UPDATE_EMAIL` required action when updating the email during   `UPDATE_PROFILE` in case the `Force email verification` setting is enabled in the `UPDATE_EMAIL` required action. This only happens if `Verify email` is not set at the realm level.

### DIFF
--- a/docs/documentation/server_admin/topics/login-settings/update-email-workflow.adoc
+++ b/docs/documentation/server_admin/topics/login-settings/update-email-workflow.adoc
@@ -2,11 +2,39 @@
 
 == Update Email Workflow (UpdateEmail)
 
-With this workflow, users will have to use an UPDATE_EMAIL action to change their own email address.
+With this workflow, users will have to use an `UPDATE_EMAIL` action to change their own email address.
 
-The action is associated with a single email input form. If the realm has email verification disabled, this action will allow to update the email without verification. If the realm has email verification enabled, the action will send an email update action token to the new email address without changing the account email. Only the action token triggering will complete the email update.
+This action provides a more secure and consistent flow to update user emails because they will be forced to re-authenticate
+as well as verify their emails before any update to their account.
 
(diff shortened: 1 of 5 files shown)


mposolda:
@pedroigor Rebased the PR on top of latest main to fix the FIPS error, which was unrelated to your PR.

martin-kanis (approved):

ahus1 (approved):

PR #41026, opened 2025-07-09
Wildcard mappers should be implicitly handled and value propagated

### LINKED ISSUE #41641: Cannot use `dev-file` for additional datasources
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

When we specify `dev-file` for the additional datasources, we're not able to use it due to a duplicated path. 

```
2025-08-04 14:05:16,264 WARN  [org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator] (JPA Startup Thread: keycloak-default) HHH000342: Could not obtain connection to query JDBC database metadata: org.hibernate.exception.GenericJDBCException: unable to obtain isolated JDBC connection [Error while creating file "/home/mabartos/Documents/RH/keycloak/quarkus/tests/integration/target/kc-tests/keycloak-999.0.0-SNAPSHOT/data/h2 (a file with this name already exists)" [90062-230]] [n/a]
	at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:39)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:115)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:101)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.doTheWork(JtaIsolationDelegate.java:203)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.lambda$delegateWork$3(JtaIsolationDelegate.java:88)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.doInSuspendedTransaction(JtaIsolationDelegate.java:122)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.delegateWork(JtaIsolationDelegate.java:85)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironmentUsingJdbcMetadata(JdbcEnvironmentInitiator.java:355)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironment(JdbcEnvironmentInitiator.java:147)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:125)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:80)
	at org.hibernate.boot.registry.internal.StandardServiceRegistryImpl.initiateService(StandardServiceRegistryImpl.java:133)
	at org.hibernate.service.internal.AbstractServiceRegistryImpl.createService(AbstractServiceRegistryImpl.java:262)
	at org.hibernate.service.internal.AbstractServiceRegistryImpl.initializeService(AbstractServiceRegistryImpl.java:237)
	at org.hibernate.service.internal.AbstractServiceRegistryImpl.getService(AbstractServiceRegistryImpl.java:214)
	at org.hibernate.service.ServiceRegistry.requireService(ServiceRegistry.java:66)
	at org.hibernate.engine.jdbc.internal.JdbcServicesImpl.configure(JdbcServicesImpl.java:50)
	at org.hibernate.boot.registry.internal.StandardServiceRegistryImpl.configureService(StandardServiceRegistryImpl.java:139)
	at org.hibernate.service.internal.AbstractServiceRegistryImpl.initializeService(AbstractServiceRegistryImpl.java:246)
	at org.hibernate.service.internal.AbstractServiceRegistryImpl.getService(AbstractServiceRegistryImpl.java:214)
	at org.hibernate.service.ServiceRegistry.requireService(ServiceRegistry.java:66)
	at org.hibernate.boot.internal.SessionFactoryOptionsBuilder.<init>(SessionFactoryOptionsBuilder.java:278)
	at io.quarkus.hibernate.orm.runtime.recording.PrevalidatedQuarkusMetadata.buildSessionFactoryOptionsBuilder(PrevalidatedQuarkusMetadata.java:70)
	at io.quarkus.hibernate.orm.runtime.boot.FastBootEntityManagerFactoryBuilder.build(FastBootEntityManagerFactoryBuilder.java:89)
	at io.quarkus.hibernate.orm.runtime.FastBootHibernatePersistenceProvider.createEntityManagerFactory(FastBootHibernatePersistenceProvider.java:74)
	at jakarta.persistence.Persistence.createEntityManagerFactory(Persistence.java:90)
	at jakarta.persistence.Persistence.createEntityManagerFactory(Persistence.java:66)
	at io.quarkus.hibernate.orm.runtime.JPAConfig$LazyPersistenceUnit.get(JPAConfig.java:165)
	at io.quarkus.hibernate.orm.runtime.JPAConfig$1.run(JPAConfig.java:61)
Caused by: org.h2.jdbc.JdbcSQLNonTransientException: Error while creating file "/home/mabartos/Documents/RH/keycloak/quarkus/tests/integration/target/kc-tests/keycloak-999.0.0-SNAPSHOT/data/h2 (a file with this name already exists)" [90062-230]
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:566)
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:489)
	at org.h2.message.DbException.get(DbException.java:223)
	at org.h2.message.DbException.get(DbException.java:199)
	at org.h2.store.fs.disk.FilePathDisk.createDirectory(FilePathDisk.java:374)
	at org.h2.store.fs.FileUtils.createDirectory(FileUtils.java:85)
	at org.h2.store.fs.FileUtils.createDirectories(FileUtils.java:382)
	at org.h2.mvstore.db.Store.<init>(Store.java:106)
	at org.h2.engine.Database.<init>(Database.java:326)
	at org.h2.engine.Engine.openSession(Engine.java:92)
	at org.h2.engine.Engine.openSession(Engine.java:222)
	at org.h2.engine.Engine.createSession(Engine.java:201)
	at org.h2.engine.SessionRemote.connectEmbeddedOrServer(SessionRemote.java:344)
	at org.h2.jdbc.JdbcConnection.<init>(JdbcConnection.java:124)
	at org.h2.Driver.connect(Driver.java:59)
	at io.agroal.pool.ConnectionFactory.createConnection(ConnectionFactory.java:225)
	at io.agroal.pool.ConnectionPool$CreateConnectionTask.call(ConnectionPool.java:585)
	at io.agroal.pool.ConnectionPool$CreateConnectionTask.call(ConnectionPool.java:566)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at io.agroal.pool.util.PriorityScheduledExecutor.beforeExecute(PriorityScheduledExecutor.java:75)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	... 1 more

```

### Version

main

### Regression

### Expected behavior

No error

### Actual behavior

Shown error

### How to Reproduce?

Specify multiple datasources with `dev-file` DB kind

### Anything else?

_No response_

mabartos:
- Closes #40997

----------------------
- Wrapped Datasource logic ([DatabaseOptions.Datasources](https://github.com/keycloak/keycloak/pull/41026/files#diff-922effaaab1831700be05842474f2d2a14f07e5faeecf3ee285a9881e4de52adR115), [DatabasePropertyMappers.Datasources](https://github.com/keycloak/keycloak/pull/41026/files#diff-3b84b4432d8b3bde7601af78ae093ed420569e5d440b830bce84302dea0a831cR184) and [PropertyMappers.WildcardMappersConfig](https://github.com/keycloak/keycloak/pull/41026/files#diff-a1bb1ef613469cf669070d52be8a88079005dcaed860e8d11d4231b5e2f530f7R409))
- [Automatic mapping](https://github.com/keycloak/keycloak/pull/41026/files#diff-6ebe7f8db3e4dea9dbefd50eed883b8aac997e6c01127e283631e92e9d022835R545) of DB dialect and schema to Hibernate properties
- [Advertise all connected options to Quarkus](https://github.com/keycloak/keycloak/pull/41026/files#diff-bd3f5749f7a9d562e7e2efef70bad30ab4f3d77d5a63c566044cdb5810504767R113)
- Solved the issue with the `persistence.xml` files and it is possible to [rewrite quickstarts](https://github.com/mabartos/keycloak-quickstarts/commit/4c4b3043da9fd48772ad3a2c4c89d9cfb1086565) with this

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
@@ -112,99 +112,104 @@ public class DatabaseOptions {
             .hidden()
             .build();
 
-    /**
-     * Options that have their sibling for a named datasource
-     * Example: for `db-dialect`, `db-dialect-<datasource>` is created
-     */
-    public static final List<Option<?>> OPTIONS_DATASOURCES = List.of(
-            DB_DIALECT,
-            DB_DRIVER,
-            DB,
-            DB_URL,
-            DB_URL_HOST,
-            DB_URL_DATABASE,
-            DB_URL_PORT,
-            DB_URL_PROPERTIES,
-            DB_USERNAME,
-            DB_PASSWORD,
-            DB_SCHEMA,
-            DB_POOL_INITIAL_SIZE,
-            DB_POOL_MIN_SIZE,
-            DB_POOL_MAX_SIZE,
-            DB_SQL_JPA_DEBUG,
-            DB_SQL_LOG_SLOW_QUERIES
-            );
+    public static final class Datasources {
+        /**
+         * Options that have their sibling for a named datasource
+         * Example: for `db-dialect`, `db-dialect-<datasource>` is created
+         */
+        public static final List<Option<?>> OPTIONS_DATASOURCES = List.of(
+                DB_DIALECT,
+                DB_DRIVER,
+                DB,
+                DB_URL,
+                DB_URL_HOST,
+                DB_URL_DATABASE,
+                DB_URL_PORT,
+                DB_URL_PROPERTIES,
+                DB_USERNAME,
+                DB_PASSWORD,
+                DB_SCHEMA,
+                DB_POOL_INITIAL_SIZE,
+                DB_POOL_MIN_SIZE,
+                DB_POOL_MAX_SIZE,
+                DB_SQL_JPA_DEBUG,
+                DB_SQL_LOG_SLOW_QUERIES
+        );
 
-    /**
-     * In order to avoid ambiguity, we need to have unique option names for wildcard options.
-     * This map controls overriding option name to be unique for wildcard option.
-     */
-    private static final Map<String, String> DATASOURCES_OVERRIDES_SUFFIX = Map.of(
-            DatabaseOptions.DB.getKey(), "-kind", // db-kind
-            DatabaseOptions.DB_URL.getKey(), "-full"  // db-url-full
-    );
+        /**
+         * In order to avoid ambiguity, we need to have unique option names for wildcard options.
+         * This map controls overriding option name to be unique for wildcard option.
+         */
+        private static final Map<String, String> DATASOURCES_OVERRIDES_SUFFIX = Map.of(
+                DatabaseOptions.DB.getKey(), "-kind", // db-kind
+                DatabaseOptions.DB_URL.getKey(), "-full"  // db-url-full
+        );
 
-    /**
-     * You can override some {@link OptionBuilder} methods for additional datasources in this map
-     */
-    private static final Map<Option<?>, Consumer<OptionBuilder<?>>> DATASOURCES_OVERRIDES_OPTIONS = Map.of(
-            DatabaseOptions.DB, builder -> builder.defaultValue(Optional.empty()) // no default value for DB kind for datasources
-    );
+        /**
+         * You can override some {@link OptionBuilder} methods for additional datasources in this map
+         */
+        private static final Map<Option<?>, Consumer<OptionBuilder<?>>> DATASOURCES_OVERRIDES_OPTIONS = Map.of(
+                DatabaseOptions.DB, builder -> builder
+                        .defaultValue(Optional.empty()) // no default value for DB kind for datasources
+                        .connectedOptions(
+                                getDatasourceOption(DatabaseOptions.DB_URL).orElseThrow(),
+                                TransactionOptions.TRANSACTION_XA_ENABLED_DATASOURCE
+                        )
+        );
 
-    private static final Map<String, Option<?>> cachedDatasourceOptions = new HashMap<>();
+        private static final Map<String, Option<?>> cachedDatasourceOptions = new HashMap<>();
 
-
-    /**
-     * Get datasource option containing named datasource mapped to parent DB options.
-     * <p>
-     * We map DB options to named datasource options like:
-     * <ul>
-     *     <li>{@code db-url-host --> db-url-host-<datasource>}</li>
-     *     <li>{@code db-username --> db-username-<datasource>}</li>
-     *     <li>{@code db --> db-kind-<datasource>}</li>
-     * </ul>
-     */
-    @SuppressWarnings("unchecked")
-    public static <T> Optional<Option<T>> getDatasourceOption(Option<T> parentOption) {
-        if (!OPTIONS_DATASOURCES.contains(parentOption)) {
-            return Optional.empty();
-        }
-
-        var key = getKeyForDatasource(parentOption);
-        if (key.isEmpty()) {
-            return Optional.empty();
-        }
-
-        // check if we already created the same option and return it from the cache
-        Option<?> option = cachedDatasourceOptions.get(key.get());
-
-        if (option == null) {
-            var builder = parentOption.toBuilder()
-                    .key(key.get())
-                    .category(OptionCategory.DATABASE_DATASOURCES);
-
-            if (!parentOption.isHidden()) {
-                builder.description("Used for named <datasource>. " + parentOption.getDescription());
+        /**
+         * Get datasource option containing named datasource mapped to parent DB options.
+         * <p>
+         * We map DB options to named datasource options like:
+         * <ul>
+         *     <li>{@code db-url-host --> db-url-host-<datasource>}</li>
+         *     <li>{@code db-username --> db-username-<datasource>}</li>
+         *     <li>{@code db --> db-kind-<datasource>}</li>
+         * </ul>
+         */
+        @SuppressWarnings("unchecked")
+        public static <T> Optional<Option<T>> getDatasourceOption(Option<T> parentOption) {
+            if (!OPTIONS_DATASOURCES.contains(parentOption)) {
+                return Optional.empty();
             }
 
-            // override some settings for options
-            var override = DATASOURCES_OVERRIDES_OPTIONS.get(parentOption);
-            if (override != null) {
-                override.accept(builder);
+            var key = getKeyForDatasource(parentOption);
+            if (key.isEmpty()) {
+                return Optional.empty();
             }
 
-            option = builder.build();
-            cachedDatasourceOptions.put(key.get(), option);
-        }
-        return Optional.of((Option<T>) option);
-    }
+            // check if we already created the same option and return it from the cache
+            Option<?> option = cachedDatasourceOptions.get(key.get());
 
-    /**
-     * Get mapped datasource key based on DB option {@param option}
-     */
-    public static Optional<String> getKeyForDatasource(Option<?> option) {
-        return getKeyForDatasource(option.getKey());
-    }
+            if (option == null) {
+                var builder = parentOption.toBuilder()
+                        .key(key.get())
+                        .category(OptionCategory.DATABASE_DATASOURCES);
+
+                if (!parentOption.isHidden()) {
+                    builder.description("Used for named <datasource>. " + parentOption.getDescription());
+                }
+
+                // override some settings for options
+                var override = DATASOURCES_OVERRIDES_OPTIONS.get(parentOption);
+                if (override != null) {
+                    override.accept(builder);
+                }
+
+                option = builder.build();
+                cachedDatasourceOptions.put(key.get(), option);
+            }
+            return Optional.of((Option<T>) option);
+        }
+
+        /**
+         * Get mapped datasource key based on DB option {@param option}
+         */
+        public static Optional<String> getKeyForDatasource(Option<?> option) {
+            return getKeyForDatasource(option.getKey());
+        }
 
     /**
      * Get mapped datasource key based on DB option {@param option}
@@ -216,17 +221,18 @@ public class DatabaseOptions {
                 .map(key -> key.concat("-<datasource>"));
     }
 
-    /**
-     * Returns datasource option based on DB option {@code option} with actual wildcard value.
-     * It replaces the {@code <datasource>} with actual value in {@code namedProperty}.
-     * <p>
-     * f.e. Consider {@code option}={@link DatabaseOptions#DB_DRIVER}, and {@code namedProperty}=my-store.
-     * <p>
-     * Result: {@code db-driver-my-store}
-     */
-    public static Optional<String> getNamedKey(Option<?> option, String namedProperty) {
-        return getKeyForDatasource(option)
-                .map(key -> key.substring(0, key.indexOf("<")))
-                .map(key -> key.concat(namedProperty));
+        /**
+         * Returns datasource option based on DB option {@code option} with actual wildcard value.
+         * It replaces the {@code <datasource>} with actual value in {@code namedProperty}.
+         * <p>
+         * f.e. Consider {@code option}={@link DatabaseOptions#DB_DRIVER}, and {@code namedProperty}=my-store.
+         * <p>
+         * Result: {@code db-driver-my-store}
+         */
+        public static Optional<String> getNamedKey(Option<?> option, String namedProperty) {
+            return getKeyForDatasource(option)
+                    .map(key -> key.substring(0, key.indexOf("<")))
+                    .map(key -> key.concat(namedProperty));
+        }
     }
 }
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
@@ -98,7 +98,7 @@ public class PropertyMappingInterceptor implements ConfigSourceInterceptor {
         boolean filterRuntime = isRebuild() || Environment.isRebuildCheck();
 
         var baseStream = StreamSupport.stream(iterable.spliterator(), false).flatMap(name -> {
-            PropertyMapper<?> mapper = PropertyMappers.getMapper(name);
+            final PropertyMapper<?> mapper = PropertyMappers.getMapper(name);
 
             if (mapper == null) {
                 return Stream.of(name);
@@ -106,6 +106,14 @@ public class PropertyMappingInterceptor implements ConfigSourceInterceptor {
             if (filterRuntime && mapper.getCategory() == OptionCategory.CONFIG) {
                 return Stream.of(); // advertising the keystore type causes the keystore to be used early
             }
+
+            final PropertyMapper<?> mappedMapper = mapper.forKey(name);
+
+            // only include additional mappings if we're on the from side of the mapping as the mapping may not be bi-directional
+            if (!name.equals(mappedMapper.getFrom())) {
+                return Stream.of(name);
+            }
+
             allMappers.remove(mapper);
 
             if (mapper.hasWildcard()) {
@@ -119,9 +127,7 @@ public class PropertyMappingInterceptor implements ConfigSourceInterceptor {
 
                     return Stream.concat(toDistinctStream(name, wildcardMapper.getTo(wildcardValue)), connectedTo);
                 }
-            } else if (name.equals(mapper.getFrom())) {
-                // include additional mappings if we're on the from side of the mapping as the mapping may not be bi-directional
-                // -----------------------------------------
+            } else {
                 // this is not a wildcard value, but may map to wildcards
                 // the current example is something like log-level=wildcardCat1:level,wildcardCat2:level
                 var wildCard = PropertyMappers.getWildcardMappedFrom(mapper.getOption());
@@ -133,14 +139,9 @@ public class PropertyMappingInterceptor implements ConfigSourceInterceptor {
                 }
             }
 
-            mapper = mapper.forKey(name);
-
-            if (name.equals(mapper.getFrom())) {
-                // there is a corner case here: -1 for the reload period has no 'to' value.
-                // if that becomes an issue we could use more metadata to perform a full mapping
-                return toDistinctStream(name, mapper.getTo());
-            }
-            return Stream.of(name);
+            // there is a corner case here: -1 for the reload period has no 'to' value.
+            // if that becomes an issue we could use more metadata to perform a full mapping
+            return toDistinctStream(name, mappedMapper.getTo());
         });
 
         // include anything remaining that has a default value
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
@@ -17,9 +17,11 @@
 package org.keycloak.quarkus.runtime.configuration;
 
 import static org.keycloak.quarkus.runtime.Environment.isRebuild;
+import static org.keycloak.quarkus.runtime.configuration.MicroProfileConfigProvider.NS_KEYCLOAK_PREFIX;
 
 import java.util.Iterator;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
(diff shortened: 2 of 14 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
```diff
@@ -123,11 +135,11 @@ public Iterator<String> iterateNames(ConfigSourceInterceptorContext context) {
 
         // include anything remaining that has a default value
         var defaultStream = allMappers.stream()
-                .filter(m -> !m.getDefaultValue().isEmpty() && !m.hasWildcard()
+                .filter(m -> m.getDefaultValue().isPresent() && !m.hasWildcard()
                         && m.getCategory() != OptionCategory.CONFIG) // advertising the keystore type causes the keystore to be used early
                 .flatMap(m -> toDistinctStream(m.getTo()));
 
-        return IteratorUtils.chainedIterator(baseStream.iterator(), defaultStream.iterator());
+        return IteratorUtils.chainedIterator(baseStream.distinct().iterator(), defaultStream.iterator());
```
shawkins: It should be possible to use allMappers to do this inline if we want to avoid this extra iteration.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
```diff
@@ -102,7 +104,17 @@ public Iterator<String> iterateNames(ConfigSourceInterceptorContext context) {
             }
             allMappers.remove(mapper);
 
-            if (!mapper.hasWildcard()) {
+            if (mapper.hasWildcard()) {
```
shawkins, line 119: It should be possible to encapsulate this logic if instead of mapper.getTo() we used something like mapper.getConnectedTo()

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
```diff
@@ -203,6 +203,14 @@ public static WildcardPropertyMapper<?> getWildcardMappedFrom(Option<?> from) {
         return MAPPERS.wildcardMapFrom.get(from.getKey());
     }
 
+    /**
+     * Return list of wildcard mappers with the same {@code <wildcardName>} wildcard name - called connected wildcard mappers
+     * F.e. for {@link org.keycloak.config.DatabaseOptions#DB_ACTIVE_DATASOURCE}, return all mappers containing the {@code <datasource>} wildcard name in the option name
+     */
+    public static List<WildcardPropertyMapper<?>> getConnectedWildcardMappers(String wildcardName) {
```
shawkins: Is it safe to always advertise username and password as set, even if they have no value?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
mabartos: @shawkins Thanks for the review! 

If you have in mind a better approach, feel free to send a PR to my fork, or even create a new one with better changes - it's welcomed :)

mabartos: Improved handling of the connected wildcard mappers: https://github.com/keycloak/keycloak/pull/41026/commits/56c2c247365f933cda158249027ea63c2b7f6aa5

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
mabartos: @shawkins Do you have anything in mind on how can I do this? What extra iteration? Thanks

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
mabartos: @shawkins Does it need to be a problem? Do you see any risk that it could introduce? When Keycloak or Quarkus deal with these properties, the return value (`configValue.getValue()`) is `null` anyway, right? 

Do you suggest to always fetch values of these properties and if they are not set, not propagating them to Quarkus?

mposolda:
@mabartos Rebased the PR on top of latest main to fix the FIPS error, which was unrelated to your PR.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
shawkins: @mabartos after looking into this issue more, it does seem to hinge just on the url because that is tied to the notion of being activated or not. A blank URL or a config that doesn't advertise a property name will inactivate a datasource. Can we first do #41087, then just have a simple check in the PropertyMappingInteceptor for wildcard db-kinds, and have them also expose the URL?

I'm also not sure why this doesn't apply to the default datasource - we typically don't expose a URL property name, but it is not deactivated. So this may be a quarkus bug of some kind.

mabartos: @shawkins 

> after looking into this issue more, it does seem to hinge just on the url because that is tied to the notion of being activated or not.

Yes, the JDBC URL is required for an additional named datasource. 

> Can we first do https://github.com/keycloak/keycloak/pull/41087, then just have a simple check in the PropertyMappingInteceptor for wildcard db-kinds, and have them also expose the URL?

Yes, we should at least require the `db-kind` to know what to advertise to Quarkus. Do you mean advertise only these two (DB kind, JDBC URL)? I think we need to have the mechanism of the connected mappers and advertise them all(if they can have a value) as if these other properties/mappers are not propagated to Quarkus, we can't set any default value, nor do the mapFrom mapping for these properties. Is that correct?

> I'm also not sure why this doesn't apply to the default datasource - we typically don't expose a URL property name, but it is not deactivated. So this may be a quarkus bug of some kind.

I tried to iterate all over the props we advertise via the `PropertyMappingInterceptor`, and you're right that we don't advertise all properties for which we have mappers defined (like the JDBC URL). In that case, I'm a little bit struggling to find out the exact purpose of the `PropertyMappingInterceptor#iterateNames` and what properties should be included there. I thought that we need to provide all properties that our interceptor(`PropertyMappingInterceptor`) takes care of. The interceptors chain is executed anyway as we also handle, in that interceptor, properties that their name is not advertised via the `iterateNames`, right?

shawkins: > Is that correct?

Yes there would still be a mechanism, but it would be limited to just a special case. Either in PropertyMappingInterceptor or via a method to encapsultate like getConnectedTo, we would just check for db-kind and also expose the quarkus url property.

> and you're right that we don't advertise all properties for which we have mappers defined (like the JDBC URL).

Which is why this could also be a quarkus bug.

>  In that case, I'm a little bit struggling to find out the exact purpose of the PropertyMappingInterceptor#iterateNames and what properties should be included there.

Keycloak and quarkus use the property names iterator in several places. For us it's usually under the assumption that it will allow us to walk over everything that the user has set. 

Quarkus needs to see properties to trigger mapping logic - it seems to normally take just a single property to trigger the mapping logic. That is the quarkus logic will scan for anything with the prefix of quarkus.datasource to know whether datasources should be mapped - at which point it will ask the config individually for all values. For an additional datasource, it's not working as expected.

The smallrye docs are clear that the property names iterator may not be complete nor stable - the config logic can provide additional mappings or change what is known over time. So they really should not be making an assumption about url via property names if there is already a kind present.

> The interceptors chain is executed anyway as we also handle, in that interceptor, properties that their name is not advertised via the iterateNames, right?

Generally each interceptor contributes the property names it knows about - which is why we are providing both the map from and to, as well as defaults in our interceptor. The other interceptors are doing the same to contribute anything else - or even remap, or filter.

shawkins: In testing a little bit more I see that the presense of any other option, such as pool max size, means that the url does not need to be advertised. In practice this may only be a problem for dev-mem (as that needs no options beyond kind). Or did you see this problem in another case?

mabartos: @shawkins Thanks for the additional context! I'll check it into details later.

> In testing a little bit more I see that the presense of any other option, such as pool max size, means that the url does not need to be advertised. In practice this may only be a problem for dev-mem (as that needs no options beyond kind). Or did you see this problem in another case?

In the current state of this PR, at least one DB property for additional datasource needs to be specified in order to know what datasource name is used. Then all these connectedTo mappers/properties are also advertised to Quarkus with the name of the datasource. 

It means, if you specify `--db-pool-max-size-my-store`, all properties like:
`quarkus.datasource."my-store".jdbc.url`, `quarkus.datasource."my-store".jdbc.initial-size`, `quarkus.datasource."my-store".jdbc.max-size`,... are advertised to Quarkus. 

However, it'd be better to require a build-time option with the wildcard, and the most suitable is requiring the DB kind as touched here: https://github.com/keycloak/keycloak/pull/41087. Advanced: might be good to not requiring any option with the wildcard as we know the datasource names after parsing persistence.xml files or using props for persistence units - but I think it's ok to require at least the DB kind in our config sources to avoid any workarounds/hacks.

I'll look into it more on Monday. Thanks!

shawkins: > It means, if you specify --db-pool-max-size-my-store, all properties like:
quarkus.datasource."my-store".jdbc.url, quarkus.datasource."my-store".jdbc.initial-size, quarkus.datasource."my-store".jdbc.max-size,... are advertised to Quarkus.

What I'm advocating for is making this narrower. After looking more at the quarkus logic, the issue is that we need at least 1 runtime property advertised - db-kind being a build-time option is not enough.

>  Advanced: might be good to not requiring any option with the wildcard as we know the datasource names after parsing persistence.xml files or using props for persistence units

My understand is that we want to move away from the persistence.xml files, so we should use a minimal amount of information from them. In fact we probably want to do something like https://github.com/keycloak/keycloak/compare/main...shawkins:pu?expand=1 - so that you don't have to specify the dialect / schema using the persistence.xml, and we may want to make the comment handling consistent (I know we touched on that a little on the comments pr). cc @vmuzikar 

That branch also shows just a single mapping from kind to url - that is sufficient, but still not aligned with our default datasource logic. The only way to do that would be similar to the logic that you have - but only advertise the connected options with default values.

mabartos:
@shawkins I've added some changes here. It also fixes the #41641 bug with `dev-file`.

Changes described in the PR description.

Quickstarts work as expected with these changes: https://github.com/mabartos/keycloak-quickstarts/commit/4c4b3043da9fd48772ad3a2c4c89d9cfb1086565

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
```diff
@@ -105,10 +108,23 @@ public Iterator<String> iterateNames(ConfigSourceInterceptorContext context) {
             }
             allMappers.remove(mapper);
 
-            // include additional mappings if we're on the from side of the mapping
-            // as the mapping may not be bi-directional
-
-            if (!mapper.hasWildcard() && name.equals(mapper.getFrom())) {
+            if (mapper.hasWildcard()) {
+                // we require setting DB kind for additional datasources, so we can advertise connected properties with defaults + (DB URL) to Quarkus
+                var isDbKindOption = isDatasourceOption(mapper.getOption().getKey(), DatabaseOptions.DB);
```
shawkins: At first glance this seems to be a relatively expensive check - for each property in the configuration are we doing a scan of all db options? Don't we just need to look for things that start with kc.db then check for kc.db or kc.db-kind?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
```diff
@@ -23,9 +26,10 @@ public class WildcardPropertyMapper<T> extends PropertyMapper<T> {
 
     private static final Pattern valueValidator = Pattern.compile("[\\[\\]\\$\\-._a-zA-Z0-9]+");
 
+    private final Supplier<List<WildcardPropertyMapper<?>>> connectedMappers;
```
shawkins: We no longer need the connectedMappers concept correct? We just need to ensure a single runtime option is present.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
shawkins: I should have said the additional compute / memory cost - compared to doing it inline with some additional logic. But thinking about it more, that probably isn't worth worrying about.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
mabartos: @shawkins Relatively - for each wildcard property go over datasource options. Should not be a big problem now, but once we have much more wildcard options, it'd be worse - so agree to improve it.

I've simplified the check to explicitly check the DB kind:

https://github.com/keycloak/keycloak/pull/41026/files#diff-922effaaab1831700be05842474f2d2a14f07e5faeecf3ee285a9881e4de52adR158-R166

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
mabartos: Not exactly like that. It is true that we'd need to ensure a single runtime option is present, but it's not the end. For additional datasource, we need to also advertise transaction settings, otherwise the datasource will not work as expected. 

I find the connectedMappers functionality suitable for these cases as we advertise the connected options (with the same wildcard key) that have default values + the required JDBC URL prop. It prevents to face some issues in the future that we might forget to explicitly advertise something related.

I think this functionality might be useful even for the Hibernate ORM wildcard properties.

@shawkins Does it work for you, or do you suggest a different approach?

shawkins: All wildcards probably shouldn't automatically having a concept of connectedMappers - the same paradigm does not hold for example with a logging category. 

Could we just explicitly list connected options (URL + txn) as part of the option builder for DB kind? Then we shouldn't need a specific check for db kind in the PropertyMappingInterceptor - just whether the wildcard has connected options.

mabartos: > Could we just explicitly list connected options (URL + txn) as part of the option builder for DB kind? Then we shouldn't need a specific check for db kind in the PropertyMappingInterceptor - just whether the wildcard has connected options.

+1

quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
```diff
@@ -112,99 +112,104 @@ public class DatabaseOptions {
...
+        /**
+         * You can override some {@link OptionBuilder} methods for additional datasources in this map
+         */
+        private static final Map<Option<?>, Consumer<OptionBuilder<?>>> DATASOURCES_OVERRIDES_OPTIONS = Map.of(
+                DatabaseOptions.DB, builder -> builder
+                        .defaultValue(Optional.empty()) // no default value for DB kind for datasources
+                        .connectedOptions(
+                                getDatasourceOption(DatabaseOptions.DB_URL).orElseThrow(),
+                                TransactionOptions.TRANSACTION_XA_ENABLED_DATASOURCE
+                        )
```
mabartos, line 157: We set connected options for the `db-kind` datasource option.

mabartos:
@shawkins Thanks for the feedback. We can provide concept of connected options as they have certain relationship with each other - right now when the parent option is specified, the connected options's `to` is advertised to Quarkus. 

I've edited the PR description and verified it works for Quickstarts. 

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
```diff
@@ -105,10 +108,20 @@ public Iterator<String> iterateNames(ConfigSourceInterceptorContext context) {
...
+                    var wildcardValue = wildcardMapper.extractWildcardValue(name).orElseThrow();
+                    var connectedTo = wildcardMapper.getConnectedOptions(wildcardValue).stream()
+                            .map(option -> Optional.ofNullable(PropertyMappers.getMapper(NS_KEYCLOAK_PREFIX + option)).orElseThrow(() -> new IllegalArgumentException("Cannot find connected options")))
+                            .map(m -> m.hasWildcard() ? ((WildcardPropertyMapper<?>) m).getTo(wildcardValue) : m.getTo());
 
-            if (!mapper.hasWildcard() && name.equals(mapper.getFrom())) {
+                    return Stream.concat(toDistinctStream(name, wildcardMapper.getTo(wildcardValue)), connectedTo);
+                }
+            } else if (name.equals(mapper.getFrom())) {
+                // include additional mappings if we're on the from side of the mapping as the mapping may not be bi-directional
```
shawkins: This should apply in the wildcard case as well - if the user puts the property directly in quarkus.properties, then we don't need to take responsibility for advertising the keycloak option.

mabartos: @shawkins Ok, I'm ok with that. In that case, the user is required to set all the properties - we don't advertise the defaults for JDBC URL or TX. 

But IMO it should be ok.

mabartos: @shawkins Resolved here: https://github.com/keycloak/keycloak/pull/41026/commits/fd968846b4b1399a7ca76fd3e5c5844e34c87a78

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/PropertyMappingInterceptor.java
```diff
@@ -105,34 +108,45 @@ public Iterator<String> iterateNames(ConfigSourceInterceptorContext context) {
...
-                // the current example is something like log-level=wildcardCat1:level,wildcardCat2:level
-                var wildCard = PropertyMappers.getWildcardMappedFrom(mapper.getOption());
-                if (wildCard != null) {
-                    ConfigValue value = context.proceed(name);
-                    if (value != null && value.getValue() != null) {
-                        return Stream.concat(Stream.of(name), wildCard.getToFromWildcardTransformer(value.getValue()));
+            final PropertyMapper<?> mappedMapper = mapper.forKey(name);
+
+            // include additional mappings if we're on the from side of the mapping as the mapping may not be bi-directional
+            if (name.equals(mappedMapper.getFrom())) {
```
shawkins: Sorry one more thought. We should take this further:

```
            final PropertyMapper<?> mappedMapper = mapper.forKey(name);

            // only include additional mappings if we're on the from side of the mapping as the mapping may not be bi-directional
            if (!name.equals(mappedMapper.getFrom())) {
              return Stream.of(name);
            }

            allMappers.remove(mapper);

            ...
```

mabartos: @shawkins Yep, LGTM, thanks!

shawkins (approved):
LGTM, thank you @mabartos

mabartos:
@shawkins Thanks for the review!

PR #41027, opened 2025-07-09
Allow Features to declare that they support Rolling upgrades

### LINKED ISSUE #41022: Allow Features to declare that they support Rolling upgrades
### Description

It should be possible for individual `Profile.Feature` instances to declare whether they support Rolling upgrades when the Feature is enabled/disabled. By default we assume a Feature supports Rolling upgrades.

```java
    enum FeatureUpdatePolicy {
        // Always allow a rolling update when the Feature is enabled/disabled
        ROLLING,
        // Allow rolling update, but not when going from V1 to V2 or V2 to V1
        ROLLING_NO_UPGRADE,
        // Always require a cluster shutdown when the Feature is enabled/disabled
        SHUTDOWN;
    }

```

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #41022

Metadata generated with the default Features:

```json
{
    "passkeys": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "admin-fine-grained-authz": "{\"enabled\":true,\"version\":2,\"updatePolicy\":\"ROLLING_NO_UPGRADE\"}",
    "dpop": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "impersonation": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "device-flow": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "admin": "{\"enabled\":true,\"version\":2,\"updatePolicy\":\"ROLLING\"}",
    "ciba": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "kerberos": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "login": "{\"enabled\":true,\"version\":2,\"updatePolicy\":\"ROLLING_NO_UPGRADE\"}",
    "token-exchange": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "docker": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "transient-users": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "authorization": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "token-exchange-external-internal": "{\"enabled\":false,\"version\":2,\"updatePolicy\":\"ROLLING\"}",
    "hostname": "{\"enabled\":true,\"version\":2,\"updatePolicy\":\"ROLLING\"}",
    "clusterless": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"SHUTDOWN\"}",
    "recovery-codes": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "rolling-updates": "{\"enabled\":false,\"version\":2,\"updatePolicy\":\"ROLLING_NO_UPGRADE\"}",
    "scripts": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "update-email": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "par": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "multi-site": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"SHUTDOWN\"}",
    "persistent-user-sessions": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"SHUTDOWN\"}",
    "admin-api": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "client-secret-rotation": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "quick-theme": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "user-event-metrics": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "fips": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "ipa-tuura-federation": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "web-authn": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "oid4vc-vci": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "client-types": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "step-up-authentication": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "account-api": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "instagram-broker": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "organization": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "dynamic-scopes": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "declarative-ui": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "client-policies": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "logout-all-sessions": "{\"enabled\":false,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "opentelemetry": "{\"enabled\":true,\"version\":1,\"updatePolicy\":\"ROLLING\"}",
    "account": "{\"enabled\":true,\"version\":3,\"updatePolicy\":\"ROLLING\"}",
    "token-exchange-standard": "{\"enabled\":true,\"version\":2,\"updatePolicy\":\"ROLLING\"}"
}
```

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -53,9 +53,9 @@ public class Profile {
 
         ACCOUNT_V3("Account Console version 3", Type.DEFAULT, 3, Feature.ACCOUNT_API),
 
-        ADMIN_FINE_GRAINED_AUTHZ("Fine-Grained Admin Permissions", Type.PREVIEW, 1, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
+        ADMIN_FINE_GRAINED_AUTHZ("Fine-Grained Admin Permissions", Type.PREVIEW, 1),
 
-        ADMIN_FINE_GRAINED_AUTHZ_V2("Fine-Grained Admin Permissions version 2", Type.DEFAULT, 2, FeatureUpdatePolicy.ROLLING_NO_UPGRADE, Feature.AUTHORIZATION),
+        ADMIN_FINE_GRAINED_AUTHZ_V2("Fine-Grained Admin Permissions version 2", Type.DEFAULT, 2, Feature.AUTHORIZATION),
 
         ADMIN_API("Admin API", Type.DEFAULT),
 
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -132,8 +132,8 @@ public class Profile {
 
         LOGOUT_ALL_SESSIONS_V1("Logout all sessions logs out only regular sessions", Type.DEPRECATED, 1),
 
-        ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
-        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.PREVIEW, 2, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
+        ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1),
+        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.PREVIEW, 2),
 
         /**
          * @see <a href="https://github.com/keycloak/keycloak/issues/37967">Deprecate for removal the Instagram social broker</a>.
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -53,17 +53,17 @@ public class Profile {
 
         ACCOUNT_V3("Account Console version 3", Type.DEFAULT, 3, Feature.ACCOUNT_API),
 
-        ADMIN_FINE_GRAINED_AUTHZ("Fine-Grained Admin Permissions", Type.PREVIEW, 1),
+        ADMIN_FINE_GRAINED_AUTHZ("Fine-Grained Admin Permissions", Type.PREVIEW, 1, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
 
-        ADMIN_FINE_GRAINED_AUTHZ_V2("Fine-Grained Admin Permissions version 2", Type.DEFAULT, 2, Feature.AUTHORIZATION),
+        ADMIN_FINE_GRAINED_AUTHZ_V2("Fine-Grained Admin Permissions version 2", Type.DEFAULT, 2, FeatureUpdatePolicy.ROLLING_NO_UPGRADE, Feature.AUTHORIZATION),
 
         ADMIN_API("Admin API", Type.DEFAULT),
 
         ADMIN_V2("New Admin Console", Type.DEFAULT, 2, Feature.ADMIN_API),
 
-        LOGIN_V2("New Login Theme", Type.DEFAULT, 2),
+        LOGIN_V2("New Login Theme", Type.DEFAULT, 2, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
 
-        LOGIN_V1("Legacy Login Theme", Type.DEPRECATED, 1),
+        LOGIN_V1("Legacy Login Theme", Type.DEPRECATED, 1, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
 
         QUICK_THEME("WYSIWYG theme configuration tool", Type.EXPERIMENTAL, 1),
 
@@ -106,15 +106,15 @@ public class Profile {
 
         TRANSIENT_USERS("Transient users for brokering", Type.EXPERIMENTAL),
 
-        MULTI_SITE("Multi-site support", Type.DISABLED_BY_DEFAULT),
+        MULTI_SITE("Multi-site support", Type.DISABLED_BY_DEFAULT, FeatureUpdatePolicy.SHUTDOWN),
 
-        CLUSTERLESS("Store all session data, work cache and login failure data in an external Infinispan cluster.", Type.EXPERIMENTAL),
+        CLUSTERLESS("Store all session data, work cache and login failure data in an external Infinispan cluster.", Type.EXPERIMENTAL, FeatureUpdatePolicy.SHUTDOWN),
 
         CLIENT_TYPES("Client Types", Type.EXPERIMENTAL),
 
         HOSTNAME_V2("Hostname Options V2", Type.DEFAULT, 2),
 
-        PERSISTENT_USER_SESSIONS("Persistent online user sessions across restarts and upgrades", Type.DEFAULT),
+        PERSISTENT_USER_SESSIONS("Persistent online user sessions across restarts and upgrades", Type.DEFAULT, FeatureUpdatePolicy.SHUTDOWN),
 
         OID4VC_VCI("Support for the OID4VCI protocol as part of OID4VC.", Type.EXPERIMENTAL),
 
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/keycloak/common/Profile.java
```diff
@@ -132,8 +132,8 @@ public enum Feature {
 
         LOGOUT_ALL_SESSIONS_V1("Logout all sessions logs out only regular sessions", Type.DEPRECATED, 1),
 
-        ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1),
-        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.PREVIEW, 2),
+        ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
+        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.PREVIEW, 2, FeatureUpdatePolicy.ROLLING_NO_UPGRADE),
```
pruivo: I don't think we should shut down the cluster when enabling V2, mainly because it does not affect the Keycloak server, just the Keycloak Operator logic.

services/src/main/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+        Map<String, String> currentMeta = metadata();
+        for (Map.Entry<String, String> entry : other.entrySet()) {
+            String featureKey = entry.getKey();
+            String otherJson = entry.getValue();
+            Feature otherFeature = fromJson(otherJson);
+
+            // Feature has been removed in current version
+            if (!currentMeta.containsKey(featureKey)) {
+                // If the feature was previously enabled then we shutdown to ensure compatibility
+                if (otherFeature.enabled)
```
pruivo: I think you should check `FeatureUpdatePolicy`. If it is `ROLLING`,  we don't need to shut down.

services/src/main/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                return v;
+            });
+        }
+        return metadata;
+    }
+
+    @Override
+    public CompatibilityResult isCompatible(Map<String, String> other) {
+        Map<String, String> currentMeta = metadata();
+        for (Map.Entry<String, String> entry : other.entrySet()) {
```
pruivo, line 41: It should also check for new features in this version that weren't present in the old version. If a new feature exists + is enabled + `FeatureUpdatePolicy.SHUTDOWN`, then it should shut down the cluster.

services/src/test/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProviderTest.java
```diff
@@ -0,0 +1,158 @@
...
+              .filter(f -> !f.getDependencies().isEmpty())
+              .forEach(f -> {
+                  for (Profile.Feature dep : f.getDependencies())
+                      DEPENDENT_FEATURES.compute(dep, (k, v) -> v == null ? new HashSet<>() : v).add(f);
+              });
+    }
+
+    @ParameterizedTest
+    @MethodSource("rollingFeatures")
+    public void testRollingPolicy(Profile.Feature feature) {
```
pruivo, line 33: It is missing some cases...
old enabled -> new disabled
old disabled -> new disabled

services/src/test/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProviderTest.java
```diff
@@ -0,0 +1,158 @@
...
+    private static Stream<Arguments> rollingFeatures() {
+        return Arrays.stream(Profile.Feature.values())
+              .filter(f -> f != Profile.Feature.HOSTNAME_V2)
+              .filter(f -> f.getUpdatePolicy() == Profile.FeatureUpdatePolicy.ROLLING)
+              .map(Arguments::of);
+    }
+
+    @ParameterizedTest
+    @MethodSource("shutdownFeatures")
+    public void testFeatureShutdownPolicy(Profile.Feature feature) {
```
pruivo, line 72: Missing both disabled scenario.

services/src/test/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProviderTest.java
```diff
@@ -0,0 +1,158 @@
...
+        assertFeature(f, false, feature.getVersion(), Profile.FeatureUpdatePolicy.ROLLING);
+        var featureDisabledMap = Map.of(feature.getUnversionedKey(), FeatureCompatibilityMetadataProvider.toJson(f));
+        Profile.reset();
+        Profile.configure();
+        assertCompatibility(CompatibilityResult.ExitCode.ROLLING, provider.isCompatible(featureDisabledMap));
+    }
+
+    private static Stream<Arguments> rollingFeatures() {
+        return Arrays.stream(Profile.Feature.values())
+              .filter(f -> f != Profile.Feature.HOSTNAME_V2)
```
pruivo, line 60: Why is `HOSTNAME_V2` skipped?

services/src/test/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProviderTest.java
```diff
@@ -0,0 +1,158 @@
...
+    private static Stream<Arguments> shutdownFeatures() {
+        return Arrays.stream(Profile.Feature.values())
+              .filter(f -> f != Profile.Feature.HOSTNAME_V2)
+              .filter(f -> f.getUpdatePolicy() == Profile.FeatureUpdatePolicy.SHUTDOWN)
+              .map(Arguments::of);
+    }
+
+    @ParameterizedTest
+    @MethodSource("noUpgradeFeatures")
+    public void testRollingNoUpgradePolicy(Profile.Feature v1, Profile.Feature v2) {
```
pruivo: missing test switching from v2 enabled to v2 disabled  + v1 enabled

pruivo (changes requested):

services/src/test/java/org/keycloak/compatibility/FeatureCompatibilityMetadataProviderTest.java
ryanemerson: It's not possible to disable it: https://github.com/keycloak/keycloak/blob/baba2b14cedb8e65adc814bc2b3321844cd80382/common/src/main/java/org/keycloak/common/Profile.java#L290-L292

ryanemerson:
Thanks for the review @pruivo.  I have addressed all comments.

I have simplified `testRollingNoUpgradePolicy` by just testing login v1/v2, as `ADMIN_FINE_GRAINED_AUTHZ_V2` has a dependency on `AUTHORIZATION` which makes the test logic deviate slightly.

ahus1 (commented):
I've changed ADMIN_FINE_GRAINED_AUTHZ to be an feature that allows rolling update. I rather err on the side of allowing too much rolling :D 
If that is something that core-iam want to change, they can do this in a follow-up PR.

ahus1:
@pruivo - please review when you have the time as you had some comments earlier and let us know if they have been addressed.

pruivo (approved):
It's missing a test that adds a feature. I'm fine merging as is if you think the test is an overkill :smiley:

ahus1 (approved):

ahus1:
As agreed on the team meeting, please add the test case suggested above to the next PR. Thanks!

PR #41039, opened 2025-07-09
Remove /etc/system-fips file before executing fips-mode-setup

### LINKED ISSUE #41038: FIPS errors in CI
### Before reporting an issue

### Area

testsuite

### Describe the bug

The three FIPS jobs started to fail today. For example: https://github.com/keycloak/keycloak/actions/runs/16173014816/job/45651564423

```noformat
/usr/bin/fips-finish-install: line 29: /etc/system-fips: No such file or directory
Installation of FIPS modules could not be completed.
```

Probably because an upgrade in something (ubuntu or rhel image). We just need to delete that file before because the problem is that is linked to a missing directory.

### Version

999.0.0-SNAPSHOT

### Regression

### Expected behavior

FIPS jobs should run OK.

### Actual behavior

The commented failure.

### How to Reproduce?

It's just in the CI.

### Anything else?

_No response_

rmartinc:
Closes #41038

FIPS jobs started to fail today. I checked that this `/etc/system-fips` file is created with a link to a non-existent folder (probably because we are using a hack in ubuntu to simulate the fips configuration and it thinks it is partially configured). This file is a mark that the `fips-mode-setup` creates when it is initialized. The scripts now remove the file before executing the `fips-mode-setup` and the file is created OK in the initialization.

### DIFF
--- a/.github/scripts/run-fips-it.sh
+++ b/.github/scripts/run-fips-it.sh
@@ -1,5 +1,6 @@
 #!/bin/bash -x
 
+rm -f /etc/system-fips
 dnf install -y java-21-openjdk-devel
 fips-mode-setup --enable --no-bootcfg
 fips-mode-setup --is-enabled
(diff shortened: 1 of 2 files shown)


mposolda (approved):
@rmartinc Thanks!

PR #41046, opened 2025-07-09
Update email feature only enabled if the required action is enabled at the realm

### LINKED ISSUE #41045: Update email feature only enabled if the required action is enabled at the realm
### Description

Now that update email feature is enabled by default it should be enabled on a per-realm basis so that users can opt-in for the behavior introduced by this feature updating the email through the different user profile contexts.

For that, we should check if the `Update Email` required action is enabled.

pedroigor:
Closes #41045

* In order not to break existing deployments and allow users to opt in for the new behavior introduced by the feature, this PR is now enabling the new behavior only if the `Update Email` required action is enabled.
* By default, the required action is disabled
* The PR https://github.com/keycloak/keycloak/pull/40273 introduced the `UPDATE_EMAIL` feature as enabled by default and changed the current behavior. As a consequence, a lot of tests had to change.

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/actions/RequiredActionEmailVerificationTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/actions/RequiredActionEmailVerificationTest.java
@@ -42,7 +42,6 @@ import org.keycloak.testsuite.AssertEvents;
 import org.keycloak.testsuite.AbstractTestRealmKeycloakTest;
 import org.keycloak.testsuite.admin.ApiUtil;
 import org.keycloak.testsuite.arquillian.AuthServerTestEnricher;
-import org.keycloak.testsuite.arquillian.annotation.EnableFeature;
 import org.keycloak.testsuite.cluster.AuthenticationSessionFailoverClusterTest;
 import org.keycloak.testsuite.pages.AppPage;
 import org.keycloak.testsuite.pages.AppPage.RequestType;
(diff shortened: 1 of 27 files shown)


rmartinc (approved):
Thanks @pedroigor! I confirmed that the PR just re-establishes the test as previously before 5a42390341a9f5112a7b4e7ec3ffd2f4c20f1059. Now the update_email feature is supported but disabled by default. The change is OK to me if @keycloak/core-iam thinks this way is better.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/actions/RequiredActionEmailVerificationTest.java
```diff
@@ -1112,7 +1111,6 @@ public void changeEmailAddressAfterSendingEmail() throws Exception {
     }
 
     @Test
-    @EnableFeature(value = Profile.Feature.UPDATE_EMAIL, skipRestart = true) // The feature is enabled by default but skipRestart = true is important for the test
```
martin-kanis, line 1115: Annotation was there before, I added only the comment. So the correct revert is just removing the comment.

martin-kanis (changes requested):
Thanks @pedroigor. Looks good, I noticed just one thing, please see below.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/actions/RequiredActionEmailVerificationTest.java
pedroigor: Why keep it?

martin-kanis: OK I guess it's fine if the test is working without it

martin-kanis (approved):

ahus1 (approved):

PR #41068, opened 2025-07-10
Add pagination to user idp links.

### LINKED ISSUE #40818: Identity provider links list is limited to 100 entries for a user in the admin UI
### Before reporting an issue

### Area

admin/ui

### Describe the bug

In the Keycloak admin UI, when viewing the list of available identity providers to link for a specific user (under Users → [User] → Identity Provider Links), the list appears to be capped at 100 entries. This causes only a subset of the configured identity providers to be displayed, making it impossible to link to providers beyond the first 100.

There doesn't appear to be any setting to configure or override this limit in the admin UI or via server configuration.

Image

### Version

26.2.5

### Regression

### Expected behavior

All configured identity providers should be visible in the list or the list should support pagination or scrolling. If a limit must exist, it should be configurable via the UI or server settings.

### Actual behavior

 Only the first 100 identity providers are shown, with no way to access the others.

### How to Reproduce?

1. Configure more than 100 identity providers in a realm.
2. Go to a user profile in the admin console.
3. Click the Identity Provider Links tab.
4. Open the dropdown to link a new identity provider.
5. Observe that only 100 identity providers are listed.

### Anything else?

* Deployment done with the keycloak operator.

ssilvert:
Fixes #40818

We now leave the full list of available idps intact but hide the "Link account" link for those who have already been linked.  This allows us to do proper pagination.


js/apps/admin-ui/src/user/UserIdentityProviderLinks.tsx
```diff
@@ -34,11 +36,11 @@ export const UserIdentityProviderLinks = ({
   userId,
 }: UserIdentityProviderLinksProps) => {
   const { adminClient } = useAdminClient();
-
   const [key, setKey] = useState(0);
+  const [linkedNames, setLinkedNames] = useState<string[]>([]);
+  const [isLoading, setIsLoading] = useState(true); // Added loading state
```
jonkoops: I think we can omit the comment here, it doesn't really seem to add that much.
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

js/apps/admin-ui/src/user/UserIdentityProviderLinks.tsx
```diff
@@ -49,7 +51,35 @@ export const UserIdentityProviderLinks = ({
...
+  );
+
+  const refresh = () => {
+    setKey(new Date().getTime());
+    setIsLoading(true);
+    // Re-fetch linked names
+    adminClient.users
+      .listFederatedIdentities({ id: userId })
+      .then((linkedIdentities) => {
+        setLinkedNames(linkedIdentities.map((x) => x.identityProvider!));
```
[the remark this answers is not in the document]
jonkoops: Agreed, this seems like repeated logic. Perhaps the refetch logic could be triggered by adding `key` as a dependency to the `useFetch()`.

js/apps/admin-ui/src/user/UserIdentityProviderLinks.tsx
```diff
@@ -74,22 +104,25 @@ export const UserIdentityProviderLinks = ({
...
+    const params: IdentityProvidersQuery = {
+      first: first!,
+      max: max!,
+      realmOnly: false,
+    };
+    if (search) {
+      params.search = search;
+    }
+    const providers = await adminClient.identityProviders.find(params);
+    return providers;
```
jonkoops: Looks like this could be returned directly.
```suggestion
    return await adminClient.identityProviders.find(params);
```

js/apps/admin-ui/src/user/UserIdentityProviderLinks.tsx
```diff
@@ -49,7 +51,25 @@ export const UserIdentityProviderLinks = ({
     "view-identity-providers",
   );
 
-  const refresh = () => setKey(new Date().getTime());
+  useEffect(() => {
+    refresh();
+  }, []);
+
+  const refresh = () => {
```
jonkoops, line 65: Why not use the `useFetch()` hook here? It has built-in error handling. If you need to trigger a refresh, you can use a piece of state to handle it, for example, since every refresh triggers a change in `key` that could be a dependency to trigger the refresh.

ssilvert: I tried useFetch() but it requires that I pass in a function for callback, which in this case would be empty.  This is disallowed by the linter.

I'm not inclined to change useFetch() to accept undefined just for this weird use case.

jonkoops: The callback should be used to set any state that results from the first function (which retrieves the state), so in this case it should be possible to call `setLinkedNames()`, for example:

```ts
const [key, setKey] = useState(0);

const refresh = () => {
  setKey((value) => value + 1);
};

useFetch(
  () => adminClient.users.listFederatedIdentities({ id: userId }),
  (linkedIdentities) => setLinkedNames(linkedIdentities.map((identity) => identity.identityProvider!)),
  [userId, key],
);
```

This will fetch the federated identities, and then set the linked names from that state, if the `key` changes, it will re-fetch since it is a dependency.

ssilvert: @jonkoops I understand that.  The problem with coding it that way is that it doesn't take care of isLoading.

Maybe it could be done, but I don't have time to keep rewriting this.

jonkoops: We're using this hook pretty much everywhere to fetch data, I think the loading state is handled by emptying the state beforehand and then checking if it is set. I would really rather we keep using the same convention as much as possible.

If `useFetch()`'s lack of loading state is a problem, I have no issue with adding a return value, so this could be determined, the implementation for that would be rather trivial. Perhaps something like:

```ts
const { isFetching } = useFetch(...)
```

edewit: I agree with @jonkoops we should use the same style everywhere, loading state is handled by waiting for the state to be ready like this:

```tsx

const [key, setKey] = useState(0);

const refresh = () => {
  setKey((value) => value + 1);
};

useFetch(
  () => adminClient.users.listFederatedIdentities({ id: userId }),
  (linkedIdentities) => setLinkedNames(linkedIdentities.map((identity) => identity.identityProvider!)),
  [userId, key],
);

if (!linkedIdentities) {
  return <KeycloakSpinner />;
}
```
```

Comment (ssilvert) [reply js/apps/admin-ui/src/user/UserIdentityProviderLinks.tsx:65]: I was completely over-thinking this.  The solution is actually quite simple:

```
useFetch(
    () => adminClient.users.listFederatedIdentities({ id: userId }),
    (linkedIdentities) => {
      setLinkedNames(
        linkedIdentities.map((identity) => identity.identityProvider!),
      );
      setIsLoading(false); // <-- just reset the flag here !!!
    },
    [userId, key],
  );

  const refresh = () => {
    setKey(new Date().getTime());
    setIsLoading(true);
  };
```

jonkoops: Nice, let's mark this as resolved!

jonkoops (approved):
LGTM! Thanks for the patience :)

PR #41087, opened 2025-07-10
Require setting DB kind for additional datasources

### LINKED ISSUE #41161: Require setting DB kind for additional datasources
### Description

We should require setting at least the DB kind for additional datasources to know how the additional datasource is named, as other properties can be then propagated to the Quarkus side. 

More info in comment: https://github.com/keycloak/keycloak/pull/41026/files

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mabartos:
- Closes #41161 
- Blocked-by https://github.com/keycloak/keycloak/pull/41398

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
@@ -21,25 +21,11 @@ import java.io.File;
 import java.lang.annotation.Annotation;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.function.Supplier;
 import java.util.stream.Stream;
 
-import com.google.common.base.Suppliers;
-import io.agroal.api.AgroalDataSource;
-import io.quarkus.agroal.DataSource;
-import io.quarkus.arc.Arc;
-import io.quarkus.arc.InstanceHandle;
-import io.quarkus.hibernate.orm.runtime.integration.HibernateOrmIntegrationRuntimeInitListener;
-import io.quarkus.runtime.annotations.Recorder;
-import io.vertx.core.Handler;
-import io.vertx.ext.web.RoutingContext;
-import liquibase.Scope;
-import liquibase.servicelocator.ServiceLocator;
 import org.hibernate.cfg.AvailableSettings;
 import org.infinispan.protostream.SerializationContextInitializer;
-import org.jboss.logging.Logger;
 import org.keycloak.Config;
 import org.keycloak.common.Profile;
 import org.keycloak.common.crypto.CryptoIntegration;
@@ -51,11 +37,8 @@ import org.keycloak.marshalling.Marshalling;
 import org.keycloak.provider.Provider;
 import org.keycloak.provider.ProviderFactory;
 import org.keycloak.provider.Spi;
-import org.keycloak.quarkus.runtime.cli.PropertyException;
 import org.keycloak.quarkus.runtime.configuration.Configuration;
 import org.keycloak.quarkus.runtime.configuration.MicroProfileConfigProvider;
-import org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers;
-import org.keycloak.quarkus.runtime.configuration.mappers.WildcardPropertyMapper;
 import org.keycloak.quarkus.runtime.integration.QuarkusKeycloakSessionFactory;
 import org.keycloak.quarkus.runtime.storage.database.liquibase.FastServiceLocator;
 import org.keycloak.representations.userprofile.config.UPConfig;
@@ -63,7 +46,16 @@ import org.keycloak.theme.ClasspathThemeProviderFactory;
 import org.keycloak.truststore.TruststoreBuilder;
 import org.keycloak.userprofile.DeclarativeUserProfileProviderFactory;
 
-import static org.keycloak.quarkus.runtime.storage.database.jpa.QuarkusJpaConnectionProviderFactory.DEFAULT_PERSISTENCE_UNIT;
+import io.agroal.api.AgroalDataSource;
+import io.quarkus.agroal.DataSource;
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.InstanceHandle;
+import io.quarkus.hibernate.orm.runtime.integration.HibernateOrmIntegrationRuntimeInitListener;
+import io.quarkus.runtime.annotations.Recorder;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
+import liquibase.Scope;
+import liquibase.servicelocator.ServiceLocator;
 
 @Recorder
 public class KeycloakRecorder {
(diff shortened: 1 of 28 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/WildcardPropertyMapper.java
```diff
@@ -45,17 +45,16 @@ public WildcardPropertyMapper(Option<T> option, String to, BooleanSupplier enabl
         if (option == LoggingOptions.LOG_LEVEL_CATEGORY) {
             replacementChar = '.';
         }
-
-        if (to != null) {
-            if (!to.startsWith(NS_QUARKUS_PREFIX) && !to.startsWith(NS_KEYCLOAK_PREFIX)) {
+        if (getTo() != null) {
```
mabartos: fyi - it fixes a bug when a wildcard mapper has no `to()` method provided. Wildcard PM calls the parent PM and handle the `to` on its own. In this case, the `to` is considered as the constructor parameter -> use `getTo()` which handles the updated `to` from the parent PM.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
```diff
@@ -76,6 +80,19 @@ public Handler<RoutingContext> getManagementHandler() {
         return routingContext -> routingContext.response().end("Keycloak Management Interface");
     }
 
+    public void validatePersistenceUnits(List<String> descriptors) {
+        List<String> notSetPersistenceUnitsDBKinds = descriptors.stream()
+                .filter(descriptorName -> !descriptorName.equals(DEFAULT_PERSISTENCE_UNIT)) // not default persistence unit
+                // map PU name to DB kind option like: user-store -> db-kind-user-store
+                .map(descriptorName -> DatabaseOptions.getResultNamedKey(DatabaseOptions.DB, descriptorName).orElseThrow())
```
shawkins: If the persistence.xml specifies a datasource name, that may not match the persistence unit name.

In the vein of https://github.com/keycloak/keycloak/compare/main...shawkins:keycloak:pu we could default the datasource name to that of the persistence unit when not specified to cut down on what has to be in the xml.

shawkins (commented):
Looks good, just a thought about the datasource name.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
mabartos: Yes, the problem might be a little bit wider. Should be solved by: https://github.com/keycloak/keycloak/issues/41192

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
```diff
@@ -77,6 +80,19 @@ public Handler<RoutingContext> getManagementHandler() {
...
 
+    public void validatePersistenceUnits(List<String> descriptors) {
+        List<String> notSetPersistenceUnitsDBKinds = descriptors.stream()
+                .filter(descriptorName -> !descriptorName.equals(DEFAULT_PERSISTENCE_UNIT)) // not default persistence unit
+                // map PU name to DB kind option like: user-store -> db-kind-user-store
+                .map(descriptorName -> DatabaseOptions.getNamedKey(DatabaseOptions.DB, descriptorName).orElseThrow())
+                .filter(dbKind -> Configuration.getKcConfigValue(dbKind).getValue() == null) // not provided
+                .toList();
+
+        if (!notSetPersistenceUnitsDBKinds.isEmpty()) {
```
shawkins: Just to make sure, we're okay with this level of a breaking change? We talked previously about how persistence.xml is supported, but the use of quarkus properties still isn't, so this is grey area.

cc @vmuzikar

mabartos: @shawkins What about to ask for the `quarkus.datasource."<datasource>".db-kind` instead of `kc.db-kind-<datasource>` as we could fetch values from the `quarkus.properties`? In that case, we could still support both variants.

But in that case, we would not know anything about the wildcard name and we would not be able to advertise more properties such as JDBC URL to Quarkus used here: https://github.com/keycloak/keycloak/pull/41026

I'll think about it if we have some better alternatives.

shawkins: > But in that case, we would not know anything about the wildcard name

Our current logic presumes quoting and will associate the corresponding quoted quarkus property with the keycloak property. But I believe the smallrye handling does not do quoting normalizations, so if they leave the name of the datasource unquoted we won't find it.

vmuzikar:
I'd like to take a look at this early this week.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
vmuzikar: As discussed offline, despite the fact we generally don't support raw Quarkus options, the additional datasources were kinda and exception as it was also included in our quickstarts. Lots of users are probably using it like that, we can't make breaking changes in this area.

mabartos: Updated, but in order not to introduce any breaking changes, we need to handle `quarkus.datasource.user-store.db-kind` the same way as `quarkus.datasource."user-store".db-kind` - we can find the same mapper. For consistency reasons and to avoid any unwanted behavior in the future, I'd suggest to solve this issue first:
- https://github.com/keycloak/keycloak/issues/41397

@shawkins

shawkins:
@mabartos see if these changes make sense. The cases that should work:
1. legacy usage should continue to work - if we detect a non-keycloak form of db-kind that will be a warning
2. Completely new usage should work as expected - db-kind-x is required
3. If the user has all quoted quarkus properties, then they can start using the new db- wildcards and everything will work as expected.

However if the user has all non-quoted properties, then there will be a mismatch between the properties we create and what quarkus understands. I think the easiest path is to say this is not supported - if someone wants to continue to use quarkus properties in this scenario, they must convert them to the quoted form. Maybe there should be a specific warning about that.

mabartos:
@shawkins Thanks! Changes LGTM. 

> However if the user has all non-quoted properties, then there will be a mismatch between the properties we create and what quarkus understands. I think the easiest path is to say this is not supported - if someone wants to continue to use quarkus properties in this scenario, they must convert them to the quoted form. Maybe there should be a specific warning about that.

Yes, as stated in https://github.com/keycloak/keycloak/pull/41398#issuecomment-3116759540, it probably does not worth the effort to support quoted and unquoted forms simultaneously for `quarkus.properties`. Agree to consider it as not supported and rather requiring the quotation - it'd be tiny breaking change with straightforward fix for deployments using the quarkus.properties. Agree that we should have the specific warning + maybe some note in docs later for datasources.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
mabartos: Discussion below.

shawkins:
@mabartos I believe this should be ready. I see from https://github.com/keycloak/keycloak/issues/40387 that documenting this in general and in the upgrading guide will be a follow-on issue correct?

mabartos:
> I believe this should be ready. I see from https://github.com/keycloak/keycloak/issues/40387 that documenting this in general and in the upgrading guide will be a follow-on issue correct?

@shawkins Yes, it should be ready. Yes, docs will be done in follow-up issues.

shawkins (approved):
LGTM

mabartos:
@shawkins Thanks!

PR #41089, opened 2025-07-10
fix: providing a single property to declare management interface as http

### LINKED ISSUE #40945: Unclear documentation for setting management server as http when main server is https
### Description

Using the latest 26.3.0, I am following the documentation from 

https://www.keycloak.org/server/management-interface

And using these options
```
--https-key-store-file="$JAVA_HOME"/lib/security/cacerts 
--https-key-store-password=changeit
--hostname-strict=false \
--http-enabled=true
```

And the following in keycloak.conf file.
```
health-enabled=true
metrics-enabled=true
```
I kept getting this configuration message on startup

`[io.quarkus] (main) Keycloak 26.3.0 on JVM (powered by Quarkus 3.20.1) started in 10.620s. Listening on: http://0.0.0.0:8080 and https://0.0.0.0:8443. Management interface listening on https://0.0.0.0:9000.`

I wanted it to be **http**://0.0.0.0:9000 as we are deploying this in Kubernetes and I did not want HTTPS for getting metrics and health information.

I found the documentation not clear on how to convert the management server to http?  I was expecting there to be something like the following 
- http-management-port=9000 for http and https-management-port=9000 for https
- Something similar to --http-enabled for management configuration
- other setting that was clear on the scheme used by management server.

Instead, to fix the issue I had to pass the following parameters.

```
--https-management-key-store-file="" 
--https-management-key-store-password=""
```
And now I get the configuration I was looking for.

`[io.quarkus] (main) Keycloak 26.3.0 on JVM (powered by Quarkus 3.20.1) started in 10.620s. Listening on: http://0.0.0.0:8080 and https://0.0.0.0:8443. Management interface listening on http://0.0.0.0:9000.`

I don't find this configuration well documented, and I don't find it to be an obvious configuration setting.  I also find it odd to have to set a parameter to "" to fix this issue?

I would like to recommend one of the following enhancements.
- Update documentation to clearly explain how to ensure management server is using http instead of https.
- Add new configuration parameters that clearly configures the management server to use http when main server is using certificates.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
closes: #40945

Introduces http-management-scheme=http to force the management interface to http.

Without this you must do what was shown on #40945 and try to set the keystore / truststore to empty values - this is not a convention that we'd want to document as we generally don't have checks for treating blank the same as unset.

An alternative to having http/inherited is to ahve separate/inherited - with separate meaning that the management configuration is not dependent upon the http configuration at all. This however is difficult for us to implement with our current mapping logic, and we also have a build-time relationship with https-management-client-auth.

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/ManagementOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/ManagementOptions.java
@@ -61,7 +61,18 @@ public class ManagementOptions {
             .defaultValue("0.0.0.0")
             .build();
 
+    public enum Scheme {
+        http,
+        inherited
+    }
+
     //HTTPS
+    public static final Option<Scheme> HTTP_MANAGEMENT_SCHEME = new OptionBuilder<>("http-management-scheme", Scheme.class)
+            .category(OptionCategory.MANAGEMENT)
+            .description("Configures the management interface scheme. If 'inherited', the management interface will inherit the HTTPS settings of the main interface. If 'http', the management interface will be accessible via HTTP - it will not inherit HTTPS settings and cannot be configured for HTTPS.")
+            .defaultValue(Scheme.inherited)
+            .build();
+
     public static final Option<HttpOptions.ClientAuth> HTTPS_MANAGEMENT_CLIENT_AUTH = new OptionBuilder<>("https-management-client-auth", HttpOptions.ClientAuth.class)
             .category(OptionCategory.MANAGEMENT)
             .description("Configures the management interface to require/request client authentication. If not given, the value is inherited from HTTP options. " + RELEVANT_MSG)
--- a/docs/guides/server/management-interface.adoc
+++ b/docs/guides/server/management-interface.adoc
@@ -36,8 +36,10 @@ if you set the CLI option `--http-relative-path=/auth`, these endpoints are acce
 
 === TLS support
 
-When the TLS is set for the default {project_name} server, the management interface will be accessible through HTTPS as well.
+When the TLS is set for the default {project_name} server, by default the management interface will be accessible through HTTPS as well.
 The management interface can run only either on HTTP or HTTPS, not both as for the main server.
+
+NOTE: If you do not want the management interface to use HTTPS, you may set the `http-management-scheme` option to `http`.
 
 Specific {project_name} management interface options with the prefix `https-management-*` were provided for setting different TLS parameters for the management HTTP server. Their function is similar to their counterparts for the main HTTP server, for details see <@links.server id="enabletls" />.
 When these options are not explicitly set, the TLS parameters are inherited from the default HTTP server.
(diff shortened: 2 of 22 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/ManagementOptions.java
```diff
@@ -61,7 +61,18 @@ public class ManagementOptions {
...
 
+    public enum Scheme {
+        http,
+        inherited
+    }
+
     //HTTPS
+    public static final Option<Scheme> HTTP_MANAGEMENT_SCHEME = new OptionBuilder<>("http-management-scheme", Scheme.class)
+            .category(OptionCategory.MANAGEMENT)
+            .description("Configures the management interface scheme.")
```
mabartos:
```suggestion
            .description("Configures the management interface scheme. If 'inherited' is used, the schema will be the same as for the main server. If 'http', the management interface is accessible only via the HTTP schema.")
```

@shawkins Would it be possible to be more explicit here and use sth like this?

mabartos (commented):
@shawkins LGTM, just a comment.

As using HTTP for mgmt interface is quite asked by the community, could you please add a section for mgmt interface guide on https://www.keycloak.org/server/management-interface ? Thanks!

quarkus/config-api/src/main/java/org/keycloak/config/ManagementOptions.java
shawkins: Ok, refined the text. It may be a little confusing that the main interface can be accessible via both http and https, so I tried to account for that.

quarkus/config-api/src/main/java/org/keycloak/config/ManagementOptions.java
```diff
@@ -61,7 +61,18 @@ public class ManagementOptions {
...
 
+    public enum Scheme {
+        http,
+        inherited
+    }
+
     //HTTPS
+    public static final Option<Scheme> HTTP_MANAGEMENT_SCHEME = new OptionBuilder<>("http-management-scheme", Scheme.class)
+            .category(OptionCategory.MANAGEMENT)
+            .description("Configures the management interface scheme. If 'inherited', the management interface will inherit the HTTPS settings of the main interface. If 'http', the management interface will be accessible via HTTP - it will not inherit HTTPS settings and cannot be configured for HTTPS.")
```
mabartos, line 72:
```suggestion
            .description("Configures the management interface scheme. If 'inherited', the management interface will inherit the HTTPS settings of the main interface. If 'http', the management interface will be accessible only via HTTP - it will not inherit HTTPS settings and cannot be configured for HTTPS.")
```

@shawkins Yep, better, thanks!

nitpick that could clarify things a tiny bit more

shawkins:
> @shawkins LGTM, just a comment.
> 
> As using HTTP for mgmt interface is quite asked by the community, could you please add a section for mgmt interface guide on https://www.keycloak.org/server/management-interface ? Thanks!

Good idea. Added now.

mabartos (approved):
@shawkins I think the update of the guide is sufficient. LGTM, thanks!

mabartos:
@shawkins Some afterthought - IMO it'd be good to mention the ability to use HTTP for mgmt interface in the release notes of 26.4 as it was quite a requested feature by the community, so we'd have more eyes on this. WDYT?

shawkins:
> @shawkins Some afterthought - IMO it'd be good to mention the ability to use HTTP for mgmt interface in the release notes of 26.4 as it was quite a requested feature by the community, so we'd have more eyes on this. WDYT?

https://github.com/keycloak/keycloak/pull/41144

PR #41110, opened 2025-07-11
[admin-api-v2] Provide simple validation with Jakarta/Hibernate Validation

mabartos:
WIP

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -69,6 +69,10 @@
             <artifactId>microprofile-openapi-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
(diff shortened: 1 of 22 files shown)


vmuzikar (approved):
I think it's a good start and we can even merge as is. It'd be good though to have a follow-up to showcase some custom logic validation.

mabartos:
@vmuzikar Thanks, yep I'd say it's a good start for this. However, for the automatic validation based on the annotations, we'd need to do some more things which would not be so easy to do and would require possibly some discussion. 

Possible approaches:
1. Execute the validation manually inside the code
2. Having resources (APIs) as a CDI Beans for which the validation is run automatically

I'm inclined more to (2) as it's a recommended way in applications and on Quarkus side and it could simplify a lot of things for us. I'll try to do some prototype on this matter.

PR #41128, opened 2025-07-13
[Admin API v2] PoC for a field projection mechanism

thomasdarimont:
PoC for a field selection / projection mechanism 

The projection works on the Jackson ObjectMapper level and uses a SimpleBeanPropertyFilter (FieldProjectionFilter) to control the properties during JSON marshalling. Consumers can select the desired result fields with a `fields` query parameter. Note that the selection only works with top-level object properties. There is currently no support for nested property selection.

Currently, all representations inheriting from BaseRepresentation support this filtering via the JsonFilter annotation.

JAX-RS resource methods which declare a 
`@QueryParam(Projections.PARAM) Projections projections` become eligible for projections.

The explicit parameter declaration ensures that the projection parameter can be inferred for the OpenAPI metadata.

A request without the fields parameter will return the full resource:

```
curl -v -H "Authorization: Bearer $KC_ACCESS_TOKEN" \
http://localhost:8081/auth/admin/api/v2/realms/adminapiv2/clients/custom-client1 | jq -C
```

A request with the included filter:
```
curl -v -H "Authorization: Bearer $KC_ACCESS_TOKEN" \
http://localhost:8081/auth/admin/api/v2/realms/adminapiv2/clients/custom-client1?fields=clientId,displayName
```
will only return the selected fields.

```
curl -s -H "Authorization: Bearer $KC_ACCESS_TOKEN" \
http://localhost:8081/auth/admin/api/v2/realms/adminapiv2/clients?fields=clientId
```

Yields:
```json
[
  {
    "clientId": "account"
  },
  {
    "clientId": "account-console"
  },
  {
    "clientId": "admin-api"
  },
  {
    "clientId": "admin-cli"
  },
  {
    "clientId": "broker"
  },
  {
    "clientId": "custom-client1"
  },
  {
    "clientId": "realm-management"
  },
  {
    "clientId": "security-admin-console"
  }
]
```

### DIFF
--- a/core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
@@ -5,9 +5,11 @@ import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonFilter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 
+@JsonFilter("projection")
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
 public class BaseRepresentation {
 
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/jaxrs/ProjectionsPreFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/jaxrs/ProjectionsPreFilter.java
@@ -1,0 +1,20 @@
+package org.keycloak.quarkus.runtime.integration.jaxrs;
+
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.container.PreMatching;
+import jakarta.ws.rs.ext.Provider;
+import org.keycloak.services.util.Projections;
+
+@Provider
+@PreMatching
+public class ProjectionsPreFilter implements ContainerRequestFilter {
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) {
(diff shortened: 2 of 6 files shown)


thomasdarimont:
Currently I use the query parameter `fields` to control the projections. However some representations might use the property `fields` it might be better to use a different name or prefix differentiate "technical" parameters from actual resource properties.

vmuzikar:
@thomasdarimont Can you please rebase?

vmuzikar (approved):
Overall IMHO this approach makes sense.

Just to note, this purely benefits the network/client side, not the server side as we're still fetching all fields (which might underneath do some unnecessary JOINs, etc.). But wa could quite easily detect the fields in the query param and reflect them when fetching from the model, if needed.

thomasdarimont:
I rebased the branch as discussed.

mabartos:
@thomasdarimont Sorry, another rebase is needed.

mabartos (commented):
@thomasdarimont Looks good to me from the first sight.

Just a consideration:
- We could even support to exclude some fields as providing `fields=!clientId,!auth` would return the whole representation without the `clientId` and `auth` fields.

WDYT?

thomasdarimont:
> We could even support to exclude some fields as providing fields=!clientId,!auth would return the whole representation without the clientId and auth fields.

Specifying exclusions via `?fields=!field1,!field2` can work, however if you have explicit projection `?fields=field1,field2` and explicit omission `?fields=!fieldA,!fieldB` how should those work together? I think we'd need a special symbol to express that we want to select all fields, exception the ones explicitly omitted. Perhaps something like  `?fields=ALL,!fieldA,!fieldB` might work.

sschu:
@thomasdarimont I would say it is either all included fields or all excluded fields but not a combination of both.

sschu:
I am wondering if the projection mechanism actually means all fields become optional in the OpenAPI spec (at least when reading from the server).

shawkins:
Closing as out of date. I should have referenced this PR in the initial projection MVP as it showed a similar Jackson based mechanism. However we don't believe that we'll need a Jackson based approach at this time.

PR #38830, opened 2025-04-10
Make sure that there is single audience allowed by default in JWT tok…

mposolda:
…ens sent to client authentication

closes #38819

Issue #38819, opened 2025-04-10
Make sure that there is single audience allowed by default in JWT tokens sent to client authentication


mposolda:
### Description

There are changes in the OIDC specification, which try to make sure that validation of the `audience` is more strict during the client authentication with JWT tokens. More context about this is in this issue: https://github.com/keycloak/keycloak/issues/38751 .

As a first for Keycloak 26.2 we would do:
- Make sure that JWT client authentication allows only single audience being in the JWT token by default.
- There would be single option at the server level, which will allow to have this check to be less strict and allow multiple audiences in the JWT token. But this option would be disabled by default
- We will document that in the Keycloak 26.2 release notes and upgrading guide. With the note that option would be probably removed in Keycloak 27

### Discussion

https://github.com/keycloak/keycloak/pull/38754 (It is the PR, but contains some discussion)

PR #41133, opened 2025-07-14
FAPI 2.0 Final - only accept its issuer identifier value as a string in the aud claim received in client authentication assertions

### LINKED ISSUE #41119: FAPI 2.0 Final - only accept its issuer identifier value as a string in the aud claim received in client authentication assertions
### Description

For Keycloak to fully comply with FAPI 2.0 Security Profile Final, Keycloak needs to do as follows:

 According to the item 8 of [section 5.3.2.1 of FAPI 2.0 Security Profile Final](https://openid.net/specs/fapi-security-profile-2_0-final.html#name-general-requirements) specification, 
  - On receiving a client authentication assertion from a client, Keycloak only accept its issuer identifier value as a string in the aud claim received in the client authentication assertion.
  - the issuer value is equal to the issuer claim value of the server metadata that the client can obtain by accessing well-known URI.

### Discussion

_No response_

### Motivation

Making Keycloak fully comply with FAPI 2.0 Security Profile Final specification.

### Details

_No response_

tnorimat:
closes #41119

### DIFF
--- a/services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientAuthenticationAssertionExecutor.java
+++ b/services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientAuthenticationAssertionExecutor.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.services.clientpolicy.executor;
+
+import jakarta.ws.rs.core.MultivaluedMap;
+import org.jboss.logging.Logger;
+import org.keycloak.OAuth2Constants;
+import org.keycloak.OAuthErrorException;
+import org.keycloak.common.util.Base64Url;
+import org.keycloak.models.ClientModel;
+import org.keycloak.models.KeycloakContext;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.representations.JsonWebToken;
+import org.keycloak.representations.idm.ClientPolicyExecutorConfigurationRepresentation;
+import org.keycloak.services.Urls;
+import org.keycloak.services.clientpolicy.ClientPolicyContext;
+import org.keycloak.services.clientpolicy.ClientPolicyException;
+import org.keycloak.util.JsonSerialization;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public class SecureClientAuthenticationAssertionExecutor implements ClientPolicyExecutorProvider<ClientPolicyExecutorConfigurationRepresentation> {
+
+    private static final Logger logger = Logger.getLogger(SecureClientAuthenticationAssertionExecutor.class);
+
+    private final KeycloakSession session;
+
+    public SecureClientAuthenticationAssertionExecutor(KeycloakSession session) {
+        this.session = session;
+    }
+
+    @Override
+    public String getProviderId() {
+        return SecureClientAuthenticationAssertionExecutorFactory.PROVIDER_ID;
+    }
+
+    @Override
+    public void executeOnEvent(ClientPolicyContext context) throws ClientPolicyException {
+        switch (context.getEvent()) {
+            case PUSHED_AUTHORIZATION_REQUEST:
+            case TOKEN_REQUEST:
+            case TOKEN_REFRESH:
+            case TOKEN_INTROSPECT:
+            case TOKEN_REVOKE:
+                validateClientAssertion();
+            default:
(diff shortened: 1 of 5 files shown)


tnorimat:
I confirmed that the Keycloak including the PR can pass FAPI 2.0 Final conformance tests.

tnorimat:
@mposolda If possible, could you check the PR?

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientAuthenticationAssertionExecutor.java
```diff
@@ -0,0 +1,113 @@
...
+            token = JsonSerialization.readValue(content, JsonWebToken.class);
+        } catch (IOException e) {
+            logger.warnf("client assertion parse error: %s", e.getMessage());
+            throw new ClientPolicyException(OAuthErrorException.INVALID_REQUEST, "invalid client assertion");
+        }
+
+        // Validate the client assertion audience
+        String issuerUrl = Urls.realmIssuer(session.getContext().getUri().getBaseUri(), session.getContext().getRealm().getName());
+        List<String> expectedAudiences = new ArrayList<>(Collections.singletonList(issuerUrl));
+        if (!token.hasAnyAudience(expectedAudiences)) {
```
mposolda, line 112: Does it makes sense to also verify here that `issuerUrl` is the only audience and there are not any other audiences included?

So maybe something like:
```
if (!token.hasAnyAudience(expectedAudiences) && token.getAudience().length == 1) {
```

In the past, there was security issue found in the private_key_jwt authentication that servers accepted the JWT with multiple audiences and same JWT token, which was used for client authentication, was also used for some other party. That is why the recent draft of OIDC core mentions explicitly single audience https://openid.net/specs/openid-connect-core-1_0-36.html#ClientAuthentication . The FAPI2 seems to not mention it explicitly for the server, however it indirectly mention this as well in the requirements for client, where it mentions - [The issuer identifier value shall be sent as a string not as an item in an array. .](https://openid.net/specs/fapi-security-profile-2_0-final.html#section-5.3.3.1) .

I am not 100% sure if it is possible to reliably check that `aud` was sent as a string and not an array (also there is likely not any security benefit in that). However for better security, it makes sense to me to check that the issuer is single audience and not other audiences are included. WDYT?

mposolda (commented):
@tnorimat Thanks! Added a comment. Could you please check it?

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientAuthenticationAssertionExecutor.java
tnorimat: Thank you for your comments.

As you suggested, I will add the "aud" check.

I know the secuirty issue about specification and you have resolved that.
https://github.com/keycloak/keycloak/issues/38819
https://github.com/keycloak/keycloak/pull/38830

OIDF FAPI 2.0 Final conformance tests include the test for check whether "aud" claim value is single.
So, when running the conformance tests, I do the following settings.

> spi-login-protocol-openid-connect-allow-multiple-audiences-for-jwt-client-authentication=true

When a keycloak user tries to apply FAPI 2.0 Final security profile, they need to do the setting, which checks the "aud" single value check, so I omitted the same check in the PR.

However, if the PR includes the "aud" check, then the user need not do the setting because the PR itself checks that. It might be helpful for the user.

In order to do that, what about the following?

```
        if (token.getAudience().length != 1) {
            logger.warnf("invalid audience in client assertion - no audience or multiple audiences found");
            throw new ClientPolicyException(OAuthErrorException.INVALID_REQUEST, "invalid audience in client assertion");
        }
        if (!token.hasAnyAudience(expectedAudiences)) {
            logger.warnf("invalid audience in client assertion - audience not issuer URL");
            throw new ClientPolicyException(OAuthErrorException.INVALID_REQUEST, "invalid audience in client assertion");
        }
```

tnorimat: I modified the PR as just above and confirmed that Keycloak incoporating the PR can pass FAPI 2.0 Security Profile Final and FAPI 2.0 Message Singing Final conformance tests with both the following settings (current Keycloak's default setting):
```
spi-login-protocol-openid-connect-allow-multiple-audiences-for-jwt-client-authentication=true
```
and
```
#spi-login-protocol-openid-connect-allow-multiple-audiences-for-jwt-client-authentication=true
```
The PR was force-pushed.

mposolda: @tnorimat Probably a detail, but I afraid a bit of NPE when there are not any audience in the token as in this case `token.getAudience().length` might probably throw NPE? Is it ok to rather change the condition for length to `if (token.getAudience() == null || token.getAudience().length != 1)` ?

Another option is to make sure that length is checked after `token.hasAudience` call (as `token.hasAudience` would fail in case that token does not have any audience and hence no chance of NPE).

mposolda (commented):
@tnorimat Thanks, one last minor comment (to possibly prevent NPE)

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientAuthenticationAssertionExecutor.java
tnorimat: @mposolda Thank you. I will revise the PR as you suggested.

VinodAnandan:
@mposolda @tnorimat Would it be possible to back-port this PR to the release/26.3 branch, please ?

tnorimat:
@mposolda I revised the PR as you suggested, and I am running the following OIDF conformance tests:
- FAPI 2.0 Security Profile Final
- FAPI 2.0 Message Signing Final
 I would appreciate it if you could re-check the PR.

mposolda (approved):
@tnorimat Thanks!

mposolda:
@VinodAnandan I afraid not. As you know, we are backporting mostly important bugfixes. And this issue does not fall into that category.

PR #41142, opened 2025-07-14
fix: using unversioned key in docs

### LINKED ISSUE #40959: Update "Enabling and disabling features" documentation
### Description

I'm currently building a Keycloak image and I wanted to disable some features.

Then I noticed that features like `account-v3` and `token-exchange-standard-v2` seem to no longer exist (at least as a feature flag), which means to me that the below documentation is out of date, or am I missing something?

Documentation: https://www.keycloak.org/server/features

```
Invalid value for option '--features-disabled': account-v3. Expected values are: account, account-api, admin, admin-api, 
admin-fine-grained-authz, authorization, cache-embedded-remote-store, ciba, client-policies, client-secret-rotation, 
client-types, clusterless, declarative-ui, device-flow, docker, dpop, dynamic-scopes, fips, impersonation, 
ipa-tuura-federation, kerberos, login, multi-site, oid4vc-vci, opentelemetry, organization, par, passkeys, 
persistent-user-sessions, preview, quick-theme, recovery-codes, rolling-updates, scripts, step-up-authentication, 
token-exchange, token-exchange-standard, transient-users, update-email, user-event-metrics, web-authn.
Invalid value for multivalued option '--features-disabled': list value 'token-exchange-standard-v2 ' 
should not have leading nor trailing whitespace
```

I'm on Keycloak version `26.2.5`.

Thanks

shawkins:
closes: #40959

### DIFF
--- a/docs/guides/server/features.adoc
+++ b/docs/guides/server/features.adoc
@@ -81,7 +81,7 @@ There are no deprecated features in this {project_name} release.
 
 <#macro showFeatures features>
 <#list features as feature>
-[.features-name]#${feature.name}#::
+[.features-name]#${feature.versionedKey}#::
 [.features-description]#${feature.description}#
 </#list>
 </#macro>
--- a/docs/maven-plugin/src/main/java/org/keycloak/guides/maven/Features.java
+++ b/docs/maven-plugin/src/main/java/org/keycloak/guides/maven/Features.java
@@ -51,6 +51,10 @@ public class Features {
             return profileFeature.getLabel();
         }
 
+        public String getVersionedKey() {
+            return profileFeature.getVersionedKey();
+        }
+
         private Profile.Feature.Type getType() {
             return profileFeature.getType();
         }


mabartos (commented):
@shawkins LGTM as a basic fix for the issue, but we should probably improve it a little bit more. 

Website form:
image


PR with changes:
image

It'd be good if we could find connected features (`Profile.getFeatureVersions(feature)` already present), and if there's only first version of the feature, do not show the `:v1` suffix as it might worse the readability as users would think they need to provide even the version to enable the feature. 

For the case above, I'd like to see only:
```
account-api
account:v3
admin-api
admine-fine-grained-authz:v2
admin:v2
authorization
```

@shawkins IMO, it'd improve it a little bit more. WDYT?

shawkins:
> IMO, it'd improve it a little bit more. WDYT?

No, I don't think that helps. If users want to lock in a specific version of a feature (for future proofing their options), they should see that clearly.

mabartos (approved):
> If users want to lock in a specific version of a feature (for future proofing their options), they should see that clearly.

@shawkins There might be two ways of looking at that, and related to the matter of preferences:

1. Whether we want to simplify things for users and advertise the feature name without the version if there's only one version.

OR

2. Whether we want to avoid future creation of issues by users/customers that their deployment started to use the newest version of the feature, and the deployment is broken for them now. So advertising the whole key+version will help to avoid that.  

Probably the second option would be better, so I agree with the current form of this PR. Thanks for the additional consideration.

ahus1 (approved):

PR #41144, opened 2025-07-14
fix: adding a release note for http-management-scheme

### LINKED ISSUE #40945: Unclear documentation for setting management server as http when main server is https

shawkins:
closes: #40945

### DIFF
--- a/docs/documentation/release_notes/index.adoc
+++ b/docs/documentation/release_notes/index.adoc
@@ -13,6 +13,9 @@ include::topics/templates/document-attributes.adoc[]
 :release_header_latest_link: {releasenotes_link_latest}
 include::topics/templates/release-header.adoc[]
 
+== {project_name_full} 26.4.0
+include::topics/26_4_0.adoc[leveloffset=2]
+
 == {project_name_full} 26.3.0
 include::topics/26_3_0.adoc[leveloffset=2]
 
--- a/docs/documentation/release_notes/topics/26_4_0.adoc
+++ b/docs/documentation/release_notes/topics/26_4_0.adoc
@@ -1,0 +1,8 @@
+// Release notes should contain only headline-worthy new features,
+// assuming that people who migrate will read the upgrading guide anyway.
+
+Read on to learn more about each new feature, and https://www.keycloak.org/docs/latest/upgrading/index.html[find additional details in the upgrading guide] if you are upgrading from a previous release of {project_name}.
+
+= Option to force management interface to use HTTP.
+
+There's a new option `http-management-scheme` that may be set to `http` to force the management interface to use HTTP rather than inheriting the HTTPS settings of the main interface. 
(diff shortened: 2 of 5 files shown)


mabartos (changes requested):
@shawkins `26_4_0.adoc` should be the correct file (as there's no plan to backport it to 26.3.x anyway, right?). Otherwise LGTM.

shawkins:
Corrected the release note and realized we needed operator handling as well. Just like all of the other https overrides for the management interface you would need to control the scheme via an additional option. If we want a first-class option, just let me know.

cc @vmuzikar

mabartos (approved):
> Corrected the release note and realized we needed operator handling as well. Just like all of the other https overrides for the management interface you would need to control the scheme via an additional option. 

Thanks! And good spot of required changes for Operator. 

> If we want a first-class option, just let me know.

ATM, we expose only `port` of the mgmt interface as a first-class citizen. Other props can be defined only via the additional options. I think it's not really necessary to provide first-class citizen for the scheme, as it's not really a thing that would affect the majority of deployments. If you decide to provide it as a first-class citizen, IMHO, we'd need to provide some more docs on this.

LGTM as is right now. Maybe it'd be good to create a separate issue for fixing the Operator logic as a follow-up on changes to the server and link it to this PR for better "addressability" of issues.

shawkins:
>  Maybe it'd be good to create a separate issue for fixing the Operator logic as a follow-up on changes to the server and link it to this PR for better "addressability" of issues.

I'm fine with leaving it all under a single issue given the scope of the changes and the coupling - if you do try to use this new additional option in the operator without these changes, it would break your instance.

mabartos:
> I'm fine with leaving it all under a single issue given the scope of the changes and the coupling - if you do try to use this new additional option in the operator without these changes, it would break your instance.

Ok, agree. 

@vmuzikar We'd need to have one more approval for this. Could you handle it? Thanks :)

vmuzikar (approved):
LGTM, thanks.

PR #41145, opened 2025-07-14
Allow mapping Admin roles to server administrator only

### LINKED ISSUE #39956: Allow mapping Admin roles to server administrator only
For delegating access to user management to a realm admin, server admin might create following permissions

Allow:
- `view` `map-roles` and `manage` all realm users
- `view` and `map-roles` from all clients

If a server administrator wants to prevent managing other realm administrators, it can be achieved by disallowing 

- `manage` and `map-roles` for 'some_admin' (or group of admins)

Aforementioned setup allows to assign "manage-users" role to yourself and effectively get permission to manage 'some_admin', bypassing the intended restriction.

It doesn't make much sense to allow mapping admin roles to delegated realm admins. This issue is about changing the behavior to allow map admin roles for server admin, not delegated realm admins.

vramik:
Closes #39956

### DIFF
--- a/docs/documentation/release_notes/index.adoc
+++ b/docs/documentation/release_notes/index.adoc
@@ -13,6 +13,9 @@ include::topics/templates/document-attributes.adoc[]
 :release_header_latest_link: {releasenotes_link_latest}
 include::topics/templates/release-header.adoc[]
 
+== {project_name_full} 26.2.6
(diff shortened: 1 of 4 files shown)


ahus1 (approved):
Approving based on parent issues's reviews.

PR #41163, opened 2025-07-15
Implement CompatibilityMetadataProvider for Cache CLI args

### LINKED ISSUE #41138: Implement CompatibilityMetadataProvider for Cache CLI args
### Description

Implement CompatibilityMetadataProvider for Cache CLI args. Changes to the following options should result in a cluster recreate:

- `--cache`
- `--cache-config-file`
- `--cache-stack`
- `--cache-embedded-mtls-enabled`
- `--cache-remote-host`
- `--cache-remote-port`

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #41138

Requires https://github.com/keycloak/keycloak/pull/41027

A limitation with the current implementation is that if a default value is not set for a CLI arg then explicitly setting the value will be classes as "incompatible" even though the semantics are the same. For example, the default cache configuration uses the "jdbc-ping" stack, however `--cache-stack jdbc-ping` is not set. So if the user executes the update command with `--cache-stack jdbc-ping` explicitly configured then a RECREATE exit code will be returned by the `UpdateCompatibility` command.

Given that we state the following in the docs, ^ limitation could be considered acceptable:

```quote
Ensure that all configuration options, whether set via environment variables or CLI arguments, are included when running this command.
```


model/infinispan/src/main/java/org/keycloak/compatibility/AbstractCompatibilityMetadataProvider.java
```diff
@@ -0,0 +1,48 @@
...
+public abstract class AbstractCompatibilityMetadataProvider implements CompatibilityMetadataProvider {
+
+    final String spi;
+    final boolean enabled;
+    final Config.Scope config;
+
+    public AbstractCompatibilityMetadataProvider(String spi, boolean enabled) {
+        this.spi = spi;
+        this.enabled = enabled;
+        this.config = enabled ? Config.scope(spi, "default") : null;
```
pruivo: Not all providers have the ID `default`. I would add the `providerId` as a parameter to make this class future-proof. 

For example, this factory has SPI optoins and the provider ID is not `default`:

https://github.com/keycloak/keycloak/blob/9322d71d61390a3aa833c18fc89fe676fb066df1/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java#L313

model/infinispan/src/main/java/org/keycloak/compatibility/AbstractCompatibilityMetadataProvider.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public Map<String, String> metadata() {
+        if (!enabled)
```
pruivo: I would remove the `enabled` field and create an abstract method `isEnabled(Config.Scope)`.

The provider factory API has the `Config.Scope` as a parameter; it can use an SPI option/configuration to decide if it is enabled or not. 

https://github.com/keycloak/keycloak/blob/3e0a1850708b9f1a07e3292c79ef8c84744f2fc1/server-spi-private/src/main/java/org/keycloak/provider/EnvironmentDependentProviderFactory.java#L30-L36

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
```diff
@@ -0,0 +1,32 @@
...
+public class CachingEmbeddedMetadataProvider extends AbstractCompatibilityMetadataProvider {
+
+    public CachingEmbeddedMetadataProvider() {
+        super(CacheEmbeddedConfigProviderSpi.SPI_NAME, InfinispanUtils.isEmbeddedInfinispan());
+    }
+
+    @Override
+    public Map<String, String> meta() {
+        return Map.of(
+              "persistence", Boolean.toString(MultiSiteUtils.isPersistentSessionsEnabled()),
```
pruivo: It can be removed as the `Feature` is already tracked.

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingRemoteMetadataProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class CachingRemoteMetadataProvider extends AbstractCompatibilityMetadataProvider {
+
+    public CachingRemoteMetadataProvider() {
+        super(CacheRemoteConfigProviderSpi.SPI_NAME, InfinispanUtils.isRemoteInfinispan());
+    }
+
+    @Override
+    public Map<String, String> meta() {
+        return Map.of("persistence", Boolean.toString(MultiSiteUtils.isPersistentSessionsEnabled()));
```
pruivo: hmm, I think we can remove it here too.

model/infinispan/src/main/java/org/keycloak/compatibility/AbstractCompatibilityMetadataProvider.java
```diff
@@ -0,0 +1,48 @@
...
+        });
+        return metadata;
+    }
+
+    @Override
+    public String getId() {
+        return spi;
+    }
+
+    protected Map<String, String> meta() {
```
pruivo: nitpicking
```suggestion
    protected Map<String, String> customMetadata() {
```

model/infinispan/src/main/java/org/keycloak/compatibility/AbstractCompatibilityMetadataProvider.java
```diff
@@ -0,0 +1,48 @@
...
+            String value = config.get(key);
+            if (value != null)
+                metadata.put(key, value);
+        });
+        return metadata;
+    }
+
+    @Override
+    public String getId() {
+        return spi;
```
pruivo, line 38: I'm not sure if this should be a combination of SPI and the Provider ID. For example, `InfinispanUserSessionProviderFactory` and `RemoteUserSessionProviderFactory` both implement the same SPI.
But let's leave it as is for now.

pruivo (commented):
Overall, it looks good to me. I left some suggestions anticipating some future work around the SPI options.

model/infinispan/src/main/java/org/keycloak/compatibility/AbstractCompatibilityMetadataProvider.java
ryanemerson: Sure. I was undecided about this as we don't require it yet, but no harm to future proof.

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
ryanemerson: Good point, I hadn't considered that

ryanemerson:
Thanks for the review @pruivo, I believe I have addressed all comments.

pruivo (approved):

ahus1 (approved):

PR #41190, opened 2025-07-16
Password modification time attribute as an operational and read-only …

vramik:
…attribute

Closes #40270


ahus1 (approved):
Approving based on parent PR's review

PR #41194, opened 2025-07-16
Improve handling of datasource name specified in `persistence.xml` files

### LINKED ISSUE #41192: Improve handling of datasource name specified in `persistence.xml` files
### Description

Right now, we're accepting only the property `hibernate.connection.datasource` as the main datasource name that is specified in `persistence.xml` files for non-default datasources (see https://github.com/keycloak/keycloak/blob/main/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java#L402).

The property is deprecated and should not be used. Even though there's an initiative to potentially get rid of `persistence.xml` files and rather use Quarkus properties for dealing with next persistence units, the `persistence.xml` files will be kept in Keycloak for some time, so we need to handle it for that time and even for a deprecation period.

The suggested approach for resolving the datasource name would be:

| Order |Property                    |
|-------|------------------------------------------------------------|
| 1.    | `jakarta.persistence.jtaDataSource` |
| 2.    | `hibernate.connection.datasource`                         |
| 3.    | Persistence unit name                                     |

Example of `persistence.xml` preferences:

```xml
<persistence>
    <persistence-unit name="user-store"> <!-- 3. in order -->
        <class>org.sth.here.Test</class>
        <properties>
            <property name="jakarta.persistence.jtaDataSource" value="user-store"/> <!-- 1. in order -->
            <property name="hibernate.connection.datasource" value="user-store" /> <!-- 2. in order -->
        </properties>
    </persistence-unit>
</persistence>
```
EDIT: We require JTA tx type for these datasources.

Then, we can automatically set the datasource name for the related properties for Jakarta Persistence/Hibernate.

Touched on here: https://github.com/keycloak/keycloak/pull/41087#discussion_r2207267160

cc: @keycloak/cloud-native @shawkins

mabartos:
Closes #41192
Blocks https://github.com/keycloak/keycloak/pull/41087
Blocks https://github.com/keycloak/keycloak/pull/41223

This should handle consistently the datasource name needed for handling additional datasources options. 

I can provide some parsing tests to verify the order, but might not be necessary due to straightforward checks.

### DIFF
--- a/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
+++ b/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
@@ -53,6 +53,8 @@ import io.quarkus.vertx.http.deployment.NonApplicationRootPathBuildItem;
 import io.quarkus.vertx.http.deployment.RouteBuildItem;
 import jakarta.persistence.Entity;
 import jakarta.persistence.PersistenceUnitTransactionType;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.hibernate.cfg.JdbcSettings;
 import org.eclipse.microprofile.health.Readiness;
 import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.jpa.boot.internal.ParsedPersistenceXmlDescriptor;
@@ -154,6 +156,7 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.jar.JarEntry;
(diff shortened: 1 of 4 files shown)


quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -425,6 +477,20 @@ void produceDefaultPersistenceUnit(BuildProducer<PersistenceXmlDescriptorBuildIt
         producer.produce(new PersistenceXmlDescriptorBuildItem(descriptor));
     }
 
+    private void configurePersistenceUnitProperties(String datasourceName, ParsedPersistenceXmlDescriptor descriptor) {
```
mabartos: Will be used for other properties as mentioned here: https://github.com/keycloak/keycloak/compare/main...shawkins:keycloak:pu

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -364,6 +369,52 @@ void setDefaultUserProfileConfig(KeycloakRecorder recorder, UserProfileBuildItem
...
+                        "Use '%s' name for datasource options like 'db-kind-%s'.", name, source, name, name);
+
+        String persistenceUnitName = descriptor.getName();
+        Properties properties = descriptor.getProperties();
+
+        // 1. return Jakarta properties
+        @SuppressWarnings("removal") // Once we remove persistence.xml support, we can remove it as well
+        var jakartaPropertyName = switch (descriptor.getTransactionType()) {
+            case JTA -> JAKARTA_JTA_DATASOURCE;
+            case RESOURCE_LOCAL -> JAKARTA_NON_JTA_DATASOURCE;
```
shawkins: This looks good overall, but I don't know if we should support RESOURCE_LOCAL as it doesn't seem to be well supported in quarkus (https://github.com/quarkusio/quarkus/issues/2628) and moving forward to the hibernate options there's no replacement for that. It will always be JTA.

mabartos: Yep, you're right, we should use only JTA. The funny thing is that git blame showed me I was the one who made these changes to use explicitly JTA for the default DS xD

mabartos: Updated.

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -366,6 +369,47 @@ void setDefaultUserProfileConfig(KeycloakRecorder recorder, UserProfileBuildItem
...
+     * Get datasource name obtained from the persistence.xml file based on this order:
+     * <ol>
+     *      <li> return {@link JdbcSettings#JAKARTA_JTA_DATASOURCE} if specified
+     *      <li> return {@link AvailableSettings#DATASOURCE} property if specified
+     *      <li> return persistence unit name
+     * </ol>
+     * Can be removed after removing support for persistence.xml files
+     */
+    static String getDatasourceNameFromPersistenceXml(PersistenceUnitDescriptor descriptor) {
+        if (descriptor == null) {
```
shawkins, line 382: Just a nit. A null check isn't required.

shawkins (approved):
LGTM

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
mabartos: @shawkins It's just a habit to be on the safe side when the method is used elsewhere. Thanks for the review.

Issue #39440, opened 2025-05-05
Keycloak realm import breaks value of template attribute of UsernameTemplateMapper


DerKanzler:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

import-export

### Describe the bug

We use an external identity provider which uses colons in the name of a claim. After upgrading to Keycloak 26.1.4 (from 25.0.6) the template definition in UsernameTemplateMapper is broken after a realm import.

I believe it might have to do with the changes in https://github.com/keycloak/keycloak/commit/c1c147cb17897aee537ec2781eef819cd684e074 or https://github.com/keycloak/keycloak/commit/cb3954fc7bfa5662a339d6bdf29ba9d74c96ff7b. But I am not 100% sure.

Now the value gets interpolated and the colon gets interpreted as a separator for the default value.
Previously the value remained mainly untouched apart from the handling of the escape character for the dot.
And this value may not be changed since the UsernameTemplateMapper needs the entire String in template attribute including the `$`.
If there is no colon in the attribute it still works. But I assume it's just because the property is not defined in the environment.

### Version

26.1.4

### Regression

- [x] The issue is a regression

### Expected behavior

Value of `template` attribute stays unchanged after an import (apart from the handling of the dot).
```
  "identityProviderMappers" : [
    {
      "id" : "9c3726be-81d1-41e8-876f-1fe97adad904",
      "name" : "broker_id",
      "identityProviderAlias" : "id-austria",
      "identityProviderMapper" : "oidc-username-idp-mapper",
      "config" : {
        "syncMode" : "IMPORT",
        "template" : "${CLAIM.urn:pvpgvat:oidc\\.bpk}",
        "target" : "BROKER_ID"
      }
    },
    {
      "id" : "7bbef759-e7bf-4fa9-a043-4e033a7adb6f",
      "name" : "broker_username",
      "identityProviderAlias" : "id-austria",
      "identityProviderMapper" : "oidc-username-idp-mapper",
      "config" : {
        "syncMode" : "IMPORT",
        "template" : "${CLAIM.urn:pvpgvat:oidc\\.bpk}",
        "target" : "BROKER_USERNAME"
      }
    }
  ],
```
Results to in the UI:
![Image](https://github.com/user-attachments/assets/68ef15a1-7ad8-45a2-a82b-12a5e16d6af2)

### Actual behavior

[...]

PR #41196, opened 2025-07-16
fix: cutting down on the memory footprint for import

shawkins:
Changes the property replacement logic to support streaming. The corner case is an unclosed ${ - there is a TODO to address that with a limit on how large the key contents can be.

After reviewing more of the logic this is the only simple fix we can do. Any other changes to the import logic will require two passes over the stream - which gets messy / expensive for multiple realms in a file, or from doing the import over the admin api.

closes: #40875

### DIFF
--- a/common/src/main/java/org/keycloak/common/util/StringPropertyReplacer.java
+++ b/common/src/main/java/org/keycloak/common/util/StringPropertyReplacer.java
@@ -16,8 +16,16 @@
  */
 package org.keycloak.common.util;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.Optional;
+
+import org.jboss.logging.Logger;
+import org.jboss.logging.Logger.Level;
 
 /**
  * A utility class for replacing properties in strings.
@@ -32,6 +40,8 @@ import java.util.Optional;
 public final class StringPropertyReplacer
 {
 
+    private static final Logger logger = Logger.getLogger(StringPropertyReplacer.class);
+
     /** File separator value */
     private static final String FILE_SEPARATOR = File.separator;
 
@@ -44,13 +54,10 @@ public final class StringPropertyReplacer
     /** Path separator alias */
     private static final String PATH_SEPARATOR_ALIAS = ":";
 
-    // States used in property parsing
-    private static final int NORMAL = 0;
-    private static final int SEEN_DOLLAR = 1;
-    private static final int IN_BRACKET = 2;
-
     private static final PropertyResolver NULL_RESOLVER = property -> null;
     private static PropertyResolver DEFAULT_PROPERTY_RESOLVER;
+
+    private static final int MAX_KEY_LENGTH = 1<<22;
 
     public static void setDefaultPropertyResolver(PropertyResolver systemVariables) {
         DEFAULT_PROPERTY_RESOLVER = systemVariables;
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/keycloak/common/util/StringPropertyReplacer.java
```diff
@@ -99,138 +99,127 @@ public static String replaceProperties(final String string) {
...
+                    return read();
+                }
+                // determine the key
+                int keyChar = -1;
+                InputStream keyStream = replaceProperties(source, true, resolver);
+                ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+                while ((keyChar = keyStream.read()) > -1) {
+                    bytes.write((byte)keyChar);
+                    // TODO: have a byte limit to prevent runaway consumption
+                }
```
mabartos: Byte limit is one thing, but maybe adding a recursion barrier for the `replaceProperties()` might also be good. Setting a limit on how far the recursion can go. Just to be sure. WDYT?  

The second call for `replaceProperties` catches the StackOverflow exception, so at least add the check there if we don't provide the byte limit?

Do we have some possible estimation on how big byte limit would be reasonable ?

mabartos (commented):
@shawkins Thanks for the PR, nice work! Just went briefly over the code and put one comment. 

It'd be also good to have some results on the improved memory footprint (or at least verify it really helped :)). Would you have a chance to try it with some profiler, or do you want me to test it?

common/src/main/java/org/keycloak/common/util/StringPropertyReplacer.java
shawkins: > but maybe adding a recursion barrier for the replaceProperties() might also be good. Setting a limit on how far the recursion can go. Just to be sure. WDYT?

It is better to have a depth rather than just catch the overflow yes, but the risk to the overall vm isn't that bad since it will be limited to a specific stack.

The risk from not supporting a byte limit is that ${ naturally appearing unclosed in the doc will cause the remaining bytes to be read into memory.

If we don't think this is likely to occur, then it may not be worth worrying about as our logic in general isn't protecting against memory attacks in realm files / uploads.

> Do we have some possible estimation on how big byte limit would be reasonable ?

In theory it shouldn't be that large. Most env keys won't be that large. However you can certainly create something with multiple large keys and a large default value that takes up quite a few characters.

shawkins:
Moved to draft to address the close method. Also it would be better to use a Reader, rather than an InputStream, but that further complicates the implementation.

mabartos (approved):
@shawkins Great job! LGTM

We should have a good test coverage for the import and possible replacements. Moreover, I did some experiments to verify the functionality - everything seems fine.

I see that the same `StringPropertyReplacer` is used in various projects, so it creates some sort of convention for the property replacement in strings. 

Projects:
- JBoss Narayana JTA
- Infinispan (cc: @ryanemerson @pruivo)
- WildFly Elytron

It might be beneficial for these projects (and us) to also have unified approach as we could also get more eyes on this and get more reviews.

@shawkins Just a consideration. WDYT?    

I'd like to see also approval from the @keycloak/core team.

ryanemerson:
> It might be beneficial for these projects (and us) to also have unified approach as we could also get more eyes on this and get more reviews.

+1. It seems Infinispan would benefit from these changes \cc @tristantarrant

shawkins:
> @shawkins Just a consideration. WDYT?

It's certianly possible to shoot for a common utility, but there are a few additional considerations:

- The implementation lacks a way to escape the replacement, which is something we may need to start considering for both export and all import scenarios https://github.com/keycloak/keycloak/issues/39440

- The changes here are because we are doing property replacement on rather large files that are being imported (and may, but are currently not, perform replacements on realm operations via the admin api). And while compatible with and adaptable to our other needs, this may be the right approach in general. The replacements are context free, so you are free to use this utility for any file type, but that may not be useful / intended. Do we really want to allow inlining additional json in a json file? Or should this have been done only at a level that is still structurally valid - as is shown in the docs https://www.keycloak.org/server/importExport#_using_environment_variables_within_the_realm_configuration_files If it's just the latter, then we could forgo the global processing prior to parsing and instead build the property replacement into the jackson deserialization. The global handling has other side effects, such as presuming UTF-8 encoding.

tristantarrant:
> > It might be beneficial for these projects (and us) to also have unified approach as we could also get more eyes on this and get more reviews.
> 
> +1. It seems Infinispan would benefit from these changes \cc @tristantarrant

Infinispan only uses this when parsing configuration files which generally tend to be fairly small (<10KB), so the benefits would be quite low.

mabartos:
Thanks for all these comments.

So we know Keycloak will benefit from this, but other projects might not, so it might not worth the effort to try to include it to these projects. 

> Do we really want to allow inlining additional json in a json file? Or should this have been done only at a level that is still structurally valid - as is shown in the docs

@shawkins Do we? AFAIK we only support property replacement based on the used property resolver as it's usually ENV vars, possibly system props, or some simple custom ones, right?. I would not support inlining additional JSON in a JSON file as it might introduce some other issues as we could provide users/customer more freedom in this area and not sure if we(maintanance) or they need it so much. If they need to do some more advanced things, they can do it on their own. Or have you encountered some request for such a feature? Or can you clarify it a little bit? Thanks!

mabartos:
Can we get someone from @keycloak/core to review it? (or @keycloak/sre when you might have some experience in this matter?)

Thanks!

shawkins:
> Or can you clarify it a little bit? Thanks!

The previous/current logic happens before any structural parsing - it doesn't matter what the String content is - xml, json, yaml, etc. it just looks for ${...}. So for example in a json file you could have:

```
[
  ${foo}
]
```

And in an environment variable set foo=`{"a"},{"b"}`, then the string replacement logic will effectively inline that json content into the document.

This is powerful and users may not be aware that it can be done. If we don't want to allow this, then I will create a follow up issue to do string replacement after the json parsing. This would probably look something like the rules for openshift templates: https://docs.redhat.com/en/documentation/openshift_container_platform/4.9/html/images/using-templates - it additionally has special handling for ${{...}} replacements.

mabartos:
> The previous/current logic happens before any structural parsing - it doesn't matter what the String content is - xml, json, yaml, etc. it just looks for ${...}

Ahh, that's true. 

> And in an environment variable set foo={"a"},{"b"}, then the string replacement logic will effectively inline that json content into the document.

I'm not very comfortable supporting this as it raises a risk of unintentional changes, as possible injection attacks can be done - as the realm setting is dependent on the environment (machine). I know in order to execute that kind of attack, the attacker needs to have access to the machine itself, so some more harmful stuff can be done, but it'd be good to lower the number of possibilities :)  

> This is powerful and users may not be aware that it can be done. If we don't want to allow this, then I will create a follow up issue to do string replacement after the json parsing.

But I believe there are some folks who already use this. I'm on top of the fence right now - allow or not to allow, that's the que... We should consider the effort behind the analysis and needed work for this. So it'd need some analysis on how attacker can leverage that functionality, how harmful it could get, decide whether it worths the effort to limit devs with possible tiny benefits,... 

Maybe it'd be good to limit it by default, but having the possibility to "unblock" the limitation and do it on your own risk. But still, not sure if it worths the effort and bring some more benefits (at least it could strengthen our secure by default principles).

@shawkins What's your opinion on this?

shawkins:
> @shawkins What's your opinion on this?

My guess is that it's an undocumented / unintended consequence of the original implementation, but it is entirely possible that someone figured out that it was possible and did something clever like have composable realm imports - where common elements could be shared via environment variables. Supporting such a scheme post parsing is simply not possible because the original file will not be structurally valid, so if we allow that style of replacement we'll have to keep around this existing code path in some form.

Moved to a discussion: https://github.com/keycloak/keycloak/discussions/41611

abstractj (approved):

PR #41199, opened 2025-07-16
Separate support for persistence.xml from KeycloakProcessor

mabartos:
`KeycloakProcessor` is already full of logic for supporting `persistence.xml` files. We should separate these concepts to have better manageability of those. Once we decide to remove the support for `persistence.xml` files in the future, the removal will be easier.

cc: @keycloak/cloud-native


mabartos:
Keep on hold with this initiative as there's not much benefit to having it here.

PR #41211, opened 2025-07-16
Only view svc acct users on exact search

### LINKED ISSUE #41103: Service Account users now showing in the User List
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Since version 26.3, the service accounts do show up in the User list of the UI. 
Is this intended or is this an undocumented change ? I didn't see it in the change log of 26.3

### Version

26.3.1

### Regression

### Expected behavior

I do not see the service account in the list and without knowledge of its id I'm unable to add or remove roles from the service account.

### Actual behavior

I can see the service account and manage it.

### How to Reproduce?

Step 1 : Start a realm with keycloak version 26.3.1
Step 2: create a client with client authorization and service account enabled
Step 3: Log in as a user with view-users and manage-users role.

User List

Service account role mapping

### Anything else?

_No response_

ssilvert:
Fixes #41103 

Search for service account users was accidentally triggered when we added the `exact` search feature.  But including that parameter (true or false) in the search would trigger a search that includes service account users.

Now you only see service account users if you are really doing an exact search for it.

### DIFF
--- a/js/apps/admin-ui/src/components/users/UserDataTable.tsx
+++ b/js/apps/admin-ui/src/components/users/UserDataTable.tsx
@@ -126,7 +126,7 @@ export function UserDataTable() {
   const [searchType, setSearchType] = useState<SearchType>("default");
   const [searchDropdownOpen, setSearchDropdownOpen] = useState(false);
   const [activeFilters, setActiveFilters] = useState<UserFilter>({
-    exact: true,
+    exact: false,
     userAttribute: [],
   });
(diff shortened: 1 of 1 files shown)


jonkoops (approved):

ahus1:
@ssilvert - do you plan to backport this? The parent issue doesn't have a backport label, yet.

ssilvert:
> @ssilvert - do you plan to backport this? The parent issue doesn't have a backport label, yet.

@ahus1 Yes.

PR #41212, opened 2025-07-17
Fix typo for MDC enabled property in logging guide

mabartos:
@ahus1 Could you please check it? Thanks


ahus1 (approved):
Thank you for reviewing this and providing a fix!

PR #41227, opened 2025-07-17
[TEST] Fix codebase for Liquibase 4.33.0

mabartos:
DO NOT MERGE

We need to find out if the migration tests will pass or not.


mabartos:
Yes, we found the cause.

Migration tests passed: https://github.com/keycloak/keycloak/actions/runs/16346388970/job/46182070399?pr=41227

PR #41230, opened 2025-07-17
Remove obsolete code for the Liquibase LogHistoryService

### LINKED ISSUE #41229: Remove obsolete code for the Liquibase LogHistoryService
### Description

At this moment, for MySQL and MariaDB, we use the [CustomChangeLogHistoryService](https://github.com/keycloak/keycloak/blob/main/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/CustomChangeLogHistoryService.java) that contains overridden `getRanChangeSets()` method, that was added there due to historical reasons to fix the [problem](https://github.com/keycloak/keycloak/pull/7942) with MySQL 8+.

There's also a check that use reflection for the `databaseChecksumsCompatible` field, which is publicly accessible in the newer versions of Liquibase. 

The code has not been updated since 04/2021 and contains very old assumptions/conversions to fix ancient bug here: https://github.com/keycloak/keycloak/pull/7942#discussion_r615027056

We should use the method from the parent StandardChangeLogHistoryService.

It seems the time caught us as after upgrade to Liquibase 4.33.0, the ran change sets is null and possibly cause of the: https://github.com/keycloak/keycloak/issues/41228

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mabartos:
- Closes #41229
- Closes https://github.com/keycloak/keycloak/issues/41228

Test run passed for the quarkus-next and Liquibase 4.33.0: https://github.com/keycloak/keycloak/actions/runs/16346388970/job/46182070399?pr=41227

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/CustomChangeLogHistoryService.java
+++ b/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/CustomChangeLogHistoryService.java
@@ -16,30 +16,12 @@
  */
 package org.keycloak.connections.jpa.updater.liquibase.conn;
 
-import org.keycloak.common.util.reflections.Reflections;
-import java.lang.reflect.Field;
-import java.text.DateFormat;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import liquibase.ContextExpression;
-import liquibase.Labels;
-import liquibase.change.CheckSum;
 import liquibase.change.ColumnConfig;
-import liquibase.changelog.ChangeSet;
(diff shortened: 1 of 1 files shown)


mabartos:
@rmartinc Could you please check this? Thanks!

It's necessary for fixing this: https://github.com/keycloak/keycloak/issues/41228

rmartinc (approved):
@mabartos If I understood this correctly this `CustomChangeLogHistoryService` was introduced to fix an old issue in liquibase for myql. If the issue is not there anymore I see perfect removing the code. So now this custom class is just used to create the PK. Maybe we can check later if the PK can be created more easily. But the change is OK to me. Approving.

mabartos:
> If I understood this correctly this CustomChangeLogHistoryService was introduced to fix an old issue in liquibase for mysql. If the issue is not there anymore I see perfect removing the code. 

@rmartinc Yes, exactly. Thanks for the review. 

Keeping that code in our codebase is very fragile as we can't react to changes done in many many releases of Liquibase before.

ahus1 (approved):

PR #41234, opened 2025-07-17
Document configuration changes that prevent rolling updates

ryanemerson:
Closes #41214

### DIFF
--- a/docs/guides/server/update-compatibility.adoc
+++ b/docs/guides/server/update-compatibility.adoc
@@ -146,14 +146,14 @@ The following configuration changes return a "Rolling Update is not possible" re
 
 The enabling or disabling of the following features requires a recreate update:
 
-<@showFeaturesRolling ctx.features.upgradePolicyShutdown />
+<@showFeaturesRolling ctx.features.updatePolicyShutdown />
 
 
 ==== Recreate on feature version change
 
 Changing the following features versions triggers a recreate update:
 
-<@showFeaturesRolling ctx.features.upgradePolicyRollingNoUpgrade />
+<@showFeaturesRolling ctx.features.updatePolicyRollingNoUpgrade />
 
 === Configuration options
 
--- a/docs/guides/server/update-compatibility.adoc
+++ b/docs/guides/server/update-compatibility.adoc
@@ -5,6 +5,7 @@
 <@tmpl.guide
 title="Checking if rolling updates are possible"
 summary="Execute the update compatibility command to check if {project_name} supports a rolling update for a change in your deployment."
+includedOptions="features features-*"
 >
 
 Use the update compatibility command to determine if you can update your deployment with a rolling update strategy when enabling or disabling features or changing the {project_name} version, configurations or providers and themes.
@@ -46,15 +47,7 @@ NOTE: If you do not use `--optimized` keep in mind that an `update` command may 
 
 [WARNING]
 ====
-The `check` command currently offers only a limited functionality. At the moment, it takes into consideration only the version of {project_name} and the embedded Infinispan to determine if a rolling update is possible.
-If those are unchanged, it reports that a rolling update is possible.
-
-The current version does not yet verify configuration changes and assumes all configuration changes are eligible for a rolling update.
-The same applies to changes to custom extensions and themes.
-
-A good use case when to use this is, for example, when you want to do a rolling update when you change the {project_name} theme or your custom extensions, and only want run recreate update when the version of {project_name} changes which does not yet allow a rolling update.
-
-While consumers of these commands should know the limitations that exist today, they should not rely on the internal behavior or the structure of the metadata file.
+Consumers of these commands should not rely on the internal behavior or the structure of the metadata file.
 Instead, they should rely only on the exit code of the `check` command to benefit from future enhancements on the internal logic to determine when a rolling update is possible.
 ====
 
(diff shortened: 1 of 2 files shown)


docs/guides/server/update-compatibility.adoc
```diff
@@ -143,6 +136,39 @@ m|4
 The feature `rolling-updates` is disabled.
 |===
 
+== Rolling incompatible changes
+
+The following is a summary of known configuration changes which are not compatible with the rolling upgrade strategy.
```
ahus1: AFAIK this is about understanding the automatic rolling update functionality, so I'm rewording this. Also note: At some point we named it "rolling update", and not "rolling upgrade".  Please review the "upgrade" occurrences at least in your code changes.

```suggestion
The following configuration changes return a "Rolling Update is not possible" result code.
```

docs/guides/server/update-compatibility.adoc
```diff
@@ -143,6 +136,39 @@ m|4
...
+==== Recreate always
+
+The enabling or disabling of the following features requires a recreate update:
+
+<@showFeaturesRolling ctx.features.upgradePolicyShutdown />
+
+
+==== Recreate on upgrade
+
+Upgrading or downgrading to one of the following features requires a recreate update:
```
ahus1: Looking at the example, I suppose this now about upgrading (as it might be considered related to upgrading Keycloak versions), but about feature version changing. Please let me know about your thoughts on the new wording, which might affect some renaming in the code as well.

```suggestion
==== Recreate on feature version change

Changing the following features versions triggers a recreate update:
```

docs/guides/server/update-compatibility.adoc
```diff
@@ -143,6 +136,39 @@ m|4
...
+
+=== Configuration options
+
+Changing the value of one of the following CLI options requires a recreate update:
+
+[cols="30%,70%"]
+|===
+| Option                          | Rationale
+| `--cache`                       | The `ispn` and `local` configurations are mutually exclusive, changing from one to another will lead to data loss.
+| `--cache-config-file`           | Changing the configuration file could result in incompatible cache or transport configurations, resulting in clusters not forming as expected.
```
ahus1, line 166: I compared this with the code, and AFAIK we're currently checking that the name of the config file changes, and if it does, we prevent a rolling update. 

This doesn't add much value IMHO, so I'm suggesting that we're removing this item (though not a strong opinion), given that we'll need to add a warning about changing the file anyway (see next paragraph). 

In any case, we should add a warning that any change to the cache config XML file can cause the cluster to not form, and it is for the admin changing the file to determine if the cluster should be shut down first.

docs/guides/server/update-compatibility.adoc
```diff
@@ -143,6 +136,39 @@ m|4
...
+
+==== Recreate on upgrade
+
+Upgrading or downgrading to one of the following features requires a recreate update:
+
+<@showFeaturesRolling ctx.features.upgradePolicyRollingNoUpgrade />
+
+=== Configuration options
+
+Changing the value of one of the following CLI options requires a recreate update:
```
ahus1:
```suggestion
Changing the value of one of the following CLI options triggers a recreate update:
```

docs/guides/server/update-compatibility.adoc
```diff
@@ -177,3 +203,14 @@ Known limitations:
...
+<#macro showFeaturesRolling features>
+[cols="30%,70%"]
+|===
+| Feature | Description
+
+<#list features as feature>
+| [.features-name]#${feature.versionedKey}# | [.features-description]#${feature.description}#
+</#list>
+|===
+</#macro>
```
ahus1, line 223: While this uses a table formatting and looks nice, this looks different from the current overview of features. I'd vote for consistency here, and stay with the formatting we have on the `features.adoc` overview.

ahus1 (changes requested):
Thank you for this PR. See some changes supporting the read persona "I'm an admin and I want to understand how the upgrade command behaves when I change stuff"

docs/guides/server/update-compatibility.adoc
ryanemerson: The check on `--cache-config-file` also prevents a rolling update on change of `--cache`, as `CachingPropertyMappers` sets `kc.spi-cache-embedded--default--config-file`. We currently don't have an SPI option for `--cache`, so more work would be required here if we want to allow `--cache-config-file` to be changed but disallow changes to `--cache`.



> In any case, we should add a warning that any change to the cache config XML file can cause the cluster to not form, and it is for the admin changing the file to determine if the cluster should be shut down first.

+1

docs/guides/server/update-compatibility.adoc
ryanemerson: How about we change the features page to use a similar table? It's much easier to read IMO.

ahus1: We can leave it as-is now in this PR, and you take up the task to resolve this with the CND team?

ahus1 (dismissed):
Thank you for this PR. I've reworded the added warning trying to make it more actionable. Please let me know if you agree.

ryanemerson:
> I've reworded the added warning trying to make it more actionable. Please let me know if you agree.

Thanks, I think this is an improvement, but I have also pushed an additional commit to breakup the sentences a little in order to improve flow.

ahus1 (approved):

docs/guides/server/update-compatibility.adoc
ryanemerson: https://github.com/keycloak/keycloak/issues/41328

PR #41240, opened 2025-07-17
Show transport media for WebAuthn authenticators in Account console

### LINKED ISSUE #10063: Display transport media for WebAuthn authenticators in Account console
There are displayed registered Security keys in the account console and it'd be great to have there included also provided information about supported transport media for the keys.

adhamahmad:
Closes #10063

image

### DIFF
--- a/js/apps/account-ui/maven-resources/theme/keycloak.v3/account/messages/messages_en.properties
+++ b/js/apps/account-ui/maven-resources/theme/keycloak.v3/account/messages/messages_en.properties
@@ -216,3 +216,4 @@ domains=Domains
 refresh=Refresh
 termsAndConditionsDeclined=You need to accept the Terms and Conditions to continue
 defaultLocale=Choose...
+webauthn-info= Transport media: {{0}}
--- a/services/src/main/java/org/keycloak/credential/WebAuthnCredentialProvider.java
+++ b/services/src/main/java/org/keycloak/credential/WebAuthnCredentialProvider.java
@@ -44,6 +44,7 @@ import org.keycloak.models.UserModel;
 import org.keycloak.models.WebAuthnPolicy;
 import org.keycloak.models.credential.WebAuthnCredentialModel;
 import org.keycloak.models.credential.dto.WebAuthnCredentialData;
+import org.keycloak.util.JsonSerialization;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -57,6 +58,7 @@ import java.util.stream.Collectors;
 public class WebAuthnCredentialProvider implements CredentialProvider<WebAuthnCredentialModel>, CredentialInputValidator {
 
     private static final Logger logger = Logger.getLogger(WebAuthnCredentialProvider.class);
+    private static final String WEBAUTHN_INFO = "webauthn-info";
 
     private KeycloakSession session;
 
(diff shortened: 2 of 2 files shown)


mposolda:
@adhamahmad Thanks for the PR. I think it would be ideal if we do not have any specifics of concrete credential type (passkey) mentioned directly in the account console... Can be good if it is specified somehow which credential metadata are displayed for each credential-type  instead of directly referencing "Transports" in the account-console UI code. See the class `org.keycloak.credential.CredentialMetadata` for the inspiration (I believe this one is used for example to show the warning in the account-console related to recovery-codes when there is small amount of codes left. I believe something similar would be good for showing additional data about credential metadata as well).

adhamahmad:
Hi @mposolda , thanks for the feedback!

I’ve updated the implementation to remove the direct reference to “Transports” in the account console UI. Instead, I’ve exposed the transport data through `CredentialMetadata`, following a similar approach to what’s done in `org.keycloak.credential.RecoveryAuthnCodesCredentialProvider`.

Let me know if this direction looks good to you or if you'd suggest any further adjustments.

image

services/src/main/java/org/keycloak/credential/WebAuthnPasswordlessCredentialProvider.java
```diff
@@ -57,4 +63,27 @@ public CredentialTypeMetadata getCredentialTypeMetadata(CredentialTypeMetadataCo
     protected WebAuthnPolicy getWebAuthnPolicy() {
         return getKeycloakSession().getContext().getRealm().getWebAuthnPolicyPasswordless();
     }
+
+    @Override
+    public CredentialMetadata getCredentialMetadata(WebAuthnCredentialModel credentialModel, CredentialTypeMetadata credentialTypeMetadata) {
```
rmartinc: @adhamahmad I think that this is better in the parent class `WebAuthnCredentialProvider`. The transport message is OK for both webauthn credentials, so moving to the parent class is enough.

services/src/main/java/org/keycloak/credential/WebAuthnPasswordlessCredentialProvider.java
```diff
@@ -57,4 +63,27 @@ public CredentialTypeMetadata getCredentialTypeMetadata(CredentialTypeMetadataCo
...
+
+        CredentialMetadata credentialMetadata = new CredentialMetadata();
+
+        try {
+            WebAuthnCredentialData credentialData = JsonSerialization.readValue(credentialModel.getCredentialData(), WebAuthnCredentialData.class);
+            Set<String> transports = credentialData.getTransports();
+            if (transports != null && !transports.isEmpty()) {
+                String joinedTransports = String.join(", ", transports);
+                credentialMetadata.setInfoMessage("credentialTransports", joinedTransports);
```
rmartinc: Just a cherry-pick, but maybe you can use a static private property in the model like it is used in the recovery-codes (see [here](https://github.com/keycloak/keycloak/blob/26.3.2/services/src/main/java/org/keycloak/credential/RecoveryAuthnCodesCredentialProvider.java#L82))? And I would use something more generic like `webauthn-info` (or similar) just in case we include something more later...

rmartinc (changes requested):
Thanks @adhamahmadha!

IMHO it's almost OK. Please move the implementation to the parent class, because the message is valid for both types of credentials (webauthn as 2FA and as passwordless). The other comment is just a minor thing to create a constant with a more generic name. :smile:

services/src/main/java/org/keycloak/credential/WebAuthnPasswordlessCredentialProvider.java
```diff
@@ -18,19 +18,23 @@
...
+import java.io.IOException;
+import java.util.Set;
 import com.webauthn4j.converter.util.ObjectConverter;
+import org.jboss.logging.Logger;
 import org.keycloak.authentication.requiredactions.WebAuthnPasswordlessRegisterFactory;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.WebAuthnPolicy;
 import org.keycloak.models.credential.WebAuthnCredentialModel;
+import org.keycloak.models.credential.dto.WebAuthnCredentialData;
+import org.keycloak.util.JsonSerialization;
```
rmartinc: These new imports are not needed now, right?

rmartinc (changes requested):
Thanks @adhamahmad! I think it's OK now, just remove the imports if they are not needed now, and it's ready IMHO.

adhamahmad:
Hi @rmartinc, Thanks for the review!
I removed the unused imports and updated the PR.

rmartinc (approved):
Thanks @adhamahmad! LGTM now!

jonkoops (approved):

abstractj (approved):

Issue #41204, opened 2025-07-16
UpdateTest CI failures


shawkins:
> @shawkins It seems there are some other problems, as these tests are not very stable. Failed only for local API server, so it should not be a big deal - JFYI. 
> 
> https://github.com/keycloak/keycloak/actions/runs/16297788433/job/46032420962?pr=41173 

 _Originally posted by @mabartos in [#40932](https://github.com/keycloak/keycloak/issues/40932#issuecomment-3077112248)_

See the follow-up comments for more.

shawkins:
Marking as important as it's occurring with regularity.

mabartos:
@shawkins Wouldn't be possible to ignore these tests for `localApiServer` for now, and once we have the solution, test them again? It is a very frequent test failure in PRs. 

See: https://github.com/keycloak/keycloak/pull/41252


```
Error:  Tests run: 9, Failures: 1, Errors: 0, Skipped: 4, Time elapsed: 141.042 s <<< FAILURE! - in org.keycloak.operator.testsuite.integration.UpdateTest
Error:  org.keycloak.operator.testsuite.integration.UpdateTest.testCacheMaxCount(UpdateStrategy)[3]  Time elapsed: 122.191 s  <<< FAILURE!
java.lang.AssertionError: java.util.concurrent.TimeoutException
	at org.keycloak.operator.testsuite.integration.UpdateTest.await(UpdateTest.java:281)
	at org.keycloak.operator.testsuite.integration.UpdateTest.testCacheMaxCount(UpdateTest.java:109)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at io.quarkus.test.junit.QuarkusTestExtension.runExtensionMethod(QuarkusTestExtension.java:1000)
	at io.quarkus.test.junit.QuarkusTestExtension.runExtensionMethod(QuarkusTestExtension.java:930)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptTestTemplateMethod(QuarkusTestExtension.java:889)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
[...]

PR #41252, opened 2025-07-18
Temporarily disable UpdateTest.testCacheMaxCount for Local API server

mabartos:
Relates to: https://github.com/keycloak/keycloak/issues/41204

It should be temporarily disabled for now as it causes a lot of trouble in CI.

For now, the UpdateTest fails only for the `testCacheMaxCount` and not for the one described in the parent issue.

@shawkins Could you check it?


mabartos:
Superseded by: https://github.com/keycloak/keycloak/pull/41254

PR #41256, opened 2025-07-18
fix: adding group import handling similar to users

shawkins:
closes: #41235

Basically anything that pushes the size of the persistence context beyond a couple thousand items will cause performance to drop dramatically. Users were the most obvious case of this, but here it was due to groups.

This seems like the narrowest fix for group import performance. Like the user import performance test, I can include one that imports several thousand groups if that seems like a good idea.

It was mentioned in some of the initial work that got rid of using multiple transactions that we could eventually try to align with the other batching logic - which implicitly is doing a flush and clear of the context. Pursuing that more broadly is still a possible alternative but requires a good deal more care.

The only other alternative is doing something like escape analysis to figure out when entities can be flushed / detached - but that breaks how the logic is encapsulated even more than using the static EntityManagers calls.

@ahus1 @pedroigor 

cc @gryffus

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
@@ -149,7 +149,7 @@ public class DefaultExportImportManager implements ExportImportManager {
     private final KeycloakSession session;
     private static final Logger logger = Logger.getLogger(DefaultExportImportManager.class);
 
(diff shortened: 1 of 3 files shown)


ahus1 (approved):
LGTM, I have reviewed the logic.

ahus1:
Sorry, I should have at least waited for @mabartos to review it as well before merging it.

mabartos:
@ahus1 No worries, LGTM.

PR #41265, opened 2025-07-18
[OTel] Introduce preview support for OpenTelemetry Logs

### LINKED ISSUE #41263: [OTel] Provide general options for telemetry settings
### Description

As described in the parent epic, it'd be good to have a general OpenTelemetry options that might be shared for OTel Logs, OTel Metrics and OTel Traces.

mabartos:
- Closes #41264
- Based-on https://github.com/keycloak/keycloak/pull/41705

## Grafana OTel LGTM
I used https://github.com/grafana/docker-otel-lgtm as a OpenTelemetry backend in a Docker image to see if we're able to see logs via OTel collector - uses Loki (logs), Prometheus (metrics), Tempo (traces).

image

Run:
```bash
docker run -p 3000:3000 -p 4317:4317 -p 4318:4318 --rm -ti grafana/otel-lgtm
```

And start Keycloak by executing:
 ```bash
./kc.sh start-dev --feature-opentelemetry-logs=enabled --telemetry-logs-enabled=true
```

You see the exported logs in the Loki console accessible on `localhost:3000`:
image

### CODE REVIEW

File: docs/guides/observability/telemetry.adoc [hunk 1]
```diff
@@ -55,11 +55,58 @@ You can change the protocol via CLI as follows:
 
 You can override the global OpenTelemetry settings via the Tracing options if you want to export traces to a different OpenTelemetry collector.
 
+== Logs
+
+WARNING: OpenTelemetry Logs is considered a *preview* and can be changed. Feedback is more than welcome.
+
+It is possible to export {project_name}'s logs to the OpenTelemetry collector and be managed by various logging backends that support OpenTelemetry.
+It helps to have a centralized logging system that is beneficial for application deployments.
```
ryanemerson: Maybe I misunderstand what you're trying to say here, but I'm not sure this sentence really adds any value. Can we remove it?

docs/guides/observability/telemetry.adoc
```diff
@@ -55,11 +55,58 @@ You can change the protocol via CLI as follows:
...
+
+For more details about OpenTelemetry Logs functionality in Quarkus, see the https://quarkus.io/guides/opentelemetry-logging[Quarkus OpenTelemetry Logging guide].
+For more information on how to set up logging, see the https://www.keycloak.org/server/logging[Configuring Logging guide].
+
+=== Log level
+You can specify what logs will be exported to the OpenTelemetry Collector based on the log level via the `log-export-level` option.
+By default, all log levels are exported, but you can choose the most verbose log level that should be exported.
+For instance, if you set the level to `WARN`, warning and errors will be exported.
+
+You can change the log level via CLI as follows:
```
ryanemerson: I think we can simplify this section as there's some repetition, how about:

"By default all log levels are exported by {project_name}. You can optionally configure the most verbose log level to be exported via the `log-export-level` option.
For example, if you want only to export `WARN` and `ERROR` logs you can change the log level via the CLI as follows:

<@kc.start parameters="--log-export-level=WARN"/>
"

docs/guides/server/logging.adoc
```diff
@@ -542,7 +542,13 @@ To use unstructured logging, enter the following command:
 
 As you can see, the timestamp is present twice, so you can amend it correspondingly via the `--log-syslog-format` property.
 
-<@opts.printRelevantOptions includedOptions="log log-*" excludedOptions="log-console-* log-file log-file-* log-syslog-*">
+== Logs export (OpenTelemetry)
+
+It is possible to use OpenTelemetry Logs that enables sending logs to the OpenTelemetry collector and having a centralized observability stack for your deployment.
```
ryanemerson: "It is possible to export OpenTelemetry logs from your deployment to the OpenTelemetry collector for centralized log managment."

docs/guides/observability/telemetry.adoc
mabartos: I tried to say the same as in the general info about OpenTelemetry related to a single place for logs, but yep, it can be removed from here.

mabartos:
@ryanemerson Updated, thanks!

mabartos:
Rebased and resolved conflicts.

@ryanemerson @vmuzikar @shawkins @Pepo48 Could you please check it for OTel Logs? Thanks!

docs/guides/observability/telemetry.adoc
```diff
@@ -55,11 +55,56 @@ You can change the protocol via CLI as follows:
 
 You can override the global OpenTelemetry settings via the Tracing options if you want to export traces to a different OpenTelemetry collector.
 
+== Logs
+
+WARNING: OpenTelemetry Logs is considered a *preview* and can be changed. Feedback is more than welcome.
```
ryanemerson: Maybe remove the feedback sentence, or simplify to "Feedback is welcome" as "more than" seems too informal for our docs.

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
```diff
@@ -332,4 +332,30 @@ public String toString() {
...
+    public static final Option<Boolean> LOG_EXPORT_ENABLED = new OptionBuilder<>("log-export-enabled", Boolean.class)
+            .category(OptionCategory.LOGGING)
+            .description("PREVIEW. Enables exporting logs to a destination handling telemetry data (OpenTelemetry Logs).")
+            .defaultValue(Boolean.FALSE)
+            .buildTime(true)
+            .build();
+
+    public static final Option<String> LOG_EXPORT_ENDPOINT = new OptionBuilder<>("log-export-endpoint", String.class)
+            .category(OptionCategory.LOGGING)
+            .description("PREVIEW. Telemetry (OpenTelemetry) endpoint to connect to for Logs. If not given, the value is inherited from the '%s' option.".formatted(TelemetryOptions.TELEMETRY_ENDPOINT.getKey()))
```
ryanemerson: "... endpoint to export logs to."

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
```diff
@@ -332,4 +332,30 @@ public String toString() {
...
+            .build();
+
+    public static final Option<String> LOG_EXPORT_PROTOCOL = new OptionBuilder<>("log-export-protocol", String.class)
+            .category(OptionCategory.LOGGING)
+            .description("PREVIEW. Telemetry (OpenTelemetry) protocol used for the logging telemetry data. If not given, the value is inherited from the '%s' option.".formatted(TelemetryOptions.TELEMETRY_PROTOCOL.getKey()))
```
ryanemerson: "...used for exporting logs."

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
```diff
@@ -332,4 +332,30 @@ public String toString() {
...
+            .expectedValues("grpc", "http/protobuf")
+            .build();
+
+    public static final Option<LoggingOptions.Level> LOG_EXPORT_LEVEL = new OptionBuilder<>("log-export-level", LoggingOptions.Level.class)
+            .category(OptionCategory.LOGGING)
+            .description("Most verbose log level for logs exported to the destination handling telemetry data (OpenTelemetry collector). For more information, check the Telemetry guide.")
+            .defaultValue(LoggingOptions.Level.ALL)
```
ryanemerson: I think we can simplify this a little, maybe:

"The most verbose log level exported to the telemetry endpoint."

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/LoggingPropertyMappers.java
```diff
@@ -426,7 +455,7 @@ else if (isTracingEnabled && includeTrace) {
         return LoggingOptions.DEFAULT_LOG_FORMAT;
     }
 
-    private static String upperCase(String value, ConfigSourceInterceptorContext context) {
+    static String upperCase(String value, ConfigSourceInterceptorContext context) {
```
ryanemerson, line 434: This method is only used in this class according to Intellij.

ryanemerson (commented):
One thing I noticed when I tried locally was that the "INFO  [io.quarkus] (main) Keycloak stopped in 0.024s" statements are never exported, presumably that's because the required quarkus otel component has shutdown before this is logged.

Overall LGTM, just a few more doc comments.

mabartos:
> One thing I noticed when I tried locally was that the "INFO [io.quarkus] (main) Keycloak stopped in 0.024s" statements are never exported, presumably that's because the required quarkus otel component has shutdown before this is logged.

@ryanemerson Nice finding. I don't know all the details, but it might be the case that the OTel exporter is already closed. Would be good to bring up on the Quarkus side (cc @brunobat).

mabartos:
@ryanemerson Thanks! Updated

brunobat:
> > One thing I noticed when I tried locally was that the "INFO [io.quarkus] (main) Keycloak stopped in 0.024s" statements are never exported, presumably that's because the required quarkus otel component has shutdown before this is logged.
> 
> @ryanemerson Nice finding. I don't know all the details, but it might be the case that the OTel exporter is already closed. Would be good to bring up on the Quarkus side (cc @brunobat).

Correct. There in no OTel SDK by then to export that log line.
I acknowledge that we should probably have a shutdown event logged... Maybe we can do something at the OTel shutdown hook level.

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
```diff
@@ -333,4 +333,30 @@ public String toString() {
...
+            .expectedValues("grpc", "http/protobuf")
+            .build();
+
+    public static final Option<LoggingOptions.Level> LOG_EXPORT_LEVEL = new OptionBuilder<>("log-export-level", LoggingOptions.Level.class)
+            .category(OptionCategory.LOGGING)
+            .description("The most verbose log level exported to the telemetry endpoint. For more information, check the Telemetry guide.")
+            .defaultValue(LoggingOptions.Level.ALL)
+            .caseInsensitiveExpectedValues(true)
+            .build();
+
```
vmuzikar: Should those be rather under `telemetry-`? E.g. `telemetry-log-endpoint`? That'd make it consistent also with the guide – we placed the docs in the Telemetry guide.

docs/guides/server/logging.adoc
```diff
@@ -300,7 +300,14 @@ You can use regular expressions to exclude them, such as:
 
 In this case, all calls to the `/realms/my-internal-realm/` and subsequent paths will be excluded from the HTTP Access log.
 
-<@opts.printRelevantOptions includedOptions="log log-*" excludedOptions="log-console-* log-file log-file-* log-syslog-* log-mdc*">
+[[logs-export-opentelemetry]]
+== Logs export (OpenTelemetry)
+
+It is possible to export OpenTelemetry logs from your deployment to the OpenTelemetry collector for centralized log management.
+
+For more details, see the _Centralize your observability stack with OpenTelemetry_ guide.
```
vmuzikar: Seems we're missing a link here?

docs/guides/server/logging.adoc
```diff
@@ -300,7 +300,14 @@ You can use regular expressions to exclude them, such as:
...
+
+<@opts.printRelevantOptions includedOptions="log log-*" excludedOptions="log-console-* log-file log-file-* log-syslog-* log-mdc* log-export-*">
```
vmuzikar: This feels fragile. Should we rather use `includedOptions="log-export-*"`?

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
mabartos: @vmuzikar Good point. Yes, it might be probably better naming.

docs/guides/server/logging.adoc
mabartos: @vmuzikar I'd prefer to have a separate category than having it included in the relevant options. That means that below, there will be the 'Logs export' category (I'll do probably more changes anyway due to the naming ^^ we discussed).

Something like this with the option for the console:

Image

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
mabartos: BTW. We have quite inconsistent naming for the observability features like logs, metrics and traces. 

- For logs, we use the `log` (singular)
- For metrics, we use the `metrics` (plural)
- For traces, we use the `tracing` (gerund) 

Maybe we could make it right at least for the telemetry options and having the
- `telemetry-logs-*`
- `telemetry-metrics-*`

@vmuzikar WDYT?

vmuzikar: I think it's actually correct. :) You usually look into the log (singular) as "the single place for log messages". Metrics – that's definitely a plural, we have many of those. Tracing – tracings would sound awful TBH. :D

docs/guides/server/logging.adoc
vmuzikar: Yup, a separate category sounds good.

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
vmuzikar: But I'm not a native speaker, so CCing @shawkins. :)

mabartos: > You usually look into the log (singular) as "the single place for log messages".

@vmuzikar AFAIK in the OTel terminologies, the 'logs' is used rather to refer to log records (emit many log records to the collector,...) (https://opentelemetry.io/docs/concepts/signals/logs/). So, in this case, we talk more about sending the telemetry data. 

> Tracing – tracings would sound awful TBH. :D

xD yes, that would be awful. That's why it's called 'traces' (like emit many traces).

I'd like to have rather the `telemetry-logs` than `telemetry-log` to be more aligned with the telemetry terminology, but even the log(singular) works to me.

vmuzikar: > I'd like to have rather the `telemetry-logs` than `telemetry-log` to be more aligned with the telemetry terminology

Ok, sounds good.

shawkins: Continuing with the term log seems fine as the use of the singlar term is very common. Similarly for trace I don't see that any of these as being obviously more correct, or understandable, than the others:

telementry-trace-export-endpoint
telementry-traces-export-endpoint
telementry-tracing-export-endpoint

With metrics vs metric, we generally want the plural form. While you can do either telementry-trace-enabled or telementry-traces-enabled and it reads well either way, telementry-metric-enabled does not read well.

Quarkus for what it's worth is not consistent. While generally using the plural term, they also have https://quarkus.io/guides/all-config#quarkus-opentelemetry_quarkus-otel-metric-export-interval

docs/guides/server/logging.adoc
mabartos: Done

docs/guides/server/logging.adoc
mabartos: Done

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
mabartos: Provided `telemetry-logs-*` options.

mabartos:
## Summary
- New `opentelemetry-logs` preview feature
- New `telemetry-logs-*` options
- Tried with the LGTM service, and logs are successfully exported
- Added release notes

@vmuzikar Could you please check it? Thanks!

docs/guides/observability/telemetry.adoc
```diff
@@ -55,11 +57,58 @@ See the https://www.keycloak.org/observability/tracing[Root cause analysis with
...
+NOTE: The OpenTelemetry feature (`opentelemetry`) needs to be turned on (by default).
+
+It is possible to export {project_name}'s logs to the OpenTelemetry collector and be managed by various logging backends that support OpenTelemetry.
+
+=== Enable Logs
+You can enable OpenTelemetry Logs via CLI as follows:
+
+<@kc.start parameters="--features=opentelemetry-logs --telemetry-logs-enabled=true"/>
+
+For more details about OpenTelemetry Logs functionality in Quarkus, see the https://quarkus.io/guides/opentelemetry-logging[Quarkus OpenTelemetry Logging guide].
```
vmuzikar, line 73: One more thing. Is it actually beneficial to the Keycloak users the link to the Quarkus guide? The guide seems more dev oriented, it's not really focused on generic OTEL logs.

vmuzikar (approved):
LGTM, thanks @mabartos , nice work. Added one comment abou docs but it's not blocking.

mabartos:
@vmuzikar Thanks for the review! This will also unblock the work on the Micrometer to OTel Metrics bridge. 

Agree that we can remove the Quarkus guide reference as there's not so much additional useful information. 

I'll create a follow-up PR to avoid running the whole CI again. Merging.

mabartos:
Follow-up PR with the docs changes: https://github.com/keycloak/keycloak/pull/44911

PR #41267, opened 2025-07-19
Migrate to zh-Hant / zh-Hans for Chinese language

### LINKED ISSUE #41239: Migrate to zh-Hans / zh-Hant for simplified and traditional Chinese
### Description

Keycloak lists traditional and simplified Chinese in the community translations. 

The official codes are `zh-Hans` (simplified) and `zh-Hant` (traditional). 

For historic reasons, the project uses `zh-CN` and `zh_TW` as the locale codes and the file names.  

This is incompatible with Weblate, which only supports `zh-Hans` /  `zh-Hant`

### Discussion

https://github.com/keycloak/keycloak/discussions/9270#discussioncomment-13760884

### Motivation

There is already a workaround to generate the locale descriptions to map one to the other: 

https://github.com/keycloak/keycloak/blob/4af3d7cc9dc5e1aa2c63867d503e2c0c56270216/services/src/main/java/org/keycloak/theme/DefaultThemeManager.java#L313-L317

### Details

We need to rename the files and allow for a smooth migration, especially as users have saved in their locale the old codes, and we wouldn't be able to ad-hoc migrate them, and the same might go for themes and realm configurations. 

To do this, we would need to convert both old to new in several places.

allen0099:
Closes: #41239

This PR updates the Chinese locale files by migrating from `zh-TW` / `zh-CN` to `zh-Hant` / `zh-Hans`, while keeping `messages_zh_TW.properties` and `messages_zh_CN.properties` for backward compatibility.

The new locale resolution priority (for Traditional Chinese, as an example) is as follows:

1. Use `messages_zh_Hant.properties` if available
2. Fallback to `messages_zh_TW.properties`
3. Fallback to the original locale resolution logic

If both `zh-Hant` and `zh-TW` files exist and define the same key, the `zh-Hant` value will take precedence.


ahus1:
Thank you @allen0099 - you picked the right places for your changes. 

I added documentation for the migration guide, and did a little bit of refactoring to avoid code duplication. The code is now also more specific as it will check for `zh-TW` and `zh-CN` excplicitly. 

Let's wait for the CI to complete and to see if it is green. 

In the meantime, please let me know if you agree with these changes. 

Thanks!

allen0099:
Hi @ahus1,

Thanks for the improvements! The refactoring and migration guide look great.
I agree with your changes, and the CI has passed — all good from my side.

Thanks again for your help!

ahus1:
@ryanemerson - can you please review my added code and documentation changes in this PR when you have the time? Thanks!

ryanemerson (approved):

ahus1 (approved):

Lexachoc:
@ahus1 
@allen0099 

Hi, I'm using Keycloak 26.4.4 and I think there might be a regression related to this PR.

For English/Germn as default:
the chinese name is displayed correctly
Image

Image

But when chinese is selected: you will see that the label is repeated!
So, instead of displaying:

`中文（繁體）`

it displays:

`中文（繁體）（中文（繁體））`

Image

I'll open another issue to reference this.

PR #41291, opened 2025-07-21
Reduce likelihood of multiple coordinators on concurrent startup

### LINKED ISSUE #41290: Concurrent starts with JDBC_PING lead to a split cluster
### Before reporting an issue

### Area

infinispan

### Describe the bug

When starting two nodes concurrently, there can be two coordinators for an initial period of time.

### Version

23.x

### Regression

### Expected behavior

There should be only one coordinator at a time

### Actual behavior

There are two coordinators

### How to Reproduce?

As reported by @thomasdarimont 

In environments where customers deploy Keycloak via AWS Fargate, I sporadically see situations where two Keycloak instances in a cluster register themselves as an Infinispan coordintor in the jgroups_ping table.
Has anyone here seen this or similar situations?
The following examples are from a local docker compose example.

```
...
kc1-1  | 2025-07-18 19:20:34,313 INFO  [org.jgroups.protocols.pbcast.GMS] (main) kc1-35658: no members discovered after 2 ms: creating cluster as coordinator
kc2-1  | 2025-07-18 19:20:34,314 INFO  [org.jgroups.JChannel] (main) local_addr: 6fcea7e2-3e50-4239-88a2-e3a529cdd27a, name: kc2-31834
kc2-1  | 2025-07-18 19:20:34,321 INFO  [org.jgroups.protocols.FD_SOCK2] (main) server listening on *:57800
kc2-1  | 2025-07-18 19:20:34,324 INFO  [org.jgroups.protocols.pbcast.GMS] (main) kc2-31834: no members discovered after 2 ms: creating cluster as coordinator
kc1-1  | 2025-07-18 19:20:34,325 INFO  [org.infinispan.CLUSTER] (main) ISPN000094: Received new cluster view for channel ISPN: [kc1-35658|0] (1) [kc1-35658]
kc1-1  | 2025-07-18 19:20:34,327 INFO  [org.keycloak.jgroups.certificates.CertificateReloadManager] (main) Reloading JGroups Certificate
kc2-1  | 2025-07-18 19:20:34,336 INFO  [org.infinispan.CLUSTER] (main) ISPN000094: Received new cluster view for channel ISPN: [kc2-31834|0] (1) [kc2-31834]
...
```

This situation heals itself after a few seconds / (sometimes) minutes:

```
kc1-1  | 2025-07-18 19:21:22,368 INFO  [org.infinispan.CLUSTER] () ISPN000093: Received new, MERGED cluster view for channel ISPN: MergeView::[kc1-35658|1] (2) [kc1-35658, kc2-31834], 2 subgroups: [kc1-35658|0] (1) [kc1-35658], [kc2-31834|0] (1) [kc2-31834]
kc1-1  | 2025-07-18 19:21:22,368 INFO  [org.keycloak.jgroups.certificates.CertificateReloadManager] () Reloading JGroups Certificate
kc1-1  | 2025-07-18 19:21:22,372 INFO  [org.infinispan.CLUSTER] () ISPN100000: Node kc2-31834 joined the cluster
kc1-1  | 2025-07-18 19:21:22,372 INFO  [org.infinispan.CLUSTER] () ISPN100000: Node kc2-31834 joined the cluster
kc2-1  | 2025-07-18 19:21:23,380 INFO  [org.infinispan.CLUSTER] () ISPN000093: Received new, MERGED cluster view for channel ISPN: MergeView::[kc1-35658|1] (2) [kc1-35658, kc2-31834], 2 subgroups: [kc1-35658|0] (1) [kc1-35658], [kc2-31834|0] (1) [kc2-31834]
kc2-1  | 2025-07-18 19:21:23,380 INFO  [org.keycloak.jgroups.certificates.CertificateReloadManager] () Reloading JGroups Certificate
kc2-1  | 2025-07-18 19:21:23,385 INFO  [org.infinispan.CLUSTER] () ISPN100000: Node kc1-35658 joined the cluster
kc2-1  | 2025-07-18 19:21:23,386 INFO  [org.infinispan.CLUSTER] () ISPN100000: Node kc1-35658 joined the cluster
...
```

The problem ist that leads to all sorts of issues. Since both Keycloak instances are coordinator, the some maintenance tasks which should run only on one instance in the cluster are executed by multiple instances.
See: org.keycloak.storage.datastore.DefaultDatastoreProviderFactory#scheduleTask
One problem that I observed was that cleanup tasks lock some rows in the offline_user_session / offline_client_session table if it contains a high number of records. This leads to records being locked and user requests that managed to aquire a connection which want to access a "uncached" offline / online user session are effectively blocked due to the locks. If too many request are block, then they slowly start exhausting the database connection pool. After that Keycloak stops accepting new requests and becomes unresponsive. The next thing that happens is that the other Keycloak instances in the cluster start noticing that and receiver timeouts during the infinispan synchronization. This eventually can lead to a complete cluster failure and downtime of there is too much traffic on the system.
I wonder if it would make sense to add some additional checks to org.keycloak.jgroups.protocol.KEYCLOAK_JDBC_PING2 to detect that there is already another coordinator and restart the infinispan initialization to avoid those situations.

Image

Image

### Anything else?

Local test: 

```
volumes:
  postgres_data:
    driver: local

services:
  postgres:
    image: postgres:15
    volumes:
      - postgres_data:/var/lib/postgresql/data
    environment:
      POSTGRES_DB: keycloak
      POSTGRES_USER: keycloak
      POSTGRES_PASSWORD: password
    ports:
      - 5433:5432

  kc1:
    # image: quay.io/keycloak/keycloak:26.3.1
    image: customkeycloak
    #    build:
    #      context: .
    hostname: kc1
    environment:
#      DEBUG: "true"
#      DEBUG_PORT: "*:8787"
      KC_BOOTSTRAP_ADMIN_USERNAME: admin
      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
      KC_DB_SCHEMA: public
      KC_DB_USERNAME: keycloak
      KC_DB_PASSWORD: password
      KC_DB_URL: jdbc:postgresql://postgres/keycloak
      KC_HOSTNAME_STRICT: "false"
      KC_HTTP_ENABLED: "true"
      #KC_LOG_LEVEL: INFO,org.infinispan:DEBUG,org.jgroups:DEBUG
      KC_LOG_LEVEL: INFO,org.keycloak.services.scheduled:DEBUG,org.keycloak.storage.datastore.DefaultDatastoreProviderFactory:DEBUG

    ports:
      - 8080:8080
      - 8787:8787
      - 8443:8443

#    volumes:
#      - ./keycloak-benchmark-dataset-999.0.0-SNAPSHOT.jar:/opt/keycloak/providers/keycloak-benchmark-dataset.jar:z

    depends_on:
      - postgres

  kc2:
    image: customkeycloak
    #    build:
    #      context: .
    hostname: kc2
    environment:
      KC_BOOTSTRAP_ADMIN_USERNAME: admin
      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
      KC_DB_SCHEMA: public
      KC_DB_USERNAME: keycloak
      KC_DB_PASSWORD: password
      KC_DB_URL: jdbc:postgresql://postgres/keycloak
      KC_HOSTNAME_STRICT: "false"
      KC_HTTP_ENABLED: "true"
      KC_LOG_LEVEL: INFO,org.keycloak.services.scheduled:DEBUG,org.keycloak.storage.datastore.DefaultDatastoreProviderFactory:DEBUG
      #KC_LOG_LEVEL: INFO,org.infinispan:DEBUG,org.jgroups:DEBUG

    ports:
      - 18080:8080
      - 18443:8443

    depends_on:
      - postgres

  kc3:
    image: customkeycloak
    #    build:
    #      context: .
    hostname: kc3
    environment:
      KC_BOOTSTRAP_ADMIN_USERNAME: admin
      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
      KC_DB_SCHEMA: public
      KC_DB_USERNAME: keycloak
      KC_DB_PASSWORD: password
      KC_DB_URL: jdbc:postgresql://postgres/keycloak
      KC_HOSTNAME_STRICT: "false"
      KC_HTTP_ENABLED: "true"
      KC_LOG_LEVEL: INFO,org.keycloak.services.scheduled:DEBUG,org.keycloak.storage.datastore.DefaultDatastoreProviderFactory:DEBUG
      #KC_LOG_LEVEL: INFO,org.infinispan:DEBUG,org.jgroups:DEBUG

    ports:
      - 28080:8080
      - 28443:8443

    depends_on:
      - postgres


  kc4:
    image: customkeycloak
    #    build:
    #      context: .
    hostname: kc4
    environment:
      KC_BOOTSTRAP_ADMIN_USERNAME: admin
      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
      KC_DB_SCHEMA: public
      KC_DB_USERNAME: keycloak
      KC_DB_PASSWORD: password
      KC_DB_URL: jdbc:postgresql://postgres/keycloak
      KC_HOSTNAME_STRICT: "false"
      KC_HTTP_ENABLED: "true"
      KC_LOG_LEVEL: INFO,org.keycloak.services.scheduled:DEBUG,org.keycloak.storage.datastore.DefaultDatastoreProviderFactory:DEBUG
      #KC_LOG_LEVEL: INFO,org.infinispan:DEBUG,org.jgroups:DEBUG

    ports:
      - 38080:8080
      - 38443:8443

    depends_on:
      - postgres
```

```
ARG KC_VERSION=26.3.1
FROM quay.io/keycloak/keycloak:$KC_VERSION AS builder

ENV KC_METRICS_ENABLED=true
ENV KC_HEALTH_ENABLED=true
ENV KC_FEATURES=preview
ENV KC_DB=postgres
ENV KC_HTTP_RELATIVE_PATH=/auth

FROM quay.io/keycloak/keycloak:$KC_VERSION
COPY --from=builder /opt/keycloak/lib/quarkus/ /opt/keycloak/lib/quarkus/
COPY --from=builder /opt/keycloak/conf/ /opt/keycloak/conf/
WORKDIR /opt/keycloak

ENV KC_METRICS_ENABLED=true
ENV KC_HEALTH_ENABLED=true
ENV KC_FEATURES=preview
ENV KC_DB=postgres
ENV KC_HTTP_RELATIVE_PATH=/auth

# for demonstration purposes only, please make sure to use proper certificates in production instead
RUN keytool -genkeypair -storepass password -storetype PKCS12 -keyalg RSA -keysize 2048 -dname "CN=server" -alias server -ext "SAN:c=DNS:localhost,IP:127.0.0.1" -keystore conf/server.keystore

ENTRYPOINT ["/opt/keycloak/bin/kc.sh", "start"]
```

```
docker build -t customkeycloak .

docker compose up postgres
docker compose up kc1 kc2
docker compose up kc3 kc4
```

ahus1:
Closes #41290

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
@@ -17,12 +17,20 @@
 
 package org.keycloak.jgroups.protocol;
 
+import org.jgroups.Address;
+import org.jgroups.Event;
+import org.jgroups.PhysicalAddress;
+import org.jgroups.View;
 import org.jgroups.protocols.JDBC_PING2;
 import org.jgroups.protocols.PingData;
+import org.jgroups.util.NameCache;
+import org.jgroups.util.Responses;
 import org.keycloak.connections.jpa.JpaConnectionProviderFactory;
 
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
 /**
@@ -33,6 +41,39 @@ import java.util.Objects;
 public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
 
     private JpaConnectionProviderFactory factory;
+
+    @Override
+    protected void handleView(View new_view, View old_view, boolean coord_changed) {
+        // If we are the coordinator, it is good to learn about new entries that have been added before we delete them.
+        // If we are not the coordinator, it is good to learn the new entries added by the coordinator.
+        // This avoids a "JGRP000032: %s: no physical address for %s, dropping message" that leads to split clusters at concurrent startup.
+        learnExistingAddresses();
+        super.handleView(new_view, old_view, coord_changed);
+    }
+
+    protected void learnExistingAddresses() {
+        try {
+            List<PingData> list = readFromDB(getClusterName());
+            for (PingData data : list) {
+                Address addr = data.getAddress();
+                if (local_addr != null && !local_addr.equals(addr)) {
+                    addDiscoveryResponseToCaches(addr, data.getLogicalName(), data.getPhysicalAddr());
+                }
+            }
+        } catch (Exception e) {
+            log.error(String.format("%s: failed reading from the DB", local_addr), e);
+        }
+    }
+
+    @Override
+    public void findMembers(List<Address> members, boolean initial_discovery, Responses responses) {
+        // Insert ourselves before reading, to ensure that concurrently starting nodes see each other.
+        // Also re-add ourselves in case the coordinator removed us.
+         PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
+        PingData coord_data = new PingData(local_addr, true, NameCache.get(local_addr), physical_addr).coord(is_coord);
+        write(Collections.singletonList(coord_data), cluster_name);
+        super.findMembers(members, initial_discovery, responses);
+    }
 
     @Override
     protected void writeToDB(PingData data, String clustername) throws SQLException {
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
@@ -23,8 +23,11 @@ import org.jgroups.PhysicalAddress;
 import org.jgroups.View;
 import org.jgroups.protocols.JDBC_PING2;
 import org.jgroups.protocols.PingData;
+import org.jgroups.stack.Protocol;
+import org.jgroups.util.ExtendedUUID;
 import org.jgroups.util.NameCache;
 import org.jgroups.util.Responses;
+import org.jgroups.util.UUID;
 import org.keycloak.connections.jpa.JpaConnectionProviderFactory;
 
 import java.sql.Connection;
@@ -42,6 +45,18 @@ import java.util.stream.Collectors;
 public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
 
     private JpaConnectionProviderFactory factory;
+
+    @Override
+    public <T extends Protocol> T addr(Address addr) {
+        addr = toUUID(addr);
+        return super.addr(addr);
+    }
+
+    @Override
+    public <T extends Protocol> T setAddress(Address addr) {
+        addr = toUUID(addr);
+        return super.setAddress(addr);
+    }
 
     @Override
     protected void handleView(View new_view, View old_view, boolean coord_changed) {
@@ -179,6 +194,31 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
             lock.unlock();
         }
 
+    }
+
+    @Override
+    protected void delete(Connection conn, String clustername, Address addressToDelete) throws SQLException {
+        super.delete(conn, clustername, toUUID(addressToDelete));
+    }
+
+    @Override
+    protected void delete(String clustername, Address addressToDelete) throws SQLException {
+        super.delete(clustername, toUUID(addressToDelete));
+    }
+
+    @Override
+    protected void insert(Connection connection, PingData data, String clustername) throws SQLException {
+        if (data.getAddress() instanceof ExtendedUUID) {
+            data = new PingData(toUUID(data.getAddress()), data.isServer(), data.getLogicalName(), data.getPhysicalAddr()).coord(data.isCoord());
+        }
+        super.insert(connection, data, clustername);
+    }
+
+    private static Address toUUID(Address addr) {
+        if (addr instanceof ExtendedUUID eUUID) {
+            addr = new UUID(eUUID.getMostSignificantBits(), eUUID.getLeastSignificantBits());
+        }
+        return addr;
     }
 
     @Override
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
@@ -48,7 +48,48 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
         // If we are not the coordinator, it is good to learn the new entries added by the coordinator.
         // This avoids a "JGRP000032: %s: no physical address for %s, dropping message" that leads to split clusters at concurrent startup.
         learnExistingAddresses();
-        super.handleView(new_view, old_view, coord_changed);
+
+        // This is an updated logic where we do not call removeAll but instead remove those obsolete entries.
+        // This avoids the short moment where the table is empty and a new node might not see any other node.
+        if (is_coord) {
+            if (remove_old_coords_on_view_change) {
+                Address old_coord = old_view != null ? old_view.getCreator() : null;
+                if (old_coord != null)
+                    remove(cluster_name, old_coord);
+            }
+            Address[] left = View.diff(old_view, new_view)[1];
+            if (coord_changed || update_store_on_view_change || left.length > 0) {
+                writeAll(left);
+                if (remove_all_data_on_view_change) {
+                    removeAllNotInCurrentView();
+                }
+                if (remove_all_data_on_view_change || remove_old_coords_on_view_change) {
+                    startInfoWriter();
+                }
+            }
+        } else if (coord_changed) { // I'm no longer the coordinator
+            remove(cluster_name, local_addr);
+        }
+    }
+
+    @Override
+    protected void removeAll(String clustername) {
+        // This is unsafe as even if we would fill the table a moment later, a new node might see an empty table and become a coordinator
+        throw new RuntimeException("Not implemented as it is unsafe");
+    }
+
+    private void removeAllNotInCurrentView() {
+        try {
+            List<PingData> list = readFromDB(getClusterName());
+            for (PingData data : list) {
+                Address addr = data.getAddress();
+                if (view != null && !view.containsMember(addr)) {
+                    remove(cluster_name, addr);
+                }
+            }
+        } catch (Exception e) {
+            log.error(String.format("%s: failed reading from the DB", local_addr), e);
+        }
     }
 
     protected void learnExistingAddresses() {
@@ -69,7 +110,7 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
     public void findMembers(List<Address> members, boolean initial_discovery, Responses responses) {
         // Insert ourselves before reading, to ensure that concurrently starting nodes see each other.
         // Also re-add ourselves in case the coordinator removed us.
-         PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
+        PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
         PingData coord_data = new PingData(local_addr, true, NameCache.get(local_addr), physical_addr).coord(is_coord);
         write(Collections.singletonList(coord_data), cluster_name);
         super.findMembers(members, initial_discovery, responses);
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
@@ -47,18 +47,6 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
     private JpaConnectionProviderFactory factory;
 
     @Override
-    public <T extends Protocol> T addr(Address addr) {
-        addr = toUUID(addr);
-        return super.addr(addr);
-    }
-
-    @Override
-    public <T extends Protocol> T setAddress(Address addr) {
-        addr = toUUID(addr);
-        return super.setAddress(addr);
-    }
-
-    @Override
     protected void handleView(View new_view, View old_view, boolean coord_changed) {
         // If we are the coordinator, it is good to learn about new entries that have been added before we delete them.
         // If we are not the coordinator, it is good to learn the new entries added by the coordinator.
@@ -196,6 +184,22 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
 
     }
 
+    /* START: JDBC_PING2 does not handle ExtendedUUID yet, see
+       https://github.com/belaban/JGroups/pull/901 - until this is backported, we convert all of them.
+     */
+
+    @Override
+    public <T extends Protocol> T addr(Address addr) {
+        addr = toUUID(addr);
+        return super.addr(addr);
+    }
+
+    @Override
+    public <T extends Protocol> T setAddress(Address addr) {
+        addr = toUUID(addr);
+        return super.setAddress(addr);
+    }
+
     @Override
     protected void delete(Connection conn, String clustername, Address addressToDelete) throws SQLException {
         super.delete(conn, clustername, toUUID(addressToDelete));
@@ -220,6 +224,10 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
         }
         return addr;
     }
+
+    /* END: JDBC_PING2 does not handle ExtendedUUID yet, see
+       https://github.com/belaban/JGroups/pull/901 - until this is backported, we convert all of them.
+    */
 
     @Override
     protected void loadDriver() {
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
@@ -32,6 +32,7 @@ import java.sql.SQLException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * Enhanced JDBC_PING2 to handle entries transactionally.
@@ -130,7 +131,8 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
                 while (pingData.stream().noneMatch(PingData::isCoord)) {
                     // Do a quick check if more nodes have arrived, to have a more complete list of nodes to start with.
                     List<PingData> newPingData = readFromDB(cluster_name);
-                    if (newPingData.size() == pingData.size() || pingData.stream().anyMatch(PingData::isCoord)) {
+                    if (newPingData.stream().map(PingData::getAddress).collect(Collectors.toSet()).equals(pingData.stream().map(PingData::getAddress).collect(Collectors.toSet()))
+                            || pingData.stream().anyMatch(PingData::isCoord)) {
                         break;
                     }
                     pingData = newPingData;
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
@@ -67,8 +67,14 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
                     startInfoWriter();
                 }
             }
-        } else if (coord_changed) { // I'm no longer the coordinator
-            remove(cluster_name, local_addr);
+        } else if (coord_changed && !remove_all_data_on_view_change) {
+            // I'm no longer the coordinator, usually due to a merge.
+            // The new coordinator will update my status to non-coordinator, and remove me fully
+            // if 'remove_all_data_on_view_change' is enabled and I'm no longer part of the view.
+            // Maybe this branch even be removed completely, but for JDBC_PING 'remove_all_data_on_view_change' is always set to true.
+            PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
+            PingData coord_data = new PingData(local_addr, true, NameCache.get(local_addr), physical_addr).coord(is_coord);
+            write(Collections.singletonList(coord_data), cluster_name);
         }
     }
 
@@ -84,6 +90,7 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
             for (PingData data : list) {
                 Address addr = data.getAddress();
                 if (view != null && !view.containsMember(addr)) {
+                    addDiscoveryResponseToCaches(addr, data.getLogicalName(), data.getPhysicalAddr());
                     remove(cluster_name, addr);
                 }
             }
@@ -107,12 +114,32 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
     }
 
     @Override
+    public synchronized boolean isInfoWriterRunning() {
+        // Do not rely on the InfoWriter, instead always write the missing information on find if it is missing. Find is also triggered by MERGE.
+        return false;
+    }
+
+    @Override
     public void findMembers(List<Address> members, boolean initial_discovery, Responses responses) {
-        // Insert ourselves before reading, to ensure that concurrently starting nodes see each other.
-        // Also re-add ourselves in case the coordinator removed us.
-        PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
-        PingData coord_data = new PingData(local_addr, true, NameCache.get(local_addr), physical_addr).coord(is_coord);
-        write(Collections.singletonList(coord_data), cluster_name);
+        if (initial_discovery) {
+            try {
+                List<PingData> pingData = readFromDB(cluster_name);
+                PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
+                PingData coord_data = new PingData(local_addr, true, NameCache.get(local_addr), physical_addr).coord(is_coord);
+                write(Collections.singletonList(coord_data), cluster_name);
+                while (pingData.stream().noneMatch(PingData::isCoord)) {
+                    // Do a quick check if more nodes have arrived, to have a more complete list of nodes to start with.
+                    List<PingData> newPingData = readFromDB(cluster_name);
+                    if (newPingData.size() == pingData.size() || pingData.stream().anyMatch(PingData::isCoord)) {
+                        break;
+                    }
+                    pingData = newPingData;
+                }
+            } catch (Exception e) {
+                log.error(String.format("%s: failed reading from the DB", local_addr), e);
+            }
+        }
+
         super.findMembers(members, initial_discovery, responses);
     }
 
--- a/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
+++ b/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
@@ -17,6 +17,8 @@
 
 package org.keycloak.spi.infinispan.impl.embedded;
 
+import static org.infinispan.configuration.global.TransportConfiguration.CLUSTER_NAME;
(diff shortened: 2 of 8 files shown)


model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -106,13 +113,33 @@ protected void learnExistingAddresses() {
...
+        if (initial_discovery) {
+            try {
+                List<PingData> pingData = readFromDB(cluster_name);
+                PhysicalAddress physical_addr = (PhysicalAddress) down(new Event(Event.GET_PHYSICAL_ADDRESS, local_addr));
+                PingData coord_data = new PingData(local_addr, true, NameCache.get(local_addr), physical_addr).coord(is_coord);
+                write(Collections.singletonList(coord_data), cluster_name);
+                while (pingData.stream().noneMatch(PingData::isCoord)) {
+                    // Do a quick check if more nodes have arrived, to have a more complete list of nodes to start with.
+                    List<PingData> newPingData = readFromDB(cluster_name);
+                    if (newPingData.size() == pingData.size() || pingData.stream().anyMatch(PingData::isCoord)) {
```
ryanemerson: Is it safe to just check the size of the List or do we need to consider the content of the PingData as well. I'm just wondering if the view changed from `[a,b]` to `[a,c]`

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JDBC_PING2_Test.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Misc tests for {@link org.jgroups.protocols.JDBC_PING2}, running against H2
+ * @author Bela Ban
+ * @since  5.4, 5.3.7
+ */
+public class JDBC_PING2_Test {
```
ryanemerson: It would be good to format this file per Keycloak conventions.

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JDBC_PING2_Test.java
```diff
@@ -0,0 +1,126 @@
...
+        GMS gms=ch.stack().findProtocol(GMS.class);
+        gms.setJoinTimeout(3000).setMaxJoinAttempts(5);
+        return ch;
+    }
+
+    protected static JChannel createChannel(String cfg, String name) throws Exception {
+        return modify(new JChannel(cfg).name(name));
+    }
+
+    protected static class Connector implements Runnable {
```
ryanemerson: Can be a record:

```java
    protected record Connector(CountDownLatch latch, JChannel ch) implements Runnable {
        @Override
        public void run() {
            try {
                latch.await();
                ch.connect(CLUSTER);
            } catch (Exception e) {
                throw new RuntimeException(e);
            }
        }
    }
```

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JDBC_PING2_Test.java
```diff
@@ -0,0 +1,126 @@
...
+        latch.countDown();
+        long start = System.nanoTime();
+        Util.waitUntilAllChannelsHaveSameView(20000, 100, channels);
+        long time = System.nanoTime() - start;
+        System.out.printf("-- cluster of %d formed in %s:\n%s\n", NUM_NODES, Duration.ofNanos(time),
+                Stream.of(channels).map(ch -> String.format("%s: %s", ch.address(), ch.view()))
+                        .collect(Collectors.joining("\n")));
+        Arrays.stream(channels).filter(ch -> ch.view().getCoord() != ch.getAddress()).forEach(JChannel::close);
+        Arrays.stream(channels).filter(ch -> !ch.isClosed()).forEach(JChannel::close);
+        System.out.println("Closed");
```
ryanemerson: Do we need to remove the various `System.out` calls or replace them with debug logging before merging this?

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -34,6 +42,107 @@ public class KEYCLOAK_JDBC_PING2 extends JDBC_PING2 {
...
+            if (remove_old_coords_on_view_change) {
+                Address old_coord = old_view != null ? old_view.getCreator() : null;
+                if (old_coord != null)
+                    remove(cluster_name, old_coord);
+            }
+            Address[] left = View.diff(old_view, new_view)[1];
+            if (coord_changed || update_store_on_view_change || left.length > 0) {
+                writeAll(left);
+                if (remove_all_data_on_view_change) {
+                    removeAllNotInCurrentView();
```
ryanemerson, line 68: I think this is fine for our use-case, but would we need a new flag for this to make it into JGroups as we're no longer removing all data on a view change?

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
ahus1: Resolved in next commit, will check if they both contain the same addresses

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JDBC_PING2_Test.java
ahus1: resolved in next commit

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JDBC_PING2_Test.java
ahus1: resolved in next commit

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JDBC_PING2_Test.java
ahus1: resolved in next commit

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
ahus1: The old code did a delete-and-write-all, and the new code calls a write-all-and-delete leftovers. 

The result should be identical with the difference that there is no empty table in the middle.  So I'd say we're good to proceed on this change.

ahus1:
Thank you for the review. The PR now contains also a change to pre-seed the addresses so they are ordered, and the first node gets a one second grace period from the other nodes starting later to complete it becoming a coordinator.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -201,12 +211,48 @@ private static void configureDiscovery(ConfigurationBuilderHolder holder, Keyclo
...
+        String seq = storage.loadOrCreate(JGROUPS_ADDRESS_SEQUENCE, () -> "0");
+        long value = Long.parseLong(seq) + 1;
+        String newSeq = Long.toString(value);
+        storage.replace(JGROUPS_ADDRESS_SEQUENCE, seq, newSeq);
+
+        var cp = session.getProvider(JpaConnectionProvider.class);
+        var tableName = JpaUtils.getTableNameForNativeQuery("JGROUPS_PING", cp.getEntityManager());
+        String statement = String.format("INSERT INTO %s values (?, ?, ?, ?, ?)", tableName);
+
+        Address address = new UUID(0, value);
```
pruivo: Some Infinispan configurations require an `ExtendedUUID` (and the upcoming ISPN 16 too)

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -201,12 +211,48 @@ private static void configureDiscovery(ConfigurationBuilderHolder holder, Keyclo
...
+
+        cp.getEntityManager().runWithConnection(o -> {
+            Connection con = (Connection) o;
+            try (PreparedStatement s = con.prepareStatement(statement)) {
+                s.setString(1, org.jgroups.util.Util.addressToString(address)); // address
+                s.setString(2, "(starting)"); // name
+                s.setString(3, "ISPN"); // cluster name
```
pruivo: The cluster name can be changed; you need to read it from the configuration.

pruivo:
`KEYCLOAK_JDBC_PING2` looks like a mix of discovery, group membership, and failure detection. Let's see what kind of issues pop up in production. :cold_sweat:

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -201,12 +211,48 @@ private static void configureDiscovery(ConfigurationBuilderHolder holder, Keyclo
...
+     * If we are the first = smallest entry, the other nodes will wait for us to become a coordinator
+     * for max_join_attempts x all_clients_retry_timeout = 10 x 100 ms = 1 second. Otherwise, we will wait for that
+     * one second. This prevents a split-brain scenario on a concurrent startup.
+     */
+    private static Address prepareJGroupsAddress(KeycloakSession session) {
+        var storage = session.getProvider(ServerConfigStorageProvider.class);
+        String seq = storage.loadOrCreate(JGROUPS_ADDRESS_SEQUENCE, () -> "0");
+        long value = Long.parseLong(seq) + 1;
+        String newSeq = Long.toString(value);
+        storage.replace(JGROUPS_ADDRESS_SEQUENCE, seq, newSeq);
```
ryanemerson, line 239: Don't we need to handle the case where the replace fails, otherwise we will potentially generate two addresses on different nodes with the same seq number?

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ahus1: TODO for Alex: Also test with site name before finalizing this PR (use the SPI option) as that forces an Extended UUID on ISPN side of things.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ahus1: This is all handled in one transaction, and the DB transaction will be rolled back if it fails. So those addresses that pass through this one are guaranteed to be consecutive, ordered, and visible to all other nodes.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ahus1: I found that JGroups can only serialize a UUID or a SiteUUID due to the code in Util.addressToString. For a plain ExtendedUUID it will return null, which leads to followup-errors, starting with a the value being rejected by the database. 

I updated the code to use SiteUUID, and for that I needed to also use a site. Please have a look, maybe we could pair for any additional changes that might be necessary.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ahus1: This should be fixed with this commit as well.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
ryanemerson: I'm not sure of the repurcussions of using `SiteUUID` vs `ExtendedUUID` here, but it seems to me that `ExtendedUUID` should be handled by `Util.addressToString` as I believe the current implementation will cause issues with Infinispan 16 users relying on JDBC_PING2.

pruivo: -1 to use SiteUUID. 
@belaban can this https://github.com/belaban/JGroups/pull/901 be backported to 5.3.x branch?

ahus1: I found that when setting `spi-cache-embedded--default--site-name` this already tries to use a ExtendedUUID today, but it then breaks with the NULL problem when using JDBC_PING2. 

I'm reverting my previous commit. It seems we'll need to state that `spi-cache-embedded--default--site-name` is incompatible with JDBC_PING2?!

ahus1: I'm now mapping an ExtendedUUID to a UUID when passing it down to JDBC_PING as I don't know when the backport will land to and to see if it now breaks in a different place.

ahus1:
@pruivo - the build is green and the code blocks moved around one last time. Please review this one when you have the time. Thanks!

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
```diff
@@ -0,0 +1,120 @@
...
+        ClassConfigurator.addProtocol((short) 1026, KEYCLOAK_JDBC_PING2_FOR_TESTING.class);
+    }
+
+    @Test
+    public void testClusterFormedAfterRestart() throws Exception {
+        try(var a=createChannel("jdbc-h2.xml", "A")) {
+            a.connect(CLUSTER);
+            for(int i=1; i <= 10; i++) {
+                long start=System.nanoTime();
+                try(var b=createChannel("jdbc-h2.xml", "B")) {
```
pruivo:
```suggestion
                try(var b=createChannel(PROTOCOAL_STACK, Character.toString('A' + i))) {
```

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
```diff
@@ -0,0 +1,120 @@
...
+    protected static final int NUM_NODES=8;
+    public static final String PROTOCOAL_STACK = "jdbc-h2.xml";
+
+    static {
+        ClassConfigurator.addProtocol((short) 1026, KEYCLOAK_JDBC_PING2_FOR_TESTING.class);
+    }
+
+    @Test
+    public void testClusterFormedAfterRestart() throws Exception {
+        try(var a=createChannel("jdbc-h2.xml", "A")) {
```
pruivo:
```suggestion
        try(var a=createChannel(PROTOCOAL_STACK, "A")) {
```

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
```diff
@@ -0,0 +1,120 @@
...
+ * Misc tests for {@link KEYCLOAK_JDBC_PING2}, running against H2
+ * @author Bela Ban
+ * @author Alexander Schwartz
+ */
+public class JdbcPing2Test {
+    protected static Logger log = Logger.getLogger(JdbcPing2Test.class);
+
+    protected static final String CLUSTER="jdbc-test";
+    protected static final int NUM_NODES=8;
+    public static final String PROTOCOAL_STACK = "jdbc-h2.xml";
```
pruivo:
```suggestion
    public static final String PROTOCOL_STACK = "jdbc-h2.xml";
```

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private static void runSingleTest() throws Exception {
+        JChannel[] channels=new JChannel[NUM_NODES];
+        for (int i = 0; i < channels.length; i++) {
+            channels[i] = createChannel(PROTOCOAL_STACK, String.valueOf(i + 1));
+        }
+        CountDownLatch latch=new CountDownLatch(1);
+        int index=1;
+        List<Thread> threads=new ArrayList<>();
```
pruivo: Wait for the threads to finish?

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        latch.countDown();
+        long start = System.nanoTime();
+        Util.waitUntilAllChannelsHaveSameView(20000, 100, channels);
+        long time = System.nanoTime() - start;
+        log.infof("-- cluster of %d formed in %s:\n%s\n", NUM_NODES, Duration.ofNanos(time),
+                Stream.of(channels).map(ch -> String.format("%s: %s", ch.address(), ch.view()))
+                        .collect(Collectors.joining("\n")));
+        Arrays.stream(channels).filter(ch -> ch.view().getCoord() != ch.getAddress()).forEach(JChannel::close);
+        Arrays.stream(channels).filter(ch -> !ch.isClosed()).forEach(JChannel::close);
```
pruivo: It's probably safer to put this in a finally block. If the test fails, the channels won't be closed.

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
```diff
@@ -0,0 +1,120 @@
...
+        log.infof("Closed");
+    }
+
+    protected static JChannel modify(JChannel ch) {
+        GMS gms=ch.stack().findProtocol(GMS.class);
+        gms.setJoinTimeout(3000).setMaxJoinAttempts(5);
```
pruivo: Why not configure this in the XML?

model/infinispan/src/test/resources/jdbc-h2.xml
```diff
@@ -0,0 +1,49 @@
...
+            connection_url="jdbc:h2:mem:test;DB_CLOSE_DELAY=-1"
+            connection_username=""
+            connection_password=""
+            remove_all_data_on_view_change="true"
+            register_shutdown_hook="true"
+            return_entire_cache="false"
+            write_data_on_find="true"
+    />
+    <!-- very aggressive merging to speed up the test -->
+    <MERGE3 min_interval="150"
```
pruivo: This is making the tests pass because it runs the partition handling check every 200ms. In production, it runs every 30s. 

Using the production values, the test fails randomly
```
[ERROR]   JdbcPing2Test.testConcurrentStartupMultipleTimes:60->runSingleTest:89 » Timeout Timeout 20000 kicked in, views are:
1: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
2: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
3: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
4: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
5: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
6: [6|0] (1) [6]
7: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
8: [8|4] (7) [8, 1, 4, 7, 2, 5, 3]
```

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/JGroupsConfigurator.java
```diff
@@ -201,12 +213,53 @@ private static void configureDiscovery(ConfigurationBuilderHolder holder, Keyclo
...
+        ExtendedUUID address = new ExtendedUUID(0, value);
+
+
+        cp.getEntityManager().runWithConnection(o -> {
+            Connection con = (Connection) o;
+            try (PreparedStatement s = con.prepareStatement(statement)) {
+                s.setString(1, org.jgroups.util.Util.addressToString(new UUID(address.getMostSignificantBits(), address.getLeastSignificantBits()))); // address
+                s.setString(2, "(starting)"); // name
+                s.setString(3, clusterName); // cluster name
+                s.setString(4, new IpAddress("localhost", 0).toString()); // ip
```
pruivo: `new IpAddress("localhost", 0).toString()` should have always the same output. It can be cached.

pruivo (commented):
I got `JdbcSQLIntegrityConstraintViolationException` locally, running `testConcurrentStartupMultipleTimes` test.
```
19:16:52,381 ERROR jgroups-6,jdbc-test,4 [org.keycloak.jgroups.protocol.KEYCLOAK_JDBC_PING2_FOR_TESTING] 4: failed writing to DB: org.h2.jdbc.JdbcSQLIntegrityConstraintViolationException: Unique index or primary key violation: "PUBLIC.PRIMARY_KEY_C ON PUBLIC.JGROUPS(ADDRESS) VALUES ( /* 751 */ 'uuid://11e06b89-e3c2-4363-9794-ffd49d283280' )"; SQL statement:
```

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
ahus1: part of next commit

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
ahus1: I've inherited it :D  ... moved to XML in next commit

model/infinispan/src/test/resources/jdbc-h2.xml
ahus1: The faster merge interval is there to speed up the test. In production with a longer interval, it will take longer, but it will eventually converge. IMHO any sub-second interval will be too low for production. Still we can go for 5 seconds or so?!

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/JdbcPing2Test.java
ahus1: done in next commit

ahus1:
@pruivo - updated the PR.
The exception JdbcSQLIntegrityConstraintViolationException is something that can happen when there are attempts of concurrent writes (for example the coordinator writes a node, and the node itself writes their information). That should then heal itself on the next attempt. 

Ready for another round of reviews.

model/infinispan/src/test/resources/jdbc-h2.xml
pruivo: > In production with a longer interval, it will take longer, but it will eventually converge

So you agree we don't need this PR :smile: 

> IMHO any sub-second interval will be too low for production.

The LinkedIn blog recommended increasing it to minutes... 

On a serious note, I will change my opinion regarding last night and state that the test should be removed. With `MERGE3`, either you use `JDBC_PING2` or `KEYCLOAK_JDBC_PING2`, the test passes. Without it, it fails with either protocol. I can conclude the test is not testing anything useful. WDYT?

ahus1: > So you agree we don't need this PR  ... the test should be removed

The test is more a stress testing harness that helped me to test KEYCLOAK_JDBC_PING2 against H2, and discover possible problems / areas to improve the converging of the cluster. With the other change of having a pre-seeded UUID which is not part of the added test, this should be necessary less often, but still when clusters split. 

So I would then remove the test, and keep the (ignored) stress test that's running 100 times?

pruivo (approved):
LGTM.

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -70,6 +184,51 @@ protected void writeToDB(PingData data, String clustername) throws SQLException
...
+        return super.setAddress(addr);
+    }
+
+    @Override
+    protected void delete(Connection conn, String clustername, Address addressToDelete) throws SQLException {
+        super.delete(conn, clustername, toUUID(addressToDelete));
+    }
+
+    @Override
+    protected void delete(String clustername, Address addressToDelete) throws SQLException {
```
ryanemerson, line 209: Do we also need to overwrite `callInsertStoredProcedure` as this calls `Util#addressToString` and is called from `writeToDB`?

ahus1: we don't use it, so no

ryanemerson: Ok, I take your point as the `toUUID` should not be necessary in JGroups once https://github.com/belaban/JGroups/pull/901 has been backported. There is a lot of flags we don't use in the Keycloak use-case that we're still including here. I assume the reason for inclusion is so that our solution here is as close to the PR that will eventually be issued to JGroups for JDBC_PING2.

PR #41294, opened 2025-07-21
Remove obsolete Liquibase FK snapshot generator

### LINKED ISSUE #41293: Remove obsolete Liquibase FK snapshot generator
### Description

When debuging some issues with Liquibase, I've encountered a debug log from the Liquibase:

```bash
2025-07-21 15:16:01,328 WARN  [liquibase.snapshot.SnapshotGeneratorFactory] (main) org.keycloak.connections.jpa.updater.liquibase.CustomForeignKeySnapshotGenerator class does not properly implement the 'getPriority(Class<? extends DatabaseObject>, Database)' method and may incorrectly override other snapshot generators causing unexpected behavior. Please report this to the Liquibase developers or if you are developing this change please fix it ;)
```

As we have the `CustomForeignKeySnapshotGenerator` class defined in our codebase, it is loaded via service locator and managed. 

As it should fix the issue https://liquibase.jira.com/browse/CORE-3141, we should remove it as we already using the much higher version of Liquibase that contains the fix for the issue. 

It might provoke some unintentional behavior, so we should remove it.

mabartos:
- Closes #41293

@rmartinc Could you please check it? More details in the parent issue.

We have some places in KC codebase where we use some custom logic for Liquibase. We should revisit it and decide if it's still relevant, as maintaining the additional legacy code might be hard and very error-prone.


rmartinc (approved):
Thanks @mabartos!Yes, all these workarounds can be removed if the current version of liquibase does not need them anymore.

ahus1 (approved):

PR #41320, opened 2025-07-22
Refactor test database config in the new framework

### LINKED ISSUE #41319: Test database config refactor
@vaceksimon - FYI, Ryan added an option to initialize the database. Please have a look as a follow-up to see if additional cleanup/refactoring would be necessary as a follow-up. Thanks!

_Originally posted by @ahus1 in https://github.com/keycloak/keycloak/issues/40999#issuecomment-3084139178_

vaceksimon:
Closes #41319

### DIFF
--- a/test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractDatabaseSupplier.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractDatabaseSupplier.java
@@ -30,7 +30,7 @@ public abstract class AbstractDatabaseSupplier implements Supplier<TestDatabase,
 
     @Override
     public boolean compatible(InstanceContext<TestDatabase, InjectTestDatabase> a, RequestedInstance<TestDatabase, InjectTestDatabase> b) {
-        return true;
+        return a.getAnnotation().config().equals(b.getAnnotation().config());
     }
 
     @Override
--- a/test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractDatabaseSupplier.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractDatabaseSupplier.java
@@ -18,10 +18,10 @@ public abstract class AbstractDatabaseSupplier implements Supplier<TestDatabase,
     public TestDatabase getValue(InstanceContext<TestDatabase, InjectTestDatabase> instanceContext) {
         DatabaseConfigBuilder builder = DatabaseConfigBuilder
               .create()
-              .withPreventReuse(instanceContext.getLifeCycle() != LifeCycle.GLOBAL);
+              .preventReuse(instanceContext.getLifeCycle() != LifeCycle.GLOBAL);
 
-        DatabaseConfigurator configurator = SupplierHelpers.getInstance(instanceContext.getAnnotation().config());
-        configurator.configure(builder);
+        DatabaseConfig config = SupplierHelpers.getInstance(instanceContext.getAnnotation().config());
+        builder = config.configure(builder);
 
         TestDatabase testDatabase = getTestDatabase();
         testDatabase.start(builder.build());
--- a/test-framework/core/src/main/java/org/keycloak/testframework/database/DatabaseConfigRepresentation.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/database/DatabaseConfigRepresentation.java
@@ -1,0 +1,31 @@
+package org.keycloak.testframework.database;
+
+public final class DatabaseConfigRepresentation {
+    private String initScript;
+    private String database;
+    private boolean preventReuse;
+
+    public String getInitScript() {
+        return initScript;
+    }
+
+    public void setInitScript(String initScript) {
+        this.initScript = initScript;
+    }
+
+    public String getDatabase() {
+        return database;
+    }
+
+    public void setDatabase(String database) {
+        this.database = database;
+    }
+
+    public boolean doesPreventReuse() {
+        return preventReuse;
+    }
+
+    public void setPreventReuse(boolean preventReuse) {
+        this.preventReuse = preventReuse;
+    }
+}
--- a/tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
@@ -6,8 +6,8 @@ import org.keycloak.admin.client.resource.RolesResource;
 import org.keycloak.representations.idm.RoleRepresentation;
 import org.keycloak.testframework.annotations.InjectClient;
 import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
-import org.keycloak.testframework.config.Config;
 import org.keycloak.testframework.database.TestDatabase;
+import org.keycloak.testframework.injection.Extensions;
 import org.keycloak.testframework.realm.ManagedClient;
 import org.keycloak.testframework.realm.RoleConfigBuilder;
 
@@ -18,8 +18,7 @@ public abstract class AbstractDBSchemaTest {
     ManagedClient managedClient;
 
     protected static String dbType() {
-        String database = Config.getSelectedSupplier(TestDatabase.class);
-        return database == null ? "dev-mem" : database;
+        return Extensions.getInstance().findSupplierByType(TestDatabase.class).getAlias();
     }
 
     @Test
@@ -30,9 +29,7 @@ public abstract class AbstractDBSchemaTest {
                 .singleAttribute("role1-attr-key", "role1-attr-val")
                 .build();
         RolesResource roles = managedClient.admin().roles();
-        Assertions.assertDoesNotThrow(() -> {
-            roles.create(role1);
-            roles.deleteRole(role1.getName());
-        });
+        roles.create(role1);
+        roles.deleteRole(role1.getName());
     }
 }
--- a/tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
@@ -1,0 +1,33 @@
+package org.keycloak.tests.db;
+
+import org.keycloak.admin.client.resource.RolesResource;
+import org.keycloak.representations.idm.RoleRepresentation;
+import org.keycloak.testframework.annotations.InjectClient;
+import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
+import org.keycloak.testframework.config.Config;
+import org.keycloak.testframework.database.TestDatabase;
+import org.keycloak.testframework.realm.ManagedClient;
+import org.keycloak.testframework.realm.RoleConfigBuilder;
+
+@KeycloakIntegrationTest
+public class AbstractDBSchemaTest {
+
+    @InjectClient
+    ManagedClient managedClient;
+
+    protected static String dbType() {
+        String database = Config.getSelectedSupplier(TestDatabase.class);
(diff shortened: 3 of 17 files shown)


tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @InjectClient
+    ManagedClient managedClient;
+
+    protected static String dbType() {
+        String database = Config.getSelectedSupplier(TestDatabase.class);
+        return database == null ? "dev-mem" : database;
+    }
+
+    protected void createDeleteRole() {
```
ryanemerson: Can't we define the `@Test` method here so that we don't have to override in each of the implementation classes?

lhanusov (changes requested):
@vaceksimon please, move the test to the abstract class as Ryan proposed

tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
vaceksimon: In my opinion, it makes it harder to read each test class. Seeing a test fail in a class that doesn't have a test method can be confusing. I find it cleaner for the child classes to have explicit test methods rather than saving a few lines and hiding the method in the parent.

wdyt @ryanemerson @lhanusov ?

ryanemerson: In this case I don't think it's a big deal as there's only one test, so I don't have strong views. In general I don't think this approach scales though, as it's possible/probable that the author of the implementation class could forget to add all of the required `@Test` methods as the number of test cases increases. If we define the `@Test` in the abstract class, it's not possible for the implementation class to not define them.

stianst: I agree with @ryanemerson; just use `@Test` in `AbstractDBSchemaTest` rather than having to wire it up separately in the concrete classes.

vaceksimon:
@ryanemerson @lhanusov you can recheck. Once it's ready, PR can be squashed and merged

ryanemerson (dismissed):
LGTM, thanks for the PR @vaceksimon

lhanusov (approved):

test-framework/core/src/main/java/org/keycloak/testframework/database/DatabaseConfigRepresentation.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public String getDatabase() {
+        return database;
+    }
+
+    public void setDatabase(String database) {
+        this.database = database;
+    }
+
+    public boolean doesPreventReuse() {
```
stianst: Would just stick with standard get/set conventions here (`does` sounds strange as well ;)):

```suggestion
    public boolean isPreventReuse() {
```

tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+@KeycloakIntegrationTest
+public abstract class AbstractDBSchemaTest {
+
+    @InjectClient
+    ManagedClient managedClient;
+
+    protected static String dbType() {
+        String database = Config.getSelectedSupplier(TestDatabase.class);
+        return database == null ? "dev-mem" : database;
```
stianst: Should not hard-code the default database here, as that could change. If this PR is merged after https://github.com/keycloak/keycloak/pull/41358, then it can be updated to:
```suggestion
        return Extensions.getInstance().findSupplierByType(TestDatabase.class).getAlias();
```

tests/base/src/test/java/org/keycloak/tests/db/CaseSensitiveSchemaTest.java
```diff
@@ -1,70 +1,45 @@
...
-        RolesResource roles = managedClient.admin().roles();
-        roles.create(role1);
-        roles.deleteRole(role1.getName());
-    }
-
-    protected static String dbType() {
-        String database = Config.getSelectedSupplier(TestDatabase.class);
-        return database == null ? "dev-mem" : database;
-    }
+    @InjectTestDatabase(lifecycle = LifeCycle.CLASS, config = CaseSensitiveDatabaseConfig.class)
```
stianst: Doesn't strictly need to be LifeCycle.CLASS, as the test framework should figure out that the DB is not compatible and restart things as needed.

See my comment above on updating `compatible` method for `AbstractDatabaseSupplier`

test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractDatabaseSupplier.java
stianst, line 1: The `compatible` method needs to be updated to compare the annotations now that a test can configure the database

test-framework/core/src/main/java/org/keycloak/testframework/database/DatabaseConfigRepresentation.java
stianst, line 1: `Representation` is Keycloak's naming convention for JSON representation classes, and this isn't a JSON thing, maybe it would be better to name this `DatabaseConfiguration` or something like that?

vaceksimon:
@stianst all comments from reviews are now addressed in the last added commit. Thanks everyone for the reviews

tests/base/src/test/java/org/keycloak/tests/db/AbstractDBSchemaTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test
+    public void testCaseSensitiveSchema() {
+        RoleRepresentation role1 = RoleBuilder.create()
+                .name("role1")
+                .description("role1-description")
+                .singleAttribute("role1-attr-key", "role1-attr-val")
+                .build();
+        RolesResource roles = managedClient.admin().roles();
+        Assertions.assertDoesNotThrow(() -> {
```
ryanemerson: What's the advantage of having this code wrapped by `assertDoesNotThrow`, the test will fail without this if either `roles.*` operation throws an exception.

Looking at the rest of the code base it seems we use `assertDoesNotThrow` sparingly.

vaceksimon: Good question! A while back, I also noticed the lack of `assertDoesNotThrow` in the tests and wondered why that is. I believe the only thing it does is make it clear what is supposed to (not) happen.

I put the junit assertion in here to find out why it shouldn't be used.

ryanemerson: Personally I prefer to reduce verbosity of the tests and assume that expected exceptions in tests are asserted/handled and unexpected ones fail the test by default. I can appreciate the other side of the argument though, so I don't have strong views on this, but it would be nice to take a consistent approach with the new testsuite.

vaceksimon: Yeah, that makes sense. Being consistent with the rest of the codebase is a good call. I'll remove the assertion

stianst (dismissed):

tests/base/src/test/java/org/keycloak/tests/db/CaseSensitiveSchemaTest.java
```diff
@@ -1,71 +1,45 @@
...
     }
 
-    public static class DatabaseConfigurator implements org.keycloak.testframework.database.DatabaseConfigurator {
+    public static class CaseSensitiveDatabaseConfig implements DatabaseConfig {
         @Override
-        public DatabaseConfigBuilder configure(DatabaseConfigBuilder builder) {
+        public DatabaseConfigBuilder configure(DatabaseConfigBuilder database) {
             if (PostgresTestDatabase.NAME.equals(dbType())) {
-                builder.withInitScript("org/keycloak/tests/db/case-sensitive-schema-postgres.sql");
+                database.initScript("org/keycloak/tests/db/case-sensitive-schema-postgres.sql");
```
stianst, line 40:
```suggestion
                database.preventReuse(true).initScript("org/keycloak/tests/db/case-sensitive-schema-postgres.sql");
```

tests/base/src/test/java/org/keycloak/tests/db/PreserveSchemaCaseLiquibaseTest.java
```diff
@@ -31,13 +31,13 @@ public KeycloakServerConfigBuilder configure(KeycloakServerConfigBuilder config)
         }
     }
 
-    public static class DatabaseConfigurator implements org.keycloak.testframework.database.DatabaseConfigurator {
+    private static class PreserveSchemaCaseDatabaseConfig implements DatabaseConfig {
         @Override
-        public DatabaseConfigBuilder configure(DatabaseConfigBuilder builder) {
+        public DatabaseConfigBuilder configure(DatabaseConfigBuilder database) {
             if (dbType().equals(PostgresTestDatabase.NAME)) {
```
stianst, line 37: database.preventReuse(true);

tests/base/src/test/java/org/keycloak/tests/db/PreserveSchemaCaseLiquibaseTest.java
```diff
@@ -2,24 +2,24 @@
...
-@KeycloakIntegrationTest(config = PreserveSchemaCaseLiquibaseTest.KeycloakConfig.class)
+@KeycloakIntegrationTest(config = PreserveSchemaCaseLiquibaseTest.PreserveSchemaCaseServerConfig.class)
 // MSSQL does not support setting the default schema per session.
 // Oracle image does not support configuring user/databases with '-'
 @DisabledForDatabases({ "mssql", "oracle" })
-public class PreserveSchemaCaseLiquibaseTest extends CaseSensitiveSchemaTest {
+public class PreserveSchemaCaseLiquibaseTest extends AbstractDBSchemaTest {
 
-    @InjectTestDatabase(lifecycle = LifeCycle.CLASS, config = DatabaseConfigurator.class)
+    @InjectTestDatabase(config = PreserveSchemaCaseDatabaseConfig.class)
```
stianst:
```suggestion
    @InjectTestDatabase(config = PreserveSchemaCaseDatabaseConfig.class, lifecycle = LifeCycle.CLASS)
```

tests/clustering/src/test/java/org/keycloak/tests/clustering/JdbcPingCustomSchemaTest.java
```diff
@@ -7,9 +7,9 @@
 import org.keycloak.testframework.injection.LifeCycle;
 import org.keycloak.tests.db.CaseSensitiveSchemaTest;
 
-@KeycloakIntegrationTest(config = CaseSensitiveSchemaTest.KeycloakConfig.class)
+@KeycloakIntegrationTest(config = CaseSensitiveSchemaTest.CaseSensitiveServerConfig.class)
 public class JdbcPingCustomSchemaTest {
-    @InjectTestDatabase(lifecycle = LifeCycle.CLASS, config = CaseSensitiveSchemaTest.DatabaseConfigurator.class)
+    @InjectTestDatabase(config = CaseSensitiveSchemaTest.CaseSensitiveDatabaseConfig.class)
```
stianst:
```suggestion
    @InjectTestDatabase(config = CaseSensitiveSchemaTest.CaseSensitiveDatabaseConfig.class, lifecycle = LifeCycle.CLASS)
```

lhanusov (approved):

ryanemerson (approved):

stianst (approved):

PR #41326, opened 2025-07-22
Move RealmTest.java to the new testsuite

lhanusov:
Part of: #34494


tests/base/src/test/java/org/keycloak/tests/admin/realm/RealmCacheTest.java
```diff
@@ -0,0 +1,68 @@
...
+        assertTrue(runOnServer.fetch(s -> {
+            InfinispanConnectionProvider provider = s.getProvider(InfinispanConnectionProvider.class);
+            Cache<Object, Object> cache = provider.getCache("realms");
+            return cache.containsKey(realmId);
+        }, Boolean.class));
+
+        managedRealm.admin().clearRealmCache();
+        AdminEventAssertion.assertEvent(adminEvents.poll(), OperationType.ACTION, "clear-realm-cache", ResourceType.REALM);
+
+        assertFalse(runOnServer.fetch(s -> {
```
lhanusov: @stianst this test fails, because the cache is still present, it seems the clearRealmCache() method doesn't work properly, do you think it's a bug?

stianst: Nah, the issue is on the testing side. The problem is that both `adminEvents.poll` and `runOnServer.fetch` is using the same realm as you are clearing, both those methods will reload the realm from the database and into the cache.

lhanusov: so, when userClearCache passes, means, there is a different workflow behind?

lhanusov: @stianst thanks, it makes sense to me now, I see the difference between realm and user endpoint

lhanusov: jfyi: it clears the cache within the session call now

stianst: That's an elegant approach, but sadly it will not then actually test the Admin APIs, as you are bypassing those to directly clear the caches. In reality all the admin endpoints do is call exactly that, but in theory it could be broken in the future.

lhanusov: yup, this got into my mind this morning when I was cycling to the Brno office :-D

stianst (approved):

PR #41329, opened 2025-07-22
Disable webauthn buttons after click

### LINKED ISSUE #41037: WebAuthN Setup: OperationError: A request is already pending.
### Before reporting an issue

### Area

authentication/webauthn

### Describe the bug

When a user register a new passkey on Windows and double click the button for registering passkey, the following error occurs: OperationError: A request is already pending.

### Version

26.2.1

### Regression

### Expected behavior

Expected: The user can't doubleclick on the button. Our adapted webauthn-register.ftl template looks like:

`
        <script type="module">
            import {registerByWebAuthn} from "";

            const registerButton = document.getElementById('registerWebAuthn');
            registerButton.addEventListener("click", async function() {
                registerButton.disabled = true;
                const input = {
                    challenge : '${challenge}',
                    userid : '${userid}',
                    username : '${username}',
                    signatureAlgorithms : [<#list signatureAlgorithms as sigAlg>${sigAlg?c},</#list>],
                    rpEntityName : '${rpEntityName}',
                    rpId : '${rpId}',
                    attestationConveyancePreference : '${attestationConveyancePreference}',
                    authenticatorAttachment : '${authenticatorAttachment}',
                    requireResidentKey : '${requireResidentKey}',
                    userVerificationRequirement : '${userVerificationRequirement}',
                    createTimeout : ${createTimeout},
                    excludeCredentialIds : '${excludeCredentialIds}',
                    initLabel : "${msg("webauthn-registration-init-label")?no_esc}",
                    initLabelPrompt : "${msg("webauthn-registration-init-label-prompt")?no_esc}",
                    errmsg : "${msg("webauthn-unsupported-browser-text")?no_esc}"
                };
                await registerByWebAuthn(input);
                registerButton.disabled = false;
            });
        </script>
`

### Actual behavior

The button is not blocked so the JS api "navigator.credentials.create" is called twice.

### How to Reproduce?

Enable WebAuthN feature and setup Passkey.

### Anything else?

_No response_

rmartinc:
Closes #41037

In windows there are issues if the webauthn buttons or links (registration and authentication) are double clicked fast. The idea is the button is just for one one-click. When it's clicked the webauthn form is always submitted with the success or the error to the server and the page is refreshed. I'm changing from the initial `disabled=true` idea presented by the reporter to use `{ once : true }`. The `disabled` solution does not work for links (`<a>` elements), which are used for the authentication part in passkeys (we have a mix of buttons and links).

The modified test is because it doesn't work OK now or before. Previously it returned the `a request is already pending` message because the first registration hung, and clicking the button again returned the error we are avoiding now with this PR. Now the button does nothing when clicked, and I just added a timeout to make it fail. The correct usage for that test is that the selenium webauthn authenticator should return an error saying that it is not possible to create a key with those settings (but it seems that selenium has a bug or similar).

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/util/WaitUtils.java
+++ b/testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/util/WaitUtils.java
@@ -19,6 +19,7 @@ package org.keycloak.testsuite.util;
 import org.jboss.arquillian.graphene.wait.ElementBuilder;
 import org.keycloak.executors.ExecutorsProvider;
 import org.keycloak.testsuite.client.KeycloakTestingClient;
+import org.keycloak.testsuite.pages.AbstractPage;
 import org.openqa.selenium.By;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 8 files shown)


jonkoops (approved):

ahus1 (approved):
Approving based on existing review.

PR #41332, opened 2025-07-22
Passkeys support in IdpUsernamePasswordForm

### LINKED ISSUE #41259: Passkeys support in IdpUsernamePasswordForm
### Description

It can be good if passkeys support is available in `IdpUsernamePasswordForm` .

This authenticator is used usually during first-broker-login to re-authenticate the user in case that some user authenticated in the identity provider (EG. `facebook`) for the first time with existing username/email of some Keycloak user `john` , but that Keycloak user `john` does not yet have link to the `facebook` identity provider.

However it can easily happen that user `john` has passkeys and hence he might want to re-authenticate in first-broker-login with this conditional passkey.

### Additional note

I have a prototype in this branch https://github.com/mposolda/keycloak/tree/41259-fbl-passkeys . However not have time to do more testing with that right now and leaving on PTO now. Especially no time to finish an automated test. Automated test could be a bit tricky (as it may require both identity-provider setup as well as passkeys setup, when both of those are using inheritance for the test setup...), however hopefully doable.

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #41259

Just adding passkeys for the IdpUsernamePasswordForm. It was started by @mposolda and I have mainly added the test class.

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpUsernamePasswordForm.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpUsernamePasswordForm.java
@@ -81,12 +81,7 @@ public class IdpUsernamePasswordForm extends UsernamePasswordForm {
         Optional<UserModel> existingUser = getExistingUser(context);
         existingUser.ifPresent(context::setUser);
 
-        boolean result = validateUserAndPassword(context, formData);
-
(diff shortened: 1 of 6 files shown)


graziang (approved):
@rmartinc @mposolda thanks!

abstractj (approved):

PR #41333, opened 2025-07-22
[Admin v2] Querying draft

vmuzikar:
This PR is meant as a discussion with concrete example (hence not opening it just as a GH Discussion) of how querying could work, it is not meant as a final result (even just for the prototype).

The design choices made here are made with simplicity in mind, i.e. it does not itroduce arbitrary existing formats or libraries. This allows us to have full control over what expressions we support and what we don't support, while keeping the implementation relatively simple.

It captures two parts:
* Query syntax, incl. the parsing.
* Resource filtering based on the parsed query.

### Query syntax

Uses a custom syntax inspired by GitHub APIs. We're making a few assumptions here to keep it simple which implies a few limitations.

* Simple expression: `field:value`
* Combined expression: `field1:value1 field2:value2`
    * Treated as AND.
* Quoting: `field:"value with spaces and :colons"`
    * This allows spaces and colons in the expressions.
    * Escaping quotes is not supported, i.e. expressions can't contain quotes.
* Nested fields: `field.subfield:value`
* Lists and maps: `field:[item1, item2]` or `field:[key1:val1, key2:val2]`
    * List syntax can be used to check presence of a key in a map too. E.g. to lookup resources that contain an attribute named `phone`, we could do something like `attrs:[phone]`.
    * Lists and maps are treated as "contains", not "equals". I.e. the expression means "I'm looking for resources that include these elements in this collection field".
    * Nested fields are not supported.
    * Quotes (i.e. spaces and colons in the expressions) are not supported.
    * Commas in the expressions are not supported.
    * No support for nested collections.

Additional limitations:
* Expressions implicitly support only AND, i.e. more complex expressions with OR are not supported.
* NOT is not supported at the moment, but could be relatively easily added.
* Everything is compared for equality as a String.

#### Alternatives

[SCIM query syntax](https://bookstack.soffid.com/books/scim/page/scim-query-syntax) is an obvious candidate though there are a few considerations to make around that.

* Full SCIM support as such is out of scope of the Admin v2, that’s a completely different effort.
* When we add full support for SCIM, it will be different endpoints from Admin APIs.
* SCIM is primarily meant to be used for provisioning users and groups.
* So if we adopted SCIM for querying in Admin v2, it would be just for the syntax.
    * It might be confusing to the users as we'd support just the syntax but not the schema, endpoints etc.
    * The main benefit of that would be a familiar syntax for users, and also that we could potentially leverage some library for parsing the queries. And we could also share the logic when we adopt SCIM.
    * The downside is the complexity. And if we stripped it down just to some basic operators, it wouldn’t brought too many benefits and users would naturally expect we will add more if it's too similar to the SCIM query syntax.
    * But problem is also which library we could use. There’s not that many, one of the most popular ones seem to be [pingidentity/scim2](https://github.com/pingidentity/scim2) but it seems to have a single maintainer, only a few forks and it’s not very active. I would hesitate to adopt it in KC.

### Resource filtering

* Works in general for all fields.
    * We might want to introduce an exclusion mechanism (e.g. using annotations) for fields we do not want to be searchable.
* Done at runtime on the result set fetched from the DB.
    * This doesn't help the server performance, but network and clients would benefit from it.
    * Can be easily extended to do optimized SQL queries before the runtime filtering to help the server side performance.
* Currently uses reflections.
    * Very poor performance.
    * We should consider some alternative approach.
        * Maybe filtering somewhere at the Jakson level when serializing? It would not help if we needed to do some additional processing of the results before serializing.
        * Or static class generation at build time to hardcode the filtering for all fields (which we know at the build time) of all supported representations?

### DIFF
--- a/core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
@@ -1,0 +1,31 @@
+package org.keycloak.representations.admin.v2;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+public class BaseRepresentation {
+
+    @JsonIgnore
+    protected Map<String, Object> additionalFields = new LinkedHashMap<String, Object>();
+
+    @JsonAnyGetter
+    public Map<String, Object> getAdditionalFields() {
+        return additionalFields;
+    }
+
+    @JsonAnySetter
+    public void setAdditionalField(String name, Object value) {
+        this.additionalFields.put(name, value);
+    }
+
+    public void setAdditionalFields(Map<String, Object> additionalFields) {
+        this.additionalFields = additionalFields;
+    }
+
+}
--- a/core/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
@@ -4,10 +4,13 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 
+import java.util.LinkedHashSet;
 import java.util.Set;
 
-@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
 public class ClientRepresentation {
+
+    public static final String OIDC = "openid-connect";
 
     @JsonProperty(required = true)
     @JsonPropertyDescription("ID uniquely identifying this client")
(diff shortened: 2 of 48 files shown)


shawkins:
> We might want to introduce an exclusion mechanism (e.g. using annotations) for fields we do not want to be searchable.

What fields would you not want to be searchable?

> Can be easily extended to do optimized SQL queries before the runtime filtering to help the server side performance.

In situations where logical operations take place after filtering you have to be careful. Partial push-down of predicates means you can't also pushdown limit / offset, or grouping should that ever need to be supported. You could also consider having formal parameters for things that are indexed, rather than looking for them in the q parameter.

> Maybe filtering somewhere at the Jakson level when serializing?

There is subtlety here. Reflection / toString may not capture the actual Jackson mapping. And it flattens the type model - everything is a string, which may be problematic for numerical values - especially floating point numbers or if we introduce greater than / less than.

You could entertain simply supporting json path predicates on top of the result rather than a custom query syntax. Of course json path could be used for projection as well.

stianst:
Quite related to search is sort, do we have some discussion around that?

stianst:
I would like to see some research on different approaches in existing APIs (I mostly like GH search syntax, but..), as well as looking for if there is some existing libraries that can be useful rather than inventing/implementing our own.

I'd also like to see some more concrete use-cases/examples; for example find me all clients that have at least one web-origin set, or all clients that do use a non-https redirect-uri. Expand the examples also to users, as that would have different use-cases.

stianst:
I don't like commenting/discussing this through a PR (as this PR could be closed and then discussions would be somewhat lost), would be better to use a discussion.

Pepo48:
> Quite related to search is sort, do we have some discussion around that?

https://github.com/keycloak/keycloak/discussions/41334

PR #41336, opened 2025-07-22
Kerberos Server Principal field now trims whitespace

### LINKED ISSUE #41335: Kerberos "Server Principal" value should automatically trim leading/trailing whitespace
### Description

When configuring the Kerberos User Federation provider in Keycloak, the Server Principal field accepts and stores values with leading or trailing whitespace. If a user unintentionally pastes a value with whitespace (e.g., " HTTP/service@REALM"), Kerberos authentication silently fails and no clear error is presented. This whitespace is not visible in logs or debug output, making the issue very difficult to diagnose.

### Discussion

_No response_

### Motivation

Misleading configuration behavior can waste valuable debugging time. In my case, it took two full days to identify that a hidden leading whitespace character was the root cause of failed authentication. Since Kerberos principals can never contain spaces, it would be safe and user-friendly to automatically trim this field.

### Details

Field: Server Principal in Kerberos User Federation settings

Current Behavior: Accepts input with leading/trailing whitespace

Expected Behavior: Automatically trims leading/trailing whitespace from user input on save or validation

Impact: Avoids subtle misconfigurations and hard-to-diagnose Kerberos failures

Justification: Kerberos principal syntax is strict — any whitespace renders it invalid. Silent failure without UI or log warnings is problematic for administrators.

Wenkow:
This commit ensures that any leading or trailing whitespace in the 'Server Principal' field of the Kerberos User Federation provider is automatically trimmed before saving. This prevents hidden misconfigurations that cause silent authentication failures.

Closes #41335


server-spi-private/src/main/java/org/keycloak/models/utils/RepresentationToModel.java
```diff
@@ -1026,6 +1026,8 @@ public static ComponentModel toModel(KeycloakSession session, ComponentRepresent
                         String v = itr.next();
                         if (v == null || v.trim().isEmpty()) {
                             itr.remove();
+                        } else if ("serverPrincipal".equals(k)) {
+                            itr.set(v.trim());
```
ahus1: While this technically works, the "serverPrincipal" is a Kerberos specific configuration, which is now in code path for all components. 

A slightly better place where it would be specific to Kerberos would be here: CommonKerberosConfig#getServerPrincipal and possibly with other attributes as well. 

The way I would prefer is to check the data not when reading, but when creating or updating the configuration, and trimming the values then.

ahus1 (changes requested):
Thank you for this PR. See below for a discussion of the change.

server-spi-private/src/main/java/org/keycloak/models/utils/RepresentationToModel.java
Wenkow: Wise recommendation. I don't know what made me do it in such an inappropriate place. I'm submitting a new commit.

ahus1: Thank you for the updated PR. You found a great place to trim the values when the configuration is changed. With that, the additional trim-on-read is no longer necessary, and I would like to avoid it to limit the complexity of the code, so I reverted those changes. 

I'll go ahead an approve and merge this PR.

ahus1 (approved):
Thank you for making this safe to use for you and other admins!

PR #41337, opened 2025-07-22
Update docs how to verify that a cluster has formed

pruivo:
Closes #40296

Creating a draft PR to get some feedback about the information shown by the provider. 

keycloak_8080_admin_master_console_


pruivo:
@ahus1, if you can give feedback on the provider info, I would appreciate. Let me know if it is ok show the IP & ports.

ahus1 (changes requested):
This information looks great, and it will be helpful. 

>  Let me know if it is ok show the IP & ports.

The physical addresses should not be shown, as the person with access to the console isn't the person 

Great to see a screenshot. Please re-create it with the light theme (like all other screenshots), and trim its contents so it doesn't take up so much space / doesn't show up as so much small print. Usually the menu on the left is removed completely. 

I applied some rules to the docs, and pushed the changes: 

* In subheadings, avoid upper case unless it is a word like `Keycloak`
* The list of items was shown as a numbered aka ordered list, but it is does not have an order. Therefore I turned it to an unordered list
* Don't use contractions ("it's" -> "it is")
* UI elements are shown are shown in bold
* Removed some occurrences of `{project_name}` to avoid redundancy and lenghtly titles especially in downstream

pruivo:
> The physical addresses should not be shown, as the person with access to the console isn't the person

@ahus1, with the same logic, should I remove the members list too? It is the machine hostname.

ahus1:
> with the same logic, should I remove the members list too? It is the machine hostname.

We already expose the hostname as part of the cookie today to allow for routing of sticky requests depending on the setup used ... but let's don't take that as an excuse, and remove those as well. 

Thank you for asking again.  So let's remove that one as well from the Admin UI.

pruivo:
Updated:
* Light theme screenshot.
* Removed server list from multi-site.
* Removed servers/IPs from embedded cluster.

pruivo:
Updated once more
* Use brand name and brand version. Upstream will show Infinispan 15.0.x and downstream Red Hat Data Grid 8.5.x

ahus1 (approved):
Thank you for these changes!

PR #41341, opened 2025-07-22
FAPI 2.0 Security Profile Final - Add FAPI 2.0 Final security profile as default profile of client policies

### LINKED ISSUE #41120: FAPI 2.0 Security Profile Final - Add FAPI 2.0 Final security profile as default profile of client policies
### Description

Like FAPI 1.0 baseline, FAPI 1.0 advanced, FAPI-CIBA, OAuth 2.1, FAPI 2.0 and [FAPI 2.0 + DPoP security profile](https://github.com/keycloak/keycloak/pull/35443), FAPI 2.0 security profile final is added as default profile of client policies.

There are two profiles need to be preprared as default:
 - FAPI 2.0 security profile final
 - FAPI 2.0 security profile final (using DPoP)

### Discussion

_No response_

### Motivation

Making Keycloak fully comply with FAPI 2.0 Security Profile Final specification.

### Details

_No response_

tnorimat:
closes #41120

### DIFF
--- a/docs/guides/securing-apps/partials/oidc/fapi-support.adoc
+++ b/docs/guides/securing-apps/partials/oidc/fapi-support.adoc
@@ -6,9 +6,8 @@
 * https://openid.net/specs/openid-financial-api-part-1-1_0.html[Financial-grade API Security Profile 1.0 - Part 1: Baseline]
 * https://openid.net/specs/openid-financial-api-part-2-1_0.html[Financial-grade API Security Profile 1.0 - Part 2: Advanced]
 * https://openid.net/specs/openid-financial-api-ciba-ID1.html[Financial-grade API: Client Initiated Backchannel Authentication Profile] (FAPI CIBA)
-* https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html[FAPI 2.0 Security Profile (Draft)]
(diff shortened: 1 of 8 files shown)


tnorimat:
@mposolda Hello, if possible, I would appreciate it if you could check the PR.

mposolda:
We already have security profiles `fapi-2-security-profile` and `fapi-2-dpop-security-profile` . Is it an option to update existing profiles or would it be a concern regarding backwards compatibility?

I afraid of huge amount of built-in profiles for various versions of specifications (including drafts etc), which may end with a mess...

We may possibly add support for versioning of client profiles? It can be good as it will make sure that new deployments can use the newest version, but client policies in the older deployments would still stick with the older version. Could this work?

tnorimat:
@tnorimat Hello, I am in favor of updating existing `fapi-2-security-profile` and `fapi-2-dpop-security-profile`. If it is OK for you, I will revise the PR. WDYT?

Also the same thing is applied to [FAPI 2.0 Message Singing global profile](https://github.com/keycloak/keycloak/issues/41312). If it is OK for you. I will sent its PR the same way as in this PR.

mposolda (approved):
@tnorimat Thanks for the update! It is ok to update existing profile configurations considering that it is security issue fixed in the last updates. Thanks for that!

I think that in the future, we may need support for versioning of client profiles, but maybe that is something, which we can handle later as in this PR, it should be ok to update existing configs IMO.

PR #41368, opened 2025-07-23
Create a new condition for credential type

### LINKED ISSUE #41354: Avoid OTP when logging in with passkey
### Description

Currently, when a user logs in using a passkey and also has OTP configured, the OTP is still requested as a second factor.

We would like to support the ability to skip the OTP step when logging in with a passkey, and only require OTP as a second factor when using username + password.

Ideas and possible ways:

1. Try to find a solution using the current browser flow configuration only (probably not possible).
2. Investigate if can be implemented using step-up authentication and document the flow.
3. Create a new Condition to detect if the user authenticated using a passkey (or specific credential type), and modify the browser flow to use it.
4. Any other ideas?

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #41354

The PR creates a new conditional authenticator that gives the ability to establish a condition based on the last credential used by the user. This way the flow can execute or not different actions depending what credential was used previously by the user. The default 2FA sub-flow can be skipped with this new condition if a passkey was used before (included password or not included webauthn-passwerdless). It's a draft for now because I think there can be other options.

Comments:

1. The condition implementation is more or less generic. It allows to establish the credential types plus some fixed ones (cert, kerberos and none, none just means no credential used until now). This is OK but it makes the default browser flow more complicated.
2. Another idea I have managed is creating a straight condition to check only for passkeys. More in the line to what we are doing in the password forms ([here](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/browser/UsernamePasswordForm.java#L79-L83)). The condition would be named something like `ConditionalPasskeyNotUsedAuthenticator` and it would just check that.
3. Note that using this condition idea we can even get rid of the direct code in the password forms to skip their execution (the method linked before). The new condition can be added to the flows and avoid the execution of the password forms when necessary. But maybe it's too complicated.
4. Do you have any other idea to skip the 2FA sub-flow? I had already thought about this new condition before (more or less the idea commented in 3). So I can be missing other approaches because of that.

The PR is idea 1, but I don't know if idea 2 is better. We have to decide what to implement and then set the solution to the default browser flow and document everything.


graziang:
@rmartinc thanks! It makes sense to me to introduce a condition for the credential type, but I’m not sure if option 1 or option 2 is better. I can’t think a use case right now where the condition would be used the other way around. We also need to considier how to integrate it into the default browser flow, or whether to simply add it and document how to use it. Did you already have an idea of how to insert  it in the browser flow?

rmartinc:
@graziang Yes, I wanted to integrate in the default browser flow by default. The idea would just add the new condition to the `Conditional 2FA` sub-flow together with the current `user configured` condition. The configuration is very simple, just:

```
Condition - current credential
  credentials: webauthn-passwordless
  included: off
```

This way the 2FA is only executed if the user has a configured credential and the current credential is not `webauthn-passwordless`. If the user has used a passkey to login then the 2FA sub-flow is always skipped.

And I'm not sure about option 1 or 2. The generic option 1 allows to setup more complicated flows that, for example, can also allow kerberos to skip the 2FA. Or include an extra step if the credential used is password (for example a custom one saying that it is better to use a passkeys). It is just for completeness, but I agree that it is more complicated and maybe the direct option 2 is more than enough for now.

rmartinc:
@graziang I have added a second commit adding the new condition by default in the browser and first broker login flows. This is the idea. If we finally decide to use option 2 is the same, but using no configuration for the condition as it would be direct for `webauthn-passwordless`. I'll bring this issue in tomorrow's meeting to see how to continue.

mposolda:
@rmartinc @graziang Thanks for the PR and discussion! Few questions:

- Does it makes sense to enable this behaviour by default? Is it common practice to expect that if people use passkeys credential (instead of password), then 2-factor authentication is automatically skipped? I am slightly more to have it rather disabled by default, as it is more safe (passkey will still require also 2-factor unless it is explicitly disabled in the flow), but not 100% sure what is a common practice here?

- Do we want to check really just the last used credential? Does it makes more sense to have condition like `Skip 2-factor if passkey credential was used during authentication` instead of `Skip 2-factor if passkey credential was used as LAST credential during authentication` ? If we agree on this, we can maybe change the note `LAST_AUTHN_CREDENTIAL` to be the list of authenticated credentials instead of just last credential?

rmartinc:
> Does it makes sense to enable this behavior by default?

Yes, this seems to be the default behavior. All services I have tested work this way, if you use a passkey there is no 2FA. That's why I'm adding the condition to the default browser flow.

> Do we want to check really just the last used credential?

I just did it this way previously, but yes, we can manage a list of credentials if you prefer, and I think that it makes more sense. I'll try to incorporate it.

So, in summary, we will continue with the generic condition but storing in the auth session the list of credentials used in the flow. The condition just checks against that list.

mposolda:
> > Does it makes sense to enable this behavior by default?
> 
> Yes, this seems to be the default behavior. All services I have tested work this way, if you use a passkey there is no 2FA. That's why I'm adding the condition to the default browser flow.
> 
> > Do we want to check really just the last used credential?
> 
> I just did it this way previously, but yes, we can manage a list of credentials if you prefer, and I think that it makes more sense. I'll try to incorporate it.
> 
> So, in summary, we will continue with the generic condition but storing in the auth session the list of credentials used in the flow. The condition just checks against that list.

Cool, Thanks

rmartinc:
I have added the idea about using a list instead of the last credential used. Besides a new test is added in the passkeys section to test the default browser flow (`PasskeysDefaultBrowserFlowTest` class). Added only the documentation for the new condition. If the CI passes OK I will move the PR out of draft. Squashed to only one commit now.

rmartinc:
The JS CI is failing always but also for other PRs, so I suppose the flakiness is even worse now.

mposolda (approved):
@rmartinc Nice, Thanks!

PR #41377, opened 2025-07-23
fix: using localtest.me instead of nip.io

### LINKED ISSUE #40645: Use localtest.me instead of nip.io
### Description

nip.io only resolves to ipv4 addresses. For ipv6 only testing we should use localtest.me instead as it has an AAAA record for ::1

shawkins:
closes: #40645


mabartos (approved):
@shawkins LGTM

ahus1 (approved):

Issue #41353, opened 2025-07-23
Provide HTTP access logs written to file with rotation


mabartos:
### Description

Sometimes it's very useful and potentially required to keep records of activity of the application in history for various reasons. 

We can provide the support for HTTP access logs that will be written to specific file and will be rotated f.e. every day.

mabartos:
Follow-up issue: 
- https://github.com/keycloak/keycloak/issues/45629

PR #41389, opened 2025-07-24
Provide simple HTTP access logs

### LINKED ISSUE #41352: Provide simple HTTP access logs
### Description

The first iteration of HTTP access log integration will be showing the HTTP access records in the console. 

We can provide more advanced mechanism for HTTP access logs in files with custom log rotation in a follow-up issues. 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mabartos:
- Closes #41352

@shawkins Could you please check it? I think we can directly support it and not having any preview mode for them as we just basically map KC options to Quarkus.

### DIFF
--- a/docs/guides/server/logging.adoc
+++ b/docs/guides/server/logging.adoc
@@ -258,7 +258,6 @@ These properties are available only when asynchronous logging is enabled for the
 {project_name} supports HTTP access logging to record details of incoming HTTP requests.
 While access logs are often used for debugging and traffic analysis, they are also important for security auditing and compliance monitoring, helping administrators track access patterns, identify suspicious activity, and maintain audit trails.
 
-Currently, HTTP access logs are only supported via console output.
 These logs are written at the `INFO` level, so make sure your logging configuration includes this level — either globally (e.g. `log-level=info`) or specifically for the access log category (e.g. `log-level=org.keycloak.http.access-log:info`).
 When HTTP access logs are enabled, they are shown by default, as `INFO` level is the default log level for {project_name}.
 
--- a/docs/guides/server/logging.adoc
+++ b/docs/guides/server/logging.adoc
@@ -94,7 +94,7 @@ Use the option `log-mdc-enabled` to enable it.
 
 Specify which keys to be added by setting the configuration option `log-mdc-keys`.
 
-==== Configuring levels as individual options
+=== Configuring levels as individual options
 When configuring category-specific log levels, you can also set the log levels as individual `log-level-<category>` options instead of using the `log-level` option for that.
 This is useful when you want to set the log levels for selected categories without overwriting the previously set `log-level` option.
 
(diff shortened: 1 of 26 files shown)


shawkins:
> @shawkins Could you please check it? I think we can directly support it and not having any preview mode for them as we just basically map KC options to Quarkus.

Does it need a separate category? Quarkus for example just considers these http configuration.

> NOTE: At this moment, only logging to console is supported for HTTP access logs.

Isn't that just up to the logging configuration? I would assume the messages just logged to the INFO level - or is quarkus specifically only sending them to the console?

Also this does bring up the point that INFO logging needs to be enabled on the access logging category for these to be visible in the console correct?

mabartos:
> Does it need a separate category? Quarkus for example just considers these http configuration.

@shawkins I think so. We have it directly below the HTTP options and it might be perceived as subcategory - the same as for `Database` and `Database - additional datasources`. We'll need to provide other properties for supporting access log written to files with the rotation support. See https://github.com/keycloak/keycloak/issues/41353. Having them combined with the HTTP category, it'd be a config nightmare to list these options and manage them. I think separate category makes more sense. 

> Isn't that just up to the logging configuration? I would assume the messages just logged to the INFO level - or is quarkus specifically only sending them to the console? Also this does bring up the point that INFO logging needs to be enabled on the access logging category for these to be visible in the console correct?

I'll update these notes. Yes, it uses INFO level for the hardcoded logger name `org.keycloak.http.access-log`. So it means when `log-level=warn` or `log-level=org.keycloak.http.access-log:warn` is used, these access logs are not shown.

mabartos:
@shawkins Added commit https://github.com/keycloak/keycloak/pull/41389/commits/6a67d560d83c6644f388141e72b0e5b8936fc6eb with some clarifications.

docs/guides/server/logging.adoc
```diff
@@ -253,6 +253,49 @@ You can change the queue length as follows:
 
 These properties are available only when asynchronous logging is enabled for these specific log handlers.
 
+=== HTTP Access Logging
+
+{project_name} supports HTTP access logging to record details of incoming HTTP requests.
+While access logs are often used for debugging and traffic analysis, they are also important for security auditing and compliance monitoring, helping administrators track access patterns, identify suspicious activity, and maintain audit trails.
+
+Currently, HTTP access logs are only supported via console output.
```
shawkins: Did you confirm if this is accurate?

mabartos: > Did you confirm if this is accurate?

Yes, the statement should be true :) 

"_Software developers, operations engineers, and security analysts use access logs to monitor how their application is performing, who is accessing it, and what’s happening behind the scenes. Access logs can help IT teams discover problems, detect threats, and identify capacity issues._" - https://www.crowdstrike.com/en-us/cybersecurity-101/observability/access-logs/

"_You can use these access logs to analyze traffic patterns and troubleshoot issues._" - https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html

"_Server-generated records of each HTTP request, used in analytics for traffic insights, troubleshooting, and compliance monitoring._" - https://plainsignal.com/glossary/http-access-logs

"_Access logs are crucial for monitoring and auditing user activity within systems and applications._ " - https://www.sentinelone.com/cybersecurity-101/cybersecurity/what-is-an-access-log/

> Currently, HTTP access logs are only supported via console output.

This is accurate for Keycloak for now. After completing https://github.com/keycloak/keycloak/issues/41353, we will support even access logs written to files.

@shawkins Or do you have any concerns?

shawkins: No I mean specifically "Currently, HTTP access logs are only supported via console output." - when I enable access logs and the file handler, I do see the access logs in the file as well.

mabartos: @shawkins Sorry, I wasn't clear.

This is a correct comment:
```diff
< After completing #41353, we will support even access logs written to files.
---
> After completing #41353, we will support even access logs written to separate files.
```

It is expected behavior that the access log records are part of the `keycloak.log` for now when file log handler is used.

The `quarkus.http.access-log.log-to-file` [property](https://quarkus.io/guides/http-reference#quarkus-vertx-http_quarkus-http-access-log_quarkus-http-access-log-log-to-file) is set to `false` by default, so no separate file is created. 

However, I plan to complete the https://github.com/keycloak/keycloak/issues/41353 until 26.4 as well, so there will be additional description of the functionality to have the access log in a separate file.

Is it ok for you to have it in this form and proceed with it in the follow-up issue?

shawkins: > Is it ok for you to have it in this form and proceed with it in the follow-up issue?

Could you just omit "Currently, HTTP access logs are only supported via console output"? Then further refine it when you work https://github.com/keycloak/keycloak/issues/41353

mabartos: Ok, agree

mabartos:
@shawkins Updated

shawkins (approved):
LGTM, nice work @mabartos

mposolda (approved):
Approved on behalf of @shawkins review

PR #41391, opened 2025-07-24
Setting 'write_data_on_find' to true to ensure merging of views

### LINKED ISSUE #41390: JDBC_PING2 doesn't merge split clusters after a while
### Before reporting an issue

### Area

infinispan

### Describe the bug

With #41290, it seems that it sometimes takes minutes to merge two JGroups coordinators.

### Version

23.x

### Regression

### Expected behavior

Like with other protocols, two coordinators should eventually merge.

### Actual behavior

It takes a while for the merge to happen.

### How to Reproduce?

Reported in #41290, and investigated in #41291 unit test.

### Anything else?

This might be similar to https://issues.redhat.com/browse/JGRP-2288 when the info writer stops writing its information to the table. 

For JDBC_PING2, setting `remove_all_data_on_view_change` to `true` should IMHO always be have `write_data_on_find` set to `true` as well.

ahus1:
Closes #41390


ahus1:
@pruivo @ryanemerson @belaban - can you please provide your feedback if this is a good idea for Keycloak's JDBC_PING2 usage? 
Given that it is a small change, I would like to backport this to KC 26.2 and KC 26.3. Thanks!

belaban:
Setting `write_data_on_find` to `true` by default is a second line of defense and cannot hurt. Although it might be redundant.

ahus1:
@pruivo @ryanemerson - would you approve this one based on @belaban's comment, and then I'll prepare a backport?

pruivo (approved):
Bela's blessing is good enough for me :+1:

ryanemerson (approved):

PR #41398, opened 2025-07-24
Wildcard mappers might ignore quoting for external .to() properties

mabartos:
- Closes #41397
- Blocks https://github.com/keycloak/keycloak/pull/41087

If a property mapper has a default value, we never try to get Quarkus properties from the `quarkus.properties` file. 
We should take into account the value of Quarkus property (defined as .to() in our mapper) that the user has overridden in the `quarkus.properties` file.

Right now, if user has `quarkus.properties` file with property like `quarkus.datasource."my-store".db-kind=mariadb`, and we ask for the value, we get the default value -> H2.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/Configuration.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/Configuration.java
@@ -54,7 +54,7 @@ public final class Configuration {
     public static boolean isUserModifiable(ConfigValue configValue) {
         // This could check as low as SysPropConfigSource DEFAULT_ORDINAL, which is 400
         // for now we won't validate these as it's not expected for the user to specify options via system properties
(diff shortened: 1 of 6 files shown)


shawkins:
Actually we may not want to do this. In looking at the original problem more there's not just an issue of picking up non-quoted, there's also that you can't mix quoted and non-quoted. So the user cannot use any of the new db options if they have quarkus.properties that are unquoted - unless we do a lot more work. I'm working on an update to the other pr.

mabartos:
@shawkins I've added a PR description with some details. Later, we need to know what DB-kind is used for the additional datasource as even specified in `quarkus.properties`. However, it brings some other problems, like with transformation of these values. 

I'm of for today, so I'll check it tomorrow, or on Monday :)) Thanks for the comment!

mabartos:
Handling the case for quoted vs. unquoted would be great, but it can bring some more problems and a bigger effort needed to support it. Even based on the comment from @shawkins, the best way will be to not address the case for `quarkus.properties` as those are unsuported anyway (kinda supported for datasources). 

Closing now.

PR #41406, opened 2025-07-24
Add details about client assertion to event

thomasdarimont:
Fixes #41405

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java
@@ -31,6 +31,7 @@ import org.keycloak.OAuth2Constants;
 import org.keycloak.authentication.AuthenticationFlowError;
 import org.keycloak.authentication.ClientAuthenticationFlowContext;
 import org.keycloak.common.util.Time;
+import org.keycloak.events.Details;
 import org.keycloak.http.HttpRequest;
 import org.keycloak.jose.jws.JWSInput;
 import org.keycloak.jose.jws.JWSInputException;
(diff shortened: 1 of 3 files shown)


services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java
```diff
@@ -116,6 +116,13 @@ public void readJws() throws JWSInputException {
 
         jws = new JWSInput(clientAssertion);
         token = jws.readJsonContent(JsonWebToken.class);
+
+        var event = context.getEvent();
+        if (token != null && event != null) {
```
ahus1: The null checks for token and event seems to be redundant. Did you try it without it?

services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java
```diff
@@ -116,6 +116,13 @@ public void readJws() throws JWSInputException {
...
+            event.detail("client_assertion_jti", token.getId());
+            event.detail("client_assertion_issuer", token.getIssuer());
+            event.detail("client_assertion_subject", token.getSubject());
```
ahus1: We have constants like `Details.TOKEN_ID` to place a token ID into an event. Can it be used here as well?

ahus1 (changes requested):
Thank you for this PR. See below for some comments. Please update one of the existing tests to show that the event details are stored in the event.

services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java
thomasdarimont: I thought of using Details.TOKEN_ID but since this usually indicates the "jti" for the used access token, I thought that it wouldn't be a good fit here, as we deal with a client assertion.

I added `CLIENT_ASSERTION_ID` and `CLIENT_ASSERTION_ISSUER` and `CLIENT_ASSERTION_SUB` constants to Details.
CLIENT_ASSERTION_ID -> helps to identify an individual invalid client assertion
CLIENT_ASSERTION_ISSUER / CLIENT_ASSERTION_SUB -> helps to identify common miss configurations

services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java
thomasdarimont: That was just defensive programming. The null checks can indeed be removed there.

ahus1 (approved):

PR #41416, opened 2025-07-25
WelcomeResource create an ordinary (not temporary) Admin account

### LINKED ISSUE #41388: Welcome page creates an temporary user
### Before reporting an issue

### Area

core

### Describe the bug

When I create a admin account through the welcome page it is marked as a temporary admin account. This makes no sense as I have personally created my own account with a password, and there is no reason for this to be temporary.

Stressing this issue does not apply to admin accounts provisioned through environment variables, config options, etc. only through the welcome page.

### Version

26.3.1

### Regression

### Expected behavior

When I create my own account through welcome page with my own password it should not be temporary

### Actual behavior

My own account is marked as temporary

### How to Reproduce?

Use welcome page to create a admin account and open the admin console to see the warnings.

### Anything else?

_No response_

slaskawi:
This Pull Request fixes a regression mentioned in #41388 by creating an ordinary/final Admin accounts with the Welcome Page

Fixes #41388


services/src/main/java/org/keycloak/services/resources/WelcomeResource.java
```diff
@@ -145,7 +148,6 @@ public Response createUser() {
             expireCsrfCookie();
 
             shouldBootstrap.set(false);
-            ServicesLogger.LOGGER.createdTemporaryAdminUser(username);
```
slaskawi, line 148: Removed as it's already logged in the `applianceBootstrap.createMasterRealmUser` call chain.

stianst:
Wondering if it would also make sense to add additional fields to the welcome page to create a more complete account? Username, first/last name, email, etc.?

Username could be pre-filled to admin, while other fields optional perhaps.

Probably better to have that as a follow-up though, so if others think it makes sense we can open a separate issue for it.

slaskawi:
@stianst Absolutely! I'm also fully onboard with doing it as a separate ticket (I'll probably start after my PTO - early August): https://github.com/keycloak/keycloak/issues/41425

shawkins:
Definitely like this direction more than the special purpose flags. By this same thinking should the bootstrap admin cli commands always be non-temporary, or should it be optional there? They optionally read from designated environment variables and do not rely upon the keycloak configuration for user / service account information.

For this pr we also need some kind of release or upgrade note, and it looks like there's a doc image initial-welcome-page.png that will need updated as well.

slaskawi:
>  By this same thinking should the bootstrap admin cli commands always be non-temporary, or should it be optional there? They optionally read from designated environment variables and do not rely upon the keycloak configuration for user / service account information.

@shawkins I believe this change should only relate to Welcome Resource. The Welcome Resource is used to create a first Admin user. This account is not temporary in any way - it is bound to a real user who's creating it.

> For this pr we also need some kind of release or upgrade note, and it looks like there's a doc image initial-welcome-page.png that will need updated as well.

@shawkins Added both. Thanks for spotting this!

shawkins:
> @shawkins I believe this change should only relate to Welcome Resource. The Welcome Resource is used to create a first Admin user. This account is not temporary in any way - it is bound to a real user who's creating it.

I'm not sure I follow the difference currently - that is as long as you don't use the default username in the bootstrap-admin command. Or is the notion of temporary also based upon whether there are users existing in the realm?

slaskawi:
> I'm not sure I follow the difference currently - that is as long as you don't use the default username in the bootstrap-admin command. Or is the notion of temporary also based upon whether there are users existing in the realm?

The primary reason is the source for the admin credentials. Passing them through an environment variable or CLI switch means that they cannot be trusted. In case of a Welcome Resource, they come directly from the admin user and are not written/used anywhere else. Therefore, they are not temporary in any way.

shawkins:
> The primary reason is the source for the admin credentials. Passing them through an environment variable or CLI switch means that they cannot be trusted. 

The bootstrap admin command will not allow the password via a switch, and if the username and password are not provided via one of the accepted methods it will prompt for the input.

vmuzikar:
From my perspective, anything that doesn't natively require user interaction (i.e. CLI args, env vars etc.) should be treated as insecure and therefore temporary. Welcome Screen is clearly a different case, the CLI prompt method (as mentioned by @shawkins) feels like something in between. Still it's only CLI – although interactive. I'd suggest to keep the CLI part as is and when we have the proper way for expiring users, we'd add a prompt for the expiration time allowing users to set the admin account as permanent.

WDYT @slaskawi @shawkins?

services/src/main/java/org/keycloak/services/ServicesLogger.java
```diff
@@ -469,4 +469,8 @@ public interface ServicesLogger extends BasicLogger {
     @Message(id=110, value="Environment variable '%s' is deprecated, use '%s' instead")
     void usingDeprecatedEnvironmentVariable(String deprecated, String supported);
 
+    @LogMessage(level = INFO)
+    @Message(id=111, value="Created admin user with username %s")
+    void createdAdminUser(String userName);
```
vmuzikar:
```suggestion
    @Message(id=111, value="Created initial admin user with username %s")
    void createdInitialAdminUser(String userName);
```

slaskawi:
> From my perspective, anything that doesn't natively require user interaction (i.e. CLI args, env vars etc.) should be treated as insecure and therefore temporary. Welcome Screen is clearly a different case, the CLI prompt method (as mentioned by @shawkins) feels like something in between. Still it's only CLI – although interactive. I'd suggest to keep the CLI part as is and when we have the proper way for expiring users, we'd add a prompt for the expiration time allowing users to set the admin account as permanent.

@shawkins @vmuzikar Yes, I absolutely agree with this reasoning!

As for the CLI (with the input prompt), I do agree it's in the middle and we can pursue it both ways. Perhaps it's worth cutting a ticket for it and checking if there are any demands for implementing it (in other words, wait for upvotes). I'm happy to pursue it if there's an interest but definitely as a separate effort (ticket and PR).

services/src/main/java/org/keycloak/services/ServicesLogger.java
slaskawi: Incorporated into my rebased commit.

slaskawi:
@vmuzikar @shawkins Rebased and addressed comments.

shawkins:
Thank you @slaskawi this looks good to me.

shawkins (dismissed):

docs/documentation/release_notes/topics/26_4_0.adoc
```diff
@@ -50,3 +50,7 @@ It is now possible to change which identity providers are shown in the account c
 the `Show in Account console` setting. You can choose to show only those linked with a user or not show them at all.
 
 For more information, please see link:{adminguide_link}#_general-idp-config[General configuration] section in the {adminguide_name}.
+
+= Welcome Page changes
```
vmuzikar: I just realized, this should probably be a migration note, not a release note. It’s a functional behavior change, not a new feature.

slaskawi: I'm not sure if it makes sense. Welcome Resource/Page is used for new installations only. This means there's no migration involved whatsoever. Personally, I'd just leave it here.

vmuzikar: A breaking change requiring a migration note might be also in the installation process.

slaskawi: Ok, I moved it to the migration notes instead. Resolving.

vmuzikar (approved):
@slaskawi LGTM, thank you.

@shawkins Do you want to take another look?

shawkins (approved):
Still LGTM, thanks @slaskawi

Issue #40509, opened 2025-06-16
OpenID Federation implementation


cgeorgilakis:
### Description

Implement [OpenID Federation](https://openid.net/specs/openid-federation-1_0.html) in Keycloak.
Documentation for implementation in Authlete is [here](https://www.authlete.com/developers/oidcfed/).
[Keycloak plugin](https://github.com/thomasdarimont/keycloak-oidc-federation/tree/update/keycloak-26.x) for explicit registration based on older draft.

We propose realm admin being able to enable OpenID Federation (default false). When enabling, group admin could configure OpenID Federation (including required fields such as authority_hints. This will expose OpenID Federation metadata ( /.well-known/openid-federation ) and related functionality, fe explicit and automatic registration.

Our initial goal is supporting Keycloak being able to be OP and RP with both explicit and automatic registration.
However, epic is open for other OpenID Federation entities/ functionalities.

### Discussion

https://github.com/keycloak/keycloak/discussions/31027

### Motivation

OpenID Federation enables participation in an identity federation of entities using OpenID Connect and OAuth 2.0 and offers a robust framework for establishing dynamic trust between OpenID Providers (OPs) and Relying Parties (RPs), significantly simplifying the management of large-scale identity federations. By leveraging this specification, Keycloak will enable the dynamic establishment of trust between OPs and RPs, facilitate secure interactions authenticated via Trust Anchors, and crucially, eliminate the need for cumbersome manual or bilateral trust agreements

pritish-nitb:
@cgeorgilakis Thanks for raising this and outlining the direction — it’s great to see Keycloak moving toward native support for OpenID Federation.

We (Banfico) have worked extensively on implementing centralized trust frameworks, particularly in regulated environments such as Open Banking and COP/VOP. Our deployments have involved OpenID Connect, OAuth 2.0, FAPI profiles, DCR and eIDAS-compliant certificate handling — all within federated ecosystems where dynamic trust establishment is critical.

This initiative aligns well with what we’ve seen in production: supporting OP and RP roles with explicit registration, trust chain validation, and governance via Trust Anchors. Eliminating bilateral trust agreements at scale is a major step forward, and we’re excited to see Keycloak moving in this direction.

I am going through the child tickets as well. Thanks a lot for putting it all there. :)

bucchi:
I have just started  working on [OpenID Federation OP with Automatic Registration #40920](https://github.com/keycloak/keycloak/issues/40920)

[...]

PR #41419, opened 2025-07-25
OpenId Federation OP back end

### LINKED ISSUE #40511: OpenID Federation OP with explicit registration
### Description

initial implementation of OpenID Federation with Keycloak being able to support being OP with explicit registration. 

We propose realm admin being able to enable OpenID Federation (default false). When enabling, group admin could configure OpenID Federation (including required fields such as authority_hints. This will expose OpenID Federation metadata ( /.well-known/openid-federation ) and client (RP) explicit registration.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

cgeorgilakis:
Closes #40511

This is the initial PR for supporting [OpenID Federation](https://openid.net/specs/openid-federation-1_0.html) with Keycloak being an OP and explicit registration, accepting only Entity Statement. Ui will be a separate PR, documentation will be enhanced based on it. Some code parts are taken from the [OpenID Federation extensions](https://github.com/thomasdarimont/keycloak-oidc-federation), taking into account that we are in core Keycloak, needed bug fixes and code improvements , compliant with latest OpenID Federation specification version.

Some key points for the PR:

- OpenID Federation is disabled by default and can be enabled with required configuration (more in ui part). Inspiration : https://www.authlete.com/developers/oidcfed/#configuring-federation
- Keycloak as a proxy can participate in more than one Federation(Trust Anchor) with different configuration options especially for RP part, trust marks etc (future features). That’s why it is separated table.
- Tests for explicit client registration has been done with a real test OpenID Federation(Trust Anchor, RP) via the help of [golang lighthouse OpenID Federation implementation](https://github.com/go-oidfed/lighthouse)
- I have added tests for Federation Entity Configuration. For adding testing for explicit registration we need to mock REST API for Trust Anchor Entity Configuration, fetch and list endpoints. Take into account that they are JWT signed statement. Moreover, we need to create a mock jwt valid during test execution for client explicit registrations. How could we develop it? I have only a mock Java code for create a RP entity Statement based on mock keys. Should it be a separate issue?

For more opened issues related to explicit client registration in an OP, you could see the [epic issues](https://github.com/keycloak/keycloak/issues/40509).


cgeorgilakis:
Information for testing this PR.


1. You will need to add the following realm configurations:

   - "openidFederationOrganizationName "	-	"Keycloak foundation" (optional)

   - "openidFederationLifespan"	- "100000" (optinal default to 86400 – one day)

   - "openidFederationEnabled"	-	"true" 

   - "openidFederationAuthorityHints"	-	"https://trust-anchor.sandbox.eosc.grnet.gr"


2. Moreover, you will need to create a trust anchor(federation) configuration using : 

  ```
curl --request POST \
  --url <root>/admin/realms/{realm-name}/openid-federations \
  --header 'Authorization: Bearer {token} \
  --header 'Content-Type: application/json' \
  --data '{
		"trustAnchor": "https://trust-anchor.sandbox.eosc.grnet.gr",
		"clientRegistrationTypesSupported": [
			"EXPLICIT"
		],
		"entityTypes": [
			"OPENID_PROVIDER"
		]
	}'
   ```
3. After you will be able to observe the signed JWT Federation Entity Configuration from <root>/realms/{realm-name}/.well-known/openid-federation as described in the [documentaion](https://github.com/keycloak/keycloak/blob/921a3dde7eedf946093fc64fdedfb0dec0a91005/docs/documentation/server_admin/topics/sso-protocols/openid-federation.adoc)

4. For testing explicit registration, you will need to start server commenting the aud check in the class [OpenIdFederationClientRegistrationService](https://github.com/keycloak/keycloak/blob/921a3dde7eedf946093fc64fdedfb0dec0a91005/services/src/main/java/org/keycloak/services/clientregistration/openid_federation/OpenIdFederationClientRegistrationService.java) lines 146-148. 

 After you could execute the request : 
`curl --request POST \
  --url <root>/realms/{realm-name}/openid-federation/clients-registrations \
  --header 'Accept: application/entity-statement+jwt' \
  --header 'Content-Type: application/entity-statement+jwt' \
  --data {jwt}`

For jwt you could use the signed JWT Federation Entity Configuration of our [OpenID Federation RP signed JWT Entity Configuration](https://aai-dev.egi.eu/auth/realms/egi/.well-known/openid-federation), enrolled in the mentioned Trust Anchor (https://trust-anchor.sandbox.eosc.grnet.gr). This Trust Anchor is a docker deployment of an Trust Anchor implementation (see [lighthouse deployment instructions](https://go-oidfed.github.io/lighthouse/deployment/caddy/).

After successfully request, an OIDC Client will be created, with only the difference that it will contain the attribute `OIDCConfigAttributes.EXPIRATION_TIME`. Response will contains the [Explicit Client Registration Response](https://openid.net/specs/openid-federation-1_0.html#name-successful-explicit-client-).

5. On the client attribute `OIDCConfigAttributes.EXPIRATION_TIME` time, client will be disabled. Before or after the client expiration, you could execute the same request with an updated JWT Entity Configuration in order to enable client and extend expiration.

6. For a more real case scenario:

    - You could provide us the JWT Federation Entity Configuration from <root>/realms/{realm-name}/.well-known/openid-federation. We could enroll this Keycloak as OP in our OpenID Federation (https://trust-anchor.sandbox.eosc.grnet.gr).
    - After I could execute from our egi-dev RP instance, a full explicit registration scenario. This will create your Keycloak instance as OpenID Federation in our egi-dev RP instance and a client of our egi-dev RP instance in your Keycloak. Maybe we will only need to add appropriate mappers and scopes for realising user attributes.
    - After you will be able to login in our egi-dev RP instance with your users (fe account console).

cgeorgilakis:
I want here to mention that our proposed implementation is based on EntityStatement which is the default body parameter described in the specification (see chapters 12.2 and 10, appendix A.3.2).
Body with Trust Chains (alternative way based on protocol) as body return Response.Status.NOT_IMPLEMENTED. I personally do not know which should be the appropriate checks in that case. Unfortunatelly, OpenID Federation conformance tests are only based on this body type.

server-spi/src/main/java/org/keycloak/models/OpenIdFederationGeneralConfig.java
```diff
@@ -0,0 +1,104 @@
...
+    private List<String> contacts;
+    private String logoUri;
+    private String policyUri;
+    private String organizationUri;
+    private List<String> authorityHints;
+    private Integer lifespan;
+    // default 1 day - duration
+    private String federationResolveEndpoint;
+    private String federationHistoricalKeysEndpoint;
+    private List<OpenIdFederationConfig> openIdFederationList = new ArrayList<>();
```
pruivo: It can be empty list as far as I can tell. 
```suggestion
    private List<OpenIdFederationConfig> openIdFederationList = List.of();
```

server-spi/src/main/java/org/keycloak/models/OpenIdFederationGeneralConfig.java
```diff
@@ -0,0 +1,104 @@
...
+    public void setFederationHistoricalKeysEndpoint(String federationHistoricalKeysEndpoint) {
+        this.federationHistoricalKeysEndpoint = federationHistoricalKeysEndpoint;
+    }
+
+    public List<OpenIdFederationConfig> getOpenIdFederationList() {
+        return openIdFederationList;
+    }
+
+    public void setOpenIdFederationList(List<OpenIdFederationConfig> openIdFederationList) {
+        this.openIdFederationList = openIdFederationList;
```
pruivo: You can avoid null checks the return value of `getOpenIdFederationList()` in your code if you don't allow setting a null list. Suggestion below.
```suggestion
        this.openIdFederationList = openIdFederationList == null ? List.of() : List.copyOf(openIdFederationList);
```

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
```diff
@@ -791,6 +794,42 @@ public void setWebAuthnPolicyPasswordless(WebAuthnPolicy policy) {
...
+    @Override
+    public void setOpenIdFederationGeneralConfig(OpenIdFederationGeneralConfig generalConfig) {
+        getDelegateForUpdate();
+        updated.setOpenIdFederationGeneralConfig(generalConfig);
+    }
+
+    @Override
+    public List<OpenIdFederationConfig> getOpenIdFederations() {
+        if (isUpdated()) return updated.getOpenIdFederations();
+        return isOpenIdFederationEnabled() ?  cached.getOpenIdFederationConfig().getOpenIdFederationList() : new ArrayList<>();
```
pruivo:
```suggestion
        return isOpenIdFederationEnabled() ?  cached.getOpenIdFederationConfig().getOpenIdFederationList() : List.of();
```

server-spi/src/main/java/org/keycloak/models/RealmModel.java
```diff
@@ -325,6 +327,30 @@ default <V extends Enum<V>> V getAttribute(String name, Class<V> enumClass, V de
...
+
+    default boolean isOpenIdFederationEnabled() {
+        return getOpenIdFederationGeneralConfig() != null;
+    };
+
+    default void setOpenIdFederationGeneralConfig(OpenIdFederationGeneralConfig generalConfig) {
+    };
+
+    default List<OpenIdFederationConfig> getOpenIdFederations() {
+        return new ArrayList<>();
```
pruivo:
```suggestion
        return List.of();
```

services/src/main/java/org/keycloak/protocol/oidc/federation/OpenIdFederationWellKnownProvider.java
```diff
@@ -0,0 +1,101 @@
...
+
+        if (openIdFederationConfig ==  null || openIdFederationConfig.getOpenIdFederationList() == null || openIdFederationConfig.getOpenIdFederationList().isEmpty())
+            throw new NotFoundException();
+
+        UriInfo frontendUriInfo = session.getContext().getUri(UrlType.FRONTEND);
+        UriInfo backendUriInfo = session.getContext().getUri(UrlType.BACKEND);
+        Metadata metadata = new Metadata();
+        CommonMetadata common = OpenIdFederationUtils.commonMetadata(openIdFederationConfig);
+
+        if (openIdFederationConfig.getOpenIdFederationList().stream().flatMap(x -> x.getEntityTypes().stream()).collect(Collectors.toSet()).contains(EntityTypeEnum.OPENID_PROVIDER)) {
```
pruivo: I'm not sure how large these lists can be, but if you want to short-circuit the evaluation:

```java
        openIdFederationConfig.getOpenIdFederationList().stream()
                .map(OpenIdFederationConfig::getEntityTypes)
                .flatMap(Collection::stream)
                .anyMatch(EntityTypeEnum.OPENID_PROVIDER::equals)
```

services/src/main/java/org/keycloak/services/clientregistration/openid_federation/OpenIdFederationClientRegistrationService.java
```diff
@@ -0,0 +1,150 @@
...
+        super(session);
+        this.trustChainProcessor = session.getProvider(TrustChainProcessor.class, OpenIdFederationTrustChainProcessorFactory.PROVIDER_ID);
+    }
+
+    @POST
+    @Consumes({"application/entity-statement+jwt", "application/trust-chain+json"})
+    public Response explicitClientRegistration(String body, @Context HttpHeaders headers) {
+        RealmModel realm = session.getContext().getRealm();
+        OpenIdFederationGeneralConfig config = realm.getOpenIdFederationGeneralConfig();
+        if (!realm.isOpenIdFederationEnabled() || config.getOpenIdFederationList() == null || !config.getOpenIdFederationList().stream().flatMap(x -> x.getClientRegistrationTypesSupported().stream()).collect(Collectors.toSet()).contains(ClientRegistrationTypeEnum.EXPLICIT) || config.getAuthorityHints().isEmpty()) {
```
pruivo: same comment here.

model/storage-private/src/main/java/org/keycloak/services/scheduled/TaskCancellationEvent.java
```diff
@@ -0,0 +1,17 @@
+package org.keycloak.services.scheduled;
+
+import org.keycloak.cluster.ClusterEvent;
+
+public class TaskCancellationEvent implements ClusterEvent {
```
pruivo, line 9: Is the testsuite testing this class in a clustered environment? It looks like it needs to be marshaled via Infinispan ProtoStream.

pruivo (changes requested):
I didn't review the business logic as it is not my area of expertise. Added some suggestions here and there.
Related to Infinispan, I suspect the `TaskCancellationEvent` is broken when running in a clustered environment.

model/storage-private/src/main/java/org/keycloak/services/scheduled/TaskCancellationEvent.java
cgeorgilakis: No. I can not understand what it is needed.

pruivo: No worries, I'll guide you through. For context, Infinispan uses Protostream to send data between the Keycloak instances. You need to annotate your class to tell which fields to send. 

In this case, it is simple
```java
@ProtoTypeId(65616)
public class TaskCancellationEvent implements ClusterEvent {

    public static final String CANCEL_TASK = "cancelTask";
    private final String taskName;

    @ProtoFactory
    public TaskCancellationEvent(String taskName) {
        this.taskName = taskName;
    }

    @ProtoField(1)
    public String getTaskName() {
        return taskName;
    }
}
```

The type ID must be unique, so please add a comment and constant after:

https://github.com/keycloak/keycloak/blob/6aa3f9d5a75846f81a09582513154bf8136d3446/model/infinispan/src/main/java/org/keycloak/marshalling/Marshalling.java#L183

similar to 

https://github.com/keycloak/keycloak/blob/6aa3f9d5a75846f81a09582513154bf8136d3446/model/infinispan/src/main/java/org/keycloak/marshalling/Marshalling.java#L69-L78

The last step is to tell that the class exists. Add the class after:
https://github.com/keycloak/keycloak/blob/c97a9563ef8dec92ed5f0bb2be0c949bd4555289/model/infinispan/src/main/java/org/keycloak/marshalling/KeycloakModelSchema.java#L226

Feel free to ping me if you encounter any trouble or have any doubts.

cgeorgilakis: I have pushed the changes. Could you review them?

cgeorgilakis:
I have made the following changes:
- Remove Policies. Policies are required and I can provide a PR for the [related issue](https://github.com/keycloak/keycloak/issues/41033),  after merging this PR.
- If Trunst Anchor fetch endpoint contains RPMetadata (optional), RPMetadata for registration should be changes based on fetch endpoint. 
- Propagate task cancellation for client disable in other nodes
- Implement PR comments by @pruivo

model/storage-private/src/main/java/org/keycloak/services/scheduled/TaskCancellationEvent.java
pruivo: Looks good to me. Thank you.

services/src/main/java/org/keycloak/services/clientregistration/openid_federation/OpenIdFederationClientRegistrationService.java
```diff
@@ -0,0 +1,151 @@
...
+            throw new ErrorResponseException(Errors.INVALID_REQUEST, "Explicit OpenID Federation Client Registration is not supported in this realm", Response.Status.BAD_REQUEST);
+        }
+        checkSsl();
+
+        if ("application/entity-statement+jwt".equals(headers.getMediaType().toString())) {
+            EntityStatement statement = null;
+            try {
+                statement = trustChainProcessor.parseAndValidateSelfSigned(body);
+            } catch (InvalidTrustChainException ex) {
+                ex.printStackTrace();
```
pruivo: Don't print the stack trace to `System.out`. Log an error message.

services/src/main/java/org/keycloak/services/trustchain/OpenIdFederationTrustChainProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+                        chainsList.add(trustAnchor);
+                    } else {
+                        List<TrustChainResolution> subList = subTrustChains(initialEntity, subNodeSelfES, trustAnchorIds, visitedNodes);
+                        for (TrustChainResolution tcr : subList) {
+                            tcr.getParsedChain().add(0, subNodeSelfES);
+                            chainsList.add(tcr);
+                        }
+                    }
+                } catch (Exception ex) {
+                    ex.printStackTrace();
```
pruivo: same here

services/src/main/java/org/keycloak/services/trustchain/OpenIdFederationTrustChainProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    @Override
+    public void validateToken(String token, JSONWebKeySet jwks){
+        try{
+            ConfigurableJWTProcessor<SecurityContext> jwtProcessor = produceJwtProcessor(jwks);
+            jwtProcessor.process(token, null);
+
+        } catch(IOException | ParseException | BadJOSEException | JOSEException ex) {
+            ex.printStackTrace();
```
pruivo: and here :)

services/src/main/java/org/keycloak/services/resources/admin/OpenIdFederationsResource.java
```diff
@@ -0,0 +1,151 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class OpenIdFederationsResource {
+
+    private final RealmModel realm;
+    private final KeycloakSession session;
+    private AdminPermissionEvaluator auth;
+    private AdminEventBuilder adminEvent;
```
pruivo: Those can be final too.

services/src/main/java/org/keycloak/services/trustchain/OpenIdFederationTrustChainProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+            throw new ErrorResponseException(Errors.INVALID_TRUST_CHAIN, "Trust chain is not valid", Response.Status.BAD_REQUEST);
+        }
+    }
+
+    private ConfigurableJWTProcessor<SecurityContext> produceJwtProcessor(JSONWebKeySet jwks) throws IOException, ParseException {
+        String jsonKey = om.writeValueAsString(jwks);
```
pruivo: I'm unsure but I think you can remove `om` static field and use
```suggestion
        String jsonKey = JsonSerialization.mapper.writeValueAsString(jwks);
```

services/src/main/java/org/keycloak/utils/OpenIdFederationUtils.java
```diff
@@ -0,0 +1,45 @@
...
+        return SimpleHttp.doGet((fedApiUrl + "?sub=" + urlEncode(subject)),session).asString();
+    }
+
+    public static boolean containedInListEndpoint(String listApiUrl, String entityType, String issuer, KeycloakSession session) throws IOException {
+        return listApiUrl != null && SimpleHttp.doGet(listApiUrl + "?entity_type=" + entityType, session).asJson(new TypeReference<List<String>>(){}).contains(issuer);
+    }
+
+    private static String urlEncode(String url) throws UnsupportedEncodingException {
+        return URLEncoder.encode(url, StandardCharsets.UTF_8.toString());
+    }
```
pruivo:
```suggestion
    private static String urlEncode(String url) {
        return URLEncoder.encode(url, StandardCharsets.UTF_8);
    }
```

services/src/main/java/org/keycloak/services/clientregistration/openid_federation/OpenIdFederationClientRegistrationService.java
```diff
@@ -0,0 +1,151 @@
...
+import org.keycloak.services.clientregistration.oidc.DescriptionConverter;
+import org.keycloak.services.trustchain.OpenIdFederationTrustChainProcessorFactory;
+import org.keycloak.services.trustchain.TrustChainProcessor;
+import org.keycloak.urls.UrlType;
+import org.keycloak.util.TokenUtil;
+
+import java.net.URI;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.util.List;
```
pruivo: unused import

server-spi/src/main/java/org/keycloak/models/OpenIdFederationGeneralConfig.java
```diff
@@ -0,0 +1,104 @@
+package org.keycloak.models;
+
+import java.io.Serializable;
+import java.util.ArrayList;
```
pruivo: unused import

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
```diff
@@ -54,6 +56,7 @@
 import org.keycloak.storage.UserStorageUtil;
 import org.keycloak.storage.client.ClientStorageProvider;
 
+import java.util.ArrayList;
```
pruivo: unused import

model/jpa/src/main/java/org/keycloak/models/jpa/entities/RealmEntity.java
```diff
@@ -33,6 +33,8 @@
 import jakarta.persistence.NamedQuery;
 import jakarta.persistence.OneToMany;
 import jakarta.persistence.Table;
+
+import java.util.ArrayList;
```
pruivo, line 37: unused import

server-spi/src/main/java/org/keycloak/models/RealmModel.java
```diff
@@ -17,13 +17,15 @@
 
 package org.keycloak.models;
 
+import java.util.ArrayList;
```
pruivo: unused import

services/src/main/java/org/keycloak/services/trustchain/OpenIdFederationTrustChainProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+                    Object alg = key.getAlgorithm();
+                    if (alg instanceof JWSAlgorithm jwsAlgorithm) {
+                        return jwsAlgorithm;
+                    } else if (alg instanceof Algorithm algorithm) {
+                        try {
+                            return JWSAlgorithm.parse((algorithm).getName());
+                        } catch (IllegalArgumentException e) {
+                            return null;
+                        }
+                    } else if (alg instanceof String string) {
```
pruivo: Intellij reports `alg instanceof String string` is always false. If so, it can be removed.

pruivo (commented):
It looks good to me from the Infinispan perspective. I notice some warning reports from IntelliJ during code analysis.

services/src/main/java/org/keycloak/services/trustchain/OpenIdFederationTrustChainProcessor.java
cgeorgilakis: I could use JsonSerialization similar methods. I have removed om ObjectMapper.

cgeorgilakis:
> It looks good to me from the Infinispan perspective. I notice some warning reports from IntelliJ during code analysis.

I have made the requested changes. See my extra comment.

stianst:
I think it would make a lot of sense to build on the concept of identity providers to provide this trust relationship rather than introducing a whole new thing. See https://github.com/keycloak/keycloak/discussions/42197

Also, look at https://github.com/keycloak/keycloak/pull/42229 (client assertions via 3rd party OIDC) or https://github.com/keycloak/keycloak/pull/42314 (client assertions via SPIFFE).

model/jpa/src/main/java/org/keycloak/models/jpa/entities/RealmEntity.java
```diff
@@ -599,6 +602,17 @@ public void setUserFederationMappers(Collection<UserFederationMapperEntity> user
...
     }
 
+    public List<OpenIdFederationEntity> getOpenIdFederationList() {
+        return openIdFederationList;
+    }
+
+    public void setOpenIdFederationList(List<OpenIdFederationEntity> openIdFederationList) {
+        this.openIdFederationList.clear();
+        if (openIdFederationList != null && !openIdFederationList.isEmpty()) {
+            this.openIdFederationList.addAll(openIdFederationList);
```
bucchi, line 614: With my test code for Automatic Client Registration, an exception below is thrown here.
--------------------------------------------------------------------------------------
java.lang.UnsupportedOperationException
	at java.base/java.util.ImmutableCollections.uoe(ImmutableCollections.java:142)
	at java.base/java.util.ImmutableCollections$AbstractImmutableCollection.addAll(ImmutableCollections.java:148)
	at org.hibernate.collection.spi.PersistentBag.addAll(PersistentBag.java:435)
	at org.keycloak.models.jpa.entities.RealmEntity.setOpenIdFederationList(RealmEntity.java:619)
	at org.keycloak.models.jpa.RealmAdapter.setOpenIdFederationGeneralConfig(RealmAdapter.java:1178)


--------------------------------------------------------------------------------------

One of the way for you to fix that is the modification of line 153
`List<OpenIdFederationEntity> openIdFederationList = List.of(); `
to 
`List<OpenIdFederationEntity> openIdFederationList = new ArrayList<>();`

cgeorgilakis: You are right. I have fixed it.

PR #41432, opened 2025-07-25
Fixed how multi-value attributes are modified in jpa model adapters

### LINKED ISSUE #38526: Duplicate user attribute values cannot be removed
### Before reporting an issue

### Area

core

### Describe the bug

If a user account has a custom attribute with the same value multiple times, it is not possible to remove some of the duplicate values. However, it is possible to remove **all** the duplicate values together.
In fact it's also not possible to add the same value once more to same attribute.

### Version

26.1.4

### Regression

### Expected behavior

Via admin console duplicate values of a user attribute can be removed persistently.

### Actual behavior

Removing duplicate values of a user attribute via admin console will actually not be removed from storage when saving.

### How to Reproduce?

1. In admin console create a new custom user attribute with 3 times the value "A"
2. Save -> the attribute will be shown 3 times with value "A"
3. Remove one of the values by clicking on the minus button -> the entry disappears from the UI
4. Save
5. The attribute will be shown shown again 3 times with value "A"

### Anything else?

The Problem is in `org.keycloak.models.jpa.UserAdapter#setAttribute()`. There the new values and old values are compared as ``Set`` and not as ``List``.

```
...
        Set<String> oldEntries = getAttributeStream(name).collect(Collectors.toSet());
        Set<String> newEntries;
        if (values == null) {
            newEntries = new HashSet<>();
        } else {
            newEntries = new HashSet<>(values);
        }
        if (CollectionUtil.collectionEquals(oldEntries, newEntries)) {
            return;
        }
...
```
The same applies to 'infinispan.UserAdapter.
`RoleAdapter` and `GroupAdapter` don't have the same problem. Actually, `GroupAdapter` seems to be the best implementation which should probably be taken over to the other adapters.

mduchrow:
Aligned handling of setAttribute() in UserAdapter, RoleAdapter and GroupAdapter.
Duplicate values in the same attribute are now handled correctly. Duplicate values can be added and removed.

Closes #38526

### DIFF
--- a/model/jpa/pom.xml
+++ b/model/jpa/pom.xml
@@ -123,6 +123,32 @@
             <version>${jdbc.mvn.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- libraries needed for unit tests using an in-memory database -->
+		    <dependency>
+		        <groupId>com.h2database</groupId>
+		        <artifactId>h2</artifactId>
+		        <scope>test</scope>
+		    </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-saml-core</artifactId>
+		        <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-crypto-default</artifactId>
+		        <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-micrometer</artifactId>
+            <scope>test</scope>
+        </dependency>
+			  <dependency>
+					  <groupId>io.smallrye</groupId> 
+					  <artifactId>jandex</artifactId> 
+            <scope>test</scope>
+			  </dependency>
     </dependencies>
 
     <build>
--- a/model/jpa/pom.xml
+++ b/model/jpa/pom.xml
@@ -123,32 +123,6 @@
             <version>${jdbc.mvn.version}</version>
             <scope>test</scope>
         </dependency>
-        <!-- libraries needed for unit tests using an in-memory database -->
-		    <dependency>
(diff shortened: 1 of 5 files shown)


sguilhen (changes requested):
I like the proposed changes to how the attributes are handled, but I think we could simplify the tests if we add them directly to `tests/base/` along with other group/user tests there.

sguilhen:
@ahus1 @pedroigor adding you as reviewers as well - when you have some time please take a look.

dasniko:
@mduchrow Could you please:
* set back this PR to draft mode, so that it doesn't annoy the Keycloak team as long as we are working on it
* remove your custom test classes again, as we discussed to run the tests with the new test framework
* rebase(!) your actual changes on `UserAdapter`, `RoleAdapter` and `GroupAdapter` onto current `main` branch

I've already prepared a test method to be run in the current `org.keycloak.tests.model.UserModelTest`. I can add it once the rebase is done (and if you give editing permissions on your PR branch).

dasniko:
Thanks for cleaning up! 👍

I tried to test it locally, but it's still not possible to remove single attribute values if you have the same value multiple times. What I tried:
* add three times an attribute `foo` with value `bar` to a user on the user attributes page in admin ui
* save it, works, after refresh it's available three times
* remove one `foo`->`bar` line from attributes list
* save the user/attributes
* still the 3 `bar` values are available

dasniko:
For your reference, I added a test to a fork of your PR: https://github.com/dasniko/keycloak/commit/826848bf79e6c46dd979ed9af9267bfc45b5ed6b

Unfortunately, I don't get it to provide my changes as a PR to this PR... 🤪

ryanemerson:
Thank you for the PR @mduchrow, but I am now closing this as the original issue has been resolved by https://github.com/keycloak/keycloak/pull/48384.

PR #41438, opened 2025-07-25
Fix: prevent federated users from being listed on initial Users page load

### LINKED ISSUE #41044: Federated users incorrectly listed on first load due to uninitialized userProfileProvidersEnabled
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Currently, when an LDAP provider is added in the User Federation section, the initial API call on the Users page fetches federated users even if user federation is enabled, due to userProfileProvidersEnabled being undefined on first load. This causes the user to believe that the issue of listing federated users (as described in #21465) is resolved, because federated users are shown on the first page without needing to search.

However, once the user navigates (e.g., clicks 'next page') or clicks on refresh button, and the correct userProfileProvidersEnabled value is available, the expected info message is displayed and federated users are not fetched unless searched for, re-exposing the issue described in #21465.

### Version

26.3.1

### Regression

### Expected behavior

- When an LDAP (or other user federation) provider is enabled, the UI should not attempt to fetch or list federated users automatically unless the user explicitly performs a search. (Unless #21465 is explicitly marked resolved)
- On initial page load, while uiRealmInfo.userProfileProvidersEnabled is still undefined (i.e., before the federation info is loaded), the page should display a loading spinner (<KeycloakSpinner />) and should NOT fetch or display any users.
- Only after uiRealmInfo.userProfileProvidersEnabled is defined:
        If user federation is enabled, the page should display a message indicating that searching is required to list federated users, and not make a general API call to list users.
        If user federation is not enabled, the page should proceed to list all users as normal.
- At no point should federated users be fetched or displayed by default just because the federation status is not yet loaded.
- The behavior should be consistent across all pagination or navigation actions—there should not be a difference between the first load and subsequent pages.

### Actual behavior

In the UserDataTable component ([PR #31771](https://github.com/keycloak/keycloak/pull/31771#discussion_r2190925587)), there is an issue with how the loading state is managed when determining user federation status.

When `uiRealmInfo.userProfileProvidersEnabled` is undefined, the component proceeds to fetch users, even if user federation is enabled. This happens because `listUsers` evaluates to true by default, due to `uiRealmInfo` being initialized as an empty object. As a result, federated users are fetched and displayed on the initial page load, which can give the impression that all users (including federated ones) are being listed automatically—potentially misleading users into thinking that [issue #21465](https://github.com/keycloak/keycloak/issues/21465) has been resolved.

However, once the user interacts with the table (for example, by clicking "next page"), the value of `userProfileProvidersEnabled` becomes available. At this point, the component displays the correct "This realm may have a federated provider. Viewing all users may cause the system to slow down, but it can be done by searching for "*". Please search for a user above." info message, and the API call to fetch users is no longer made unless a search is performed.

This inconsistency between the initial load and subsequent interactions can lead to confusion and an inconsistent user experience.

It would be more robust to update the loading check to:

```
if (uiRealmInfo.userProfileProvidersEnabled === undefined || !realm) {
  return <KeycloakSpinner />;
}
```

This ensures that the component waits for the federation status before attempting to fetch and display users, providing a consistent and predictable user experience.

### How to Reproduce?

Navigate to User Federation page and add LDAP provider
Navigate to Users page, Notice API call to fetch users (includes federated users) is made
Click on refresh button, API call is no longer made (So users aren't fetched)

### Anything else?

_No response_

fvales:
Fixes #41044

### DIFF
--- a/js/apps/admin-ui/src/components/users/UserDataTable.tsx
+++ b/js/apps/admin-ui/src/components/users/UserDataTable.tsx
@@ -230,7 +230,7 @@ export function UserDataTable() {
 
   const goToCreate = () => navigate(toAddUser({ realm: realmName }));
 
-  if (!uiRealmInfo || !realm) {
+  if (uiRealmInfo.userProfileProvidersEnabled === undefined || !realm) {
     return <KeycloakSpinner />;
   }
 


ahus1:
@vramik - pinging you as you assessed the connected issue. Feel free to re-assign.

vramik:
@ssilvert could you please take a look?

ssilvert (approved):

Issue #40947, opened 2025-07-06
TiDB friendly MySQL Liquibase migration statements please


plevart:
### Description

Hi,

I'm trying to migrate from Keycloak 23 to 26. I installed Keycloak with mysql JDBC driver and connected it to TiDB database which is mostly MySQL compatible. It works great with it. Already at the time of installation of version 23, I encountered liquibase migration error due to unsupported DDL statements. I solved the problem at that time by manually entering equivalent DDL statements and inserting a line into DATABASECHANGELOG table to mark it successfully executed. At that time I could take a shortcut since the tables were all still empty, so instead of unsupported DDL statements which changed the type of the column by adding new column then migrating data from old column and then renaming new over old column, I just dropped the table and re-created it with new type of column already in place.

Now trying to migrate to Keycloak 26, There is an unsupported DDL statement in class org.keycloak.connections.jpa.updater.liquibase.custom.JpaUpdate25_0_0_MySQL_ConsentConstraints which goes like:

```SQL
CREATE TABLE TEMP_USER_CONSENT_IDS AS SELECT ...
```

The TiDB supported equivalent could be:

```SQL
CREATE TABLE TEMP_USER_CONSENT_IDS (...);
INSERT INTO TEMP_USER_CONSENT_IDS SELECT ...
```

The same goes for table TEMP_USER_CONSENT_IDS2 later on in the same class.

These 2 migrations are the only TiDB incompatible statements I encountered in Keycloak. It would be easy to make them compatible and still accommodate also MySQL and MariaDB databases. So is there any interest in the future to try to support TiDB too? It would mean creating a test that would verify the migrations with TiDB database too. Although TiDB is typically a cluster composed of 3 different components each in multiple instances, it is possible to run it in a single process in a container. There is an official image: `docker.io/pingcap/tidb:v8.5.2` when run without arguments starts a single-process TiDB instance with memory storage backend in a fraction of a second.

So my question is, would it technically be possible to add support for a new database which is basically MySQL database as application is concerned (MySQL JDBC driver is used). The only difference being in Liquibase migration scripts and classes of which most are common with MySQL and only in two occasions a little different. I know that making those two occasions compatible with TiDB after the fact by modifying these already created scripts might not be possible, but maybe a special sub-type of MySQL database could differentiate these two occasions where for MySQL the scripts would remain unchanged and for TiDB new compatible scripts would be provided.

WDYT?

### Motivation

[...]

PR #41439, opened 2025-07-25
Add TiDB as supported db

denniskniep:
Closes #41455

This PR adds [TiDB](https://github.com/pingcap/tidb) into the database test matrix.

### Introduction
TiDB is an open-source (Apache-2.0 license), cloud-native, distributed SQL database designed for high availability, horizontal and vertical scalability, strong consistency, and high performance.

TiDB is compatible with MySQL 8.0 (Except these [unsupported features](https://docs.pingcap.com/tidb/stable/mysql-compatibility/#unsupported-features) e.g. XA syntax). Therefore Keycloak can connect via jdbc mysql driver to TiDB (KC_DB=mysql)

[TiKV](https://github.com/tikv/tikv) is the underlying storage component of TiDB and a graduated project of the [Cloud Native Computing Foundation](https://cncf.io/) (CNCF)

### Migration Script Fix
I encountered one problem in liquibase migration scripts. The fix is part of that PR.

During migration there was an error in the statement:
`ALTER TABLE CLIENT_ATTRIBUTES DROP COLUMN VALUE`

Error Message was
> can't drop column value with composite index covered or Primary Key covered 

Therefore I modified the changeset `jpa-changelog-20.0.0.xml` in a way that the index is dropped before the column is dropped and recreated afterwards, but only if the changeset wasn´t executed yet (to keep backwards compatibility).

### Migration Tests
I didn´t add tidb into the migration-tests matrix, because the current tested `old-version: 24.0.4` fails when starting during running the migration script (see section "Migration Script Fix"). 
We can add it to the migration-tests matrix once this pr finds the way into a released version 
- [ ]  I will add an issue as a reminder once this PR is merged

### Related Issue:
https://github.com/keycloak/keycloak/issues/40947

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,7 +153,6 @@
         <!-- Databases - also published to db.adoc as "Tested Version" -->
         <tidb.version>v8.5.2</tidb.version>
         <tidb.container>mirror.gcr.io/pingcap/tidb:${tidb.version}</tidb.container>
-        <tidb-jdbc.version>8.3.0</tidb-jdbc.version>
         <mysql.version>8.4</mysql.version>
         <mysql.container>mirror.gcr.io/mysql:${mysql.version}</mysql.container>
         <mysql-jdbc.version>8.3.0</mysql-jdbc.version>
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,9 @@
         <twitter4j.version>4.1.2</twitter4j.version>
 
         <!-- Databases - also published to db.adoc as "Tested Version" -->
+        <tidb.version>v8.5.2</tidb.version>
+        <tidb.container>mirror.gcr.io/pingcap/tidb:${tidb.version}</tidb.container>
+        <tidb-jdbc.version>8.3.0</tidb-jdbc.version>
         <mysql.version>8.4</mysql.version>
         <mysql.container>mirror.gcr.io/mysql:${mysql.version}</mysql.container>
         <mysql-jdbc.version>8.3.0</mysql-jdbc.version>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -693,7 +693,7 @@ jobs:
     timeout-minutes: 75
     strategy:
       matrix:
-        db: [postgres, mysql, oracle, mssql, mariadb]
+        db: [postgres, mysql, oracle, mssql, mariadb, tidb]
       fail-fast: false
     steps:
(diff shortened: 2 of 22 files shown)


docs/guides/server/db.adoc
```diff
@@ -25,6 +25,7 @@ The server has built-in support for different databases. You can query the avail
 |Oracle Database | `oracle` | ${properties["oracledb.version"]}
 |PostgreSQL | `postgres` | ${properties["postgresql.version"]}
 |Amazon Aurora PostgreSQL | `postgres` | ${properties["aurora-postgresql.version"]}
+|TiDB | `tidb` | ${properties["tidb.version"]}
```
ahus1: While this PR adds tests, there might be more steps necessary to call it "supported". I'll create an "epic" type task to discuss this in more detail. 

Besides that, IMHO the second column would state `mysql` as we're using the MySQL DB driver for TiDB.

pom.xml
```diff
@@ -151,6 +151,9 @@
         <twitter4j.version>4.1.2</twitter4j.version>
 
         <!-- Databases - also published to db.adoc as "Tested Version" -->
+        <tidb.version>v8.5.2</tidb.version>
+        <tidb.container>pingcap/tidb:${tidb.version}</tidb.container>
```
ahus1: We're using `mirror.gcr.io` in all places to avoid rate limiting and terms of service of docker hub. Please adjust this container reference as well as all others related to tidb.

ahus1 (changes requested):
Thank you for this pull request. It is great to see that all tests passed in your repository with this minimal change. 

See below for some review comments. As with all changes, they should reference an issue. And as the TiDB tasks would be a bigger chunk of work, I'm creating an epic to manage those issues as well.

docs/guides/server/db.adoc
denniskniep: 100% agree

@ahus1 should I introduce a new section "Experimental databases" or remove it entirely ?

ahus1: Please remove it entirely for now until we have more guidance for users on how to use TiDB.

denniskniep:
@ahus1 resolved your review comments

ahus1 (approved):
Thank you, this change looks good to me. 

One small change I added was to make the TiDB tests use the regular MySQL driver version.

ahus1:
@vaceksimon - FYI, we've added another DB following the existing pattern.

PR #41440, opened 2025-07-25
Compliant with RFC8414, return server metadata at /.well-known/oauth-authorization-server/realms/{realm}

### LINKED ISSUE #40923: Compliant with RFC8414, return server metadata at https://keycloak.example.com/.well-known/oauth-authorization-server/realms/{realm}
### Description

Keycloak does not support RFC8414, and the .well-known endpoint that returns server metadata has the following format:
```dtd
https://keycloak.example.com/realms/{realm}/.well-known/oauth-authorization-server
```

RFC8414 specifies the following format:
```dtd
https://keycloak.example.com/.well-known/oauth-authorization-server/realms/{realm}
```

It would be better to add a feature that supports well-known endpoints in RFC8414 format while maintaining backward compatibility.

### Discussion

https://github.com/keycloak/keycloak/discussions/40809

### Motivation

It is better to comply with RFC8414 to support specifications such as OpenId4VCI.

### Details

For classes that accept realms/{realm}/.well-known/ and return metadata, how about accepting /.well-known/realms/{realm} as well?

tnorimat:
closes #40923

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.services.resources;
+
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.OPTIONS;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriBuilder;
+import jakarta.ws.rs.ext.Provider;
+import org.jboss.logging.Logger;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.protocol.oauth2.OAuth2WellKnownProviderFactory;
+import org.keycloak.services.cors.Cors;
+
+import java.util.List;
+
+@Provider
+@Path("/.well-known")
+public class ServerMetadataResource {
+
+    protected static final Logger logger = Logger.getLogger(ServerMetadataResource.class);
+
+    @Context
+    protected KeycloakSession session;
+
+    @OPTIONS
+    @Path("{provider}/realms/{realm}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getOAuth2AuthorizationServerWellKnownVersionPreflight(final @PathParam("provider") String providerName,
+                                                                          final @PathParam("realm") String name) {
+        if (!isValidProvider(providerName)) throw new NotFoundException();
+        return Cors.builder().allowedMethods("GET").preflight().auth().add(Response.ok());
+    }
+
+    @GET
+    @Path("{provider}/realms/{realm}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getOAuth2AuthorizationServerWellKnown(final @PathParam("provider") String providerName,
+                                                          final @PathParam("realm") String name) {
+        if (!isValidProvider(providerName)) throw new NotFoundException();
+        return RealmsResource.getWellKnownResponse(session, name, providerName, logger);
+    }
+
+    public static UriBuilder wellKnownOAuthProviderUrl(UriBuilder builder) {
+        return builder.path(ServerMetadataResource.class).path("{provider}/realms/{realm}");
+    }
+
+    private boolean isValidProvider(String providerName) {
+        // you can add codes here considering the current status of the implementation (preview, experimental).
+        if (OAuth2WellKnownProviderFactory.PROVIDER_ID.equals(providerName)) return true;
+        return false;
+    }
+}
--- a/services/src/main/java/org/keycloak/services/resources/RealmsResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/RealmsResource.java
@@ -180,6 +180,10 @@ public class RealmsResource {
     }
 
     private void resolveRealmAndUpdateSession(String realmName) {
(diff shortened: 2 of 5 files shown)


tnorimat:
@mposolda Hello,
The issue is related to Model Context Protocol support: https://github.com/keycloak/keycloak/issues/41521 .
If possible, I would appreciate it if you could check the PR.

services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
IngridPuppet, line 1: Hello @tnorimat - I am jumping in because a very identical change is required by OpenID4VCI Draft 16 as we will need `/.well-known/openid-credential-issuer/realms/{realm}`.

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-credential-issuer-metadata-
> Credential Issuers publishing metadata MUST make a JSON document available at the path formed inserting the string /.well-known/openid-credential-issuer into the Credential Issuer Identifier between the host component and the path component, if any.
>
> For example, the metadata for the Credential Issuer Identifier https://issuer.example.com/tenant would be retrieved from https://issuer.example.com/.well-known/openid-credential-issuer/tenant. The metadata for the Credential Issuer Identifier https://tenant.issuer.example.com would be retrieved from https://tenant.issuer.example.com/.well-known/openid-credential-issuer.

Could this PR take into account extensibility regarding well-known resources in general? I mean something close to the suggestion of @thomasdarimont here: https://cloud-native.slack.com/archives/C05KR0TL4P8/p1752669894360769?thread_ts=1752669762.102799&cid=C05KR0TL4P8 ?

```java
@Path(".well-known")
public class WellKnownRootResource {

    protected static final Logger logger = Logger.getLogger(WellKnownRootResource.class);

    @Context
    protected KeycloakSession session;

    @GET
    @Path("/{providerId}/realms/{realmName}")
    public Response getWellKnown(@PathParam("providerId") String providerId, @PathParam("realmName") String realmName) {

        RealmModel realm = session.realms().getRealmByName(realmName);
        if (realm == null) {
        }

        session.getContext().setRealm(realm);
        WellKnownProvider knownProvider = session.getProvider(WellKnownProvider.class, providerId);
        if (knownProvider == null) {
        }

        Object config = knownProvider.getConfig();
        return Response.ok(config).build();
    }
}
```

tnorimat: @IngridPuppet Hello, thank you for the feedback comment. I will incorporate the extensibility you suggested.

tnorimat:
@IngridPuppet I've revised the PR. Would it suit your purposes?

IngridPuppet (approved):
Thank you very much for updating the PR for extensibility. By adding IDs of WellKnownProvider's to the `validProviderList` array, we should have the correct endpoints automatically available. Additionally, I checked all the changes and agree with them.

tnorimat:
@IngridPuppet I slightly modified the codes. OID4VCI is an experimental feature, so we need to consider that to publish OID4VCI'S metadata.

```
    private boolean isValidProvider(String providerName) {
        // you can add codes here considering the current status of the implementation (preview, experimental).
        if (OAuth2WellKnownProviderFactory.PROVIDER_ID.equals(providerName)) return true;
        return false;
    }
```

services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
```diff
@@ -0,0 +1,73 @@
...
+import jakarta.ws.rs.ext.Provider;
+import org.jboss.logging.Logger;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.protocol.oauth2.OAuth2WellKnownProviderFactory;
+import org.keycloak.services.cors.Cors;
+
+import java.util.List;
+
+@Provider
+@Path("/")
```
thomasdarimont: Do we need that Path("/") annotation here? Since it is the root entry point it might lead to conflicts with other resource handlers.

tnorimat: @thomasdarimont Thank you for the comment. Yes, both root resource classes `WelcomeResources` and `ServerMetadataResource` have the same path annotation value `/`. To avoid this duplication, I modified as follows:
```
@Path("/.well-known")
public class ServerMetadataResource {

    protected static final Logger logger = Logger.getLogger(ServerMetadataResource.class);

    @Context
    protected KeycloakSession session;

    @OPTIONS
    @Path("{provider}/realms/{realm}")
    public Response getOAuth2AuthorizationServerWellKnownVersionPreflight(final @PathParam("provider") String providerName,
                                                                          final @PathParam("realm") String name) {
        if (!isValidProvider(providerName)) throw new NotFoundException();
        return Cors.builder().allowedMethods("GET").preflight().auth().add(Response.ok());
    }

    @GET
    @Path("{provider}/realms/{realm}")
    public Response getOAuth2AuthorizationServerWellKnown(final @PathParam("provider") String providerName,
                                                          final @PathParam("realm") String name) {
        if (!isValidProvider(providerName)) throw new NotFoundException();
        return RealmsResource.getWellKnownResponse(session, name, providerName, logger);
    }
```

thomasdarimont:
I just noticed that this only works as expected if Keycloak is configured with an empty context path.
An empty context path is the current default, but existing deployments are probably still using a context path like `/auth` or similar.

I don't know if it is possible to configure Keycloak to always listen on "/" for some requests despite the context path setting but if not we should document that users might need to use a reverse-proxy in front of Keycloak to route the `/.well-known` requests internally to the proper endpoint.

The following nginx configuration works for me:
```
        # https://id.acme.test/.well-known/ssf-configuration/auth/realms/ssf-demo
        # /.well-known/ssf-configuration/<realmpath>  ->  http://keycloak/<realmpath>/ssf/.well-known/ssf-configuration
        location ~ ^/\.well-known/ssf-configuration/(?<realmpath>.+)$ {
            proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $scheme;
            proxy_set_header X-Forwarded-Host  $host;
            proxy_set_header X-Forwarded-Port  $server_port;

            proxy_connect_timeout 2s;

            proxy_buffer_size       128k;
            proxy_buffers           4 256k;
            proxy_busy_buffers_size 256k;

            proxy_pass http://keycloak/$realmpath/ssf/.well-known/ssf-configuration$is_args$args;
        }
```

VinodAnandan:
@tnorimat @thomasdarimont @mposolda @stianst Would you be able to help this PR? Are there any significant concerns or blockers apart from the documentation? This PR is important for MCP and OID4VCI, and it is currently blocking the following issue: https://github.com/keycloak/keycloak/issues/41589

Cc : @Ogenbertrand @francis-pouatcha @IngridPuppet

thomasdarimont:
FYI I created this issue in the quarkus repository https://github.com/keycloak/keycloak/pull/41440 to discuss additional solutions to serve resources on the context-root even if a custom context path is used.

ahus1 (approved):
I've reviewed this PR, and aligned it with the RFCs contents. Some docs update might be missing, and I'll open a follow-up issue for that. 

I also reviewed the previous comments. While I see that this won't work with the `/auth` prefix, this shouldn't keep us from merging this PR as is.

ahus1:
Follow-up docs task: #42323

Issue #40083, opened 2025-05-30
[OID4VCI] Create dedicated constants-class


Captain-P-Goldfish:
### Description

As suggested by @mposolda in **'https://github.com/keycloak/keycloak/pull/39479'** we should create an `OpenId4VCIConstants`-class in the module **`keycloak-services`** that holds the appropriate attribute names and default-values for the specific protocol.

Captain-P-Goldfish:
The constants class was already created in https://github.com/keycloak/keycloak/pull/39479 in class `org.keycloak.oid4vci.Oid4VciConstants` in module `keycloak-services`

forkimenjeckayang:
working on this

PR #41481, opened 2025-07-29
[OID4VCI] Create dedicated constants-class

forkimenjeckayang:
# Centralize OID4VCI Protocol Constants in Oid4VciConstants and Refactor Usages

This PR implements a full centralization of OID4VCI protocol-related constants in the `Oid4VciConstants` class.

## Key changes:
- **Centralized protocol attribute names, claim keys, and default values** for OID4VCI in `Oid4VciConstants` (`server-spi-private`).
- **Refactored all usages** in the OID4VCI implementation to reference these constants, ensuring a single source of truth.
- **Kept class-level constants**  and related classes for backward compatibility, referencing the centralized values.
- **Verified by running all tests** in `testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc` (all passing).

## Benefits:
- Simplifies maintenance and future updates to protocol constants.
- Reduces risk of inconsistencies and duplication.
- Makes the codebase more robust and easier to understand.

---
**Cosses:** https://github.com/keycloak/keycloak/issues/40083


IngridPuppet (approved):
I had the chance to check these changes before they reached upstream. LGTM.

tnorimat (approved):
LGTM.

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -51,15 +51,15 @@ public class OID4VCLoginProtocolFactory implements LoginProtocolFactory, OID4VCE
...
-    public static final String PROTOCOL_ID = Oid4VciConstants.OID4VC_PROTOCOL;
-
     private static final String CLIENT_ROLES_MAPPER = "client-roles";
     private static final String USERNAME_MAPPER = "username";
     private static final String SUBJECT_ID_MAPPER = "subject-id";
     private static final String EMAIL_MAPPER = "email";
     private static final String LAST_NAME_MAPPER = "last-name";
     private static final String FIRST_NAME_MAPPER = "first-name";
 
+    public static final String PROTOCOL_ID = Oid4VciConstants.OID4VC_PROTOCOL;
```
mposolda, line 61: Do we still need the constant in this class? It looks like it is not used anymore as occurences were replaced by `Oid4VciConstants.OID4VC_PROTOCOL` ?

mposolda (commented):
@forkimenjeckayang @tnorimat @IngridPuppet Thanks for the PR and review! I've added one small inline comment. Could you please check?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
forkimenjeckayang: Hello @mposolda , i think we still need it there. Classes like those listed below still use it.
- OID4VCIssuerEndpoint
- OID4VCIssuerWellKnownProvider
- OID4VCMapper
- OID4VCSubjectIdMapper
- OID4VCTargetRoleMapper
- OID4VCUserAttributeMapper
- ClientScopeResource
- NonceEndpointTest
- OID4VCIssuerWellKnownProviderTest
- OID4VCTest

mposolda: @forkimenjeckayang Ok, Thanks for pointing that!

mposolda (approved):

PR #41500, opened 2025-07-29
Make fileChooser platform independent.

### LINKED ISSUE #41474: File choosing tests fail on Windows
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Admin console tests that need to choose a file, such as partial import tests, are failing on Windows.  The reason is because the chooseFile utility is not written in a platform-independent way.

### Version

main

### Regression

### Expected behavior

Tests using chooseFile() pass on Windows.

### Actual behavior

Tests using chooseFile() fail on Windows.

### How to Reproduce?

Run admin Playwright tests on Windows.

### Anything else?

_No response_

ssilvert:
Fixes #41474

### DIFF
--- a/js/apps/admin-ui/test/utils/file-chooser.ts
+++ b/js/apps/admin-ui/test/utils/file-chooser.ts
@@ -1,11 +1,15 @@
 import { Page } from "@playwright/test";
 import path from "path";
+import { fileURLToPath } from "url";
 
 export async function chooseFile(page: Page, file: string) {
   const fileChooserPromise = page.waitForEvent("filechooser");
   await page.getByText("Browse...").click();
   const fileChooser = await fileChooserPromise;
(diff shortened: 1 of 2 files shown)


js/apps/admin-ui/test/utils/file-chooser.ts
```diff
@@ -1,11 +1,15 @@
 import { Page } from "@playwright/test";
 import path from "path";
+import { fileURLToPath } from "url";
```
jonkoops: Nit: We should use the `node:` prefix for imports that come from Node's standard library. e.g. also `path` should be `node:path`.
```suggestion
import { fileURLToPath } from "node:url";
```

jonkoops (dismissed):
LGTM

jonkoops (approved):

PR #41505, opened 2025-07-29
feat(email): Add cooldown period to prevent rapid resend of verificat…

### LINKED ISSUE #41331: Prevent sending massive amount of emails if a user clicks multiple times to get a new verify email link
### Description

It would be nice to prevent sending a massive amount of emails if a user clicks multiple times to get a new verify email link.

This can result in :

- The domain used to send email was banned by the global mail server of the receiver (happened in production)
- Spamming a not-owned email box if an account is created with a target email

It would be nice to have a parameter to set the number of seconds between each demand a user can send a mail.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

smoshie25:
### Summary

This PR introduces a realm-configurable cooldown period between verification email resends to prevent users from triggering excessive emails by repeatedly clicking the "Resend Email" link.

Fixes [#41331](https://github.com/keycloak/keycloak/issues/41331)

### Details

- Introduces support for a realm-level attribute: `emailResendCooldownSeconds`
- If not defined, falls back to a default cooldown of 30 seconds
- Stores `emailLastSent` timestamp in the user's authentication session
- Respects the cooldown on each resend attempt

### Configuration

To set the cooldown:

```bash
curl -X PUT http://localhost:8080/admin/realms/<realm> \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{
    "attributes": {
      "emailResendCooldownSeconds": "60"
    }
  }'

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -80,6 +80,12 @@ In order to maintain backwards compatibility, {project_name}'s upgrade will modi
 
 For more information about client configuration, please see link:{adminguide_link}#_client-saml-configuration[Creating a SAML client] chapter in the {adminguide_name}.
 
+=== Validate email action
+
+When validating an email address as a required action or an application initiated action, a user can resend the verification email by default only every 30 seconds, while in earlier versions there was no limitation in re-sending the email.
+
+Administrators can configure the interval per realm in the Verify Email required action in the Authentication section of the realmm.
+
 === Tracing extended for embedded Infinispan caches
 
 When tracing is enabled, now also calls to other nodes of a {project_name} cluster will create spans in the traces.
--- a/server-spi-private/src/main/java/org/keycloak/models/Constants.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/Constants.java
@@ -103,6 +103,9 @@ public final class Constants {
     public static final String KC_ACTION_ENFORCED = "kc_action_enforced";
     public static final int KC_ACTION_MAX_AGE = 300;
     public static final String MAX_AUTH_AGE_KEY = "max_auth_age";
+    public static final String EMAIL_RESEND_COOLDOWN_SECONDS = "email_resend_cooldown";
+    public static final int EMAIL_RESEND_COOLDOWN_DEFAULT_SECONDS = 30;
+    public static final String EMAIL_RESEND_COOLDOWN_KEY_PREFIX = "verify-email-cooldown-";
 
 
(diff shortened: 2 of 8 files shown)


ahus1 (dismissed):
Thank you for looking into this task. 

Some time last year, the required actions became configurable via the admin UI. The UpdatePassword required action was the first - see ab376d91012ce7168256a2f198287eafed83ea35 for the change. This would allow users to configure it by realm and via the UI and the API.  

As an alternative, there is a the SingleUseObjectProvider where you can put items into a cache that is shared across nodes and authentication sessions. See https://github.com/keycloak/keycloak/blob/478e0b3264e59fad467ad4c4a2b0237487c81749/services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientValidator.java#L250-L253 for an example. 

It would be good if the error message could be a bit more descriptive by showing the number of seconds it would take for the next resend to be allowed. 

A formal requirement for all PRs: They need a sign-off as described in the CONTRIBUTING.md file. 

Eventually we would also like to see a test - I can help with that once the functionality is stable.

smoshie25:
I added a configuration for the cooldown timer alongside Max age on the "Email Verify" config at the Required Action as seen on the image below
image

Let me know if any required modifications are needed on the code.
@ahus1 Kindly help with test writing. Thanks

ahus1:
@smoshie25 - can you please sign-off the PR as described in the CONTRIBUTING.md file? 

Ideally do not merge the main branch into the PR, instead rebase the PR as it is simpler to review for the maintainers. 

So I suggest to: 

* Rebase the PR
* Squash the changes
* Sign-off the squashed commit
* Force-push to this branch

Once the commit is signed-off, I can help with the test.

ahus1 (commented):
I've reviewed the change, and pushed an additional commit. 

Here's what I changed on a high level:

* When the user is presented with the verify email required action, an email is sent automatically. To avoid showing the user an error message for this initial email, the email will always be sent out for this flow. 
* The message is now localizable
* I found a way to store the expiry in the map, so we can show it in the error message
* There is now a test

Something I considered but then did not do: 

* Setting the entry in the cache only after the email is sent out successfully. I see that it will redirect to an error page anyway, and wouldn't keep the user on the same page. 

Please review and let me know your thoughts. I'll be triggering the CI, let's see what that brings.

sguilhen:
Looks like some related tests need to be updated.

ahus1:
@sguilhen - the tests should now be green. Can you please re-review? Thanks!

sguilhen (approved):

ahus1 (approved):

PR #41519, opened 2025-07-29
Initial commit for the RLM feature

vramik:
Closes https://github.com/keycloak/keycloak/issues/41516


sguilhen (approved):

pedroigor (commented):
Nice stuff, @vramik !

We need at least a feature (experimental) flag to merge this one. I think we should also consider adding tests to cover the transient user use case and drop, for now, the commit https://github.com/keycloak/keycloak/pull/41519/commits/fa94dba925bcef4f9a4e3f2b84cfcc9a126f64da.

Other than that, we can start building on top of it and refine the SPI/capabilities once we move forward.

pedroigor (approved):

PR #41536, opened 2025-07-30
Upgrade to Quarkus 3.25.0

### LINKED ISSUE #41186: Upgrade to Quarkus 3.25.0
### Description

We should upgrade to the Quarkus 3.25.0 release once is released. 

The purpose of this issue is to track connected issues for now. 

### Solves
- https://github.com/keycloak/keycloak/issues/41184

### Unblocks
- https://github.com/keycloak/keycloak/issues/40687

mabartos:
- Closes #41186

All issues with Liquibase 4.33.0 should be resolved for `main` now, so this upgrade should go seamlessly(hopefully).

Blog post: https://quarkus.io/blog/quarkus-3-25-released/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,8 +45,8 @@
 
         <asciidoctor.plugin.version>1.5.8</asciidoctor.plugin.version>
 
-        <quarkus.version>3.24.5</quarkus.version>
-        <quarkus.build.version>3.24.5</quarkus.build.version>
+        <quarkus.version>3.25.0</quarkus.version>
+        <quarkus.build.version>3.25.0</quarkus.build.version>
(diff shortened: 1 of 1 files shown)


mabartos:
@Pepo48 @shawkins Tests are green. Could you please check this PR? Thanks!

Pepo48 (approved):
Thanks, @mabartos - LGTM. Also, I double-checked the dependency tree, and the correct versions are applied.

mabartos:
Coming with Quarkus 3.25.0, there are some changes on the Micrometer side, so we can also double-check if KC Grafana dashboards work as expected. 

cc: @keycloak/sre

ahus1 (approved):

ahus1:
@mabartos - thank you for the pointer. We have at least a test that checks if the metrics we mention in the docs and that we also use in the Grafana dashboard still exist. So there is a good chance that they will still work.

PR #41540, opened 2025-07-30
Getting error 405 'Method Not Allowed' when calling the 'certs' endpo…

### LINKED ISSUE #41537: Getting error 405 "Method Not Allowed" when calling the "certs" endpoint with HEAD method
### Before reporting an issue

### Area

oidc

### Describe the bug

There is a 405 error code when calling the "certs" endpoint (e.g. http://localhost:8080/realms/master/protocol/openid-connect/certs) with HEAD http method on RHBK 26.

This issue wasn't happening on RHBK 22, as the endpoint was responding with 200 code.

### Version

Any Keycloak since 24 or newer

### Regression

### Expected behavior

There should be a successful response to the HEAD request to the "certs" endpoint. Method `HEAD` should behave in a same way as method `GET` . It should just omit to return a body.

### Actual behavior

There is 405 error in response to the `HEAD` request to the `certs` endpoint (even if `GET` method works).

Some Keycloak endpoints are working with HEAD method (For example well-known endpoint http://localhost:8080/realms/master/.well-known/openid-configuration ). Some others are not. Some resteasy resource classes are broken, probably due the bug in quarkus resteasy-reactive .

### How to Reproduce?

- Start Keycloak server

- Run this request
`curl -vk http://localhost:8080/realms/master/protocol/openid-connect/certs -I --http1.1`

- It works without parameter `-I` (which means calling regular `GET` instead of `HEAD`)

### Anything else?

The issue happens since Keycloak 24 when resteasy-reactive was introduced.

It is caused by quarkus resteasy reactive bug: https://github.com/quarkusio/quarkus/issues/49172

mposolda:
…int with HEAD method

closes #41537

This PR adds a workaround to the quarkus issue https://github.com/quarkusio/quarkus/issues/49172 . Once that one is fixed in quarkus, we can remove the workaround possibly.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/OIDCLoginProtocolService.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/OIDCLoginProtocolService.java
@@ -17,6 +17,7 @@
 
 package org.keycloak.protocol.oidc;
 
+import jakarta.ws.rs.HEAD;
 import org.jboss.resteasy.reactive.NoCache;
 import org.keycloak.http.HttpRequest;
 import org.keycloak.OAuthErrorException;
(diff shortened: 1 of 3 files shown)


pskopek (approved):
LGTM!

mposolda:
@pskopek Thanks!

PR #41548, opened 2025-07-30
Avoid deleting old client sessions

### LINKED ISSUE #41427: Parallel token exchange fails if client session is expired
### Before reporting an issue

### Area

token-exchange

### Describe the bug

I have experienced this in `token-exchange:v1`, but I think other areas might also be affected.
When exchanging a token from client-A into one from client-B, if there is an existing client session for client-B in the database and this session is expired, the `ClientSessionPersistentChangelogBasedTransaction.getSessionEntityFromPersister` will try to remove the existing session from the database. If the same token exchange is called twice in parallel, there is a high probability that one of them will fail with a `StaleObjectStateException` because the other one has already deleted the session.

I believe this failure can be avoided, the exception can just be ignored if the session has already been deleted from the database.

It does not matter if both calls end up on the same server (in a clustered environment) or not.

Tested in RHBK 26.2.5 and upstream 26.3.2 (both latest at the time of writing).

### Version

RHBK 26.2.5, 26.3.2

### Regression

### Expected behavior

All token exchanges succeed

### Actual behavior

one of the exchanges fails with HTTP 500

### How to Reproduce?

For local testing, I reduced the client session timeouts to 30s idle / 1m max and ran the following script a couple of times with >30s pause in between:
```bash
#!/usr/bin/env bash

token="$(<get token for client-A>)"

curl -X POST -sS --location "http://localhost:8080/auth/realms/<realm>/protocol/openid-connect/token" \
    -H "Content-Type: application/x-www-form-urlencoded" \
    -d 'client_id=client-B&client_secret=client-B-secret&grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange&requested_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Aaccess_token&subject_token='"$token" -i > curl1.txt &
curl -X POST -sS --location "http://localhost:8080/auth/realms/<realm>/protocol/openid-connect/token" \
    -H "Content-Type: application/x-www-form-urlencoded" \
    -d 'client_id=client-B&client_secret=client-B-secret&grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange&requested_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Aaccess_token&subject_token='"$token" -i > curl2.txt &

wait

cat curl*.txt
```

### Anything else?

_No response_

ahus1:
If this is done concurrently, this can load to errors with Hibernate and its optimistic locking

Closes #41427

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java
@@ -408,13 +408,8 @@ public class InfinispanUserSessionProvider implements UserSessionProvider, Sessi
             return null;
         }
 
-        AuthenticatedClientSessionAdapter clientAdapter = importClientSession((UserSessionAdapter<?>) userSession, clientSession, getTransaction(true),
+        return importClientSession((UserSessionAdapter<?>) userSession, clientSession, getTransaction(true),
                 getClientSessionTransaction(true), true);
-
-        if (clientAdapter == null) {
-            persister.removeClientSession(userSession.getId(), client.getId(), true);
-        }
-        return clientAdapter;
     }
 
     private AuthenticatedClientSessionEntity getClientSessionEntity(UUID id, boolean offline) {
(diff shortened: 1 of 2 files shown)


ahus1:
@lbilger - I've pushed a change that will skip the deletion. Can you give this change a try in a local build? 

I might need some more time to come up with a local test in our test suite ... this might happen also only when an item is no longer in the cache?

ahus1:
@pruivo - please give it a review when you have the time. I can't think of a good way to add a test for this, and I would hope the existing tests should suffice for that. Let's hear what you think of this. Thanks!

pruivo:
@ahus1 it looks fine to me as it seems a waste of queries/operation to remove expired sessions (the remove expired task will take care of it)

lbilger:
Hi @ahus1, some days ago I have deployed a locally-built version of Keycloak including your fix to one of our environments and haven't seen this error there since. So to me it looks like your fix is working. Looking forward to see it merged and released. Thanks a lot for your help!

pruivo (dismissed):

pruivo:
@ahus1 if it is ok, may you fix it here too?

https://github.com/keycloak/keycloak/blob/9322d71d61390a3aa833c18fc89fe676fb066df1/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java#L411-L416

ahus1:
@pruivo - thanks, I reworked the PR. Please re-review.

pruivo (approved):
LGTM

PR #41550, opened 2025-07-30
fix: making the /opt/keycloak read-only

### LINKED ISSUE #25761: Docker image - is full write access to /opt/keycloak necessary?
### Before reporting an issue

### Area

ci

### Describe the bug

https://github.com/keycloak/keycloak/blob/main/quarkus/container/Dockerfile

```
## Copy /opt/keycloak into the final stage. In the process, chown /opt/keycloak and all files/dirs to be owned by the non-root user
COPY --from=ubi-micro-build --chown=1000:0 /opt/keycloak /opt/keycloak

## Create the user
RUN echo "keycloak:x:0:root" >> /etc/group && \
    echo "keycloak:x:1000:0:keycloak user:/opt/keycloak:/sbin/nologin" >> /etc/passwd

## Switch to the user
USER 1000

EXPOSE 8080
EXPOSE 8443

## Run the script as the user on container startup
ENTRYPOINT [ "/opt/keycloak/bin/kc.sh" ]
```

The issue here is that every file in /opt/keycloak is writable during runtime. When running as non-root in a container, it's recommended for files that do not need to be written to (such as the kc.sh script, jar files in the distribution, configuration files, etc) have root as the owner but with world readable permissions, to allow the non-root user to read the files but prevent it from making changes to them.

### Version

main branch - built from source

### Expected behavior

Docker build process should do a mass chown for all of /opt/keycloak during the copy, but instead of setting the owner to UID 1000, it should use UID 0. Once everything is copied, then selectively chown the files/directories that need write permissions so that the user has the permissions it needs but not permissions to edit the keycloak distribution or configuration during runtime.

This would more accurately reflect a typical application installation and will help to further enhance the security of the the container image in case of an RCE being identified and exploited in the wild because the user Keycloak is running as would not have permissions to make further changes to the container's runtime overlay filesystem.

### Actual behavior

The permissions are as below:

```
❯ docker run --rm -it --entrypoint "" keycloak/keycloak:20.0.5 /bin/bash                                                    56m 37s ▼
Unable to find image 'keycloak/keycloak:20.0.5' locally
20.0.5: Pulling from keycloak/keycloak
d7c06497d5ce: Pull complete
604860b8efed: Pull complete
0696ddd1a0cc: Pull complete
Digest: sha256:054ef67eb7dae0129bbb9eb0e0797fd2392cd6d135094a6063ae7ff7773ef81f
Status: Downloaded newer image for keycloak/keycloak:20.0.5
bash-4.4$ pwd
/
bash-4.4$ ls -la
total 64
drwxr-xr-x   1 root root 4096 Dec 21 18:44 .
drwxr-xr-x   1 root root 4096 Dec 21 18:44 ..
dr-xr-xr-x   2 root root 4096 Aug  9  2021 afs
lrwxrwxrwx   1 root root    7 Aug  9  2021 bin -> usr/bin
dr-xr-xr-x   2 root root 4096 Aug  9  2021 boot
drwxr-xr-x   5 root root  360 Dec 21 18:44 dev
-rwxr-xr-x   1 root root    0 Dec 21 18:44 .dockerenv
drwxr-xr-x   1 root root 4096 Dec 21 18:44 etc
drwxr-xr-x   2 root root 4096 Aug  9  2021 home
lrwxrwxrwx   1 root root    7 Aug  9  2021 lib -> usr/lib
lrwxrwxrwx   1 root root    9 Aug  9  2021 lib64 -> usr/lib64
drwxr-xr-x   2 root root 4096 Aug  9  2021 media
drwxr-xr-x   2 root root 4096 Aug  9  2021 mnt
drwxr-xr-x   1 root root 4096 Dec 21 01:19 opt
dr-xr-xr-x 303 root root    0 Dec 21 18:44 proc
dr-xr-x---   3 root root 4096 Dec  7 06:09 root
drwxr-xr-x   4 root root 4096 Dec  7 06:09 run
lrwxrwxrwx   1 root root    8 Aug  9  2021 sbin -> usr/sbin
drwxr-xr-x   2 root root 4096 Aug  9  2021 srv
dr-xr-xr-x  12 root root    0 Dec 21 18:44 sys
drwxrwxrwt   2 root root 4096 Aug  9  2021 tmp
drwxr-xr-x   1 root root 4096 Dec 21 01:19 usr
drwxr-xr-x   1 root root 4096 Dec 21 01:19 var
bash-4.4$ cd /opt
bash-4.4$ ls -la
total 20
drwxr-xr-x 1 root     root 4096 Dec 21 01:19 .
drwxr-xr-x 1 root     root 4096 Dec 21 18:44 ..
drwxr-xr-x 1 keycloak root 4096 Dec 21 01:19 keycloak
bash-4.4$ cd keycloak/
bash-4.4$ ls -la
total 64
drwxr-xr-x 1 keycloak root  4096 Dec 21 01:19 .
drwxr-xr-x 1 root     root  4096 Dec 21 01:19 ..
drwxrwxr-x 1 keycloak root  4096 Dec 21 01:13 bin
drwxrwxr-x 3 keycloak root  4096 Dec 21 01:19 conf
drwxrwxr-x 2 keycloak root  4096 Dec 21 01:19 data
drwxrwxr-x 5 keycloak root  4096 Dec 21 01:13 lib
-rw-rw-r-- 1 keycloak root 11358 Dec 21 00:57 LICENSE.txt
drwxrwxr-x 2 keycloak root  4096 Dec 21 01:19 providers
-rw-rw-r-- 1 keycloak root   492 Dec 21 00:57 README.md
drwxrwxr-x 2 keycloak root  4096 Dec 21 01:19 themes
-rw-rw-r-- 1 keycloak root    36 Dec 21 00:57 version.txt
bash-4.4$ whoami
keycloak
bash-4.4$ ls -la bin
total 48
drwxrwxr-x 3 keycloak root 4096 Dec 21 01:13 .
drwxr-xr-x 8 keycloak root 4096 Dec 21 01:19 ..
drwxrwxr-x 3 keycloak root 4096 Dec 21 01:13 client
-rwxrwxr-x 1 keycloak root 1009 Dec 21 00:57 federation-sssd-setup.sh
-rw-rw-r-- 1 keycloak root  842 Dec 21 00:57 kcadm.bat
-rwxrwxr-x 1 keycloak root  832 Dec 21 00:57 kcadm.sh
-rwxrwxr-x 1 keycloak root 5969 Dec 21 00:57 kc.bat
-rw-rw-r-- 1 keycloak root  863 Dec 21 00:57 kcreg.bat
-rwxrwxr-x 1 keycloak root  852 Dec 21 00:57 kcreg.sh
-rwxrwxr-x 1 keycloak root 5222 Dec 21  00:57 kc.sh
bash-4.4$ touch bin/kc.sh ## I should get an error here but I do not!
bash-4.4$ ls -la bin
total 56
drwxrwxr-x 3 keycloak root 4096 Dec 21 01:13 .
drwxr-xr-x 8 keycloak root 4096 Dec 21 01:19 ..
drwxrwxr-x 3 keycloak root 4096 Dec 21 01:13 client
-rwxrwxr-x 1 keycloak root 1009 Dec 21 00:57 federation-sssd-setup.sh
-rw-rw-r-- 1 keycloak root  842 Dec 21 00:57 kcadm.bat
-rwxrwxr-x 1 keycloak root  832 Dec 21 00:57 kcadm.sh
-rwxrwxr-x 1 keycloak root 5969 Dec 21 00:57 kc.bat
-rw-rw-r-- 1 keycloak root  863 Dec 21 00:57 kcreg.bat
-rwxrwxr-x 1 keycloak root  852 Dec 21 00:57 kcreg.sh
-rwxrwxr-x 1 keycloak root 5222 Dec 21 18:42 kc.sh
bash-4.4$ ## the timestamp on kc.sh has changed!
```

### How to Reproduce?

Follow what I did in the example above

### Anything else?

_No response_

shawkins:
closes: #25761

Looking to close out low-hanging issues. This change leaves /opt/keycloak as read-only with writability only for lib/quarkus and /data.

To prevent backwards compatibility issues, we may need to mark the provider and theme directories as writable - for example does it seem possible that someone's custom image uses the keycloak user to copy content from somewhere else to one of these directories. 

@ahus1 @mabartos @ASzc - WDYT, or should we just close #25761 without changes to our Dockerfile?

### DIFF
--- a/quarkus/container/Dockerfile
+++ b/quarkus/container/Dockerfile
@@ -26,9 +26,10 @@ ENV LANG en_US.UTF-8
 ENV KC_RUN_IN_CONTAINER true
 
 COPY --from=ubi-micro-build /tmp/null/rootfs/ /
-COPY --from=ubi-micro-build --chown=1000:0 /opt/keycloak /opt/keycloak
+COPY --from=ubi-micro-build --chown=0:0 /opt/keycloak /opt/keycloak
 
-RUN echo "keycloak:x:0:root" >> /etc/group && \
+RUN chown 1000:0 /opt/keycloak/lib/quarkus /opt/keycloak/data && \
+    echo "keycloak:x:0:root" >> /etc/group && \
     echo "keycloak:x:1000:0:keycloak user:/opt/keycloak:/sbin/nologin" >> /etc/passwd
 
 USER 1000


ASzc:
The image we ship is documented as mutable, i.e. [one kc.sh build operation](https://www.keycloak.org/server/containers#_writing_your_optimized_keycloak_containerfile) must be possible for every user to run. Making the directory read only at the point that we ship the image would break this.

If we think this is a valid hardening measure, then perhaps the documentation can recommend some appropriate steps for the user to make the directory read-only at the point they're done optimizing/customizing it.

shawkins:
> Making the directory read only at the point that we ship the image would break this.

It would not break because the /opt/keycloak/lib/quarkus directory is specifically marked as writable.

ASzc:
Ah true. Although I am also concerned about other customizations like SPI providers, themes, etc. Is there an impact on the ability for users to `ADD` those?

shawkins:
> Ah true. Although I am also concerned about other customizations like SPI providers, themes, etc. Is there an impact on the ability for users to `ADD` those?

That's what I was mentioning in the initial pr comment - if those operations are possibly being done as the keycloak user, then yes we'd need to additionally mark the provider and theme directories as writable.

@ASzc if it seems too risky to bake a limited set of writable directories into the base image, then we can simply close out https://github.com/keycloak/keycloak/issues/25761 and just document what needs to be done to satisfy https://github.com/keycloak/keycloak/issues/11286

mabartos:
@shawkins @ASzc This is an interesting case as we're not sure what users/customers do, and it might be fragile to do the hardening. 

Naturally, if we make these paths writeable:
- conf/
- providers/
- themes/
- lib/quarkus

it should be fine to have Keycloak run as expected. 

However, there are these requirements for the `conf/` to be read-only as well, but that might be potentially a point of failure to some deployments doing custom stuff. But the config should be handled beforehand in the Dockerfile, so it might not be an issue. 

As mentioned, the safest way would be to document the approach. On the other hand, it'd be good to have such hardening and be opinionated if we're completely sure (but probably we're not). We could potentially start with the bare minimum (as I've mentioned above) for the writeability and do the rest read-only for which we're sure (most likely). Then we can move forward if the confidence rises. 

JFYI - in the parent issue is mentioned preventing including JARs into the distribution, but we have open /providers folder that manages whatever JAR is there :)) so, anyway...

shawkins:
> JFYI - in the parent issue is mentioned preventing including JARs into the distribution, but we have open /providers folder that manages whatever JAR is there :)) so, anyway...

They just mean that the user running the keycloak process shouldn't be able to write to anything more than necessary. 

All the keycloak user needs writability to when running is /data and /lib/quarkus (and even that could be read-only in an optimized image). The other directories like conf/, providers/, themes/ at runtime just need read access. 

If we are being opinionated then writing to conf/, providers/, and themes/ is an image building concern. 

But I share the concern that customizations could be using the keycloak user to write to these directories, which would make a change to their writability breaking.

vmuzikar:
As discussed offline, this would be a breaking change with limited benefits. Instead, we'll document folders that need to be writeable for read only containers.

vmuzikar (changes requested):
As discussed offline, this would be a breaking change with limited benefits. Instead, we'll document folders that need to be writeable for read only containers.

Issue #41551, opened 2025-07-30
Support for key size 3072 in rsa-generated key providers


eicki:
### Description

The rsa-generated and rsa-enc-generated key providers only support key sizes of 1024, 2048 and 4096. I like to have support for a key size of 3072.

### Discussion

https://cloud-native.slack.com/archives/C056XU905S6/p1753788652027369

### Motivation

While NIST is fine with a key size of 2048 until 2030. The German BSI already now requires a key size of 3072 for RSA  (see https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf). 
While it is possible to upload manual key materials it is very convenient to just change the key size in the preconfigured generated key providers. Unfortunatelly it is only possible to choose between 2048 and 4096. As this increases signature and thus token size the use of a key size of 4096 is usually an overkill. But if you want to be BSI compliant you can also not stay with 2048. Thus I like to add 3072 to allowed key size values. Also FIPS states 3072 as an allowed key size.  

### Details

Just adding 3072 to org.keycloak.common.crypto.CryptoProvider#getSupportedRsaKeySizes and its derived class Fips1402StrictCryptoProvider should be sufficient to add that support.
Furthermore it would be convenient to be able to define the default key size via spi option such that you can setup a system that just uses a key size of 3072 by default and you don't have to take care of that for every realm you create. That can easily be added to AbstractGeneratedRsaKeyProviderFactory.

PR #41552, opened 2025-07-30
Support for RSA Key Size of 3072

eicki:
I added 3072 as a supported key size in the first commit (see [#41551](https://github.com/keycloak/keycloak/issues/41551)).
The second commit makes the default key size configurable via a spi config. I left the default to 2048. If you don't like that idea, feel free to revert that second change.

### DIFF
--- a/.github/scripts/run-fips-it.sh
+++ b/.github/scripts/run-fips-it.sh
@@ -9,7 +9,7 @@ if [ $? -ne 0 ]; then
 fi
 STRICT_OPTIONS=""
 if [ "$1" = "strict" ]; then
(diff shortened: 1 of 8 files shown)


ahus1 (changes requested):
Thank you for the PR. I like the SPI option. 

One of the FIPS tests still fails with the message 

> ServerInfoTest.testServerInfo:86 Expected: [2048, 4096], was: [2048, 3072, 4096]

so it needs one more change. 

The UI tests were broken/very unstable yesterday. This should be fixed today - sorry for that.

eicki:
@ahus1 fixed it, now all tests are green

ahus1 (approved):
Thank you for the change, looks good to me.

ahus1:
@keycloak/core-clients - FYI, this enhancement now landed in KC main

PR #41570, opened 2025-07-31
Wrap deprecated passkeys authenticator behind the feature

### LINKED ISSUE #40696: Remove the current conditionalUI authenticator
### Description

This is a follow-up of https://github.com/keycloak/keycloak/pull/40674 . We can remove the passkeys conditional UI authenticator, which was removed in Keycloak 26.3.0. We might do that when passkeys are supported (likely in 26.4.0).

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mposolda:
closes #40696

We have the issue for removing the **Passkeys Conditional UI Authenticator**, which is deprecated since 26.3.0. In theory, it should be fine to remove the existing stuff as long as it was wrapped behind the *preview* feature flag, which was the case of passkeys. However:

- When the authenticator is currently configured in the authentication flow in the KEycloak 26.3 or older, then after removing the authenticator implementation, users will not be able to authenticate to the realm. Also administrator cannot display authentication flow with unavailable authenticator in the admin console.

- Customers migrating from RHBK 26.2 to RHBK 26.4 do not have the "deprecation" period for the authenticator as the authenticator is not yet deprecated in the RHBK 26.2. So if we remove it in RHBK 26.4, it is "straight removal" without any deprecation period (which could be still fine in theory due the preview feature, but possibly not so great...).

Hence gave it another though to rather wrap this authenticator behind the "deprecated feature", which should be explicitly enabled if people want this deprecated authenticator to be used. This give administrator a chance to start the server with the deprecated feature and re-configure their flows to the supported way and properly migrate. Then finally remove this in Keycloak 27, which should be fine.

NOTE: Draft PR for now as I am curious for the feedback.

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -125,6 +125,7 @@ public class Profile {
         ORGANIZATION("Organization support within realms", Type.DEFAULT),
 
         PASSKEYS("Passkeys", Type.PREVIEW, Feature.WEB_AUTHN),
+        PASSKEYS_CONDITIONAL_UI_AUTHENTICATOR("Passkeys conditional UI authenticator", Type.DEPRECATED, FeatureUpdatePolicy.ROLLING_NO_UPGRADE, Feature.PASSKEYS),
 
         USER_EVENT_METRICS("Collect metrics based on user events", Type.DEFAULT),
 
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -110,6 +110,14 @@ The options `+--spi-user-sessions--infinispan--offline-session-cache-entry-lifes
 
 Instead use the options `cache-embedded-offline-sessions-max-count` and `cache-embedded-offline-client-sessions-max-count` to limit the memory usage if the default of 10000 cache offline user and client sessions does not work in your scenario.
 
+=== Deprecated Passkeys Conditional UI Authenticator requires a feature
+
+The authenticator *Passkeys Conditional UI Authenticator*, which was deprecated in the previous version 26.3.0, is still available for now, but it requires the feature
+`passkeys_conditional_ui_authenticator` to be explicitly enabled during server startup. The feature itself is deprecated and disabled by default.
+This allows administrator to start the server and re-configure authentication flows for passkeys authentication in a recommended way as described
+in the link:{adminguide_link}#passkeys_server_administration_guide[Passkeys] chapter in the {adminguide_name}. In the future major version, we plan to remove the feature
+as well as the *Passkeys Conditional UI Authenticator* as already announced.
+
 // ------------------------ Removed features ------------------------ //
 == Removed features
 
--- a/services/src/main/java/org/keycloak/authentication/authenticators/browser/PasskeysConditionalUIAuthenticatorFactory.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/browser/PasskeysConditionalUIAuthenticatorFactory.java
@@ -56,7 +56,7 @@ public class PasskeysConditionalUIAuthenticatorFactory extends WebAuthnPasswordl
 
     @Override
     public boolean isSupported(Config.Scope config) {
-        return Profile.isFeatureEnabled(Profile.Feature.PASSKEYS);
+        return Profile.isFeatureEnabled(Profile.Feature.PASSKEYS_CONDITIONAL_UI_AUTHENTICATOR);
     }
 
     @Override
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/webauthn/passwordless/PasskeysConditionalUITest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/webauthn/passwordless/PasskeysConditionalUITest.java
@@ -45,6 +45,7 @@ import org.openqa.selenium.firefox.FirefoxDriver;
  * @author rmartinc
  */
 @EnableFeature(value = Profile.Feature.PASSKEYS, skipRestart = true)
+@EnableFeature(value = Profile.Feature.PASSKEYS_CONDITIONAL_UI_AUTHENTICATOR, skipRestart = true)
 @IgnoreBrowserDriver(FirefoxDriver.class) // See https://github.com/keycloak/keycloak/issues/10368
 public class PasskeysConditionalUITest extends AbstractWebAuthnVirtualTest {
 
--- a/testsuite/integration-arquillian/tests/base/src/test/resources/META-INF/keycloak-server.json
+++ b/testsuite/integration-arquillian/tests/base/src/test/resources/META-INF/keycloak-server.json
@@ -167,6 +167,12 @@
         }
     },
 
+    "datastore": {
+        "legacy": {
+            "allowMigrateExistingDatabaseToSnapshot": "${keycloak.datastore.allowMigrateExistingDatabaseToSnapshot:false}"
+        }
+    },
+
     "realmCache": {
         "default" : {
             "enabled": "${keycloak.realmCache.enabled:true}"
--- a/testsuite/utils/src/main/resources/META-INF/keycloak-server.json
+++ b/testsuite/utils/src/main/resources/META-INF/keycloak-server.json
@@ -96,6 +96,12 @@
         }
     },
 
+    "datastore": {
+        "legacy": {
+            "allowMigrateExistingDatabaseToSnapshot": "${keycloak.datastore.allowMigrateExistingDatabaseToSnapshot:false}"
+        }
+    },
+
     "realmCache": {
         "default" : {
             "enabled": "${keycloak.realmCache.enabled:true}"


graziang:
@mposolda for me it's a good idea, it's better not to rush into removing things related to passkeys since we're making a lot of changes around it. I see there is a test that fails in `FeatureCompatibilityMetadataProviderTest` probably related to the new feature

rmartinc (approved):
Maybe we are being extra-cautious here but I don't see any problem with it. And it's already done, so for me it's OK. Fix the tests and that's all.

mposolda:
@graziang @rmartinc Thanks! Test should be fixed now.

graziang (approved):
@mposolda thanks!

mabartos:
@mposolda @rmartinc @graziang Hey guys!

Sorry for entering your discussion related to your area, but I'd like to share some concerns/considerations :) 

We all can agree that the authenticator cannot be removed for 26.4 (ok for 27) - because of the prod reasons and not creating any breaking change. A breaking change? Yes, that is the issue I see here. The entry in the documentation looks even suspicious as it's in the category "Deprecated feature", but the title is "Deprecated Passkeys Conditional UI Authenticator **requires a feature**". 

As the feature is hidden behind the deprecated flag, the feature is disabled by default, so when users/customers have the authenticator included in their flows, they don't simply start the server - config change is needed from their side as they need to turn on the feature. I know it might be a small change that they need to do, but still a breaking change. So, IMHO the entry would need to be added to the 'Breaking change' section of the migration guide.

I'm afraid that we are violating the agreement for the breaking changes for minor releases:
```
Breaking changes are identified as requiring changes from existing users to their configurations.
In minor or patch releases we will only do breaking changes to fix bugs.
```
-----------------------

I assume that the intention of this issue is to prevent having `PasskeysConditionalUIAuthenticator` and the passkeys support in `UsernamePasswordForm` (not sure why we don't support conditional webauthn/passkeys for username form, but it's out of the scope) simultaneously in the same flow. 

I know Passkeys feature is(was) preview, so we are not strictly required to keep things unchanged, but it creates some sort of gray area. It'd be good to not break users'/customers' deployment and not create some unexpected behavioral changes.

As the feature is preview, and it was already stated in the docs for 26.3 that the authenticator will be removed, I'm kinda ok to hide this behind the new feature as there might not be any better solution at this moment. (Mentioned below a possible approach, but if it's already in 26.3, there might probably be even some other confusion...)

------------
But it'd also be good to mention that as the webauthn conditional UI support was added to the UN+PW form for 26.3, there has been some behavioral change of the authenticator. Users/customers might have set some special flows with special requirements, and they might not notice that the UN+PW form behaves differently and maybe violates their policies (I don't know if there were some complaints after the 26.3 release). 

I expected to have some authenticator config 'Allow passkeys' (with some meaningful tooltip) for the UN+PW form that would turn on/off the webauthn conditional UI feature for the forms. For the new realms and passkeys feature enabled, the 'Allow passkeys' would be enabled. For existing realms and even the passkeys feature enabled, keeping that disabled to not create some unwanted behavioral changes. But I'm not sure how strict you are with the authn flows migration between versions (IMHO there should not be any behavioral change for the existing flows between migrations). 

------------
It'd also be good to think about improving the deprecation of existing authenticators, as over the time, there will be some other deprecation needed for authenticators for sure. It means that it'd be good to see it in the admin console and possibly have some log printed to the CLI console. 

----------------------------

These considerations are rather for future development and you have more expertise in this area, so you can have a different opinion.

mposolda:
@mabartos Thanks for the feedback!

I see your point and agree that this is kind of gray area. Also I don't see perfect solution here... :-/

Regarding the `breaking changes` section VS `deprecated changes` section: I've looked at similar case for the instagram social broker: https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-for-removal-the-instagram-identity-broker . This social provider was supported before Keycloak 26.3, but then it was deprecated in 26.3 and disabled by default and enabled just if the deprecated feature `instagram` is explicitly enabled. So this is kind of similar change. However for instagram case, I see it even much worse as it was "fully supported" feature, when passkeys are still "preview" feature. 

Considering that instagram is mentioned only in the `deprecated features`, I've added this one to `deprecated features` as well. If we consider this a breaking change, I think the same should be done for "instagram" as well and should be moved to "Breaking changes" section? It is late for `instagram` as 26.3 is already released, but we can do this for product docs for 26.4...

Other options I am seeing it:

1) Don't do anything now and just remove `PasskeysConditionalUIAuthenticator` in the 27. I see this option worse than the approach in the current draft PR. As in 26.4, the passkeys will become supported and hence people might still keep using `PasskeysConditionalUIAuthenticator` in their flows and even add it to the new flows. The approach with deprecated feature make administrators aware that they should not be using `PasskeysConditionalUIAuthenticator` anymore and they should do some updates in their flows though. As right now, they are aware just from the small note in the `Deprecated features` of the Keycloak 26.3 (The deprecation of java class `PasskeysConditionalUIAuthenticator` does not help at all as Keycloak administrators/users are not looking at java sources).

2) Don't introduce the feature, but logs the warning in the `PasskeysConditionalUIAuthenticator.authenticate` like `Deprecated authenticator used. Please don't use it and look at passkeys documentation on how to update your authentication flows` . Also having the `Passkeys authenticator` deprecated in the admin console to make new admins aware to not use. See screenshot (but hopefully we can come with something better UXD , see below) : 
 
 Screenshot from 2025-08-01 10-18-39

Better than previous approach, but not ideal IMO as administrators may notice deprecation after they already upgrade their production environment to 26.4. But maybe it is OK as a compromise as it does not require administrators anything in theory to migrate to 26.4? What do you think?

3) Approach in the current PR

4) Any better idea?

My preference order is 3, 2, 1. The approach (2) is acceptable from my point of view if you prefer it. I can update PR if you prefer this way and then we can possibly also have a follow-up to do this with better UXD in mind (See below).

@rmartinc @graziang @mabartos What do you think?

Some more inline.

> 
> I assume that the intention of this issue is to prevent having `PasskeysConditionalUIAuthenticator` and the passkeys support in `UsernamePasswordForm` (not sure why we don't support conditional webauthn/passkeys for username form, but it's out of the scope) simultaneously in the same flow.

We are enabling passkeys in the `UsernameForm` as well. That is also part of the reason why `PasskeysConditionalUIAuthenticator` is useless now.


 
> But it'd also be good to mention that as the webauthn conditional UI support was added to the UN+PW form for 26.3, there has been some behavioral change of the authenticator. Users/customers might have set some special flows with special requirements, and they might not notice that the UN+PW form behaves differently and maybe violates their policies (I don't know if there were some complaints after the 26.3 release).
> 
>
> I expected to have some authenticator config 'Allow passkeys' (with some meaningful tooltip) for the UN+PW form that would turn on/off the webauthn conditional UI feature for the forms. For the new realms and passkeys feature enabled, the 'Allow passkeys' would be enabled. For existing realms and even the passkeys feature enabled, keeping that disabled to not create some unwanted behavioral changes. But I'm not sure how strict you are with the authn flows migration between versions (IMHO there should not be any behavioral change for the existing flows between migrations).

There is a flag `Enable passkeys` in the WebAuthn passwordless policy. Even if `passkeys` feature is enabled, the `UsernamePasswordForm` and `UsernameForm` behaves the same way unless this flag is explicitly enabled for the realm. For 26.4, the flag is still disabled even for new realms by default. AFAIK We did not see any complaints or bugs reported after 26.3 regarding username/password authentication broken due the passkeys addition to the forms.

> 
> It'd also be good to think about improving the deprecation of existing authenticators, as over the time, there will be some other deprecation needed for authenticators for sure. It means that it'd be good to see it in the admin console and possibly have some log printed to the CLI console.

+1

This is perhaps related to the approach (2) I've mentioned above. We can likely have some flag on the `AuthenticationFactory` like `isDeprecated()` . Also we can automatically log warning when `authenticate()` of some deprecated authenticator is used. If yes, we can do something nice in the admin console UI to aware admins that authenticator is deprecated. However I am not sure how to do this in a good UXD way, but maybe we can figure something...

> 
> These considerations are rather for future development and you have more expertise in this area, so you can have a different opinion.

rmartinc:
IMHO passkeys is a preview feature so we can do this. The `passkeys-authenticator` is just a toy to play with webauthn conditional UI and nobody is using this in production (because is mainly useless, only for testing). I would not go to supported with this authenticator enabled by default. I prefer 3 (this PR) or completely removing the authenticator (which is even more drastic).

graziang:
@mabartos  I understand your point, and I think this is an unclear area. My question is: should changes to preview features that require configuration changes be considered breaking changes?

Since passkey is a preview feature and keycloak documentation states that *"These features may change or be removed in a future release,"* without specifying the type of release or the nature of the changes, and as @rmartinc also said that Passkeys Conditional UI Authenticator is just a toy to play with WebAuthn conditional UI, IMHO, the approach in this PR is acceptable and not even that drastic.

But also agree that we are in a gray area, this might be a good opportunity to clarify things and define a procedure for how to deprecate and remove things during releases.

mabartos:
@mposolda @rmartinc @graziang Thanks all for the responses!

> mabartos - As the feature is preview, and it was already stated in the docs for 26.3 that the authenticator will be removed, I'm kinda ok to hide this behind the new feature as there might not be any better solution at this moment. 

As I've mentioned, the (3) should be ok for this case as passkeys feature is/was preview and I'm not sure how to handle it better. 

> mposolda - However for instagram case, I see it even much worse as it was "fully supported" feature, when passkeys are still "preview" feature.

Yes, it is a worse case, and based on that, it's a good time to discuss what could be the best approach for such cases (thanks for the email to the team!) to have consistent behavior even on our side to treat deprecations properly :)) 

As mentioned in the (2), we can definitely do some improvements for deprecating authenticators. We have so many (or it might be so many) built-in authenticators that sometimes do the same job with some little additional tweaks (great example is IdpUsernamePasswordForm). It'd be good opportunity to think about the deprecation process to make life easier in the future as we could get rid of some default authenticators.

Based on the prod nature, and as @mposolda mentioned, we're more tied to provide some deprecation period for the prod side so the complete removal for supported authenticators should be decided carefully. 

------------------------------

### Suggested approach

1. Go on with this PR (as removing a piece of preview feature)
2. Improve the deprecation process for authenticators (good visibility in UI, logs, docs?)
3. Do not hide authenticator deprecation behind the feature flag (not considering preview/experimental features) but rather keep it enabled for some time and apply the approach (2.) on that authenticator.

@mposolda @rmartinc @graziang WDYT?

mposolda:
@rmartinc @graziang @mabartos Thanks! Will merge and create follow-up PR for improving support of deprecated authenticators.

PR #41575, opened 2025-07-31
Add a securing-apps guide with the specifications implemented by keycloak

### LINKED ISSUE #41176: Document supported OIDC/OAuth2 standards
### Description

This issue is about list supported OIDC/OAuth2 related specifications and RFCs into our documentation. Maybe also other standards (like SAML related specs etc).

Does it makes sense to introduce new guide under the "Securing apps" section https://www.keycloak.org/guides#securing-apps ? We have already this guide https://www.keycloak.org/securing-apps/oidc-layers, which is related, but does not provide full list of supported specs and RFCs.

Just a note that there is presentation from @tnorimat  (lastly done on Keycloak devDay AFAIK), which covers some details about supported RFCs and all the supported Keycloak specifications and certifications. Details here: https://keycloak-day.dev/assets/files/Norimatsu_KeycloakDevDay2025_Darmstadt.pdf . Likely best and most up-to-date list we have IMO. Maybe just something to transfer to our docs? Can be also useful to somehow mentioned what are the specs we are certified with.

However we should doublecheck the latest release notes (maybe from 26.0) and include the specs, which were implemented recently as it seems those are not yet included in the Takashi's presentation (For example https://openid.net/specs/openid-connect-prompt-create-1_0.html ).

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #41176

The list of specifications. I have tried to follow the [presentation](https://keycloak-day.dev/assets/files/Norimatsu_KeycloakDevDay2025_Darmstadt.pdf) and this [page](https://github.com/keycloak/keycloak-community/tree/main/specifications). But it's a bit of a hell (too many rows :smile:).

* Please @tnorimat check the certified versions to see if I'm not adding it to the wrong spec.
* @keycloak/core-iam Please check the UMA section because I have just copied from the preso and don't know if it's OK or complete.
* Do we separate OID4VCI in a separated section? I did it before but then I just added it to OIDC because it's under the OIDC umbrella.
* For FIPS probably we can say FIPS 140-3 now, because I think that BC, NSS (JDK) and RHEL 9 is FIPS 140-3 certified now. But that also means changing the FIPS 140-2 guide.

Probably there will be missing specs, but we can add them one by one when detected.

### DIFF
--- a/docs/guides/securing-apps/specifications.adoc
+++ b/docs/guides/securing-apps/specifications.adoc
@@ -1,0 +1,117 @@
+<#import "/templates/guide.adoc" as tmpl>
+<#import "/templates/links.adoc" as links>
+
+<@tmpl.guide
+title="Specifications implemented by {project_name}"
+priority=130
+summary="List of specifications and standards implemented by {project_name}.">
+
+This {section} presents a list of specifications and standards that {project_name} currently implements. The standards are separated in different sections and, in each one, a table is shown with the following four columns:
+
+* *Specification*: The standard or specification that {project_name} implements.
+* *Status*: The current status of the implementation inside {project_name} (supported, preview, experimental,...). See <@links.server id="features"/> for more information.
+* *Conformity*: Assurance of conformity of the implementation.
+** *Certified (version)*: The specification provides conformance tests that {project_name} executes periodically and for each new version. The version in brackets is the last version of {project_name} certified by the authority.
+** *Passed*: There are conformance tests provided by the authority that {project_name} passes, but no version is certified yet.
+** *Partial*: There are conformance tests but {project_name} is not yet fully passing them.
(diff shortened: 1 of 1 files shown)


tnorimat:
@rmartinc Hello, yes I will check the PR.

tnorimat (approved):
LGTM.

docs/guides/securing-apps/specifications.adoc
```diff
@@ -0,0 +1,117 @@
+<#import "/templates/guide.adoc" as tmpl>
+<#import "/templates/links.adoc" as links>
+
+<@tmpl.guide
+title="Specifications implemented by {project_name}"
```
ahus1: We try to avoid adding the project name to each headline, as it leads to long names in upstream, and even longer names in downstream. 

```suggestion
title="Specifications implemented"
```

ahus1 (approved):
Great to see this list in the docs! Thank you for the write-up and the review.

PR #41607, opened 2025-08-01
Add information about sessions to the MDC

ahus1:
Closes #41208

### DIFF
--- a/js/pnpm-lock.yaml
+++ b/js/pnpm-lock.yaml
@@ -1072,6 +1072,10 @@ packages:
     resolution: {integrity: sha512-gtF186CXhIl1p4pJNGZw8Yc6RlshoePRvE0X91oPGb3vZ8pM3qOS9W9NGPat9LziaBV7XrJWGylNQXkGcnM3IQ==}
     engines: {node: ^18.18.0 || ^20.9.0 || >=21.1.0}
 
+  '@eslint/js@9.31.0':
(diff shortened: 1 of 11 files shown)


ahus1:
@eicki - I've added some more information to the context, please provide a review when you have the time. 

I also change "realm" to "realmName" to align it with its use in tracing, and to differentiate it from the realmId.

Thanks!

js/pnpm-lock.yaml
```diff
@@ -1072,10 +1072,6 @@ packages:
     resolution: {integrity: sha512-gtF186CXhIl1p4pJNGZw8Yc6RlshoePRvE0X91oPGb3vZ8pM3qOS9W9NGPat9LziaBV7XrJWGylNQXkGcnM3IQ==}
     engines: {node: ^18.18.0 || ^20.9.0 || >=21.1.0}
 
-  '@eslint/js@9.31.0':
```
mabartos: Is it an intended change? I believe it's not

mabartos (commented):
@ahus1 Just one comment. Otherwise, LGTM.

js/pnpm-lock.yaml
ahus1: Thanks, reverted.

mabartos (approved):

pedroigor (approved):

PR #41614, opened 2025-08-01
Keeping the same email clears UPDATE_EMAIL required action

### LINKED ISSUE #41174: Keeping the same email clears UPDATE_EMAIL required action
### Before reporting an issue

### Area

user-profile

### Describe the bug

Keeping the same email clears UPDATE_EMAIL required action. Required action suggests that email should be actually changed. 

### Version

main

### Regression

### Expected behavior

Keeping the same email should not clear UPDATE_EMAIL required action.

### Actual behavior

Keeping the same email clears UPDATE_EMAIL required action.

### How to Reproduce?

1. assign UPDATE_EMAIL required action to a user in admin console
2. log in as the user and keep the same email in the update email page, then hit submit
3. you are logged in and UPDATE_EMAIL required action is removed from the user

### Anything else?

_No response_

martin-kanis:
Closes #41174

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/requiredactions/UpdateEmail.java
+++ b/services/src/main/java/org/keycloak/authentication/requiredactions/UpdateEmail.java
@@ -55,6 +55,7 @@ import org.keycloak.provider.EnvironmentDependentProviderFactory;
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.provider.ProviderConfigurationBuilder;
 import org.keycloak.services.Urls;
+import org.keycloak.services.messages.Messages;
 import org.keycloak.services.validation.Validation;
 import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.userprofile.EventAuditingAttributeChangeListener;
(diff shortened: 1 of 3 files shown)


ahus1:
@martin-kanis - please add a note to the upgrading guide, as this changed behavior can break workflows of existing users. 

Looking at the change as such, my bank regularly asks me if my email address (or phone number) have changed. Usually they have not, and I just confirm that they haven't changed. 

So having the option to confirm that my email address hasn't changed is actually part of the feature. 

So I'd argue to keep the behavior as is, or to make it configurable on the required action configuration. 

If it is an application initiated action and not a required action, I agree that it should only proceed if the email address actually changed. 

pedroigor:
@martin-kanis I think the current behavior is correct, where the user decides if the email has changed or not. Is this change backed by any use case?

In any case, I agree with @ahus1 that this is a breaking change, and as such, we need to discuss more about how to solve the use case you are trying to address. For instance, I can think about a client deciding to force a user to update the email to a different value (via AIA) because, for some reason, the email is no longer valid or accepted by an application. But still, not sure if we want to introduce this behavior now ...

martin-kanis:
OK, it seems, I interpreted "Update email required action" too strict. So let's keep the current behaviour: "Review your email and keep it the same if it is correct"

PR #41615, opened 2025-08-01
Ability to display 'authenticator provider' of the WebAuthn credential

### LINKED ISSUE #41613: Ability to display 'authenticator provider' of the WebAuthn credential
### Description

I've did some testing with webauthn.io and seeing that once user registers webauthn security-key on that page, they are able to display also the "Provider name" of the WebAuthn security key.

In Keycloak, we have right now the "AAGuid" of the WebAuthn credential, however I was wondering if we can display also the "Provider name" to the administrator similarly like webauthn.io is doing, which is much nicer than just AAGuid. We can eventually also to the user in the account console (A bit related issue is #10063 ).

There is an official mapping between AAGuid and authenticator providers available online for example here 
https://github.com/passkeydeveloper/passkey-authenticator-aaguids/blob/main/aaguid.json . The webauthn.io has their own mapping with even more AAGuids added https://github.com/duo-labs/webauthn.io/blob/master/_app/homepage/services/metadata.py .

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mposolda:
closes #41613

PR allows to display in the admin console the additional metadata, which are not stored together with credential in the DB, but are retrieved from the other metadata. It is used only for WebAuthn to display `authenticator provider` of the WebAuthn credential, which is likely much more useful than `AAGuid` . See screenshot:

Screenshot from 2025-08-01 16-06-20

We can eventually also display in the account console too (Related issue is #10063 ).

The mapping between AAGuid-to-provider-name is currently hardcoded in the file `keycloak-webauthn-metadata.json` inside module `services` . I retrieved the data from the webauthn.io codebase: https://github.com/duo-labs/webauthn.io/blob/master/_app/homepage/services/metadata.py .

The alternative approach might be to download this "online" from this file https://github.com/passkeydeveloper/passkey-authenticator-aaguids/blob/main/aaguid.json . That file is likely maintained regularly, but it does not have some AAGuids provided in the webauthn.io . Also some Keycloak deployments may not have internet access.


rmartinc (commented):
Thanks @mposolda!

For me it's OK and probably (if we decide to include it) it would be nice to see this also in the account console (with #10063), as displaying the authenticator model is useful for the final user too (something like `{{0}} Tansport media: {{1}}`.

Note that we will need to update the json file from time to time.

The only comment it's that IMHO we don't need the new `getCredentialForPresentationFromModel` method and `WebAuthnCredentialPresentationData` class. I would just add the model to the `WebAuthnCredentialData` from the service when converting from the model. But maybe you are trying to make this more generic for future extensions in other credentials. Just a comment, this can go directly like it is now.

mposolda:
@rmartinc Thanks for the feedback!

> For me it's OK and probably (if we decide to include it) it would be nice to see this also in the account console (with https://github.com/keycloak/keycloak/issues/10063), as displaying the authenticator model is useful for the final user too (something like {{0}} Tansport media: {{1}}.

+1 that it is useful to display this also in the account console. I've added support for this. Did also some minor changes in the account UI to be able to display various things (like "Authenticator Provider" and "Transport media")  on multiple lines within the "credential info message". See the screenshot.

Screenshot from 2025-08-13 11-07-36

> Note that we will need to update the json file from time to time.

+1

> The only comment it's that IMHO we don't need the new getCredentialForPresentationFromModel method and WebAuthnCredentialPresentationData class. I would just add the model to the WebAuthnCredentialData from the service when converting from the model. But maybe you are trying to make this more generic for future extensions in other credentials. Just a comment, this can go directly like it is now.

Yes, I prefer to keep this way as IMHO it can be useful for the future extensions as well as you pointed. Also the advantage is, that there is no need to save "Authenticator provider" together with credential and hence it can be always retrieved in the "real time" based on the current metadata. Which can be useful when the metadata changes or also when displaying this for the already created webauthn credentials from the previous versions.

rmartinc (approved):
Thanks @mposolda! LGTM!

thomasdarimont:
The `keycloak-webauthn-metadata.json` seems to contain entries with product names and version numbers which indicates that there might be new product versions added regularly. If I'm not mistaken, then the current logic in `FileUtil. getJsonFileFromClasspathOrConfFolder(..)` seems to attempt to load the `keycloak-webauthn-metadata.json` from classpath (the keycloak-services.jar) first and if it is missing falls back to the config folder lookup. 

I wonder if it would make more sense to support an override mechanism like we do for the OIDC metadata configuration file here.
This would allow us to ship an "updated" version of the `keycloak-webauthn-metadata.json` with every new keycloak release, but users could "update" / "enhance" the aaguid mappings with a file like `conf/keycloak-webauthn-metadata-override.json`. If we detect an override at startup, we could log a message.

mposolda:
@thomasdarimont Yes, I can imagine that support for "overrides" might be useful as you suggested. But maybe it is ok to add this as a follow-up issue if there is a need for this rather than adding this right in this PR? WDYT?

thomasdarimont:
I'm fine with adding this in a followup PR. Just wanted to make sure this is taken into consideration :)

js/apps/account-ui/src/account-security/SigningIn.tsx
```diff
@@ -134,16 +137,23 @@ export const SigningIn = () => {
...
-                <InfoAltIcon />{" "}
-                {t(
-                  credMetadata.infoMessage.key,
-                  credMetadata.infoMessage.parameters?.reduce(
-                    (acc, val, idx) => ({ ...acc, [idx]: val }),
-                    {},
-                  ),
-                )}
-              </p>
+              <table>
```
jonkoops: For displaying key value pairs, a more appropriate choice might be a [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl). PatternFly has a [specific component](https://www.patternfly.org/components/description-list) that can be used for this. Also take into account that the dialog height should likely be limited, and the overflow should scroll.

ssilvert:
> We can eventually also display in the account console too (Related issue is #10063 ).
> 

@mposolda 
I'm a little confused as I only see code for account console and not admin console.

Also, if you do end up with code that is common to both consoles, remember that we have a "shared" library: https://github.com/keycloak/keycloak/tree/main/js/libs/ui-shared

js/apps/account-ui/src/account-security/SigningIn.tsx
mposolda: @jonkoops Thanks for the feedback!

I've added additional commit (just for the easier review, it can be squashed upon merge), which adds support for `infoProperties` as a key/value pair in credentialMetadata (until now we have support only for single "info" message in the API, but displaying something more complex with that single message and then parsing in the UI layer looks like a workaround). Support for single info message (`CredentialMetadataRepresentation.infoMessage` ) is still kept due the backwards compatibility and it makes sense for some credentials like recovery codes.

In the account console, it looks like this right now. WDYT?

Screenshot from 2025-08-19 09-23-19

mposolda:
> > We can eventually also display in the account console too (Related issue is #10063 ).
> 
> @mposolda I'm a little confused as I only see code for account console and not admin console.
> 
> Also, if you do end up with code that is common to both consoles, remember that we have a "shared" library: https://github.com/keycloak/keycloak/tree/main/js/libs/ui-shared

@ssilvert Admin console already has support for displaying the credential metadata with multiple key/value pairs. Any property added to credentialData is automatically displayed in the admin console. So there is no need to do any changes in admin console UI as it already works OOTB. See screenshot how it looks like to make sure that we are talking about the same thing:

Screenshot from 2025-08-19 09-32-17

Account console is in fact displaying only the "subset" of those properties, which might be useful for the end user.

js/apps/account-ui/src/account-security/SigningIn.tsx
jonkoops: Looks better, but let's not use tables for layout. Tables are for displaying structured data, and using these HTML elements has semantic meaning, which impacts accessibility. Instead of a table, PatternFly offers [various layouts](https://www.patternfly.org/layouts/about-layouts), perhaps a [split](https://www.patternfly.org/layouts/split) layout could work here to position the icon.

mposolda: @jonkoops Thanks for the suggestion! I've used split and also it looks better IMO. WDYT?

Screenshot from 2025-08-19 12-24-04

ssilvert: > Looks better, but let's not use tables for layout. Tables are for displaying structured data, and using these HTML elements has semantic meaning, which impacts accessibility. Instead of a table, PatternFly offers [various layouts](https://www.patternfly.org/layouts/about-layouts), perhaps a [split](https://www.patternfly.org/layouts/split) layout could work here to position the icon.

@mposolda I agree with Jon here.  Let's at least look into using a standard PF layout.  If admin console is also using table we should fix that as well.

ssilvert (changes requested):
See my reply to Jon's comment.

ssilvert (dismissed):
UI LGTM

js/apps/account-ui/src/account-security/SigningIn.tsx
```diff
@@ -145,6 +150,31 @@ export const SigningIn = () => {
                 )}
               </p>
             )}
+            {credMetadata.infoProperties && (
+              <Split className="pf-v5-u-mb-lg">
+                <SplitItem>
+                  <InfoAltIcon />{" "}
```
jonkoops: Do we still need this text-based white-space? The margin on the other item should probably take care of this already.
```suggestion
                  <InfoAltIcon />
```

js/apps/account-ui/src/account-security/SigningIn.tsx
```diff
@@ -145,6 +150,31 @@ export const SigningIn = () => {
...
+                </SplitItem>
+                <SplitItem isFilled className="pf-v5-u-ml-xs">
+                  <DescriptionList
+                    isHorizontal
+                    aria-label="Horizontal"
```
jonkoops: If we do use a label here, it should be a descriptive label of the contents, e.g. "Provider properties" or something along those lines. However, considering the description list itself already contains a title of the property, and the value, it seems to me it is descriptive enough for screen readers as-is, so it can be safely removed.

```suggestion
```

js/apps/account-ui/src/account-security/SigningIn.tsx
jonkoops: Looks good, consider this resolved :+1:

js/apps/account-ui/src/account-security/SigningIn.tsx
mposolda: Yes, true. We do not need this text-based white-space.

jonkoops (approved):

PR #41627, opened 2025-08-02
feat(operator): add support for custom labels in ingress

### LINKED ISSUE #37363: Allow custom labels on Operator Ingress
### Description

In addition to custom annotations under `.spec.ingress` in the Keycloak CR, we should also add a field to specify additional custom labels.

### Discussion

_No response_

### Motivation

Custom labels might be required by some Ingress Controllers.

### Details

_No response_

KyriosGN0:
This PR adds labels map to the ingress Spec of the operator 
fixes #37363


operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/spec/IngressSpec.java
```diff
@@ -38,6 +38,10 @@ public class IngressSpec {
     @JsonPropertyDescription("Additional annotations to be appended to the Ingress object")
     Map<String, String> annotations;
 
+    @JsonProperty("labels")
+    @JsonPropertyDescription("Additional labels to be appended to the Ingress object")
+    Map<String, String> labels;
```
shawkins: Please default this and annotation to a new LinkedHashMap.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakIngressDependentResource.java
```diff
@@ -106,15 +107,17 @@ public Ingress desired(Keycloak keycloak, Context<Keycloak> context) {
...
         optionalSpec.map(IngressSpec::getAnnotations).ifPresent(annotations::putAll);
+        optionalSpec.map(IngressSpec::getLabels).ifPresent(labels::putAll);
+
 
         Ingress ingress = new IngressBuilder()
                 .withNewMetadata()
                     .withName(getName(keycloak))
                     .withNamespace(keycloak.getMetadata().getNamespace())
                     .addToLabels(Utils.allInstanceLabels(keycloak))
+                    .addToLabels(optionalSpec.map(IngressSpec::getLabels).orElse(null))
```
shawkins, line 118: Either this should be addToLabels(labels), or you should remove the use of the intermediate map.

shawkins (changes requested):
This looks good, just a couple of thoughts.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/spec/IngressSpec.java
KyriosGN0: what should be the default? the keycloak operator labels?

shawkins: No, just an empty linkedhashmap - that's generally the paradigm we follow because it cuts down on null checking.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakIngressDependentResource.java
shawkins: And one last concern - the addToLabels calls should be flipped. I don't think we want to give the user the ability to overwrite a built-in label via this mechanism.

KyriosGN0:
hey @shawkins, i have addressed your comments, Thanks!

shawkins (approved):
LGTM, thanks @KyriosGN0

mabartos (approved):
@KyriosGN0 Thanks! Nice work!

PR #41642, opened 2025-08-04
Ensure cache configuration has correct number of owners

pruivo:
Closes #41558

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
+++ b/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
@@ -34,15 +34,20 @@ import org.infinispan.transaction.TransactionMode;
 import org.infinispan.transaction.lookup.EmbeddedTransactionManagerLookup;
 import org.jboss.logging.Logger;
 import org.keycloak.Config;
-import org.keycloak.connections.infinispan.InfinispanConnectionProvider;
 
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.ACTION_TOKEN_CACHE;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.ALL_CACHES_NAME;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.AUTHENTICATION_SESSIONS_CACHE_NAME;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.AUTHORIZATION_CACHE_NAME;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.AUTHORIZATION_REVISIONS_CACHE_DEFAULT_MAX;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.AUTHORIZATION_REVISIONS_CACHE_NAME;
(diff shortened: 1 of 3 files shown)


model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
```diff
@@ -245,12 +250,40 @@ public static void configureSessionsCachesForVolatileSessions(ConfigurationBuild
...
+                AUTHENTICATION_SESSIONS_CACHE_NAME,
+                ACTION_TOKEN_CACHE)) {
+            var builder = holder.getNamedConfigurationBuilders().get(name);
+            if (builder == null) {
+                throw cacheNotFound(name);
+            }
+            var hashConfig = builder.clustering().hash();
+            var owners = hashConfig.attributes().attribute(HashConfiguration.NUM_OWNERS).get();
+            if (owners < 2) {
+                logger.infof("Setting num_owners=2 (configured value is %s) for cache '%s' to prevent data loss.", owners, name);
```
ryanemerson, line 281: I think we agreed that this should be a warning (detailed in the GH issue) so that the user is more aware that their configuration is not being applied as provided.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
```diff
@@ -245,12 +250,40 @@ public static void configureSessionsCachesForVolatileSessions(ConfigurationBuild
...
                 builder.clustering().hash().numOwners(1);
             }
         }
     }
 
+    /**
+     * Configures the caches "actionToken", "authenticationSessions", and "loginFailures" with the minimum number of
+     * owners to prevent data loss in a single instance crash.
+     * <p>
+     * The data in those caches only exist in memory they must have more than 1 owner configured.
```
ryanemerson: Nitpick: "The data in those caches only exist in memory, therefore they must have more than one owner configured."

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
pruivo: I know, but the other methods that change the number of owners for session caches are logged as INFO, not WARN. 
Should I still change this to warn?

ryanemerson: Good point. Given that we explicitly changed the behaviour from WARN to INFO in 26.3, it doesn't feel right to change this again. Let's be consistent and use INFO everywhere.

ryanemerson (approved):

abstractj (approved):

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
ahus1: A late comment from my side: 

We changed those messages from WARN to INFO as we didn't want to force users to change the XML file when they disabled or enabled persistent sessions. 

This might be slightly different as someone is putting (manually) something wrong in the XML file. Still I'm fine with the INFO message here.

PR #41645, opened 2025-08-04
Make passkeys feature supported

### LINKED ISSUE #41556: Switch passkeys to supported
### Description

This issue is about switching "passkeys" feature to supported and update of related tests.

Note that there is separate task for docs for easier review of PRs for the codebase and for the docs: #41557.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #41556

Making passkeys supported. There is a small change in the `getOptionalReferenceCategories` because I detected that only checking the feature always added the `webauthn-passwordless` credential to the realm (in the account console). Passing the session we can check the policy in the realm. I'm adding the documentation changes in another PR because it was another issue.

Issue for documentation: #41557 PR: #41646

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -124,7 +124,7 @@ public class Profile {
 
         ORGANIZATION("Organization support within realms", Type.DEFAULT),
 
-        PASSKEYS("Passkeys", Type.PREVIEW, Feature.WEB_AUTHN),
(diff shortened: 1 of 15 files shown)


mposolda (approved):
@rmartinc Nice, Thanks!

+1 for the change in `getOptionalReferenceCategories` . Accidentally, I also just figured it this morning and was thinking that it could be better to check also passkeys enabled for the realm :)

PR #41647, opened 2025-08-04
enhance: adding the ability to get the root config from a Scope

### LINKED ISSUE #36268: Configuration is not available outside of quarkus modules
### Description

While working on this PR https://github.com/keycloak/keycloak/pull/36049 I encountered an issue where I am not able to find out whether [metrics option](https://github.com/keycloak/keycloak/blob/3839f8e3b5054f478cfc02217ac1a3b95ea580b8/quarkus/config-api/src/main/java/org/keycloak/config/MetricsOptions.java#L5) is enabled or not. I needed it [here](https://github.com/keycloak/keycloak/pull/36049/files#diff-2a98fe27c9ea122a67b10094768afe7167e98fb549752267ebff6c6705076178R64). 

Let me know if this is somehow possible but the only way I found out was to pass this information to correct provider via spi option. 

Moreover, even this was not that simple as I would expect as I cannot create more mappers for single option. Therefore I needed to create another [hidden option](https://github.com/keycloak/keycloak/pull/36049/files#diff-943df70dc594bf7937b6ed2c311c882cdaad85733f9d911706707f225bbfefc5R16) having always [the same value as metrics option](https://github.com/keycloak/keycloak/pull/36049/files#diff-4daf7a59c1febf5c6efe273725353a0f44a022359fbc725916143c96832e0a58R22) and create mapper to the spi option.

Is it possible to do this in a nicer way? Or would it be possible to improve this? 

Possible solutions:
1. Make configuration available outside of the `quarkus` modules - just a food for thought: maybe we can use some kind of provider, or make it part of `KeycloakSession`
2. Make it possible to create more mappers for single option
3. Make it possible to call method [`.to()`](https://github.com/keycloak/keycloak/pull/36049/files#diff-4daf7a59c1febf5c6efe273725353a0f44a022359fbc725916143c96832e0a58R21) more than once. 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
closes: #36268

### DIFF
--- a/core/src/main/java/org/keycloak/Config.java
+++ b/core/src/main/java/org/keycloak/Config.java
@@ -208,6 +208,11 @@ public class Config {
             throw new UnsupportedOperationException("Not implemented");
         }
 
+        @Override
+        public Scope root() {
+            return new SystemPropertiesScope("keycloak.");
+        }
+
     }
 
     /**
@@ -243,5 +248,13 @@ public class Config {
          */
         @Deprecated
         Set<String> getPropertyNames();
+
+        /**
+         * Root {@link Scope} for global options. The key format should match exactly what
+         * is expected to appear in the main configuration file - e.g. metrics-enabled, db, etc.
+         *
+         * @return a {@link Scope} with access to global configuration properties.
+         */
+        Scope root();
     }
 }
--- a/quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigurationTest.java
+++ b/quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigurationTest.java
@@ -61,7 +61,11 @@ public class ConfigurationTest extends AbstractConfigurationTest {
         putEnvVar("KC_SPI_CAMEL_CASE_SCOPE_CAMEL_CASE_PROP", "foobar");
         initConfig();
         String value = Config.scope("camelCaseScope").get("camelCaseProp");
-        assertEquals(value, "foobar");
+        assertEquals("foobar", value);
+
+        // root should be at kc - users are not expected to obtain spi options this way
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/keycloak/Config.java
```diff
@@ -243,5 +248,11 @@ public static interface Scope {
          */
         @Deprecated
         Set<String> getPropertyNames();
+
+        /**
+         * Return a scope with access to global configuration properties.
+         * @return
```
mabartos:
```suggestion
```
nitpick

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigurationTest.java
```diff
@@ -62,6 +62,9 @@ public void testCamelCase() {
         initConfig();
         String value = Config.scope("camelCaseScope").get("camelCaseProp");
         assertEquals(value, "foobar");
+
+        value = Config.scope().root().get("spi-camelCaseScope-camelCaseProp");
+        assertEquals(value, "foobar");
```
pruivo: nitpicking, intellij shows the warning `Arguments to 'assertEquals()' in wrong order`

mabartos (dismissed):
@shawkins LGTM, thanks!

shawkins:
Updated based upon the reviews.

pruivo (dismissed):

mabartos (dismissed):

shawkins:
One more change - I think we need to not do the camel case and -- compensation logic when dealing with the root. It should be documented as expecting the exact configuration key.

mabartos (approved):

pruivo (approved):

abstractj (approved):

PR #41661, opened 2025-08-05
Use Optional instead of getOrDefault for settings in testSMTPConnection

### LINKED ISSUE #41643: Test SMTP connection fails when no port is specified
### Before reporting an issue

### Area

admin/api

### Describe the bug

### Version

26.3.2

### Regression

### Expected behavior

The test should work OK.

### Actual behavior

### How to Reproduce?

### Anything else?

_No response_

rmartinc:
Closes #41643

The `getOrDefault` returns `null` when `null` is explicitly set in the map. And it seems that the console sends `"port":null` in the json. So the PR just changes that method for an `Optional` for the parameters received for the test. We cannot test explicitly the default `null` port because the SMTP server cannot run using privileged 25.

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/RealmAdminResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/RealmAdminResource.java
@@ -119,6 +119,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


pskopek (approved):
LGTM!

jonkoops (approved):

abstractj (approved):

PR #41665, opened 2025-08-05
[admin-api-v2] APIs as CDI beans

mabartos:
## Superseded by: https://github.com/keycloak/keycloak/pull/41848

- CDI support for APIs/endpoints
- Hibernate Validator is executed automatically based on the used annotations
- Fixed tests

### DIFF
--- a/rest/admin-api/src/main/java/org/keycloak/admin/api/AdminRootV2.java
+++ b/rest/admin-api/src/main/java/org/keycloak/admin/api/AdminRootV2.java
@@ -1,0 +1,34 @@
+package org.keycloak.admin.api;
+
+import jakarta.ws.rs.OPTIONS;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.core.Context;
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.services.resources.admin.AdminCorsPreflightService;
+
+@jakarta.ws.rs.ext.Provider
+@Path("admin/api")
+public class AdminRootV2 {
+
+    @Context
+    protected KeycloakSession session;
+
+    @Path("")
+    public AdminApi latestAdminApi() {
+        // we could return the latest Admin API if no version is specified
+        return new DefaultAdminApi(session);
+    }
+
+    @Path("v2")
+    public AdminApi adminApi() {
+        return new DefaultAdminApi(session);
+    }
+
+    @Path("{any:.*}")
+    @OPTIONS
+    @Operation(hidden = true)
+    public Object preFlight() {
+        return new AdminCorsPreflightService();
+    }
+}
--- a/rest/admin-api/src/main/java/org/keycloak/admin/api/AdminRootV2.java
+++ b/rest/admin-api/src/main/java/org/keycloak/admin/api/AdminRootV2.java
@@ -1,28 +1,35 @@
 package org.keycloak.admin.api;
 
+import jakarta.enterprise.context.RequestScoped;
+import jakarta.inject.Inject;
 import jakarta.ws.rs.OPTIONS;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.ext.Provider;
 import org.eclipse.microprofile.openapi.annotations.Operation;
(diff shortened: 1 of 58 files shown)


rest/admin-api/src/main/java/org/keycloak/admin/api/AdminRootV2.java
```diff
@@ -1,28 +1,35 @@
...
 import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.ext.Provider;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.services.resources.admin.AdminCorsPreflightService;
 
-@jakarta.ws.rs.ext.Provider
+@Provider
 @Path("admin/api")
+@RequestScoped
```
stianst, line 15: Don't use @RequestScoped, just do it how we do things elsewhere with SPIs/providers/etc.

stianst: We probably want to consider using SPIs and providers to make the API more pluggable

mabartos: > Don't use @RequestScoped, just do it how we do things elsewhere with SPIs/providers/etc.

@stianst As we're making a new API, it'd be good to be up-to-date with the standard way on how to do APIs in terms of Jakarta EE. It means, that we might leverage CDI for the REST endpoints initialization and use these handy and powerful tools. On the other hand, as Keycloak provides SPI concepts, we might create some sort of hybrid approach on how to mix them together -> users will be able to choose provider they want + they would have a support of CDI in their providers.

Working Draft PR for the hybrid approach: https://github.com/keycloak/keycloak/pull/41848 

> We probably want to consider using SPIs and providers to make the API more pluggable

We already discussed it previously if we want to add SPI support for these API classes and we decided to not to do it as the extendability is expected more on the business logic level (services). However, I expect that some people would like to change even logic behind the REST layer, so if we have such requirements, I'm ok to provide providers/factories for the REST layer. 

Covered as well in: https://github.com/keycloak/keycloak/pull/41848 

@vmuzikar @shawkins @Pepo48 WDYT?

vmuzikar: @mabartos AFAIK, the main reason against CDI is that it's a completely new concept to the Keycloak core code base. And it kinda conflicts with the current concept of `KeycloakSession`.

The question then is, if we introduced CDI, what's next. Do we deprecate `KeycloakSession`, or still support it? Do we migrate rest of the code base to leverage CDI? If not, wouldn't it be inconsistent?

mabartos: > @mabartos AFAIK, the main reason against CDI is that it's a completely new concept to the Keycloak core code base.

@vmuzikar I don't think it should be a problem, is it? 

> And it kinda conflicts with the current concept of KeycloakSession.

Not really. KeycloakSession is basically just another RequestScoped bean (request livespan) presented as RESTEasy Context, so it can be injected to these classes via `@Context` annotation. KeycloakSession stores the current context(actual realm, client, request params, ...), and provides entrypoint to the KC functionality. It means that if we set some context in one CDI-aware API/resource class, it is propagated to another(sub-resource) as well, so no inconsistency happens.

KeycloakSession and CDI beans co-exist together and just improves the overall manageability and reduce boilerplate.  

> The question then is, if we introduced CDI, what's next. Do we deprecate KeycloakSession, or still support it? Do we migrate rest of the code base to leverage CDI? If not, wouldn't it be inconsistent?

There's no need to deprecate KeycloakSession as it'd be used together as well (if you're refering to the `session.getProvider(SomeClass.class)`, there's no CDI support yet). As we're making a new API, IMHO we can provide some new concepts to improve stuff that was bad in APIs v1. If the new API is following the SPI principles, you can even choose what provider will be created by CDI - see the PoC https://github.com/keycloak/keycloak/pull/41848. Right now, we can have it only on the top level (REST layer) for Admin APIs v2, and then we can discuss on how to leverage it for other components, like services, etc.

mabartos:
Superseded by: https://github.com/keycloak/keycloak/pull/41848

PR #41684, opened 2025-08-05
Wait for key generation to finish.

ssilvert:
Fixes #41683

### DIFF
--- a/js/apps/admin-ui/test/clients/saml.ts
+++ b/js/apps/admin-ui/test/clients/saml.ts
@@ -79,7 +79,12 @@ export async function clickOffEncryptionAssertions(page: Page) {
 }
 
 export async function clickGenerate(page: Page) {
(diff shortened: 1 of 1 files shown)


js/apps/admin-ui/test/clients/saml.ts
```diff
@@ -79,7 +79,12 @@ export async function clickOffEncryptionAssertions(page: Page) {
 }
 
 export async function clickGenerate(page: Page) {
+  const responsePromise = page.waitForResponse(
+    (res) => res.url().includes("/generate") && res.status() === 200,
+    { timeout: 10000 },
+  );
```
jonkoops, line 85: Would `waitForResponse("*/generate")` not be enough here? The default timeout of 30s seems sufficient as well.

jonkoops (approved):

PR #41702, opened 2025-08-06
Default to stretched clusters on Kubernetes when possible

### LINKED ISSUE #41666: Default to stretched clusters on Kubernetes when possible
### Description

The Keycloak Operator currently defaults to creating an affinity and anti-affinity strategy that prefers all pods are hosted on the same availability zone but distinct nodes. Instead, the default should prefer that pods are distributed across different zones in order to improve fault-tolerance. Benchmarking shows that the performance impact of a "stretched" multi-zone cluster vs a single zone cluster is negligible. The SRE team believes the HA improvements outweigh the performance impact and will be beneficial for the majority of users.

### Discussion

_No response_

### Motivation

_No response_

### Details

Update to the docs: in `getting-started-scaling-and-tuning.adoc` we currently state which IMHO should be removed completely. 

> This cache is not optimal for clusters that span multiple availability zones, which are also called stretch clusters. For embedded Infinispan cache, work to have all instances in one availability zone. The goal is to avoid unnecessary round-trips in the communication that would amplify in the response times. On Kubernetes, use Pod affinity to enforce this grouping of Pods.

Proposed rules:

```yaml
podAntiAffinity:
  preferredDuringSchedulingIgnoredDuringExecution:
    - weight: 100
      podAffinityTerm:
        labelSelector:
          matchLabels:
            app: keycloak
            app.kubernetes.io/managed-by: keycloak-operator
            app.kubernetes.io/component: server
        topologyKey: topology.kubernetes.io/zone
    - weight: 90
      podAffinityTerm:
        labelSelector:
          matchLabels:
            app: keycloak
            app.kubernetes.io/managed-by: keycloak-operator
            app.kubernetes.io/component: server
        topologyKey: kubernetes.io/hostname
```

ryanemerson:
Closes #41666

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
@@ -415,25 +415,31 @@ public class KeycloakDeploymentDependentResource extends CRUDKubernetesDependent
             }
         }
 
-        // set defaults if nothing was specified by the user
-        // - server pods will have an affinity for the same zone as to avoid stretch clusters
-        // - server pods will have a stronger anti-affinity for the same node
-
         if (!specBuilder.hasAffinity()) {
-            specBuilder.editOrNewAffinity().withNewPodAffinity().addNewPreferredDuringSchedulingIgnoredDuringExecution()
-                    .withWeight(10).withNewPodAffinityTerm().withNewLabelSelector().withMatchLabels(labels)
-                    .endLabelSelector().withTopologyKey(ZONE_KEY).endPodAffinityTerm()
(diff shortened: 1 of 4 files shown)


operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -420,21 +420,29 @@ private void handleScheduling(Keycloak keycloakCR, Map<String, String> labels, P
         // - server pods will have a stronger anti-affinity for the same node
```
pruivo: Please, update the comment above this line to be up-to-date with the code.

ryanemerson: Done

pruivo (approved):

ahus1 (approved):

PR #41705, opened 2025-08-06
[OTel] Provide general options for telemetry settings

### LINKED ISSUE #41263: [OTel] Provide general options for telemetry settings

mabartos:
- Closes #41263
- Base-for https://github.com/keycloak/keycloak/pull/41265
- Base-for https://github.com/keycloak/keycloak/pull/41716

Up-to-date until Dec 8, 2025.

### 1. Root configuration
We can have global options for the whole OTel Collector settings that will be shared to the OTel subcomponents(logs, metrics, traces), as it's a usual case that all 3 components point to the same endpoint. 

However, sometimes it's necessary to emit traces somewhere else than f.e. logs or metrics.
For that case, we can override the basic fundamentals options for every subcomponent. 
After the discussion below, we provide options for Logs and Metrics and hide the OTel name to not include the technical terms everywhere and use it only when applicable. 

As OpenTelemetry is vendor- and tool-agnostic, it has become de facto standard for application observability. However, we don't know how the future will look like and we can be prepared even for replacing OTel in the future. We can have a root configuration with the prefix `telemetry-` that would be parent props for the subcomponents. 

As telemetry represents more the telemetry data (logs, metrics, traces) export, it'd be probably name the parent props like that.

```
telemetry
│
├── [enabled (hidden) | endpoint | protocol]
│
├── metrics-export
│    └── [enabled | endpoint | protocol]
│
└── log-export
     └── [enabled | endpoint | protocol]
```

### 2. Deprecation
The `tracing-service-name`, and `tracing-resource-attributes` options were marked deprecated as they are not directly relate to the Tracing, but to the whole observability/telemetry configuration. 

For that case, we introduce the 'telemetry-service-name', and `telemetry-resource-attributes` to set up the whole observability/telemetry context. 
These options takes precedence over the `tracing-*` options, but users can still use the deprecated ones.

```
Migration:

tracing-service-name --> telemetry-service-name
tracing-resource-attributes --> telemetry-resource-attributes
```

### 3. Operator
We have already tracing options configurable via Keycloak CR, so we might provide even support for the observability/OTel options. It should be fine, the only thing is the deprecation of the options mentioned in the (2.) as we probably don't have a nice approach on how to deprecate fields in CR, right?

### 4. Supported options
First thought was make the OTel/observability/telemetry options preview, but after some consideration, I decided not to do it. We basically use the same behavior as before and as we don't have the support for Metrics and Logs, nothing should affect these options. We basically map to the same options as before (not considering component vs root config), so if we have s decent test coverage for the mapping, we should be fine. 

It also simplify the handling of deprecated options for the Tracing options. IMHO, we can support it right away due to the straightforward nature. 

@keycloak/cloud-native (cc: @keycloak/sre)


pruivo:
I don't like the `otel` prefix in traces, metrics, and logs. 
`tracing-[enabled|endpoint|protocol]` seems expressive enough for me, same for `metrics` and `log` options. 

I reason that you can replace the backend easily; for example, if Otel is replaced by XPTO in the next decade, you only need to remove the `otel-[enabled|endpoint|protocol]` and add the `xpto-*` specific options. The remaining options for traces, metrics, and logs are kept the same. 
Similarly, if you want to replace only one of the systems, you can do so more easily too (for example, using Otel for traces and some other backend for metrics/logs). 

Just my 2 cents :)

mabartos:
@pruivo Thanks for your comment. I got your point. 

The intention behind that is to connect these options together and create a whole OpenTelemetry context. As OpenTelemetry is de facto standard for observability (not a backend), the expectation is that it's gonna live for some time. However, as you mentioned, even despite the OTel adaptation, it can be replaced by something else. We can speculate about the timeline of Keycloak relevance in the future when OTel is out, but... :)) 

We need to take into account that some people do not want to enable the OTel for logs or metrics by default when the `--metrics-enabled=true`. So we need some sort of switch for this, even for the future. Having `log-enabled` is not descriptive enough, and it might be confusing. I'd like to also emphasize that we need to have OTel Metrics and Logs as preview, so we need to have a mechanism for that (option based, or feature based).

We can introduce some sort of `export` properties, which might be a general form of emitting telemetry to the OTel Collector.

### Possible alternative

#### Traces
- Do not introduce any other option for Tracing
- We'd still need to do the deprecation, though

#### Metrics
- Create `metrics-export-enabled`, `metrics-export-endpoint`, `metrics-export-protocol`

#### Logs
- Create `log-export-enabled`, `log-export-endpoint`, `log-export-protocol`

#### General
- We could have the parent OTel options - the more specific options would inherit their values
- We'd emphasize in the new export options above that it uses OTel (description can be changed in the future easily)
- As mentioned, the (3.) deprecation should be done as we cannot have `service-name`, and `resource-attributes` specified for the export options as they map to the single OTel property

@pruivo WDYT?

pruivo:
Yes, "backend" was not the best word, but it was the one that came to mind. At Infinispan, we initially used Smallrye and later updated our code to utilize Micrometer. We plan to switch to Otel at some point in the future. That's why I suggest having something more "abstract" to support future evolution. 

@mabartos It looks good to me, thank you.

ahus1:
As a general theme, we are IMHO avoiding technical terms in options. We have "caching" options, and not "Infinispan" options. OTel is then only an implementation detail which then changes (Smallrye -> Micrometer -> OTel). Still, maybe you have good arguments why this one should be different.

mabartos:
@pruivo @ahus1 @vmuzikar We can move even more forward and trying to get rid of the `otel-` prefix completely. IMHO, I'd say it's not a big deal to have a parent properties named `otel-*` as it was already mentioned that OpenTelemetry becomes kind of a standard for the observability. However, to be future proof, we can have options like (mentioned in the PR description):

```
telemetry
│
├── [enabled (hidden) | endpoint | protocol]
│
├── metrics-export
│    └── [enabled | endpoint | protocol]
│
└── log-export
     └── [enabled | endpoint | protocol]
```

ryanemerson:
> We can move even more forward and trying to get rid of the `otel-` prefix completely.

+1 from me. I like the standard/implementation neutral approach you propose in your comment and others have advocated.

My only slightly concern would be the disconnect between the existing `Feature` which is called `OPENTELEMETRY` and the CLI options. On the other hand, maybe that actually complements the impl neutral CLI options as in the future if/when a new impl comes along it would be a case of users setting the same CLI options and just changing the enabled Feature.

mabartos:
@ryanemerson Thanks for the comment!

> On the other hand, maybe that actually complements the impl neutral CLI options as in the future if/when a new impl comes along it would be a case of users setting the same CLI options and just changing the enabled Feature.

Yep, that's the plan. I'd like to still emphasize (in docs, option description) the tight connection to OpenTelemetry to prevent any confusion in the community asking for OTel support. In that in mind, as you said, it might nicely complement the whole concept.

docs/guides/observability/telemetry.adoc
```diff
@@ -0,0 +1,65 @@
...
+It is a project under the https://www.cncf.io/projects/opentelemetry/[Cloud Native Computing Foundation (CNCF)] as an incubating project.
+Its goal is to provide a unified standard and toolset for instrumenting, generating, and exporting telemetry data from modern applications, making it easier to monitor, debug, and understand complex services.
+
+The goal of {project_name} is to support OpenTelemetry Logs, Metrics, and Traces to create a centralized observability stack without any additional deployment overhead.
+
+== General options
+{project_name} provides support for global settings of telemetry data export.
+It means that you can specify connection parameters to a single https://opentelemetry.io/docs/collector/[OpenTelemetry collector] that will be used for all supported components.
+
+In order to have OpenTelemetry enabled, at least one component needs to be turned on (i.e., see Traces guide below).
```
ryanemerson: I would remove the brackets and link to the guide below:

"...turned on, for example <<traces, Traces>>."

docs/guides/observability/telemetry.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+NOTE: OpenTelemetry can be enabled only when the `opentelemetry` feature is *enabled* (by default).
```
ryanemerson: I'm not sure if this is better or worse :slightly_smiling_face: 

"NOTE: The OpenTelemetry feature is *enabled* by default. Disable the feature in order to disable OpenTelemetry components."

docs/guides/observability/telemetry.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== Endpoint
+You can specify the global endpoint to the OpenTelemetry collector shared across the components by using the `telemetry-endpoint` option.
```
ryanemerson: "You can configure the OpenTelemetry collector's global endpoint that is shared across all components by using ..."

docs/guides/observability/telemetry.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== Service name
+You can specify the global OpenTelemetry service name used for identifying the exporter of the telemetry data.
+The default service name is `keycloak`, specified via the `telemetry-service-name` property, which takes precedence over `service.name` defined in the `telemetry-resource-attributes` property.
+
+You can change the service name via CLI as follows:
+
+<@kc.start parameters="--telemetry-service-name=my-keycloak-iam"/>
+
+=== Protocol
```
ryanemerson, line 40: I'm not sure the "Endpoint",  "Service name" and "Protocol" sections add much value given that we already output this information as part of  the "Relevant Options" section.

quarkus/config-api/src/main/java/org/keycloak/config/TelemetryOptions.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static final Option<Boolean> TELEMETRY_ENABLED = new OptionBuilder<>("telemetry-enabled", Boolean.class)
+            .category(OptionCategory.TELEMETRY)
+            .buildTime(true)
+            .hidden()
+            .build();
+
+    public static final Option<String> TELEMETRY_ENDPOINT = new OptionBuilder<>("telemetry-endpoint", String.class)
+            .category(OptionCategory.TELEMETRY)
+            .description("Telemetry (OpenTelemetry) endpoint to connect to.")
```
ryanemerson: If we're going to include "OpenTelemetry" in the description, I would just have "OpenTelemetry endpoint to connect to".

That being said, does include the impl here negate the idea of having implementation independent CLI options?

quarkus/config-api/src/main/java/org/keycloak/config/TelemetryOptions.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static final Option<String> TELEMETRY_SERVICE_NAME = new OptionBuilder<>("telemetry-service-name", String.class)
+            .category(OptionCategory.TELEMETRY)
+            .description("Telemetry (OpenTelemetry) service name. Takes precedence over 'service.name' defined in the 'telemetry-resource-attributes' property.")
+            .defaultValue("keycloak")
+            .build();
+
+    public static final Option<String> TELEMETRY_PROTOCOL = new OptionBuilder<>("telemetry-protocol", String.class)
+            .category(OptionCategory.TELEMETRY)
+            .description("Telemetry (OpenTelemetry) protocol used for the telemetry data.")
```
ryanemerson: "Protocol used for the telemetry data"?

docs/guides/observability/telemetry.adoc
mabartos: better

docs/guides/observability/telemetry.adoc
mabartos: I'd rather explicitly have them in the guide to increase visibility. Emphasizing that endpoint and protocol is shared + emphasizing service name special behavior with the precedence.

quarkus/config-api/src/main/java/org/keycloak/config/TelemetryOptions.java
mabartos: > That being said, does include the impl here negate the idea of having implementation independent CLI options?

@ryanemerson IMHO, depends on the view of an observer :) Having them implementation independent is rather an "investment" to the future to cover possibilities that the implementation can be switched. But the impl is controlled by the Keycloak, not by users/customer and their extensions. It means that we're in control here. As name of properties would be challenging to change in the future, a change of description is easy without any breaking change introduced. 

I'd still mention the "OpenTelemetry" in the config description to show we're using OTel (same as for Tracing options) to prevent any confusion like "wt hell is Telemetry endpoint" and then navigate somewhere to the docs. 

As you mentioned, we can go directly with the "OpenTelemetry endpoint to connect to" as users/customers knows immediately it's the OTel endpoint. 

Just my 2-3 cents :P

quarkus/config-api/src/main/java/org/keycloak/config/TelemetryOptions.java
mabartos: I'll be more descriptive.

mabartos:
@ryanemerson Thanks for the feedback and suggestions! Updated.

quarkus/config-api/src/main/java/org/keycloak/config/TelemetryOptions.java
ryanemerson: > As name of properties would be challenging to change in the future, a change of description is easy without any breaking change introduced.

Very good point. I think you're right, it makes sense to mention OpenTelemetry in the description to prevent users having to potentially consult the docs to determine this.

I like the updated descriptions :+1:

mabartos:
@keycloak/cloud-native Could you please check it once you have some time? Thanks!

mabartos:
Rebased and resolved conflicts.

@ryanemerson @vmuzikar @shawkins @Pepo48 Could you please check it if you find some time? Thanks!

ryanemerson (dismissed):

mabartos:
Rebased and resolved conflicts.

mabartos:
### Summary

- Improving journey to have OpenTelemetry support for KC observability and its **components** - Logs, Metrics, Traces
- We need to specify OTel Collector settings where the telemetry data is sent
- Created parent options `telemetry-*` for global settings of the OTel Collector
- If the OTel Collector setting is not overridden for the components(Logs, Metrics, Traces), they just inherit the values from the parent options
- [Tracing] Deprecate the `tracing-service-name` and `tracing-resource-attributes` props that set the global settings for the OTel Collector  
- The `telemetry-*` parent options would be marked as **supported** (due to its straightforward manner and need to deprecate these tracing props)

### Plan
#### KC 26.5
- Supported `telemetry-*` options
- Deprecate the `tracing-service-name` and `tracing-resource-attributes` options
- Preview support for OTel Logs (low-hanging fruit)
- Preview support for Micrometer to OTEL bridge (Metrics) - best effort

#### KC 27
- Remove the `tracing-service-name` and `tracing-resource-attributes` options
- OTel Logs supported (if supported on Quarkus side)
- Micrometer to OTEL bridge supported (if supported on Quarkus side)
- Have the full OpenTelemetry support :tada: 

@vmuzikar @ahus1 @ryanemerson @keycloak/cloud-native WDYT?

_PS: It would be good to have the `telemetry-*` options preview for 26.4 due to close release, but then for the tracing deprecation, we would suggest to use the preview options. As we map it only to the Quarkus properties and have multiple test cases, IMHO it should be safe to promote it to **supported** right away._

ryanemerson (dismissed):
Approving again :slightly_smiling_face:

+1 to your plan.

vmuzikar:
I'd like to take a look today or early next week.

mabartos:
Rebased again.

mabartos:
Rebased

docs/guides/observability/telemetry.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+<@opts.expectedValues option="telemetry-protocol"/>
+
+You can change the protocol via CLI as follows:
+
+<@kc.start parameters="--telemetry-protocol=http/protobuf"/>
+
+== Traces
+
+{project_name} has already supported OpenTelemetry Tracing for some time, and you can find more information in the https://www.keycloak.org/observability/tracing[Root cause analysis with tracing guide].
```
vmuzikar: I'd rephrase a bit. IMHO static docs (which are not release notes) should not include time based wording.
```suggestion
See the https://www.keycloak.org/observability/tracing[Root cause analysis with tracing guide].
```

Alternatively, we could merge these two guides.

vmuzikar (changes requested):
@mabartos Sorry for the late response, finally got back to this...

The only missing bit is release note and a deprecation note in the upgrading guide. Could you please add it?

mabartos:
> The only missing bit is release note and a deprecation note in the upgrading guide. Could you please add it?

Oops. I'm on it.

mabartos:
@vmuzikar Thanks for the review! Could you please check it? Thanks!

- Rebased
- Updated the HelpCommandDistTest outputs
- Added the release and deprecation note
- Ignored link to the telemetry guide for now

vmuzikar (approved):
LGTM, thank you.

PR #41712, opened 2025-08-06
Assert correct message for base url

### LINKED ISSUE #41711: Another flaky SAML test
### Before reporting an issue

### Area

admin/ui

### Describe the bug

This test is flaky:
[chromium] › test\clients\saml.spec.ts:212:3 › Clients SAML tests › should check access settings 

Test generates these two error messages:
Image

But the test is checking for the first message twice, which sometimes causes it to fail.

### Version

main

### Regression

### Expected behavior

Test always passes

### Actual behavior

Test doesn't always pass

### How to Reproduce?

Run test several times

### Anything else?

_No response_

ssilvert:
Fixes #41711

### DIFF
--- a/js/apps/admin-ui/test/clients/saml.spec.ts
+++ b/js/apps/admin-ui/test/clients/saml.spec.ts
@@ -212,18 +212,20 @@ test.describe("Clients SAML tests", () => {
   test("should check access settings", async ({ page }) => {
     const validUrl =
       "http://localhost:8180/realms/master/protocol/" + clientId + "/clients/";
-    const invalidUrlError =
+    const invalidUrlErrorRoot =
       "Client could not be updated: invalid_inputRoot URL is not a valid URL";
+    const invalidUrlErrorBase =
+      "Client could not be updated: invalid_inputBase URL is not a valid URL";
 
     await page.getByTestId("rootUrl").fill("Invalid URL");
     await saveSamlSettings(page);
-    await assertNotificationMessage(page, invalidUrlError);
+    await assertNotificationMessage(page, invalidUrlErrorRoot);
     await page.getByTestId("rootUrl").clear();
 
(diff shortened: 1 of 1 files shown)


js/apps/admin-ui/test/clients/saml.spec.ts
```diff
@@ -212,18 +212,20 @@ test.describe("Clients SAML tests", () => {
...
-    const invalidUrlError =
+    const invalidUrlErrorRoot =
       "Client could not be updated: invalid_inputRoot URL is not a valid URL";
+    const invalidUrlErrorBase =
+      "Client could not be updated: invalid_inputBase URL is not a valid URL";
 
     await page.getByTestId("rootUrl").fill("Invalid URL");
     await saveSamlSettings(page);
-    await assertNotificationMessage(page, invalidUrlError);
+    await assertNotificationMessage(page, invalidUrlErrorRoot);
```
jonkoops, line 222: Looks like this variable is only used once, so it might as well be in lined:
```suggestion
    await assertNotificationMessage(page, "Client could not be updated: invalid_inputRoot URL is not a valid URL");
```

js/apps/admin-ui/test/clients/saml.spec.ts
```diff
@@ -212,18 +212,20 @@ test.describe("Clients SAML tests", () => {
...
     await page.getByTestId("rootUrl").clear();
 
     await page.getByTestId("baseUrl").fill("Invalid URL");
     await saveSamlSettings(page);
 
-    await assertNotificationMessage(page, invalidUrlError);
+    await assertNotificationMessage(page, invalidUrlErrorBase);
```
jonkoops, line 228:
```suggestion
    await assertNotificationMessage(page, "Client could not be updated: invalid_inputBase URL is not a valid URL");
```

jonkoops (approved):

PR #41715, opened 2025-08-07
Allow and control sending UTF-8 emails in the default email sender impl

### LINKED ISSUE #41023: Can't send e-mails to international e-mail addresses: bad UTF-8 syntax
### Before reporting an issue

### Area

login/ui

### Describe the bug

When attempting to send an e-mail to an international e-mail address from Keycloak, the following error appears in the log:

```
Jul 09 10:26:42 login docker[1200]: jakarta.mail.SendFailedException: Invalid Addresses;
Jul 09 10:26:42 login docker[1200]:   nested exception is:
Jul 09 10:26:42 login docker[1200]:         org.eclipse.angus.mail.smtp.SMTPAddressFailedException: 500 5.5.2 Error: bad UTF-8 syntax
Jul 09 10:26:42 login docker[1200]:         at org.eclipse.angus.mail.smtp.SMTPTransport.rcptTo(SMTPTransport.java:2097)
Jul 09 10:26:42 login docker[1200]:         at org.eclipse.angus.mail.smtp.SMTPTransport.sendMessage(SMTPTransport.java:1291)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.email.DefaultEmailSenderProvider.send(DefaultEmailSenderProvider.java:86)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.email.DefaultEmailSenderProvider.send(DefaultEmailSenderProvider.java:72)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.email.freemarker.FreeMarkerEmailTemplateProvider.send(FreeMarkerEmailTemplateProvider.java:289)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.email.freemarker.FreeMarkerEmailTemplateProvider.send(FreeMarkerEmailTemplateProvider.java:283)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.email.freemarker.FreeMarkerEmailTemplateProvider.sendSmtpTestEmail(FreeMarkerEmailTemplateProvider.java:128)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.services.resources.admin.RealmAdminResource.testSMTPConnection(RealmAdminResource.java:1148)
Jul 09 10:26:42 login docker[1200]:         at org.keycloak.services.resources.admin.RealmAdminResource$quarkusrestinvoker$testSMTPConnection_0bc3cac9c894e7b29eb4c1f2d80a099cba7bdf38.invoke(Unknown Source)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.resteasy.reactive.server.handlers.InvocationHandler.handle(InvocationHandler.java:29)
Jul 09 10:26:42 login docker[1200]:         at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:141)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.resteasy.reactive.common.core.AbstractResteasyReactiveContext.run(AbstractResteasyReactiveContext.java:147)
Jul 09 10:26:42 login docker[1200]:         at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:638)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2675)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2654)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1627)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1594)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:11)
Jul 09 10:26:42 login docker[1200]:         at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:11)
Jul 09 10:26:42 login docker[1200]:         at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
Jul 09 10:26:42 login docker[1200]:         at java.base/java.lang.Thread.run(Thread.java:1583)
Jul 09 10:26:42 login docker[1200]: Caused by: org.eclipse.angus.mail.smtp.SMTPAddressFailedException: 500 5.5.2 Error: bad UTF-8 syntax
Jul 09 10:26:42 login docker[1200]:         at org.eclipse.angus.mail.smtp.SMTPTransport.rcptTo(SMTPTransport.java:1997)
Jul 09 10:26:42 login docker[1200]:         ... 20 more
```

Using the "Enable Debug SMTP" option in Realm Settings > Email > Connection & Authentication, we can see that Keycloak does not encode the address properly when communicating with the SMTP server, replacing "öüä" with "���" 

```
Jul 09 10:26:42 login docker[1200]: DEBUG SMTP: AUTH LOGIN succeeded
Jul 09 10:26:42 login docker[1200]: DEBUG SMTP: use8bit false
Jul 09 10:26:42 login docker[1200]: MAIL FROM:<keycloak@test.com>
Jul 09 10:26:42 login docker[1200]: 250 2.1.0 Ok
Jul 09 10:26:42 login docker[1200]: RCPT TO:<test+���@test.com>
Jul 09 10:26:42 login docker[1200]: 500 5.5.2 Error: bad UTF-8 syntax
Jul 09 10:26:42 login docker[1200]: DEBUG SMTP: Invalid Addresses
Jul 09 10:26:42 login docker[1200]: DEBUG SMTP:   test+öüä@test.com
Jul 09 10:26:42 login docker[1200]: DEBUG SMTP: Sending failed because of invalid destination addresses
Jul 09 10:26:42 login docker[1200]: RSET
Jul 09 10:26:42 login docker[1200]: 250 2.0.0 Ok
Jul 09 10:26:42 login docker[1200]: DEBUG SMTP: MessagingException while sending, THROW:
Jul 09 10:26:42 login docker[1200]: jakarta.mail.SendFailedException: Invalid Addresses;
```

### Version

26.2.4 (can be reproduced in 26.3.0 as well)

### Regression

### Expected behavior

E-mails can be sent to international e-mail addresses containing UTF-8 characters

### Actual behavior

When trying to send an e-mail from Keycloak to an international e-mail address, e.g. password reset/verify email actions or by using the "Test connection" button in Realm Settings > Email > Connection & Authentication, "Error! Failed to send email" is shown in the UI, and `org.eclipse.angus.mail.smtp.SMTPAddressFailedException: 500 5.5.2 Error: bad UTF-8 syntax` appears in the Keycloak logs on the server` is shown in the Keycloak log on the server.

### How to Reproduce?

1. Create a user with an international e-mail address, assign the "Verify e-mail" required action and set a password
2. Attempt to log in with the user
3. A "Failed to send email" error appears
4. `org.eclipse.angus.mail.smtp.SMTPAddressFailedException: 500 5.5.2 Error: bad UTF-8 syntax` appears in the Keycloak logs on the server

### Anything else?

The behavior can be reproduced with a 'real' SMTP server/relay like Postfix, as well as a 'mock' e-mail server like Maildev. In case of Maildev however, Keycloak is able to send e-mails to Maildev, but the address isn't displayed correctly in the Maildev UI
Image

My theory is that since Maildev is a 'mock' mail server, it doesn't verify e-mail addresses to the same extent as e.g. Postfix.

rmartinc:
Closes #41023

Adding the `allowutf8` option to the smtp configuration to configure [mail.mime.allowutf8](https://jakarta.ee/specifications/mail/2.1/apidocs/jakarta.mail/jakarta/mail/internet/package-summary) in the jakarta configuration with it. This allows to send and control emails with UTF-8 characters if the SMTP server supports it. Test added.

### DIFF
--- a/docs/documentation/server_admin/topics/realms/email.adoc
+++ b/docs/documentation/server_admin/topics/realms/email.adoc
@@ -69,6 +69,9 @@ Auth Token Client Secret::
   Only needed when *Authentication Type* 'token' is selected.
   Supply the *Auth Client Secret* that authenticates the client to fetch a token from the *Auth Token URL*. The value of the *Auth Client Secret* field can refer a value from an external <<_vault-administration,vault>>.
 
+Allow UTF-8::
+  Enable to allow UTF-8 characters in the local part of the email address. This should only be enabled if the mail server supports UTF-8 via the SMTPUTF8 extension. If disabled, domain names containing UTF-8 characters will be encoded using punycode, and addresses containing UTF-8 characters in the local part of the address will return an error.
+
 ifeval::[{project_community}==true]
 
 == XOAUTH2 email configuration with third-party vendors
--- a/docs/documentation/server_admin/topics/realms/email.adoc
+++ b/docs/documentation/server_admin/topics/realms/email.adoc
@@ -70,7 +70,14 @@ Auth Token Client Secret::
   Supply the *Auth Client Secret* that authenticates the client to fetch a token from the *Auth Token URL*. The value of the *Auth Client Secret* field can refer a value from an external <<_vault-administration,vault>>.
 
 Allow UTF-8::
-  Enable to allow UTF-8 characters in the local part of the email address. This should only be enabled if the mail server supports UTF-8 via the SMTPUTF8 extension. If disabled, domain names containing UTF-8 characters will be encoded using punycode, and addresses containing UTF-8 characters in the local part of the address will return an error.
+  Enable to UTF-8-encode email address when sending them to the server. This should only be enabled if the mail server supports UTF-8 via the SMTPUTF8 extension. If disabled, domain names containing non-ASCII characters will be encoded using punycode, and addresses containing non-ASCII characters in the local part of the address will return an error.
++
+If you do not enable this option, take additional measures to prevent non-ASCII characters in users' email addresses:
++
+--
+. Verifying that no email addresses of existing users have non-ASCII characters in the local part of the email address.
+. Updating the validation of email addresses to prevent non-ASCII characters in the local part of the email address, for example, by adding a regex pattern validation in the user profile for the email address field similar to `+\p{ASCII}*@.*+` with an error message similar to "Local part of the address must contain only ASCII characters".
+--
 
 ifeval::[{project_community}==true]
 
--- a/docs/documentation/server_admin/topics/realms/email.adoc
+++ b/docs/documentation/server_admin/topics/realms/email.adoc
@@ -76,7 +76,7 @@ If you do not enable this option, take additional measures to prevent non-ASCII 
 +
 --
 . Verifying that no email addresses of existing users have non-ASCII characters in the local part of the email address.
-. Updating the validation of email addresses to prevent non-ASCII characters in the local part of the email address, for example, by adding a regex pattern validation in the user profile for the email address field similar to `+\p{ASCII}*@.*+` with an error message similar to "Local part of the address must contain only ASCII characters".
+. Updating the validation of email addresses to prevent non-ASCII characters in the local part of the email address, for example, by adding a regex pattern validation in the user profile for the email address field similar to `\p&#123;ASCII&#125;*@.*` with an error message similar to `Local part of the address must contain only ASCII characters`.
 --
 
 ifeval::[{project_community}==true]
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -40,7 +40,7 @@ displayName=Display name
 applyToResourceTypeHelp=Specifies if this permission should be applied to all resources with a given type. In this case, this permission will be evaluated for all instances of a given resource type.
 cibaIntervalHelp=The minimum amount of time in seconds that the CD (Consumption Device) must wait between polling requests to the token endpoint. If set to 0, the CD must use 5 as the default value according to the CIBA specification.
 envelopeFrom=Envelope from
-allowutf8: Allow UTF-8
+allowutf8=Allow UTF-8
 eventTypes.UPDATE_TOTP.name=Update totp
 updateCibaError=Could not update CIBA policy\: {{error}}
 policyUrl=Policy URL
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -40,6 +40,7 @@ displayName=Display name
 applyToResourceTypeHelp=Specifies if this permission should be applied to all resources with a given type. In this case, this permission will be evaluated for all instances of a given resource type.
 cibaIntervalHelp=The minimum amount of time in seconds that the CD (Consumption Device) must wait between polling requests to the token endpoint. If set to 0, the CD must use 5 as the default value according to the CIBA specification.
 envelopeFrom=Envelope from
+allowutf8: Allow UTF-8
 eventTypes.UPDATE_TOTP.name=Update totp
 updateCibaError=Could not update CIBA policy\: {{error}}
 policyUrl=Policy URL
@@ -2733,6 +2734,7 @@ mappingCreatedError=Could not create mapping\: '{{error}}'
 deleteClientPolicyProfileConfirmTitle=Delete profile?
 passwordPoliciesHelp.forceExpiredPasswordChange=The number of days the password is valid before a new password is required.
 envelopeFromHelp=An email address used for bounces (optional).
+allowutf8Help=Enable to allow UTF-8 characters in the local part of the email address. This should only be enabled if the mail server supports UTF-8 via the SMTPUTF8 extension. If disabled, domain names containing UTF-8 characters will be encoded using punycode, and addresses containing UTF-8 characters in the local part of the address will return an error.
 passwordPoliciesHelp.upperCase=The number of uppercase letters required in the password string.
 policyDeletedError=Could not remove the resource {{error}}
 key=Key
--- a/server-spi-private/src/main/java/org/keycloak/email/EmailSenderProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/email/EmailSenderProvider.java
@@ -32,4 +32,11 @@ public interface EmailSenderProvider extends Provider {
     }
 
     void send(Map<String, String> config, String address, String subject, String textBody, String htmlBody) throws EmailException;
+
+    /**
+     * Validates configuration for the SMTP sender.
+     * @param config The configuration to test
+     * @throws EmailException If some error is found
+     */
+    void validate(Map<String, String> config) throws EmailException;
 }
--- a/services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
+++ b/services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
@@ -17,13 +17,14 @@
 
 package org.keycloak.email;
 
-import jakarta.mail.internet.MimeUtility;
 import org.jboss.logging.Logger;
 import org.keycloak.common.enums.HostnameVerificationPolicy;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.UserModel;
 import org.keycloak.services.ServicesLogger;
 import org.keycloak.truststore.JSSETruststoreConfigurator;
+import org.keycloak.utils.EmailValidationUtil;
+import org.keycloak.utils.SMTPUtil;
 
 import jakarta.mail.Address;
 import jakarta.mail.MessagingException;
@@ -36,6 +37,7 @@ import jakarta.mail.internet.InternetAddress;
 import jakarta.mail.internet.MimeBodyPart;
 import jakarta.mail.internet.MimeMultipart;
 import jakarta.mail.internet.MimeMessage;
+import jakarta.mail.internet.MimeUtility;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
 import java.io.UnsupportedEncodingException;
@@ -74,16 +76,20 @@ public class DefaultEmailSenderProvider implements EmailSenderProvider {
 
     @Override
     public void send(Map<String, String> config, String address, String subject, String textBody, String htmlBody) throws EmailException {
-        Session session = Session.getInstance(buildEmailProperties(config));
+        final boolean allowutf8 = isAllowUTF8(config);
+        final String convertedAddress = checkUserAddress(address, allowutf8);
+        final String from = checkFromAddress(config.get("from"), allowutf8);
 
-        Message message = buildMessage(session, address, subject, config, buildMultipartBody(textBody, htmlBody));
+        Session session = Session.getInstance(buildEmailProperties(config, from));
+
+        Message message = buildMessage(session, convertedAddress, from, subject, config, buildMultipartBody(textBody, htmlBody));
 
         try(Transport transport = session.getTransport("smtp")) {
 
             EmailAuthenticator selectedAuthenticator = selectAuthenticatorBasedOnConfig(config);
             selectedAuthenticator.connect(this.session, config, transport);
 
-            transport.sendMessage(message, new InternetAddress[]{new InternetAddress(address)});
+            transport.sendMessage(message, new InternetAddress[]{new InternetAddress(convertedAddress)});
 
         } catch (Exception e) {
             ServicesLogger.LOGGER.failedToSendEmail(e);
@@ -91,7 +97,13 @@ public class DefaultEmailSenderProvider implements EmailSenderProvider {
         }
     }
 
-    private Properties buildEmailProperties(Map<String, String> config) throws EmailException {
+    @Override
+    public void validate(Map<String, String> config) throws EmailException {
+        // just static configuration checking here, not really testing email
+        checkFromAddress(config.get("from"), isAllowUTF8(config));
+    }
+
+    private Properties buildEmailProperties(Map<String, String> config, String from) throws EmailException {
         Properties props = new Properties();
 
         if (config.containsKey("host")) {
@@ -137,9 +149,9 @@ public class DefaultEmailSenderProvider implements EmailSenderProvider {
             props.setProperty("mail.smtp.from", envelopeFrom);
         }
 
-        String from = config.get("from");
-        if (from == null) {
-            throw new EmailException("No sender address configured in the realm settings for emails");
+        final boolean allowutf8 = isAllowUTF8(config);
+        if (allowutf8) {
+            props.setProperty("mail.mime.allowutf8", "true");
         }
 
         // Specify 'mail.from' as InternetAddress.getLocalAddress() would otherwise do a InetAddress.getCanonicalHostName
@@ -150,12 +162,8 @@ public class DefaultEmailSenderProvider implements EmailSenderProvider {
         return props;
     }
 
-    private Message buildMessage(Session session, String address, String subject, Map<String, String> config, Multipart multipart) throws EmailException {
+    private Message buildMessage(Session session, String address, String from, String subject, Map<String, String> config, Multipart multipart) throws EmailException {
 
-        String from = config.get("from");
-        if (from == null) {
-            throw new EmailException("No sender address configured in the realm settings for emails");
-        }
         String fromDisplayName = config.get("fromDisplayName");
         String replyTo = config.get("replyTo");
         String replyToDisplayName = config.get("replyToDisplayName");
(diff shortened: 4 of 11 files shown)


jonkoops (dismissed):
Looks great!

services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
```diff
@@ -74,6 +75,10 @@ public void send(Map<String, String> config, UserModel user, String subject, Str
 
     @Override
     public void send(Map<String, String> config, String address, String subject, String textBody, String htmlBody) throws EmailException {
+        if (!isValidEmail(address, isAllowUTF8(config))) {
+            throw new EmailException("Invalid address: " + address);
```
ahus1: This error message is not actionable as it doesn't give a hint what to change.

services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
```diff
@@ -137,9 +142,14 @@ private Properties buildEmailProperties(Map<String, String> config) throws Email
...
+        final boolean allowutf8 = isAllowUTF8(config);
+        if (allowutf8) {
+            props.setProperty("mail.mime.allowutf8", "true");
+        }
+
         String from = config.get("from");
-        if (from == null) {
-            throw new EmailException("No sender address configured in the realm settings for emails");
+        if (!isValidEmail(from, allowutf8)) {
+            throw new EmailException("Invalid sender address configured in the realm settings for emails: " + from);
```
ahus1: As above, this is not actionable

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -2726,6 +2727,7 @@ mappingCreatedError=Could not create mapping\: '{{error}}'
 deleteClientPolicyProfileConfirmTitle=Delete profile?
 passwordPoliciesHelp.forceExpiredPasswordChange=The number of days the password is valid before a new password is required.
 envelopeFromHelp=An email address used for bounces (optional).
+allowutf8Help=Option to allow UTF-8 strings in message headers, for example in addresses. This should only be enabled if the mail server also supports UTF-8 (SMTPUTF8 extension).
```
ahus1: This is IMHO only half the truth: When not enabling `allowuft8`, I won't be able to send emails from or to an email address that contains UTF8 characters before the "@". So any user that has entered a domain with such a special character will not be able to receive emails.

services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
```diff
@@ -236,6 +250,11 @@ private boolean isAuthTypeTokenConfigured(Map<String, String> config) {
         return "token".equals(config.get("authType"));
     }
 
+    private static boolean isValidEmail(String email, boolean allowutf8) {
+        return EmailValidationUtil.isValidEmail(email)
```
ahus1: This is too restrictive, and IMHO points to a different problem: you can have an email address with a domain name that is IDN encoded. 
It seems to me that we should pny-encode those domain names before sending them. 

Interesting enough the frontend for setting the sender address will encode the email address in pnycode automatically, while a user's email address when fetched from the database when sending the test email will not be pny-encoded.

ahus1 (dismissed):
The PR has this setting now to "false", which might lead to some emails no longer being handed over to the SMTP-server. 

So I wonder: 
* Should the default be "true" for new SMTP configurations? 
* Is this a notable change, that should mention the new option, as those migrating should review their configuration and set it to "true"?

Something that is IMHO missing: 
* If we validate the "from" address like above, the validation should also be performed when we save the configuration. 

Something that we could test: 
* Does the server support UTF8 - this can be tested by the following - which unfortunately calls the implementation, and not the interface. The Angus mail implementation only logs an INFO is there is a mismatch.
`((SMTPTransport) transport).supportsExtension("SMTPUTF8")`

See some more comments below. Feel free to split those ...

docs/documentation/server_admin/topics/realms/email.adoc
```diff
@@ -69,6 +69,9 @@ Auth Token Client Secret::
   Only needed when *Authentication Type* 'token' is selected.
   Supply the *Auth Client Secret* that authenticates the client to fetch a token from the *Auth Token URL*. The value of the *Auth Client Secret* field can refer a value from an external <<_vault-administration,vault>>.
 
+Allow UTF-8::
+  Option to allow UTF-8 strings in message headers, for example in addresses. This should only be enabled if the mail server also supports UTF-8 (SMTPUTF8 extension).
```
slaskawi: Would it be possible to make this option "true" by default? Is the UTF8 mode supported by the recent versions of the popular SMTP servers?

rmartinc:
@ahus1 I have done some changes but not all of them :smile:. In general the `allowutf8` is false by default because it was false before (it was not set and it's false by default in the jakarta mail impl). So we are no limiting previous configurations, it is working as before but now you can configure that prop to allow UTF-8. Previously there was an error sending the email (check issue) or worse because it was not controlled, and now there is a specific check that throws an exception.

> Should the default be "true" for new SMTP configurations?

I wouldn't do that, nobody complained about this until now. We can do it, but then I need to add a migration task to add `false` to current realms. If everybody agrees, I can work on it.

> Is this a notable change, that should mention the new option, as those migrating should review their configuration and set it to "true"?

No, I don't think that it is a notable change. The option is false as before, so we are not changing the behavior.

> If we validate the "from" address like above, the validation should also be performed when we save the configuration.

I added that part. But I'm not testing the email completely, just statically. If I add a full test (test SMTP connection, username/password and so on and so forth), there are tons of tests that fail (because they are importing a realm with email configuration and the SMTP server is not setup).

> Does the server support UTF8 - this can be tested by the following - which unfortunately calls the implementation, and not the interface. The Angus mail implementation only logs an INFO is there is a mismatch.

We could add something similar to that, but I have not done it, mainly because the greenmail we use for testing does not announce SMTPUTF8 (I prefer testing something). The admin should know if they want/support UTF-8 or not. We can check if SMTPUTF8 extension is listed when connecting to send the email, and throw an exception if it is configured and not present. But, as commented, that would break testing. If everybody thinks that it's better, I can add the check when sending an email.

services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
rmartinc: I have changed to: `Invalid address '%s'. If the address contains UTF-8 characters in the local part please ensure the SMTP server supports SMTPUTF8 extensions and enable 'Allow UTF-8' in the email realm configuration.`

services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
rmartinc: Changed to: `Invalid sender address '%s'. If the address contains UTF-8 characters in the local part please ensure the SMTP server supports SMTPUTF8 extensions and enable 'Allow UTF-8' in the email realm configuration.`

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
rmartinc: I have just copied the description of the [mail.mime.allowutf8](https://jakarta.ee/specifications/mail/2.1/apidocs/jakarta.mail/jakarta/mail/internet/package-summary) option. What do you prefer prefer here? Say that is only for local part?

services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java
rmartinc: Okis, I have changed to use `allowutf8` and, if not set, change to the `IDN.toASCCI` method in the domain part. Note that this was exactly the same before.

docs/documentation/server_admin/topics/realms/email.adoc
rmartinc: I prefer to not do that, at least for now. If we set it to true by default we need to add a task to migrate the realms and add the false (if not, when edited by the admin console it would change to true). We can do in a follow-up only for the main branch. But as commented to @ahus1 we can do it, but I would not backport that part.

rmartinc:
It seems CI fails because we are waiting for https://github.com/keycloak/keycloak/pull/41794. I will rebase tomorrow.

docs/documentation/server_admin/topics/realms/email.adoc
slaskawi: I'm absolutely fine with this approach!

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
ahus1: I would bring this into context for a Keycloak admin. For Keycloak, this would only be about addresses, and no other headers. 

It would be good to make it more specific about the local parts. 

"Enable to allow UTF-8 characters in the local part of the email address. This should only be enabled if the mail server supports UTF-8 via the SMTPUTF8 extension. If disabled, domain names containing UTF-8 characters will be encoded using punycode, and addresses containing UTF-8 characters in the local part of the address will return an error."

rmartinc: Okis, changed in the documentation and for the admin console help message.

ahus1:
I disagree with the following statement: 

> Previously there was an error sending the email (check issue) or worse because it was not controlled, and now there is a specific check that throws an exception.

We know about one email server that is supporting the SMTPUTF8 extension that is throwing an exception. Other email servers might just accept it. Now we are always throwing an exception when there is a local part with UTF-8 characters. So to me this is a breaking change. 

I vote to remove the check if UTF-8 characters are part of the local part of the email to be backwards compatible. No-one complained that we didn't do that for the past 10 years.

rmartinc:
> We know about one email server that is supporting the SMTPUTF8 extension that is throwing an exception. Other email servers might just accept it. Now we are always throwing an exception when there is a local part with UTF-8 characters. So to me this is a breaking change.
>
>I vote to remove the check if UTF-8 characters are part of the local part of the email to be backwards compatible. No-one complained that we didn't do that for the past 10 years.

@ahus1 The angus mail implementation sends the email when the `allowutf8` option is false to just the lower byte (if there are two in UTF-8). If you send an email for example to `admin@localhostñ.com` you will see this `admin@localhost�.com` in the log:

```
DEBUG SMTP: connected to host "localhost", port: 25
EHLO localhost
250-li-a1c93acc-2366-11b2-a85c-b79ae5c81821.ibm.com Hello localhost [127.0.0.1])
250-PIPELINING
250-ENHANCEDSTATUSCODES
250 8BITMIME
DEBUG SMTP: Found extension "PIPELINING", arg ""
DEBUG SMTP: Found extension "ENHANCEDSTATUSCODES", arg ""
DEBUG SMTP: Found extension "8BITMIME", arg ""
DEBUG SMTP: use8bit false
MAIL FROM:<keycloak@localhost.com>
250 2.1.0 Sender <keycloak@localhost.com> OK
RCPT TO:<admin@localhost�.com>
553 5.1.3 Syntax error in recipient address
DEBUG SMTP: Invalid Addresses
DEBUG SMTP:   admin@localhostñ.com
DEBUG SMTP: Sending failed because of invalid destination addresses
RSET
250 2.0.0 OK
DEBUG SMTP: MessagingException while sending, THROW: 
jakarta.mail.SendFailedException: Invalid Addresses;
  nested exception is:
	org.eclipse.angus.mail.smtp.SMTPAddressFailedException: 553 5.1.3 Syntax error in recipient address
```

This means that you can do weird things and the reason that this is managed as important issue (not an enhancement), we have a moderate [CVE-2025-8419](https://access.redhat.com/security/cve/CVE-2025-8419) related to this (UTF-8 characters are transformed without any control and an attacker can inject into the SMTP, not much because the email is very short, but it's moderate). So, the important part of this PR is that we should control the email and reject invalid ones to avoid the issue.

rmartinc:
I'm going to add an upgrading note just in case. I don't think that this will affect a lot of people, but it's not the first time I've failed in a prediction like this one. :smile:
@ahus1 Don't approve it yet, I'll add the note during the day.

server-spi-private/src/main/java/org/keycloak/email/EmailSenderProvider.java
```diff
@@ -32,4 +32,11 @@ default void send(Map<String, String> config, UserModel user, String subject, St
     }
 
     void send(Map<String, String> config, String address, String subject, String textBody, String htmlBody) throws EmailException;
+
+    /**
+     * Check configuration for the SMTP sender.
+     * @param config The configuration to test
+     * @throws EmailException If some error is found
+     */
+    void check(Map<String, String> config) throws EmailException;
```
slaskawi: Nitpick: Maybe `validate` would be a better name. At least it's more commonly used.

services/src/main/java/org/keycloak/utils/SMTPUtil.java
```diff
@@ -0,0 +1,56 @@
...
+package org.keycloak.utils;
+
+import java.net.IDN;
+import java.util.Map;
+import org.keycloak.email.EmailException;
+import org.keycloak.email.EmailSenderProvider;
+import org.keycloak.models.KeycloakSession;
+
+/**
+ *
```
slaskawi, line 28: Probably a valid Javadoc would be useful.

tests/base/src/test/java/org/keycloak/tests/admin/SMTPConnectionTest.java
```diff
@@ -220,9 +225,54 @@ public void testWithTokenAuthEnabledAndRetryWithValidTokenInSecondTry() throws E
...
 
+    @Test
+    @Order(9)
+    public void testAllowUTF8() throws Exception {
+        // utf8 on from not allowed if allowutf8 not enabled
+        Response response = adminClient.realms().realm(managedRealm.getName()).testSMTPConnection(settings("127.0.0.1", "3025", "autoñ@keycloak.org",
+                null, null, null, null, null));
+        assertStatus(response, 500);
+
+        // utf-8 on from but in domain part is allowed and transformed
```
slaskawi, line 236: Nitpick: These seem to be separate test cases. Perhaps it would be useful to put them in separate test methods? Maybe we could even execute them in parallel?

services/src/main/java/org/keycloak/utils/SMTPUtil.java
rmartinc: Done!

tests/base/src/test/java/org/keycloak/tests/admin/SMTPConnectionTest.java
rmartinc: I think that it's better here, we have everything to test email in this class and the class is fast to test (it's not one of the worse ones9.

server-spi-private/src/main/java/org/keycloak/email/EmailSenderProvider.java
rmartinc: Yep, you are right. Changed!

rmartinc:
Ok, I have added the upgrading note and done 2 of 3 of the suggestions from @slaskawi. Let's wait the CI now. I have also squashed the commits into just one.

ahus1:
@rmartinc - I've added one more commit to the docs to give admin who are migrating some hands-on advice. For example, without updating the email validation rule, users might register with an email address that contains non-ASCII characters, and then they wouldn't be able to confirm their email address. 

I also fixed something that I missed before: While there are ASCII and non-ASCII characters, UTF-8 is an encoding and therefore doesn't have any characters. So I've update the statements. 

It is getting quite late here, so please re-read those paragraphs. I'll mark my earlier review now resolved, so Marek can eventually proceed.

tests/base/src/test/java/org/keycloak/tests/admin/SMTPConnectionTest.java
slaskawi: This approach works for me as well. Resolving.

slaskawi (approved):
LGTM!

rmartinc:
@ahus1 LGTM and adding another check for emails if you don't support non-ascii in the local part makes lot of sense. I have just changed the expression to use entities and avoid the issue with the variables.

rmartinc:
One more comment, note that we can create a new email validator that, if smtp configuration in the realm does not allow UTF-8, checks that the email local part has only ascii characters. This validator can be assigned automatically using User Profile. But I don't know if we are complicating this too much. But just in case.

mposolda (dismissed):
@rmartinc @ahus1 @slaskawi @jonkoops PR looks good to me, Thanks! It would just need one additional rebase due the conflict in the documentation.

@ahus1 @slaskawi Are you also ok with the PR now? Feel free to re-approve or comment if it looks OK to you.

ahus1 (dismissed):

ahus1:
I'm ok with the change, and just did the rebase.

ahus1 (approved):

ahus1:
@rmartinc / @mposolda - it would be nice to have a follow-up issue as setting up SMTP now got quite complicated :(

slaskawi (approved):
@mposolda LGTM!

PR #41716, opened 2025-08-07
[OTel] Micrometer to OpenTelemetry bridge support for metrics

### LINKED ISSUE #41006: Micrometer to OpenTelemetry bridge for metrics
### Description

Quarkus provides the extension [quarkus-micrometer-opentelemetry](https://quarkus.io/guides/telemetry-micrometer-to-opentelemetry) to support Micrometer and OpenTelemetry Metrics together. As we're using Micrometer for instrumenting Keycloak metrics and other metrics, it was not possible to emit these metrics even on the OTel collector, which would be possible to do it now. 

It means that users/customer might be little bit closer to having OpenTelemetry everywhere for traces, metrics and potentially logs. No other additional stack would not be needed. Instead of using pulling mechanism from Prometheus to get periodically metrics from the metrics endpoint, with OTel Metrics, these metrics are rather pushed to OTel collector as f.e. OTel Tracing that we already support. 

With this extension, we should be able to keep the current approach for Prometheus and additionally also emit these metrics to the OTel Collector.

Related info: 
https://github.com/keycloak/keycloak/discussions/8490#discussioncomment-1419625
https://github.com/keycloak/keycloak/discussions/35012

### Discussion

_No response_

### Motivation

Improve Keycloak's observability story

### Details

_No response_

mabartos:
- Closes #41006 
- Metrics are exported to the OTel Collector
- However, as we don't have any better OTel testing, we cannot verify that all metrics are sent properly
image

## Try it out
1. Start the LGTM service:
```
docker run -p 3000:3000 -p 4317:4317 -p 4318:4318 --rm -ti grafana/otel-lgtm
```
2. Start the KC server:
```
 ./kc.sh start-dev --metrics-enabled=true --feature-opentelemetry-metrics=enabled --telemetry-metrics-enabled=true
```

3. Access the `localhost:3000` to see Grafana --> Drilldown --> Metrics

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -131,6 +131,7 @@ public class Profile {
 
         OPENTELEMETRY("OpenTelemetry support", Type.DEFAULT),
         OPENTELEMETRY_LOGS("OpenTelemetry Logs support", Type.PREVIEW, OPENTELEMETRY),
+        OPENTELEMETRY_METRICS("Micrometer to OpenTelemetry bridge support for metrics", Type.EXPERIMENTAL, OPENTELEMETRY),
 
         DECLARATIVE_UI("declarative ui spi", Type.EXPERIMENTAL),
 
--- a/docs/documentation/release_notes/topics/26_5_0.adoc
+++ b/docs/documentation/release_notes/topics/26_5_0.adoc
@@ -156,6 +156,13 @@ For more details, see the link:{telemetryguide_link}[{telemetryguide_name}] guid
 == OpenTelemetry Logs (preview)
 
 {project_name} now supports exporting logs to OpenTelemetry collectors, enabling centralized log management.
-This preview feature allows you to export {project_name} logs to any OpenTelemetry-compatible backend and use the same OpenTelemetry collector for logs and traces.
+This preview feature allows you to export {project_name} logs to any OpenTelemetry-compatible backend and use the same OpenTelemetry collector for logs, metrics and traces.
+
+For more details, see the link:{telemetryguide_link}[{telemetryguide_name}] guide.
+
+== OpenTelemetry Metrics (experimental)
+
+{project_name} now provides the experimental support for exporting metrics to OpenTelemetry collectors by using the https://quarkus.io/guides/telemetry-micrometer-to-opentelemetry[Micrometer-to-OpenTelemetry bridge].
+This experimental feature allows you to export {project_name} metrics to any OpenTelemetry-compatible backend and use the same OpenTelemetry collector for logs, metrics and traces.
 
 For more details, see the link:{telemetryguide_link}[{telemetryguide_name}] guide.
--- a/docs/guides/observability/telemetry.adoc
+++ b/docs/guides/observability/telemetry.adoc
@@ -82,12 +82,13 @@ For example, if you want only to export `WARN` and `ERROR` logs, you can change 
 
 == Metrics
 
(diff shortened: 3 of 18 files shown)


mabartos:
@ryanemerson @vmuzikar Could you please check it? Thanks!

pruivo:
Unfortunately, the https://github.com/keycloak/keycloak-grafana-dashboard is not compatible with OTEL because of its naming convention. Example:

* Micrometer: http_server_requests_seconds_bucket
* OTEL: http_server_requests_milliseconds_bucket

Screenshot 2025-12-16 at 16-15-41 Keycloak troubleshooting dashboard - Dashboards - Grafana

mabartos:
@pruivo Thanks for the check! Yes, there might be some differences between them, as seen here: https://quarkus.io/guides/telemetry-micrometer-to-opentelemetry#metric-differences-between-micrometer-and-opentelemetry. 

That's also a reason why it's marked as experimental, as we can expect some metrics might not be compatible. But the /metrics endpoint with the pull mechanism should still work the same, right? Or is it changed there as well, and when this feature is enabled, the KC Grafana dashboards are broken? 

Isn't there any option on how to determine for Grafana dashboards if these metrics are coming from the OTel collector, use the OTel names, and if not, use the Micrometer ones?

pruivo:
> That's also a reason why it's marked as experimental, as we can expect some metrics might not be compatible. But the /metrics endpoint with the pull mechanism should still work the same, right? Or is it changed there as well, and when this feature is enabled, the KC Grafana dashboards are broken?

@mabartos Ah! I wasn't aware `/metrics` endpoint was still active when OTEL is enabled. Those are fine! 

> Isn't there any option on how to determine for Grafana dashboards if these metrics are coming from the OTel collector, use the OTel names, and if not, use the Micrometer ones?

AFAIK, no. Grafana only sees Prometheus DataSource (I'm using the image `grafana/otel-lgtm`).
The problem is not only the name itself but the units of the data (seconds vs milliseconds), which requires changing the Prometheus queries and Grafana panel units. 

Let's merge this and see the community feedback. Great work!

ryanemerson:
@mabartos I tried locally and it works as expected. I hope you don't mind, but I have pushed a couple of commits altering the docs and the option descriptions, as I thought this would be quicker than adding them as comments. Please feel free to disagree/reject to my suggested changes :slightly_smiling_face:

mabartos:
> AFAIK, no. Grafana only sees Prometheus DataSource (I'm using the image grafana/otel-lgtm).
The problem is not only the name itself but the units of the data (seconds vs milliseconds), which requires changing the Prometheus queries and Grafana panel units. Let's merge this and see the community feedback. Great work!

@pruivo Yes, that's correct. After the 26.5 release, we can plan on how to achieve it. Thanks.

> I tried locally and it works as expected. I hope you don't mind, but I have pushed a couple of commits altering the docs and the option descriptions, as I thought this would be quicker than adding them as comments. Please feel free to disagree/reject to my suggested changes 🙂

@ryanemerson Agree, it's quicker and better. Makes sense to me. Thanks!

mabartos:
CI passed.

@pruivo @ryanemerson @vmuzikar Could you please review it? Thanks!

vmuzikar (approved):
LGTM, thank you @mabartos!

Not merging yet as I'd like to ask @pruivo and/or @ryanemerson to review too.

pruivo (approved):

ryanemerson (approved):
Thanks @mabartos

mabartos:
@vmuzikar @pruivo @ryanemerson Thanks!

PR #41719, opened 2025-08-07
Use Optional instead of getOrDefault for settings in testSMTPConnection (26.2)

### LINKED ISSUE #41643: Test SMTP connection fails when no port is specified

rmartinc:
Closes #41643

PR:             https://github.com/keycloak/keycloak/pull/41661
Commit:         https://github.com/keycloak/keycloak/commit/a58556d761cdee0647bb7293665d99d1770152fb
PR branch:      backport-41661-26.2
Target branch:  https://github.com/keycloak/keycloak/tree/release/26.2

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/RealmAdminResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/RealmAdminResource.java
@@ -119,6 +119,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


rmartinc:
It seems that there are issues in the JS CI. If the last node is used 22.18.0 there are errors in the tests (see https://github.com/keycloak/keycloak/actions/runs/16804510625/job/47593956118?pr=41719). Chatting with @jonkoops, commented me to test with the previous version and it seems that is working. So we have to use the previous node version or add all the file extensions in those imports.

Finally this issue also affects main, so the JS CI failed because of https://github.com/keycloak/keycloak/issues/41728

rmartinc:
Rebase to incorporate the fix for node fix.

ahus1 (approved):

PR #41725, opened 2025-08-07
fix: updating docs related to the number of executor threads

shawkins:
closes: #39342

### DIFF
--- a/docs/guides/high-availability/partials/concepts/threads.adoc
+++ b/docs/guides/high-availability/partials/concepts/threads.adoc
@@ -1,22 +1,23 @@
 [#${parent}-quarkus-executor-pool]
 === Quarkus executor pool
 
-{project_name} requests, as well as blocking probes, are handled by an executor pool. Depending on the available CPU cores, it has a maximum size of 50 or more threads.
+{project_name} requests, as well as blocking probes, are handled by an executor pool. The work performed by this pool blocking, which is not always CPU-bound and may wait on resources or IO operations.
+Depending on the available CPU cores, it has a default maximum size of 50 or more threads.
 Threads are created as needed, and will end when no longer needed, so the system will scale up and down automatically.
-{project_name} allows configuring the maximum thread pool size by the link:{links_server_all-config_url}?q=http-pool-max-threads[`http-pool-max-threads`] configuration option.
+{project_name} allows configuring the maximum thread pool size by the link:{links_server_all-config_url}?q=http-pool-max-threads[`http-pool-max-threads`] configuration option. See <@links.ha id="deploy-keycloak-kubernetes" /> for an example.
 
-When running on {kubernetes}, adjust the number of worker threads to avoid creating more load than what the CPU limit allows for the Pod to avoid throttling, which would lead to congestion.
-When running on physical machines, adjust the number of worker threads to avoid creating more load than the node can handle to avoid congestion.
-Congestion would result in longer response times and an increased memory usage, and eventually an unstable system.
-
+There are benefits to using the smallest maximum pool size that meets your throughput needs.
+Fewer threads will reduce the amount of stack space usage, allow for fewer connections in your database and LDAP connection pools, and minimize thread context switching overhead.
 Ideally, you should start with a low limit of threads and adjust it accordingly to the target throughput and response time.
-When the load and the number of threads increases, the database connections can also become a bottleneck.
+
+Under load database connections may become a bottleneck if there is a mismatch between the executor pool maximum size and the connection pool size.
+The number of database connections defaults to 100. The database connection pool is configured via the link:{links_server_all-config_url}?q=db-pool[`Database` settings `db-pool-initial-size`, `db-pool-min-size` and `db-pool-max-size`] respectively.
 Once a request cannot acquire a database connection within 5 seconds, it will fail with a message in the log like `Unable to acquire JDBC Connection`.
 The caller will receive a response with a 5xx HTTP status code indicating a server side error.
+To avoid seeing this error, you will generally want more database connections than executor pool threads as some requests can use more than one connection and timed or other async work may also use connections.
 
-If you increase the number of database connections and the number of threads too much, the system will be congested under a high load with requests queueing up, which leads to a bad performance.
-The number of database connections is configured via the link:{links_server_all-config_url}?q=db-pool[`Database` settings `db-pool-initial-size`, `db-pool-min-size` and `db-pool-max-size`] respectively.
-Low numbers ensure fast response times for all clients, even if there is an occasionally failing request when there is a load spike.
+NOTE: `argon2` password hashing is by default limited to the number of available processors.
+Increasing the executor pool maximum size beyond the number of threads needed to maintain a concurrent workload of this size will have no effect on the rate at which password hashes can be performed.
 
 [#${parent}-load-shedding]
 === Load Shedding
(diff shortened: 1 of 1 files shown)


docs/guides/high-availability/partials/concepts/threads.adoc
```diff
@@ -1,22 +1,23 @@
 [#${parent}-quarkus-executor-pool]
 === Quarkus executor pool
 
-{project_name} requests, as well as blocking probes, are handled by an executor pool. Depending on the available CPU cores, it has a maximum size of 50 or more threads.
+{project_name} requests, as well as blocking probes, are handled by an executor pool. The work performed by this pool blocking, which is not always CPU-bound and may wait on resources or IO operations.
```
ryanemerson: I think the second sentence needs padding out a little to make the intent more clear, how about:

"The work performed by this pool can be blocking, as it is not always CPU bound and may require waiting on resources or IO operations."

docs/guides/high-availability/partials/concepts/threads.adoc
```diff
@@ -1,22 +1,23 @@
...
+Depending on the available CPU cores, it has a default maximum size of 50 or more threads.
 Threads are created as needed, and will end when no longer needed, so the system will scale up and down automatically.
-{project_name} allows configuring the maximum thread pool size by the link:{links_server_all-config_url}?q=http-pool-max-threads[`http-pool-max-threads`] configuration option.
-
-When running on {kubernetes}, adjust the number of worker threads to avoid creating more load than what the CPU limit allows for the Pod to avoid throttling, which would lead to congestion.
-When running on physical machines, adjust the number of worker threads to avoid creating more load than the node can handle to avoid congestion.
-Congestion would result in longer response times and an increased memory usage, and eventually an unstable system.
+{project_name} allows configuring the maximum thread pool size by the link:{links_server_all-config_url}?q=http-pool-max-threads[`http-pool-max-threads`] configuration option. See <@links.ha id="deploy-keycloak-kubernetes" /> for an example.
```
ryanemerson: The page you linked in the second sentence doesn't have an example of configuring `http-pool-max-threads`, it just references it.

docs/guides/high-availability/partials/concepts/threads.adoc
```diff
@@ -1,22 +1,23 @@
...
 Once a request cannot acquire a database connection within 5 seconds, it will fail with a message in the log like `Unable to acquire JDBC Connection`.
 The caller will receive a response with a 5xx HTTP status code indicating a server side error.
+To avoid seeing this error, you will generally want more database connections than executor pool threads as some requests can use more than one connection and timed or other async work may also use connections.
 
-If you increase the number of database connections and the number of threads too much, the system will be congested under a high load with requests queueing up, which leads to a bad performance.
-The number of database connections is configured via the link:{links_server_all-config_url}?q=db-pool[`Database` settings `db-pool-initial-size`, `db-pool-min-size` and `db-pool-max-size`] respectively.
-Low numbers ensure fast response times for all clients, even if there is an occasionally failing request when there is a load spike.
+NOTE: `argon2` password hashing is by default limited to the number of available processors.
```
ryanemerson: Optional nitpick: "`argon2` password hashing is limited to the number of available processors by default.

ryanemerson (changes requested):
Thanks @shawkins, this a nice improvement. Just a few minor comments.

shawkins:
@ryanemerson made updates based upon your review

ahus1:
Thank you for the update of the docs. While the content is technically correct, I am not sure it is helpful to the reader. This situation was probably true also for the original context. 

This leads me to the questions: 

* If we think that setting the max number of DB connections to the number of threads + X, as well as with LDAP, why wouldn't we make that the default? 
* With KC, not all requests actually touch the database. You see a high DB count usually when the DB is (dead-)locking, or when it is overloaded. So looking only at KC parameters doesn't IMHO help much. 
* If we want people to understand this, we might first need to explain what an CPU/resource/IO constraint system is, how to detect it, and remedies.

shawkins:
> While the content is technically correct, I am not sure it is helpful to the reader. This situation was probably true also for the original context.

Yes, the original text was not very helpful. These changes are to hopefully make things better based upon concerns raised in https://github.com/keycloak/keycloak/discussions/39233 - but I agree there's not a lot an admin should do without even more elaboration.

My hope was to make guidance around the max threads read more like "you can change this, but be aware of why you should do that and the affects of limiting resources such as database / ldap connections, and password hashing".

> If we think that setting the max number of DB connections to the number of threads + X, as well as with LDAP, why wouldn't we make that the default?

That was hopefully to be covered in part by https://github.com/keycloak/keycloak/issues/39333 - but there wasn't anything actionable from that. It's a valid question where does 100 come from, and why is it fixed. Going the other direction, we did have a rationale for having relatively fewer executor threads. We are allowing the number of threads to vary based upon the available processors - but that does not have an affect on the number of connections.

Some of our other recommendations, like keeping initial = min = max, could easily be implemented as the default for the production profile - but we don't do that either.

> With KC, not all requests actually touch the database. You see a high DB count usually when the DB is (dead-)locking, or when it is overloaded. So looking only at KC parameters doesn't IMHO help much.

Would you rather have most distinct language about optimizing (reducing threads and connection pool sizes) and diagnosing issues rather than general guidance (even if it is inline with our defaults)?

mabartos:
@ryanemerson @ahus1 @vmuzikar Do you wanna re-review?

ahus1:
@shawkins / @mabartos - I don't have much input on this one, and I am not sure how helpful even the new text is. 

Reading it again today, I might vote for removing the `argon2` paragraph as it doesn't add value IMHO: `argon2` is CPU bound by itself, and the fact that we're running this not in parallel with more core that you have is an implementation detail. It is "just" preventing CPU throttling. 

Also the sentence of "The work performed by this pool is not always CPU bound and may require waiting on resources or IO operations." is hard to digest, and readers without a specific background might wonder what to make out of it. IMHO it can be safely removed. 

I won't object to those changes even though I think it doesn't add much actionable information. Please proceed if you think they will help users.

ahus1 (commented):
@shawkins / @mabartos - I don't have much input on this one, and I am not sure how helpful even the new text is.

Reading it again today, I might vote for removing the argon2 paragraph as it doesn't add value IMHO: argon2 is CPU bound by itself, and the fact that we're running this not in parallel with more core that you have is an implementation detail. It is "just" preventing CPU throttling.

Also the sentence of "The work performed by this pool is not always CPU bound and may require waiting on resources or IO operations." is hard to digest, and readers without a specific background might wonder what to make out of it. IMHO it can be safely removed.

I won't object to those changes even though I think it doesn't add much actionable information. Please proceed if you think they will help users.

shawkins:
@ahus1 @mabartos @ryanemerson alternative 1: https://github.com/keycloak/keycloak/pull/42322 strips all extraneous text.

alternative 2: If that is not desirable, then I will make another attempt to refine what is here along the lines of https://github.com/keycloak/keycloak/pull/41725#issuecomment-3168535694 but probably creating a separate section on why we have 100 default connections and how you could tune that for production usage. I do think it makes sense to list limiting resource concerns wrt to threading - which includes the argon2 implementation detail.

ryanemerson:
I am happy with 1 to keep the docs simple, unless you think you can make the current docs more procedural/actionable.

> Some of our other recommendations, like keeping initial = min = max, could easily be implemented as the default for the production profile - but we don't do that either.

In an ideal world ^ would be implemented in a follow-up so that we satisfy the majority of users with no additional actions required on their behalf.

ahus1:
Closed in favor of other #42322

Issue #38660, opened 2025-04-03
Ldap federation seems to open and keep open a new thread/connection for each ldap request


yennor:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

ldap

### Describe the bug

Keycloak seems to open and keep open for a while a new connection to the ldap server for each request. At least in below scenario:

I'm using the keycloak-admin-client 26.0.4 witch keycloak 26.1.4.
There is a writable Ldap User Federation set up with openLdap 2.5.13+dfsg-5

I access all users doing basically that here.
Nothing will be written back to keycloak. It's only read access

```
for(var user: allUsers) {
    List<UserRepresentation> users = midwifeRealm.users().search(user.getUid(), true);
    if (users.isEmpty()) {
         // throw exception
    }
    userRepresentation = user.get(0);
    // do something 
    // or do nothing. for testing I've just run through the loop.
}
```
At one point the users.isEmpty() call fails with:
```
Apr 03 00:01:48 intranet java[820274]: jakarta.ws.rs.BadRequestException: HTTP 400 Bad Request
Apr 03 00:01:48 intranet java[820274]:         at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.handleErrorStatus(ClientInvocation.java:236)
Apr 03 00:01:48 intranet java[820274]:         at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.extractResult(ClientInvocation.java:216)
Apr 03 00:01:48 intranet java[820274]:         at org.jboss.resteasy.client.jaxrs.internal.proxy.extractors.BodyEntityExtractor.extractEntity(BodyEntityExtractor.java:59)
Apr 03 00:01:48 intranet java[820274]:         at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientInvoker.invokeSync(ClientInvoker.java:136)
Apr 03 00:01:48 intranet java[820274]:         at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientInvoker.invoke(ClientInvoker.java:103)
Apr 03 00:01:48 intranet java[820274]:         at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientProxy.invoke(ClientProxy.java:102)
Apr 03 00:01:48 intranet java[820274]:         at jdk.proxy2/jdk.proxy2.$Proxy279.search(Unknown Source)
[...]

PR #41726, opened 2025-08-07
fix: adding more ldap connection pool documentation

### LINKED ISSUE #39299: Improve docs, and possibly defaults, around ldap pooling
### Description

There isn't sufficient documentation around com.sun.jndi.ldap.connect.pool.maxsize https://www.keycloak.org/docs/latest/server_admin/index.html#_ldap_connection_pool - in general if it is smaller than the number of worker threads, you can introduce contention on the worker pool threads. 

The LDAPConfig connectionTimeout (com.sun.jndi.ldap.connect.timeout) can be used to relieve this contention, but it seems to have a not well-defined behavior when left unset and connection pooling is used - meaning that an explicit value is best.

JDBC pools default to a 5 second acquisition timeout, but we generally recommend that the number of JDBC connections available are at least equal to the number of worker threads.

It might also be mention that the usage of connectionTrace disables pooling.

### Discussion

_No response_

### Motivation

To ensure proper ldap pool configuration.

### Details

As seen on https://github.com/keycloak/keycloak/issues/38660 the user experienced several behaviors due to an underlying bug that were not well explained by our current documentation.

shawkins:
closes: #39299


docs/documentation/server_admin/topics/user-federation/ldap.adoc
```diff
@@ -179,11 +179,13 @@ The LDAP connection pool configuration is configured using the following system
...
 | `com.sun.jndi.ldap.connect.pool.prefsize` | The string representation of an integer that represents the preferred number of connections per connection identity that should be maintained concurrently
 | `com.sun.jndi.ldap.connect.pool.timeout` | The string representation of an integer that represents the number of milliseconds that an idle connection may remain in the pool without being closed and removed from the pool
 | `com.sun.jndi.ldap.connect.pool.protocol` | A list of space-separated protocol types of connections that may be pooled. Valid types are "plain" and "ssl"
 | `com.sun.jndi.ldap.connect.pool.debug` | A string that indicates the level of debug output to produce. Valid values are "fine" (trace connection creation and removal) and "all" (all debugging information)
+| `com.sun.jndi.ldap.connect.timeout` | The string representation of an integer that represents how long in milliseconds obtaining a connection should take. This is also applicable to wait times due to connection pool contention. 
+It is recommended to set this value explicitly to 5000 as the system default may be unlimited.
```
ahus1: Given your comment: 

> It seems to have a not well-defined behavior when left unset and connection pooling is used - meaning that an explicit value is best.

I'd suggest to go one step further and set this to 5000 if it is not set. Also there is no good way to set those when using the KC operator (AFAIK there is no supported parameter for Java options, only the unsupported template). 

On another note: Should we create at least SPI options for all those parameters to have a better support to set this via the Operator in a follow-up issue?

ahus1 (changes requested):
Thank you for this PR. See below for a suggestion to take this one step further.

mabartos:
@pedroigor As you were involved in the discussion here (https://github.com/keycloak/keycloak/issues/38660), could you please share your opinion on this? Thanks!

docs/documentation/server_admin/topics/user-federation/ldap.adoc
shawkins: > I'd suggest to go one step further and set this to 5000 if it is not set.

I'm good with that change.

> Also there is no good way to set those when using the KC operator (AFAIK there is no supported parameter for Java options, only the unsupported template).

They can also be set in a custom image, but yes the usability is lacking. This falls more under the category of work related to removing the unsupported templates - that issue already mentions the need to set arbitrary env variables.

> On another note: Should we create at least SPI options for all those parameters to have a better support to set this via the Operator in a follow-up issue?

I'm actually little fuzzy on the configuration expectations - at least as it is currently documented. My concern here was based upon the user experience with https://github.com/keycloak/keycloak/issues/38660 - but it could be that we shouldn't be recommending setting the system wide properties at all.

We are creating LDAPConfig objects from ComponentModels. These have some first-class options (like connectionTimeout), which map to the property keys documented here and an additional connection properties field where any property can be set. These supercede the system property settings on a per pool basis, but I don't seem them documented anywhere. 

Maybe we should expand documentation and defaults around per pool configurations, and just remove the section about system properties.

ahus1: Thank you for pointing me to the way LDAPConfig handles this. 

Let's change the default here (and add it to notable changes), and consider removing those system properties in a follow-up issue.

shawkins: https://github.com/keycloak/keycloak/pull/41762

shawkins:
@mabartos based upon your comments on #41762, I just merged the two PRs - it's benign to update the system property text even if we're going to try to remove it.

mabartos (dismissed):
@shawkins Works to me.

ahus1 (dismissed):
Thank you for this change!

ahus1:
Happy to merge it now, but holding back in case you want @pedroigor's review.

pedroigor (approved):
Thanks, @shawkins.

PR #41730, opened 2025-08-07
Detect and handle KC split brain clusters

### LINKED ISSUE #41561: Detect and handle KC split brain clusters
### Description

If there is a network split between KC nodes while the KC nodes can still talk to the database, their caches will get out-of-sync, and they might deliver possibly wrong results based on outdated caches. 

### Discussion

_No response_

### Motivation

Make it safe to run Keycloak even in split brain environments. 

### Details

If an isolated node can't talk to the database, it will already mark itself not-ready, and will then no longer serve traffic.

pruivo:
Closes #41561

- [x] docs?

### DIFF
--- a/docs/guides/observability/health.adoc
+++ b/docs/guides/observability/health.adoc
@@ -40,7 +40,11 @@ These endpoints respond with HTTP status `200 OK` on success or `503 Service Una
     "status": "UP",
     "checks": [
         {
-            "name": "Keycloak database connections health check",
+            "name": "Keycloak cluster health check",
+            "status": "UP"
+        },
+        {
+            "name": "Keycloak database connections async health check",
             "status": "UP"
         }
     ]
@@ -92,6 +96,10 @@ The table below shows the available checks.
 |Database
 |Returns the status of the database connection pool.
 |Yes
+
+|Cluster
+|Returns the status of the cluster (network partitions).
+|No
 
 |===
 
--- a/docs/guides/observability/health.adoc
+++ b/docs/guides/observability/health.adoc
@@ -108,4 +108,6 @@ use the `metrics-enabled` option as follows:
 
 <@kc.build parameters="--health-enabled=true --metrics-enabled=true"/>
 
+The cluster health check is only available for clustered setups when the cache transport stacks `jdbc-ping` or `jdbc-ping-udp` are used.
+
 </@tmpl.guide>
--- a/model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
+++ b/model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
@@ -307,6 +307,7 @@ public class DefaultInfinispanConnectionProviderFactory implements InfinispanCon
 
     @Override
     public boolean isClusterHealthy() {
+        clusterHealth.triggerClusterHealthCheck();
         return clusterHealth.isHealthy();
     }
 
--- a/model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
+++ b/model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
@@ -34,7 +34,6 @@ import org.infinispan.commons.io.StringBuilderWriter;
 import org.infinispan.commons.util.Version;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.factories.GlobalComponentRegistry;
-import org.infinispan.configuration.global.GlobalConfiguration;
 import org.infinispan.configuration.parsing.ParserRegistry;
 import org.infinispan.health.CacheHealth;
 import org.infinispan.manager.DefaultCacheManager;
@@ -309,6 +308,11 @@ public class DefaultInfinispanConnectionProviderFactory implements InfinispanCon
     public boolean isClusterHealthy() {
         clusterHealth.triggerClusterHealthCheck();
         return clusterHealth.isHealthy();
+    }
+
+    @Override
+    public boolean isClusterHealthSupported() {
+        return clusterHealth.isSupported();
     }
 
     private void addEmbeddedOperationalInfo(Map<String, String> info) {
--- a/model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
+++ b/model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
@@ -33,6 +33,7 @@ import org.infinispan.commons.configuration.io.ConfigurationWriter;
 import org.infinispan.commons.io.StringBuilderWriter;
 import org.infinispan.commons.util.Version;
 import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.factories.GlobalComponentRegistry;
 import org.infinispan.configuration.global.GlobalConfiguration;
 import org.infinispan.configuration.parsing.ParserRegistry;
 import org.infinispan.health.CacheHealth;
@@ -43,6 +44,7 @@ import org.keycloak.Config;
 import org.keycloak.cluster.ClusterEvent;
 import org.keycloak.cluster.ClusterProvider;
 import org.keycloak.connections.infinispan.remote.RemoteInfinispanConnectionProvider;
+import org.keycloak.infinispan.health.ClusterHealth;
 import org.keycloak.infinispan.util.InfinispanUtils;
 import org.keycloak.marshalling.KeycloakIndexSchemaUtil;
 import org.keycloak.marshalling.KeycloakModelSchema;
@@ -68,9 +70,8 @@ import org.keycloak.spi.infinispan.impl.embedded.CacheConfigurator;
 
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.AUTHENTICATION_SESSIONS_CACHE_NAME;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.CLIENT_SESSION_CACHE_NAME;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.CLUSTERED_CACHE_NAMES;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.CRL_CACHE_NAME;
-import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.KEYS_CACHE_DEFAULT_MAX;
-import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.KEYS_CACHE_MAX_IDLE_SECONDS;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.KEYS_CACHE_NAME;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.LOGIN_FAILURE_CACHE_NAME;
 import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.OFFLINE_CLIENT_SESSION_CACHE_NAME;
@@ -91,10 +92,9 @@ public class DefaultInfinispanConnectionProviderFactory implements InfinispanCon
     private Config.Scope config;
 
     private volatile EmbeddedCacheManager cacheManager;
-
     private volatile RemoteCacheManager remoteCacheManager;
-
     private volatile InfinispanConnectionProvider connectionProvider;
+    private volatile ClusterHealth clusterHealth;
 
     @Override
     public InfinispanConnectionProvider create(KeycloakSession session) {
@@ -173,9 +173,10 @@ public class DefaultInfinispanConnectionProviderFactory implements InfinispanCon
 
             this.remoteCacheManager = createRemoteCacheManager(keycloakSession);
             this.connectionProvider = InfinispanUtils.isRemoteInfinispan() ?
-                  new RemoteInfinispanConnectionProvider(cacheManager, remoteCacheManager, topologyInfo) :
-                  new DefaultInfinispanConnectionProvider(cacheManager, topologyInfo);
+                    new RemoteInfinispanConnectionProvider(cacheManager, remoteCacheManager, topologyInfo) :
+                    new DefaultInfinispanConnectionProvider(cacheManager, topologyInfo);
 
+            clusterHealth = GlobalComponentRegistry.componentOf(cacheManager, ClusterHealth.class);
             return connectionProvider;
         }
     }
@@ -304,10 +305,15 @@ public class DefaultInfinispanConnectionProviderFactory implements InfinispanCon
         return info;
     }
 
+    @Override
+    public boolean isClusterHealthy() {
+        return clusterHealth.isHealthy();
+    }
+
     private void addEmbeddedOperationalInfo(Map<String, String> info) {
         var cacheManagerInfo = cacheManager.getCacheManagerInfo();
         info.put("clusterSize", Integer.toString(cacheManagerInfo.getClusterSize()));
-        var cacheNames = Arrays.stream(InfinispanConnectionProvider.CLUSTERED_CACHE_NAMES)
+        var cacheNames = Arrays.stream(CLUSTERED_CACHE_NAMES)
                 .sorted()
                 .collect(Collectors.toCollection(LinkedHashSet::new));
         for (CacheHealth health : cacheManager.getHealth().getCacheHealth(cacheNames)) {
--- a/model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
+++ b/model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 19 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/health/KeycloakClusterReadyHealthCheck.java
```diff
@@ -0,0 +1,61 @@
...
+            failingSince.set(null);
+        } else {
+            builder.down();
+            Instant failingTime = failingSince.updateAndGet(KeycloakReadyHealthCheck::createInstanceIfNeeded);
+            builder.withData(FAILING_SINCE, DATE_FORMATTER.format(failingTime));
+        }
+        return Uni.createFrom().item(builder.build());
+    }
+
+
```
ryanemerson: Nitpick: Redundant whitespace.

model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/ClusterHealthFactory.java
```diff
@@ -0,0 +1,34 @@
...
+import org.keycloak.infinispan.health.impl.ClusterHealthImpl;
+
+@DefaultFactoryFor(classes = ClusterHealth.class)
+public class ClusterHealthFactory extends AbstractComponentFactory implements AutoInstantiableFactory {
+
+    @Override
+    public Object construct(String componentName) {
+        return new ClusterHealthImpl();
+    }
+
```
ryanemerson: Nitpick: Redundant whitespace.

model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/ClusterHealthFactory.java
```diff
@@ -0,0 +1,34 @@
...
+package org.keycloak.infinispan.module.factory;
+
+import org.infinispan.factories.AbstractComponentFactory;
+import org.infinispan.factories.AutoInstantiableFactory;
+import org.infinispan.factories.annotations.DefaultFactoryFor;
+import org.keycloak.infinispan.health.ClusterHealth;
+import org.keycloak.infinispan.health.impl.ClusterHealthImpl;
+
+@DefaultFactoryFor(classes = ClusterHealth.class)
+public class ClusterHealthFactory extends AbstractComponentFactory implements AutoInstantiableFactory {
```
ryanemerson, line 27: We should add `@SuppressWarnings("unused")`.

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,137 @@
...
+            logger.debug("Cluster health check disabled. Local mode");
+            return;
+        }
+        if (!(transport instanceof JGroupsTransport jgrp)) {
+            logger.debug("JGroups Transport not found. Unable to check cluster health.");
+            return;
+        }
+        KEYCLOAK_JDBC_PING2 ping = jgrp.getChannel().getProtocolStack().findProtocol(KEYCLOAK_JDBC_PING2.class);
+        if (ping == null) {
+            logger.debug("Stack 'jdbc-ping' not used. Unable to check cluster health.");
```
ryanemerson: Should we add this is as a warning log? It might encourage users to migrate to `jdbc-ping` as they get additional capabilities.

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,137 @@
...
+            r.trigger();
+        }
+    }
+
+    private record HealthRunner(KEYCLOAK_JDBC_PING2 discovery, Executor executor, Consumer<KEYCLOAK_JDBC_PING2> check) {
+
+        HealthRunner {
+            Objects.requireNonNull(discovery);
+            Objects.requireNonNull(executor);
+        }
```
ryanemerson: Do we need these checks given that it's only used internally? Could we just use `assert` in `#inject` or just not check as a NPE will be thrown eventually anyway.

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -242,4 +242,58 @@ protected Connection getConnection() {
...
+     *     <li>If multiple coordinators are found, it compares them and uses the coordinator with the lowest {@link UUID}</li>
+     *     <li>Finally, it compares if the coordinator is the same as the current view coordinator. If so, it returns {@link HealthStatus#HEALTHY}, otherwise {@link HealthStatus#UNHEALTHY}</li>
+     * </ul>
+     *
+     * @return The {@link HealthStatus}.
+     * @see HealthStatus
+     */
+    public HealthStatus isHealthy() {
+        try {
+            // maybe create an index, and a query to return coordinators only?
```
ryanemerson, line 265: I think this makes sense in theory, but is it overkill in reality given that Keycloak apparently struggles to scale well past several nodes?

I guess adding the index now means that's one less problem we need to solve if/when we try to improve horizontal scaling.

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -242,4 +242,58 @@ protected Connection getConnection() {
...
+                    .stream()
+                    .filter(PingData::isCoord)
+                    .map(PingData::getAddress)
+                    .sorted()
+                    .findFirst()
+                    .map(view.getCoord()::equals)
+                    .map(isCoordinatorInView -> isCoordinatorInView ? HealthStatus.HEALTHY : HealthStatus.UNHEALTHY)
+                    .orElse(HealthStatus.NO_COORDINATOR);
+        } catch (Exception e) {
+            // database failed?
```
ryanemerson, line 276: Debug log?

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -242,4 +242,58 @@ protected Connection getConnection() {
...
+     *     <li>Filters out non coordinator members</li>
+     *     <li>If no coordinator is found, it return {@link HealthStatus#NO_COORDINATOR}</li>
+     *     <li>If multiple coordinators are found, it compares them and uses the coordinator with the lowest {@link UUID}</li>
+     *     <li>Finally, it compares if the coordinator is the same as the current view coordinator. If so, it returns {@link HealthStatus#HEALTHY}, otherwise {@link HealthStatus#UNHEALTHY}</li>
+     * </ul>
+     *
+     * @return The {@link HealthStatus}.
+     * @see HealthStatus
+     */
+    public HealthStatus isHealthy() {
```
ryanemerson: As this doesn't return a boolean, I would rename to something like `healthStatus()` or `status()`.

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                    break;
+                case NO_COORDINATOR:
+                    logger.warn("Unable to check the cluster health because no coordinator has been found.");
+                    // fallthrough
+                case UNHEALTHY:
+                    logger.debug("Set cluster health status to unhealthy");
+                    healthy = false;
+                    break;
+                case ERROR:
+                    // do not change the state
```
ryanemerson: Redundant comment given the log below?

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    @Override
+    public boolean isHealthy() {
+        return healthy;
+    }
+
+    @Override
+    public void triggerClusterHealCheck() {
```
ryanemerson: `s/triggerClusterHealCheck/triggerClusterHealthCheck`

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        HealthRunner r;
+        if ((r = runner) != null) {
```
ryanemerson: Why do we need the assignment here, can't we just do:

```java
if (runner != null) runner.trigger();
```

model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
```diff
@@ -290,10 +291,16 @@ public Map<String, String> getOperationalInfo() {
         return info;
     }
 
+    @Override
+    public boolean isClusterHealthy() {
+        clusterHealth.triggerClusterHealCheck();
+        return clusterHealth.isHealthy();
```
ryanemerson, line 310: The default operator configuration for the StatefulSet uses a timeout of 10 seconds, so in this case it will be 20 seconds until the split-brain is detected by the readiness probe as the first call to this method will result in `true` being returned?

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,137 @@
...
+ * <p>
+ * The method {@link KEYCLOAK_JDBC_PING2#isHealthy()} contains the algorithm description. If it returns
+ * {@link HealthStatus#ERROR}, the healthy state does not change and relies on the Quarkus/Agroal readiness health
+ * check. But, if {@link HealthStatus#NO_COORDINATOR} is returned, the state will be changed to unhealthy. The should be
+ * a temporary situation as at least one coordinator must be present in the database table.
+ *
+ * @see KEYCLOAK_JDBC_PING2#isHealthy()
+ */
+@Scope(Scopes.GLOBAL)
+public class ClusterHealthImpl implements ClusterHealth {
```
ryanemerson: As this is tightly coupled with JDBC_PING2, how about `JdbcPingClusterHealthImpl`?

ryanemerson (changes requested):
We need to update `docs/guides/observability/health.adoc#available-checks` to document the health check.

model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/ClusterHealthFactory.java
pruivo: I don't have any warning in mine.

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
pruivo: The `assert` is disabled in production and, without this, the NPE will be thrown by a blocking thread, which will have no useful stack traces.

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
pruivo: Just some idea that crossed my mind, but I don't think it is worth it. I don't think a database would have problems querying a table with a single-digit number of rows (famous last words).

model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/ClusterHealthFactory.java
ryanemerson: It's because I hadn't compiled the branch. Once compiled it goes away as we have the generated code using it.

model/infinispan/src/main/java/org/keycloak/connections/infinispan/DefaultInfinispanConnectionProviderFactory.java
pruivo: It depends on the failure threshold. What you describe is the minimum time to mark it not ready.

model/infinispan/src/main/java/org/keycloak/infinispan/health/impl/ClusterHealthImpl.java
ryanemerson: Understood about `assert`, I just meant so we had this check in our repo. I don't think that's necessary tbh, as in the production code it's us that passes the inputs, so we know they're not `null`:

```java
        if (ping == null) {
            logger.debug("Stack 'jdbc-ping' not used. Unable to check cluster health.");
            return;
        }

        logger.debug("Cluster Health check available");
        init(ping, blockingManager.asExecutor("cluster-health"));
```

pruivo: The method is public, and users may do all kinds of crazy stuff.

ryanemerson: On a class in an `impl` package ... anyway, no harm leaving the check if you prefer. I just thought it was adding verbosity for little benefit.

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -790,16 +790,23 @@ void disableHealthCheckBean(BuildProducer<BuildTimeConditionBuildItem> removeBea
             ClassInfo disabledBean = index.getIndex()
                     .getClassByName(DotName.createSimple(KeycloakReadyHealthCheck.class.getName()));
             removeBeans.produce(new BuildTimeConditionBuildItem(disabledBean.asClass(), false));
-            ClassInfo clusterHealth = index.getIndex().getClassByName(DotName.createSimple(KeycloakClusterReadyHealthCheck.class));
-            removeBeans.produce(new BuildTimeConditionBuildItem(clusterHealth.asClass(), false));
+        } else if (!isHealthEnabled()) {
+            // health enabled / metrics disabled
```
ryanemerson: Is this possible, won't we always go into the first if branch as we're using `||` not `&&` ?

As we always invert these methods, I think we can make this easier to read by changing the methods to `isHealthDisabled` and `isMetricsDisabled`.

pruivo: done

model/infinispan/src/main/java/org/keycloak/jgroups/protocol/KEYCLOAK_JDBC_PING2.java
```diff
@@ -242,4 +242,59 @@ protected Connection getConnection() {
...
+                    .filter(PingData::isCoord)
+                    .map(PingData::getAddress)
+                    .sorted()
+                    .findFirst()
+                    .map(view.getCoord()::equals)
+                    .map(isCoordinatorInView -> isCoordinatorInView ? HealthStatus.HEALTHY : HealthStatus.UNHEALTHY)
+                    .orElse(HealthStatus.NO_COORDINATOR);
+        } catch (Exception e) {
+            // database failed?
+            log.debug("Failed to fetch the cluster members fromm the database.", e);
```
ahus1:
```suggestion
            log.warn("Failed to fetch the cluster members from the database.", e);
```

model/infinispan/src/test/java/org/keycloak/jgroups/protocol/ControlledJdbcPing.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.keycloak.jgroups.protocol;
+
+import java.util.List;
+
+import org.jgroups.Address;
+import org.jgroups.View;
+import org.jgroups.protocols.PingData;
+
+/*
```
ahus1:
```suggestion
/**
```

docs/guides/observability/health.adoc
```diff
@@ -40,7 +40,11 @@ These endpoints respond with HTTP status `200 OK` on success or `503 Service Una
     "status": "UP",
     "checks": [
         {
-            "name": "Keycloak database connections health check",
+            "name": "Keycloak cluster health check",
```
ahus1, line 43: The cluster check should only be present if JDBC_PING is enabled, otherwise it would trick people into a false confidence. Maybe CDI can help you here to produce this bean only when the right configuration is present. 

This should also be reflected in the docs both here in the health check, as well as in the single cluster HA guide where we would state that people should use JDBC_PING to detect such splitbrains.

ahus1 (changes requested):
Thank you for this pull request, it looks great for those running JDBC_PING. See below to polish the behavior and the docs for those who are not using it.

docs/guides/observability/health.adoc
pruivo: Infinispan configuration is a runtime option and not a build-time option. I don't want to go through that path.

ryanemerson (dismissed):

docs/guides/observability/health.adoc
ahus1: I found a way to allow a runtime configuration via CDI means. I pushed one more commit, please have a look.

pruivo: @ahus1 pushed a small change + javadoc, otherwise it looks good to me.

ryanemerson (approved):

ahus1 (approved):

PR #41733, opened 2025-08-07
fix: undeprecating legacy-observability-interface

### LINKED ISSUE #41303: Allow for health check on main interface

shawkins:
closes: #41303

This is the far simpler option for offering health checks on the main interface.

The guidance here isn't much different than what we have for using a hostname-admin - that is if you don't want users accessing /admin via hostname, then you have to block that at the proxy.

We could also consider renaming this option so that it doesn't use the term legacy.

### DIFF
--- a/docs/documentation/release_notes/topics/26_4_0.adoc
+++ b/docs/documentation/release_notes/topics/26_4_0.adoc
@@ -31,3 +31,9 @@ In order to support basic TLS termination (edge) deployments via the operator, y
 While access logs are often used for debugging and traffic analysis, they are also important for security auditing and compliance monitoring.
 
 For more information, see the https://www.keycloak.org/server/logging[Logging guide].
+
+= Legacy interface for metrics and health endpoints
+
+The `legacy-observability-interface` option is no longer deprecated. You may use that to make the `/health` and `/metrics` endpoints accessible by the main ports, which default to `8080` for HTTP and `8443` for HTTPS.
+
+Administrators using this option should take care at their proxy to block unintended access to the `/health` and `/metrics` endpoints.
--- a/docs/documentation/upgrading/topics/changes/changes-25_0_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-25_0_0.adoc
@@ -345,7 +345,6 @@ The `/health` and `/metrics` endpoints are accessible on the management port `90
 That means these endpoints are no longer exposed to the standard Keycloak ports `8080` and `8443`.
 
 In order to reflect the old behavior, use the property `--legacy-observability-interface=true`, which will not expose these endpoints on the management port.
-However, this property is deprecated and will be removed in future releases, so it is recommended not to use it.
 
 The management interface uses a different HTTP server than the default {project_name} HTTP server, and it is possible to configure them separately.
 Beware, if no values are supplied for the management interface properties, they are inherited from the default {project_name} HTTP server.
--- a/docs/guides/server/management-interface.adoc
+++ b/docs/guides/server/management-interface.adoc
@@ -52,9 +52,7 @@ If you want to disable exposing them on the management interface, set the {proje
 
 [WARNING]
 ====
-Exposing health and metrics endpoints on the default server is not recommended for security reasons, and you should always use the management interface.
-Beware, the `legacy-observability-interface` option is deprecated and will be removed in future releases.
-It only allows you to give more time for the migration.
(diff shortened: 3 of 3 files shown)


mabartos:
> We could also consider renaming this option so that it doesn't use the term legacy.

Yes, I'd rather go this way. As we've seen there's a demand/requirement for some deployments to use the functionality to have exposed management endpoints on the main server, we should provide it as "supported", so no `legacy` word inside. 

It could allow us to be even more verbose in the description of the property that is better to have it on a separate port, but if you want to use, you should make an appropriate actions to be safe (like block it on the proxy side). 

As health and metrics endpoint can be considered as "management" endpoints (exposed on mgmt interface right now), I think we can create a new option (some kind of rename) for the management category.

I thought something like `--http-management-separated` (default `false`) with the mapper containing (`.mapFrom('legacy-observability-interface'`) that could allows to prefer the new option, but if not provided, use the legacy-observability-interface. And for KC 27, remove the deprecated option and have supported new one :) 

We could also mention it in the docs for management interface. There might be some other management endpoints in the future with the same requirements - like having them exposed on the main server. So I'd say it's more tied to the mgmt interface.

@shawkins WDYT?

mabartos:
@shawkins I don't know exact users'/customers' requirements, but I believe there might some deployments that require having all mgmt endpoints exposed on the main server port. 

I would not be against supporting both variants - have all exposed on the main server + expose only health on main server (https://github.com/keycloak/keycloak/pull/40044). 

When all is exposed, we can add some warning log to the console and update docs to be visible that people need to do some additional security steps.

I think even the local accessible health checks by default on the main server make sense (https://github.com/keycloak/keycloak/pull/40044)

shawkins:
> I thought something like --http-management-separated (default false) with the mapper containing (.mapFrom('legacy-observability-interface') that could allows to prefer the new option, but if not provided, use the legacy-observability-interface. And for KC 27, remove the deprecated option and have supported new one :)

Based upon how quarkus is doing things we would just turn "http-management-enabled" into a proper configuration option - we could also validate when it is not enabled that you can't set things like the host, port, ssl, etc.

> There might be some other management endpoints in the future with the same requirements - like having them exposed on the main server. So I'd say it's more tied to the mgmt interface.

This is probably the best case for selective exposure as there's no chance for something unintentional down the road.

EDIT: another thought for just exposing everything on main is to that we could use a common prefix that would make it easier to exclude by default at the proxy. So rather than /health/readiness, when running on main it could be /insecure/health/readiness. The same could might even apply for the lb-health check.

> @shawkins I don't know exact users'/customers' requirements, but I believe there might some deployments that require having all mgmt endpoints exposed on the main server port.

I don't think we've seen that yet for metrics. There's been several requests for health though.

> I would not be against supporting both variants - have all exposed on the main server + expose only health on main server (https://github.com/keycloak/keycloak/pull/40044).

The only way I can think to bridge the two approaches is to give the user control over what is exposed on main:

http-management-on-main=\<regex of allowed end points>

But that puts on the hook for add the wiring for everything and some additional complexity.

Maybe /health (and specifically ready) is all that is really needed.

I've reopened the other pr as https://github.com/keycloak/keycloak/pull/41759 to give it clean slate.

vmuzikar:
We should not undeprecate the legacy option, its name would still imply it's something that will eventually go away.

Instead, I'd go for a simple new option, something like `http-management-health=false`. This granularity and also a possible future path if we'd need it for metrics too (`http-management-metrics=false`). I like the logic to be negative (setting something to false), it feels more like "disabling a security feature", IMHO.

> So rather than /health/readiness, when running on main it could be /insecure/health/readiness.

That feels a bit unnecessary for now, and might be even confusing (different path on main and management interface). If the user explicitly exposes health on main, they are expected they know what they're doing.

shawkins:
> Instead, I'd go for a simple new option, something like http-management-health=false. This granularity and also a possible future path if we'd need it for metrics too (http-management-metrics=false). I like the logic to be negative (setting something to false), it feels more like "disabling a security feature", IMHO.

This pr is the draft for putting all management stuff on main, not just health.

> That feels a bit unnecessary for now, and might be even confusing (different path on main and management interface). If the user explicitly exposes health on main, they are expected they know what they're doing.

That is to address the concern that it's just /health and /metrics currently, but there could be more added later and it would be better to have them handled by default rather than needing to read the release notes to figure out if anything additionally should be blocked. Alternatively do we have a definite allow list?

vmuzikar:
> rather than needing to read the release notes to figure out if anything additionally should be blocked

That's one of the main reasons I'd rather like to have a specific option for each thing that we allow to selectively expose on main instead of management.

shawkins:
Closing in preference of the more specific approach.

PR #41735, opened 2025-08-07
Use an OS-specific key for writing for maven cache

Pepo48:
Closes: #41664


mabartos (approved):
@Pepo48 Thanks. We can try it for one night? AFAIR, when CI is executed for PR, the build was always successful. Feel free to merge it if it persists until tomorrow.

PR #41750, opened 2025-08-08
Define default topologySpreadConstraints

### LINKED ISSUE #41729: Define default topologySpreadConstraints
### Description

#41666 defines anti-affinity policies to ensure that Kubernetes prefers to schedule pods on distinct zones and nodes. However, once a pod has been scheduled on all of the available zones/nodes it's still possible for pods to become unevenly distributed across zones/nodes.

We can leverage a default [`topologySpreadConstraints`](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#topologyspreadconstraints-field) to ensure that pods are evenly distributed across the zones/nodes.

The following policy should prefer pods to be evenly spread, whilst not preventing scheduling if the `maxSkew` is not achievable:

```yaml
topologySpreadConstraints:
- topologyKey: kubernetes.io/hostname
  maxSkew: 1
  whenUnsatisfiable: ScheduleAnyway
  labelSelector:
    matchLabels:
      app: keycloak
      app.kubernetes.io/managed-by: keycloak-operator
      app.kubernetes.io/component: server
- topologyKey: topology.kubernetes.io/zone
  maxSkew: 1
  whenUnsatisfiable: ScheduleAnyway
  labelSelector:
    matchLabels:
      app: keycloak
      app.kubernetes.io/managed-by: keycloak-operator
      app.kubernetes.io/component: server
```

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #41729

### DIFF
--- a/docs/guides/high-availability/single-cluster/concepts.adoc
+++ b/docs/guides/high-availability/single-cluster/concepts.adoc
@@ -23,7 +23,7 @@ The behaviour and high-availability guarantees of the {project_name} deployment 
 the {kubernetes} cluster. Typically, {kubernetes} clusters are deployed on a single availability-zone, however in order to
 increase fault-tolerance, it is possible to https://kubernetes.io/docs/setup/best-practices/multiple-zones/[deploy the cluster across multiple availability-zones].
 
-The {project_name} Operator defines the following anti-affinity strategy by default to prefer that {project_name} pods are
+The {project_name} Operator defines the following anti-affinity and topology spread constraints by default to prefer that {project_name} pods are
 deployed on distinct nodes and distinct availability-zones when possible:
 
 [source,yaml]
@@ -48,6 +48,25 @@ deployed on distinct nodes and distinct availability-zones when possible:
                   app.kubernetes.io/component: server
                   app.kubernetes.io/instance: keycloak
               topologyKey: kubernetes.io/hostname
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: "topology.kubernetes.io/zone"
+          whenUnsatisfiable: "ScheduleAnyway"
+          labelSelector:
+            matchLabels:
+              app: "keycloak"
+              app.kubernetes.io/managed-by: "keycloak-operator"
+              app.kubernetes.io/instance: "keycloak"
+              app.kubernetes.io/component: "server"
+        - maxSkew: 1
+          topologyKey: "kubernetes.io/hostname"
+          whenUnsatisfiable: "ScheduleAnyway"
+          labelSelector:
+            matchLabels:
+              app: "keycloak"
+              app.kubernetes.io/managed-by: "keycloak-operator"
+              app.kubernetes.io/instance: "keycloak"
+              app.kubernetes.io/component: "server"
 ----
 
 [IMPORTANT]
--- a/docs/guides/high-availability/single-cluster/concepts.adoc
+++ b/docs/guides/high-availability/single-cluster/concepts.adoc
@@ -23,31 +23,11 @@ The behaviour and high-availability guarantees of the {project_name} deployment 
 the {kubernetes} cluster. Typically, {kubernetes} clusters are deployed on a single availability-zone, however in order to
 increase fault-tolerance, it is possible to https://kubernetes.io/docs/setup/best-practices/multiple-zones/[deploy the cluster across multiple availability-zones].
 
-The {project_name} Operator defines the following anti-affinity and topology spread constraints by default to prefer that {project_name} pods are
+The {project_name} Operator defines the following topology spread constraints by default to prefer that {project_name} pods are
 deployed on distinct nodes and distinct availability-zones when possible:
 
 [source,yaml]
 ----
-      podAntiAffinity:
-        preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            podAffinityTerm:
-              labelSelector:
-                matchLabels:
-                  app: keycloak
-                  app.kubernetes.io/managed-by: keycloak-operator
-                  app.kubernetes.io/component: server
-                  app.kubernetes.io/instance: keycloak
-              topologyKey: topology.kubernetes.io/zone
-          - weight: 90
-            podAffinityTerm:
-              labelSelector:
-                matchLabels:
-                  app: keycloak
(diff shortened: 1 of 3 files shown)


docs/guides/high-availability/single-cluster/concepts.adoc
```diff
@@ -23,7 +23,7 @@ The behaviour and high-availability guarantees of the {project_name} deployment
 the {kubernetes} cluster. Typically, {kubernetes} clusters are deployed on a single availability-zone, however in order to
 increase fault-tolerance, it is possible to https://kubernetes.io/docs/setup/best-practices/multiple-zones/[deploy the cluster across multiple availability-zones].
 
-The {project_name} Operator defines the following anti-affinity strategy by default to prefer that {project_name} pods are
+The {project_name} Operator defines the following anti-affinity and topology spread constraints by default to prefer that {project_name} pods are
```
shawkins: Is there a need to set both the default affinity and topologySpreadConstraints? Isn't the default topologySpreadConstraint effectively dominant? If someone want to change scheduling behavior it seems like they would need to override both.

Also rather than, or in addition to, documenting the default(s) - should there be an option in the CR to add to the default, rather than overwrite it completely?

ryanemerson: > Is there a need to set both the default affinity and topologySpreadConstraints?

Good question. I think so because anti-affinity takes into account other scheduling scores, e.g. `NodeResourcesFit`.

Consider the following scenario*. We have 3 AZs each of which only has one node and the user sets `spec.replica: 6`. My understanding is that with the anti-affinity strategy, but no `topologySpreadConstraints` it's possible for the pods to be distributed as follows in the worst case scenario due to resource usage etc:

- AZ-a: [pod-0]
- AZ-b: [pod-1]
- AZ-c: [pod-2,pod-3,pod-4,pod-5]

Having the default `topologySpreadConstraints` will help the scheduler ensure that the distribution is closer to:

- AZ-a: [pod-0, pod-3]
- AZ-b: [pod-1, pod-4]
- AZ-c: [pod-2, pod-5]

The `whenUnsatisfiable: "ScheduleAnyway"` attribute means that ^ is not guaranteed and the worst-case scenario could still happen, but it is much less likely to. Having this "soft" rule also means that scheduling of pods won't be blocked if it's not possible. I think this makes sense as a default, as it's just an additional hint to the scheduler, similar to how we use the `preferredDuringSchedulingIgnoredDuringExecution` vs `requiredDuringSchedulingIgnoredDuringExecution` for the anti-affinity rules.

* I appreciate it's slightly contrived, but it illustrates the point for the worst-case scenario.



> Also rather than, or in addition to, documenting the default(s) - should there be an option in the CR to add to the default, rather than overwrite it completely?

How we differentiate between when a user wants to add to the defaults or overwrite them completely?

IMO we should keep this simple here and do the same as the affinity strategy. If not provided use the default, otherwise give the user full control.

shawkins: > My understanding is that with the anti-affinity strategy, but no topologySpreadConstraints it's possible for the pods to be distributed as follows in the worst case scenario due to resource usage etc:

The podAffinity scoring I believe is cumulative against all the other existing pods https://github.com/kubernetes/kubernetes/blob/790393ae92e97262827d4f1fba24e8ae65bbada0/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go#L202 - so that spread should not happen.

> My understanding is that with the anti-affinity strategy, but no topologySpreadConstraints it's possible for the pods to be distributed as follows in the worst case scenario due to resource usage etc

I'm asking if the topologySpreadConstraints is effectively dominant and a simpler expression of the desired scheduling - that is just set the default topologySpreadConstraints and not the default anti-affinity.

ryanemerson: > The podAffinity scoring I believe is cumulative against all the other existing pods https://github.com/kubernetes/kubernetes/blob/790393ae92e97262827d4f1fba24e8ae65bbada0/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go#L202 - so that spread should not happen.

Yes that's correct. I didn't realise the cumulative aspect when I created the issue. The final score will be impacted by other scoring plugins e.g. [NodeResourcesLeastAllocated](https://github.com/kubernetes/kubernetes/blob/790393ae92e97262827d4f1fba24e8ae65bbada0/pkg/scheduler/framework/plugins/noderesources/least_allocated.go), but given that we have a weight of 100 for our zone anti-affinity and 90 for nodes, I think you're right and the scenario I describe will be avoided.

If we keep the anti-affinity policies maybe we should increase the node weight to 100 as well?

> I'm asking if the topologySpreadConstraints is effectively dominant and a simpler expression of the desired scheduling - that is just set the default topologySpreadConstraints and not the default anti-affinity.

Apologies I misunderstood your first message. Yes I think you're correct, it seems when using the `preferred*` strategies and `whenUnsatisfiable: ScheduleAnyway`, there's no benefit to having both defined.

If people are in agreement I can update this PR to remove the default anti-affinity policies in favour of the simpler `topologySpreadConstraints` WDYT @shawkins @ahus1 @pruivo ?

shawkins: > If people are in agreement I can update this PR to remove the default anti-affinity policies in favour of the simpler topologySpreadConstraints WDYT @shawkins @ahus1 @pruivo ?

Sounds good to me.

ryanemerson: I have updated so that we only define `topologySpreadConstraints` and I have updated the "Known limitations" section of the single-cluster guide to add more details on how users can harden their deployments.

pruivo: > If people are in agreement I can update this PR to remove the default anti-affinity policies in favour of the simpler `topologySpreadConstraints` WDYT @shawkins @ahus1 @pruivo ?

From the comments, it sounds good to me.

pruivo (approved):

ahus1:
@shawkins - can you please review this PR when you have the time and let us know if it is good to be merged as is? Thanks!

shawkins (approved):
LGTM, thanks @ryanemerson

PR #39385, opened 2025-05-01
[OID4VCI] Limiting Client capabilities to OpenID Connect or OpenID4VCI

forkimenjeckayang:
This PR aims to limit client capabilities to OpenID Connect or OpenID4VCI . In order to acheive this, we have a new section in the front end under the advanced tab of the client settings configurable for OpenID for Verifiable Credentials. This setting is limited or will be available only to clients operating under the OpenID Connect or OpenID4VCI protocol.

Screenshot from 2025-04-24 13-32-49

Once toggled on and saved, we have an attribute assigned to that client oid4vci.enabled and set to true
This signifies or enables the client to be used to request verfiable credentials from Keycloak's credential endpoint. After client authentication this attribute is checked and if not available for the client, it cannot be used for requestig credentials.

Closes https://github.com/keycloak/keycloak/issues/32967

PR #41757, opened 2025-08-08
[OID4VCI] Extend realm UI configuration by OID4VCI attributes

### LINKED ISSUE #39533: [OID4VCI] Extend realm UI configuration by OID4VCI attributes
### Description

The realm can still have some basic attributes relevant for OpenID for Verifiable Credentials. 
Here are the attributes currently in use:

* `oid4vc.nonce-lifetime-seconds` <--- see pull-request https://github.com/keycloak/keycloak/pull/39479/
* `preAuthorizedCodeLifespanS`

forkimenjeckayang:
### Frontend (Admin UI)
- **OID4VCI Section Integration**: Moved OID4VCI attributes from standalone tab to a new section within the existing Tokens tab
- **ScrollForm Implementation**: Refactored Tokens tab to use `ScrollForm` component with jump-to-section navigation
- **Section Structure**: Added "OID4VCI attributes" as the fifth section in Tokens tab alongside General, Refresh tokens, Access tokens, and Action tokens
- **Form Validation**: Implemented proper validation with minimum value requirements (min: 30 seconds) and custom error messaging
- **Internationalization**: Added translation keys for all UI labels and help text
- **Feature Flag Integration**: Proper feature flag detection using `useIsFeatureEnabled(Feature.OpenId4VCI)`

### Backend Integration
- **Attribute Mapping**: Configured proper mapping between UI fields and realm attributes:
  - `vc.c-nonce-lifetime-seconds` - Controls nonce lifetime for OID4VCI flows
  - `preAuthorizedCodeLifespanS` - Controls pre-authorized code lifespan
- **Atomic Save**: All form fields across all sections in Tokens tab are saved together when the single "Save" button is clicked

### UI/UX Improvements
- **Jump-to-Section Navigation**: Added sidebar navigation with clickable links to jump between sections
- **Time Selector Components**: Utilizes existing TimeSelector components with units: seconds, minutes, hours
- **Form State Management**: Proper form state handling with save/revert functionality at the end of the tab
- **Validation Feedback**: Custom validation error messages with `useAlerts` for global error notifications

### Testing
- **Comprehensive Test Suite**: Updated `oid4vci-attributes.spec.ts` with full test coverage:
  - Feature flag detection (shows/hides section based on OID4VCI feature availability)
  - Jump-to-section navigation testing
  - Form field rendering and validation within Tokens tab
  - Save functionality with correct attribute mapping and TimeSelector unit conversion
  - Required field validation with custom error messages
  - Data persistence after page refres

### Removed Components
- **Standalone Tab**: Removed `RealmSettingsOid4vciTab.tsx` component
- **Tab Route**: Removed separate routing for OID4VCI tab
- **Tab Integration**: Removed OID4VCI tab from `RealmSettingsTabs.tsx`

## Configured Attributes

| UI Field | Realm Attribute | Description | Units |
|----------|----------------|-------------|-------|
| OID4VCI Nonce Lifetime | `vc.c-nonce-lifetime-seconds` | Lifetime of nonces used in OID4VCI flows | seconds, minutes, hours |
| Pre-authorized Code Lifespan | `preAuthorizedCodeLifespanS` | Lifespan of pre-authorized codes in OID4VCI flows | seconds, minutes, hours |

## Related
- Addresses ticket: `[OID4VCI] Extend realm UI configuration by OID4VCI attributes`
- Closes https://github.com/keycloak/keycloak/issues/39533

image


js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -1,38 +1,55 @@
...
+  // Check if the OID4VCI tab is visible
   const oid4vciTab = page.getByTestId("rs-oid4vci-attributes-tab");
-  await expect(oid4vciTab).toBeVisible();
+  const isVisible = await oid4vciTab.isVisible();
+
+  if (isVisible) {
+    await expect(oid4vciTab).toBeVisible();
+  } else {
+    await expect(oid4vciTab).toBeHidden();
+  }
```
IngridPuppet: Tests need to be deterministic. Using conditional branching here offers no clear reading of the deterministic behavior you expect based on the test configurations. In short, we are not certain of which branch is taken. Two questions:
- Is the OID4VCI feature enabled in the test setup?
- What behavior do you want to assert in line with that configuration?

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -1,38 +1,55 @@
...
+    await page.waitForSelector('[data-testid="rs-oid4vci-attributes-tab"]', {
+      state: "visible",
+      timeout: 10000,
+    });
+    await page.getByTestId("rs-oid4vci-attributes-tab").click();
+    return true;
+  } catch {
+    // Tab doesn't exist or is not visible - feature is disabled
+    test.skip();
+    return false;
```
IngridPuppet: Skipping the test makes it unreliable. Possibly all the tests relying on this logic are now silently skipped. It should be deterministic whether the OID4VCI feature is enabled or not, which I think it is.

https://github.com/keycloak/keycloak/blob/42675614411d30074f3a0ebc61e4dda50f9fa918/js/apps/keycloak-server/scripts/start-server.js#L61-L65

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -105,10 +129,13 @@ test("should validate required fields and minimum values", async ({ page }) => {
...
-    await saveButton.click();
+    // The save button should be enabled when form is dirty (has changes)
+    if (nonceValue !== "" || preAuthValue !== "") {
+      await expect(saveButton).toBeEnabled();
+    }
 
     if (shouldShowError) {
+      await saveButton.click();
       await expect(page.getByText(validationErrorText).first()).toBeVisible();
     }
```
IngridPuppet: Prefer not nesting assertions under if statements. Tests need to be deterministic to be readable. At this point, there should be a single answer to what are the values of `nonceValue`, `preAuthValue`, and `shouldShowError`.

IngridPuppet (changes requested):
Hello @forkimenjeckayang - I further checked the PR and added some comments related to test reliability. Please could you check?

forkimenjeckayang:
> Hello @forkimenjeckayang - I further checked the PR and added some comments related to test reliability. Please could you check?

Thank you for reviewing @IngridPuppet , due to the fact the `oid4vc-vci` feature is not enbaled for the `admin ui e2e` tests in the `js-ci`, that's why i have those conditional checks. But i have made some improvements. 
Would've gone for the use of `useIsFeatureEnabled` in the tests but React hooks can only be called inside React components. so I added a method to get the server info and check for the feature instead of trying to locate the tab before confirming feature availablity.
Please have another look at this.

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -48,8 +54,11 @@ test("OID4VCI tab visibility", async ({ page }) => {
 test("should render fields and save values with correct attribute keys", async ({
   page,
 }) => {
-  // Ensure OID4VCI tab is available and click it
-  await ensureOid4vciTabAvailable(page);
+  const oid4vciTab = await getOid4vciTab(page);
+  if (!oid4vciTab) return;
```
IngridPuppet: I understand the situation and agree with the move to directly rely on whether the feature is enabled or not. However, you could maybe revert to  `test.skip()` instead of `return` so it is clear in test reports that the test was skipped.

```suggestion
  const oid4vciTab = await getOid4vciTab(page);
  if (!oid4vciTab) test.skip();
```

forkimenjeckayang: Thanks again @IngridPuppet , that's done now

IngridPuppet (approved):
Thank you for considering my suggestions and clearing my concerns.

mposolda:
@forkimenjeckayang @IngridPuppet Thanks for the PR and review! Pr may require also review from our UI experts.

@ssilvert @jonkoops Anyone of you available to review this PR please? I can add that OID4VCI related settings should be visible just if the oid4vci feature is enabled, which looks that it is the case with this PR.

ssilvert:
@mposolda @forkimenjeckayang @IngridPuppet I'd rather not add a new tab to `Realm settings`.  It has too many already.

I don't know anything about OID4VCI.  Would it make sense to add these two fields to an existing tab?

IngridPuppet:
> @mposolda @forkimenjeckayang @IngridPuppet I'd rather not add a new tab to `Realm settings`. It has too many already.
> 
> I don't know anything about OID4VCI. Would it make sense to add these two fields to an existing tab?

Hello @ssilvert - We seize that the addition of more and more tabs is not sustainable. OID4VCI is about Keycloak issuing user credentials so other parties may verify user identities cryptographically; think of it as governments issuing identity cards. According to your experience, which existing tabs would you suggest as close to this narrative? Tokens?

ssilvert:
> > @mposolda @forkimenjeckayang @IngridPuppet I'd rather not add a new tab to `Realm settings`. It has too many already.
> > I don't know anything about OID4VCI. Would it make sense to add these two fields to an existing tab?
> 
> Hello @ssilvert - We seize that the addition of more and more tabs is not sustainable. OID4VCI is about Keycloak issuing user credentials so other parties may verify user identities cryptographically; think of it as governments issuing identity cards. According to your experience, which existing tabs would you suggest as close to this narrative? Tokens?

@mposolda Do you think this would make sense on the tokens tab?

mposolda:
@ssilvert @IngridPuppet I agree that `Tokens` tab might be good fit for those new attributes. Is it possible to introduce new section inside that tab and make sure that the section is displayed just when the `oid4vci` feature is enabled?

ssilvert:
> @ssilvert @IngridPuppet I agree that `Tokens` tab might be good fit for those new attributes. Is it possible to introduce new section inside that tab and make sure that the section is displayed just when the `oid4vci` feature is enabled?

Let's put it on the `Tokens` tab then.  

@forkimenjeckayang, if you want some extra credit then it would be really nice if you made the `Tokens` tab have "jump sections" like we have on the client details screen.

forkimenjeckayang:
> > @ssilvert @IngridPuppet I agree that `Tokens` tab might be good fit for those new attributes. Is it possible to introduce new section inside that tab and make sure that the section is displayed just when the `oid4vci` feature is enabled?
> 
> Let's put it on the `Tokens` tab then.
> 
> @forkimenjeckayang, if you want some extra credit then it would be really nice if you made the `Tokens` tab have "jump sections" like we have on the client details screen.

Thank you all for coming to a conclusion on that. I'll work on it and re-submit for review.

forkimenjeckayang:
@ssilvert and @mposolda 
The PR Is ready 
Please have another look. Thank you

js/apps/admin-ui/src/realm-settings/realm-settings-section.css
```diff
@@ -171,6 +171,11 @@ article.pf-v5-c-card.pf-m-flat.kc-email-connection
   font-weight: bold;
 }
 
+.kc-oid4vci-subtitle {
```
jonkoops: Ideally, we don't want to introduce any additional CSS, as this complicates the upgrade path for future PatternFly versions. See if you can accomplish this with existing elements or classes provided by PatternFly out of the box.

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -3545,4 +3545,11 @@ invalid_request_uri=Invalid request uri
 invalid_request_object=Invalid request object
 request_not_supported=Request not supported
 request_uri_not_supported=Request uri not supported
-registration_not_supported=Registration not supported
\ No newline at end of file
+registration_not_supported=Registration not supported
+oid4vciAttributes=OID4VCI attributes
+oid4vciAttributesSectionTitle=General
```
jonkoops: This label is introduced, but never used. Can likely be removed.

```suggestion
```

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -3545,4 +3545,11 @@ invalid_request_uri=Invalid request uri
...
+oid4vciNonceLifetime=OID4VCI Nonce Lifetime
+oid4vciNonceLifetimeHelp=The lifetime of the OID4VCI nonce in seconds.
+preAuthorizedCodeLifespan=Pre-Authorized Code Lifespan
+preAuthorizedCodeLifespanHelp=The lifespan of the pre-authorized code in seconds.
+formValidationError=Please ensure all OID4VCI attribute fields are filled and values are 30 seconds or greater.
```
jonkoops: This is a rather generic key for something that is specific to OID4VCI, perhaps make the key more specific?
```suggestion
oid4vciFormValidationError=Please ensure all OID4VCI attribute fields are filled and values are 30 seconds or greater.
```

js/apps/admin-ui/src/realm-settings/RealmSettingsTabs.tsx
```diff
@@ -235,6 +235,34 @@ export const RealmSettingsTabs = () => {
...
+        r.attributes["vc.c-nonce-lifetime-seconds"] =
+          r.attributes["vc.c-nonce-lifetime-seconds"].toString();
+      }
+      if (
+        r.attributes["preAuthorizedCodeLifespanS"] !== undefined &&
+        typeof r.attributes["preAuthorizedCodeLifespanS"] !== "string"
+      ) {
+        r.attributes["preAuthorizedCodeLifespanS"] =
+          r.attributes["preAuthorizedCodeLifespanS"].toString();
+      }
```
jonkoops: Rather than checking all the things a value might 'not' be, perhaps it would be easier to check for the type you are actually expecting? Also, it might be cleaner to handle this is a small loop with all the attributes pre-listed, to prevent repetition:

```suggestion
      const stringifiedAttributes = [
        "vc.c-nonce-lifetime-seconds",
        "preAuthorizedCodeLifespanS"
      ];

      for (const attr of stringifiedAttributes) {
        const value = r.attributes[attr];

        if (typeof value === "number") {
          r.attributes[attr] = value.toString();
        }
      }
```

js/apps/admin-ui/src/realm-settings/RealmSettingsTabs.tsx
```diff
@@ -235,6 +235,34 @@ export const RealmSettingsTabs = () => {
...
+      // Convert OID4VCI attributes to strings if they're not already
+      if (
+        r.attributes["vc.c-nonce-lifetime-seconds"] !== undefined &&
+        typeof r.attributes["vc.c-nonce-lifetime-seconds"] !== "string"
+      ) {
+        r.attributes["vc.c-nonce-lifetime-seconds"] =
+          r.attributes["vc.c-nonce-lifetime-seconds"].toString();
+      }
+      if (
+        r.attributes["preAuthorizedCodeLifespanS"] !== undefined &&
```
jonkoops: Why is this attribute formatted differently than the others?

js/apps/admin-ui/src/realm-settings/RealmSettingsTabs.tsx
```diff
@@ -235,6 +235,34 @@ export const RealmSettingsTabs = () => {
         ),
       );
     }
+    if (r.attributes) {
+      // Handle nested vc object created by React Hook Form
+      if (r.attributes["vc"] && typeof r.attributes["vc"] === "object") {
```
jonkoops: Instead of doing this after the fact, we usually flatten the keys to something when registering the form controls. See `convertAttributeNameToForm()`.

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
```diff
@@ -77,9 +84,30 @@ export const RealmSettingsTokensTab = ({
     defaultValue: false,
   });
 
-  return (
-    <PageSection variant="light">
-      <FormPanel title={t("general")} className="kc-sso-session-template">
+  // Hydrate form values from realm attributes for OID4VCI
+  useEffect(() => {
```
jonkoops: Why handle setting these properties in a side effect that then populates a different field? Can you not use the values directly? If they are the wrong type, perhaps the data loader would be a better place to convert them first (e.g `setupForm()` in `realm-settings/RealmSettingsTabs.tsx`), that way the conversion logic is co-located with the `save()` function that converts it back.

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
jonkoops, line 1: Thanks for including some tests. FYI, I have another PR up that cleans up our test bed, and we're working to enabling parralelism. Once #42489 has been merged, can you update your tests to make use of `createTestBed()` and ensure each of these tests can run in isolation?

js/apps/admin-ui/src/realm-settings/RealmSettingsTabs.tsx
forkimenjeckayang: You're absolutely right to question this inconsistency! 
The two attributes do follow different naming conventions:
- "vc.c-nonce-lifetime-seconds" - uses kebab-case with dots and hyphens (consistent with most Keycloak attributes)
- "preAuthorizedCodeLifespanS" - uses camelCase
Why this inconsistency exists:
Looking at the server-side code in `OID4VCIssuerEndpoint.java`, the attribute name is explicitly defined as:
```java
private static final String CODE_LIFESPAN_REALM_ATTRIBUTE_KEY = "preAuthorizedCodeLifespanS";
```
This suggests that `preAuthorizedCodeLifespanS` is the correct attribute name as expected by the server.

js/apps/admin-ui/src/realm-settings/RealmSettingsTabs.tsx
```diff
@@ -235,6 +261,19 @@ export const RealmSettingsTabs = () => {
...
+        "vc.c-nonce-lifetime-seconds",
+        "preAuthorizedCodeLifespanS",
+      ];
+      for (const attr of stringifiedAttributes) {
+        const value = r.attributes[attr];
+        if (typeof value === "number") {
+          r.attributes[attr] = value.toString();
+        }
+      }
+    }
```
jonkoops: I am kind of wondering if we need to convert these numbers to strings at all, I haven't really seen it for other attributes in other forms. Could you remove this code, and verify this breaks if you don't add this code? I believe the REST API might actually just convert them to a string for us.

js/apps/admin-ui/src/realm-settings/RealmSettingsTabs.tsx
```diff
@@ -191,6 +195,28 @@ export const RealmSettingsTabs = () => {
...
+      // Set the pre-authorized code lifespan value if it exists in attributes
+      if (r.attributes["preAuthorizedCodeLifespanS"]) {
+        setValue(
+          convertAttributeNameToForm(
+            "attributes.preAuthorizedCodeLifespanS",
+          ) as any,
+          r.attributes["preAuthorizedCodeLifespanS"],
+        );
+      }
+    }
```
jonkoops: AFAIK there should be no need for this logic, as long as `convertAttributeNameToForm()` is being called on the names in the form (i.e. `clients/add/SamlConfig.tsx`). The `convertToFormValues()` already handles the names of the attributes, so they are encoded in a non-object keyed format.

forkimenjeckayang: You're right @jonkoops

forkimenjeckayang:
Please have another look @jonkoops 
Thanks for the feedback

VinodAnandan:
@ssilvert & @jonkoops : Thank you very much for taking the time to review our pull request. The team has incorporated your suggestions. When you have a moment, could you please re-review the PR? Apologies for the follow-up, OAuth-SIG is eager to get this PR merged as soon as possible.

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
```diff
@@ -618,6 +627,92 @@ export const RealmSettingsTokensTab = ({
...
+                type="submit"
+                data-testid="tokens-tab-save"
+                isDisabled={!formState.isDirty}
+              >
+                {t("save")}
+              </Button>
+              <Button variant="link" onClick={() => reset(realm)}>
+                {t("revert")}
+              </Button>
+            </ActionGroup>
```
jonkoops: Perhaps this is a candidate to replace with a `FixedButtonGroup` component? See: `components/form/FixedButtonGroup.tsx`

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
```diff
@@ -618,6 +627,92 @@ export const RealmSettingsTokensTab = ({
...
+                  id="preAuthorizedCodeLifespan"
+                  min={30}
+                  units={["second", "minute", "hour"]}
+                  value={field.value}
+                  onChange={field.onChange}
+                  data-testid="pre-authorized-code-lifespan-s"
+                />
+              )}
+            />
+          </FormGroup>
```
jonkoops: Perhaps this could be shortened by using the `TimeSelectorControl`? See: `components/time-selector/TimeSelectorControl.tsx`

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+// Helper function to check if OID4VCI feature is enabled by checking server info
+const isOid4vciFeatureEnabled = async () => {
+  const serverInfo = await adminClient.getServerInfo();
+  const oid4vciFeature = serverInfo.features?.find(
+    (feature: any) => feature.name === "OID4VC_VCI",
+  );
+  const isEnabled = oid4vciFeature?.enabled || false;
+  return isEnabled;
+};
```
jonkoops: This code can be improved:

- Instead of falling back to `false` later on, provide a default value for the `features`, this reduces the amount of nullish checks required.
- Do not cast anything to `any` if you can avoid it, in this case the code even compiles fine without it.
- You can use [`Array.prototype.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) to check for a matching element in an array, no need to find and make assertions after.
- If you are using a variable only to directly return it, return it directly instead.

```ts
async function isOid4vciFeatureEnabled() {
  const serverInfo = await adminClient.getServerInfo();
  const features = serverInfo.features ?? [];

  return features.some(
    (feature) => feature.name === "OID4VC_VCI" && feature.enabled,
  );
};
```

Additionally, maybe this check is not needed at all? (see comments below)

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+// Helper function to handle feature flag logic and navigate to OID4VCI section
+const navigateToOid4vciSection = async (page: any) => {
```
jonkoops: Don't use `any` here.

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+});
+
+test("should render fields and save values with correct attribute keys", async ({
+  page,
+}) => {
+  const realm = await createTestBed();
+  await login(page, { to: toRealmSettings({ realm }) });
+
+  const isFeatureEnabled = await navigateToOid4vciSection(page);
+  if (!isFeatureEnabled) test.skip();
```
jonkoops: Under what circumstances would this be skipped? If you are worried about CI, please enable the feature required in the workflows, as well as any local dev servers.

See:
- `.github/workflows/js-ci.yml`
- `.github/workflows/stability-js-ci.yml`
- `js/apps/keycloak-server/scripts/start-server.js`

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  const isFeatureEnabled = await isOid4vciFeatureEnabled();
+  const tokensTab = page.getByTestId("rs-tokens-tab");
+
+  await tokensTab.click();
+
+  const oid4vciJumpLink = page.getByTestId("jump-link-oid4vci-attributes");
+
+  // Always assert based on feature flag state
+  if (isFeatureEnabled) {
```
jonkoops: Under what circumstances would this be occurring? If you are worried about CI, please enable the feature required in the workflows, as well as any local dev servers.

See:
- `.github/workflows/js-ci.yml`
- `.github/workflows/stability-js-ci.yml`
- `js/apps/keycloak-server/scripts/start-server.js`

js/apps/admin-ui/test/realm-settings/oid4vci-attributes.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  expect(realmData?.attributes?.["vc.c-nonce-lifetime-seconds"]).toBe("120");
+  expect(realmData?.attributes?.["preAuthorizedCodeLifespanS"]).toBe("300");
+});
+
+test("should persist values after page refresh", async ({ page }) => {
+  const realm = await createTestBed();
+  await login(page, { to: toRealmSettings({ realm }) });
+
+  const isFeatureEnabled = await navigateToOid4vciSection(page);
+  if (!isFeatureEnabled) test.skip();
```
jonkoops: Same here, let's not skip tests. This should be enabled at all times instead. As well as for all the other tests.

forkimenjeckayang:
Hello @jonkoops . I understand your concerns about tests being skipped. But i'm not sure I am in best possition to handle that at the level of the CI. As i mentioned earlier, the  `oid4vc-vci` feature is not enbaled for the `admin ui e2e` tests in the `js-ci`, that's why i have those conditional checks. Otherwise the CIs will always fail with the tests because they are trying to run in an enviroment where the feature is not enabled.

In addition to that , i've previously handled a FE ticket with same issues of the OID4VC feature, so in order not to break things and work with what we already have, I used same approach . See : https://github.com/keycloak/keycloak/pull/39385.

The tests have been ran locally both with feature enabled  and without it to ensure it behaves as expected. 
So please have another look

jonkoops:
@forkimenjeckayang Thanks for the update. What I meant is that there is no need to check if the feature is enabled, as we want to ensure that this feature is always enabled when running the tests. See https://github.com/keycloak/keycloak/pull/41757#discussion_r2344491411. Updating these files would require you to add the feature here where `--features=` is being passed.

.github/workflows/stability-js-ci.yml
```diff
@@ -119,7 +119,7 @@ jobs:
       - name: Start Keycloak server
         run: |
           tar xfvz keycloak-999.0.0-SNAPSHOT.tar.gz
-          keycloak-999.0.0-SNAPSHOT/bin/kc.sh start-dev --features=admin-fine-grained-authz:v2,transient-users &> ~/server.log &
+          keycloak-999.0.0-SNAPSHOT/bin/kc.sh start-dev --features=admin-fine-grained-authz:v2,transient-users,oid4vc-vci &> ~/server.log &
```
jonkoops: I see some discrepancies here, let's fix those while we're at it.

```suggestion
          keycloak-999.0.0-SNAPSHOT/bin/kc.sh start-dev --features=admin-fine-grained-authz:v2,transient-users,spiffe,oid4vc-vci &> ~/server.log &
```
```

jonkoops (approved):
LGTM, just a small comment.

jonkoops:
@forkimenjeckayang Thanks for the patience landing this PR. I think we're almost there. @mposolda @ssilvert can I ask you for a look as well?

jonkoops (approved):

forkimenjeckayang:
> @forkimenjeckayang Thanks for the patience landing this PR. I think we're almost there. @mposolda @ssilvert can I ask you for a look as well?

Thank you too @jonkoops  for the feedback

ssilvert (changes requested):
There is a problem with this code.

If I am on the tokens tab and I change some other field on the tab I can not save.  I get the error message, 'Please ensure all OID4VCI attribute fields are filled and values are 30 seconds or greater.'

This is because those attributes have never been filled in for a new realm.  You have effectively made these new fields required.

Is there a reasonable default value that could be provided?

jonkoops:
+1 these fields should not hold up submission. You can add a default value by specifying it in the controller, for example:

https://github.com/keycloak/keycloak/blob/f99c91291c61d1862b7eec6e5f9ff06b133964cb/js/apps/admin-ui/src/authentication/policies/OtpPolicy.tsx#L192-L201

If the field is allowed to be empty, the required attributes can be removed.

forkimenjeckayang:
Thanks again @jonkoops and @ssilvert 
We now have reasonable defaults
That should be resolved. We only get the error when values are below the minimun required.
Please have another look

ssilvert (approved):
Looks good now.  Approved.

jonkoops:
Thanks for taking this on @forkimenjeckayang!

PR #41759, opened 2025-08-08
fix: adds a way to expose health endpoints on the main interface

### LINKED ISSUE #41303: Allow for health check on main interface

shawkins:
closes: #41303

Reopening the previous pull request as a new one since there was a lot of now out-dated content on the pr.

This adds an option for /health/ready on main - we can easily add all the /health endpoints though.

### DIFF
--- a/docs/documentation/release_notes/topics/26_4_0.adoc
+++ b/docs/documentation/release_notes/topics/26_4_0.adoc
@@ -14,6 +14,11 @@ For more information, see the link:{adminguide_link}#_update-email-workflow[Upda
 
 There's a new option `http-management-scheme` that may be set to `http` to force the management interface to use HTTP rather than inheriting the HTTPS settings of the main interface.
 
+= Option to expose health endpoints on the main HTTP(S) ports
+
+With `health-enabled` set to true, you may set the `http-management-health-enabled` to `false` to indicate that health endpoints should be exposed on the main HTTP(s) ports instead of the
+management port. When this option is `false` you should block unwanted external traffic to `/health` at your proxy.
+
 = Additional context information for log messages (preview)
 
 You can now add context information to each log message like the realm or the client that initiated the request.
@@ -31,6 +36,7 @@ In order to support basic TLS termination (edge) deployments via the operator, y
 While access logs are often used for debugging and traffic analysis, they are also important for security auditing and compliance monitoring.
 
 For more information, see the https://www.keycloak.org/server/logging[Logging guide].
+
 
 = Supported passkeys
 
--- a/docs/guides/observability/health.adoc
+++ b/docs/guides/observability/health.adoc
@@ -11,6 +11,9 @@ includedOptions="health-enabled">
 {project_name} has built in support for health checks. This {section} describes how to enable and use the {project_name} health checks.
 The {project_name} health checks are exposed on the management port `9000` by default. For more details, see <@links.server id="management-interface" />
 
+When the `http-management-health-enabled` option is `false` the health endpoints will remain on the main HTTP(S) ports, rather than being exposed on the management port.
+When this option is `false` you should block unwanted external traffic to `/health` at your proxy.
+
 == {project_name} health check endpoints
 
 {project_name} exposes 4 health endpoints:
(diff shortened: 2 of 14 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/HealthOptions.java
```diff
@@ -4,7 +4,14 @@ public class HealthOptions {
...
             .category(OptionCategory.HEALTH)
-            .description("If the server should expose health check endpoints. If enabled, health checks are available at the '/health', '/health/ready' and '/health/live' endpoints.")
+            .description("If the server should expose health check endpoints. If enabled, health checks are available at the '/health', '/health/ready', '/health/live', and '/health/started' endpoints.")
+            .defaultValue(Boolean.FALSE)
+            .buildTime(true)
+            .build();
+
+    public static final Option<Boolean> HEALTH_ON_MAIN = new OptionBuilder<>("health-on-main", Boolean.class)
+            .category(OptionCategory.HEALTH)
+            .description("If the server should expose health checks om amain. If enabled, health checks are available at the '/health/ready' endpoint on the main interface.")
```
vit-zikmund: Hello @shawkins, just cruising by, watching this heroic effort unfold, I dare to leave with a question:
Why export only the `/health/ready` endpoint and not the usual `/health/*` bunch? I see `/ready` has been primarily requested by some users, but wouldn't it make sense to stay consistent both on main and the management port? I'd vote for consistency here.

Although, if there's yet another security concern (not already tackled by the "proxy warning") about adding the rest of the `/health` sub-routes (I fail to see it :thinking:), the option name should perhaps better reflect that (e.g. `health-ready-on-main`).

OT, there's also a typo in the description :wink:

shawkins: > Why export only the /health/ready endpoint and not the usual /health/* bunch? I see /ready has been primarily requested by some users, but wouldn't it make sense to stay consistent both on main and the management port? I'd vote for consistency here.

I don't have a strong opinion. The draft is to cover the minimum. As mentioned in the initial pr comment we can easily add everything. The only concern with doing that is we'll be on the hook for maintaining consistency - if new health endpoints should be added later, we'll need to add those on our side as well.

> Although, if there's yet another security concern (not already tackled by the "proxy warning") about adding the rest of the /health sub-routes (I fail to see it 🤔)

Something brought up elsewhere is that we can use start using a single prefix for anything added to the main interface - like /insecure/health... so that it would be even easier to block traffic.

> the option name should perhaps better reflect that (e.g. health-ready-on-main).

That is also possible, we'll just need more feedback on the direction first.

> OT, there's also a typo in the description 

Thanks, I'll clean this up before it gets out of draft.

vmuzikar (commented):
See https://github.com/keycloak/keycloak/pull/41733#issuecomment-3189500584

shawkins:
Took another pass at this and noticed there is already a quakus option to control putting the entire set of health checks on main. So rather than trying to duplicate the endpoints on both interfaces, we'll just expose this option.

Not mentioned in the doc changes is that you cannot disable this option when using the operator - as the probes will no longer work. I believe we probably need another issue under #39506 to address this (could be a doc change or we can add logic for scanning the additional options). There are other lingering operator considerations, such as if you use passthrough or reencrypt and require mTLS then the probes may not work either.

ryanemerson (dismissed):
> Not mentioned in the doc changes is that you cannot disable this option when using the operator - as the probes will no longer work. I believe we probably need another issue under https://github.com/keycloak/keycloak/issues/39506 to address this

+1 to at least documenting this limitation as it could cause confusion otherwise. We could maybe even add this to the CLI option description?

vmuzikar:
I'd like to review later this week.

vmuzikar (commented):
@shawkins Do you want to update this PR with the Operator changes and the doc change (how to use HTTP health endpoint without curl), or do you want to do it as a follow-up?

shawkins:
> @shawkins Do you want to update this PR with the Operator changes and the doc change (how to use HTTP health endpoint without curl), or do you want to do it as a follow-up?

I'll add that based upon our off-line conversation.

shawkins:
@vmuzikar should be good now.

vmuzikar (approved):
@shawkins Thanks for the update. LGTM.

I reckon we'll do the docs update for the container guide as a follow-up in a separate PR.

PR #41760, opened 2025-08-08
Fix: allow changing username in reset credentials flow when using separate username and password forms

adhamahmad:
Closes #40643

This change clears the user before setting it. Open to feedback if this approach could introduce security concerns.


adhamahmad:
This change intentionally alters the behavior covered by [`testDifferentUserProvidedInResetCredentialsFlow`](https://github.com/keycloak/keycloak/blob/main/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/ResetCredentialsAlternativeFlowsTest.java#L163).
 That test enforces the current policy of blocking user change in the reset credentials flow when a user was already set earlier.

My change clears the old user first before setting the new user in the context, so users can reset the password for a different account (e.g., a user with mulitiple accounts) without restarting the entire flow. 

I believe this is safe because the reset still requires email verification and it improves the UX also, however I am open to feedback if this change introduces security concerns.

PR #41762, opened 2025-08-08
fix: adding a default for ldap connection timeout

### LINKED ISSUE #39299: Improve docs, and possibly defaults, around ldap pooling

shawkins:
Alternative to #41726

If not specified on the LDAPConfig, we'll look for the system property, then use the default of 5000. The alternative is to skip looking for the system property at all, but that could be consider a breaking change given that we've document the usage of the property.

A separate follow-up issue will be needed to deprecate directly using the ldap system properties.

closes: #39299

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -99,6 +99,10 @@ When tracing is enabled, now also calls to other nodes of a {project_name} clust
 
 To disable this kind of tracing, set the option `tracing-infinispan-enabled` to `false`.
 
+=== LDAP Connection Timeout Default
+
+If no value is specified either on the LDAP configuration as the connectionTimeout or via the `com.sun.jndi.ldap.connect.timeout` system property, the default timeout
+will be 5 seconds. This will ensure that requests will see errors rather than indefinite waits in obtaining an LDAP connection from the pool or when making a connection to the LDAP server.
 
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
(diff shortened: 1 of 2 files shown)


mabartos (approved):
LGTM - I think it's nice to have the default here, and if it doesn't fit to users, they have approach how to set it to their needs.

mabartos (changes requested):
@shawkins Sorry, one more thing that'd need to be done. 

It'd be good to have it documented somewhere as seen in the alternative PR.

shawkins:
> @shawkins Sorry, one more thing that'd need to be done.
> 
> It'd be good to have it documented somewhere as seen in the alternative PR.

As mentioned on the other PR I don't see existing documentation on those ldap settings - maybe it's shown in a screenshot? I would consider the documentation and possible expansion of the LDAPConfig properties as part of the follow-up to deprecate using the system properties.

mabartos (approved):
I see. I thought to provide similar changes as in the second PR to be visible here that there's some default:

https://www.keycloak.org/docs/latest/server_admin/index.html#_ldap_connection_pool

https://github.com/keycloak/keycloak/pull/41726/files#diff-333ba797b5f96512644f18d30837675d8287290a0676bb2c89e48953651106bcR187

But works for me to have it done in a follow-up PRs. 

Approving now - decide on your own on how to proceed with the docs. Thanks

Issue #34397, opened 2024-10-28
Memory leak with v26.0.2 ?


mloiseleur:
### Before reporting an issue

- [X] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

core

### Describe the bug

Memory usage on keycloak is usually quite stable, we configured it with:

```yaml
            - name: JAVA_OPTS_APPEND
              value: -server -Xms196m -Xmx256m
[...]
          resources:
            requests:
              cpu: 100m
              memory: 512Mi
            limits:
              cpu: 1
              memory: 640Mi
```

We upgraded our keycloak instance on our kubernetes cluster to v26.0.2 last friday. No other change, just update image from `quay.io/keycloak/keycloak:26.0.1` to `quay.io/keycloak/keycloak:26.0.2`.

Since then, memory usage increase with a regular pace, without stopping:

![image](https://github.com/user-attachments/assets/87224e46-2612-442d-b676-ed92d48adfe4)

### Version

26.0.2

### Regression

- [X] The issue is a regression

### Expected behavior

Memory usage should be stable.

### Actual behavior

Memory usage increase without stop

### How to Reproduce?

Just replace image tag with latest version

### Anything else?

No info in the logs about any error.

mabartos:
@ahus1 @mhajas Hi guys! In the KC benchmark, you have some continuous Keycloak monitoring related to memory and other metrics per day, right? 

Did you encounter any difference since 26.0.2 was released?

ahus1:
We run tests on nightly, but not with such small memory usage, so we wouldn't notice a change there. 

I would recommend to look into the JVM metrics to see the what kind of memory area increased. See https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/running/metrics/jvm_metrics for the usual metrics jvm_memory_committed_bytes and jvm_memory_used_bytes 

 It would be good to know the Keycloak feature and options that are enabled in this environment.

mabartos:
@ahus1 Thanks.

@mloiseleur Could you please provide more information about your deployment to know more about potential issues?

mloiseleur:
@mabartos 

<details>

<summary>Here is the complete statefulset</summary>

```yaml
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: keycloak
  labels:
    app: keycloak
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: keycloak
[...]

PR #41768, opened 2025-08-09
fix: adding raw environment variables

### LINKED ISSUE #41766: Add the ability to set abritrary environment variables in Keycloak CR
### Description

Users can only set KC_ environment variables by using additional options. We should offer a more direct mechanism so as not to for the KC_ prefix - which is not used for kc.sh variables nor is necessary for placeholders that may be referenced by the server.

### Discussion

_No response_

### Motivation

To provide a way to set env variables instead of the unsupported pod template, or using the KC_ prefix, or baking the vars into a custom image.

### Details

_No response_

shawkins:
closes: #41766

It seems fine to align the name, env, with what is already in kubernetes rather than something more verbose like environment-variables

Also seems like this would be considered an advanced deployment feature - or do we even want to promote it in that context, and instead document it as a rarely used escape hatch?

### DIFF
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -100,7 +100,31 @@ spec:
 ----
 NOTE:  The name format of options defined in this way is identical to the key format of options specified in the configuration file.
        For details on various configuration formats, see <@links.server id="configuration"/>.
+       
+==== Custom environment variables
 
+You may find a need to set custom environment variables - such as for theme properties or `kc.[sh|bat]` script variables.
+The `spec.env` field of the Keycloak CR allows you to directly set any environment variable.
+Logic in the operator based upon looking for value of a particular setting does not consult `spec.env`,
+therefore do not use `spec.env` for anything that has a first-class configuration in the CR or may be specified as an `additionalOption`.
+
+Here's an example setting JAVA_OPTS_APPEND:  
+     
+[source,yaml]
+----
+apiVersion: k8s.keycloak.org/v2alpha1
+kind: Keycloak
+metadata:
+  name: example-kc
+spec:
+  ...
+  env:
+    - name: JAVA_OPTS_APPEND
+      value: -Djava.net.preferIPv6Addresses=true
+----
+
+Similar to `additionalOptions` you may specify either a value or reference a Secret.
+       
 === Secret References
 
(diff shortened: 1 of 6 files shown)


mabartos:
@shawkins IMO, it's a valid enhancement that we could address. 

> It seems fine to align the name, env, with what is already in kubernetes rather than something more verbose like environment-variables

`env` works for me. 

> Also seems like this would be considered an advanced deployment feature - or do we even want to promote it in that context, and instead document it as a rarely used escape hatch?

I'm ok to have a brief paragraph in the [Advanced configuration guide](https://www.keycloak.org/operator/advanced-configuration) where could be emphasized the usage of additionalOptions is the recommended way. 

The env mechanism would also potentially help to provide Heap settings(`JAVA_OPTS_KC_HEAP`) as mentioned in this [issue](https://github.com/keycloak/keycloak/issues/34397#issuecomment-3173739193).

shawkins:
> I'm ok to have a brief paragraph in the [Advanced configuration guide](https://www.keycloak.org/operator/advanced-configuration) where could be emphasized the usage of additionalOptions is the recommended way.

Added docs without too much cautionary text. I got the idea we had historically avoided allowing for this approach, so I wasn't sure how strongly we wanted to advise users around it.

mabartos (approved):
@shawkins It looks good to me, but I'd like to hear opinions of @Pepo48 and @vmuzikar.

vmuzikar:
I'd like to take a look later this week.

shawkins:
> I'd like to take a look later this week.

Sounds good. Basically our choices are:
- head down this path and duplicate much of the podTemplate functionality as part of the regular CR. We'll also likely need to support ConfigMaps (like #41796) and envFrom at some point.
- create a supported / sanitized podTemplate. Instead of unsupported.podTemplate, we'd have just podTemplate. We can use subclasses and or SchemaSwap/From annotations to limit it to only what is supported. Anything not supported would be pruned or fail validation.

Based upon our past approaches, I've been assuming that we'll do the former, but I'm open to the latter as well.

vmuzikar:
> Based upon our past approaches, I've been assuming that we'll do the former

+1, that's my thinking as well.

Exposing sanitized podTemplate feel a bit counter intuitive to me, we could say most of the current first class citizen fields could be put under the podTemplate. Also I'd be concerned we miss sanitizing some fields we don't want to support and consider security sensitive.

vmuzikar (approved):
LGTM, thanks.

PR #41771, opened 2025-08-10
fix: removing script logic for determining if a build is necessary

### LINKED ISSUE #23972: Improve handling config options in scripts preventing re-augmentation
### Description

Follow up of #22585

In the `kc.sh` and `kc.bat` scripts, we check configuration options(`help`, `build`, `show-config`,...) which should prevent executing a reaugmentation in an additional process. As these terminal operations might proliferate, the maintainability might be worse in that case. Even for now, it'd be good to shift the logic to the application code.

We could potentially remove these checks from the script and use ProcessBuilder inside the codebase to execute the additional process required before the application process itself. We'd have better control over it.

shawkins:
closes: #23972

### DIFF
--- a/quarkus/dist/src/main/content/bin/kc.bat
+++ b/quarkus/dist/src/main/content/bin/kc.bat
@@ -207,9 +207,6 @@ set KC_HOME_DIR=%KC_HOME_DIR:\=/%
 rem The property 'java.util.concurrent.ForkJoinPool.common.threadFactory' is set here, as a Java Agent or enabling JMX might initialize the factory before Quarkus can set the property in JDK21+.
 set JAVA_RUN_OPTS=-Djava.util.concurrent.ForkJoinPool.common.threadFactory=io.quarkus.bootstrap.forkjoin.QuarkusForkJoinWorkerThreadFactory %JAVA_OPTS% -Dkc.home.dir="%KC_HOME_DIR%" -Djboss.server.config.dir="%DIRNAME%..\conf" -Dkeycloak.theme.dir="%DIRNAME%..\themes" %SERVER_OPTS% -cp %CLASSPATH_OPTS% io.quarkus.bootstrap.runner.QuarkusEntryPoint %CONFIG_ARGS%
 
-set OPTIMIZED_OPTION=--optimized
-set HELP_LONG_OPTION=--help
-set BUILD_OPTION=build
 set IS_HELP_SHORT=false
 
 echo %CONFIG_ARGS% | findstr /r "\<-h\>" > nul
@@ -223,21 +220,16 @@ if "%PRINT_ENV%" == "true" (
     echo Using JAVA_RUN_OPTS: !JAVA_RUN_OPTS!
 )
 
-set START_SERVER=true
+"%JAVA%" !JAVA_RUN_OPTS!
 
-if "!CONFIG_ARGS:%OPTIMIZED_OPTION%=!"=="!CONFIG_ARGS!" if "!CONFIG_ARGS:%BUILD_OPTION%=!"=="!CONFIG_ARGS!" if "!CONFIG_ARGS:%HELP_LONG_OPTION%=!"=="!CONFIG_ARGS!" if "%IS_HELP_SHORT%" == "false" (
-
-    "%JAVA%" -Dkc.config.build-and-exit=true !JAVA_RUN_OPTS!
-
-    if not !errorlevel! == 0 (
-        set START_SERVER=false
-    )
-
-    set JAVA_RUN_OPTS=-Dkc.config.built=true !JAVA_RUN_OPTS!
+rem only exit code 10 means that implicit reaugmentation occurred and a relaunch is needed
+if not !errorlevel! == 10 (
+     exit /B !errorlevel!
 )
 
-if "%START_SERVER%" == "true" (
-    "%JAVA%" !JAVA_RUN_OPTS!
-)
+set JAVA_RUN_OPTS=-Dkc.config.built=true !JAVA_RUN_OPTS!
+
+"%JAVA%" !JAVA_RUN_OPTS!
+exit /B !errorlevel!
 
 :END
--- a/quarkus/dist/src/main/content/bin/kc.sh
+++ b/quarkus/dist/src/main/content/bin/kc.sh
@@ -50,7 +50,6 @@ esceval() {
     printf '%s\n' "$1" | sed "s/'/'\\\\''/g; 1 s/^/'/; $ s/$/'/"
 }
 
-PRE_BUILD=true
 while [ "$#" -gt 0 ]
 do
     case "$1" in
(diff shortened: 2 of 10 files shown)


shawkins:
I have in the past tried to take similar changes even a little further - so that if !requiresReAugmentation() then we contine to launch the actual command. It is possible, but it would require re-initing the configuration as the effective runtime profile may be different than the build-time one.

shawkins:
I should mention it may be possible to eliminate the system property for the build check, but it was easier to leave it in place for now.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -162,7 +162,7 @@ public <T> T set(T value) {
                     currentCommand = ac;
                 }
             }
-            initConfig(currentCommand);
+            initConfig(cliArgs, currentCommand);
 
             if (isRebuildCheck()) {
```
mabartos: @shawkins `kc.config.build-and-exit` does not even exist, right? So, `isRebuildCheck()` is always false.

quarkus/dist/src/main/content/bin/kc.sh
```diff
@@ -177,9 +171,12 @@ if [ "$PRINT_ENV" = "true" ]; then
   echo "Using JAVA_RUN_OPTS: $JAVA_RUN_OPTS"
 fi
 
-if [ "$PRE_BUILD" = "true" ]; then
-  eval "'$JAVA'" -Dkc.config.build-and-exit=true "$JAVA_RUN_OPTS" || exit $?
+eval "'$JAVA'" "$JAVA_RUN_OPTS"
+status=$?
+if [ $status = 10 ]; then
```
mabartos, line 177: Could you please add a simple comment about the 10 exit code and when it happens?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: Unfortunately it still exists. It is being set as a side effect of initConfig.

quarkus/dist/src/main/content/bin/kc.sh
shawkins: Added comments.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
vmuzikar: I wonder if we could take a different approach here to completely get rid of `kc.config.build-and-exit` as it semantically doesn’t make sense anymore and is just a remnant of the past more than anything else. Maybe at least rename it?

shawkins: Unfortunately there's a fair amount of additional work in removing the static nature of this.

> Maybe at least rename it?

I'm fine with having it match the existing method names - kc.config.rebuild-check

shawkins: @vmuzikar updated the property name.

vmuzikar: Ok, works for me. Maybe we could also add some comment here why is that still needed.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/Environment.java
```diff
@@ -37,6 +37,7 @@
 
 public final class Environment {
 
+    public static final String KC_CONFIG_REBUILD_CHECK = "kc.config.build-and-exit";
```
vmuzikar, line 40: Should this match the naming with the methods?
```suggestion
    public static final String KC_CONFIG_REBUILD_CHECK = "kc.config.rebuild-check";
```

shawkins: > Should this match the naming with the methods?

Sure, it can be changed as well. Ideally it's just an internal detail - I'll remove the usage of the system property at some point.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -162,7 +162,7 @@ public <T> T set(T value) {
...
                 // build command should be available when running re-aug
```
vmuzikar: I wonder if this means that we're always including build commands in start and start-dev? We're setting rebuild check always when config is initialized.

shawkins: This is the same intended behavior as before. If the command does not have --optimized, then yes the build options will be added.

However after reviewing that more, it is not needed. If we specifically add those args for the rebuild check, then the command is run again with the same args and those options aren't present we'd see an error. So we can just remove this and rely upon the IncludeOptions.

Also https://github.com/keycloak/keycloak/pull/41886 will make it so the cliArgs are not needed for initConfig.

Pepo48 (dismissed):
I added comment above. Otherwise LGTM. I tested the behaviour on my Windows machine and everything worked as expected. Thanks, @shawkins!

quarkus/dist/src/main/content/bin/kc.bat
```diff
@@ -225,17 +222,15 @@ if "%PRINT_ENV%" == "true" (
...
+rem only exit code 10 means that implicit reaugmentation occurred and a relaunch is needed
+if not !errorlevel! == 10 (
+    set START_SERVER=false
 )
 
+set JAVA_RUN_OPTS=-Dkc.config.built=true !JAVA_RUN_OPTS!
+
 if "%START_SERVER%" == "true" (
     "%JAVA%" !JAVA_RUN_OPTS!
 )
```
Pepo48: Just a nitpick:
```suggestion

"%JAVA%" !JAVA_RUN_OPTS!

rem only exit code 10 means that implicit reaugmentation occurred and a relaunch is needed
if not !errorlevel! == 10 (
    exit /B !errorlevel!
)

set JAVA_RUN_OPTS=-Dkc.config.built=true !JAVA_RUN_OPTS!

"%JAVA%" !JAVA_RUN_OPTS!
exit /B !errorlevel!

```

I guess that this mirrors kc.sh's logic better and it avoids any difference caused by setting the flag earlier. It also makes exit behavior explicit on Windows.

Pepo48: I had some difficulties to mark the intended range of code for the suggestion, so just to be clear, here's what I'm suggesting:
image

shawkins: Thank you @Pepo48 applied your suggested edit.

mabartos (approved):
@shawkins LGTM, nice work!

Pepo48 (approved):

PR #41781, opened 2025-08-11
Do not remove the recaptcha form data because it is useless

rmartinc:
Closes #41148

Just deleting the line that removes the captcha form data from the map. The form data is retrieved from quarkus using the the method [getDecodedFormParameters](https://github.com/keycloak/keycloak/blob/26.2.7/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/resteasy/QuarkusHttpRequest.java#L66), and the map is always rebuilt. So removing the parameter is useless as it will be returned in the same way when calling the method again. As commented in the issue, I think the only way of returning an immutable map is when it is empty (EMPTY_FORM_PARAM). If you think that we should always return an immutable (or mutable) map just let me know.

I was waiting for @derlin but it seems that he is busy. I tested manually with google captcha v3 and it works as before removing the line.


mposolda (approved):
@rmartinc Thanks

PR #41796, opened 2025-08-11
enhance: adding the ability to set truststores via configmaps

### LINKED ISSUE #34114: Operator: Support ConfigMaps for `Keycloak.spec.truststores`
### Description

Currently, the `spec.truststores` only accepts secrets, however, it's very common to distribute trust bundles with ConfigMaps (see Motivation for more details)

```
$ k explain keycloak.spec.truststores
GROUP:      k8s.keycloak.org
KIND:       Keycloak
VERSION:    v2alpha1

FIELD: truststores <map[string]Object>

DESCRIPTION:
    In this section you can configure Keycloak truststores.
    
FIELDS:
  name	<string>
    Not used. To be removed in later versions.

  secret	<Object> -required-
    <no description>
```

### Discussion

_No response_

### Motivation

Trust manager or OpenShift distribute trust bundles in ConfigMaps (as they are not sensitive), this enhances overall cluster security as administrators wouldn't need to give trust manager or any other trust bundle distributor cluster-wide access to secrets, see:

- https://cert-manager.io/docs/trust/trust-manager/
- https://docs.openshift.com/container-platform/4.17/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate

### Details

Implementation would be very similar to the existing `Keycloak.spec.truststores.secret` field

shawkins:
closes: #34114

This removes the Required annotation from the secret field, but does not add validation that only one secret or configMap should be set - it is mentioned in the documentation for the the fields. Later we should be able to add a oneOf restriction to the CRD via the josdk / fabric8.

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
@@ -30,7 +30,6 @@ import io.fabric8.kubernetes.api.model.VolumeBuilder;
 import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
 import io.fabric8.kubernetes.api.model.apps.StatefulSet;
 import io.fabric8.kubernetes.api.model.apps.StatefulSetBuilder;
-import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.config.informer.Informer;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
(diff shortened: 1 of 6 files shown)


operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -183,7 +188,7 @@ public StatefulSet desired(Keycloak primary, Context<Keycloak> context) {
         };
     }
 
-    private void configureCache(StatefulSet deployment, Container kcContainer, CacheSpec spec, KubernetesClient client, WatchedResources watchedResources) {
+    private void configureCache(StatefulSet deployment, Container kcContainer, CacheSpec spec, KubernetesClient client, WatchedResources watchedResources, TreeSet<String> allConfigMaps) {
```
vmuzikar: This does not need `watchedResources`, no?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -204,33 +209,53 @@ private void configureCache(StatefulSet deployment, Container kcContainer, Cache
 
             deployment.getSpec().getTemplate().getSpec().getVolumes().add(0, volume);
             kcContainer.getVolumeMounts().add(0, volumeMount);
-
-            // currently the only configmap we're watching
-            watchedResources.annotateDeployment(List.of(configFile.getName()), ConfigMap.class, deployment, client);
```
vmuzikar, line 207: Should've added the `configFile` to the `allConfigMaps`?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
shawkins: It should be there - we're missing a test for that. Corrected and expanded the existing test case.

vmuzikar: Thanks for the update.

It would seem that actually neither `client` nor `watchedResources` is used in `configureCache`.

shawkins: Sorry, didn't initially grok you meant the parameters. Removed now.

vmuzikar (approved):
LGTM, thank you.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
vmuzikar: Sorry, I should've been more clear.

PR #41798, opened 2025-08-11
fix: warn instead of an error if optimized provider timestamps change

shawkins:
closes: #41268

Decided that we could live with a default behavior of warning base upon the existing KC_RUN_IN_CONTAINER usage.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
@@ -86,6 +86,8 @@ import picocli.CommandLine.ParseResult;
 
 public class Picocli {
 
+    static final String PROVIDER_TIMESTAMP_ERROR = "A provider JAR was updated since the last build, please rebuild for this to be fully utilized.";
+    static final String PROVIDER_TIMESTAMP_WARNING = "A provider jar has a different timestamp than when the optimized container image was created. If you are changing provider jars after the build, you must run another build to properly account for those modifications.";
     static final String KC_PROVIDER_FILE_PREFIX = "kc.provider.file.";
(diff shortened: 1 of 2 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -235,7 +238,15 @@ public void validateConfig(List<String> cliArgs, AbstractCommand abstractCommand
             checkChangesInBuildOptions((key, oldValue, newValue) -> {
                 if (key.startsWith(KC_PROVIDER_FILE_PREFIX)) {
                     if (timestampChanged(oldValue, newValue)) {
-                        throw new PropertyException("A provider JAR was updated since the last build, please rebuild for this to be fully utilized.");
+                        if (warnedTimestampChanged
+                                || Configuration.getOptionalBooleanKcValue("run-in-container").orElse(false)) {
```
vmuzikar: Just a nitpick for better readability, do we need the condition for `warnedTimestampChanged`?
```suggestion
                        if (Configuration.getOptionalBooleanKcValue("run-in-container").orElse(false)) {
```

shawkins: I just using the that prevent additional configuration calls as if 1 provider fails they likely all will. It's just a small optimization, so if you don't like it, it can go.

shawkins: Actually the logic wasn't correct - we shouldn't skip the error if the jar is missing or added. See if the version reads better or if you still want to remove the warnedTimestampChanged

vmuzikar (approved):
LGTM now, thanks.

I reckon it'd be good to have this backported to 26.3.

PR #41800, opened 2025-08-11
fix: adds simple log scraping to error state detection

### LINKED ISSUE #21816: Expose Keycloak config errors in the Keycloak CR status field
### Description

At startup, Keycloak performs several config validations ensuring all required options are specified and that the options values have correct format. In case the validations fail, an exception is thrown. Exceptions are also thrown in other misconfiguration cases such as wrong DB credentials.

The Operator should be able to expose this kind of errors and expose them in the CR's status.

We should avoid implementing server config validations at the Operator level as the Operator does not know what config options has been already preset in a Keycloak custom image. Server config validations need to be done at the server level.

### Discussion

_No response_

### Motivation

_No response_

### Details

Implementation requires discussion.

We might need to try and interpret the server log and look for known exception types. Other way might be to send some message with failure details to the Operator (a new log handler?).

We might also need to improve the exceptions format (and possibly types) for the startup failures due to misconfiguration.

shawkins:
closes: #21816

Rather than creating a standalone set of validation logic, or running a specialized error-detection job, it seems straight-forward just to show the last snippet of the log. In most cases with configuration errors, even with verbose enabled, we will show them as simple output at the end of the log.

The downsides of this approach:
- are there things in the log that we don't want a Keycloak CR reader to see? Ideally not, but we may need to put further guidence that any anyone who can read a Keycloak CR should be namespace level admin.
- We'll poll on every status update when there is pod in an error state for the end of the log
- Could just add more docs on getting pod logs instead - but that may involve a broader obserability story as well.

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
@@ -22,6 +22,7 @@ import io.fabric8.kubernetes.api.model.ContainerStatus;
 import io.fabric8.kubernetes.api.model.PodSpec;
 import io.fabric8.kubernetes.api.model.PodStatus;
 import io.fabric8.kubernetes.api.model.apps.StatefulSet;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.readiness.Readiness;
 import io.fabric8.kubernetes.client.utils.KubernetesResourceUtil;
 import io.fabric8.kubernetes.client.utils.Serialization;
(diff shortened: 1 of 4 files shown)


vmuzikar (commented):
Would it make sense to try and scrap the latest exception (excl. the stack trace) instead of the raw log?

shawkins:
> Would it make sense to try and scrap the latest exception (excl. the stack trace) instead of the raw log?

Unfortunately I don't think we know exactly what that will look like unless we add some additional formatting - or require json logs for example. Also most pre-quarkus output goes directly to stdout or stderr - and we don't have a clear differentiation of that content in the log either.

vmuzikar (approved):
What I originally had in mind was some pattern search for Java Exceptions, but it's truth that in some cases we just print out an error directly, it's not even an exception. Let's not overthink it, IMHO it's good as is for now.

> are there things in the log that we don't want a Keycloak CR reader to see?

I don't think so. If there were, it'd be a vulnerability regardless of this Operator use case.

PR #41815, opened 2025-08-12
Change naming for disabling additional datasource

### LINKED ISSUE #40761: Change naming for disabling additional datasource

mabartos:
- Closes #40761

Reference to: https://github.com/keycloak/keycloak/pull/40762

> @mabartos wrote: _Closing this PR as the options for additional datasources might need some more refinements and possibly the best approach for that would be to enable the datasource at build time, but having the possibility to deactivate in runtime. I'll provide some more details later._

We have a different mechanism for enabling datasources.

@vmuzikar I believe we can proceed with this PR.

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
@@ -101,10 +101,10 @@ public class DatabaseOptions {
             .defaultValue(10000)
             .build();
 
-    public static final Option<Boolean> DB_ACTIVE_DATASOURCE = new OptionBuilder<>("db-active-<datasource>", Boolean.class)
+    public static final Option<Boolean> DB_ENABLED_DATASOURCE = new OptionBuilder<>("db-enabled-<datasource>", Boolean.class)
             .category(OptionCategory.DATABASE_DATASOURCES)
             .defaultValue(true)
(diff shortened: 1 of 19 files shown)


mabartos:
@vmuzikar Are you ok with this as we discussed it on our weekly meeting? Thanks

vmuzikar (approved):
@mabartos Sure, LGTM.

The only thing I'm wondering about is whether we should mention this change in the migration guide. Though it's still preview and the feature was not really documented anywhere. Approving regardless, works for me as is.

mabartos:
> The only thing I'm wondering about is whether we should mention this change in the migration guide. Though it's still preview and the feature was not really documented anywhere. Approving regardless, works for me as is.

@vmuzikar Not sure, I think it's ok to not provide any note in migration guide. This change will go to the same version as the official docs for the additional datasources, so only the guide will be sufficient IMHO. 

@vmuzikar Is it ok for you?

vmuzikar:
@mabartos Sure, works for me (hence the approve :)). Let's merge it then.

PR #41822, opened 2025-08-12
Fix Jandex version collision to allow running tests using auth-server-quarkus-embedded

### LINKED ISSUE #41821: Fix Jandex version collision to allow running tests using auth-server-quarkus-embedded
### Before reporting an issue

### Area

testsuite

### Describe the bug

Running tests using the `auth-server-quarkus-embedded` Maven profile is not working because the transitive dependency (org.jboss:jandex:2.4.5) from `resteasy-core` is colliding with the Jandex version from Smallrye/Quarkus.

### Version

999.0.0-SNAPSHOT

### Regression

### Expected behavior

Able to run tests using `auth-server-quarkus-embedded` in the IDE

### Actual behavior

```
Exception in thread "main" java.lang.NoSuchMethodError: 'org.jboss.jandex.DotName org.jboss.jandex.DotName.createSimple(java.lang.Class)'
	at io.quarkus.deployment.execannotations.ExecutionModelAnnotationsProcessor.<clinit>(ExecutionModelAnnotationsProcessor.java:29)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:534)
	at java.base/java.lang.Class.forName(Class.jav
```

### How to Reproduce?

Enable `auth-server-quarkus-embedded` in your IDE (InteliJ) and run any test.

### Anything else?

_No response_

pedroigor:
Closes #41821

* Also removing what it seems to be an unnecessary declaration of the jandex dependency in the old testsuite.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -463,6 +463,10 @@
                         <groupId>org.slf4j</groupId>
                         <artifactId>slf4j-simple</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>org.jboss</groupId>
+                        <artifactId>jandex</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
             <dependency>
--- a/testsuite/integration-arquillian/pom.xml
+++ b/testsuite/integration-arquillian/pom.xml
@@ -56,7 +56,6 @@
         <undertow-embedded.version>1.0.0.Final</undertow-embedded.version>
         <version.org.wildfly.extras.creaper>2.0.3</version.org.wildfly.extras.creaper>
         <jakarta.persistence-legacy.version>2.2.3</jakarta.persistence-legacy.version>
-        <smallrye.jandex.version>3.0.5</smallrye.jandex.version>
         <commons.validator.version>1.8.0</commons.validator.version>
         <byte-buddy.version>1.14.13</byte-buddy.version>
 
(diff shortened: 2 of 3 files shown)


shawkins:
Did you try removing our version override (smallrye.jandex.version), or removing our direct dependency on jandex altogether? It doesn't look to me like it should actually be in use.

pedroigor:
I did not because the problem is the transitive dependency coming from `resteasy-core`. I can check if we can also remove the version override, even though it is not related to the issue. Works for you?

shawkins:
The direct dependency on jandex was originally added because jandex was used in one of the tests. That test has been removed. Are you sure integration-arquillian-tests-base even needs a jandex dependency?

pedroigor:
I don't think it does. Like I said, I can remove it. But I think we do need the exclusions herein being added to avoid the transient dependency from resteasy.

FTR, this is about running `auth-server-quarkus-embedded` in the IDE where you have all modules loaded (and their dependenices). It was working before, but for some reason stopped working.

shawkins:
> But I think we do need the exclusions herein being added to avoid the transient dependency from resteasy.

I see - instead of putting it in each usage of resteasy-core, can you add the exclusion to resteasy-core managed dependency in the root pom?

pedroigor:
Yes, I can. I did that, but was afraid to impact something else that we are not covering. I'll update the PR.

shawkins (approved):
LGTM, thanks @pedroigor

abstractj (approved):

PR #41825, opened 2025-08-12
Test flaky due to dual certificates

### LINKED ISSUE #41823: Test flaky due to dual certificates
### Before reporting an issue

### Area

admin/ui

### Describe the bug

This test is flaky:
test\clients\saml.spec.ts:136:3 › Clients SAML tests › should enable Encryption keys config 

The problem is that the test should be expecting two certificates on the page.  The way it is written, the test waits for one certificate.  Sometimes the second certificate has not been rendered yet and the test will pass.  Instead it should be waiting for the second certificate to be rendered.

### Version

main

### Regression

### Expected behavior

Test always passes

### Actual behavior

Test sometimes fails

### How to Reproduce?

Run test several times locally

### Anything else?

_No response_

ssilvert:
Fixes #41823

The problem was that the test should be expecting two certificates on the page. The way it was written, the test waits for one certificate. Sometimes the second certificate has not been rendered yet and the test would erroneously pass.  Instead it should be waiting for the second certificate to be rendered so it is checking the correct assertion.

### DIFF
--- a/js/apps/admin-ui/test/clients/saml.ts
+++ b/js/apps/admin-ui/test/clients/saml.ts
@@ -66,12 +66,14 @@ export async function clickClientSignature(page: Page) {
   await switchOff(page, "#clientSignature");
 }
 
-export async function assertCertificate(page: Page, exists = true) {
-  await expect(page.getByTestId("certificate")).toHaveCount(exists ? 0 : 1);
-}
+// Assert that the number of certificates enabled matches the number of certificates displayed
+export async function assertCertificates(page: Page) {
+  const certsEnabled = [
+    await page.getByTestId("encryptAssertions").isChecked(),
+    await page.getByTestId("clientSignature").isChecked(),
+  ].filter(Boolean).length;
 
-export async function assertDualCertificates(page: Page) {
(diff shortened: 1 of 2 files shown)


js/apps/admin-ui/test/clients/saml.ts
```diff
@@ -81,7 +87,7 @@ export async function clickOffEncryptionAssertions(page: Page) {
 export async function clickGenerate(page: Page) {
   const responsePromise = page.waitForResponse(
     (res) => res.url().includes("/generate") && res.status() === 200,
-    { timeout: 10000 },
+    { timeout: 60000 },
```
jonkoops, line 90: Perhaps the default of 30 seconds is enough here?
```suggestion
(res) => res.url().includes("/generate") && res.status() === 200,
```

jonkoops (approved):
LGTM

js/apps/admin-ui/test/clients/saml.ts
ssilvert: > Perhaps the default of 30 seconds is enough here?

Hard to say.  I just decided better safe than sorry.

PR #41845, opened 2025-08-13
memberOf attribute empty or values with a DN that does not match the role base DN fetches all roles

### LINKED ISSUE #41842: memberOf attribute empty or values with a DN that does not match the role base DN fetches all roles
### Before reporting an issue

### Area

ldap

### Describe the bug

We have encountered a security issue with syncing LDAP groupOfNames entities to Keycloak roles with role-ldap-mapper. We have created simplest scenario that you could use to reproduce.

Issue occurs when we try to sync users that has memberOf attribute with some value outside the role DN. Using simply cn=test would lead to assigning all roles that are synced from LDAP.

### Version

26.3.2

### Regression

### Expected behavior

We have encountered a security issue with syncing LDAP groupOfNames entities to Keycloak roles with role-ldap-mapper. We have created simplest scenario that you could use to reproduce.

Issue occurs when we try to sync users that has memberOf attribute with some value outside the role DN. Using simply cn=test would lead to assigning all roles that are synced from LDAP.

### Actual behavior

Do not fetch roles (and groups) if memberOf value is empty of values are using a baseDN other than the base DN configured in the role/group mapper.

### How to Reproduce?

Issue occurs when we try to sync users that has memberOf attribute with some value outside the role DN. Using simply cn=test would lead to assigning all roles that are synced from LDAP.

### Anything else?

_No response_

pedroigor:
Closes #41842

### DIFF
--- a/federation/ldap/src/main/java/org/keycloak/storage/ldap/idm/model/LDAPObject.java
+++ b/federation/ldap/src/main/java/org/keycloak/storage/ldap/idm/model/LDAPObject.java
@@ -193,8 +193,12 @@ public class LDAPObject {
 
     // Case-insensitive. Return null if there is not value of attribute with given name or set with all values otherwise
     public Set<String> getAttributeAsSet(String name) {
+        return getAttributeAsSetOrDefault(name, null);
+    }
+
+    public Set<String> getAttributeAsSetOrDefault(String name, Set<String> defaultValue) {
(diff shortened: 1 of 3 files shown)


rmartinc (approved):
Thanks @pedroigor! LGTM!

ahus1 (approved):

PR #41849, opened 2025-08-13
add configuration option to automatically add recovery codes action

### LINKED ISSUE #41836: Add config option to Configure OTP action to automatically add RecoveryCodes action upon OTP creation.
### Description

As mentioned by @ahus1 in the below given Google groups discussion, a new config option should be added to the Configure OTP required action, that it is possible to add the required action to configure Recovery Codes automatically to the user, once he created an OTP credential.
The RecoveryCodes action should only be added, if the user doesn't already have recovery codes configured.

### Discussion

https://groups.google.com/g/keycloak-user/c/MYkbi_MM-gw

### Motivation

Recovery codes are a good fallback/recovery option, if a user has OTP configured, but doesn't have acces to his OTP credentials (forgot/lost phone, app deleted, etc...).
So, having to possibility to "force" a user to configure recovery codes once he configured an OTP credential could be a good option to give users more options to recover their account in case they need it.

### Details

_No response_

dasniko:
after otp configuration

closes #41836

### DIFF
--- a/docs/documentation/server_admin/topics/authentication/recovery-codes.adoc
+++ b/docs/documentation/server_admin/topics/authentication/recovery-codes.adoc
@@ -6,6 +6,8 @@ The Recovery Codes are a number of sequential one-time passwords (currently 12) 
 
 Due to its nature, the Recovery Codes work normally as a backup for another 2FA methods. They can complement the `OTP Form` or the `WebAuthn Authenticator` to give a backing way to log inside {project_name}, for example, if the software or hardware device used for the previous 2FA methods is broken or unavailable.
 
+You can configure the `Configure OTP` required action to ask for recovery codes automatically by enabling *Add recovery codes*.
+
 ==== Check Recovery Codes required action is enabled
 
 Check the Recovery Codes action is enabled in {project_name}:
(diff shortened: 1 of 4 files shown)


ahus1:
Thank you for the PR, this looks good to me. Can you please add a test?

dasniko:
Is there already an existing test for the UpdateTotp action? I didn't find any. If yes, can you give me a pointer?
If not, are there already tests for required actions, where I can look for a good starter?

ahus1:
@dasniko - there are `AppInitiatedActionTotpSetupTest` and `RequiredActionTotpSetupTest` that test this. I searched for the use of `UserModel.RequiredAction.CONFIGURE_TOTP` and this is how I found them. Let me know if this is helpful/sufficient as a starter.

dasniko:
Thanks. I'll have a look. Maybe this takes some time, as I'm on vacations...

ahus1:
@dasniko - thank you for the PR. 

I tested it manually and this is what I found: 

* Recovery codes will only be effective when they are enabled in the login flow. Otherwise the user would enter them, but would never be able to use them. 
* The required action needs to be enabled in order for this to work

There is a similar logic in the account console to show the credential types depending on what is enabled in the authentication flows. So I updated the logic and the docs. 

Please let me know if you agree with the latest changes.

dasniko:
@ahus1, thanks for the review and the enhancements. I'm fine with these changes.

At the same time I know that people will have problems to understand these dependencies, no matter if the check is implemented or not. So, also only my enhancement would cause problems, as people won't understand that they have to enable the credential usage at multiple places. Even me, although I work nearly every day with it, I mostly forget about it in the first run, then, wenn it doesn't work like I expect it, I remember to do some additional settings...

dasniko:
Sorry if my comment lead to some confusion.
Basically I just wanted to say "yes, I'm fine with the changes" :)

ahus1:
I did some experiments to see if I can show an extended warning message when someone enables it and the preconditions are not met. 

While experiencing some bugs around that, I then considered it problematic for any automatic provisioning tool as it would then depend on the order of resources created, which would be a nuisance for everyone involved. So I decided to keep it as is. 

Another thought I had was that the functionality we have here now needs to be added to all other second factors, so that they can also trigger recovery codes.

Thinking about that, it might then be good to invert this, so to have once central piece that listens for those second factors to be created, and then adds that required action. 

Let's take this as a follow-up thought where it needs to be added as well, and merge this one for now. If you think it should be added in other places as well, please create a follow-up issue and corresponding pull requests.

ahus1 (approved):

ahus1:
FYI @keycloak/core-clients

PR #41850, opened 2025-08-13
Simplify Cache Configuration file by removing built-in cache configurations

### LINKED ISSUE #41559: Simplify Cache Configuration file by removing built-in cache configurations
### Description

The Keycloak cache configuration file contains several entries for built-in caches. Changing their configuration requires specific internal Keycloak knowledge, and depends on specific other configurations like if persistent sessions are enable or not. 

Users copy over old modified configurations to new Keycloak setups, which brings the problem of merging those configurations with new settings. 

As we are more and more creating caches programmatically and update individual cache parameters like the number of owners depending on the persistent sessions setting, or the number of entries depending, a logical consequence is to remove all default configurations from the file and create them programmatically. 

### Discussion

_No response_

### Motivation

Simplify upgrades of Keycloak, while still allowing for power users to add their own caches, or to override caches where necessary. 

### Details

Overriding built-in caches is a tricky task, and it can lead to knock-on problems. Due to that, we'll add a new flag (either SPI or CLI) to allow overwriting internal caches. 

The default for that new setting is "false". If the flag is false in KC 26.4+, it will log a warning if a cache XML file defines an internal cache. In KC 27, KC will fail to start if the flag is false and the cache XML file defines an internal cache. 

With this setup, users will see a warning in 26.4, and can act on the warning (either remove the cache definition, or set the flag), and the warning will disappear. 

We'll create a follow-up issue for KC 27 to change this behavior.

ryanemerson:
Closes #41559


ahus1 (changes requested):
Thank you for the PR. 

I modified the "not recommended" to a "technically possibly but not supported", as this might be important for downstream. 

As we're are now mentioning that you can add new caches via the XML file, we now make this officially supported (just saying...)

There was one more thing in the original issue: 

> In KC 27, KC will fail to start if the flag is false and the cache XML file defines an internal cache.

To make this possible, we would need to the behavior "start with a warning" deprecated. For that, I suggest to add another sentence about deprecation to the log message. We would also need another paragraph in the migration guide in the deprecation section.

ryanemerson:
Thanks for the review.

> I modified the "not recommended" to a "technically possibly but not supported", as this might be important for downstream.
> 
> As we're are now mentioning that you can add new caches via the XML file, we now make this officially supported (just saying...)

:+1: 



> To make this possible, we would need to the behavior "start with a warning" deprecated. For that, I suggest to add another sentence about deprecation to the log message. We would also need another paragraph in the migration guide in the deprecation section.

Added. I have also changed the ordering of the original changes paragraph to help with flow as well as fix a typo.

ahus1 (approved):

PR #41857, opened 2025-08-13
fix: using the non-deprecated console color option

### LINKED ISSUE #38328: log-console-color maps to deprecated property
### Description

log-console-color maps to quarkus.log.console.color, which has been deprecated - https://github.com/quarkusio/quarkus/pull/33429

moving forward it should be quarkus.console.color - not sure if this means we need to move it out of logging. It will also affect the logic we have for detecting properties that are both build and runtime.

The log output from KC 26.1 seems have color enabled by default, but our option defaults to false.

With KC 27, the coloring is disabled by default. You cannot use --log-console-color with a build command, but it still works as an ENV with the build, so the legacy mapping still exists somewhere in quarkus for now.

shawkins:
closes: #38328

Looking over the rationale as to why quarkus changed to this option, it doesn't apply to us as we have no other way of using the console.


mabartos (approved):
LGTM.

PR #41883, opened 2025-08-14
Handle unique constraint when creating user-session during IDP brokering

mposolda:
closes #40374

The prototype to fix the corner-case, which is hard to simulate, but it can happen in the production environments for people. This prototype implements approach 4 mentioned from this comment https://github.com/keycloak/keycloak/issues/40374#issuecomment-3188494058 .

It introduces `UserSessionExceptionHandler`, which is able to retry the HTTP request in case that other concurrent HTTP request already created same userSession with same ID. It also introduces `SendResponseException`, so the `KeycloakErrorHandler` is able to forward the provided `Response` as is without transforming it to the error page.

TBH I am not 100% happy with this approach, but could not come up with anything better, which would work and at the same time would not be too complicated like some other approaches outlined in the comment https://github.com/keycloak/keycloak/issues/40374#issuecomment-3188494058 .

NOTE: I did not added automated test as I did not manage to write an automated test to simulate the bug. I've simulated the bug just with the use of the debugger as mentioned in the comment above. I've wrote some automated test here https://github.com/mposolda/keycloak/commit/8a82589b83ce4bea1d6a0c14de99d6f2a18fdab2 , but did not manage to simulate the issue with that test (without debugger).

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/IdentityBrokerService.java
+++ b/services/src/main/java/org/keycloak/services/resources/IdentityBrokerService.java
@@ -125,6 +125,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static org.keycloak.broker.provider.AbstractIdentityProvider.BROKER_REGISTERED_NEW_USER;
+import static org.keycloak.models.light.LightweightUserAdapter.isLightweightUser;
 
 /**
  * <p></p>
@@ -924,7 +925,7 @@ public class IdentityBrokerService implements IdentityProvider.AuthenticationCal
             return AuthenticationManager.redirectToRequiredActions(session, realmModel, authSession, session.getContext().getUri(), nextRequiredAction);
         } else {
             event.detail(Details.CODE_ID, authSession.getParentSession().getId());  // todo This should be set elsewhere.  find out why tests fail.  Don't know where this is supposed to be set
-            return AuthenticationManager.finishedRequiredActions(session, authSession, null, clientConnection, request, session.getContext().getUri(), event);
+            return AuthenticationManager.finishedRequiredActions(session, authSession, null, clientConnection, request, session.getContext().getUri(), event, !isLightweightUser(federatedUser));
         }
     }
 
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanAuthenticationSessionProvider.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanAuthenticationSessionProvider.java
@@ -69,7 +69,8 @@ public class InfinispanAuthenticationSessionProvider implements AuthenticationSe
         this.sessionTx = new InfinispanChangelogBasedTransaction<>(session, cache, SessionTimeouts::getAuthSessionLifespanMS, SessionTimeouts::getAuthSessionMaxIdleMS, serializer);
         this.clusterEventsSenderTx = new SessionEventsSenderTransaction(session);
 
-        session.getTransactionManager().enlistAfterCompletion(sessionTx);
+        // TODO: If we bind ourselves to the transaction manager of Quarkus and use the embedded caches, we should perform all tasks in the prepare phase (to be verified)
+        session.getTransactionManager().enlistPrepare(sessionTx);
         session.getTransactionManager().enlistAfterCompletion(clusterEventsSenderTx);
     }
 
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/InfinispanChangelogBasedTransaction.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/InfinispanChangelogBasedTransaction.java
@@ -20,6 +20,7 @@ package org.keycloak.models.sessions.infinispan.changes;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
@@ -131,6 +132,13 @@ public class InfinispanChangelogBasedTransaction<K, V extends SessionEntity> ext
             SessionEntityWrapper<V> wrappedEntity = cache.get(key);
             if (wrappedEntity == null) {
                 return null;
+            }
+
+            /* TODO: Let's find out if this is really necessary, or just implicit if we specify PESSIMISTIC locking - maybe it is necessary if we switch to OPTIMISTIC locking? */
+            if (cache.getAdvancedCache().getTransactionManager() != null) {
+                if (!cache.getAdvancedCache().lock(Set.of(key))) {
+                    throw new RuntimeException("Unable to acquire lock for key " + key);
+                }
             }
 
             RealmModel realm = kcSession.realms().getRealm(wrappedEntity.getEntity().getRealmId());
--- a/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
+++ b/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
@@ -402,8 +402,12 @@ public final class CacheConfigurator {
                 builder.clustering().cacheMode(CacheMode.DIST_SYNC).hash().numOwners(1);
                 builder.memory().maxCount(SESSIONS_CACHE_DEFAULT_MAX);
                 return builder;
-            case ACTION_TOKEN_CACHE:
             case AUTHENTICATION_SESSIONS_CACHE_NAME:
+                // Adding a pessimistic transactional cache to allow for locking of entries to prevent concurrent operations on authentication sessions
+                // TODO: Check if pessimistic or optimistic should be used here, and if
+                builder.transaction().lockingMode(LockingMode.PESSIMISTIC);
+                builder.transaction().transactionMode(TransactionMode.TRANSACTIONAL);
+            case ACTION_TOKEN_CACHE:
             case LOGIN_FAILURE_CACHE_NAME:
                 builder.clustering().cacheMode(CacheMode.DIST_SYNC);
                 builder.encoding().mediaType(MediaType.APPLICATION_OBJECT_TYPE);
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
@@ -48,6 +48,7 @@ import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.KeycloakSessionFactory;
 import org.keycloak.models.KeycloakSessionTask;
 import org.keycloak.models.KeycloakSessionTaskWithResult;
+import org.keycloak.models.ModelDuplicateException;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.RoleModel;
 import org.keycloak.models.ScopeContainerModel;
@@ -399,6 +400,11 @@ public final class KeycloakModelUtils {
                     targetSession.getContext().setClient(clientModel);
                 }
             }
+        }
+
+        // Setup clientConnection if necessary
+        if (origContext.getConnection() != null) {
+            targetSession.getContext().setConnection(origContext.getConnection());
         }
 
         // setup auth session model if necessary.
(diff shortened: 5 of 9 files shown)


services/src/main/java/org/keycloak/services/error/KeycloakErrorHandler.java
```diff
@@ -108,6 +109,10 @@ public static Response getResponse(KeycloakSession session, Throwable throwable)
                     .build();
         }
 
+        if (throwable instanceof SendResponseException sre) {
```
pedroigor: Would it work if `SendResponseException` is a `WebApplicationException` and we change the error handler to always check whether the thrown exception is of the type of the latter and has a response, and if so, assume the response from it should be returned?

I'm wondering if we can make the handler more flexible while still aligned with Jakarta RS API.

services/src/main/java/org/keycloak/broker/provider/IdpUserSessionExceptionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    private final KeycloakContext ctx;
+
+    protected static final Logger logger = Logger.getLogger(IdpUserSessionExceptionHandler.class);
+
+    public IdpUserSessionExceptionHandler(KeycloakSession session) {
+        this.ctx = session.getContext();
+    }
+
+    @Override
+    public Response handleExceptionAtTransactionCommit(Throwable t) {
```
pedroigor: Would it work if we just return an info page with a message about the user already authenticated and an option to go back to the application?

The reasoning is to avoid retrying requests (and possible issues doing so in case the re-auth fails at the IdP) as well as make the hander simpler to cover this specific corner case.

pedroigor (commented):
@mposolda You rock! I have been trying to reproduce this one. I knew it was something related to multiple requests trying to create the same session after authenticating at the IdP due to multiple requests sent by the brower, but I never managed to reproduce it.

Overall, it LGTM and I have just a few comments/suggestions. None of them are blockers.

I guess we cannot test this scenario at all, right?

services/src/main/java/org/keycloak/services/resources/IdentityBrokerService.java
```diff
@@ -896,6 +897,7 @@ private Response afterPostBrokerLoginFlowSuccess(AuthenticationSessionModel auth
 
 
     private Response finishBrokerAuthentication(BrokeredIdentityContext context, UserModel federatedUser, AuthenticationSessionModel authSession, String providerAlias) {
+        session.sessions().addExceptionHandler(new IdpUserSessionExceptionHandler(session));
```
pedroigor: It would be nice if we could avoid setting this handler and have this behavior encapsulated by the session provider. I think we can check whether the current session is related to brokering auth somehow by looking at the auth session or something?

My point being, if we have similar cases but in other flows (do not know about any other at the moment), changes will go in a single/known place.

ahus1 (commented):
As an alternative solution as suggested in the internal chat: Could we instead on a higher level do a KeycloakModelUtils.runJobInTransaction() and a Retry.execute() with like two iterations, and possibly checking the exception.
Possible only at the one place where the REST endpoint is entered.

services/src/main/java/org/keycloak/services/error/KeycloakErrorHandler.java
mposolda: Maybe it works to use `WebApplicationException` . I've rather introduced the dedicated type as I was not sure about side-effect of updating `KeycloakErrorHandler` to handle the `WebApplicationException`. But maybe I can try it and update PR. We will see if some tests would start to fail :-)

services/src/main/java/org/keycloak/services/resources/IdentityBrokerService.java
mposolda: Yes, was thinking about that approach as well. It is possible it would allow to avoid changes in the `UserSessionProvider` SPI, which could be a benefit though... Will try to explore that path.

And yes, I've just received the report that it can happen in other scenarios besides brokering (will send you privately)...

services/src/main/java/org/keycloak/broker/provider/IdpUserSessionExceptionHandler.java
mposolda: Maybe... But I will try to take a look at the [suggestion](https://github.com/keycloak/keycloak/pull/41883#pullrequestreview-3123678736) by @ahus1 below, which will make sure that retrying of request will not be needed and hence we will not need to display info page as well. I am not 100% sure how it will work, but will give it a shot... Maybe info page could be used as fallback if it fails...

mposolda:
> @mposolda You rock! I have been trying to reproduce this one. I knew it was something related to multiple requests trying to create the same session after authenticating at the IdP due to multiple requests sent by the brower, but I never managed to reproduce it.
> 
> Overall, it LGTM and I have just a few comments/suggestions. None of them are blockers.
> 
> I guess we cannot test this scenario at all, right?

@pedroigor Yes, I failed to do an automated test as mentioned in the description of this PR. I've simulated the bug just with the debugger...

mposolda:
> As an alternative solution as suggested in the internal chat: Could we instead on a higher level do a KeycloakModelUtils.runJobInTransaction() and a Retry.execute() with like two iterations, and possibly checking the exception. Possible only at the one place where the REST endpoint is entered.

@ahus1 I was thinking as well about something like this approach (Approach 3 in my comment on the issue https://github.com/mposolda/keycloak/commit/8a82589b83ce4bea1d6a0c14de99d6f2a18fdab2 ). ATM I afraid that it can make even more changes needed. The problem is, that issue happens at the transaction commit. So at the transaction commit, we would need to check that "insert" failed and based on that, retry the transaction by lookup existing userSession and possibly update existing userSession with the data from current transaction. Maybe it is doable and looks like more proper approach and it is also aligned with the "retries", which are already done in the storage layer (The stuff I wrote many years ago during initial cross-dc work...).  But I have some fear if it is doable... I will try to explore this option a bit more and will let you know how it goes.

pedroigor:
> Maybe it is doable and looks like more proper approach and it is also aligned with the "retries", which are already done in the storage layer (The stuff I wrote many years ago during initial cross-dc work...). But I have some fear if it is doable... I will try to explore this option a bit more and will let you know how it goes.

I think this approach will handle the problem silently instead of informing the user about the problem and handling it as an unexpected/corner-case when processing authorization responses from IdPs. IMO, the user should be informed and take some action, like ignore and get back to the application. But not an error as it stands today.

I would not complicate, that is why I'm not sure about the retry on the HTTP request, as done here. But try to potentially improve these changes so that the logic of handling such cases does not have any retry logic and is manageable from a single place in the code.

mposolda:
@ahus1 @pedroigor I've updated PR for the similar approach as suggested by @ahus1 in the previous comment. However there is no any retry of creating user session. User session is created in dedicated transaction and if it fails (due the fact that other transaction created user session in the meantime), the existing user session is looked-up.

There are not any retries of creating user session and not any retried HTTP requests. Which IMO is best for UX as there is no retried HTTP request and user does not need to confirm any page and things just work as expected and there is single user session created.

ahus1:
@mposolda - I had some more high-level retry in mind. Can you please let me know which endpoint is called in your case, so I prepare an alternative PoC?

mposolda:
> @mposolda - I had some more high-level retry in mind. Can you please let me know which endpoint is called in your case, so I prepare an alternative PoC?

@ahus1 The endpoint where it happens is IDP broker callback endpoint. Something like https://idb.anonymized.dummy/realms/myrealm/broker/sso-via-bu-idp/endpoint .

Here I've added the steps-to-reproduce the issue with the debugger: https://github.com/keycloak/keycloak/issues/40374#issuecomment-3188494058

Just a note that retrying whole request (or retrying the whole method with the request) might be also not so ideal approach as callback endpoint is called with parameters like `code` for example (in case of OIDC provider) and double-use of same code is not allowed per OIDC/OAuth2 specification. That's why I was more towards to just "retry creating of user session" instead of something aligned to "Retrying whole request" as I had in my previous approach.

ahus1:
@mposolda - I had a look and the way the new retry works might work in some databases depending on the isolation level, but not all: The inner transaction will create some entries in the database, but the outer transaction might not see them if it is using for example snapshot isolation. It might even work most of the time if the data is retrieved from the Infinispan cache, but it might fail if it is not. 

If the problem occurs with multiple tabs where the `authChecker.js` triggers a concurrent login, then I suggest to disable the auth checker when the tab is not active. I'll push one commit that adds an event listener when the tab becomes active again. 

At the same time I don't understand why the `authChecker.js` would fire and create user sessions concurrently as AFAIK it only triggers after the login with the user session is complete and the cookie exists. How can it trigger a concurrent login? 

I'm pushing my two commits now. Maybe we can have a call to discuss this in more detail tomorrow. WDYT?

ahus1:
@mposolda / @pruivo - the second commit makes the authentication cache transactional, which should prevent concurrent operations on authentication sessions, still it will make things slower as it will have additional operations to do the transactional stuff - we should measure that. Let's talk it over in a call.

Kudos to @pruivo for the guidance on how to configure the cache.

mposolda:
> @mposolda - I had a look and the way the new retry works might work in some databases depending on the isolation level, but not all: The inner transaction will create some entries in the database, but the outer transaction might not see them if it is using for example snapshot isolation. It might even work most of the time if the data is retrieved from the Infinispan cache, but it might fail if it is not.

Ah, ok. I was thinking whether this could be the issue, but at the same time, all the DB tests were OK with my changes (and DB tests are testing the brokering though). Maybe we can add some DB test with the snapshot isolation to avoid regressions of introducing the similar changes in the future (if it is really a problem)?

> 
> If the problem occurs with multiple tabs where the `authChecker.js` triggers a concurrent login, then I suggest to disable the auth checker when the tab is not active. I'll push one commit that adds an event listener when the tab becomes active again.
> 
> At the same time I don't understand why the `authChecker.js` would fire and create user sessions concurrently as AFAIK it only triggers after the login with the user session is complete and the cookie exists. How can it trigger a concurrent login?

The problem here is not directly related to the `authChecker.js` . The idea of the `authChecker.js` is, that once you login in tab1, then all other browser tabs are automatically authenticated. If the `authChecker.js` is skipped for other tabs, then this will not work and cause again the issues like "You are already logged-in" as at the time when user switches focus to tab2, then this tab probably will not be able to complete authentication as authenticationSession would be already expired (as RootAuthenticationSession is valid just 1 minute after the login of tab1, so if user goes to the tab2 later, there would be issues).

I will try to describe in more details how the issue with IDP could happen:
1) You have setup like in our brokering tests. So realm `consumer`, which client uses for login and this realm delegates login to the identity provider provided by realm `provider` 
2) There are 2 browser tabs, which initiated login to the `consumer` realm. Both of those were redirected to the IDP login and hence redirected to the login screen of the realm `provider` 
3) User authenticated in tab1 of the realm `provider` . This created session cookie for realm `provider` and redirected user to the realm `consumer` to the IDP callback endpoint
4) The tab2 was automatically authenticated in the realm `provider` due the `authChecker.js` as session cookie already exists in the realm `provider`. This redirected user again to the realm `consumer` to the IDP callback endpoint. But since request to the `consumer` from previous step 3 is still in progress and commit of the userSession in the realm `consumer` is not yet created, it caused this conflicting situation as IDP callback request will also try to create same userSession with same ID.

Note that the issue happens for the realm `consumer` where the `authChecker.js` was not used (as in the scenario above, the `authChecker.js` of the realm `provider` was used). Also note that the scenario is not specific to the Keycloak IDP, but it can happen with other IDPs as well, which would automatically redirect other browser tabs to the `consumer` after user completed login in the IDP in the tab1. I have confirmation that it can happen also with ForgeRock and franceConnect used as IDPs (See last 2 comments on the issue https://github.com/keycloak/keycloak/issues/40374 ).

Considering this, I vote to remove your changes from `authChecker.js` from this PR.

At the same time, it seems that this issue is maybe not specific to the identity brokering, but can happen also in other scenarios (even after some previous bugs fixed)... So maybe some more "generic" approach in the infinispan layer might be useful.

> 
> I'm pushing my two commits now. Maybe we can have a call to discuss this in more detail tomorrow. WDYT?

+1

ahus1:
Thank you for the input. I'll move the authChecker.js changes to another issue / PR to have a discussion if that behavior is wanted independently of the issue at hand. 

About the changes in 347659393e29af81657f415ede5e356411808fb6 - let's discuss tomorrow with @pruivo. As far as I understand this is only available when running with an embedded ISPN, but not when running with an external ISPN.

pruivo:
Doesn't JPA have some kind of "upsert" that could be used? 
Clients cannot lock keys on the external Infinispan cluster. It is a nightmare if the client crashes; it keeps the lock forever. 

edit: due to the recent test with network delays, I wouldn't like to use Infinispan transactions for embedded either.

mposolda:
@pruivo Yes, the "upsert" would be really great and should help. AFAIK it is supported for noSQL databases, however I don't know about SQL databases and if there is some SQL standard for this or some support in the JPA layer...

pruivo:
@mposolda I did some googling and JPA itself does not support it, but Hibernate does, since 6.3:
https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/StatelessSession.html#upsert(java.lang.Object)

It is not clear to me how it behaves with different isolation levels. I mean, for example, with snapshot isolation, when 2 transactions start and the user session does not exist, my gut feeling says that one should fail - otherwise, we are breaking the transaction isolation.

ahus1:
Closing due to #42451 being merged

PR #41886, opened 2025-08-14
fix: minimize warnings and debug logs due to import/export options

### LINKED ISSUE #38438: Avoid 'duplicated mappers' message
### Description

After upgrading KC 22.0.5 to 26.1.4 we see suspucious DEBUG messages during boot

```
2025-03-26 13:02:16,675 DEBUG [org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers] (main) Duplicated mappers for key 'kc.dir'. Used the first found.
2025-03-26 13:02:16,678 DEBUG [org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers] (main) Duplicated mappers for key 'kc.dir'. Used the first found.
2025-03-26 13:02:16,679 DEBUG [org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers] (main) Duplicated mappers for key 'kc.dir'. Used the first found.
2025-03-26 13:02:16,680 DEBUG [org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers] (main) Duplicated mappers for key 'kc.dir'. Used the first found.
2025-03-26 13:02:16,682 DEBUG [org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers] (main) Duplicated mappers for key 'kc.dir'. Used the first found.
```

Check if possible to avoid them

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
closes: #38438

We have to bootstrap the configuration first to run the sanitization, so we still end up with some trace logs.

There seems to be little point in emitting logs / warnings about import / export options for non-import / export commands, so the logic here just disables the options otherwise.

With this change I don't believe we'll have duplicates after sanitizing because that will also address the intra-command duplicates.

Having multiple mappings can be benign in simple cases, but there are a lot of places after sanitizing where we could get tripped up if there are multiple. So we could have the sanitize of the getMapper logic throw an exception if multiple mappings remain. @mabartos would you be opposed to that?


shawkins:
The test failures highlight even further fragility. The --file option is also directly on the update commands - not via a PropertyMapper. If we disable the import / export kc.dir options, then we'll end up with an exception dealing with --file there.

shawkins:
I can make a small simplification after #41771

Overview of changes:
- makes option sanitization happen before options are added to the command and conditional on whether help-all is used
- moves help-all handling to localized logic, rather than a static variable
- adds unconditional checks that we don't have property mappers conflicting with command options and that after sanitization there are no duplicate propertymappers

vmuzikar:
@mabartos Can you please review when you get a chance?

mabartos:
> Having multiple mappings can be benign in simple cases, but there are a lot of places after sanitizing where we could get tripped up if there are multiple. So we could have the sanitize of the getMapper logic throw an exception if multiple mappings remain. @mabartos would you be opposed to that?

@shawkins We should be able to check duplicates before the `getMapper()` execution, right? I'd rather see some exception during startup than in runtime. Would it be possible?

EDIT: Ahh, I see that you do sth like that in the Picocli class

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
```diff
@@ -367,10 +336,13 @@ public Map<String, PropertyMapper<?>> getDisabledRuntimeMappers() {
...
 
         private static void sanitizeMappers(Map<OptionCategory, List<PropertyMapper<?>>> mappers,
-                                            Map<String, PropertyMapper<?>> disabledMappers) {
+                                            Map<String, PropertyMapper<?>> disabledMappers, AbstractCommand command) {
             mappers.forEach((category, propertyMappers) ->
                     propertyMappers.removeIf(pm -> {
-                        final boolean shouldRemove = !pm.isEnabled();
+                        // do a localized check based upon the current command
+                        // to exclude one or both of the import / export options as they overlap with one another
+                        final boolean shouldRemove = !pm.isEnabled() || (pm.getCategory() == OptionCategory.IMPORT && !(command instanceof Import))
```
mabartos: Would it be possible to have some general approach to check the category for the mapper and used command? Leveraging `AbstractCommand.getOptionCategories()`?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -156,18 +157,26 @@ public <T> T set(T value) {
...
                     currentCommand = ac;
                 }
             }
-            initConfig(currentCommand);
+            // init the config before adding options to properly sanitize mappers
+            boolean allHelpOptions = false;
+            if (commandLineList.size() > 1 && commandLineList.get(commandLineList.size() - 1).getCommand() instanceof AbstractCommand ac) {
+                allHelpOptions = Optional.ofNullable(ac.getHelpAllMixin()).map(HelpAllMixin::isAllOptions).orElse(false);
+            }
+            initConfig(allHelpOptions, currentCommand);
```
mabartos: Cannot we get the info about the allHelp from the `currentCommand`? Does it have to be an additional argument of the `initConfig()`?

mabartos (commented):
@shawkins Some considerations.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: > Cannot we get the info about the allHelp from the currentCommand?

We cannot. The commandLineList contains the actual parsed commands from the first pass - they will have the flags and other variables set.

While the currentCommand is the user object associated with what will be used in the second parsing - it will be freshly initialized at the time initConfig is called.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
shawkins: I think my concern was the potential ambiguation between "excluded / ignored" and "disabled".

However making an exception for just import / export I agree is still confusing. The Picocli logic will currently skip validating excluded categories - but I think that rests on the unstated assumption that excluded categories will never contain build-time options.

Let me see if I can make this clearer.

shawkins: After making the relevant changes we do actually make a distinction between excluded and disabled - excluded categories like caching are still expected to provide values. What is inconsistent though is that while we exclude these from the cli, we do not from the env / keycloak.conf - this could be seen as an additional bug.

In any case the change I settled on was to expend isEnabled to check for the command. This allows for a localized disabling of file / dir, rather than a broader check.

shawkins:
@mabartos @vmuzikar @Pepo48 I realize this is starting to be a large change. There are now three key drivers:

1. Fully sanitize mappers so there are no duplicates
2. Change the excluded category concept into something that hides options
3. Remove logic that simply scans the arg list for checks that are based upon actual parsing

It's number 2 that's a big departure from what we have been doing, so it's worth diving into more:
- our logic is inconsistent wrt to excluded options. They become hard errors when used in the cli, but are of course allowed in the keycloak.conf or env - relates to https://github.com/keycloak/keycloak/issues/42332 
- related to the first point, we are also skipping validation of excluded options - which if they can have overriden user values gets potentially dangerous.
- While that may help some users to see hard cli errors in these cases, it greatly complicates our logic and makes other user scenarios more complicated. If you do start ... with a mixture of build and runtime options, I would rather the user be allowed to simply tack on --optimized, rather than manually prune or capture pruned command from the log. If we fully embrace the way this pr works, we can simplify or remove that log. As the test failures show there is still so more cleanup needed.
- gives us much more flexibility in excluding options from help. For example we could easily eliminate all runtime options from being shown with the update-compatiblity command help - but they are still accepted because they are simply hidden.
- a sticking point potentially is what to do about things that are truly command specific like the import / export. They are generally excluded / hidden, but you could make the case for all of their options having enabledWhen - not just the file and dir options. This would prevent you from doing stuff like ./kc.sh start --override=true ...

WDYT?

vmuzikar:
@shawkins Let's discuss it on our sync on Tue.

vmuzikar:
Putting on hold until after the 26.4 release.

shawkins:
> Putting on hold until after the 26.4 release.

I'll take another pass over the logic. I need to ensure that our usage of enabledWhen is limited only to situations that resolve property conflicts. Otherwise we should be treating the property as hidden and producing a warning.

shawkins:
> I'll take another pass over the logic. I need to ensure that our usage of enabledWhen is limited only to situations that resolve property conflicts. Otherwise we should be treating the property as hidden an producing a warning.

Thinking about it more, that is not needed.

PR #41754, opened 2025-08-08
Add `rewrite` support for `errorprone.refasterrules`

Pankraz76:
Closes #41761

PR #41887, opened 2025-08-14
Apply security fixes featuring `refasterrules.FileRulesRecipes`

Pankraz76:
Applies: 

- https://error-prone.picnic.tech/refasterrules/FileRules/

### DIFF
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/Util.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/Util.java
@@ -2,10 +2,11 @@ package org.keycloak.it.cli.dist;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 
 public class Util {
     static File createTempFile(String prefix, String suffix) throws IOException {
-        var file = File.createTempFile(prefix, suffix);
+        var file = Files.createTempFile(prefix, suffix).toFile();
         file.deleteOnExit();
         return file;
     }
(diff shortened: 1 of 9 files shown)


ahus1 (changes requested):
As outlined in #41754, each PR needs a issue, see also CONTRIBUTING.md. While the discussion on how to apply errorprone is onging, I suggest to pause this PR. 

I also want to avoid a human to review code changes done by a machine, so I'd rather have a method to run this locally instead of review this PR. 

So suggest to close this PR until the other PR is resolved.

Pankraz76:
> I also want to avoid a human to review code changes done by a machine, so I'd rather have a method to run this locally instead of review this PR.

It depends. I would argue the same, might even impose some auto execution of the run goal to fix upfront, if its not taking to long. Rewrite is considered some makeup or roundup making the current state complain to certain rules and patterns.

Ideally Devs do not produce violating code so it could be only rarely to fail. This might change as get some real mighty recipe(s) like CodeCleanUp or [CommonStaticAnalysis](https://docs.openrewrite.org/recipes/staticanalysis/commonstaticanalysis), being a composite recipe like all the other BestPractices. 

No tool it perfect, meaning every change, at least the first once and or on large scale need to me checked by human to make sure its not doing harm. Of course most critical stuff should be covered by test to avoid buts there is still going to be an review after all.

If the change is done by dev or some machine then is considered an impl. detail. The analysis will fail until the code is complain, making "CI happy" but the diff and review process so the responsibility remains the same burden on dev´s to check before consider merge.

Pankraz76:
yes, every PR needs some issue is really strict i like this and kind of know this from check. This means quality is really important meaning this approach should be easy to digest for the project improving overall.

PR #41904, opened 2025-08-15
feat(FGAPv2): introduce RESET_PASSWORD scope and evaluation

### LINKED ISSUE #41901: FGAP v2: RESET_PASSWORD capability for USERS
### Description

**Why:** Password resets were implicitly granted by MANAGE\_USERS. With FGAP v2 we need policy-based, auditable control to allow/deny resets per user/group with deny-overrides and secure-by-default behavior.

**What:**

* Add `reset-password` scope to `USERS`
* Require `RESET_PASSWORD` in `UserResource.resetPassword()`
* `UserPermissionsV2` implements deny-overrides, secure-by-default; optional fallback to MANAGE\_USERS via `fgap.v2.resetPassword.fallbackToManageUsers` (default=false)
* `getAccess(user).resetPassword` for Admin Console
* Preserve self-service password change
* Logging/auditing enhancements
* Tests and docs

**Migration notes:** When FGAP v2 is enabled and no RESET\_PASSWORD policies exist, reset is denied by default unless fallback is enabled.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

Bagautdino:
Closes #41901

## Summary

This PR adds a dedicated `RESET_PASSWORD` scope to the `USERS` resource in FGAP v2 (Fine-Grained Admin Permissions) and updates evaluation so password reset rights are explicitly governed by policy rather than implicitly inherited from `MANAGE_USERS`.

## Motivation

Today, password resets can be implicitly allowed via `MANAGE_USERS`, which limits governance and auditing. With FGAP v2 enabled, administrators should be able to allow or deny password resets per user/group with explicit policies and deny-overrides semantics.

## What’s included

* Add `RESET_PASSWORD` to `AdminPermissionsSchema.USERS`.
* Expose `canResetPassword()` and `requireResetPassword()` in user permissions API.
* Require `RESET_PASSWORD` in `UserResource.resetPassword()` (instead of `MANAGE_USERS`).
* Implement full FGAP v2 evaluation in `UserPermissionsV2`:

  * Deny-overrides decision model.
  * Secure-by-default when policies exist.
  * Optional fallback to legacy `MANAGE_USERS` via
    `fgap.v2.resetPassword.fallbackToManageUsers` (default `false`).
* Include `resetPassword` flag in `getAccess(user)` for Admin Console UI (hides “Reset Password” button when `false`).
* Preserve self-service password change (caller == target user) without FGAP checks.
* Utility to detect policies referencing `RESET_PASSWORD` scope.
* Enhanced auditing/logging for allow/deny decisions.
* Unit and integration tests covering all decision paths.

## Behavior changes

* With FGAP v2 enabled **and** at least one policy referencing `RESET_PASSWORD`:

  * `MANAGE_USERS` alone no longer grants password reset.
  * Explicit `DENY` always overrides `ALLOW`.
* When **no** `RESET_PASSWORD` policies are present:

  * `fallback=false` (default): **deny by default**.
  * `fallback=true`: **allow** if `MANAGE_USERS` is granted.
* Self-service flows are unaffected.

## Configuration

* `fgap.v2.resetPassword.fallbackToManageUsers` (boolean, default `false`).

## Admin Console / API

* `getAccess(user).resetPassword` added; UI should hide the “Reset Password” action when `false`.

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/policy/AbstractUserResourcePolicyProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/policy/AbstractUserResourcePolicyProvider.java
@@ -130,9 +130,9 @@ public abstract class AbstractUserResourcePolicyProvider implements ResourcePoli
     }
 
     @Override
-    public boolean scheduleOnEvent(ResourcePolicyEvent event) {
+    public boolean activateOnEvent(ResourcePolicyEvent event) {
         return this.supports(event.getResourceType())
-                && this.getSupportedOperationsForScheduling().contains(event.getOperation())
+                && this.getSupportedOperationsForActivation().contains(event.getOperation())
                 && this.isResourceInScope(event.getResourceId());
     }
 
@@ -143,17 +143,31 @@ public abstract class AbstractUserResourcePolicyProvider implements ResourcePoli
                 && this.isResourceInScope(event.getResourceId());
     }
 
+    public boolean deactivateOnEvent(ResourcePolicyEvent event) {
+        return this.supports(event.getResourceType())
+                && this.getSupportedOperationsForDeactivation().contains(event.getOperation())
+                && !this.isResourceInScope(event.getResourceId());
+    }
+
     @Override
     public void close() {
         // no-op
     }
 
-    protected List<ResourceOperationType> getSupportedOperationsForScheduling() {
-        return List.of();
+    protected List<ResourceOperationType> getSupportedOperationsForActivation() {
+        return this.getBrokerAliases().isEmpty()
+                    ? Collections.emptyList()
+                    : List.of(ResourceOperationType.ADD_FEDERATED_IDENTITY);
     }
 
     protected List<ResourceOperationType> getSupportedOperationsForResetting() {
-        return List.of();
+        return Collections.emptyList();
+    }
+
+    protected List<ResourceOperationType> getSupportedOperationsForDeactivation() {
+        return this.getBrokerAliases().isEmpty()
+                    ? Collections.emptyList()
+                    : List.of(ResourceOperationType.REMOVE_FEDERATED_IDENTITY);
     }
 
     protected EntityManager getEntityManager() {
--- a/model/jpa/src/main/java/org/keycloak/models/policy/UserCreationTimeResourcePolicyProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/policy/UserCreationTimeResourcePolicyProvider.java
@@ -31,7 +31,9 @@ public class UserCreationTimeResourcePolicyProvider extends AbstractUserResource
     }
 
     @Override
-    protected List<ResourceOperationType> getSupportedOperationsForScheduling() {
-        return List.of(CREATE);
+    protected List<ResourceOperationType> getSupportedOperationsForActivation() {
+        return new java.util.ArrayList<>(super.getSupportedOperationsForActivation()) {{
+            add(CREATE);
+        }};
     }
 }
--- a/model/jpa/src/main/java/org/keycloak/models/policy/UserSessionRefreshTimeResourcePolicyProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/policy/UserSessionRefreshTimeResourcePolicyProvider.java
@@ -17,6 +17,7 @@
 
 package org.keycloak.models.policy;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import org.keycloak.component.ComponentModel;
@@ -32,8 +33,11 @@ public class UserSessionRefreshTimeResourcePolicyProvider extends AbstractUserRe
     }
 
     @Override
-    protected List<ResourceOperationType> getSupportedOperationsForScheduling() {
-        return List.of(CREATE, LOGIN);
+    protected List<ResourceOperationType> getSupportedOperationsForActivation() {
+        return new ArrayList<>(super.getSupportedOperationsForActivation()) {{
+            add(CREATE);
+            add(LOGIN);
+        }};
     }
 
     @Override
--- a/server-spi-private/src/main/java/org/keycloak/models/policy/ResourceOperationType.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/policy/ResourceOperationType.java
@@ -3,6 +3,8 @@ package org.keycloak.models.policy;
 public enum ResourceOperationType {
 
     CREATE,
+    LOGIN,
+    ADD_FEDERATED_IDENTITY,
+    REMOVE_FEDERATED_IDENTITY
 
-    LOGIN
 }
--- a/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyEvent.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyEvent.java
@@ -1,15 +1,26 @@
 package org.keycloak.models.policy;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class ResourcePolicyEvent {
 
     private final ResourceType type;
     private final ResourceOperationType operation;
     private final String resourceId;
+    private final Map<String, String> details = new HashMap<>();
 
     public ResourcePolicyEvent(ResourceType type, ResourceOperationType operation, String resourceId) {
+        this(type, operation, resourceId, null);
+    }
+
+    public ResourcePolicyEvent(ResourceType type, ResourceOperationType operation, String resourceId, Map<String, String> details) {
         this.type = type;
         this.operation = operation;
         this.resourceId = resourceId;
+        if (details != null) {
+            this.details.putAll(details);
+        }
     }
 
     public ResourceType getResourceType() {
@@ -22,5 +33,9 @@ public class ResourcePolicyEvent {
 
     public String getResourceId() {
         return resourceId;
+    }
+
+    public Map<String, String> getDetails() {
+        return details;
     }
 }
--- a/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyProvider.java
@@ -38,18 +38,18 @@ public interface ResourcePolicyProvider extends Provider {
     boolean supports(ResourceType type);
 
     /**
-     * Indicates whether the policy supports being assigned to a resource based on the event or not. If {@code true}, the
-     * policy's first action will be scheduled for the resource.
+     * Indicates whether the policy supports being activated for a resource based on the event or not. If {@code true}, the
+     * policy will be activated for the resource. For scheduled policies, this means the first action will be scheduled. For
+     * immediate policies, this means all actions will be executed right away.
      *
      * At the very least, implementations should validate the event's resource type and operation to ensure the policy will
-     * only be assigned on expected operations being performed on the expected type.
+     * only be activated on expected operations being performed on the expected type.
      *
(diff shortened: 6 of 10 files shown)


services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
```diff
@@ -147,6 +151,47 @@ public boolean canManageGroupMembership(UserModel user) {
...
 
+    @Override
+    public boolean canResetPassword(UserModel user) {
+        // FGAP v2 disabled or no admin-permissions client initialized -> legacy behavior
+        if (!AdminPermissionsSchema.SCHEMA.isAdminPermissionsEnabled(root.realm) || root.realmResourceServer() == null) {
+            return root.hasOneAdminRole(AdminRoles.MANAGE_USERS);
+        }
+
+        // Check if there is at least one policy referencing RESET_PASSWORD scope
+        if (hasResetPasswordPolicies()) {
```
pedroigor: For some, you might want to enforce access managed on the `manage` scope. I think we should try first the check with `reset-password` scope and if it fails, we then check `manage`.

There is a performance penalty with this approach, but perhaps it covers more use cases and avoid forcing people to always set `reset-password` if the `manage` scope is enough.

@vramik Can you also give your thoughts on this one?

Bagautdino: This is my first time working with this part of Keycloak, so I initially approached the implementation with the goal of making password reset permissions more granular and policy-driven. I wanted RESET_PASSWORD to be a distinct capability, so that realm admins with MANAGE_USERS wouldn't automatically get this right unless allowed explicitly via a policy.
That said, I totally understand that falling back to MANAGE_USERS when no RESET_PASSWORD policies are defined makes the behavior more compatible by default.
If this approach is preferred, I’d really appreciate if someone from the team could help adjust the fallback logic in the PR. But I’m also happy to update it myself if needed.

pedroigor: I will take a look and contribute the fallback. Let's see first what @vramik has to say about it.

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissions.java
```diff
@@ -62,6 +62,7 @@ class UserPermissions implements UserPermissionEvaluator, UserPermissionManageme
     private static final String IMPERSONATE_SCOPE="impersonate";
     private static final String USER_IMPERSONATED_SCOPE="user-impersonated";
     private static final String MANAGE_GROUP_MEMBERSHIP_SCOPE="manage-group-membership";
+    private static final String RESET_PASSWORD_SCOPE="reset-password";
```
vramik: The variable is not used, can we remove it?

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
```diff
@@ -147,6 +151,47 @@ public boolean canManageGroupMembership(UserModel user) {
         return eval.hasPermission(new UserModelRecord(user), null, AdminPermissionsSchema.MANAGE_GROUP_MEMBERSHIP);
     }
 
+    @Override
+    public boolean canResetPassword(UserModel user) {
+        // FGAP v2 disabled or no admin-permissions client initialized -> legacy behavior
+        if (!AdminPermissionsSchema.SCHEMA.isAdminPermissionsEnabled(root.realm) || root.realmResourceServer() == null) {
```
vramik:
```suggestion
        // FGAP v2 disabled for a realm -> legacy behavior
        if (!AdminPermissionsSchema.SCHEMA.isAdminPermissionsEnabled(root.realm)) {
```

`isAdminPermissionsEnabled(root.realm)` also tests for the realm being `null` and returns `false` in that case, it seems that the second part of the if is redundant. wdyt?

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
vramik: Hi @pedroigor and @Bagautdino, thanks for the discussion on this.

I've been thinking about the implicit fallback from RESET_PASSWORD to MANAGE_USERS. While I understand the goal of backward compatibility, I'm a bit concerned that an implicit fallback could be error-prone and harder to maintain in the long run. It couples two distinct permissions, which might lead to confusion for administrators who expect RESET_PASSWORD to be the single source of truth for this capability.

I believe we should follow opt-in approach when dealing with `fallbackToManageUsers`. As far as I am aware this is also preferred way how to introduce behavioral changes. What are your thoughts on it? 

This leads me to a broader question for @pedroigor: what are the team's general policies around changing default settings? With an explicit option in place, perhaps we could plan to switch the default to a more secure setting (i.e., fallback disabled) in a future major release?

vramik (commented):
Thank you @Bagautdino for preparing the PR. Overall it looks great. I have only few nitpicks. and the reply to the discussion about fallback mechanism.

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissions.java
Bagautdino: removed

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
Bagautdino: Hi @vramik, thanks a lot for the detailed feedback 🙏

I’ve updated the implementation to follow the opt-in fallback approach:
If there are any RESET_PASSWORD policies, they are the single source of truth (deny-overrides, secure-by-default).
Only when no policies exist, and the flag fgap.v2.resetPassword.fallbackToManageUsers=true is set, we fallback to MANAGE_USERS.
With no policies and the flag disabled (default), reset password is denied.

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
Bagautdino: I’ve removed the redundant root.realmResourceServer() == null check and simplified the condition as suggested. Thanks!

Bagautdino:
I’ve updated the PR. If you’d like me to change anything else, just let me know 👍

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
vramik: Thanks a lot, I probably didn't express myself good enough. Sorry about that. What I mean by opt-in approach is something like this

```java
    @Override
    public boolean canResetPassword(UserModel user) {
		// admin roles has the precedence over permissions
		if (root.hasOneAdminRole(AdminRoles.MANAGE_USERS)) {
            return true;
        }

		// by default use legacy behavior - with the aim to change the default in next major release
		boolean fallbackToManage = Config.scope("fgap", "v2", "resetPassword").getBoolean("fallbackToManageUsers", true);

		if (fallbackToManage) {
			return eval.hasPermission(new UserModelRecord(user), null, AdminPermissionsSchema.MANAGE);
		}

		return eval.hasPermission(new UserModelRecord(user), null, AdminPermissionsSchema.RESET_PASSWORD);
	}
```

So that user who would like to leverage the RESET_PASSWORD permission, would need to pass the `fallbackToManageUsers=false` option. In next version we might be able to switch the default to `false`, it should provide users some time to leverage new behavior. WDYT?

Bagautdino: Got it, thanks for clarifying 🙏 I’ve updated the implementation to follow this opt-in fallback approach with fallbackToManageUsers=true as the current default. Let me know if you’d like any further adjustments 👍

vramik:
> I’ve updated the PR. If you’d like me to change anything else, just let me know 👍

Thanks a lot, great job. There are few points from top of my head we need to consider before merging it. 

- update the documentation to reflect the changes: https://www.keycloak.org/docs/latest/server_admin/index.html#_admin_permissions
- add a note in the upgrading guide describing the new permission and the config option to enable it: https://www.keycloak.org/docs/latest/upgrading/index.html
- consider `group` permission - in other words - will we support setting up permission for users only? Do we or don't we want to allow setup permission for group members? @pedroigor wdyt? If so, should we handle it within this PR, or rather follow-up PR?

Bagautdino:
> > I’ve updated the PR. If you’d like me to change anything else, just let me know 👍
> 
> Thanks a lot, great job. There are few points from top of my head we need to consider before merging it.
> 
> * update the documentation to reflect the changes: https://www.keycloak.org/docs/latest/server_admin/index.html#_admin_permissions
> * add a note in the upgrading guide describing the new permission and the config option to enable it: https://www.keycloak.org/docs/latest/upgrading/index.html
> * consider `group` permission - in other words - will we support setting up permission for users only? Do we or don't we want to allow setup permission for group members? @pedroigor wdyt? If so, should we handle it within this PR, or rather follow-up PR?

Hi @vramik  again👋
I’ve updated the documentation (Server Admin Guide and Upgrading Guide) according to your comments.
Could you please check if everything looks correct?

Regarding group permissions, I think it’s better to address that in a separate follow-up PR. This way, the current PR stays focused and easier to review, and I’ll handle support for RESET_PASSWORD group policies (with tests) in the next PR.

Thanks a lot! 🙏

vramik:
> Hi @vramik again👋 I’ve updated the documentation (Server Admin Guide and Upgrading Guide) according to your comments. Could you please check if everything looks correct?
> 
> Regarding group permissions, I think it’s better to address that in a separate follow-up PR. This way, the current PR stays focused and easier to review, and I’ll handle support for RESET_PASSWORD group policies (with tests) in the next PR.
> 
> Thanks a lot! 🙏

Awesome job @Bagautdino, I agree that possible addition to allow setup reset-password for group resource should be handled in follow-up PR.

The PR looks almost ready to merge. I just realized it's missing test coverage (my apologies for not catching that sooner). Do you have the capacity to add tests?

Alternatively I can contribute the tests, probably as a PR to you branch.

services/src/main/java/org/keycloak/services/resources/admin/fgap/UserPermissionsV2.java
pedroigor: Sorry for the late reply. I'm not sure about introducing the `fallbackToManageUsers`. For me, we can just check if the reset password scope is set, and if not we check the manager scope.

Or is there any backward compatibility issue with this approach?

pedroigor:
@vramik @Bagautdino Sorry, I'm working on some changes to your branch, and GH closed the PR. Will let you know when I'm done.

pedroigor:
@vramik @Bagautdino I did some updates to your branch:

* Added the fallback, where the fallback is only executed if there are no permissions associated with the target scope (`reset-password`).
* Added a test

Sorry for the mess, but if you can, please re-open your PR or send another one. I appreciate it.

pedroigor:
@Bagautdino I really messed up your branch. As you are working on main, I accidentally pushed the wrong changeset to your branch.

Hopefully you can force-push the commits in your local branch?

pedroigor:
@Bagautdino Sent https://github.com/keycloak/keycloak/pull/42298. Can you please check it out? Please, feel free to rebase your branch with that one if you prefer to send/re-open a PR from your side.

Sorry ...

Bagautdino:
Okay! I'll do! just give me some time

Bagautdino:
Follow-up continues in #42307, which restores and replaces this PR

PR #41905, opened 2025-08-15
Adding TiDB dialect for Quarkus

### LINKED ISSUE #41897: Unstable tidb DB tests
### Before reporting an issue

### Area

ci

### Describe the bug

The test showed to be unstable in https://github.com/keycloak/keycloak/actions/runs/16966315188/job/48115619560?pr=41882

```
2025-08-14T18:21:00.1129051Z 2025-08-14 18:21:00,047 WARN  [managed.keycloak] (pool-2-thread-1) [liquibase.changelog.DatabaseChangeLog] (main) Due to mysql SQL limitations, setNullable will lose primary key/autoincrement/not null/comment settings explicitly redefined in the change. Use <sql> or <modifySql> to re-specify all configuration if this is the case
2025-08-14T18:21:37.2154982Z 2025-08-14 18:21:37,209 WARN  [managed.keycloak] (main) [org.hibernate.orm.jdbc.error] (executor-thread-1) HHH000247: ErrorCode: 1146, SQLState: 42S02
2025-08-14T18:21:37.2175552Z 2025-08-14 18:21:37,209 WARN  [managed.keycloak] (main) [org.hibernate.orm.jdbc.error] (executor-thread-1) Table 'test.ue1_0' doesn't exist
2025-08-14T18:21:37.2177621Z 2025-08-14 18:21:37,209 ERROR [managed.keycloak] (main) [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-1) Uncaught server error: org.keycloak.models.ModelException: Database operation failed
```

It seems that the import of the realm was not successful as a table was missing, and then later part of the realm were imported anyway? 

### Version

main

### Regression

### Expected behavior

DB migration should be stable, and failed imports should roll back

### Actual behavior

A failed import wasn't rolled back. During the first import, a table was probably not visible.

### How to Reproduce?

Unknown.

### Anything else?

It might be worth investigating the isolation levels both from a performance point of view, and also why the table was not visible?

https://docs.pingcap.com/tidb/stable/transaction-isolation-levels/

With PostgreSQL, we use read committed everywhere

ahus1:
Closes #41897

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/database/Database.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/database/Database.java
@@ -192,6 +192,18 @@ public final class Database {
                         getProperty(DatabaseOptions.DB_URL_PROPERTIES, namedProperty)),
                 List.of("org.keycloak.connections.jpa.updater.liquibase.UpdatedMySqlDatabase")
         ),
+        TIDB("tidb",
+                "com.mysql.cj.jdbc.MysqlXADataSource",
+                "com.mysql.cj.jdbc.Driver",
+                "org.hibernate.dialect.TiDBDialect",
+                // default URL looks like this: "jdbc:mysql://${kc.db-url-host:localhost}:${kc.db-url-port:3306}/${kc.db-url-database:keycloak}${kc.db-url-properties:}"
+                (namedProperty, alias) -> "jdbc:mysql://%s:%s/%s%s".formatted(
+                        getProperty(DatabaseOptions.DB_URL_HOST, namedProperty, "localhost"),
+                        getProperty(DatabaseOptions.DB_URL_PORT, namedProperty, "3306"),
+                        getProperty(DatabaseOptions.DB_URL_DATABASE, namedProperty, "keycloak"),
+                        getProperty(DatabaseOptions.DB_URL_PROPERTIES, namedProperty)),
+                List.of("org.keycloak.connections.jpa.updater.liquibase.UpdatedMySqlDatabase")
+        ),
         MARIADB("mariadb",
                 "org.mariadb.jdbc.MariaDbDataSource",
                 "org.mariadb.jdbc.Driver",
--- a/quarkus/config-api/src/main/java/org/keycloak/config/database/Database.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/database/Database.java
@@ -195,7 +195,7 @@ public final class Database {
         TIDB("tidb",
                 "com.mysql.cj.jdbc.MysqlXADataSource",
                 "com.mysql.cj.jdbc.Driver",
-                "org.hibernate.dialect.TiDBDialect",
+                "org.hibernate.community.dialect.TiDBDialect",
                 // default URL looks like this: "jdbc:mysql://${kc.db-url-host:localhost}:${kc.db-url-port:3306}/${kc.db-url-database:keycloak}${kc.db-url-properties:}"
                 (namedProperty, alias) -> "jdbc:mysql://%s:%s/%s%s".formatted(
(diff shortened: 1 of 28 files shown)


quarkus/deployment/pom.xml
```diff
@@ -158,8 +158,20 @@
...
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-hibernate-orm-dev</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-community-dialects</artifactId>
+        </dependency>
```
mabartos:
```suggestion
```

Having that in `quarkus/runtime/pom.xml` should be sufficient.

quarkus/server/pom.xml
```diff
@@ -21,6 +21,10 @@
             <groupId>org.keycloak</groupId>
             <artifactId>keycloak-quarkus-server</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-community-dialects</artifactId>
+        </dependency>
```
mabartos:
```suggestion
```

Having that in `quarkus/runtime/pom.xml` should be sufficient.

mabartos:
@ahus1 As you mentioned you have some troubles with Quarkus, could you please share a log with the exception? CI does not show anything as well for now. Thanks

quarkus/config-api/src/main/java/org/keycloak/config/database/Database.java
```diff
@@ -192,6 +192,18 @@ private String amendH2(String jdbcUrl) {
                         getProperty(DatabaseOptions.DB_URL_PROPERTIES, namedProperty)),
                 List.of("org.keycloak.connections.jpa.updater.liquibase.UpdatedMySqlDatabase")
         ),
+        TIDB("tidb",
+                "com.mysql.cj.jdbc.MysqlXADataSource",
+                "com.mysql.cj.jdbc.Driver",
+                "org.hibernate.dialect.TiDBDialect",
```
denniskniep: @ahus1  I think their docs have a typo.
Try to use `org.hibernate.community.dialect.TiDBDialect`. This worked for me.

denniskniep: And this needs to be changed to successfully run the arquillian tests:

https://github.com/denniskniep/keycloak/blob/0e0dc6fdd34da5e8a44c4b31026b768bf794277d/testsuite/integration-arquillian/pom.xml#L422

ahus1:
Thank you for the pointer @denniskniep - I opened an issue with TiDB docs, see https://github.com/pingcap/docs/issues/21618

ahus1:
@ryanemerson - could you please help out here to fix the remaining CI issues? Thanks! Also please review if adding the feature makes sense, of if the hidden flag to pass in the dialect should be used until we officially support TiDB.

denniskniep:
@ahus1 This works for me: https://github.com/denniskniep/keycloak/commit/bf5532386f848728eaa4fd65bb56b37324b8927a

see https://github.com/denniskniep/keycloak/actions/runs/17221143802

ryanemerson (approved):

pedroigor (approved):

ahus1:
@vmuzikar - we added one more Hibernate JAR for this. I don't know if it will be productized or not; maybe you can check in advance. 

After all, we don't plan to make tidb supported in 26.4, so we're good to exclude it from downstream if it is not productized. 

cc: @mabartos

vmuzikar:
@ahus1 The `org.hibernate.orm:hibernate-community-dialects` seems to be productized.

PR #41908, opened 2025-08-15
fix: removing the possiblity of an npe

### LINKED ISSUE #40069: Possible dereference of Null
### Description

Hi, everyone!
I'm continuing to perform a static analysis of the project.
Found a vulnerability potentially leading to dereferencing of a null variable not handled in a try/catch block.

The `rs` variable, which is checked for null value compliance and can potentially match it, is called on the line:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/services/src/main/java/org/keycloak/authentication/authenticators/x509/CertificateValidator.java#L740
which can lead to null dereference

The logger here:
https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/services/src/main/java/org/keycloak/authentication/authenticators/x509/CertificateValidator.java#L735
will not terminate the program, therefore, in the context of `rs == null` and `_ocspFailOpen == false`, dereference of null will occur when proceeding to execute the next if block.
I suggest changing the condition from `if` to `else if` in order to avoid an error

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
closes: #40069

This is mostly a non-issue. The actual implementation seems to either return a response or throw an exception - null doesn't currently seem possible.

Feel free to take a different direction, such as removing the null check altogether, or just closing this and #40069

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/x509/CertificateValidator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/x509/CertificateValidator.java
@@ -229,6 +229,7 @@ public class CertificateValidator {
         public CRLLoaderProxy(X509CRL crl) {
             _crl = crl;
         }
+        @Override
         public Collection<X509CRL> getX509CRLs() throws GeneralSecurityException {
             return Collections.singleton(_crl);
         }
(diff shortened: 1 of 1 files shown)


vmuzikar (approved):

vmuzikar:
@mposolda @rmartinc Do you want to review as well?

rmartinc (approved):
Thanks @shawkins! LGTM!

PR #41910, opened 2025-08-15
fix: disable devservices api server

shawkins:
closes: #41903

because of how we are taking over manually for creating the operator and don't have prior usage of devservices, we can just disable it for now.

At some point we can revisit doing this in a more standard way.

### DIFF
--- a/operator/src/test/resources/application.properties
+++ b/operator/src/test/resources/application.properties
@@ -2,7 +2,7 @@ quarkus.test.hang-detection-timeout=30m
 
 # we start operator manually during tests
 quarkus.operator-sdk.start-operator=false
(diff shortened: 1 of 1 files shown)


operator/src/test/resources/application.properties
```diff
@@ -2,6 +2,8 @@ quarkus.test.hang-detection-timeout=30m
 
 # we start operator manually during tests
 quarkus.operator-sdk.start-operator=false
+# we take responsibility for startin the apiserver
```
vmuzikar:
```suggestion
# we take responsibility for starting the apiserver
```

vmuzikar (approved):
I fixed only a small typo, LGTM otherwise.

PR #41919, opened 2025-08-15
Do not try to lookup the user if the sub claim is missing from the admin bearer token

### LINKED ISSUE #41098: Locked out after upgrade to 26.3.1 due to missing sub in lightweight access token
### Before reporting an issue

### Area

admin/ui

### Describe the bug

I upgraded from 26.2.5 to 26.3.1 and could not load security-admin-console. I was getting a 500 and [this stack trace](https://gist.github.com/mmelvin0/9caac40d2c71c23be33852e7b805c16a) on the my `/admin/serverinfo` endpoint.

Was able to determine that the `sub` claim was missing from my access token and that was causing the issue. Similar issue #40249.

After downgrading to 26.2.5 I found I do have the `sub` mapper in the `basic` scope and it was set to add to regular access tokens but not lightweight access tokens. I also noted that my security-admin-console client has the "always use lightweight access token" turned on. I believe that these are both at default settings added by earlier migrations and I haven't modified them, but cannot be 100% sure. I'm pretty confident though as it affected two different realms the same way, my master realm, which I have barely changed any options on except to lock down, and another realm which I have heavily customized.

I flipped the switch to "Add to lightweight access token" on the `sub` mapper and after that upgrading back to 26.3.1 I was able to log in to `security-admin-console`.

### Version

26.3.1

### Regression

### Expected behavior

To be able to login in to the Keycloak administration UI.

### Actual behavior

After logging in to the administration UI, I am greeted with a blank white page.

### How to Reproduce?

Upgrade from Keycloak 26.2.5 to 26.3.1 with default `basic` client scope and `security-admin-console` configurations.

### Anything else?

_No response_

pedroigor:
Closes #41098
Closes #41799

### DIFF
--- a/services/src/main/java/org/keycloak/services/DefaultKeycloakContext.java
+++ b/services/src/main/java/org/keycloak/services/DefaultKeycloakContext.java
@@ -304,9 +304,10 @@ public abstract class DefaultKeycloakContext implements KeycloakContext {
             String issuer = jwt.getIssuer();
             String realmName = issuer.substring(issuer.lastIndexOf("/") + 1);
             RealmModel realm = session.realms().getRealmByName(realmName);
+            String id = jwt.getSubject();
 
-            if (realm != null) {
-                user = session.users().getUserById(realm, jwt.getSubject());
+            if (realm != null && id != null) {
+                user = session.users().getUserById(realm, id);
             }
         }
 


mposolda (approved):

mposolda:
@pedroigor Nice, Thanks!

PR #41920, opened 2025-08-15
Add Support for UUID v7 Generation for User IDs

### LINKED ISSUE #30459: Is there any plan to migrate user id from uuid v4 to uuid v7?
### Description

Hi,

There are a lot of discussion about performance issue in database with uuid v4 and many applications are moving from uuid v4 to uuid v7 to reduce that performance impact.

Is there any plan to migrate user ids from v4 to v7?

Or is there any settings available to choose it by server administration now?

### Discussion

_No response_

### Motivation

Personally, in my app I have uuid v7 selected for all records, and in some cases there are records that get the id from the jwt authenticated user instead of the local v7 function, so in the database there is a mix of v4 and v7 IDs I have that it makes the performance issue.

### Details

_No response_

kcrandall:
<html><head></head><body><h1>Add Support for UUID v7 Generation for User IDs</h1>
<h2>Summary</h2>
<p>This PR introduces optional UUID v7 generation for user creation in Keycloak, providing significant database performance improvements while maintaining full backward compatibility through an environment variable configuration.</p>
<p><strong>Fixes #30459</strong></p>
<h2>What changed?</h2>
<ul>
<li>Added <code>generateIdv7()</code> method to <code>KeycloakModelUtils</code> for UUID v7 generation</li>
<li>Modified <code>UsersPartialImport.create()</code> method to conditionally use UUID v7 based on environment variable <code>KC_USER_UUID_V7</code></li>
<li>Maintained 100% backward compatibility - existing deployments are unaffected</li>
</ul>
<h2>Motivation</h2>
<h3>Primary Use Case: External Database Performance</h3>
<p>This enhancement primarily targets deployments where <strong>Keycloak user IDs are used as primary keys in external databases</strong>. Many organizations use Keycloak user IDs as foreign keys or primary keys in their application databases, creating performance bottlenecks due to random UUID v4 insertion patterns.</p>
<h3>Database Performance Issues with UUID v4</h3>
<p>When Keycloak user IDs are used as primary keys in external databases, UUID v4 (random UUIDs) create several performance challenges:</p>
<ol>
<li><strong>Index Fragmentation</strong>: Random UUIDs cause B-tree index fragmentation as new entries are inserted randomly throughout the index structure</li>
<li><strong>Page Splits</strong>: Frequent page splits in database indexes due to non-sequential inserts</li>
<li><strong>Poor Cache Locality</strong>: Random ordering reduces database buffer pool efficiency</li>
<li><strong>Slower Inserts</strong>: Non-sequential primary keys require more disk I/O for index maintenance</li>
<li><strong>Increased Storage Overhead</strong>: Fragmented indexes consume more disk space</li>
<li><strong>Cross-Table Join Performance</strong>: Random primary keys degrade join performance in related tables</li>
</ol>
<h3>Benefits of UUID v7</h3>
<p>UUID v7 addresses these issues by incorporating a timestamp component while maintaining UUID compatibility:</p>
<ul>
<li><strong>Sequential Ordering</strong>: UUIDs are naturally ordered by creation time, improving insert performance</li>
<li><strong>Reduced Fragmentation</strong>: New entries append to the end of indexes instead of random insertion</li>
<li><strong>Better Performance</strong>: 20-40% improvement in insert performance for high-volume scenarios</li>
<li><strong>Improved Queries</strong>: Range queries by creation time become more efficient</li>
<li><strong>Smaller Indexes</strong>: Reduced fragmentation leads to more compact index storage</li>
<li><strong>External Database Benefits</strong>: Dramatically improves performance when Keycloak user IDs are used as primary keys in application databases</li>
</ul>
<h2>Technical Details</h2>
<h3>UUID v7 Structure</h3>
<pre><code> 0                   1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
├─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┤
|                           unix_ts_ms                          |
├─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┤
|          unix_ts_ms           |  ver  |       rand_a          |
├─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┤
|var|                        rand_b                             |
├─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┼─┤
|                            rand_b                             |
└─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┘
</code></pre>
<ul>
<li><strong>48-bit timestamp</strong>: Milliseconds since Unix epoch (good until year 10889)</li>
<li><strong>4-bit version</strong>: Always <code>0111</code> (7)</li>
<li><strong>12-bit random A</strong>: High randomness for uniqueness</li>
<li><strong>2-bit variant</strong>: Always <code>10</code></li>
<li><strong>62-bit random B</strong>: Additional randomness ensuring uniqueness</li>
</ul>
<h3>Backward Compatibility Strategy</h3>
<p>The implementation uses an <strong>opt-in approach</strong> to ensure zero impact on existing deployments:</p>
<ol>
<li><strong>Default Behavior Unchanged</strong>: Without the environment variable, Keycloak continues using UUID v4</li>
<li><strong>Environment Variable Gated</strong>: Only activated when <code>KC_USER_UUID_V7=true</code> is explicitly set</li>
<li><strong>Same Format</strong>: UUID v7 maintains the standard 36-character UUID format</li>
<li><strong>Database Compatible</strong>: Existing UUID columns work unchanged with UUID v7 values</li>
<li><strong>No Migration Required</strong>: Existing users keep their original UUIDs</li>
</ol>

<p><em>Note: Actual results depend on database engine, hardware, and workload patterns</em></p>
<h2>Configuration</h2>
<h3>Environment Variable</h3>
<pre><code class="language-bash">export KC_USER_UUID_V7=true
</code></pre>
<h2>Safety Considerations</h2>
<h3>Why Environment Variable is Safe</h3>
<ol>
<li><strong>Explicit Opt-in</strong>: Administrators must consciously enable the feature</li>
<li><strong>No Automatic Activation</strong>: Feature remains dormant unless explicitly configured</li>
<li><strong>Clear Documentation</strong>: Environment variable name clearly indicates its purpose</li>
<li><strong>Reversible</strong>: Can be disabled by removing/changing the environment variable</li>
<li><strong>Testing Friendly</strong>: Easy to test in staging environments before production</li>
</ol>
<h3>Migration Strategy for Existing Deployments</h3>
<p>For organizations wanting to adopt UUID v7:</p>
<ol>
<li><strong>New Installations</strong>: Set environment variable from day one</li>
<li><strong>Existing Deployments</strong>:
<ul>
<li>Test in staging environment first</li>
<li>Enable for new users only (existing users keep UUID v4)</li>
<li>Mixed UUID versions are fully supported</li>
<li>Monitor database performance improvements</li>
</ul>
</li>
</ol>
<h3>UUID Version Coexistence</h3>
<p>The system gracefully handles mixed UUID versions:</p>
<ul>
<li>Existing users retain their UUID v4 IDs</li>
<li>New users get UUID v7 IDs (if enabled)</li>
<li>All Keycloak features work identically with both versions</li>
<li>Database queries work seamlessly across both formats</li>
</ul>
<h2>Implementation Details</h2>
<h3>Files Modified</h3>
<ul>
<li><code>server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java</code></li>
<li><code>services/src/main/java/org/keycloak/partialimport/UsersPartialImport.java</code></li>
</ul>
<h3>New Methods Added</h3>
<ul>
<li><code>generateIdv7()</code> - New UUID v7 generator method</li>
</ul>
<h3>Modified Methods</h3>
<ul>
<li><code>UsersPartialImport.create()</code> - Added conditional UUID v7 generation based on <code>KC_USER_UUID_V7</code> environment variable</li>
</ul>
<h2>Testing</h2>
<h3>Unit Tests</h3>
<ul>
<li>[x] UUID v7 format validation (<code>testUUIDv7()</code>)</li>
<li>[x] Timestamp extraction accuracy</li>
<li>[x] Version and variant bit validation</li>
<li>[x] Basic functionality verification</li>
</ul>

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
@@ -71,7 +71,9 @@ import java.security.Key;
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.security.PublicKey;
+import java.security.SecureRandom;
 import java.security.cert.X509Certificate;
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Calendar;
@@ -100,6 +102,7 @@ import static org.keycloak.utils.StreamsUtil.closing;
 public final class KeycloakModelUtils {
 
     private static final Logger logger = Logger.getLogger(KeycloakModelUtils.class);
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
 
     public static final String AUTH_TYPE_CLIENT_SECRET = "client-secret";
     public static final String AUTH_TYPE_CLIENT_SECRET_JWT = "client-secret-jwt";
@@ -119,6 +122,39 @@ public final class KeycloakModelUtils {
      */
     public static String generateId() {
         return UUID.randomUUID().toString();
+    }
+
+    /**
+     * Generate a UUID v7 (time-ordered UUID)
+     * 
+     * @return UUID v7 instance
+     */
+    public static String generateIdv7() {
+        long timestamp = Instant.now().toEpochMilli();
+        
+        // Generate random bytes for the remaining parts
+        byte[] randomBytes = new byte[10];
+        SECURE_RANDOM.nextBytes(randomBytes);
+        
+        // Construct MSB: 48-bit timestamp + 4-bit version + 12-bit random
+        long msb = (timestamp << 16) |
+                   (0x7L << 12) |
+                   (((long) randomBytes[0] & 0x0F) << 8) |
+                   ((long) randomBytes[1] & 0xFF);
+        
+        // Construct LSB: 2-bit variant + 62-bit random
+        long lsb = (0x8L << 60) |
+                   ((long) (randomBytes[2] & 0x3F) << 56) |
+                   ((long) (randomBytes[3] & 0xFF) << 48) |
+                   ((long) (randomBytes[4] & 0xFF) << 40) |
+                   ((long) (randomBytes[5] & 0xFF) << 32) |
+                   ((long) (randomBytes[6] & 0xFF) << 24) |
+                   ((long) (randomBytes[7] & 0xFF) << 16) |
+                   ((long) (randomBytes[8] & 0xFF) << 8) |
+                   (randomBytes[9] & 0xFF);
+
+        UUID uuid = new UUID(msb, lsb);
+        return uuid.toString();
     }
 
     /**
--- a/server-spi-private/src/test/java/org/keycloak/models/utils/KeycloakModelUtilsTest.java
+++ b/server-spi-private/src/test/java/org/keycloak/models/utils/KeycloakModelUtilsTest.java
@@ -18,8 +18,11 @@
 package org.keycloak.models.utils;
 
 import java.nio.ByteBuffer;
+import java.time.Instant;
 import java.util.Base64;
 import java.util.UUID;
+import java.util.regex.Pattern;
+
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -28,6 +31,11 @@ import org.junit.Test;
  * @author rmartinc
  */
 public class KeycloakModelUtilsTest {
+
+    private static final Pattern UUID_PATTERN = Pattern.compile(
+            "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+    );
+
 
     @Test
     public void testGenerateId() {
@@ -53,5 +61,45 @@ public class KeycloakModelUtilsTest {
         final long msb = bb.getLong();
         final long lsb = bb.getLong();
         return new UUID(msb, lsb);
+    }
+
+    @Test
+    public void testUUIDv7() {
+        try {
+            String id = KeycloakModelUtils.generateIdv7();
+            
+            // First, validate it's a proper UUID format
+            Assert.assertNotNull("Generated ID should not be null", id);
+            Assert.assertEquals("Generated ID should be 36 characters", 36, id.length());
+            Assert.assertTrue("Generated ID should match UUID format", UUID_PATTERN.matcher(id).matches());
+            
+            // Parse as UUID to ensure it's valid
+            UUID uuid = UUID.fromString(id);
+            Assert.assertNotNull("UUID should parse successfully", uuid);
+            
+            // Direct test for UUID v7 since we're calling generateIdv7() directly
+            Assert.assertEquals("Should generate UUID v7", 7, uuid.version());
+            
+            // Test UUID v7 variant bits (should be 10xxxxxx in binary)
+            long lsb = uuid.getLeastSignificantBits();
+            long variantBits = (lsb >>> 62) & 0x3;
+            Assert.assertEquals("UUID v7 variant bits should be 10 (binary)", 2, variantBits);
+            
+            // Test timestamp extraction
+            long timestamp = uuid.getMostSignificantBits() >>> 16;
+            long currentTime = Instant.now().toEpochMilli();
+            Assert.assertTrue("Timestamp should be recent", 
+                            Math.abs(currentTime - timestamp) < 5000); // Within 5 seconds
+            Assert.assertTrue("Timestamp should be positive", timestamp > 0);
+            
+            // Test that timestamp is reasonable (after year 2020, before year 2100)
+            long year2020 = 1577836800000L; // 2020-01-01 00:00:00 UTC
+            long year2100 = 4102444800000L; // 2100-01-01 00:00:00 UTC
+            Assert.assertTrue("Timestamp should be after 2020", timestamp > year2020);
+            Assert.assertTrue("Timestamp should be before 2100", timestamp < year2100);
+            
+        } catch (IllegalArgumentException e) {
+            Assert.fail("Generated ID should be a valid UUID: " + e.getMessage());
+        }
     }
 }
--- a/services/src/main/java/org/keycloak/partialimport/UsersPartialImport.java
+++ b/services/src/main/java/org/keycloak/partialimport/UsersPartialImport.java
@@ -41,6 +41,9 @@ public class UsersPartialImport extends AbstractPartialImport<UserRepresentation
     // so we cache the created id here.
     private final Map<String, String> createdIds = new HashMap<>();
 
+    private static final boolean USE_USER_UUID_V7 = Boolean.parseBoolean(System.getenv("KC_USER_UUID_V7")
+);
+
     @Override
     public List<UserRepresentation> getRepList(PartialImportRepresentation partialImportRep) {
         return partialImportRep.getUsers();
@@ -111,7 +114,11 @@ public class UsersPartialImport extends AbstractPartialImport<UserRepresentation
     @Override
     public void create(RealmModel realm, KeycloakSession session, UserRepresentation user) {
         if (user.getId() == null) {
-            user.setId(KeycloakModelUtils.generateId());
+            if (USE_USER_UUID_V7) {
+                user.setId(KeycloakModelUtils.generateIdv7());
+            } else {
+                user.setId(KeycloakModelUtils.generateId());
+            }
         }
         UserModel userModel = RepresentationToModel.createUser(session, realm, user);
         if (userModel == null) throw new RuntimeException("Unable to create user " + getName(user));


pedroigor (commented):
I need to catch up on UUID v7. Is your implementation based on https://www.rfc-editor.org/rfc/rfc9562.html?

I think we need more changes to make sure this change affects and possibly make sure we have a single place to decide which UUID version to use so that everywhere is affected.

We also probably want this configuration either as a user provider configuration. By doing this, you should be able to set the version via CLI, env, or sys properties.

I'll check with others in the team about the risks/concerns of such a change. I agree with you initially that it should not break existing deployments, and different UUID versions can co-exist. But I'm not 100% sure.

kcrandall:
I'm pretty sure its that RFC but its the version included in Postgres 18 
See:
https://www.postgresql.org/docs/18/release-18.html#RELEASE-18-FUNCTIONS
https://www.thenile.dev/blog/uuidv7

I am unfamiliar with how Keycloak is architected so I tried to make minimal changes to the repo, but another option would be to change the env variable to something like:
KC_UUID_VERSION = 7

This would be more portable to future versions that may come out in next 10 years, right now there is only v4 and v7 that really matter. v4 is more secure because its more random and v7 is ordered so its easier to do a brute force attack (still very very hard because last part is still random) to know a users id if you know the time when the uuid was created. Thats why it should be users choice which uuid they want to use in Keycloak. 

The other reason why i named the env var with USER is because i didnt want to let users think this affects UUID in the entire app because i didnt want to override the current generateId() method as that might have side affects so i made a new one. But you could inject the environment variable into that function is my guess and I assume it would change it everywhere if the rest of the codebase uses that utility.

pedroigor:
> I'm pretty sure its that RFC but its the version included in Postgres 18 See: https://www.postgresql.org/docs/18/release-18.html#RELEASE-18-FUNCTIONS https://www.thenile.dev/blog/uuidv7
> 
> I am unfamiliar with how Keycloak is architected so I tried to make minimal changes to the repo, but another option would be to change the env variable to something like: KC_UUID_VERSION = 7
> 
> This would be more portable to future versions that may come out in next 10 years, right now there is only v4 and v7 that really matter. v4 is more secure because its more random and v7 is ordered so its easier to do a brute force attack (still very very hard because last part is still random) to know a users id if you know the time when the uuid was created. Thats why it should be users choice which uuid they want to use in Keycloak.
> 
> The other reason why i named the env var with USER is because i didnt want to let users think this affects UUID in the entire app because i didnt want to override the current generateId() method as that might have side affects so i made a new one. But you could inject the environment variable into that function is my guess and I assume it would change it everywhere if the rest of the codebase uses that utility.

It is a sensitive change. Perhaps we want to pick a version when creating a database. There is also concerns about the lack of support from Java e perhaps we want to wait until we get v7 supported.  

From a security PoV, exposing the creation date for resources, like users or sessions, can make brute force easier? We need to evaluate the tradeoffs with careful even if we make it configurable.

By using it everywhere, I meant when persisting data. I guess the main arguments for this change are improvements to persisting/fetching data from the database. Some benchmarks based on a database with loads of users, for instance, would be helpful.

kcrandall:
From a user point of view the most important thing is the user id so you can use that as your primary key in your db so the auth layer doesn’t need to do a SELECT to match an internal uuid with keycloak. Every other resource in keycloak doesn’t matter to me personally. It’s just because the user id is used in so so many tables as a foreign key in almost every db. 

Basically more bytes in uuid are based on the timestamp in v7 vs v4. The first many bytes are directly convertible to a timestamp where the remaining part is completely random. Uuid4 most of the bytes are random and not based on time. I think the risks are low especially with a user id. By risks I basically meant you know with an integer primary key you can just integrate 1,2,3,… on a rest api? On this you could use the first half and then brute force the second half to try to find users who were created on the second half of the uuid. It would still take forever. I said “technically” below is what Claude told me the number of combinations you’d have to generate for the exact millisecond.

“ If you know the exact timestamp in a UUID v7, you significantly reduce the total entropy but still retain substantial uniqueness. UUID v7 uses a 48-bit Unix timestamp in milliseconds in the most significant bits, while the remaining 74 bits are random (with some bits reserved for version and variant identifiers). This means that if you know the precise millisecond timestamp, you’re left with approximately 2^74 possible combinations, which equals roughly 1.9 × 10^22 unique values. Even with this reduced entropy, the number of possible UUIDs for any given millisecond is astronomically large - far exceeding what any single system could generate in that timeframe, ensuring that UUID v7 maintains its collision-resistant properties even when the timestamp component is known.​​​​​​​​​​​​​​​​”


https://github.com/equenum/postgre_uuid_performance
# PostgreSQL UUID v7 vs v4 Performance Comparison

## Performance Metrics Table

|Task                             |UUID v4         |UUID v7         |Improvement             |Source            |
|---------------------------------|----------------|----------------|------------------------|------------------|
|**Insert Performance (10M rows)**|615 seconds     |410 seconds     |**34.8% faster**        |Dev.to benchmark  |
|**TPS during inserts**           |2,670 tps       |3,420 tps       |**28% higher**          |Ardent Performance|
|**Index Size (10M rows)**        |777 MB          |603 MB          |**22% smaller**         |Dev.to benchmark  |
|**Total Disk Usage (10M rows)**  |1.2 GB          |1.025 GB        |**175 MB smaller**      |Dev.to benchmark  |
|**Point Lookup Query Time**      |2.4-2.6 ms      |0.1-0.4 ms      |**85% faster**          |Multiple sources  |
|**Range Scan Performance**       |Slower execution|Faster execution|**Significantly faster**|Dev.to benchmark  |
|**B-tree Page Splits**           |Frequent        |Minimal         |**Dramatically reduced**|Multiple sources  |
|**Buffer Cache Efficiency**      |Poor locality   |Good locality   |**Major improvement**   |Ardent Performance|
|**Planning Time (queries)**      |0.5-12 ms       |0.2-0.5 ms      |**~50% faster**         |Seven’s Blog      |
|**Execution Time (queries)**     |2.3-2.8 ms      |0.1-0.4 ms      |**80-90% faster**       |Seven’s Blog      |

## Key Performance Characteristics

### **Insert Performance**

- **UUID v4**: Causes random B-tree insertions leading to frequent page splits
- **UUID v7**: Sequential insertions with minimal page splits due to timestamp ordering
- **Result**: UUID v7 is consistently 25-35% faster for bulk inserts

### **Storage Efficiency**

- **UUID v4**: Index fragmentation leads to ~20-25% larger indexes
- **UUID v7**: Better fill factor and less fragmentation
- **Result**: UUID v7 uses significantly less disk space

### **Query Performance**

- **Point Lookups**: UUID v7 shows 80-90% improvement in execution time
- **Range Scans**: UUID v7 benefits from sequential data layout
- **Index Scans**: Better cache locality with UUID v7

### **Memory Usage**

- **Buffer Cache**: UUID v4 requires more buffer cache space due to scattered index pages
- **Working Set**: UUID v7 reduces total working set size by improving data locality

## Test Environment Details

**Hardware Specifications:**

- CPU: 8 cores / AMD EPYC Genoa 3.7 GHz
- Memory: 16 GB RAM
- Storage: NVMe SSD / 16k IOPS gp3
- Database: PostgreSQL 16/17 (with UUID v7 patch)

**Test Parameters:**

- Dataset size: 10 million rows
- Batch size: 10,000 rows per insert batch
- Concurrent queries during inserts
- B-tree indexes (default PostgreSQL primary key index type)

## Performance Scaling Observations

1. **Small datasets (<100k rows)**: Minimal difference between v4 and v7
1. **Medium datasets (1M+ rows)**: UUID v7 advantages become noticeable
1. **Large datasets (10M+ rows)**: UUID v7 shows dramatic performance benefits
1. **Write-heavy workloads**: UUID v7 advantages are most pronounced

## Sources

- [PostgreSQL UUID Performance Benchmarking](https://dev.to/umangsinha12/postgresql-uuid-performance-benchmarking-random-v4-and-time-based-v7-uuids-n9b)
- [UUID Benchmark War - Ardent Performance](https://ardentperf.com/2024/02/03/uuid-benchmark-war/)
- [Comparing UUIDv4 vs UUIDv7 - Seven’s Blog](https://www.dujinfang.com/2023/09/02/uuidv4-and-uuidv7.html)
- [PostgreSQL UUID Primary Key Performance](https://maciejwalkowiak.com/blog/postgres-uuid-primary-key/)
- [UUID Downsides in PostgreSQL - Cybertec](https://www.cybertec-postgresql.com/en/unexpected-downsides-of-uuid-keys-in-postgresql/)

**Note**: PostgreSQL 18 will include native UUID v7 support via the `uuidv7()` function. Prior versions require third-party libraries or extensions to generate UUID v7 values.

kcrandall:
With your comment at Java support build in, I don’t think it’s a big of a deal the algorithm to generate it is pretty small. Can just swap out the util for official version if it ever comes out. Don’t think it’s worth waiting for an official release in Java. 

Here is a lib, but I think it’s better just to make a simple util for something so simple.
https://github.com/cowtowncoder/java-uuid-generator

C# got it last year with dotnet 9, but I’m not sure if Java stays up to date as fast as dotnet does. Before the offial uuidv7 people were using ulid which is similar, uuidv7 is just the official version trying to solve the same thing that ulid has been around for since like 2016. They all are just random strings with different structures so usually pretty easy to interchange. They all fit into the same data type.

stianst (changes requested):
We shouldn't add UUIDv7 support until it's an accepted standard, at the moment it's a proposed standard. Ideally we should also wait for Java to support it, but if there's enough demand we can always do it earlier.

Bear in mind also that UUIDv7 are longer than the current UUIDs we generate, this affects UI, tokens, etc.

We'd probably also want to consider using UUIDv7 everywhere, or at least for one type of entities like users. As in this PR it is only using UUIDv7 when partially importing users into an existing realm, which makes no sense to only have it there.

Finally, we don't use environment variables to expose configuration options. Not sure how we would want to make this configurable, but not through environment variables that's for sure.

stianst:
Maybe what would make sense is to add an SPI/provider for ID generation? Default provider can by like we do today, and we could add an experimental one for UUIDv7? That makes it possible for folks to use custom generators, and also provides a nice way to configure it, including the ability to have experimental, preview, providers.

kcrandall:
> Bear in mind also that UUIDv7 are longer than the current UUIDs we generate, this affects UI, tokens, etc.

UUIDv7 is the same length as UUIDv4. they are completely interchangeable. you can insert a v4 into a database table column with all v7s and the reverse. they are both 128-bit. the only difference is how the bits are generated

stianst:
> > Bear in mind also that UUIDv7 are longer than the current UUIDs we generate, this affects UI, tokens, etc.
> 
> UUIDv7 is the same length as UUIDv4. they are completely interchangeable. you can insert a v4 into a database table column with all v7s and the reverse. they are both 128-bit. the only difference is how the bits are generated

You're right, my bad. I just looked at an example UUIDv7 and thought that's long(er) without actually checking ;)

pedroigor:
An SPI makes more sense for now, but we still need to look at the impact of changing the format and usage when designing its contract.

@kcrandall Wdyt?

kcrandall:
Ya I think that’s fine. Especially if keycloak was shipped with a way to configure a few options such as uuid4 and uuid7 that you could configure easily and then the option to provide your own as well. might be a bit over engineering something that is pretty simple. 

The only thing I question is will it ever be used for anything other than uuid versions? Like I could see that being more useful if you could change user ids lots of different things like ints or other data types.

I do feel like just have a configuration for uuid version(3,4,7) and marking it as experimental for now might be more straightforward unless you see a future where this would have more options then just uuid versions. 

Also I’m not that familiar with how that is done in Java. In dotnet you would make an interface and register with DI IService container with extension methods. I assume it’s something similar in Java.

stianst:
SPI/Provider is really not over-engineering it, on the contrary really as there is no need to introduce config options, etc. I would imagine we'd start really simple and only support generating IDs and use the same provider for all IDs. It really is 10-15 minutes of effort to do. What might make it a bit more complicated is that KeycloakModelUtils.generateId doesn't have access to the KeycloakSession, which would require a bit more fiddling with the code to be able to obtain the provider, etc. Not particularly hard though.

What would make things more complicated is when we want to use different ID formats for different entities (say users have UUIDv7, while tokens or clients use something else).

pedroigor:
@kcrandall Perhaps you can start looking at this branch https://github.com/pedroigor/keycloak/tree/uuid-spi. 

It provides the bare minimum to introduce a `UUIDSpi`. Implementations of it will be based on implementing a `UUIDProviderFactory` and a `UUIDProvider`. As you can see from there, there is a `UUIDv4Provider|Factory` implementation that will generate UUIDs based on v4.

You can follow the same idea to implement a UUID v7 provider.

To set which provider the server should use at runtime, you can set a configuration option like `spi-uuid--provider-default=uuid-v7`.

In that branch, I've also updated the `UserCreateTest` to run using the UUID v7 provider. Just to allow you to check how the mechanism works when choosing a specific SPI provider implementation.

There are quite a few things yet to look at but I hope this will give you something to start with. As Stian said, we might want to have UUID versions set on a per-resource-type basis (e.g.: users, realm, clients, tokens, sessions, etc) and a mechanism to decide which provider to use based on the resource type.

However, if we can start with users, perhaps we can already come up with something that you and others can evaluate.

kcrandall:
@pedroigor that branch looks very good, I was missunderstanding how this flow would work in the project. I think starting with users is the best way to go. Its the use case that 99% of people would care about the most. I think extending other objects would be something to do as people ask for it. It looks like you added the uuidv7 provider so this looks like its good to go?

pedroigor:
> It looks like you added the uuidv7 provider so this looks like its good to go?

It depends on how we want to use this SPI with different resource types. If we want to start with users while at the same time be prepared to expand to other realm resource types (clients, groups, sessions, etc), tokens (not sure if it makes sense), etc, we need more work.

I would like to have an issue open for discussion where we provide at least an overview of where UUIDs are used, so that we can build a plan about how we are going to refactor the code to allow changing the UUID format. It should help to get more insights in the design of this SPI and how we are going to configure it.

Another thing we need to consider is testing. And I would like to run some benchmarks using the `keycloak-benchmark` tool.

But yeah, if we want to start with users. I think that should be enough. I also need to check with @stianst and others how much we should prioritize it.

stianst:
Why start with users and not just do all UUIDs? Simplest approach is just to use UUIDv7 everywhere, they have the same format, so shouldn't really matter, and if it helps databases to deal with them more efficient then that's a win for everything?

kcrandall:
I would have no problem with doing it everywhere. Whatever is easier

stianst:
@pedroigor I think we should just start with it everywhere, to see if that works fine and as you said if we can get any perf benefits from it. It would be the non-default provider, and marked as experimental for now I'd say/

pedroigor:
> I think we should just start with it everywhere

By everywhere, we are talking about realm resource types like:

* Users
* Roles
* Groups
* Clients
* Components
* Etc

But not:

* Tokens
* Sessions
* Need to check for more

For tokens, I don't think it makes much sense because we do not persist them. For sessions, using UUIDv7 can potentially help with session fixation|enumeration|hijack?

That is why I would like to make it clearer the places where this SPI would apply and start with users to evaluate the benefits of using it to later expand to others.

But if we can at least decide on the resources we want to allow to change the UUID, I'm also open to having it "everywhere".

stianst:
By everywhere I mean everywhere. 

Tokens, sessions, events, etc. are (or can be) stored in the database as well. UUIDv7 still has sufficient randomness, and are unique, we also don't ever rely on just a UUID to prove ownership of anything like a token or session. We use signatures for that.

I'd be more than open to have the SPI support different types of IDs in the future, perhaps something like `#generateId(IdType type)`.

However, the first thing we should really do is just have all IDs replaced by UUID v7 with an experimental provider, so some benchmarking can be done easily to see if there is any real value.

ahus1:
@stianst - I disagree with your statement as described in #38663

> Tokens, sessions, events, etc. are (or can be) stored in the database as well. UUIDv7 still has sufficient randomness, and are unique, we also don't ever rely on just a UUID to prove ownership of anything like a token or session. We use signatures for that.

The requirements of NIST and ANSSI differ here: In some places at least 128 bits or randomness are required, and UUIDv7 only has 74 random bits while the others are a timestamp. So UUIDv7 is not a fit for those places.

stianst:
> @stianst - I disagree with your statement as described in #38663
> 
> > Tokens, sessions, events, etc. are (or can be) stored in the database as well. UUIDv7 still has sufficient randomness, and are unique, we also don't ever rely on just a UUID to prove ownership of anything like a token or session. We use signatures for that.
> 
> The requirements of NIST and ANSSI differ here: In some places at least 128 bits or randomness are required, and UUIDv7 only has 74 random bits while the others are a timestamp. So UUIDv7 is not a fit for those places.

Could you elaborate? We shouldn't be using UUIDs in places where it's being used as more than an identifier. Take tokens for instances it's the signature that provides the security, not the jti.

`KeycloakModelUtils.generateId` is to generate a unique ID, not a secure reference. If it's used to generate "secrets" that's incorrect use.

stianst:
After discussing with @ahus1 I think this needs a bit more consideration, and review of the codebase before we introduce something like this

pedroigor:
> After discussing with @ahus1 I think this needs a bit more consideration, and review of the codebase before we introduce something like this

+1

francescapiacentini-satispay:
Hi, do you have an ETA for when this PR will be completed? Also, do you know which release it is targeted for?

ahus1:
Thank you for reminding me of this issue. 

Given some latest security requirements and reviews, we don't want people to guess any generated IDs or see collisions. Given the requirements stated above for example for tokens, this would bei 128bit of randomness. UUIDv7 doesn't show that.

Adding v7 for generating UUIDs for users and other entities would reveal their creation time. This would leak information, and might be used as OSINT. 

With that, I'm refusing the change to  UUIDv7 for user IDs and other entities. 

The only place where this could be used would be places where we already include the timestamp. The only place where this would then be a fit would be the IDs of user and admin events. 

As this PR is about users, I'm closing it. I'll be updating the parent issue that this might be a fit for those IDs. At the same time, this will be low priority as long term and for security reasons, storing admin and user events for auditing reasons in Keycloak's own database is a bad idea.

PR #41922, opened 2025-08-15
Checking password and OTP are present and can be converted

vivekksk:
closes #41640


sguilhen:
@pedroigor can you trigger the CI for this one?

sguilhen:
Hi @vivekksk, thank you for being the first to submit a fix for this issue - your approach of validating the required fields and throwing a `ModelException` is the correct one, and you identified it almost a year ago.

PR #49316 independently arrived at the same approach and also includes an integration test, so we'll be moving forward with that one. We're closing this PR as superseded, but want to acknowledge that you got the fix right first. Thank you for the contribution!

PR #41923, opened 2025-08-16
Add `org.owasp.dependencycheck`

Pankraz76:
As in email mentioned.

### DIFF
--- a/.github/workflows/owasp-analysis.yml
+++ b/.github/workflows/owasp-analysis.yml
@@ -1,0 +1,26 @@
+name: OWASP
+
+on: 
+  workflow_dispatch:
+
+env:
+  MAVEN_ARGS: "-B -nsu -Daether.connector.http.connectionMaxTtl=25"
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
+jobs:
+  analysis:
+    name: OWASP dependency-check
+    runs-on: ubuntu-latest
+    if: github.repository == 'keycloak/keycloak'
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Build Keycloak
+        uses: ./.github/actions/build-keycloak
+      - name: OWASP dependency-check
+        run: ./mvnw -e --no-transfer-progress dependency-check:check
--- a/.github/workflows/schedule-nightly.yml
+++ b/.github/workflows/schedule-nightly.yml
@@ -41,6 +41,7 @@ jobs:
         - codeql-analysis.yml
         - snyk-analysis.yml
         - trivy-analysis.yml
+        - owasp-analysis.yml
 
     steps:
       - name: Run workflow
@@ -58,6 +59,8 @@ jobs:
       matrix:
         workflow:
           - snyk-analysis.yml
+          - trivy-analysis.yml
+          - owasp-analysis.yml
 
     steps:
       - run: echo ${{ needs.setup.outputs.latest-release-branch }}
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -1314,6 +1314,22 @@
...
+                        <failBuildOnCVSS>3</failBuildOnCVSS>
+                    </configuration>
+                    <executions>
+                        <execution>
+                            <phase>verify</phase>
+                            <goals>
+                                <goal>check</goal>
+                            </goals>
+                        </execution>
+                    </executions>
```
Pankraz76:
```suggestion
```
as executed on CI, right?

Pankraz76:
here it's surprisingly no problem, how it should be the desired case:

- https://github.com/apache/kafka/pull/20364

stianst (changes requested):
I'm not sure what email you are referring to, but don't think we would want OWASP DependencyCheck integrated in this way.

We're also already using Snyk and Trivy. OWASP will most likely just report the same CVEs again. From what I found in the past OWASP also reports a number of false positives (for example matching dependencies with a different dependency because some parts of the name is shared).

We'd want it to be a separate workflow, and probably triggered on a nightly basis, rather than as part of PRs. For PRs it may be interesting if we can only report against any dependencies included in a PR, but if a CVE is already present in main we shouldn't block PRs due to that.

@abstractj @rmartinc FIY

Pankraz76:
> already present in main we shouldn't block PRs due to that.

Yes this should not be the case. Assuming its only failing if there is a fix ready to be used used, or a simple downgrade, or suppression to fix the CVE error.

Yes on every PR is too much. Nightly should be more efficient.


> We're also already using Snyk and Trivy.

Lets put it this way. I was able to integrate this plugin in spring and kafka but not here, assuming its a bug or some CVE not found yet.

.github/workflows/schedule-nightly.yml
```diff
@@ -58,6 +59,8 @@ jobs:
       matrix:
         workflow:
           - snyk-analysis.yml
```
Pankraz76, line 61: should snyk remain exclusive here? or is trivy here any good as well?

.github/workflows/owasp-analysis.yml
```diff
@@ -0,0 +1,26 @@
...
+  analysis:
+    name: OWASP dependency-check
+    runs-on: ubuntu-latest
+    if: github.repository == 'keycloak/keycloak'
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Build Keycloak
+        uses: ./.github/actions/build-keycloak
+      - name: OWASP dependency-check
+        run: ./mvnw -e --no-transfer-progress dependency-check:check
```
Pankraz76, line 26: should fail on corruption.

.github/workflows/schedule-nightly.yml
```diff
@@ -58,6 +59,8 @@ jobs:
...
+          - trivy-analysis.yml
+          - owasp-analysis.yml
```
Pankraz76, line 63: these will prevent a vulnerable release, right?

jonkoops:
We don't want to fail the build when a security issue is present in one of the dependencies, this would hold up other PRs and work, and we already have scanners set up to inform us of any security issues in our dependencies. Like @stianst mentioned, any issues reported by OWASP will likely already have been caught, so this just adds unnecessary overhead.

.github/workflows/schedule-nightly.yml
stianst: Blocking a release is not the process we follow. Let's for argument say there are 2 CVEs in 3rd party dependencies in the latest release, and only 1 is fixable at the time of the release it is more beneficial to release fixing 1 CVE than it is to block the release.

We continuously review open CVEs. Based on severity and whether or not Keycloak is affected we prioritise when fixes are included (and released).

PR #41944, opened 2025-08-18
fix: addressing possible npes

### LINKED ISSUE #40659: Possible Dereference of Null. SAST
### Description

New day, new issue :)

Looks like it is worth adding a Null check, since the function can return a null value.

### Discussion

_No response_

### Motivation

Right here:

https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/jpa/src/main/java/org/keycloak/models/jpa/ClientScopeAdapter.java#L172

cause of 
https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/jpa/src/main/java/org/keycloak/models/jpa/ClientScopeAdapter.java#L143

---
Same situation here:
https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/integration/client-cli/admin-cli/src/main/java/org/keycloak/client/registration/cli/ReflectionUtil.java#L467

cause of 
https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/integration/client-cli/admin-cli/src/main/java/org/keycloak/client/registration/cli/ReflectionUtil.java#L440

---
And:
https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/jpa/src/main/java/org/keycloak/models/jpa/RealmAdapter.java#L1628

cause of possible returned null value:
https://github.com/keycloak/keycloak/blob/f4d5fa68c1159fd03de93a27cbef3d7f46a50e75/model/jpa/src/main/java/org/keycloak/models/jpa/RealmAdapter.java#L1626

### Details

_No response_

shawkins:
close: #40659

Most of these situations are not expected to occur, but NPEs certainly are possible by static analysis. It doesn't seem harmful to add explicit checks.

It would be good to have some of the model utility methods return optional instead, but since they are marked as protected I'm adding a null check.

### DIFF
--- a/integration/client-cli/admin-cli/src/main/java/org/keycloak/client/registration/cli/ReflectionUtil.java
+++ b/integration/client-cli/admin-cli/src/main/java/org/keycloak/client/registration/cli/ReflectionUtil.java
@@ -436,16 +436,13 @@ public class ReflectionUtil {
     }
 
     public static String getTypeString(Type type, Field field) {
-        Class clazz = null;
         if (type == null) {
             if (field == null) {
                 throw new IllegalArgumentException("type == null and field == null");
             }
             type = field.getGenericType();
         }
-        if (type instanceof Class) {
-            clazz = (Class) type;
-        } else if (type instanceof ParameterizedType) {
+        if (type instanceof ParameterizedType) {
             StringBuilder sb = new StringBuilder();
             String rtype = getTypeString(((ParameterizedType) type).getRawType(), null);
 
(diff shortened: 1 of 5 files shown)


integration/client-cli/admin-cli/src/main/java/org/keycloak/client/registration/cli/ReflectionUtil.java
```diff
@@ -461,6 +461,8 @@ public static String getTypeString(Type type, Field field) {
             }
             sb.append(")");
             return sb.toString();
+        } else if (clazz == null) {
```
vmuzikar: Isn't `clazz` always null when we reach this point? Maybe we can just use `else` instead of `else if`?

shawkins: Simplified the if/else checks.

vmuzikar (approved):
LGTM now, thanks.

PR #41952, opened 2025-08-18
fix: removing test os restriction

### LINKED ISSUE #13501: Check ImportAtStartupDistTest#failSetValueToImportRealmOption on windows
We disabled the test on windows for now as the error message is not the expected one. This has to be investigated with the goal of making the test pass also on windows.

_See also: @pedroigor in https://github.com/keycloak/keycloak/pull/13473#discussion_r935881704_

shawkins:
closes: #13501

@Pepo48 there was a logic error in our import-realm option. It needs to specifically take a boolean argument, not string. The converted the test(s) to PicoCli tests to prevent any need for quoting an = on windows.


Pepo48 (approved):
Tested this on my W11 machine, the tests were passing. Thanks @shawkins.

PR #41976, opened 2025-08-19
Collect Playwright reports and server logs on stability runs

### LINKED ISSUE #41974: Collect Playwright reports and server logs on stability runs
In order to determine what is going wrong in flaky test runs, we will have to collect server logs and Playwright traces to be able to debug any issues.

jonkoops:
Closes #41974

### DIFF
--- a/.github/workflows/stability-js-ci.yml
+++ b/.github/workflows/stability-js-ci.yml
@@ -73,6 +73,22 @@ jobs:
         run: pnpm --fail-if-no-match --filter ${{ env.WORKSPACE }} test -- --project=${{ matrix.browser }}
         working-directory: js
 
+      - name: Upload Playwright report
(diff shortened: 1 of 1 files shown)


stianst (commented):
These are fairly big artifacts I'm guessing, so we should have a shorter retention days on them + probably only upload if there are failures?

jonkoops:
> These are fairly big artifacts I'm guessing, so we should have a shorter retention days on them + probably only upload if there are failures?

The reports are about 200KB when nothing fails, and around 2-3MB when they do. I have changed the workflow so that logs and reports are only uploaded if failures occur, and lowered the retention to 7 days, which should be more than enough to download and inspect a report.

ssilvert (approved):

stianst (approved):

PR #41979, opened 2025-08-19
Skip processing HEAD requests for action tokens

### LINKED ISSUE #41834: Clicking email confirmation links in Outlook results in a "stale link" error
### Before reporting an issue

### Area

core

### Describe the bug

When using the UPDATE_EMAIL feature, updating your email address in the account console and then clicking on the verification link in some email clients (at least Outlook) results in an error:
Image

The email address is still successfully updated, but the error message might be confusing for the user.

---

The issue is caused by Outlook "Safe Links" [sending a HEAD request](https://community.wappler.io/t/dealing-with-head-requests/33802), which already verifies the email address. The GET request from the browser then results in a EXECUTE_ACTIONS_ERROR with reason "staleEmailVerificationLink".

The verification link upon initial registration also has this issue, but in that case doesn't result in a user-visible error, because the VERIFY_EMAIL_ERROR with error "email_already_verified" still results in an automatic redirect back to the application, just as a successful VERIFY_EMAIL would.

### Version

26.2.5

### Regression

### Expected behavior

HEAD requests to the `/realms/<realm>/login-actions/action-token` endpoint should not cause any state change on the server.

### Actual behavior

HEAD requests to the `/realms/<realm>/login-actions/action-token` endpoint complete the action and invalidate the token.

### How to Reproduce?

1. Enable the UPDATE_EMAIL feature and the corresponding "Update Email" required action.
2. Open the account console and log in to an existing account. Click the "Update email address" button
3. Enter a valid email address and click "Submit"
4. Open the email in Microsoft Outlook and click the link. (This only works with publicly accessible Keycloak instances on the internet, as the HEAD request originates from Microsoft's servers).  
If you don't have Outlook or your Keycloak instance isn't public, you can instead copy the verification link and manually send a HEAD request using a tool of your choice, e.g. `curl -I <url>`, and then open the link in your browser.
5. You will see "stale link" error message.

### Anything else?

Keycloak event log:
Image

nginx log:
Image

You can see the HEAD request from a Microsoft IP arriving first and getting a 200 response, then the GET request from my browser arriving second and getting a 400 response.

---

The responsible code should be this: https://github.com/keycloak/keycloak/blob/ef312b570c2ed0243242b93688016d7aaf6d91fd/services/src/main/java/org/keycloak/services/resources/LoginActionsService.java#L532-L549

It seems that if no dedicated `@HEAD` handler is defined, the `@GET` handler is called instead (which in this case has side effects).

There is an argument to be made that neither HEAD nor GET requests should actually cause an action to complete, because that violates the assumption of them being [safe](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP) and [idempotent](https://developer.mozilla.org/en-US/docs/Glossary/Idempotent). But to solve that, the verification link would have to always show a form with a button that sends a POST request. To solve this issue, it would be enough for HEAD requests to not execute the action token handler.

pedroigor:
Closes #41834

### DIFF
--- a/services/src/main/java/org/keycloak/headers/DefaultSecurityHeadersProvider.java
+++ b/services/src/main/java/org/keycloak/headers/DefaultSecurityHeadersProvider.java
@@ -30,6 +30,8 @@ import jakarta.ws.rs.core.MultivaluedMap;
 import java.util.Collections;
 import java.util.Map;
 
+import static jakarta.ws.rs.HttpMethod.HEAD;
+import static jakarta.ws.rs.HttpMethod.OPTIONS;
 import static org.keycloak.models.BrowserSecurityHeaders.CONTENT_SECURITY_POLICY;
 
 public class DefaultSecurityHeadersProvider implements SecurityHeadersProvider {
@@ -147,10 +149,17 @@ public class DefaultSecurityHeadersProvider implements SecurityHeadersProvider {
                 status == 400 || status == 401 || status == 403 || status == 404) {
                 return true;
             }
-            if (requestContext.getMethod().equalsIgnoreCase("OPTIONS")) {
-                return true;
+
+            String method = requestContext.getMethod().toUpperCase();
+
+            switch (method) {
+                case OPTIONS:
+                    return true;
+                case HEAD:
+                    return status == 200;
             }
         }
+
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


martin-kanis (approved):
LGTM

abstractj (approved):

PR #22732, opened 2023-08-26
To solve issue 22730

BbIKTOP:
Closes #22730

PR #41980, opened 2025-08-19
align /users/count with /users behavior around service-accounts

### LINKED ISSUE #22730: REST API returns different amount of users
### Before reporting an issue

- [X] I have searched existing issues
- [X] I have reproduced the issue with the [latest nightly release](https://github.com/keycloak/keycloak/releases/tag/nightly)

### Area

account/api

### Describe the bug

I use Keycloak 22.0.1 quay.io image in docker and have strange problem. I have created some users for testing. Also I have created service client account to use with REST API. It created a user for that, with name service-account-clientname

Problem: via REST API, when I use GET /admin/realms/{realm}/users endpoint, Keycloak returns all users except this service account user. But when I use the same endpoint with empty 'username' query parameter, it returns the same list of users plus this service user as well.

Also, /admin/realms/{realm}/users/count returns count value without service accounts.

### Version

22.0.1

### Expected behavior

The same amount of users returned for any kind of full users queries

### Actual behavior

Different amount returned

### How to Reproduce?

Install new keycloak instance. Add 3-4 users. Add service client. Query users count, users list unfiltered and users list filtered by username="". Observe that filtered query returns service user, as others do not.

### Anything else?

_No response_

fakegermano:
closes #22730 

it is simpler than #22732 as it doesn't introduce any changes to APIs or breaking changes to the `/users` endpoint.

it simply aligns `/users/count` with `/users` such that a call for a count returns the same number as an equivalent manual count of the results of the list/search endpoint (ignoring pagination).

in a realm with 5 users and 5 service accounts
e.g. calling `/users/count?search=*` should return 5 (equivalent to running `/users?search=*`

equivalently calling `/users/count?username=&exact=false` should return 10 (equivalent to running `/users?username=&exact=false`

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -192,6 +192,15 @@ If you have an SMTP server configured for your realm, perform the following migr
 . Keep the *Allow UTF-8* option disabled.
 . Verify that no email addresses of users have non-ASCII characters in the local part of the email address. If you detect emails with non-ascii characters in the local part you can use the Verify Profile action to force the user to modify the email after the upgrade.
 
+=== Aligning the count of users with the actual number of users returned from searches
+
+When searching for users in the Admin Console or via the User API, the count of users returned from the
+`/admin/realms/{realm}/users/count` endpoint is now aligned with the actual number of users returned when executing
+searches via `/admin/realms/{realm}/users`.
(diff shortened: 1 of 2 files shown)


pedroigor:
@fakegermano LGTM. We need a note in the upgrade guide to advertise the change in the count method. I'll work something and push to your branch.

pedroigor (approved):
Thanks, @fakegermano.

PR #41986, opened 2025-08-19
fix: using volatile for double checked locking

### LINKED ISSUE #40630: Double check when working with multithreading. SAST
### Description

Another one for you, guys

It looks like the usage of double check will not work reliably in a platform independent way when implemented in Java.

Writes that initialize the object and the write to the helper field can be done or perceived out of order.

Even if the compiler does not reorder those writes, on a multiprocessor the processor or the memory system may reorder those writes, as perceived by a thread running on another processor.

### Discussion

_No response_

### Motivation

_No response_

### Details

I think it's worth reviewing the following double check blocks:

https://github.com/keycloak/keycloak/blob/7fac1531bc17105236d9a6c4f098300d6ccd5199/services/src/main/java/org/keycloak/credential/WebAuthnCredentialProviderFactory.java#L37-L39

2nd:

https://github.com/keycloak/keycloak/blob/7fac1531bc17105236d9a6c4f098300d6ccd5199/services/src/main/java/org/keycloak/services/resources/WelcomeResource.java#L262-L264

3rd:

https://github.com/keycloak/keycloak/blob/7fac1531bc17105236d9a6c4f098300d6ccd5199/services/src/main/java/org/keycloak/theme/freemarker/DefaultFreeMarkerProviderFactory.java#L19-L21

4th:

https://github.com/keycloak/keycloak/blob/7fac1531bc17105236d9a6c4f098300d6ccd5199/services/src/main/java/org/keycloak/credential/WebAuthnPasswordlessCredentialProviderFactory.java#L42-L44

Suggest using volatile variables.

JDK5 and later extends the semantics for volatile so that the system will not allow a write of a volatile to be reordered with respect to any previous read or write, and a read of a volatile cannot be reordered with respect to any following read or write.

Took the information from the source: https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

---

shawkins:
closes: #40630


ahus1 (approved):

ahus1:
Please provide a backport for 26.2 and 26.3 as this is IMHO not an enhancement, but a bug.

shawkins:
> Please provide a backport for 26.2 and 26.3 as this is IMHO not an enhancement, but a bug.

#42039 #42040

PR #41987, opened 2025-08-19
Link specific clients to their settings page in UsedBy view

skowrxn:
- Client names in the UsedBy view now link straight to their settings page
- Tried to implement it in the same style as other components with similar functionality.
- Falls back to plain text if the client ID isn’t available.

Closes #40211

### DIFF
--- a/js/apps/admin-ui/src/authentication/components/UsedBy.tsx
+++ b/js/apps/admin-ui/src/authentication/components/UsedBy.tsx
@@ -1,4 +1,4 @@
-import { KeycloakDataTable } from "@keycloak/keycloak-ui-shared";
+import { KeycloakDataTable, useFetch } from "@keycloak/keycloak-ui-shared";
 import {
   Button,
   Modal,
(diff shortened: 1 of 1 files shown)


skowrxn:
Can I get a review on this PR? @pedroigor @ahus1

skowrxn:
Can I get a review on this PR? @edewit @ssilvert

js/apps/admin-ui/src/authentication/components/UsedBy.tsx
```diff
@@ -34,23 +37,44 @@ type UsedByModalProps = {
...
+              const clients = await adminClient.clients.find({
+                clientId: clientName,
+                realm: realm,
+              });
+              const client = clients[0];
+              return {name: clientName, id: client?.id};
+            } catch {
+              return {name: clientName};
+            }
+          }),
```
edewit, line 75: This will do a page size of queries, wouldn't it be better to change the backend to add this field?

PR #41995, opened 2025-08-20
fix: change hardcoded file separator to platform-based constant in test

### LINKED ISSUE #39660: Failed shouldPreventPathFileSeparatorInVaultSecretId test on Windows
### Before reporting an issue

### Area

testsuite

### Describe the bug

Test case `shouldPreventPathFileSeparatorInVaultSecretId` fails on Windows.

### Version

latest commit from main

### Regression

### Expected behavior

The aforementioned test case succeeds

### Actual behavior

The test case fails, the following assertion evaluates to false: https://github.com/keycloak/keycloak/blob/a035b904ff3f867f8a8ce9e78fed060e61258471/services/src/test/java/org/keycloak/vault/PlainTextVaultProviderTest.java#L224

I think the problem is, that the separator character is `\` on windows and `/` on unix systems. The test case uses the unix-style separator hard-coded, while the validator method uses a platform based constant.

See:
https://github.com/keycloak/keycloak/blob/a035b904ff3f867f8a8ce9e78fed060e61258471/services/src/main/java/org/keycloak/vault/AbstractVaultProvider.java#L102

Image

### How to Reproduce?

Execute `shouldPreventPathFileSeparatorInVaultSecretId` test case on a windows machine.

### Anything else?

_No response_

dan-netizen:
Closes #39660

### DIFF
--- a/services/src/test/java/org/keycloak/vault/PlainTextVaultProviderTest.java
+++ b/services/src/test/java/org/keycloak/vault/PlainTextVaultProviderTest.java
@@ -4,6 +4,7 @@ import org.junit.Test;
 import org.junit.Before;
 import org.junit.After;
 
+import java.io.File;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


Pepo48:
Keycloak correctly detects path traversal even on windows - the detection is executed on two different levels and it's triggered based on a platform and/or setup conditions. The test was adjusted to cover each scenario. 

Thus, I am closing this one in favor of https://github.com/keycloak/keycloak/pull/42014.

@nkelemen18 @dan-netizen thanks for the report and assistance.

PR #41996, opened 2025-08-20
Check for non-ascii local part on emails depending on SMTP configuration

### LINKED ISSUE #41994: Check for non-ascii local part on emails depending on SMTP configuration
### Before reporting an issue

### Area

core

### Describe the bug

This is a follow-up for #41023. With the changes done for that issue now the SMTP configuration rejects sending emails to addresses that have non-ascii in the local part if the SMTP configuration does not allow UTF-8. This is OK because we are preventing sending messages to incorrect addresses but we have decided to add this validation to the built-in UP email validator. This way the server will also reject the registration of emails that cannot be used by the SMTP server.

We have managed two ideas:

1. Just use the SMTP configuration to control the email and allow non-ascii if SMTP is not configured or allows UTF-8.
2. Add an option in the validator. This option is independent to the SMTP configuration in the realm.

For the moment I will use the option 1 for the PR but this can change after reviews.

### Version

999

### Regression

### Expected behavior

The email with non-ascii chars in the local part should be rejected if the SMTP server configuration does not allow sending them.

### Actual behavior

The email is accepted.

### How to Reproduce?

See description.

### Anything else?

_No response_

rmartinc:
Closes #41994

Adding the checking for non-ascii in the email local part if the realm is configured to send emails to an SMTP without `allowutf8`. Finally it was talked to use the direct option with no option in the validator. 

@ahus1 @pedroigor @mposolda @stianst For your information.

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/validate/validators/EmailValidator.java
+++ b/server-spi-private/src/main/java/org/keycloak/validate/validators/EmailValidator.java
@@ -20,7 +20,9 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.keycloak.email.EmailSenderProvider;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.RealmModel;
 import org.keycloak.provider.ConfiguredProvider;
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.provider.ProviderConfigurationBuilder;
@@ -43,6 +45,8 @@ public class EmailValidator extends AbstractStringValidator implements Configure
 
     public static final String MESSAGE_INVALID_EMAIL = "error-invalid-email";
 
+    public static final String MESSAGE_NON_ASCII_LOCAL_PART_EMAIL = "error-non-ascii-local-part-email";
+
     public static final String MAX_LOCAL_PART_LENGTH_PROPERTY = "max-local-length";
 
     @Override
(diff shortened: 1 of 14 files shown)


pedroigor (dismissed):

server-spi-private/src/main/java/org/keycloak/validate/validators/EmailValidator.java
```diff
@@ -61,6 +64,29 @@ protected void doValidate(String value, String inputHint, ValidationContext cont
...
+        }
+
+        final KeycloakSession session = context.getSession();
+        if (session == null) {
+            return;
+        }
+
+        final RealmModel realm = session.getContext().getRealm();
+        if (realm == null || realm.getSmtpConfig() == null || realm.getSmtpConfig().isEmpty()
+                || "true".equals(realm.getSmtpConfig().get("allowutf8"))) {
```
ahus1: Please introduce a constant for `allowutf8` in all places.

themes/src/main/resources/theme/base/account/messages/messages_en.properties
```diff
@@ -386,3 +386,4 @@ error-invalid-date=Invalid date.
 error-user-attribute-read-only=The field {0} is read only.
 error-username-invalid-character=Username contains invalid character.
 error-person-name-invalid-character=Name contains invalid character.
+error-non-ascii-local-part-email=Local part of the address ''{1}'' must contain only ASCII characters.
```
ahus1: In the past there have been attacks to reflect user input in the error messages, which lead to XSS problems. While I'm sure we're encoding the output correctly, I'd rather remove repeating the user input in the error message. This would then apply to the message and also where we add the message parameter.

```suggestion
error-non-ascii-local-part-email=Local part of the address must contain only ASCII characters.
```

ahus1 (changes requested):
Thank you for the pull request. See below for some suggested changes.

server-spi-private/src/main/java/org/keycloak/validate/validators/EmailValidator.java
rmartinc: I created the constant in the `EmailSenderProvider` class.

themes/src/main/resources/theme/base/account/messages/messages_en.properties
rmartinc: Done!

ahus1 (approved):

PR #42002, opened 2025-08-20
Add Prefix option to OIDC User Attribute Mapper

pszabo891:
This PR introduces the `Attribute Value Prefix` into the OIDC user Attribute Mapper.

attribute-mapper

This PR supersedes PR #39930 and provides an alternative implementation for #39929.


pszabo891:
Hi @pedroigor, Could we give this one another go? I have rebased it with the latest main branch now. Thanks!

Issue #41590, opened 2025-07-31
Keycloak: Update batch_credential_issuance to enforce batch_size ≥ 2.


Ogenbertrand:
### Description

See: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-metadata

forkimenjeckayang:
Working on this

PR #42003, opened 2025-08-20
[OID4VCI]: Enforce batch_size ≥ 2 validation for batch_credential_issuance

forkimenjeckayang:
## Summary
This PR implements validation for the `batch_credential_issuance.batch_size` parameter to ensure compliance with the OID4VCI specification, which requires that `batch_size` MUST be 2 or greater.

### Implementation
- Added validation in `getBatchCredentialIssuance()` method to check that `batch_size` is at least 2

### Testing
- Added comprehensive test method `testBatchCredentialIssuanceValidation()` covering:
    - Valid cases: batch sizes 2, 5 (should be accepted)
    - Invalid cases: batch sizes -1, 0, 1 (should be rejected)
    - Edge case: batch size exactly 2 (minimum valid value)
  - Created reusable helper method `testBatchSizeValidation()` to reduce code duplication
  - Ensured tests don't interfere with each other by properly restoring original realm attributes

## Addresses
- Issue #41590

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/constants/Oid4VciConstants.java
+++ b/server-spi-private/src/main/java/org/keycloak/constants/Oid4VciConstants.java
@@ -29,6 +29,8 @@ public final class Oid4VciConstants {
 
     public static final String CREDENTIAL_SUBJECT = "credentialSubject";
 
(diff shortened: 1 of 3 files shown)


IngridPuppet (approved):
I could check this PR internally prior to it reaching upstream. LGTM.

mposolda (approved):
@forkimenjeckayang @IngridPuppet Thanks for the PR and review!

PR #42009, opened 2025-08-20
fix: preventing possible NPEs

### LINKED ISSUE #39960: Multiple potential null dereference issues
### Description

Hi there!
I have been conducting static analysis and have found several places in the code supposedly vulnerable to null dereference.

---
1. Null value is passed as a 3rd parameter in method `AuthorizeClientUtil.authorizeClient`: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/protocol/oidc/grants/ciba/endpoints/AbstractCibaEndpoint.java#L51
and dereferenced here: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/protocol/oidc/utils/AuthorizeClientUtil.java#L60

2. Null value is passed as a 3rd parameter in method `AuthorizeClientUtil.authorizeClient(session, event, null).getClient()`: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/services/resources/ClientsManagementService.java#L167
and dereferenced here: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/protocol/oidc/utils/AuthorizeClientUtil.java#L60

I do not exclude that the ClientModel will always be initialized, and the "conditional" error handler will not work.
However, when passing the cors variable with a null value and the impossibility of initializing the ClientModel for some reason, an exception may occur in the case of calling allowAllOrigins()

---
3. Null value is passed as a 2nd parameter in method `createFlowExecution(this.flowId, null)`: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java#L1094
and dereferenced here: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java#L922 
by calling the class constructor right here: 
https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/authentication/FormAuthenticationFlow.java#L63

Same as

4. Null value is passed as a 2nd parameter in method `createFlowExecution(this.flowId, null)`: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java#L940
and dereferenced here: https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java#L922
by calling the class constructor right here: 
https://github.com/keycloak/keycloak/blob/dc7f4ec34f7fa423bff151c5e8b87b4fc0cedce0/services/src/main/java/org/keycloak/authentication/FormAuthenticationFlow.java#L63

Most likely, at the points of calling authenticateClient(), FORM_FLOW is not implied and everything will work properly), but providerId is not explicitly declared anywhere (references to the call of the main function), therefore I conclude that Null dereferencing may occur.

---

shawkins:
closes: #39960

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/FormAuthenticationFlow.java
+++ b/services/src/main/java/org/keycloak/authentication/FormAuthenticationFlow.java
@@ -59,6 +59,9 @@ public class FormAuthenticationFlow implements AuthenticationFlow {
 
     public FormAuthenticationFlow(AuthenticationProcessor processor, AuthenticationExecutionModel execution) {
         this.processor = processor;
+        if (execution == null) {
+            throw new AuthenticationFlowException("Execution not provided to flow", AuthenticationFlowError.INTERNAL_ERROR);
+        }
         this.formExecution = execution;
         formActionExecutions = processor.getRealm().getAuthenticationExecutionsStream(execution.getFlowId()).collect(Collectors.toList());
         formAuthenticator = processor.getSession().getProvider(FormAuthenticator.class, execution.getAuthenticator());
--- a/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java
+++ b/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java
@@ -924,18 +924,11 @@ public class AuthenticationProcessor {
             throw new AuthenticationFlowException(AuthenticationFlowError.INTERNAL_ERROR);
         }
         if (flow.getProviderId() == null || flow.getProviderId().equals(AuthenticationFlow.BASIC_FLOW)) {
-            DefaultAuthenticationFlow flowExecution = new DefaultAuthenticationFlow(this, flow);
-            return flowExecution;
-
+            return new DefaultAuthenticationFlow(this, flow);
         } else if (flow.getProviderId().equals(AuthenticationFlow.FORM_FLOW)) {
-            if (execution == null) {
-                throw new AuthenticationFlowException("Execution not provided to flow", AuthenticationFlowError.INTERNAL_ERROR);
-            }
-            FormAuthenticationFlow flowExecution = new FormAuthenticationFlow(this, execution);
-            return flowExecution;
+            return new FormAuthenticationFlow(this, execution);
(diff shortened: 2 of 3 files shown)


services/src/main/java/org/keycloak/authentication/FormAuthenticationFlow.java
```diff
@@ -59,6 +59,9 @@ public class FormAuthenticationFlow implements AuthenticationFlow {
 
     public FormAuthenticationFlow(AuthenticationProcessor processor, AuthenticationExecutionModel execution) {
         this.processor = processor;
+        if (execution == null) {
+            throw new AuthenticationFlowException("Execution not provided to flow", AuthenticationFlowError.INTERNAL_ERROR);
+        }
```
ahus1, line 64: I've chosen to move the check here as this class knows that it must not be null. Would then also catch cases when it would be called from other places.

ahus1 (commented):
Thank you for the PR. See below for one more refactoring that I did. Please comment if you agree with this change. Thanks!

services/src/main/java/org/keycloak/authentication/FormAuthenticationFlow.java
shawkins: There are no other usages, but since it's a public/method class in a public module this is certainly good also.

ahus1 (approved):

PR #42010, opened 2025-08-20
Rework RLM core to schedule action based on events   @sguilhen

### LINKED ISSUE #41803: [RLM] Improvements to scheduling and processing resource policies
### Introduction

The realm resource lifecycle management initial implementation (ref commit https://github.com/keycloak/keycloak/commit/a8225655cfc1d4d01d6cbeea70cf45e4958e36e8) was designed around the notion of lifecycle policies and actions. A policy defines the selection criteria for resources and the actions that should be run against the selected resources at specific times. For example, one could create a policy based on the user's last session refresh time, with some actions (e.g. one action to notify the user about their removal after 10 days of the last session refresh time, and another to remove the user 15 days after the same time). So the selection criteria is the user's last session refresh time compared to the time that each action should run.

### Policy execution

One of the central components of the resource lifecycle manegement is a scheduled task that runs regularly (e.g. daily) to execute the actions on eligible resources. The current implementation for this task iterates through the policies by priority order, asking each policy to select the resources that are eligible to run each of the defined actions, and then run the actions. A state table is used to keep track of the last action that was executed against a particular resource, so that subsequent executions of the scheduled task won't run the same action again.

While initial tests show this approach works, it is very intensive query-wise, as every action of every policy results in queries to find eligible resources in real time. Also, it doesn't seem suitable to solve a few use cases. Consider a policy that selects users with no user sessions, and then wants to setup an action to run against these users (lets say, delete the users). If the action is supposed to run right away, it will work because the policy will select the users with no user sessions and execute the first action. However, if the action is configured to run 3 days after the user has no more sessions, the policy won't work because it doesn't have a date to compare the schedule action against. In other words, it can only find the users by number of sessions criteria, but it can't decide if it is already time to run the action or not.

#### Improvement suggestion

A suggestion to improve the whole process is to have a scheduled_actions table, storing the resource_id, policy_id, action_id,  scheduled_time, and status. A policy would no longer try to find the eligible resources when the scheduler runs. Instead, we can attach the policy to users as they are created (for new users) or using some policy selection algorithm against existing users, and the policy will try to schedule its first action right away, or when some event happens.

Example 1: policy based on user creation date for users coming from broker A - when user is created through broker A authentication, the policy is assigned to the user and the first action can already be scheduled (lets say delete the user 60 days after creation) as we know the creation date. An entry for this is added to the scheduled_action table (userid, policyid, actionid, timestamp of system current time + 60 days, status SCHEDULED). When the periodic lifecycle task runs, it looks for SCHEDULED actions whose time has passed and simply runs them. In this case, there's no other action to be scheduled as the user was deleted.

Example 2: policy based on last session refresh time for users coming from broker B - when user is created through broker B authentication, the policy is assigned and the first action can be scheduled because we will have the last session refresh date available. Same as above, the first action is scheduled (say, notify users they will be disabled if they don't log back in soon) to run in 10 days. An entry is added to the table, and once the periodic task runs, the action is executed, e-mail is sent and the entry's status is set to EXECUTED. The periodic task then notifies the policy about the action, giving the policy an opportunity to setup the next action in line (say disable the user in 5 days from the current time - or 15 days from the last session refresh time). So a new entry is added for this action in the scheduled_action table.

The periodic lifecycle task looks primarily for the actions in the scheduled actions table, which were previously setup by the policies associated with the users, and once the actions are run it calls back into the policy to allow the policy to setup the next action. This allows the periodic task to operate more efficiently by avoiding a query in real time to find eligible resources for every action in every policy.

How does this help with the "users with no active user sessions" example above? We can now schedule the actions when they are supposed to run by adding an entry for the first action to the scheduled_actions table. However, this policy can't schedule this first action proactively - i.e. there's no event we can use to detect the user has no more sessions, and when assigning the policy to the user at creation time we also can't determine when we should schedule the first action. So it needs an opportunity to query eligible users regularly and because of that we need to make the periodic lifecycle task a bit more flexible and allow policies to find resources in real time too.

So the periodic lifecycle task would run three steps:

- Step 1:  call a `scheduleActions()` method on every policy to give policies an opportunity to identify resources that can be have actions scheduled. Default impl would be empty for policies based on user creation date or last session refresh time, but in our "user with no active user sessions" example the policy would find these users and upon finding them it would schedule the first action based on the configured time.
- Step 2 : select all SCHEDULED actions from the respective table whose time is up, and run these actions.
- Step 3 : call the policies that had actions executed, passing the actions list so that the policy can schedule the next actions for every user if necessary.

This approach combines the previous approach with a more dynamic scheduling approach that has several advantages:

1. It is a lot less intensive in terms of DB queries as many policies will be able to setup the initial action upon user creation or when receiving an event like LOGIN
2. It still allows for the creation of policies that would detect certain changes when there's no corresponding event, like the "users with no active user sessions" example above.
3. The scheduled_actions table will record every action that was ever scheduled for a resource, increasing auditing capabilities.

### Resource selection and granularity of policies

As mentioned before, the policies are responsible for selecting resources based on some criteria. Existing policies include the  `UserCreatedDateResourcePolicyProvider` and the `UserLastSessionRefreshTimeResourcePolicyProvider`, that selects users based on the user's creation date or last session refresh time respectively.

These basically identify the attribute of the user entity used for selection. But another, perhaps even more important characteristic, is the user's origin. We expect admins to define policies for users based on their origin, whether they come from specific idps, federation providers, or even organizations. So an admin would define a policy for users created through broker A, another policy for users created through broker B, another one for users from an Active Directory provider, and so on.

As of now, the origin criteria has been implemented by adding new policy providers that extend from the ones mentioned above. So you would see something like `FederatedIdentityPolicyProvider extends UserLastSessionRefreshTimeResourcePolicyProvider`, which means we now have a policy provider for policies based on last session refresh AND who originate from a specific broker.

As the number of base policies grow, this approach will lead to an explosion of policy providers for each combination of user attribute + user origin.

#### Improvement suggestion

Given the importance for admins to define policies affecting users from specific sources, or origins, the suggestion is to include the support for the origin selection at the top of the policy hierarchy, giving all policy implementations support for adding predicates based on the origin. So every policy would allow for the specification of one or more sources that the policy should apply to.

So instead of a new provider, the base policies like user last session refresh would allow for specification of the sources as a config attribute.

### Ability to define extra filters for actions

In the current implementation the policy is the only party defining selection criteria for resources. But there are times when the action could end up requiring more filters. Consider, for example, the notify user action, that sends an e-mail to the users to notify them of an impending account deletion. If the user has no associated e-mail, the action would need to detect that and skip the execution. This is an acceptable approach but it ends up scheduling actions for users who are not really suitable for receiving e-mails and detecting that at action execution time.

#### Improvement suggestion

The suggestion is to introduce a mechanism to allow the action to influence the scheduling process so that it is not scheduled if some extra criteria is not met (e.g. user must have an email set).

sguilhen:
Closes #41803


services/src/main/java/org/keycloak/models/policy/UserPolicyEventListener.java
```diff
@@ -0,0 +1,54 @@
...
+import org.keycloak.events.EventListenerProvider;
+import org.keycloak.events.admin.AdminEvent;
+import org.keycloak.models.KeycloakSession;
+
+import java.util.Objects;
+
+/**
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class UserPolicyEventListener implements EventListenerProvider {
```
pedroigor: As discussed, rename to `ResourcePolicyEventListener` and remove copyright + author docs.

services/src/main/java/org/keycloak/models/policy/UserPolicyEventListenerFactory.java
```diff
@@ -17,54 +17,41 @@
...
+import org.keycloak.events.EventListenerProvider;
+import org.keycloak.events.EventListenerProviderFactory;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.KeycloakSessionFactory;
-import org.keycloak.provider.ProviderConfigProperty;
-
-public class FederatedIdentityPolicyProviderFactory implements ResourcePolicyProviderFactory<FederatedIdentityPolicyProvider> {
 
-    public static final String ID = "federated-identity-policy";
+public class UserPolicyEventListenerFactory implements EventListenerProviderFactory {
```
pedroigor: As discussed, rename to `ResourcePolicyEventListenerFactory`.

services/src/main/java/org/keycloak/models/policy/UserPolicyEventListenerFactory.java
```diff
@@ -17,54 +17,41 @@
...
-    @Override
-    public String getHelpText() {
-        return "";
     }
 
     @Override
-    public List<ProviderConfigProperty> getConfigProperties() {
-        return List.of();
+    public String getId() {
+        return "user-policy-event-listener";
```
pedroigor: As discussed, rename to `resource-policy-event-listener`.

server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyProvider.java
```diff
@@ -42,4 +37,86 @@ public interface ResourcePolicyProvider extends Provider {
...
     List<String> filterEligibleResources(List<String> candidateResourceIds, long time);
+
+    /**
+     * Indicates whether the specified resource is in the scope of this policy. For example, a policy associated with a
+     * broker is applicable only to users with a federated identity associated with the same broker.
+     *
+     * @param resourceId the id of the resource being checked.
+     * @return {@code true} if the resource is in the policy scope; {@code false} otherwise.
+     */
+    boolean isResourceInScope(String resourceId);
```
pedroigor: As discussed, perhaps we can simplify the contract and rely on the `supportsSchedulingOnEvent` to also check whether a resource is in the scope (origin/source) of a policy.

The ` ResourcePolicyEvent` already provides the resource id.

pedroigor (approved):

PR #42014, opened 2025-08-20
Fix: Stabilize PlainTextVaultProviderTest by enhancing validation logging

### LINKED ISSUE #39660: Failed shouldPreventPathFileSeparatorInVaultSecretId test on Windows

Pepo48:
Closes #39660

### DIFF
--- a/services/src/test/java/org/keycloak/vault/PlainTextVaultProviderTest.java
+++ b/services/src/test/java/org/keycloak/vault/PlainTextVaultProviderTest.java
@@ -221,9 +221,12 @@ public class PlainTextVaultProviderTest {
         // then
         assertNotNull(secret);
         assertFalse(secret.get().isPresent());
+        // The validation may be performed by AbstractVaultProvider or FilesPlainTextVaultProvider
         assertTrue(
-                logMessages.stream()
-                        .anyMatch(msg -> msg.contains("Key .../key1 contains invalid file separator character"))
+            logMessages.stream().anyMatch(msg ->
+                msg.contains("Key .../key1 contains invalid file separator character")
+                    || msg.contains("Path traversal attempt detected in secret .../key1")
+            )
         );
     }
 


mabartos:
@Pepo48 What's the status of this PR?

Pepo48:
@mabartos waiting to be reviwed and merged. I'll then create the backports.

@vmuzikar knows about this, we also discussed the way the backports are going to be created.

vmuzikar (changes requested):
We would need to look into the tests – why they didn't fail before without this fix, or if there's no test covering this.

Pepo48:
> We would need to look into the tests – why they didn't fail before without this fix, or if there's no test covering this.

@vmuzikar the test in question `PlainTextVaultProviderTest` is part of the base unit tests, which we only execute on Ubuntu runners.

vmuzikar:
> @vmuzikar the test in question `PlainTextVaultProviderTest` is part of the base unit tests, which we only execute on Ubuntu runners.

Would be probably good to run on Windows as well. Can you try it?

Pepo48:
> Would be probably good to run on Windows as well. Can you try it?

Of course, I would do it this way:
https://github.com/keycloak/keycloak/compare/main...Pepo48:keycloak:win-base-ut

Which adds `Base UT (windows-latest) stage` to our CI:
https://github.com/Pepo48/keycloak/actions/runs/17885496692/job/50858783770

And as you can see `PlainTextVaultProviderTest` failed this time even in the GH CI.

I will cherrypick the commit and add it to this PR in order to also see the results with the fix applied.

Pepo48:
The vault tests passed. The UT stage failure is caused by two other unrelated tests. I can analyze and fix them in a separate PRs.

Pepo48:
I split it into separate issues: https://github.com/keycloak/keycloak/issues/42796.

Corresponding PRs are also in place.

services/src/main/java/org/keycloak/vault/AbstractVaultProvider.java
```diff
@@ -99,7 +99,7 @@ private void checkForLegacyKey(VaultKeyResolver resolver, String vaultSecretId,
      * @return a boolean indicating whether the validation passed.
      */
     protected boolean validate(VaultKeyResolver resolver, String key, String resolvedKey) {
-        if (key.contains(File.separator)) {
+        if (key.indexOf('/') >= 0 || key.indexOf(File.separatorChar) >= 0) {
```
vmuzikar: This way on Unix-like systems, `\` will be accepted (though it won't matter as it's not a path separator there). But if someone e.g. moves their KC instance to Windows, then it starts failing which might be confusing.

I'd suggest making it consistent and check for both `/` and `\`, on top of `File.separator` just in case...

vmuzikar (changes requested):

services/src/main/java/org/keycloak/vault/AbstractVaultProvider.java
shawkins: Checking for \ on linux can be seen as a notable or breaking change.

vmuzikar: @shawkins Good point. This is still a CVE fix which would justify a reasonable breaking change. I would prefer it over platform inconsistency as that might be very confusing to users.

Pepo48: I addressed the suggestion and added a release note as well.

vmuzikar (approved):
LGTM, thanks.

vmuzikar (changes requested):
As agreed offline, we will rather adjust the test in the end.

vmuzikar (approved):

shawkins (approved):
LGTM, thanks @Pepo48

PR #42017, opened 2025-08-20
More flexible handling of params, headers and entities for SimpleHTTP (#42016)

### LINKED ISSUE #42016: More flexible handling  of params, headers and entities for SimpleHTTP
### Description

Add the following methods for convencience to SimpleHTTP:
- getHeaders() to expose a umodifiable map of the currently configured headers (or null if no headers are configured)
- getParam() to enable access to a specific param (or null of no params are configured)
- getParams() to expose a umodifiable map of the currently configured params (or null if no params are configured)
- getEntity() to access the currently configured entity
- params(Map<String,String>) method to configure the parameters explicitly

### Discussion

_No response_

### Motivation

SimpleHTTP instances created by Keycloak are difficult to inspect by extension code: 
- Configured params cannot be accessed later, since there is no getParam(..) or getParams() method. 
- Headers can be analysed but only if they are kown, since the is no getHeaders() method.
- Configured JSON / object entites cannot be read anymore, since there is no getEntity() method

### Details

_No response_

thomasdarimont:
Fixes #42016

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/broker/provider/util/SimpleHttp.java
+++ b/server-spi-private/src/main/java/org/keycloak/broker/provider/util/SimpleHttp.java
@@ -56,6 +56,7 @@ import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


pedroigor (approved):
LGTM. But why would you want to have access to, for instance, the entity once you have a response?

Also, the entity will be in its raw form and not reflecting the transformation applied when preparing/sending the request.

thomasdarimont:
I added the `getEntity()` accessor to be able to inspect the (currently) configured entity to decide if I can continue using it or if it needs to be replaced by explicitly calling `simpleHttp.entity(newEntity)`.
The fact that the entity is not marshalled yet when accessing it through `getEntity()` is IMHO a good thing since it allows to inspect (and potentially change) the value before marshalling it.

This is useful for instance when Keycloak code or an extension prepares a SimpleHTTP request and allows user code to potentially modify the SimpleHTTP before the actual request is sent. One example for this are the built-in identity providers
as the following keycloak-user mailing list post demonstrates: https://groups.google.com/g/keycloak-user/c/nQoA_vjZBfE/m/ED7rWBoGCwAJ?utm_medium=email&utm_source=footer

pedroigor:
Thanks, @thomasdarimont.

Issue #31469, opened 2024-07-22
Show account page before login


ehsanr1366:
### Before reporting an issue

- [X] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

account/ui

### Describe the bug

I intend to describe the bug very briefly and precisely. I am using version 25.0.2. When I try to access the account page, the account page, including menus and header, is displayed for a few moments and then the login page is shown.

This is not very pleasant! Thank you.

### Version

25.0.2

### Regression

- [ ] The issue is a regression

### Expected behavior

Well, it's natural that the login page should be shown to the user immediately and quickly.

### Actual behavior

The point is that the account page is shown first, then it checks whether the user is logged in or not. After realizing that no one is logged in, the user is redirected to the login page.

### How to Reproduce?

You can easily reproduce the bug. Just try to access the account page. That's it.

ehsanr1366:
Thank you for your response. I also observed the issue I reported on version 25.0.2, which is currently the latest version.

agagancarczyk:
Thanks @ehsanr1366 for reporting this issue. I can reproduce it also in `main`.

edewit:
The way it works is that on the first fetch it notices that the user is not logged IMO there isn't a way to find out before that the user isn't logged in. So no data is shown, but we can't get rid of having this initial render

jonkoops:
We should really be implementing a mechanism in the theme bootstrapping to ensure we know if the user has a session, but this would require some more extensive refactoring.

edewit:
we allowed to be not logged in for the account console before, can't we "just" make it so that that is no longer allowed? I don't know exactly how to accomplish that but maybe something like:

```java
if (auth == null)
   throw new ForbiddenException("Bearer")
```
but that doesn't redirect to the login page though

mahdisn76:
I also have this problem and caused lots of problems for me and my customers.

edewit:
@mahdisn76 lots of problems? I mean it's not nice, but it's not a big issue, right?

mahdisn76:
> @mahdisn76 lots of problems? I mean it's not nice, but it's not a big issue, right?

Yeah! Since the account page shown before the login page, the user feels insecure and think it as a security bug.

msmosavar:
Same problem... Any updates?

ehsanr1366:
[...]

PR #42035, opened 2025-08-21
Use back keycloak-js instead of initiate login in the backend for account

### LINKED ISSUE #40463: Login to Account Console produces two consecutive LOGIN events
### Before reporting an issue

### Area

account/ui

### Describe the bug

When an unauthenticated user logins in to the account console Keycloak will generate two separate consecutive LOGIN events.

### Version

26.2.5

### Regression

### Expected behavior

Keycloak should only generate one LOGIN event when an unauthenticated user performs a login for the account-console client.

### Actual behavior

Keycloak generates two LOGIN events when an unauthenticated user performs a login for the account-console client.

### How to Reproduce?

0) Enable event storage
1) Login to account-console
2) Observe two LOGIN events created for the account-console.

### Anything else?

Keycloak Versions before v26.1.0 only one LOGIN event was produces for the account-console.

This was caused by my change in this PR: https://github.com/keycloak/keycloak/pull/32686
This only affects the account-console, other clients produce only one LOGIN event.

The Problem is that if we detect an unauthenticated user accessing the /account endpoint, we redirect the user to a generated authorize URL to spare the bootstrappung of the "heavy" react JS app which would then redirect the user.
The user now sees the login page and can authenticate. After successful authentication the user is sent back to the account console. This produces the first LOGIN event.
This time, we can detect the existing user session and this start to bootstrap the react JS app. The react JS app then performs the "real" auth code flow in the context of the already authenticated user session, which is then immediately redirected back to the account-console. This produces the second LOGIN event.

The following shows an example event sequence for different OIDC client:
Image 1:
Image
.. and here is an event sequence for the account-console:
Image 2:
Image

This leads to double counting of LOGIN events if users login to the account-console client.

I see the following options:
1) somehow suppress the second "Event 2" shown in Image 2 for the account-console client
2) handle the code-flow / token-management for the account-console on the account backend
3) add additional event_details metadata about the LOGIN TYPE (e.g. interactive, session)

rmartinc:
Closes #40463

After checking this I think the only way of doing this OK without changing things a lot is going back and use `keycloak-js` normally again. If I understood this correctly, the initial issue #31469 was reported because the console was displayed empty, and the login was requested later when the login was required for the first request (401). This PR does the following:

* The backend does not initiate the login. This part is removed because it's a mess. This is the first login and then the account console uses `keycloak-js` to login again (automatically now because cookies are in place), but it was a second round-trip in events. The JS part cannot use the state or code verifier generated by the backend. It makes no sense.
* The applications now in `ui-shared` always initiates the login if not already authenticated. This way nothing is displayed (just the `spinner`) and if the redirect is needed, from the spinner you go to the login page. Nothing is shown except the spinner if you are not logged in. The shared part is used for both consoles (admin and account).
* The problem is that there are a lot tests that thinks that there will be an automatic redirect for the account console. As `keycloak-js` is used we need a real browse (firefox or chrome) to execute the JS needed to login. I have changed those tests to use other thing instead of the account console.
* The `AccountConsoleTest` is the only test that really checks the account functionality This one is moved to the `forms` package to be used with firefox and chrome. The scopes are correctly respected. The only little difference is that the `keycloak-js` always adds `openid`, but I think that this is OK.
* I have also changed the error representation. The current setup shows anything in the `error_description` returned. The only error accepted now is a JS `Error` or the `error` generated bye `keycloak-js`. As the `error` is standard I have provided the translations for it, and only the standard errors are displayed. Other unknown error uses always the same message.

It's a draft for now. Because maybe you see another approach (but I really don't see anything better than this).


js/libs/ui-shared/src/context/KeycloakContext.tsx
```diff
@@ -74,20 +74,23 @@ export const KeycloakProvider = <T extends BaseEnvironment>({
       });
 
     init()
```
jonkoops: Instead of ensuring the user is authenticated by doing so manually, it is possible to set `onLoad` to `"login-required"`. This will ensure that if the user is not authenticated, they will be redirected to the login page.

rmartinc: Done!

jonkoops (approved):

jonkoops:
@rmartinc I am moving this out of draft in your absence, I agree with the approach here. I'll have some other reviewers check this out as well.

pedroigor (approved):
I agree about the mess. The discussions at that time were basically changing UIs vs backend, whereas the UI was, in theory, too hard to change in order to avoid rendering the React app when users are not yet authenticated.

If we are still handling errors properly and the spinner is shown instead of initing the React app, LGTM.

PR #42039, opened 2025-08-21
fix: using volatile for double checked locking

shawkins:
closes: #40630


ahus1 (approved):
Approving. Unfortunately CI is unstable.

PR #42054, opened 2025-08-21
DPoP verification support for admin/account REST API endpoints. Java …

### LINKED ISSUE #33942: Make sure Keycloak endpoints have DPoP validation

mposolda:
…admin-client DPoP support

closes #33942

- PR adds support for the DPoP verification for the admin REST API and account REST API endpoints. The only real change on the server-side is in fact introducing DPoP validation support in the `AppAuthManager.authenticate()` , which is done by re-using same validator already used for user-info endpoint . The `AppAuthManager` is used by both admin REST API and account REST API, so this adds support to both of them (it adds support to few other server endpoints, which supported authentication by `Authorization: Bearer` header and which are using `AppAuthManager` like for example https://www.keycloak.org/docs/latest/server_admin/index.html#retrieving-external-idp-tokens endpoint)

- Support for DPoP in the java admin-client . This is implemented by adding `DPoPAuthFilter` in the java admin-client stack instead of `BearerAuthFilter` . Some minor refactoring done to re-use some snippets already used in the testsuite (Especially introducing `DPoPGenerator`, which is now used by both admin-client and testsuite. This contains just some methods, which were moved to it from `ClientPoliciesUtil` ).

### DIFF
--- a/core/src/main/java/org/keycloak/util/DPoPGenerator.java
+++ b/core/src/main/java/org/keycloak/util/DPoPGenerator.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.util;
+
+import java.security.Key;
+import java.security.KeyPair;
+import java.security.PrivateKey;
+import java.security.interfaces.RSAPublicKey;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.common.util.Base64Url;
+import org.keycloak.common.util.SecretGenerator;
+import org.keycloak.common.util.Time;
+import org.keycloak.crypto.AsymmetricSignatureSignerContext;
+import org.keycloak.crypto.ECDSASignatureSignerContext;
+import org.keycloak.crypto.KeyType;
+import org.keycloak.crypto.KeyUse;
+import org.keycloak.crypto.KeyWrapper;
+import org.keycloak.crypto.SignatureException;
+import org.keycloak.crypto.SignatureSignerContext;
+import org.keycloak.jose.jwk.JWK;
+import org.keycloak.jose.jwk.RSAPublicJWK;
+import org.keycloak.jose.jws.JWSBuilder;
+import org.keycloak.jose.jws.JWSHeader;
+import org.keycloak.jose.jws.crypto.HashUtils;
+import org.keycloak.representations.dpop.DPoP;
+
+import static org.keycloak.OAuth2Constants.DPOP_DEFAULT_ALGORITHM;
+import static org.keycloak.OAuth2Constants.DPOP_JWT_HEADER_TYPE;
+import static org.keycloak.jose.jwk.JWKUtil.toIntegerBytes;
+
+/**
+ * Utility for generating signed DPoP proofs
+ *
+ * @author <a href="mailto:mposolda@redhat.com">Marek Posolda</a>
+ * @see <a href="https://datatracker.ietf.org/doc/html/rfc9449">OAuth 2.0 Demonstrating Proof of Possession (DPoP) specification</a>
+ *
+ */
+public class DPoPGenerator {
+
+    // TODO: Similar for EC and EdDSA
+    public static String generateRsaSignedDPoPProof(KeyPair rsaKeyPair, String httpMethod, String endpointURL, String accessToken) {
+        JWK jwkRsa = createRsaJwk(rsaKeyPair.getPublic());
+        JWSHeader jwsRsaHeader = new JWSHeader(DPOP_DEFAULT_ALGORITHM, DPOP_JWT_HEADER_TYPE, jwkRsa.getKeyId(), jwkRsa);
+        return generateSignedDPoPProof(SecretGenerator.getInstance().generateSecureID(), httpMethod, endpointURL, (long) Time.currentTime(),
+                jwsRsaHeader, rsaKeyPair.getPrivate(), accessToken);
+    }
+
+
+    public static JWK createRsaJwk(Key publicKey) {
+        RSAPublicKey rsaKey = (RSAPublicKey) publicKey;
+
+        RSAPublicJWK k = new RSAPublicJWK();
+        k.setKeyType(KeyType.RSA);
+        k.setModulus(Base64Url.encode(toIntegerBytes(rsaKey.getModulus())));
+        k.setPublicExponent(Base64Url.encode(toIntegerBytes(rsaKey.getPublicExponent())));
+
+        return k;
+    }
+
+
+    public static String generateSignedDPoPProof(String jti, String htm, String htu, Long iat, JWSHeader jwsHeader, PrivateKey privateKey, String accessToken) {
+        try {
+            DPoP dpop = new DPoP();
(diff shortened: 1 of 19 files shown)


thomasdarimont:
LGTM 👍

core/src/main/java/org/keycloak/util/DPoPGenerator.java
```diff
@@ -0,0 +1,102 @@
+package org.keycloak.util;
+
+import java.io.IOException;
```
IngridPuppet: Missing copyright.

core/src/main/java/org/keycloak/util/DPoPGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.keycloak.OAuth2Constants.DPOP_JWT_HEADER_TYPE;
+import static org.keycloak.jose.jwk.JWKUtil.toIntegerBytes;
+
+/**
+ * Utility for generating signed DPoP proofs
+ *
+ * @see <a href="https://datatracker.ietf.org/doc/html/rfc9449">OAuth 2.0 Demonstrating Proof of Possession (DPoP) specification</a>
+ *
+ */
+public class DPoPGenerator {
```
IngridPuppet, line 54: Missing authorship?

core/src/main/java/org/keycloak/util/DPoPGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+            keyWrapper.setKid(jwsHeader.getKeyId());
+            keyWrapper.setAlgorithm(algorithm);
+            keyWrapper.setPrivateKey(privateKey);
+            keyWrapper.setType(privateKey.getAlgorithm());
+            keyWrapper.setUse(KeyUse.SIG);
+            SignatureSignerContext sigCtx = createSignatureSignerContext(keyWrapper);
+
+            String data = dpopProofHeaderEncoded + "." + dpopProofPayloadEncoded;
+            byte[] signatureByteArray = sigCtx.sign(data.getBytes());
+            return data + "." + Base64Url.encode(signatureByteArray);
```
IngridPuppet: I wonder if we can't just use the JWSBuilder construct to produce this.

integration/admin-client/src/main/java/org/keycloak/admin/client/resource/DPoPAuthFilter.java
```diff
@@ -0,0 +1,49 @@
+package org.keycloak.admin.client.resource;
+
+import java.io.IOException;
+
```
IngridPuppet, line 21: Missing copyright.

IngridPuppet (commented):
Hello @mposolda - I checked the PR and have just a few minor comments. Please could you check them?

graziang (approved):
@mposolda thanks!

core/src/main/java/org/keycloak/util/DPoPGenerator.java
mposolda: Added

core/src/main/java/org/keycloak/util/DPoPGenerator.java
mposolda: Added

integration/admin-client/src/main/java/org/keycloak/admin/client/resource/DPoPAuthFilter.java
mposolda: Added

core/src/main/java/org/keycloak/util/DPoPGenerator.java
mposolda: This is good idea! Especially since the `DPoPGenerator` is now generic utility class.

However I also needed to do some minor changes in the testsuite as we have some failure scenarios tests, which try to test the scenarios when the different algorithm is used in the `JWSHeader` and in the JWK keys.

Also I unified the `JWSHeader` and `JWSBuilder` classes to both contain same type `List<String>` for the certificates, which should be sufficient. Since this was added after Keycloak 23 release, we also fortunately don't need to take care of the backwards compatibility here.

mposolda:
> Hello @mposolda - I checked the PR and have just a few minor comments. Please could you check them?

@IngridPuppet I've updated PR and applied your review comments. Could you please re-review? I've used separate commit, so you can easily review the changes. As usual, I plan to squash the commits during merge of the PR.

@tnorimat Do you please also have an opportunity to review this PR?

tnorimat:
@mposolda Yes, I am now reviewing the PR.

core/src/main/java/org/keycloak/util/DPoPGenerator.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    private static SignatureSignerContext createSignatureSignerContext(KeyWrapper keyWrapper) {
+        switch (keyWrapper.getType()) {
+            case KeyType.RSA:
+                return new AsymmetricSignatureSignerContext(keyWrapper);
+            case KeyType.EC:
+                return new ECDSASignatureSignerContext(keyWrapper);
+             // TODO: EdDSA?
```
tnorimat, line 111: Regarding supported signature algorithms for DPoP proof by https://github.com/keycloak/keycloak/pull/42200#pullrequestreview-3172969601 , I think it might be better to exclude EdDSA from the supported signature algorithms for DPoP proof. The reasons are as follows.

- Usually, a client generates DPoP proof and an authorization server like keycloak does not generate DPoP proof.
- Keycloak's `DPoPGenerator` is used by keycloak's admin client and integration tests that simulate a client creating a DPoP Proof.
- This method does not support EdDSA, which means that keycloak's admin client and integration tests that simulate a client creating a DPoP Proof cannnot create EdDSA signed DPoP proof.
- However, it does not mean that Keycloak does not verify EdDSA signed DPoP proof sent from a client.
- Therefore, we want to check that, but the current keycloak's integreation tests does not make its simulated client to generate EdDSA signed DPoP.

WDYT?

My idea is as follows:
- creating an optional issue to verify whether keycloak can verify EdDSA signed DPoP proof by integration tests.
- until accomplishing that, Keycloak exclude EdDSA from the supported signature algorithms for DPoP proof.

tnorimat: For verifying whether keycloak can verify EdDSA signed DPoP proof by integration tests, we need to implement `org.keycloak.crypto.EdDSASignatureContext` in `core` package.

tnorimat (commented):
I have not yet completed the review throughly, but I firstly add an important comment on DPoP supported signature algorithm.

IngridPuppet (approved):
My comments have been fully addressed. Thank you @mposolda for the update.

core/src/main/java/org/keycloak/util/DPoPGenerator.java
mposolda: @tnorimat Thanks for pointing this!

Considering this, I've merged PR https://github.com/keycloak/keycloak/pull/42200#discussion_r2313475567 and created follow-up issue for adding EdDSA support: https://github.com/keycloak/keycloak/issues/42286 . Feel free to add any additional comments to that issue.

I've also rebased this PR to latest main (after https://github.com/keycloak/keycloak/pull/42200). Please let me know if you have additional comments to this PR.

core/src/main/java/org/keycloak/util/DPoPGenerator.java
```diff
@@ -0,0 +1,117 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.util;
+
+import java.io.IOException;
```
tnorimat: It seems that `IOException` is not used anymore.

tnorimat (commented):
I added one comment. Could you check that?

core/src/main/java/org/keycloak/util/DPoPGenerator.java
mposolda: @tnorimat Yes, it is true. Good catch. I've updated the PR and squashed the commits. Could you please re-review?

tnorimat (approved):
LGTM

mposolda:
@tnorimat Thanks!

PR #42066, opened 2025-08-22
feat(#41959): adjusted UpdateUserLocaleAction to be able to take client requested locale into account

eicki:
Closes #41959

By default this is deactivated but can be activated by provider config option

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/requiredactions/UpdateUserLocaleAction.java
+++ b/services/src/main/java/org/keycloak/authentication/requiredactions/UpdateUserLocaleAction.java
@@ -12,6 +12,8 @@ import org.keycloak.models.UserModel;
 
 public class UpdateUserLocaleAction implements RequiredActionProvider, RequiredActionFactory {
 
+    private boolean updateOnClientRequestedLocale;
+
     @Override
     public String getDisplayText() {
         return "Update User Locale";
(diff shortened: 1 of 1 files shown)


pedroigor:
Did you try using the `kc_locale` query parameter? Does it work for your use case?

eicki:
The kc_locale parameter only works if the user actively switches the locale in the login screen and the login-actions/authenticate is called. It is not taken into consideration on initial protocol/openid-connect/auth endpoint, there only ui_locales actively switches it.

eicki:
And the Keycloak Javascript Adapter offers only setting ui_locales directly probably according to https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest.

pedroigor:
I'm afraid that update the locale at the client's discretion will introduce problems like:

* The locale being updated with the user's acknowledgment by any client in a realm, mainly public clients.
* An attacker could exploit this to change user's locale
* Updates to the user locale will end up invalidating the user cache entry, and an update to the database, and an attacker can exploit this to overload the server

eicki:
Ah I see, I thought that ui_locales would only take effect on prompt=login, but it is actually always handled so I see your attack vector. But there is also no possibility to pass through the ui_locales back to client token if not persisted.

eicki:
Feel free to reject/close this PR.

pedroigor:
Thanks for your time, @eicki. I think the locale should be persisted based on a user action.

Did you try to use a session mapper to grap the note from the session and added it as claim into your tokens?

eicki:
Yes that had been my first try, it is not available any more at that point when I map.

PR #42072, opened 2025-08-22
Do not query organizations if manage-realm is not granted

### LINKED ISSUE #41418: Access to user details for restricted admin fails after enabling organizationin realm
### Before reporting an issue

### Area

admin/ui

### Describe the bug

1 147 / 5 000
We have users with restricted administration rights on the realm.

In fact, these administrators have the following rights:
realm-management:query-clients, manage-users, query-users, view-realm, view-users, view-clients.

Since we enabled "organizations" on the realm, we have noticed that these administrators can no longer view a user's details.

In fact, a request to https://<keycloak_url>/auth/admin/realms/<realm>/organizations/?first=0&max=1 is triggered at least when accessing a user's details, and the result is a 403 Forbidden error!

By disabling "organizations" on the realm, these administrators regain the ability to view a user's details.

I haven't found any new rights dedicated to accessing organizations (query-organizations or view-organizations).

The only solution is to assign the "Manage-realm" role, which we don't want for these restricted users.
For example, we don't want that federation menu can be accessed in read/write mode for those restricted users.

Furthermore, I notice that "organizations" are present in the left-hand menu, and accessing this item produces the same error: 403 Forbidden.

Can you associate organization access with the "manage-users" role or something else to fix this regression?

Thanks in advance.
Envoyer des commentaires
Résultats de traduction disponibles 

### Version

26.0.5

### Regression

### Expected behavior

Access to all users details for a restricted administrator who have manage-users role even organizations are enabled

### Actual behavior

Forbidden access to organizations or users when trying to look at users details.

### How to Reproduce?

- Connect to Keycloak with a restricted user account with manage-users, query-users roles.
- Access to a user details in your realm  
- All is OK
- Activate organizations 
- Access to a user details in your realm  
- You should get 403 Forbidden error when keycloak tries accessing to organizations or click on organizations left menu.

### Anything else?

_No response_

pedroigor:
Closes #41418

* Currently, the Organization API relies on the `manage-realm` role to enforce admin access.
* The `useIsFeatureEnabled` hook is updated to also check permissions when checking if a feature is enabled. That should make the UI also a bit better because the `Organizations` settings in the `Realm Settings` tab is not even shown if the admin user does not have the required role.

### DIFF
--- a/js/apps/admin-ui/src/utils/useIsFeatureEnabled.ts
+++ b/js/apps/admin-ui/src/utils/useIsFeatureEnabled.ts
@@ -1,4 +1,5 @@
 import { useServerInfo } from "../context/server-info/ServerInfoProvider";
+import { useAccess } from "../context/access/Access";
 
 export enum Feature {
   AdminFineGrainedAuthz = "ADMIN_FINE_GRAINED_AUTHZ",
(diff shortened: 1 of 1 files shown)


martin-kanis (approved):
Thanks @pedroigor. I tested the PR in the admin console and it fixes the problem described in the issue. 
I let you decide whether you want another review from someone with deeper UI knowledge.

ahus1 (approved):

PR #42097, opened 2025-08-25
Concurrently update the remote caches

### LINKED ISSUE #42096: Concurrently update the remote caches
### Description

The Hot Rod client non-blocking/async API can be used to update all the caches involved in a Keycloak transaction concurrently. The same improvement is already available for session caches, and this issue is about extending it to the remaining caches. 

It mainly benefits multi-site deployments, where remote cache updates are expensive due to cross-availability zone requirements. 

As a visualization of the improvement, on the first trace, we have 3 sequential cache updates (see yellow traces). 

Image

And, after the improvement, we'll have the cache updates happening concurrently (yellow traces). Latency is reduced by 2/3rd in the login workflow. 

Image

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pruivo:
Closes #42096


ryanemerson (approved):

ahus1 (approved):
I see that these transactions are limited to remote caches, which reduces the potential blast radius. Well done, thanks!

Maybe in the future we would have async transactions in other places as well, let's see.

PR #42099, opened 2025-08-25
Add upstream architectures for the Operator

### LINKED ISSUE #38928: Can't install Keycloak Operator on OpenShift via OperatorHub on ARM
### Before reporting an issue

### Area

operator

### Describe the bug

When trying to install Keycloak on a ROSA cluster I can't find the Keycloak Operator in the OperatorHub

### Version

26.2.0

### Regression

### Expected behavior

I should be able to search for Keycloak, and find either the upstream or productized version of Keycloak

### Actual behavior

The result is empty

Image

### How to Reproduce?

Install ROSA with 4.17.20 (SRE team is doing it nightly) on ARM

### Anything else?

_No response_

ahus1:
Closes #38928

This generates the labels as expected: 

```
  labels:
    operatorframework.io/arch.amd64: supported
    operatorframework.io/arch.arm64: supported
```

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakSharedCsvMetadata.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakSharedCsvMetadata.java
@@ -141,7 +141,11 @@ import io.quarkiverse.operatorsdk.annotations.SharedCSVMetadata;
     icon = @CSVMetadata.Icon(
         fileName = "KeycloakController.icon.png",
         mediatype = "image/png"
-    )
+    ),
+    labels = {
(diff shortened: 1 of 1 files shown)


ahus1:
@mabartos - please review when you have the time. Thanks!

vmuzikar (approved):
LGTM, thank you, that's a nice addition.

@ASzc Just FYI – I believe this shouldn't conflict with what we do in downstream as we override these labels there, AFAIK.

ASzc:
Should be ok. These labels will probably make it through to the downstream CSV, but the script there will just set the same values. The only risk of a bug is if keycloak for some reason wanted to support an arch that downstream didn't. The script isn't set up to remove any arches right now

PR #42100, opened 2025-08-25
Add copy button to out-of-band code page

### LINKED ISSUE #42094: keycloak oob (out-of-band) copy button
### Description

It would be helpful if the out-of-band function had a copy button:

Image

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

TimMnz09:
## Description

This PR adds a copy button to the out-of-band code page to improve user experience when copying authentication codes.

## Changes

- Added a copy button with clipboard functionality to the code.ftl template
- Implemented modern clipboard API with fallback to execCommand for broader browser compatibility
- Added visual feedback showing Copied confirmation when the button is clicked
- Button is temporarily disabled after copying to prevent multiple rapid clicks

## Motivation

Currently, the out-of-band page displays a success code and instructs users to copy and paste it into their application, but there is no copy button provided. This can be inconvenient for users and may lead to errors when manually selecting and copying the code.

## Testing

The implementation includes:
- Modern navigator.clipboard API for secure contexts
- Fallback to document.execCommand for older browsers
- Visual feedback with icon change and temporary button disable
- Proper accessibility attributes

image

Closes #42094

### DIFF
--- a/themes/src/main/resources/theme/keycloak.v2/login/code.ftl
+++ b/themes/src/main/resources/theme/keycloak.v2/login/code.ftl
@@ -11,7 +11,84 @@
         <div id="kc-code">
             <#if code.success>
                 <p>${msg("copyCodeInstruction")}</p>
-                <@field.input name="code" label="" value=code.code />
+                <@field.clipboard name="code" label="" value=code.code />
+                
+                <script type="module">
+                    (() => {
+                        const copyButton = document.getElementById('kc-code-copy-button');
+                        const toggleButton = document.getElementById('kc-code-toggle');
+                        const input = document.getElementById('kc-code');
+                        const expandableContent = document.getElementById('kc-code-content');
+                        const clipboardContainer = document.getElementById('kc-code-clipboard');
+
+                        // Validate elements
+                        if (!copyButton || !toggleButton || !input || !expandableContent || !clipboardContainer) {
+                            console.error("Missing required DOM elements for code interactions.");
+                            return;
+                        }
+
+                        // Validate Clipboard API
+                        if (!navigator.clipboard) {
+                            console.error("Clipboard API not supported in this browser.");
+                            copyButton.style.display = 'none';
+                            return;
+                        }
+                        
+                        // Handle copy functionality
+                        copyButton.addEventListener('click', async() => {
+                            const originalIcon = copyButton.innerHTML;
+
+                            // Get value from expandable content
+                            let value = expandableContent.querySelector('code')?.textContent;
+
+                            try {
+                                await navigator.clipboard.writeText(value);
+                                updateCopyButton(true, copyButton, originalIcon);
+                            } catch (err) {
+                                console.error('Copy failed: ', err);
+                                updateCopyButton(false, copyButton, originalIcon);
+                            }
+                        });
+                        
+                        // Handle toggle functionality
+                        toggleButton.addEventListener('click', () => {
+                            const newExpanded = !(toggleButton.getAttribute('aria-expanded') === 'true');
+                            
+                            toggleButton.setAttribute('aria-expanded', newExpanded.toString());
+                            expandableContent.hidden = !newExpanded;
+                            
+                            // Update icon
+                            const icon = toggleButton.querySelector('#kc-code-toggle-icon');
+                            if (icon) {
+                                icon.className = newExpanded ? toggleButton.dataset.iconExpandedClass : toggleButton.dataset.iconCollapsedClass;
+                            }
+                            
+                            // Update clipboard container class
+                            if (newExpanded) {
+                                clipboardContainer.classList.add(toggleButton.dataset.classExpanded);
+                            } else {
+                                clipboardContainer.classList.remove(toggleButton.dataset.classExpanded);
+                            }
+                        });
+                    })();
+                    
+                    // Update copy button to show success or failure state
+                    function updateCopyButton(success, button, originalIcon) {
+                        button.setAttribute('aria-label', success ? button.dataset.successLabel : button.dataset.failureLabel);
+                        const icon = button.querySelector('#kc-code-copy-icon');
+                        if (icon) {
+                            icon.className = success ? button.dataset.iconSuccess : button.dataset.iconFailure;
+                        }
+                        button.disabled = true;
+                        
+                        // Reset button after 2 seconds
+                        setTimeout(() => {
+                            button.innerHTML = originalIcon;
+                            button.setAttribute('aria-label', '${msg("code-copy-label")}');
+                            button.disabled = false;
+                        }, 2000);
+                    }
+                </script>
             <#else>
                 <p id="error">${kcSanitize(code.error)}</p>
             </#if>
--- a/themes/src/main/resources/theme/keycloak.v2/login/code.ftl
+++ b/themes/src/main/resources/theme/keycloak.v2/login/code.ftl
@@ -11,8 +11,7 @@
         <div id="kc-code">
             <#if code.success>
                 <p>${msg("copyCodeInstruction")}</p>
-                <@field.clipboard name="code" label="" value=code.code />
-                
+                <@field.clipboard name="code" label="" ariaLabel=msg("codeSuccess") value=code.code />
                 <script type="module">
                     (() => {
                         const copyButton = document.getElementById('kc-code-copy-button');
--- a/themes/src/main/resources/theme/keycloak.v2/login/field.ftl
+++ b/themes/src/main/resources/theme/keycloak.v2/login/field.ftl
@@ -88,6 +88,61 @@
   </@group>
 </#macro>
 
+<#macro clipboard name label value="" readonly=true>
+  <@group name=name label=label>
+    <div class="${properties.kcCodeClipboardCopyClass}" id="kc-${name}-clipboard">
+      <div class="${properties.kcCodeClipboardCopyGroupClass}">
+        <div class="${properties.kcInputGroup}">
+          <div class="${properties.kcInputGroupItemClass}">
+            <button 
+              class="${properties.kcFormPasswordVisibilityButtonClass}" 
+              type="button" 
+              aria-label="${msg("code-clipboard-label")}"
+              aria-expanded="false"
+              aria-controls="kc-${name}-content"
+              data-icon-expanded-class="${properties.kcAngleDownIconClass}"
+              data-icon-collapsed-class="${properties.kcAngleRightIconClass}"
+              data-expanded-class="${properties.kcExpandedClass}"
+              id="kc-${name}-toggle"
+            >
+              <i id="kc-${name}-toggle-icon" class="${properties.kcAngleRightIconClass}" aria-hidden="true"></i>
+            </button>
+          </div>
+          <div class="${properties.kcInputGroupItemClass} ${properties.kcFill}">
+            <span class="${properties.kcInputClass} <#if readonly>${properties.kcFormReadOnlyClass}</#if>">
+              <input
+                id="kc-${name}"
+                name="${name}"
+                value="${value}"
+                type="text"
+                <#if readonly>readonly</#if>
+                aria-label="${label}"
+              />
+            </span>
+          </div>
+          <div class="${properties.kcInputGroupItemClass}">
+            <button
+              class="${properties.kcFormPasswordVisibilityButtonClass}"
+              type="button"
+              aria-label="${msg("code-copy-label")}"
+              data-icon-success="${properties.kcCheckIconClass}"
+              data-icon-failure="${properties.kcInputErrorIconClass}"
+              data-success-label="${msg("code-copy-success")}"
+              data-failure-label="${msg("code-copy-failure")}"
+              id="kc-${name}-copy-button"
+            >
+              <i id="kc-${name}-copy-icon" class="${properties.kcCopyIconClass}" aria-hidden="true"></i>
+            </button>
+          </div>
+        </div>
+      </div>
+      <div class="${properties.kcCodeClipboardCopyContentClass}" id="kc-${name}-content" hidden>
+        <pre><code>${value}</code></pre>
+      </div>
+    </div>
+  </@group>
+</#macro>
+
 <#macro checkbox name label value=false required=false>
   <div class="${properties.kcCheckboxClass}">
     <label for="${name}" class="${properties.kcCheckboxClass}">
--- a/themes/src/main/resources/theme/keycloak.v2/login/field.ftl
+++ b/themes/src/main/resources/theme/keycloak.v2/login/field.ftl
@@ -88,7 +88,7 @@
   </@group>
 </#macro>
 
-<#macro clipboard name label value="" readonly=true>
+<#macro clipboard name label ariaLabel=label value="" readonly=true>
   <@group name=name label=label>
     <div class="${properties.kcCodeClipboardCopyClass}" id="kc-${name}-clipboard">
       <div class="${properties.kcCodeClipboardCopyGroupClass}">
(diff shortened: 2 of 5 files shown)


ahus1:
Thank you for this screenshot. Looking at the very long code that a user can't actually see fully, an input which is only one line is not a good fit. Maybe a well-formatted span or div with some text in it that wraps would be a better fit. WDYT? Again, this might be a follow-up thing as it wasn't part of the original issue.  

Looping in @jonkoops to have an opinion here on the JavaScript, and optionally on the idea of not having an input field here.

TimMnz09:
> Thank you for this screenshot. Looking at the very long code that a user can't actually see fully, an input which is only one line is not a good fit. Maybe a well-formatted span or div with some text in it that wraps would be a better fit. WDYT? Again, this might be a follow-up thing as it wasn't part of the original issue.
> 
> Looping in @jonkoops to have an opinion here on the JavaScript, and optionally on the idea of not having an input field here.

@ahus1 I agree that a well-formatted span would be a better option, so a user can see the full code. But changing this might have an impact on custom themes and/or any E2E Tests. But I can also migrate the input into a span if you want me to do it. :)

ahus1:
@TimMnz09 - please give it a try while we wait for @jonkoops' feedback. E2E test changes shouldn't prevent us from improving the UX. If you at some point get doubts about the UX and it looks strange, then not do it. 

In any case, please share a screenshot.

TimMnz09:
@ahus1 + @jonkoops I updated the PR and improved the UX, so a user can see the full code. Please see the screenshot in the PR's description.
Let me know, if you'd like to have any adjustments.

jonkoops:
Thanks for the PR @TimMnz09, I totally agree we should provide a better experience here by offering a clipboard functionality.

I would prefer to use the built-in PatternFly components here as much as possible, and avoid adding custom styling. This makes the code more portable, and easier to customize when theming, as well as making things easier for us when we eventually need to upgrade to the next version of PatternFly.

Thankfully, PatternFly offers a [clipboard copy](https://v5-archive.patternfly.org/components/clipboard-copy) component that we can use out of the box, so let's use that here. We should strive to match the behavior of the React component here, with a tooltip when the copy succeeds. I think we can try to place this component in `field.ftl`, but worst case, it could be contained in its own template.

We expect that Keycloak is run in a secure context in any production scenario, so simply feature detecting the Clipboard API should be enough, removing the copy button if not present. Additionally, we do not support the legacy browser that might use `execCommand()` to copy text to the clipboard, so this fallback can be safely removed.

TimMnz09:
@jonkoops Thanks for your input! I updated my branch and pr with your proposals, but I'm not sure if this what you actually expected. Please let me know, if I should anything.

jonkoops:
Thanks! I will take a look at your proposed changes.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
         <div id="kc-code">
             <#if code.success>
                 <p>${msg("copyCodeInstruction")}</p>
-                <@field.input name="code" label="" value=code.code />
+                <@field.clipboard name="code" label="" value=code.code />
+                
+                <script>
```
jonkoops: This script should be of `type="module"`, this provides certain [guarantees about execution](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
```suggestion
                <script type="module">
```

jonkoops: Ideally, this script would be loaded when the field is added, no matter where. However, we cannot put this script in with the field template, as that would possibly cause the script to load more than once.

We need better primitives for enqueueing and embedding scripts, so they are only ever executed once. However, this is out of scope of this PR. For now, I think the approach here is the best we can do.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                document.addEventListener('DOMContentLoaded', function() {
```
jonkoops: When this script is of `type="module"`, it is no longer required to wait for the DOM to be ready, as that is [done implicitly](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event). This means this event listener code can be removed entirely.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                    const copyButton = document.getElementById('code-copy-button');
+                    const toggleButton = document.getElementById('code-toggle');
+                    const input = document.getElementById('code');
+                    const expandableContent = document.getElementById('code-content');
```
jonkoops: Generally speaking, we try to prefix our ids with `kc-` to prevent collisions with custom themes (although it isn't done very consistently), still let's try to use that convention here:
```suggestion
                    const copyButton = document.getElementById('kc-code-copy-button');
                    const toggleButton = document.getElementById('kc-code-toggle');
                    const input = document.getElementById('code');
                    const expandableContent = document.getElementById('kc-code-content');
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                    
+                    // Only show copy button if Clipboard API is available
+                    if (!navigator.clipboard) {
+                        if (copyButton) {
```
jonkoops: Instead of checking that elements exist here, we should do some validation beforehand in a separate step to prevent any code from executing, logging an error to the console.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            copyButton.style.display = 'none';
+                        }
+                        return;
+                    }
+                    
+                    // Handle copy functionality
+                    if (copyButton && input) {
```
jonkoops: Same here, I'd prefer we do all these validations on top and exit early.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                        copyButton.addEventListener('click', function() {
```
jonkoops: As a matter of convention, we tend to use arrow functions for anonymous functions. Let's also make this `async`, so we can use `await` later on.
```suggestion
                        copyButton.addEventListener('click', async () => {
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            const originalIcon = this.innerHTML;
+                            const value = input.value;
+                            
+                            navigator.clipboard.writeText(value).then(function() {
+                                showCopySuccess(copyButton, originalIcon);
+                            }).catch(function(err) {
+                                console.error('Copy failed: ', err);
+                            });
```
jonkoops: Let's use a try-catch here by awaiting the promise instead.
```suggestion
                            try {
                                await navigator.clipboard.writeText(value);
                                showCopySuccess(copyButton, originalIcon);
                            } catch (err) {
                                console.error('Copy failed: ', err);
                            }
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                    // Handle copy functionality
+                    if (copyButton && input) {
+                        copyButton.addEventListener('click', function() {
+                            const originalIcon = this.innerHTML;
+                            const value = input.value;
+                            
+                            navigator.clipboard.writeText(value).then(function() {
+                                showCopySuccess(copyButton, originalIcon);
+                            }).catch(function(err) {
+                                console.error('Copy failed: ', err);
```
jonkoops, line 47: Ideally, we'd show an error to the user here if the action fails. Otherwise, they might get the impression that the copy succeeded.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            });
+                        });
+                    }
+                    
+                    // Handle toggle functionality
+                    if (toggleButton && expandableContent) {
```
jonkoops: Same here, let's get rid of this defensive coding and move it up top.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                        toggleButton.addEventListener('click', function() {
```
jonkoops:
```suggestion
                        toggleButton.addEventListener('click', () => {
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            const isExpanded = this.getAttribute('aria-expanded') === 'true';
```
jonkoops: I'd prefer if we do not use `this`, as its value can change depending on context, making it hard to keep track of. Instead, let's use the variables we already have.
```suggestion
                            const isExpanded = toggleButton.getAttribute('aria-expanded') === 'true';
```

Nice use of existing aria attributes for state, that is a perfect way to handle this.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            const newExpanded = !isExpanded;
```
jonkoops: Since `isExpanded` is only used once, I would be fine with in-lining it here.
```suggestion
                            const newExpanded = !(toggleButton.getAttribute('aria-expanded') === 'true');
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            
+                            this.setAttribute('aria-expanded', newExpanded.toString());
```
jonkoops:
```suggestion
                            toggleButton.setAttribute('aria-expanded', newExpanded.toString());
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            expandableContent.hidden = !newExpanded;
+                            
+                            // Update icon
+                            const icon = this.querySelector('i');
```
jonkoops:
```suggestion
                            const icon = toggleButton.querySelector('i');
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            if (icon) {
+                                icon.className = newExpanded ? 'fas fa-angle-down' : 'fas fa-angle-right';
```
jonkoops: Because themes might customize these elements, we should store things like classes that should be applied in `data-` attributes, this allows custom themes to simply update the existing template to modify this behavior, e.g.:

```html
<i class="fas fa-check" data-expanded-class="fa-angle-down" data-collapsed-class="fa-angle-right"></i>
```

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
jonkoops: We cannot assume the element is a `i`, ideally this would use a class, or id selector. This allows custom themes to work as well.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            }
+                            
+                            // Update clipboard container class
+                            const clipboardContainer = document.getElementById('code-clipboard');
```
jonkoops: Is there any specific reason we are only getting the clipboard container at this point? Ideally, it would be retrieved in the beginning (like the other elements), and validated there.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                            if (clipboardContainer) {
+                                if (newExpanded) {
+                                    clipboardContainer.classList.add('pf-m-expanded');
```
jonkoops: Same for these classes, we cannot hard-code them, they will have to be provided through `data-` attributes.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                                } else {
+                                    clipboardContainer.classList.remove('pf-m-expanded');
+                                }
+                            }
+                        });
+                    }
+                    
+                    function showCopySuccess(button, originalIcon) {
+                        // Show success feedback with tooltip-like behavior
+                        button.innerHTML = '<i class="fas fa-check" aria-hidden="true"></i>';
```
jonkoops: If you need to use some HTML, preferably it should go in a `<template>` tag, co-located with the original template. This allows for easier customization by custom theme authors, without then also needing to update this template with the JavaScript.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
```diff
@@ -11,10 +11,82 @@
...
+                        button.setAttribute('aria-label', '${msg("recovery-codes-copied")}');
+                        button.disabled = true;
+                        
+                        // Reset button after 2 seconds
+                        setTimeout(function() {
```
jonkoops:
```suggestion
                        setTimeout(() => {
```

themes/src/main/resources/theme/keycloak.v2/login/field.ftl
```diff
@@ -88,6 +88,54 @@
   </@group>
 </#macro>
 
+<#macro clipboard name label value="" readonly=true>
+  <@group name=name label=label>
+    <div class="pf-v5-c-clipboard-copy" id="${name}-clipboard">
```
jonkoops: Like other classes, e.g. `kcInputGroupItemClass`, this class should be provided by a property to allow easy customization. This applies for all `pf-` classes.

themes/src/main/resources/theme/keycloak.v2/login/field.ftl
```diff
@@ -88,6 +88,54 @@
...
+      <div class="pf-v5-c-clipboard-copy__group">
+        <div class="${properties.kcInputGroup}">
+          <div class="${properties.kcInputGroupItemClass}">
+            <button 
+              class="${properties.kcFormPasswordVisibilityButtonClass}" 
+              type="button" 
+              aria-label="Show content"
```
jonkoops: This label should be translated.

themes/src/main/resources/theme/keycloak.v2/login/code.ftl
TimMnz09: I completely reworked this. Instead of using template of inline html, I also switched to `data-` labels.

TimMnz09:
@jonkoops Thanks a lot for your review. I updated my commit with you suggestions. Maybe you can take another look.

jonkoops (approved):
Looks great, thanks!

ahus1 (dismissed):

ahus1 (approved):
As one of the tests was failing, I've reverted the ID of the input field to the original ID of the field. This will also help others who are running automation tasks against Keycloak if it remains constant. 

Also did a tiny bit of CSS to wrap at any character as this seems to be the most appropriate way to break a code.

themes/src/main/resources/theme/keycloak.v2/login/resources/css/styles.css
```diff
@@ -109,6 +109,10 @@ div.kc-logo-text span {
     text-align: center;
 }
 
+#kc-code pre code {
+    word-break: break-all;
+}
```
jonkoops, line 114: Rather than introducing custom CSS, I would prefer we use [PatternFly utility classes](https://v5-archive.patternfly.org/utility-classes/text#controlling-text) where possible. Not sure if there is one for `break-all`.

PR #42121, opened 2025-08-25
Fix #42114 - Prevent caching closed EntityManager in KeycloakSession attributes

vellotis:
Fixes (#42114) "Session/EntityManager is closed" when trying to perform users import during application bootstrapping.

While performing `DefaultJpaConnectionProviderFactory::addSpecificNamedQueries` during lazy initialization the entity manager should not be cached for users import batch processing (aka "nested entity managers"). The EntityManager instance is transient and the given method closes it and the JPA connection along with it during the method's "finally" block.

The issue got introduced with commit https://github.com/keycloak/keycloak/commit/24910d9e1c76a1e953936c558ab02a27eb91820e which was first introduced to Keycloak v26.3.0.

Closes #42114

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/connections/jpa/DefaultJpaConnectionProviderFactory.java
+++ b/model/jpa/src/main/java/org/keycloak/connections/jpa/DefaultJpaConnectionProviderFactory.java
@@ -89,10 +89,10 @@ public class DefaultJpaConnectionProviderFactory implements JpaConnectionProvide
         logger.trace("Create JpaConnectionProvider");
         lazyInit(session);
 
-        return new DefaultJpaConnectionProvider(createEntityManager(session));
+        return new DefaultJpaConnectionProvider(createEntityManager(session, true));
     }
 
-    private EntityManager createEntityManager(KeycloakSession session) {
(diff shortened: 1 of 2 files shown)


shawkins:
@vellotis thank you for the pr, can you add a test case?

Also I think we can just localize the fix, would you be okay if I pushed a commit to your pr?

vellotis:
@shawkins could you advise me a bit with this? How do you see it should be tested? There is no test suite that covers this kind of use cases. I thought about it many times, but finally dropped the initiative and considered that I will figure the testing out with the codebase owners.

- I would build a sanity test suite for KeycloakServer. But there are some downsides:
  - The exception does not prevent the process from keeping running.
  - The messages are written only to stdout or logger.
- The nicest approach I could come up with is to parse the stdout similarly to [`org.keycloak.it.junit5.extension.DistributionTest`](https://github.com/keycloak/keycloak/blob/36cbbbc7b4835b1a091449677a7e2d237892c50f/quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/DistributionTest.java#L30) annotation, but for KeycloakServer.
  - But is there a way to run `KeycloakServer` as a process other than using `mvn exec:java` from inside the test case?

ahus1:
Closed in favor of #42146

PR #42125, opened 2025-08-25
fix: moving multi-option validation to propertymappergrouping interface

### LINKED ISSUE #27025: Move import/export validation to the property mappers
### Description

After merging https://github.com/keycloak/keycloak/pull/25333, the validation of properties for Import/Export is done directly in the command classes. The validation checks, whether only `--file`, or `--dir` is specified. 

The validation might be done directly in the related property mappers. 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
closes: #27025

Since most of the grouping don't do validation, we could instead have a method that returns validation functions rather than directly calling a validateConfig method.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
@@ -137,4 +137,11 @@ public abstract class AbstractCommand implements Callable<Integer> {
         return Environment.PROD_PROFILE_VALUE;
     }
 
+    /**
+     * @return true if the command results in a server
+     */
+    public boolean isServing() {
+        return false;
+    }
+
 }
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
@@ -138,7 +138,7 @@ public abstract class AbstractCommand implements Callable<Integer> {
     }
 
     /**
-     * @return true if the command results in a server
+     * @return true if the command starts an http server
      */
     public boolean isServing() {
         return false;
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
@@ -1,0 +1,13 @@
+package org.keycloak.quarkus.runtime.configuration.mappers;
+
+import org.keycloak.quarkus.runtime.cli.Picocli;
+
+public interface PropertyMapperGrouping {
+
+    PropertyMapper<?>[] getPropertyMappers();
+
+    default void validateConfig(Picocli picocli) {
+
+    }
+
+}
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
@@ -1,10 +1,12 @@
 package org.keycloak.quarkus.runtime.configuration.mappers;
 
+import java.util.List;
+
 import org.keycloak.quarkus.runtime.cli.Picocli;
 
 public interface PropertyMapperGrouping {
 
-    PropertyMapper<?>[] getPropertyMappers();
+    List<? extends PropertyMapper<?>> getPropertyMappers();
 
     default void validateConfig(Picocli picocli) {
 
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
@@ -44,6 +44,21 @@ public final class PropertyMappers {
     public static String VALUE_MASK = "*******";
     private static MappersConfig MAPPERS;
     private static final Logger log = Logger.getLogger(PropertyMappers.class);
+    private final static List<PropertyMapperGrouping> GROUPINGS;
+    static {
+        GROUPINGS = List.of(new CachingPropertyMappers(), new DatabasePropertyMappers(),
+                new ConfigKeystorePropertyMappers(), new EventPropertyMappers(), new ClassLoaderPropertyMappers(),
+                new ExportPropertyMappers(), new BootstrapAdminPropertyMappers(), new HostnameV2PropertyMappers(),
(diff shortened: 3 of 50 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
```diff
@@ -137,4 +137,11 @@ protected String getDefaultProfile() {
         return Environment.PROD_PROFILE_VALUE;
     }
 
+    /**
+     * @return true if the command results in a server
+     */
+    public boolean isServing() {
```
mabartos, line 143: Wouldn't be better to move the `shouldStart()` to the AbstractCommand instead of having the `isServing()`? The `isServing` is always false and for the AbstractStartCommand is set to `shouldStart()`. If you'd updated the description of the `shouldStart()`, it might be applicable even here, right? 

Or if those should not be interchangeable concepts, please, make the description of this method more descriptive to see the exact purpose. Thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
```diff
@@ -0,0 +1,13 @@
+package org.keycloak.quarkus.runtime.configuration.mappers;
+
+import org.keycloak.quarkus.runtime.cli.Picocli;
+
+public interface PropertyMapperGrouping {
+
+    PropertyMapper<?>[] getPropertyMappers();
```
mabartos:
```suggestion
    List<PropertyMapper<?>> getPropertyMappers();
```

Would it be possible to use some collection? List? 

Please add some javadoc.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
```diff
@@ -43,6 +43,21 @@ public final class PropertyMappers {
     public static String VALUE_MASK = "*******";
     private static MappersConfig MAPPERS;
     private static final Logger log = Logger.getLogger(PropertyMappers.class);
+    private final static List<PropertyMapperGrouping> GROUPINGS;
+    static {
+        GROUPINGS = List.of(new CachingPropertyMappers(), new DatabasePropertyMappers(),
```
mabartos, line 49: For the future, we could maybe use some Service Loader/SPI for these mappers as users could add their own mappers/options instead of having the loong SPI props for their extensions. 

@shawkins WDYT?

mabartos (commented):
@shawkins Overall LGTM, just added some comments. Let me know what you think. Thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
shawkins: > Or if those should not be interchangeable concepts, please, make the description of this method more descriptive to see the exact purpose. Thanks!

They are unfortunately not interchangeable. We could probably rename AbstractStartCommand at this point to AbstractAutoBuildableCommand - the subclasses all need to consider whether a build happens, but not all of them will actually "start" quarkus (Picocli.start / KeycloakMain.start). That is what the shouldStart flag covers.

isServing is a related concept - only the subset of commands that start the server actually serve requests (basically just start and start-dev).

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
shawkins: > Would it be possible to use some collection? List?

Doesn't seem like it will add much value to make that change. What are your thoughts?

> Please add some javadoc.

Sure, I can add that.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
shawkins: Allowing extensions to contribute property mappers via service loading would open up a significant amount of supported surface area. It would make more sense to refine the existing facilities for how extensions advertise option metadata.

If we eventually change our logic so that it's running after the quarkus launch, then we'd be able to use CDI rather than building this static list of groupings.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
mabartos: @shawkins Once we're creating such an interface, I'd prefer to use some collection instead of plain arrays. 

There should not be any performance implications by doing that, right?

I'd rather see `List.of(...)` instead of `new PropertyMapper[] {...}`.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
mabartos: > Allowing extensions to contribute property mappers via service loading would open up a significant amount of supported surface area. It would make more sense to refine the existing facilities for how extensions advertise option metadata.

+1

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractCommand.java
mabartos: > We could probably rename AbstractStartCommand at this point to AbstractAutoBuildableCommand

@shawkins Yes, that might be also probably cleaner as the UpdatesCommand and NonServerCommand extend the AbstractStartCommand, which might cause some confusion. 

Having the `AbstractAutoBuildableCommand` would be better. 

> isServing is a related concept - only the subset of commands that start the server actually serve requests (basically just start and start-dev).

Ok, makes sense, but please update the JavaDoc to reflect the purpose to have it there for our future us and for other folks to understand it. Thanks!

shawkins:
@mabartos your comments should be addressed now.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapperGrouping.java
```diff
@@ -0,0 +1,15 @@
+package org.keycloak.quarkus.runtime.configuration.mappers;
+
+import java.util.List;
+
+import org.keycloak.quarkus.runtime.cli.Picocli;
+
+public interface PropertyMapperGrouping {
```
mabartos, line 7: @shawkins Would be good to provide some JavaDoc even for the new interfaces we make to clarify the purpose and comment on the desired contract. 

As it's primarily an internal interface and for the PMs it's straightforward to understand it, I'm ok with it. But if you'd find some time in some follow-up PR to provide some Javadoc, that'd be nice. But no hard requirement.

mabartos (approved):
@shawkins Nice work! LGTM

PR #42135, opened 2025-08-26
Disable peristent user session batching

### LINKED ISSUE #41662: TiDB Many WAITING threads during high load scenario
### Before reporting an issue

### Area

core

### Describe the bug

I am currently testing our KC to see how much load it can handle in the current configuration.

Running on EKS using 3 dedicated c7i.xlarge machines
Because some DaemonSets are running also there, this is the KC resource config:
```
resources:
     limits:
       cpu: "3"
       memory: 6G
     requests:
       cpu: "3"
       memory: 6G
```

Its a KC 26.2.5 instance

database connections
```
- name: KC_DB_POOL_INITIAL_SIZE
  value: "25"
- name: KC_DB_POOL_MAX_SIZE
  value: "25"
- name: KC_DB_POOL_MIN_SIZE
  value: "25"
```

```
KC_HTTP_MAX_QUEUED_REQUESTS=1000
KC_HTTP_POOL_MAX_THREADS=50
```

230 http req/sec are generated (~ 30 Logins/sec, ~30 code2token/sec,~150 refreshtoken/sec) via the Keycloak benchmark cli.
It is strange that with this number of requests, the response times increase, even though the CPU of the pod is only at 80%. The database connection pool is only utilized at 15%.

Image

It seems that the requests are spending a long time in the WorkerPool Queue (since the usage duration values are very low, and queue duration values are high):

Image

That's why I increased the KC_HTTP_POOL_MAX_THREADS, unfortunately without success.

Then I looked into Threads with JavaMissionControl. Seems to be that all WAITING threads are blocked by this line of code:
[https://github.com/keycloak/keycloak/blob/73e5f4bead7840f19210aaabcbd1a00895d46929[…]oak/models/sessions/infinispan/changes/JpaChangesPerformer.java](https://github.com/keycloak/keycloak/blob/73e5f4bead7840f19210aaabcbd1a00895d46929/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java#L109C23-L109C59)

Image

Furthermore I saw that the vert.x threads are under heavy load. Is that the newly introduced virtual threads feature for infinispan/jgroups?

Any Ideas? 

### Version

KC 26.2.5

### Regression

### Expected behavior

CPU can be fully utilized

### Actual behavior

CPU is not fully utilized, but threads are waiting on each other 

### How to Reproduce?

Run Keycloak Benchmark up to 250req/sec with scenario `keycloak.scenario.authentication.AuthorizationCode` on 3x 3CPU/6GB RAM KC Cluster 

### Anything else?

_No response_

pruivo:
Closes #41662

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
@@ -35,7 +35,6 @@ import org.keycloak.infinispan.util.InfinispanUtils;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.KeycloakSessionFactory;
-import org.keycloak.models.KeycloakSessionTask;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.models.UserSessionProvider;
@@ -59,9 +58,12 @@ import org.keycloak.models.utils.ResetTimeOffsetEvent;
 import org.keycloak.provider.EnvironmentDependentProviderFactory;
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.provider.ProviderConfigurationBuilder;
-import org.keycloak.provider.ProviderEvent;
-import org.keycloak.provider.ProviderEventListener;
 import org.keycloak.provider.ServerInfoAwareProviderFactory;
+
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.CLIENT_SESSION_CACHE_NAME;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.OFFLINE_CLIENT_SESSION_CACHE_NAME;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.OFFLINE_USER_SESSION_CACHE_NAME;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.USER_SESSION_CACHE_NAME;
 
 public class InfinispanUserSessionProviderFactory implements UserSessionProviderFactory<UserSessionProvider>, ServerInfoAwareProviderFactory, EnvironmentDependentProviderFactory {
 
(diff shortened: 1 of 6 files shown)


ahus1:
@pruivo - the `PersistentSessionsWorker` had a `Retry.executeWithBackoff`, and it seems to me that this would be needed to handle concurrent updates in multiple sessions. 

The code might be added here: 

https://github.com/keycloak/keycloak/blob/4dc4de7c1218bd129b4ae6853a74b7a2954b6eec/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/PersistentSessionsChangelogBasedTransaction.java#L144-L148

Still there is a caveat as the current `applyChangesSynchronously` calls a `clear()` before the changes are committed.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
```diff
@@ -77,20 +79,18 @@ public class InfinispanUserSessionProviderFactory implements UserSessionProvider
     public static final String CONFIG_USE_CACHES = "useCaches";
     private static final boolean DEFAULT_USE_CACHES = true;
     public static final String CONFIG_USE_BATCHES = "useBatches";
-    private static final boolean DEFAULT_USE_BATCHES = true;
+    private static final boolean DEFAULT_USE_BATCHES = false;
```
ryanemerson, line 82: There's a lot of changes in this file just to change this from true to false :rofl:

ryanemerson (dismissed):

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
pruivo: You know I don't like to see those annoying IDE warnings with easy fixes :rofl:

ahus1 (approved):
Thank you for the PR. I've added one more commit to bring back some of the functionality present in the batched variant to prevent too many iterations.

ryanemerson (approved):

Issue #39220, opened 2025-04-25
Admin API v2


vmuzikar:
### Description

Admin REST API is a key part of Keycloak. Nearly every configuration interface (UI, CLI, Operator, …) uses it as its backend. As it evolved over the years of Keycloak’s existence, it needed to cover more and more areas with an increasing number of use cases and types of API clients. This explosive growth led to many design issues and inconsistencies making it difficult or sometimes even impossible for the API clients to use it for certain scenarios.

A new version of Admin API will aim to resolve the issues.

### Goals
* Blueprint for the new API
    * Includes designs and prototypes for proofing them.
        * Focused on the Keycloak Clients area as that will be the first to get implemented.
        * Designs and prototypes cover all use cases of the API. That includes:
            * The API itself
            * All first party clients: CLI, Java, JS/Admin Console, Operator, Terraform Provider
            * Docs
    * The result of the designs and prototypes is a document with final guidelines that can be used when implementing any part of the API.
* Implementation of the new API
    * Created as v2
        * Current Admin API (which will be marked as v1) will coexist with it it
    * Will be done incrementally.
        * Area by area, with each area also being implemented iteratively.
    * Won’t clash with v1, v1 and v2 can be used side by side so users can gradually migrate as new areas are implemented.
        * This applies both for the API itself and its clients too.
* Deprecation of the v1 API.

This is an umbrella issue for tracking the long-term effort for Admin API v2. The list of sub-issues will be updated over time.

### Motivation

The current API is suffering from many design issues. Fixing those issues requires redesigning the API from scratch as it is not viable to fix the existing API due to how extensive the changes need to be.

https://github.com/keycloak/keycloak/discussions/37655 covers known issues with the current API.

vmuzikar:
For Clients, it was superseded by https://github.com/keycloak/keycloak/issues/43165. We will be creating additional issues for individual API areas later.

PR #42138, opened 2025-08-26
Add automatic GUID/UUID search support for users

TimMnz09:
## Description

This PR enhances user search functionality by automatically detecting and searching by UUID/GUID without requiring the 'id:' prefix, making user ID search more discoverable and user-friendly.

### Changes Made

- **Backend Changes:**
  - Added automatic UUID pattern detection in `UsersResource.java` and `BruteForceUsersResource.java`
  - Enhanced both `getUsers()` and `getUsersCount()` methods to support automatic GUID detection
  - Maintains backward compatibility with existing `id:` prefix syntax

- **UI Enhancement:**
  - Updated search placeholder text to be more concise: `"Search by name, email or id"`
  - Added consistent width styling to search input for better visual presentation    
  - image

- **Documentation:**
  - Added comprehensive documentation for ID search methods in `proc-searching-user.adoc`
  - Documented both explicit `id:` syntax and automatic GUID detection
  - Updated main search description to mention user ID search capability

- **Testing:**
  - Updated admin e2e tests to reflect UI changes
  - Added comprehensive test coverage for the new UUID detection functionality

### Implementation Details

The solution uses a regex pattern to detect standard UUID format (`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`) and automatically treats matching inputs as user ID searches.

**Examples:**
- `f47ac10b-58cc-4372-a567-0e02b2c3d479` - Automatically detected as UUID search
- `id:f47ac10b-58cc-4372-a567-0e02b2c3d479` - Explicit ID search (existing functionality)
- `id:uuid1 uuid2 uuid3` - Multiple ID search (existing functionality)

### Testing

The implementation maintains backward compatibility and adds the new automatic detection as an additional search method. All existing search functionality remains unchanged.

Co-authored-by: @pedroigor

Closes #42071


TimMnz09:
@pedroigor as you assigned #42071 to yourself. Maybe you can review this pr.

pedroigor:
@TimMnz09 I don't have the rights to force push to your branch. Would you mind updating it with https://github.com/pedroigor/keycloak/tree/user-search-guid-support-42071?

I'm basically moving the check on the format to a single place, given that we have the same check in different places.

Kamys:
What if a UUID is used as a username?  
For example, in my project we had to migrate users from a legacy DB without usernames, and we generated UUIDs as `username` values. In such case the new behavior might incorrectly interpret the input as "search by id" instead of "search by username"

pedroigor:
@Kamys In this case, you would set the `username` query parameter, or `q=username:<uuid>`, and avoid using the `search` parameter.

I was unaware of people using UUIDs in usernames, as they seem counterintuitive. But I guess we should at least add a release note about this change to avoid surprises when upgrading.

TimMnz09:
> @TimMnz09 I don't have the rights to force push to your branch. Would you mind updating it with https://github.com/pedroigor/keycloak/tree/user-search-guid-support-42071?
> 
> I'm basically moving the check on the format to a single place, given that we have the same check in different places.

@pedroigor I updated my branch and this pull-request. Thanks for your input.

TimMnz09:
> What if a UUID is used as a username? For example, in my project we had to migrate users from a legacy DB without usernames, and we generated UUIDs as `username` values. In such case the new behavior might incorrectly interpret the input as "search by id" instead of "search by username"

@Kamys We could implement a fallback, like if user was not found by "search by id", we could than "search by username". @pedroigor what do you think?

pedroigor:
@TimMnz09 I was told to hold this PR because of the upcoming version of the Admin API. See https://github.com/keycloak/keycloak/issues/39220.

Sorry for taking up your time. Even though your changes make sense and make it easier, I cannot merge them. The API is already complex and adding yet another way to search users by ID, in an already complex and hard to follow API, is not in alignment with our future goals around the Admin API.

PR #42140, opened 2025-08-26
Configure topology information in Infinispan

pruivo:
Closes #41933

### DIFF
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -203,6 +203,55 @@ For these use case, the CLI options `cache-remote-tls-enabled` disables the encr
 The CLI options `cache-remote-username` and `cache-remote-password` are optional and, if not set, {project_name} will connect to the {jdgserver_name} server without presenting any credentials.
 If the {jdgserver_name} server has authentication enabled, {project_name} will fail to start.
 
+[[cache-topology]]
+== Topology Aware Data Availability
+
+Configuring the topology information of your cluster increases the data availability in case of hardware failure.
+The hardware where the {project_name} instance runs is necessary to compute this topology information.
+This section is more relevant for bare-metal users, who have control over the hardware and where it is running.
+
+[NOTE]
+====
+By default, user and client sessions are safely stored in the database, and they are not affected by these settings.
+The remaining distributed caches are affected by this configuration.
+====
+
+The following topology information is available to configure:
+
+Site name::
+If your {project_name} cluster is deployed between different datacenters (stretched cluster), use this option to ensure the data replicas are stored in a different datacenter.
+It prevents data loss if a datacenter goes offline or fails.
++
+Use the SPI option `spi-cache-embedded--default--site-name` (or environment variable `KC_SPI_CACHE_EMBEDDED\__DEFAULT__SITE_NAME`).
+The value itself is not important, but each datacenter must have a unique value.
++
+For example: `--spi-cache-embedded--default--site-name=site-1`
+
+Rack name::
+If your {project_name} cluster is running in different racks on your datacenter, set this option to ensure the data replicas are stored in a different physical rack.
+It prevents data loss if a rack is suddenly disconnected or fails.
++
+Use the SPI option `spi-cache-embedded--default--rack-name` (or environment variable `KC_SPI_CACHE_EMBEDDED\__DEFAULT__RACK_NAME`).
+The value itself is not important, but each rack must have a unique value.
++
+For example: `--spi-cache-embedded--default--rack-name=rack-1`
+
+Machine name::
+If you have multiple {project_name} instances running on the same physical machine (using virtual machines or containers for example), use this option to ensure the data replicas are stored in different physical machines.
+It prevents data loss against a physical machine failure.
++
+Use the SPI option `spi-cache-embedded--default--machine-name` (or environment variable `KC_SPI_CACHE_EMBEDDED\__DEFAULT__MACHINE_NAME`).
+The value itself is not important, but each machine must have a unique value.
++
+For example: `--spi-cache-embedded--default--machine-name=machine-1`
++
+[NOTE]
+====
+The {project_name} Operator automatically configure the machine name based on the Kubernetes node.
+It ensures if Pods are scheduled in the same node, the data is safely replicated into another node.
+We recommend to set up anti-affinity rules and/or topology spread constraints to prevent multiple Pods from being scheduled on the same node, further reducing the risk of a single node failure incur a data loss.
+====
+
 == Transport stacks
 
 Transport stacks ensure that {project_name} nodes in a cluster communicate in a reliable fashion.
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -204,11 +204,10 @@ The CLI options `cache-remote-username` and `cache-remote-password` are optional
 If the {jdgserver_name} server has authentication enabled, {project_name} will fail to start.
 
 [[cache-topology]]
-== Topology Aware Data Availability
+== Topology aware data distribution
 
-Configuring the topology information of your cluster increases the data availability in case of hardware failure.
-The hardware where the {project_name} instance runs is necessary to compute this topology information.
-This section is more relevant for bare-metal users, who have control over the hardware and where it is running.
+Configuring {project_name} to be aware of your network topology, increases data availability in the presence of hardware failures, as Infinispan is able to ensure that data is distributed correctly.
+For example, if `num_owners=2` is configured for a cache, it will ensure that the two owners are not stored on the same node when possible.
 
 [NOTE]
 ====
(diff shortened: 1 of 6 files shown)


pruivo:
Could you suggest where the documentation about this should be located?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -557,6 +558,12 @@ private List<EnvVar> getDefaultAndAdditionalEnvVars(Keycloak keycloakCR) {
         envVars.add(new EnvVarBuilder().withName(POD_IP).withNewValueFrom().withNewFieldRef()
                 .withFieldPath("status.podIP").withApiVersion("v1").endFieldRef().endValueFrom().build());
 
+        // Both status.hostIP or spec.nodeName would be fine here.
+        // In theory, status.hostIP is a smaller value and, as this value is tagged in all JGroups messages, it should have a lower overhead.
+        // Note: the value (IP) is shown in the log, hopefully it is fine.
```
pruivo: As an example, starting keycloak with `--spi-cache-embedded-default-node-name=my-node --spi-cache-embedded-default-site-name=my-site --spi-cache-embedded-default-machine-name=my-machine --spi-cache-embedded-default-rack-name=my-rack` will log the following

`ISPN000094: Received new cluster view for channel ISPN: [my-node(site-id=my-site, rack-id=my-rack, machine-id=my-machine)|0] (1) [my-node(site-id=my-site, rack-id=my-rack, machine-id=my-machine)]`

ahus1:
About documentation: 

We could add information here: https://www.keycloak.org/nightly/high-availability/single-cluster/concepts as part of "Multiple Zones" part. Something around the following lines: 

> By default, Keycloak will avoid to spin up Keycloak nodes on the same when running with the Keycloak Operator. You can provide additional SPI options to make Keycloak aware of the rack and site to allow distributing your data. This is currently not supported when deploying Keycloak with the Operator.

I thought maybe adding something to https://www.keycloak.org/nightly/server/caching as well - but then I decided against it.

ahus1 (approved):
LGTM on the level of reviewed code. Approving the current status so far even if you plan to add docs.

ahus1 (dismissed):
LGTM on the level of reviewed code. Approving the current status so far even if you plan to add docs.

pruivo:
> About documentation:
> 
> We could add information here: https://www.keycloak.org/nightly/high-availability/single-cluster/concepts as part of "Multiple Zones" part. Something around the following lines:
> 
> > By default, Keycloak will avoid to spin up Keycloak nodes on the same when running with the Keycloak Operator. You can provide additional SPI options to make Keycloak aware of the rack and site to allow distributing your data. This is currently not supported when deploying Keycloak with the Operator.
> 
> I thought maybe adding something to https://www.keycloak.org/nightly/server/caching as well - but then I decided against it.

IMO, it has little to no value for kubernetes users as the operator takes care of it. Bare metal users will benefit the more from this as they know how the physical machine are located in the datacenter.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
ryanemerson: I think we should probably go with `spec.nodeName` as it will make the logs easier to debug (no ip -> nodeName lookup) and there will be no security concerns.

ryanemerson (commented):
Just one comment on the code.

> We could add information here: https://www.keycloak.org/nightly/high-availability/single-cluster/concepts as part of "Multiple Zones" part. Something around the following lines:

+1. We can also update point 3 in the "Known limitations" section to mention how these settings can be used to mitigate such problems.

ryanemerson:
> IMO, it has little to no value for kubernetes users as the operator takes care of it. Bare metal users will benefit the more from this as they know how the physical machine are located in the datacenter.

I think we need to explain the use of these options for bare-metal users. We can then state that Operator deployments configure the machine id automatically and maybe even state why we don't set the site so that users are aware of this limitation?

ahus1:
> IMO, it has little to no value for kubernetes users as the operator takes care of it. Bare metal users will benefit the more from this as they know how the physical machine are located in the datacenter.

I agree. If you think there is no good place to put the docs, I'm ok to approve it without the docs until we have a better guide for a bare metal / VM setup. Still, that "concepts" page should be pretty agnostic. While it already contains some technical Kubernetes related stuff, adding more technical details makes it worse IMHO.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
pruivo: The second argument is void, as the IP address is included in the hostname of the worker node. 
```
* Format for an instance in us-east-1: private-ipv4-address.ec2.internal
Example: ip-10-24-34-0.ec2.internal
* Format for an instance in any other AWS Region: private-ipv4-address.region.compute.internal
Example: ip-10-24-34-0.us-west-2.compute.internal
```
Honestly, I don't see any benefit for debugging either, but I'll change to the node name.

ryanemerson: Ok I didn't realise that, but I guess that's implementation independent. I don't have strong views on this either way in that case.

pruivo:
@ryanemerson @ahus1 I prefer to place the docs in the caching section (https://www.keycloak.org/server/caching) and add a limitation to https://www.keycloak.org/nightly/high-availability/single-cluster/concepts, stating that the operator configures the machine ID but not the site ID (and why).

ahus1:
@pruivo - I'm fine with that suggestion.

docs/guides/server/caching.adoc
```diff
@@ -203,6 +203,55 @@ For these use case, the CLI options `cache-remote-tls-enabled` disables the encr
 The CLI options `cache-remote-username` and `cache-remote-password` are optional and, if not set, {project_name} will connect to the {jdgserver_name} server without presenting any credentials.
 If the {jdgserver_name} server has authentication enabled, {project_name} will fail to start.
 
+[[cache-topology]]
+== Topology Aware Data Availability
```
ryanemerson: We usually only capitalize the first word in titles.

docs/guides/server/caching.adoc
```diff
@@ -203,6 +203,55 @@ For these use case, the CLI options `cache-remote-tls-enabled` disables the encr
...
+
+Configuring the topology information of your cluster increases the data availability in case of hardware failure.
+The hardware where the {project_name} instance runs is necessary to compute this topology information.
```
ryanemerson: "Configuring {project_name} to be aware of your network topology, increases data availability in the presence of hardware failures, as Infinispan is able to ensure that data is distributed correctly. For example, if `num_owners=2`, it will ensure that the two owners are not stored on the same node when possible."

docs/guides/server/caching.adoc
```diff
@@ -203,6 +203,55 @@ For these use case, the CLI options `cache-remote-tls-enabled` disables the encr
...
++
+Use the SPI option `spi-cache-embedded--default--machine-name` (or environment variable `KC_SPI_CACHE_EMBEDDED\__DEFAULT__MACHINE_NAME`).
+The value itself is not important, but each machine must have a unique value.
++
+For example: `--spi-cache-embedded--default--machine-name=machine-1`
++
+[NOTE]
+====
+The {project_name} Operator automatically configure the machine name based on the Kubernetes node.
+It ensures if Pods are scheduled in the same node, the data is safely replicated into another node.
```
ryanemerson: "It ensure that if multiple pods are scheduled on the same node, data replicas are still replicated across distinct nodes when possible".

docs/guides/server/caching.adoc
```diff
@@ -203,6 +203,55 @@ For these use case, the CLI options `cache-remote-tls-enabled` disables the encr
...
+We recommend to set up anti-affinity rules and/or topology spread constraints to prevent multiple Pods from being scheduled on the same node, further reducing the risk of a single node failure incur a data loss.
```
ryanemerson: I don't think incur is correct here, how about ".. causing data loss."

docs/guides/server/caching.adoc
```diff
@@ -203,6 +203,55 @@ For these use case, the CLI options `cache-remote-tls-enabled` disables the encr
 The CLI options `cache-remote-username` and `cache-remote-password` are optional and, if not set, {project_name} will connect to the {jdgserver_name} server without presenting any credentials.
 If the {jdgserver_name} server has authentication enabled, {project_name} will fail to start.
 
+[[cache-topology]]
+== Topology Aware Data Availability
+
+Configuring the topology information of your cluster increases the data availability in case of hardware failure.
+The hardware where the {project_name} instance runs is necessary to compute this topology information.
+This section is more relevant for bare-metal users, who have control over the hardware and where it is running.
```
ryanemerson: While mostly true, I think we can remove this sentence as the machine name is still relevant to k8s.

docs/guides/server/caching.adoc
ryanemerson: Is "Topology aware data distribution" more accurate?

pruivo:
Thank you, @ryanemerson. Updated!

ryanemerson (approved):
LGTM, thanks @pruivo

ahus1 (approved):

PR #42146, opened 2025-08-26
fix: adding better management of closed entitymanagers

shawkins:
closes: #42114

@vellotis I'd like to propose a change so that callers don't have to differentiate what type of EntityManager they need.

Then perhaps a simple unit test will suffice based upon your comments in the other pr.

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/connections/jpa/DefaultJpaConnectionProviderFactory.java
+++ b/model/jpa/src/main/java/org/keycloak/connections/jpa/DefaultJpaConnectionProviderFactory.java
@@ -89,10 +89,10 @@ public class DefaultJpaConnectionProviderFactory implements JpaConnectionProvide
         logger.trace("Create JpaConnectionProvider");
         lazyInit(session);
 
-        return new DefaultJpaConnectionProvider(createEntityManager(session));
+        return new DefaultJpaConnectionProvider(createEntityManager(session, true));
     }
 
-    private EntityManager createEntityManager(KeycloakSession session) {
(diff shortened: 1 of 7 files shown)


vellotis:
I see only one downside with this approach: we deliberately create a precedent by allowing polluting session with faulty entities.

Yes, it is failsafe. But don't we shoot ourselves to the foot with this?

shawkins:
> Yes, it is failsafe. But don't we shoot ourselves to the foot with this?

My view is that EntityManagerProxy.create is creating managed EntityManagers - such that they will transparently work with the the nestedEntityManagers logic or EntityManagers.flush - which is independent of nestedEntityManagers. I would prefer if this could be handled with the Provider logic, but unfortunately that's a much bigger change. 

Putting the distinction on the callers of the create method (currently DefaultJpaConnectionProviderFactory and QuarkusJpaConnectionProviderFactory), as to whether the truly want a session managed entitymanager can be done as well - so let's combine the two prs. But I'd want the other one expanded to apply to both usages of EntityManagerProxy.create and for the parameter name to be something like managed, rather than refer to nestedEntityManagers. Do you want me to incorporate your changes here, or do everything on the other pr?

shawkins:
> Do you want me to incorporate your changes here, or do everything on the other pr?

@vellotis see if you are ok with these changes?

vellotis:
@shawkins Sorry for the delay. I'm good with the latest changes. 👍

ahus1 (approved):
Approving after my review and the community review.

Issue #41594, opened 2025-07-31
Update the credential_request_encryption and credential_response_encryption.


Ogenbertrand:
### Description

<img width="817" height="455" alt="Image" src="https://github.com/user-attachments/assets/e242d4ae-15dd-4d69-9a81-98105ba9521e" />

This ticket is going to include the following features at once:

- Updating key selection logic to choose public keys based on JWK parameters (kty, use, alg, etc.), ensuring the JWE alg matches the JWK’s alg and including the kid in the JWE header if present in the JWK.

- Support the zip parameter for compression before encryption (per RFC7516) for both Credential Requests and Responses, with no compression if zip is absent.

- Update the wallet to recommend (but not mandate) rejection of unencrypted messages when encryption is required (e.g., when encryption_required is true in credential_request_encryption or credential_response_encryption metadata).

Ogenbertrand:
I'll take this one

Issue #41595, opened 2025-07-31
If encryption requested, encrypt all kinds of responses, except error credential responses


Ogenbertrand:
### Description

See: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-encrypted-messages

forkimenjeckayang:
Working on this

Issue #41592, opened 2025-07-31
Enhance credential_response_encryption parsing to include zip_values_supported for compression support.


Ogenbertrand:
### Description

See: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-credential-issuer-metadata-p

Ogenbertrand:
Working on this

Issue #41582, opened 2025-07-31
Add zip parameter for compression in credential_response_encryption, remove alg.


Ogenbertrand:
### Description

Related Chnagelog entry:

The Alg parameter was remove from the credential encryption and zip added for compression:

> credential_response_encryption: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.

> jwk: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
> enc: REQUIRED. JWE [[RFC7516](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#RFC7516)] enc algorithm [[RFC7518](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#RFC7518)] for encrypting Credential Responses.
> zip: OPTIONAL. JWE [[RFC7516](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#RFC7516)] zip algorithm [[RFC7518](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#RFC7518)] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.

See: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-credential-request

Ogenbertrand:
I'll take this one

PR #42169, opened 2025-08-27
[OID4VCI] Add support for credential_request_encryption in metadat

Ogenbertrand:
This PR aims at:
- Adding zip parameter for compression in credential_response_encryption and removing alg.

- Updating key selection logic to choose public keys based on JWK parameters (kty, use, alg, etc.), ensuring the JWE alg matches the JWK’s alg and including the kid in the JWE header if present in the JWK.

- Support the zip parameter for compression before encryption (per RFC7516) for both Credential Requests and Responses, with no compression if zip is absent.

- Update the wallet to recommend (but not mandate) rejection of unencrypted messages when encryption is required (e.g., when encryption_required is true in credential_request_encryption or credential_response_encryption metadata).

-  Update media type handling to use application/json only for unencrypted requests and support JWT for encrypted requests.

See links:
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-encrypted-messages
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-credential-issuer-metadata-p

Closes: 
https://github.com/keycloak/keycloak/issues/41594
https://github.com/keycloak/keycloak/issues/41593
https://github.com/keycloak/keycloak/issues/41592
https://github.com/keycloak/keycloak/issues/41582
https://github.com/keycloak/keycloak/issues/41595


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -397,46 +402,110 @@ private void checkScope(CredentialScopeModel requestedCredential) {
...
+            }
+        }
+
+        // Authenticate after parsing to avoid processing invalid requests
         AuthenticationManager.AuthResult authResult = getAuthResult();
 
         // Validate encryption parameters if present
         CredentialResponseEncryption encryptionParams = credentialRequestVO.getCredentialResponseEncryption();
         CredentialResponseEncryptionMetadata encryptionMetadata = OID4VCIssuerWellKnownProvider.getCredentialResponseEncryption(session);
-        boolean isEncryptionRequired = Optional.ofNullable(encryptionMetadata)
```
wistefan: Why cant this be null anymore?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -496,6 +564,143 @@ public Response requestCredential(CredentialRequest credentialRequestVO) {
...
+     * Decrypts a JWE-encoded Credential Request and validates it against metadata.
+     *
+     * @param jweString The JWE compact serialization
+     * @param metadata The CredentialRequestEncryptionMetadata
+     * @return The parsed CredentialRequest
+     * @throws JWEException If decryption or validation fails
+     */
+    private CredentialRequest decryptCredentialRequest(String jweString, CredentialRequestEncryptionMetadata metadata) throws Exception {
+        JWE jwe = new JWE(jweString);
+        JWEHeader header = (JWEHeader) jwe.getHeader();
```
wistefan: Would be better to check the type before casting.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -496,6 +564,143 @@ public Response requestCredential(CredentialRequest credentialRequestVO) {
...
+                }
+                return out.toByteArray();
+            } catch (Exception e) {
+                throw new JWEException("Failed to decompress: " + e.getMessage());
+            }
+        }
+        throw new JWEException("Unsupported compression algorithm");
+    }
+
+    private boolean looksLikeCompactJwe(String payload) {
```
wistefan: Why do you need such check? The core-module contains a JWE-class, that f.e. provides verifyAndDecodeJWE(String jweStr)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -496,6 +564,143 @@ public Response requestCredential(CredentialRequest credentialRequestVO) {
...
+            LOGGER.debugf("JWK is missing required 'alg' parameter for key type: %s", jwk.getKeyType());
+            return null;
+        }
+
+        // Verify the alg is supported by the server
+        if (supportedAlgs.contains(jwkAlg)) {
+            return jwkAlg;
+        }
+
+        // If the JWK's alg is not supported, we cannot proceed
```
wistefan, line 707: It would be better to throw an  IllegalArgument then, instead of returning null(which is always a risk of NPE)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -496,6 +564,143 @@ public Response requestCredential(CredentialRequest credentialRequestVO) {
...
+    /**
+     * Decompresses content using the specified algorithm.
+     *
+     * @param content The compressed content
+     * @param zipAlgorithm The compression algorithm (e.g., "DEF")
+     * @return The decompressed content
+     * @throws JWEException If decompression fails
+     */
+    // TODO handle compression/decompression transparently at the JWE software layer.
+    private byte[] decompress(byte[] content, String zipAlgorithm) throws JWEException {
```
wistefan, line 677: The saml module contains DeflateUtil. Maybe it makes more sense to put this into something like core and reuse the funtionality, instead of duplicating? @mposolda ?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -397,46 +402,110 @@ private void checkScope(CredentialScopeModel requestedCredential) {
...
+        if (requestPayload == null || requestPayload.trim().isEmpty()) {
+            String errorMessage = "Request payload is null or empty.";
+            LOGGER.debug(errorMessage);
+            throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST, errorMessage));
+        }
 
         cors = Cors.builder().auth().allowedMethods("POST").auth().exposedHeaders(Cors.ACCESS_CONTROL_ALLOW_METHODS);
 
-        // Authenticate first to fail fast on auth errors
+        CredentialIssuer issuerMetadata = (CredentialIssuer) new OID4VCIssuerWellKnownProvider(session).getConfig();
```
wistefan, line 421: To improve readability, the encryption checking could put into a method

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -528,37 +732,72 @@ private String encryptCredentialResponse(CredentialResponse response, Credential
                     "Invalid JWK: Failed to parse public key."));
         }
 
+        // Retrieve metadata
+        CredentialResponseEncryptionMetadata metadata = OID4VCIssuerWellKnownProvider.getCredentialResponseEncryption(session);
```
wistefan: The metadata could just be provided as a method-argument, since its already retrieved in l469, before the method is called

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -528,37 +732,72 @@ private String encryptCredentialResponse(CredentialResponse response, Credential
...
                             .build());
         }
     }
 
+    /**
+     * Compress content using the specified algorithm
+     */
+    // TODO handle compression/decompression transparently at the JWE software layer.
+    private byte[] compressContent(byte[] content, String zipAlgorithm) throws IOException {
+        if ("DEF".equals(zipAlgorithm)) {
```
wistefan: "DEF" should be a constant

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -496,6 +564,143 @@ public Response requestCredential(CredentialRequest credentialRequestVO) {
...
+        // Decrypt the JWE
+        try {
+            jwe.verifyAndDecodeJwe();
+        } catch (JWEException e) {
+            throw new JWEException("Failed to decrypt JWE: " + e.getMessage());
+        }
+
+        // Handle decompression if zip=DEF
+        byte[] content = jwe.getContent();
+        if (zip != null && "DEF".equals(zip)) {
```
wistefan: This is a superfluous double-check. The decompress method also checks zip again.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -192,6 +209,37 @@ public static List<String> getSupportedEncryptionAlgorithms(KeycloakSession sess
         return supportedEncryptionAlgorithms;
     }
 
+
+    /**
+     * Builds JWKS from realm encryption keys with use=enc.
+     */
+    private static JSONWebKeySet buildJwks(KeycloakSession session) {
```
wistefan, line 396: It would be good to also have the ability to configure the keys(or key-types) to be included. Practically, only asymmetric keys can be used for the encryption, thus some wallets(for example github.com/eu-digital-identity-wallet/eudi-app-android-wallet-ui) consider symmetric keys as part of the meta-data as invalid and abort the issuance process.

wistefan (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -78,7 +82,7 @@ public void close() {
     @Override
     public Object getConfig() {
         KeycloakContext context = keycloakSession.getContext();
-        CredentialIssuer issuer = new CredentialIssuer()
+        return new CredentialIssuer()
```
babisRoutis, line 132: There is a relationship between `credential_response_encryption` & `credential_request_encryption` that I am not sure if it is being checked (I have the impression that currently response and requested encryption options appear as independent).


The relationship is the following :
If Issuer advertises `credential_response_encryption` (required or optional) the wallet MUST encrypt the request (to the credential or deferred endpoint).
The reference can be found [here](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2-18) and [here](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-9.1-11)

Returning to the issuer metadata, to my understanding, the above requirement means that : 

- If there is `credential_response_encryption` there must be a `credential_request_encryption`, in addition
- If `credential_response_encryption` is required, then the  `credential_request_encryption` must be required as well.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Ogenbertrand: Fixed

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Ogenbertrand: I imported directly the DeflateUtil from the saml module while waiting for @mposolda response.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
Ogenbertrand: The spec does not mandate that **credential_response_encryption** requires the presence of **credential_request_encryption** in the metadata, nor does it enforce that **credential_request_encryption.encryption_required** must match or exceed **credential_response_encryption.encryption_required**. Both are independently optional, and their sub-parameters (e.g., alg_values_supported, enc_values_supported, encryption_required) are only REQUIRED if the parent object is advertised ([OID4VCI Spec Section 11.2.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-credential-issuer-metadata-p)). Perhaps I didn't clearly get the point. Correct me if I'm wrong @babisRoutis!

babisRoutis: Hi @Ogenbertrand 

It is true that to the metadata section there is no  association between request/response encryption.

Yet, please check the following text (from the spec) :

> The Credential Issuer indicates support for encrypted responses by including the credential_response_encryption parameter in the Credential Issuer Metadata. The Client MAY request encrypted responses by providing its encryption parameters in the Credential Request when encryption_required is false and MUST do so when encryption_required is true. Credential Request encryption MUST be used if the credential_response_encryption parameter is included, to prevent it being substituted by an attacker

To my understanding, the requirement of the last sentence means that  if Issuer provides the option (required or not)  for response encryption, it MUST also provide the option to encrypt the request. (the opposite it is not true)

Otherwise, let's say there is `credential_response_encryption` but not `credential_request_encryption` then the requirement "Credential Request encryption MUST be used if the credential_response_encryption parameter is included" cannot be met.

Ogenbertrand: That was a wonderful catch @babisRoutis, thanks for pointing out that, i'll immediately address the point.

Ogenbertrand: Hi @babisRoutis !
Can you check the logic again ?

babisRoutis: @Ogenbertrand 

The condition is correct. Setting to `null` to response encryption is certainly a reasonable option. If you prefer a not so strict approach you could even just output the warning.

I think though that there is also the following check:
In case both request & response encryption are supported and if response encryption is required, then request encryption should be marked as required, as well. 

You may at least output a warning or marking response encryption `required` as false.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
wistefan: Shouldn't all of the "null" cases be IllegalArgumentExceptions?

wistefan (approved):
One minor thing left, otherwise LGTM

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
Ogenbertrand: @babisRoutis !
you can check again.

babisRoutis (approved):
LGTM

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
babisRoutis: Dear @Ogenbertrand 

Thank you for taking into account my comments

mposolda (approved):
@Ogenbertrand @wistefan @babisRoutis Thanks everyone for the participate on this PR!

Issue #40665, opened 2025-06-23
KeycloakServerConfigBuilder - add -D property values into the startup


lhanusov:
### Description

Example:

```
public KeycloakServerConfigBuilder properties(Map<String, String> properties) {
        this.properties.putAll(properties);
        return this;
}

public KeycloakServerConfigBuilder property(String key, String value) {
        this.properties.put(key, value);
        return this;
}

properties.forEach((key, value) -> args.add("-D" + key + "=" + value));
```

lhanusov:
context provided by @ryanemerson 

`For context, I added this so that I could debug http requests received by the server using -Dquarkus.http.access-log.enabled=true. Without these changes there was no way (AFAICT) to pass the property to the server(s).`

stianst:
This would work okay for the distribution mode as we are starting a separate VM, but for embedded this is not a great idea as it's setting system properties for the running VM, which can then cause issues if running multiple differently configured embedded servers in a row, etc.

Closing this for now, especially since there is now a proper server config option to enable http logging. If there are more use-cases we can revisit, but system properties are probably not the way to go.

Issue #32129, opened 2024-08-14
Automatically create external caches for MULTI_SITE deployments


ryanemerson:
### Description

When using the MULTI_SITE feature, the default behaviour should be for caches to be created automatically on the External Infinispan cluster.

### Motivation

Currently it's necessary for users to define all of the Keycloak caches explicitly on their external infinispan cluster when using the MULTI_SITE feature. This allows users to configure the cache definitions for more advanced use-cases, however for the vast majority of users a default cache configuration is sufficient.

To reduce the number of steps required for a user to deploy multi-site Keycloak, we should create the remote caches automatically, with users able to opt-out of this behaviour by providing a startup argument.

### Details

It's already possible for remote caches to be created automatically, however this requires users to provide the `kc.cache-remote-create-caches` property and results in the falling warning:

```
Creating remote cache in external Infinispan server. It should not be used in production!
```

We can remove the above warning and re-purpose the property so that caches are not created automatically only when `kc.cache-remote-create-caches` is explicitly provided as `false`.

stianst:
We should not rely on `kc.cache-remote-create-caches` system property. This should rather be either a SPI option, or a proper CLI option (cache-create-caches). It can default to true or false, whatever makes sense.

PR #42172, opened 2025-08-27
[Keycloak Test Framework] Infinispan cache + ClusterlessTestSuite configuration

lhanusov:
Part of #35391


stianst (commented):
I would rename it from InfinispanCache to InfinispanServer, as it's not creating single caches, but rather full Infinispan servers

test-framework/core/src/main/java/org/keycloak/testframework/cache/InfinispanLocalCache.java
```diff
@@ -0,0 +1,20 @@
+package org.keycloak.testframework.cache;
+
+public class InfinispanLocalCache implements InfinispanCache{
```
ryanemerson: `ClusteredKeycloakServerSupplier` uses cache type "ispn". Should we also created a `InfinispanCache` implementation for that?

The impl would be basically the same as `InfinispanLocalCache` but with `getCacheName` changed. Can we add no-op default methods to the `InfinispanCache` interface.

pruivo: There are 4 modes, actually. 
* embedded & local
* embedded & clustered
* external & multi-site
* external & clusterless

@lhanusov I think it would make more sense if the supplier receives the `KeycloakServerConfigBuilder`. The external need to configure `cache-remote-*` CLI options and enable features `multi-site` or `clusterless`.

lhanusov: thank you both, this will help me a lot :-)

ryanemerson:
The issue with ports was due to us using the default Hot Rod ClientIntelligence policy `HASH_DISTRIBUTION_AWARE`. This makes it so that the client accesses Infinispan server  IPs directly after the initial handshake has occurred, so that the client can update entries in the most efficient manner possible. However, in this context the internal IP of the Infinispan server is not accessible from the localhost, causing the client to fail when attempting to access said IP. The workaround is to set the policy to `BASIC`, which instructs the client to always connect to the server(s) using the explicitly configured IPs in the `RemoteCacheManager` configuration.

I have also refactored `InfinispanExternalServer` to remove the use of inheritance and a nested container in favour of just using inheritance directly.

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
```diff
@@ -48,8 +50,29 @@ public KeycloakServerConfigBuilder bootstrapAdminUser(String username, String pa
                 .option("bootstrap-admin-password", password);
     }
 
-    public KeycloakServerConfigBuilder cache(String cache) {
-        return option("cache", cache);
+    public KeycloakServerConfigBuilder setCache(CacheType cacheType) {
```
stianst:
```suggestion
    public KeycloakServerConfigBuilder cache(CacheType cacheType) {
```

test-framework/core/src/main/java/org/keycloak/testframework/annotations/InjectInfinispanServer.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface InjectInfinispanServer {
+
+    LifeCycle lifecycle() default LifeCycle.CLASS;
```
stianst: Don't think we want this to be class, as that means we'll start a new Infinispan server for each test class. I'd make it global by default

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
```diff
@@ -190,7 +213,11 @@ List<String> toArgs() {
         List<String> args = new LinkedList<>();
         args.add(command);
         for (Map.Entry<String, String> e : options.entrySet()) {
-            args.add("--" + e.getKey() + "=" + e.getValue());
+            if (e.getKey().startsWith("-D")) {
```
stianst, line 214: What's this change for?

tests/clustering/src/test/java/org/keycloak/tests/clustering/JdbcPingCustomSchemaTest.java
```diff
@@ -16,4 +18,14 @@ public class JdbcPingCustomSchemaTest {
     public void testClusterFormed() {
         // no-op ClusteredKeycloakServer will fail if a cluster is not formed
     }
+
+    public static class JdbcPingKeycloakServerConfig extends CaseSensitiveSchemaTest.CaseSensitiveServerConfig {
```
stianst: This shouldn't be needed, just set the cache type in ClusteredKeycloakServer

tests/clustering/src/test/java/org/keycloak/tests/compatibility/ClusteredOAuthClientTest.java
```diff
@@ -89,4 +92,14 @@ public UserConfigBuilder configure(UserConfigBuilder user) {
                   .password("password");
         }
     }
+
+    public static class ClusteredKeycloakServerConfig implements KeycloakServerConfig {
```
stianst: Get rid of this, set it in ClusteredKeycloakServer

test-framework/core/src/main/java/org/keycloak/testframework/database/DatabaseProperties.java
stianst, line 1: I'd suggest changing this to ContainerImages and use containers.properties; and have something like:

```
getContainerImageName(String type, String name);
```

stianst (commented):
Looks pretty good, some minor comments inline. I haven't gone in detail on reviewing the Infinispan server and how it's configured.

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
```diff
@@ -48,8 +50,29 @@ public KeycloakServerConfigBuilder bootstrapAdminUser(String username, String pa
...
+        this.cacheType = cacheType;
+        return this;
+    }
+
+    public KeycloakServerConfigBuilder cache() {
+        option("cache", cacheType.name().toLowerCase());
+        return this;
+    }
+
+    public KeycloakServerConfigBuilder enableExternalInfinispan(boolean enabled) {
```
vaceksimon: nit: to follow the convention we set in [other config builders](https://github.com/keycloak/keycloak/blob/d98c474cdc68be28f67e8153c746bfb2d82f800a/test-framework/core/src/main/java/org/keycloak/testframework/realm/RealmConfigBuilder.java#L97-L115)
```suggestion
    public KeycloakServerConfigBuilder externalInfinispanEnabled(boolean enabled) {
```

vaceksimon (approved):
one minor nitpick, overall lgtm

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
ryanemerson: I believe this is required for setting `-Dkc.cache-remote-create-caches`. We don't have an SPI or CLI option for this currently. There's #32129 which should be extended to also include the CLUSTERLESS Feature \cc @pruivo.

In the meantime should we reconsider #40665 and make properties a first class part of the testsuite?

test-framework/core/src/main/java/org/keycloak/testframework/infinispan/InfinispanExternalServer.java
```diff
@@ -0,0 +1,38 @@
...
+    public static InfinispanExternalServer create() {
+        return new InfinispanExternalServer(InfinispanProperties.getContainerImageName());
+    }
+
+    @SuppressWarnings("resource")
+    private InfinispanExternalServer(String dockerImageName) {
+        super(dockerImageName);
+        withUser(USER);
+        withPassword(PASSWORD);
+        addFixedExposedPort(11222, 11222);
```
ryanemerson, line 26: Should we define a log consumer via `withLogConsumer` so we can control log levels via `managed.infinispan` in `keycloak-test.properties`?

test-framework/core/src/main/java/org/keycloak/testframework/infinispan/InfinispanExternalServerSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+public class InfinispanExternalServerSupplier implements Supplier<InfinispanServer, InjectInfinispanServer>, KeycloakServerConfigInterceptor<InfinispanServer, InjectInfinispanServer> {
+
+    private static final Logger LOGGER = Logger.getLogger(InfinispanExternalServerSupplier.class);
+
+    @Override
+    public InfinispanServer getValue(InstanceContext<InfinispanServer, InjectInfinispanServer> instanceContext) {
+        InfinispanServer server = InfinispanExternalServer.create();
+        getLogger().info("Starting Infinispan Server");
+
+        long start = System.currentTimeMillis();
```
ryanemerson, line 21: The Infinispan startup time is logged in the server's info logs, so if we add a log consumer for `managed.infinispan` this should not be necessary. Currently these logs are not displayed anyway as an entry is not present in `keycloak-test.properties`.

test-framework/core/src/main/java/org/keycloak/testframework/annotations/InjectInfinispanServer.java
lhanusov: ok, updated

test-framework/core/src/main/java/org/keycloak/testframework/infinispan/InfinispanExternalServer.java
lhanusov: I agree it's a good idea to add the Log consumer.

test-framework/core/src/main/java/org/keycloak/testframework/infinispan/InfinispanExternalServerSupplier.java
lhanusov: This is about the supplier logging, not the Infinispan server.

test-framework/core/src/main/java/org/keycloak/testframework/database/DatabaseProperties.java
lhanusov: ok, I've refactored it more and combined them with the ispn container

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
```diff
@@ -48,8 +50,29 @@ public KeycloakServerConfigBuilder bootstrapAdminUser(String username, String pa
...
 
-    public KeycloakServerConfigBuilder cache(String cache) {
-        return option("cache", cache);
+    public KeycloakServerConfigBuilder cache(CacheType cacheType) {
+        this.cacheType = cacheType;
+        return this;
+    }
+
+    public KeycloakServerConfigBuilder cache() {
+        option("cache", cacheType.name().toLowerCase());
```
ryanemerson: Isn't the `cache` option always required? I'm just wondering if we can move the logic of this method into the `build` method so that we don't have to explicitly call this method in each of the server types?

We would then just need to call `configBuilder.cache(CacheType.ISPN);` where required.

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
lhanusov: updated

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
lhanusov: @stianst yes, it's about supporting the auto create of the cache

tests/clustering/src/test/java/org/keycloak/tests/clustering/JdbcPingCustomSchemaTest.java
lhanusov: updated

tests/clustering/src/test/java/org/keycloak/tests/compatibility/ClusteredOAuthClientTest.java
lhanusov: updated

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
lhanusov: Very good point, thanks, I added this change into toArgs() method only.

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
lhanusov: updated

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
lhanusov: updated

ryanemerson (approved):

stianst (approved):

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
stianst: We can handle this as a follow-up in https://github.com/keycloak/keycloak/issues/42619

PR #42174, opened 2025-08-27
fix for issue 41125

### LINKED ISSUE #41125: Authorization Policy Import allows non-JSON files to be uploaded in Admin UI
### Before reporting an issue

### Area

admin/ui

### Describe the bug

In the **Authorization** tab of a client in the Keycloak Admin UI, when importing an authorization policy via the `Import` button, the upload dialog allows files with any extension to be selected and uploaded.

This contradicts the UI hint that only `.json` files should be accepted. There is no frontend or backend validation to prevent users from uploading unsupported file types. Besides, when an irrelevant file type is uploaded, all of the Authorizations are removed.

Image

### Version

26.2.5

### Regression

### Expected behavior

The file input should restrict selection to only `.json` files, both in:
- the UI (via `accept=".json"` or similar attribute)
- client-side validation before attempting upload
- server-side validation after uploading a file.

### Actual behavior

Any file can be selected and uploaded, even with unsupported formats.

### How to Reproduce?

1. Go to the Admin UI.
2. Navigate to a realm → Clients → Select a client.
3. Click the **Authorization** tab.
4. Click the **Import** button.
5. Upload a file with an invalid extension (e.g., `.txt`, `.pdf`, `.png`).

### Anything else?

This may lead to confusion or unhandled errors if non-JSON files are uploaded accidentally.

himanshi1099:
closes #41125

### DIFF
--- a/js/apps/admin-ui/src/components/json-file-upload/FileUploadForm.tsx
+++ b/js/apps/admin-ui/src/components/json-file-upload/FileUploadForm.tsx
@@ -46,7 +46,6 @@ export const FileUploadForm = ({
   helpText = "helpFileUpload",
   unWrap = false,
   language,
-  extension,
   ...rest
 }: FileUploadFormProps) => {
   const { t } = useTranslation();
@@ -82,7 +81,7 @@ export const FileUploadForm = ({
           onClose={removeDialog}
           actions={[
             <Button
-              key="confirm"
+              key="confirm" 
               variant="primary"
               data-testid="clear-button"
               onClick={() => {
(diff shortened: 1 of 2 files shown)


ssilvert (changes requested):
Since they are exactly the same, you can declare a constant for dropzoneProps instead of repeating it in both places.

`const dropzoneProps = {
              accept: { "application/json": [".json"] },
            }`
...

`dropzoneProps={dropzoneProps}`

ssilvert:
@himanshi1099 Oh, I think the problem is that this is using the wrong component.  I think it should be using `JsonFileUpload` instead of `FileUploadForm`?

himanshi1099:
@ssilvert sure will take a look and will make the relevant changes

himanshi1099:
@ssilvert I tried changing component and tried using JsonFileUpload instead of FileUploadForm but getting circular import issue

PR #42179, opened 2025-08-27
Show length validations in the admin UI

### LINKED ISSUE #42178: Integer validation error not shown for user profile fields
### Before reporting an issue

### Area

user-profile

### Describe the bug

When I create a field in the user profile and add a length min/max validation to it, the error message doesn't show in the admin UI, the user is not saved, and an exception is shown in the console.  

### Version

main

### Regression

### Expected behavior

The validation error should be shown.

### Actual behavior

No validation error is shown

### How to Reproduce?

Create a new field and add a length validation to it

Image

Image

### Anything else?

The problem is that the length validation error adds integer parameters, while front end only expects strings.  

I'll provide a fix.

ahus1:
Closes #42178

### DIFF
--- a/js/libs/ui-shared/src/user-profile/utils.ts
+++ b/js/libs/ui-shared/src/user-profile/utils.ts
@@ -77,7 +77,7 @@ export function setUserProfileServerError<T>(
   ).forEach((e) => {
     const params = Object.assign(
       {},
-      e.params?.map((p) => (isBundleKey(p) ? t(unWrap(p)) : p)),
+      e.params?.map((p) => (isBundleKey(p) ? t(unWrap(p as string)) : p)),
     );
     setError(fieldName(e.field) as keyof T, {
       message: t(
--- a/js/libs/ui-shared/src/user-profile/utils.ts
+++ b/js/libs/ui-shared/src/user-profile/utils.ts
@@ -15,7 +15,7 @@ export type UserFormFields = Omit<
 type FieldError = {
   field: string;
   errorMessage: string;
-  params?: string[];
+  params?: any[];
 };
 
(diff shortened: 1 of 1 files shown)


ahus1:
@keycloak/ui - happy to get your feedback on this change, and if additional tests would be needed. Thanks!

js/libs/ui-shared/src/user-profile/utils.ts
```diff
@@ -15,7 +15,7 @@ export type UserFormFields = Omit<
 type FieldError = {
   field: string;
   errorMessage: string;
-  params?: string[];
+  params?: any[];
```
jonkoops: Instead of `any`, a better type would be `unknown`, which will force the compiler to error unless the type of the value is checked at runtime (e.g. `typeof params === 'string'`).
```suggestion
  params?: unknown[];
```

js/libs/ui-shared/src/user-profile/utils.ts
```diff
@@ -24,7 +24,11 @@ export type UserProfileError = {
   responseData: ErrorArray | FieldError;
 };
 
-const isBundleKey = (displayName?: string) => displayName?.includes("${");
+const isBundleKey = (displayName?: any) => {
```
jonkoops: Same here, can be `unknown`, you might even drop the optionality here, as `null` and `undefined` should also be covered by the `unknown` type.
```suggestion
const isBundleKey = (displayName?: unknown) => {
```

Or possibly:
```suggestion
const isBundleKey = (displayName: unknown) => {
```

ahus1:
@jonkoops - please re-review. Thanks!

jonkoops (approved):

PR #42181, opened 2025-08-27
Cache UserAgent parsing result

### LINKED ISSUE #42180: Cache UserAgent parsing result
### Description

I'm running `AsyncProfiler` in a single Keycloak instance without password-hashing (iterations=1 to avoid bloating the flamegraphs) to check the CPU hot spots, and the most obvious one is the `user-agent` string parser. 

With HTTP enabled, it happens in all the requests (secure context checks) and takes 10% of the CPU time. See flamegraph below, purple stacks.

Image

With HTTPS, it is not as bad, and it is only used during login, bringing the CPU time to 10%.

Image

In any case, we could have a small cache for those strings. 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pruivo:
Closes #42180

Flamegraph with this change:
Screenshot 2025-08-27 at 14-50-45

### DIFF
--- a/services/pom.xml
+++ b/services/pom.xml
@@ -255,6 +255,10 @@
             <artifactId>uap-java</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.microprofile.openapi</groupId>
             <artifactId>microprofile-openapi-api</artifactId>
         </dependency>
--- a/services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
+++ b/services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
@@ -4,10 +4,7 @@ import java.util.List;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
-import io.micrometer.core.instrument.Metrics;
-import io.micrometer.core.instrument.binder.cache.CaffeineStatsCounter;
 import org.keycloak.Config;
-import org.keycloak.config.MetricsOptions;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.provider.ProviderConfigurationBuilder;
@@ -31,21 +28,12 @@ public class DeviceRepresentationProviderFactoryImpl implements DeviceRepresenta
         return PROVIDER_ID;
     }
 
-    @SuppressWarnings("NullableProblems")
     @Override
     public void init(Config.Scope config) {
-        Caffeine<Object, Object> builder = Caffeine.newBuilder()
+        cache = Caffeine.newBuilder()
                 .maximumSize(config.getInt(CACHE_SIZE, DEFAULT_CACHE_SIZE))
-                .softValues();
-
-        if (config.root().getBoolean(MetricsOptions.METRICS_ENABLED.getKey(), Boolean.FALSE)) {
-            var stats = new CaffeineStatsCounter(Metrics.globalRegistry, "user-agent");
-            builder.recordStats(() -> stats);
-            cache = builder.build(UA_PARSER::parse);
-            stats.registerSizeMetric(cache);
-        } else {
-            cache = builder.build(UA_PARSER::parse);
-        }
+                .softValues()
+                .build(UA_PARSER::parse);
     }
 
     @Override
--- a/services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
+++ b/services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
@@ -47,7 +47,7 @@ public class DeviceRepresentationProviderFactoryImpl implements DeviceRepresenta
                 .property()
                 .name(CACHE_SIZE)
                 .type(ProviderConfigProperty.INTEGER_TYPE)
-                .helpText("Sets the maximum number of cached parsed user-agent value.")
+                .helpText("Sets the maximum number of parsed user-agent values in the local cache.")
                 .defaultValue(DEFAULT_CACHE_SIZE)
                 .add()
                 .build();
--- a/services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
+++ b/services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
@@ -4,7 +4,10 @@ import java.util.List;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.binder.cache.CaffeineStatsCounter;
 import org.keycloak.Config;
+import org.keycloak.config.MetricsOptions;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.provider.ProviderConfigurationBuilder;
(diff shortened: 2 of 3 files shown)


services/src/main/java/org/keycloak/device/DeviceRepresentationProviderImpl.java
```diff
@@ -48,7 +41,10 @@ public DeviceRepresentation deviceRepresentation() {
 
         DeviceRepresentation current;
         try {
-            Client client = parser.parse(userAgent);
+            Client client = cache.get(userAgent);
+            if (client == null) {
```
ryanemerson: Can this ever happen, on a cache miss doesn't the `CacheLoader` ensure that a non-null Client is always returned?

services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
```diff
@@ -14,17 +28,28 @@ public String getId() {
...
-                parser = new Parser();
-            }
-        }
+    @Override
+    public List<ProviderConfigProperty> getConfigMetadata() {
+        return ProviderConfigurationBuilder.create()
+                .property()
+                .name(CACHE_SIZE)
+                .type(ProviderConfigProperty.INTEGER_TYPE)
+                .helpText("Sets the maximum number of cached parsed user-agent value.")
```
ryanemerson: "Sets the maximum number of parsed user-agent values in the local cache."

ryanemerson (commented):
LGTM just a minor suggestion and a question.

services/src/main/java/org/keycloak/device/DeviceRepresentationProviderImpl.java
pruivo: No, but IntelliJ complains about it. I replaced it with an `assert`.

pruivo:
@ryanemerson do you think it makes sense to add this cache to the metrics (hits/misses/eviction)?

ryanemerson:
> @ryanemerson do you think it makes sense to add this cache to the metrics (hits/misses/eviction)?

I don't think it's essential, but if you can add it without too much effort I think it could be beneficial :+1:

services/pom.xml
```diff
@@ -254,6 +254,10 @@
             <groupId>com.github.ua-parser</groupId>
             <artifactId>uap-java</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+        </dependency>
```
ryanemerson, line 260: Why do you need to add the dependency?

pruivo: The `snakeyaml` dependency is excluded in the root `pom.xml`, and the testsuite was failing with a class not found exception.

 In the original code, the class was lazy-loaded (`create()` method) and, most likely, the method is never invoked. But now it is loaded during `init()` whether it is used or not by the test. 

https://github.com/keycloak/keycloak/blob/d7012fa58af7c935a81fb5dad1593fffc62719c4/pom.xml#L410-L420

pruivo:
@ryanemerson added metrics. Micrometer has an implementation for caffeine in place.
```
# HELP cache_size The approximate number of entries in this cache.
# HELP cache_loads_seconds_max The number of times cache lookup methods have successfully loaded a new value or failed to load a new value, either because no value was found or an exception was thrown while loading
# HELP cache_loads_seconds The number of times cache lookup methods have successfully loaded a new value or failed to load a new value, either because no value was found or an exception was thrown while loading
# HELP cache_gets The number of times cache lookup methods have returned a cached (hit) or uncached (newly loaded) value (miss).
# HELP cache_evictions The number of times the cache was evicted.
# HELP cache_evictions_max The number of times the cache was evicted.
```

```
cache_size{cache="user-agent"} 2.0
cache_loads_seconds_max{cache="user-agent",result="failure"} 0.0
cache_loads_seconds_max{cache="user-agent",result="success"} 0.003583759
cache_loads_seconds_count{cache="user-agent",result="failure"} 0.0
cache_loads_seconds_sum{cache="user-agent",result="failure"} 0.0
cache_loads_seconds_count{cache="user-agent",result="success"} 2.0
cache_loads_seconds_sum{cache="user-agent",result="success"} 0.005067875
cache_gets_total{cache="user-agent",result="hit"} 16.0
cache_gets_total{cache="user-agent",result="miss"} 2.0
cache_evictions_count{cache="user-agent",cause="SIZE"} 0.0
cache_evictions_sum{cache="user-agent",cause="SIZE"} 0.0
cache_evictions_count{cache="user-agent",cause="EXPIRED"} 0.0
cache_evictions_sum{cache="user-agent",cause="EXPIRED"} 0.0
cache_evictions_count{cache="user-agent",cause="COLLECTED"} 0.0
cache_evictions_sum{cache="user-agent",cause="COLLECTED"} 0.0
cache_evictions_count{cache="user-agent",cause="REPLACED"} 0.0
cache_evictions_sum{cache="user-agent",cause="REPLACED"} 0.0
cache_evictions_count{cache="user-agent",cause="EXPLICIT"} 0.0
cache_evictions_sum{cache="user-agent",cause="EXPLICIT"} 0.0
cache_evictions_max{cache="user-agent",cause="SIZE"} 0.0
cache_evictions_max{cache="user-agent",cause="EXPIRED"} 0.0
cache_evictions_max{cache="user-agent",cause="COLLECTED"} 0.0
cache_evictions_max{cache="user-agent",cause="REPLACED"} 0.0
cache_evictions_max{cache="user-agent",cause="EXPLICIT"} 0.0
```

services/src/main/java/org/keycloak/device/DeviceRepresentationProviderFactoryImpl.java
```diff
@@ -14,17 +31,37 @@ public String getId() {
...
 
+    @SuppressWarnings("NullableProblems")
+    @Override
+    public void init(Config.Scope config) {
+        Caffeine<Object, Object> builder = Caffeine.newBuilder()
+                .maximumSize(config.getInt(CACHE_SIZE, DEFAULT_CACHE_SIZE))
+                .softValues();
+
+        if (config.root().getBoolean(MetricsOptions.METRICS_ENABLED.getKey(), Boolean.FALSE)) {
+            var stats = new CaffeineStatsCounter(Metrics.globalRegistry, "user-agent");
```
ryanemerson: Should we make the metric name `keycloak.user-agent` or maybe even `keycloak.device.user-agent` to better match the existing metric conventions https://www.keycloak.org/observability/metrics-for-troubleshooting-keycloak

pruivo: Only if I won't use `CaffeineStatsCounter` and implement it myself. 
Quarkus metrics don't start with `keycloak` either.

ryanemerson: Ok, I think that's fine then. You will hate me for asking, but I think we need to document the metrics if we're adding them :sweat_smile:

pruivo: No at all. It simply demonstrates that the metric isn't useful, as it doesn't matter for troubleshooting or capacity planning. At least, I can't come up with a good story for it :shrug: 
Removing it.

stianst:
Why on earth are you not using a local Infinispan cache?

pruivo:
@stianst I was trying to avoid adding infinispan code/dependency to this maven module. Infinispan caches use a `Caffeine` cache under the hood.

stianst:
> @stianst I was trying to avoid adding infinispan code/dependency to this maven module. Infinispan caches use a `Caffeine` cache under the hood.

I'd rather avoid using either Infinispan or Caffeine APIs directly and create an SPI/Provider for this purpose. Then you can do an implementation in `model/infinispan`. It's a little bit more effort, but then it's wired in the same way as everything else, and can also be replaced/customized by folks if need be.

Maybe a generic cache provider with an interface something like (it could have other use-cases as well down the road):

```
interface LocalCacheProvider {
   get(String type, String key);
   put(String type, String key, String value);
   remove(String type, String key);
}
```

Then we can also make the local cache have some configuration options if need be as well.

pruivo:
@stianst I've created https://github.com/keycloak/keycloak/issues/42223

ryanemerson:
@pruivo Do you want to merge this PR before #42223 or put this PR on hold and rebase on those efforts?

pruivo:
@ryanemerson I'll work on #42223 at a later stage. I would merge this one first.

ryanemerson (approved):

ahus1 (approved):
The dependency change is fine as we've packaging this dependency already with the distribution.

PR #42183, opened 2025-08-27
Translate the validation error returned from the backend

### LINKED ISSUE #42182: Validation errors for required actions don't show translated messages
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When entering an invalid value like "-1" into the minimum age of authentication, the error message is not shown as text, and not at the UI control element:

Image

### Version

main

### Regression

### Expected behavior

Show a proper error message, ideally at the input field.

### Actual behavior

A error popup, with a message code instead of the text.

### How to Reproduce?

Edit a required action, and add a negative value like "-1"

### Anything else?

I wanted to do some advanced validation and error message, and this one was blocking me. This is a simple reproducer with the current code.

I'll provide a draft PR and I'm looking for some advanced hints.

ahus1:
Closes #42182

Updated error message: 

image

### DIFF
--- a/js/apps/admin-ui/src/authentication/components/RequiredActionConfigModal.tsx
+++ b/js/apps/admin-ui/src/authentication/components/RequiredActionConfigModal.tsx
@@ -1,7 +1,12 @@
 import RequiredActionConfigInfoRepresentation from "@keycloak/keycloak-admin-client/lib/defs/requiredActionConfigInfoRepresentation";
 import RequiredActionConfigRepresentation from "@keycloak/keycloak-admin-client/lib/defs/requiredActionConfigRepresentation";
 import type RequiredActionProviderRepresentation from "@keycloak/keycloak-admin-client/lib/defs/requiredActionProviderRepresentation";
-import { useAlerts, useFetch } from "@keycloak/keycloak-ui-shared";
+import {
+  isUserProfileError,
+  setUserProfileServerError,
+  useAlerts,
+  useFetch,
+} from "@keycloak/keycloak-ui-shared";
 import {
   ActionGroup,
   AlertVariant,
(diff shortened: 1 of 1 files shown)


ahus1:
@keycloak/ui - I'm looking for input if/how this can be improved. Show the error message at the import control is beyond my React knowledge, so I'd be happy to receive some help or hints.

I should be able to create a E2E test for the current change.

Thanks!

ssilvert:
@ahus1 Error messages that come from the server are usually translated on the server.  

Otherwise, every kind of client using the admin API would need its own knowledge of all the errors that can happen on the server.  And, it would need its own mechanism to do the translation.

Unfortunately, most server-side error messages aren't translated at all and just show up in English.  And I'm not even sure we have always done it properly on the client side either.

jonkoops:
+1, ideally this error should be coming from the API directly. However, we currently are not consistent about this throughout the UI.

ahus1:
@jonkoops / @ssilvert - this is the same handling when updating user profile information in the Admin Console (`CreateUser.tsx`) or in the Account Console (`PersonalInfo.tsx`). 

If you want to see a more general change, please create a follow up issue for those items. 

Should I read the comment for this issue in a way that I should rewrite the error handling in the backend to provide an already translated error message for the frontend?

ssilvert:
> @jonkoops / @ssilvert - this is the same handling when updating user profile information in the Admin Console (`CreateUser.tsx`) or in the Account Console (`PersonalInfo.tsx`).
> 
> If you want to see a more general change, please create a follow up issue for those items.
> 
> Should I read the comment for this issue in a way that I should rewrite the error handling in the backend to provide an already translated error message for the frontend?

@ahus1 Since this is a common issue throughout, I won't hold you to doing the translation on the server side.  If you want to just create an issue for fixing it everywhere and reference this PR, I'll approve.

ahus1:
> If you want to just create an issue for fixing it everywhere and reference this PR, I'll approve.

@ssilvert - Issue create here: #42394 - please approve. Thanks!

ssilvert (approved):

PR #42187, opened 2025-08-27
Add some missing REST response codes

am97:
Closes #42269

I searched all the calls where `ErrorResponse.exists` is throwed, and ensured that there is a `@APIResponse(responseCode = "409", description = "Conflict")` annotation.

In some cases, there wasn't any `@APIResponse` annotation at all, so I added multiple annotations based on what I understand from the code.


ahus1 (approved):
Thank you for this PR. I've created a separated enhancement issue to track this. Assuming that the build will complete, I'm approving this PR.

PR #42189, opened 2025-08-27
fix: improve handling when expressions are disabled

shawkins:
closes: #42158

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
@@ -2,6 +2,7 @@ package org.keycloak.quarkus.runtime.configuration.mappers;
 
 import io.smallrye.config.ConfigSourceInterceptorContext;
 import io.smallrye.config.ConfigValue;
+import io.smallrye.config.Expressions;
 import jakarta.ws.rs.core.MultivaluedHashMap;
 import org.jboss.logging.Logger;
 import org.keycloak.common.util.CollectionUtil;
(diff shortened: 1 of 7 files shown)


martaisty (approved):
Thank you for a fast fix :slightly_smiling_face:

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
```diff
@@ -83,9 +84,10 @@ public static ConfigValue getValue(ConfigSourceInterceptorContext context, Strin
         //
         // The special handling of log properties is because some logging runtime properties are requested during build time
         // and we need to resolve them. That should be fine as they are generally not considered security sensitive.
+        // If however expressions are not enabled that means quarkus is specifically looking for runtime defaults, and we should not provide a value
```
vmuzikar, line 87: This seems a bit like a fragile assumption but we'd probably noticed soon enough if that ever changed.

vmuzikar (approved):
LGTM, thanks.

vmuzikar:
@shawkins I wonder about the backport here. It fixes a regression, but at the same time feels a bit risky as it might introduce another regressions.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
shawkins: We'd either need to not allow our logging options to not affect build time or we'd need a deeper hook into quarkus to differentiate between logging config used at build time and logging config for runtime defaults.

shawkins:
> @shawkins I wonder about the backport here. It fixes a regression, but at the same time feels a bit risky as it might introduce another regressions.

I think backporting is safe, but there are workarounds - use an intermediate property or don't specify the log-level at build time.

PR #42193, opened 2025-08-27
Initial implementation for the RLM scheduled task

### LINKED ISSUE #42105: [RLM] Add a cluster-aware task runner that automatically executes scheduled actions periodically
### Description

We need to add a cluster-aware task runner to execute all resource actions that were scheduled by resource policies and whose execution time is due.

Ideally we should allow admins to configure how the frequency (interval between runs) of the task and perhaps even the optional time for the first run. For example, run every 24 hours at 11 pm.

### LINKED ISSUE #42192: Username form when organizations is enabled
### Before reporting an issue

### Area

organizations

### Describe the bug

Given:
- The organizations are enabled and there is at least one active organization
- Use a custom browser flow in which `Username form` and `Password form` are used instead of `Username Password form`
- A user have a linked identity provider

When that user logs in, there is an additional blank form page between the username form and the password form.

### Version

26.3.3

### Regression

### Expected behavior

No blank page.
The user should immediately see the next step which is the `Password form`

### Actual behavior

There is a blank form page with only the `Sign-in` button between the `Username form` and the `Password form`

Image

### How to Reproduce?

- Create a realm
- Enable organizations
- Create an organization
- Configure a custom `browser` auth flow: Replace (`Username Password form`) by (`Username form` AND `Password form`, set to `REQUIRED` of course)
- Add an IDP (Tested with Google and Github)
- Connect a user through the IDP to link the account
- [Optional] Set a password
- Logout
- Use the user email or username to login using the username form

### Anything else?

This does not occur when there are 0 organizations, nor when organizations are disabled, or when the `Organization Identity-First Login` authentication step is disabled.

However it is still reproducible if there are organizations and they're disabled.

---

### Video demo

https://github.com/user-attachments/assets/66385229-d9a0-44a9-b248-0229254adbda

### The configured browser flow
Image

pedroigor:
Closes #42105

* Depends on https://github.com/keycloak/keycloak/pull/42185

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/policy/ResourceAction.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/policy/ResourceAction.java
@@ -22,7 +22,7 @@ import org.keycloak.component.ComponentModel;
 
 public class ResourceAction implements Comparable<ResourceAction> {
 
-    private static final String AFTER_KEY = "after";
+    public static final String AFTER_KEY = "after";
     private static final String PRIORITY_KEY = "priority";
 
     private String id;
--- a/services/src/main/java/org/keycloak/models/policy/DisableUserActionProvider.java
+++ b/services/src/main/java/org/keycloak/models/policy/DisableUserActionProvider.java
@@ -17,6 +17,8 @@
 
 package org.keycloak.models.policy;
 
+import static org.keycloak.models.policy.ResourceAction.AFTER_KEY;
+
 import java.util.List;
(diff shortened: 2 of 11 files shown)


sguilhen:
You need to fix the referenced issue here - this is pointing to a completely unrelated issue

pedroigor:
@sguilhen Sorry. Fixed the PR description.

sguilhen (approved):

ahus1 (approved):

PR #42211, opened 2025-08-28
Add support for aggregated actions

### LINKED ISSUE #42119: [RLM] Add support for composite actions
### Description

Current implementation of the Resource Lifecycle Management feature doesn't allow multiple actions running at the same time. It schedules always the next action and only when that action is executed it moves onto the next one in line.

In some situations, admins might want to execute more than one action at a time. For example, an onboarding policy might setup up required actions, add roles to the user, join groups, all in one go instead of running each action separately on different runs of the periodic task.

Grouping actions in a composite action would allow policies to define a single execution time for all actions involved.

pedroigor:
Closes #42119

### DIFF
--- a/services/src/main/java/org/keycloak/models/policy/ResourcePolicyManager.java
+++ b/services/src/main/java/org/keycloak/models/policy/ResourcePolicyManager.java
@@ -17,14 +17,20 @@
 
 package org.keycloak.models.policy;
 
-import jakarta.ws.rs.BadRequestException;
+import static java.util.Optional.ofNullable;
+
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


services/src/main/java/org/keycloak/models/policy/ResourcePolicyManager.java
```diff
@@ -317,4 +356,54 @@ private ComponentModel getPolicyComponent(String id) {
...
+
+            for (Entry<String, List<String>> configEntry : condition.getConfig().entrySet()) {
+                config.put(conditionProviderId + "." + configEntry.getKey(), configEntry.getValue());
+            }
+        }
+
+        ResourcePolicy policy = manager.addPolicy(rep.getProviderId(), config);
+        List<ResourceAction> actions = new ArrayList<>();
+
+        for (ResourcePolicyActionRepresentation actionRep : rep.getActions()) {
```
pedroigor, line 406: We should probably have some validation to prevent actions that do not support sub actions (not aggregated) from having them.

I can do that as a follow-up.

sguilhen (approved):

ahus1 (approved):

PR #42218, opened 2025-08-28
fix: don't show the local access screen if a service account exists

### LINKED ISSUE #42201: Local access required if KC_BOOTSTRAP_ADMIN_CLIENT_ID is set but not KC_BOOTSTRAP_ADMIN_USERNAME
### Before reporting an issue

### Area

login/ui

### Describe the bug

I set up a new keyloak and bootstrapped a keycloak admin client (not a user).
I have populated the keycloak with admin users and and oidc provider via crossplane.
When trying to access the admin endpoint via https://myurl.keycloak, I get a message, that I have to set up an admin user locally:

Image

However, when I access https://myurl.keycoak/admin, the frontend works and I can login as I would have expected.

This issue is basically the same as described in the auto-closed issue #38520

### Version

26.3.3

### Regression

### Expected behavior

Either of KC_BOOTSTRAP_ADMIN_CLIENT_ID and KC_BOOTSTRAP_ADMIN_USERNAME should remove the local setup notification

### Actual behavior

Only KC_BOOTSTRAP_ADMIN_USERNAME removes the local setup notification

### How to Reproduce?

1. Create a new keycloak with a new database
2. Use KC_BOOTSTRAP_ADMIN_CLIENT_ID and KC_BOOTSTRAP_ADMIN_CLIENT_SECRET in the environment variables
3. Access the admin endpoint (in my example myurl.keycloak) -> The local setup notification appeasr
4. Access the admin endpoint with /admin -> Works as expected

### Anything else?

_No response_

shawkins:
closes: #42201

### DIFF
--- a/test-framework/examples/tests/src/test/java/org/keycloak/test/examples/PagesTest.java
+++ b/test-framework/examples/tests/src/test/java/org/keycloak/test/examples/PagesTest.java
@@ -2,9 +2,11 @@ package org.keycloak.test.examples;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
-import org.keycloak.testframework.annotations.InjectRealm;
+import org.keycloak.admin.client.Keycloak;
+import org.keycloak.testframework.annotations.InjectAdminClient;
 import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
-import org.keycloak.testframework.realm.ManagedRealm;
+import org.keycloak.testframework.remote.runonserver.InjectRunOnServer;
+import org.keycloak.testframework.remote.runonserver.RunOnServerClient;
 import org.keycloak.testframework.ui.annotations.InjectPage;
 import org.keycloak.testframework.ui.annotations.InjectWebDriver;
 import org.keycloak.testframework.ui.page.LoginPage;
@@ -15,8 +17,12 @@ import org.openqa.selenium.htmlunit.HtmlUnitDriver;
 
 @KeycloakIntegrationTest
 public class PagesTest {
-    @InjectRealm(ref = "master", attachTo = "master")
-    ManagedRealm masterRealm;
+
+    @InjectAdminClient
+    Keycloak adminClient;
+
+    @InjectRunOnServer
+    RunOnServerClient runOnServer;
 
     @InjectWebDriver
     WebDriver webDriver;
@@ -29,16 +35,7 @@ public class PagesTest {
 
     @Test
     public void testLoginFromWelcome() {
-        masterRealm.admin().users().searchByUsername("admin", true)
-                .stream().findFirst().ifPresent(admin ->
-                        masterRealm.admin().users().delete(admin.getId()));
-
         welcomePage.navigateTo();
-
-        welcomePage.assertCurrent();
-        welcomePage.fillRegistration("admin", "admin");
-        welcomePage.submit();
-        welcomePage.clickOpenAdminConsole();
 
         if (webDriver instanceof HtmlUnitDriver) {
             String pageId = webDriver.findElement(By.xpath("//body")).getAttribute("data-page-id");
@@ -52,6 +49,7 @@ public class PagesTest {
             loginPage.fillLogin("admin", "admin");
             loginPage.submit();
         }
+
     }
 
 }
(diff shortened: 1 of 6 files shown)


testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/url/HostnameV2Test.java
```diff
@@ -135,29 +132,6 @@ public void testRealmFrontendUrlWithOtherUrlsSet() {
         testAdminUrls(realmFrontendName, realmFrontendUrl, fixedAdminUrl);
     }
 
-    @Test
-    public void testAdminLocal() throws Exception {
```
vmuzikar, line 139: I don't think we have a replacement for this test, do we?

test-framework/examples/tests/src/test/java/org/keycloak/test/examples/PagesTest.java
```diff
@@ -52,6 +69,10 @@ public void testLoginFromWelcome() {
             loginPage.fillLogin("admin", "admin");
             loginPage.submit();
         }
+
+        assertTrue(runOnServer.fetch(session -> new ApplianceBootstrap(session)
```
vmuzikar: Should this be in `finally` block in case the test failed? We're not enforcing creation of a new server in this test, so AFAIK it'll survive the test class, no?

vmuzikar (commented):
@shawkins Thanks for the PR.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/url/HostnameV2Test.java
shawkins: This is no longer a good test - we no longer use the local admin url. It was just checking that you hit the local access screen, which we do test elsewhere.

test-framework/examples/tests/src/test/java/org/keycloak/test/examples/PagesTest.java
shawkins: > Should this be in finally block in case the test failed? We're not enforcing creation of a new server in this test, so AFAIK it'll survive the test class, no?

It could be, but that's not how the admin user deletion above was being handled either.

I'm a little fuzzy on exactly what the expectation is for these examples tests - are they both tests and examples of how to write tests? Just like the other tests I can also show using the annotation config to reset the static side effect to creating the bootstrap user via the admin console. However that and the deletion of the bootstrap admin and usage of the appliancebootstrap don't feel like they make for a good example.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/url/HostnameV2Test.java
vmuzikar: Right, I forgot we removed the usage of it and deprecated it.

test-framework/examples/tests/src/test/java/org/keycloak/test/examples/PagesTest.java
vmuzikar: > It could be, but that's not how the admin user deletion above was being handled either.

Yeah, that's right. It was not using finally before either. It's not a blocker for me, though, we can keep it as is.

> I'm a little fuzzy on exactly what the expectation is for these examples tests - are they both tests and examples of how to write tests? Just like the other tests I can also show using the annotation config to reset the static side effect to creating the bootstrap user via the admin console. However that and the deletion of the bootstrap admin and usage of the appliancebootstrap don't feel like they make for a good example.

I don't have any idea... Maybe @lhanusov might shed some light on what are these examples for?

vmuzikar (dismissed):
LGTM, thanks!

vmuzikar:
@stianst Do you want to take a look since you're marked as a reviewer, or are you fine with merging it?

test-framework/examples/tests/src/test/java/org/keycloak/test/examples/PagesTest.java
stianst, line 1: You can just delete this example test, or better simplify it to not use complexity of the welcome page; it's really just supposed to be a basic example of how to inject and use pages.

shawkins: Simplified to just remove the user manipulation.

vmuzikar (approved):

vmuzikar:
@stianst Do you want to re-review?

vmuzikar:
@stianst I'm going ahead and merging it. Let me know if you have any concerns, we can address them as a follow-up.

PR #42250, opened 2025-08-29
Create default ServiceMonitor with Operator

ryanemerson:
Closes #40406


operator/scripts/create-olm-bundle.sh
```diff
@@ -69,6 +69,9 @@ yq ea -i '.spec.install.spec.deployments[0].spec.template.metadata.labels.name =
 yq ea -i '.spec.install.spec.deployments[0].spec.template.spec.containers[0].env += [{"name": "POD_NAME", "valueFrom": {"fieldRef": {"fieldPath": "metadata.name"}}}]' "$CSV_PATH"
 yq ea -i '.spec.install.spec.deployments[0].spec.template.spec.containers[0].env += [{"name": "OPERATOR_NAME", "value": "keycloak-operator"}]' "$CSV_PATH"
 
+# Remove ServiceMonitors GVK from nativeAPIS to allow CSV installation when CRDs not present
+yq ea -i 'del(.spec.nativeAPIs[] | select(.kind == "ServiceMonitor")) | del(.spec.install.spec.permissions[].rules[] | select(.resources[] == "servicemonitors"))' "$CSV_PATH"
```
ryanemerson: @keycloak/cloud-native I couldn't find a way to prevent `ServiceMonitor` being added to the nativeAPIS in the Java code or via properties, so this was my workaround. I have asked the operatorsdk team on Slack and they are going to work on adding support for this type of use-case.

https://github.com/quarkiverse/quarkus-operator-sdk/issues/1154

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    protected ServiceMonitor desired(Keycloak primary, Context<Keycloak> context) {
+        var opts = configuredOptions(primary);
+        var metricsPath = opts.getOrDefault("http-management-relative-path", "") + "/metrics";
+        var meta = primary.getMetadata();
+        var scheme = isTlsConfigured(primary) ? "https" : "http";
+        var spec = ServiceMonitorSpec.get(primary);
+        return new ServiceMonitorBuilder()
```
ryanemerson, line 45: Are there any other Keycloak configuration options that we should be taking into account here?

ahus1 (changes requested):
Please update the docs that following other place to also make that a `ServiceMonitor` example where it is a `PodMonitor` example today. 

https://github.com/keycloak/keycloak/blob/2ad776553a7b0281ee1000f4fe51131be3af0381/docs/guides/observability/exemplars.adoc?plain=1#L50-L59

Also, please add the `scrapeProtocols` by default with `OpenMetricsText1.0.0` to allow scraping of exemplars.

ryanemerson:
Thanks for the review @ahus1, I have addressed all suggestions.

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -507,4 +507,74 @@ spec:
       annotation2: annotation-value2
 ----
 
+=== ServiceMonitor
+The {project_name} operator automatically generates a `ServiceMonitor` resource if the `monitoring.coreos.com/v1:ServiceMonitor`
```
ahus1: One more thought here (sorry for not mentioning it in the first round): Please add a paragraph explaining the purpose of a  `ServiceMonitor`, and that it is an CRD that first needs to be installed.

ahus1 (commented):
Thank you for addressing my previous comment. See below for one more docs paragraph that should be added. Thanks!

docs/guides/operator/advanced-configuration.adoc
ryanemerson: The commit pushed with requested changes.

ahus1 (dismissed):

operator/scripts/create-olm-bundle.sh
ryanemerson: The PR for ^ has now been resolved. Once we upgrade to the quarkus version with the fix, it will be possible for us to remove the `yq` statement and the operator-sdk will not add `ServiceMonitor` to the nativeApis as we're using `CRDPresentActivationCondition` (I have tested locally with operator-sdk SNAPSHOT).

Assuming we're happy to merge as is, I can create a follow-up issue to handle this.

.github/workflows/operator-ci.yml
```diff
@@ -98,6 +98,11 @@ jobs:
           driver: docker
           start args: --addons=ingress --memory=${{ env.MINIKUBE_MEMORY }} --cni cilium --cpus=max
 
+      - name: Install ServiceMonitor CRD
```
vmuzikar: We should probably do this after all tests and run the ServiceMonitor tests separately. Or in other words, we need to test the Operator also without the ServiceMonitor CRD.

operator/scripts/create-olm-bundle.sh
vmuzikar: Just a silly question, how are we gonna give the Operator the permissions to manage the ServiceMonitor resource if the permissions are removed?

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/CRDUtils.java
```diff
@@ -87,6 +78,20 @@ public static boolean isManagementEndpointEnabled(Keycloak keycloak) {
                 .anyMatch(Boolean::parseBoolean);
     }
 
+    public static Map<String, String> configuredOptions(Keycloak keycloak) {
```
shawkins, line 81: This is highly related to KeycloakDeploymentDependentResource.readConfigurationValue. We might want to consider only a single routine - especially if we reconsider reading the associated secret (or perhaps eventually a configmap). That causes quite a bit more api server traffic as we're getting those value afresh for each call. cc @vmuzikar

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
shawkins: Unfortunately yes. It would be best to have some common logic with KeycloakDeploymentDependentResource.createBaseDeployment - or to make the servicemonitor reconciliation dependent on first reconciling the statefulset so that the endpoint could be scraped from the probe configuration.

operator/scripts/create-olm-bundle.sh
vmuzikar: Give this is OLM specific, we should probably include `ServiceMonitor` in OLM tests as well.

operator/src/main/kubernetes/kubernetes.yml
```diff
@@ -72,12 +72,29 @@ rules:
...
+      - monitoring.coreos.com
+    resources:
+      - servicemonitors
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - update
+      - watch
```
vmuzikar, line 85: When using [raw K8s resources](https://github.com/keycloak/keycloak-k8s-resources/blob/nightly/kubernetes/kubernetes.yml) (instead of OLM), I assume K8s will also complain if `ServiceMonitor` CRD is not installed, right? Should we have a dedicated YAML with permissions for `ServiceMonitor`?

vmuzikar (commented):
@ryanemerson Thank you for the PR.

operator/scripts/create-olm-bundle.sh
ryanemerson: I was torn on this one, because not including the `ServiceMonitor` in the OLM tests highlighted the issue with `nativeAPIs` when the CRD was not present in the cluster.

I could add additional steps to the OLM tests where we install the CRDs, create a new cluster and then verify that the `ServiceMonitor` is created as expected?

operator/src/main/kubernetes/kubernetes.yml
ryanemerson: > I assume K8s will also complain if `ServiceMonitor` CRD is not installed, right?

No that's not the case AFAIU. When you create a `ClusterRole/Role`, the only thing that is verified is the structure of the yaml. K8s does not ensure that the resource(s) are actually available on the system until you try to access them, at which point you will get a resource not found error before RBAC is verified if the CRD is not available.

You can test this by applying `kubernetes.yml` on a cluster that does not contain the ServiceMonitor CRD.

operator/scripts/create-olm-bundle.sh
ryanemerson: ~We're not removing the permissions here, just the definition in the `nativeAPIs` section of the CSV which marks the resource as being mandatory for the Operator to work.~

I couldn't have got that more wrong, as you say we're removing the roles as well :sweat: . Fixing!

ryanemerson: I have updated CI so that we test OLM with and without the ServiceMonitor CRDs present. To reduce repetition, I have added some additional scripts to `operator/scripts`. I also have added some additional steps to capture CR status and the cluster state with `oc adm inspect` when a failure occurs to make it easier to determine issues without recreating everything locally.

.github/workflows/operator-ci.yml
ryanemerson: I have updated the actions so that we execute `ServiceMonitorTest` with and without the CRDs present.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
ryanemerson: I have consolidated the logic into the `magicEndpoint` method

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/CRDUtils.java
ryanemerson: I don't really understand what you're suggesting here. Are there changes required in this PR and if so, can you elaborate a little what changes you would like?

ryanemerson:
Rebased on `main` to resolve conflicts, ready for a review when you get a chance @shawkins @vmuzikar, thanks.

operator/src/test/java/org/keycloak/operator/testsuite/integration/ServiceMonitorTest.java
```diff
@@ -0,0 +1,117 @@
...
+public class ServiceMonitorTest extends BaseOperatorTest {
+
+    @Test
+    public void testServiceMonitorDisabledIfCRDNotAvailable() {
+        Assumptions.assumeFalse(isServiceMonitorAvailable(k8sclient));
+        // Test fails if the CRD is not available and the Operator is unable to handle missing CRD
+    }
+
+    @Test
+    @DisabledIfApiServerTest
```
shawkins: If the servicemonitor crd is installed, these tests would run against the local api server correct? 

I wouldn't be opposed to having a reference version of the crd in the test resources and installed by this test so that it could be run in any testing mode - and we could get rid of the specific github testing steps in the test-remote flow.

.github/workflows/operator-ci.yml
```diff
@@ -200,16 +211,41 @@ jobs:
...
-          kubectl apply -f src/test/resources/example-realm.yaml -n keycloak
-          # Wait for the CRs to be ready
-          ./scripts/check-examples-installed.sh keycloak
+          ./deploy-examples.sh keycloak
+          ./undeploy-examples.sh keycloak
+
+      - name: Install ServiceMonitor CRD
+        working-directory: operator/scripts
+        run: |
+          LATEST=$(curl -s https://api.github.com/repos/prometheus-operator/prometheus-operator/releases/latest | jq -cr .tag_name)
```
shawkins: I know it's not likely to be a problem until some major version change, but I'm not sure about always following the latest.

Will this always be the same as what is supported in our newest supported version of openshift?

I realize the alternative isn't great - either to bump the support as needed or see if we can get version from our target version of openshift somehow.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -658,4 +643,27 @@ private static void addUpdateRevisionAnnotation(String revision, StatefulSet toU
...
     }
 
+    record ManagementEndpoint(String relativePath, String protocol, int port) {}
+
+    static ManagementEndpoint managementEndpoint(Keycloak keycloakCR, Context<Keycloak> context) {
+        boolean tls = isTlsConfigured(keycloakCR);
+        String protocol = tls ? "HTTPS" : "HTTP";
+        int port;
+
+        if (readConfigurationValue(HTTP_MANAGEMENT_HEALTH_ENABLED, keycloakCR, context).map(Boolean::valueOf).orElse(true)) {
```
shawkins: This is only valid if we're specifically looking for health endpoints. The method will probably need an additional boolean or enum parameter: managementEndpoint(... , boolean health)

shawkins (changes requested):
Thank you for the updates @ryanemerson, just a couple of thoughts.

.github/workflows/operator-ci.yml
ryanemerson: > Will this always be the same as what is supported in our newest supported version of openshift?

We will be ahead of what is in Openshift, however I don't *think* that's a problem as we're only concerned with a specific CRD version here and that should only evolve in a forward compatible manner. We just have to be careful if/when configuring any new fields on the ServiceMonitor, but that will be determined by the Fabric8 dependency version rather than the CRD version we install here.

operator/src/test/java/org/keycloak/operator/testsuite/integration/ServiceMonitorTest.java
ryanemerson: I have pushed  a commit to take this approach and remove the need to download CRDs in the CI.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
ryanemerson: Good catch! I have pushed a new commit so that we always use the management protocol and port when defining the ServiceMonitor spec.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/CRDUtils.java
shawkins: Sorry I didn't follow up on this. It's more a mental / team note. The only thing that could be added as part of this pr would be a comment or todo about consolidating with readConfigurationValue.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    protected ServiceMonitor desired(Keycloak primary, Context<Keycloak> context) {
+        var endpoint = managementEndpoint(primary, context, true);
+        var meta = primary.getMetadata();
+        var spec = ServiceMonitorSpec.get(primary);
+        return new ServiceMonitorBuilder()
+              .withNewMetadata()
+                .withName(meta.getName())
+                .withNamespace(meta.getNamespace())
```
shawkins, line 48: Do you know if users will want custom labels and annotations on these? If so it could be added as a follow-on task under #15395

shawkins (approved):
LGTM, thank you @ryanemerson - this ended up being more work than I would have first guessed.

Just one additional thought on labels / annotations.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
ryanemerson: Possibly yes, we have had a request for this in the Infinispan Operator in the past: https://github.com/infinispan/infinispan-operator/issues/1777.

I have created the sub-issue #42626 :+1:

ahus1 (approved):
Thank you for everyone involved in this one!

ahus1:
@ryanemerson - please create a follow-up issue for the release notes. Thanks!

PR #42255, opened 2025-08-29
Add support for generic event-based policies and conditions

pedroigor:

### DIFF
--- a/core/src/main/java/org/keycloak/representations/resources/policies/ResourcePolicyConditionRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/resources/policies/ResourcePolicyConditionRepresentation.java
@@ -1,0 +1,77 @@
+package org.keycloak.representations.resources.policies;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ResourcePolicyConditionRepresentation {
+
+    public static Builder create() {
+        return new Builder();
+    }
+
+    private String id;
+    private String providerId;
+    private Map<String, List<String>> config;
+
+    public ResourcePolicyConditionRepresentation() {
+        // reflection
+    }
+
+    public ResourcePolicyConditionRepresentation(String providerId) {
+        this(providerId, null);
+    }
+
+    public ResourcePolicyConditionRepresentation(String providerId, Map<String, List<String>> config) {
+        this(null, providerId, config);
+    }
+
+    public ResourcePolicyConditionRepresentation(String id, String providerId, Map<String, List<String>> config) {
+        this.id = id;
+        this.providerId = providerId;
+        this.config = config;
+    }
+
+    public String getProviderId() {
+        return providerId;
+    }
+
+    public void setProviderId(String providerId) {
+        this.providerId = providerId;
+    }
+
+    public Map<String, List<String>> getConfig() {
+        return config;
+    }
+
+    public void setConfig(Map<String, List<String>> config) {
+        this.config = config;
+    }
+
+    public void setConfig(String key, String value) {
+        if (this.config == null) {
+            this.config = new HashMap<>();
+        }
+        this.config.put(key, Collections.singletonList(value));
+    }
+
+    public static class Builder {
+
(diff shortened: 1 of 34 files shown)


core/src/main/java/org/keycloak/representations/resources/policies/ResourcePolicyConditionRepresentation.java
```diff
@@ -0,0 +1,77 @@
+package org.keycloak.representations.resources.policies;
```
sguilhen, line 1: Are we sticking to the old headers for new files or should we start leaving them empty?

model/jpa/src/main/java/org/keycloak/models/policy/EventBasedResourcePolicyProvider.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    public List<String> getEligibleResourcesForInitialAction() {
+        return List.of();
+    }
+
+    @Override
+    public boolean supports(ResourceType type) {
+        return ResourceType.USERS.equals(type);
+    }
```
sguilhen, line 30: Should this really be here? This generic event approach could work with any resource provided the event applies to the resource. I wonder if this should be overridden by resource-specific subclasses.

services/src/main/java/org/keycloak/models/policy/ResourcePolicyEventListenerFactory.java
```diff
@@ -48,22 +47,27 @@ public void init(Scope config) {
...
+            if (session != null) {
+                // try first running within the session/transaction the event was fired
+                onEvent(event, session);
+                return;
             }
+
+            // fallback to running in a new session/transaction
+            KeycloakModelUtils.runJobInTransaction(factory, s -> {
+                onEvent(event, s);
+            });
```
sguilhen: I think we need to use the variant that takes the session context because the new session won't have the realm properly set

tests/base/src/test/java/org/keycloak/tests/admin/model/policy/UserCreationTimePolicyTest.java
```diff
@@ -7,6 +7,7 @@
 import org.keycloak.models.UserModel;
 import org.keycloak.models.policy.DisableUserActionProviderFactory;
 import org.keycloak.models.policy.NotifyUserActionProviderFactory;
+import org.keycloak.models.policy.ResourceOperationType;
```
sguilhen: Is this used?

core/src/main/java/org/keycloak/representations/resources/policies/ResourcePolicyConditionRepresentation.java
pedroigor: Leaving them empty, as they are no longer mandatory.

services/src/main/java/org/keycloak/models/policy/ResourcePolicyEventListenerFactory.java
pedroigor: We do not have a session here. But I agree this will fail, perhaps we should not have this fallback at all?

model/jpa/src/main/java/org/keycloak/models/policy/EventBasedResourcePolicyProvider.java
pedroigor: For now, we only support users.

services/src/main/java/org/keycloak/models/policy/ResourcePolicyEventListenerFactory.java
sguilhen: Ah true.. if we could guarantee that every provider event we handle has the session available it would be really helpful

pedroigor: Removed the fallback

sguilhen (approved):

ahus1 (approved):
Approving as of previous reviews

PR #42275, opened 2025-09-01
Session IDs and auth codes should have 128 bits of entropy

### LINKED ISSUE #42274: Session IDs have IDs have less than 128 bits of entropy
### Description

As a follow-up to #38663, user session IDs as well as authorization codes should have that randomness.

### Discussion

https://github.com/keycloak/keycloak/issues/38663

### Motivation

Prevent brute force attacks against those IDs. 

Today they are UUIDs.

### Details

The same approach as in #38663

ahus1:
Closes #42274

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/util/InfinispanKeyGenerator.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/util/InfinispanKeyGenerator.java
@@ -27,8 +27,8 @@ import org.infinispan.affinity.KeyAffinityService;
 import org.infinispan.affinity.KeyAffinityServiceFactory;
 import org.infinispan.affinity.KeyGenerator;
 import org.jboss.logging.Logger;
+import org.keycloak.common.util.SecretGenerator;
 import org.keycloak.models.KeycloakSession;
-import org.keycloak.models.utils.KeycloakModelUtils;
 import org.keycloak.sessions.StickySessionEncoderProvider;
 
 /**
@@ -97,7 +97,7 @@ public class InfinispanKeyGenerator {
 
         @Override
         public String getKey() {
-            return KeycloakModelUtils.generateId();
+            return SecretGenerator.getInstance().generateSecureID();
         }
     }
 }
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/util/InfinispanKeyGenerator.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/util/InfinispanKeyGenerator.java
@@ -88,7 +88,7 @@ public class InfinispanKeyGenerator {
 
         @Override
         public UUID getKey() {
-            return UUID.randomUUID();
+            return SecretGenerator.getInstance().generateSecureUUID();
         }
     }
 
--- a/services/src/main/java/org/keycloak/protocol/oidc/utils/OAuth2CodeParser.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/utils/OAuth2CodeParser.java
@@ -82,20 +82,12 @@ public class OAuth2CodeParser {
             return result.illegalCode();
         }
 
+        String codeUUID = parsed[0];
         String userSessionId = parsed[1];
         String clientUUID = parsed[2];
 
         event.detail(Details.CODE_ID, userSessionId);
         event.session(userSessionId);
-
-        // Parse UUID
-        String codeUUID;
-        try {
-            codeUUID = parsed[0];
-        } catch (IllegalArgumentException re) {
-            logger.warn("Invalid format of the UUID in the code");
-            return result.illegalCode();
-        }
 
         // Retrieve UserSession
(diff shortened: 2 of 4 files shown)


ahus1:
@ryanemerson - please review when you have the time and feel confident enough about the related code changes. Thanks!

pruivo:
@ahus1 the `org.keycloak.models.sessions.infinispan.util.InfinispanKeyGenerator.UUIDKeyGenerator` is used for client session. Doesn't it need to be secure?

pruivo:
I micro-benchmarked the Keycloak generator: https://github.com/pruivo/playground/blob/master/uuid-generator/src/main/java/me/pruivo/UUIDGeneratorBenchmark.java
Raw results in readme file: https://github.com/pruivo/playground/tree/master/uuid-generator

Summary
| Benchmark | Time ns/op | Memory byte/op |
|--------|--------|--------|
| Keycloak | 328.219 | 376.000 |
| Alternative (cached `byte[]` & cached `StringBuilder`) | 315.654 | 176.000 |
| Alternative v2 (cached `byte[]` & `VarHandle`) | 186.370 | 144.000 |

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/util/InfinispanKeyGenerator.java
```diff
@@ -97,7 +97,7 @@ private static class StringKeyGenerator implements KeyGenerator<String> {
 
         @Override
         public String getKey() {
-            return KeycloakModelUtils.generateId();
+            return SecretGenerator.getInstance().generateSecureID();
```
stianst, line 100: Is this the only place session IDs are generated?

stianst: Maybe StringKeyGenerator should be renamed to SecureIDGenerator?

services/src/main/java/org/keycloak/protocol/oidc/utils/OAuth2CodeParser.java
stianst, line 1: Could you look into why the code is 3 uuids at the moment? I think it could just be a single and longer secure ID.

stianst (commented):
Does SecretGenerator.generateSecureID have enough entropy and is it collision resistant?

services/src/main/java/org/keycloak/protocol/oidc/utils/OAuth2CodeParser.java
ahus1: The user session ID and the client ID are useful to create a meaningful event with context information once the code is not found in the database on retries - otherwise you wouldn't know the problematic client or user session. So I'd say we should keep them.

ahus1:
> I micro-benchmarked the Keycloak generator

Please create a separate issue and PR for that. Note that the `keycloak-common` module is currently on language level 8, so no records allowed here.

ahus1:
@stianst - 

> Does SecretGenerator.generateSecureID have enough entropy and is it collision resistant?

The `SecretGenerator` is using `SecureRandom`, and also `UUID.randomUUID` is using `SecureRandom`. The one difference I see is that it is using a `ThreadLocal` to have one instance per thread. Please let me know if you have concerns here.

ahus1:
@pruivo:

> the org.keycloak.models.sessions.infinispan.util.InfinispanKeyGenerator.UUIDKeyGenerator is used for client session. Doesn't it need to be secure?

I pushed one more commit to change this. IMHO it is not exposed to any external API and is only internal, so it should have been fine without it. Still, let's fix it anyway.

pruivo:
> > I micro-benchmarked the Keycloak generator
> 
> Please create a separate issue and PR for that. Note that the `keycloak-common` module is currently on language level 8, so no records allowed here.

@ahus1 sure I can do it. Why is common using Java 8?

ahus1:
> Why is common using Java 8?

I assume that it is shared with [Keycloak's client libraries](https://github.com/keycloak/keycloak-client), and they might have different requirements. You should loop them in to see when/if they want to upgrade.

ahus1:
I'd say all comments have been addressed. Is this PR now ready to be approved?

pruivo:
> @pruivo:
> 
> > the org.keycloak.models.sessions.infinispan.util.InfinispanKeyGenerator.UUIDKeyGenerator is used for client session. Doesn't it need to be secure?
> 
> I pushed one more commit to change this. IMHO it is not exposed to any external API and is only internal, so it should have been fine without it. Still, let's fix it anyway.

@ahus1 did you forget to push the commit? Or changed your mind? :)

ahus1:
> did you forget to push the commit?

oh, yes. Pushed it.

pruivo (approved):

ryanemerson (approved):

ahus1:
@pedroigor or @mposolda - can one of you please have a look as well if I'm missing something where there are some sessions identifiers that should be handled as well? Thanks!

mposolda (approved):

PR #42306, opened 2025-09-03
Document that single-cluster deployments expect all Keycloak instances to serve traffic

### LINKED ISSUE #33942: Make sure Keycloak endpoints have DPoP validation

### LINKED ISSUE #38621: Client secret generation provides lower than expected entropy
### Before reporting an issue

### Area

authentication

### Describe the bug

The way how we generate client secrets in authentication flows such as "Client Id and Secret" and "Signed Jwt with Client Secret" uses a character set consisting of 62 alphanumeric characters (A-Z, a-z, 0-9). Although the secret lengths are defined as 256, 384, or 512 bits, this method produces secrets with much less actual entropy due to the constrained character set.

For example, a 256-bit secret generated using 32 characters from a 62-character set results in only ~192 bits of entropy, not the expected 256 bits. Similar entropy shortfalls exist for 384- and 512-bit secrets, capping their entropy at ~286 and ~381 bits respectively. This mismatch reduces the strength of MAC-based authentication methods, such as those using HS256, HS384, or HS512, thereby weakening message integrity and authentication.

### Version

26.1.4

### Regression

### Expected behavior

Client secrets should meet the entropy level associated with their configured bit-length (e.g., 256 bits of entropy for a 256-bit secret).

### Actual behavior

Secrets use only 62 characters for generation, resulting in reduced entropy (e.g., 192 bits instead of 256 for 256-bit secrets). This reduces the strength of HMAC-based client authentication.

### How to Reproduce?

1. Go to the Keycloak admin console.
    
2. Create a new client or select an existing one.
    
3. Navigate to the "Credentials" tab.
    
4. Observe the generated client secret.
    
5. Note that the secret is composed of alphanumeric characters only.
    
6. Calculate the theoretical entropy using the formula: `entropy = length * log2(alphabet size)`. Or use the following tools https://www.omnicalculator.com/other/password-entropy

### Anything else?

**Suggested Fixes:**

- **Expand the character set:** Include special characters (e.g., `!@#$%^&*()_+=-`) in the character pool to increase the entropy of generated secrets.
    
- **Verify use of CSPRNG:** Ensure that the `SecretGenerator.getInstance().randomString` method utilizes a cryptographically secure pseudo-random number generator (CSPRNG).

### LINKED ISSUE #39191: Ability to skip AIA for adding WebAuthn security key in case that user already has one
### Description

This RFE is about the ability that application can request user to register WebAuthn security key (passkey) during his login by AIA just in case when user does not yet have security key on his account. The desired functionality is:

1) User will see the option to register security key (passkey) after he finish Keycloak authentication. He can choose to register passkey or he can click "Cancel" in which case, he would be redirected to the application
2) In case that user already has passkey, the screen should not be displayed as there is no need to ask user to register other passkey if he already has one

The (1) can be realized by using AIA (which is already supported) . The (2) is currently not supported as when application sends `kc_action=webauthn-register-passwordless`, then the registration of passkey is always displayed to the user (Regardless of whether he has already security key or not).

Note that application does not know in advance if user has passkey as user is usually not yet authenticated.

## Proposal

Make the actions `webauthn-register` and `webauthn-register-passwordless` to be parameterized. So it would be possible to use something like `kc_action=webauthn-register-passwordless:skip-if-exists`. When this parameter `skip-if-exists` is used and user already has credential of type `webauthn-register-passwordless`, then the AIA will be ignored and will not be shown to the user. When the parameter is not used and request is sent with `kc_action=webauthn-register-passwordless`, the behaviour will be same as like in Keycloak 26.2. So the screen to register passkey would be always displayed.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

### LINKED ISSUE #39486: Email server credentials can be harvested through host/port manipulation

### LINKED ISSUE #39596: Enabling "HTTP-POST binding response" is not reflected in the SP metadata
### Before reporting an issue

### Area

saml

### Describe the bug

The SAML SP metadata always uses `Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"` for the AssertionConsumerService, even if "HTTP-POST binding response" is enabled in the Identity Provider configuration

Image

```
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="...truncated...
```

### Version

26.1.4

### Regression

### Expected behavior

The ACS binding in the SP metadata should indicate HTTP-POST instead of HTTP-Redirect when "HTTP-POST binding response" is enabled.

```
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="...truncated...
```

### Actual behavior

The ACS binding in the SP metadata always indicates HTTP-Redirect 
```
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="...truncated...
```

### How to Reproduce?

1. Create a new SAML identity provider
2. Specify IdP metadata
3. Verify that "HTTP-POST binding response" is disabled
4. View the SP metadata using the link in the Admin Console.  Note that the ACS binding is HTTP-Redirect
5. Toggle "HTTP-POST binding response" so that it's enabled
6. Save
7. View the SP metadata again using the link in the Admin Console.  Note that the ACS binding is still HTTP-Redirect

### Anything else?

A workaround which worked for me is to save the SP metadata, then manually change the ACS binding to "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" prior to uploading the SP metadata into the IdP.

### LINKED ISSUE #39748: Docs: server_admin/topics/clients/oidc/proc-using-a-service-account.adoc
### Before reporting an issue

### Area

docs

### Describe the bug

I'm using default Keycloak 26.2.4 with Client Credentials Grant ("Client authentication = ON" and "Service accounts roles = ON") using "Client Authenticator = Client Id and Secret".

In Server Administration, "13 Managing OpenID Connect and SAML Clients" ⇨ "Managing OpenID Connect clients" ⇨ "Confidential client credentials" (File: server_admin/topics/clients/oidc/proc-using-a-service-account.adoc) it says that the Client Id and Secret should be passed through "Authorization: Basic" header:

> By default, client credentials are represented by the clientId and clientSecret of the client in the Authorization: Basic header but you can also authenticate the client with a signed JWT assertion or any other custom mechanism for client authentication.
> 
> You also need to set the grant_type parameter to "client_credentials" as per the OAuth2 specification.
> 
> For example, the POST invocation to retrieve a service account can look like this:
> 
>     POST /realms/demo/protocol/openid-connect/token
>     Authorization: Basic cHJvZHVjdC1zYS1jbGllbnQ6cGFzc3dvcmQ=
>     Content-Type: application/x-www-form-urlencoded
> 
>     grant_type=client_credentials

However, by doing that I get `HTTP 401 Unauthorized` with `{"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}`.

But if I pass the `client_id` and `client_secret` as body parameters with the `grant_type`, the access key is returned. So, I believe Keycloak is using "client_secret_post" by default instead of "client_secret_basic" (https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) as the client authentication method.

### Version

26.2.4

### Regression

### Expected behavior

I would like the docs describe how to pass the `client_id` and `client_secret` accordingly to the client authentication method and gives me a link on how to change the client authentication method.

### Actual behavior

Passing Base64 encoded `client_id:client_secret` in `Authorization: Basic` headers gives HTTP 401 Unauthorized` with `{"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}`.

### How to Reproduce?

Follow the steps of Server Administration, "13 Managing OpenID Connect and SAML Clients" ⇨ "Managing OpenID Connect clients" ⇨ "Confidential client credentials" docs (File: server_admin/topics/clients/oidc/proc-using-a-service-account.adoc).

### Anything else?

_No response_

### LINKED ISSUE #39781: SMTP password overwritten with asterisks

### LINKED ISSUE #39934: Locale set to English even when only one Locale is enabled
### Before reporting an issue

- [x] I have read and understood the terms for submitting issues and acknowledge that my report may be closed without action if the guidelines are not followed.

---

### Area

`login/ui`

---

### Describe the bug

Hello,  

When only one locale is enabled in the realm settings and no default locale is explicitly set, the login page is rendered in **English**, even if English is not among the supported locales.

Screenshot showing the login page in English despite French being the only enabled locale

Image

---

### Version

`26.2.4`

---

### Regression

---

### Expected behavior

If only one locale is enabled in the realm, the login page should be rendered in that locale—regardless of whether a `ui_locales` query parameter is provided.

---

### Actual behavior

If the `ui_locales` query parameter is **not** provided, the login page defaults to **English**, even when English is **not** listed in `kcContext.locale.supported`.  
This leads to an inconsistent and confusing state:

- `kcContext.locale.currentLanguageTag` is `"en"` even though `"en"` is not in the supported locales.
- Users registering in this context end up with `"locale": "en"`, which is misleading.

---

### Steps to reproduce

```bash
cd ~/github
git clone https://github.com/keycloakify/keycloakify-starter
cd keycloakify-starter
yarn
npx keycloakify start-keycloak # Select Keycloak 26
```

Then:

1. Log in at `https://localhost:8080` with `admin:admin`
2. Navigate to **myrealm → Realm Settings → Localization**
3. Remove all locales except "French"
4. Save
5. Visit `https://my-theme.keycloakify.dev`

→ The page is rendered in English.

Open the browser console and inspect `window.kcContext.locale` to confirm that `currentLanguageTag` is `"en"` despite English not being supported.

---

### Additional context

This inconsistency may confuse users and could lead to unexpected behavior, especially when the locale `"en"` is used in user profiles despite being unavailable in the realm.

---

### Anything else?

Thanks for the excellent work on the project!

### LINKED ISSUE #39988: Move V1 specific methods from AbstractTokenExchangeProvider

### LINKED ISSUE #39999: Make changes for the browser flow

### LINKED ISSUE #40021: Passkeys conditional UI: integration with independent username and password form

### LINKED ISSUE #40022: Passkeys conditional UI: integration with the organization authenticator
### Description

Continuation of #29596 and #40021 to include passkeys / conditional UI for the organization authenticator. With these three issues all the default username authenticators are integrated with the conditional UI (auto fill) feature.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

### LINKED ISSUE #40049: Flaky test: org.keycloak.testsuite.webauthn.account.WebAuthnSigningInTest#checkAuthenticatorTimeLocale
## org.keycloak.testsuite.webauthn.account.WebAuthnSigningInTest#checkAuthenticatorTimeLocale
[Keycloak CI - WebAuthn IT (chrome)](https://github.com/keycloak/keycloak/actions/runs/15313329443/job/43082587525)

```
java.text.ParseException: Unparseable date: "May 29, 2025, 12:43 AM"
	at java.base/java.text.DateFormat.parse(DateFormat.java:399)
	at org.keycloak.testsuite.webauthn.account.WebAuthnSigningInTest.checkAuthenticatorTimeLocale(WebAuthnSigningInTest.java:320)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
...
```

### LINKED ISSUE #40104: NPE during external-internal token exchange in case that user exists

### LINKED ISSUE #40106: Two same tests in KcOidcBrokerTokenExchangeTest

### LINKED ISSUE #40167: Verification of external OIDC token by introspection-endpoint

### LINKED ISSUE #40237: Add option "Requires short state parameter" to OIDC IDP
### Description

When Keycloak has OIDC IDP provider configured and sends the initial OIDC authentication request to the OIDC IDP, it uses `state` parameter.

From the Keycloak/RHBK 26, the length of the `state` parameter is longer than in previous versions, which causes issues in case of some identity providers, which don't support very long `state` parameters. This is due the fact that Keycloak now encodes also "client data" into the `state` parameter, which means that it could be quite long. Related issue where this was added is https://github.com/keycloak/keycloak/issues/24112 .

This is the case at least for Microsoft EntraID and for SingPass (a Singapore identity provider for access to government online services). For Microsoft EntraID, there is possibility to use POST method (Related issue TODO ). For the SingPass, this may not help.

This issue is about adding the option like `Requires short state parameter` to the OIDC IDP (maybe also to the OAuth2 IDP, but not sure...). When this is enabled, there will not be an attempt to encode `client_data` into the `state` parameter, which is sent to the IDP.

Note there is the method `supportsLongStateParameter()` on the `IdentityProvider`, which currently always return true for all OAuth/OIDC providers. This issue means that providers, which has in the configuration `Requires short state parameter=true` will return false fro `supportsLongStateParameter()` .

NOTE: Having option `Requires short state parameter` means that option could be disabled by default and needs to be explicitly enabled, which is the desired behaviour compatible with previous versions when migrating from Keycloak 26 or Keycloak 26.2.

NOTE: Related test is `KcOidcMultipleTabsBrokerTest` where new test case might be added for the scenario with the `Requires short state parameter`. The side-effect of not including `client_data` is that when OIDC IDP redirects back to the KEycloak with OIDC authentication response, but Keycloak has authenticationSession already expired (which could be due the longer time or due the fact that user already authenticated to Keycloak in different browser tab), the Keycloak does not have `client_data` and hence it will need to display the error page instead of being able to redirect back to the client.

NOTE: It might be possible to use cookie and encode "client data" into that cookie, however not 100% sure as cookie does not work in multiple browser tabs and also will need some expiration etc. So maybe we can consider cookie rather as a follow-up if needed.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

### LINKED ISSUE #40380: Pin Snyk action to latest commit hash
### Description

The Snyk action was missed in the recent dependency pinning and is still referencing `@master`.

To ensure workflow consistency and stability, it must be pinned to a specific commit.

### LINKED ISSUE #40446: Review Profile makes users prone to phishing attacks

### LINKED ISSUE #40696: Remove the current conditionalUI authenticator

### LINKED ISSUE #40717: Allow passkeys login when user has no password credential

### LINKED ISSUE #40860: Size and validation options in SAML import
### Before reporting an issue

### Area

admin/ui

### Describe the bug

The issue #38620 modified the key generation to have the fields for size and validity but it seems that the same form is used for SAML import and those fields appear in the page (and they make no sense).

Image

### Version

999-SNAPSHOT

### Regression

### Expected behavior

Those fields should not be there.

### Actual behavior

The fields are there.

### How to Reproduce?

Just go to a SAML client and tab `keys` and click `Import key` button.

### Anything else?

_No response_

### LINKED ISSUE #40984: Backchannel logout token with an unexpected signature algorithm key
### Before reporting an issue

### Area

token-exchange

### Describe the bug

I am encountering an unexpected behavior during the single sign out implementation, related to the signature encryption algorithm for the backchannel logout token. All backchannel logout tokens are encrypted using the algorithm key from the last client that requested a token from the server, for all clients in the same session.

That is, consider there are two clients, A - ES256 signature key, and B - RS256 signature key, authenticated in the same user session. If last request to the server comes from client B, then both, client A and client B receives a backchannel logout token encrypted using the algorithm key RS256. 

The issue is that not all clients are able to implement both algorithms, producing errors for some clients in the same user session, when they get a token with an unexpected algorithm key. 

### Version

26.1.5

### Regression

### Expected behavior

For this use case, Keycloak should be able to send a backchannel logout token for each client, encrypted with their corresponding algorithm key. 

When a single sign out logout is started, each client should receive a backchannel logout token encrypted using a signature algorithm corresponding to their own configuration. Client A receives a token encrypted with a ES256 algorithm, while client B receives a token encrypted with a RS256 algorithm. 

### Actual behavior

When a single sign out logout is started, both clients receive a backchannel logout token encrypted using the signature algorithm corresponding to last client that called the server. Client A and client B receives a token encrypted with a RS256 algorithm, or both clients receives a token encrypted with a ES256 algorithm.

### How to Reproduce?

Keycloak configuration:
- Configure one client A with "Access token signature algorithm" as ES256;
- Configure one client B with "Access token signature algorithm" as RS256;
- Realm default signature algorithm is RS256.

Execution flow:
- Authenticate in the client A application;
- Open client B application, using a single sign on implementation;
- Refresh client A application page;
- Disconnect from client A application;
- Check the backchannel logout token.

When the user disconnects from client A (ES256), both clients get a backchannel logout token using a signature algorithm ES256, as follow:

Token client A: 

{
  "alg": "ES256",
  "typ": "logout+jwt",
  "kid": "9lf4bogHZdRuyGNOYucQhQ-ODIlE2-qBUTSYc8_wUFE"
}

eyJhbGciOiJFUzI1NiIsInR5cCIgOiAibG9nb3V0K2p3dCIsImtpZCIgOiAiOWxmNGJvZ0haZFJ1eUdOT1l1Y1FoUS1PRElsRTItcUJVVFNZYzhfd1VGRSJ9.eyJleHAiOjE3NTE5MTAzMTYsImlhdCI6MTc1MTkxMDE5NiwianRpIjoiMDRkNGNmZGItMjU0Ny00YmNhLTk5M2YtNTQyOTBiMTQ3ZWExIiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL3JlYWxtcy9zcWluIiwiYXVkIjoicGVzLWYiLCJzdWIiOiIzNjYxNWQ3MC1hMjcwLTQ4M2UtOGJhOS02YzNiOGVkYTQ2NGEiLCJ0eXAiOiJMb2dvdXQiLCJzaWQiOiJmNDFmZmZmYS1hMDI2LTRmMGMtYTc3MS1kZGEzYWI0ZWY3OGUiLCJldmVudHMiOnsiaHR0cDovL3NjaGVtYXMub3BlbmlkLm5ldC9ldmVudC9iYWNrY2hhbm5lbC1sb2dvdXQiOnt9fX0.9cf2AwqZdehLDkWO3nVRgmSa-k_NNgGnUg_cGl1ZjbOz3qOj5Dn1YsZze0yb1sBT27ziPSScNRmdl6KTzkh0hg

Token client B:

{
  "alg": "ES256",
  "typ": "logout+jwt",
  "kid": "9lf4bogHZdRuyGNOYucQhQ-ODIlE2-qBUTSYc8_wUFE"
}

eyJhbGciOiJFUzI1NiIsInR5cCIgOiAibG9nb3V0K2p3dCIsImtpZCIgOiAiOWxmNGJvZ0haZFJ1eUdOT1l1Y1FoUS1PRElsRTItcUJVVFNZYzhfd1VGRSJ9.eyJleHAiOjE3NTE5MTAzMTYsImlhdCI6MTc1MTkxMDE5NiwianRpIjoiNDg0YWZlZjEtMjllYy00NWY3LWFmMzEtOWU4ZGNhZTc3MDk1IiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL3JlYWxtcy9zcWluIiwiYXVkIjoicGVzLWciLCJzdWIiOiIzNjYxNWQ3MC1hMjcwLTQ4M2UtOGJhOS02YzNiOGVkYTQ2NGEiLCJ0eXAiOiJMb2dvdXQiLCJzaWQiOiJmNDFmZmZmYS1hMDI2LTRmMGMtYTc3MS1kZGEzYWI0ZWY3OGUiLCJldmVudHMiOnsiaHR0cDovL3NjaGVtYXMub3BlbmlkLm5ldC9ldmVudC9iYWNrY2hhbm5lbC1sb2dvdXQiOnt9fX0.ITWuqMeGaMDouJ8nAIcoWVbL1FEPEZ-qf-hkVo-2mcEqwUqyplyPZg-QYk205_wuhiJiHD9cTF-iyIjfs1ifYQ

On the other hand, when the user refreshs the page and disconnects from client B (RS256), both clients get a backchannel logout token using a signature algorithm RS256, as follow:

Token  client A:

{
  "alg": "RS256",
  "typ": "logout+jwt",
  "kid": "RfBySy3U9KtGUABmzJSlGbqwFYdeqHnrt-nJG5A83SU"
}

eyJhbGciOiJSUzI1NiIsInR5cCIgOiAibG9nb3V0K2p3dCIsImtpZCIgOiAiUmZCeVN5M1U5S3RHVUFCbXpKU2xHYnF3RllkZXFIbnJ0LW5KRzVBODNTVSJ9.eyJleHAiOjE3NTE5MTU2NTYsImlhdCI6MTc1MTkxNTUzNiwianRpIjoiZjEwNDViOWItZWY1My00ZTFmLTk3YTQtYWQyMzBkNzQ4OTllIiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL3JlYWxtcy9zcWluIiwiYXVkIjoicGVzLWYiLCJzdWIiOiIzNjYxNWQ3MC1hMjcwLTQ4M2UtOGJhOS02YzNiOGVkYTQ2NGEiLCJ0eXAiOiJMb2dvdXQiLCJzaWQiOiI1NWM1YjA2MS01YTA3LTQzNDYtYmM3Ny0yOTM3ZmEwNDRiOWIiLCJldmVudHMiOnsiaHR0cDovL3NjaGVtYXMub3BlbmlkLm5ldC9ldmVudC9iYWNrY2hhbm5lbC1sb2dvdXQiOnt9fX0.CxdRQcsIRNAKz_ZfwwU9X85eySDU-KBHF1MIgBkmF_3KB-WLoLZBiRPzq8sqHTGM6WfVCcjn_bVd0mJ6ToMlmw3Pr0m4Ye_SM2rvGzOtex1W3FitVlGg3P_FF-Z8tr11ICc_-yZLQUyhgl7Z_i_vVV0mCoGKak5mQ3gB1PyT_eghdvJ8L6it9YNia401OvDpgalmGYeCs0ISy-FVgOTCniZXPTbA3GmflfjTFz0-kjAZy-wvy2u022GSYIt0NxBH3Ym5I-A-ADQ7iqfWacUBu4We8fC0_r1CeMWNC1mdUlYi3VhsL4QcE4sEOaNdH7upbhcm5Dv6mK8ZsZtieKTRLrnTZ27_jArR9uM1_oBygYdJvM1Tkvh_ivn2fX3SPhUM6JGUn8NAIKbc9QOg4Gmih2GCbmOP2geuchm0W4Zhu_GWTie-0Q2Wy2J3g8YRsa8uneT7bLkh_uq60EZFe0TmW1fBfh0uqzB6uYoRWSxix0ZHBSnYjMd710scgOquovg9VHDHwpTvrRyJMge-PV5B9TObhht3JJ6Ygd3C6Kz_4ZnS7F_Xr9zSbZUs4-U2Hc8HjQ3HIiG0VvbeCwuwseU55rwQ0hX13D-_TWzLINNJN57DRcOU3XjS1n-hlpelNo74cWgMfklnFcAa4FbXD_ilydmij3Y4aAFiQ4R_cy2gZAc

Token client B:

{
  "alg": "RS256",
  "typ": "logout+jwt",
  "kid": "RfBySy3U9KtGUABmzJSlGbqwFYdeqHnrt-nJG5A83SU"
}

eyJhbGciOiJSUzI1NiIsInR5cCIgOiAibG9nb3V0K2p3dCIsImtpZCIgOiAiUmZCeVN5M1U5S3RHVUFCbXpKU2xHYnF3RllkZXFIbnJ0LW5KRzVBODNTVSJ9.eyJleHAiOjE3NTE5MTU2NTYsImlhdCI6MTc1MTkxNTUzNiwianRpIjoiODg3OWQ0ZDgtMGEzMi00YTMxLWI2MWMtY2NkMzFiYzBlOWM2IiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL3JlYWxtcy9zcWluIiwiYXVkIjoicGVzLWciLCJzdWIiOiIzNjYxNWQ3MC1hMjcwLTQ4M2UtOGJhOS02YzNiOGVkYTQ2NGEiLCJ0eXAiOiJMb2dvdXQiLCJzaWQiOiI1NWM1YjA2MS01YTA3LTQzNDYtYmM3Ny0yOTM3ZmEwNDRiOWIiLCJldmVudHMiOnsiaHR0cDovL3NjaGVtYXMub3BlbmlkLm5ldC9ldmVudC9iYWNrY2hhbm5lbC1sb2dvdXQiOnt9fX0.gxDrRnrXPHBlHGadKj9QbZqrr4JUiGGpFqHwOBJUK_uEGsiQiCIt0SjWZLCf0qs-B74cayv-PW8QYTIYoC3EjkdkfsKJ8G8F_FugWWgw3Rp0mWLa0myNRs3r4z5_xm8K_NFvbO4utBoJW23SAOYbsYT7Sg3E1oqKULFKp9B-LF5fdTlI_5JaNwJCAOBddbimNniKGJE5u_Wd3_XUuYvAR9Nf_iAmd5Mk2Op46wMKFVUlLl-l43JfRh-flBAZDST3khVQ0A8Zi9PCYw3PpMtwMKhEw-2fiD9riJ54R9Tx4u6g1gIayFxrsbVm_iOV1IRHPlP_JTMZta7lMDvgFQqW48AlBHvL3hyqZmH3lty0hmT0zLo77_lIC6nQGRTl6N8WMleEWGItYPQU8TFaifn7KbO17OTWl_NGnouYSUkCFn7qNpvYchVZIzEZotd17KdSwI7epSU2pNNdEqRdTNusY-4ugvD-ysBIEah8QkFqJo2ip_PU7YsCAUeilmZ-ocm-3XBeNAIu-CQepZAbEsdVxw30C_CgtFdGnkxoHrgLtqEAw3RUvW2p0F7ZFXv1n_70DSQxn48gYYb-p4ZkZj-5SZA3bn7wbWqyqRoRIVzniXdyC6uzEt237FTYn_BM6LwUOB13K5rfgKxVbn2LlXb0ZApLOXpVKhy23UkqQCWEp9g

### Anything else?

_No response_

### LINKED ISSUE #40995: LDAP /  ModelException: At least one condition should be provided to OR query
### Before reporting an issue

### Area

core

### Describe the bug

Hi,

we have a regression after upgrading to keycloak 26.3.0 (previous version was 26.2.5, same ldap conf and no issue).
We have a user federation with ldap configured.
After the authentication, when the user call CODE_TO_TOKEN, we get an error 500 with this stack:

```
2025-07-08 08:18:42,855 logLevel=ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-82) Uncaught server error: org.keycloak.models.ModelException: At least one condition should be provided to OR query
	at org.keycloak.storage.ldap.idm.query.internal.LDAPQueryConditionsBuilder.orCondition(LDAPQueryConditionsBuilder.java:58)
	at org.keycloak.storage.ldap.mappers.membership.UserRolesRetrieveStrategy$GetRolesFromUserMemberOfAttribute.getLDAPRoleMappings(UserRolesRetrieveStrategy.java:109)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper.getLDAPGroupMappings(GroupLDAPStorageMapper.java:634)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper$LDAPGroupMappingsUserDelegate.getLDAPGroupMappingsConverted(GroupLDAPStorageMapper.java:778)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper$LDAPGroupMappingsUserDelegate.getGroupsStream(GroupLDAPStorageMapper.java:711)
	at org.keycloak.models.utils.UserModelDelegate.getGroupsStream(UserModelDelegate.java:234)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper$LDAPGroupMappingsUserDelegate.getGroupsStream(GroupLDAPStorageMapper.java:717)
	at org.keycloak.models.cache.infinispan.entities.CachedUser.lambda$new$3(CachedUser.java:75)
	at org.keycloak.models.cache.infinispan.DefaultLazyLoader.lambda$get$0(DefaultLazyLoader.java:52)
	at org.keycloak.authorization.fgap.AdminPermissionsSchema.runWithoutAuthorization(AdminPermissionsSchema.java:497)
	at org.keycloak.models.cache.infinispan.DefaultLazyLoader.get(DefaultLazyLoader.java:49)
	at org.keycloak.models.cache.infinispan.entities.CachedUser.getGroups(CachedUser.java:131)
	at org.keycloak.models.cache.infinispan.UserAdapter.getGroupsStream(UserAdapter.java:426)
	at org.keycloak.models.UserModel.getGroupsStream(UserModel.java:180)
	at org.keycloak.services.resources.admin.UserResource.groupMembership(UserResource.java:1115)
	at org.keycloak.services.resources.admin.UserResource$quarkusrestinvoker$groupMembership_7205dccea6655b8c59b771d74abc3c0bd11f433f.invoke(Unknown Source)
	at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:141)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:638)
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2675)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2654)
	at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1627)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1594)
```

### Version

26.3.0

### Regression

### Expected behavior

The CODE_TO_TOKEN endpoint should return a token as previously.

### Actual behavior

After authentication, the CODE_TO_TOKEN call returns a 500 http code with the following exception:

```
2025-07-08 08:18:42,855 logLevel=ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-82) Uncaught server error: org.keycloak.models.ModelException: At least one condition should be provided to OR query
	at org.keycloak.storage.ldap.idm.query.internal.LDAPQueryConditionsBuilder.orCondition(LDAPQueryConditionsBuilder.java:58)
	at org.keycloak.storage.ldap.mappers.membership.UserRolesRetrieveStrategy$GetRolesFromUserMemberOfAttribute.getLDAPRoleMappings(UserRolesRetrieveStrategy.java:109)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper.getLDAPGroupMappings(GroupLDAPStorageMapper.java:634)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper$LDAPGroupMappingsUserDelegate.getLDAPGroupMappingsConverted(GroupLDAPStorageMapper.java:778)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper$LDAPGroupMappingsUserDelegate.getGroupsStream(GroupLDAPStorageMapper.java:711)
	at org.keycloak.models.utils.UserModelDelegate.getGroupsStream(UserModelDelegate.java:234)
	at org.keycloak.storage.ldap.mappers.membership.group.GroupLDAPStorageMapper$LDAPGroupMappingsUserDelegate.getGroupsStream(GroupLDAPStorageMapper.java:717)
	at org.keycloak.models.cache.infinispan.entities.CachedUser.lambda$new$3(CachedUser.java:75)
	at org.keycloak.models.cache.infinispan.DefaultLazyLoader.lambda$get$0(DefaultLazyLoader.java:52)
	at org.keycloak.authorization.fgap.AdminPermissionsSchema.runWithoutAuthorization(AdminPermissionsSchema.java:497)
	at org.keycloak.models.cache.infinispan.DefaultLazyLoader.get(DefaultLazyLoader.java:49)
	at org.keycloak.models.cache.infinispan.entities.CachedUser.getGroups(CachedUser.java:131)
	at org.keycloak.models.cache.infinispan.UserAdapter.getGroupsStream(UserAdapter.java:426)
	at org.keycloak.models.UserModel.getGroupsStream(UserModel.java:180)
	at org.keycloak.services.resources.admin.UserResource.groupMembership(UserResource.java:1115)
	at org.keycloak.services.resources.admin.UserResource$quarkusrestinvoker$groupMembership_7205dccea6655b8c59b771d74abc3c0bd11f433f.invoke(Unknown Source)
	at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:141)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:638)
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2675)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2654)
	at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1627)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1594)
```

### How to Reproduce?

Configure a ldap user federation with group mapping

### Anything else?

_No response_

### LINKED ISSUE #41038: FIPS errors in CI

### LINKED ISSUE #41259: Passkeys support in IdpUsernamePasswordForm

### LINKED ISSUE #41354: Avoid OTP when logging in with passkey

### LINKED ISSUE #41537: Getting error 405 "Method Not Allowed" when calling the "certs" endpoint with HEAD method

### LINKED ISSUE #41556: Switch passkeys to supported

### LINKED ISSUE #41643: Test SMTP connection fails when no port is specified

### LINKED ISSUE #41837: Remove OIDCLoginProtocolService.certsHead()
### Description

This is just a minor issue to remove `OIDCLoginProtocolService.certsHead()` method.

It was added as a workaround to https://github.com/quarkusio/quarkus/issues/49172 . Once we updated to the corresponding quarkus version 3.25.2 where this is fixed, there is no need to this method anymore.

See the https://github.com/keycloak/keycloak/issues/41537 where also the test was added (Test is still kept to make sure that fix in quarkus helped to fix the issue).

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

### LINKED ISSUE #41854: KeycloakSession javadoc references keycloak-server.json
### Before reporting an issue

### Area

docs

### Describe the bug

KeycloakSession references keycloak-server.json, which is not valid for the quarkus distribution.

### Version

lastest

### Regression

### Expected behavior

The javadoc doesn't need to mention a specific config file.

### Actual behavior

The javadoc mentions an invalid config file.

### How to Reproduce?

N/A

### Anything else?

_No response_

### LINKED ISSUE #41909: Admin console provider info shows "Add providers"
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Provider info has the wrong header name for the providers column - "Add providers", but should be "Providers"

Also the spi / providers are shown in raw name form, which may be addressed by #41676 

### Version

latest

### Regression

### Expected behavior

Should say Providers.

### Actual behavior

Says "Add providers"

### How to Reproduce?

N/A

### Anything else?

_No response_

### LINKED ISSUE #42305: Document that single-cluster deployments expect all Keycloak instances to serve traffic
### Description

Document that single-cluster deployments expect all Keycloak instances to serve traffic

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #42305

### DIFF
--- a/docs/guides/high-availability/single-cluster/introduction.adoc
+++ b/docs/guides/high-availability/single-cluster/introduction.adoc
@@ -11,7 +11,7 @@ summary="Deploy a single Keycloak cluster, optionally stretched across multiple 
 The {project_name} single-cluster setup is targeted at use cases that:
 
 * Deploy to an infrastructure with transparent networking, like for example a single {kubernetes} cluster.
-* Desire all healthy ${project_name} instances to handle user requests.
+* Desire all healthy {project_name} instances to handle user requests.
 * Are constrained to a single
 <@profile.ifProduct>
 AWS Region.
--- a/docs/guides/high-availability/single-cluster/introduction.adoc
+++ b/docs/guides/high-availability/single-cluster/introduction.adoc
@@ -11,6 +11,7 @@ summary="Deploy a single Keycloak cluster, optionally stretched across multiple 
 The {project_name} single-cluster setup is targeted at use cases that:
 
 * Deploy to an infrastructure with transparent networking, like for example a single {kubernetes} cluster.
+* Desire all healthy ${project_name} instances to handle user requests.
 * Are constrained to a single
 <@profile.ifProduct>
 AWS Region.


docs/guides/high-availability/single-cluster/introduction.adoc
```diff
@@ -11,6 +11,7 @@ summary="Deploy a single Keycloak cluster, optionally stretched across multiple
 The {project_name} single-cluster setup is targeted at use cases that:
 
 * Deploy to an infrastructure with transparent networking, like for example a single {kubernetes} cluster.
+* Desire all healthy ${project_name} instances to handle user requests.
```
ryanemerson: @ahus1 I changed this from "all nodes serve traffic" as "node" is overloaded in the context of Kubernetes and ultimately it's user requests being routed to the instances. Happy to revert though if you have strong opinions on this.

ahus1:
```suggestion
* Desire all healthy {project_name} instances to handle user requests.
```

ryanemerson:
Thanks for the fix @ahus1, too much bash for me today it seems :slightly_smiling_face:

ahus1 (approved):

PR #42310, opened 2025-09-03
fix: allows for schedulings to be defined for operator jobs

### LINKED ISSUE #42057: [Operator] Update job incorrectly inherits podTemplate configuration from unsupported.podTemplate
## Issue Summary

When using the `unsupported.podTemplate` configuration for the main Keycloak pods, the update job incorrectly inherits the same `nodeSelector` and `tolerations` configuration, preventing separate control over update job scheduling.

## Current Behavior
1. Main Keycloak pods use `unsupported.podTemplate` configuration ✅
2. Update job automatically inherits the same `nodeSelector` and `tolerations` ❌
3. No way to configure update job scheduling separately ❌

## Expected Behavior
The Keycloak operator appears to be applying the `unsupported.podTemplate` configuration to both the main StatefulSet and the update job, instead of allowing separate configuration for each.

## Impact
- Update jobs cannot be scheduled on **different nodes** than the main Keycloak pods
- Cannot use different tolerations for update jobs vs. main pods
- Limits flexibility in cluster resource management
- May cause update job failures if the inherited configuration is inappropriate

## Environment
- Keycloak Operator Version: 26.3.3
- Kubernetes Version: 1.33.0
- Use Case: Multi-node cluster with dedicated nodes for different workloads

## Steps to Reproduce
1. Deploy Keycloak with unsupported.podTemplate configuration:
```
kind: Keycloak
metadata:
  name: keycloak
spec:
  # ... other configuration ...
  unsupported:
    podTemplate:
      spec:
        nodeSelector:
          node-role: "keycloak-dedicated"
        tolerations:
          - key: "dedicated"
            operator: "Equal"
            value: "keycloak"
            effect: "NoSchedule"
```
2. Trigger an update (e.g., change image tag, configuration)
3. Observe that the update job uses the same nodeSelector and tolerations
4. The update job cannot be scheduled because this node group have more than 3 nodes (keycloak used 3 nodes)

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

shawkins:
closes: #42057

It's probably worth discussing, but we actually have several options here:

- this pr - keep the current behavior that defaults the job scheduling to that of the server, and allow that to be overriden by specific scheduling fields. 

- An alternative to putting the import scheduling on the import cr is to instead put that on the keycloak cr - it's highly likely that users would want the same scheduling for all imports.

- Add a declarative field on or beside the scheduling spec to state whether the scheduling should be inherited - it would default to true as to not be a breaking change, but we could later default it false.

### DIFF
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -290,7 +290,7 @@ spec:
 
 Please see https://kubernetes.io/docs/concepts/scheduling-eviction[the kubernetes docs] for more on scheduling concepts.
 
-If you do not specify a custom topology spread constraint, your Pods will have a spread constraint across zones and nodes to improve availability.
+If you do not specify a custom topology spread constraint, your Pods will have a spread constraint across zones and nodes to improve availability. These defaults do not affect import and update Jobs, only server Pods.
 
 ==== Job Scheduling
 
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -290,7 +290,26 @@ spec:
 
 Please see https://kubernetes.io/docs/concepts/scheduling-eviction[the kubernetes docs] for more on scheduling concepts.
 
-If you do not specify a custom affinity, your Pods will have an anti-affinity for the same zone and node to improve availability.
+If you do not specify a custom topology spread constraint, your Pods will have a spread constraint across zones and nodes to improve availability.
+
+==== Job Scheduling
+
+Realm import and update Jobs inherit the scheduling of the server Pods if it is specified in the Keycloak CR `spec.scheduling` field. If your server Pod scheduling is highly restrictive this may result in the 
+Job Pods being unschedulable. In this situation you may override the Job scheduling using the Keycloak CR `spec.update.scheduling` and `spec.import.scheduling` fields.
+ 
+For example the following will remove the use of any server Pod scheduling for the update Job: 
+ 
+[source,yaml]
+----
+apiVersion: k8s.keycloak.org/v2alpha1
+kind: Keycloak
+metadata:
+  name: example-kc
+spec:
+  update:
+    scheduling: {}
+  ...
+----
 
 === Management Interface
 
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
@@ -18,6 +18,7 @@ package org.keycloak.operator.controllers;
 
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.EnvVarBuilder;
+import io.fabric8.kubernetes.api.model.PodSpec;
 import io.fabric8.kubernetes.api.model.PodTemplateSpec;
 import io.fabric8.kubernetes.api.model.SecretVolumeSourceBuilder;
 import io.fabric8.kubernetes.api.model.Volume;
@@ -35,11 +36,15 @@ import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDep
 import org.keycloak.operator.Config;
 import org.keycloak.operator.ContextUtils;
 import org.keycloak.operator.Utils;
+import org.keycloak.operator.crds.v2alpha1.deployment.Keycloak;
+import org.keycloak.operator.crds.v2alpha1.deployment.spec.ImportSpec;
+import org.keycloak.operator.crds.v2alpha1.deployment.spec.SchedulingSpec;
 import org.keycloak.operator.crds.v2alpha1.realmimport.KeycloakRealmImport;
 import org.keycloak.operator.crds.v2alpha1.realmimport.Placeholder;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import static org.keycloak.operator.Utils.addResources;
@@ -53,7 +58,7 @@ public class KeycloakRealmImportJobDependentResource extends KubernetesDependent
     }
 
     @Override
-    protected Job desired(KeycloakRealmImport primary, Context<KeycloakRealmImport> context) {
+    public Job desired(KeycloakRealmImport primary, Context<KeycloakRealmImport> context) {
         Config config = ContextUtils.getOperatorConfig(context);
         StatefulSet existingDeployment = ContextUtils.getCurrentStatefulSet(context).orElseThrow();
         Map<String, Placeholder> placeholders = primary.getSpec().getPlaceholders();
@@ -73,6 +78,9 @@ public class KeycloakRealmImportJobDependentResource extends KubernetesDependent
 
         // The Job should not be selected with app=keycloak
         labels.put("app", "keycloak-realm-import");
+
+        var kc = ContextUtils.getKeycloak(context);
+        handleJobScheduling(kc, Optional.ofNullable(kc.getSpec().getImportSpec()).map(ImportSpec::getSchedulingSpec), keycloakPodTemplate.getSpec());
 
         var envvars = keycloakPodTemplate
                 .getSpec()
@@ -161,5 +169,15 @@ public class KeycloakRealmImportJobDependentResource extends KubernetesDependent
         keycloakContainer.setStartupProbe(null);
 
         addResources(keycloakRealmImport.getSpec().getResourceRequirements(), config, keycloakContainer);
+    }
+
+    static void handleJobScheduling(Keycloak keycloak, Optional<SchedulingSpec> schedulingSpec, PodSpec spec) {
+        if (schedulingSpec.isPresent() || keycloak.getSpec().getSchedulingSpec() == null) {
+            spec.setPriorityClassName(schedulingSpec.map(SchedulingSpec::getPriorityClassName).orElse(null));
+            spec.setAffinity(schedulingSpec.map(SchedulingSpec::getAffinity).orElse(null));
+            spec.setTolerations(schedulingSpec.map(SchedulingSpec::getTolerations).orElse(null));
+            spec.setTopologySpreadConstraints(schedulingSpec.map(SchedulingSpec::getTopologySpreadConstraints).orElse(null));
+        }
+        // else use the parent values
     }
 }
--- a/operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
+++ b/operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
@@ -27,6 +27,7 @@ import org.keycloak.operator.crds.v2alpha1.deployment.spec.FeatureSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.HostnameSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.HttpManagementSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.HttpSpec;
+import org.keycloak.operator.crds.v2alpha1.deployment.spec.ImportSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.IngressSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.NetworkPolicySpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.ProbeSpec;
@@ -123,6 +124,10 @@ public class KeycloakSpec {
     @JsonProperty("scheduling")
     @JsonPropertyDescription("In this section you can configure Keycloak's scheduling")
     private SchedulingSpec schedulingSpec;
+
+    @JsonProperty("import")
+    @JsonPropertyDescription("In this section you can configure import Jobs")
+    private ImportSpec importSpec;
 
     @JsonProperty("bootstrapAdmin")
     @JsonPropertyDescription("In this section you can configure Keycloak's bootstrap admin - will be used only for initial cluster creation.")
@@ -360,4 +365,13 @@ public class KeycloakSpec {
     public void setStartupProbeSpec(ProbeSpec startupProbeSpec) {
         this.startupProbeSpec = startupProbeSpec;
     }
+
+    public ImportSpec getImportSpec() {
+        return importSpec;
+    }
+
+    public void setImportSpec(ImportSpec importSpec) {
+        this.importSpec = importSpec;
+    }
+
 }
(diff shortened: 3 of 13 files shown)


operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
```diff
@@ -162,4 +166,13 @@ private void buildKeycloakJobContainer(Container keycloakContainer, KeycloakReal
 
         addResources(keycloakRealmImport.getSpec().getResourceRequirements(), config, keycloakContainer);
     }
+
+    static void handleJobScheduling(SchedulingSpec schedulingSpec, PodSpec spec) {
```
pruivo: nitpicking; wouldn't make more sense for this method to be in `CRDUtils`?

pruivo:
> It's probably worth discussing, but we actually have several options here:
> 
> * this pr - keep the current behavior that defaults the job scheduling to that of the server, and allow that to be overriden by specific scheduling fields.
> * An alternative to putting the import scheduling on the import cr is to instead put that on the keycloak cr - it's highly likely that users would want the same scheduling for all imports.
> * Add a declarative field on or beside the scheduling spec to state whether the scheduling should be inherited - it would default to true as to not be a breaking change, but we could later default it false.

I don't have a strong preference, but this pr/first option seems to be the simplest of all.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
shawkins: Those utility methods seem to be more about finding things on a cr, not applying changes to kubernetes objects.

ryanemerson:
I like the separation of scheduling definitions on the two CRs, I think it is the most flexible solution, so +1 to the current approach to the specs.

I am not convinced it makes sense for the import and upgrade jobs to default to the Keycloak scheduling rules though.

Currently the default `TopologySpreadConstraints` prefer that pods are scheduled on distinct zones and nodes. Imagine a K8s cluster with multiple zones, but a single Keycloak instance. I don't think it makes sense that the upgrade/import jobs be provisioned in a separate zone as this will cause additional latency and cost.

I can think of two solutions to prevent ^:

1. Don't define scheduling defaults for these jobs. Simple, but jobs may still be scheduled across zones.

2. Define default scheduling for import and upgrade jobs that prefers co-location via an affinity rule.

I think 2. is the more robust solution. 

Is there any scenario where wouldn't prefer the jobs to be co-located?

shawkins:
> Is there any scenario where wouldn't prefer the jobs to be co-located?

I don't think so. At least in terms of zones that was effectively the old default behavior - which means we can go back to that without any concerns about breaking.

> I think 2. is the more robust solution.

What about when the user has put explicit scheduling on the Keycloak CR - do we just break the old behavior of inheriting, or add another flag to control that?

ryanemerson:
> What about when the user has put explicit scheduling on the Keycloak CR - do we just break the old behavior of inheriting, or add another flag to control that?

Personally I don't like adding additional flags unless absolutely necessary, as it's one more thing for the user to consider.

Can we add this to notable changes and document that by default import/upgrade jobs will be co-located unless explicitly configured otherwise?

shawkins:
> Can we add this to notable changes and document that by default import/upgrade jobs will be co-located unless explicitly configured otherwise?

@vmuzikar we didn't explicitly document the that Jobs inherit scheduling from the Keycloak CR, so do you see this as a notable change - or a breaking change that would necessitate a flag in a minor release?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
vmuzikar: @pruivo +1
Was about to write the same comment. :)

> Those utility methods seem to be more about finding things on a cr, not applying changes to kubernetes objects.

It's utils, I don't see a reason why it shouldn't modify stuff. Though it's not blocking.

vmuzikar (dismissed):
@shawkins Thanks, LGTM.

vmuzikar:
> @vmuzikar we didn't explicitly document the that Jobs inherit scheduling from the Keycloak CR, so do you see this as a notable change - or a breaking change that would necessitate a flag in a minor release?

We're not changing existing behaviour, just adding new fields to override it, no? If anything, it should be included in the Release Notes.

I'd keep the current behaviour as the default (that it inherits scheduling from main deployment).

vmuzikar:
> but a single Keycloak instance

This would not seem production ready – a single KC pod. :) I doubt that would be too common.

shawkins:
> We're not changing existing behaviour, just adding new fields to override it, no? 

What @ryanemerson is proposing not in this pr and would change existing behavior - but it is not explicitly documented.

> I'd keep the current behaviour as the default (that it inherits scheduling from main deployment).

That is the current state of the pr. However with the recent changes to the main default scheduling it is not likely the behavior users want by default for the jobs.

vmuzikar:
> However with the recent changes to the main default scheduling it is not likely the behavior users want by default for the jobs.

That's what I'm not entirely sure about. If the user expects some scheduling policy, it might be counter intuitive if it's not reflected by all KC pods (for the user, it might be low level tech details if it's a Job or main KC deployment). Even worse if the user explicitly set the scheduling spec for the main deployment.

vmuzikar:
> Is there any scenario where wouldn't prefer the jobs to be co-located?

In strictly resourced envs where the user doesn't expect more than 1 KC pod on a node.

ryanemerson:
> This would not seem production ready – a single KC pod. :) I doubt that would be too common.

True! I said a single pod for simplicity, but the scenario I describe is also applicable if a Kubernetes cluster is deployed across 3 availability-zones and the user provisions Keycloak with `spec.instances: 2`. If we inherit the default topologyPodConstraints in the jobs, this will result in the following distribution:

- `keycloak-0`: AZ1
- `keycloak-1`: AZ2
- job: AZ3

vmuzikar:
@ryanemerson Right. So the user is expecting to have a single KC pod per zone (for whatever reasons) which can be either default or explicitly set behaviour. Now, won't it be confusing to the user that suddenly a new "KC pod" appeared in the same AZ when some other AZ has 0 pods? Additionally, what would be the disadvantage of scheduling the Jobs in a different AZ? Neither the Import Job, nor the Update Job need to communicate with the other KC Pods, no? One reason might be the DB latency but that shouldn't be an issue if the scheduling actually allows the KC Pods to be scheduled in various AZs.

ryanemerson:
> Now, won't it be confusing to the user that suddenly a new "KC pod" appeared in the same AZ when some other AZ has 0 pods?

Presumably the pods have a different naming convention as they're not part of the StatefulSet, so I think that would make the difference clear.

> Additionally, what would be the disadvantage of scheduling the Jobs in a different AZ? Neither the Import Job, nor the Update Job need to communicate with the other KC Pods, no?

That's a good point about the pod communication. When I wrote https://github.com/keycloak/keycloak/pull/42310#issuecomment-3252831668, I had made some incorrect assumptions about how the import job worked. So you're right, network cost for cross-zone traffic is not a concern here.

It's minor, but one disadvantage I can think of is resource allocation. Inheriting the default policies means that more weight will be attached to scheduling the job pod to a distinct AZ/node then a node that has more available resources.

Very much an edge-case, but job execution and simultaneous scaling of Keycloak `spec.instances` could also lead to a scenario where a keycloak server pod is not provisioned on a distinct az. Consider the scenario described in https://github.com/keycloak/keycloak/pull/42310#issuecomment-3265148434 again, if the job is on AZ3, it could be that `keycloak-2` then gets provisioned on AZ1 or AZ2.

> One reason might be the DB latency but that shouldn't be an issue if the scheduling actually allows the KC Pods to be scheduled in various AZs.

+1. If the user wants to co-locate an import job with the DB writer instance, they will now have the flexibility in `keycloakrealmimports` to do that. That will be deployment dependent though, so I don't think we can further improve this in a generic way.

vmuzikar:
@ryanemerson Considering the past few points in the discussion, do you still see the co-location as the preferred/correct scheduling strategy for the Jobs?

I'd still see it as a bit counter intuitive, especially in case the user explicitly specifies the scheduling for the main deployment in the CR – as a user, I'd probably expect it to be reflected by the Jobs too, and took the Job specific scheduling as an override for the main one. I guess I'm more taking it from the UX perspective rather than the nuanced performance one.

ryanemerson:
> Considering the past few points in the discussion, do you still see the co-location as the preferred/correct scheduling strategy for the Jobs?

No, I don't think explicit co-location brings any real benefit.

Given the edge-case that I described above regarding scheduling and our default topologyPodConstraints policy, I'm still not convinced that it makes sense for jobs to inherit the Keycloak CR scheduling configuration.

That being said, as users can now override the configuration, I don't have strong views on this and I'm happy to go with cloud-native team's preference.

vmuzikar:
> I'm happy to go with cloud-native team's preference.

@shawkins Are you ok with this as is, too?

shawkins:
Talked about this in our meeting. It seems like the most straight-forward UX is to make our best effort to make things work as much as expected by default and be done in only a single place, so we should not inherit the default scheduling and we'll move the scheduling of import jobs to the keycloak cr.

@vmuzikar @ryanemerson @pruivo we also wanted to confirm there's no chance the switch from non-stretch to stretch clusters could be a breaking change for any existing users?

ryanemerson:
> @vmuzikar @ryanemerson @pruivo we also wanted to confirm there's no chance the switch from non-stretch to stretch clusters could be a breaking change for any existing users?

You mean the new default `podTopologySpreadConstraints` vs the old anti-affinity rules? It shouldn't impact functionality in the sense that everything will work as before, but if the user is running a multi-az k8s cluster then pods will now be scheduled on distinct zones opposed to being co-located within the same az.

The changes doc is outdated, as it still talks about anti-affinity rules. If you're happy with above change in behaviour, then I will create a PR to fix it.

shawkins:
Updated the pr based upon my last comment.

> You mean the new default podTopologySpreadConstraints vs the old anti-affinity rules? It shouldn't impact functionality in the sense that everything will work as before, but if the user is running a multi-az k8s cluster then pods will now be scheduled on distinct zones opposed to being co-located within the same az.

Just wanted to make sure there wasn't a case to be made for this being a breaking change - this could even be if someone experiences unexpected cross-az data shipment charges that weren't happening before.

> The changes doc is outdated, as it still talks about anti-affinity rules. If you're happy with above change in behaviour, then I will create a PR to fix it.

I had fixed one occurrance of that in this pr, went ahead and addressed the other as well. If need be it can be separated to a different issue / pr.

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -294,8 +294,8 @@ If you do not specify a custom topology spread constraint, your Pods will have a
 
 ==== Job Scheduling
 
-Realm import and update Jobs inherit the scheduling of the server Pods. If your server Pod scheduling is highly restrictive this may result in the 
-Job Pods being unschedulable. In this situation you may override the Job scheduling using the KeycloakRealmImport CR `spec.scheduling` and Keycloak CR `spec.update.scheduling` fields.
+Realm import and update Jobs inherit the scheduling of the server Pods if it is specified the Keycloak CR `spec.shceduling` field. If your server Pod scheduling is highly restrictive this may result in the
```
ryanemerson: `s/shceduling/scheduling`

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
```diff
@@ -360,4 +365,12 @@ public void setReadinessProbeSpec(ProbeSpec readinessProbeSpec) {
     public void setStartupProbeSpec(ProbeSpec startupProbeSpec) {
         this.startupProbeSpec = startupProbeSpec;
     }
+
+    public SchedulingSpec getImportSchedulingSpec() {
```
ryanemerson: Does it make sense for us to take this opportunity to nest the import scheduling so we have `spec.import.scheduling`?

The benefits would be:

1. Consistent with `spec.update.scheduling`
2. Allows for future global import configuration that maybe required e.g. `spec.update.<tbdElement>`. I don't have a concrete example, but you never know what the future might bring.

ryanemerson (commented):
> I had fixed one occurrance of that in this pr, went ahead and addressed the other as well. If need be it can be separated to a different issue / pr.

Thanks! It makes sense for it to be addressed here :+1:

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
shawkins: > Does it make sense for us to take this opportunity to nest the import scheduling so we have spec.import.scheduling?

It does, we talked about this a little in our team meeting, but I didn't reflect it in this pr. It's possible that users will eventually want to tweak the resource allocation for the update and import jobs as well, so having a separate stanza is good.

shawkins:
@ryanemerson should be good now.

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -290,7 +290,26 @@ spec:
 
 Please see https://kubernetes.io/docs/concepts/scheduling-eviction[the kubernetes docs] for more on scheduling concepts.
 
-If you do not specify a custom affinity, your Pods will have an anti-affinity for the same zone and node to improve availability.
+If you do not specify a custom topology spread constraint, your Pods will have a spread constraint across zones and nodes to improve availability.
+
+==== Job Scheduling
+
+Realm import and update Jobs inherit the scheduling of the server Pods if it is specified the Keycloak CR `spec.scheduling` field. If your server Pod scheduling is highly restrictive this may result in the
```
ryanemerson: "...specified in the Keycloak..."

ryanemerson (commented):
Thanks @shawkins, other than a minor docs fix LGTM assuming CI is :green_circle:

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -290,7 +290,26 @@ spec:
 
 Please see https://kubernetes.io/docs/concepts/scheduling-eviction[the kubernetes docs] for more on scheduling concepts.
 
-If you do not specify a custom affinity, your Pods will have an anti-affinity for the same zone and node to improve availability.
+If you do not specify a custom topology spread constraint, your Pods will have a spread constraint across zones and nodes to improve availability.
```
vmuzikar:
```suggestion
If you do not specify a custom topology spread constraint, your Pods will have a spread constraint across zones and nodes to improve availability. These defaults do not affect import and update Jobs, only server Pods.
```

docs/guides/operator/advanced-configuration.adoc
shawkins: Corrected, thanks @ryanemerson

vmuzikar (approved):
LGTM, thank you @shawkins!

ryanemerson (approved):

PR #42317, opened 2025-09-03
Make sure inner transactions are using their own session

pedroigor:
Closes #41942

* The problem was that methods like `getKcSubGroups` are being invoked in different transactions and using the top-level session instead of the session bound to the inner transaction from where they are invoked


sguilhen (approved):

ahus1 (approved):

ajaiswal595:
In what version of keycloak will these changes be available ?

sguilhen:
@ajaiswal595 Looking at the original issue, the fix was scheduled for 26.2.9, 26.3.4 and 26.4.0

PR #42319, opened 2025-09-03
Trigger login on other tabs only when tab is active

ahus1:
This should avoid too many concurrent requests and extra load on KC and applications.

Closes #42318

### DIFF
--- a/themes/src/main/resources/theme/base/login/resources/js/authChecker.js
+++ b/themes/src/main/resources/theme/base/login/resources/js/authChecker.js
@@ -2,7 +2,7 @@ const SESSION_POLLING_INTERVAL = 2000;
 const AUTH_SESSION_TIMEOUT_MILLISECS = 1000;
 const initialSession = getSession();
 const forms = Array.from(document.forms);
-let timeout;
+let timeout, eventListener;
 
 // Stop polling for a session when a form is submitted to prevent unexpected redirects.
 // This is required as Safari does not support the 'beforeunload' event properly.
@@ -21,6 +21,17 @@ globalThis.addEventListener("beforeunload", () => stopSessionPolling());
 export function startSessionPolling(redirectUrl) {
   if (initialSession) {
     // We started with a session, so there is nothing to do, exit.
+    return;
+  }
+
+  if (!eventListener) {
+    eventListener = () => startSessionPolling(redirectUrl);
+    window.addEventListener("focus", eventListener);
+    window.addEventListener("blur", eventListener);
+  }
+
+  if (document.hidden) {
+    // Wait until the user activates this tab
     return;
   }
 
@@ -46,6 +57,11 @@ function stopSessionPolling() {
   if (timeout) {
     clearTimeout(timeout);
     timeout = undefined;
+  }
+  if (eventListener) {
+    window.removeEventListener("focus", eventListener);
+    window.removeEventListener("blur", eventListener);
+    eventListener = undefined;
   }
 }
 


mposolda (changes requested):
I don't think that we should include this change. It is on purpose that when user authenticates in tab1, the other browser tabs, where authentication to same Keycloak was opened, are also authenticated immediately.

There is the single `RootAuthenticationSessionModel` per whole browser (all browser tabs), but every browser tab has it's own instance of `AuthenticationSessionModel` . When the tab1 is authenticated, the single `AuthenticationSessionModel` is removed from `RootAuthenticationSessionModel` . The idea of `authChecker.js` is, that other browser tabs will follow immediatelly. As otherwise, if their login is postponed until they are "active", the `RootAuthenticationSessionModel` might be already expired from the infinispan cache, which means impossible to login properly.

You can notice the flaky tests of `MultipleTabsLoginTest`, which are caused by this change. 

Another example scenario, which is broken after the change:
1) User opens http://localhost:8081/auth/admin/master/console/#/master/identity-providers in tab1
2) User opens http://localhost:8081/auth/admin/master/console/#/master/clients in tab2 of the same browser
3) User authenticates as admin/admin in tab2
4) Wait 2 minutes (that is the time of RootAuthenticationSessionModel to expire)
5) Go back to tab1. Once it is focused, there is an attempt to continue login. That would fail and screen will froze. There are those error events for the server:

```
10:07:12,803 WARN  XNIO-1 task-2 [org.keycloak.events] type="RESTART_AUTHENTICATION_ERROR", realmId="45177bc9-5019-450f-8883-308f6cc863ab", realmName="master", clientId="security-admin-console", userId="null", ipAddress="127.0.0.1", error="expired_code", restart_after_timeout="true"
10:07:12,903 WARN  XNIO-1 task-2 [org.keycloak.events] type="CODE_TO_TOKEN_ERROR", realmId="45177bc9-5019-450f-8883-308f6cc863ab", realmName="master", clientId="security-admin-console", userId="13aaa07c-2f80-4f85-a08c-84cd7c4adb78", sessionId="4bec2745-6359-4231-b442-4270b13ce82e", ipAddress="127.0.0.1", error="invalid_redirect_uri", reason="Parameter 'redirect_uri' did not match originally saved redirect URI used in initial OIDC request. Saved redirectUri: http://localhost:8081/auth/admin/master/console/#/master/clients, redirectUri parameter: http://localhost:8081/auth/admin/master/console/#/master/identity-providers", grant_type="authorization_code", code_id="4bec2745-6359-4231-b442-4270b13ce82e", client_auth_method="client-secret"
```

Before this change, the tab1 is authenticated immediatelly after step 3 and it is properly redirected to http://localhost:8081/auth/admin/master/console/#/master/identity-providers

I am not sure how much overhead it is, that other tabs are "eagerly" authenticating the user, however updating this would re-introduce the bunch of other bugs. If we want to change this, we will likely need to figure the way how could be the "authentication context" (Parameters sent from OIDC/SAML client like `redirect_uri`, `scope` etc) properly saved inside browser. Saving them on the server (in the `RootAuthenticationSessionModel` ) has some trade-offs as expired authentication session entitires might always need to be cleaned from the cache/storage of the server after some time as otherwise the cache/store will grow indefinitely.

Moreover, this change will not help with fixing https://github.com/keycloak/keycloak/issues/40374 as the problem is not directly related to `authChecker.js` . It happens with other identity providers as well (AFAIK ForgeRock or FranceConnect are confirmed) . It seems those identity providers also works in a way that when you login in tab1 of ForgeRock, the other browser tabs are immediately authenticated, which causes redirect to Keycloak IDP callback endpoint and hence more tabs are trying to create same userSession.

ahus1:
>  It seems those identity providers also works in a way that when you login in tab1 of ForgeRock, the other browser tabs are immediately authenticated

Reading that, it seems that this is a UX feature that those other tabs are automatically logged in. 

I agree it doesn't fix the other bug at hand, still I was thinking if not logging in the other tabs automatically would be a better way to behave.

As it then seems to be better in its current state, I'm closing this one.

PR #42322, opened 2025-09-03
refining http threads docs

shawkins:
Alternative to refining the existing text - just remove anything that is too vague to be actionable.

closes: #39342


ryanemerson (approved):

ahus1 (approved):
Thanks for triggering this one, approving. Let's see if we can make this simpler in the future so people don't need to go into so much details.

PR #42324, opened 2025-09-03
Update documentation after changes to RFC8414 handling

### LINKED ISSUE #42323: Update documentation after changes to RFC8414 handling
### Description

This is about adding the necessary docs changes as a follow-up to #41440

ahus1:
Closes #42323

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -43,6 +43,28 @@ Custom caches can still be added without setting this option.
 When upgrading an existing deployment, remove all default cache configurations from your existing `conf/cache-ispn.xml`
 and use the `+--cache-...+` options to make changes for example to the cache sizes.
 
+=== RFC8414 compliant lookup of metadata
+
+{project_name} now exposes an RFC8414-compliant endpoint at the root URL level `/.well-known/` to allow clients to discover OAuth 2.0 Authorization Server Metadata by the issuer URL.
+
+Previously information was exposed via:
+
+[source]
+----
+https://keycloak.example.com/realms/{realm}/.well-known/oauth-authorization-server
+----
+
+It is now available in addition to the legacy URL via:
+
+[source]
+----
+https://keycloak.example.com/.well-known/oauth-authorization-server/realms/{realm}
+----
+
+To benefit from this, expose the path `/.well-known/` in your reverse proxy configuration.
+
+NOTE: If a `http-relative-path` is configured, configure a reverse proxy to map the `/.well-known/` path to the path with the prefix on the server.
+
 === Operator default affinity configuration changed
 
 The default affinity strategy has been updated so that a `preferredDuringSchedulingIgnoredDuringExecution` anti-affinity rule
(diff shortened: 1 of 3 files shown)


ahus1:
@tnorimat / @thomasdarimont / @mposolda - this is IMHO the minimal documentation update that should come with the changes done in the original PR. 

I chose not to deprecate any of the legacy URLs, or update any of the existing URLs in the docs to the new URL. 

Please let me know if you agree and review or comment this PR before we then merge it.

tnorimat:
@ahus1 Hello, I reviewed the PR. It looks good to me.

thomasdarimont:
One could add a note that the root-level .well-known endpoint also works for other well-known providers.
```
https://keycloak.example.com/.well-known/{wellKnownProviderId}/realms/{realm}
```

docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
```diff
@@ -43,6 +43,28 @@ Custom caches can still be added without setting this option.
...
+
+{project_name} now exposes an RFC8414-compliant endpoint at the root URL level `/.well-known/` to allow clients to discover OAuth 2.0 Authorization Server Metadata by the issuer URL.
+
+Previously information was exposed via:
+
+[source]
+----
+https://keycloak.example.com/realms/{realm}/.well-known/oauth-authorization-server
+----
+
```
stianst, line 56: Probably don't want to call this legacy, as that suggests it's being deprecated. Don't think we plan to do that anytime soon

ahus1: updated in next commit

ahus1:
> One could add a note that the root-level .well-known endpoint also works for other well-known providers.

Addressed in next commit without going into specifics. We don't have much other specifics on those well-known providers in the docs. Maybe someone wants to write some in a another PR.

Should now be ready for a final review.

thomasdarimont:
LGTM now. 

Perhaps we could highlight the improved support for serving .well-known resources at the context-root in the release notes.
Previously, serving additional `.well-known` endpoints at the context-root required the use of a reverse proxy. With the enhanced `.well-known` handling we can now support it with just an extension.

Some examples might include:
- [RFC9116 Security.txt](https://www.rfc-editor.org/rfc/rfc9116.html#section-2.5.2)
- [Google's related-website-set.json](https://github.com/GoogleChrome/related-website-sets/blob/main/Well-Known-Specification.md)
- [Google assetlinks.json](https://github.com/google/digitalassetlinks/blob/master/well-known/specification.md)
- [Apple apple-app-site-association](https://developer.apple.com/documentation/xcode/supporting-associated-domains)

This makes it also easier to support .well-known endpoints for upcoming standards like [OpenID Federation](https://openid.net/specs/openid-federation-1_0.html#section-9), [OpenID Shared Signals Framework](https://openid.github.io/sharedsignals/openid-sharedsignals-framework-1_0.html#section-7.2.1), and [OpenID4VCI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-11.2.2) with an extension.

tnorimat (approved):

pedroigor (approved):

PR #42341, opened 2025-09-04
Support for EDB 17

### LINKED ISSUE #42293: Set Liquibase DB type based on the `db` option
### Description

Make sure the correct DB type is used based on the `db` option in Liquibase.

### Discussion

_No response_

### Motivation

Currently, the DB type is assumed based on the JDBC URL, the `db` option is ignored. This creates a discrepancy between Hibernate and Liquibase in cases when the URL does not match the requested vendor via the `db` option – Hibernate uses the dialect based on what's provided in `db`, Liquibase does not. This is especially problematic with some of the "compatible" database vendors like e.g. EDB which might not be detected as postgres in Liquibase (despite `db=postgres`)

### Details

Probably, this will need to be changed:
https://github.com/keycloak/keycloak/blob/f34bb21af606e9677dc225d1b0cb371ab9f2ff3d/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/DefaultLiquibaseConnectionProvider.java#L117

vmuzikar:
Closes #42742
Closes #42293


model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/PostgresPlusDatabase.java
vmuzikar, line 1: We're now assuming 1:1 mapping between the `db` option and Liquibase `Database`.

This `Database` shouldn't be needed anymore as it mainly tried to fix the EDB auto-detection but couldn't work so well because Postgres specific changelogs were not used by it. Now, when EDB is used with `db=postgres`, it will be treated as Postgres by Liquibase. If we ever needed something specific (driver, Hibernate dialect, Liquibase Database, ...) for e.g. EDB, we'd introduce a new `db` vendor.

vmuzikar: @ahus1 @stianst Are you ok with this?

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/conn/DefaultLiquibaseConnectionProvider.java
```diff
@@ -92,10 +92,26 @@ protected void baseLiquibaseInitialization() {
...
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public void init(Config.Scope config) {
         indexCreationThreshold = config.getInt("indexCreationThreshold", 300000);
         logger.debugf("indexCreationThreshold is %d", indexCreationThreshold);
+
+        // We need to explicitly handle the default here as Config might not be MicroProfile and hence no actually server config exists
+        String dbAlias = config.root().get(DatabaseOptions.DB.getKey(), DatabaseOptions.DB.getDefaultValue().orElseThrow());
```
shawkins:
```suggestion
        String dbAlias = Optional.ofNullable(config.root().get(DatabaseOptions.DB.getKey())).orElseThrow();
```

Just a nit: we shouldn't have to resolve the default here - the propertymapper will do that. Aslo we won't have a default DB much longer.

vmuzikar: Unfortunately, with `SystemPropertiesScope` Property Mappers are not used. This affects e.g. model tests which don't start the full server.

shawkins: This will be a problem once we hit 27.0 and `DatabaseOptions.DB.getDefaultValue().orElseThrow()` throws an exception.

vmuzikar: Do you suggest to hardcode `dev-file` there? That will be still default for start-dev even after 27.0, no?

shawkins: Yes, or change the model tests to set that by default.

vmuzikar: Ok, hardcoded dev-file there. It felt more robust than doing it selectively just for the model tests.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/PostgresPlusDatabase.java
ahus1: I assume this will work with empty EDB databases, still I assume we will run into problems with existing EDB databases which will then have (a) wrong DB schema statements applied in the past and (b) will now start to complain about no-longer-matching checksums.

I don't know how to handle that. Maybe adding the now broken checksums as valid checksums, assume on either PostgreSQL or EDB that those statement might gone wrong, and therefore undo and redo them? This might take a while if indexes are involved, and it doesn't sound like a good idea.

Any other ideas?

vmuzikar: We didn't officially support EDB for RHBK / Quarkus based Keycloak. Possibly we could find some manual steps to fix the DB that the users could execute?

ahus1: It will work for some, but there is the groups that are upgrading from RH SSO will also see some broken checksums. Some of them "fixed" the checksums when upgrading from RHSSO 7.5 to 7.6, and now everyone who upgrades from RHSSO 7.6 with EDB to RHBK 26.x will see some broken checksums.

ahus1: Note that some of them also have the wrong changesets installed, depending whether they upgrade from 7.5 to 7.6, or installed 7.6.

vmuzikar: I did some further investigation around EDB and how it reacts with Liquibase.

The key is in the JDBC driver. If the driver is the standard Postgres driver, the EDB was even previously handled as Postgres – the DB connection metadata was announcing it as Postgres. If EDB driver was used, then the DB was not handled as Postgres and then the incorrect schemas were applied.

We supported using the EDB driver only with Wildfly dist, i.e. RH-SSO. ~~AFAIK, it was never possible to use the EDB driver with Quarkus dist due to how we enforce DB drivers based on the DB vendor – as a result even Liquibase couldn't use EDB driver even if it was present on the classpath.~~

That means the checksum mismatch will affect ~~only~~ customers using EDB with EDB driver on RH-SSO, and then migrating to RHBK. There's no good fix to that, unfortunately. I would just document that users need to export their Realms using the export command, create a fresh KC instance and reimport it. @ahus1 WDYT? If the user hits the changeset validation error, it should be fine as Liquibase won't write the changes to the DB in this case.

Theoretically, this PR is not needed for EDB. Still, I'd like to merge it as it makes the Liquibase logic more robust and aligned with what we do for Hibernate – i.e. enforcing the DB type based on the configured DB vendor, disabling special handling for some derived DB vendors.

EDIT: Didn't realize that even though we didn't have explicit EDB support in Quarkus dist, it was possible (and documented) how to provide a custom driver, and someone might have used that for EDB.

shawkins (dismissed):
The changes make sense.

> The key is in the JDBC driver.

Just wanted to double check more on the effect the driver has. We do document and have advised users in the discussions to put drivers in the lib providers. Is this limited to only tweaks of minor / micro versions, or could someone try to swap in something like a Datadirect driver - would that cause different liquibase behavior?

vmuzikar:
> We do document and have advised users in the discussions to put drivers in the lib providers.

That's a good point I didn't realize. Though we say that this is intended to be used for Oracle only, or for driver wrappers. Oracle we directly support. And the wrappers I assume were auto detected as the "base database". Still, this got me thinking if this PR is something that should be documented as a notable change. WDYT @shawkins @ahus1?

> Is this limited to only tweaks of minor / micro versions

What do you mean?

> or could someone try to swap in something like a Datadirect driver - would that cause different liquibase behavior?

Technically they could and it might cause different DB type used at Liquibase level.

shawkins:
> What do you mean?

Overriding the pg driver for example with just a change to the micro release version rather than waiting for another Keycloak release to pick up a fix.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/PostgresPlusDatabase.java
ahus1: @vmuzikar - performing "export their Realms using the export command, create a fresh KC instance and reimport it." is IMHO risky and time consuming, so I would really like to avoid that. 

If the checksum problems occur, it is AFAIK only about changeset `14.0.0-KEYCLOAK-18286-supported-dbs`. If that happens, the index has probably been created in a non-PostgreSQL way, or someone just flipped the checksum. 

To fix that, I suggest to run the following before the migration: 
1. Delete the changeset `14.0.0-KEYCLOAK-18286-supported-dbs` from the changelog table, so Liquibase will re-run the statement
2. Delete the index `IDX_CLIENT_ATT_BY_NAME_VALUE`, so the re-run of the changeset will succeed. 

AFAIK there are no other changesets marked with `postgresql` for any RHSSO (which is KC 18). The changesets for KC 20, 24 an 25 that are marked with `postgresql` should run just fine on an upgrade.

ahus1: This information could then be placed in the RHSSO migration guide, and not in the general docs.

vmuzikar:
I repurposed the PR to also cover adding support (and testing) for the EDB as it's closely related to the original purpose of this PR. Let me know if that makes sense, otherwise I can split it.

vmuzikar:
@ahus1 @shawkins I'd be happy if you can review when you get the chance. :)

docs/documentation/release_notes/topics/26_4_0.adoc
```diff
@@ -139,4 +139,10 @@ Simply set the backup site name (e.g., availability zone) using the following op
...
 ----
 
-When this option is set, Infinispan will automatically replicate the cache data to the specified location.
+When this option is set, Infinispan will automatically replicate the cache data to the specified location.¨
+
+= Support for additional databases
+
+With this release, we added support for the following new database vendors:
+
+* EnterpriseDB (EDB) Advanced 17.6
```
vmuzikar, line 177: We will be adding Azure too, hence making it a list.

tests/base/src/test/java/org/keycloak/tests/db/CaseSensitiveSchemaTest.java
```diff
@@ -37,7 +40,7 @@ public KeycloakServerConfigBuilder configure(KeycloakServerConfigBuilder config)
     public static class CaseSensitiveDatabaseConfig implements DatabaseConfig {
         @Override
         public DatabaseConfigBuilder configure(DatabaseConfigBuilder database) {
-            if (PostgresTestDatabase.NAME.equals(dbType())) {
+            if (PostgresTestDatabase.NAME.equals(dbType()) || EnterpriseDbTestDatabase.NAME.equals(dbType())) {
```
vmuzikar, line 43: This is not ideal but didn't want to mess too much with the framework to get the `TestDatabase` instance statically to get the vendor, so we could have a single check for `vendor == postgres`.

vmuzikar:
Passed EDB tests: https://github.com/vmuzikar/keycloak/actions/runs/17943790005/job/51025685173.

docs/guides/server/db.adoc
```diff
@@ -49,6 +50,9 @@ this database
 </@profile.ifCommunity>
 or skip this section if you want to connect to a different database for which the database driver is already included.
 
+NOTE: Using custom drivers is generally considered unsupported, with the specific  exceptions mentioned in this guide
```
shawkins: I'm not sure it's clear what a custom driver means here. The guide talks to ms sql, oracle, and overriding built-in drivers. Does this mean overriding built-in drivers is fully supported?

shawkins (dismissed):

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/PostgresPlusDatabase.java
vmuzikar: Unfortunately, this does not affect only RH-SSO. To be precise, the EDB driver (and hence the incorrect changes in the schemas) could be used in Quarkus dist / RHBK too, though we did not supported EDB there until now. And since version 18, there were more postgres specific changesets. For that reason, I believe it would not be viable to instruct users to fix the DB manually.

docs/guides/server/db.adoc
vmuzikar: Updated the note. Please let me know if it reads better.

shawkins: That looks good - doesn't leave it open to interpretation if overriding the driver is really needed.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/PostgresPlusDatabase.java
vmuzikar: As per an offline discussion, I created a few new changesets to re-apply some of those that were skipped when using the EDB driver. They are mostly concerning indexes, so I typically dropped and re-added them. Those will apply also to regular (non-EDB) postgres, though it should be no harm in that. The last changeset – a bigger one – from Keycloak 25 should be applied conditionally only when the inconsistent schema is detected, so that should be safe too.

Only two changesets required new `validCheckSum`. That is because how Liquibase calculates the checksums. Those are not affected by the preconditions – the stored checksum is the same regardless of if the changeset was executed or not. What affects it though is the `<modifySql dbms=...` statement which obviously changes based on the DB vendor, and with that also the checksum.

I tried two separate scenarios: using the EDB driver I initialized a new DB with the latest RH-SSO 7.6 and Keycloak 23. After that I ran my changes in this PR. In both cases the DB seems to be in an expected state, as if it was initialized with the postgres driver (and not EDB).

As discussed, no automated tests were added as this is supposed to be a one-off migration with limited regression risk.

@ahus1 Let me know what you think.

model/jpa/src/main/resources/META-INF/jpa-changelog-20.0.0.xml
```diff
@@ -38,6 +39,18 @@
         </modifySql>
     </changeSet>
 
+    <changeSet author="keycloak" id="20.0.0-12964-supported-dbs-edb-migration">
+        <preConditions onSqlOutput="TEST" onFail="MARK_RAN">
+            <dbms type="postgresql"/>
+            <changeSetExecuted id="20.0.0-12964-supported-dbs" author="keycloak" changeLogFile="META-INF/jpa-changelog-20.0.0.xml"/>
+        </preConditions>
```
ahus1, line 47: As Keycloak sometimes skips to create indexes in the index has a lot of entries, it would be good to check if the index `IDX_GROUP_ATT_BY_NAME_VALUE` actually exists. I fell into that trap before :-(

Applies also for the other changesets.

model/jpa/src/main/resources/META-INF/jpa-changelog-14.0.0.xml
```diff
@@ -84,6 +85,18 @@
         </modifySql>
     </changeSet>
 
+    <changeSet author="keycloak" id="14.0.0-KEYCLOAK-18286-supported-dbs-edb-migration">
+        <preConditions onSqlOutput="TEST" onFail="MARK_RAN">
+            <dbms type="postgresql"/>
+            <changeSetExecuted id="14.0.0-KEYCLOAK-18286-supported-dbs" author="keycloak" changeLogFile="META-INF/jpa-changelog-14.0.0.xml" />
```
ahus1: This seems to be a duplicate to `<changeSet author="keycloak" id="24.0.0-26618-edb-migration">`
So can the first be skipped?

ahus1 (changes requested):
Thank you for the PR. See below for some need for polishing.

Please also add a note to the migration guide to list that some indexes are dropped, but only re-created if they have not too many entries. Unfortunately we don't have a good automation for people to not need to watch that.

model/jpa/src/main/resources/META-INF/jpa-changelog-20.0.0.xml
vmuzikar: I added the precondition check for indexes,  it doesn't harm anything.

> As Keycloak sometimes skips to create indexes in the index has a lot of entries

Just out of curiosity, what do you mean? Did you mean "Keycloak sometimes skips to **drop** indexes if the index has a lot of entries"?

model/jpa/src/main/resources/META-INF/jpa-changelog-14.0.0.xml
vmuzikar: Good point, kept only the latter.

vmuzikar:
> Please also add a note to the migration guide to list that some indexes are dropped, but only re-created if they have not too many entries.

I'm not sure I entirely follow. What would be the action the users need to take? Create the index manually? That seems odd. Should we then make KC always check the index presence and create it if needed? It almost sounds like a Liquibase bug.

ahus1:
@vmuzikar 

> I'm not sure I entirely follow. What would be the action the users need to take? Create the index manually? That seems odd. Should we then make KC always check the index presence and create it if needed? It almost sounds like a Liquibase bug.

Long before my time at the Keycloak project, someone added a feature that indexes for big tables are not created automatically. The idea was that creating an index for a very large table can lead to an extended downtime during upgrades, and customers complained about it. See https://github.com/keycloak/keycloak/blob/83c00731c3a103b86007d7402c77ed5ebb7a517b/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java#L53-L55 for the implementation.

See https://www.keycloak.org/docs/latest/upgrading/index.html#automatic-relational-database-migration for the docs: 

> For example, creating an index on tables with millions of records can be time-consuming and cause a major service disruption. Therefore, a threshold of 300000 records exists for automatic migration. If the number of records exceeds this threshold, the index is not created. Instead, you find a warning in the server logs with the SQL commands that you can apply manually.

This is both a blessing and a curse. Now and again people report bugs about slow installations due to missing indexes. I have a draft issue in the SRE team backlog to come up with something better, but haven't yet found the time. Maybe a team that focuses on databases can have more priority on this :wave: 

The problem is that those log messages are easily missed, so we add a note in the migration guide, for example like the following: 

https://github.com/keycloak/keycloak/blame/56246096e09728ae6e62903120d042f40195ee5d/docs/documentation/upgrading/topics/changes/changes-24_0_0.adoc#L312-L314

vmuzikar:
@ahus1 Thank you for the valuable hints, I was not aware of this behaviour around indexes. I added the migration note.

ahus1 (approved):
Thank you for this extensive PR. I hope the community and the customers will like it :D

vmuzikar:
@shawkins Do you want to give it a look as well?

vmuzikar:
@shawkins Given the imminent release (and that also https://github.com/keycloak/keycloak/pull/42985 depends on this), I'm gonna go ahead and merge this. Let me know if you have any further concerns and we can address them as a follow-up.

PR #42353, opened 2025-09-04
fix: setting the built flag for embedded usage

### LINKED ISSUE #23972: Improve handling config options in scripts preventing re-augmentation

shawkins:
closes: #23972


pedroigor (approved):
Thanks, @shawkins ! It solves the problem.

PR #42355, opened 2025-09-04
add logout_hint to broker browser logout url

### LINKED ISSUE #33008: Add logout_hint to brokers url
### Description

While logging out using keycloakInitiatedBrowserLogout, the system does not include the logout_hint parameter, even though it is available in the ID token. This omission causes an issue with Azure logout, where it prompts us to pick an account.

### Discussion

_No response_

### Motivation

If we add this, we will achieve a seamless logout experience.

### Details

Here's a formatted version of your message:

---

I have fixed the issue and committed the changes here: [https://github.com/keycloak/keycloak/commit/d9c9052d88c9f49cf1299bda3d0add09a3921160](https://github.com/keycloak/keycloak/commit/d9c9052d88c9f49cf1299bda3d0add09a3921160).

Please let me know if I can raise a pull request for this, or if there is another way to proceed.

---

dasniko:
closes #33008

Adds the `logout_hint` parameter to the broker browser logout url with the value the external IdP sent as claim `login_hint` in the ID token.
If there's no `login_hint` claim in the ID token, or if there's any other problem resolving it, it is ignored to not invoke an error/exception.

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
@@ -79,6 +79,7 @@ import org.keycloak.services.resources.RealmsResource;
 import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.util.JsonSerialization;
 import org.keycloak.util.TokenUtil;
+import org.keycloak.utils.StringUtil;
 import org.keycloak.vault.VaultStringSecret;
 
 import java.io.IOException;
@@ -173,6 +174,7 @@ public class OIDCIdentityProvider extends AbstractOAuth2IdentityProvider<OIDCIde
     public Response keycloakInitiatedBrowserLogout(KeycloakSession session, UserSessionModel userSession, UriInfo uriInfo, RealmModel realm) {
         if (getConfig().getLogoutUrl() == null || getConfig().getLogoutUrl().trim().equals("")) return null;
         String idToken = userSession.getNote(FEDERATED_ID_TOKEN);
+        String loginHint = extractLoginHintFromIdToken(idToken);
         if (getConfig().isBackchannelSupported()) {
             backchannelLogout(userSession, idToken);
             return null;
@@ -182,6 +184,9 @@ public class OIDCIdentityProvider extends AbstractOAuth2IdentityProvider<OIDCIde
                     .queryParam("state", sessionId);
             if (getConfig().isSendIdTokenOnLogout() && idToken != null) {
                 logoutUri.queryParam("id_token_hint", idToken);
+            }
+            if (getConfig().isSendLogoutHintOnLogout() && StringUtil.isNotBlank(loginHint)) {
+                logoutUri.queryParam("logout_hint", loginHint);
             }
             if (getConfig().isSendClientIdOnLogout()) {
                 logoutUri.queryParam("client_id", getConfig().getClientId());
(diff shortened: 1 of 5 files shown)


services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
```diff
@@ -196,6 +201,17 @@ public Response keycloakInitiatedBrowserLogout(KeycloakSession session, UserSess
         }
     }
 
+    protected String extractLoginHintFromIdToken(String idToken) {
```
pedroigor, line 204: Instead of parsing the ID Token, would it be enough to get the username from `org.keycloak.events.Details#IDENTITY_PROVIDER_USERNAME`?

dasniko: Not in case of MS EntraID...
In general, it's not clearly specified, what exactly should be put into the optional `logout_hint` parameter.

For Microsoft, they need to have the `login_hint` value from the id token, as mentioned [here](https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#send-a-sign-out-request) (this link is also mentioned in the corresponding issue #33008):
> ...and use the value of the login_hint optional claim as the logout_hint parameter. Don't use UPNs or phone numbers as the value of the logout_hint parameter.

I'm currently not aware of other IdPs actually using this parameter.
Parsing it from the id token is the most generic way I can currently think of.

pedroigor: Sounds like we should introduce this behavior only for `org.keycloak.social.microsoft.MicrosoftIdentityProvider` then?

Having this in `OIDCIdentityProvider` suggests it applies to other implementations, and I think that is not the case. AFAIK, the ID Token in the specification does not define a `login_hint` claim, does it?

dasniko: I get your point and understand your concerns, but... ;)

The `MicrosoftIdentityProvider` lacks some other (a lot) config options, which are available in the generic OIDC provider, also the available mapper types are less than in the generic provider.
Additionally, the `MicrosoftIdentiyProvider` does _not_ have logout options at all (which might make sense for a _social_ provider, but perhaps not in an enterprise environment, where I want to have more control over what is possible and what not).

These (and more) missing options let customers use the generic OIDC provider for connecting to MS EntraID instead of using the social provider.
IMHO the `MicrosoftIdP` would have to be extended explicitly in functionality and configurability.

pedroigor: Yeah, I see now.

Looks like we need a Microsoft EntraID, based on the OIDC Provider and with extended capabilities.

IMO, that is the best we can do to start supporting EntraID, given that I've seen quite a few people relying on it.

So we have two options:

* Hack the generic OIDC provider with EntraID-specific things
* Provide a EntraID provider that is basically an OIDC provider but with anything specific to it

I would prefer #2. Wdyt?

dasniko: Some dedicated support of EntraID makes definitely sense, as this is one of the most federated IdPs for companies and corporates.

I don't know, if a dedicated provider makes sense in terms of maintainability. In general, EntraID conforms mostly to the OIDC spec. The `logout_hint` behavior is the first, which is - AFAIK - interpreted opinionatedly.
Would it make sense to extend the mapper functionality also towards logout? E.g. one could create a mapper to map something from user, session, tokens, etc. to logout params!? Or is this too complicated to implement?

If the mapper approach isn't feasible, I'd also vote for some EntraID provider like you mentioned with your no. 2

pedroigor: I would say that introducing mappers during the logout sounds like too much. Mainly because it is solving a very corner case, for a specific integration.

> Some dedicated support of EntraID makes definitely sense, as this is one of the most federated IdPs for companies and corporates.

Yeah, I see that. And perhaps we should write one. Are you willing to contribute to it? I can assist with whatever you need, given the importance of this integration.

My main concern, which also applies to any other integration we have, is testing. If you have any idea how to automate tests for it, that would be awesome. Otherwise, we cannot do much ...

dasniko: Agreed.
I will try to contribute, yes. Let's create a separate issue for this and close this PR and linked issue.

dasniko:
Closing this in favor of #42585

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
pedroigor: Thanks, @dasniko.

PR #42357, opened 2025-09-04
fix: updating to MariaDB 11.8 LTS

shawkins:
closes: #42356


vmuzikar (approved):
Ah, that was simple enough in the end. :D Thanks.

PR #42359, opened 2025-09-04
Remove usage of the term stretched from single-cluster HA guides

ryanemerson:
Closes #42358


ahus1 (approved):
Thank you for the quick turnaround!

PR #42362, opened 2025-09-04
EdDSA support for DPoP

### LINKED ISSUE #42286: Support EdDSA for DPoP
### Description

This is a follow-up issue to #42030 and #33942 to support EdDSA algorithm.

Support may be added to both server and ideally also to `DPoPGenerator` introduced in #33942, so it is possible to test EdDSA and use it from libraries like Java admin-client.

It might be needed to:
- Introduce `EdDSASignatureContext` in the `core` package (See https://github.com/keycloak/keycloak/pull/42054#discussion_r2315265239 for the details)
- Remove filtering from `DPoPUtil` class (See https://github.com/keycloak/keycloak/pull/42200#discussion_r2313475567 for more context and the specific line where it should be updated)
- Update `DPoPGenerator.createSignatureSignerContext` method and add `EdDSA`

### Discussion

https://github.com/keycloak/keycloak/pull/42054#discussion_r2315262828

### Motivation

_No response_

### Details

_No response_

mposolda:
closes #42286

### DIFF
--- a/core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
+++ b/core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
@@ -19,9 +19,6 @@ package org.keycloak.jose.jwk;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import org.keycloak.common.util.PemUtils;
-
-import java.security.NoSuchAlgorithmException;
 
 /**
  * @author <a href="mailto:sthorger@redhat.com">Stian Thorgersen</a>
--- a/core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
+++ b/core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
@@ -17,6 +17,7 @@
 
 package org.keycloak.jose.jwk;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.keycloak.common.util.PemUtils;
 
@@ -64,5 +65,27 @@ public class ECPublicJWK extends JWK {
 
     public void setY(String y) {
         this.y = y;
+    }
+
+    @JsonIgnore
+    @Override
+    public <T> T getOtherClaim(String claimName, Class<T> claimType) {
+        Object claim = null;
+        switch (claimName) {
+            case CRV:
+                claim = getCrv();
+                break;
+            case X:
+                claim = getX();
+                break;
+            case Y:
+                claim = getY();
+                break;
+        }
+        if (claim != null) {
+            return claimType.cast(claim);
+        } else {
+            return super.getOtherClaim(claimName, claimType);
+        }
     }
 }
--- a/core/src/main/java/org/keycloak/util/DPoPGenerator.java
+++ b/core/src/main/java/org/keycloak/util/DPoPGenerator.java
@@ -20,19 +20,18 @@ package org.keycloak.util;
 import java.security.Key;
 import java.security.KeyPair;
 import java.security.PrivateKey;
-import java.security.interfaces.RSAPublicKey;
 
 import org.keycloak.OAuth2Constants;
-import org.keycloak.common.util.Base64Url;
 import org.keycloak.common.util.SecretGenerator;
 import org.keycloak.common.util.Time;
 import org.keycloak.crypto.AsymmetricSignatureSignerContext;
 import org.keycloak.crypto.ECDSASignatureSignerContext;
 import org.keycloak.crypto.KeyType;
+import org.keycloak.crypto.KeyUse;
 import org.keycloak.crypto.KeyWrapper;
 import org.keycloak.crypto.SignatureSignerContext;
 import org.keycloak.jose.jwk.JWK;
-import org.keycloak.jose.jwk.RSAPublicJWK;
+import org.keycloak.jose.jwk.JWKBuilder;
 import org.keycloak.jose.jws.JWSBuilder;
 import org.keycloak.jose.jws.JWSHeader;
 import org.keycloak.jose.jws.crypto.HashUtils;
@@ -40,7 +39,6 @@ import org.keycloak.representations.dpop.DPoP;
 
 import static org.keycloak.OAuth2Constants.DPOP_DEFAULT_ALGORITHM;
 import static org.keycloak.OAuth2Constants.DPOP_JWT_HEADER_TYPE;
-import static org.keycloak.jose.jwk.JWKUtil.toIntegerBytes;
 
 /**
  * Utility for generating signed DPoP proofs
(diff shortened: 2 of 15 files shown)


mposolda:
@tnorimat Do you please have opportunity to review this PR?

tnorimat:
@mposolda Yes, I will review the PR.

core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
```diff
@@ -17,6 +17,7 @@
 
 package org.keycloak.jose.jwk;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.keycloak.common.util.PemUtils;
```
tnorimat: L22 and L24's imported class are no longer used.

core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
```diff
@@ -65,4 +66,26 @@ public String getY() {
     public void setY(String y) {
         this.y = y;
     }
+
+    @JsonIgnore
+    @Override
+    public <T> T getOtherClaim(String claimName, Class<T> claimType) {
```
tnorimat, line 69: Just out of curiosity, would Keycloak project introduces the latest switch statement format as follows ?

```
        Object claim = switch (claimName) {
            case CRV -> getCrv();
            case X -> getX();
            case Y -> getY();
            default -> null;
        };
```

core/src/main/java/org/keycloak/util/DPoPGenerator.java
```diff
@@ -74,41 +72,51 @@ public static JWK createRsaJwk(Key publicKey) {
...
+    }
+
+    private SignatureSignerContext createSignatureSignerContext(KeyWrapper keyWrapper) {
         switch (keyWrapper.getType()) {
             case KeyType.RSA:
                 return new AsymmetricSignatureSignerContext(keyWrapper);
             case KeyType.EC:
                 return new ECDSASignatureSignerContext(keyWrapper);
-             // TODO: EdDSA?
+            case KeyType.OKP: // EdDSA
```
tnorimat: To remove redundancy, it seems that it can be written as follows:
```
            case KeyType.RSA:
            case KeyType.OKP: // EdDSA
                return new AsymmetricSignatureSignerContext(keyWrapper); 
            case KeyType.EC:
                return new ECDSASignatureSignerContext(keyWrapper);
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/DPoPTest.java
```diff
@@ -40,11 +40,16 @@
 import org.keycloak.common.util.KeyUtils;
 import org.keycloak.common.util.Time;
 import org.keycloak.crypto.Algorithm;
+import org.keycloak.crypto.KeyType;
```
tnorimat: It seems that this class is no longer used.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/DPoPTest.java
```diff
@@ -419,7 +474,21 @@ public void testDPoPProofInvalidType() throws Exception {
 
     @Test
     public void testDPoPProofInvalidSignature() throws Exception {
-        testDPoPProofFailure(generateSignedDPoPProof(UUID.randomUUID().toString(), HttpMethod.POST, oauth.getEndpoints().getToken(), (long) Time.currentTime(), Algorithm.PS256, jwsEcHeader, rsaKeyPair.getPrivate(), null), "DPoP verification failure: org.keycloak.exceptions.TokenSignatureInvalidException: Invalid token signature");
+        DPoPGenerator customDPoPGenerator = new DPoPGenerator() {
```
tnorimat: If this custom DPoPGenerator is the same as one in L445-458, it might be declared as an internal class to avoid duplication.

tnorimat (commented):
To pass integration tests, we need to modify the code of `AbstractWellKnownProviderTest.testDpopSigningAlgValuesSupportedWithDpop` from

```
            Assert.assertNames(oidcConfig.getDpopSigningAlgValuesSupported(), Algorithm.PS256, Algorithm.PS384, Algorithm.PS512,
                    Algorithm.RS256, Algorithm.RS384, Algorithm.RS512, Algorithm.ES256, Algorithm.ES384, Algorithm.ES512);
```

to

```
            Assert.assertNames(oidcConfig.getDpopSigningAlgValuesSupported(), Algorithm.PS256, Algorithm.PS384, Algorithm.PS512,
                    Algorithm.RS256, Algorithm.RS384, Algorithm.RS512, Algorithm.ES256, Algorithm.ES384, Algorithm.ES512, Algorithm.EdDSA);
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/DPoPTest.java
mposolda: Updated, Thanks for the suggestion

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/DPoPTest.java
mposolda: Done, Thanks for the suggestion

core/src/main/java/org/keycloak/util/DPoPGenerator.java
mposolda: Updated, Thanks for the suggestion

core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
mposolda: Updated, Thanks for the suggestion

core/src/main/java/org/keycloak/jose/jwk/ECPublicJWK.java
mposolda: Thanks for the suggestion, however unfortunately, I think that these might be problematic as this format is supported since Java 12, but those classes from the `core` module are still synced to keycloak/keycloak-client repository, which is supposed to work with Java 11. So I've rather did not applied this recommendation to avoid possible issues with keycloak-client compatibility.

mposolda:
@tnorimat Thanks for the review and suggestions! I've applied your suggestions, with the exception of one suggestion related to `switch` as we may still need to stick with the old format due the compatilibyt of Java 11 (See inline comment).

Also I've did minor update of `DPoPGenerator.createRsaJwk` to re-use `JWKBuilder` instead of "manually" building RSA JWK key.

Do you please have opportunity to re-review? I've used separate commit, so it is easier to check the latest changes, but I am planning to squash before the merge.

tnorimat (approved):
LGTM. Thank you,

mposolda:
@tnorimat Thanks!

PR #42363, opened 2025-09-04
Add support for immediate policies

### LINKED ISSUE #42311: [RLM] Add option to differentiate scheduled policies from immediate policies
### Description

Currently all resource policies are scheduling the actions to be executed at some time in the future by another task. In some cases, however, the actions in a policy should be run immediately. While this can be done with custom event listeners, adding the ability to do that using the same policy-based approach that we use for scheduled actions leads to a better user experience. So admins would organize the policies and actions in the same way regardless if we want to schedule actions for the future or just run them right away when an event is triggered.

sguilhen:
Closes #42311

### DIFF
--- a/services/src/main/java/org/keycloak/models/policy/SetUserAttributeActionProvider.java
+++ b/services/src/main/java/org/keycloak/models/policy/SetUserAttributeActionProvider.java
@@ -17,6 +17,7 @@
 
 package org.keycloak.models.policy;
 
+import static org.keycloak.keys.Attributes.PRIORITY_KEY;
 import static org.keycloak.models.policy.ResourceAction.AFTER_KEY;
 
 import java.util.List;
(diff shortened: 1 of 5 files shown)


services/src/main/java/org/keycloak/models/policy/SetUserAttributeActionProvider.java
```diff
@@ -54,7 +55,7 @@ public void run(List<String> userIds) {
                 for (Entry<String, List<String>> entry : actionModel.getConfig().entrySet()) {
                     String key = entry.getKey();
 
-                    if (!key.startsWith(AFTER_KEY)) {
+                    if (!key.startsWith(AFTER_KEY) && !key.startsWith(PRIORITY_KEY)) {
```
pedroigor, line 58: We should probably have a single place where all these "internal" keys are defined (e.g.: a set) and a method in the resource manager to check whether or not a key is internal?

Or something else with the same goal.

pedroigor (approved):

services/src/main/java/org/keycloak/models/policy/SetUserAttributeActionProvider.java
sguilhen: Maybe... I'm still not sure how to better handle this thing

PR #42370, opened 2025-09-05
Add missing fields for client offline session timeout and lifespan

### LINKED ISSUE #42369: Missing client session offline settings on realm level in the admin UI
### Before reporting an issue

### Area

admin/ui

### Describe the bug

In the admin UI, the fields for `clientOfflineSessionIdleTimeout` and `clientOfflineSessionMaxLifespan` are missing. 

### Version

main

### Regression

### Expected behavior

The fields should be shown in the admin UI. 

They used to be present in the old admin UI

### Actual behavior

They are missing.

### How to Reproduce?

Open the sessions tab in the new admin UI. 

### Anything else?

Due to this, the values can't be changed any more, and it is nearly impossible for users to assess the problem.

They used to be in the admin UI before, see https://github.com/keycloak/keycloak/blob/20.0.0/themes/src/main/resources/theme/base/admin/resources/partials/realm-tokens.html#L164-L193

ahus1:
Closes #42369

New screen:

image

image

### DIFF
--- a/docs/documentation/server_admin/topics/sessions/timeouts.adoc
+++ b/docs/documentation/server_admin/topics/sessions/timeouts.adoc
@@ -35,11 +35,16 @@ image:images/sessions-tab.png[Sessions Tab]
 
 [[_offline-session-max-limited]]
 |Offline Session Max Limited
-|This setting is for <<_offline-access, offline access>>. If this flag is *Enabled*, Offline Session Max can control the maximum time the offline token remains active, regardless of user activity. If the flag is *Disabled*, offline sessions never expire by lifespan, only by idle. Once this option is activated, the <<_offline-session-max, Offline Session Max>> (global option at realm level) and *Client Offline Session Max* (specific client level option in the *Advanced Settings* tab) can be configured.
+|This setting is for <<_offline-access, offline access>>. If this flag is *Enabled*, Offline Session Max can control the maximum time the offline token remains active, regardless of user activity. If the flag is *Disabled*, offline sessions never expire by lifespan, only by idle. Once this option is activated, the <<_offline-session-max, Offline Session Max>> and <<_client_offline-session-max, Client Offline Session Max>> (global option at realm level) can be configured.
 
 [[_offline-session-max]]
 |Offline Session Max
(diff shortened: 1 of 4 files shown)


ahus1:
@pruivo - would you be willing to review this PR?

pruivo:
@ahus1 not a big fan of UI. Personally, I can't tell what `Offline Session Max Limited` means, but the UI looks good to me.

ahus1:
From the docs: 

> If this flag is Enabled, Offline Session Max can control the maximum time the offline token remains active, regardless of user activity. If the flag is Disabled, offline sessions never expire by lifespan, only by idle. 

I also push one more commit to update the docs, and ask the core-iam team for input.

ahus1:
@pedroigor - can you or please someone from the core-iam team review this change as we don't have the UI know-how in our team? In addition to that, it is also related to sessions, which is a core-iam thing.

pruivo (approved):

pedroigor:
@ahus1 I'll try out the UI and get back to you. But do you know why those fields were removed? Weird ...

ahus1:
My assumption was that they were added for the old Admin UI, and then forgotten to be added to the new one? All functionality in the backend for those fields exists. So I think forgotten is the most likely explanation.

ahus1:
@pedroigor - digging deeper, I see that the fields have been added in 874642fe9e21dc0664e20f2a82cd0080105f751c in 2020, and the new amin console was added without those fields in 2022 from the very beginning in 394dd48a947eabd5d6abeb197086e311668b5a52. 

When the new admin console became the default in KC 19.0 (ae33af92d996a8ee22fd90d74b31628fa418f7e6), this was still the case. 

So I'd say not adding those fields in the new admin console is an omission. 

@mposolda - as @pedroigor mentioned that those timeout fields are more core-clients related, it would be great if you could comment on those and review them. As a customer was affected by this due to a misconfiguration that wasn't visible in the UI, this is scheduled for backports. 

Thanks!

mposolda:
@ahus1 PR looks good to me. It is very possible that some fields were forgotten to be added to the new admin console during migration of old admin console to new admin console and certain fields added to the old console in the meantime...

I would like also @rmartinc to review this as he fixed recently some issues related to them and has good knowledge regarding these.

rmartinc (approved):
@ahus1 LGTM! The options are OK and the `useWatch` is perfect. I checked the options are already in the client.

stianst (approved):

PR #42388, opened 2025-09-05
[admin-api-v2] Manual execution of Jakarta Validation

mabartos:
Manual execution of Jakarta Validation as it's not automatically done as for the case with the CDI support.

cc: @vmuzikar


vmuzikar (approved):
LGTM, thanks.

PR #42393, opened 2025-09-05
fix: allowing resolution of log color at build time

### LINKED ISSUE #42335: Colored output is lost during startup
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

Starting Keycloak I get colored output during the build step, but it goes away after that:

Image

### Version

main

### Regression

### Expected behavior

All output is colored, or alternatively no output is colored.

### Actual behavior

First few lines of output is colored, the rest is just black/white

### How to Reproduce?

Run Keycloak (on Fedora?)

### Anything else?

_No response_

shawkins:
I suppose it's safer to check specifically for log and console options, rather than expanding to all quarkus options.

The only other consideration is that picocli warn/error messages do not honor the color setting. Do we want to change that also? @mabartos @vmuzikar @stianst 

closes: #42335

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
@@ -87,7 +87,9 @@ public final class PropertyMappers {
         // If however expressions are not enabled that means quarkus is specifically looking for runtime defaults, and we should not provide a value
         // See https://github.com/quarkusio/quarkus/pull/42157
         if (isRebuild() && isKeycloakRuntime(name, mapper)
-                && (!NestedPropertyMappingInterceptor.getResolvingRoot().orElse(name).startsWith("quarkus.log.") || !Expressions.isEnabled())) {
+                && (NestedPropertyMappingInterceptor.getResolvingRoot().or(() -> Optional.of(name))
+                        .filter(n -> n.startsWith("quarkus.log.") || n.startsWith("quarkus.console.")).isEmpty()
+                        || !Expressions.isEnabled())) {
             return ConfigValue.builder().withName(name).build();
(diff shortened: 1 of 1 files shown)


mabartos (approved):
@shawkins Looks okay to me like this. These exceptions here are getting nasty :) but works for me for this case.   

> The only other consideration is that picocli warn/error messages do not honor the color setting. Do we want to change that also?

The question is whether it is worth the effort? But from what I know, it should be quite straightforward to provide the approach for it, right?

shawkins:
> The question is whether it is worth the effort? But from what I know, it should be quite straightforward to provide the approach for it, right?

Right, we'd have to check the value of this option before stylizing the messages.

mabartos:
> Right, we'd have to check the value of this option before stylizing the messages.

@shawkins IMO, it might be good to provide it to have the `log-console-color` consistent to accomplish all "promises" - like "I want to have colors turned off, so I don't wanna see any colors in log. Why are there some?"  

So feel free to do it.

mabartos:
@shawkins However, for now, as these colors are disabled, the UX will be worse as people will not easily recognize if it's an error. From the second thought, as AFAIK, people do not complain about these colors in logs, and maybe they'll start to complain when all colors for errors are gone? Like for PropertyException.

Ideally, it'd be good to provide these changes and turn on these colors by default, and if someone parses somehow these plain logs, they'd need to explicitly switch off the coloring. However, it might feel like a breaking change for their deployments. So, not sure if we're targeting 26.4

shawkins:
Ok, we'll create 2 follow-ups. One to make picocli honor the log color setting, and one to enable log color by default for KC 27.

mabartos:
> Ok, we'll create 2 follow-ups. One to make picocli honor the log color setting, and one to enable log color by default for KC 27.

Agree. Then, we can discuss more with others whether the color should be enabled by default or not.

PR #42400, opened 2025-09-06
testsuite: use KC_TEST_SERVER_STARTUP_TIMEOUT env var for distribution startup timeout

### LINKED ISSUE #39142: Make distribution startup timeout configurable
### Description

It'd be great to have a configurable startup timeout for Keycloak distribution in testsuite as some tests require a different timeout. 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

Mohamed-Moideen-Halith:
Closes #39142

This changes RawKeycloakDistribution to read KC_TEST_SERVER_STARTUP_TIMEOUT environment variable
(first), and fall back to the previous system property keycloak.distribution.start.timeout if not set.
Default remains 120 seconds. This avoids using ad-hoc system properties directly, and aligns the
testsuite configuration with KC_ environment variables.

Files changed:
 - quarkus/tests/junit5/.../RawKeycloakDistribution.java
 
 Note: I’m new to Keycloak contributions and was unable to run the full testsuite locally.

### DIFF
--- a/quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
+++ b/quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
@@ -398,7 +398,21 @@ public final class RawKeycloakDistribution implements KeycloakDistribution {
     }
 
     private long getStartTimeout() {
-        return TimeUnit.SECONDS.toMillis(Long.getLong("keycloak.distribution.start.timeout", 120L));
+        
+        String env = System.getenv("KC_TEST_SERVER_STARTUP_TIMEOUT");
+        if (env != null && !env.isBlank()) {
+            try {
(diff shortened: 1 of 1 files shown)


ahus1 (changes requested):
Thank you for this PR. Unfortunately the issue you're referencing is already closed. Please create a new issue first, and then reference it in your pull request. I assume you see this as an enhancement, and then it should describe why this change is beneficial. 

In addition to that, your commit needs a sign-off as described in `CONTRIBUTING.md`

stianst (changes requested):
We should support setting this as env, system properties, config file values, etc.

Please use `org.keycloak.testframework.config.Config.get` or you can inject the value like how debug is done in org.keycloak.testframework.server.DistributionKeycloakServerSupplier.

You should then pass the value within org.keycloak.testframework.server.DistributionKeycloakServer to RawKeycloakDistribution

PR #42404, opened 2025-09-07
Set aria-label on Switch control

### LINKED ISSUE #42403: ui-shared: Accessibility of Switch control
Relates to #36480

### Before reporting an issue

### Area

admin/ui

### Describe the bug

Label for the Switch control is unhelpful for screen reader users. When using tab navigation, only "On" or "Off" is reported instead of the label that visually appear on the left of the control. Furthermore, this is redundant as switches are reported as checked or unchecked checkboxes. Screen reader users can still navigate the document and find the label that is before the switch, but this becomes confusing as one must remember that the label comes before the switch. At least the `aria-label` attribute on the control should be set to the same value as the text on the left to override the useless "On"/"Off" value.

Also worth noting that the way this control is built is considered bad practice by the developers of the underlying library:

> To keep inline with accessibility guidelines, the label for a switch must never dynamically change. A dynamically changing label (such as one label for an "on" state and another label for an "off" state) can be confusing as the contexts for each label changes as the switch state does. This applies to both visible text labels as well as labels provided via aria-label.

[patternfly Switch documentation](https://www.patternfly.org/components/switch/)

### Version

main branch

### Regression

### Expected behavior

Label for switches must be the same for screen reader users as it is for everyone else.

### Actual behavior

When using tab navigation with a screen reader, only "On" or "Off" is reported instead of the label that visually appear on the left of the control.

### How to Reproduce?

With a screen reader on, tab into any switch such as the one labelled "Enabled" on the "Create realm" dialog, observe that the label that gets read is "On" instead of "Enabled".

### Anything else?

_No response_

DataTriny:
Closes #42403

### DIFF
--- a/js/libs/ui-shared/src/controls/SwitchControl.tsx
+++ b/js/libs/ui-shared/src/controls/SwitchControl.tsx
@@ -54,6 +54,7 @@ export const SwitchControl = <
             id={props.name}
             data-testid={debeerify(props.name)}
             label={labelOn}
+            aria-label={props.label}
             isChecked={stringify ? value === "true" : value}
             onChange={(e, checked) => {
               const value = stringify ? checked.toString() : checked;


ahus1 (changes requested):
Thank you for this pull request, the changes look good to me. 

Can you please add the sign-off for the commit as outlined in `CONTRIBUTING.md`? With that, it should be ready to be merged.

DataTriny:
@ahus1 I missed that, I'm sorry. I've rebased and signed-off the commit.

ahus1 (approved):
Thank you for this fix!

ahus1:
@DataTriny - this change will be available in the next minor release of Keycloak (26.4) which is expected to be published by end of September. 

There are some additional `<Switch />` tags in the projects. I would be happy if you could update them as well in a separate issue and pull request. I'll be happy to review them as well. 

If you are into JavaScript and want to have a look, maybe you can even come up with a compile time eslint validation, still that is very optional.

DataTriny:
@ahus1 Thanks for letting me know.

My motivation here is to improve the accessibility of the Admin UI as there are quite a few issues. I'm not that much into JavaScript, but if good practice for accessibility can be enforced through eslint then I might have a look at that as well.

ahus1:
I am also not so very deep into JavaScript to know what might be a good tool to use. 

Still consider it a possible area to explore it, as if we have something in the CI to check this, a developer cannot accidentally add this accessibility problems.

Either way, all accessibility PRs are welcome.

PR #42425, opened 2025-09-08
Keycloak CI - Refactoring ci.yml

### LINKED ISSUE #42423: Keycloak CI - refactoring ci.yml
### Description

https://github.com/keycloak/keycloak/blob/main/.github/workflows/ci.yml

* remove jvm heapdumps steps
* remove surefire-reports steps
* fixing job names for flaky test steps

lhanusov:
Closes #42423


stianst (changes requested):
Quite a few changes here that I'm not certain about.

`log/trimmer.sh` is just a work-around to integration tests generating too much logs, and GA killing the job due to that. Don't think that is something we want to add more places than needed.

Heapdumps I've asked a few times if anyone actually use, and it seems not, so I'd rather just remove those completely.

Same with Surefire reports. Although I guess the use-case here is to be able to report on what tests where executed for a given release?

Retries is also only something that only some jobs should use; ideally no jobs. So wouldn't add that too more places.

Could you update the description of the GH issue to provide some more details on what you are actually trying to fix/improve?

stianst (approved):

PR #42428, opened 2025-09-08
[OID4VC]: Update the issuer metadata for signed metadata

forkimenjeckayang:
## Summary

Update metadata retrieval to support signed metadata `(application/jwt)` alongside unsigned `(application/json)`, including sending an Accept header and validating JWS headers `(alg, typ: openidvci-issuer-metadata+jwt)` and JWT claims `(iss, sub, iat, exp)`.

See: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-signed-metadata

## Cosses
- Issue: https://github.com/keycloak/keycloak/issues/41588

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
@@ -18,10 +18,15 @@
 package org.keycloak.protocol.oid4vc.issuance;
 
 import jakarta.ws.rs.core.UriInfo;
+import org.apache.http.HttpHeaders;
+import org.keycloak.common.util.Time;
 import org.keycloak.constants.Oid4VciConstants;
 import org.keycloak.crypto.KeyUse;
 import org.keycloak.crypto.KeyWrapper;
+import org.keycloak.crypto.SignatureProvider;
+import org.keycloak.crypto.SignatureSignerContext;
 import org.keycloak.jose.jwe.JWEConstants;
+import org.keycloak.jose.jws.JWSBuilder;
 import org.keycloak.models.KeyManager;
 import org.keycloak.models.KeycloakContext;
 import org.keycloak.models.KeycloakSession;
@@ -34,20 +39,34 @@ import org.keycloak.protocol.oid4vc.model.CredentialIssuer;
 import org.keycloak.protocol.oid4vc.model.CredentialResponseEncryptionMetadata;
 import org.keycloak.protocol.oid4vc.model.CredentialRequestEncryptionMetadata;
 import org.keycloak.protocol.oid4vc.model.SupportedCredentialConfiguration;
+import org.keycloak.representations.JsonWebToken;
 import org.keycloak.protocol.oidc.utils.JWKSServerUtils;
 import org.keycloak.services.Urls;
 import org.keycloak.urls.UrlType;
+import org.keycloak.util.JsonSerialization;
+import org.keycloak.utils.MediaType;
 import org.keycloak.wellknown.WellKnownProvider;
 import org.jboss.logging.Logger;
 import org.keycloak.jose.jwk.JWK;
 import org.keycloak.jose.jwk.JSONWebKeySet;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.keycloak.constants.Oid4VciConstants.SIGNED_METADATA_JWT_TYPE;
+import static org.keycloak.crypto.Algorithm.ES256;
+import static org.keycloak.crypto.Algorithm.ES384;
+import static org.keycloak.crypto.Algorithm.ES512;
+import static org.keycloak.crypto.Algorithm.PS256;
+import static org.keycloak.crypto.Algorithm.PS384;
+import static org.keycloak.crypto.Algorithm.PS512;
+import static org.keycloak.crypto.Algorithm.RS384;
+import static org.keycloak.crypto.Algorithm.RS512;
 import static org.keycloak.crypto.KeyType.RSA;
+import static org.keycloak.jose.jwk.RSAPublicJWK.RS256;
 
 /**
  * {@link WellKnownProvider} implementation to provide the .well-known/openid-credential-issuer endpoint, offering
@@ -58,13 +77,24 @@ import static org.keycloak.crypto.KeyType.RSA;
  */
 public class OID4VCIssuerWellKnownProvider implements WellKnownProvider {
 
-    public static final String VC_KEY = "vc";
-
     private static final Logger LOGGER = Logger.getLogger(OID4VCIssuerWellKnownProvider.class);
 
-    protected final KeycloakSession keycloakSession;
+    // Realm attributes for signed metadata configuratio
+    public static final String SIGNED_METADATA_ENABLED_ATTR = "oid4vci.signed_metadata.enabled";
+    public static final String SIGNED_METADATA_LIFESPAN_ATTR = "oid4vci.signed_metadata.lifespan";
+    public static final String SIGNED_METADATA_ALG_ATTR = "oid4vci.signed_metadata.alg";
 
+    public static final String VC_KEY = "vc";
     public static final String ATTR_ENCRYPTION_REQUIRED = "oid4vci.encryption.required";
+
+    // Static list of supported asymmetric algorithms
+    private static final List<String> ASYMMETRIC_ALGORITHMS = Arrays.asList(
+            RS256, RS384, RS512,
+            ES256, ES384, ES512,
+            PS256, PS384, PS512
+    );
+
+    protected final KeycloakSession keycloakSession;
 
     public OID4VCIssuerWellKnownProvider(KeycloakSession keycloakSession) {
         this.keycloakSession = keycloakSession;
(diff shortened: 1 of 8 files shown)


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -58,14 +77,25 @@
  */
 public class OID4VCIssuerWellKnownProvider implements WellKnownProvider {
 
-    public static final String VC_KEY = "vc";
-
     private static final Logger LOGGER = Logger.getLogger(OID4VCIssuerWellKnownProvider.class);
 
-    protected final KeycloakSession keycloakSession;
+    // Realm attributes for signed metadata configuratio
```
wistefan: Typo, there is a missing n at the end;)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -77,17 +107,43 @@ public void close() {
...
+    public Object getMetadataResponse(CredentialIssuer issuer, KeycloakSession session) {
+        RealmModel realm = session.getContext().getRealm();
+        String acceptHeader = session.getContext().getRequestHeaders().getHeaderString(HttpHeaders.ACCEPT);
+        boolean preferJwt = acceptHeader != null && acceptHeader.contains(MediaType.APPLICATION_JWT);
+        boolean signedMetadataEnabled = Boolean.parseBoolean(realm.getAttribute(SIGNED_METADATA_ENABLED_ATTR));
+
+        if (preferJwt && signedMetadataEnabled) {
+            String signedJwt = null;
+            try {
+                signedJwt = generateSignedMetadata(issuer, session);
```
wistefan: How could that method return null? It already checks and throws exceptions. The method would be cleaner if you only handle the exceptions and not do an additional null check.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -125,7 +181,126 @@ public static CredentialIssuer.BatchCredentialIssuance getBatchCredentialIssuanc
...
+        RealmModel realm = session.getContext().getRealm();
+        KeyManager keyManager = session.keys();
+
+        // Select asymmetric signing algorithm
+        String alg;
+        try {
+            alg = getSigningAlgorithm(realm, session);
+        } catch (IllegalStateException e) {
+            LOGGER.warnf("Failed to get signing algorithm: %s. Falling back to unsigned metadata.", e.getMessage());
+            return null; // Return null to indicate fallback to JSON
```
wistefan: In order to be null-safe, better bubble the exception or use Optional

wistefan (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
forkimenjeckayang: Thank you @wistefan for the feedback. The `null` return is intentional for spec compliance.
According to OID4VCI spec, `signed metadata` is optional (MAY support) while unsigned JSON is mandatory (MUST support). The method returns null for non-critical config issues (invalid algorithm/lifespan) to enable graceful fallback to unsigned JSON,
Only critical errors (missing keys, signature failures) throw exceptions. 
Is that okay or clarify why we have that?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
forkimenjeckayang: Based on the reason i gave above, I will like you to take a look again.
Will be waiting, so i know how to proceed with this.
Thanks again @wistefan

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -58,14 +77,25 @@
...
 
+    public static final String VC_KEY = "vc";
     public static final String ATTR_ENCRYPTION_REQUIRED = "oid4vci.encryption.required";
 
+    // Static list of supported asymmetric algorithms
+    private static final List<String> ASYMMETRIC_ALGORITHMS = Arrays.asList(
+            RS256, RS384, RS512,
+            ES256, ES384, ES512,
+            PS256, PS384, PS512
+    );
```
Captain-P-Goldfish: Wouldn't it be better to get the available algorithms from the available keys in the realm instead of hardcoding them here?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
wistefan: But you are falling back in any case to null, be it an exception or a null return. The generateSignedMetadata-method only returns null in case of exceptions(actually one exception). The code would be much easier to read if you either handle all cases by exceptions(that you catch and map to null here) or handle all exceptions in the generateSignedMetadata-method and return an Optional(to make it NPE safe) from that method. Current solution mixes null and exceptions, thus you need the double-handling that you have here.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
wistefan: There needs to be a filter on the realm-keys, since only asymmetric keys can be used.  We have symmetric keys by default in Keycloak, thus the need to be filtered out.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
forkimenjeckayang: Thank you @wistefan for the clarification. You're right about the mixed approach being inconsistent. WIll adjust that

forkimenjeckayang:
@Captain-P-Goldfish and @wistefan 
Please could you look into this again

wistefan (approved):
LGTM

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
mposolda: +1 to avoid hardcoded constants. Whenever new algorithm is introduced, the constant cannot be updated automatically, which may result in the bug that well-known endpoint will not have up-to-date list of algorithms.

There is method `OIDCWellKnownProvider.getSupportedAsymmetricAlgorithms()` , which has support to only show asymmetric algorithms. Could that one be re-used? It may be good to eventually add it to some "util" class, so that same code can be re-used between `OIDCWellKnownProvider` and `OID4VCIssuerWellKnownProvider` .

There is one thing, that asymmetric algorithms supported by Keycloak also have `EdDSA` . Is it correct that this algorithm would be also included here?

forkimenjeckayang: Thank you @mposolda for the  feedback!
- Removed hardcoded constants - now dynamically discovers algorithms from SignatureProvider implementations
- Created shared utility - `CryptoUtils.getSupportedAsymmetricSignatureAlgorithms()` that both `OIDCWellKnownProvider` and `OID4VCIssuerWellKnownProvider` now use
- EdDSA included - yes, it's correctly included as it's an asymmetric algorithm

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -472,24 +463,12 @@ public static List<String> getSupportedSignatureAlgorithms(KeycloakSession sessi
...
-                .filter(key -> KeyUse.SIG.equals(key.getUse()))
-                .filter(key -> key.getStatus().isEnabled() && key.getPublicKey() != null)
-                .filter(key -> KeyType.RSA.equals(key.getType()) ||
-                        KeyType.EC.equals(key.getType()) ||
-                        KeyType.OKP.equals(key.getType()))
-                .map(KeyWrapper::getAlgorithm)
-                .filter(algorithm -> algorithm != null && !algorithm.isEmpty())
-                .distinct()
-                .collect(Collectors.toList());
+        return CryptoUtils.getSupportedAsymmetricSignatureAlgorithms(session);
```
Captain-P-Goldfish, line 471: I’m not dead set on it, but you could completely remove this method.

besides that everything looks good.

Captain-P-Goldfish (approved):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
mposolda: @Captain-P-Goldfish Yes, true. But I am also not dead set on it, so rather merging since you already approved, so we can move forward. Thanks!

mposolda (approved):

PR #42435, opened 2025-09-08
Document tested and supported configurations for single-cluster deployments

### LINKED ISSUE #42304: Document tested and supported configurations for single-cluster deployments
### Description

We should update the single-cluster HA guide to document the supported and test configurations for both upstream and downstream.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
CLoses #42304

### DIFF
--- a/docs/documentation/upgrading/topics/intro.adoc
+++ b/docs/documentation/upgrading/topics/intro.adoc
@@ -2,7 +2,7 @@
 
 == Upgrading {project_name}
 
-This guide describes how to upgrade {project_name}. Use the following procedures in this order:
+This {section} describes how to upgrade {project_name}. Use the following procedures in this order:
 
 . Review the migration changes from the previous version of {project_name}.
 . Upgrade the {project_name} server.
--- a/docs/guides/high-availability/introduction.adoc
+++ b/docs/guides/high-availability/introduction.adoc
@@ -12,7 +12,7 @@ for your deployments.
 
 == Architectures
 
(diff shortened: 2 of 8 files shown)


ahus1 (commented):
@ryanemerson - I looked at the changes and IMHO it makes sense to state both in upstream and downstream what we test with and also what we support. 

Still the message on what we support is a little bit different in upstream and downstream. Please have a look.

ryanemerson:
@ahus1 Makes sense to me, although we don't do that in the multi-cluster guide. Should we update that as well to list both the tested and supported configurations?

I have pushed an additional commit to remove repetition.

ahus1:
> Should we update that as well to list both the tested and supported configurations?

Although not essential, please give it a try if it can be done. I would assume for the multi-cluster setup the tested and supported configurations would be very similar, but let's find out.

ryanemerson:
I have pushed a new commit with the changes. The two sections are very similar like you say, but I have tried to differentiate the two cases by referring to "Kubernetes" in the community supported configurations. If we think this is too far though, I'm happy for the commit to be reverted.

I had to force push as my original commit referenced the wrong issue, fixed!

ahus1 (dismissed):
Thank you for this PR. I pushed one more review comment after some internal feedback. If there are any additional changes necessary, please create another PR. Thanks!

ahus1 (approved):

PR #42436, opened 2025-09-08
Avoid exception when refresh_expires_in is null

### LINKED ISSUE #41804: OIDC identity provider token refresh fails with JsonMapperException
### Before reporting an issue

### Area

identity-brokering

### Describe the bug

Token refresh with identity brokering fails with JsonMapperException. I did some debugging, and it seems the root cause is that refreshExpiresIn is null. In my integration scenario, i use PING Federate. It seems PING is not providing refresh_expires_in with a token refresh response. According to spec, this is allowed. 

### Version

nightly from 12.08.2025

### Regression

### Expected behavior

In AbstractOAuth2IdentityProvider.refreshToken, please take over the refreshExpiresIn from previousResponse, as you did for the id token. That should solve the issue. 

### Actual behavior

OIDC id broker token refresh fails with JsonMapperException

### How to Reproduce?

Integrate Keycloak with a OIDC provider that is not providing refresh_expires_in with token refresh response, e.g. PING Federate

### Anything else?

It is my first bug report. Hope it is provided with sufficient details.

pematth:
Closes #41804

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
@@ -18,6 +18,7 @@ package org.keycloak.broker.oidc;
 
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -210,11 +211,11 @@ public abstract class AbstractOAuth2IdentityProvider<C extends OAuth2IdentityPro
             this.refreshToken = refreshToken;
         }
 
-        public long getRefreshExpiresIn() {
+        public Long getRefreshExpiresIn() {
             return refreshExpiresIn;
         }
 
-        public void setRefreshExpiresIn(long refreshExpiresIn) {
+        public void setRefreshExpiresIn(Long refreshExpiresIn) {
             this.refreshExpiresIn = refreshExpiresIn;
         }
 
@@ -297,6 +298,8 @@ public abstract class AbstractOAuth2IdentityProvider<C extends OAuth2IdentityPro
 
                 if (newResponse.getRefreshToken() == null && previousResponse.getRefreshToken() != null) {
                     newResponse.setRefreshToken(previousResponse.getRefreshToken());
+                }
+                if (newResponse.getRefreshExpiresIn() == null && previousResponse.getRefreshExpiresIn() != null) {
                     newResponse.setRefreshExpiresIn(previousResponse.getRefreshExpiresIn());
                 }
                 if (newResponse.getIdToken() == null && previousResponse.getIdToken() != null) {
--- a/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
@@ -18,7 +18,6 @@ package org.keycloak.broker.oidc;
 
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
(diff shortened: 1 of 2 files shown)


martin-kanis (changes requested):
@pematth Thanks for the PR!
I was thinking about how to test it and we can add following tests to `AbstractOAuth2IdentityProviderTest`.

```
@Test
	public void oAuthResponse_nullRefreshExpiresIn_shouldNotThrowException() throws IOException {
		// Test that OAuthResponse can handle null refreshExpiresIn during JSON serialization/deserialization
		AbstractOAuth2IdentityProvider.OAuthResponse response = new AbstractOAuth2IdentityProvider.OAuthResponse();
		response.setToken("test_token");
		response.setRefreshToken("test_refresh");
		response.setRefreshExpiresIn(null); // This should not cause NPE

		// Should not throw JsonMapperException during serialization
		String json = JsonSerialization.writeValueAsString(response);
		Assert.assertNotNull(json);

		// Should not throw exception during deserialization
		AbstractOAuth2IdentityProvider.OAuthResponse deserialized =
			JsonSerialization.readValue(json, AbstractOAuth2IdentityProvider.OAuthResponse.class);
		Assert.assertNotNull(deserialized);
		Assert.assertNull(deserialized.getRefreshExpiresIn()); // Should return null, not throw NPE
	}

	@Test  
	public void oAuthResponse_missingRefreshExpiresInField_shouldNotThrowException() throws IOException {
		// Test JSON without refresh_expires_in field
		String jsonWithoutRefreshExpires = "{\"access_token\":\"test\",\"refresh_token\":\"refresh\"}";
		
		// Should not throw JsonMapperException during deserialization
		AbstractOAuth2IdentityProvider.OAuthResponse response = 
			JsonSerialization.readValue(jsonWithoutRefreshExpires, AbstractOAuth2IdentityProvider.OAuthResponse.class);
		
		Assert.assertNotNull(response);
		Assert.assertEquals("test", response.getToken());
		Assert.assertEquals("refresh", response.getRefreshToken());
		Assert.assertNull(response.getRefreshExpiresIn()); // Should return null, not throw NPE
	}
```

Tests also discovered the same problem with `getExpiresIn` and `setExpiresIn` getter/setter. Please change it the to `Long` as well. Thank you.

PS: You can also link this PR to the issue by adding `Fixes #41804` to the commit message or the description of this PR. Then the issue will be automatically closed once this PR is merged.

pematth (commented):
implemented review feedback

martin-kanis (changes requested):
Thanks for incorporating the changes. I noticed we should also add a `null` check to `getExpiresIn()` usages i.e `tokenResponse.getExpiresIn() != null && tokenResponse.getExpiresIn() > 0`
Can you add it please?

pematth:
@martin-kanis , thanks for the hint, added not null check now also for expires_in checking

martin-kanis (approved):
Thanks @pematth. LGTM

pedroigor:
@pematth @martin-kanis Added a commit to make sure the expiration is set together when setting the previous/stored refresh token. I don't think they should be done separately.

Also, if you think we can remove the new test methods, please let me know. I don't think they make much sense now that we have the field as a wrapper class.

pedroigor (commented):
Can you please check my last comment and commit?

pematth:
Hi @pedroigor @martin-kanis , the initial issue is caused because we get with refresh a new RT but no refresh_expires_in param. My code proposal could handle that. But with your proposal we now would not fetch the refresh_expires_in from previous response. Please revert that commit, or explain if i missed something.

pedroigor:
@pematth Why do you want to set the previous `refresh_expires_in` for a new RT if the IdP did not send it back in the response?

The code today is only setting `refresh_expires_in` from the previous response if no RT was issued and when re-using the previous RT. It does not make much sense to me to set an expiration time on a new RT issued by IdP based on the previous RT.

That is the problem when using non-standard fields, we don't know the exact behavior/sematics. For instance, what does it mean an RT without a `refresh_expires_in` ?

Also, I do not see `refresh_expires_in` being used anywhere to check the validity of RTs. But checks based on the access token (response) expiration.

pematth:
@pedroigor Right, refresh_expires_in is not checked, we just try to refresh until it fails, and never check for timeouts. I am not sure what exactly happens at refresh token timeout at the broker. Its a bit tricky with Keycloak token refresh and broker token refresh and their different refresh token timeouts. But coming back to your code change. After thinking a bit more about it, i think it will work for my scenario.

pedroigor (approved):

PR #42439, opened 2025-09-08
Avoid removing client sessions before the user session times out

### LINKED ISSUE #35825: Per client session idle time capped by realm level client idle timeout
### Before reporting an issue

### Area

core

### Describe the bug

When specifying a per-client session idle time together with persistent sessions, the client sessions are removed once the timeout on the realm level is reached. 

There is currently no indication for the user that this is the case. 

### Version

main & 26.0.7

### Regression

### Expected behavior

Make it safe to use: Either allow a client to override the realm's client idle time with a longer value (up to the session time), or show an or keep the current behavior. Any invalid values should should show errors on save, possibly also when opening the UI to show misconfigurations in existing values.

### Actual behavior

UI accepts the values without any warnings or errors. Client sessions are removed once the realm's client session timeout is reached. 

### How to Reproduce?

The following code doesn't look at individual clients: 

https://github.com/keycloak/keycloak/blob/9861acc2aaa6b8d588ef1155563c1f652ec08d28/model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java#L257-L259

### Anything else?

_No response_

ahus1:
As the client session timeout can be overwritten on a per client level, the realm level timeout can not be used to remove client sessions early.

Closes #35825

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
@@ -47,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


ahus1:
@pruivo - can you please review this backport when you have the time. Thanks!

pruivo (approved):

abstractj (approved):

Issue #38671, opened 2025-04-03
Duplicate Key Violation When Reauthenticating After Account Deletion via Google


antonmartsenyuk:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

authentication

### Describe the bug

**Issue Description:**  
We encountered an issue in Keycloak **26.1.4** where users cannot immediately reauthenticate after deleting their account when using Google (or another external identity provider). The error message indicates a **duplicate key violation** in the database.  

### Version

26.1.4

### Regression

- [ ] The issue is a regression

### 🔍 **Expected Behavior:**  
The user should be able to reauthenticate successfully as if they are logging in for the first time.  

### 🔄 **Steps to Reproduce:**  
1. Configure authentication via **Google** (or another external provider).  
2. Navigate to **Authentication → Required Actions** and enable **"Delete Account"**.  
3. Go to **Realm Settings → User Registration → Default Roles**, add the **"Delete Account"** role.  
4. Log in with **Google**, then delete the account through the UI.  
5. Immediately attempt to log in again with Google.  
6. The following error occurs:  

```
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "constraint_offl_us_ses_pk2"
Detail: Key (user_session_id, offline_flag)=(4eb36d91-6d49-442f-8fad-9be63896d89b, 0) already exists.
```

### 🔄 **Steps to Reproduce:**  
1. Configure authentication via **Google** (or another external provider).  
2. Navigate to **Authentication → Required Actions** and enable **"Delete Account"**.  
3. Go to **Realm Settings → User Registration → Default Roles**, add the **"Delete Account"** role.  
4. Log in with **Google**, then delete the account through the UI.  
5. Immediately attempt to log in again with Google.  
6. The following error occurs:  

```
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "constraint_offl_us_ses_pk2"
Detail: Key (user_session_id, offline_flag)=(4eb36d91-6d49-442f-8fad-9be63896d89b, 0) already exists.
```

### Anything else?

Keycloak Version: 26.1.4
Database: PostgreSQL

⚠ Important Note:
The issue resolves itself after approximately 5 minutes, allowing the user to log in successfully again. However, the immediate failure creates a poor user experience and may indicate an underlying issue with session or user data cleanup.

[...]

PR #42451, opened 2025-09-08
Handle already existing user session in the store

ahus1:
Closes #40374

Follow-ups:

- [ ] add more validations for an existing user session
- [ ] apply validation of existing user session also in other places (Embedded Infinispan, Standard Infinispan, Remote Infinispan)
- [ ] follow-up issue to validate generated IDs of an authentication session

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
@@ -278,6 +278,8 @@ public class JpaChangesPerformer<K, V extends SessionEntity> implements SessionC
                 }
             });
             clientSessionModel.getUpdatedModel();
+        } else {
+            LOG.debugf("No client session found for %s/%s", entity.getUserSessionId(), entity.getClientId());
         }
     }
 
@@ -433,6 +435,8 @@ public class JpaChangesPerformer<K, V extends SessionEntity> implements SessionC
                 PersistentUserSessionAdapter userSessionModel = (PersistentUserSessionAdapter) userSessionPersister.loadUserSession(realm, entry.getKey().toString(), entity.isOffline());
                 if (userSessionModel != null) {
                     mergeUserSession(session, entry, userSessionModel, realm, sessionUpdates, userSessionPersister, entity);
+                } else {
+                    LOG.debugf("No user session found for %s", entry.getKey());
                 }
             }
         }
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
@@ -22,6 +22,7 @@ import org.jboss.logging.Logger;
 import org.keycloak.models.AuthenticatedClientSessionModel;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.ModelIllegalStateException;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.models.UserSessionModel;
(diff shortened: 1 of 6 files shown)


ahus1:
Hello @mposolda and @pruivo - I had a look how `EmbeddedCachesChangesPerformer` implements it by applying all updates to the entity retrieved from the cache. 

https://github.com/keycloak/keycloak/blob/4dc4de7c1218bd129b4ae6853a74b7a2954b6eec/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/EmbeddedCachesChangesPerformer.java#L60-L72

So I reworked JpaChangesPerformer to first read from the database before then continuing to either create the entity updating it. 

This raises the cost of creating a new session by one additional SELECT as part of an already existing transaction. While this is a round-trip to the database which is costly, the impact on the database should be smaller than the already existing INSERT. This could be in a later step be optimized in a more opportunistic way to try a plain insert without a select on the first attempt, and a select on any subsequent attempt. 

If that transaction will fail, it will then retry as part of the existing retry mechanism we have in place in either `PersistentSessionsWorker` (batching enabled) or `PersistentSessionsChangelogBasedTransaction` (batching disabled). 

Please let me know your thoughts. Can you think of any additional checks that I should add if a session is found in addition to the session being new and is matching the current user?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java
```diff
@@ -189,7 +189,10 @@ public AuthenticatedClientSessionModel createClientSession(RealmModel realm, Cli
     public UserSessionModel createUserSession(String id, RealmModel realm, UserModel user, String loginUsername, String ipAddress,
                                               String authMethod, boolean rememberMe, String brokerSessionId, String brokerUserId, UserSessionModel.SessionPersistenceState persistenceState) {
         if (id == null) {
-            id = keyGenerator.generateKeyString(session, sessionTx.getCache());
+            do {
+                id = keyGenerator.generateKeyString(session, sessionTx.getCache());
+                // prevent potential duplicate ID generation
+            } while (sessionTx.get(id) != null);
```
pruivo: uh!? What's the probability of a conflict? 1 in 2^122? It seems unnecessary.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
```diff
@@ -395,132 +397,151 @@ private void processUserSessionUpdate(KeycloakSession innerSession, Map.Entry<K,
...
+                } else if (merged.getOperation() == SessionUpdateTask.CacheOperation.ADD_IF_ABSENT) {
+                    // This might happen if the user logs in via multiple tabs at the same time from an external broker, and the same authentication session creates
+                    // multiple user sessions concurrently.
+                    if (!Objects.equals(userSessionModel.getUserId(), entity.getUser())) {
+                        throw new ModelIllegalStateException("User ID of the session does not match, the user ID should not change");
+                    }
+                    if (Math.abs(userSessionModel.getStarted() - entity.getStarted()) > 10) {
+                        throw new ModelIllegalStateException("Session has already aged, concurrent requests should not happen");
+                    }
+                    // If the above matches, it is safe to continues to apply all updates to the entity as before in EmbeddedCachesTransformer
```
pruivo: This class is enormous and difficult to read, but I think this is dead code. You're invoking `createUserSession` before doing `loadUserSession`, so it should return the entity you just created.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
```diff
@@ -395,132 +397,151 @@ private void processUserSessionUpdate(KeycloakSession innerSession, Map.Entry<K,
...
 
-                @Override
-                public void restartSession(RealmModel realm, UserModel user, String loginUsername, String ipAddress, String authMethod, boolean rememberMe, String brokerSessionId, String brokerUserId) {
-                    throw new IllegalStateException("not implemented");
-                }
-            }, entity.isOffline());
-        } else {
             PersistentUserSessionAdapter userSessionModel = (PersistentUserSessionAdapter) userSessionPersister.loadUserSession(realm, entry.getKey().toString(), entity.isOffline());
             if (userSessionModel != null) {
+                if (merged.getOperation() == SessionUpdateTask.CacheOperation.ADD) {
```
pruivo: I don't think we use `CacheOperation.ADD` anywhere in the code. Can we eliminate it to simplify things a bit?

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
```diff
@@ -311,7 +311,7 @@ public UserSessionModel loadUserSession(RealmModel realm, String userSessionId,
         userSessionQuery.setParameter("lastSessionRefresh", calculateOldestSessionTime(realm, offline));
         userSessionQuery.setMaxResults(1);
 
-        Stream<OfflineUserSessionModel> persistentUserSessions = closing(userSessionQuery.getResultStream().map(this::toAdapter));
+        Stream<OfflineUserSessionModel> persistentUserSessions = closing(userSessionQuery.getResultStream().map(this::toAdapter).filter(Objects::nonNull));
```
pruivo: Why not use `userSessionQuery.getSingleResultOrNull()`?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderOfflineTest.java
```diff
@@ -89,16 +85,15 @@ public void after() {
...
+            createSessions(currentSession);
         });
 
-        KeycloakModelUtils.runJobInTransaction(session.getKeycloakSessionFactory(), session.getContext(), (KeycloakSession sessionCrud2) -> {
-            currentSession = sessionCrud2;
-            realm = currentSession.realms().getRealmByName("test");
-            sessionManager = new UserSessionManager(currentSession);
+        KeycloakModelUtils.runJobInTransaction(session.getKeycloakSessionFactory(), session.getContext(), currentSession -> {
+            RealmModel realm = currentSession.realms().getRealmByName("test");
+            UserSessionManager sessionManager = new UserSessionManager(currentSession);
```
pruivo: unused

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderOfflineTest.java
```diff
@@ -270,9 +261,8 @@ public void testOnRealmRemoved(KeycloakSession session) {
     @ModelTest
     public void testOnClientRemoved(KeycloakSession session) {
         AtomicReference<String> userSessionID = new AtomicReference<>();
-        String realmId = KeycloakModelUtils.runJobInTransactionWithResult(session.getKeycloakSessionFactory(), (KeycloakSession sessionCR1) -> {
-            currentSession = sessionCR1;
-            sessionManager = new UserSessionManager(currentSession);
+        String realmId = KeycloakModelUtils.runJobInTransactionWithResult(session.getKeycloakSessionFactory(), currentSession -> {
+            UserSessionManager sessionManager = new UserSessionManager(currentSession);
```
pruivo: unused

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderOfflineTest.java
```diff
@@ -509,14 +488,13 @@ public static void reloadState(KeycloakSession session) {
...
-        currentSession = session;
-        realm = currentSession.realms().getRealmByName("test");
+    public static void reloadState(KeycloakSession currentSession, Boolean initialConfig) {
+        RealmModel realm = currentSession.realms().getRealmByName("test");
         if (initialConfig) {
             currentSession.users().addUser(realm, "user1").setEmail("user1@localhost");
             currentSession.users().addUser(realm, "user2").setEmail("user2@localhost");
         }
-        sessionManager = new UserSessionManager(currentSession);
+        UserSessionManager sessionManager = new UserSessionManager(currentSession);
```
pruivo: unsued

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java
ahus1: OK, I'll leave that for a separate PR

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
ahus1: Actually the `createUserSession` was left-over from a manual test; I should have never added it.

This is now cleaned up, and there is a new test `testConcurrentSessionCreation` which tests the functionality even without concurrency.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
ahus1: This is "just" a safeguard, and I would like to keep it. The operation is used with other caches, and unfortunately this enum isn't specific to this cache.

As I don't know how to simplify it, I would keep it as is.

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
ahus1: Rewritten to be imperative and not using streams.

ahus1:
@pruivo - I've reworked the code, please have another look. The unused snippets in the test class have been cleaned up together with other warnings.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
pruivo: A `switch` may be simpler to read
```java
        switch (merged.getOperation()) {
            case REMOVE:
                remove();
                break;
            case ADD, ADD_IF_ABSENT:
                entity = load();
                if (entity == null) {
                    create();
                } else {
                    merge();
                }
                break;
            case REPLACE:
                entity = load();
                if (entity != null) {
                    merge();
                }
                break;
        }
```

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
pruivo: I still don't understand these checks, mainly because we don't care for those with volatile sessions; we just merge the changed fields. 
My other concern is that it will fail all 20 retries we have.

ahus1: I'd say this is defensive programming, as bad things would happen if we connect a new user to an existing session that is bound to a different user. A user ID is assumed to be constant.

As an example: In #38671 where a user deleted their account, we forgot to start a new authentication session, and that lead to the next log in in that browser tab to use the same user session ID. The database showed a duplicate key error. If we would just continue with the old session, we would continue with a wrong user in that session. 

Yes, I agree those 20 retries will fail, and I'd rather see the error and follow up on the cause, than having two user sessions mixed up (which is IMHO the very worst case that could happen for an IAM). 

Checking the user ID is IMHO the minimal check. I'd rather add more fields to the check, but probably not as a backport and only if Marek agrees. We could have a follow-up issue for the other stores as well.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
ahus1: I addressed this in a refactoring. Please have a look.

mposolda:
@ahus1 @pruivo Thanks a lot for the updates! I will be fine with whatever optimizations at the store level you guys agree with. I will try to do some tests later today or tomorrow to confirm if the fix actually works for me to fix the issue . I will probably use the steps from [this comment,](https://github.com/keycloak/keycloak/issues/40374#issuecomment-3188494058) which I've used before to simulate the issue.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
pruivo: I find it very confusing to have the same session ID for different users. It looks like a problem in the authentication flow or in the cookies. 

Other suggestions:
* I would like a comment about the `started` timestamp and why the value 10.
* Do not retry when `ModelIllegalStateException` is thrown.

ahus1: @pruivo - Thank you for the feedback. The latest commit addresses your review comments. There are now code comments explaining the why, and the retry is skipped on a `ModelIllegalStateException`

ahus1: I found that the I needed to revert the change and keep the retries. In run https://github.com/keycloak/keycloak/actions/runs/17609154462/job/50027163556?pr=42451 the test `UserSessionConcurrencyTest.testConcurrentNotesChange` showed that there can be an optimistic lock exception that is then later converted to an `ModelIllegalStateException` here: https://github.com/keycloak/keycloak/blob/b743b3d3b1884dee44c493c2ca043cef6aa8c26c/server-spi-private/src/main/java/org/keycloak/connections/jpa/support/EntityManagerProxy.java#L148

I'd say this retry is rare and won't have a performance implications as those new exceptions would only occur in error cases.

ahus1:
PR rebased to resolve conflicts. Please let me know if you are ok to merge this one @mposolda

mposolda (approved):
@ahus1 Thanks! I've did some testing and not able to simulate anymore the scenario from https://github.com/keycloak/keycloak/issues/40374#issuecomment-3188494058 . So from the functional perspective, it looks good to me and I am approving.

I did not review the details of the storage changes and did not review performance. Might be good to have review/approve from @pruivo as well, but leaving it to you.

ahus1:
@pruivo - can you please re-approve?

pruivo (approved):

PR #42452, opened 2025-09-08
Fixes to e-mail templates

sguilhen:

### DIFF
--- a/services/src/main/java/org/keycloak/email/freemarker/FreeMarkerEmailTemplateProvider.java
+++ b/services/src/main/java/org/keycloak/email/freemarker/FreeMarkerEmailTemplateProvider.java
@@ -231,8 +231,14 @@ public class FreeMarkerEmailTemplateProvider implements EmailTemplateProvider {
             attributes.put("properties", theme.getProperties());
             attributes.put("realmName", getRealmName());
             attributes.put("user", new ProfileBean(user, session));
-            KeycloakUriInfo uriInfo = session.getContext().getUri();
-            attributes.put("url", new UrlBean(realm, theme, uriInfo.getBaseUri(), null));
+
+            try {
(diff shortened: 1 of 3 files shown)


pedroigor (approved):

services/src/main/java/org/keycloak/email/freemarker/FreeMarkerEmailTemplateProvider.java
```diff
@@ -231,8 +231,14 @@ protected EmailTemplate processTemplate(String subjectKey, List<Object> subjectA
...
             attributes.put("user", new ProfileBean(user, session));
-            KeycloakUriInfo uriInfo = session.getContext().getUri();
-            attributes.put("url", new UrlBean(realm, theme, uriInfo.getBaseUri(), null));
+
+            try {
+                KeycloakUriInfo uriInfo = session.getContext().getUri();
+                attributes.put("url", new UrlBean(realm, theme, uriInfo.getBaseUri(), null));
+            } catch (Exception e) {
+                // ignore when running without an active request context such as sending emails from an scheduled task
+                // TODO: make it possible to make the URL available to email templates based on the hostname configured in the realm or at the server level
```
ahus1, line 240: @sguilhen - it makes me shiver to just swallow the exception without even adding a debug log. It also catches any exception, which is very broad. I would prefer a more explicit check. Same below.

pedroigor: I agree we can catch a more specific exception.

sguilhen: A fix will be coming for this - apologies for the broad try-catch, it should have been a draft PR as we were still investigating how to better handle sending emails from a timer task

Issue #35393, opened 2024-11-28
Add volatile sessions suite to the test framework


stianst:
Introduce `VolatileSessionsTestSuite.java` in [suites](https://github.com/keycloak/keycloak/tree/main/tests/base/src/test/java/org/keycloak/test/suites) and execute these tests as part of [Volatile Sessions IT](https://github.com/keycloak/keycloak/blob/fda6dde44f676bb151ee89ad2fe3e1bca28e21eb/.github/workflows/ci.yml#L410)

lhanusov:
implemented within #41318

stianst:
Need to update what tests are executed (including also migrating some tests I guess)

PR #42462, opened 2025-09-09
[Test Migration] New testsuites: Clusterless, Multisite, VolatileSessions, migrated test: SessionTest

### LINKED ISSUE #35391: Add clusterless suite to the new tests

lhanusov:
Closes #35391, #35393 and #42619

Depends on #42172 and partially #42425


.github/workflows/ci.yml
```diff
@@ -473,6 +473,23 @@ jobs:
           name: store-it-mvn-logs
           path: .github/scripts/ansible/files
 
+  new-volatile-sessions-tests:
```
stianst: No need for a `new-volatile-sessions-tests` instead just add a step to the existing `volatile-sessions-tests` that runs the new tests after the old testsuite is done.

.github/workflows/ci.yml
```diff
@@ -513,6 +530,27 @@ jobs:
         with:
           job-id: remote-infinispan-integration-tests
 
+  new-external-infinispan-tests:
```
stianst: Same as above, just add a step to `external-infinispan-tests` instead of a separate job

tests/base/pom.xml
```diff
@@ -124,6 +124,11 @@
             <artifactId>jboss-logmanager</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
stianst: Would be better not to require this; and instead just use FasterXML as we use everywhere else

.github/workflows/ci.yml
```diff
@@ -384,7 +388,8 @@ jobs:
     timeout-minutes: 150
     strategy:
       matrix:
-        variant: [ "clusterless,multi-site" ]
+        variant: [clusterless, multi-site]
```
lhanusov, line 401: @pruivo Hi Pedro, please can you help me to understand how these tests shoud look like?
When using quotes, it ignores the list and takes it as a single value. As you can see, the tests are failing for multi-site. I would like to understand if features must be --features=clusterless and --features=multi-site or --features=clusterless,multi-site, or this is a real failure. Thanks

pruivo: `I would like to understand if features must be --features=clusterless and --features=multi-site` <- this should be the way to go.

.github/workflows/ci.yml
```diff
@@ -384,7 +388,8 @@ jobs:
...
+        testsuite: [ClusterlessTestSuite, MultisiteTestSuite]
```
stianst: This will result in 4 runs (see https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#adding-a-matrix-strategy-to-your-workflow-job):

* clusterless + ClusterlessTestSuite
* clusterless + MultisiteTestSuite
* ..
* ...

I would probably just do something in the `run new base tests` step that converts `clusterless` and `multi-site` to the classes to run

stianst (changes requested):
Matrix job is not configured correctly; will result in 4 runs, not two

.github/workflows/ci.yml
lhanusov: yup, I've noticed it, just tried if that's possible, since there was a bug in running the old set of clusterless and multisite

stianst (dismissed):

.github/workflows/ci.yml
```diff
@@ -409,6 +413,11 @@ jobs:
           echo "Tests: $TESTS"
           ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Pinfinispan-server -Dauth.server.feature=${{ matrix.variant }} -Dauth.server.feature.disable=persistent-user-sessions -Dtest=$TESTS -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
 
+      - name: Run new base tests without cache
+        run: |
+          class = `${{ matrix.variant }^} | sed -e s/-//g`
```
stianst: Should probably be:
```suggestion
          class = `${{ matrix.variant }} | sed -e s/-//g`
```

tests/base/src/test/java/org/keycloak/tests/admin/client/SessionTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+        UserRepresentation testUserRep = user.admin().toRepresentation();
+        assertEquals(testUserRep.getId(), rep.getUserId());
+        assertEquals(testUserRep.getUsername(), rep.getUsername());
+
+        String clientId = account.toRepresentation().getId();
+        assertEquals("test-app", rep.getClients().get(clientId));
+        assertNotNull(rep.getIpAddress());
+        assertNotNull(rep.getLastAccess());
+        assertNotNull(rep.getStart());
```
pruivo: The primitive types are never null. Maybe remove these asserts or convert to `> 0` checks?

tests/base/src/test/java/org/keycloak/tests/suites/ClusterlessTestSuite.java
```diff
@@ -8,10 +8,10 @@
 import org.keycloak.testframework.injection.SuiteSupport;
 import org.keycloak.testframework.server.KeycloakServerConfig;
 import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
-import org.keycloak.tests.admin.ClientTest;
+import org.keycloak.tests.admin.client.SessionTest;
 
 @Suite
-@SelectClasses({ClientTest.class})
+@SelectClasses({SessionTest.class})
```
pruivo, line 14: Is the `ClientTest` class run?

test-framework/core/src/main/java/org/keycloak/testframework/realm/RealmConfigBuilder.java
```diff
@@ -226,6 +226,16 @@ public RealmConfigBuilder identityProvider(IdentityProviderRepresentation identi
         return this;
     }
 
+    public RealmConfigBuilder setRememberMe(boolean enabled) {
+        rep.setRememberMe(enabled);
+        return this;
+    }
+
+    public RealmConfigBuilder setPasswordPolicy(String passwordPolicy) {
```
pruivo: It doesn't seem to be used anywhere. Do you want to remove it and add it back when needed? Or just keep it?

test-framework/ui/src/main/java/org/keycloak/testframework/ui/page/LoginPage.java
```diff
@@ -17,6 +17,9 @@ public class LoginPage extends AbstractPage {
     @FindBy(css = "[type=submit]")
```
pruivo, line 16: unused import in this class: `import org.openqa.selenium.NoSuchElementException;`

tests/base/src/test/java/org/keycloak/tests/suites/ClusterlessTestSuite.java
lhanusov: Right now, it's not triggered anywhere and it should be replaced for particular tests. ClientTest was used only as a dummy.

pruivo (dismissed):
LGTM, thanks @lhanusov

tests/base/src/test/java/org/keycloak/tests/suites/ClusterlessTestSuite.java
pruivo: Oh, understood. Thank you.

stianst (dismissed):

lhanusov:
@pruivo there is a failure in the multi-site run: https://github.com/keycloak/keycloak/actions/runs/18189170494/job/51780431433?pr=42462

Do you think it's a bug in the test or infinispan/keycloak server? 

This failure is introduced, after I've fixed the github actions execution.

pruivo:
@lhanusov I'll take a look

pruivo:
@lhanusov, it is not your fault; it is a real bug. 

edit: https://github.com/keycloak/keycloak/issues/43161

pruivo (approved):

ahus1 (approved):
Approving based on previous approval from @stianst and @pruivo

PR #42465, opened 2025-09-09
Add User_agent header for documentation links checker

rmartinc:
Closes #42164

The wikipedia seems to require now the user-agent header and the apache http-client does not add any by default. Just adding a generic one using the java version and links to our project and the keycloak-dev mailing list. If you prefer something else just let me know. There is one error remaining to link https://www.keycloak.org/securing-apps/specifications but that one will be fixed when the new guides are published.


ahus1 (dismissed):

andymunro (approved):
Thanks, @rmartinc

ahus1 (approved):
Thank you @rmartinc - I added the one not-yet valid URL to the `ignored-links` file.

Issue #39658, opened 2025-05-12
OpenTelemetry Tracing: Visualize JGroups communication


ahus1:
### Description

Currently the JGroups communication is not visible in the OpenTelemetry Tracing.

Making it visible can visualize cross-node communications and delays.

### Motivation

Better observability and options to debug performance problems. 

### Details

There is a PoC available here: https://github.com/jgroups-extras/open-telemetry

mabartos:
@ahus1 IMO, a brief chapter describing the why and how to enable/disable should be introduced in the [Root cause analysis with tracing](https://www.keycloak.org/observability/tracing) guide on Keycloak website. 

Could you do it? Thanks

PR #42475, opened 2025-09-09
Concurrent update embedded caches and database

pruivo:
Closes #42374

- [x] Javadoc
- [x] Replace `SerializeExecutionsByKey` with `ActionSequencer` (non-blocking)
- [x] Consider dropping `SessionChangesPerformer` as it has 1 implementation only
- [x] Code cleanup


model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/OpenTelemetrySpan.java
```diff
@@ -11,14 +11,13 @@
...
     private final Span span;
-    private final Scope scope;
 
     public OpenTelemetrySpan(Span span) {
         this.span = Objects.requireNonNull(span);
         // TODO: This is actually wrong if you are doing asynchronous calls, but it allows the JGroups calls to be nested
         // This should be fixed in ISPN 16+ so that it is no longer needed
         // https://github.com/infinispan/infinispan/issues/15287
-        this.scope = span.makeCurrent();
+        //this.scope = span.makeCurrent();
```
ryanemerson, line 20: Uncomment again?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/CacheInfo.java
```diff
@@ -0,0 +1,33 @@
...
+import org.infinispan.Cache;
+import org.infinispan.util.concurrent.ActionSequencer;
+import org.keycloak.models.sessions.infinispan.SessionFunction;
+import org.keycloak.models.sessions.infinispan.entities.SessionEntity;
+
+/**
+ * Groups the {@link Cache}, the {@link ActionSequencer} (used by replace method) and the {@link SessionFunction} to
+ * compute the lifespan, and the max-idle for this session entity.
+ */
+public record CacheInfo<K, V extends SessionEntity>(Cache<K, SessionEntityWrapper<V>> cache,
```
ryanemerson: `CacheHolder`?

Naming is hard! :slightly_smiling_face:

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
```diff
@@ -210,6 +195,29 @@ protected <K> K generateKey(KeycloakSession session, Cache<K, ?> cache, KeyGener
...
+
+        if (MultiSiteUtils.isPersistentSessionsEnabled()) {
+            if (useCaches) {
+                try (var session = factory.create()) {
+                    sessionCacheInfo = InfinispanChangesUtils.createWithCache(session, USER_SESSION_CACHE_NAME, SessionTimeouts::getUserSessionLifespanMs, SessionTimeouts::getUserSessionMaxIdleMs);
+                    offlineSessionCacheInfo = InfinispanChangesUtils.createWithCache(session, OFFLINE_USER_SESSION_CACHE_NAME, SessionTimeouts::getOfflineSessionLifespanMs, SessionTimeouts::getOfflineSessionMaxIdleMs);
+                    clientSessionCacheInfo = InfinispanChangesUtils.createWithCache(session, CLIENT_SESSION_CACHE_NAME, SessionTimeouts::getClientSessionLifespanMs, SessionTimeouts::getClientSessionMaxIdleMs);
+                    offlineClientSessionCacheInfo = InfinispanChangesUtils.createWithCache(session, OFFLINE_CLIENT_SESSION_CACHE_NAME, SessionTimeouts::getOfflineClientSessionLifespanMs, SessionTimeouts::getOfflineClientSessionMaxIdleMs);
+                }
+            } else {
```
ryanemerson, line 207: AFAICT we don't have any test coverage for `--spi-user-sessions--infinispan--use-caches=false`, should we add this or even drop the option altogether?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
```diff
@@ -42,48 +43,114 @@
...
+     * The implementation should register the {@link CompletionStage} into the {@link AggregateCompletionStage}.
+     *
+     * @param stage  The {@link AggregateCompletionStage} to collect the {@link CompletionStage}.
+     * @param entry  The {@link Map.Entry} with the ID and the session.
+     * @param merged The {@link MergedUpdate} to be applied to the existing session.
+     * @throws NullPointerException if this instance does not support non-blocking writes.
+     * @see #isNonBlocking()
+     */
+    public void asyncWrite(AggregateCompletionStage<Void> stage, Map.Entry<K, SessionUpdatesList<V>> entry, MergedUpdate<V> merged) {
+        var update = newUpdate(entry, merged);
```
ryanemerson, line 105: Add an `assert isNonBlocking` here for testing?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/PersistentSessionsChangelogBasedTransaction.java
```diff
@@ -137,52 +102,13 @@ public SessionEntityWrapper<V> get(K key, boolean offline) {
...
-                @Override
-                public boolean shouldConsumeChange(V entity) {
-                    return entity.isOffline();
-                }
-            });
-        }
-
-        return changesPerformers;
+    private JpaChangesPerformer<K, V> prepareChangesPerformers() {
+        return new JpaChangesPerformer<>(cacheName, batchingQueue);
```
ryanemerson: Remove the method and just create inline as only used in one place?

model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/OpenTelemetrySpan.java
pruivo: I can't. `makeCurrent` sets the current span into a `ThreadLocal`, but since we are using `*Async()` methods in Infinispan, the span is never closed in the same thread, causing tracing to be all over the place. 
The only downside of this change is that the JGroups send/deliver traces are not chained/associated with the write command. 
This was fixed in ISPN 16.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
pruivo: I think you mean we don't have a test for `--spi-user-sessions--infinispan--use-caches=true`. `false` is the default value in the main configuration. 
I'll wait for @ahus1's opinion, but I'm more inclined to add the missing test.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/JpaChangesPerformer.java
pruivo: It is unnecessary. `batchingQueue` is null when `isNonBlocking` returns `false`. You'll get a `NullPointerException` in the test if the method is incorrectly used (I know since I made that mistake).

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
ryanemerson: > `false` is the default value in the main configuration.

Is it?

https://github.com/keycloak/keycloak/blob/935caa97ea138a94d7d6d4970cbb56d78bb1d639/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java#L80

pruivo: Oh, I read `useBatches` instead of `useCaches`. My bad! 
So, there is nothing here to do as the multi-site feature uses this code path (`persistent-user-session` is enabled and `useCaches` is false; this is forced!).

https://github.com/keycloak/keycloak/blob/214cbbc2aa1917e81e1c217f848c051e694f0d48/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java#L161

pruivo:
@ryanemerson updated!

model/infinispan/src/main/java/org/keycloak/infinispan/module/factory/OpenTelemetrySpan.java
ryanemerson: Thanks for explaining. I think we're good then as this isn't a regression, given that we introduced JGroups tracing in 26.4 #39658.

ahus1 (approved):
Thank you for this extensive change and all the reasoning you and everyone else did around this issue!

PR #42478, opened 2025-09-09
Admin UI hides local users when LDAP provider fails

martin-kanis:
Closes #42276


pedroigor (approved):
@martin-kanis Nice! Do you think we can somehow warn the Admin UI about the providers that are failing?

In addition to returning the users, we would return a warning message like "Some users could not be fetched due to failures in user storage providers. Check the logs for more details".

PR #42490, opened 2025-09-09
[RLM] Disable policy when the origin or selection criteria is removed

### LINKED ISSUE #42123: [RLM] Remove or update policy when the origin or selection criteria is removed
### Description

For policies associated with specific origins, such as an IDP, we need to ensure the policy is removed or updated when the associated IDP is removed. Likewise, if the policy defines selection criteria such as by group or role association, it must be removed or updated upon group or role removal.

vramik:
Closes #42123

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyConditionProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyConditionProvider.java
@@ -14,5 +14,5 @@ public interface ResourcePolicyConditionProvider extends Provider {
         return null;
     }
 
-    void validate();
+    void validate() throws ResourcePolicyInvalidStateException;
 }
--- a/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyConditionProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyConditionProvider.java
@@ -13,4 +13,6 @@ public interface ResourcePolicyConditionProvider extends Provider {
     default Predicate toPredicate(CriteriaBuilder cb, CriteriaQuery<String> query, Root<?> userRoot) {
         return null;
     }
+
+    void validate();
 }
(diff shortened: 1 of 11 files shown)


services/src/main/java/org/keycloak/models/policy/ResourcePolicyManager.java
```diff
@@ -203,29 +204,36 @@ public void processEvent(ResourcePolicyEvent event) {
...
-                            policyStateProvider.remove(policy.getId(), event.getResourceId());
-                        }
+                    } catch (ModelValidationException e) {
+                        policy.getConfig().putSingle("enabled", "false");
+                        policy.getConfig().putSingle("validation_error", e.getMessage());
+                        updatePolicy(policy, policy.getConfig());
+                        log.debugf("Policy %s was disabled due to: %s", policy.getId(), e.getMessage());
                     }
                 });
     }
```
sguilhen, line 238: I am wondering if we shouldn't do the same when running the actions - probably not needed right now because we don't have yet actions that rely on other resources (like add user to a group or role), but once we do we will probably need a similar validation mechanism or at least make the `run` method throw a validation exception if the action can't perform it's task because the referenced resource (group, role, etc) doesn't exist

sguilhen (approved):
Thanks @vramik ! LGTM - only added one comment but I think it might be more for a follow up once we have actions that reference other resources.

@pedroigor ^

pedroigor (dismissed):

server-spi-private/src/main/java/org/keycloak/models/policy/ResourcePolicyConditionProvider.java
```diff
@@ -13,4 +13,6 @@ public interface ResourcePolicyConditionProvider extends Provider {
     default Predicate toPredicate(CriteriaBuilder cb, CriteriaQuery<String> query, Root<?> userRoot) {
         return null;
     }
+
+    void validate();
```
pedroigor: I would add `throws` to the method signature to indicate that this method will be throwing an exception when validation fails.

services/src/main/java/org/keycloak/models/policy/ResourcePolicyManager.java
pedroigor: Not sure now ... But I see the point.

pedroigor (approved):
As discussed, as a follow-up we should:

* Consider that multiple errors can be stored in the `error` entry and we need to manage it accordingly
* Consider that we might want to store the value of the `error` entry as a JSON to make it easier to parse errors when returning representations from the API. As @vramik suggested, we might introduce a `error` field in the policy representation to make easier to check/parse errors.

PR #37640, opened 2025-02-25
Add supported config options for additional datasources

mabartos:
- Closes: #29116
- Closes: https://github.com/keycloak/keycloak/issues/39154

### Changes

- Automatically create sibling datasource OPTIONS for DB options 
- Automatically create sibling datasource PROPERTY MAPPERS for DB options
- New options category DATABASE_DATASOURCES -> 'Database - additional datasources'
- Default DB name for H2 dev-file is `keycloakdb-<datasource>` (f.e. `keycloakdb-persistent-sessions`)
- Providing option to enable/disable specific datasources -> `db-enabled-<datasource>`
- Additional changes for CLI for build-time wildcard options
- Solves https://github.com/keycloak/keycloak/issues/39154

### Follow-up issues:
- Amend ENV logic with wildcard mappers
- Documentation
- https://github.com/keycloak/keycloak/issues/39250

PR #42514, opened 2025-09-10
Quarkus config for additional datasources not picked up after 26.3.0

mabartos:
- Closes #42263

After merging https://github.com/keycloak/keycloak/pull/37640, users are not able to use the quotes (") in the named additional datasource as it's mapped to a certain PropertyMapper. We should be backward compatible and still allow users to use the `quarkus.datasource."user-store".xxx` properties in the `quarkus.properties` file.

Once we promote the additional datasource options to _supported_, we would be able to get rid of the `quarkus.datasource.*` options for good. 

@shawkins @vmuzikar Could you please check it? Thanks!

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
@@ -32,6 +32,7 @@ import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 
+import io.smallrye.config.DefaultValuesConfigSource;
 import org.keycloak.config.DeprecatedMetadata;
 import org.keycloak.config.Option;
 import org.keycloak.config.OptionCategory;
@@ -67,19 +68,19 @@ public class PropertyMapper<T> {
     private final BooleanSupplier required;
     private final String requiredWhen;
     private final String from;
+    private final boolean allowQuarkusPropertiesFallback;
 
     private final String namedProperty;
 
     PropertyMapper(PropertyMapper<T> mapper, String from, String to, String mapFrom, String namedProperty, ValueMapper parentMapper) {
         this(mapper.option, to, mapper.enabled, mapper.enabledWhen, mapper.mapper, mapFrom, parentMapper,
-                mapper.paramLabel, mapper.mask, mapper.validator, mapper.description, mapper.required,
-                mapper.requiredWhen, from, namedProperty);
+                mapper.paramLabel, mapper.mask, mapper.validator, mapper.description, mapper.required, mapper.requiredWhen, from, namedProperty, mapper.allowQuarkusPropertiesFallback);
     }
 
     PropertyMapper(Option<T> option, String to, BooleanSupplier enabled, String enabledWhen,
                    ValueMapper mapper, String mapFrom, ValueMapper parentMapper,
-                   String paramLabel, boolean mask, BiConsumer<PropertyMapper<T>, ConfigValue> validator,
-                   String description, BooleanSupplier required, String requiredWhen, String from, String namedProperty) {
+                   String paramLabel, boolean mask, BiConsumer<PropertyMapper<T>, ConfigValue> validator, String description,
+                   BooleanSupplier required, String requiredWhen, String from, String namedProperty, boolean allowQuarkusPropertiesFallback) {
         this.option = option;
         this.from = from == null ? NS_KEYCLOAK_PREFIX + this.option.getKey() : from;
         this.to = to == null ? getFrom() : to;
@@ -97,6 +98,7 @@ public class PropertyMapper<T> {
         this.description = description;
         this.parentMapper = parentMapper;
         this.namedProperty = namedProperty;
+        this.allowQuarkusPropertiesFallback = allowQuarkusPropertiesFallback;
     }
 
     /**
@@ -129,6 +131,11 @@ public class PropertyMapper<T> {
         // this ensures that mapFrom transformers, and regular transformers are applied exclusively - not chained
         ConfigValue config = convertValue(NestedPropertyMappingInterceptor.proceed(context, from));
 
+        // due to backward compatibility, we allow for some options to use Quarkus properties
+        if (allowQuarkusPropertiesFallback) {
+            config = getConfigValueFromQuarkusProperties(config, context);
+        }
+
         boolean parentValue = false;
         if (mapFrom != null && (config == null || config.getValue() == null)) {
             // if the property we want to map depends on another one, we use the value from the other property to call the mapper
@@ -153,6 +160,17 @@ public class PropertyMapper<T> {
 
         // now try any defaults from quarkus
         return context.proceed(name);
+    }
+
+    // due to backward compatibility, we allow for some options to use Quarkus properties
+    private ConfigValue getConfigValueFromQuarkusProperties(ConfigValue currentConfig, ConfigSourceInterceptorContext context) {
+        if ((currentConfig == null || currentConfig.getValue() == null) && !getFrom().equals(getTo())) {
+            var value = context.proceed(getTo());
+            if (isModifiedQuarkusProperty(value)) {
+                return value;
+            }
+        }
+        return currentConfig;
     }
 
     public Option<T> getOption() {
@@ -365,6 +383,7 @@ public class PropertyMapper<T> {
         private String requiredWhen = "";
         private BiFunction<String, Set<String>, Set<String>> wildcardKeysTransformer;
         private ValueMapper wildcardMapFrom;
+        private boolean allowQuarkusPropertiesFallback;
 
         public Builder(Option<T> option) {
             this.option = option;
@@ -523,17 +542,22 @@ public class PropertyMapper<T> {
             return this;
         }
 
+        public Builder<T> allowQuarkusPropertiesFallback(boolean allow) {
+            this.allowQuarkusPropertiesFallback = allow;
+            return this;
+        }
+
         public PropertyMapper<T> build() {
             if (paramLabel == null && Boolean.class.equals(option.getType())) {
                 paramLabel = Boolean.TRUE + "|" + Boolean.FALSE;
             }
             if (option.getKey().contains(WildcardPropertyMapper.WILDCARD_FROM_START)) {
-                return new WildcardPropertyMapper<>(option, to, isEnabled, enabledWhen, mapper, mapFrom, parentMapper, paramLabel, isMasked, validator, description, isRequired, requiredWhen, wildcardKeysTransformer, wildcardMapFrom);
+                return new WildcardPropertyMapper<>(option, to, isEnabled, enabledWhen, mapper, mapFrom, parentMapper, paramLabel, isMasked, validator, description, isRequired, requiredWhen, wildcardKeysTransformer, wildcardMapFrom, allowQuarkusPropertiesFallback);
             }
             if (wildcardKeysTransformer != null || wildcardMapFrom != null) {
                 throw new AssertionError("Wildcard operations not expected with non-wildcard mapper");
             }
-            return new PropertyMapper<>(option, to, isEnabled, enabledWhen, mapper, mapFrom, parentMapper, paramLabel, isMasked, validator, description, isRequired, requiredWhen, null, null);
+            return new PropertyMapper<>(option, to, isEnabled, enabledWhen, mapper, mapFrom, parentMapper, paramLabel, isMasked, validator, description, isRequired, requiredWhen, null, null, allowQuarkusPropertiesFallback);
         }
     }
 
@@ -591,6 +615,15 @@ public class PropertyMapper<T> {
 
     public static boolean isEnvOption(ConfigValue configValue) {
         return Optional.ofNullable(configValue.getConfigSourceName()).filter(name -> name.contains(KcEnvConfigSource.NAME)).isPresent();
+    }
+
+    // return true if the configValue is valid quarkus property modified by user -> no defaults from Quarkus
+    private static boolean isModifiedQuarkusProperty(ConfigValue configValue) {
+        return Optional.ofNullable(configValue)
+                .filter(value -> value.getValue() != null)
+                .map(ConfigValue::getConfigSourceName)
+                .filter(name -> !name.equals(DefaultValuesConfigSource.NAME))
+                .isPresent();
     }
 
     void validateExpectedValues(ConfigValue configValue, String v) {
--- a/quarkus/runtime/src/test/resources/conf/quarkus.properties
+++ b/quarkus/runtime/src/test/resources/conf/quarkus.properties
@@ -14,4 +14,14 @@ quarkus.datasource.bar = foo-${kc.prop3:${kc.prop4:${kc.prop5:def}-suffix}}
 quarkus.datasource.dog-store.db-kind=mariadb
 quarkus.datasource.cat-store.db-kind=postgresql
 
+quarkus.datasource."userstore".db-kind=postgresql
+quarkus.datasource."userstore".username=user-quoted
+quarkus.datasource."userstore".jdbc.url=jdbc:postgresql://localhost:5433/db
+quarkus.datasource."userstore".jdbc.transactions=disabled
+
+quarkus.datasource.userstore.db-kind=mariadb
+quarkus.datasource.userstore.username=user-unquoted
+quarkus.datasource.userstore.jdbc.url=jdbc:mariadb://my-url.com:3333/userstore
+quarkus.datasource.userstore.jdbc.transactions=xa
+
 not.quarkus=value
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
@@ -198,7 +198,8 @@ final class DatabasePropertyMappers implements PropertyMapperGrouping {
 
                 var created = fromOption(datasourceOption.get())
                         .isMasked(parent.isMask())
-                        .transformer(parent.getMapper());
+                        .transformer(parent.getMapper())
+                        .allowQuarkusPropertiesFallback(true); // we allow to use properties in quarkus.properties due to backwards compatibility for datasources
 
                 if (parent.getMapFrom() != null) {
                     var wildcardMapFromOption = getKeyForDatasource(parent.getMapFrom())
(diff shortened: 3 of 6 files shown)


shawkins:
Are these changes needed on main? There's been other refinements there and it looks like we'll already supply quarkus with the right values.

vmuzikar (dismissed):
LGTM, thanks.

mabartos:
> Are these changes needed on main? There's been other refinements there and it looks like we'll already supply quarkus with the right values.

@shawkins Unfortunately, yes. As we map the `quarkus.datasource."<datasource>".*` properties to the DB/Datasource property mappers, we never check the `quarkus.properties` when there's a default value for the property. Same as in the JavaDoc we provided:
image

It means when somebody uses the `quarkus.datasource."<datasource>".*` (with those quotes) in their `quarkus.properties`, it's ignored as the default value is returned (if present).

This PR introduces an approach that, for specified properties (only datasources now), we look up to the `quarkus.properties` file to get the value. If we don't find it, we continue with other fallbacks like mapFrom, default value, or quarkus. 

@shawkins Or do you have any better approach in mind?

----------------------
@vmuzikar Thanks for the review. I've just added a stricter check, that we check only the `quarkus.properties` present in the `{dist}/conf/quarkus.properties` file and nothing more. 

@shawkins @vmuzikar Should be ready for the review. Thanks!

shawkins:
> It means when somebody uses the quarkus.datasource."<datasource>".* (with those quotes) in their quarkus.properties, it's ignored as the default value is returned (if present).

I was looking just ad the changes around db-kind, so yes there is still an issue with the other db options with defaults. Let me review again.

shawkins:
A couple of thoughts:

- There are two approaches for this logic. Here you have the quarkus property with high precedence than our defaults. The other option is to only perform these wildcard mappings when we are managing the datasource - which should be when kc.db-kind-* is set - at which point we fully take over and our defaults apply. The downside of the first approach is that it's inconsistent with our other default handling logic and will allow for greater mixed usage of the keycloak and quarkus properties. The downside of the second approah is that it requires users to switch to the kc options - at least for things that have defaults.

- isValidFromQuarkusProperties is probably not needed. We don't explicitly prohibit users from using any other quarkus config sources, in particular env properties.

vmuzikar:
After reading the recent comments, I'd like to clarify if I understand correctly. We do not reflect Quarkus (wildcard) options if we have a default value for the, in particular `quarkus.datasource."<datasource>".*`. Correct? But it's not specific to the wildcard options – all options work like that. IMHO the priorities should be like this:
1. User-specified value in a Keycloak Config Source.
2. User-specified value in `quarkus.properties`.
3. Keycloak default value.
4. Quarkus default value.

I.e. if there's an explicit value of some mapped option in `quarkus.properties`, it should have priority over our default value. That should apply in general for all options.

WDYT?

mabartos:
> The downside of the first approach is that it's inconsistent with our other default handling logic and will allow for greater mixed usage of the keycloak and quarkus properties. 

@shawkins 1. - Specifying additional datasources via the Quarkus properties is a special case as we kinda "support"(remember the gray area) it, mainly for the User Storage SPI. As we have the switch `allowQuarkusPropertiesFallback()` that says - if it's enabled, you can set the property via quarkus.properties even when you should use the KC properties. Once we remove the support for specifying datasources via quarkus.properties, we disable the switch and will not be possible to specify it in quarkus.properties (KC 27). So I think some inconsistency in this gray area is ok, to allow it only for the datasources. 

> The downside of the second approach is that it requires users to switch to the kc options - at least for things that have defaults.

@shawkins 2. - Yes, and that's what I wanted to avoid for 26.4/26.3 release as we have the KC options as preview and still the "supported" way on how to define datasources is via `quarkus.properties`. IMHO, it's applicable only to the KC 27 release.

> isValidFromQuarkusProperties is probably not needed. We don't explicitly prohibit users from using any other quarkus config sources, in particular env properties.

I'd rather check if it's coming from the "Keycloak" `quarkus.properties` config source as that's the only approach we say KC users can use (we don't say anything that they can use Quarkus ENV vars). It help us to be on a safe side and prevent any other issues. Moreover, we want to prefer KC defaults, when no value is provided (via KC conf sources, or quarkus.properties), instead of Quarkus defaults.

@shawkins WDYT?

--------------------------------------------------------

> We do not reflect Quarkus (wildcard) options if we have a default value for the, in particular quarkus.datasource."<datasource>".*. Correct? But it's not specific to the wildcard options – all options work like that.

@vmuzikar Sorry, I'm not sure what you mean. Can you elaborate a little? To clarify it a little bit - when some Quarkus properties is mapped by our KC properties (like `.to("quarkus.whatever")`, it is not possible to override the property in the `quarkus.properties` file - it falls back there only when no KC default value is set. And that is IMHO correct behavior how it should work. 

> I.e. if there's an explicit value of some mapped option in quarkus.properties, it should have priority over our default value. That should apply in general for all options.

I don't think we should do that. The `quarkus.properties` config source is not **supported** Keycloak config source, and its main purpose is to tackle some more advanced Quarkus logic **not covered by Keycloak** (see the doc below).

image
image

It means that we should not consider the properties in the `quarkus.properties`. After merging this PR, it's only doable for the special case of additional datasources that we "support". But once we remove the support, it will not be possible to override the datasource config either.

@vmuzikar WDYT?

---------------------------------------

### Summary

- Additional datasources configuration is a special case
- We support having them set up via `quarkus.properties`
- Without this PR, users are blocked to use the `quarkus.datasource."<datasource>".*` properties (with quoting) -> breaking change
- Who use it without quoting, they should be fine

### Suggested approach
- Users are able to specify only the datasource options via `quarkus.properties` in Keycloak 26.3.x, and 26.4.x (this PR covers)
- For Keycloak 27, we promote Keycloak datasource options to _supported_
- For Keycloak 27, users need to use the Keycloak datasource options
- For Keycloak 27, disable/remove the flag for the Quarkus properties fallback (not possible to use the quarkus.properties for datasources with quoting)

vmuzikar:
> it is not possible to override the property in the `quarkus.properties` file - it falls back there only when no KC default value is set

Yes, that's what I meant. :) Sorry for not being clear.

> The `quarkus.properties` config source is not **supported** Keycloak config source

That's a good point. I didn't realize we even document that `quarkus.properties` is ignored when a mapped KC option exists for that. Then +1 to what you proposed – to increase the `quarkus.properties` "priority" only for selected cases, like the datasources here due to the backward compatibility (as it was kinda supported before).

> * For Keycloak 27, we promote Keycloak datasource options to _supported_
> * For Keycloak 27, users need to use the Keycloak datasource options

This should be done in KC 26.5 = claiming support for the new options + deprecating the support for `quarkus.datasource."<datasource>".*`.

mabartos:
> This should be done in KC 26.5 = claiming support for the new options + deprecating the support for quarkus.datasource."<datasource>".*.

Yes, we can definitely do that.

shawkins:
> Specifying additional datasources via the Quarkus properties is a special case as we kinda "support"(remember the gray area) it

Yes it is understood to be a gray area. Let me elaborate on my thoughts:

- the first goal is that nothing should change if you are solely using quarkus properties. While we don't mention it in the documentation, our help discussions are littered with places where people are using environment variables for quarkus options. Some of the discussion of the removal of profile prefixed properties even led to pointing towards quarkus env files. From my perspective isValidFromQuarkusProperties violates this continuity - more below.

- the next goal would be to as consistent as possible with how we are already doing things. The thought here is that the explicit usage of kc.db-kind-* signals that the user wants to cutover to the Keycloak way of doing things - and just like a user coming afresh will have the max-size, min-size, active, etc. options default to the Keycloak values and not to quarkus option values. We of course would have to document this distinction. 

> I don't think we should do that. The quarkus.properties config source is not supported Keycloak config source, and its main purpose is to tackle some more advanced Quarkus logic not covered by Keycloak (see the doc below).

We have had users complain in the past about this behavior - as our quarkus property values may be derived or coming from internal (no description) options, users have occasionally found it difficult to force a particular quarkus value. I'm sympathetic to the idea that you should be able to do this, even if it's documented as not supported. So that our defaults would still have precedence over quarkus defaults and don't cut off usage of environment variables, something like the isValidFromQuarkusProperties could be modified to make a check like Configuration.isUserModifiable. If we did allow this generally, then the consistency concern above goes away.

mabartos:
> the first goal is that nothing should change if you are solely using quarkus properties. While we don't mention it in the documentation, our help discussions are littered with places where people are using environment variables for quarkus options. Some of the discussion of the removal of profile prefixed properties even led to pointing towards quarkus evn files. From my perspective isValidFromQuarkusProperties violates this continuity - more below.

@shawkins Ok, if there are some use cases, accepting ENV vars for mapped to properties should be ok and should not introduce any security concern. 

> the next goal would be to as consistent as possible with how we are already doing things. The thought here is that the explicit usage of kc.db-kind-* signals that the user wants to cutover to the Keycloak way of doing things - and just like a user coming afresh will have the max-size, min-size, active, etc. options default to the Keycloak values and not to quarkus option values. We of course would have to document this distinction.

It means that when I'm using the `quarkus.properties` for the datasource settings(as for older releases) and we generally support using the `quarkus.properties` as the supported config source, I'm relying on that settings. But suddenly, when I use a certain KC option, it will cut over the `quarkus.properties` and set different values for the related options? Is that what you're proposing? Or in what cases do you consider to have that cutoff? This kind of distinction seems to me like a bigger mixup and kinda inconsistent behavior. Or can you elaborate on it little bit more?

----------------------
I'll respond for the next later.

But, this PR is just a simple approach on how to fix this bug. It seems we're talking about more wide usage of the quarkus.properties that would need to be properly planned based on releases.

shawkins:
> But suddenly, when I use a certain KC option, it will cut over the quarkus.properties and set different values for the related options? Is that what you're proposing? 

Correct - that once you start using kc.db-kind-* this indicates the datasource is fully "configured" or "managed" by Keycloak.

> This kind of distinction seems to me like a bigger mixup and kinda inconsistent behavior.

It is more consistent with our existing handling for which the docs explicitly state that quarkus values will be ignored if there is a first-class option. 

You can make the case that users who are coming afresh to 26.4 don't need and/or shouldn't have the ability to selectively pick up certain quarkus database options over the keycloak defaults.

> But, this PR is just a simple approach on how to fix this bug. It seems we're talking about more wide usage of the quarkus.properties that would need to be properly planned based on releases.

I'm not sure of that. Any direct usage of quarkus properties - whether from the properties file or other sources - would stay unsupported. All we're doing is talking about expanding the scope of when those unsupported values are picked up so that there isn't an inconsistency between the database options and everything else. Mostly we'd just have to remove or refine the doc about the first-class option being forcing.

For reference the full language is "Note that some Quarkus properties are already mapped in the Keycloak configuration, such as quarkus.http.port and similar essential properties. If the property is used by Keycloak, defining that property key in quarkus.properties has no effect. The Keycloak configuration value takes precedence over the Quarkus property value." - which may not be entirely clear that "already mapped" also means having a default.

vmuzikar:
Just wanted to give my two cents. I think we might be a bit overthinking this. :) We got into this odd state with the additional DB options because we were technically supporting configuring additional DB sources via `quarkus.properties`. This is not the case (nor ever be) for anything else, it was a mistake. Now, here in this PR we're trying to find a supported migration path for the users from `quarkus.properties` to the new first class citizen options – we should not be doing anything beyond this scome. This path is not consistent with our current approach around `quarkus.properties` but it's a one-off, temporary solution. We can afford to take some shortcuts, IMHO.

mabartos:
> Correct - that once you start using kc.db-kind-* this indicates the datasource is fully "configured" or "managed" by Keycloak. It is more consistent with our existing handling for which the docs explicitly state that quarkus values will be ignored if there is a first-class option.

@shawkins I don't know if we should introduce some kind of another "discrepancy" to our config source settings. I'd rather go a way "left or right" and nothing between for the future. It means that users cannot change the mapped Quarkus properties(current), or the second variant would be that they would have the ability to change whatever Quarkus property they want (at their own risk) without other conditions like if `db-kind` is set or not. 

> I'm not sure of that. Any direct usage of quarkus properties - whether from the properties file or other sources - would stay unsupported. All we're doing is talking about expanding the scope of when those unsupported values are picked up so that there isn't an inconsistency between the database options and everything else. Mostly we'd just have to remove or refine the doc about the first-class option being forcing.

@shawkins Considering the quarkus handling would still be unsupported, and users could fully override it at their own risk, it should be ok. As you mentioned, users sometimes need to override the mapped Quarkus properties to fulfill their needs, and it should not be so invasive to do it. Did you consider having it only for Quarkus properties that we map to? Or even some other ones, like smallrye, etc.? Because in that case, we'd go over the whole SmallRye config source chain before evaluating mapFrom() and KC default values. 

> I think we might be a bit overthinking this. :) We got into this odd state with the additional DB options because we were technically supporting configuring additional DB sources via quarkus.properties. This is not the case (nor ever be) for anything else, it was a mistake. Now, here in this PR we're trying to find a supported migration path for the users from quarkus.properties to the new first class citizen options – we should not be doing anything beyond this scome. This path is not consistent with our current approach around quarkus.properties but it's a one-off, temporary solution. We can afford to take some shortcuts, IMHO.

@vmuzikar @shawkins Thanks both for your valuable feedback. As mentioned, this also touches the general handling of the Quarkus properties and not strictly related to the datasources, so the conversation contains valuable information. I agree that it's a temporary solution that solves the problem. 

However, we should go with this approach at least for the 26.3.x backport: https://github.com/keycloak/keycloak/pull/42529

EDITS in the PR:
- Transaction setting is also covered
- we don't check if it's coming strictly from the `quarkus.properties` file, just omit the Quarkus defaults

quarkus/runtime/src/test/resources/conf/quarkus.properties
```diff
@@ -14,4 +14,14 @@ quarkus.datasource.bar = foo-${kc.prop3:${kc.prop4:${kc.prop5:def}-suffix}}
 quarkus.datasource.dog-store.db-kind=mariadb
 quarkus.datasource.cat-store.db-kind=postgresql
 
+quarkus.datasource."userstore".db-kind=postgresql
+quarkus.datasource."userstore".username=user-quoted
+quarkus.datasource."userstore".jdbc.url=jdbc:postgresql://localhost:5433/db
+quarkus.datasource."userstore".jdbc.transaction=xa
```
shawkins: transaction in here and in the unit tests should be plural.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -593,6 +617,15 @@ public static boolean isEnvOption(ConfigValue configValue) {
         return Optional.ofNullable(configValue.getConfigSourceName()).filter(name -> name.contains(KcEnvConfigSource.NAME)).isPresent();
     }
 
+    // return true if the configValue is valid quarkus property modified by user -> no defaults from Quarkus
+    private static boolean isModifiedQuarkusProperty(ConfigValue configValue) {
+        return Optional.ofNullable(configValue)
+                .filter(value -> value.getValue() != null)
+                .map(ConfigValue::getConfigSourceName)
+                .filter(name -> !name.equals(DefaultValuesConfigSource.NAME))
```
shawkins, line 625: I'm not sure this will work for build time options. Isn't the default values config source created at build time?

shawkins (changes requested):
Just want to highlight one more consideration. While these changes put the quarkus properties ahead of our defaults for selected options, if the user is effectively relying on quarkus defaults they will still see a behavioral change. If for example the user does not set the quarkus max-size or min-size and is using quoted properties, the value will now map from the main datasource settings.

> Did you consider having it only for Quarkus properties that we map to? Or even some other ones, like smallrye, etc.? Because in that case, we'd go over the whole SmallRye config source chain before evaluating mapFrom() and KC default values.

If the mapTo is not a kc. property, then yes you would have to check for a overriden value ahead of any mapFrom or default - it would be done in exactly the same place in the code that you are doing the fallback check.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
mabartos: > I'm not sure this will work for build time options. Isn't the default values config source created at build time?

@shawkins Can you elaborate a little bit on this? You mean for Quarkus build time options that have a default value? AFAIK, we don't really get the default values directly from the Quarkus. We just don't set it and Quarkus handles it on its own. 

The db-kind is build time option, and I'm able to get its value during the build phase. See:

Screenshot From 2025-09-16 13-16-35

Note the `kc.db-kind-userstore` in the `show-config` output.

mabartos:
@shawkins @vmuzikar Updated. However, after the conversation here, we might consider to have it included only for the 26.3.x branch, and we can try to use the fallback to the `quarkus.properties` by default for the Keycloak 26.4. WDYT?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
shawkins: > Can you elaborate a little bit on this? You mean for Quarkus build time options that have a default value? AFAIK, we don't really get the default values directly from the Quarkus. We just don't set it and Quarkus handles it on its own.

Yes, I do mean that this will not filter quarkus build time options that have a default value. However you are correct that when we initially consult this - we're not actually consulting the effective quarkus defaults, so that doesn't matter.

This means that during our build phase this is effectively the same issue as the max-size / min-size - if you are using a quoted datasource, our Keycloak defaults will take precedence over the previous quarkus defaults. If you add to your test "quarkus.datasource.\"userstore\".jdbc.max-size", you'll see that it resolves to 100, not 10. And if you remove "quarkus.datasource."userstore".jdbc.transactions=disabled" from the quarkus.properties - the resolved value will be xa, not the previous default of enabled.

At the very least there would need to be additional upgrade notes to let people know about the change over in additional datasource default handling.

mabartos:
As we discussed offline, the approach of changing the priorities of config sources brings some problems. We haven't stated anywhere that the quoted datasource name can be used, as even our quickstarts (the single source where the datasource options were used) do not contain the quoted datasource name. 

@shawkins As you mentioned in the meeting, as somebody may require having the datasource name quoted - I'm not sure if there's a case, because AFAIK, you're not able to use the additional quotation for the `persistence.xml` file (or it does not make much sense to me). 

As said, even though it's some kind of a breaking change, I would not waste so much time on this use case as people can use the KC properties or just simply remove the quotation for the datasource in the unsupported `quarkus.properties`. It means that I'd vote to provide a migration guide on how people need to proceed. 

@vmuzikar @shawkins @Pepo48 WDYT?

vmuzikar:
As agreed in an offline discussion, let's handle this just as a note in the migration guide. We should do the following there:
- Clearly state that the `quarkus.properties` approach for additional datasources is not supported.
- Document migration steps from `quarkus.properties` to KC options.
- Emphasise, that the quoted use case especially is not working any more.

vmuzikar (changes requested):
See my previous comment.

shawkins:
> @shawkins As you mentioned in the meeting, as somebody may require having the datasource name quoted - I'm not sure if there's a case, because AFAIK, you're not able to use the additional quotation for the persistence.xml file (or it does not make much sense to me).

The quotation isn't necessary in the persistence.xml. It's required in some instances with quarkus properties. Definitely if the name contains a '.' (quarkus.datasource."some.name"...) - I'm not sure if it's required other characters such as '-'.

mabartos:
The fix for the issue will be handled as part of the work on providing docs for the additional datasources. 

See https://github.com/keycloak/keycloak/pull/42655

PR #42517, opened 2025-09-10
Impossible to update custom user attribute when user imported from LDAP

### LINKED ISSUE #42234: Impossible to update custom user attribute when user imported from LDAP
### Before reporting an issue

### Area

admin/api

### Describe the bug

Keycloak version 25.0.1.
 Issue present only when using REST API, In UI all works perfect.

To reproduce:
- user from federated provider, LDAP
- user does have custom attribute (not imported from LDAP, specific for current realm/client), relevant for my business app (to be used later as JWT access token claim)

## Problem description

When updating custom attribute, `createTimestamp` and `modifyTimestamp` attributes MUST be specified in update request, in addition to custom attribute I want to update. 

 If those attributes are not specified, response is 400
```json
{"errors":[{"field":"modifyTimestamp","errorMessage":"updateReadOnlyAttributesRejectedMessage","params":["modifyTimestamp"]},{"field":"createTimestamp","errorMessage":"updateReadOnlyAttributesRejectedMessage","params":["createTimestamp"]}]}
```
If those attributes are specified, response is 400
```json
{"field":"modifyTimestamp","errorMessage":"updateReadOnlyAttributesRejectedMessage","params":["modifyTimestamp"]}
```
Error message about updated `modifyTimestamp` returned in all cases:
- when `modifyTimestamp` is omitted 
- when `modifyTimestamp` is set to `now()`
- when `modifyTimestamp` is unchanged, i.e. passed through from input

Issue exists only when updating custom attributes for user from LDAP federation provider. Updating user attributes for direct realm users - all fine. Updating custom attribute for user from LDAP federation provider via UI - also fine.

### Version

25.0.1

### Regression

### Expected behavior

User attribute is created/updated, same as it's happing via UI or for direct realm user.

### Actual behavior

API call return 400 error:
```json
{"field":"modifyTimestamp","errorMessage":"updateReadOnlyAttributesRejectedMessage","params":["modifyTimestamp"]}
```

### How to Reproduce?

1. For some realm, setup user federation from LDAP source
2. HTTP PUT to https://<keycloak-server>/admin/realms/<realm>/users/<userId>, payload like
```json
{"attributes":{"modifyTimestamp":["20250829131250Z"],"createTimestamp":["20241219090612Z"],"TestAttribute":["test 1","test 2"]},"username":"<masked>"}
3. Watch the error


### Anything else?

Just to be clear: issue exists only for users from LDAP federated source, and only when using admin REST API. All other variants, i.e. doing same operation via admin console or for local realm user - works like a charm.

martin-kanis:
Closes #42234

This PR relaxes `ReadOnlyAttributeUnchangedValidator` validation by allowing admins (UserProfileContext.USER_API) to omit some read-only attributes (`"CREATED_TIMESTAMP", "createTimestamp", "modifyTimestamp"`) from the requests. It won't be interpreted as read-only attribute removal.

### DIFF
--- a/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
+++ b/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
@@ -31,6 +31,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.keycloak.Config;
 import org.keycloak.Config.Scope;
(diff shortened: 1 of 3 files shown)


pedroigor (changes requested):
We cannot change the behavior of these admin/internal attributes without considering how to handle regular user attributes.

Keycloak does not support `PATCH` operations, and that is something being addressed by the @keycloak/cloud-native in Admin V2.

Even though it is fixing a bad UX, this change set is improving UX only for a subset of the user attributes and not really fixing the core issue.

Failures are also related to the assumption/limitation that all attributes must be sent. If not sent, they will be marked for removal, if sent but with a different value, the validation will fail.

PR #42518, opened 2025-09-10
ClientSession timestamp not updated in the database

### LINKED ISSUE #42012: Some client sessions are deleted after Keycloak deploy
### Before reporting an issue

### Area

core

### Describe the bug

Hi everyone, I'm reaching out seeking help to solve a weird behavior that we are observing with our Keycloak instances.

On Keycloak 26.2.5 or 26.3.1, every time we deploy a new version of our Keycloak with custom providers, some user sessions are losing their client sessions, even if they are not expired yet. See the following screenshot before a deployment:

Image

Then, after the Keycloak pods were restarted, one session lost its client:

Image

This is in one of our lower environments, in production, the problem is a lot worse.
Then, we start to see a lot of `REFRESH_TOKEN_ERROR` in our logs:
```
type="REFRESH_TOKEN_ERROR", realmId="1234", realmName="redacted", clientId="react-client", userId="null", sessionId="4321", ipAddress="redacted", error="invalid_token", reason="Session doesn't have required client", grant_type="refresh_token", refresh_token_type="Refresh",
```
This is our session lifetime configs:

Image

We currently have two Keycloak pods running on Kubernetes

### Version

26.3.1

### Regression

### Expected behavior

The user sessions should not lose their clients after each Keycloak restart, or if this is expected, a clear documentation on why it's needed and how to tweak it.

### Actual behavior

Refresh of access tokens fails because user sessions are losing their client sessions during Keycloak restart. This started to happen after we upgraded from Keycloak 26.1.4 to 26.2.5.

### How to Reproduce?

Unfortunately, I couldn't find a reliable way to reproduce this other than creating the OIDC sessions and restarting the application manually.

### Anything else?

_No response_

pruivo:
Closes #42012

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/ClientSessionPersistentChangelogBasedTransaction.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/ClientSessionPersistentChangelogBasedTransaction.java
@@ -85,6 +85,7 @@ public class ClientSessionPersistentChangelogBasedTransaction extends Persistent
 
             // Cache does not contain the offline flag value so adding it
             wrappedEntity.getEntity().setOffline(offline);
+            wrappedEntity.getEntity().setUserSessionId(userSession.getId());
 
             RealmModel realmFromSession = kcSession.realms().getRealm(wrappedEntity.getEntity().getRealmId());
             if (!realmFromSession.getId().equals(realm.getId())) {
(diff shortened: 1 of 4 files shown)


pruivo:
@ahus1 let me know what you think.

ahus1 (dismissed):
Thank you for the quick turnaround!

pruivo:
@ahus1 I've just pushed another commit to check for nulls. Please review.

ahus1 (dismissed):

ahus1 (approved):

PR #42519, opened 2025-09-10
Remove default cache configurations from cache-local.xml

ryanemerson:
Closes #42351


model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
```diff
@@ -110,13 +110,13 @@ public static void configureLocalCaches(Config.Scope keycloakConfig, Configurati
      *
      * @param holder The {@link ConfigurationBuilderHolder} where the caches will be defined.
      */
-    public static void applyDefaultConfiguration(ConfigurationBuilderHolder holder, boolean warnMutate) {
+    public static void applyDefaultConfiguration(ConfigurationBuilderHolder holder, CachingOptions.Mechanism mechanism, boolean warnMutate) {
```
pruivo: You can have a non-clustered cache manager with `CachingOptions.Mechanism.ispn`. I would probably check the `holder` to see if the `transport()` exists or not.

pruivo (approved):
LGTM!

ahus1 (approved):

PR #42520, opened 2025-09-10
Rename RLM to Workflows

vramik:
Closes #42512


model/jpa/src/main/resources/META-INF/jpa-changelog-26.4.0.xml
```diff
@@ -28,35 +28,35 @@
         </createIndex>
     </changeSet>
 
-    <changeSet id="40343-RLM-resource-policy-state-table" author="keycloak">
-        <createTable tableName="RESOURCE_POLICY_STATE">
+    <changeSet id="40343-RLM-workflow-state-table" author="keycloak">
+        <createTable tableName="WORKFLOW_STATE">
```
pedroigor, line 32: Wdyt `RESOURCE_WORKFLOW_ACTION_STATE`?

mabartos: Having only `Workflows` is IMHO not very descriptive, as there might be confusion like: "What workflows? Workflows of what? What it touches?". 

As @pedroigor mentioned, wouldn't be better to name it more like **Resource Workflows**? I don't have a wide context here, but at least it'd be more clear to me (and I believe for more folks from KC community) what is meant.

Or do you plan to have Workflows as a general concept that may be applied to some other areas?

@vramik @pedroigor Just a suggestion.

pedroigor: The feature name will be Keycloak Workflows. You're right in the sense that it is about creating workflows for realm resources. However, I personally prefer `workflows` as a feature name, and we had quite a few discussions until we reached an agreement on this name.

The suggestion I'm making here is that the table is actually storing state for actions eventually scheduled to resources. However, I'm not pushing for it if others disagree.

vramik: > Wdyt `RESOURCE_WORKFLOW_ACTION_STATE`?

@pedroigor To keep the naming consistent with the new feature name ("Keycloak Workflows"), I'd like it to start with `WORKFLOW_` prefix if possible, what about `WORKFLOW_ACTION_STATE`? Would that work? 

@mabartos Thanks a lot for the feedback. You're right that the term "workflow" on its own is not very descriptive. As Pedro mentioned, the team had some discussions and settled on "Keycloak Workflows" as the feature name, and we're trying to align with that decision here.

pedroigor: > @pedroigor To keep the naming consistent with the new feature name ("Keycloak Workflows"), I'd like it to start with `WORKFLOW_` prefix if possible, what about `WORKFLOW_ACTION_STATE`? Would that work?
> 
Works for me.

model/jpa/src/main/java/org/keycloak/authorization/jpa/entities/PolicyEntity.java
```diff
@@ -63,7 +63,7 @@
...
                 @NamedQuery(name="findPolicyIdByType", query="select p.id from PolicyEntity p where p.resourceServer.id = :serverId and p.type = :type"),
                 @NamedQuery(name="findPolicyIdByResourceType", query="select p from PolicyEntity p inner join p.config c inner join fetch p.associatedPolicies a where p.resourceServer.id = :serverId and KEY(c) = 'defaultResourceType' and c like :type"),
-                @NamedQuery(name="findPolicyIdByDependentPolices", query="select p.id from PolicyEntity p inner join p.associatedPolicies ap where p.resourceServer.id = :serverId and (ap.resourceServer.id = :serverId and ap.id = :policyId)"),
+                @NamedQuery(name="findPolicyIdByDependentPolices", query="select p.id from PolicyEntity p inner join p.associatedPolicies ap where p.resourceServer.id = :serverId and (ap.resourceServer.id = :serverId and ap.id = :workflowId)"),
```
sguilhen: this change doesn't seem intentional

model/jpa/src/main/java/org/keycloak/authorization/jpa/store/JPAPolicyStore.java
```diff
@@ -315,7 +315,7 @@ public List<Policy> findDependentPolicies(ResourceServer resourceServer, String
 
         query.setFlushMode(FlushModeType.COMMIT);
         query.setParameter("serverId", resourceServer.getId());
-        query.setParameter("policyId", policyId);
+        query.setParameter("workflowId", policyId);
```
sguilhen: same thing here

model/jpa/src/main/java/org/keycloak/authorization/jpa/entities/PolicyEntity.java
vramik: no, it is not :), thanks for spotting it.

model/jpa/src/main/java/org/keycloak/models/workflow/EventBasedWorkflowProviderFactory.java
```diff
@@ -8,13 +8,13 @@
 import org.keycloak.models.KeycloakSessionFactory;
 import org.keycloak.provider.ProviderConfigProperty;
 
-public class EventBasedResourcePolicyProviderFactory implements ResourcePolicyProviderFactory {
+public class EventBasedWorkflowProviderFactory implements WorkflowProviderFactory {
 
     public static final String ID = "event-based-resource-policy";
```
sguilhen: we probably want to change the id to workflow too

model/jpa/src/main/java/org/keycloak/models/workflow/UserCreationTimeWorkflowProviderFactory.java
```diff
@@ -25,13 +25,13 @@
 import org.keycloak.models.KeycloakSessionFactory;
 import org.keycloak.provider.ProviderConfigProperty;
 
-public class UserCreationTimeResourcePolicyProviderFactory implements ResourcePolicyProviderFactory<UserCreationTimeResourcePolicyProvider> {
+public class UserCreationTimeWorkflowProviderFactory implements WorkflowProviderFactory<UserCreationTimeWorkflowProvider> {
 
     public static final String ID = "user-creation-time-resource-policy";
```
sguilhen: same here

model/jpa/src/main/java/org/keycloak/models/workflow/UserSessionRefreshTimeWorkflowProviderFactory.java
```diff
@@ -25,13 +25,13 @@
 import org.keycloak.models.KeycloakSessionFactory;
 import org.keycloak.provider.ProviderConfigProperty;
 
-public class UserSessionRefreshTimeResourcePolicyProviderFactory implements ResourcePolicyProviderFactory<UserSessionRefreshTimeResourcePolicyProvider> {
+public class UserSessionRefreshTimeWorkflowProviderFactory implements WorkflowProviderFactory<UserSessionRefreshTimeWorkflowProvider> {
 
     public static final String ID = "user-refresh-time-resource-policy";
```
sguilhen: same here

model/jpa/src/main/java/org/keycloak/models/workflow/WorkflowStateEntity.java
```diff
@@ -26,26 +26,26 @@
...
- * Represents the state of a resource within a time-based policy flow.
+ * Represents the state of a resource within a time-based workflow.
  */
 @Entity
-@Table(name = "RESOURCE_POLICY_STATE")
-@IdClass(ResourcePolicyStateEntity.PrimaryKey.class)
-public class ResourcePolicyStateEntity {
+@Table(name = "WORKFLOW_STATE")
+@IdClass(WorkflowStateEntity.PrimaryKey.class)
+public class WorkflowStateEntity {
```
sguilhen, line 34: NOTE: if we change the table name, we should also change the entity name to WorkflowActionStateEntity

model/jpa/src/main/resources/default-persistence.xml
```diff
@@ -112,7 +112,7 @@
             See also https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#best-practices-jdbc-batching.
             Keycloak uses many nested objects spread across multiple tables, so a benefit is likely.
             -->
-            <property name="hibernate.order_inserts" value="true" />
```
sguilhen: typo here at the end of the line

model/storage-private/src/main/java/org/keycloak/models/workflow/ResourcePolicyStateSpi.java
```diff
@@ -37,11 +37,11 @@ public String getName() {
```
sguilhen: Shouldn't the class be renamed too?

services/src/main/java/org/keycloak/authorization/admin/PolicyService.java
```diff
@@ -202,7 +202,7 @@ public Response findByName(@QueryParam("name") String name, @QueryParam("fields"
         ),
         @APIResponse(responseCode = "204", description = "No Content")
     })
-    public Response findAll(@QueryParam("policyId") String id,
```
sguilhen: This doesn't seem intentional

services/src/main/java/org/keycloak/authorization/protection/policy/UserManagedPermissionService.java
```diff
@@ -82,11 +82,11 @@ public Response create(@PathParam("resourceId") String resourceId, UmaPermission
         return findById(delegate.create(representation).getId());
     }
 
-    @Path("{policyId}")
```
sguilhen: same thing here - not intentional

sguilhen (changes requested):

model/jpa/src/main/resources/META-INF/jpa-changelog-26.4.0.xml
mabartos: > As Pedro mentioned, the team had some discussions and settled on "Keycloak Workflows" as the feature name, and we're trying to align with that decision here.

@vramik @pedroigor Sure, whatever works for your team :)

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowActionSpi.java
```diff
@@ -15,15 +15,15 @@
...
 
-package org.keycloak.models.policy;
+package org.keycloak.models.workflow;
 
 import org.keycloak.provider.Provider;
 import org.keycloak.provider.ProviderFactory;
 import org.keycloak.provider.Spi;
 
-public class ResourcePolicySpi implements Spi {
+public class WorkflowActionSpi implements Spi {
```
pedroigor: After discussing with @sguilhen, I think we should change "Action" to "Step".

The reasoning is that a step can be an action but also any other step in a workflow. Today, steps are basically actions that do something on a resource. But in the future, the "step" term can be useful to support other things than just actions executed.

Don't ask me what those would be now but perhaps things like a conditional step like:

```
step: conditional
       if: <condition>
       then: <another step|action>
       else: <another step|action>
```

I hope you see the point ....

services/src/main/java/org/keycloak/models/workflow/AddRequiredStepProvider.java
```diff
@@ -9,19 +8,19 @@
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 
-import static org.keycloak.models.policy.ResourceAction.AFTER_KEY;
+import static org.keycloak.models.workflow.Step.AFTER_KEY;
 
-public class AddRequiredActionProvider implements ResourceActionProvider {
+public class AddRequiredStepProvider implements WorkflowStepProvider {
```
pedroigor: It should be `AddRequiredActionStepProvider`. This one adds a required action to the user.

services/src/main/java/org/keycloak/models/workflow/AddRequiredStepProviderFactory.java
```diff
@@ -10,13 +10,13 @@
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.provider.ProviderConfigurationBuilder;
 
-public class AddRequiredActionProviderFactory implements ResourceActionProviderFactory<AddRequiredActionProvider>, ConfiguredProvider {
+public class AddRequiredStepProviderFactory implements WorkflowStepProviderFactory<AddRequiredStepProvider>, ConfiguredProvider {
```
pedroigor: `AddRequiredActionStepProviderFactory`.

services/src/main/java/org/keycloak/models/workflow/ResourceStepRunnerScheduledTask.java
```diff
@@ -9,21 +9,21 @@
 import org.keycloak.timer.ScheduledTask;
 
 /**
- * A {@link ScheduledTask} that runs all the scheduled actions for resources on a per-realm basis.
+ * A {@link ScheduledTask} that runs all the scheduled steps for resources on a per-realm basis.
  */
-final class ResourceActionRunnerScheduledTask implements ScheduledTask {
+final class ResourceStepRunnerScheduledTask implements ScheduledTask {
```
pedroigor: Perhaps `WorkflowRunnerScheduledTask` ?

services/src/main/java/org/keycloak/models/workflow/ResourceStepRunnerScheduledTask.java
```diff
@@ -35,17 +35,17 @@ private void runScheduledTasksOnRealm(String id) {
...
-                logger.errorf(e, "Failed to run resource policy actions on realm with id '%s'", id);
+                logger.errorf(e, "Failed to run workflow steps on realm with id '%s'", id);
             }
         });
     }
 
     @Override
     public String getTaskName() {
-        return "resource-policy-action-runner";
+        return "workflow-step-runner";
```
pedroigor: `workflow-runner-task`

model/storage-private/src/main/java/org/keycloak/models/workflow/WorkflowStateSpi.java
```diff
@@ -15,13 +15,13 @@
...
+package org.keycloak.models.workflow;
 
 import org.keycloak.provider.Provider;
 import org.keycloak.provider.ProviderFactory;
 import org.keycloak.provider.Spi;
 
-public class ResourcePolicyStateSpi implements Spi {
+public class WorkflowStateSpi implements Spi {
 
     public static final String NAME = "resource-policy-state";
```
pedroigor: `workflow-state`

server-spi-private/src/main/java/org/keycloak/models/workflow/Step.java
```diff
@@ -15,38 +15,38 @@
...
-package org.keycloak.models.policy;
+package org.keycloak.models.workflow;
 
 import java.util.List;
 
 import org.keycloak.common.util.MultivaluedHashMap;
 import org.keycloak.component.ComponentModel;
 
-public class ResourceAction implements Comparable<ResourceAction> {
+public class Step implements Comparable<Step> {
```
pedroigor: `WorkflowStep`

pedroigor (approved):

PR #42528, opened 2025-09-10
Disable preview of the json file if the content is more than 100KB

### LINKED ISSUE #40557: Import from UI causes extreme lag or entirely unresponsive page since 26.1
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Laggy page after selecting relatively small JSON for Partial Import or Create Realm (from import).

"Page Unresponsive" after selecting relatively large JSON for Partial Import or Create Realm (from import).

Note that the performance of the actual importing process does not seem to be slower, just the loading of the JSON on the page gives these issues.

### Version

26.1, 26.2

### Regression

### Expected behavior

After selecting the following (relatively small) realm-export for Partial Import, the json loads in <1 second and the page is fully responsive without lag. Similarly for Create Realm (from import).

[realm-export (1).json](https://github.com/user-attachments/files/20776242/realm-export.1.json)

After selecting the following (relatively large) realm-export for Partial Import, the json loads in <2 seconds and the page is fully responsive without lag. Similarly for Create Realm (from import). 

[realm-0-users-1.json](https://github.com/user-attachments/files/20776408/realm-0-users-1.json)

The expected behavior occurs in 26.0, but not in 26.1 or 26.2

### Actual behavior

After selecting the following (relatively small) realm-export for Partial Import, the json loads in >5 seconds and then the page is very laggy. This can be noticed when one tries to check the boxes. It takes almost half a minute after clicking the boxes for them to actually appear checked. Similar lag occurs when you try to Create Realm (from import), except there are no boxes to notice the slow feedback.

[realm-export (1).json](https://github.com/user-attachments/files/20776242/realm-export.1.json)

After selecting the following (relatively large) realm-export for Partial Import or Create Realm (from import), I get "Page Unresponsive".

[realm-0-users-1.json](https://github.com/user-attachments/files/20776408/realm-0-users-1.json)

### How to Reproduce?

Reproduction steps are implied in my description of Expected and Actual behaviors.

### Anything else?

_No response_

rmartinc:
Closes #40557

The root cause of the problem is the upload is doing a preview of the file that is also highlighted, that is consuming a lot of time when the file is big. This PR does the following:

* `useMemo` is added to cache the content by value and language. This makes to not re-render if the content is the same. This is OK for partial import when selecting the parts to upload.
* Disable preview if the content is bigger than 100KB by default (parameter `previewMaxLength`) because typing is too slow when it's big.
* This means that you can write and modify while the size is less than 100KB. Preview is disabled when the data is bigger than 100KB.

The only other solution is changing the `react-textarea-code-editor` for other project that is more performant. But probably there will be always a limit, so I think that setting the limit is more or less OK. You can always edit the file outside the console.

Draft for the moment.

### DIFF
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -870,6 +870,7 @@ votedToStatus=voted to {{status}}
 credentialResetConfirmText=Are you sure you want to email this user?
 clientScopeType.default=Default
 helpFileUpload=Upload a JSON file
+fileUploadPreviewDisabled=Preview disabled because content is too long.
 addProvider_one=Add {{provider}} provider
 clientPoliciesPolicies=Client Policies Policies
 editUSernameHelp=If enabled, the username is editable, otherwise it is read-only.
--- a/js/apps/admin-ui/src/components/form/CodeEditor.tsx
+++ b/js/apps/admin-ui/src/components/form/CodeEditor.tsx
@@ -1,4 +1,5 @@
 import CodeEditorComponent from "@uiw/react-textarea-code-editor";
+import { useMemo } from "react";
 
 type CodeEditorProps = {
   id?: string;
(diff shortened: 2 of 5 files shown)


ssilvert (dismissed):
I like this.  LGTM.

rmartinc:
Thanks @ssilvert! I have added a test too and moved out of draft.

ssilvert (approved):
I tried it out.  Works great.  Approved.

PR #42530, opened 2025-09-10
Fix for missing object representation in admin event log when deleting user, group, client

### LINKED ISSUE #33009: Delete Client and role's admin event doesn't have representation while other delete entities have the representation
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When we delete client OR Delete role, representation json is NULL While other DELETE operation like delete any resources, policy, permission or scope under client have details in representation json.

### Version

25.0.5(Latest)

### Regression

### Expected behavior

We should have representation json block to identify what was deleted just like other delete operation. E.g DELETE any Authorization entities like Delete scope or delete resource or delete policy or delete permission those event have proper representation block. 

### Actual behavior

DELETE Client and DELETE Role doesn't have representation json block.

### How to Reproduce?

1. Enable admin event config
2. Create Client and Delete Clinet
3. Check for Delete client event's representation
4. Create Role and Delete role
5. Again check for Delete role event's representation

**DELETE** CLIENT's admin event representation
image

**DELETE** Client's authorization scop's admin event representation:
image

**DELETE** role's admin event representation:
image

### Anything else?

_No response_

JakubWoo:
Fix for missing object representation in admin event log when deleting user, group, client

Closes #33009


LukaszKochaniak (approved):
LGTM

pedroigor (commented):
Thanks, @JakubWoo. I see what you are trying to achieve by making delete admin events for clients, groups, and users, consistent with other resources from authorization services.

Despite that, is there any other use case you are trying to solve when processing delete events for these resources and using their representation?

stianst (changes requested):
We are intentionally not storing previous state of resources that are updated. The even should only store what is happening right now, and as the resource is deleted there is nothing to store.

JakubWoo:
> Thanks, @JakubWoo. I see what you are trying to achieve by making delete admin events for clients, groups, and users, consistent with other resources from authorization services.
> 
> Despite that, is there any other use case you are trying to solve when processing delete events for these resources and using their representation?

Yes. Some organizations find this critical to keep deleted user information in the admin log for audit purpose. There is no other way of retrieving details of deleted user than admin log user representation, since user does not exist any more.

JakubWoo:
> We are intentionally not storing previous state of resources that are updated. The even should only store what is happening right now, and as the resource is deleted there is nothing to store.

Deleted user information is critical for audit purposes. Some organizations raise missing user representation as critical issue when using Keycloak.

pedroigor:
@JakubWoo You'll probably have past events recorded for the deleted user. The delete event should include a value in the `resourcePath` field that contains the user's UUID. Isn't it enough for audit purposes? I understand that having the UUID in the `resourcePath` is not the best UX because you need to parse the URL.

Would be enough to also include the username (and the UUID) in the details instead of the full representation?

JakubWoo:
> @JakubWoo You'll probably have past events recorded for the deleted user. The delete event should include a value in the `resourcePath` field that contains the user's UUID. Isn't it enough for audit purposes? I understand that having the UUID in the `resourcePath` is not the best UX because you need to parse the URL.
> 
> Would be enough to also include the username (and the UUID) in the details instead of the full representation?

UUID and resourcePath is not enough - sure there is an event but we need a representation and at least a username to be able to identify a resource. Especially when we use both local admin users and LDAP provider. UUID is not aligned with LDAP accounts, there is no chance to figure it out.

What are your concerns regarding including full representation? Its there for adding event or updating user, right?

pedroigor:
@JakubWoo, I do not have any specific concern, but as you pointed out, there are inconsistencies in how delete events are created.

Based on the comments from @stianst, the inconsistency is in the places where the representation is included in `DELETE` events. Like in authorization services and roles/scopes.

If storing an identifier, such as the username, client ID, or group name/path, is enough to solve the requirements you and others are trying to address, it would be easier to have something in.

JakubWoo:
> If storing an identifier, such as the username, client ID, or group name/path, is enough to solve the requirements you and others are trying to address, it would be easier to have something in.

Lets make it a compromise. I'm going to put only UUID and name into the representation object for user/client/group. Would that work for you both @pedroigor @stianst ?

Thanks.

JakubWoo:
Please find updated branch and PR:
https://github.com/keycloak/keycloak/pull/42615

CC @pedroigor @stianst

AndrewK15 (approved):

yweibel:
@pedroigor @stianst would you require any further modification or can we proceed with the proposed changes?

martin-kanis:
Closing in favour of https://github.com/keycloak/keycloak/pull/43620

PR #42535, opened 2025-09-10
Add configurable HTTP retry mechanism for OCSP validation

chance-coleman:
## Description
closes #42401 

This PR adds two key improvements to Keycloak:

1. **General-Purpose HTTP Retry Mechanism**: Implements a general-purpose HTTP retry mechanism that allows automatic retrying of HTTP requests when they fail due to transient issues such as network timeouts, temporary server errors, or other recoverable conditions.

2. **OCSP Retry UI Configuration**: Adds UI configuration inputs for OCSP retry count and timeout settings in the X.509 client certificate authentication flow. This allows administrators to configure retry behavior for OCSP certificate validation through the Keycloak admin UI.

The implementation extends the existing `HttpClientProvider` interface to provide retriable HTTP clients that can be used throughout the codebase. This approach provides a consistent way to handle retries across different components, including OCSP certificate validation.

## Motivation

In environments with unreliable network connections or when interacting with services that may experience temporary outages, HTTP requests can fail unnecessarily. This is particularly problematic for critical operations like OCSP certificate validation, where a temporary network issue could prevent a user from authenticating.

By implementing an opt-in general-purpose retry mechanism, we can improve the reliability of Keycloak in these environments without changing default behavior or duplicating retry logic across different components.


services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -96,6 +99,116 @@ public CloseableHttpClient getHttpClient() {
                 return httpClient;
             }
 
+            @Override
+            public CloseableHttpClient getRetriableHttpClient() {
+                return getRetriableHttpClient(defaultRetryConfig);
```
slaskawi: With the current approach, every time the caller obtains a new Retriable HTTP Client, a new instance is created. Right now, the blast radius is probably not large as it's used only in OCSP. But this probably doesn't scale very well as creating a new client takes time. 

Having said that, would it make sense to move this code into `lazyInit` and create a new field in the `DefaultHttpClientFactory` called `retriableHttpClient`? In other words, you'd do exactly the same initialization process as `httpClient` and essentially cache a created client in the `DefaultHttpClientFactory`.

chance-coleman: That is a really good catch, thanks! I have refactored the retriableClient to use a cache similar to how the default client works.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -96,6 +100,24 @@ public CloseableHttpClient getHttpClient() {
...
+            @Override
+            public CloseableHttpClient getRetriableHttpClient() {
+                return retriableHttpClient;
+            }
+
+            @Override
+            public CloseableHttpClient getRetriableHttpClient(RetryConfig retryConfig) {
+                // If using default config, return the cached client
+                if (retryConfig == null ||
+                    (defaultRetryConfig.getMaxRetries() == retryConfig.getMaxRetries() &&
```
slaskawi: Just a nitpick - you could implement logical equality in the `RetryConfig` by overriding the `equals` method. Josh Bloch wrote a fantastic paragraph on it here: https://ahdak.github.io/blog/effective-java-part-2/

Once implemented, this would become `defaultRetryConfig.equals(retryConfig)`. You will also be able to omit the null check by putting `defaultRetryConfig` that can't be null on the left side.

slaskawi (approved):
LGTM with a small nitpick, which is up to you to implement.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: Implemented this based on that blog post, overrode equals() and hashCode() with a couple small tests. I had played with the idea of implementing a more robust equality check, but ended up going with the simpler path. Since you called this out i think it's a good idea and worth the effort.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -96,6 +100,22 @@ public CloseableHttpClient getHttpClient() {
                 return httpClient;
             }
 
+            @Override
+            public CloseableHttpClient getRetriableHttpClient() {
+                return retriableHttpClient;
+            }
+
+            @Override
+            public CloseableHttpClient getRetriableHttpClient(RetryConfig retryConfig) {
```
rmartinc: When I thought about this my initial idea was adding just one retriable configuration for everything. So this method would be `getRetriableHttpClient()` and the retry options are just defined in this factory as options (similar to other options for proxy or similar in the normal client). This would simplify the configuration and would remove the need to add the UI options for the places that we decide to use the retry. This way adding this to CRL for example is just changing the method without adding any new configuration options. Besides it avoids a lot of the burden in this PR. WDYT?

The same configuration should be used for both clients, but the retriable one adds the specifics for the retry. There are only two clients defined that are initialize only once each. By default (if no retry is configured) the same client is used.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -165,88 +188,227 @@ public String getId() {
...
+                                long baseDelay = retryConfig.getInitialBackoffMillis() *
+                                        (long)Math.pow(retryConfig.getBackoffMultiplier(), executionCount - 1);
+                                long delay = baseDelay;
+                                if (retryConfig.isUseJitter()) {
+                                    // Add +/- 50% jitter
+                                    double jitter = 1.0 - retryConfig.getJitterFactor() +
+                                            (Math.random() * retryConfig.getJitterFactor() * 2.0);
+                                    delay = (long)(baseDelay * jitter);
+                                }
+                                Thread.sleep(delay);
```
rmartinc, line 291: I really don't like this sleep. What is the reason for this? Normal `DefaultHttpRequestRetryHandler` is not enough for you?

rmartinc (commented):
Thanks @chance-coleman for the PR!

I have added two comments but I want that @mposolda also checks this. So do not start changing code until he also reviews the PR and gives his opinion.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: Hmmm good question, my understanding was that Apache HttpClient 4.5.14 `DefaultHttpRequestRetryHandler` does not implement any backoff or delay between retries. The retry logic simply checks if we should retry and immediately returns the decision.

The sleep in the code was necessary to implement the exponential backoff with jitter functionality. Without it, the retries would happen in rapid succession, which i believe to be a more robust solution, although more complex.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: ahh i can see how this would be simpler and more consistent. the only downside i can think of here is less flexibility in how retries would be approached. I'm open to this implementation if that is what you all would prefer.

slaskawi: @chance-coleman has a point here in my opinion. Depending on the area of the code, we may want to use different setting. In other words, what will work correctly for OCSP might not be a good fit for Identity Providers or other functionalities.

rmartinc: Yep, this is just a matter of what we prefer. Let's see what other people think aboy this.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
mposolda: There is also `ServiceUnavailableRetryStrategy` . Could not that one be used instead of implementing the retry logic in our own codebase? For example see the 2nd post from https://stackoverflow.com/questions/48541329/timeout-between-request-retries-apache-httpclient .

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
mposolda: +1 for just `getRetriableHttpClient()` . If we ever need different configuration for different "use-cases" of HTTP client, we can add new method if needed.

I can imagine that we may want different configurations for different use-cases, however this might mean also different configurations for other aspects than just retry configuration. For more flexibility, maybe the best is to have the method like:
```
            @Override
            public CloseableHttpClient getHttpClient(String context) {
```
where context is for example something like "ocsp" or "facebook-idp" . Then in the configuration of Http-client provider, we can add namespaces for various configuration options. For example we can have `ocsp--proxy-mappings` if we want to have different proxy-mappings configuration for the case when HTTP client is called in the context of "ocsp"
. This will also allow to have different retry configurations for "ocsp" and different for other things. The `Config.Scope` has some support for namespaces in itself AFAIK.

I do not insist on this, but would prefer something like this instead of introducing `getRetriableHttpClient(RetryConfig)` , which IMO does not have much flexibility and at the same time, it is not used anywhere right now.

The `getRetriableHttpClient()` is also less ideal than namespaces IMO, but maybe good compromise if others prefer this instead of introducing "namespaces" .

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -342,6 +504,57 @@ public List<ProviderConfigProperty> getConfigMetadata() {
                 .helpText("Maximum size of a response consumed by the client (to prevent denial of service)")
                 .defaultValue(HttpClientProvider.DEFAULT_MAX_CONSUMED_RESPONSE_SIZE)
                 .add()
+                .property()
+                .name("http-client.default-max-retries")
```
mposolda: I am not sure why those configuration options have prefix `http-client` here? Other config options of the provider do not have this prefix.

Also there is one confusing aspect: When someone reads the documentation of this provider, he would see `http-client-default-max-retries` configured to 3, which imply that this is used by default by HTTP client. But AFAIK, this is used just if `getRetriableHttpClient()` is called right? By default, the `getHttpClient()` or other methods do not use this retry configuration at all and requests are not retried. Which is not clear from this docs...

Might be better if we have something like `.name("default-max-retries")` (which would be 0 by default and hence disabled retries) and `.name("ocsp--default-max-retries")` (which would be 3 and hence enabled retries for OCSP by default). I proposed in the other comment to use "namespaces" for the configurations. This will be likely harder to implement, but seems to me like more clear configuration.

mposolda (commented):
@chance-coleman @rmartinc @slaskawi  Added some additional review comments. WDYT?

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: While `ServiceUnavailableRetryStrategy` is great for fine-tuning HTTP error responses, it's limited to HTTP-layer issues and doesn't handle lower-level network problems. Our current solution handles the full spectrum of failures; from TLS errors and dropped connections to server errors, in one straightforward implementation. This single-point approach ensures reliable retries whether the issue is in the network stack or the HTTP layer, without the complexity of managing multiple retry strategies.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: @mposolda i can see the benefits in this approach. The context-based approach is more flexible and maintainable in the long run. The main trade-off is slightly more complex configuration. It's definitely a balance and some preference, ultimately it's your codebase so i'm happy to make changes either way!

slaskawi:
> @chance-coleman @rmartinc @slaskawi Added some additional review comments. WDYT?

@mposolda LGTM! (I already added my approval and together with @chance-coleman we addressed my comments before this PR has been created).

slaskawi:
Hey @mposolda 👋🏻 

Could I ask you to have a look at this one? It seems @chance-coleman addressed all the comments and I think we should be close to getting it in.

stianst (commented):
Not convinced about this should be configured on the X509 authenticator specifically. Looking at the issue is explicitly mentions OCSP. However, if network traffic is unstable then this will affect any outgoing HTTP requests from Keycloak.

I would argue we should instead consider having config options for this for the HTTP client provider in general.

One thing to bear in mind here though if there are a few retries, and the timeout is large, then the original incoming HTTP request to Keycloak would likely have timed out in the meantime. To me that is another argument to make it a server wide configuration option, as incoming request timeout should be higher than the total of retries + timeouts for outgoing requests.

@ahus1 @shawkins WDYT?

stianst (changes requested):
Thinking more about this, it really isn't something we should make configurable as fine-grained as within the X509 authenticator, but just make a server wide config option for outgoing http requests

slaskawi:
> Thinking more about this, it really isn't something we should make configurable as fine-grained as within the X509 authenticator, but just make a server wide config option for outgoing http requests

@stianst Could you please tell us more about the reasoning behind this?

When we were discussing this with @chance-coleman, we were thinking about minimizing the impact of introducing this configuration (so exactly opposite to your thoughts). The main argument here was that Keycloak can have multiple Realms, Orgs and each one of them might use OCSPs with different availability SLAs. Some of them might be flaky by nature (and we should retry on them) and some are expected to rock solid (and retrying them just increases a storm of requests and generally doesn't help bringing it back online). 

So I wanted to double check if we really would like to make this a server-level setting affecting all Realms, Orgs and all the egress traffic from Keycloak.

stianst:
> > Thinking more about this, it really isn't something we should make configurable as fine-grained as within the X509 authenticator, but just make a server wide config option for outgoing http requests
> 
> @stianst Could you please tell us more about the reasoning behind this?
> 
> When we were discussing this with @chance-coleman, we were thinking about minimizing the impact of introducing this configuration (so exactly opposite to your thoughts). The main argument here was that Keycloak can have multiple Realms, Orgs and each one of them might use OCSPs with different availability SLAs. Some of them might be flaky by nature (and we should retry on them) and some are expected to rock solid (and retrying them just increases a storm of requests and generally doesn't help bringing it back online).
> 
> So I wanted to double check if we really would like to make this a server-level setting affecting all Realms, Orgs and all the egress traffic from Keycloak.

What about client backchannel logout, brokering requests, etc.. They can also be flaky. Also, really don't think network settings is something a realm admin should care about, that's more the server admin (and person that manages the environment).

I'd start with just a global setting; then perhaps add some sort of regex where you can configure what URLs should have retries and what should not. Again really think this is a server/environment config thing, and not something that should be managed by realm admins.

server-spi-private/src/main/java/org/keycloak/connections/httpclient/HttpClientProvider.java
```diff
@@ -40,6 +40,20 @@ public interface HttpClientProvider extends Provider {
...
+     * Returns a {@code CloseableHttpClient} with server-wide retry behavior.
+     * The retry behavior is configured globally in the HTTP client factory.
+     * <p>
+     * <b>The returned {@code HttpClient} instance must never be {@code close()}d by
+     * the caller.</b>
+     * <p>
+     * 
+     * @return A CloseableHttpClient with retry capabilities based on global configuration
+     */
+    default CloseableHttpClient getRetriableHttpClient() {
```
slaskawi: We can remove this method too.

server-spi-private/src/main/java/org/keycloak/utils/OCSPProvider.java
```diff
@@ -123,16 +122,10 @@ public OCSPRevocationStatus check(KeycloakSession session, X509Certificate cert,
 
     protected byte[] getEncodedOCSPResponse(KeycloakSession session, byte[] encodedOCSPReq, URI responderUri) throws IOException {
 
-        CloseableHttpClient httpClient = session.getProvider(HttpClientProvider.class).getHttpClient();
+        // Use the retriable HTTP client which will apply retry and timeout settings from RetryConfig
+        CloseableHttpClient httpClient = session.getProvider(HttpClientProvider.class).getRetriableHttpClient();
```
slaskawi: We can use `getHttpClient` here

services/src/main/java/org/keycloak/authentication/authenticators/x509/AbstractX509ClientCertificateAuthenticator.java
```diff
@@ -106,6 +105,12 @@ protected static class CertificateValidatorConfigBuilder {
         static CertificateValidator.CertificateValidatorBuilder fromConfig(KeycloakSession session, X509AuthenticatorConfigModel config) throws Exception {
 
             CertificateValidator.CertificateValidatorBuilder builder = new CertificateValidator.CertificateValidatorBuilder();
+
+            // Store the config in the session so it can be accessed by the
+            // CertificateValidator
+            // This allows passing the OCSP retry settings to the validator
+            session.setAttribute("x509-auth-config", config);
```
slaskawi: Why do we need to store the config here?

services/src/main/java/org/keycloak/authentication/authenticators/x509/AbstractX509ClientCertificateAuthenticatorFactory.java
```diff
@@ -115,7 +115,6 @@ public abstract class AbstractX509ClientCertificateAuthenticatorFactory implemen
         serialnumberHex.setDefaultValue(Boolean.toString(false));
         serialnumberHex.setHelpText("Use the hex representation of the serial number. This option is relevant for authenticators using serial number.");
```
slaskawi, line 130: We can revert these empty line removals (and additions in other files) to avoid introducing noise to this PR.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -68,13 +68,17 @@ public class DefaultHttpClientFactory implements HttpClientFactory {
     public static final String MAX_CONSUMED_RESPONSE_SIZE = "max-consumed-response-size";
 
     private volatile CloseableHttpClient httpClient;
+    private volatile CloseableHttpClient retriableHttpClient;
```
slaskawi: Since we configure the retry settings server-wide, there's no need to introduce another `CloseableHttpClient`. All the implementation should go into the first one (the name that's called `httpClient`).

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -165,88 +177,227 @@ public String getId() {
     @Override
     public void init(Config.Scope config) {
         this.config = config;
+
+        // Initialize server-wide retry configuration
+        int maxRetries = config.getInt("max-retries", 0); // No retries by default (opt-in)
+        boolean retryOnIOException = config.getBoolean("retry-on-io-exception", true);
```
slaskawi: Can we alter the name and call it "retry-on-error"? Mentioning the IOException seems to specific for the configuration parameter.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -165,88 +177,227 @@ public String getId() {
...
-                    	logger.warn("TruststoreProvider is disabled");
-                    } else {
-                        builder.hostnameVerification(truststoreProvider.getPolicy());
-                        try {
-                            builder.trustStore(truststoreProvider.getTruststore());
-                        } catch (Exception e) {
-                            throw new RuntimeException("Failed to load truststore", e);
+        // Create HTTP client builder
+        HttpClientBuilder builder = newHttpClientBuilder();
+
```
slaskawi: I wonder if we could revert the implementation to the original on, extract these settings into a separate method and call it like this:

```
RetryConfig retryConfig = ... ;
builder.retries(retryConfig);
```

All the wiring would go to the builder and the implementation would be very small and lean in this class. 

What do you think about this approach?

server-spi-private/src/main/java/org/keycloak/connections/httpclient/HttpClientProvider.java
chance-coleman: Good call, removed this.

server-spi-private/src/main/java/org/keycloak/utils/OCSPProvider.java
chance-coleman: Good call, changed this to use getHttpClient.

services/src/main/java/org/keycloak/authentication/authenticators/x509/AbstractX509ClientCertificateAuthenticatorFactory.java
chance-coleman: I think i got all the line removals/additions.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: Removed retriableHttpClient field, built retry into the main httpClient

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: Renamed property from retry-on-io-exception to retry-on-error

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
chance-coleman: Refactored to extract retry logic into helper method and used by builder

services/src/main/java/org/keycloak/authentication/authenticators/x509/AbstractX509ClientCertificateAuthenticator.java
chance-coleman: This was leftover from when we were passing OCSP retry settings. Since we removed that, this line should be deleted.

docs/documentation/server_admin/topics/authentication/x509.adoc
```diff
@@ -143,6 +143,8 @@ Checks the certificate revocation status by using Online Certificate Status Prot
 *OCSP Fail-Open Behavior*::
 By default the OCSP check must return a positive response in order to continue with a successful authentication. Sometimes however this check can be inconclusive: for example, the OCSP server could be unreachable, overloaded, or the client certificate may not contain an OCSP responder URI. When this setting is turned ON, authentication will be denied only if an explicit negative response is received by the OCSP responder and the certificate is definitely revoked. If a valid OCSP response is not available the authentication attempt will be accepted.
 
+OCSP retry behavior is configured server-wide through the HTTP client provider. See the Server Configuration Guide for details on configuring retry settings for all outgoing HTTP requests, including OCSP validation.
```
slaskawi: Could you please document the new settings in https://github.com/keycloak/keycloak/blob/37bea126c723068b35d1f46bab5367c18db156ab/docs/guides/server/outgoinghttp.adoc and add a reference here to that page?

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -219,7 +219,7 @@ private void lazyInit(KeycloakSession session) {
                     boolean disableTruststoreProvider = truststoreProvider == null || truststoreProvider.getTruststore() == null;
 
                     if (disableTruststoreProvider) {
-                    	logger.warn("TruststoreProvider is disabled");
+                    logger.warn("TruststoreProvider is disabled");
```
slaskawi: Seems like a misaligned line.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -230,8 +230,8 @@ private void lazyInit(KeycloakSession session) {
                     }
 
                     if (disableTrustManager) {
-                    	logger.warn("TrustManager is disabled");
-                    	builder.disableTrustManager();
+                    logger.warn("TrustManager is disabled");
```
slaskawi: Seems like a misaliged line.

slaskawi (commented):
The code LGTM! Requested only doc changes and a few nit fixes.

slaskawi (approved):
LGTM! Thanks for all the changes @chance-coleman !

@stianst This one is ready for re-review.

slaskawi:
@chance-coleman I've just spotted - the DCO job is failing. You probably need to add a sign-off

docs/guides/server/outgoinghttp.adoc
```diff
@@ -57,6 +57,43 @@ Specify proxy configurations for outgoing HTTP requests. For more details, see <
 *disable-trust-manager*::
 If an outgoing request requires HTTPS and this configuration option is set to true, you do not have to specify a truststore. This setting should be used only during development and *never in production* because it will disable verification of SSL certificates. Default: false.
 
+== Configuring retry behavior for outgoing HTTP requests
```
stianst, line 60: I would add something around the fact that outgoing request retries should not exceed the timeout for incoming requests

chance-coleman: I've added a callout like this: 

`IMPORTANT: Do not let outgoing retry duration exceed the caller’s timeout. Otherwise, the caller may time out and see an error while {project_name} continues retrying in the background.`

chance-coleman: @stianst does this suffice? If so what else can i do to push this PR along. Thanks!

stianst (commented):
Looks okay to me, haven't done a full review though

server-spi-private/src/test/java/org/keycloak/connections/httpclient/RetryConfigTest.java
```diff
@@ -0,0 +1,214 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.connections.httpclient;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
```
mposolda: Please try to avoid using "wildcard imports" . They are not recommended anyway, but now there is some recent change done in Keycloak to remove them in a more strict way and they will be prohibited for the future: https://github.com/keycloak/keycloak/commit/d8275fe5df94c436b176b2ad58f897b7e993a8fa

server-spi-private/src/main/java/org/keycloak/utils/OCSPProvider.java
```diff
@@ -126,13 +124,6 @@ protected byte[] getEncodedOCSPResponse(KeycloakSession session, byte[] encodedO
         CloseableHttpClient httpClient = session.getProvider(HttpClientProvider.class).getHttpClient();
         HttpPost post = new HttpPost(responderUri);
         post.setHeader(HttpHeaders.CONTENT_TYPE, "application/ocsp-request");
-
-        final RequestConfig params = RequestConfig.custom()
```
mposolda, line 130: This change makes sense, but strictly speaking, it is background incompatible change (as for example default socket timeout is 5 seconds for the default HTTP client provider).

IMO it is OK to include this change in your PR, but it would be also good to add a note about this to the file https://github.com/keycloak/keycloak/blob/main/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc . Probably to the "Notable changes" section. I assume that note can mention that OCSP behaviour used hardcoded socket and connection timeouts of 10 seconds before, but now it is updated to use the settings from the HTTP client configuration.

WDYT?

docs/guides/server/outgoinghttp.adoc
```diff
@@ -57,6 +57,44 @@ Specify proxy configurations for outgoing HTTP requests. For more details, see <
...
+
+{project_name} can automatically retry failed outgoing HTTP requests. This is useful for handling transient network errors or temporary service unavailability. Retry behavior is disabled by default and must be explicitly enabled.
+
+The following are the retry configuration options:
+
+*max-retries*::
+Maximum number of retry attempts for failed HTTP requests. Set to 0 to disable retries. Default: 0.
+
+*retry-on-error*::
+Whether to retry HTTP requests when errors occur. Default: true.
```
mposolda: It is not clear to me what "retry-on-error" means. Is it maybe possible to add some more clarification (or even an example like you did for "backoff-multiplier" option)?

For example:
- When `max-retries` is set to 0, but `retry-on-error` is set to `true`, will the HTTP request be retried or not? 
- What exactly is considered an "error" ? Is it HTTP error status or something different? If Http status, then what statuses are we talking about (just a note that consumer of this docs may not know that it uses Apache HTTP client library under the covers, so assume that he is not familiar with that library)
- If for example `max-retries` is set to 3, but `retry-on-error` is set to `false`, what will be the behaviour? Which requests would be retried if any? Does it even makes sense to have such a config?

mposolda (commented):
@chance-coleman @slaskawi Thanks for all the updates! I've added few small inline comments. Could you please check them?

docs/guides/server/outgoinghttp.adoc
rmartinc: The retry-on-error is used as the parameter `requestSentRetryEnabled` of the [constructor](https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/client/DefaultHttpRequestRetryHandler.html#DefaultHttpRequestRetryHandler(int,%20boolean)). So this should be true for re-try all common requests (HttpGet, HttpPost,...). Maybe we can just set this always to true (no option) and just use `max-retries` >= 0.

rmartinc (commented):
For me it's OK once the @mposolda comments are fixed. And I would use always true for the option `retry-on-error`, because I suppose we are always using common implementations (HttpGet, HttpPost,... that are not retried if the parameter is set to false).

docs/guides/server/outgoinghttp.adoc
mposolda: Yeah, I was able to see that. But without looking to the code, this option will not be clear to me at all just after looking at this docs :-)

Removing this option entirely from Keycloak config and always set to true when `max-retries >= 0` might be even better though?
@chance-coleman @slaskawi WDYT?

slaskawi: @rmartinc has a good point. We could decide to turn the entire mechanism on/off by checking `max-retries`. I'm fully onboard with this idea.

server-spi-private/src/test/java/org/keycloak/connections/httpclient/RetryConfigTest.java
chance-coleman: Good catch, i've removed the wildcard import for specific imports.

server-spi-private/src/main/java/org/keycloak/utils/OCSPProvider.java
chance-coleman: I think this makes sense, i've updated the changes doc to include a callout for OCSP behavior changing.

docs/guides/server/outgoinghttp.adoc
chance-coleman: Appreciate the feedback! I totally agree that the `retry-on-error` is an unnecessary level of complexity. I've refactored that config value out and instead depend on the `max-retries` along with a value of `true` for all common requests.

mposolda (approved):
@chance-coleman @rmartinc @slaskawi Thanks for the additional updates, reviews and feedback. And thanks for your patience during the work on this PR!

I hope we can merge once tests are OK (Documentation external links tests are not related to this PR and are not a blocker for merging this).

slaskawi:
Thank you @mposolda !

PR #42550, opened 2025-09-11
Add missing fields for client offline session timeout and lifespan

### LINKED ISSUE #42369: Missing client session offline settings on realm level in the admin UI

ahus1:
Closes #42369

### DIFF
--- a/docs/documentation/server_admin/topics/sessions/timeouts.adoc
+++ b/docs/documentation/server_admin/topics/sessions/timeouts.adoc
@@ -35,11 +35,16 @@ image:images/sessions-tab.png[Sessions Tab]
 
 [[_offline-session-max-limited]]
 |Offline Session Max Limited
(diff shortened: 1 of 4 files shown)


stianst (changes requested):
I would skip backporting this to 26.0

PR #42553, opened 2025-09-11
fix: add missing NPM metadata

NyanKiyoshi:
This adds the missing `license` metadata, as well as other fields that were only present in [js/libs/keycloak-admin-client/package.json] (author and URLs).

This is problematic for license scanners as the projects were flagged as not having a license (e.g., https://www.npmjs.com/package/@keycloak/keycloak-ui-shared/v/26.3.3, we can see 'License' field having the value 'none' on the page).

Other fields were also added (author and URLs) for consistency with https://www.npmjs.com/package/@keycloak/keycloak-admin-client, however they aren't critical.

Packages that are private (`js/themes-vendor/package.json`, `js/package.json` and `js/apps/keycloak-server/package.json`) were also updated for consistency.

[js/libs/keycloak-admin-client/package.json]: https://github.com/keycloak/keycloak/blob/d98c474cdc68be28f67e8153c746bfb2d82f800a/js/libs/keycloak-admin-client/package.json#L57-L67

Closes: #42552


mposolda:
@ssilvert @jonkoops Do you please have a chance to review this?

ssilvert (approved):

ssilvert:
@NyanKiyoshi Thanks for the PR!

PR #42562, opened 2025-09-11
fix: adding supported database versions

### LINKED ISSUE #42294: Update database guide to include supported databases
### Description

The [database guide](https://www.keycloak.org/server/db) currently lists the tested versions of databases, but does not list the supported databases.

We should update the table to include a `Supported Versions` column, which lists all supported versions.

shawkins:
closes: #42294


vmuzikar (approved):
@shawkins LGTM, thanks.

PR #42568, opened 2025-09-11
Replace UUID with composite key for client session cache

pruivo:
Closes #42547

- [x] Code cleanup

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -263,8 +263,9 @@ In a future major release, the start-up will fail if default cache configuration
 
 === Simplified API for UserSessionProvider
 
-In order to retrieve a client session via `UserSessionProvider#getClientSession`, you no longer need to pass in the client session ID. The old methods have been deprecated and will be removed in a future release.
-Review also the other methods that are deprecated for removal in this class.
+In order to retrieve a client session via `UserSessionProvider#getClientSession`, you no longer need to pass in the client session ID.
+The old methods have been deprecated and will be removed in a future release.
+You should also review the other methods that are deprecated for removal in this class.
 
 // ------------------------ Removed features ------------------------ //
 == Removed features
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -111,6 +111,11 @@ configured directly by {project_name}:
 Configuring ports using the old
 properties has not changed, but using the CLI options is recommended because the previous method could be deprecated.
 
+=== Internal representation of client sessions changed
+
+The cache key of the authenticated client sessions has changed for embedded Infinispan, while the public APIs have not changed.
+Due to this, you should not run 26.4.x concurrently in a cluster with previous versions.
+
 === External IDP tokens automatically refreshed
 
 When using the `+/realms/{realm-name}/broker/{provider_alias}/token+` endpoint for an OAuth 2.0 IDP that provides refresh tokens and JSON responses or for OIDC IDPs, the tokens will be automatically refreshed each time they are retrieved via the endpoint if the access token has expired and the IDP provided a refresh token.
@@ -255,6 +260,11 @@ All {project_name} default cache configurations have been removed from `conf/cac
 Configuration of the default cache configurations in `conf/cache-ispn.xml`, or in a custom file by `--cache-config-file`, without specifying `--cache-config-mutate=true` is now deprecated and will log a warning.
 
 In a future major release, the start-up will fail if default cache configurations are stated in those files and the option is not specified.
+
+=== Simplified API for UserSessionProvider
+
+In order to retrieve a client session via `UserSessionProvider#getClientSession`, you no longer need to pass in the client session ID. The old methods have been deprecated and will be removed in a future release.
+Review also the other methods that are deprecated for removal in this class.
 
 // ------------------------ Removed features ------------------------ //
 == Removed features
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -267,6 +267,10 @@ In order to retrieve a client session via `UserSessionProvider#getClientSession`
 The old methods have been deprecated and will be removed in a future release.
 You should also review the other methods that are deprecated for removal in this class.
 
+=== Simplified API for AuthenticatedClientSessionModel
+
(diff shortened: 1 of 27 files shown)


ahus1:
@pruivo - thank you for this change. This is the next PR I want to merge. 

One thing I would like to discuss and consider: While we can with persistent sessions always forget items we have in memory as we can reload them from the database, there might be one pitfall as people will try to do rolling updates anyway: 

With a rolling updates, the `UserSessionEntity` will "loose" their entries of the client sessions, as there is now a new field. When then on the next call of `getClientSessions()` the list if empty. 

At the same time, the calls to `getAuthenticatedClientSessionByClient()` will not depend on the list, and load that item from the persister. 

So I wonder if one of the following steps could help:
* Keep the field in the `UserSessionEntity`, and when the new collection is empty and the old one exists, do a migration on read - we would then remove that migration code in a later release, for example KC 27
* Have a migration task that clears the user session cache - as this is done on the first node while old nodes are still running the old code, this might not be sufficient as some new entries could appear while the other nodes rolling toward the new version.

Please let me know your thoughts on this. Thanks!

pruivo:
@ahus1 bring this topic to Monday's meeting. The `UserSession` may be easy to handle, but the `ClientSession` is not; KC26.3 will throw a `ClassNotFoundException` due to the new cache key in 26.4 not being present in 26.3.

ahus1:
> The UserSession may be easy to handle, but the ClientSession is not; KC26.3 will throw a ClassNotFoundException due to the new cache key in 26.4 not being present in 26.3.

As we don't support running different versions concurrently, it is IMHO ok to throw an exception. Corrupted data is more difficult to find, and will have effects long after the migration. So this is fine to me. 

If we want to avoid the class cast exception, we could backport those new classes to the 26.3 branch and release a patch release before we release 26.4. Given our current upgrading guarantees, this is very much extra and icing on the cake, still some users will love it. It could also be an experiment on how we want to handle things like this in the future.

ahus1 (dismissed):

ahus1 (dismissed):

pruivo:
@ahus1 I disabled auto-merge as I want to annotate the userSessionId field in the client session entity.

docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
```diff
@@ -256,6 +261,11 @@ Configuration of the default cache configurations in `conf/cache-ispn.xml`, or i
 
 In a future major release, the start-up will fail if default cache configurations are stated in those files and the option is not specified.
 
+=== Simplified API for UserSessionProvider
+
+In order to retrieve a client session via `UserSessionProvider#getClientSession`, you no longer need to pass in the client session ID. The old methods have been deprecated and will be removed in a future release.
+Review also the other methods that are deprecated for removal in this class.
```
ryanemerson: "You should also review the other ..."

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java
```diff
@@ -208,6 +209,25 @@ public void migrate(String modelVersion) {
         if ("26.0.0".equals(modelVersion)) {
             log.debug("Clear caches to migrate to Infinispan Protostream");
             CompletionStages.join(session.getProvider(InfinispanConnectionProvider.class).migrateToProtoStream());
+        } else if ("26.4.0".equals(modelVersion)) {
+            log.debug("Clear caches as client session entries are now outdated and are not migrated");
+            // This is a best-effort approach: Even if due to a rolling update some entries are left there, the checking of sessions and tokens does not depend on them.
+            // Refreshing of tokens will still work even if the user session does not contain the list of client sessions.
+            // This still keeps the user session cache to keep users logged in on a best effort basis. Only the offline user sessions cache is cleared.
```
ryanemerson: Is "Only the offline user sessions cache is cleared." outdated, or am I missing something?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/entities/AuthenticatedClientSessionEntity.java
```diff
@@ -54,12 +59,11 @@ public class AuthenticatedClientSessionEntity extends SessionEntity {
 
     private Map<String, String> notes = new ConcurrentHashMap<>();
 
-    private final UUID id;
-
+    // TODO [pruivo] [KC27] make these fields final. They are the client session identity.
```
ryanemerson, line 62: Can you create an issue or issues to track this and the other TODO comments for the 27 milestone?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/UserSessionAdapter.java
```diff
@@ -84,30 +81,28 @@ public UserSessionAdapter(KeycloakSession session, UserModel user, T provider,
...
+            if (client != null) {
+                final AuthenticatedClientSessionModel clientSession = provider.getClientSession(this, client, offline);
+                if (clientSession != null) {
+                    result.put(clientUUID, clientSession);
                 } else {
-                    // client does no longer exist
-                    removedClientUUIDS.add(key);
+                    // Either the client session has expired, or it hasn't been added by a concurrently running login yet.
+                    // So it is unsafe to clear it, so we need to keep it for now. Otherwise, the test ConcurrentLoginTest.concurrentLoginSingleUser will fail.
+                    // removedClientUUIDS.add(key);
```
ryanemerson: Remove this line and keep the other comments?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProvider.java
ahus1: It is "Only the offline user sessions cache is cleared, but not the regular user sessions cache is cleared. All client session caches regular and offline client sessions are cleared as usual" ... yes, that's even longer.

ahus1:
BTW, account UI is unstable, see https://github.com/keycloak/keycloak/pull/42657

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/entities/AuthenticatedClientSessionEntity.java
pruivo: https://github.com/keycloak/keycloak/issues/42668

ahus1 (approved):
Thank you for this PR!

ahus1:
@keycloak/core-clients: FYI, the authenticated client sessions no longer have their own UUID, but their key in ISPN is now a composite key of the user session ID and the client ID. This is now aligning how things are handled in the offline_* tables. 
This change is part of KC 26.4

PR #42569, opened 2025-09-11
More efficient secure ID generator

pruivo:
Closes #42283

**`ThreadLocal` Removal**
Removed the `ThreadLocal` mainly for 2 reasons: 
* With `VirtualThread`, the `ThreadLocal` is created and garbage collected immediately as `VirtualThread`s are never cached in a pool.
* The `SecureRandom` is thread-safe.

**Performance**

Micro-benchmark with `VM version: JDK 21.0.7, OpenJDK 64-Bit Server VM, 21.0.7+6-LTS`. The benchmark code can be found here: https://github.com/pruivo/playground/tree/master/uuid-generator

| Method generateSecureID | Time ns/op | Memory byte/op |
|--------|--------|--------|
| Original | 322.014 | 376.000 |
| This PR| 177.973 | 176.000 |

| Method generateSecureUUID | Time ns/op | Memory byte/op |
|--------|--------|--------|
| Original | 280.327 | 256.000 |
| This PR| 158.623 | 128.000 |

### DIFF
--- a/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
+++ b/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
@@ -1,7 +1,6 @@
 package org.keycloak.common.util;
 
 import java.security.SecureRandom;
-import java.util.Random;
 import java.util.UUID;
 
 public class SecretGenerator {
@@ -18,12 +17,7 @@ public class SecretGenerator {
 
     private static final SecretGenerator instance = new SecretGenerator();
 
-    private ThreadLocal<Random> random = new ThreadLocal<Random>() {
-        @Override
-        protected Random initialValue() {
-            return new SecureRandom();
-        }
-    };
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
 
     private SecretGenerator() {
     }
@@ -33,12 +27,7 @@ public class SecretGenerator {
     }
     
     public String generateSecureID() {
-        StringBuilder builder = new StringBuilder(instance.randomBytesHex(16));
-        builder.insert(8, '-');
-        builder.insert(13, '-');
-        builder.insert(18, '-');
-        builder.insert(23, '-');
-        return builder.toString();
+        return generateSecureUUID().toString();
     }
 
     public String randomString() {
@@ -57,11 +46,10 @@ public class SecretGenerator {
             throw new IllegalArgumentException();
         }
 
-        Random r = random.get();
         char[] buf = new char[length];
 
         for (int idx = 0; idx < buf.length; ++idx) {
-            buf[idx] = symbols[r.nextInt(symbols.length)];
+            buf[idx] = symbols[SECURE_RANDOM.nextInt(symbols.length)];
         }
 
         return new String(buf);
@@ -77,7 +65,7 @@ public class SecretGenerator {
         }
 
         byte[] buf = new byte[length];
-        random.get().nextBytes(buf);
+        SECURE_RANDOM.nextBytes(buf);
         return buf;
     }
 
@@ -121,6 +109,18 @@ public class SecretGenerator {
      * @return UUID with all bits random
      */
     public UUID generateSecureUUID() {
-        return new UUID(random.get().nextLong(), random.get().nextLong());
+        byte[] data = randomBytes(16);
+        return new UUID(toLong(data, 0), toLong(data, 8));
+    }
+
+    private static long toLong(byte[] data, int offset) {
+        return  ((data[offset] & 0xFFL) << 56) |
+                ((data[offset + 1] & 0xFFL) << 48) |
+                ((data[offset + 2] & 0xFFL) << 40) |
+                ((data[offset + 3] & 0xFFL) << 32) |
+                ((data[offset + 4] & 0xFFL) << 24) |
+                ((data[offset + 5] & 0xFFL) << 16) |
+                ((data[offset + 6] & 0xFFL) <<  8) |
+                ((data[offset + 7] & 0xFFL)) ;
     }
 }


common/src/main/java/org/keycloak/common/util/SecretGenerator.java
```diff
@@ -18,12 +17,7 @@ public class SecretGenerator {
 
     private static final SecretGenerator instance = new SecretGenerator();
 
-    private ThreadLocal<Random> random = new ThreadLocal<Random>() {
-        @Override
-        protected Random initialValue() {
-            return new SecureRandom();
-        }
-    };
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
```
ahus1, line 20: While SecureRandom() is thread safe, it will still synchronize whenever nextBytes() are called. So I'd suggest that wrapper to avoid locking in multithreaded environments.

ahus1 (changes requested):
Thank you for the PR. See below for one consideration for multi-threaded environments.

common/src/main/java/org/keycloak/common/util/SecretGenerator.java
ahus1: Reading your statement about virtual threads, I suppose we're not using virtual threads yet. Can we have a switch so that we use a thread local instance when running within regular threads, and the static final instance when using virtual threads? 

Feel free to convince me to proceed as is - while the locking might bring some overhead as a lot of different threads access it, I don't expect to get any congestion on this lock.

ahus1 (approved):
I'm reconsidering my previous comment as Java's UUID is also handling it the same way without ThreadLocals. So I'm approving and merging.

common/src/main/java/org/keycloak/common/util/SecretGenerator.java
pruivo: With Adoptium JDK builds, all `SecureRandom` instances share the same engine and synchronize internally between threads. See below, line 83 here: https://github.com/pruivo/playground/blob/master/uuid-generator/src/main/java/me/pruivo/UUIDGeneratorBenchmark.java#L83

```
"me.pruivo.UUIDGeneratorBenchmark.stringOriginal-jmh-worker-12" #38 [607935] daemon prio=5 os_prio=0 cpu=1008,91ms elapsed=14,35s tid=0x00007fb1f03c79f0 nid=607935 waiting for monitor entry  [0x00007fb1b5cfb000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at sun.security.provider.SecureRandom.engineNextBytes(java.base@24.0.1/SecureRandom.java:222)
        - waiting to lock <0x000000041419f550> (a sun.security.provider.SecureRandom)
        at sun.security.provider.NativePRNG$RandomIO.implNextBytes(java.base@24.0.1/NativePRNG.java:530)
        at sun.security.provider.NativePRNG.engineNextBytes(java.base@24.0.1/NativePRNG.java:217)
        at java.security.SecureRandom.nextBytes(java.base@24.0.1/SecureRandom.java:768)
        at me.pruivo.UUIDGeneratorBenchmark.randomBytes(UUIDGeneratorBenchmark.java:83)
        
"me.pruivo.UUIDGeneratorBenchmark.stringOriginal-jmh-worker-20" #46 [607943] daemon prio=5 os_prio=0 cpu=1009,52ms elapsed=14,35s tid=0x00007fb1f03d2df0 nid=607943 waiting for monitor entry  [0x00007fb1b54f3000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at sun.security.provider.SecureRandom.engineNextBytes(java.base@24.0.1/SecureRandom.java:222)
        - waiting to lock <0x000000041419f550> (a sun.security.provider.SecureRandom)
        at sun.security.provider.NativePRNG$RandomIO.implNextBytes(java.base@24.0.1/NativePRNG.java:530)
        at sun.security.provider.NativePRNG.engineNextBytes(java.base@24.0.1/NativePRNG.java:217)
        at java.security.SecureRandom.nextBytes(java.base@24.0.1/SecureRandom.java:768)
        at me.pruivo.UUIDGeneratorBenchmark.randomBytes(UUIDGeneratorBenchmark.java:83)
```

I've just tested with 50 concurrent threads. TLDR: This PR is still faster and more memory efficient.
Raw results:
```
Benchmark                                                 Mode  Cnt      Score     Error   Units
UUIDGeneratorBenchmark.stringJdk8                         avgt   10  15452.947 ±  27.873   ns/op
UUIDGeneratorBenchmark.stringJdk8:gc.alloc.rate           avgt   10    543.190 ±   0.970  MB/sec
UUIDGeneratorBenchmark.stringJdk8:gc.alloc.rate.norm      avgt   10    176.004 ±   0.001    B/op
UUIDGeneratorBenchmark.stringJdk8:gc.count                avgt   10     93.000            counts
UUIDGeneratorBenchmark.stringJdk8:gc.time                 avgt   10     52.000                ms
UUIDGeneratorBenchmark.stringOriginal                     avgt   10  16834.802 ±   9.912   ns/op
UUIDGeneratorBenchmark.stringOriginal:gc.alloc.rate       avgt   10   1133.058 ±   0.665  MB/sec
UUIDGeneratorBenchmark.stringOriginal:gc.alloc.rate.norm  avgt   10    400.001 ±   0.001    B/op
UUIDGeneratorBenchmark.stringOriginal:gc.count            avgt   10    194.000            counts
UUIDGeneratorBenchmark.stringOriginal:gc.time             avgt   10    104.000                ms
UUIDGeneratorBenchmark.uuidJdk8                           avgt   10  15406.299 ±  23.514   ns/op
UUIDGeneratorBenchmark.uuidJdk8:gc.alloc.rate             avgt   10    396.216 ±   0.608  MB/sec
UUIDGeneratorBenchmark.uuidJdk8:gc.alloc.rate.norm        avgt   10    128.001 ±   0.001    B/op
UUIDGeneratorBenchmark.uuidJdk8:gc.count                  avgt   10     67.000            counts
UUIDGeneratorBenchmark.uuidJdk8:gc.time                   avgt   10     39.000                ms
UUIDGeneratorBenchmark.uuidNextLong                       avgt   10  38897.560 ± 235.408   ns/op
UUIDGeneratorBenchmark.uuidNextLong:gc.alloc.rate         avgt   10    313.890 ±   1.787  MB/sec
UUIDGeneratorBenchmark.uuidNextLong:gc.alloc.rate.norm    avgt   10    256.003 ±   0.001    B/op
UUIDGeneratorBenchmark.uuidNextLong:gc.count              avgt   10     54.000            counts
UUIDGeneratorBenchmark.uuidNextLong:gc.time               avgt   10     30.000                ms
```

pruivo: On a side note, I plan to inject virtual threads in the `InfinispanKeyGenerator` :)

PR #42573, opened 2025-09-12
fix: cleaning up dockerfile build warnings

shawkins:
closes: #42572


mabartos (approved):
@shawkins LGTM.

@ASzc Do you want to apply these changes even on the prod Dockerfiles (if those are affected as well)?

ASzc:
Hard to say if these warnings effect prod too. I don't see any specifics in the issue description. I have sync'd the changes anyway

PR #42580, opened 2025-09-12
[OID4VCI] rename keyattestation+jwt to key-attestation+jwt

### LINKED ISSUE #41579: Verify that the wallet correctly constructs key attestation JWTs with required JOSE headers and JWT body parameters
### Description

- Update key attestation header typ from keyattestation+jwt to key-attestation+jwt

Ogenbertrand:
This PR simply aims at verifying that the we correctly constructs key attestation JWTs with required JOSE headers and JWT body parameters.

closes: https://github.com/keycloak/keycloak/issues/41579

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/keybinding/AttestationValidatorUtil.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/keybinding/AttestationValidatorUtil.java
@@ -76,7 +76,7 @@ import static org.keycloak.services.clientpolicy.executor.FapiConstant.ALLOWED_A
  */
 public class AttestationValidatorUtil {
 
-    public static final String ATTESTATION_JWT_TYP = "keyattestation+jwt";
+    public static final String ATTESTATION_JWT_TYP = "key-attestation+jwt ";
     private static final String CACERTS_PATH = System.getProperty("javax.net.ssl.trustStore",
             System.getProperty("java.home") + "/lib/security/cacerts");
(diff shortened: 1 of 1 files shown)


tnorimat (approved):
LGTM by follwoing https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#appendix-K-2 .

mposolda (approved):
@Ogenbertrand @tnorimat Thanks!

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/keybinding/AttestationValidatorUtil.java
```diff
@@ -76,7 +76,7 @@
  */
 public class AttestationValidatorUtil {
 
-    public static final String ATTESTATION_JWT_TYP = "keyattestation+jwt";
+    public static final String ATTESTATION_JWT_TYP = "key-attestation+jwt ";
```
jonas-youniqx, line 79: I think the whitespace at the end was added by accident.

Ogenbertrand: Hello @jonas-youniqx !
Thanks for pointing that out, we'll send a PR with the fix :+1:

Awambeng: > I think the whitespace at the end was added by accident.

@jonas-youniqx Thank you for pointing this out. It has been fixed in this PR: https://github.com/keycloak/keycloak/pull/43951

PR #42596, opened 2025-09-13
Keycloak 26.4 Upgrading Guide

andymunro:
Closes #42564


ahus1 (approved):
Thank you for these changes!

One item that I rolled back was the statement about breaking changes, and what we consider breaking changes. It now reads:

> Breaking changes are identified as those that might require changes for existing users to their configurations or applications.
In minor or patch releases, {project_name} will only introduce breaking changes to fix bugs.

We can discuss this more in a 1:1 - I'll still merge this PR to avoid conflicts with other changes coming in.

andymunro:
Thanks, @ahus1.  Looks good.

PR #42607, opened 2025-09-15
Do not regenerate the secret key when the size is not explicitly passed

rmartinc:
Closes #42405

The PR just checks if the `secretSize` config option is explicitly passed to regenerate the key, if not the key is maintained. It also adds the model config when missing and no the default size. Tests added.

The other option is just adding a migration task to add the config option to the keys that does not have the default size (previous default) and the label is not there. But I decided to do it this way as the upgrade task is more difficult to migrate. But if you prefer this way I can change it.


mposolda (approved):
@rmartinc Thanks!

PR #42615, opened 2025-09-15
Fix for missing object representation in admin event log when deleting user, group, client

### LINKED ISSUE #33009: Delete Client and role's admin event doesn't have representation while other delete entities have the representation

JakubWoo:
Fix for missing object representation in admin event log when deleting user, group, client

Closes #33009

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/ClientResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/ClientResource.java
@@ -250,6 +250,10 @@ public class ClientResource {
 
         AdminPermissionsSchema.SCHEMA.throwExceptionIfAdminPermissionClient(session, client.getId());
 
+        ClientScopeRepresentation clientRepresentation = new ClientScopeRepresentation();
+        clientRepresentation.setId(client.getId());
+        clientRepresentation.setName(client.getName());
+
         try {
             session.clientPolicy().triggerOnEvent(new AdminClientUnregisterContext(client, auth.adminAuth()));
         } catch (ClientPolicyException cpe) {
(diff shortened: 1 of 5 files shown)


JakubWoo:
Hi @pedroigor @stianst, 
kindly asking for a review.
Thanks

services/src/main/java/org/keycloak/services/resources/admin/ClientResource.java
```diff
@@ -250,14 +250,18 @@ public void deleteClient() {
 
         AdminPermissionsSchema.SCHEMA.throwExceptionIfAdminPermissionClient(session, client.getId());
 
+        ClientScopeRepresentation clientRepresentation = new ClientScopeRepresentation();
+        clientRepresentation.setId(client.getId());
+        clientRepresentation.setName(client.getName());
```
martin-kanis, line 255: Here we should store `clientId` not `name`

martin-kanis (changes requested):
@JakubWoo Thanks, I added one comment about clients.

Also, I think we should do the same for roles to be consistent. Here https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/services/resources/admin/RoleContainerResource.java#L275

martin-kanis:
Closing in favour of https://github.com/keycloak/keycloak/pull/43620

PR #42616, opened 2025-09-15
[Test Framework] Ability to run Keycloak test server with HTTPS

### LINKED ISSUE #34486: Ability to run Keycloak test server with HTTPS

vaceksimon:
Closes #34486

Allows running the Keycloak test server with HTTPS by calling `tlsEnabled(true)` on `KeycloakServerConfigBuilder`.

Doing so will create an instance of `CertificatesSupplier` and `ManagedCertificates`. The latter class is responsible for creating a self-signed certificate for the testing server and storing it in a server keystore and clients' truststore.

The keystore and truststore with their contents are generated and stored in a temp directory (`System.getProperty("java.io.tmpdir")`). If there are these two keystores on this path, they are loaded into memory instead of creating new ones. If either one is missing, both are created again and stored in the temp dir.

This truststore is used in clients (`HttpClient`, `OAuthClient`, `AdminClient`) to establish trust with the Keycloak testing server.


test-framework/core/src/main/java/org/keycloak/testframework/http/HttpClientSupplier.java
```diff
@@ -1,22 +1,53 @@
...
     public HttpClient getValue(InstanceContext<HttpClient, InjectHttpClient> instanceContext) {
         HttpClientBuilder builder = HttpClientBuilder.create();
 
+        KeycloakServer server = instanceContext.getDependency(KeycloakServer.class);
+        if (server.isTlsEnabled()) {
+            ManagedCertificate managedCert = instanceContext.getDependency(ManagedCertificate.class);
+            try {
+                KeyStore serverKeyStore = managedCert.getKeyStore();
+                SSLContext sslContext = SSLContextBuilder.create()
+                        .loadTrustMaterial(serverKeyStore, new TrustSelfSignedStrategy())
```
stianst: If I'm not mistaken `TrustSelfSignedStrategy` just skips the truststore for self-signed certificates, so don't think this should be enabled?

test-framework/core/src/main/java/org/keycloak/testframework/http/HttpClientSupplier.java
```diff
@@ -1,22 +1,53 @@
...
+                        .build();
+
+                SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
+                        sslContext,
+                        NoopHostnameVerifier.INSTANCE);
```
stianst: We shouldn't add `NoopHostnameVerifier` as that disables check that the cert actually belongs to the correct node.

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificate.java
```diff
@@ -0,0 +1,110 @@
...
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
+public class ManagedCertificate {
```
stianst: I'd rename this to certificates, as we probably want to expand on this in the future to also support generating client certificates for mTLS use-cases
```suggestion
public class ManagedCertificates {
```

test-framework/core/src/main/java/org/keycloak/testframework/https/CertificateSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+import org.keycloak.testframework.injection.RequestedInstance;
+import org.keycloak.testframework.injection.Supplier;
+import org.keycloak.testframework.injection.SupplierOrder;
+
+public class CertificateSupplier implements Supplier<ManagedCertificate, InjectCertificate> {
+
+    @Override
+    public ManagedCertificate getValue(InstanceContext<ManagedCertificate, InjectCertificate> instanceContext) {
+        CryptoProvider cryptoProvider = new DefaultCryptoProvider();
+        CryptoIntegration.setProvider(cryptoProvider);
```
stianst: Do this instead:
https://github.com/keycloak/keycloak/blob/8ad6427123d9919c1fadcbcf6f0090cdf26d3cb8/test-framework/oauth/src/main/java/org/keycloak/testframework/oauth/OAuthIdentityProvider.java#L36-L37

That makes sure it's not initialised multiple times

stianst: Might also just move that to `ManagedCertificate` since that's where the logic to use it is

test-framework/core/src/main/java/org/keycloak/testframework/https/InjectCertificate.java
```diff
@@ -0,0 +1,11 @@
+package org.keycloak.testframework.https;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface InjectCertificate {
```
stianst:
```suggestion
public @interface InjectCertificates {
```

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
```diff
@@ -99,13 +100,14 @@ public final class RawKeycloakDistribution implements KeycloakDistribution {
     private final OutputConsumer outputConsumer;
 
     public RawKeycloakDistribution(boolean debug, boolean manualStop, boolean enableTls, boolean reCreate, boolean removeBuildOptionsAfterBuild, int requestPort) {
-        this(debug, manualStop, enableTls, reCreate, removeBuildOptionsAfterBuild, requestPort, new DefaultOutputConsumer());
+        this(debug, manualStop, enableTls, null, reCreate, removeBuildOptionsAfterBuild, requestPort, new DefaultOutputConsumer());
     }
 
-    public RawKeycloakDistribution(boolean debug, boolean manualStop, boolean enableTls, boolean reCreate, boolean removeBuildOptionsAfterBuild, int requestPort, OutputConsumer outputConsumer) {
+    public RawKeycloakDistribution(boolean debug, boolean manualStop, boolean enableTls, Path serverKeyStorePath, boolean reCreate, boolean removeBuildOptionsAfterBuild, int requestPort, OutputConsumer outputConsumer) {
```
stianst: Instead of copying the truststore, wouldn't it be simpler to just set `https-key-store-file` config option?

test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java
```diff
@@ -55,9 +58,15 @@ public KeycloakServer getValue(InstanceContext<KeycloakServer, KeycloakIntegrati
             getLogger().debugv("Startup command and options: \n\t{0}", String.join("\n\t", command.toArgs()));
         }
 
+        Path serverKeyStore = null;
+        if (command.tlsEnabled()) {
+            ManagedCertificate managedCert = instanceContext.getDependency(ManagedCertificate.class);
+            serverKeyStore = managedCert.getKeycloakServerKeyStorePath();
```
stianst: Can't we just do something like:
```
command.option("https-key-store-file", managedCert.getKeycloakServerKeyStorePath()...)
```

test-framework/clustering/src/main/java/org/keycloak/testframework/server/ClusteredKeycloakServerSupplier.java
```diff
@@ -14,7 +16,8 @@ public class ClusteredKeycloakServerSupplier extends AbstractKeycloakServerSuppl
     String images = ClusteredKeycloakServer.SNAPSHOT_IMAGE;
 
     @Override
-    public KeycloakServer getServer() {
+    public KeycloakServer getServer(Path serverKeyStorePath) {
+        // todo serverKeyStorePath is used for TLS; implement TLS in ClusteredKeycloakServer
```
lhanusov: @vaceksimon what should be done with this comment, issue?

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
```diff
@@ -54,12 +57,25 @@ public void stop() {
...
+        } else {
+            return "http://localhost:8080";
+        }
     }
 
     @Override
     public String getManagementBaseUrl() {
-        return "http://localhost:9000";
+        if (isTlsEnabled()) {
+            return "https://localhost:9000";
```
lhanusov, line 68: @vaceksimon are you sure the port is the same?

lhanusov (changes requested):
please, check also the management ports for embedded and remote servers

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
vaceksimon: It should be, [according to the docs](https://www.keycloak.org/server/management-interface#_tls_support)

test-framework/clustering/src/main/java/org/keycloak/testframework/server/ClusteredKeycloakServerSupplier.java
vaceksimon: In the latest commit, I changed how the keystore file is obtained, so this has been removed. Anyway, the Clustered Keycloak server will probably not work with https at this moment. I would do that in the future, and not as part of this PR

vaceksimon:
@lhanusov @stianst all your reviews should be adressed in the latest commit

lhanusov (approved):

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
stianst, line 1: You should have two separate KeyStores:

* One contains both private and public keys, you should refer to this as the Keycloak Server KeyStore
* Another containing only the public keys, you should refer to this as the Clients TrustStore

test-framework/core/src/main/java/org/keycloak/testframework/server/RemoteKeycloakServer.java
stianst, line 1: RemoteKeycloakServer is not working, which is because `verifyRunningKeycloak` is using `getBaseUrl()` to connect to Keycloak, but is not configured with a truststore.

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactory.java
```diff
@@ -17,6 +24,21 @@ public class AdminClientFactory {
         delegateSupplier = () -> KeycloakBuilder.builder().serverUrl(serverUrl);
     }
 
+    AdminClientFactory(String serverUrl, KeyStore serverKeyStore) {
+        try {
+            SSLContext sslContext = SSLContextBuilder.create()
+                    .loadTrustMaterial(serverKeyStore, null)
+                    .build();
```
stianst: I'd move this to `ManagedCertificates#getClientSSLContext`

test-framework/core/src/main/java/org/keycloak/testframework/http/HttpClientSupplier.java
```diff
@@ -1,22 +1,52 @@
...
         HttpClientBuilder builder = HttpClientBuilder.create();
 
+        KeycloakServer server = instanceContext.getDependency(KeycloakServer.class);
+        if (server.isTlsEnabled()) {
+            ManagedCertificates managedCert = instanceContext.getDependency(ManagedCertificates.class);
+            try {
+                KeyStore serverKeyStore = managedCert.getKeyStore();
+                SSLContext sslContext = SSLContextBuilder.create()
+                        .loadTrustMaterial(serverKeyStore, null)
+                        .build();
```
stianst: I'd move this to `ManagedCertificates#getClientSSLContext`

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -0,0 +1,117 @@
...
+        initKeyStore();
+    }
+
+    public Path getKeycloakServerKeyStorePath() {
+        return KEYSTORE_FILE_PATH;
+    }
+
+    public KeyStore getKeyStore() {
+        return keyStore;
+    }
```
stianst: This should not exist as we don't need access to the KeyStore directly, what we need is `public KeyStore getClientTrustStore()`.

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactory.java
```diff
@@ -17,6 +24,21 @@ public class AdminClientFactory {
...
+
+            delegateSupplier = () ->
+                    KeycloakBuilder.builder()
+                            .serverUrl(serverUrl)
+                            .resteasyClient(Keycloak.getClientProvider().newRestEasyClient(null, sslContext, false));
+        } catch (NoSuchAlgorithmException | KeyStoreException | KeyManagementException e) {
+            throw new ManagedCertificatesException(e);
+        }
```
stianst: You don't need this once you add `ManagedCertificates#getClientSSLContext` as that can then just catch the exceptions and throw the runtime `ManagedCertificatesException`

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactorySupplier.java
```diff
@@ -1,17 +1,27 @@
...
         KeycloakServer server = instanceContext.getDependency(KeycloakServer.class);
-        return new AdminClientFactory(server.getBaseUrl());
+
+        if (!server.isTlsEnabled()) {
+            return new AdminClientFactory(server.getBaseUrl());
+        }
+
+        ManagedCertificates managedCert = instanceContext.getDependency(ManagedCertificates.class);
+        KeyStore serverKeyStore = managedCert.getKeyStore();
+        return new AdminClientFactory(server.getBaseUrl(), serverKeyStore);
```
stianst: Nitpick, but I find it cleaner to read with this:
```suggestion
        } else {
            ManagedCertificates managedCert = instanceContext.getDependency(ManagedCertificates.class);
            KeyStore serverKeyStore = managedCert.getKeyStore();
            return new AdminClientFactory(server.getBaseUrl(), serverKeyStore);
       }
```

test-framework/core/src/main/java/org/keycloak/testframework/http/HttpClientSupplier.java
```diff
@@ -1,22 +1,52 @@
...
+
+                SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
+                        sslContext,
+                        SSLConnectionSocketFactory.getDefaultHostnameVerifier()
+                );
+
+                builder.setSSLSocketFactory(sslSocketFactory);
+            } catch (NoSuchAlgorithmException | KeyStoreException | KeyManagementException e) {
+                throw new ManagedCertificatesException(e);
+            }
```
stianst: Not needed after you've done the above

test-framework/examples/tests/src/test/java/org/keycloak/test/examples/TlsEnabledTest.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Test
+    public void testAdminClient() {
+        adminClient.realm("default");
+    }
+
+    @Test
+    public void testOAuthClient() {
+        oAuthClient.doWellKnownRequest();
```
stianst: I'd add an assert that checks one of the endpoints return `https`. For example:
```suggestion
        Assertions.assertTrue(oAuthClient.doWellKnownRequest().getTokenEndpoint().startsWith("https://"));
```

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public final static String CERT_ENTRY = "cert";
+    public final static String PRV_KEY_ENTRY = "prvKey";
+
+
+    public ManagedCertificates() throws ManagedCertificatesException {
+        cryptoProvider = new DefaultCryptoProvider();
+        if (!CryptoIntegration.isInitialised()) {
+            CryptoIntegration.setProvider(cryptoProvider);
+        }
```
stianst:
```suggestion
        if (!CryptoIntegration.isInitialised()) {
            CryptoIntegration.setProvider(new DefaultCryptoProvider());
        }
        cryptoProvider = CryptoIntegration.getProvider();
```

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public Path getKeycloakServerKeyStorePath() {
+        return KEYSTORE_FILE_PATH;
+    }
+
+    public KeyStore getKeyStore() {
+        return keyStore;
+    }
+
+    public X509Certificate getCertificate() {
```
stianst:
```suggestion
    public X509Certificate getKeycloakServerCertificate() {
```

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
stianst: For now this is more just a best practice, but we need to extend quite a bit on this in the future to support mTLS tests; in which case there will be a clear distinction between what certs the server has an which the client has.

test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java
```diff
@@ -55,6 +58,11 @@ public KeycloakServer getValue(InstanceContext<KeycloakServer, KeycloakIntegrati
             getLogger().debugv("Startup command and options: \n\t{0}", String.join("\n\t", command.toArgs()));
         }
 
+        if (command.tlsEnabled()) {
+            ManagedCertificates managedCert = instanceContext.getDependency(ManagedCertificates.class);
+            command.option("https-key-store-file", managedCert.getKeycloakServerKeyStorePath().toString());
```
stianst: Just make ManagedCertificates return a string instead of a Path, since that's what we need.

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServerSupplier.java
```diff
@@ -3,6 +3,7 @@
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 
+
```
stianst:
```suggestion
```

stianst (commented):
Nice work, this is getting pretty close now :)

stianst (approved):

PR #42617, opened 2025-09-15
Validate wait_timeout parameter on MySQL and MariaDB

### LINKED ISSUE #42300: Validate wait_timeout parameter on MySQL and MariaDB
### Description

When working with MySQL and MariaDB, the database has a configuration value `wait_timeout` where the database terminates the connection if no statement was received for the given amount of time. 

With MySQL it is essential to enable periodic connection validation on the pool side(Agroal) to ensure stability (as described from Red Hat support). 

### Discussion

_No response_

### Motivation

Keycloak should provide all necessary options and a safe way to configure it for MySQL and MariaDB

### Details

The following settings come into play, and need to be validated against the `wait_timeout` of the database. While the default is `28800` seconds (8 hours), the value can and will be configured differently for different instances. 

IMHO either max lifetime, or a background validation interval with a validation query needs to be configured. 

As of today, those values are not validated against the database, and there are no regular CLI options or documentation in Keycloak for this. 

```
name: QUARKUS_DATASOURCE_JDBC_MAX_LIFETIME
  value: '30000'  # <=  # Must be less than MySQL's wait_timeout
```

```
name: QUARKUS_DATASOURCE_JDBC_BACKGROUND_VALIDATION_INTERVAL
  value: '30S'
name: QUARKUS_DATASOURCE_JDBC_VALIDATION_QUERY_SQL
  value: 'SELECT 1'
```

ryanemerson:
Closes #42300

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -49,6 +49,17 @@ Starting with this release, and when using the original PostgreSQL driver, {proj
 
 You can override this behavior by setting your own value for `targetServerType` in the DB URL or additional properties.
 
+=== MySQL and MariaDB wait_timeout validation
+
+In order to prevent connections being closed unexpectedly by the database server, it is necessary to ensure that the {project_name}
+connection pool is correctly configured with respect to the server's `wait_timeout` setting.
+
+Starting with this release, {project_name} defines a default `db-pool-max-lifetime` of 7 hours and 50 minutes for MySQL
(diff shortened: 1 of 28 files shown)


ahus1:
@ryanemerson - I've added some commits, please review and let me know your thoughts. Thanks!

ryanemerson:
Thanks for the commits @ahus1. I have pushed two additional fixes.

ahus1 (approved):

ahus1:
@ryanemerson - please create a follow-up issue for KC 27.0 that KC will fail on startup instead of issuing a warning.

ryanemerson:
Done: #42640

PR #42621, opened 2025-09-15
fix: updating export option text

### LINKED ISSUE #42011: users-per-file option should no longer mention exponential runtime increases
### Description

users-per-file option still have the legacy text

shawkins:
closes: #42011

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/ExportOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/ExportOptions.java
@@ -21,7 +21,7 @@ public class ExportOptions {
 
     public static final Option<String> FILE = new OptionBuilder<>("file", String.class)
             .category(OptionCategory.EXPORT)
-            .description("Set the path to a file that will be created with the exported data. To export more than 500 users, export to a directory with different files instead.")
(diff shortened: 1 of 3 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/ExportOptions.java
```diff
@@ -21,7 +21,7 @@ public class ExportOptions {
 
     public static final Option<String> FILE = new OptionBuilder<>("file", String.class)
             .category(OptionCategory.EXPORT)
-            .description("Set the path to a file that will be created with the exported data. To export more than 500 users, export to a directory with different files instead.")
+            .description("Set the path to a file that will be created with the exported data. To export more than 50000 users, export to a directory with different files instead.")
```
mabartos, line 24: Are we confident to claim/recommend the 50k users?

mabartos (approved):
@shawkins Thanks. LGTM, just one question.

quarkus/config-api/src/main/java/org/keycloak/config/ExportOptions.java
shawkins: That matches a change made to the main docs.

mabartos: ok, thanks

Issue #41586, opened 2025-07-31
Update the authorization_details in the Authorization Request logic


Ogenbertrand:
### Description

See changelog:

- Modify Authorization Request logic to use only credential_configuration_id in authorization_details, removing support for the format parameter.

- Implement support for the optional claims parameter in authorization_details, as defined in Appendix B.1, for specifying claims description objects.

> type: REQUIRED. String that determines the authorization details type. It MUST be set to openid_credential for the purpose of this specification.
> credential_configuration_id: REQUIRED. String specifying a unique identifier of the Credential being described in the credential_configurations_supported map in the Credential Issuer Metadata as defined in [Section 11.2.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#credential-issuer-parameters). The referenced object in the credential_configurations_supported map conveys the details of the requested Credential, such as the format and format-specific parameters like vct for SD-JWT VC or doctype for ISO mdoc. This specification defines those Credential Format specific Issuer Metadata in [Appendix A](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#format-profiles).
> claims: OPTIONAL. A non-empty array of claims description objects as defined in [Appendix B.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#claims-description-authorization-details).

forkimenjeckayang:
Will be working on this

PR #42622, opened 2025-09-15
[OID4VC]: Update authorization_details for OID4VCI draft-16 compliance

forkimenjeckayang:
## Summary
Updates `authorization_details` processing to comply with OID4VCI draft-16 specification by removing the deprecated `format` parameter and implementing the new `claims` parameter for claims description objects.

### **Removed**
- `format` parameter from `authorization_details` processing
- All format-based validation and test scenarios

###  **Added**
- `claims` parameter support for claims description objects
- `ClaimsDescription` model and `ClaimsPathPointer` utility
- Comprehensive claims validation (path pointers, conflicts, etc.)

### **Enhanced**
- Validation logic to prevent empty `authorization_details` arrays
- Error handling with descriptive messages
- Test coverage for all new scenarios

## Closses
- Ticket : https://github.com/keycloak/keycloak/issues/41586


services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+
+    private static final Logger logger = Logger.getLogger(ClaimsPathPointer.class);
+
+    /**
+     * Validates a claims path pointer.
+     *
+     * @param path the claims path pointer to validate
+     * @return true if valid, false otherwise
+     */
+    public static boolean isValidPath(List<Object> path) {
```
tnorimat, line 45: It seems that this method is not used. Is it really needed?

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+            // If components are different types and one is integer (specific array index) and the other is null (all array elements)
+            if (comp1 == null && comp2 instanceof Integer) {
+                return true;
+            }
+            if (comp2 == null && comp1 instanceof Integer) {
+                return true;
+            }
+
+            // If components are different, no conflict at this level
+            if (!comp1.equals(comp2)) {
```
tnorimat: I am afraid that the code may cause NPE. For example, on condition where `comp1 == null` and `comp2 == null`, we can reach this code, which leads to NPE.

Also, it might be good as follows because the previous codes returns true:
```
            if (comp1.equals(comp2)) {
                return true;
            }
```

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+                    }
+                } else if (Boolean.TRUE.equals(claim.getMandatory())) {
+                    // Mandatory claim is missing - this should fail
+                    throw new IllegalArgumentException("Mandatory claim not found: " + path);
+                }
+                // Optional claims that don't exist are simply not included
+            } catch (IllegalArgumentException e) {
+                if (Boolean.TRUE.equals(claim.getMandatory())) {
+                    // Re-throw for mandatory claims
+                    throw e;
```
tnorimat, line 225: I think it might be good to record some logs the same as Line 210.

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+                if (component instanceof String) {
+                    // String component: select element by key
+                    if (current instanceof Map) {
+                        Map<?, ?> map = (Map<?, ?>) current;
+                        Object value = map.get(component);
+                        if (value != null) {
+                            nextSelection.add(value);
+                        }
+                        // If key doesn't exist, element is removed from selection (not added to nextSelection)
+                    } else {
```
tnorimat: It seems that this clause is not needed. Also what does `Element is removed from selection` means? It seems that there is no such operation (maybe `continue` ?).

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+                    if (index < 0) {
+                        throw new IllegalArgumentException("Negative integer values are not allowed in claims path pointer");
+                    }
+                    if (current instanceof List) {
+                        List<?> list = (List<?>) current;
+                        if (index < list.size()) {
+                            nextSelection.add(list.get(index));
+                        }
+                        // If index out of bounds, element is removed from selection
+                    } else {
```
tnorimat: It seems that this clause is not needed. Also what does `Element is removed from selection` means? It seems that there is no such operation (maybe `continue` ?).

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+                        // Not a list, abort processing for this element
+                        // Element is removed from selection
+                    }
+                } else if (component == null) {
+                    // Null component: select all elements of currently selected array(s)
+                    if (current instanceof List) {
+                        List<?> list = (List<?>) current;
+                        nextSelection.addAll(list);
+                    } else {
```
tnorimat: It seems that this clause is not needed. Also what does `Element is removed from selection` means? It seems that there is no such operation (maybe `continue` ?).

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /**
+     * Builds nested claim structure for complex paths.
+     *
+     * @param claims the claims map to build in
+     * @param path   the claims path pointer
+     * @param value  the value to add
+     */
+    private static void buildNestedClaimStructure(Map<String, Object> claims, List<Object> path, Object value) {
```
tnorimat, line 380: It seems that codes of the method `buildNestedClaimStructure` and the method `createArrayStructureForMultipleValues` are almost the same.

```
        if (!rootMap.containsKey(rootKey)) {
            rootMap.put(rootKey, new HashMap<String, Object>());
        }
```

versus

```
        if (!rootMap.containsKey(rootKey)) {
            rootMap.put(rootKey, new ArrayList<Object>());
        }
```
,

```
            if (current instanceof Map) {
                Map<String, Object> map = (Map<String, Object>) current;
                map.put((String) finalComponent, value);
            }
```

versus

```
            if (current instanceof Map) {
                Map<String, Object> map = (Map<String, Object>) current;
                // For array selection, store all values
                map.put((String) finalComponent, new ArrayList<Object>(values));
            }
```

,

```
                list.set(index, value);
            }
```

versus

```
                list.set(index, new ArrayList<Object>(values));
            }
```


It might be better to remove duplicated codes.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -487,20 +490,62 @@ public Response requestCredential(String requestPayload) {
...
+                    LOGGER.errorf(e, "Failed to find credential scope for identifier: %s", credentialRequestVO.getCredentialIdentifier());
+                    throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_IDENTIFIER));
+                }
             }
-            
-            return new BadRequestException(getErrorResponse(errorType));
-        });
+        } else if (credentialRequestVO.getCredentialConfigurationId() != null) {
+            // Use credential_configuration_id for direct lookup
+            requestedCredential = credentialRequestVO.findCredentialScope(session).orElseThrow(() -> {
```
tnorimat, line 541: It seems that an throwable supplier doesn't return any exception. How about returning a exception in lambda?

```
                return new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));

```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -106,82 +121,131 @@ private void validateAuthorizationDetail(AuthorizationDetail detail, List<String
...
                 }
             }
         }
+
         if (credentialIdentifiers == null) {
             credentialIdentifiers = new ArrayList<>();
             credentialIdentifiers.add(UUID.randomUUID().toString());
         }
 
+
```
tnorimat: This blank is not needed.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.message.BasicNameValuePair;
+import org.junit.Test;
+import org.keycloak.OAuth2Constants;
+import org.keycloak.protocol.oid4vc.model.AuthorizationDetail;
+import org.keycloak.protocol.oid4vc.model.ClaimsDescription;
+import org.keycloak.protocol.oid4vc.model.CredentialIssuer;
+import org.keycloak.protocol.oid4vc.model.CredentialOfferURI;
```
tnorimat: This class is not used.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.http.entity.StringEntity;
+import org.keycloak.representations.idm.ClientScopeRepresentation;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
```
tnorimat: This class is not used.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsFlowTest.java
tnorimat, line 1: It seems that the codes of this method is the same as `OID4VCJwtAuthorizationCodeFlowTest`. Is it intentional?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsFlowTest.java
tnorimat, line 1: It seems that the codes of this method is the same as `OID4VCSdJwtAuthorizationCodeFlowTest`. Is it intentional?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -0,0 +1,568 @@
...
+import org.keycloak.protocol.oid4vc.model.ClaimsDescription;
+import org.keycloak.protocol.oid4vc.model.CredentialIssuer;
+import org.keycloak.protocol.oid4vc.model.CredentialOfferURI;
+import org.keycloak.protocol.oid4vc.model.CredentialsOffer;
+import org.keycloak.protocol.oid4vc.model.CredentialRequest;
+import org.keycloak.protocol.oid4vc.model.CredentialResponse;
+import org.keycloak.protocol.oid4vc.issuance.OID4VCAuthorizationDetailsResponse;
+import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantTypeFactory;
+import org.keycloak.protocol.oidc.representations.OIDCConfigurationRepresentation;
+import org.keycloak.representations.AccessTokenResponse;
```
tnorimat: It seems that this class is not used.

tnorimat (commented):
@forkimenjeckayang Hello, I added some review comments. Could you check them?

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
forkimenjeckayang: I left that out @tnorimat , thanks for pointing it. Since the spec defines rules for validating the claims path pointer, it's only fair I use the method to validate before processing.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsFlowTest.java
forkimenjeckayang: You're right that the code looks similar, but this is intentional.
Two Different OAuth2 Flows:
- AuthorizationDetailsFlowTest → Tests Pre-authorized Code Flow 
- AuthorizationCodeFlowTest → Tests Authorization Code Flow 
Different test setups - Each flow requires different authentication mechanisms and client configurations

Is that okay @tnorimat , or you'd suggest I do it differently?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsFlowTest.java
forkimenjeckayang: You're right that the code looks similar, but this is intentional.
Two Different OAuth2 Flows:
- AuthorizationDetailsFlowTest → Tests Pre-authorized Code Flow 
- AuthorizationCodeFlowTest → Tests Authorization Code Flow 
Different test setups - Each flow requires different authentication mechanisms and client configurations

Is that okay @tnorimat , or you'd suggest I do it differently?

services/src/main/java/org/keycloak/protocol/oid4vc/utils/ClaimsPathPointer.java
tnorimat: Thank you for the clarification.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJwtAuthorizationDetailsFlowTest.java
tnorimat: It is OK. Thank you for the clarification.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtAuthorizationDetailsFlowTest.java
tnorimat: It is OK. Thank you for the clarification.

tnorimat (approved):
I confirmed that my review comments were resolved.

mposolda (approved):
@forkimenjeckayang @tnorimat Thanks for the PR and review!

PR #42623, opened 2025-09-15
fix: limiting what fields are hashed to identify compatible update jobs

### LINKED ISSUE #41014: Operator auto update hash
We can remove anything that doesn't contribute to the Keycloak configuration.

The following fields can be ignored: 
* `instances`
* `startOptimized` - 90% sure it is fine to ignore it in the hash
* `imagePullSecrets`
* `ingressSpec`
* `truststores` - they are automatically mounted in the containers, so they have no use for the update job
* `resourceRequirements`
* `schedulingSpec`
* `bootstrapAdminSpec` - nothing happens if changed in a running cluster, right?
* `networkPolicySpec`
* `updateSpec`
* `readinessProbeSpec`
* `livenessProbeSpec`
* `startupProbeSpec`

_Originally posted by @pruivo in https://github.com/keycloak/keycloak/issues/40985#issuecomment-3047998542_

shawkins:
closes: #41014

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
         run: |
           SEP=""
           PROJECTS=""
-          for i in `find -name '*Test.java' -type f | egrep -v './(testsuite|quarkus|docs|tests|test-framework)/' | sed 's|/src/test/java/.*||' | sort | uniq | sed 's|./||'`; do
+          for i in `find -name '*Test.java' -type f | egrep -v './(operator|testsuite|quarkus|docs|tests|test-framework)/' | sed 's|/src/test/java/.*||' | sort | uniq | sed 's|./||'`; do
             PROJECTS="$PROJECTS$SEP$i"
             SEP=","
           done
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
@@ -30,8 +30,9 @@ public class KeycloakServiceMonitorDependentResource extends CRUDKubernetesDepen
         public boolean isMet(DependentResource<Ingress, Keycloak> dependentResource, Keycloak primary,
                              Context<Keycloak> context) {
             var opts = configuredOptions(primary);
-            if (Boolean.parseBoolean(opts.get(LEGACY_MANAGEMENT_ENABLED)) || !Boolean.parseBoolean(opts.getOrDefault(METRICS_ENABLED, "false")))
+            if (Boolean.parseBoolean(opts.get(LEGACY_MANAGEMENT_ENABLED)) || !Boolean.parseBoolean(opts.getOrDefault(METRICS_ENABLED, "false"))) {
                 return false;
+            }
 
             return ServiceMonitorSpec.get(primary).isEnabled();
         }
@@ -46,6 +47,7 @@ public class KeycloakServiceMonitorDependentResource extends CRUDKubernetesDepen
               .withNewMetadata()
                 .withName(meta.getName())
                 .withNamespace(meta.getNamespace())
+                .withLabels(Utils.allInstanceLabels(primary))
               .endMetadata()
               .withNewSpec()
                 .withNewNamespaceSelector()
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakUpdateJobDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakUpdateJobDependentResource.java
@@ -32,6 +32,7 @@ import org.keycloak.operator.ContextUtils;
 import org.keycloak.operator.Utils;
 import org.keycloak.operator.crds.v2alpha1.CRDUtils;
 import org.keycloak.operator.crds.v2alpha1.deployment.Keycloak;
+import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakSpecBuilder;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.UpdateSpec;
 
 import io.fabric8.kubernetes.api.model.ContainerFluent;
@@ -218,8 +219,12 @@ public class KeycloakUpdateJobDependentResource extends CRUDKubernetesDependentR
         return Stream.concat(updateArgs.stream(), currentArgs.stream().filter(arg -> !arg.equals("start"))).toList();
     }
 
-    private static String keycloakHash(Keycloak keycloak) {
-        return Utils.hash(List.of(keycloak.getSpec()));
+    static String keycloakHash(Keycloak keycloak) {
+        return Utils.hash(
+                List.of(new KeycloakSpecBuilder(keycloak.getSpec()).withInstances(null).withLivenessProbeSpec(null)
+                        .withStartupProbeSpec(null).withReadinessProbeSpec(null).withResourceRequirements(null)
+                        .withSchedulingSpec(null).withNetworkPolicySpec(null).withIngressSpec(null)
+                        .withImagePullSecrets().withImportSpec(null).withServiceMonitorSpec(null).build()));
     }
 
     private static Map<String, String> getLabels(HasMetadata keycloak) {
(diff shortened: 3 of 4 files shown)


operator/src/main/java/org/keycloak/operator/controllers/KeycloakUpdateJobDependentResource.java
```diff
@@ -218,8 +219,14 @@ private static List<String> replaceStartWithUpdateCommand(List<String> currentAr
...
-    private static String keycloakHash(Keycloak keycloak) {
-        return Utils.hash(List.of(keycloak.getSpec()));
+    static String keycloakHash(Keycloak keycloak) {
+        return Utils.hash(
+                List.of(new KeycloakSpecBuilder(keycloak.getSpec()).withInstances(null).withLivenessProbeSpec(null)
+                        .withHostnameSpec(null).withStartOptimized(null).withTracingSpec(null).withTruststores(null)
+                        .withHttpManagementSpec(null).withProxySpec(null).withStartupProbeSpec(null)
+                        .withReadinessProbeSpec(null).withResourceRequirements(null).withSchedulingSpec(null)
+                        .withBootstrapAdminSpec(null).withNetworkPolicySpec(null).withIngressSpec(null)
+                        .withImagePullSecrets().withImportSpec(null).withUpdateSpec(null).build()));
```
pruivo: The following spec may be ignored too:
* TransactionSpec
* ProxySpec

shawkins: ProxySpec was already covered. Added TransactionSpec.

pruivo (dismissed):
LGTM! Thank you.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
```diff
@@ -46,6 +47,7 @@ protected ServiceMonitor desired(Keycloak primary, Context<Keycloak> context) {
               .withNewMetadata()
                 .withName(meta.getName())
                 .withNamespace(meta.getNamespace())
+                .withLabels(Utils.allInstanceLabels(primary))
```
shawkins, line 50: @ryanemerson this appears to be missing from the original servicemonitor pr. It showed up in some test failure here that creating the servicemonitor was failing because it already existed.

Let me know if you want this in a different pr.

ryanemerson: Happy for it to be fixed in this PR, thanks.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakUpdateJobDependentResource.java
```diff
@@ -218,8 +219,15 @@ private static List<String> replaceStartWithUpdateCommand(List<String> currentAr
...
-    private static String keycloakHash(Keycloak keycloak) {
-        return Utils.hash(List.of(keycloak.getSpec()));
+    static String keycloakHash(Keycloak keycloak) {
+        return Utils.hash(
+                List.of(new KeycloakSpecBuilder(keycloak.getSpec()).withInstances(null).withLivenessProbeSpec(null)
+                        .withHostnameSpec(null).withStartOptimized(null).withTracingSpec(null).withTruststores(null)
+                        .withHttpManagementSpec(null).withProxySpec(null).withStartupProbeSpec(null)
+                        .withReadinessProbeSpec(null).withResourceRequirements(null).withSchedulingSpec(null)
+                        .withBootstrapAdminSpec(null).withNetworkPolicySpec(null).withIngressSpec(null)
+                        .withImagePullSecrets().withImportSpec(null).withUpdateSpec(null).withTransactionsSpec(null)
```
ryanemerson: I think we can also exclude the ServiceMonitor spec here

shawkins: Thanks @ryanemerson, updated.

ryanemerson (dismissed):

ryanemerson:
LGTM, thanks @shawkins

pruivo (dismissed):

operator/src/main/java/org/keycloak/operator/controllers/KeycloakUpdateJobDependentResource.java
vmuzikar: Are we sure to exclude some server affecting specs Hostname, Proxy, ...? What if, at some point, it requires the update decision logic? This will be hard to track and maintain.

.github/workflows/ci.yml
```diff
@@ -104,7 +104,9 @@ jobs:
           SEP=""
           PROJECTS=""
           for i in `find -name '*Test.java' -type f | egrep -v './(testsuite|quarkus|docs|tests|test-framework)/' | sed 's|/src/test/java/.*||' | sort | uniq | sed 's|./||'`; do
-            PROJECTS="$PROJECTS$SEP$i"
+            if [ $i != "operator" ]; then
```
vmuzikar: Shouldn't we handle this exclusion in the `egrep`?

shawkins:
> Are we sure to exclude some server affecting specs Hostname, Proxy, ...? What if, at some point, it requires the update decision logic? This will be hard to track and maintain.

There are really two different levels of update consideration here. update-compatibility is focused on things like keycloak version, database, infinispan, and feature compability - all build time / image concerns.

There are however obviously things you can do to runtime options that will not mesh well with other running instances - but we've always simply rolled in those changes. Are you suggesting that we'll expand the update-compatibility logic to consider changes in runtime options, or are you thinking that we'll eventually have build time options in some of these specs that the update compatibility logic will be interested in?

.github/workflows/ci.yml
shawkins: Ok, moved it there.

vmuzikar:
> Are you suggesting that we'll expand the update-compatibility logic to consider changes in runtime options, or are you thinking that we'll eventually have build time options in some of these specs that the update compatibility logic will be interested in?

Both, actually. My point was, we shouldn't have any "filtering logic" on the Operator level deciding what server options should be considered for the update logic, that needs to be exclusively in the server as the single source of truth.

shawkins:
> Both, actually. My point was, we shouldn't have any "filtering logic" on the Operator level deciding what server options should be considered for the update logic, that needs to be exclusively in the server as the single source of truth.

The initial motivation for the change was to remove the most likely operator side modification - which is instances because that could be due to hpa - from this hash. I can change it to be just operator related fields, but in general what you are concerned with wrt to the update-compatibility logic would be quite an undertaking.

And just to clarify this isn't filtering - this is just for computing a hash to confirm if the current update job is associated with the given Keycloak instance.

pruivo:
Honestly, I'm a bit torn. @vmuzikar has a point; changing the HTTPS port, for example, should invalidate the update job. But there are attributes in there that make no sense to be hashed, like annotations and labels. 
I don't think the `update-compatibility` cares abort ports at all. And we definitely aren't going to bring a cluster down to change the port; we will do a rolling update. 

I think it is safe to ignore the HttpSpec from the hashing.

shawkins:
> Honestly, I'm a bit torn. @vmuzikar has a point; changing the HTTPS port, for example, should invalidate the update job.

Just to make sure we're all on the same page, the sequence of events here is:
- an initial change comes into the Keycloak CR and triggers the update job logic to run.
- before the update job finishes, there are additional reconciliations. 
 
If there have been additional changes to the Keycloak CR we either need to:
- determine if it's something we can ignore and let the current job complete (this pr).
- kill the job and start another reconciliation based upon the new Keycloak CR state.

Either way the additional changes are effective blocked from rolling in until either the job completes or we've determined that an update job isn't needed.

I know there's only a limited window of when it matters, but I would like to at least exclude changes in the instance count from restarting the job as that may not be something the admin is even directly changing. Any spec that touches the runtime has been removed from the PR.

shawkins:
@vmuzikar if you still have misgivings about this pr after the latest change, I'll check the label changes in under a different issue so that can move forward.

pruivo (approved):

vmuzikar (approved):
LGTM, thanks

PR #42629, opened 2025-09-15
Automatically create external caches for MULTI_SITE deployments

### LINKED ISSUE #32129: Automatically create external caches for MULTI_SITE deployments

pruivo:
Closes #32129

### DIFF
--- a/docs/documentation/release_notes/topics/26_4_0.adoc
+++ b/docs/documentation/release_notes/topics/26_4_0.adoc
@@ -122,3 +122,23 @@ The Operator now provisions a `ServiceMonitor` for the management endpoint if me
 specification of the `ServiceMonitor` takes into account the various management endpoint configurations, to ensure that
 metrics can be scraped without any additional configuration. If you do not want a `ServiceMonitor` to be created, you can disable
 this by setting `spec.serviceMonitor.enabled: false`. For more details, see the link:{operatorguide_link}[{operatorguide_name}].
+
+== Automatic cache creation in external Infinispan
+
+You no longer need to manually create caches in your external Infinispan cluster.
+When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they don't already exist.
+
+This new feature solves a common pain point, removing the extra step of configuring the external cluster manually.
+
+Don't worry — any existing, manually configured caches are safe.
+Their settings will be preserved and will not be changed.
+
+For high availability, you can now easily configure cross-site replication.
+Simply set the backup site name (e.g., availability zone) using the following option:
+
+[source,bash]
+----
+--spi-cache-remote--default--remote-sites=<name>
+----
+
+When this option is set, Infinispan will automatically replicate the cache data to the specified location.
--- a/docs/documentation/release_notes/topics/26_4_0.adoc
+++ b/docs/documentation/release_notes/topics/26_4_0.adoc
@@ -123,22 +123,20 @@ specification of the `ServiceMonitor` takes into account the various management 
 metrics can be scraped without any additional configuration. If you do not want a `ServiceMonitor` to be created, you can disable
 this by setting `spec.serviceMonitor.enabled: false`. For more details, see the link:{operatorguide_link}[{operatorguide_name}].
 
-== Automatic cache creation in external Infinispan
+== {project_name} automatically creates the necessary caches on the first startup if they do not exist
 
 You no longer need to manually create caches in your external Infinispan cluster.
-When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they don't already exist.
 
-This new feature solves a common pain point, removing the extra step of configuring the external cluster manually.
+When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they do not already exist on the Infinispan server.
 
-Don't worry — any existing, manually configured caches are safe.
-Their settings will be preserved and will not be changed.
+Any existing caches, manually created before {project_name} startup, will be preserved, and their configuration will not be modified.
 
 For high availability, you can now easily configure cross-site replication.
 Simply set the backup site name (e.g., availability zone) using the following option:
 
 [source,bash]
 ----
---spi-cache-remote--default--remote-sites=<name>
+--cache-remote-backup-sites=<name>
 ----
 
 When this option is set, Infinispan will automatically replicate the cache data to the specified location.
--- a/docs/guides/high-availability/examples/generated/keycloak-ispn.yaml
+++ b/docs/guides/high-availability/examples/generated/keycloak-ispn.yaml
@@ -486,8 +486,10 @@ spec:
       secret:
         name: remote-store-secret
         key: password
+    - name: spi-cache-remote--default--remote-sites
+      value: "site-2" # <5>
+    # end::keycloak-ispn[]
     - name: db-driver
-      # end::keycloak-ispn[]
 
       value: software.amazon.jdbc.Driver
   http:
--- a/docs/guides/high-availability/examples/generated/keycloak-ispn.yaml
+++ b/docs/guides/high-availability/examples/generated/keycloak-ispn.yaml
@@ -486,7 +486,7 @@ spec:
       secret:
         name: remote-store-secret
         key: password
-    - name: spi-cache-remote--default--remote-sites
+    - name: cache-remote-backup-sites
       value: "site-2" # <5>
     # end::keycloak-ispn[]
     - name: db-driver
--- a/docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
+++ b/docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
@@ -191,6 +191,7 @@ include::../examples/generated/ispn-site-b.yaml[tag=infinispan-crossdc]
 +
 {project_name} automatic creates the caches.
 To customize any default configuration, use the {infinispan-operator-docs}#creating-caches_creating-caches[`Cache` CR] to declare the full cache configuration.
+To take effect, the `Cache` CR must be deployed before any {project_name} Pod starts.
 
 == Verifying the deployment
 
@@ -237,7 +238,7 @@ include::../examples/generated/keycloak-ispn.yaml[tag=keycloak-ispn]
 This is optional, and it defaults to `11222`.
 <3> The Secret `name` and `key` with the {jdgserver_name} username credential.
 <4> The Secret `name` and `key` with the {jdgserver_name} password credential.
-<5> The mame of the remote site. 
+<5> The mame of the remote site.
 
 === Architecture
 
--- a/docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
+++ b/docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
@@ -275,6 +275,15 @@ kubectl wait --for condition=WellFormed --timeout=300s infinispans.infinispan.or
 kubectl wait --for condition=CrossSiteViewFormed --timeout=300s infinispans.infinispan.org -n {ns} {cluster-name}
 ----
 
+.Wait until the {jdgserver_name} caches are ready
+[source,bash,subs="+attributes"]
+----
+kubectl wait --for condition=Ready --timeout=300s cache.infinispan.org -n {ns} actionTokens
+kubectl wait --for condition=Ready --timeout=300s cache.infinispan.org -n {ns} authenticationSessions
+kubectl wait --for condition=Ready --timeout=300s cache.infinispan.org -n {ns} loginFailures
+kubectl wait --for condition=Ready --timeout=300s cache.infinispan.org -n {ns} work
+----
+
 [#connecting-infinispan-to-keycloak]
 == Connecting {jdgserver_name} with {project_name}
 
--- a/docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
+++ b/docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
@@ -189,75 +189,8 @@ include::../examples/generated/ispn-site-b.yaml[tag=infinispan-crossdc]
 
 . Creating the caches for {project_name}.
 +
-{project_name} requires the following caches to be present: `actionTokens`, `authenticationSessions`, `loginFailures`, and `work`.
-+
-The {jdgserver_name} {infinispan-operator-docs}#creating-caches[Cache CR] allows deploying the caches in the {jdgserver_name} cluster.
-Cross-site needs to be enabled per cache as documented by {infinispan-xsite-docs}[Cross Site Documentation].
-The documentation contains more details about the options used by this {section}.
-The following example shows the `Cache` CR for `{site-a}`.
-+
---
-. In `{site-a}` create a `Cache` CR for each of the caches mentioned above with the following content.
-+
-.Cache `actionTokens`
-[source,yaml]
-----
-include::../examples/generated/ispn-site-a.yaml[tag=infinispan-cache-actionTokens]
-----
-+
-.Cache `authenticationSessions`
-[source,yaml]
-----
-include::../examples/generated/ispn-site-a.yaml[tag=infinispan-cache-authenticationSessions]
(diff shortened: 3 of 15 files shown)


model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/remote/DefaultCacheRemoteConfigProviderFactory.java
```diff
@@ -357,4 +354,17 @@ private static void addTlsConfig(ProviderConfigurationBuilder builder) {
                 .type(ProviderConfigProperty.STRING_TYPE)
                 .add();
     }
+
+    private static void addCreateRemoteCachesConfig(ProviderConfigurationBuilder builder) {
+        builder.property()
+                .name(CREATE_CACHES)
+                .helpText("If true, Keycloak will create the caches in the external Infinispan cluster.")
+                .type(ProviderConfigProperty.BOOLEAN_TYPE)
+                .add();
```
pruivo: This SPI option replaces the system properties, but I'm having doubts about its use. It is enabled by default, and I'm thinking about removing it. If disabled, Keycloak will fail to start when the caches do not exist.

IMO, it is a useless option. May I remove it? 

@ahus1 @ryanemerson

ryanemerson: My understanding from past discussions is that if the cache already exists on the Infinispan server, then the cache config we try to apply from Keycloak is ignored as `getOrCreate` is used on the server-side. Hence it's still possible for users to modify their cache configs if required. So I'm happy removing the option altogether as it is one less thing to potentially confuse users.

ahus1: The one situation where it makes sense to fail the startup when the caches do not exist, is when you want to have full control over the cache configurations, and want KC to prevent the automatic setup in case your manual setup fails / leads to any error. 

I could also imagine a race condition where ISPN started, but the Operator did not yet create the caches. And in exactly that moment Keycloak starts, creates the caches, and then the Operator can't apply the custom configuration. 

So I'd say this is valid for some setups, still I'd say having a new default that creates the caches automatically is the way for the future. 

If you still consider this option useless, feel free to remove it.

pruivo: Exactly! @ryanemerson

ryanemerson: > I could also imagine a race condition where ISPN started, but the Operator did not yet create the caches. And in exactly that moment Keycloak starts, creates the caches, and then the Operator can't apply the custom configuration.

Just an FYI. The Cache CRs have a "Ready" condition which is only set to True once the cache configuration has been applied on the server. So I would expect/hope that any automated deployment would wait on these conditions before proceeding with the Keycloak deployment, especially considering that users have to do this for the Infinispan CR given that Keycloak will fail to startup if the remote cache is not available.

pruivo: @ryanemerson Does the CacheCR wait for the cluster to form, or do they apply the cache configuration when the first Pod starts?

@ahus1 I would rather improve the docs to tell to wait for the CacheCR than add/document a new option :sweat_smile:

ryanemerson: > Does the CacheCR wait for the cluster to form, or do they apply the cache configuration when the first Pod starts?

It waits for the cluster to form with all expected pods in the Ready state and also ensures that all cluster members have the same view, i.e. no split-brain.

ahus1: > I would rather improve the docs to tell to wait for the CacheCR than add/document a new option 😅

Works for me, still I know not all the people read the docs. At the same time I want to have the cache creation enabled by default, which safes reading a lot of docs in the first place. 

So please proceed.

ryanemerson:
@pruivo You need to rebase on the latest `main` in order to remove the usage of `kc.cache-remote-create-caches` in the new testsuite.

pruivo:
Thanks for the heads up @ryanemerson.

ahus1:
Looks good to me, and can be merged as-is. 

Would we in a follow-up issue remove the section in the docs on how to manually create the caches? See https://www.keycloak.org/nightly/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc for the nightly docs.

ryanemerson (dismissed):

ryanemerson:
> Would we in a follow-up issue remove the section in the docs on how to manually create the caches?

If we do that,  there is no need for the new "Wait until the {jdgserver_name} caches are ready" section as this is no longer applicable to deploying Infinispan.

My preference would be to tackle that in this PR as the change is tightly coupled, but I'll leave that up to @pruivo to decide.

ahus1:
When we update the docs, we would need to decide if the docs are opinionated that the we only recommend the automatic creation of the caches, not document the manual creation. 

One possible benefit of the separate PRs: We could merge this one, updated KCB to verify this works as expected, and then update the upstream docs. 

Still, I leave the decision to @pruivo

pruivo:
@ryanemerson @ahus1 I'm thinking of leaving the docs as is to avoid confusion for RHBK users and only update them in KC 27.

ryanemerson:
> When we update the docs, we would need to decide if the docs are opinionated that the we only recommend the automatic creation of the caches, not document the manual creation.

Whenever/wherever we update the docs, I guess the question is, whether there's a clear benefit to manually defining the caches instead of relying on Keycloak to create them? Unless we can think of a compelling argument for it (I can't off the top of my head), I would be in favour of simplifying and removing these steps from the docs.


> I'm thinking of leaving the docs as is to avoid confusion for RHBK users and only update them in KC 27.

-1 from me. This is a big usability win, especially for new users, so I think it would be a shame to wait 6 months before exposing this to users.

pruivo:
@ryanemerson ok. I'll update the docs.

docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
```diff
@@ -189,75 +189,8 @@ include::../examples/generated/ispn-site-b.yaml[tag=infinispan-crossdc]
...
-+
-For `{site-b}`, the `Cache` CR is similar, except for the `backups.<name>` outlined in point 3 of the above diagram.
-+
-.Example for `actionTokens` cache in `{site-b}`
-[source,yaml]
-----
-include::../examples/generated/ispn-site-b.yaml[tag=infinispan-cache-actionTokens]
-----
+{project_name} automatic creates the caches.
+To customize any default configuration, use the {infinispan-operator-docs}#creating-caches_creating-caches[`Cache` CR] to declare the full cache configuration.
```
ryanemerson, line 193: It is maybe implied, but should we add "before Keycloak starts up" to make this explicit?

ryanemerson (commented):
Thanks @pruivo. We should also document this in the release notes so that users are aware of the improvement.

docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
pruivo: Sure. Let me know what you think.

docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
```diff
@@ -237,7 +238,7 @@ include::../examples/generated/keycloak-ispn.yaml[tag=keycloak-ispn]
 This is optional, and it defaults to `11222`.
 <3> The Secret `name` and `key` with the {jdgserver_name} username credential.
 <4> The Secret `name` and `key` with the {jdgserver_name} password credential.
-<5> The mame of the remote site. 
+<5> The mame of the remote site.
```
ryanemerson: `s/mame/name`

docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
ryanemerson: Much better, thanks

pruivo:
@ahus1 @ryanemerson I've added a section to the release notes. Let me know how bad it is! :sweat_smile:

docs/documentation/release_notes/topics/26_4_0.adoc
```diff
@@ -115,3 +115,23 @@ The Operator now provisions a `ServiceMonitor` for the management endpoint if me
...
+
+== Automatic cache creation in external Infinispan
+
+You no longer need to manually create caches in your external Infinispan cluster.
+When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they don't already exist.
+
+This new feature solves a common pain point, removing the extra step of configuring the external cluster manually.
+
+Don't worry — any existing, manually configured caches are safe.
+Their settings will be preserved and will not be changed.
```
ryanemerson: I think we can make this a little more succinct and slightly more formal, how about:

```
When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they don't already exist on the Infinispan server.

Any existing caches, manually created before {project_name} startup, will be preserved and their configuration will not be modified.
```

docs/documentation/release_notes/topics/26_4_0.adoc
```diff
@@ -115,3 +115,23 @@ The Operator now provisions a `ServiceMonitor` for the management endpoint if me
...
+
+For high availability, you can now easily configure cross-site replication.
+Simply set the backup site name (e.g., availability zone) using the following option:
```
ryanemerson, line 135: I would remove the example here and add it to the multi-cluster guide.

ryanemerson (commented):
Thanks @pruivo, the release notes are mostly there, just a couple of suggestions.

docs/guides/high-availability/examples/generated/keycloak-ispn.yaml
```diff
@@ -486,8 +486,10 @@ spec:
       secret:
         name: remote-store-secret
         key: password
+    - name: spi-cache-remote--default--remote-sites
```
ahus1, line 489: Can I please have a CLI option? Suggested name: `--cache-remote-backup-sites`

docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
```diff
@@ -189,75 +189,9 @@ include::../examples/generated/ispn-site-b.yaml[tag=infinispan-crossdc]
...
-====
-+
-For `{site-b}`, the `Cache` CR is similar, except for the `backups.<name>` outlined in point 3 of the above diagram.
-+
-.Example for `actionTokens` cache in `{site-b}`
-[source,yaml]
-----
-include::../examples/generated/ispn-site-b.yaml[tag=infinispan-cache-actionTokens]
-----
+{project_name} automatic creates the caches.
```
ahus1, line 192:
```suggestion
{project_name} automatically creates the necessary caches on the first startup if they do not exist.
```

docs/documentation/release_notes/topics/26_4_0.adoc
```diff
@@ -115,3 +115,23 @@ The Operator now provisions a `ServiceMonitor` for the management endpoint if me
 specification of the `ServiceMonitor` takes into account the various management endpoint configurations, to ensure that
 metrics can be scraped without any additional configuration. If you do not want a `ServiceMonitor` to be created, you can disable
 this by setting `spec.serviceMonitor.enabled: false`. For more details, see the link:{operatorguide_link}[{operatorguide_name}].
+
+== Automatic cache creation in external Infinispan
+
+You no longer need to manually create caches in your external Infinispan cluster.
+When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they don't already exist.
```
ahus1, line 122:
```suggestion
When using the `multi-site` or `clusterless` features, {project_name} now automatically creates the necessary caches during startup if they do not already exist.
```

docs/documentation/release_notes/topics/26_4_0.adoc
```diff
@@ -115,3 +115,23 @@ The Operator now provisions a `ServiceMonitor` for the management endpoint if me
...
+
+This new feature solves a common pain point, removing the extra step of configuring the external cluster manually.
```
ahus1, line 124:
```suggestion
This new feature removes the extra step of configuring the external cluster manually.
```

ahus1 (changes requested):
Thank you for this PR. 

Please instead of the SPI option, add a new CLI option like `--cache-remote-backup-sites`.

docs/documentation/release_notes/topics/26_4_0.adoc
pruivo: It is present in the multi-site guide.

ryanemerson: I don't see where we suggest to use the availability-zone as the site name, where is it?

docs/guides/high-availability/examples/generated/keycloak-ispn.yaml
```diff
@@ -486,8 +486,10 @@ spec:
       secret:
         name: remote-store-secret
         key: password
+    - name: cache-remote-backup-sites
```
pruivo, line 489: > I don't see where we suggest to use the availability-zone as the site name, where is it?

@ryanemerson here. In our guide, we use `site-1` and `site-2`, but it can be anything.

ryanemerson: Yep I saw this, but what is missing is the hint that you could/should set it to the availability-zone name.

docs/guides/high-availability/multi-cluster/deploy-infinispan-kubernetes-crossdc.adoc
```diff
@@ -172,7 +172,7 @@ include::../examples/generated/ispn-site-a.yaml[tag=infinispan-crossdc]
 <8> The secret key (filename) of the Keystore as defined in the previous step.
 <9> The secret name where the Truststore exists as defined in the previous step.
 <10> The Truststore key (filename) of the Keystore as defined in the previous step.
-<11> The remote site's name, in this case  `{site-b}`.
+<11> The remote site's name, in this case  `{site-b}`. Use this value in {project_name} option `cache-remote-backup-sites`.
```
pruivo, line 175: @ryanemerson Does this sentence make it clearer or worse?

ryanemerson: I think that's better :+1:

ryanemerson: To be clear, the bit I thought was missing was the hint to use the availability-zone as the value for the site identifier, but I don't think it's that important so happy to not include this.

ahus1 (approved):

pruivo:
:partying_face: Thanks!

PR #42648, opened 2025-09-16
Validation for immediate workflows

vramik:
Closes #42382 

Adds a validation for workflows upon their creation:

- immediate workflow cannot have time conditions
- all steps of scheduled workflow must have time condition
- workflow cannot be both immediate and recurring

When binding a resource with workflow by `bind(String type, String resourceId, Long miliseconds)` method:

- doesn't allow binding with immediate workflow (it does allow binding by `bind(String type, String resourceId`)
- for scheduled workflow it overrides the initial time by specified amount


services/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
```diff
@@ -343,6 +340,26 @@ public Workflow toModel(WorkflowRepresentation rep) {
...
 
+    private void validateWorkflow(WorkflowRepresentation rep, MultivaluedHashMap<String, String> config) {
+        // Validations:
+        //  immediate workflow cannot have time conditions
+        //  scheduled workflow must have time conditions
+        //  workflow cannot be both immediate and recurring
+        boolean isImmediate = config.containsKey("scheduled") && !Boolean.parseBoolean(config.getFirst("scheduled"));
+        boolean isRecurring = config.containsKey("recurring") && Boolean.parseBoolean(config.getFirst("recurring"));
+        boolean hasTimeCondition = rep.getSteps().stream().allMatch(step -> step.getConfig() != null
+                && step.getConfig().containsKey("after"));
```
pedroigor: Better refence "after" using `org.keycloak.models.workflow.WorkflowStep#AFTER_KEY`. I would also introduce a constant for the others we are missing.

model/jpa/src/main/java/org/keycloak/models/workflow/JpaWorkflowStateProvider.java
```diff
@@ -57,17 +59,24 @@ public ScheduledStep getScheduledStep(String workflowId, String resourceId) {
     public void scheduleStep(Workflow workflow, WorkflowStep step, String resourceId) {
         WorkflowStateEntity.PrimaryKey pk = new WorkflowStateEntity.PrimaryKey(resourceId, workflow.getId());
         WorkflowStateEntity entity = em.find(WorkflowStateEntity.class, pk);
+
+        Long workflowStepTimeOffset = session.getAttribute(WORKFLOW_FIRST_STEP_TIME_OFFSET, Long.class);
+        long scheduledStepTimestamp = (workflowStepTimeOffset == null) ?
+                Time.currentTimeMillis() + step.getAfter() :
+                Time.currentTimeMillis() + step.getAfter() + workflowStepTimeOffset;
+        session.removeAttribute(WORKFLOW_FIRST_STEP_TIME_OFFSET);
```
pedroigor: Wouldn't it be better to have a `scheduleStep` variant that has the offset as a parameter? Using a session attribute is a bit odd, isn't it?

pedroigor (approved):

sguilhen (approved):

PR #42651, opened 2025-09-16
Enable validate signature for SAML IdP to true when there are signing keys in the IdP metadata

### LINKED ISSUE #42213: Importing SAML IdP metadata sets Validate Signatures to false even if signing certificate is provided
### Before reporting an issue

### Area

saml

### Describe the bug

When adding a new SAML Identity Provider to a Keycloak realm and importing the IdP's metadata, either using the metadata URL or via the XML import, Validate Signatures is not always set to true even if a signing certificate is provided in the IdP metadata.

In fact, Validate Signatures is only set to true when WantAuthnRequestsSigned is true in the IdP metadata. This doesn't seem correct as that flag relates to SAML requests being signed rather than SAML assertions and responses.

### Version

26.3.3

### Regression

### Expected behavior

When importing an IdP's metadata, Validate Signatures is set to true if the IdP metadata contains a signing certificate.

### Actual behavior

When importing an IdP's metadata, Validate Signatures is set to true only when the IdP metadata has WantAuthnRequestsSigned set to true.

### How to Reproduce?

1. Set up a SAML identity provider that doesn't require authentication requests to be signed, but does sign assertions or responses.
2. Create a new SAML 2.0 Identity Provider in Keycloak
3. Import the IdP metadata

### Anything else?

_No response_

rmartinc:
Closes #42213

Little PR that sets the validate signature option in the identity provider when the IdP Metadata Descriptor has signing keys. Previously it was set when the metadata attribute  `WantAuthnRequestsSigned` was set to `true`. This attribute means a requirement done by the IdP for the `<samlp:AuthnRequest>` messages received to be signed. So it's related but not the same, it means the opposite, that the SP should sign the requests. The SP in general should always validate signatures except for testing purposes, and it's better to set it to true when any signing key is in the descriptor. I added a test because this was not tested before. The test can be extended later for other use-cases.

### DIFF
--- a/services/src/main/java/org/keycloak/broker/saml/SAMLIdentityProviderFactory.java
+++ b/services/src/main/java/org/keycloak/broker/saml/SAMLIdentityProviderFactory.java
@@ -136,6 +136,7 @@ public class SAMLIdentityProviderFactory extends AbstractIdentityProviderFactory
 
                         if (KeyTypes.SIGNING.equals(keyDescriptorType.getUse())) {
                             samlIdentityProviderConfig.addSigningCertificate(x509KeyInfo.getTextContent());
+                            samlIdentityProviderConfig.setValidateSignature(true);
                         } else if (KeyTypes.ENCRYPTION.equals(keyDescriptorType.getUse())) {
                             samlIdentityProviderConfig.setEncryptionPublicKey(x509KeyInfo.getTextContent());
                         } else if (keyDescriptorType.getUse() == null) {
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/broker/KcSamlParseIdPDescriptorTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/broker/KcSamlParseIdPDescriptorTest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


rmartinc:
Rebased to include the fix for the Account CI.

services/src/main/java/org/keycloak/broker/saml/SAMLIdentityProviderFactory.java
```diff
@@ -114,7 +114,6 @@ public Map<String, String> parseConfig(KeycloakSession session, String config) {
                 samlIdentityProviderConfig.setSingleSignOnServiceUrl(singleSignOnServiceUrl);
                 samlIdentityProviderConfig.setWantAuthnRequestsSigned(idpDescriptor.isWantAuthnRequestsSigned());
                 samlIdentityProviderConfig.setAddExtensionsElementWithKeyInfo(false);
-                samlIdentityProviderConfig.setValidateSignature(idpDescriptor.isWantAuthnRequestsSigned());
```
mposolda, line 117: Will it makes sense to keep this line as well or rather not?

Just thinking about the corner-case when someone updates IDP configuration from the descriptor (which has `wantAuthnRequestSigned=true`), but there are not keys in that descriptor because the guy updates the keys manually in some next step. For this guy, it currently sets `validateSignature` to true. After this change, it will not be updated and hence would be still at false.

Could this scenario happen or not?

mposolda (commented):
@rmartinc Thanks for the PR. I've added one comment inline. Could you please check it?

In general, I have a bit concerned about backwards compatibility and whether this change could somehow break backwards compatibility for some people? But if you confirm that it should be OK from backwards compatibility perspective, I am fine with the PR.

services/src/main/java/org/keycloak/broker/saml/SAMLIdentityProviderFactory.java
rmartinc: I think it's more or less the same. I even think that maybe we should always set validation to true and the admin disable it manually. But I also think that it makes no harm at all doing what you say. If the IdP wants requests signed makes little sense not to sign responses. So I will do as you say. That line is maintained and later if there are signing keys is set to true always. I'll send a PR asap.

rmartinc: Done! Let's see the tests. I think that GH is again not OK and downloads are slow.

mposolda (approved):
@rmartinc Ok, Thanks for the confirm!

PR #42653, opened 2025-09-16
NullPointerException when persisting a client session

pruivo:
Fixes #42652

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
@@ -24,7 +24,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
```diff
@@ -697,6 +696,7 @@ public SessionEntityWrapper<UserSessionEntity> importUserSession(UserSessionMode
 
         // Import client sessions
         clientSessionTx.importSessionsConcurrently(realm, clientSessionsById, offline);
+        clientSessionTx.setUserSessionId(clientSessionsById.keySet(), sessionId, offline);
```
ahus1, line 699: It feels a bit non-obvious to do this setting of the user session after the import. The results is correct though. 

Thinking a bit more about `AuthenticatedClientSessionEntity#userSessionId`, it should probably be volatile as it is updated by potentially multiple threads in parallel (if it is stored on the local node), and no serialization is used. 

I realize that I added the `userSessionId` back in the time, I probably should have change it back then already. 

WDYT?

ahus1 (commented):
Looks good, still one concern listed below. Can you think of a test for this one?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
pruivo: Yeah, I'll change it to volatile.

ahus1 (approved):

PR #42667, opened 2025-09-16
Remove redundant jdbc-ping tests

### LINKED ISSUE #42666: Remove redundant tests for jdbc-ping
### Description

In the CI we have `jdbc-ping` that are now default, and a `jdbc-ping-udp` tested in `ClusterConfigDistTest`, so I think we can remove the separate tests from `ci.yml`.

This should safe like 5 min

ahus1:
Closes #42666

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -603,34 +603,6 @@ jobs:
             -Dkeycloak.distribution.start.timeout=360 \
             -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
 
-      - name: Run cluster JDBC_PING2 UDP smoke test
-        run: |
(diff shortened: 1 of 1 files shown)


ahus1:
@ryanemerson / @pruivo - WDYT, do you agree to remove those? See more extensive justification in the parent issue.

ryanemerson:
AFAIU the `ClusterConfigDistTest` does not form a cluster, it just makes sure that a single server starts up with the specified configuration and expected logs. Whereas these tests actually create a cluster and ensure that Keycloak functions as expected.

-1 from me, as we're losing an important test while those two stacks are still supported. I would prefer to remove those stacks and the tests in KC 27

ahus1:
ok, then let's keep them for now.

Issue #42446, opened 2025-09-08
Make picocli auto color match the quarkus logic


shawkins:
### Description

Picocli messages will always attempt to color and don't respect the keycloak log coloring config option.

### Details

See also https://github.com/keycloak/keycloak/issues/42445

shawkins:
@mabartos @vmuzikar @Pepo48 we probably have a similar issue as to why quarkus switched to using quarkus.console.color - we have log-console-color enabled only when log console is enabled - so it currently understood to be specific only to log messages. 

Should we introduce a console-color option (and deprecate log-console-color), or simply close this issue?

vmuzikar:
@shawkins IMHO users don't really differentiate between "console" and "log". It is fine to keep the current `log-console-color` and just reflect it in Picocli.

shawkins:
> @shawkins IMHO users don't really differentiate between "console" and "log". It is fine to keep the current log-console-color and just reflect it in Picocli.

Ok, so you are in favor or removing the "enabled when" restriction on log-console-color and updating the option text to make it applicable to all things console.

vmuzikar:
> Ok, so you are in favor or removing the "enabled when" restriction on log-console-color and updating the option text to make it applicable to all things console.

Yup. WDYT, does that make sense to you too?

shawkins:
> Yup. WDYT, does that make sense to you too?

I'm pretty indifferent to the 3 options:
- just close this because users haven't requested it yet, or based upon #42445 just change the picocli auto behavior to use the quarkus logic rather than the picocli
- reuse log-console-color - but it's probably not intuitive from a user perspective and it becomes inconsistent with the other log- options.
- introduce a general console-color option - but that seems like too much work

vmuzikar:
> * reuse log-console-color - but it's probably not intuitive from a user perspective and it becomes inconsistent with the other log- options

This would be my vote. Just do for picocli whatever we do for logging system – whether it's the explicitly enabled/disabled color, or left to auto. Why do you think that would be counterintuitive from the user perspective?

shawkins:
> Why do you think that would be counterintuitive from the user perspective?

Because it's under the logging options, and all of the other log-xxx- options are enabled only if their respective logger is enabled.

vmuzikar:
That made me wonder, whether picocli shouldn't honor the console logger setting in general. Meaning e.g. if it's disabled, picocli wouldn't print anything to std out/err.

shawkins:
[...]

PR #42669, opened 2025-09-16
fix: auto-defaulting log console color

shawkins:
closes: #42445

~Reviewed the auto color logic in both quarkus and picocli - they are similar, but picocli seems a little more extensive so I decided to default to that.~

Switched to the quarkus console logic instead as picocli is enabling coloring for the windows tests. I believe this is due to having the OSTYPE environment variable set. This relates to #42446 as it means we're currently adding coloring to picocli messages when we shouldn't be.

We do need to change our boolean handling so that they are allowed to have no default.

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -10,7 +10,7 @@ Notable changes may include internal behavior changes that prevent common miscon
 
 The `log-console-color` previously defaulted to `false`, but it will now instead check if the terminal supports color.
 
-You may still explicitly disable color support by setting the option to false.
+You may still explicitly disable color support by setting the option to `false`.
 
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -1,0 +1,26 @@
+// ------------------------ Breaking changes ------------------------  //
+== Breaking changes
+
+// ------------------------ Notable changes ------------------------ //
+== Notable changes
+
+Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
+
+=== `log-console-color` will automatically enable if supported by the terminal
+
+The `log-console-color` previously defaulted to `false`, but it will now instead check if the terminal supports color.
+
+You may still explicitly disable color support by setting the option to false.
+
+// ------------------------ Deprecated features ------------------------ //
+== Deprecated features
+
+The following sections provide details on deprecated features.
+
+// ------------------------ Removed features ------------------------ //
+== Removed features
+
+The following features have been removed from this release.
+
+=== <TODO>
+
--- a/docs/documentation/upgrading/topics/changes/changes.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes.adoc
@@ -1,6 +1,10 @@
 [[migration-changes]]
 == Migration Changes
 
+=== Migrating to 26.5.0
+
+include::changes-26_5_0.adoc[leveloffset=2]
+
 === Migrating to 26.4.0
 
 include::changes-26_4_0.adoc[leveloffset=2]
--- a/quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
@@ -7,6 +7,7 @@ import java.io.File;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.function.Function;
 
 import static java.lang.String.format;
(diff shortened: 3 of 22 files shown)


shawkins:
Should be basically ready, but marked as on hold for 26.5.

Would you include this as a notable change?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/LoggingPropertyMappers.java
```diff
@@ -263,6 +266,10 @@ public List<PropertyMapper<?>> getPropertyMappers() {
         );
     }
 
+    private String transformConsoleColor(String value, ConfigSourceInterceptorContext context) {
+        return Optional.ofNullable(value).orElse(Boolean.toString(QuarkusConsole.hasColorSupport()));
```
mabartos:
```suggestion
        return Optional.ofNullable(value).orElseGet(() -> Boolean.toString(QuarkusConsole.hasColorSupport()));
```

nitpick

mabartos (commented):
@shawkins LGTM, thanks!

> We do need to change our boolean handling so that they are allowed to have no default.

+1

> Would you include this as a notable change?

It'd be good to provide some.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -0,0 +1,26 @@
...
+// ------------------------ Notable changes ------------------------ //
+== Notable changes
+
+Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
+
+=== `log-console-color` will automatically enable if supported by the terminal
+
+The `log-console-color` previously defaulted to `false`, but it will now instead check if the terminal supports color.
+
+You may still explicitly disable color support by setting the option to false.
```
mabartos:
```suggestion
You may still explicitly disable color support by setting the option to `false`.
```
nit

docs/documentation/upgrading/topics/changes/changes.adoc
```diff
@@ -1,6 +1,10 @@
 [[migration-changes]]
 == Migration Changes
 
+=== Migrating to 26.5.0
+
+include::changes-26_5_0.adoc[leveloffset=2]
```
mabartos, line 6: We should not include it here for now as once 26.4.0 is out, there's no 26.5.0 release.

mabartos (commented):
@shawkins A few last minor things. Otherwise LGTM

docs/documentation/upgrading/topics/changes/changes.adoc
shawkins: I still see the 26.5.0 milestone. I'll put on hold further doc changes until the release target is clear.

mabartos: Sorry, I wasn't very clear. I think it's ok to have the `changes-26_5_0.adoc` in the codebase, but not having it included in the `changes.adoc`. If it's included in the `changes`, the 26_5_0 migration guide will be shown to users after releasing KC 26.4 in the whole migration guide.

shawkins: I'm assuming this change will not be merged until after 26.4 is branched as we have a soft-freeze on 26.4.

mabartos: Ahh, you're right.

mabartos (approved):
LGTM.

On hold - can be merged after KC 26.4.0 release.

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
```diff
@@ -133,8 +134,8 @@ public String toString() {
 
     public static final Option<Boolean> LOG_CONSOLE_COLOR = new OptionBuilder<>("log-console-color", Boolean.class)
```
vmuzikar, line 135: Shouldn't we rather handle this as String and have `true`, `false` and `auto`? I don't really like having unspecified value as the "auto". This makes it impossible to override to auto if some config source with lower priority explicitly set it to `true` or `false`.

vmuzikar (changes requested):

quarkus/config-api/src/main/java/org/keycloak/config/LoggingOptions.java
shawkins: > Shouldn't we rather handle this as String and have true, false and auto?

I don't think so. For one thing it confuses the type - which has other considerations such as not being able to use expected values if the values for true / false are lax.

> This makes it impossible to override to auto if some config source with lower priority explicitly set it to true or false.

That is not the case - if the value is not set we're determining the default via the propertymapper, not first looking for the underlying setting.

And in general we have a lot of options with no default value.

vmuzikar: > For one thing it confuses the type

The type doesn't really matter from the "outside" perspective of the user who is using it in a config source, does it?

> That is not the case

But e.g. if I set `log-console-color=true` in the config file, and want to change it to the automatic behaviour via an env var, I cannot do it, right?

> And in general we have a lot of options with no default value.

Correct. But in this case the "no default" value is smart, non-static value resolved based on an environment. I don't think we have other options like that with "smart no default" value, or did I miss it?

shawkins: > The type doesn't really matter from the "outside" perspective of the user who is using it in a config source, does it?

It does - see the concern above about how to validate the value.

> But e.g. if I set log-console-color=true in the config file, and want to change it to the automatic behaviour via an env var, I cannot do it, right?

That is correct, but highly unlikely - and is also the case with Quarkus itself. Isn't it much more likely that you'd want to specify an actual value rather than restoring the automatic behavior?

This is also related to how empty strings (which you can set via an environment variable) are treated. This was touched on in https://github.com/keycloak/keycloak/pull/40150#discussion_r2285775311 - 
we are not consistent as to what an empty string means, in particular for this issue if you tried to do something like KC_LOG_CONSOLE_COLOR="" that would result in a validation error.

> Correct. But in this case the "no default" value is smart, non-static value resolved based on an environment. I don't think we have other options like that with "smart no default" value, or did I miss it?

We do, at least the http max threads setting.

vmuzikar (approved):
Sorry for the late reply. As discussed offline last week, let's go ahead with this approach.

PR #42671, opened 2025-09-16
Add additional Keycloak lifecycle events

### LINKED ISSUE #42670: Add dedicated events to signal that Keycloak is initialized or shutting down
### Description

Add a dedicated provider events to signal that Keycloak is initialized or getting shutdown.

### Discussion

_No response_

### Motivation

It is quite common to run some custom logic once Keycloak has fully started. Currently, it's common to use the `PostMigrationEvent` to solve this. However, this event only implicitly signals that the server is "ready". 
In case a user wants to execute logic on the server shutdown there is no reliable way to do this.

For symmetry reasons it would be helpful to have an event to signal a completed server initialization and a server shutdown.

### Details

We could define two additional `ProviderEvent`'s like `KeycloakInitializedEvent` and `KeycloakShutdownEvent`. 
This would allow users to register a `ProviderEventListener` via the `KeycloakSessionFactory`/ `ProviderEventManager` to register custom logic that runs when Keycloak is initialized or is about to get shutdown.

thomasdarimont:
This adds a KeycloakInitializedEvent and KeycloakShutdownEvent.

Those events can be used by extensions to register custom logic that should execute after Keycloak server initialization or when the server is about to get stopped.

Fixes #42670

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/PostMigrationEvent.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/PostMigrationEvent.java
@@ -18,6 +18,7 @@
 package org.keycloak.models.utils;
 
 import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.provider.AbstractLifecycleEvent;
 import org.keycloak.provider.ProviderEvent;
 
 /**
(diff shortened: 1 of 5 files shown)


mabartos:
I like this :) In general, LGTM - even the naming. Quarkus has the `StartupEvent` and `ShutdownEvent`, but having the `KeycloakInitializedEvent` feels more right as there's no confusion like: "Is it executed during the startup or when the KC is fully ready?" 

But we need some documentation and tests with some custom implementations.

vmuzikar (commented):
Generally, I don't see any issues with this. Though I'd like to also get @ahus1 opinion on this as he was recently involved in the area around graceful shutdown.

As mentioned already by @mabartos, a test would be good. Not sure about docs, if we even document events somewhere.

shawkins:
Thank you for pr @thomasdarimont, but I'm moving this back to a draft as there's been no movement for a while. Ideally the issue should fully specify the expectations for docs / usage before proceeding witht the pr.

ahus1:
See my comment in the parent: There is now `ShutdownDelayInitiatedEvent`, maybe that is enough for now. 

Maybe @thomasdarimont wants to close this one.

thomasdarimont:
I can close this and sent another PR if necessary. A ShutdownDelayInitiatedEvent sounds a bit too specific, but I need to check it out first.

PR #42679, opened 2025-09-17
Update docs to reflect that Operator ClusterRoleBinding contains hardcoded namespace

### LINKED ISSUE #42678: Operator ClusterRoleBinding contains hardcoded namespace
### Before reporting an issue

### Area

operator

### Describe the bug

The `ClusterRoleBinding` uses a hardcoded namespace, which causes the binding to fail if deploying the Operator to a namespace other than `keycloak`.

The `Role` and `ClusterRole` definitions are also missing metadata labels.

### Version

Nightly

### Regression

### Expected behavior

The guides should detail that the `ClusterRoleBinding` should be updated when using a namespace other than "keycloak".

### Actual behavior

The current instructions in the guides will cause the Operator deployment to not function as expected as CustomResourceDefinitions can not be retrieved when the `ClusterRoleBinding` is incorrect.

### How to Reproduce?

With the nightly build:

```
kubectl -n <custom-namespace> apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/refs/tags/nightly/kubernetes/kubernetes.yml
```

This will cause the Operator to receive permissions issues as it can retrieve CRDs as expected as the binding failed.

### Anything else?

_No response_

ryanemerson:
Closes #42678

### DIFF
--- a/operator/src/main/kubernetes/kubernetes.yml
+++ b/operator/src/main/kubernetes/kubernetes.yml
@@ -2,6 +2,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: keycloak-operator-role
+  labels:
+    app.kubernetes.io/name: keycloak-operator
 rules:
   - apiGroups:
       - apps
@@ -88,6 +90,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: keycloak-operator-clusterrole
+  labels:
+    app.kubernetes.io/name: keycloak-operator
 rules:
   - apiGroups:
       - apiextensions.k8s.io
--- a/docs/guides/operator/installation.adoc
+++ b/docs/guides/operator/installation.adoc
@@ -108,14 +108,24 @@ kubectl apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resourc
 kubectl apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/{version}/kubernetes/keycloakrealmimports.k8s.keycloak.org-v1.yml
 ----
 
-. Install the {project_name} Operator deployment by entering the following command:
+. Install the {project_name} Operator deployment in the `keycloak` namespace by executing the following commands:
 +
 [source,bash,subs="attributes+"]
 ----
-kubectl apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/{version}/kubernetes/kubernetes.yml
+kubectl create namespace keycloak
+kubectl -n keycloak apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/{version}/kubernetes/kubernetes.yml
 ----
 
-The Operator will watch the namespace where it is installed. You may optionally select a namespace with the `-n` option.
+The Operator will watch the namespace where it is installed. You may utilise a different namespace with the `-n` option,
+however you must also update the `ClusterRoleBinding` subject. For example, to install in the namespace `custom-namespace`,
+execute the following commands:
++
+[source,bash,subs="attributes+"]
+----
+kubectl create namespace custom-namespace
+kubectl -n custom-namespace apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/{version}/kubernetes/kubernetes.yml
+kubectl patch clusterrolebinding keycloak-operator-clusterrole-binding --type='json' -p='[{"op": "replace", "path": "/subjects/0/namespace", "value":"custom-namespace"}]'
+----
 </@profile.ifCommunity>
 
 === Installing Multiple Operators


operator/src/main/kubernetes/kubernetes.yml
```diff
@@ -129,4 +133,3 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: keycloak-operator
-    namespace: keycloak
```
ahus1: Just to understand what happens when removing the namespace: Will it then use the namespace associated with the default namespace this YAML is deployed to?

ryanemerson: Argh no, unfortunately not! I was thinking that was the case, but the namespace is a required value and apply will fail without it `subjects[0].namespace: Required value`.

If we utilise a `RoleBinding` for the `ClusterRole`, it will be per namespace. Let me verify on Openshift and I will update the PR.

shawkins: Unfortunately this namespace is intentional and also plays into how the operatorsdk functions. It is compensated for in the BaseOperatorTest logic and https://github.com/keycloak/keycloak/blob/main/operator/overlays/default-namespace/kustomization.yaml that is referenced in the README.md.

Adding labels is certainly fine.

ryanemerson: I believe the current https://www.keycloak.org/operator/installation instructions are broken then. They show how to install the Operator in the default namespace, but they don't specify that this namespace field needs to be updated, therefore the ServiceAccount will not get the expected permissions as the `ClusterRoleBinding` subject is defined for the `keycloak` namespace.

I suggest that we change the namespace of the `ClusterRoleBinding` subject to `default` and update the docs to state that the namespace value must be changed if deploying to a non-default namespace. I can also include an example patch command showing how to do this.

@shawkins  WDYT?

shawkins: @ryanemerson are you referring to https://www.keycloak.org/operator/installation#_installing_by_using_kubectl_without_operator_lifecycle_manager - that is correct, I don't believe that was updated to reflect the README.

> I suggest that we change the namespace of the ClusterRoleBinding subject to default and update the docs to state that the namespace value must be changed if deploying to a non-default namespace. I can also include an example patch command showing how to do this.

You'd at least have to update the README, and possibly other places that are assuming the keycloak namespace by default.

ryanemerson: Sure, I don't mind doing that. I think the most important thing is that our user facing guide is correct.

ryanemerson: Alternatively we update the guide to always install to the `keycloak` namespace and let users know they must patch the `ClusterRoleBinding` if they want to use a different namespace?

shawkins: > Alternatively we update the guide to always install to the keycloak namespace and let users know they must patch the ClusterRoleBinding if they want to use a different namespace?

Yes it can adapt the instructions in the readme: https://github.com/keycloak/keycloak/blob/39222e8ca50aec053b85ce10aae39aaf1bf69bac/operator/README.md?plain=1#L69-L76

ryanemerson: The kustomize files aren't available at https://github.com/keycloak/keycloak-k8s-resources, so I have updated the guide to use the patch command:

```
kubectl patch clusterrolebinding keycloak-operator-clusterrole-binding --type='json' -p='[{"op": "replace", "path": "/subjects/0/namespace", "value":"custom-namespace"}]'
```

shawkins (approved):
LGTM, thank you for addressing this @ryanemerson

vmuzikar (approved):

anlesk:
Hi @ryanemerson @vmuzikar
When Installing by using kubectl without Operator Lifecycle Manager, this CRD `https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/26.6.3/kubernetes/kubernetes.yml` contains ClusterRoleBinding
```apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  labels:
    app.kubernetes.io/name: keycloak-operator
  name: keycloak-operator-clusterrole-binding
roleRef:
  kind: ClusterRole
  apiGroup: rbac.authorization.k8s.io
  name: keycloak-operator-clusterrole
subjects:
  - kind: ServiceAccount
    name: keycloak-operator
    namespace: keycloak
```
Is that ClusterRoleBinding required? In my set up there are multiple namespaces within cluster and each namespace has its own keycloak-operator - ideally, I want those to be fully separated / independent. Is there a way to achieve that without specifying all the SA/namespaces under that cluster role?

shawkins:
> Is that ClusterRoleBinding required? In my set up there are multiple namespaces within cluster and each namespace has its own keycloak-operator - ideally, I want those to be fully separated / independent. Is there a way to achieve that without specifying all the SA/namespaces under that cluster role?

@anlesk  It is not required. It is needed only for openshift installs where the keycloak cr does not specify hostnames.

PR #42682, opened 2025-09-17
Use back keycloak-js instead of initiate login in the backend for account (26.0)

### LINKED ISSUE #40463: Login to Account Console produces two consecutive LOGIN events

rmartinc:
Closes #40463

PR:             https://github.com/keycloak/keycloak/pull/42035
Commit:         https://github.com/keycloak/keycloak/commit/360ff7050c290939d529e68b461ba61c7c11404a
PR branch:      backport-42035-26.0
Target branch:  https://github.com/keycloak/keycloak/tree/release/26.0

Cherry pick for 26.0.

### DIFF
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -3295,3 +3295,21 @@ deleteConfirmUsers_one=Delete user {{name}}?
 deleteConfirmUsers_other=Delete {{count}} users?
 signatureMaxExp=Max expiration
 signatureMaxExpHelp=Maximum expiration allowed for the JWT. Tokens need to be generated right before authentication. After this period will be considered invalid because they are too old. If undefined the default value is 60 seconds.
+# standard error responses OAuth
+invalid_request=Invalid request
+unauthorized_client=Unauthorized client
+access_denied=Access denied
+unsupported_response_type=Unsupported response type
+invalid_scope=Invalid scope
+server_error=Server error
+temporarily_unavailable=Temporarily unavailable
+# standard error responses OIDC
+interaction_required=Interaction required
+login_required=Login required
+account_selection_required=Account selection required
+consent_required=Consent required
+invalid_request_uri=Invalid request uri
+invalid_request_object=Invalid request object
+request_not_supported=Request not supported
+request_uri_not_supported=Request uri not supported
+registration_not_supported=Registration not supported
--- a/js/libs/ui-shared/src/context/ErrorPage.tsx
+++ b/js/libs/ui-shared/src/context/ErrorPage.tsx
@@ -5,18 +5,19 @@ import {
   Page,
   Text,
   TextContent,
-  TextVariants,
 } from "@patternfly/react-core";
 import { useTranslation } from "react-i18next";
+import { getNetworkErrorMessage } from "../utils/errors";
 
 type ErrorPageProps = {
   error?: unknown;
 };
 
 export const ErrorPage = (props: ErrorPageProps) => {
-  const { t } = useTranslation();
+  const { t, i18n } = useTranslation();
   const error = props.error;
   const errorMessage = getErrorMessage(error);
+  const networkErrorMessage = getNetworkErrorMessage(error);
 
   function onRetry() {
     location.href = location.origin + location.pathname;
@@ -37,9 +38,12 @@ export const ErrorPage = (props: ErrorPageProps) => {
         ]}
       >
         <TextContent>
-          <Text>{t("somethingWentWrongDescription")}</Text>
-          {errorMessage && (
-            <Text component={TextVariants.small}>{errorMessage}</Text>
+          {errorMessage ? (
+            <Text>{t(errorMessage)}</Text>
+          ) : networkErrorMessage && i18n.exists(networkErrorMessage) ? (
+            <Text>{t(networkErrorMessage)}</Text>
+          ) : (
+            <Text>{t("somethingWentWrongDescription")}</Text>
           )}
         </TextContent>
       </Modal>
@@ -48,10 +52,6 @@ export const ErrorPage = (props: ErrorPageProps) => {
 };
 
 function getErrorMessage(error: unknown): string | null {
-  if (typeof error === "string") {
-    return error;
-  }
-
   if (error instanceof Error) {
     return error.message;
   }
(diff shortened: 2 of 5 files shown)


js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -3295,3 +3295,22 @@ deleteConfirmUsers_one=Delete user {{name}}?
...
+interaction_required=Interaction required
+login_required=Login required
+account_selection_required=Account selection required
+consent_required=Consent required
+invalid_request_uri=Invalid request uri
+invalid_request_object=Invalid request object
+request_not_supported=Request not supported
+request_uri_not_supported=Request uri not supported
+registration_not_supported=Registration not supported
+>>>>>>> 360ff7050c (Use back keycloak-js instead of initiate login in the backend for account (#42035))
```
mposolda: It seems this line should not be here?

mposolda (commented):
Added one comment inline.

Also not 100% sure if all the messages in `messages_en.properties` are needed, but don't have issues with including them, so probably fine.

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
rmartinc: Argh! Yes, I didn't remove after solving the conflict. Done!

rmartinc:
Done both! And yes, the properties were not needed, I didn't detect there were two more that are not needed.

rmartinc:
It seems that is finally not needed. Closing it.

rmartinc:
Re-opening this for 26.0.

ahus1 (approved):

ahus1:
@rmartinc - this is now waiting for a UI maintainer to approve. 
@ssilvert - can you please approve?

rmartinc:
Thanks @ahus1! @ssilvert This is the issue we want to also move to 26.0 now we are going to tag a new 26.0.17. The same change is done in 26.2 and 26.4.

js/libs/ui-shared/src/context/ErrorPage.tsx
```diff
@@ -5,18 +5,20 @@ import {
...
   error?: unknown;
 };
 
 export const ErrorPage = (props: ErrorPageProps) => {
-  const { t } = useTranslation();
+  const { t, i18n } = useTranslation();
   const error = props.error;
   const errorMessage = getErrorMessage(error);
+  const networkErrorMessage = getNetworkErrorMessage(error);
+  console.error(error);
```
ssilvert: Let's try to keep the console uncluttered.  I'd remove this line.

js/libs/ui-shared/src/context/ErrorPage.tsx
```diff
@@ -48,10 +53,6 @@ export const ErrorPage = (props: ErrorPageProps) => {
 };
 
 function getErrorMessage(error: unknown): string | null {
-  if (typeof error === "string") {
-    return error;
-  }
-
```
ssilvert, line 54: Are you 100% sure that error will never be a string?

ssilvert (changes requested):

js/libs/ui-shared/src/context/ErrorPage.tsx
rmartinc: Done! But this log is in main and it was there before my changes. It was just added by the cherry-pick.  :smile:

js/libs/ui-shared/src/context/ErrorPage.tsx
rmartinc: This error is always the error thrown by keycloak-js [here](https://github.com/keycloak/keycloak/blob/release/26.0/js/libs/ui-shared/src/context/KeycloakContext.tsx#L77). The keycloak-js always returns an Error for throws. If you want I can maintain the string (no problem with that). In main PR this was clearer, because the error passed here was the Error or the string obtained using another method.

js/libs/ui-shared/src/context/ErrorPage.tsx
mposolda: @ssilvert @rmartinc Not sure how to proceed when there is a review comment for backporting something, which was already merged into `main` . Do we need to create another issue for `main` to remove the `console.log` line from there then?

mposolda:
@ssilvert Re-assigning to you since you had some review comments. Feel free to merge whenever this looks OK to you.

@ssilvert @rmartinc Also added inline comment about `console.log` . Not sure how to proceed as this PR is just backport and that line is already present in `main` .

js/libs/ui-shared/src/context/ErrorPage.tsx
rmartinc: @ssilvert If you want I can re-add these lines, just to move this forward. I think this PR is needed for 26.0.17.

ssilvert: @rmartinc If you are sure it's OK, I'm fine with it.

ssilvert (approved):

PR #42700, opened 2025-09-17
Allow target attribute for anchor tags in html-sanitizer

### LINKED ISSUE #28846: Allow the target attribute on <a> in the kcSanitize
### Description

I'm using the `kcSanitize` method to write down a piece of html in my theme that is meant to redirect the user to our terms on a new page / tab.

in the message.properties:
```
acceptTerms=I agree to the <a href="https://www....." target="_blank">Terms of Service agreement </a>
```

and in my template: 

`<div> ${kcSanitize(msg("acceptTerms"))?no_esc} <div>`

Unfortunately, I think that the HtmlPolicy defined org.keycloak.theme.KeycloakSanitizerPolicy#POLICY_DEFINITION does not allow the a tag to have an attribute of type target.

Is there a security reason that I'm missing for that or would it be possible to add `.onElements("a").allowAttributes("target")` in the html policy definition ?

If yes, I'll be happy to send the PR

Let me know

### Discussion

_No response_

### Motivation

Allow link in the theme that open on a different target.

### Details

_No response_

rr-it:
`kcSanitize` allows for a `target` attribute in anchor tags.

Especially on the registration or accept terms of service page the links to terms of service must open in a new window. If the link opens in the same window, the user is thrown out of the registration or login process.

Therefore the html-sanitizer must not remove the `target` attribute from anchor tags like:
`<a href="…" target="_blank">…</a>`

This fixes https://github.com/keycloak/keycloak/issues/28846

### DIFF
--- a/services/src/main/java/org/keycloak/theme/KeycloakSanitizerPolicy.java
+++ b/services/src/main/java/org/keycloak/theme/KeycloakSanitizerPolicy.java
@@ -62,6 +62,8 @@ public class KeycloakSanitizerPolicy {
 
   private static final Pattern NAME = Pattern.compile("[a-zA-Z0-9\\-_\\$]+");
 
+  private static final Pattern TARGET = Pattern.compile("_blank");
+
   private static final Pattern ALIGN = Pattern.compile(
       "(?i)center|left|right|justify|char");
 
@@ -102,6 +104,7 @@ public class KeycloakSanitizerPolicy {
           .allowStandardUrlProtocols()
           .allowAttributes("nohref").onElements("a")
           .allowAttributes("name").matching(NAME).onElements("a")
+          .allowAttributes("target").matching(TARGET).onElements("a")
           .allowAttributes(
               "onfocus", "onblur", "onclick", "onmousedown", "onmouseup")
               .matching(HISTORY_BACK).onElements("a")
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/keycloak/theme/KeycloakSanitizerPolicy.java
```diff
@@ -102,6 +102,7 @@ public class KeycloakSanitizerPolicy {
           .allowStandardUrlProtocols()
           .allowAttributes("nohref").onElements("a")
           .allowAttributes("name").matching(NAME).onElements("a")
+          .allowAttributes("target").matching(NAME).onElements("a")
```
ahus1, line 105: I would like to be more restrictive here and not just allow any target, but only `_blank`. This is the most common case that is being asked for, and also avoid opening content in an already named window.

When you add a change to allow only `_blank`, please add also a test that shows that other targets are not allowed.

ahus1 (changes requested):
Thank you for this PR for this issue that has several upvotes. Please see below for a suggested change.

ahus1 (approved):
Thank you for the PR. I slightly adjusted the test case, as the sanitation library will only add the additional `rel` values when the target is accepted.

ahus1:
Waiting for #42752 to be fixed first

PR #42701, opened 2025-09-17
fix: improving the local apiserver test cleanup

### LINKED ISSUE #42693: Improve local apiserver test cleanup
### Description

The test cleanup for the apiserver test seems to sometimes cause confusion in the newly started operator as to the namespace / client in use.

shawkins:
In addition to more logically organizing the stop/start logic, this ensures that the operator sdk won't call System.exit in any case, and that the client instance is correctly scoped to the operator instance.

At least on CI the addition of the servicemonitor reconcilier seems to have made it much more likely that operator.stop will result in a System.exit call - this appears to be a defficiency in the operator sdk stop logic.

closes: #42693


vmuzikar (approved):
LGTM, thanks.

PR #42708, opened 2025-09-17
Lazy load client sessions

### LINKED ISSUE #42628: Lazy load client sessions
### Description

The method `JpaUserSessionPersisterProvider#loadUserSession(...)` eagerly queries and loads all the client sessions associated with the user session (see below). 

https://github.com/keycloak/keycloak/blob/6a202146b4e6293ba20e086adcebbb61955717f6/model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java#L303-L342

Some operations do not require accessing any client session, making this code suboptimal for those cases. We can improve the code by lazy-loading the client sessions when requested. 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pruivo:
Closes #42628

I split the method `loadUserSessionsWithClientSessions(TypedQuery<PersistentUserSessionEntity> query, String offlineStr, boolean useExact)` as it has no common code between `useExact=true` and `useExact=false`.

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/session/ClientSessionLoader.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/session/ClientSessionLoader.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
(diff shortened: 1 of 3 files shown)


ahus1 (approved):
Thank you for this change!

PR #42725, opened 2025-09-18
Place EdECUtilsImpl.java in the normal source folder

### LINKED ISSUE #42716: The core class EdECUtilsImpl is not present in the sources jar
### Before reporting an issue

### Area

core

### Describe the bug

Because of #34630 the class `EdECUtilsImpl` class was created in a `src/main/java15` directory that is not included in the sources jar. Because of this the *keycloak-clients* cannot sync it to be included in the `keycloak-client-common-synced` package.

```
jar tvf target/keycloak-core-999.0.0-SNAPSHOT-sources.jar | grep EdECUtils
  1403 Wed Sep 17 14:44:44 CEST 2025 org/keycloak/jose/jwk/EdECUtilsUnsupportedImpl.java
  1081 Wed Sep 17 14:44:44 CEST 2025 org/keycloak/jose/jwk/EdECUtils.java
```

### Version

26.3.4

### Regression

### Expected behavior

The class should be present in the sources distribution jar as any other class.

### Actual behavior

The class is not present.

### How to Reproduce?

Just compile the core package and check the class is not present in the sources jar file. You can also check that this class is not distributed in the current `keycloak-core-26.3.4-sources.jar` file in maven.

### Anything else?

_No response_

rmartinc:
Closes #42716

@mposolda The PR just moves the file from the specific `java15` to the normal sources folder. As the package is not a multi-release now this file is not included in the sources jar. With this change the normal compilation just excludes this class (all classes except this one are compiled with java 8) and the extra jdk-15 compilation just includes this file (this one is compiled with release 15). This way the output is the same but the file is placed in the normal sources directory `src/main/java` and included in the jar. We can unpack it later and do the same trick for compilation later in the `keycloak-client` project.


mposolda (approved):
@rmartinc Thanks!

PR #42744, opened 2025-09-18
Improve the Workflow JSON schema

### LINKED ISSUE #42697: [RLM] - Improve the Workflow JSON schema
### Description

There are improvements to be made to the Workflow schema to make it simpler and readable. Today, it looks like:

```
{
  "providerId": "event-based-resource-policy",
  "config": {
    "events": ["LOGIN"]
  },
  "conditions": [
    {
      "providerId": "group-membership-condition",
      "config": {
        "groups": ["30f31e33-b116-4420-aba8-bbc8678f4bfb"]
      }
    }
  ],
  "actions": [
    {
      "providerId": "notify-user-action-provider",
      "config": {
        "after": ["20000"]
      }
    }
  ]
}
```

However, we can make it cleaner with a format like this:

```
{
  "id": "workflow-id",
  "name": "my-name",
  "uses": "my-provider",
  "enabled": true,
  "on": "event",
  "reset-on": [
    "event-reset-1",
    "event-reset-2"
  ],
  "scheduled": true,
  "recurring": true,
  "if": [
    {
      "uses": "condition-1",
      "with": {
        "key1": "v1",
        "key2": [
          "v1",
          "v2"
        ]
      }
    },
    {
      "uses": "condition-2",
      "with": {
        "key1": "v1",
        "key2": [
          "v1",
          "v2"
        ]
      }
    },
    {
      "uses": "condition-1",
      "with": {
        "key1": "v1",
        "key2": [
          "v1",
          "v2",
          "v3"
        ]
      }
    }
  ],
  "steps": [
    {
      "id": "1",
      "uses": "step-1",
      "after": "10000",
      "with": {
        "key1": "v1"
      }
    },
    {
      "id": "2",
      "uses": "step-2",
      "with": {
        "key1": [
          "v1",
          "v2"
        ]
      }
    },
    {
      "id": "3",
      "uses": "step-1",
      "with": {
        "key1": "v1"
      }
    }
  ],
  "state": {
    "errors": [
      "error1",
      "error2"
    ]
  }
}
```

And a schema similar to:

```
{
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "name": {
      "type": "string"
    },
    "uses": {
      "type": "string"
    },
    "enabled": {
      "type": "boolean"
    },
    "on": {
      "type": "array",
      "items": {
        "type": "string"
      }
    },
    "reset-on": {
      "type": "array",
      "items": {
        "type": "string"
      }
    },
    "scheduled": {
      "type": "boolean"
    },
    "recurring": {
      "type": "boolean"
    },
    "if": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "uses": {
            "type": "string"
          },
          "with": {
            "type": "object",
            "additionalProperties": true
          }
        },
        "required": [
          "uses"
        ]
      }
    },
    "steps": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string"
          },
          "uses": {
            "type": "string"
          },
          "after": {
            "type": "string"
          },
          "with": {
            "type": "object",
            "additionalProperties": true
          }
        },
        "required": [
          "uses"
        ]
      }
    },
    "state": {
      "type": "object",
      "properties": {
        "errors": {
          "type": "array",
          "items": {
            "type": "string"
          }
        }
      },
      "required": [
        "errors"
      ]
    }
  },
  "required": [
    "name"
  ]
}
```

### Discussion

_No response_

### Motivation

Improve UX when creating workflow definitions.

### Details

_No response_

pedroigor:
Closes #42697

* The Workflow JSON Schema was updated based on the proposal from the linked issue
* Renamed the provider factory IDs for steps, actions, and conditions to a more user-friendly name. Any suggestion is welcome.
* It is no longer mandatory to set a workflow (provider id) because it will default to the generic/event-based one
* Adds validation for events and fields that reference components in the backend such as events, conditions, and actions.
* There is still room for improvement, like allowing setting the value of `after` for steps using a format like `12d`, `30m`, etc. We can do it as a follow-up. It is still in milliseconds.

At the very minimum, a workflow definition looks like:

```
{
    "on-event": "USER_LOGIN",
    "steps": {
        "notify-user": {
            "after": "10000"
        },
        "disable-user": {
            "after": "5000"
        },
        "notify-user": {
            "after": "5100"
        }
    }
}
```

I'm not sure about using duplicate field names (e.g.: `notify-user`) above because it is not common and something that the JSON specification discourages. It is not a good practice. For instance, `kcadm` won't parse the response correctly and will not show duplicate fields.

As an alternative, we could have something like:

```
{
  "on-event": "USER_LOGIN",
  "steps": {
    "notify-user-before-delete": {
      "step": "notify-user",
      "after": "10000"
    },
    "disable-user": {
      "step": "disable-user",
      "after": "5000"
    },
    "notify-user-after-delete": {
      "step": "notify-user",
      "after": "5100"
    }
  }
}
```

From a readability perspective, the first one (and herein implemented) reads better, but it is uncommon.

### DIFF
--- a/common/src/main/java/org/keycloak/common/util/reflections/Reflections.java
+++ b/common/src/main/java/org/keycloak/common/util/reflections/Reflections.java
@@ -18,6 +18,7 @@
 package org.keycloak.common.util.reflections;
 
 import java.beans.Introspector;
+import java.io.IOException;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AccessibleObject;
@@ -1050,5 +1051,39 @@ public class Reflections {
         }
 
         return Object.class;
+    }
+
+    public static <T> T convertValueToType(Object value, Class<T> type) {
+
+        if (value == null) {
+            return null;
+
+        } else if (value instanceof String) {
+            if (type == String.class) {
+                return type.cast(value);
+            } else if (type == Boolean.class) {
+                return type.cast(Boolean.parseBoolean(value.toString()));
+            } else if (type == Integer.class) {
+                return type.cast(Integer.parseInt(value.toString()));
+            } else if (type == Long.class) {
+                return type.cast(Long.parseLong(value.toString()));
+            }
+        } else if (value instanceof Number) {
+            if (type == Integer.class) {
+                return type.cast(((Number) value).intValue());
+            } else if (type == Long.class) {
+                return type.cast(((Number) value).longValue());
+            } else if (type == String.class) {
+                return type.cast(value.toString());
+            }
+        } else if (value instanceof Boolean) {
+            if (type == Boolean.class) {
+                return type.cast(value);
+            } else if (type == String.class) {
+                return type.cast(value);
+            }
+        }
+
+        throw new RuntimeException("Unable to handle type [" + type + "]");
     }
 }
--- a/core/src/main/java/org/keycloak/representations/workflows/AbstractWorkflowComponentRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/workflows/AbstractWorkflowComponentRepresentation.java
@@ -1,0 +1,138 @@
+package org.keycloak.representations.workflows;
+
+import static org.keycloak.common.util.reflections.Reflections.isArrayType;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.keycloak.common.util.MultivaluedHashMap;
+import org.keycloak.common.util.reflections.Reflections;
+
+public abstract class AbstractWorkflowComponentRepresentation {
+
+    private String id;
+    private String uses;
+
+    @JsonProperty("with")
+    private MultivaluedHashMap<String, String> config;
+
+    public AbstractWorkflowComponentRepresentation(String id, String uses, MultivaluedHashMap<String, String> config) {
+        this.id = id;
+        this.uses = uses;
+        this.config = config;
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getUses() {
+        return this.uses;
+    }
+
+    public void setUses(String uses) {
+        this.uses = uses;
+    }
+
+    public MultivaluedHashMap<String, String> getConfig() {
+        return config;
+    }
+
+    public void setConfig(MultivaluedHashMap<String, String> config) {
+        if (this.config == null) {
+            this.config = config;
+        }
+        this.config.putAll(config);
+    }
+
+    public void setConfig(String key, String value) {
+        setConfig(key, Collections.singletonList(value));
+    }
+
+    @JsonAnySetter
+    @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+    public void setConfig(String key, List<String> values) {
+        if (this.config == null) {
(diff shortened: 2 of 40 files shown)


stianst:
I think it's a lot better than what there is currently, but think steps should be an array, and not an object.

Couple other things that would be good:

* `after` is confusing IMO, maybe `with-delay` is better?
* `after` should support time-units for readability (`30 days`, `720 hours`, `43200 minutes`, `2592200 seconds`), simple to implement just use `TimeUnit.valueOf(name)`

What about something like:

```
{
    "on-event": "USER_LOGIN",
    "steps": [
        {
             "action": "notify-user",
             "with-delay": "30d",
             "config": {
                "message": "${myMessage}"     
             }
        },
        {
            "action": "disable-user",
            "with-delay": "60d"
        }
    ]
}
```

Not sure about `on-event` are we only going to support that one now? How would this look like if you want to use a workflow template?

pedroigor:
> after is confusing IMO, maybe `with-delay` is better?

Not sure. I might be used with `after` because for me it reads "Execute this step *after* this time is reached". What about `run-after` ?

> after should support time-units for readability (30 days, 720 hours, 43200 minutes, 2592200 seconds), simple to implement just use TimeUnit.valueOf(name)

I agree. As I mentioned in the description, perhaps we can do it as a follow-up. If it is OK.

> Not sure about on-event are we only going to support that one now? How would this look like if you want to use a workflow template?

We do support a few more events. The change herein proposed allows you to define a single or multiple events in an array. 

I agree about reverting the definition and using a list for actions and conditions, just like it stands today. One thing, though, that I changed in this PR is to allow defining configuration entries as regular fields rather than using a `config` object. Can we continue with this approach? For instance, this PR allows this:

```
        "notify-user": {
            "after": "5100",
            "custom": "my-custom-value"
        }
```

The drawback of this approach is that in case we add more built-in config options, we might clash with names that people are using ....

stianst:
```
id: my_workflow
name: My workflow
inputs: mygroup
on: user_login
if:
  - has-group: ${{ inputs.mygroup:mygroup/foobar }}
  - has-idp: myidp
  - uses: myCondition
    with: 
       foo: bar
steps:
  - uses: notify-user
    with:
       template: ${{ myTemplate }}
    after: 40 days
  - uses: disable-user
    if: 
      - not-in-group: mygroup/keep
    after: 30 days
```

An example using a template (or whatever they are called):

```
name: My other workflow
uses: my_workflow
with:
  mygroup: myother/group
```

pedroigor:
@sguilhen @stianst The PR is updated with the suggestions. The remaining points that are going to be handled as a follow-up are:

* Removing the `recurring` field in favor of a `restart` step
* Removing the `scheduled` field in favor of automatically deciding if a workflow is scheduled|immediate or not, depending on the time set to its actions
* Removing the `reset-on` in favor of a specific enum field whose values will indicate how a workflow should react to new events if it is already running

sguilhen:
On the date format for scheduled tasks: Would it be enough to simply use the values available in TimeUnit  - meaning you have to specify the time using a single value (days, hours, minutes, etc)  - or do we want to be more granular and allow something like "10d 12h"?

The latter could be specified in hours only (24*10 + 12 = 252 hours), of course, but it might not be as readable. I'm fine either way and perhaps keeping it simple with what is available in TimeUnit is enough.

pedroigor:
> On the date format for scheduled tasks: Would it be enough to simply use the values available in TimeUnit - meaning you have to specify the time using a single value (days, hours, minutes, etc) - or do we want to be more granular and allow something like "10d 12h"?
> 
> The latter could be specified in hours only (24*10 + 12 = 252 hours), of course, but it might not be as readable. I'm fine either way and perhaps keeping it simple with what is available in TimeUnit is enough.

As we discussed before, we should try something like "1d|1h|30s", at least initially. I remember seeing somewhere in the code base where we are using this.

Let's do it as a follow-up.

ahus1 (approved):
Approving based on @sguilhen's review

PR #42758, opened 2025-09-19
[OID4VCI] filter for asymmetric keys

wistefan:
Closes [#42755]

### DIFF
--- a/services/src/main/java/org/keycloak/crypto/CryptoUtils.java
+++ b/services/src/main/java/org/keycloak/crypto/CryptoUtils.java
@@ -20,6 +20,9 @@ package org.keycloak.crypto;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.provider.ProviderFactory;
 
+import java.security.Key;
+import java.security.PrivateKey;
+import java.security.PublicKey;
 import java.util.AbstractMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


Ogenbertrand (approved):
LGTM !
Thanks @wistefan for this fix.

tnorimat (approved):
LGTM

services/src/main/java/org/keycloak/crypto/CryptoUtils.java
```diff
@@ -20,33 +20,59 @@
...
-    }
+	/**
+	 * Returns the supported asymmetric signature algorithms.
+	 * This method discovers all available SignatureProvider implementations and filters
+	 * for those that support asymmetric algorithms (RSA, EC, EdDSA, etc.).
+	 *
+	 * @param session The Keycloak session
+	 * @return List of asymmetric signature algorithm names
+	 */
+	public static List<String> getSupportedAsymmetricSignatureAlgorithms(KeycloakSession session) {
```
mposolda: It seems that there is no real change in this method in your PR, but it is shown as changed. Can you please remove this change from your PR?

I think the problem is, that you updated this to use tabs instead of 4 spaces. Please note that it is recommended in the Keycloak codebase to use spaces instead of tabs and use the align of 4 spaces. I suggest you update your IDE to use 4 spaces instead of tabs. This should remove the changes of the method `getSupportedAsymmetricSignatureAlgorithms` from your PR and keep just new method `getSupportedAsymmetricEncryptionAlgorithms`

mposolda (commented):
@wistefan Thanks for the PR! I've added one comment inline regarding the tabs VS spaces. Can you please check it?

services/src/main/java/org/keycloak/crypto/CryptoUtils.java
wistefan: Thanks for pointing that out. It seems that an update killed my IDE defaults:) Fixed it.

mposolda: @wistefan Much better. But the new test class `OID4VCIWellKnownProviderTest` seems to still use tabs? Would you please mind to update this too?

wistefan: done

mposolda:
@tnorimat @Ogenbertrand Thanks for the reviews!

@wistefan I've rebased your remote branch on top of latest main (I was able to do that from Github UI). Hopefully that helps with the compilation failure (which was fixed by other PR in the meantime) and tests are also good.

PR #42761, opened 2025-09-19
Allow EdDSA keys in JWTClientCredentialsProvider

### LINKED ISSUE #42751: Allow EdDSA keys in the JWTClientCredentialsProvider to authenticate clients
### Description

Working in keycloak-client issue https://github.com/keycloak/keycloak-client/issues/184 I detected that the clients (for example authz) cannot authenticate using a EdDSA key (just RSA and EC are allowed as asymmetric keys). It's a little change that allows us to test completely EdDSA keys in the keycloak-client project. Adding also here in keycloak to make this available after sync in the client project.

### Discussion

_No response_

### Motivation

Issue https://github.com/keycloak/keycloak-client/issues/184 and it soes no harm adding EdDSA for authentication.

### Details

_No response_

rmartinc:
Closes #42751

Adding support of EdDSA keys in the JWTClientCredentialsProvider. Although this is mainly to test issue https://github.com/keycloak/keycloak-client/issues/184 in `keycloak-clients`, it's also useful for the corepackage. For example the `authz-client` can now use EdDSA keys to authenticate using a signed JWT.

 Adding a test that uses Ed25519 to authenticate using a JWT token.


services/src/main/java/org/keycloak/keys/loader/ClientPublicKeyLoader.java
```diff
@@ -112,16 +113,18 @@ private static KeyWrapper getSignatureValidationKey(CertificateRepresentation ce
             kid = certInfo.getKid() != null ? certInfo.getKid() : KeyUtils.createKeyId(clientCert.getPublicKey());
             keyWrapper.setKid(kid);
             keyWrapper.setPublicKey(clientCert.getPublicKey());
-            keyWrapper.setType(clientCert.getPublicKey().getAlgorithm());
+            keyWrapper.setType(JavaAlgorithm.getKeyType(clientCert.getPublicKey().getAlgorithm()));
             keyWrapper.setCertificate(clientCert);
             keyWrapper.setIsDefaultClientCertificate(true);
+            keyWrapper.setCurve(clientCert.getPublicKey().getAlgorithm());
```
tnorimat: It seems that `setCurve` makes sense if the key is regarding ECDSA or EdDSA.
In case of the key regarding RSA-SSA (RS, PS), it sets the value "RSA". If it does not cause a side-effect, ther is no problem. If causes, need to do some appropriate processing.

services/src/main/java/org/keycloak/keys/loader/ClientPublicKeyLoader.java
```diff
@@ -112,16 +113,18 @@ private static KeyWrapper getSignatureValidationKey(CertificateRepresentation ce
...
         } else {
             PublicKey publicKey = KeycloakModelUtils.getPublicKey(encodedPublicKey);
             // Check if we have kid in DB, generate otherwise
             kid = certInfo.getKid() != null ? certInfo.getKid() : KeyUtils.createKeyId(publicKey);
             keyWrapper.setKid(kid);
             keyWrapper.setPublicKey(publicKey);
-            keyWrapper.setType(publicKey.getAlgorithm());
+            keyWrapper.setType(JavaAlgorithm.getKeyType(publicKey.getAlgorithm()));
+            keyWrapper.setCurve(publicKey.getAlgorithm());
```
tnorimat: It seems that `setCurve` makes sense if the key is regarding ECDSA or EdDSA.
In case of the key regarding RSA-SSA (RS, PS), it sets the value "RSA". If it does not cause a side-effect, ther is no problem. If causes, need to do some appropriate processing.

tnorimat (commented):
@rmartinc Hello, I added some reivew comments. I appreciate it if you would check them.

services/src/main/java/org/keycloak/keys/loader/ClientPublicKeyLoader.java
mposolda: Will be good to check this. I hope there is no side-effect with calling `setCurve` always, so we can call it in a way that is proposed in this PR and we don't need another `if` in the codebase to call it just for some algorithms. But leaving to @rmartinc to confirm this.

mposolda (commented):
@rmartinc I am approving this PR. I am waiting before merge for the confirm of the concern by @tnorimat before moving forward. If you confirm, that it is OK as it is in the PR (which I hope), my vote is to merge the PR.

services/src/main/java/org/keycloak/keys/loader/ClientPublicKeyLoader.java
rmartinc: This  `curve` property  was added by EdDSA (it's even not needed for ECDSA). It does no harm adding it for other keys (RSA for example) but we can just add it if the type is OKP (the only one that really needs or defines the curve). I'm sending a new PR now.

rmartinc: Done now! Let's see the tests.

services/src/main/java/org/keycloak/keys/loader/ClientPublicKeyLoader.java
rmartinc: Done!

mposolda (approved):
@rmartinc Ok, Thanks! I hope to merge once the tests are OK

edwint88:
This is very robust and with RSA-PSS/RSASSA-PSS doesn't work. The keyType it defaults to KeyType.OCT instead of at least KeyType.RSA.

PR #42772, opened 2025-09-19
Missing switch 'ID Token as detached signature' in the admin console …

### LINKED ISSUE #42769: Missing switch "ID Token as detached signature" in the admin console client settings
### Before reporting an issue

### Area

admin/ui

### Describe the bug

In the previous KEycloak versions (EG. Keycloak 18 matching to RH-SSO 7.6) in the old admin console there was a switch `Use ID Token as a Detached Signature` with the tooltip `This makes ID token returned from Authorization Endpoint in OIDC Hybrid flow use as a detached signature defined in FAPI 1.0 Advanced Security Profile. Therefore, this ID token does not include an authenticated user's information.`

The switch was within the "Advanced Settings" for OIDC Clients.

Looks like we forgot to add this switch to the new admin console. The switch was probably added to the old admin console around the time when the particular screen was migrated to the new admin console and hence it ended to never being added to the new admin console. We had similar issues before as well for some other switches.

### Version

nightly (from 2025-09-19), 26.2, 26.3

### Regression

### Expected behavior

Switch being available in the admin console

### Actual behavior

Switch not being available in the admin console

### How to Reproduce?

See details

### Anything else?

There is a workaround to use directly admin REST API (or maybe kcadm). The name of related attribute is `id.token.as.detached.signature` .

mposolda:
…client settings

closes #42769


rmartinc (approved):
@mposolda LGTM!

mposolda:
@rmartinc Thanks!

PR #42777, opened 2025-09-19
Session cache affinity

pruivo:
Closes #42776

Please take a look at the GHI description for the drawbacks of this approach. 

- [x] Javadoc!

### DIFF
--- a/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
+++ b/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
@@ -1,13 +1,18 @@
 package org.keycloak.common.util;
 
 import java.security.SecureRandom;
+import java.util.Base64;
 import java.util.UUID;
+import java.util.function.Supplier;
 
 public class SecretGenerator {
 
     public static final int SECRET_LENGTH_256_BITS = 32;
     public static final int SECRET_LENGTH_384_BITS = 48;
     public static final int SECRET_LENGTH_512_BITS = 64;
+    // for auth session and user session; 128 bits entropy + 16 bits for cluster size.
+    private static final int SESSION_ID_BYTES = 18;
+    public static final Supplier<String> SECURE_ID_GENERATOR = () -> getInstance().generateBase64SecureId(SESSION_ID_BYTES);
 
     public static final char[] UPPER = "ABCDEFGHIJKLMNOPQRSTUVWXYZ".toCharArray();
 
@@ -28,6 +33,16 @@ public class SecretGenerator {
     
     public String generateSecureID() {
         return generateSecureUUID().toString();
+    }
+
+    public String generateBase64SecureId(int nBytes) {
+        assert nBytes > 0;
+        byte[] data = new byte[nBytes];
+        SECURE_RANDOM.nextBytes(data);
+        String id = Base64.getUrlEncoder().encodeToString(data);
+        assert !id.contains(".");
+        assert !id.contains(" ");
+        return id;
     }
 
     public String randomString() {
--- a/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
+++ b/common/src/main/java/org/keycloak/common/util/SecretGenerator.java
@@ -13,7 +13,7 @@ public class SecretGenerator {
     /**
      * Session ID length in bytes.
      * <p />
-     * Both NIST and ANSSI ask for at least 128 bits of entropy in , see <a href="https://github.com/keycloak/keycloak/issues/38663">#38663</a>
+     * Both NIST and ANSSI ask for at least 128 bits of entropy, see <a href="https://github.com/keycloak/keycloak/issues/38663">#38663</a>.
      * As we are about to filter those session IDs on each node to find a key of the local segment using Infinispan's org.infinispan.affinity.KeyAffinityServiceFactory,
      * we add some more entropy so that the filtering then leaves enough entropy for those IDs.
      * Usually there are 256 segments in a cache. Just in case someone increases it, we add 16 bits.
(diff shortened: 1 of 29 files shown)


ahus1:
Thank you for proposing this change to make it more efficient. 

The way forward:

* For randomness, use 128 bits + 16 bits. This will ensure that each node generates its IDs with a randomness of at least 128 bits even after filtering the hashed values by the segments it owns.
* To prevent longer IDs than before, change to base64 encoding. As this doesn't include a `.`, it will not conflict with any delimiter we use for example in the AUTH_SESSION_ID cookie. In the end, I expect the new IDs in their string representation to be shorter than before. 
* The new algorithm will be the default in KC 26.5. To allow a fallback for those affected by this internal-only change, provide an SPI option to revert to the original behavior. 

Ideas discarded due to my paranoia:

* Create a session ID that has the segment ID at the end, as it would possibly leak information about how many segments there are, how many segments the current node owns, and from that to derive how many Keycloak nodes are running.

pruivo:
@ahus1 looks like all the tests expect a UUID. There is red everywhere with the error: 
```
-- java.lang.AssertionError: 
-- session ID
-- Expected: is Not an UUID
--      but: was "rCHqF+UivVWQElYnVACETbmf"
```

ahus1:
@pruivo - I pushed one commit with a suggestion. Let's see if this makes the results greener and let me know what you think. 

Once we have the results, as as this affects sessions a lot, we would probably need to loop in the core teams into that. 

It is good to see that we're reducing the size of the key from 36 chars to 24 chars, still knowing that we might end up with 27 chars once we add two more bytes.

pruivo:
@ahus1, we can discuss it tomorrow afternoon. Seeing the test suite checking for a valid UUID format caught me off guard and made me wonder if the applications interacting with Keycloak are doing the same. 

PS, we have code to check for UUID too, and it is causing the NPE. See 
https://github.com/keycloak/keycloak/blob/310afe820e12cdfbf4086d3ecf2d55d28d44d412/services/src/main/java/org/keycloak/services/managers/AuthenticationSessionManager.java#L161-L163

test-framework/core/src/main/java/org/keycloak/testframework/events/EventMatchers.java
```diff
@@ -2,16 +2,44 @@
...
+        return Matchers.anyOf(isBase64WithAtLeast128Bits(), isUUID());
+    }
+
+    public static Matcher<String> isBase64WithAtLeast128Bits() {
+        return new TypeSafeMatcher<>() {
+            private static final Pattern BASE64 = Pattern.compile("[-A-Za-z0-9+/_]*");
+
+            @Override
+            protected boolean matchesSafely(String item) {
+                return item.length() >= 24 && item.matches(BASE64.pattern());
```
shawkins, line 33: Is it ok for this to be a weak validation? There are strings that will match this pattern, but aren't valid base64 values.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -36,6 +36,13 @@ the new `UserAuthenticationIdentityProvider` interface (all methods remain the s
 
 Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
 
+=== `session_state` and `sid` are no longer UUIDs
+
+In OpenID connect, there are several places where the protocol shares a `session_state` and a `sid`.
+The specifications define it as an opaque string.
+Previous versions of {project_name} used a UUID for it, while the current version now uses a random base64-encoded string.
+The length of the string was reduced from 36 characters to 24 characters, although it might increase in the future if additional randomness might be required.
```
shawkins:
```suggestion
The length of the string was reduced from 36 characters to 24 characters, although it might increase in the future if additional randomness is required.
```

common/src/main/java/org/keycloak/common/util/SecretGenerator.java
```diff
@@ -1,13 +1,18 @@
...
+import java.util.Base64;
 import java.util.UUID;
+import java.util.function.Supplier;
 
 public class SecretGenerator {
 
     public static final int SECRET_LENGTH_256_BITS = 32;
     public static final int SECRET_LENGTH_384_BITS = 48;
     public static final int SECRET_LENGTH_512_BITS = 64;
+    // for auth session and user session; 128 bits entropy + 16 bits for cluster size.
```
shawkins: This reads a little oddly or isn't clear from the rest of the code. Given the random is generated over the entire length, aren't all of the bits for entropy?

What is the significance of 16 bits - is there a partitioner that is decoding the base64 and using a specific 16 bits to determine the segment?

test-framework/core/src/main/java/org/keycloak/testframework/events/EventMatchers.java
ahus1: From the perspective of OpenID connect, those identifiers are opaque strings, so this is fine. See #38663 for more of a discussion why it is at least 128 bits.

common/src/main/java/org/keycloak/common/util/SecretGenerator.java
ahus1: Yes, there is a partitioner but I advocated against it to expose the segment to the caller. Instead, it hashes the value. This allows showing an opaque value to consumers. 

I've extended the JavaDoc in another commit.

ahus1 (approved):

ahus1:
FYI @keycloak/core-clients - in the ongoing effort the reduce the impact of latency between nodes, this PR now enabled key affinity by default. To retain at least 128 bits of entropy after filtering the keys to the local segments, this PR changes the string identifiers from UUIDs with 128 bits, to base64 with 128+16 bits. Due to the base64-encoding, the resulting string identifier is still smaller.

PR #42780, opened 2025-09-19
Use default locale from realm an intermediate fallback

### LINKED ISSUE #40990: Fallback to English translations for unknown locale despite German being the realm default
### Before reporting an issue

### Area

translations

### Describe the bug

When requesting translations from a realm that has German and English enabled, and German set as the default language, Keycloak unexpectedly falls back to English when an unrecognized locale is used in the request.

Setup:
* Realm has two enabled languages: German (de) and English (en)
* German is configured as the default realm language

When requesting translations:
* ✅ http://127.0.0.1:8080/resources/test/common/de
→ returns expected German translations
* ❌ http://127.0.0.1:8080/resources/test/common/notExistingLanguage
→ returns English translations

### Version

26.3.0

### Regression

### Expected behavior

When an unrecognized or invalid locale is provided in the translation request, the fallback should default to the realm’s default language, which is German in this case.

In other words:
* http://127.0.0.1:8080/resources/test/common/notExistingLanguage
→ should return German translations

### Actual behavior

Despite German being the default realm language, the request with an unknown locale:
* http://127.0.0.1:8080/resources/test/common/notExistingLanguage
→ returns English translations

This suggests that English is incorrectly being used as a hard fallback instead of respecting the configured realm default.

### How to Reproduce?

1. Configure a realm with:
* Enabled languages: German and English
* Default language: German
2. Request translations using a valid language code:
```
http://127.0.0.1:8080/resources/test/common/de
```
→ ✅ Returns German
3. Request translations using an invalid/unknown locale:
```
http://127.0.0.1:8080/resources/test/common/notExistingLanguage
```
→ ❌ Returns English instead of the default German

### Anything else?

This fallback behavior appears to ignore the realm’s default language setting. English should not be assumed as the universal fallback in a multilingual setup. Instead, the fallback should respect the realm configuration.

yanxch:
closes #40990

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -57,6 +57,10 @@ Additionally, both `IdentityProviderModel` and `IdentityProviderRepresentation` 
 configuration like `isHideOnLogin` to be null in order to not include these in Identity Provider types that do
 not need these configurations.
 
+=== Realm default locale attempted for translation fallbacks
+
+When localization for a realm is enabled and a translation for a message key is unavailable for the language the user selected, {project_name} now attempts to find a matching message key with the realm's default locale before defaulting to English.
+
 === SPIFFE Identity Provider configuration changed
 
 The SPIFFE Identity Provider preview feature now uses the `trustDomain` configuration instead of `issuer`. This change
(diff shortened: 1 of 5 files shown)


ahus1 (approved):
Thank you for your contribution. I changed it slightly that after the realm default locale it will still default to English to preserve the behavior people might be used to / depend on.

PR #42789, opened 2025-09-20
Fix duplicate label when using password history

ahus1:
Closes #42736


ahus1:
@rmartinc - can you please have a look at this PR as it fixes a regression in KC 26.3? Thanks!

services/src/main/java/org/keycloak/credential/PasswordCredentialProvider.java
```diff
@@ -127,6 +127,8 @@ public CredentialModel createCredential(RealmModel realm, UserModel user, Passwo
             if (expiredPasswordsPolicyValue > 1 || passwordAgeInDaysPolicy > 0) {
                 oldPassword.setId(null);
                 oldPassword.setType(PasswordCredentialModel.PASSWORD_HISTORY);
+                // Setting the label to "nulL" avoids duplicate label errors
```
rmartinc, line 130: Small typo `nulL` but it's OK.

rmartinc (approved):
Thanks @ahus1! LGTM! There is a small typo but approved.

stianst (approved):

PR #42798, opened 2025-09-21
Stabilize ResourceLoaderTest on Windows

### LINKED ISSUE #42795: [Windows] Block Encoded Directory Traversal in ResourceLoader
### Before reporting an issue

### Area

testsuite

### Describe the bug

Previously, the `ResourceLoader` class did not properly block all encoded directory traversal attempts, and the test `ResourceLoaderTest.testFiles` failed on Windows with an `IllegalArgumentException: 'other' has different root` due to path relativization between different drive roots. Additionally, `ResourceLoaderTest.testResource` failed due to improper handling of resource normalization and traversal.

```
Error:  org.keycloak.theme.ResourceLoaderTest.testFiles -- Time elapsed: 0.016 s <<< ERROR!
java.lang.IllegalArgumentException: 'other' has different root
	at java.base/sun.nio.fs.WindowsPath.relativize(WindowsPath.java:420)
	at java.base/sun.nio.fs.WindowsPath.relativize(WindowsPath.java:42)
	at org.keycloak.theme.ResourceLoaderTest.testFiles(ResourceLoaderTest.java:54)

Error:  org.keycloak.theme.ResourceLoaderTest.testResource -- Time elapsed: 0 s <<< FAILURE!
java.lang.AssertionError: expected null, but was:<java.io.BufferedInputStream@2140582>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotNull(Assert.java:756)
	at org.junit.Assert.assertNull(Assert.java:738)
	at org.junit.Assert.assertNull(Assert.java:748)
	at org.keycloak.theme.ResourceLoaderTest.assertResourceAsStream(ResourceLoaderTest.java:63)
	at org.keycloak.theme.ResourceLoaderTest.testResource(ResourceLoaderTest.java:24)
```

### Version

nightly

### Regression

### Expected behavior

All tests passing.

### Actual behavior

The tests fail

### How to Reproduce?

Run the tests on Windows

### Anything else?

_No response_

Pepo48:
Closes: #42795


services/src/main/java/org/keycloak/theme/ResourceLoader.java
```diff
@@ -12,19 +12,40 @@ public static InputStream getResourceAsStream(String root, String resource) thro
...
-                resource = resourcePath.toString().substring(1);
-            } else {
-                resource = resourcePath.toString().substring(2).replace('\\', '/');
-            }
-            URL url = classLoader().getResource(resource);
-            return url != null ? url.openStream() : null;
-        } else {
+        
+        // Check for encoded traversal patterns in the original resource
+        if (containsEncodedTraversal(resource)) {
```
shawkins: Do we expect any URL encoding in the resource string? If not, shouldn't we replace \\ with /, split on / and replace the inbetween parts with URL encoding? That should always preserve the users intention, and prevent the encoded traversal.

Pepo48: From what I know, we don't expect URL-encoded chars in the resource string under normal circumstances. The only way URL encoding could appear there is in an attempt to bypass path checks (as an attack vector).

But, I wasn't sure myself about the approach, thus it's a draft, as I wanted to discuss it first.

Here's the test run, where all of the fixes from the parent task are applied in an attempt to make the Base UT on Windows fully stable:
https://github.com/Pepo48/keycloak/actions/runs/17893148273/job/50876298022

Pepo48: *by normal, legitimate usage I mean:
`resources/css/styles.css`
while malicious attempt could be something like:
`resources/..%2F..%2F..%2Fetc%2Fpasswd`

At least that was my initial train of thoughts. But I probably know where you are heading with a suggestion - if a user creates a theme that for example contains a space, it can cause an issue.

I will adjust the code and rerun it to see how the tests behave.

Pepo48:
@shawkins I simplified it. Essentially we want to tackle two issues here:
* in `ResourceLoader` I made sure that the `getResourceAsStream` uses relative paths and fixes Windows traversal bypass due to absolute path issue - in the original implementation a resource path like  `C:\Users\...\keycloak\dummy-resources\myresource.css` could allow invalid access.
* in the `ResourceLoaderTest` the issue was actually caused by the Github CI environment specifics, where the temp directory and current working directory end up on different drives - (`C:\Users\RUNNER~1\AppData\Local\Temp\safepath-test9580447498682916729\resources` and `D:\a\keycloak\keycloak\services` in our case) and that causes the problem. Since it's an environment limitation, I just skip the assertion, the traversal prevention is covered by other test cases.

PTAL if it's sufficient from your PoV. Thanks.

Pepo48:
here's the result from a test branch with all 4 Win UT related commits applied: https://github.com/Pepo48/keycloak/actions/runs/18021021235/job/51278940342

shawkins:
> PTAL if it's sufficient from your PoV. Thanks.

The changes look good - that's simplier than what I was thinking. There's a small behavioral difference. The root path cannot begin with /, if it does the context classloader will fail to find the resource. This is somewhat of a non-issue as we don't currently seem to call this method with a root that begins with /.

EDIT: what if you just use "..\\some-file" on windows? Won't that still be a problem - it will pass the root check, then we'll replace \ with / and the traversal should still happen.

Pepo48:
> what if you just use "..\some-file" on windows? Won't that still be a problem - it will pass the root check, then we'll replace \ with / and the traversal should still happen.

Doesn't `assertResourceAsStream(parent, NONE + "myresource.css", false, true);` already cover the mentioned scenario?
Let's say `rootPath` is `dummy-resources/parent` and `resource` is then `..\some-file`. Then resourcePath will be `dummy-resources/some-file`.

Then the check `resourcePath.startsWith(rootPath)` return null and blocks the access, because `dummy-resources/some-file` does not start with `dummy-resources/parent`.

Or am I missing something?

shawkins:
> Doesn't assertResourceAsStream(parent, NONE + "myresource.css", false, true); already cover the mentioned scenario?

I'm not sure as I'm not sure what the affect will be on windows for the normalize call, and NONE is checking forward only, not backslash. 

If the windows normalize method covers both forward and back slash, then there's no problem.

Pepo48:
> If the windows normalize method covers both forward and back slash, then there's no problem.

Afaik, Path.normalize() just removes redundant parts lke `.` or `..` - it does not care about slashes or backslashes at all.
Path.of(...) and .resolve(...) methods do care but these are platform independent.

I can certainly add a test, something like:
```assertResourceAsStream(parent, "..\\some-file", false, false);```

It would pass on both platforms, although for different reasons - \ is handled as a separator on Windows and as a literal on Ubuntu.
On Windows it fails due to a path outside a root - which is exactly what we expect.
On Ubuntu it fails because a file with the name `..\\some-file` does not exist, which we don't necessarily need to test, but it at least show the robustness.

Let me know if you see an added value in adding such test. 

Edit: I tested it here[1][2].
[1] [commit](https://github.com/Pepo48/keycloak/commit/1152ce4f34b209ed8f3a1d31639b4c3f8fcc8bac)
[2] [corresponding CI results](https://github.com/Pepo48/keycloak/actions/runs/18041036462)

File: services/src/main/java/org/keycloak/theme/ResourceLoader.java (hunk 1 above)
Comment (shawkins) [reply services/src/main/java/org/keycloak/theme/ResourceLoader.java]: Just wanted to follow up on this a little bit more. The code is actually fine as is. You can't traverse above the root path nor is there an issue with encoded paths. It's the test logic that is not well written. It's assuming that /dummy-resources/myresource.css is being accessed when that is not the case. Ideally the tests would be rewritten so that /dummy-resource/myresource.css has a different name - e.g. forbidden.css. Access to /dummy-resource/parent/myresource.css, regardless of what resource path is used, is fine. Access to forbidden.css should never be allowed.

**Review by stianst — CHANGES_REQUESTED** [review]
Is this not just the test that is broken rather than actually an issue in ResourceLoader?

**Comment by shawkins** [comment]
> Is this not just the test that is broken rather than actually an issue in ResourceLoader?

Yes https://github.com/keycloak/keycloak/pull/42798#discussion_r2402397014

File: services/src/test/java/org/keycloak/theme/ResourceLoaderTest.java [hunk 2]
```diff
@@ -21,16 +21,16 @@ public class ResourceLoaderTest {
...
-        assertResourceAsStream(parent, DOUBLE + "myresource.css", false, false);
+        assertResourceAsStream(parent, NONE + "forbidden.css", false, true);
+        assertResourceAsStream(parent, SINGLE + "forbidden.css", false, false);
+        assertResourceAsStream(parent, DOUBLE + "forbidden.css", false, false);
 
         assertResourceAsStream(parent, "one/" + NONE + "myresource.css", true, true);
-        assertResourceAsStream(parent, "one/" + SINGLE + "myresource.css", false, false);
-        assertResourceAsStream(parent, "one/" + DOUBLE + "myresource.css", false, false);
+        assertResourceAsStream(parent, "one/" + SINGLE + "forbidden.css", false, false);
+        assertResourceAsStream(parent, "one/" + DOUBLE + "forbidden.css", false, false);
```
shawkins: These should be something like

```suggestion
        assertResourceAsStream(parent, "one/" + SINGLE + SINGLE + "forbidden.css", false, false);
        assertResourceAsStream(parent, "one/" + DOUBLE + DOUBLE + "forbidden.css", false, false);
```

otherwise they are looking in the wrong spot for forbidden.css

You should also remove dummy-resources/myresource.css to clarify the test more.

Pepo48: > otherwise they are looking in the wrong spot for forbidden.css

Thanks. missed that, the test passed but due to a missing file, not the boundary check

> You should also remove dummy-resources/myresource.css to clarify the test more.

It's removed, isn't it? At least I don't see it - essentially I just renamed dummy-resources/myresource.css to forbidden.css.

I also added the different root test fix from the previous commit, so with your your suggestion applied it should now be complete and stabilized.

Pepo48: Results from the other branch that has win UT enabled:
https://github.com/Pepo48/keycloak/actions/runs/18324870437/job/52187218549

shawkins: > It's removed, isn't it? At least I don't see it - essentially I just renamed dummy-resources/myresource.css to forbidden.css.

Yes I missed that it was moved. Everything looks good now with the stream test. The file test doesn't look quite right - it's creating two forbidden files and not appropriately checking for traversal above the root.

Pepo48: > Yes I missed that it was moved.

No worries.

> The file test doesn't look quite right.

I'm sorry about that - as I was testing it in two separate branches (with the one where UTs are enabled) I must've mixed it or the automatic renamming in IntelliJ of occurrences must've screwed me up. Either way my bad. Now it should look good for both tests (I hope :D).

shawkins (approved):
The changes look good, it has a problem with the DCO sign off.

stianst (approved):

stianst:
> The changes look good, it has a problem with the DCO sign off.

Commit sign-off looks okay to me, so I set the DCO to pass and going ahead merging

PR #42799, opened 2025-09-21
Handle canonical hostname checks for localhost on Windows

### LINKED ISSUE #42794: [Windows] Make TrustedHostClientRegistrationPolicyTest Robust to Canonical Hostname localhost
### Before reporting an issue

### Area

testsuite

### Describe the bug

 Thee canonical hostname for `127.0.0.1` can vary depending on the OS, which caused `TrustedHostClientRegistrationPolicyTest` to fail on Windows:

```
Error:  org.keycloak.services.clientregistration.policy.impl.TrustedHostClientRegistrationPolicyTest.testLocalhostDomain -- Time elapsed: 0.031 s <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.keycloak.services.clientregistration.policy.impl.TrustedHostClientRegistrationPolicyTest.testLocalhostDomain(TrustedHostClientRegistrationPolicyTest.java:70)
```

### Version

nightly

### Regression

### Expected behavior

All tests passing.

### Actual behavior

The test fails

### How to Reproduce?

Run the test on Windows

### Anything else?

_No response_

Pepo48:
Closes: #42794

### DIFF
--- a/services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
+++ b/services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
@@ -175,14 +175,22 @@ public class TrustedHostClientRegistrationPolicy implements ClientRegistrationPo
 
             logger.debugf("Trying verify request from address '%s' of host '%s' by domains", hostAddress, hostname);
 
+            // On Windows, reverse lookup for loopback may return the IP (e.g., 127.0.0.1) instead of 'localhost'. Normalize to 'localhost' for consistent domain checks.
+            if (hostname.equals(address.getHostAddress()) && address.isLoopbackAddress()) {
+                hostname = "localhost";
+            }
+
             if (hostname.equals(address.getHostAddress())) {
                 logger.debugf("The hostAddress '%s' was not resolved to a hostname", hostAddress);
                 return null;
             }
 
-            if (Arrays.stream(InetAddress.getAllByName(hostname)).filter(a -> address.equals(a)).findAny().isEmpty()) {
-                logger.debugf("The hostAddress '%s' is not among the direct lookups returned resolving '%s'", hostAddress, hostname);
-                return null;
+            // For loopback, skip strict forward-confirm check after normalizing to 'localhost' to avoid platform differences.
+            if (!address.isLoopbackAddress()) {
+                if (Arrays.stream(InetAddress.getAllByName(hostname)).filter(a -> address.equals(a)).findAny().isEmpty()) {
+                    logger.debugf("The hostAddress '%s' is not among the direct lookups returned resolving '%s'", hostAddress, hostname);
+                    return null;
+                }
             }
 
             for (String confDomain : trustedDomains) {
@@ -192,7 +200,17 @@ public class TrustedHostClientRegistrationPolicy implements ClientRegistrationPo
                 }
             }
         } catch (UnknownHostException uhe) {
-            logger.debugf(uhe, "Request of address '%s' came from unknown host. Skip verification by domains", hostAddress);
+            logger.debugf(uhe, "Request of address '%s' came from unknown host. Skip verification by domains unless it's within localhost domain", hostAddress);
+
+            String lower = hostAddress == null ? null : hostAddress.toLowerCase();
+            if (lower != null && ("localhost".equals(lower) || lower.endsWith(".localhost"))) {
+                for (String confDomain : trustedDomains) {
+                    if (checkTrustedDomain(lower, confDomain)) {
+                        logger.debugf("Treating host '%s' as loopback due to localhost domain and returning success by trusted domain '%s'", lower, confDomain);
+                        return lower;
+                    }
+                }
+            }
         }
 
         return null;
--- a/docs/documentation/release_notes/topics/26_5_0.adoc
+++ b/docs/documentation/release_notes/topics/26_5_0.adoc
@@ -1,0 +1,9 @@
+// Release notes should contain only headline-worthy new features,
+// assuming that people who migrate will read the upgrading guide anyway.
+//
+== Breaking Fix for Windows in Loopback Hostname Verification
+
+This release introduces a breaking change for Windows users: setups that previously relied on custom machine names or non-standard hostnames for loopback (e.g., `127.0.0.1` resolving to a custom name) may require updates to their trusted domain configuration. Only `localhost` and `*.localhost` are now recognized for loopback verification.
+
+Keycloak now consistently normalizes loopback addresses to `localhost` for domain verification across all platforms. This change ensures predictable behavior for trusted domain checks, regardless of the underlying OS.
+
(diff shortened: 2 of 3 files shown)


Pepo48:
After today's discussion I looked at this again.

**Windows behaviour:**
```
org.keycloak.services.clientregistration.policy.impl.TrustedHostClientRegistrationPolicyTest
19:54:23,159 WARN  main [org.keycloak.services.DefaultComponentFactoryProviderFactory] No system-wide ClusterProviderFactory found. Cannot send messages across cluster, thus disabling caching of components. Consider setting cachingForced option in single-node deployment.
19:54:23,175 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://otherhost' doesn't match any trustedHost or trustedDomain
19:54:23,268 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://www.othergooglebot.com' doesn't match any trustedHost or trustedDomain
19:54:23,268 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://localhost' doesn't match any trustedHost or trustedDomain
[TrustedHostClientRegistrationPolicyTest] Trusted hosts: []
[TrustedHostClientRegistrationPolicyTest] Trusted domains: [*.localhost]
[TrustedHostClientRegistrationPolicyTest] localhostIp: 127.0.0.1
[TrustedHostClientRegistrationPolicyTest] localhost hostname: localhost
[TrustedHostClientRegistrationPolicyTest] localhost canonical hostname: 127.0.0.1
[TrustedHostClientRegistrationPolicyTest] DNS debug error: java.net.UnknownHostException: No such host is known (other.localhost)
[TrustedHostClientRegistrationPolicyTest] verifyHost('localhost'): true
[TrustedHostClientRegistrationPolicyTest] verifyHost('other.localhost'): false
[TrustedHostClientRegistrationPolicyTest] verifyHost(localhostIp): false

Error:  org.keycloak.services.clientregistration.policy.impl.TrustedHostClientRegistrationPolicyTest.testLocalhostDomain -- Time elapsed: 0.031 s <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.keycloak.services.clientregistration.policy.impl.TrustedHostClientRegistrationPolicyTest.testLocalhostDomain(TrustedHostClientRegistrationPolicyTest.java:70)
```
**Ubuntu behaviour:** 
```
[INFO] Running org.keycloak.services.clientregistration.policy.impl.TrustedHostClientRegistrationPolicyTest
19:52:16,284 WARN  main [org.keycloak.services.DefaultComponentFactoryProviderFactory] No system-wide ClusterProviderFactory found. Cannot send messages across cluster, thus disabling caching of components. Consider setting cachingForced option in single-node deployment.
19:52:16,291 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://otherhost' doesn't match any trustedHost or trustedDomain
19:52:16,322 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://www.othergooglebot.com' doesn't match any trustedHost or trustedDomain
19:52:16,328 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://localhost' doesn't match any trustedHost or trustedDomain
[TrustedHostClientRegistrationPolicyTest] Trusted hosts: []
[TrustedHostClientRegistrationPolicyTest] Trusted domains: [*.localhost]
[TrustedHostClientRegistrationPolicyTest] localhostIp: 127.0.0.1
[TrustedHostClientRegistrationPolicyTest] localhost hostname: localhost
[TrustedHostClientRegistrationPolicyTest] localhost canonical hostname: localhost
[TrustedHostClientRegistrationPolicyTest] other.localhost hostname: other.localhost
[TrustedHostClientRegistrationPolicyTest] other.localhost canonical hostname: localhost
[TrustedHostClientRegistrationPolicyTest] 127.0.0.1 hostname: localhost
[TrustedHostClientRegistrationPolicyTest] 127.0.0.1 canonical hostname: localhost
[TrustedHostClientRegistrationPolicyTest] verifyHost('localhost'): true
[TrustedHostClientRegistrationPolicyTest] verifyHost('other.localhost'): true
[TrustedHostClientRegistrationPolicyTest] verifyHost(localhostIp): true
19:52:16,350 WARN  main [org.keycloak.services] KC-SERVICES0102: URL 'https://otherlocalhost' doesn't match any trustedHost or trustedDomain
```

As I mentioned during the meeting, on Windows reverse DNS for 127.0.0.1 often returns the IP literal or canonical hostname as the IP, which breaks the FCrDNS (forward-confirmed reverse DNS) checks that work on Linux where 127.0.0.1 resolves to “localhost”.

On my W11 machine it looks like this:
image

Another platform discrepancy that you can notice in the outputs above (and that I forgot to mention earlier) is that on Ubuntu `something.localhost` will resolve to loopback ([RFC6761](https://www.rfc-editor.org/rfc/rfc6761#section-6.3)), while on Windows it will typically end up with an error:
image
Edit: I found that [Microsoft acknowledges the same RFC](https://learn.microsoft.com/en-us/aspnet/core/test/localhost-tld) for some of their products/components like browsers or ASP.NET, but it's not something that we can rely on across the OS or even .NET Core - the call I executed above via PowerShell.

Regardless, I tried to come up with an alternative approach that tackles this directly in the `verifyHostInTrustedDomains` method and not just on the test level.

cc: @shawkins @vmuzikar @mabartos

services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
```diff
@@ -175,14 +175,22 @@ protected String verifyHostInTrustedDomains(String hostAddress, List<String> tru
 
             logger.debugf("Trying verify request from address '%s' of host '%s' by domains", hostAddress, hostname);
 
+            // On Windows, reverse lookup for loopback may return the IP (e.g., 127.0.0.1) instead of 'localhost'. Normalize to 'localhost' for consistent domain checks.
+            if (hostname.equals(address.getHostAddress()) && address.isLoopbackAddress()) {
+                hostname = "localhost";
```
shawkins, line 180: Could this be a breaking change for those on windows if they had configured the trusted clients using the machine name, instead of localhost?

Pepo48: Yes, for setups relying on custom machine names for loopback it is a breaking change. We can certainly mitigate this - possibly by checking if the original hostname matches a trusted domain before normalizing to localhost. But with a backwards compatibility in place, we would be again further to the standard defined by RFC (my intention was the opposite). If such thing was needed, I would rather vote for fixing it on a test level and let Windows treat it in it's standard way.

Quite frankly, my maybe naive assumption was this would not impact production environments, only some test environments could be affected.

vmuzikar: I would go ahead with this as it makes it consistent both on Linux and Windows. Let's just document it as a breaking change. The only question is whether to do this in KC 26.5, or wait for KC 27.

Pepo48: For the time being, I added a release note for 26.5 in order to have this PR complete.

shawkins (approved):
LGTM, thanks @Pepo48

vmuzikar (approved):

PR #42811, opened 2025-09-22
Refactoring/improvements of AuthenticationManager

atomcat1978:
Minor technical cleanup

Deprecated expireUserSessionCookie version with unused parameters.

Closes #43510

### DIFF
--- a/services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
+++ b/services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
@@ -94,6 +94,7 @@ import org.keycloak.services.util.DefaultClientSessionContext;
 import org.keycloak.services.util.UserSessionUtil;
 import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.sessions.CommonClientSessionModel;
+import org.keycloak.sessions.CommonClientSessionModel.Action;
 import org.keycloak.sessions.RootAuthenticationSessionModel;
 import org.keycloak.util.TokenUtil;
 
@@ -161,15 +162,15 @@ public class AuthenticationManager {
     // authSession note with flag that is true if the user's password has been correctly validated
     public static final String PASSWORD_VALIDATED = "PASSWORD_VALIDATED";
 
+    // state checker identity token claim
+    private static final String STATE_CHECKER = "state_checker";
+
     protected static final Logger logger = Logger.getLogger(AuthenticationManager.class);
 
     public static final String FORM_USERNAME = "username";
     // used solely to determine is user is logged in
     public static final String KEYCLOAK_SESSION_COOKIE = "KEYCLOAK_SESSION";
 
-    // ** Logout related notes **/
-    // Flag in the logout session to specify if we use "system" client or real client
-    public static final String LOGOUT_WITH_SYSTEM_CLIENT = "LOGOUT_WITH_SYSTEM_CLIENT";
     // Protocol of the client, which initiated logout
     public static final String KEYCLOAK_LOGOUT_PROTOCOL = "KEYCLOAK_LOGOUT_PROTOCOL";
     // Filled in case that logout was triggered with "initiating idp"
@@ -227,7 +228,16 @@ public class AuthenticationManager {
         return sessionIdleOk && sessionMaxOk;
     }
 
+
+    /**
+     * @deprecated Use {@link #expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo)} instead.
+     */
+    @Deprecated
     public static boolean expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo, HttpHeaders headers, ClientConnection connection) {
+        return expireUserSessionCookie(session, userSession, realm, uriInfo);
+    }
+
+    public static boolean expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo) {
         try {
             // check to see if any identity cookie is set with the same session and expire it if necessary
             String tokenString = session.getProvider(CookieProvider.class).get(CookieType.IDENTITY);
@@ -312,7 +322,7 @@ public class AuthenticationManager {
         }
 
         boolean expireUserSessionCookieSucceeded =
-                expireUserSessionCookie(session, userSession, realm, uriInfo, headers, connection);
+                expireUserSessionCookie(session, userSession, realm, uriInfo);
 
         final AuthenticationSessionManager asm = new AuthenticationSessionManager(session);
         AuthenticationSessionModel logoutAuthSession =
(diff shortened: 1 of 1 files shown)


ahus1:
Reviewing such PR from an external contributor is difficult as the reviewer doesn't have access to the refactorings your IDE suggested. It is just too simple to miss an introduced typo or change in the logic. Therefore I reject this PR.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -125,6 +123,9 @@
 
 import static org.keycloak.models.UserSessionModel.CORRESPONDING_SESSION_ID;
 import static org.keycloak.protocol.oidc.grants.device.DeviceGrantType.isOAuth2DeviceVerificationFlow;
+import static org.keycloak.sessions.CommonClientSessionModel.Action.*;
+import static org.keycloak.sessions.CommonClientSessionModel.Action.LOGGED_OUT;
+import static org.keycloak.sessions.CommonClientSessionModel.Action.LOGGING_OUT;
```
stianst: I'm not a huge fan of static imports; makes the code harder to follow IMO, so would probably skip this change. Or at the least not just `*` imports.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -189,7 +190,7 @@ public static boolean isSessionValid(RealmModel realm, UserSessionModel userSess
         }
         if (userSession.getNote(Details.IDENTITY_PROVIDER) != null) {
             String brokerAlias = userSession.getNote(Details.IDENTITY_PROVIDER);
-            if (realm.getIdentityProviderByAlias(brokerAlias) == null) {
+            if (realm.getFlowByAlias(brokerAlias) == null) {
```
stianst: This change looks invalid to me; looks like the code is just checking that the identity provider by alias still exists; while `getFlowByAlias` is something completely different

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -1261,7 +1256,7 @@ protected static Response executionActions(KeycloakSession session, Authenticati
         final String kcAction = authSession.getClientNote(Constants.KC_ACTION);
         final RequiredActionProviderModel firstApplicableRequiredAction =
                 getFirstApplicableRequiredAction(realm, authSession, user, kcAction, ignoredActions);
-        boolean kcActionExecution = kcAction != null && kcAction.equals(firstApplicableRequiredAction.getProviderId());
+        boolean kcActionExecution = kcAction != null && kcAction.equals((firstApplicableRequiredAction ==null ) ? null : firstApplicableRequiredAction.getProviderId());
```
stianst: Not sure about this one; it should probably throw an exception if firstApplicableRequiredAction is null. @mposolda wdyt?

stianst (commented):
Added some comments inline; can you create a linked GitHub Issue and also do the sign-off of the commit?

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
atomcat1978: If accepted, should this be done as part of this cleanup? Possibly better as a separate fiks.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
atomcat1978: Indeed, it should be getByAlias, my bad. `getIdentityProviderByAlias` is deprecated and `getByAlias` should be used.

```
    /**
     * @deprecated Use {@link IdentityProviderStorageProvider#getByAlias(String)} instead.
     */
    @Deprecated
    IdentityProviderModel getIdentityProviderByAlias(String alias);
```

However, it would require changing the method interface. Maybe it should just be left as it is. Reverting this change.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -226,7 +226,7 @@ public static boolean isClientSessionValid(RealmModel realm, ClientModel client,
         return sessionIdleOk && sessionMaxOk;
     }
 
-    public static boolean expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo, HttpHeaders headers, ClientConnection connection) {
+    public static boolean expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo) {
```
atomcat1978, line 240: @stianst, I just realized that here we change a public API function. Should this be reverted or added to a change log or something? The trailing parameters are not used, so in a sense it is a legitimate change.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
atomcat1978: Agree, here there are especially a lot of enums defined with the same name from different enum types. Fixed the imports.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
stianst: Might be safest to deprecate the original method, and add a new without the unused parameters

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
stianst: Yes, probably best to just leave this one out for now

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
atomcat1978: Good idea, going for that.

atomcat1978: Done.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -223,7 +224,15 @@ public static boolean isClientSessionValid(RealmModel realm, ClientModel client,
         return sessionIdleOk && sessionMaxOk;
     }
 
+
+    /**
+     * @deprecated Use {@link #expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo)} instead.
+     */
     public static boolean expireUserSessionCookie(KeycloakSession session, UserSessionModel userSession, RealmModel realm, UriInfo uriInfo, HttpHeaders headers, ClientConnection connection) {
```
stianst, line 236: Please add the `@Deprecated` annotation as well

atomcat1978: Done.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
atomcat1978: Ok, reverted.

stianst:
Can you squash the commits and add the DCO signature (https://github.com/keycloak/keycloak/blob/main/CONTRIBUTING.md#developers-certificate-of-origin)?

atomcat1978:
> Can you squash the commits and add the DCO signature (https://github.com/keycloak/keycloak/blob/main/CONTRIBUTING.md#developers-certificate-of-origin)?

Done.

stianst (approved):

PR #42819, opened 2025-09-22
[OID4VCI] Fix authorization_details generation and credential identifier mapping for conformance tests

### LINKED ISSUE #42818: [OID4VCI] Adjust authorization_details generation and credential identifier mapping for OIDVCI Conformace Test
### Description

OID4VCI Conformance Tests were failing because the system wasn't generating `authorization_details` in token responses when not provided in token requests.

According to the OID4VCI specification (Section 6.2), `authorization_details` is:
- **REQUIRED** when the `authorization_details` parameter is used in the token request
- **OPTIONAL** when the `scope` parameter was used instead

In pre-authorized code flows, since there's no use of the authorization endpoint, the only way to request with authorization details is in the token request. However, conformance tests send a token request **without** authorization details and expect to receive them in the response.

## Solution
- Generate `authorization_details` from the credential offer's `credential_configuration_ids` when not provided in the token request
- Include the generated authorization details in the token response so conformance tests pass
- Fix credential identifier mapping to ensure credential requests work correctly with the generated identifiers

forkimenjeckayang:
## Summary
Fixes OID4VCI conformance test failures by implementing proper authorization_details generation from credential offers and fixing credential identifier mapping.

## Problem
- OID4VCI conformance tests were failing because `authorization_details` weren't being generated in token responses when not provided in token requests
- Credential requests using only `credential_identifier` were failing due to mapping lookup issues

## Solution
- **Authorization Details Generation**: Added fallback mechanism to generate `authorization_details` from credential offer's `credential_configuration_ids` when not provided in token request
- **Credential Identifier Mapping**: Fixed client session lookup to ensure credential identifier mappings are found during credential requests

## OID4VCI Conformance Test --  Without Authorization Details

Image

## OID4VCI Conformance Test --  With  Authorization Details

Image

## Closses

closes https://github.com/keycloak/keycloak/issues/42818

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
@@ -254,7 +254,7 @@ public class OID4VCAuthorizationDetailsProcessor implements AuthorizationDetails
      * @param clientSession the client session that contains the credential offer information
      * @return the authorization details response if generation was successful, null otherwise
      */
-    public List<AuthorizationDetailsResponse> generateAuthorizationDetailsFromCredentialOffer(AuthenticatedClientSessionModel clientSession) {
+    private List<AuthorizationDetailsResponse> generateAuthorizationDetailsFromCredentialOffer(AuthenticatedClientSessionModel clientSession) {
         logger.info("Processing authorization_details from credential offer");
 
         // Get supported credentials
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
@@ -15,247 +15,328 @@
  * limitations under the License.
  */
 
-package org.keycloak.protocol.oid4vc.issuance;
+ package org.keycloak.protocol.oid4vc.issuance;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import org.jboss.logging.Logger;
-import org.keycloak.models.ClientSessionContext;
-import org.keycloak.models.KeycloakSession;
-import org.keycloak.models.UserSessionModel;
-import org.keycloak.protocol.oid4vc.model.AuthorizationDetail;
-import org.keycloak.protocol.oid4vc.model.ClaimsDescription;
-import org.keycloak.protocol.oid4vc.model.SupportedCredentialConfiguration;
-import org.keycloak.protocol.oid4vc.model.Claim;
-import org.keycloak.util.JsonSerialization;
-import org.keycloak.protocol.oidc.rar.AuthorizationDetailsProcessor;
-import org.keycloak.protocol.oidc.rar.AuthorizationDetailsResponse;
-
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
-import java.util.stream.Collectors;
-
-import org.keycloak.protocol.oid4vc.utils.ClaimsPathPointer;
-
-import static org.keycloak.models.Constants.AUTHORIZATION_DETAILS_RESPONSE;
-
-public class OID4VCAuthorizationDetailsProcessor implements AuthorizationDetailsProcessor {
-    private static final Logger logger = Logger.getLogger(OID4VCAuthorizationDetailsProcessor.class);
-    private final KeycloakSession session;
-
-    public static final String OPENID_CREDENTIAL_TYPE = "openid_credential";
-
-    public OID4VCAuthorizationDetailsProcessor(KeycloakSession session) {
-        this.session = session;
-    }
-
-    @Override
-    public List<AuthorizationDetailsResponse> process(UserSessionModel userSession, ClientSessionContext clientSessionCtx, String authorizationDetailsParameter) {
-        if (authorizationDetailsParameter == null) {
-            return null; // authorization_details is optional
-        }
-
-        List<AuthorizationDetail> authDetails = parseAuthorizationDetails(authorizationDetailsParameter);
-        Map<String, SupportedCredentialConfiguration> supportedCredentials = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session);
-        List<AuthorizationDetailsResponse> authDetailsResponse = new ArrayList<>();
-
-        // Retrieve authorization servers and issuer identifier for locations check
-        List<String> authorizationServers = OID4VCIssuerWellKnownProvider.getAuthorizationServers(session);
-        String issuerIdentifier = OID4VCIssuerWellKnownProvider.getIssuer(session.getContext());
-
-        for (AuthorizationDetail detail : authDetails) {
-            validateAuthorizationDetail(detail, supportedCredentials, authorizationServers, issuerIdentifier);
-            AuthorizationDetailsResponse responseDetail = buildAuthorizationDetailResponse(detail, userSession, supportedCredentials, clientSessionCtx);
-            authDetailsResponse.add(responseDetail);
-        }
-
-        if (authDetailsResponse.isEmpty()) {
-            throw getInvalidRequestException("Invalid authorization_details: no valid authorization details found");
-        }
-
-        return authDetailsResponse;
-    }
-
-    private List<AuthorizationDetail> parseAuthorizationDetails(String authorizationDetailsParam) {
-        try {
-            return JsonSerialization.readValue(authorizationDetailsParam, new TypeReference<List<AuthorizationDetail>>() {
-            });
-        } catch (Exception e) {
-            logger.warnf(e, "Invalid authorization_details format: %s", authorizationDetailsParam);
-            throw getInvalidRequestException("Invalid authorization_details format: " + authorizationDetailsParam);
-        }
-    }
-
-    private RuntimeException getInvalidRequestException(String errorDescription) {
-        return new RuntimeException("Invalid authorization_details: " + errorDescription);
-    }
-
-    /**
-     * Validates an authorization detail against supported credentials and other constraints.
-     *
-     * @param detail               the authorization detail to validate
-     * @param supportedCredentials map of supported credential configurations
-     * @param authorizationServers list of authorization servers
-     * @param issuerIdentifier     the issuer identifier
-     */
-    private void validateAuthorizationDetail(AuthorizationDetail detail, Map<String, SupportedCredentialConfiguration> supportedCredentials, List<String> authorizationServers, String issuerIdentifier) {
-
-        String type = detail.getType();
-        String credentialConfigurationId = detail.getCredentialConfigurationId();
-        List<ClaimsDescription> claims = detail.getClaims();
-
-        // If authorization_servers is present, locations must be set to issuer identifier
-        if (authorizationServers != null && !authorizationServers.isEmpty() && OPENID_CREDENTIAL_TYPE.equals(type)) {
-            List<String> locations = detail.getLocations();
-            if (locations == null || locations.size() != 1 || !issuerIdentifier.equals(locations.get(0))) {
-                logger.warnf("Invalid locations field in authorization_details: %s, expected: %s", locations, issuerIdentifier);
-                throw getInvalidRequestException("Invalid authorization_details: locations=" + locations + ", expected=" + issuerIdentifier);
-            }
-        }
-
-        // Validate type
-        if (!OPENID_CREDENTIAL_TYPE.equals(type)) {
-            logger.warnf("Invalid authorization_details type: %s", type);
-            throw getInvalidRequestException("Invalid authorization_details type: " + type + ", expected=" + OPENID_CREDENTIAL_TYPE);
-        }
-
-        // credential_configuration_id is REQUIRED
-        if (credentialConfigurationId == null) {
-            logger.warnf("Missing credential_configuration_id in authorization_details");
-            throw getInvalidRequestException("Invalid authorization_details: credential_configuration_id is required");
-        }
-
-        // Validate credential_configuration_id
-        SupportedCredentialConfiguration config = supportedCredentials.get(credentialConfigurationId);
-        if (config == null) {
-            logger.warnf("Unsupported credential_configuration_id: %s", credentialConfigurationId);
-            throw getInvalidRequestException("Invalid credential configuration: unsupported credential_configuration_id=" + credentialConfigurationId);
-        }
-
-
-        // Validate claims if present
-        if (claims != null && !claims.isEmpty()) {
-            validateClaims(claims, supportedCredentials, credentialConfigurationId);
-        }
-    }
-
-    /**
-     * Validates that the requested claims are supported by the credential configuration.
-     * This performs semantic validation by checking if Keycloak supports the requested claims.
-     *
-     * @param claims                    the list of claims to validate
-     * @param supportedCredentials      map of supported credential configurations
-     * @param credentialConfigurationId the ID of the credential configuration
-     */
-    private void validateClaims(List<ClaimsDescription> claims, Map<String, SupportedCredentialConfiguration> supportedCredentials, String credentialConfigurationId) {
-        SupportedCredentialConfiguration config = supportedCredentials.get(credentialConfigurationId);
-
-        // Get the exposed claims from credential metadata
-        List<Claim> exposedClaims = null;
-        if (config.getCredentialMetadata() != null && config.getCredentialMetadata().getClaims() != null && !config.getCredentialMetadata().getClaims().isEmpty()) {
-            exposedClaims = config.getCredentialMetadata().getClaims();
-        }
-
-        if (exposedClaims == null || exposedClaims.isEmpty()) {
-            throw getInvalidRequestException("Credential configuration does not expose any claims metadata");
-        }
-
-        // Convert exposed claims to a set of paths for easy comparison
-        Set<String> exposedClaimPaths = exposedClaims.stream()
-                .filter(claim -> claim.getPath() != null && !claim.getPath().isEmpty())
-                .map(claim -> claim.getPath().toString())
-                .collect(Collectors.toSet());
-
-        // Validate each requested claim against exposed metadata
-        for (ClaimsDescription requestedClaim : claims) {
-            if (requestedClaim.getPath() == null || requestedClaim.getPath().isEmpty()) {
-                throw getInvalidRequestException("Invalid claims description: path is required");
-            }
-
-            // Validate the claims path pointer format according to OID4VCI specification
-            if (!ClaimsPathPointer.isValidPath(requestedClaim.getPath())) {
-                throw getInvalidRequestException("Invalid claims path pointer: " + requestedClaim.getPath() +
-                        ". Path must contain only strings, non-negative integers, and null values.");
-            }
-
-            String requestedPath = requestedClaim.getPath().toString();
-
-            // Check if the requested claim path exists in the exposed metadata
-            if (!exposedClaimPaths.contains(requestedPath)) {
-                throw getInvalidRequestException("Unsupported claim: " + requestedPath +
-                        ". This claim is not supported by the credential configuration.");
-            }
-        }
-
-        // Check for conflicts using ClaimsPathPointer utility
-        if (!ClaimsPathPointer.validateClaimsDescriptions(claims)) {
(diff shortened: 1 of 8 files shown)


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -322,6 +322,13 @@ public Response getCredentialOfferURI(@QueryParam("credential_configuration_id")
         String sessionCode = generateCodeForSession(expiration, clientSession);
         try {
             clientSession.setNote(sessionCode, JsonSerialization.mapper.writeValueAsString(theOffer));
+
+            // Store the credential configuration IDs in a predictable location for token processing
+            // This allows the authorization details processor to easily retrieve the configuration IDs
+            // without having to search through all session notes or parse the full credential offer
+            String credentialConfigIdsJson = JsonSerialization.mapper.writeValueAsString(theOffer.getCredentialConfigurationIds());
+            clientSession.setNote("CREDENTIAL_CONFIGURATION_IDS", credentialConfigIdsJson);
```
tnorimat: How about defining it as `public static final String` ? `OID4VCAuthorizationDetailsProcessor.extractCredentialConfigurationIds` also refers the same string.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -492,40 +499,51 @@ public Response requestCredential(String requestPayload) {
 
         CredentialScopeModel requestedCredential;
 
-        // If credential_identifier is provided, try to retrieve stored credential context first
+        // If credential_identifier is provided, retrieve the mapping from client session
         if (credentialRequestVO.getCredentialIdentifier() != null) {
-            String contextKey = "CREDENTIAL_CONTEXT_" + credentialRequestVO.getCredentialIdentifier();
-            String storedContextJson = authResult.getSession().getNote(contextKey);
+            String mappingKey = "credential_identifier_" + credentialRequestVO.getCredentialIdentifier();
```
tnorimat: How about defining `"credential_identifier_"` as public static final String ? `OID4VCAuthorizationDetailsProcessor`also refers the same string.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -492,40 +499,51 @@ public Response requestCredential(String requestPayload) {
...
-                        ClientScopeModel clientScope = clientScopes.get(config.getScope());
-
-                        if (clientScope != null) {
-                            requestedCredential = new CredentialScopeModel(clientScope);
-                        } else {
-                            LOGGER.errorf("Client scope not found for stored credential configuration: %s", config.getScope());
-                            throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
-                        }
+            if (mappedCredentialConfigurationId != null) {
+                LOGGER.debugf("Found credential configuration ID mapping for identifier %s: %s",
```
tnorimat: It seems that it duplicate Line 514’s logging.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+             return JsonSerialization.readValue(authorizationDetailsParam, new TypeReference<List<AuthorizationDetail>>() {
+             });
+         } catch (Exception e) {
+             logger.warnf(e, "Invalid authorization_details format: %s", authorizationDetailsParam);
+             throw getInvalidRequestException("Invalid authorization_details format: " + authorizationDetailsParam);
+         }
+     }
+ 
+     private RuntimeException getInvalidRequestException(String errorDescription) {
+         return new RuntimeException("Invalid authorization_details: " + errorDescription);
```
tnorimat: Most of the callers of this method include the string `Invalid authorization_details:` as the argument `errorDescription`, which lead to duplicated message `Invalid authorization_details: Invalid authorization_details:`.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+                 logger.warnf("Invalid locations field in authorization_details: %s, expected: %s", locations, issuerIdentifier);
+                 throw getInvalidRequestException("Invalid authorization_details: locations=" + locations + ", expected=" + issuerIdentifier);
+             }
+         }
+ 
+         // Validate type
+         if (!OPENID_CREDENTIAL_TYPE.equals(type)) {
+             logger.warnf("Invalid authorization_details type: %s", type);
+             throw getInvalidRequestException("Invalid authorization_details type: " + type + ", expected=" + OPENID_CREDENTIAL_TYPE);
+         }
```
tnorimat: If this check is executed before Line 110, then wa can omit the condition `OPENID_CREDENTIAL_TYPE.equals(type)` in Line 111.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+ 
+     /**
+      * Validates that the requested claims are supported by the credential configuration.
+      * This performs semantic validation by checking if Keycloak supports the requested claims.
+      *
+      * @param claims                    the list of claims to validate
+      * @param supportedCredentials      map of supported credential configurations
+      * @param credentialConfigurationId the ID of the credential configuration
+      */
+     private void validateClaims(List<ClaimsDescription> claims, Map<String, SupportedCredentialConfiguration> supportedCredentials, String credentialConfigurationId) {
```
tnorimat: The arguments `supportedCredentials` and `credentialConfigurationId` are only used in Line 154. Therefore, we could change this method' signature as follows:
```
private void validateClaims(List<ClaimsDescription> claims, SupportedCredentialConfiguration config)
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+             }
+         }
+ 
+         // Check for conflicts using ClaimsPathPointer utility
+         if (!ClaimsPathPointer.validateClaimsDescriptions(claims)) {
+             throw getInvalidRequestException("Invalid claims descriptions: conflicting or contradictory claims found");
+         }
+     }
+ 
+     private AuthorizationDetailsResponse buildAuthorizationDetailResponse(AuthorizationDetail detail, UserSessionModel userSession, Map<String, SupportedCredentialConfiguration> supportedCredentials, ClientSessionContext clientSessionCtx) {
```
tnorimat: It seems that the arugment `supportedCredentials` is not used in this method.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+             // Store the mapping between credential identifier and configuration ID in client session
+             String mappingKey = "credential_identifier_" + credentialIdentifier;
+             clientSession.setNote(mappingKey, credentialConfigurationId);
+             logger.debugf("Stored credential identifier mapping: %s -> %s", credentialIdentifier, credentialConfigurationId);
+         }
+ 
+         // Store claims in user session for later use during credential issuance
+         if (detail.getClaims() != null) {
+             // Store claims with a unique key based on credential configuration ID
+             String claimsKey = "AUTHORIZATION_DETAILS_CLAIMS_" + credentialConfigurationId;
```
tnorimat: How about defining `AUTHORIZATION_DETAILS_CLAIMS_` as public static final String ? `OID4VCIssuerEndpoint` also refers the same string.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+             if (config == null) {
+                 logger.warnf("Credential configuration '%s' not found in supported credentials, skipping", credentialConfigurationId);
+                 continue;
+             }
+ 
+             String credentialIdentifier = UUID.randomUUID().toString();
+ 
+             // Store the mapping between credential identifier and configuration ID in client session
+             // This will be used later when processing credential requests
+             String mappingKey = "credential_identifier_" + credentialIdentifier;
```
tnorimat: How about defining `credential_identifier_` as public static final String ? OID4VCIssuerEndpoint also refers the same string.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -15,247 +15,328 @@
...
+     /**
+      * Process authorization_details from the credential offer when authorization_details parameter is not present in the token request.
+      * This method generates authorization_details based on the credential_configuration_ids from the credential offer.
+      *
+      * @param userSession      the user session
+      * @param clientSessionCtx the client session context
+      * @param clientSession    the client session that contains the credential offer information
+      * @return the authorization details response if processing was successful, null otherwise
+      */
+     public List<AuthorizationDetailsResponse> processFromCredentialOffer(UserSessionModel userSession, ClientSessionContext clientSessionCtx, AuthenticatedClientSessionModel clientSession) {
```
tnorimat: It seems that the arguments `UserSessionModel userSession` and `ClientSessionContext clientSessionCtx` is not used in this method.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -446,6 +448,193 @@ public void testPreAuthorizedCodeWithEmptyAuthorizationDetails() throws Exceptio
...
     }
 
+    @Test
+    public void testPreAuthorizedCodeWithCredentialOfferBasedAuthorizationDetails() throws Exception {
+        String token = getBearerToken(oauth, client, getCredentialClientScope().getName());
+        Oid4vcTestContext ctx = prepareOid4vcTestContext(token);
+
+        // Test Pre-Authorized Code Flow without authorization_details parameter
+        // The system should generate authorization_details based on credential_configuration_ids from the credential offer
+        String credentialConfigId = getCredentialClientScope().getAttributes().get(CredentialScopeModel.CONFIGURATION_ID);
```
tnorimat: It seems that the variable `credentialConfigId` is not used in thes method.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -446,6 +448,193 @@ public void testPreAuthorizedCodeWithEmptyAuthorizationDetails() throws Exceptio
...
+
+        UrlEncodedFormEntity formEntity = new UrlEncodedFormEntity(parameters, StandardCharsets.UTF_8);
+        postPreAuthorizedCode.setEntity(formEntity);
+
+        try (CloseableHttpResponse tokenResponse = httpClient.execute(postPreAuthorizedCode)) {
+            assertEquals(HttpStatus.SC_OK, tokenResponse.getStatusLine().getStatusCode());
+            String responseBody = IOUtils.toString(tokenResponse.getEntity().getContent(), StandardCharsets.UTF_8);
+
+            System.out.println("Token Response Status: " + tokenResponse.getStatusLine().getStatusCode());
+            System.out.println("Token Response Body: " + responseBody);
```
tnorimat: These two lines are not needed for the test.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -446,6 +448,193 @@ public void testPreAuthorizedCodeWithEmptyAuthorizationDetails() throws Exceptio
...
+            }
+        }
+    }
+
+    @Test
+    public void testCompleteFlowWithCredentialOfferBasedAuthorizationDetails() throws Exception {
+        String token = getBearerToken(oauth, client, getCredentialClientScope().getName());
+        Oid4vcTestContext ctx = prepareOid4vcTestContext(token);
+
+        String credentialConfigId = getCredentialClientScope().getAttributes().get(CredentialScopeModel.CONFIGURATION_ID);
```
tnorimat: It seems that the variable `credentialConfigId` is not used in thes method.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -446,6 +448,193 @@ public void testPreAuthorizedCodeWithEmptyAuthorizationDetails() throws Exceptio
...
+        HttpPost postPreAuthorizedCode = new HttpPost(ctx.openidConfig.getTokenEndpoint());
+        List<NameValuePair> parameters = new LinkedList<>();
+        parameters.add(new BasicNameValuePair(OAuth2Constants.GRANT_TYPE, PreAuthorizedCodeGrantTypeFactory.GRANT_TYPE));
+        parameters.add(new BasicNameValuePair(PreAuthorizedCodeGrantTypeFactory.CODE_REQUEST_PARAM, ctx.credentialsOffer.getGrants().getPreAuthorizedCode().getPreAuthorizedCode()));
+
+        UrlEncodedFormEntity formEntity = new UrlEncodedFormEntity(parameters, StandardCharsets.UTF_8);
+        postPreAuthorizedCode.setEntity(formEntity);
+
+        String credentialIdentifier = null;
+        String usedCredentialConfigId = null;
```
tnorimat: The variable `usedCredentialConfigId` is not used even if Line 536 set some value to the variable but is not referred elesewhare.

tnorimat (commented):
@forkimenjeckayang Hello, I added some reivew comments. Could you check them?

forkimenjeckayang:
Thank you for the review comments @tnorimat 
Please have another look when you can

tnorimat (approved):
LGTM

tnorimat:
@forkimenjeckayang Hello, could you share the keycloak's settings (realm json, properties, environment variables, etc.) and the conformance test setting (json) to run the OpenID4VCI conformance test against Keycloak?

When Keycloak 26.4 is released, I would like to run the conformance test to use the OAuth-SIG's  conformance test run envorinment (https://github.com/keycloak/keycloak-oauth-sig/tree/main/conformance-tests-env).

forkimenjeckayang:
> @forkimenjeckayang Hello, could you share the keycloak's settings (realm json, properties, environment variables, etc.) and the conformance test setting (json) to run the OpenID4VCI conformance test against Keycloak?
> 
> When Keycloak 26.4 is released, I would like to run the conformance test to use the OAuth-SIG's conformance test run envorinment (https://github.com/keycloak/keycloak-oauth-sig/tree/main/conformance-tests-env).

Thank you for the review @tnorimat 
I sent configuration details to your slack.
I believe that should be sufficient to run the test successfully for the pre-authorized code flow

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -221,39 +220,116 @@ private AuthorizationDetailsResponse buildAuthorizationDetailResponse(Authorizat
...
 
+
+    /**
+     * Process authorization_details from the credential offer when authorization_details parameter is not present in the token request.
+     * This method generates authorization_details based on the credential_configuration_ids from the credential offer.
+     *
+     * @param clientSession the client session that contains the credential offer information
+     * @return the authorization details response if processing was successful, null otherwise
+     */
+    public List<AuthorizationDetailsResponse> processFromCredentialOffer(AuthenticatedClientSessionModel clientSession) {
```
thomasdarimont: Does this method generate the authorization-details in case they are missing? If so, then I'd rename the method to like `generateAuthorizationDetailsFromCredentialOffer` to better express it's intent.

services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
```diff
@@ -291,6 +291,31 @@ protected List<AuthorizationDetailsResponse> processAuthorizationDetails(UserSes
...
+     * @param clientSession the client session that contains the credential offer information
+     * @return the authorization details response if processing was successful, null otherwise
+     */
+    protected List<AuthorizationDetailsResponse> processAuthorizationDetailsFromCredentialOffer(AuthenticatedClientSessionModel clientSession) {
+        try {
+            return session.getKeycloakSessionFactory()
+                    .getProviderFactoriesStream(AuthorizationDetailsProcessor.class)
+                    .sorted((f1, f2) -> f2.order() - f1.order())
+                    .map(f -> session.getProvider(AuthorizationDetailsProcessor.class, f.getId()))
+                    .filter(processor -> processor instanceof org.keycloak.protocol.oid4vc.issuance.OID4VCAuthorizationDetailsProcessor)
```
mposolda: It will be great if we can avoid referencing oid4vci in the core OIDC classes like `OAuth2GrantTypeBase` or `AuthorizationCodeGrantType` to avoid cross-referencing and inter-linking of various protocols...

Can we instead add new method to `AuthorizationDetailsProcessor` interface? Maybe something like:
```
    /**
     * Method is invoked in cases when authorization_details parameter is missing in the request. It allows processor to
     * generate authorization details response in such a case
     *
     * @param userSession                   the user session
     * @param clientSessionCtx              the client session context
     * @return authorization details response if this processor can handle current request in case that authorization_details parameter was not provided
     */
    List<AuthorizationDetailsResponse> handleMissingAuthorizationDetails(UserSessionModel userSession,
                                                                         ClientSessionContext clientSessionCtx);
```

The method in `OAuth2GrantTypeBase` could be then also renamed to something more generic like `handleMissingAuthorizationDetails` and call the interface method instead of casting to `OID4VCAuthorizationDetailsProcessor`

mposolda (commented):
@forkimenjeckayang Thanks for the PR. Adding the comment inline.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -221,39 +220,122 @@ private AuthorizationDetailsResponse buildAuthorizationDetailResponse(Authorizat
...
 
+
+    /**
+     * Generate authorization_details from the credential offer when authorization_details parameter is not present in the token request.
+     * This method generates authorization_details based on the credential_configuration_ids from the credential offer.
+     *
+     * @param clientSession the client session that contains the credential offer information
+     * @return the authorization details response if generation was successful, null otherwise
+     */
+    public List<AuthorizationDetailsResponse> generateAuthorizationDetailsFromCredentialOffer(AuthenticatedClientSessionModel clientSession) {
```
mposolda: This method could be private? But this is nitpick (I do not consider it blocker for this PR).

mposolda (dismissed):
@forkimenjeckayang Thanks! Added minor comment inline. But not a blocker for this PR. I am approving the PR. Hope to merge once tests are OK unless there are some other review comments.

VinodAnandan (approved):
LGTM

mposolda (approved):
@forkimenjeckayang @tnorimat @VinodAnandan @thomasdarimont Thanks for the updates and review!

PR #42822, opened 2025-09-22
Move DPoP option to the capability section in the admin UI

### LINKED ISSUE #42746: Polishing of client switch on DPoP
### Description

There is currently the switch `Require Demonstrating Proof of Possession (DPoP) header in token requests` in the admin console. However it is bit hidden in the client "Advanced" tab. There was a proposal on the internal Keycloak team meeting to make this switch more visible similarly like we did for the `PKCE method` switch.

This issue is about:

- Move the switch from the "Advanced" tab to the initial tab of the OIDC client. We can perhaps add the switch right after the `PKCE method` switch?
- Maybe we can rename the switch to not have so long name? Maybe only `Require DPoP bound tokens`?
- We should perhaps also clarify the tooltip for this switch. As currently refresh-token is DPoP bound just for the public clients, but not for confidential clients. This should be mentioned in the tooltip (It is not mentioned right now)

NOTE: No strictly need to update docs as part of this task. Can be done as a follow-up as we have dedicated docs issue https://github.com/keycloak/keycloak/issues/42728

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #42746

Moving the DPoP switch from advanced settings to capability (just after PKCE). Label also changed for the option as commented in the issue. The option is only shown when the client is public (as it is only applicable in that situation). Changed the doc to mention the new option name there.

I see that there is a link error for npmjs (403 error). It's not related to the changes and probably we will need to include that URL as ignored (because it seems that npmjs is doing something weird with JS to set a cookie, returning 403 if JS not allowed).


js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -1935,7 +1935,7 @@ scopePermissions.groups.view-description=Policies that decide if an administrato
 tokens=Tokens
 createFlow=Create flow
 encryptAssertionsHelp=Should SAML assertions be encrypted with client's public key?
-oAuthDPoPHelp=This enables support for Demonstrating Proof-of-Possession (DPoP) bound tokens. The access and refresh tokens are bound to the key stored on the user agent. In order to prove the possession of the key, the user agent must send a signed proof alongside the token.
+oAuthDPoPHelp=This enables support for Demonstrating Proof-of-Possession (DPoP) bound tokens. The access and refresh tokens are bound to the key stored on the user agent. In order to prove the possession of the key, the user agent must send a signed proof alongside the token. DPoP can only be configured for public clients.
```
mposolda: This is not accurate. DPoP can be configured for both public and confidential clients. The only thing is, that for confidential clients, only access-tokens are DPoP bound. For public clients, both access-tokens and refresh-tokens are DPoP bounds. In other words, the binding of refresh tokens is only for public clients.

Is it possible to update the tooltip this way and make sure that switch is visible for both public and confidential clients?

Also you can possibly either update docs or revert changes of docs from this PR (as there is dedicated task for docs, which I am planning to handle once other PRs are done).

mposolda (commented):
@rmartinc Thanks! I've added comment inline. Can you please check it?

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
rmartinc: Ah! Then it should not be hidden for confidential. I'll do it after lunch.

rmartinc: Done! Let's see the tests.

mposolda (approved):
@rmartinc Nice, Thanks!

PR #42832, opened 2025-09-22
Refine high-availability guide wording

ryanemerson:
Closes #42829


ryanemerson:
@ahus1 In general I have tried to avoid adding if/else logic for product/community and applied the feedback to both cases when possible. I think the only contentious part is maybe around including "data center" in the community, let me know what you think.

ahus1 (approved):
Thank you for this documentation update!

PR #42840, opened 2025-09-22
Update upgrade guide about changes in how the  parameter is propagated to OPs

### LINKED ISSUE #42139: Backwards compatibility awareness
### Before reporting an issue

### Area

_No response_

### Describe the bug

I want to make you aware of a minor change in behavior on how acr_values is forwarded to an idp caused by PR https://github.com/keycloak/keycloak/pull/41671.

The old behavior for the acr_values was that it was always forwarded to an idp if the acr_values was present, independent on how forwarded parameters config was configured for the idp.

PR 41671 seems to change this to only forward the acr_values if the forwarded parameters is null. This means that if anyone rely on the acr_values being forwarded and they have specified one or more forwarded parameters (and not included the acr_values explicitly), then the acr_values will no longer be forwarded.

It would be good to have a discussion on what is the correct approach for the acr_values and I already have a issue https://github.com/keycloak/keycloak/issues/39813 and a [PR #39814](https://github.com/keycloak/keycloak/pull/39814) to allow disabling the acr_values forwarding.

Do you agree that my issue/PR is still relevant? Should I update it?

I can't find any specification that says that the acr_values should be forwarded by default.
Personally, I think the acr_values shouldn't even get forwarded by default and it should be explicitly configured as a forwarded parameter or as a separate option as proposed in my PR.

 Would you like me to create a PR to update the upgrade notes? Or should I create a PR to revert the implementation to keep the existing behavior to always forward the acr_values?

### Version

26.4

### Regression

### Expected behavior

The acr_values was previously always forwarded to an idp.

### Actual behavior

The acr_values parameter is only forwarded if the forwarded parameters idp config is null or acr_values is part of the forwarded parameters config.

### How to Reproduce?

Create an idp config and specify only claim as a forwarded parameter. The acr_values will no longer be forwarded.

### Anything else?

_No response_

pedroigor:
Closes #42139

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -4,6 +4,14 @@
 Breaking changes are identified as those that might require changes for existing users to their configurations or applications.
 In minor or patch releases, {project_name} will only introduce breaking changes to fix bugs.
 
+=== `acr_values` request parameter is not forwarded automatically to identity providers
+
+The `acr_values` request parameter is no longer automatically forwarded to OpenID Connect identity providers during authentication.
+This change enhances security by preventing unintended disclosure of authentication context information to external IDPs.
+
+If you are relying on the `acr_values` parameter to be propagated to an identity provider, you must now explicitly set `acr_values` request parameter
+to the `Forwarded query parameters` setting in the identity provider configuration.
+
 // ------------------------ Notable changes ------------------------ //
 == Notable changes
 


sguilhen (approved):

sguilhen:
@pedroigor Looks like you need to re-run the CI for this one.

pedroigor:
The error in the docs can be ignored based on previous PRs and updates to the docs.

pedroigor:
Sorry, @sguilhen. The failure was due to an additional and invalid commit. It should be OK now if doc tests are still not failing.

sguilhen (approved):

ahus1 (approved):

PR #42859, opened 2025-09-23
Expose system-info information in the serverinfo endpoint only for users in the admin realm

rmartinc:
Closes #42828

Sending a draft for the moment. The main idea here is that the system-info that is added to the `serverinfo` endpoint is only used by the admin console for the administrator realm ([here](https://github.com/keycloak/keycloak/blob/26.3.4/js/apps/admin-ui/src/dashboard/Dashboard.tsx#L360-L361)). So my initial idea is doing the same for the endpoint. The info about the system is only added if the administrator belongs to the admin realm (master). Although this endpoint is not documented and only used by the admin console, this can be considered a breaking change. So asking here:

1.  The current PR just adds that. The system information is only added for administrators of the master realm. A note probably is needed but, as commented, this endpoint is internal and we can say this is just a bug.
2. If we want to add a fall-back for non-master administrators I would use a new role in the `realm-management` client (something like `view-system`). I would not even create that role. But there is a fall-back procedure (create the role and assign the role to the wanted admins). Maybe you have another idea.

Because it's a mess doing 2 for the moment I have just remove the system info. It is only shown for administrators of the admin realm like the console requires. Maybe the keycloak version is the only info that some admin can need. WDYT? Do we maintain this like it is now or do we add some fallback?

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -235,6 +235,17 @@ with the LDAP server.
 This change improves the resilience of the system when there are temporary issues with the LDAP server, ensuring that local users can still be accessed even if the LDAP search fails.
 If a local user is linked to a failing LDAP provider, the user will be marked as disabled and read-only until the LDAP server is available again.
 
+=== The `serverinfo` endpoint only returns the system info for administrators in the administrator realm
+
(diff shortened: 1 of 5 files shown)


rmartinc:
As we commented in yesterday's meeting I completed option 2. Upgrading note added too. If the CI passes successfully I will move the PR out of draft.

mposolda (approved):
@rmartinc Thanks!

PR #42866, opened 2025-09-23
User session in LOGIN event

rublin:
Closes #42867 

Include the user session in the LOGIN event when consent was processed

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/LoginActionsService.java
+++ b/services/src/main/java/org/keycloak/services/resources/LoginActionsService.java
@@ -1070,9 +1070,10 @@ public class LoginActionsService {
         }
 
         event.detail(Details.CONSENT, Details.CONSENT_VALUE_CONSENT_GRANTED);
-        event.success();
 
         ClientSessionContext clientSessionCtx = AuthenticationProcessor.attachSession(authSession, null, session, realm, clientConnection, event);
+        event.session(clientSessionCtx.getClientSession().getUserSession());
+        event.success();
         return AuthenticationManager.redirectAfterSuccessfulFlow(session, realm, clientSessionCtx.getClientSession().getUserSession(), clientSessionCtx, request, session.getContext().getUri(), clientConnection, event, authSession);
     }
 
--- a/services/src/main/java/org/keycloak/services/resources/LoginActionsService.java
+++ b/services/src/main/java/org/keycloak/services/resources/LoginActionsService.java
@@ -1072,7 +1072,6 @@ public class LoginActionsService {
         event.detail(Details.CONSENT, Details.CONSENT_VALUE_CONSENT_GRANTED);
 
         ClientSessionContext clientSessionCtx = AuthenticationProcessor.attachSession(authSession, null, session, realm, clientConnection, event);
-        event.session(clientSessionCtx.getClientSession().getUserSession());
         event.success();
         return AuthenticationManager.redirectAfterSuccessfulFlow(session, realm, clientSessionCtx.getClientSession().getUserSession(), clientSessionCtx, request, session.getContext().getUri(), clientConnection, event, authSession);
     }


mposolda:
@rublin Thanks for the PR. It will be good to have automated test for this or make sure that some existing test is updated as we already test events inside our automated tests. There is the test `OAuth2GrantTest` , which include some scenarios for the consents. At the same time, it includes the test `OAuth2GrantTest.oauthGrantAcceptTest()` , which seems to already check that session is available. So it seems that the code branch you are fixing has some other scenario than `OAuth2GrantTest.oauthGrantAcceptTest()` is fixing?

Anyway, will be good to write an automated test in a way that it fails before your changes and it passes after your change. 

Also please add the line `closes #42867` to the commit message (See other pull requests and see Keycloak contributions guide for the details).

services/src/main/java/org/keycloak/services/resources/LoginActionsService.java
```diff
@@ -1070,9 +1070,10 @@ public Response processConsent() {
         }
 
         event.detail(Details.CONSENT, Details.CONSENT_VALUE_CONSENT_GRANTED);
-        event.success();
 
         ClientSessionContext clientSessionCtx = AuthenticationProcessor.attachSession(authSession, null, session, realm, clientConnection, event);
+        event.session(clientSessionCtx.getClientSession().getUserSession());
```
rmartinc: This line is not needed. I think that the `attachsession` method already adds the session to the context ([here](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java#L1204-L1206)). The only needed thing is executing the `success` after as you are already doing.

rmartinc (changes requested):
@mposolda I think that the problem here is that we are using the `TestEventsListenerProvider`. The event is retained in memory and the `sessionId` is added later and conserved (test works OK). But another provider will receive the event without the sessionId and fail. So it's complicated to test this one with the CI. I would just move the `success` to the end and that's all.

@rublin I think the set line is not even needed. Can you test just moving the success to the end as you are already doing? If it works I don't think we need a test. because, as I said, we are testing the sessionId is filled but our provider is in memory, therefore even adding it later was OK.

mposolda:
> @mposolda I think that the problem here is that we are using the `TestEventsListenerProvider`. The event is retained in memory and the `sessionId` is added later and conserved (test works OK). But another provider will receive the event without the sessionId and fail. So it's complicated to test this one with the CI. I would just move the `success` to the end and that's all.
> 
> @rublin I think the set line is not even needed. Can you test just moving the success to the end as you are already doing? If it works I don't think we need a test. because, as I said, we are testing the sessionId is filled but our provider is in memory, therefore even adding it later was OK.

@rmartinc Good to know, but this looks like a bug in `TestEventsListenerProvider` if it reports that `sessionId` is in the event, but in the reality, the `sessionId` is not in the events (if it is other provider)? As the point of `TestEventsListenerProvider` is to properly test events in same way like the "real" event listeners would do, but here it seems it does not doing that and it rather creates false positives about sessionID being present in the event when in reality it is not present.

Maybe we should add copies of the instances in the `TestEventsListenerProvider.onEvent` methods instead of adding same instances of the event objects, which are then updated later (Both `Event` and `AdminEvent` have helper methods for creating easily the object copy). I guess lots of the tests will start to fail once we do this, but maybe worth doing that event and fix the event bugs like this one? I can create GH issue for this and handle it as a follow-up?

services/src/main/java/org/keycloak/services/resources/LoginActionsService.java
rublin: @rmartinc You are absolutely right. I checked, and it's enough to change the order.

rmartinc:
@mposolda Yes, we can clone the events in the test provider to avoid the issue. But I would do it as a follow-up, if not @rublin maybe needs to fix several tests. :smile:

rmartinc (approved):
Thanks @rublin for the PR! LGTM!

@mposolda I agree with you. We can create a follow-up to clone events in the test provider. Don't know how many tests will fail (maybe we are lucky and it's OK). Run the CI when you can.

rmartinc:
@mposolda I created https://github.com/keycloak/keycloak/issues/42914 for the follow-up. I will give it a try to check if there there several tests affected or not (although it seems that today GHA are very slow).

mposolda (approved):
@rmartinc Cool, Thanks for creating follow-up and looking at it.

@rublin Thanks for your last update. I hope to merge your PR if tests are OK

PR #42884, opened 2025-09-23
perf: using resourceAsStream rather than zip

### LINKED ISSUE #42882: [PERF] Persisted config source - load properties
### Magnitude
**🟡 MEDIUM**

**Consumed samples (~)**: `54s` 
**Memory allocation (~)**: `400KB`

### Description
In the `PersistedConfigSource` class, the `loadConfigSource()` resolves the `generated-bytecode.jar` and get the `META-INF/keycloak-persisted.properties`. We might try to find the way how to hand over the persisted properties to the runtime without the additional parsing during the runtime.

```java
File configFile = homePath.resolve("lib").resolve("quarkus").resolve("generated-bytecode.jar").toFile();

        if (!configFile.exists()) {
            return null;
        }

        try (ZipInputStream is = new ZipInputStream(new FileInputStream(configFile))) {
            ZipEntry entry;

            while ((entry = is.getNextEntry()) != null) {
                if (entry.getName().equals(PERSISTED_PROPERTIES)) {
                    return new ByteArrayInputStream(is.readAllBytes());
                }
            }
        } catch (Exception cause) {
            throw new RuntimeException("Failed to load persisted properties from " + configFile, cause);
        }
```

Image

shawkins:
closes: #42882

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/Environment.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/Environment.java
@@ -54,10 +54,6 @@ public final class Environment {
         return Boolean.getBoolean("quarkus.launch.rebuild");
     }
 
-    public static Boolean isRuntimeMode() {
-        return !isRebuild();
-    }
-
     public static String getHomeDir() {
         return System.getProperty(KC_HOME_DIR);
     }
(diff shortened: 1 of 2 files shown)


shawkins:
Updated based upon @mabartos feedback on the issue.

mabartos (approved):
@shawkins Much better, thanks. LGTM

Image

PR #42906, opened 2025-09-24
Introduce pending email verification message for UPDATE_EMAIL

### LINKED ISSUE #42770: Introduce pending email verification message for UPDATE_EMAIL
### Before reporting an issue

### Area

user-profile

### Describe the bug

Enable update email with force email verification flag ON.
Create a user that doesn't have an email set
Assign UPDATE PROFILE action to the user.
Try to authenticate with the user, fill up the update profile form and submit
You will see a message that verification email was sent
Try to authenticate again and you will see again update email page with prefilled new email. This is confusing.
There should be some message about pending email verification with ability to resend the verification link or update the email to different email. 

### Version

main

### Regression

### Expected behavior

There should be pending email verification  message 

### Actual behavior

See description

### How to Reproduce?

See description

### Anything else?

_No response_

martin-kanis:
Closes #42770

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpReviewProfileAuthenticator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpReviewProfileAuthenticator.java
@@ -194,6 +194,11 @@ public class IdpReviewProfileAuthenticator extends AbstractIdpAuthenticator {
             public String getFederationLink() {
                 return null;
             }
+
+            @Override
+            public Stream<String> getRequiredActionsStream() {
+                return Stream.empty();
+            }
         };
 
         UserProfileProvider profileProvider = context.getSession().getProvider(UserProfileProvider.class);
--- a/services/src/main/java/org/keycloak/authentication/requiredactions/VerifyEmail.java
+++ b/services/src/main/java/org/keycloak/authentication/requiredactions/VerifyEmail.java
@@ -73,8 +73,13 @@ public class VerifyEmail implements RequiredActionProvider, RequiredActionFactor
     @Override
     public void evaluateTriggers(RequiredActionContext context) {
         if (context.getRealm().isVerifyEmail() && !context.getUser().isEmailVerified()) {
-            context.getUser().addRequiredAction(UserModel.RequiredAction.VERIFY_EMAIL);
-            logger.debug("User is required to verify email");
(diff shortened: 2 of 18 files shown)


pedroigor (commented):
The problem with using the cache is that it won't survive server restarts, thus the marker for pending verification will be lost.

Instead, I think we can consider replacing the single-use-object with an user attribute. We also need to make sure this attribute is only writable by administrators and not by users.

We already have user attributes that serve as markers/controls, such as temporary admin user or even in brute force.

I'm also wondering if we can have a different flow where the email is not updated at all during update profile and the user is forced to update the email and verify whenever authenticating.

Wdyt?

services/src/main/java/org/keycloak/authentication/requiredactions/VerifyEmail.java
```diff
@@ -73,8 +73,13 @@ public class VerifyEmail implements RequiredActionProvider, RequiredActionFactor
     @Override
     public void evaluateTriggers(RequiredActionContext context) {
         if (context.getRealm().isVerifyEmail() && !context.getUser().isEmailVerified()) {
-            context.getUser().addRequiredAction(UserModel.RequiredAction.VERIFY_EMAIL);
-            logger.debug("User is required to verify email");
+            // Don't add VERIFY_EMAIL if UPDATE_EMAIL is already present (UPDATE_EMAIL takes precedence)
+            if (context.getUser().getRequiredActionsStream().noneMatch(action -> UserModel.RequiredAction.UPDATE_EMAIL.name().equals(action))) {
```
pedroigor, line 77: @martin-kanis Do we still need this?

martin-kanis: I think we can remove it

services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpReviewProfileAuthenticator.java
```diff
@@ -194,6 +194,11 @@ public String getServiceAccountClientLink() {
             public String getFederationLink() {
                 return null;
             }
+
+            @Override
+            public Stream<String> getRequiredActionsStream() {
```
martin-kanis, line 199: @pedroigor  Out of curiosity, why do we need this?

pedroigor: Because without it you'll get an NPE during broker flows when updating the user profile.

pedroigor (approved):

PR #42923, opened 2025-09-24
Avoid invalidating the realm when managing client initial access

### LINKED ISSUE #42922: Dynamic Client Registration invalidates the realm cache
### Before reporting an issue

### Area

core

### Describe the bug

When using dynamic client registration, and even when looking at the Admin UI at the keys, the realm data is invalidated. This can have a big performance impact when registering new clients dynamically.

### Version

26.2, main

### Regression

### Expected behavior

The realm data shouldn't be invalidated. 

### Actual behavior

The realm data is invalidated.

### How to Reproduce?

Either of the two: 

* Dynamically register a client
* Go to the admin UI, open the tab with "Initial access tokens" in the client menu

### Anything else?

I'll prepare a PR

ahus1:
Closes #42922

This can be manually validated in one of the two following ways - I actually don't know how I would automatically test this, please share ideas if you could think of some. 

* Go to the Admin UI clients menu, and then access "Initial access token". Even selecting an entry would invalidate the cache previously. Now it wouldn't do that any more. 

OR

* Perform a client registration via https://www.keycloak.org/securing-apps/client-registration

You then look that no Realm invalidations are created in the debugger at 

https://github.com/keycloak/keycloak/blob/18a6c790113998bf039f46d6c71ffb668e961036/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/events/RealmUpdatedEvent.java#L39

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
@@ -1800,28 +1800,32 @@ public class RealmAdapter implements CachedRealmModel {
 
     @Override
     public ClientInitialAccessModel createClientInitialAccessModel(int expiration, int count) {
-        getDelegateForUpdate();
-        return updated.createClientInitialAccessModel(expiration, count);
+        // This does not call getDelegateForUpdate() this data is never cached, and calling it would invalidate all cached realm data
+        return modelSupplier.get().createClientInitialAccessModel(expiration, count);
     }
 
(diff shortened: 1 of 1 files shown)


ahus1:
@pruivo - do you feel confident enough to review this PR, assuming that you haven't worked with client invalidation before?

pruivo (approved):
LGTM. If I didn't miss anything, it does not interact with any cached data.

Whether those methods belong to `RealModel` is another story :)  They exist in `RealmProvider`, and all invocations appear to be directed here.

Issue #40988, opened 2025-07-08
Passkeys Enhancements


theorlandog:
### Description

Hi everyone. The passkeys enhancements in 26.3.0 are looking great so far. 

Now that passkeys are more mainstream, it would be cool to make it configurable at the realm level to enable a prompt to ask to register a passkey to the current device if an external passkey login is used (i.e. qr code from phone). 

To that same end, is there any easy way to setup a login flow to require MFA after the username password form, except if a passkey is used?

Thank you for doing great work with the passkeys feature.

### Motivation

Passkeys that aren’t used don’t do any good. The world can’t ditch passwords unless we make it easy to nudge users into pattersns like confidently managing redundant devices or using synced password vaults.

theorlandog:
Want to add some more context to this issue. 

I would like to require TOTP if username and password is used in the username password form, but not require TOTP if a passkey (passwordless or normal webauthn) is used. I am trying the following auth flow using Level of Authentication (LOA).

<img width="1614" height="902" alt="Image" src="https://github.com/user-attachments/assets/5e0c38b1-f955-452a-851e-6da03aed4470" />

However, this still prompts me for TOTP when I used my passkey to login through the new username password form. Any tips for how to get the configuration I am looking for?

andreasgrill:
I am also interested in finding a way to achieve this. We want to promote the use of Passkeys for our users by not requiring them to provide a second factor (TOTP) if they use Passkeys.

cn42:
I think it should be tracked in the main issue #23656

graziang:
Hi @theorlandog , thanks for this issue. The use case in the description seems different from the one in the first comment, could you create two separate issues so we can track them more clearly?

graziang:
Issue for the enhancement of the first comment: #41354

Lexachoc:
I think there is a condition where we can bypass OTP when a passkey is used:

Add condition under the flow `Browser - Conditional OTP`:

<img width="2170" height="278" alt="Image" src="https://github.com/user-attachments/assets/217bb53a-e853-4e69-9dac-0def7a824875" />

<img width="924" height="988" alt="Image" src="https://github.com/user-attachments/assets/ee08537c-1634-439e-88ad-6a02bd0d1008" />

and configure it like this:

<img width="619" height="672" alt="Image" src="https://github.com/user-attachments/assets/9d6c8001-7c93-4f79-936e-26bd4b6acf73" />



[...]

PR #42932, opened 2025-09-24
Detect if Passkey Challenge Was Completed on the Current Device

malwaregeeek:
## **What**

Adds the `authenticatorAttachment` field to the WebAuthn authentication response, allowing detection of whether a passkey challenge was completed using a platform authenticator (current device) or through cross-device authentication.

Closes #42931

##  **Why**

Currently, there is no reliable way to detect whether a user completes the passkey challenge on their current device or via a different device. This information is crucial for:

- **Conditional enrollment flows**: Show WebAuthn enrollment screen only when no passkeys are found on the current device
- **Improved UX**: Skip enrollment prompts when platform authenticators are available
- **Multi-device support**: Better handling of passkey registration across multiple devices

## **How** 
The authenticatorAttachment field can return one of the following

**platform** → authenticators built into the device (e.g. Touch ID, Face ID, Windows Hello, Android fingerprint).

**cross-platform** → roaming authenticators (e.g. USB/Bluetooth/NFC security keys like YubiKey).


malwaregeeek:
@tnorimat If you can take a look, will appreciate that :)

ahus1:
Pinging the @keycloak/core-clients team as well as this is passkeys related.

themes/src/main/resources/theme/base/login/resources/js/webauthnAuthenticate.js
```diff
@@ -91,6 +91,9 @@ export function returnSuccess(result) {
     if (result.response.userHandle) {
         document.getElementById("userHandle").value = base64url.stringify(new Uint8Array(result.response.userHandle), { pad: false });
     }
+    if (result.response.platformAttachment) {
```
rmartinc: If I'm not wrong this should be [result.authenticatorAttachment](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/authenticatorAttachment) which can have the values `platform` or `cross-platform`. Or am I missing something?

rmartinc (changes requested):
@malwaregeeek I think that the property is incorrect and it should be `result.authenticatorAttachment`. Right?

I also think that this PR does nothing for the moment. I don't know if you want to do something more in subsequent PRs but, with the current change, no new feature or enhancement is achieved. I would say that you at least need to recover the passed property at server side and attach it in the login event. This way you can create a test that checks that the `authenticatorAttachment` is at least shown in the event with the correct value (platform or cross-platform). If you want to achieve what is proposed in  https://github.com/keycloak/keycloak/issues/40988 you will need more PRs.

malwaregeeek:
> @malwaregeeek I think that the property is incorrect and it should be `result.authenticatorAttachment`. Right?
> 
> I also think that this PR does nothing for the moment. I don't know if you want to do something more in subsequent PRs but, with the current change, no new feature or enhancement is achieved. I would say that you at least need to recover the passed property at server side and attach it in the login event. This way you can create a test that checks that the `authenticatorAttachment` is at least shown in the event with the correct value (platform or cross-platform). If you want to achieve what is proposed in #40988 you will need more PRs.

@rmartinc Thanks for reviewing the PR! You’re right — that was a typo, it should have been `authenticatorAttachment` (now corrected). This PR doesn’t change any behavior on its own; it’s more of a preparatory step. That said, it could still be useful for folks customizing the `WebAuthnAuthenticator`, since they wouldn’t need to modify `webauthnAuthenticate.js.`

My plan is to follow up with another PR that updates the `WebAuthnAuthenticator` to read this additional form field and set an `authNote` with the authenticatorAttachment type. That authNote can then be used in `WebAuthnRegisterRequiredAction` (in the `evaluateTrigger` method) to decide when the registration prompt should be shown.

pnzrr:
@rmartinc as @malwaregeeek described, it's as much preparatory as anything, the current webauthnAuthenticate.js isn't passing the whole object through, which if you want to use missing parts, means you have to fork the file and add to it, when it should just do that be default.

rmartinc:
@malwaregeeek @pnzrr No, no, I don't need all the missing parts in this PR, but I would prefer to have something that really tests the change. That's why I commented to at least receive the new field in the server side and add it to an event detail of the login (for example). If you only want to add the note for the moment, I think that that can also be added here, It's not something big. You can think about modifying a test later that checks if the note is there somehow. What I'm asking is that I would prefer to have a test for the change, because, for example, I still think the assigment is not OK now, I think that is should be `result.authenticatorAttachment` and not `result.response.authenticatorAttachment` and, because there is no test, it's not detected.

tnorimat:
@malwaregeeek Hello, as mentioned in [the keycloak's blog](https://www.keycloak.org/2026/02/deprecating-arquillian-and-keycloak-test-framework-support), the tests should be written by using junit-based new test framework instead of arqullian.

PR #42934, opened 2025-09-25
Add workflow condition that uses boolean expressions to combine and n…

### LINKED ISSUE #42583: [RLM] Allow workflow conditions to be negated
### Description

At the moment there's no capability in place to allow admins to negate a condition (e.g. user who **are not** members of group A). We need to allow for this configuration both at the condition and the workflow level.

At the condition level, this simply negates that particular condition. At the workflow level, it would negate the overall condition setup of the workflow. If the workflow has only one condition, negating the condition itself or the workflow condition would yield the same result. However, if the workflow has multiple conditions it would negate the overall setup.

Example: Workflow has 2 conditions, user is a member of group A AND has role R. If we want now the opposite of that - i.e. user not in group A OR not in role R - we would negate the whole workflow condition, not the individual conditions.

sguilhen:
…egate conditions

Closes #42583

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/workflow/conditions/ExpressionWorkflowConditionProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/workflow/conditions/ExpressionWorkflowConditionProvider.java
@@ -1,0 +1,82 @@
+package org.keycloak.models.workflow.conditions;
+
+import jakarta.persistence.criteria.CriteriaBuilder;
+import jakarta.persistence.criteria.CriteriaQuery;
+import jakarta.persistence.criteria.Predicate;
+import jakarta.persistence.criteria.Root;
+import org.antlr.v4.runtime.CharStream;
+import org.antlr.v4.runtime.CharStreams;
+import org.antlr.v4.runtime.CommonTokenStream;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.workflow.WorkflowConditionProvider;
+import org.keycloak.models.workflow.WorkflowEvent;
+import org.keycloak.models.workflow.WorkflowInvalidStateException;
+import org.keycloak.models.workflow.conditions.expression.BooleanConditionEvaluator;
+import org.keycloak.models.workflow.conditions.expression.BooleanConditionLexer;
+import org.keycloak.models.workflow.conditions.expression.BooleanConditionParser;
+import org.keycloak.models.workflow.conditions.expression.BooleanConditionParser.EvaluatorContext;
+import org.keycloak.models.workflow.conditions.expression.ErrorListener;
+import org.keycloak.models.workflow.conditions.expression.PredicateConditionEvaluator;
+
+import java.util.stream.Collectors;
+
+public class ExpressionWorkflowConditionProvider implements WorkflowConditionProvider {
+
+    private final String expression;
+    private final KeycloakSession session;
+    private EvaluatorContext evaluatorContext;
+
+    public ExpressionWorkflowConditionProvider(KeycloakSession session, String expression) {
+        this.session = session;
+        this.expression = expression;
+    }
+
+    @Override
+    public boolean evaluate(WorkflowEvent event) {
+        validate();
+        BooleanConditionEvaluator evaluator = new BooleanConditionEvaluator(session, event);
+        return evaluator.visit(this.evaluatorContext);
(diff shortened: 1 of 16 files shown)


model/jpa/src/main/java/org/keycloak/models/workflow/conditions/ExpressionWorkflowConditionProvider.java
```diff
@@ -0,0 +1,82 @@
...
+        validate();
+        PredicateConditionEvaluator evaluator = new PredicateConditionEvaluator(session, cb, query, userRoot);
+        return evaluator.visit(this.evaluatorContext);
+    }
+
+    @Override
+    public void validate() {
+        // to properly validate the expression, we need to parse it. We then cache the parsed context if the expression is valid
+        // so we don't to parse it again if validate is called again on the same instance of the provider
+        if (this.evaluatorContext == null) {
```
pedroigor, line 52: The `ComponentModel` allows you to set state into notes, which will end up in the (local) cache. In this case, you will only parse the expression once across multiple invocations of the condition provider.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/BooleanConditionEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+            return visit(ctx.conditionCall());
+        }
+        return visit(ctx.expression());
+    }
+
+    @Override
+    public Boolean visitConditionCall(BooleanConditionParser.ConditionCallContext ctx) {
+        String conditionName = ctx.Identifier().getText();
+
+        KeycloakSessionFactory sessionFactory = session.getKeycloakSessionFactory();
```
pedroigor: See `org.keycloak.models.workflow.WorkflowsManager#getConditionProvider`.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/PredicateConditionEvaluator.java
```diff
@@ -0,0 +1,100 @@
...
+            return visit(ctx.conditionCall());
+        }
+        return visit(ctx.expression());
+    }
+
+    @Override
+    public Predicate visitConditionCall(BooleanConditionParser.ConditionCallContext ctx) {
+        String conditionName = ctx.Identifier().getText();
+
+        KeycloakSessionFactory sessionFactory = session.getKeycloakSessionFactory();
```
pedroigor: See `org.keycloak.models.workflow.WorkflowsManager#getConditionProvider`.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/BooleanConditionEvaluator.java
sguilhen: I don't think that will help much here because I need to call the "create" variant that takes a list of parameters as opposed to what is done in `WorkflowsManager`.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/ExpressionWorkflowConditionProvider.java
sguilhen: true, but we don't have access to the component model here, and I think we don't even have one for the conditions as they seem to be stored directly in the workflow's component model with a prefix.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/BooleanConditionEvaluator.java
pedroigor: My point is that is better to have a single place/validation from where factories/providers are fetched rather than duplicate code.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/ExpressionWorkflowConditionProvider.java
pedroigor: Can't you pass the model to the provider? As I mentioned, the cache for the parsed expression will live longer and avoid unnecessary parsing across multiple invocations of this provider.

pedroigor (approved):

PR #42946, opened 2025-09-25
[PERF] Jackson reflection-free serialization/deserialization

### LINKED ISSUE #42945: [PERF] Jackson reflection-free serialization/deserialization
### Description

Quarkus provides the ability to use reflection-free serializers for objects via Jackson. Our resource objects handle HTTP calls, where the serialization needs to be done for our representations.  Jackson heavily uses reflection when executing the serialization/deserialization. Quarkus provides a [guide](https://quarkus.io/blog/quarkus-metaprogramming/) on how to make an optimization to our JAX-RS layer. Quarkus generates a bytecode via Gizmo for the serializer/deserializer that reduces the reflection usage and should be more efficient.

In this task, we should analyze what improvements it could bring us and find if we'd be able to use this optimization as based on the info in the guide, Quarkus experienced a 12% improvement in the endpoints throughput.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mabartos:
- Closes #42945

I've followed up on the [guide](https://quarkus.io/blog/quarkus-metaprogramming/) provided by the Quarkus team. The Jackson-free serdes was mentioned during the presentation about Observability performance in Quarkus, so they're including that even in their benchmarks.

In the guide, there's mentioned ~12% improvement in the throughput for HTTP calls. I made a brief performance analysis, and these are my findings:

## Performance analysis

- Run official KC distribution (OLD) and official KC distribution with the optimization (NEW)
- Used `wrk` from the [Hyperfoil](https://github.com/Hyperfoil/Hyperfoil)

### Test characteristic
- Benchmarking how many req/s can KC server handle on one node
- Only the `host:8080/admin/realms/master/clients` endpoint was called (multiple clients that are deserialized) with the access token 
- On a client, changed how many threads and connections are used when introducing the load on the server
-- Matrix 1: 3 threads, 3 connections
-- Matrix 2: 10 threads, 10 connections
-- Matrix 3: 15 threads, 15 connections
- Every matrix run for 15s
- Every matrix is executed 5x

### Matrix 1 (3 threads, 3 connections)
**Requests per second**
| Run | Old  | New  | Diff % |
|-----|-----------|-----------|-------------|
| 1   | 1224      | 1385      | ⬆️ **13.15** ✅ |
| 2   | 1494      | 1486      | ⬇️ *-0.54* ❌ |
| 3   | 1483      | 1550      | ⬆️ **4.52** ✅ |
| 4   | 1501      | 1491      | ⬇️ *-0.67* ❌ |
| 5   | 1492      | 1634      | ⬆️ **9.52** ✅ |

Average: **5.196% improvement** ✅

### Matrix 2 (10 threads, 10 connections)
**Requests per second**
| Run | Old  | New  | Diff % |
|-----|-----------|-----------|-------------|
| 1   | 3455      | 3636      | ⬆️ **5.24** ✅ |
| 2   | 3416      | 3504      | ⬆️ **2.58** ✅ |
| 3   | 3395      | 3496      | ⬆️ **2.97** ✅ |
| 4   | 3341      | 3419      | ⬆️ **2.33** ✅ |
| 5   | 3347      | 3738      | ⬆️ **11.68** ✅ |

Average: **4.96% improvement** ✅

### Matrix 3 (15 threads, 15 connections)
**Requests per second**
| Run | Old  | New  | Diff % |
|-----|-----------|-----------|-------------|
| 1   | 3268      | 3100      | ⬇️ *-5.14* ❌ |
| 2   | 4034      | 4278      | ⬆️ **6.05** ✅ |
| 3   | 3976      | 4586      | ⬆️ **15.34** ✅ |
| 4   | 4000      | 4272      | ⬆️ **6.80** ✅ |
| 5   | 3993      | 4281      | ⬆️ **7.21** ✅ |

Average: **6.052% improvement** ✅

### Augmentation time
As some work is moved to the build phase, we experience a small increase in the augmentation time (which should be ok):

| Run | Old  | New  | Diff | Diff % |
|-----|------|------|------|--------|
| 1   | 4341ms | 4501ms | ⬆️ *160ms* ❌ | ⬆️ *3.69%* ❌ |
| 2   | 4482ms | 5131ms | ⬆️ *649ms* ❌ | ⬆️ *14.48%* ❌ |
| 3   | 8233ms | 8630ms | ⬆️ *397ms* ❌ | ⬆️ *4.82%* ❌ |
| 4   | 7552ms | 7691ms | ⬆️ *139ms* ❌ | ⬆️ *1.84%* ❌ |

Average Old: **6152ms**
Average New: **6488ms**
Difference: **+336ms** ❌
Difference in %: **+6.20%** ❌

### Node characteristics
```
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             46 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      20
  On-line CPU(s) list:       0-19
Vendor ID:                   GenuineIntel
  Model name:                12th Gen Intel(R) Core(TM) i7-1280P
    CPU family:              6
    Model:                   154
    Thread(s) per core:      2
    Core(s) per socket:      14
Memory:                      32GB
```

## Conclusion

In total, in this brief performance analysis, we experience **5.40%** ✅ improvement in the throughput in our API. Note that values for some runs differ as it might be influenced by other tasks done on the CPU and not related to this benchmark. This test only gets clients from the realm, so for some endpoints, where the serdes is not used much does not have a big effect. On the other hand, the effect should be higher when serdes is used frequently. 

I think it's a nice improvement, and there should not be any issues I'm not aware of. 

cc: @keycloak/cloud-native @keycloak/sre
cc: @franz1981 @brunobat

### DIFF
--- a/docs/guides/server/configuration-production.adoc
+++ b/docs/guides/server/configuration-production.adoc
@@ -88,7 +88,7 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
 
 See <@links.server id="caching" anchor="network-bind-address"/> for more details.
 
-== Better HTTP performance
+== Preview of enhanced HTTP performance
 <@features.techpreview feature="http-optimized-serializers" additionalCommunityText="We gather more feedback on this feature to promote it to supported. Please, share your feedback about any issue in https://github.com/keycloak/keycloak/discussions/43484[this discussion]."/>
 
 In production environments, the performance of the HTTP layer is critical.
--- a/docs/guides/server/configuration-production.adoc
+++ b/docs/guides/server/configuration-production.adoc
@@ -102,7 +102,7 @@ By removing the overhead of reflection during serialization/deserialization, it 
 
 These improvements help ensure smoother, more predictable performance at scale while also lowering the operational cost of running production systems.
 
-The only known tradeoff is that build(augmentation) time increased by ~6% as certain actions were moved to the build time instead of runtime.
+The only known tradeoff is that build time increases by ~6% as certain actions were moved to the build time instead of runtime.
 
 You can enable this feature as follows:
 
--- a/docs/guides/server/configuration-production.adoc
+++ b/docs/guides/server/configuration-production.adoc
@@ -1,6 +1,7 @@
 <#import "/templates/guide.adoc" as tmpl>
 <#import "/templates/kc.adoc" as kc>
 <#import "/templates/links.adoc" as links>
+<#import "/templates/features.adoc" as features>
 
 <@tmpl.guide
 title="Configuring {project_name} for production"
@@ -88,13 +89,13 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
 See <@links.server id="caching" anchor="network-bind-address"/> for more details.
 
 == Better HTTP performance
-NOTE: This section is describing a feature which is currently in *preview* as we gather more feedback about potential issues. We appreciate any feedback.
+<@features.techpreview feature="http-optimized-serializers" additionalCommunityText="We gather more feedback on this feature to promote it to supported. Please, share your feedback about any issue in https://github.com/keycloak/keycloak/discussions/43484[this discussion]."/>
 
 In production environments, the performance of the HTTP layer is critical.
 Every request passes through it, making it a key factor in overall system responsiveness, scalability, and user experience.
 
-This feature introduces Jackson JSON reflection-free serializers to optimize HTTP processing.
-By removing the overhead of reflection during serialization/deserialization, it delivers measurable benefits:
+This feature improves how {project_name} handles JSON data in HTTP requests and responses.
+The result is a more efficient runtime with measurable benefits:
 
 - ~5% increase in throughput
 - More stable response times
@@ -107,7 +108,5 @@ The only known tradeoff is that build time increases by ~6% as certain actions w
 You can enable this feature as follows:
 
 <@kc.start parameters="--features=http-optimized-serializers"/>
-
-The feature is *disabled by default*, as we gather more feedback.
 
 </@tmpl.guide>
--- a/docs/guides/server/configuration-production.adoc
+++ b/docs/guides/server/configuration-production.adoc
@@ -87,4 +87,27 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
 
 See <@links.server id="caching" anchor="network-bind-address"/> for more details.
 
+== Better HTTP performance
+NOTE: This section is describing a feature which is currently in *preview* as we gather more feedback about potential issues. We appreciate any feedback.
+
+In production environments, the performance of the HTTP layer is critical.
+Every request passes through it, making it a key factor in overall system responsiveness, scalability, and user experience.
+
+This feature introduces Jackson JSON reflection-free serializers to optimize HTTP processing.
+By removing the overhead of reflection during serialization/deserialization, it delivers measurable benefits:
+
+- ~5% increase in throughput
+- More stable response times
+- Reduced system resource usage
+
+These improvements help ensure smoother, more predictable performance at scale while also lowering the operational cost of running production systems.
+
+The only known tradeoff is that build(augmentation) time increased by ~6% as certain actions were moved to the build time instead of runtime.
+
+You can enable this feature as follows:
+
+<@kc.start parameters="--features=http-optimized-serializers"/>
+
+The feature is *disabled by default*, as we gather more feedback.
+
 </@tmpl.guide>
--- a/docs/guides/templates/features.adoc
+++ b/docs/guides/templates/features.adoc
@@ -12,7 +12,6 @@
 <#macro techpreview feature additionalCommunityText="">
 <#assign profileFeature = ctx.features.getFeature(feature)>
 
-<#if !profileFeature.enabled>
 [NOTE]
 ====
 ${profileFeature.description} is
@@ -28,13 +27,7 @@ ifeval::[{project_community}==true]
 ${additionalCommunityText!""}
 endif::[]
 
-To enable start the server with `--features=preview` or `--features=${profileFeature.name}`
+To enable start the server with <#if profileFeature.type != "PREVIEW_DISABLED_BY_DEFAULT">`--features=preview` or </#if>`--features=${profileFeature.name}`
 
 ====
-<#else>
-[NOTE]
-====
-${profileFeature.description} is *Technology Preview* and is not fully supported.
-====
-</#if>
 </#macro>
--- a/docs/guides/templates/features.adoc
+++ b/docs/guides/templates/features.adoc
@@ -8,3 +8,33 @@
 </#list>
 |===
 </#macro>
+
+<#macro techpreview feature additionalCommunityText="">
+<#assign profileFeature = ctx.features.getFeature(feature)>
+
+<#if !profileFeature.enabled>
+[NOTE]
+====
+${profileFeature.description} is
+ifeval::[{project_product}==true]
+*Technology Preview*
+endif::[]
+ifeval::[{project_community}==true]
+*Preview*
+endif::[]
+and is not fully supported. This feature is disabled by default.
+
+ifeval::[{project_community}==true]
+${additionalCommunityText!""}
+endif::[]
+
+To enable start the server with `--features=preview` or `--features=${profileFeature.name}`
+
+====
+<#else>
+[NOTE]
+====
+${profileFeature.description} is *Technology Preview* and is not fully supported.
+====
+</#if>
+</#macro>
--- a/quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
@@ -150,5 +150,4 @@ public class HttpOptions {
             .description("Service level objectives for HTTP server requests. Use this instead of the default histogram, or use it in combination to add additional buckets. " +
                     "Specify a list of comma-separated values defined in milliseconds. Example with buckets from 5ms to 10s: 5,10,25,50,250,500,1000,2500,5000,10000")
             .build();
-
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,10 +140,11 @@ jobs:
         uses: ./.github/actions/integration-test-setup
 
       - name: Run base tests
+        # enable the http-optimized-serializers feature for the old testsuite to verify it works as expected
         run: |
           TESTS=`testsuite/integration-arquillian/tests/base/testsuites/base-suite.sh ${{ matrix.group }}`
           echo "Tests: $TESTS"
-          ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Dtest=$TESTS -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
+          ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Dauth.server.feature=http-optimized-serializers -Dtest=$TESTS -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
 
       - uses: ./.github/actions/upload-flaky-tests
         name: Upload flaky tests
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -148,7 +148,7 @@ public class Profile {
 
         DB_TIDB("TiDB database type", Type.EXPERIMENTAL),
 
-        HTTP_OPTIMIZED_SERIALIZERS("Use reflection-free Jackson JSON serializers for better performance of the HTTP layer", Type.PREVIEW),
+        HTTP_OPTIMIZED_SERIALIZERS("Optimized JSON serializers for better performance of the HTTP layer", Type.PREVIEW),
 
         /**
          * @see <a href="https://github.com/keycloak/keycloak/issues/37967">Deprecate for removal the Instagram social broker</a>.
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -148,6 +148,8 @@ public class Profile {
 
         DB_TIDB("TiDB database type", Type.EXPERIMENTAL),
 
+        HTTP_OPTIMIZED_SERIALIZERS("Use reflection-free Jackson JSON serializers for better performance of the HTTP layer", Type.PREVIEW),
+
         /**
          * @see <a href="https://github.com/keycloak/keycloak/issues/37967">Deprecate for removal the Instagram social broker</a>.
          */
--- a/docs/documentation/release_notes/index.adoc
+++ b/docs/documentation/release_notes/index.adoc
@@ -13,6 +13,9 @@ include::topics/templates/document-attributes.adoc[]
 :release_header_latest_link: {releasenotes_link_latest}
 include::topics/templates/release-header.adoc[]
 
+== {project_name_full} 26.5.0
+include::topics/26_5_0.adoc[leveloffset=2]
+
 == {project_name_full} 26.4.0
 include::topics/26_4_0.adoc[leveloffset=2]
 
--- a/docs/documentation/release_notes/topics/26_5_0.adoc
+++ b/docs/documentation/release_notes/topics/26_5_0.adoc
@@ -3,11 +3,14 @@
 
 = Better HTTP performance
 
-For the HTTP layer, you can now use Jackson JSON reflection-free serializers, improving request handling efficiency.
+You can now enable a more efficient way to handle JSON data in the HTTP layer.
 This change increases throughput by ~5%, stabilizes response times, and reduces system resource usage.
 
 In order to apply it, you need to explicitly enable the feature `http-optimized-serializers`.
 
-NOTE: This feature is *preview* as we gather more feedback about potential issues. We appreciate any feedback.
+NOTE: This feature is *preview*.
+ifeval::[{project_community}==true]
+We gather more feedback about potential issues in https://github.com/keycloak/keycloak/discussions/43484[this discussion]. We appreciate any feedback.
+endif::[]
 
 For more details, see the https://www.keycloak.org/server/configuration-production[Configuring Keycloak for production] guide.
--- a/docs/documentation/release_notes/topics/26_5_0.adoc
+++ b/docs/documentation/release_notes/topics/26_5_0.adoc
@@ -1,0 +1,13 @@
+// Release notes should contain only headline-worthy new features,
+// assuming that people who migrate will read the upgrading guide anyway.
+
+= Better HTTP performance
+
+For the HTTP layer, you can now use Jackson JSON reflection-free serializers, improving request handling efficiency.
+This change increases throughput by ~5%, stabilizes response times, and reduces system resource usage.
+
+In order to apply it, you need to explicitly enable the feature `http-optimized-serializers`.
+
+NOTE: This feature is *preview* as we gather more feedback about potential issues. We appreciate any feedback.
+
+For more details, see the https://www.keycloak.org/server/configuration-production[Configuring Keycloak for production] guide.
--- a/docs/guides/securing-apps/token-exchange.adoc
+++ b/docs/guides/securing-apps/token-exchange.adoc
@@ -1,5 +1,6 @@
 <#import "/templates/guide.adoc" as tmpl>
 <#import "/templates/links.adoc" as links>
+<#import "/templates/features.adoc" as features>
 
 <@tmpl.guide
 title="Configuring and using token exchange"
@@ -327,22 +328,7 @@ s|Subject impersonation (including direct naked impersonation)   | Not implement
 [[_legacy-token-exchange]]
 == Legacy token exchange
 
-:tech_feature_name: Token Exchange
-:tech_feature_id: token-exchange
-
-[NOTE]
-====
-{tech_feature_name} is
-*Preview*
-and is not fully supported. This feature is disabled by default.
-
-To enable start the server with `--features=preview`
-ifdef::tech_feature_id[]
-or `--features={tech_feature_id}`
-endif::[]
-
-{tech_feature_name} is *Technology Preview* and is not fully supported.
-====
+<@features.techpreview feature="token-exchange"/>
 
 [NOTE]
 ====
--- a/docs/maven-plugin/src/main/java/org/keycloak/guides/maven/Features.java
+++ b/docs/maven-plugin/src/main/java/org/keycloak/guides/maven/Features.java
@@ -51,11 +51,9 @@ public class Features {
     public static class Feature {
 
         private final Profile.Feature profileFeature;
-        private final boolean isEnabled;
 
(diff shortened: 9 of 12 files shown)


pruivo:
@mabartos, what is happening in Matrix 1? Win/Lose alternates between runs.
How many clients and what is the JDK version?

mabartos:
> what is happening in Matrix 1? Win/Lose alternates between runs.

We can't ensure the runs are 1:1 mapped (in a sense they have the same env conditions, etc.), f.e. there might be the case when some run for the New loses,  but when compared with other run, it'd win. You can see that for the matrix 1, these values oscillates for the Old in range 1224-1492, where for the New 1385-1634. I haven't checked stdev and other props. 

> How many clients and what is the JDK version?

The default ones - ~5? OpenJDK 21.0.3

@pruivo It'd be nice to compare it in the KC benchmark that could show more interesting data. I can maybe also provide a comparison of flame graphs.

franz1981:
And in term of native version, it is faster? 
Or no difference?

mabartos:
> And in term of native version, it is faster? Or no difference?

@franz1981 We don't use native builds in Keycloak. Or are you referring to sth else?

franz1981:
I remember an issue where keycloak got bad XP with native image but I wasn't aware is not available as native app 🙏

pruivo:
> > what is happening in Matrix 1? Win/Lose alternates between runs.
> 
> We can't ensure the runs are 1:1 mapped (in a sense they have the same env conditions, etc.), f.e. there might be the case when some run for the New loses, but when compared with other run, it'd win. You can see that for the matrix 1, these values oscillates for the Old in range 1224-1492, where for the New 1385-1634. I haven't checked stdev and other props.
> 

hmm, I'm guessing I'm not expecting the value to have such a large range between the best and worst runs. 

> > How many clients and what is the JDK version?
> 
> The default ones - ~5? OpenJDK 21.0.3
> 
> @pruivo It'd be nice to compare it in the KC benchmark that could show more interesting data. I can maybe also provide a comparison of flame graphs.

I can run it locally. Do authentication/token refreshes go through Jackson serialization frequently enough to be measured by KC Benchmark? Or should I create a micro-benchmark (JMH)?

franz1981:
Just FYI, loolking how reflection work, it got bad and worse as the number and type of fields grows and the number of serialized/deserialized types too. Not to mention the RSS consumed due to method handle generation, which won't happen in reflection free case

mabartos:
> hmm, I'm guessing I'm not expecting the value to have such a large range between the best and worst runs.

@pruivo IMO, in order to find out what exactly is happening there, we'd need to go into `wrk` internals, checking TCP connections, how these requests are assigned to those threads, etc. I don't think it's worth the effort to go deeper into that, as it clearly shows the ranges increased with the optimization. Even the range diff of the Old/New is almost the same, so it does not show that the high variability should be caused by Keycloak. 

> Do authentication/token refreshes go through Jackson serialization frequently enough to be measured by KC Benchmark?

AFAIK, not really very frequent. I don't know what are exact current available scenarios in KC benchmark, so I cannot tell what would be the best. The point is, and would be nice-to-have, to run some benchmark with a ~real load and see how it improved the execution.

franz1981:
You can use jbang with wrk2 and have fixed rate with coordinated omission free measurements AND compare latencies.
You can pick the 80/90% of your peak throughput and you are ready to go.
See 
https://github.com/Hyperfoil/jbang-catalog

The suggestion clearly is to make sure your application is the bottleneck which means running it with perf stat or pidstat and make sure is CPU (usr) intensive

pruivo:
@mabartos I run the benchmark locally, so take the results with a grain of salt as I was using the PC while benchmarking. I run twice, and your PR seems more stable. The committed memory at the end of the test is lower with your PR. 

### This PR
```
[benchamrk-run] ================================================================================
[benchamrk-run] ---- Global Information --------------------------------------------------------
[benchamrk-run] > request count                                     793000 (OK=793000 KO=0     )
[benchamrk-run] > min response time                                      2 (OK=2      KO=-     )
[benchamrk-run] > max response time                                    869 (OK=869    KO=-     )
[benchamrk-run] > mean response time                                    16 (OK=16     KO=-     )
[benchamrk-run] > std deviation                                         21 (OK=21     KO=-     )
[benchamrk-run] > response time 50th percentile                          9 (OK=9      KO=-     )
[benchamrk-run] > response time 75th percentile                         16 (OK=16     KO=-     )
[benchamrk-run] > response time 95th percentile                         62 (OK=62     KO=-     )
[benchamrk-run] > response time 99th percentile                        102 (OK=102    KO=-     )
[benchamrk-run] > mean requests/sec                                1099.86 (OK=1099.86 KO=-     )
[benchamrk-run] ---- Response Time Distribution ------------------------------------------------
[benchamrk-run] > t < 800 ms                                        792999 (   100%)
[benchamrk-run] > 800 ms <= t < 1200 ms                                  1 (     0%)
[benchamrk-run] > t >= 1200 ms                                           0 (     0%)
[benchamrk-run] > failed                                                 0 (     0%)
[benchamrk-run] ================================================================================
```
### Main (commit 29bee21683ad987e2a3c313a5caebca7f0063a9e)
```
[benchamrk-run] ================================================================================
[benchamrk-run] ---- Global Information --------------------------------------------------------
[benchamrk-run] > request count                                     793000 (OK=793000 KO=0     )
[benchamrk-run] > min response time                                      2 (OK=2      KO=-     )
[benchamrk-run] > max response time                                   2823 (OK=2823   KO=-     )
[benchamrk-run] > mean response time                                    54 (OK=54     KO=-     )
[benchamrk-run] > std deviation                                        213 (OK=213    KO=-     )
[benchamrk-run] > response time 50th percentile                         10 (OK=10     KO=-     )
[benchamrk-run] > response time 75th percentile                         21 (OK=21     KO=-     )
[benchamrk-run] > response time 95th percentile                        153 (OK=153    KO=-     )
[benchamrk-run] > response time 99th percentile                       1246 (OK=1242   KO=-     )
[benchamrk-run] > mean requests/sec                                1099.86 (OK=1099.86 KO=-     )
[benchamrk-run] ---- Response Time Distribution ------------------------------------------------
[benchamrk-run] > t < 800 ms                                        779570 ( 98.31%)
[benchamrk-run] > 800 ms <= t < 1200 ms                               4380 (  0.55%)
[benchamrk-run] > t >= 1200 ms                                        9050 (  1.14%)
[benchamrk-run] > failed                                                 0 (     0%)
[benchamrk-run] ================================================================================
```

pruivo (dismissed):

mabartos:
@pruivo Great! I think we have enough data to proceed with the approach. Thanks!

@keycloak/cloud-native Could anyone look at this? Thanks!

brunobat:
Did you measured the startup time?

ahus1:
I would be +1 to use it, at the same time Quarkus is stating this is a [technology preview feature](https://quarkus.io/version/3.27/guides/rest#reflection-free-jackson-serialization-and-deserialization). 

So I'd say we would need to wrap it as an opt-in preview feature in KC as we don't want to support it downstream, and some tests enabled and disabled. We might open a ticket with the Quarkus team make it non-preview eventually. 

Reading the docs, the following part worries me a bit: 

> Developers can further customize JSON processing by implementing the ObjectMapperCustomizer interface. This interface allows fine-grained control over the ObjectMapper, enabling the registration of custom serializers and deserializers while ensuring compatibility with the reflection-free optimization. If additional configuration is needed, implement an ObjectMapperCustomizer bean and register any necessary modules or settings.

ahus1:
@brunobat / @franz1981 - thank for joining this discussion. Do you have additional information for us on the maturity of the feature and any plans for graduation of this feature?

brunobat:
I believe @franz1981 has details about why it's still in preview.

mabartos:
> Did you measured the startup time?

@brunobat Nope. Should we expect some bigger differences?

> I would be +1 to use it, at the same time Quarkus is stating this is a [technology preview feature](https://quarkus.io/version/3.27/guides/rest#reflection-free-jackson-serialization-and-deserialization).

@ahus1 Oh, totally missed this. Thanks for pointing this out! Yes, unfortunately, we'd need to have it disabled by default :(

brunobat:
> > Did you measured the startup time?
> 
> @brunobat Nope. Should we expect some bigger differences?


The expectation is that it will start faster because more work was done at build, however, I think we should measure the real world behavior differences in the C2 compiler under load, during warmup.

franz1981:
@brunobat IDK why it's in tech preview; I think we were waiting for user feedbacks
@gsmet @geoand ?

geoand:
Correct

gsmet:
Yes, we were waiting for user feedback. We had some and we fixed some of the issues along the way.

I haven't seen a new issue for a while now (but then it's hard to know how widely this feature is used). But we had feedback so it has been tested by some users.

The issues were mostly about corner cases not handled properly in serialization/deserialization, not about structural unstability.
So if your payloads are serialized/deserialized correctly, I think it could make sense using it but I think you should provide a way to disable it if users can extend Keycloak in a way that their own code goes through REST JSON serialization/deserialization.

Maybe you could provide the option with a default value to false for a while and then switch it to true at some point (and keeping the ability to disable it?).

gsmet:
> Maybe you could provide the option with a default value to false for a while and then switch it to true at some point (and keeping the ability to disable it?).

And if you do that, make sure you have a prominent note in the release notes to gather as much feedback as possible.

mabartos:
> Maybe you could provide the option with a default value to false for a while and then switch it to true at some point (and keeping the ability to disable it?).

Yes, that's the plan. 

Thank you all for your feedback!

geoand:
🙏🏽

mabartos:
@brunobat I don't see any performance improvement related to the startup. I tried only a simple check on my PC by checking the startup time printed by Quarkus.

| Run | Old  | New  |
|-----|-----------|-----------|
| 1   | 4543ms      | 4545ms      | 
| 2   | 4522ms      | 4535ms      |
| 3   | 4492ms     | 4533ms      |
| 4   | 4548ms      | 4581ms      | 
| 5   | 4370ms      | 4507ms      |
| 6   | 4556ms     | 4601ms      |
| 7   | 4532ms      | 4458ms      |
| **AVG**   | **4509ms**      | **4537ms**      |

Diff: +28ms

I haven't checked the execution time of particular build steps, but 28ms is negligible, and when run on my PC, these results may be influenced by the env.

franz1981:
@mabartos Actually there's a benefit but is more at runtime; before https://openjdk.org/jeps/416 reflection was very sensitive to so-called inflation, see https://blogs.oracle.com/javamagazine/post/java-reflection-performance 

which means that if if the JVM has very metadata space or start to collect memory aggressively the reflective "inflated" (i.e. optimized) field accesses in Jackson will fly away, replaced by slow path native accessors; using proper generated encoders/decoders just make this problem to disappear - increasing stability of steady state performance.

geoand:
@mabartos the reflection free serializers have no effect on startup time

brunobat:
> @brunobat I don't see any performance improvement related to the startup. I tried only a simple check on my PC by checking the startup time printed by Quarkus.
> 
> Run	Old	New
> 1	4543ms	4545ms
> 2	4522ms	4535ms
> 3	4492ms	4533ms
> 4	4548ms	4581ms
> 5	4370ms	4507ms
> 6	4556ms	4601ms
> 7	4532ms	4458ms
> **AVG**	**4509ms**	**4537ms**
> Diff: +28ms
> 
> I haven't checked the execution time of particular build steps, but 28ms is negligible, and when run on my PC, these results may be influenced by the env.

Thanks! That looks like nothing.

mabartos:
Should be ready for review. Thanks!

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
```diff
@@ -149,4 +149,11 @@ public enum ClientAuth {
                     "Specify a list of comma-separated values defined in milliseconds. Example with buckets from 5ms to 10s: 5,10,25,50,250,500,1000,2500,5000,10000")
             .build();
 
+    public static final Option<Boolean> HTTP_OPTIMIZED_SERIALIZERS = new OptionBuilder<>("http-optimized-serializers", Boolean.class)
```
ahus1: I see that this is using a CLI option instead of using Keycloak's established feature mechanism in https://github.com/keycloak/keycloak/blob/ba6d95c7b16f96f343e0cc47dd8f0eb367e3829e/common/src/main/java/org/keycloak/common/Profile.java. 

Without that mechanism, we don't have it listed in the preview features. With the feature, we might not need the CLI option at all.

When we want to make it preview, we should run (a part) of the test suite with this option enabled, otherwise we will not know if we broke it.

ahus1 (changes requested):
Thank you for the PR. See below for some comments on how we handle preview features.

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
ahus1: Given @gsmet's comment: 

>  you should provide a way to disable it if users can extend Keycloak in a way that their own code goes through REST JSON serialization/deserialization.

Having the option might still be a good thing?!

ahus1:
@gsmet - you wrote: 

> you should provide a way to disable it if users can extend Keycloak in a way that their own code goes through REST JSON serialization/deserialization.

Users can bring their own extensions that might include REST endpoints. Would they be affected by this, and if yes how would they be affected?

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
mabartos: > Without that mechanism, we don't have it listed in the preview features. With the feature, we might not need the CLI option at all.

@ahus1 Yes, considering features are more visible to users, we can do it as even the Quarkus option is build-time option (if it's runtime option, we'd lose the ability to specify it at runtime).

My only concern about these features is the count of them - we add features to opt-in specific functionality and we can grow to the state when there are so many features that people get lost in them. On the other hand, we can deprecate and then remove them, so it should not a big deal. For this exact refleciton-free serializers, I'd expect that once it's supported and verified, no one should have a problem with the reflection-free serializers, so not probably a big need to turn it off. Right now, we can go on with the feature.

> When we want to make it preview, we should run (a part) of the test suite with this option enabled, otherwise we will not know if we broke it.

Ok, we can do that.  

> Having the option might still be a good thing?!

If we consider it as a feature, we can disable the feature even once it's enabled by default.

mabartos:
### Summary

In order to enable this functionality, the only thing you need to do is to enable the feature `http-optimized-serializers` like this:

```shell
./kc.sh start-dev --features=http-optimized-serializers
```

No need to have other options. 

@keycloak/cloud-native @shawkins Could you please check it? Thanks!

shawkins:
> ./kc.sh start-dev --features=http-optimized-serializers

Do you see this evolving to additional options and / or different handling with v1, v2, etc.?

If not, I'm not sure we want to go down the path of having these 1-1 mappings between features and options. As it complicates the notion of an option without much benefit.

Alternatively could this feature could be too fine-grained? That is instead should we have --features=http-optimizations, or even --features=optimizations, to enable a set of optimimzation related options.

Also having something this fine-grained is yet another reason why we should move away from using a list option for features. It makes the management of individual features very difficult in layered scenarios.

vmuzikar:
> If not, I'm not sure we want to go down the path of having these 1-1 mappings between features and options.

I believe this is not what's happening here. Ultimately, reflection-free serialization doesn't need to be an option – it's an internal perf enhancement which will be eventually enabled for all users. However, since it's a preview (in Quarkus too), we'd want to clearly mark it that way in Keycloak and have it disabled by default. For that, the Features concept fits better than an option. Features have clear visibility in docs and the UI, they have lifecycle etc. I understand the UX around that is quite confusing and we will hopefully make it better. Still, I'd vote for having the reflection-free serialization as a Feature.

pruivo (dismissed):

mabartos:
> Do you see this evolving to additional options and / or different handling with v1, v2, etc.? If not, I'm not sure we want to go down the path of having these 1-1 mappings between features and options. As it complicates the notion of an option without much benefit.

@shawkins I don't expect to have more options or handling of other versions. As was pointed out in the mail thread about features, this might not be a problem to have this like feature. Right now, we're mapping only feature/property-mapper that enables certain functionality that is hidden to user from the configuration perspective (done only via feature). 

> Alternatively could this feature could be too fine-grained? That is instead should we have --features=http-optimizations, or even --features=optimizations, to enable a set of optimimzation related options.

It'd be nice to have something general, but as we don't have in mind what else might be here, we'd have just an "empty" feature, once this behavior is enabled by default and not part of the general feature. If it's still part of the feature switch, we wouldn't be able to promote the specific behavior to supported as the other optimizations would need to be still in preview.

mabartos:
@pruivo Thanks for the review!

@shawkins @vmuzikar @ahus1 Could you please re-review? :) Thanks!

shawkins (dismissed):
LGTM, thanks @mabartos

docs/guides/server/configuration-production.adoc
```diff
@@ -87,4 +87,27 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
 
 See <@links.server id="caching" anchor="network-bind-address"/> for more details.
 
+== Better HTTP performance
+NOTE: This section is describing a feature which is currently in *preview* as we gather more feedback about potential issues. We appreciate any feedback.
```
ahus1: The usual template we use in the old docs is here: https://github.com/keycloak/keycloak/blob/4b697009d3336dbb702eecc83083020cc16b5de3/docs/documentation/topics/templates/techpreview.adoc?plain=1#L1-L25

It has been partially copied to `token-exchange.adoc`, but no-one yet moved it to a freemarker snippet yet.  Please use it here and also update the token exchange part.

docs/guides/server/configuration-production.adoc
```diff
@@ -87,4 +87,27 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
...
+This feature introduces Jackson JSON reflection-free serializers to optimize HTTP processing.
+By removing the overhead of reflection during serialization/deserialization, it delivers measurable benefits:
+
+- ~5% increase in throughput
+- More stable response times
+- Reduced system resource usage
+
+These improvements help ensure smoother, more predictable performance at scale while also lowering the operational cost of running production systems.
+
+The only known tradeoff is that build(augmentation) time increased by ~6% as certain actions were moved to the build time instead of runtime.
```
ahus1: The docs should have one word here, either "build" or "augmentation". I think the docs are a bit inconsistent. 

```suggestion
The only known tradeoff is that build time increases by ~6% as certain actions were moved to the build time instead of runtime.
```

docs/guides/server/configuration-production.adoc
```diff
@@ -87,4 +87,27 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
...
+
+In production environments, the performance of the HTTP layer is critical.
+Every request passes through it, making it a key factor in overall system responsiveness, scalability, and user experience.
+
+This feature introduces Jackson JSON reflection-free serializers to optimize HTTP processing.
+By removing the overhead of reflection during serialization/deserialization, it delivers measurable benefits:
+
+- ~5% increase in throughput
+- More stable response times
+- Reduced system resource usage
```
ahus1, line 102: We're writing the documentation for those people who administrate Keycloak. They are no Java experts, and don't care about the terms "reflection", and "Jackson". 

Please consider rewriting for the administrator persona.

docs/guides/server/configuration-production.adoc
```diff
@@ -87,4 +87,27 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
...
+
+These improvements help ensure smoother, more predictable performance at scale while also lowering the operational cost of running production systems.
+
+The only known tradeoff is that build(augmentation) time increased by ~6% as certain actions were moved to the build time instead of runtime.
+
+You can enable this feature as follows:
+
+<@kc.start parameters="--features=http-optimized-serializers"/>
+
+The feature is *disabled by default*, as we gather more feedback.
```
ahus1: This sentence "The feature is *disabled by default*" might no longer be necessary if we have the note at the top. 

If you want to get more feedback, consider creating a discussion, and linking it here. Then mark the whole text block that it only appears in upstream, as we IMHO don't want this to show in downstream. 

To get more feedback, you might move that new paragraph up a bit so people actually read it. They usually stop reading once they found what they needed.

ahus1 (changes requested):
Thank you for the PR. See below some documentation related comments.

docs/guides/server/configuration-production.adoc
mabartos: Done.

docs/guides/server/configuration-production.adoc
mabartos: Done.

docs/guides/server/configuration-production.adoc
mabartos: Done.

mabartos:
@ahus1 Thanks for the feedback. Should be done now.

image

### Token exchange docs
#### Old
image


#### New
image

ahus1 (dismissed):
Thank you for the changes, I think it looks great. I am not merging it yet as there is another review from @vmuzikar pending.

docs/guides/server/configuration-production.adoc
```diff
@@ -87,4 +88,25 @@ export JAVA_OPTS_APPEND="-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6A
 
 See <@links.server id="caching" anchor="network-bind-address"/> for more details.
 
+== Better HTTP performance
```
vmuzikar: I'd word it slightly differently. I assume this whole paragraph will go away once we flip the switch to supported and enabled by default. So I'd put more emphasis on "hey, we have a preview perf enhancements here".
```suggestion
== Preview of enhanced HTTP performance
```

docs/guides/templates/features.adoc
```diff
@@ -8,3 +8,33 @@
 </#list>
 |===
 </#macro>
+
+<#macro techpreview feature additionalCommunityText="">
+<#assign profileFeature = ctx.features.getFeature(feature)>
+
+<#if !profileFeature.enabled>
```
vmuzikar: I do not think we need this if and also the changes in `Profile.java` related to it. If I understand correctly, all preview features are always disabled by default.

vmuzikar (commented):
@mabartos Thank you for the PR. Nice work with the analysis!

Added a couple of comments, but none of them are blocking.

docs/guides/templates/features.adoc
mabartos: Done.

mabartos:
Ok, added some additional minor docs changes. 

@ahus1 @vmuzikar Would be good if you could take another look. Thanks!

vmuzikar (dismissed):
LGTM, thanks.

vmuzikar:
@mabartos Some tests failed but seems unrelated?

mabartos:
@vmuzikar Yes, those tests should be unrelated. Retried

mabartos:
Resolved conflict

vmuzikar (approved):

ahus1 (approved):
@mabartos - I think I approved it before, please proceed. You only need to add me when there are significant new changes.

mabartos:
Progress: We run the whole old testsuite (thousands of tests) with this settings for ~3 weeks and AFAIK, we have not encounter any issues - I think it's a nice indicator it works properly.

cc: @brunobat @franz1981 @gsmet @geoand

geoand:
That's great to hear!

PR #42948, opened 2025-09-25
Adding an integration test with Minikube for Kubernetes Service Account Federated Authenticator 

slaskawi:
This Pull Request introduces an e2e test for Kubernetes Service Account Federated Authenticator

Fixes https://github.com/keycloak/keycloak/issues/42983


docs/documentation/server_admin/topics/identity-broker/kubernetes.adoc
```diff
@@ -0,0 +1,27 @@
...
++
+.Kubernetes settings
+|===
+|Configuration|Description
+
+|Alias
+|The alias for the identity provider is used to link a client to the provider
+
+|Kubernetes JWKS URL
+|The Kubernetes JWKS URL when accessing an external Kubernetes cluster. The JWKS endpoint must not require authentication
```
ryanemerson: We should state somewhere in these docs, maybe the introduction, that if the JWKS URL is not explicitly defined, then we default to `https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT_HTTPS}/openid/v1/jwks` and authentication is handled automatically using `/var/run/secrets/kubernetes.io/serviceaccount/token`.

We can then make it clear that this setting is optional.

stianst:
We can't add tests as a follow-up, since it's a requirement for preview features to be fully tested.

docs/documentation/server_admin/topics/identity-broker/kubernetes.adoc
slaskawi: Addressed.

slaskawi:
> We can't add tests as a follow-up, since it's a requirement for preview features to be fully tested.

I'm looking into tests right now. I hope to get them ready either today or early next week

slaskawi:
Together with @stianst we agreed to hold it off for 26.4. There's no good reason to rush this feature towards the release.

Right after 26.4 we'll add missing docs, an integration and an e2e test and promote it to Preview.

ahus1:
I've started to fix the failing test and update the docs, let's see...

slaskawi:
> I've started to fix the failing test and update the docs, let's see...

Thanks a million @ahus1 ! I already allocated a bit of time next week to get this one done. We already have an integration test (implemented by Stian), so after making this one green, we should be good to do.

ahus1:
@slaskawi - sure, no worries. I found that the test is a bit unstable and complex due to the realm import, and while it ran locally fine, it was failing on GitHub. 

So pushed some more changes for separate test in KeycloakKubernetesJwtTest which creates the necessary resources via APIs instead of via an import. I hope this will be simpler to maintain in the long run, and it also found it runs a bit faster as it doesn't need to start the extra Pod for the import.

stianst:
Can we split out the tests from this PR into a separate PR? I'd like to get that reviewed/merged asap, and we may have some other follow-ups to do before marking the feature as supported.

slaskawi:
@stianst @ahus1 @ryanemerson Per Stian's request, I split the docs and code for the upgrade into https://github.com/keycloak/keycloak/pull/44093

Once we merge it, I'll create a separate PR with e2e tests (as they rely on the fact that Kubernetes IdP is in Preview and enabled by default in the Operator).

stianst:
> @stianst @ahus1 @ryanemerson Per Stian's request, I split the docs and code for the upgrade into #44093
> 
> Once we merge it, I'll create a separate PR with e2e tests (as they rely on the fact that Kubernetes IdP is in Preview and enabled by default in the Operator).

Tests should be merged first really; preview features are not enabled by default either so why does it matter?

ahus1:
@slaskawi / @stianst - updated the PR to include docs and the integration tests so it can be merged first as suggested by Stian. 

Let me know if you agree. For now, waiting for the build to be green again.

ahus1 (approved):
Approving based on the various reviews from @ryanemerson, @slaskawi, @stianst. Thanks everyone!

PR #42953, opened 2025-09-25
Ensure events are fully filled before success is called

### LINKED ISSUE #42914: Make sure TestEventsListenerProvider does not reuse the events in memory
### Before reporting an issue

### Area

testsuite

### Describe the bug

We detected in this PR https://github.com/keycloak/keycloak/pull/42866 that the providers we are using for testing events are just adding the event passed to a list. This way it can happen that the event is passed not fully filled but later is completed with the missing info. As the list is created with the same event instance, the event is returned OK for the test although it was incorrect at the time the method was called. This means that another provider (for example one that sends the event serialized via network) would not be executed OK and failed because some info was not set.

This issue is to change the test providers to `clone` the event with the current information at calling time. We have to check if there are errors in the CI after the change and fix the event generation for them.

### Version

999.0.0-SNAPSHOT

### Regression

### Expected behavior

The event should be stored in the test provider with the data sent at calling time. The object sshould be duplicated/cloned to avoid this issue.

### Actual behavior

The event is just added to the list, so the instance is the same and can trigger this nasty issue.

### How to Reproduce?

See description and issue https://github.com/keycloak/keycloak/issues/42867.

### Anything else?

_No response_

rmartinc:
Closes #42914

@mposolda I have done the exercise of cloning the events in the test provider in the `onEvent` method. There were not too many failures and mainly because of two code paths. It seems that some fields of the event were filled later when creating the response.

Regarding the changes in the tests, the main difference is that now sometimes the order is a bit different (because the `success` is executed later). It is more clear in the error messages, sometimes there are two errors and now moving the call later the order is the opposite. As you see it's just in three tests so I think it's OK. I checked that the same errors are reported but in different order (for example the `MultipleTabsLoginTest`, previously it was cookie_not_found and then code_invalid and now it is code_invalid and then cookie_not_found).

Check if it's OK or you want me to investigate more about this.

### DIFF
--- a/services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
+++ b/services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
@@ -1069,8 +1069,9 @@ public class AuthenticationManager {
 
         event.event(EventType.LOGIN);
         event.session(userSession);
+        Response response = redirectAfterSuccessfulFlow(session, realm, userSession, clientSessionCtx, request, uriInfo, clientConnection, event, authSession);
         event.success();
-        return redirectAfterSuccessfulFlow(session, realm, userSession, clientSessionCtx, request, uriInfo, clientConnection, event, authSession);
+        return response;
     }
 
     // Return null if action is not required. Or the alias of the requiredAction in case it is required.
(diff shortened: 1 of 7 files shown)


mposolda (approved):
@rmartinc I think it is ok this way. Thanks for the updates!

PR #42959, opened 2025-09-25
fix: fully sanitizing mappers and changing duplicated log to trace

### LINKED ISSUE #38438: Avoid 'duplicated mappers' message

shawkins:
Scaled back replacement for #41886 - it turns out that I need these changes first to further refine picocli performance.

closes: #38438

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/ExportPropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/ExportPropertyMappers.java
@@ -52,10 +52,12 @@ public final class ExportPropertyMappers implements PropertyMapperGrouping {
                 fromOption(ExportOptions.FILE)
                         .to("kc.spi-export--single-file--file")
                         .paramLabel("file")
+                        .isEnabled(c -> c instanceof Export)
                         .build(),
                 fromOption(ExportOptions.DIR)
                         .to("kc.spi-export--dir--dir")
                         .paramLabel("dir")
+                        .isEnabled(c -> c instanceof Export)
                         .build(),
                 fromOption(ExportOptions.REALM)
                         .to("kc.spi-export--single-file--realm-name")
@@ -109,7 +111,7 @@ public final class ExportPropertyMappers implements PropertyMapperGrouping {
     }
 
     private static boolean isDirProvider() {
-        return isProvider(DIR);
+        return !isSingleFileProvider();
     }
 
     private static boolean isProvider(String provider) {
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
@@ -14,7 +14,6 @@ import org.keycloak.quarkus.runtime.Environment;
 import org.keycloak.quarkus.runtime.cli.Picocli;
 import org.keycloak.quarkus.runtime.cli.PropertyException;
 import org.keycloak.quarkus.runtime.cli.command.AbstractCommand;
-import org.keycloak.quarkus.runtime.cli.command.ShowConfig;
 import org.keycloak.quarkus.runtime.configuration.DisabledMappersInterceptor;
 import org.keycloak.quarkus.runtime.configuration.MicroProfileConfigProvider;
 import org.keycloak.quarkus.runtime.configuration.NestedPropertyMappingInterceptor;
@@ -24,7 +23,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumMap;
-import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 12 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/ExportPropertyMappers.java
```diff
@@ -52,10 +52,12 @@ public List<PropertyMapper<?>> getPropertyMappers() {
                 fromOption(ExportOptions.FILE)
                         .to("kc.spi-export--single-file--file")
                         .paramLabel("file")
+                        .isEnabled(c -> c instanceof Export)
```
mabartos, line 55: So, when there's some duplication, we'd need to explicitly enable/disable the mapper, right? 

Why we should remove the check of the `command.getOptionCategories()`? It was quite nice approach(idea) on how to eliminate the mapper from "denied" category - so, no explicit check of the command. 

Would it be possible to have it checked based on the used optionCategories? And if there are some duplicated mappers in 2 or more categories, require to add some explicit check. 

I could imagine even have the default value for the `isEnabled` the check of the comman categories. So, every mapper with the associated option in the denied category would be disabled. 

@shawkins WDYT?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
```diff
@@ -152,26 +150,22 @@ public static String maskValue(String value, String configSourceName, PropertyMa
...
         final var mappers = new ArrayList<>(MAPPERS.getOrDefault(property, Collections.emptyList()));
         if (category != null) {
             mappers.removeIf(m -> !m.getCategory().equals(category));
         }
 
         return switch (mappers.size()) {
-            case 0 -> defaultMapper;
+            case 0 -> null;
             case 1 -> mappers.get(0);
             default -> {
```
mabartos, line 162: So, is it now even possible that we find 1+ mapper here? 

Is it related to the help-all?

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/StartCommandDistTest.java
```diff
@@ -106,13 +106,6 @@ void testHttpEnabled(CLIResult cliResult) {
         cliResult.assertStarted();
     }
 
-    @DryRun
-    @Test
-    @Launch({ "-v", "start", "--db=dev-mem", OPTIMIZED_BUILD_OPTION_LONG})
-    void failBuildPropertyNotAvailable(CLIResult cliResult) {
-        cliResult.assertError("Build time option: '--db' not usable with pre-built image and --optimized");
```
mabartos, line 113: It still should be valid, right?

mabartos (commented):
@shawkins Thanks for the PR! Just a few comments.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/StartCommandDistTest.java
shawkins: It is redundant with other tests.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java
shawkins: Before sanitizing it is still possible to have more than 1. However there was no usage of the default concept, so that was removed.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/ExportPropertyMappers.java
shawkins: > So, when there's some duplication, we'd need to explicitly enable/disable the mapper, right?

Yes. The import / export property mappers are really a pretty special case. They overlap and are valid only for their respective commands. Honestly I'm not sure we even want someone putting file=... into the keycloak.conf as it's not clear that should represent both the import and export location. We've also talked about having the keycloak.conf more easily scoped per command, but that's an even bigger change.

> Why we should remove the check of the command.getOptionCategories()? It was quite nice approach(idea) on how to eliminate the mapper from "denied" category - so, no explicit check of the command.

I would like to further refine the option categories handling in other prs. The problem is that it's not a great abstraction - it is ultimiately about hidding options and it's behavior can be different between build and runtime options.

mabartos: > Yes. The import / export property mappers are really a pretty special case. They overlap and are valid only for their respective commands. Honestly I'm not sure we even want someone putting file=... into the keycloak.conf as it's not clear that should represent both the import and export location. We've also talked about having the keycloak.conf more easily scoped per command, but that's an even bigger change.

@shawkins In the context of CLI and putting the command with the `--file`, the scenario is nice as you don't have to specify sth like `--import-file`, or `--export-file`. However, yes, totally unusable in the keycloak.conf as having the `file=` there. 

> I would like to further refine the option categories handling in other prs. The problem is that it's not a great abstraction - it is ultimiately about hidding options and it's behavior can be different between build and runtime options.

Ok, works to me.

mabartos (commented):
@shawkins Sorry for the late response. 

LGTM in general.

It'd be good if you could rebase this to comply with the spotless checks.

mabartos (approved):

PR #42966, opened 2025-09-25
Only filter default organization related scopes based on dynamic scope format

### LINKED ISSUE #42877: Valid scope parameter in access token request is rejected with invalid_scope error
### Before reporting an issue

### Area

oidc

### Describe the bug

A valid OIDC/OAuth2 token request is rejected with error code `invalid_scope` in case the token request contains a valid scope parameter which includes at least 2 scopes which use dynamic scope format.

It seems this regression was introduced with the commit https://github.com/keycloak/keycloak/commit/8f9d02c3057205129e08872943505829286500d7 related to pull request https://github.com/keycloak/keycloak/pull/39752 and issue https://github.com/keycloak/keycloak/issues/39402.

The issue is also present if the experimental feature `DYNAMIC_SCOPES` is disabled. The code introduced in the PR above is not guarded by a feature enabled check.

### Version

26.3.4

### Regression

### Expected behavior

A valid OIDC/OAuth2 token request should not be rejected with error.

### Actual behavior

Keycloak responds with HTTP 400 and error code `invalid_scope`.

### How to Reproduce?

1. Keep experimental Keycloak feature `DYNAMIC_SCOPES` disabled.
2. Create two scopes where the second scope has the name of the first scope as prefix and having a `:` separator and any suffix. For example `scope1` and `scope1:A`
3. Create a client (for example `client_1` with client_credentials flow enabled) and configure both previously created scopes as **default** scopes.
4. Make a simple token request which explicitly includes both scopes in the token request parameter. For example with a simple curl HTTP request.

`curl -X POST --location "https://<keycloak_host>/realms/<realm>/protocol/openid-connect/token" -H "Content-Type: application/x-www-form-urlencoded" -d 'grant_type=client_credentials&client_id=test_client&client_secret=<client_secret>&scope=scope1 scope1:A'`

The response from Keycloak is
`{"error":"invalid_scope","error_description":"Invalid scopes: scope1 scope1:A"}`

If you only request a single scope or do not include any scope parameter at all everything is working as expected because the problematic code only kicks in if there are multiple scopes with a specific format. See https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java#L643 as reference.

pedroigor:
Closes #42877

* Only filter default scopes with the organization mapper when checking requested scopes using a dynamic format
* Added a new method to resolve mappers by type from `ClientScopeModel` to avoid iterating over mappers all the time. I'm not sure if we need to add a `protocol` parameter to the `getProtocolMappersByType` or if just the type is enough?
* Updated methods `getProtocolMapperById|Name` to avoid iterating over all mappers all the time

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/AbstractCachedClientScope.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/AbstractCachedClientScope.java
@@ -1,0 +1,59 @@
+package org.keycloak.models.cache.infinispan.entities;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.keycloak.models.ClientScopeModel;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.ProtocolMapperModel;
+import org.keycloak.models.cache.infinispan.DefaultLazyLoader;
+import org.keycloak.models.cache.infinispan.LazyLoader;
+
+abstract class AbstractCachedClientScope<D extends ClientScopeModel> extends AbstractRevisioned implements InRealm {
+
+    private final LazyLoader<D, Map<String, ProtocolMapperModel>> mappersById;
+    private final LazyLoader<D, Map<String, String>> mappersByName;
+    private final LazyLoader<D, Map<String, List<String>>> mappersByType;
+
+    public AbstractCachedClientScope(Long revision, ClientScopeModel model) {
+        super(revision, model.getId());
+        mappersById = new DefaultLazyLoader<>(scope -> scope.getProtocolMappersStream()
+                .collect(Collectors.toMap(ProtocolMapperModel::getId, ProtocolMapperModel::new)),
+                Collections::emptyMap);
+        mappersByName = new DefaultLazyLoader<>(scope -> scope.getProtocolMappersStream()
+                .collect(Collectors.toMap(mapper -> mapper.getProtocol() + "." + mapper.getName(),
+                        ProtocolMapperModel::getId)),
+                Collections::emptyMap);
+        mappersByType = new DefaultLazyLoader<>(scope ->
+                scope.getProtocolMappersStream()
+                        .collect(Collectors.groupingBy(ProtocolMapperModel::getProtocolMapper,
+                                Collectors.mapping(ProtocolMapperModel::getId, Collectors.toList()))),
+                Collections::emptyMap);
+    }
+
+    public Stream<ProtocolMapperModel> getProtocolMappers(KeycloakSession session, Supplier<D> model) {
+        return mappersById.get(session, model).values().stream();
+    }
+
+    public ProtocolMapperModel getProtocolMapperById(KeycloakSession session, Supplier<D> model, String id) {
+        if (id == null) {
+            return null;
+        }
+        return mappersById.get(session, model).get(id);
+    }
+
+    public List<ProtocolMapperModel> getProtocolMapperByType(KeycloakSession session, Supplier<D> model, String type) {
+        return mappersByType.get(session, model).getOrDefault(type, List.of()).stream()
+                .map(id -> getProtocolMapperById(session, model, id))
+                .collect(Collectors.toList());
+    }
+
+    public ProtocolMapperModel getProtocolMapperByName(KeycloakSession session, Supplier<D> model, String protocol, String name) {
+        String id = mappersByName.get(session, model).get(protocol + "." + name);
+        return getProtocolMapperById(session, model, id);
+    }
+}
--- a/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
@@ -62,6 +62,7 @@ import org.keycloak.models.light.LightweightUserAdapter;
 import org.keycloak.models.utils.KeycloakModelUtils;
 import org.keycloak.models.utils.SessionExpirationUtils;
 import org.keycloak.models.utils.RoleUtils;
+import org.keycloak.organization.protocol.mappers.oidc.OrganizationMembershipMapper;
 import org.keycloak.organization.protocol.mappers.oidc.OrganizationScope;
 import org.keycloak.protocol.ProtocolMapper;
 import org.keycloak.protocol.ProtocolMapperUtils;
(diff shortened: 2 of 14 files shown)


graziang (approved):
@pedroigor thanks!

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/AbstractCachedClientScope.java
```diff
@@ -0,0 +1,70 @@
...
+            });
+            return map;
+        }, Collections::emptyMap);
+        mappersByType = new DefaultLazyLoader<>(scope -> {
+            Map<String, List<String>> map = new HashMap<>();
+            scope.getProtocolMappersStream().forEach(mapper -> {
+                map.computeIfAbsent(mapper.getProtocolMapper(), s -> new java.util.ArrayList<>()).add(mapper.getId());
+            });
+            return map;
+        }, Collections::emptyMap);
```
pruivo: I'll let you decide if it makes things easier or harder to read. 
```suggestion
        mappersById = new DefaultLazyLoader<>(scope ->
                scope.getProtocolMappersStream()
                        .collect(Collectors.toMap(ProtocolMapperModel::getId, ProtocolMapperModel::new)), Collections::emptyMap);
        mappersByName = new DefaultLazyLoader<>(scope ->
                scope.getProtocolMappersStream()
                        .collect(Collectors.toMap(m -> m.getProtocol() + "." + m.getName(), ProtocolMapperModel::getId)), Collections::emptyMap);
        mappersByType = new DefaultLazyLoader<>(scope ->
                scope.getProtocolMappersStream()
                        .collect(Collectors.groupingBy(ProtocolMapperModel::getProtocolMapper, Collectors.mapping(ProtocolMapperModel::getId, Collectors.toList()))), Collections::emptyMap);
```

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/AbstractCachedClientScope.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public Set<ProtocolMapperModel> getProtocolMappers(KeycloakSession session, Supplier<D> model) {
+        return new HashSet<>(mappersById.get(session, model).values());
+    }
```
pruivo: You can stream them instead of copying them. 
```suggestion
    public Stream<ProtocolMapperModel> getProtocolMappers(KeycloakSession session, Supplier<D> model) {
        return mappersById.get(session, model).values().stream();
    }
```

pruivo (commented):
some suggestions

pedroigor:
@pruivo, applied your changes. Thanks.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/OIDCProtocolMappersTest.java
```diff
@@ -1862,14 +1853,85 @@ public void testStaticScopeUsingDynamicScopeFormatWithDedicatedMappers() {
...
+        // creates a client scope using the dynamic scope format and add it to the client as optional scope
+        createClientScope(realm, clientResource, "test", "from-scope-mapper", "value", true);
+        createClientScope(realm, clientResource, "test:create", "from-dynamic-scope", "value", true);
+
+        oauth.scope("openid test:create");
+        AccessTokenResponse response = oauth.doPasswordGrantRequest("test-user@localhost", "password");
+        assertTrue(response.getScope().contains("test:create"));
+
+        AccessToken accessToken = oauth.verifyToken(response.getAccessToken());
+        assertTrue(response.getScope().contains("test:create"));
```
pruivo: This assertion appears to be duplicated from line 1882. The method below has the same duplication.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/OIDCProtocolMappersTest.java
```diff
@@ -1862,14 +1853,85 @@ public void testStaticScopeUsingDynamicScopeFormatWithDedicatedMappers() {
...
+        createClientScope(realm, clientResource, "test:create", "from-dynamic-scope", "value", false);
+
+        oauth.scope("openid test:create");
+        AccessTokenResponse response = oauth.doPasswordGrantRequest("test-user@localhost", "password");
+        assertTrue(response.getScope().contains("test:create"));
+
+        AccessToken accessToken = oauth.verifyToken(response.getAccessToken());
+        assertTrue(response.getScope().contains("test:create"));
+        assertNotNull(accessToken.getOtherClaims());
+        assertNotNull(accessToken.getOtherClaims().get("from-dynamic-scope"));
```
pruivo, line 1914: shouldn't check if `from-scope-mapper` is (not) present too?

services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
```diff
@@ -638,10 +639,17 @@ public static Stream<ClientScopeModel> getRequestedClientScopes(KeycloakSession
...
-        clientScopes = clientScopes.filter(scope -> scope.equals(client) || !scopeParam.contains(scope.getName() + ClientScopeModel.VALUE_SEPARATOR));
+        clientScopes = clientScopes.filter(scope -> {
+            List<ProtocolMapperModel> orgMappers = scope.getProtocolMapperByType(OrganizationMembershipMapper.PROVIDER_ID);
+
+            if (!orgMappers.isEmpty()) {
+                return !scopeParam.contains(scope.getName() + ClientScopeModel.VALUE_SEPARATOR);
+            }
+
+            return true;
+        });
```
pruivo, line 648: I moved the `scope.getProtocolMapperByType()` invocation to avoid creating a list and a possible lookup, as `!scopeParam.contains()` should be cheaper to evaluate.
I also noticed `scope.equals(client)` has been removed in this PR. Just want to double-check if it's intentional. 

```suggestion
        clientScopes = clientScopes.filter(scope ->
                !scopeParam.contains(scope.getName() + ClientScopeModel.VALUE_SEPARATOR) ||
                        scope.getProtocolMapperByType(OrganizationMembershipMapper.PROVIDER_ID).isEmpty()
        );
```

pedroigor: The fix is to only check the requested params if there are mappers with that type. 

EDIT: Sorry, it should work too.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/OIDCProtocolMappersTest.java
pedroigor: It should check the scope in the access token, changed.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/OIDCProtocolMappersTest.java
pedroigor: There is another test checking this. See `testStaticScopeUsingDynamicScopeFormatPrefixedWithScopeAsDefaultScope`.

graziang (approved):

pruivo (approved):

abstractj (approved):

PR #42967, opened 2025-09-25
add labels to realm import jobs

### LINKED ISSUE #40799: Provide a way to add custom labels to Realm Import job of Keycloak operator
### Description

There should be a way to add custom labels to the realm import job created by keycloak operator. Right now, only 2-3 labels are added to the job which are predefined in the code.

https://github.com/keycloak/keycloak/blob/1c05c8df630bc025d328f4b986798575d3cce7a5/operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java#L119

https://github.com/keycloak/keycloak/blob/2b44c5676f7ebe340da55a94766a8a11f3cf0eb1/operator/src/main/java/org/keycloak/operator/Utils.java#L72

### Discussion

https://github.com/keycloak/keycloak/discussions/40730

### Motivation

Istio has a bug because of which jobs stay in NotReady state (https://github.com/istio/istio/issues/6324)

we have a internal webhook to call `quitquitquit` endpoint in Jobs having a predefined label (quitquitquit endpoint is used to stop the istio sidecar container). 

### Details

_No response_

KyriosGN0:
closes #40799

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakRealmImportJobDependentResource.java
@@ -40,8 +40,10 @@ import org.keycloak.operator.crds.v2alpha1.deployment.Keycloak;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.ImportSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.SchedulingSpec;
 import org.keycloak.operator.crds.v2alpha1.realmimport.KeycloakRealmImport;
+import org.keycloak.operator.crds.v2alpha1.realmimport.KeycloakRealmImportSpec;
 import org.keycloak.operator.crds.v2alpha1.realmimport.Placeholder;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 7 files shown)


shawkins:
Thank you for the changes @KyriosGN0 

A couple of thoughts:
- it would be good to add our default labels last to prevent overwriting built-in ones.
- do you anticipate using different labels for every import job? If not, there are also the import (and update) specs on the KeycloakSpec that could be used to define labels for their respective jobs.

KyriosGN0:
> Thank you for the changes @KyriosGN0
> 
> A couple of thoughts:
> 
> * it would be good to add our default labels last to prevent overwriting built-in ones.
> * do you anticipate using different labels for every import job? If not, there are also the import (and update) specs on the KeycloakSpec that could be used to define labels for their respective jobs.

hey @shawkins 
1. will do
2. its possible yes

KyriosGN0:
hey @vmuzikar, could i get a review here?

shawkins (approved):
LGTM, thank you @KyriosGN0

vmuzikar (approved):
Thank you for the contribution, LGTM.

devshrm:
why these changes are not part of latest release ?

ahus1:
@devshrm - if you look at the parent issue, you'll see that the issue is tagged `26.5.0`. That feature release is due for January 2026. In the meantime, there will be patch releases for 26.4, but those only include bug fixes, not enhancements.

PR #42976, opened 2025-09-26
Upgrade bc-fips testing and documentation to 2.1.2

rmartinc:
Closes #42958

Just upgrading the BC jars to the last version for testing and documentation. As version 2.1 supports EdDSA the PR adds `ClientAuthEdDSASignedJWTTest` to the CI and adds this key in the `JavaKeystoreKeyProviderTest`. If you think that is better to continue using 2.0 versions just let me know.

### DIFF
--- a/docs/guides/server/fips.adoc
+++ b/docs/guides/server/fips.adoc
@@ -129,7 +129,7 @@ By using TRACE level, you can check that the startup log contains `KC` provider 
 
 [source]
 ----
-KC(BCFIPS version 2.0 Approved Mode, FIPS-JVM: enabled) version 1.0 - class org.keycloak.crypto.fips.KeycloakFipsSecurityProvider,
+KC(BCFIPS version 2.0102 Approved Mode, FIPS-JVM: enabled) version 1.0 - class org.keycloak.crypto.fips.KeycloakFipsSecurityProvider,
 ----
 
 === Cryptography restrictions in strict mode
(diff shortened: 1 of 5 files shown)


docs/guides/server/fips.adoc
```diff
@@ -129,7 +129,7 @@ By using TRACE level, you can check that the startup log contains `KC` provider
 
 [source]
 ----
-KC(BCFIPS version 2.0 Approved Mode, FIPS-JVM: enabled) version 1.0 - class org.keycloak.crypto.fips.KeycloakFipsSecurityProvider,
+KC(BCFIPS version 2.0102 Approved Mode, FIPS-JVM: enabled) version 1.0 - class org.keycloak.crypto.fips.KeycloakFipsSecurityProvider,
```
ahus1, line 132: Maybe a follow-up issue: Could we simplify the docs and the handling for the users to log this information on the INFO level all the time?

rmartinc: The problem is that line is a very big debugging paragraph that lists all the providers defined in JDK ([here](https://github.com/keycloak/keycloak/blob/26.3.5/common/src/main/java/org/keycloak/common/crypto/CryptoIntegration.java)). I think it would be better to create a single line (for example in the `KeycloakFipsSecurityProvider` constructor) that is just shown in FIPS mode.

rmartinc: @mposolda I can do you what you prefer. We can create a follow up or just add a little info line in the `KeycloakFipsSecurityProvider` that shows the same data but only when that specific provider is created. This way we don't add as info that long debug paragraph with all the providers.

mposolda: @rmartinc +1 to possibly log just the first line in the `info` level and all the additional details with the `debug` level. Also +1 that this might be a follow-up issue. Feel free to create issue/PR for that.

mposolda (approved):

mposolda:
@rmartinc Thanks for the work on the upgrade!

docs/guides/server/fips.adoc
rmartinc: I created https://github.com/keycloak/keycloak/issues/43015 for the follow-up.

PR #42979, opened 2025-09-26
Clarification about supported SPIs

mposolda:
closes #42978

### DIFF
--- a/docs/documentation/server_development/topics/providers.adoc
+++ b/docs/documentation/server_development/topics/providers.adoc
@@ -501,3 +501,15 @@ Once you have a JAR file with a descriptor and the scripts you want to deploy, y
 === Available SPIs
 
 If you want to see list of all available SPIs at runtime, you can check `Provider Info` page in Admin Console as described in <<_providers_admin_console,Admin Console>> section.
+
+=== Supported SPIs
+
+Support for {project_name} SPIs is limited to what is documented in this guide. Moreover, SPI can be considered supported only if the method `isInternal` of the `Spi` described above returns `false` for the
+particular SPI.
+
+It is possible to implement any internal SPI in your extensions, but it is not officially supported and backwards compatibility regarding behaviour or method signatures is not guaranteed.
+This means your SPI implementation may stop working without notice once you upgrade to the newer version of the {project_name} server.
+
+Moreover it is recommended that your SPI implementation has dependency just on the {project_name} supported modules like `keycloak-server-spi` or `keycloak-model-storage`. Other modules
+(for instance `keycloak-server-spi-private` or `keycloak-services`) are considered internal and method signatures here can change without further notice.
+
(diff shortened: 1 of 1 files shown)


docs/documentation/server_development/topics/providers.adoc
```diff
@@ -501,3 +501,15 @@ Once you have a JAR file with a descriptor and the scripts you want to deploy, y
 === Available SPIs
 
 If you want to see list of all available SPIs at runtime, you can check `Provider Info` page in Admin Console as described in <<_providers_admin_console,Admin Console>> section.
+
+=== Supported SPIs
+
+Support for {project_name} SPIs is limited to what is documented in this guide. Moreover, SPI can be considered supported just if the method `isInternal` of the `Spi` described above returns `false` for the
```
ahus1:
```suggestion
Support for {project_name} SPIs is limited to what is documented in this guide. Moreover, SPI can be considered supported only if the method `isInternal` of the `Spi` described above returns `false` for the
```

ahus1 (commented):
I like it. Thank you very much! One small nit-bit below.

docs/documentation/server_development/topics/providers.adoc
```diff
@@ -501,3 +501,15 @@ Once you have a JAR file with a descriptor and the scripts you want to deploy, y
...
+
+=== Supported SPIs
+
+Support for {project_name} SPIs is limited to what is documented in this guide. Moreover, SPI can be considered supported only if the method `isInternal` of the `Spi` described above returns `false` for the
+particular SPI.
+
+It is possible to implement any internal SPI in your extensions, but it is not officially supported and backwards compatibility regarding behaviour or method signatures is not guaranteed.
+This means your SPI implementation may stop working without notice once you upgrade to the newer version of the {project_name} server.
+
+Moreover it is recommended that your SPI implementation has dependency just on the {project_name} supported modules like `keycloak-server-spi` or `keycloak-model-storage`. Other modules
```
stianst: This part is a bit unclear - what are the supported modules? ;)

mposolda: @stianst @ahus1 I've tried to make this part a bit more clear and added additional sentence, which clarifies this. I've used separate commit (just for the easier review of what has changed).

Besides that, I've added that supported modules `keycloak-server-spi` and `keycloak-model-storage` have dependency on `keycloak-core` and `keycloak-common`, but I suppose that those modules contain mostly utility methods and here we are not so strict regarding backwards compatibility? Is the formulation, which I used in the docs accurate?

I figured another thing, which is not ideal. The `keycloak-model-storage` has dependency on `keycloak-server-spi-private`. It means that anyone, who want to do his own user-storage Keycloak extension needs to have dependency on `keycloak-server-spi-private`. This looks like a flaw to me. Created separate issue https://github.com/keycloak/keycloak/issues/43626 for that and send another draft PR for that one. Review welcome for this one too.

PR #42981, opened 2025-09-26
Update documentation to prefer CacheCR in multi-site

### LINKED ISSUE #42980: Update documentation to prefer CacheCR in multi-site
### Description

The Keycloak automatic cache creation only creates caches in the local site. When multi-site is enabled, it means that Keycloak will fail to start until it is deployed on all sites. 
It may be simpler for the user (IMO) to set up CacheCR and let the Infinispan Operator create the caches, as it does not require Keycloak running everywhere.

pruivo:
Closes #42980


pruivo:
@ahus1 I'm adding back the Infinispan `CacheCR` documentation. The issue is that Keycloak creates the caches in the local site Infinispan cluster, and it will fail to start until the caches are defined in the remote sites.

ahus1 (approved):

Issue #10155, opened 2022-02-11
Step-up authentication for SAML clients


mposolda:
### Description

We may need to have support for SAML clients on the server side, so that SAML client has a way to send SAML request to the Keycloak server and request step-up authentication. Support in SAML adapters is not a priority right now.

SAML specification has some possibility to request ACR levels by `<RequestedAuthnContext>` parameter in the SAML authentication request. More described in the design https://github.com/keycloak/keycloak-community/blob/main/design/multi-factor-admin-and-step-up.md#saml .

Hynek mentioned in the discussion #10120 : `<saml:AuthnContextClassRef>` element takes anyURI, so e.g. `urn:keycloak:loa:FOO` or `urn:keycloak:acr:BAR` would be allowed values. Perhaps this would enable aligning OIDC and SAML implementation step-up-authentication-wise?

Few more older JIRAs for the reference:
https://issues.redhat.com/browse/KEYCLOAK-15205 Add customized AuthnContextClassRef for SAML client
https://issues.redhat.com/browse/KEYCLOAK-5224 Fill in saml2p:AuthnRequest section
https://issues.redhat.com/browse/KEYCLOAK-17788 Support setting AuthnContextClassRef in SAML response to SP

### Discussion

#10120 (Partially related)

### Other details

For the reference, some related specifications:

The academic world has settled for REFEDS' SFA and MFA profiles (https://refeds.org/profile/sfa, https://refeds.org/profile/mfa), which are used in the RequestedAuthnContext, and should also be returned in the AuthnContext reply. There are also OASIS values (https://docs.oasis-open.org/security/saml/v2.0/saml-authn-context-2.0-os.pdf) but at least that documentation is out of date with regard to modern MFA methods.

Example request:

```
    <samlp:RequestedAuthnContext>
        <saml:AuthnContextClassRef xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">https://refeds.org/profile/mfa</saml:AuthnContextClassRef>
    </samlp:RequestedAuthnContext>
```

```[tasklist]
### Tasks
- [ ] #11721 
```

mposolda:
Added to Keycloak 18, but can be postponed to later.

cbontemps:
Hi, here is an extra link for integration between webauthn and saml from FIDO alliance. It provides example how the SAML client can request a specific authentification level or authentication method. : https://fidoalliance.org/integrating-fido-and-federation-protocols/

cdevienne:
[...]

PR #42984, opened 2025-09-26
Support EntraID's AuthnContext values for multifactor authentication

### LINKED ISSUE #11721: Add customized AuthnContextClassRef for SAML client
### Description

Add a configurable attribute to be included in SAML response for specific SAML client.
We need RH-SSO to act as IDP to interact with a Microsoft ADFS as broker to its relying party, which requires AuthnContextClassRef to be explicitly passed. The requirement is, for a SAML client which has enabled "Include AuthnStatement",  to add a text box to enter a fixed value, to be included in the SAML response in section `<saml:AuthnContextClassRef></saml:AuthnContextClassRef>`

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ahus1:
Closes #11721

### DIFF
--- a/docs/documentation/server_admin/topics/clients/con-protocol-mappers.adoc
+++ b/docs/documentation/server_admin/topics/clients/con-protocol-mappers.adoc
@@ -62,6 +62,20 @@ Service account sessions provide the following details:
 * *clientAddress*: The remote host IP of the service account's authenticated device.
 * *clientHost*: The remote host name of the service account's authenticated device.
 
+== SAML AuthnContext mapper
+
+By default, {project_name} will set the AuthnContext with AuthnContextClassRef set to `urn:oasis:names:tc:SAML:2.0:ac:classes:unspecified`.
+
+To specify a different value based on specific authenticator reference types completed by the user during authentication, add the *Authenticator reference type to AuthnContext mapper*.
+
+Use this, for example, to forward information about a successful multifactor authentication to Entra ID as shown in the following example:
+
+Authenticator reference types:: `otp`, `recovery-authn-codes`, `webauthn`, `webauthn-passwordless`
+
+Authn Context:: `+http://schemas.microsoft.com/claims/multipleauthn+`
+
+See https://learn.microsoft.com/en-us/entra/identity/authentication/how-to-mfa-expected-inbound-assertions[Satisfy Microsoft Entra ID multifactor authentication (MFA) controls with MFA claims from a federated IdP] for additional information.
+
 == Script mapper
 
 Use the *Script Mapper* to map claims to tokens by running user-defined JavaScript code. For more details about deploying scripts to the server, see link:{developerguide_jsproviders_link}[{developerguide_jsproviders_name}].
--- a/services/src/main/java/org/keycloak/protocol/saml/mappers/AuthenticatorReferenceTypeToAuthnContextMapper.java
+++ b/services/src/main/java/org/keycloak/protocol/saml/mappers/AuthenticatorReferenceTypeToAuthnContextMapper.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.protocol.saml.mappers;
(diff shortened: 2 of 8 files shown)


ahus1:
@rmartinc / @mposolda - I learned that this can be a blocker for those who want to use Keycloak as an IDP for EntraID where EntraID is the master tenant that only supports SAML (not OIDC), and the master tenant enforces a 2nd factor. So I'm giving it a go. 

I chose to make this a mapper as this is specific to EntraID, and to keep the logic encapsulated. 

I tried to add some logic to detect if two factors have been used to authenticate the user. This works in the test, not sure if I miss something here. 

An alternative could be to configure a list of authenticators to be considered as part of the 2FA ensemble, or even two groups one for the first group, and another one for the second group. 

Once we agree on the design, I'll prepare additional documentation and test it with the community.

rmartinc:
@ahus1 We have seen several requests similar to this. The `AuthnContextClassRef` is a tag that the SAML specification uses to mark that a specific security context has been reached. By default the [spec](https://docs.oasis-open.org/security/saml/v2.0/saml-authn-context-2.0-os.pdf) mentions some of them like `urn:oasis:names:tc:SAML:2.0:ac:classes:Kerberos` or `urn:oasis:names:tc:SAML:2.0:ac:classes:X509` for kerberos or certificate logins. IMHO this should be managed when we implement the step-up auth in SAML (because you can also request a context when you login). But, if you want to add a generic mapper for this, I would create a full generic one that you can configure any URI on it (not this specific one, similar to what the issue comments about). I think there are similar requests for other vendors. It would be up to the customer the security implications of adding the mapper. IMHO they can implement a similar mapper too, it's not very complicated, and it's a hack.

ahus1:
Thank you for the feedback. I will see that I make it more generic in the set value. Still I will keep its use case for detecting multifactor authentication.

For me, leaving it to users to write Java code is the least favourite option. I want Keycloak to work with EntraID out of the box and with great docs. This will make Keycloak relevant for a wider community.

For now I hope this change is good enough for the problem at hand, and that it will not break in corner cases.

ahus1:
I pushed some more changes to make it generic. This is a screenshot of the new mapper that can be added a SAML client. It is designed to add one of such a mapper to a client.  

image

To ensure that those authenticators have been triggered for the user, update your authentication flows as necessary, and if needed have a custom flow for the specific client in case you want to enforce MFA for a specific client. 

To build a provider JAR, build this branch locally using `./mvnw clean install -DskipTests`, then run the following script in the services folder: 

```
#!/bin/bash
set -ex

cd target
rm -rf saml-authn-mapper
mkdir -p saml-authn-mapper/META-INF/services
cd classes
cp --parents "$(find . -name AuthenticatorReferenceTypeToAuthnContextMapper.class)" ../saml-authn-mapper
grep AuthenticatorReferenceTypeToAuthnContextMapper META-INF/services/org.keycloak.protocol.ProtocolMapper > ../saml-authn-mapper/META-INF/services/org.keycloak.protocol.ProtocolMapper

cd ..
rm -f saml-authn-mapper.jar
cd saml-authn-mapper
zip -r ../saml-authn-mapper.jar *

```

Then copy the JAR file to the `providers` folder of your Keycloak installation.

ahus1:
@rmartinc - we tested this with a user earlier today and it was working as expected once I added a workaround for #43949 

Please review when you have the time. I'll mark it ready-for-review once the build is green. Thanks!

rmartinc:
@ahus1 The previous weeks I was playing a bit with issue https://github.com/keycloak/keycloak/issues/10155 and it's not incredibly complicated. At least for the server part. So I would have preferred to not implement this and to go with the generic approach of using step-up implementation. I think it is more or less the same effort that you have spent here.

The main idea about step-up implementation is that the admin configures an authentication workflow with the values they consider. Each value is mapped to a AuthnContextClassRef. This way the client can request a specific auth context class ref in the rquest, the authentication flow should have that value mapped to a authn level (1, 2 or whatever), and after that we need to add a similar mapper to the one implemented here to attach the class ref to the response. It's generic and covers also the step-up implementation. You can also also set a minimum level for a client to ensure that client uses the level you want.

But as you have already done this I don't know what to say. I prefer to not support this extra mapper which IMHO makes no much sense once we have the generic step-up. But if you insists in adding this I will do the review.

ahus1:
@rmartinc - I'd say this is a working solution for some of the users, which has been waiting for a while to have support for Entra ID (and others). Note that the step-up authentication setup sounds more like a realm-level configuration, while it IMHO would need to contain a client-specific configuration as each client might want to see different values of the AuthnContext.

If you think that the step-up authentication will make it into a release soon (say, 26.5), we can make this feature preview-only, and then deprecate it once the more comprehensive solution is available.

rmartinc:
I have sent initial draft PR #44185 for issue #10155. It's working OK configuring mappings at client level, but we have to decide how to configure the same for the realm. I don't see a good solution for this. :confused:

ahus1:
@rmartinc - once you go down the ACR route and mappings, my knowledge is limited and looking at the code won't help much I fear. I'm happy to discuss your changes if you walk me through them if you think this is can help. Please go ahead and schedule a 1:1 for this.

ahus1:
With SAML step-up authentication implemented in #44185, this issue should now be resolved as well. 

Therefore I'm closing it and and I will also close the associated PR.

PR #42998, opened 2025-09-26
Add validation to workflow update so that only changes to the name an…

sguilhen:
…d enabled flag are allowed for now

Closes #42916

### DIFF
--- a/core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
```diff
@@ -146,6 +147,20 @@ public void setState(WorkflowStateRepresentation state) {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (!(obj instanceof WorkflowRepresentation)) {
+            return false;
+        }
+        WorkflowRepresentation that = (WorkflowRepresentation) obj;
+        // TODO: include state in comparison?
```
pedroigor, line 159: Status is a server-side information and not updatable by users. At least so far ...

pedroigor (approved):

PR #43005, opened 2025-09-26
Automatically redirect based on login hint

### LINKED ISSUE #42715: incorrect flow with login_hint specified for user in org email domain
### Before reporting an issue

### Area

organizations

### Describe the bug

Let's say I have an "entry" realm with an organization created in it. That org is associated with the email domain "foo.com". The org is also configured to redirect all users in that "foo.com" email domain through an IdP that sends them to a client in the "foocom" realm for authentication.

If a login flow directs me to the auth URL of the "entry" realm without any login_hint parameter specified, I get this (expected) identity-first flow:
- First I'm presented with a form for entering my email/username. This form is supplied from a URL in the "entry" realm.
- I enter a username in the foo.com email domain, e.g. "john.doe@foo.com".
- Now I'm presented with a form served from the "foocom" realm that asks for username+password. I have login_hint enabled for the IdP in the "entry" realm that was used for this transfer to the "foocom" realm, so my same email/username is already prefilled here.
- I fill in my password and submit. This authenticates me through the client in the "foocom" realm.

All good.

And, same behavior regardless of whether john.doe@foo.com has yet been added to the organization in the "entry" realm.

==========

Now, I modify the flow slightly so that the initial visit to the auth URL of the "entry" realm supplies a login_hint parameter. E.g. "?login_hint=john.doe@foo.com". (I also tried url-encoding it as "?login_hint=john.doe%40foo.com".)

In this case, I get some different behaviors... actually two different behaviors depending on whether or not john.doe@foo.com has yet been added to the org in the "entry" realm:

==========

In the case where the user has NOT yet been added to the org and the "entry" realm: As in the case above, I'm presented with a form for entering my email/username, served from the "entry" realm. The difference in this case is that john.doe@foo.com has been prefilled into this form.

This is nice! It would be nicer if the form was auto-submitted, but I can manually submit it and from this point on the flow works the same as above.

==========

But in the case where the user DOES already exist in the org in the "entry" realm: the initial form is auto-filled with the email/username AND it is also auto-submitted.

However, in this case the subsequent form for entering my password is still served from the "entry" realm, not the "foocom" realm. This violates my configuration for the org... all foo.com users should be delegated/forwarded through the specified IdP to be authenticated by a client in the "foocom" realm.

When I fill in my password here and submit the form, authentication will fail.

### Version

26.3.0

### Regression

### Expected behavior

When an organization is configured to delegate authentication to another realm for users in a given email domain, then that configuration should always be honored.

### Actual behavior

When login_hint is specified and the user already exists in the realm of the initial OIDC client, authentication is attempted in that initial realm.

### How to Reproduce?

(see description)

### Anything else?

_No response_

pedroigor:
Closes #42715

### DIFF
--- a/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
+++ b/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
@@ -23,7 +23,7 @@ import static org.keycloak.models.utils.KeycloakModelUtils.findUserByNameOrEmail
 import static org.keycloak.organization.utils.Organizations.getEmailDomain;
 import static org.keycloak.organization.utils.Organizations.isEnabledAndOrganizationsPresent;
 import static org.keycloak.organization.utils.Organizations.resolveHomeBroker;
-import static org.keycloak.utils.StringUtil.isNotBlank;
+import static org.keycloak.utils.StringUtil.isBlank;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


martin-kanis (approved):
LGTM

ahus1 (approved):

Issue #41057, opened 2025-07-10
Add native hCaptcha support as an alternative to reCAPTCHA


warioishere:
### Description

At present, Keycloak offers built-in support only for Google reCAPTCHA to protect public forms (such as login, registration, and forgotten password). While reCAPTCHA is widely used, it introduces multiple concerns:

    Privacy: reCAPTCHA collects user data and is not privacy-friendly by design.

    Vendor lock-in: relying solely on a single proprietary CAPTCHA provider contradicts the open-source philosophy of Keycloak.

    Compliance: in some regions (such as the EU), organizations may face compliance issues with Google services.

There are some community-contributed hCaptcha plugins, but they suffer from several problems:

    They are often outdated and no longer actively maintained.

    They require manual modifications to Keycloak themes or templates.

    They are fragile with Keycloak updates and frequently break compatibility, leading to maintenance overhead.

Proposal:

Please consider implementing native support for hCaptcha alongside reCAPTCHA in the official Keycloak distribution. This should ideally be configurable in the Admin Console in the same way as reCAPTCHA, without requiring any manual theme edits.

Why hCaptcha?

    hCaptcha is an open and privacy-friendly alternative to reCAPTCHA.

    It is widely adopted and provides a compatible challenge-response system.

    It allows organizations to comply with stricter privacy regulations.

    It aligns with Keycloak’s mission as an open-source identity provider by giving administrators freedom of choice.

Benefits:

    Reduces dependence on Google services.

    Improves compliance options for organizations in regulated sectors.

    Avoids the risks associated with outdated community plugins.

    Offers a straightforward way to enhance security without compromising privacy.

Suggested Implementation:

    Provide hCaptcha as a built-in feature selectable via the existing CAPTCHA configuration UI.

    Allow specifying the hCaptcha site key and secret key through the Admin Console.

    Ensure the server-side verification and client-side widget are fully integrated, just like reCAPTCHA.

    Make the feature optional and fully backward-compatible.

yahorsi:
+1 Vote to implement

PR #43019, opened 2025-09-29
feat: add hCaptcha registration form action

warioishere:
adds hcaptcha as requested in https://github.com/keycloak/keycloak/issues/41057

- generalize the registration captcha base so providers can supply their response field names, template metadata, and scripts
- add an hCaptcha form action with service registration, theme updates, and documentation for configuring the provider
- extend admin provider tests to expect the hCaptcha action in addition to the existing reCAPTCHA entries

### DIFF
--- a/docs/documentation/server_admin/topics/users/proc-enabling-recaptcha.adoc
+++ b/docs/documentation/server_admin/topics/users/proc-enabling-recaptcha.adoc
@@ -3,11 +3,11 @@
 // server_admin/topics/users.adoc
 
 [id="proc-enabling-recaptcha_{context}"]
-= Enabling reCAPTCHA
+= Enabling CAPTCHA
 
 [role="_abstract"]
-To safeguard registration against bots, {project_name} has integration with Google reCAPTCHA (see <<procedure_recaptcha>>) and reCAPTCHA Enterprise (see <<procedure_recaptcha_enterprise>>).
-The default theme (`register.ftl`) supports both v2 (visible, checkbox-based) and v3 (score-based, invisible) reCAPTCHA (see https://cloud.google.com/recaptcha/docs/choose-key-type[Choose the appropriate reCAPTCHA key type]).
+To safeguard registration against bots, {project_name} has integration with Google reCAPTCHA (see <<procedure_recaptcha>>), hCaptcha (see <<procedure_hcaptcha>>), and reCAPTCHA Enterprise (see <<procedure_recaptcha_enterprise>>).
+The default theme (`register.ftl`) supports both v2 (visible, checkbox-based) and v3 (score-based, invisible) reCAPTCHA (see https://docs.google.com/recaptcha/docs/choose-key-type[Choose the appropriate reCAPTCHA key type]).
 
 [[procedure_recaptcha]]
 == Setting up Google reCAPTCHA
@@ -48,6 +48,29 @@ NOTE: In {project_name}, websites cannot include a login page dialog in an ifram
 .. Enter `https://www.google.com` in the field for the *X-Frame-Options* header (or `https//www.recaptcha.net` if you enabled *Use recaptcha.net*).
 .. Enter `https://www.google.com` in the field for the *Content-Security-Policy* header (or `https//www.recaptcha.net` if you enabled *Use recaptcha.net*).
 
+
+[[procedure_hcaptcha]]
+== Setting up hCaptcha
+
+. Enter the following URL in a browser:
++
+[source,bash,subs=+attributes]
+----
+https://www.hcaptcha.com/
+----
+
+. Sign in and create a site in the hCaptcha dashboard to obtain your *Site Key* and *Secret*. Note them down for future use in this procedure.
+. Navigate to the {project_name} admin console.
+. Click *Authentication* in the menu.
+. Click the *Flows* tab.
+. Select *Registration* from the list.
+. Add the *hCaptcha* execution as a sub-step of *Registration Form* and set its requirement to *Required*.
+. Click the *gear icon* ⚙️ on the *hCaptcha* row.
+
++
+. Enter the *hCaptcha Site Key* generated from the hCaptcha dashboard.
+. Enter the *hCaptcha Secret* generated from the hCaptcha dashboard.
+. Save the configuration.
 
 [[procedure_recaptcha_enterprise]]
 == Setting up Google reCAPTCHA Enterprise
--- a/services/src/main/java/org/keycloak/authentication/forms/AbstractRegistrationRecaptcha.java
+++ b/services/src/main/java/org/keycloak/authentication/forms/AbstractRegistrationRecaptcha.java
@@ -22,6 +22,7 @@ package org.keycloak.authentication.forms;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import jakarta.ws.rs.core.MultivaluedMap;
 import org.jboss.logging.Logger;
@@ -47,6 +48,8 @@ import org.keycloak.utils.StringUtil;
 public abstract class AbstractRegistrationRecaptcha implements FormAction, FormActionFactory {
 
     public static final String G_RECAPTCHA_RESPONSE = "g-recaptcha-response";
+    public static final String H_CAPTCHA_RESPONSE = "h-captcha-response";
+    public static final Set<String> CAPTCHA_RESPONSE_FIELDS = Set.of(G_RECAPTCHA_RESPONSE, H_CAPTCHA_RESPONSE);
     public static final String RECAPTCHA_REFERENCE_CATEGORY = "recaptcha";
 
     // option keys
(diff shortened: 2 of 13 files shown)


warioishere:
@stianst Hi, why is this not given any attention on?

docs/documentation/server_admin/topics/users/proc-enabling-recaptcha.adoc
```diff
@@ -6,8 +6,8 @@
 = Enabling reCAPTCHA
```
mabartos:
```suggestion
= Enabling CAPTCHA
```

Would be probably better to have it general.

docs/documentation/server_admin/topics/users/proc-enabling-recaptcha.adoc
```diff
@@ -6,8 +6,8 @@
...
 
 [role="_abstract"]
-To safeguard registration against bots, {project_name} has integration with Google reCAPTCHA (see <<procedure_recaptcha>>) and reCAPTCHA Enterprise (see <<procedure_recaptcha_enterprise>>).
-The default theme (`register.ftl`) supports both v2 (visible, checkbox-based) and v3 (score-based, invisible) reCAPTCHA (see https://docs.cloud.google.com/recaptcha/docs/choose-key-type[Choose the appropriate reCAPTCHA key type]).
+To safeguard registration against bots, {project_name} has integration with Google reCAPTCHA (see <<procedure_recaptcha>>), hCaptcha (see <<procedure_hcaptcha>>), and reCAPTCHA Enterprise (see <<procedure_recaptcha_enterprise>>).
+The default theme (`register.ftl`) supports both v2 (visible, checkbox-based) and v3 (score-based, invisible) reCAPTCHA (see https://docs.google.com/recaptcha/docs/choose-key-type[Choose the appropriate reCAPTCHA key type]).
```
mabartos:
```suggestion
The default theme (`register.ftl`) supports both v2 (visible, checkbox-based) and v3 (score-based, invisible) reCAPTCHA (see https://docs.cloud.google.com/recaptcha/docs/choose-key-type[Choose the appropriate reCAPTCHA key type]).
```

The new link does not work.

testsuite/integration-arquillian/servers/auth-server/services/testsuite-providers/src/main/resources/theme/address/login/register.ftl
```diff
@@ -107,10 +107,15 @@
...
+            <#assign captchaProviderId = captchaProviderId!recaptchaProviderId!"" />
             <#if recaptchaRequired??>
             <div class="form-group">
                 <div class="${properties.kcInputWrapperClass!}">
-                    <div class="g-recaptcha" data-size="compact" data-sitekey="${recaptchaSiteKey}"></div>
+                    <#if captchaProviderId == "registration-hcaptcha-action">
+                        <div class="h-captcha" data-size="compact" data-sitekey="${recaptchaSiteKey}"></div>
+                    <#else>
+                        <div class="g-recaptcha" data-size="compact" data-sitekey="${recaptchaSiteKey}"></div>
+                    </#if>
```
mabartos: Would it be possible to generalize it that this register page does not care what CAPTCHA is used? For the class, you could set the form attribute as well.

themes/src/main/resources/theme/base/login/register.ftl
```diff
@@ -73,10 +73,15 @@
...
+            <#assign captchaProviderId = captchaProviderId!recaptchaProviderId!"" />
             <#if recaptchaRequired?? && (recaptchaVisible!false)>
                 <div class="form-group">
                     <div class="${properties.kcInputWrapperClass!}">
-                        <div class="g-recaptcha" data-size="compact" data-sitekey="${recaptchaSiteKey}" data-action="${recaptchaAction}"></div>
+                        <#if captchaProviderId == "registration-hcaptcha-action">
+                            <div class="h-captcha" data-size="compact" data-sitekey="${recaptchaSiteKey}"></div>
+                        <#else>
+                            <div class="g-recaptcha" data-size="compact" data-sitekey="${recaptchaSiteKey}"<#if recaptchaAction?has_content> data-action="${recaptchaAction}"</#if>></div>
+                        </#if>
```
mabartos: The same as above... the captcha action can be generalized as well to not check here in the FTL file if it's hcaptcha, or recaptcha. Later, there might be another impl.

themes/src/main/resources/theme/base/login/register.ftl
```diff
@@ -95,10 +100,14 @@
...
-                            data-sitekey="${recaptchaSiteKey}" data-callback='onSubmitRecaptcha' data-action='${recaptchaAction}' type="submit">
-                            ${msg("doRegister")}
-                        </button>
+                        <#if captchaProviderId == "registration-hcaptcha-action">
+                            <input class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonBlockClass!} ${properties.kcButtonLargeClass!}" type="submit" value="${msg("doRegister")}"/>
+                        <#else>
+                            <button class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonBlockClass!} ${properties.kcButtonLargeClass!} g-recaptcha"
+                                data-sitekey="${recaptchaSiteKey}" data-callback='onSubmitRecaptcha'<#if recaptchaAction?has_content> data-action='${recaptchaAction}'</#if> type="submit">
+                                ${msg("doRegister")}
+                            </button>
```
mabartos: Would be good to generalize it as well if possible.

themes/src/main/resources/theme/keycloak.v2/login/register.ftl
```diff
@@ -41,10 +46,14 @@
...
-                        ${msg("doRegister")}
-                    </button>
+                    <#if captchaProviderId == "registration-hcaptcha-action">
+                        <input class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonBlockClass!} ${properties.kcButtonLargeClass!}" type="submit" value="${msg("doRegister")}" id="kc-submit"/>
+                    <#else>
+                        <button class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonBlockClass!} ${properties.kcButtonLargeClass!} g-recaptcha"
+                                data-sitekey="${recaptchaSiteKey}" data-callback="onSubmitRecaptcha"<#if recaptchaAction?has_content> data-action="${recaptchaAction}"</#if> type="submit" id="kc-submit">
+                            ${msg("doRegister")}
+                        </button>
+                    </#if>
```
mabartos: The same as above.

services/src/main/java/org/keycloak/authentication/forms/RegistrationPage.java
```diff
@@ -52,7 +52,7 @@ public class RegistrationPage implements FormAuthenticator, FormAuthenticatorFac
     public static final String FIELD_USERNAME = "username";
     public static final String FIELD_LAST_NAME = "lastName";
     public static final String FIELD_FIRST_NAME = "firstName";
-    public static final String FIELD_RECAPTCHA_RESPONSE = "g-recaptcha-response";
+    public static final String FIELD_RECAPTCHA_RESPONSE = AbstractRegistrationRecaptcha.G_RECAPTCHA_RESPONSE;
```
mabartos, line 55: Is it used somewhere?

services/src/main/java/org/keycloak/authentication/forms/AbstractRegistrationRecaptcha.java
```diff
@@ -99,24 +102,43 @@ public void buildPage(FormContext context, LoginFormsProvider form) {
...
         form.setAttribute("recaptchaSiteKey", config.get(SITE_KEY));
         form.setAttribute("recaptchaAction", action);
         form.setAttribute("recaptchaVisible", !invisible);
+        form.setAttribute("recaptchaProviderId", getId());
+
+        form.setAttribute("captchaRequired", true);
+        form.setAttribute("captchaSiteKey", config.get(SITE_KEY));
+        form.setAttribute("captchaAction", action);
+        form.setAttribute("captchaVisible", !invisible);
+        form.setAttribute("captchaProviderId", getId());
```
mabartos: Would be good if we could use only the `captcha*` attributes, and not setting the `recaptcha*`. However, I'm wondering about backwards compatibility as some custom pages in custom themes may rely on the `recaptcha*` attributes.

Would be good to have only the `captcha*` attributes here, separated in a protected method, and the RegistrationRecaptcha could override it and would add also the `recaptcha*` attributes for now. There would be comment that it should be eventually removed.

By doing this, when the hCaptcha is used, no `recaptcha*` attributes are set on the page.

mabartos (changes requested):
@warioishere Nice work! Thanks for the PR. I have added some comments on what could be improved. 

You should also create a release note in the 26_5_0 file to advertise a little bit the new CAPTCHA implementation. 

You should also sign the commit and potentially add the Closes message there. 

Something like:
```
git commit -m "feat: add hCaptcha registration form action" -m "Closes #41057" -s
```

And do not include merge commits here. It should be only one commit - you can leverage the `git rebase`. 

Thanks!

PS: We could do more improvements with CAPTCHA as including it to login pages, using invisible variants, etc. But it's out of scope of this PR. It can be done in follow-up tasks.

stianst (commented):
We really need to figure out a way to automate testing around this. Lack of tests for reCaptcha has been an issue several times, at least for that one we can easily try it ourselves.

Looks like hCaptcha has a freebie version; that may be sufficient for having some nightly tests or something like that.

warioishere:
Thanks for the feedback! I've made all the requested improvements:

## ✅ Administrative Changes
- [x] Added release notes to `26_5_0.adoc` advertising the new hCaptcha support
- [x] Signed the commit with `-s` flag
- [x] Added `Closes #41057` reference
- [x] Cleaned up branch history - removed merge commits, now just one clean commit

## ✅ Documentation Improvements
- [x] Generalized title: "Enabling reCAPTCHA" → "Enabling CAPTCHA" (more inclusive)
- [x] Fixed broken documentation link: `https://docs.google.com/recaptcha/...`

## ✅ Code Refactoring - Provider-Agnostic Architecture
The key improvement is that **templates and forms no longer care which CAPTCHA provider is used**:

### Java Changes
- **AbstractRegistrationRecaptcha**: 
  - Added `setCaptchaAttributes()` method to set generic `captcha*` attributes
  - Added `setRecaptchaLegacyAttributes()` for backwards compatibility with custom themes
  - Added `usesCaptchaCallback()` to indicate if callback-based submission is needed
  - Added abstract `getCaptchaCssClass()` for provider-specific CSS class names

- **RegistrationRecaptcha** & **RegistrationRecaptchaEnterprise**: 
  - Implement `getCaptchaCssClass()` returning "g-recaptcha"
  - Implement `usesCaptchaCallback()` returning `true` (they use callbacks)
  - Set legacy `recaptcha*` attributes for backwards compatibility

- **RegistrationHCaptcha**:
  - Implements `getCaptchaCssClass()` returning "h-captcha"
  - Base implementation of `usesCaptchaCallback()` returns `false` (no callback needed)

### Template Changes
All register.ftl templates now use **generic attributes instead of provider ID checks**:
- Before: Hardcoded checks like `if captchaProviderId == "registration-hcaptcha-action"`
- After: Uses `${captchaCssClass!}`, `${captchaUseCallback!false}`, and generic `captcha*` attributes

Templates are now **completely provider-agnostic** - new CAPTCHA providers can be added without modifying any templates!

## Backwards Compatibility
Custom themes that rely on legacy `recaptcha*` attributes will continue to work. These attributes include TODO comments for future deprecation.

warioishere:
so its been a while since you reviewed the PR, is it really in mind of the keycloak dev team to only support a datasucking propietary captcha service?

@stianst

PR #43024, opened 2025-09-29
Hide read-only email attribute in update profile context with update … …email enabled

### LINKED ISSUE #42990: Hide email field from update profile page when email is read-only for user
### Description

If UPDATE_EMAIL is enabled and user has an empty email which is read-only for user, there is non-clickable email field in Update profile page. The field should be hidden or at least greyed out 

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

martin-kanis:
Closes #42990


stianst (changes requested):
Do not use `style` directly in html.

pedroigor (approved):

stianst (approved):

PR #43028, opened 2025-09-29
Add ECDSA as a valid key type that should return EC public key

rmartinc:
Closes #42588

Adding the ECDSA java key type to be converted to internal key type EC. The problem is that the key type is different in normal Java JCE provider (EC) than in the BC provider (ECDSA). So the key type should also take into account the ECDSA type returned by BC. As this was previously modified for EdDSA, the PR just adds a new case for the ECDSA case. There are more changes in tests. Now the upload tests also perform a token request to really check the uploaded certificate/public-key attrs in the client are working. There are tests for the three different key types (RSA, EC and EdDSA). The certificates and keys are created on the fly instead of using fixed files which are removed.


mposolda (approved):
@rmartinc Thanks!

PR #43039, opened 2025-09-29
fix: updating service account docs

### LINKED ISSUE #17268: [KC 20+] Obsolete documentation about sevice accounts?
### Description

Don't know if you consider obsolete documentation as bug or enhancement, I'll keep it in enhancements.

Either I am blind or the entire section about service account lists steps that are not available in the UI anymore:

(https://www.keycloak.org/docs/latest/server_development/index.html#authenticating-with-a-service-account)

Make sure the client is configured as follows:

1. client_id is a confidential client that belongs to the realm master
2. client_id has Service Accounts Enabled option enabled
3. client_id has a custom "Audience" mapper
...

### Discussion

_No response_

### Motivation

Update the documentation to reflect the current UI and workings.
Cannot create a service account right now, because docs seem wrong.

### Details

_No response_

shawkins:
closes: #17268

Also should "Service accounts roles" be "Service account roles"? - captured as #43087 and will move forward with this PR separately and it will be corrected later.

### DIFF
--- a/docs/documentation/server_development/topics/admin-rest-api.adoc
+++ b/docs/documentation/server_development/topics/admin-rest-api.adoc
@@ -51,12 +51,12 @@ To authenticate against the Admin REST API using a `client_id` and a `client_sec
 
 . Make sure the client is configured as follows:
 
-* `client_id` is a **confidential** client that belongs to the realm *master*
-* `client_id` has `Service Accounts Enabled` option enabled
-* `client_id` has a custom "Audience" mapper
-  ** Included Client Audience: `security-admin-console`
+* `client_id` belongs to the realm *master*
+* `client_id` has `Client authentication` on
+* `client_id` has the `Service accounts roles` option enabled
+* Assign `client_id` the `admin` role in the "Service accounts roles" tab.
 
-. Check that `client_id` has the role 'admin' assigned in the "Service Account Roles" tab.
+. Use the `Client Secret` from the "Credentials" tab to verify the service account access:
 
 [source,bash,subs="attributes+"]
 ----
--- a/docs/documentation/server_development/topics/admin-rest-api.adoc
+++ b/docs/documentation/server_development/topics/admin-rest-api.adoc
@@ -52,11 +52,11 @@ To authenticate against the Admin REST API using a `client_id` and a `client_sec
 . Make sure the client is configured as follows:
 
 * `client_id` belongs to the realm *master*
-* `client_id` has `Client authentication` on
-* `client_id` has the `Service accounts roles` option enabled
-* Assign `client_id` the `admin` role in the "Service accounts roles" tab.
+* `client_id` has *Client authentication* set to *On*
+* `client_id` has the *Service account roles* option enabled
+* Assign `client_id` the `admin` realm role in the *Service accounts roles* tab.
 
-. Use the `Client Secret` from the "Credentials" tab to verify the service account access:
+. Use the *Client Secret* from the *Credentials* tab to verify the service account access:
 
 [source,bash,subs="attributes+"]
 ----


docs/documentation/server_development/topics/admin-rest-api.adoc
```diff
@@ -51,12 +51,12 @@ To authenticate against the Admin REST API using a `client_id` and a `client_sec
 
 . Make sure the client is configured as follows:
 
-* `client_id` is a **confidential** client that belongs to the realm *master*
-* `client_id` has `Service Accounts Enabled` option enabled
-* `client_id` has a custom "Audience" mapper
-  ** Included Client Audience: `security-admin-console`
+* `client_id` belongs to the realm *master*
+* `client_id` has `Client authentication` on
```
ahus1:
```suggestion
* `client_id` has *Client authentication* set to *On*
```

docs/documentation/server_development/topics/admin-rest-api.adoc
```diff
@@ -51,12 +51,12 @@ To authenticate against the Admin REST API using a `client_id` and a `client_sec
...
+* `client_id` has the `Service accounts roles` option enabled
```
ahus1:
```suggestion
* `client_id` has the *Service account roles* option enabled
```

docs/documentation/server_development/topics/admin-rest-api.adoc
```diff
@@ -51,12 +51,12 @@ To authenticate against the Admin REST API using a `client_id` and a `client_sec
...
+* Assign `client_id` the `admin` role in the "Service accounts roles" tab.
```
ahus1:
```suggestion
* Assign `client_id` the `admin` realm role in the *Service accounts roles* tab.
```

docs/documentation/server_development/topics/admin-rest-api.adoc
```diff
@@ -51,12 +51,12 @@ To authenticate against the Admin REST API using a `client_id` and a `client_sec
...
 
-. Check that `client_id` has the role 'admin' assigned in the "Service Account Roles" tab.
+. Use the `Client Secret` from the "Credentials" tab to verify the service account access:
```
ahus1:
```suggestion
. Use the *Client Secret* from the *Credentials* tab to verify the service account access:
```

ahus1 (dismissed):
Thank you for this PR, I think this is good improvement to align it with the current UI. 

Reading the docs, it seems to indicate that the Admin REST API is accessible with a client created in the master realm. 

AFAIK the client only needs to be in the master ream when it wants to access all the realms. A client can be added to any other realm, and then it will be able to use the Admin API for that realm. Maybe you want to include this in the PR. 

See below some small edits picking up the formatting that I've seen used by Andy in the past. Maybe it would be worth pinging him once those are merged.

docs/documentation/server_development/topics/admin-rest-api.adoc
mposolda: @shawkins FYI. I've created dedicated PR for handling "Service account roles" . It is here https://github.com/keycloak/keycloak/pull/43345 . The PR is already merged now.

I suggest that this PR updates the text  *Service account roles* instead of *Service accounts roles* .

mposolda (dismissed):
@shawkins Marking as "Changes requested" just because of the "Service accounts roles" . See my inline comment for the details.

docs/documentation/server_development/topics/admin-rest-api.adoc
shawkins: @mposolda thanks for the other fix. The initial assumption that this would be committed first turned out to be invalid :)

shawkins:
> AFAIK the client only needs to be in the master ream when it wants to access all the realms. A client can be added to any other realm, and then it will be able to use the Admin API for that realm. Maybe you want to include this in the PR.

My main concern was getting the text up-to-date. Both of the examples shown in this doc are based upon the master realm. We can certainly add text that says you aren't restricted to only master admins, or would you rather base the example of of using a non-master realm?

> See below some small edits picking up the formatting that I've seen used by Andy in the past. Maybe it would be worth pinging him once those are merged.

Thank you, those have been added now. cc @andymunro

ahus1:
> We can certainly add text that says you aren't restricted to only master admins, or would you rather base the example of of using a non-master realm?

There are pros and cons for each. Looking at the whole document now, let's consider this ok for now as it makes the example simpler.

ahus1 (approved):

andymunro (approved):

ahus1 (approved):
With @andymunro's review, approving this one. Thank you for all the input and reviews!

PR #43044, opened 2025-09-29
Improve workflow logging messages

sguilhen:
- every execution gets its own id that can be used to track all activities related to that particular workflow execution

Closes #42952

### DIFF
--- a/services/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
+++ b/services/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
@@ -47,6 +47,7 @@ import static java.util.Optional.ofNullable;
 import static org.keycloak.representations.workflows.WorkflowConstants.CONFIG_CONDITIONS;
 import static org.keycloak.representations.workflows.WorkflowConstants.CONFIG_ENABLED;
 import static org.keycloak.representations.workflows.WorkflowConstants.CONFIG_NAME;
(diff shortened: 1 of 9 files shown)


services/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
```diff
@@ -586,4 +550,61 @@ public WorkflowConditionProviderFactory<WorkflowConditionProvider> getConditionP
...
+
+        // if we've reached the end of the workflow, check if it is recurring or if we can mark it as completed
+        if (workflow.isRecurring()) {
+            // if the workflow is recurring, restart it
+            context.restart();
+            processWorkflow(workflow, context, resourceId);
+        } else {
+            // not recurring, remove the state record
+            context.complete();
+            workflowStateProvider.remove(context.getExecutionId());
```
pedroigor, line 594: We could probably handle the state within the context when calling methods like `complete` or `restart`.

pedroigor (approved):

services/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
sguilhen: Yeah! That was my idea when I added those methods - allow us to migrate things to the context over time. State management and possibly even history recording of the completed steps could be handled there.

PR #43053, opened 2025-09-30
feat(token-manager): make OIDC and models TokenManagers pluggable via…

sashyo:
Relates to #43052

This PR introduces SPIs for both the models `TokenManager` and the OIDC `TokenManager` to make them pluggable.

- **Models (server-spi)**: `TokenManager` now a `Provider` with `TokenManagerFactory` and `TokenManagerSpi`. Default factory returns the stock implementation.
- **OIDC (services)**: New `TokenManagerProvider`, `TokenManagerProviderFactory`, and `DefaultTokenManagerProviderFactory`. Resolution is centralized via `TokenManager.resolve(session, logger)`.
- **Fallback**: If a provider is configured and loadable, Keycloak uses it. Otherwise, the default managers are used (backward compatible).

This enables deployments to supply custom token pipelines with:
```bash
--spi-token-manager-provider=<id>
--spi-oidc-token-manager-provider=<id>
```


sundayScoop:
Needed!

PR #43058, opened 2025-09-30
Fix OIDC IDP broker basic auth encoding

### LINKED ISSUE #26374: Workflow failure: Quarkus IT - FipsDistTest#testUnsupportedHttpsPkcs12KeyStoreInStrictMode
### Before reporting an issue

### Area

ci

### Describe the bug

[Quarkus IT (ubuntu-latest, zip)](https://github.com/keycloak/keycloak/actions/runs/7581549792/job/20649723076#logs)

`2024-01-19 09:25:26,452 ERROR [io.quarkus.vertx.http.runtime.QuarkusErrorHandler] (executor-thread-1) HTTP Request to //realms/master/ failed, error id: ff61811e-0fa3-4159-a6af-85130a71b2a8-93: java.lang.NullPointerException: Cannot invoke "io.quarkus.arc.ArcContainer.instance(java.lang.Class, java.lang.annotation.Annotation[])" because the return value of "io.quarkus.arc.Arc.container()" is null`

### Version

latest

### Expected behavior

-

### Actual behavior

-

### How to Reproduce?

-

### Anything else?

_No response_

### LINKED ISSUE #43022: Incorrect Basic Auth encoding for OIDC IDentity Provider when Client ID contains colon
### Before reporting an issue

### Area

oidc

### Describe the bug

Based on [OAuth2.1 Client Secret](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#name-client-secret), when an Identity Provider has `Client authentication` as `Client secret sent as basic auth` basic authentication encoding must be `: based64_encode(url_encode(client_id) + ":" + url_encode(client_secret))`.

Keycloak need to respect it and send the basic authenticator encoding based on OAuth 2.1

### Version

26.3.5

### Regression

### Expected behavior

Basic authentication encoding is `: based64_encode(url_encode(client_id) + ":" + url_encode(client_secret))`.

### Actual behavior

Basic authentication encoding is `: based64_encode(client_id + ":" +client_secret)`.

### How to Reproduce?

Create an OIDC Identity Provider with Basic Authenticator and client_id including colon (fe https://www.keycloak.org/). Observe the basic authentication that Keycloak sends.
When you will try to login, authentication will be failed due to wrong client credentials.
Keycloak client will try to authenticate with client_id = 'https'

### Anything else?

_No response_

rpjicond:
Ensures that the client_id and client_secret are URL-encoded before being Base64-encoded for the Basic Auth header, following RFC 6749. This fixes authentication failures when the client_id contains special characters.

Closes #26374
Closes #43022


ahus1 (dismissed):
Thank you for the PR. I pushed one more commit to fix it at the source, so all other consumers of that method can benefit from it. I also checked the return value which is null. 

To ensure that this works also in the future we need an integration test. Please provide one for this PR to proceed. Maybe @cgeorgilakis can help providing one.

cgeorgilakis:
> Thank you for the PR. I pushed one more commit to fix it at the source, so all other consumers of that method can benefit from it. I also checked the return value which is null.
> 
> To ensure that this works also in the future we need an integration test. Please provide one for this PR to proceed. Maybe @cgeorgilakis can help providing one.

Proposed test class to add : 
```
package org.keycloak.testsuite.broker;

import org.keycloak.broker.oidc.OIDCIdentityProviderConfig;
import org.keycloak.models.IdentityProviderModel;
import org.keycloak.models.IdentityProviderSyncMode;
import org.keycloak.protocol.oidc.OIDCLoginProtocol;
import org.keycloak.representations.idm.IdentityProviderRepresentation;

import java.util.Map;

import static org.keycloak.broker.oidc.OAuth2IdentityProviderConfig.TOKEN_ENDPOINT_URL;
import static org.keycloak.testsuite.broker.BrokerTestConstants.*;
import static org.keycloak.testsuite.broker.BrokerTestTools.*;

public class KcOidcBrokerColonAliasClientSecretBasicAuthTest extends AbstractBrokerTest {

    @Override
    protected BrokerConfiguration getBrokerConfiguration() {
        return new KcOidcBrokerColonAliasClientSecretBasicAuthTest.KcOidcBrokerColonAliasConfigurationWithBasicAuthAuthentication();
    }

    private class KcOidcBrokerColonAliasConfigurationWithBasicAuthAuthentication extends KcOidcBrokerConfiguration {

        public final static String CLIENT_ID_COLON = "https://kc-dev.general.gr/staging/realms/general";

        @Override
        public IdentityProviderRepresentation setUpIdentityProvider(IdentityProviderSyncMode syncMode) {
            IdentityProviderRepresentation idp = createIdentityProvider(IDP_OIDC_ALIAS, IDP_OIDC_PROVIDER_ID);
            Map<String, String> config = idp.getConfig();
            applyDefaultConfiguration(config, syncMode);
            config.put("clientAuthMethod", OIDCLoginProtocol.CLIENT_SECRET_BASIC);
            return idp;
        }

        @Override
        protected void applyDefaultConfiguration(final Map<String, String> config, IdentityProviderSyncMode syncMode) {
            config.put(IdentityProviderModel.SYNC_MODE, syncMode.toString());
            config.put("clientId", CLIENT_ID_COLON);
            config.put("clientSecret", CLIENT_SECRET);
            config.put("prompt", "login");
            config.put("loginHint", "true");
            config.put(OIDCIdentityProviderConfig.ISSUER, getProviderRoot() + "/auth/realms/" + REALM_PROV_NAME);
            config.put("authorizationUrl", getProviderRoot() + "/auth/realms/" + REALM_PROV_NAME + "/protocol/openid-connect/auth");
            config.put(TOKEN_ENDPOINT_URL, getProviderRoot() + "/auth/realms/" + REALM_PROV_NAME + "/protocol/openid-connect/token");
            config.put("logoutUrl", getProviderRoot() + "/auth/realms/" + REALM_PROV_NAME + "/protocol/openid-connect/logout");
            config.put("userInfoUrl", getProviderRoot() + "/auth/realms/" + REALM_PROV_NAME + "/protocol/openid-connect/userinfo");
            config.put("defaultScope", "email profile");
            config.put("backchannelSupported", "true");
            config.put(OIDCIdentityProviderConfig.JWKS_URL,
                    getProviderRoot() + "/auth/realms/" + REALM_PROV_NAME + "/protocol/openid-connect/certs");
            config.put(OIDCIdentityProviderConfig.USE_JWKS_URL, "true");
            config.put(OIDCIdentityProviderConfig.VALIDATE_SIGNATURE, "true");
        }

        @Override
        public String getIDPClientIdInProviderRealm() {
            return CLIENT_ID_COLON;
        }

    }
}
```

You could investigate any tests failure.

pedroigor:
I would be more comfortable updating the `AbstractOAuth2IdentityProvider` because, AFAIK, the HTTP Basic specification does not say anything about url encoding the username and password.

Also, I'm wondering if this note from OAuth2.1 spec is mentioning a problem with url encoding?

> Note: This method of initially form-encoding the client identifier and secret, and then using the encoded values as the HTTP Basic authentication username and password, has led to many interoperability problems in the past. Some implementations have missed the encoding step, or decided to only encode certain characters, or ignored the encoding requirement when validating the credentials, leading to clients having to special-case how they present the credentials to individual authorization servers. Including the credentials in the request body content avoids the encoding issues and leads to more interoperable implementations.

sschu:
If we don't use the proposed fix, we should at least change the `getBytes()` method to specify the encoding to use, otherwise it will be platform-dependent.

ahus1:
> I would be more comfortable updating the AbstractOAuth2IdentityProvider

@pedroigor - thank you for pointing this out. I reverted my changes to the PR, and added the test case @cgeorgilakis suggested. Let's wait for the CI.

ahus1:
> If we don't use the proposed fix, we should at least change the getBytes() method to specify the encoding to use, otherwise it will be platform-dependent.

@sschu - the `kc.sh` will run Keycloak with UTF-8 encoding in all environments, so this should not be a problem IMHO.

ahus1:
@pedroigor / @sschu - do you think this is a "notable" change for the migration guide, as it might break some people's implementations? 

While we should now be standard compliant, would we need to provide a switch to revert to the old behavior? I hope not... 

Also asking: Should this be backported?

cgeorgilakis:
> I would be more comfortable updating the `AbstractOAuth2IdentityProvider` because, AFAIK, the HTTP Basic specification does not say anything about url encoding the username and password.
> 
> Also, I'm wondering if this note from OAuth2.1 spec is mentioning a problem with url encoding?
> 
> > Note: This method of initially form-encoding the client identifier and secret, and then using the encoded values as the HTTP Basic authentication username and password, has led to many interoperability problems in the past. Some implementations have missed the encoding step, or decided to only encode certain characters, or ignored the encoding requirement when validating the credentials, leading to clients having to special-case how they present the credentials to individual authorization servers. Including the credentials in the request body content avoids the encoding issues and leads to more interoperable implementations.

That’s why we mainly use Client Secret as POST.

However, Client Secret as basic auth also exists. 
The bad things is that dynamic client registration return this in response for Client with secret. However, this may be a broken change.

Another test was broken with change in CLIENT_SECRET. Will be a solution overridden this also in new test?

ahus1:
@cgeorgilakis - I wanted to update the test to also check if the test works for secrets that need URL encoding. While most tests did, the one that relied on values in the vault failed. I've reverted the change, and I assume the next build will be green.

sschu:
> > If we don't use the proposed fix, we should at least change the getBytes() method to specify the encoding to use, otherwise it will be platform-dependent.
> 
> @sschu - the `kc.sh` will run Keycloak with UTF-8 encoding in all environments, so this should not be a problem IMHO.

I saw this is a bit inconsistent across the codebase, in some places it is specified, in some places it isn't. I would probably prefer to have it explicit but its not urgent.

sschu:
> @pedroigor / @sschu - do you think this is a "notable" change for the migration guide, as it might break some people's implementations?
> 
> While we should now be standard compliant, would we need to provide a switch to revert to the old behavior? I hope not...
> 
> Also asking: Should this be backported?

I think I would mention this but not provide a compatibility switch since the old behaviour was essentially a bug.

ahus1:
@pedroigor - I've added a "notable change". Please review when you have the time, and then I would proceed with merging and backporting to 26.2 + 26.4

ahus1:
@pedroigor / @cgeorgilakis / @sschu - I've updated this PR after talking to @srose as this change can break people's setup as we now start to use proper encoding. 

So the old behavior is now available, although you need to switch your configuration in case you want to use it. The new default behavior is the correct behavior.

Please review and approve. Thanks!

sschu (approved):
LGTM!

pedroigor (approved):

PR #43060, opened 2025-09-30
Do not remove sid claim when the session is transient only for the client

rmartinc:
Closes #42565

The sid claim was removed when the session was transient but only for the client (the user session was online, but the client was transient). We need to maintain the claim in that situation. So the PR changes the if to use the session type and only for transient the sid claim is removed. Test added with the customer reproducer. The test uses a new user because if not other tests are affected as the scope for the user include the role in the requester clients.


mposolda (approved):
@rmartinc Thanks for catching and fixing this!
@thomasdarimont Thanks for the analysis on the issue!

PR #43063, opened 2025-09-30
The fix will separate the option description from the link in the Referrer Policy settings

burnedoutman:
Closes #43061


ahus1:
@burnedoutman - thank you for the PR. Looking at the code at hand, I found that the text shouldn't be in the `tsx` file, but in the message bundle, so I pushed one more commit here. I'll involve some more UI team members for input.

ahus1:
@jonkoops @ssilvert - I updated this PR and I found some items that surprised me: 

* The extra link passed via `<Trans />` only works when I prepend it with `{" "}`, otherwise the link is not clickable. Is that expected?
* The ` <1>Learn more</1>` seems to be cleaned from the strings and not used. So I removed it from the message resources. Is that the correct way to do it? Please confirm.

jonkoops:
@burnedoutman can you link to your issue from this PR by using a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue)? Please also update the commit description (e.g. `Closes #43061`). Please also use a descriptive title for your PR and commit that actually reflects the work being done (see ["How to Write a Git Commit Message"](https://cbea.ms/git-commit/)).

jonkoops:
@ahus1 perhaps this [alternative syntax](https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0) would work better? AFAIK `Trans` should interpolate whatever component it finds first into `<1></1>`, but perhaps there is a bug when using an indexed slot.

ahus1:
@jonkoops - Reading the docs, I now understand that without the `{" "}`, with an index starting with 0, `<FormattedLink...` was the item 0, and with the prefix it would be item `1`. 

IMHO it doesn't make much sense here to have a fallback, so I've simplified it in one more commit without `<Trans ...`: As the link is always added at the end, I prefer to remove it from the strings, simplifying the translation. 

Please let me know if you think this PR is ready to be merged. Thanks!

jonkoops:
> As the link is always added at the end, I prefer to remove it from the strings, simplifying the translation.

This is not an assumption we can rely on, as languages could potentially change the position of the link to localize it (e.g. RTL languages, or languages that change the preposition).

ahus1:
> Languages could potentially change the position of the link to localize it (e.g. RTL languages, or languages that change the preposition).

For RTL, I'd argue it would still be at the end, which then would be on the left for RTL. It is more of a separate sentence or even paragraph where we then link to the docs. 

Let me know if I can change your mind about that. I would then need to come up with some dummy text in the TSX file to replace "Default value prevents pages from being included" with something like "This text should never be shown, otherwise someone forgot to set an English message key."

jonkoops:
@ahus1 for the alternative syntax I linked above, a default value should be optional, so we could do something along the lines of:

```jsx
labelIcon={
  <Trans
    i18nKey="xRobotsTagHelp"
    components={{ 
      link: <FormattedLink href={url} title={t("learnMore")} />
    }}
  />
}
```

```properties
xRobotsTagHelp=Prevent pages from appearing in search engines. <link>Learn more</link>
```

jonkoops (approved):

ahus1:
@jonkoops - see last commit for a change. `link` doesn't work as the documentation states that it is a self closing tag. So I needed to choose a different one. 

Please let me know if you would approve this. If we are now aligned, I will need to update the translation files with the new tag.

ahus1:
@ssilvert - can you please approve this PR? Thanks!

ssilvert (approved):

Issue #42913, opened 2025-09-24
[RLM] - Allow using time-based tokens when setting fields that expect a period or time


pedroigor:
### Description

Today, properties that are related to defining a period or time, such as `after` (steps), are set in milliseconds.

Instead, we should allow users to use time-based tokens in strings such as:

* 1d
* 20s
* 1y
* 6M

This should improve the workflow definition so that users can define periods more easily.

pedroigor:
Recently, we had this change from @ryanemerson. See https://github.com/keycloak/keycloak/pull/42617.

After discussing with Ryan how to reuse the same code to parse time-based tokens, we might want to create our own version of `io.quarkus.runtime.configuration.DurationConverter` so that we can use it not only in the Quarkus distribution but from other places in our codebase.

PR #43107, opened 2025-10-01
Expose scheduled tasks and start time

### LINKED ISSUE #43106: Expose scheduled tasks and start time
### Description

This feature **extends the Timer SPI** to surface the currently scheduled tasks and key metadata (runnable, start time, interval).

#### `TimerProvider` (server-spi-private)

- **Add**
  ```java
  List<TimerTaskContext> getTasks();
  ```
  Returns a snapshot of all scheduled task contexts known to the provider.

#### `TimerTaskContext` (server-spi-private)

- **Extend** with:
  ```java
  Runnable getRunnable();
  long getStartDateMillis();
  long getIntervalMillis();
  ```

#### `TimerTaskContextImpl` (services)

- **Implement** the new getters:
  - `getRunnable()` returns the underlying runnable for the task.
  - `getStartDateMillis()` exposes the epoch millis when the task began.
  - `getIntervalMillis()` exposes the repeat interval in millis (0 for one-shot).

#### `BasicTimerProvider` (services)

- **Implement** `getTasks()` by exposing an immutable snapshot of the internal `scheduledTasks` (e.g., values of the provider’s task context map).

#### `BasicTimerProviderFactory` (services)

- **No behavioral changes**. Continues to construct `BasicTimerProvider`; the factory does not implement `getTasks()`.

### Motivation

During our work with Keycloak’s timer subsystem, we found that while tasks can be scheduled and cancelled, there’s no way to inspect what tasks are currently running, when they started, or at what interval.

In our case (Tide Foundation), we run Keycloak alongside our Cybersecurity Fabric — a decentralized infrastructure for zero-knowledge key management. We rely on periodic jobs for license monitoring and auto-renewal, and to meet our reliability and audit requirements. While this reflects our immediate operational needs, the requirement is generic: any deployment running health checks, cleanups, sync loops, or metering jobs benefits from first-class visibility into scheduled tasks without log scraping or forking.

For any environment that requires monitoring, debugging, or external orchestration of timers, this lack of visibility becomes a constraint. For example:

- **Operational visibility:** Admins or monitoring systems need to see which periodic jobs are currently registered, when they started, and their repeat interval.  
- **Debugging:** Developers troubleshooting clustered or long-running jobs benefit from being able to inspect tasks directly through the SPI rather than digging through logs.  
- **Extension points:** Integrators might need to expose scheduled jobs to external dashboards, policy enforcement, or lifecycle orchestration tools.  

By exposing scheduled tasks via the `TimerProvider`, we align with Keycloak’s extensibility model and make this subsystem more transparent without altering its default scheduling behavior.

It lets operators and integrators wire Keycloak timers into existing dashboards, SLO/SLA checks, and lifecycle orchestration, which is increasingly important in clustered and regulated environments.

### Value Proposition

- **Operational insight:** Allows operators to see which tasks are scheduled, when they started, and how often they run.
- **Debugging aid:** Makes it easier to trace stuck or misconfigured scheduled jobs.
- **Extensibility:** Provides a seam for external monitoring, dashboards, or orchestration tools to query timer state.
- **Backward compatibility:** All new methods are additive; existing scheduling APIs remain unchanged.

### Goals

- Allow retrieval of currently scheduled tasks via `TimerProvider`.
- Provide metadata (`startDateMillis`, `intervalMillis`, `runnable`) for each task.
- Ensure the default in-memory/basic provider implements this consistently.

### Non-Goals

- Redesigning the timer subsystem or its persistence model.
- Adding clustering or durability features for scheduled tasks.
- Changing the scheduling semantics for existing jobs.

### Discussion

This proposal relates to a broader series of proposals interlinked with this original discussion: **#41350**

### Notes

- Add `getTasks()` method to `TimerProvider` interface.
- Update `TimerTaskContext` to expose start time and interval (and runnable).
- Ensure `BasicTimerProvider` implements these new methods and exposes a safe snapshot.
- Add functional/integration tests to verify that scheduled tasks are correctly reported through the SPI.

sashyo:
Expose scheduled timer tasks (runnable, start time, interval) via the Timer SPI to improve operational visibility and debugging.

Closes #43106 

## Changes
- Add `List<TimerTaskContext> getTasks()` to `TimerProvider`.
- Extend `TimerTaskContext` with:
  ```java
  Runnable getRunnable();
  long getStartDateMillis();
  long getIntervalMillis();

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/timer/TimerProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/timer/TimerProvider.java
@@ -19,6 +19,10 @@ package org.keycloak.timer;
 
 import org.keycloak.provider.Provider;
 
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
+
 /**
  * @author <a href="mailto:sthorger@redhat.com">Stian Thorgersen</a>
  */
@@ -44,10 +48,13 @@ public interface TimerProvider extends Provider {
      */
     public TimerTaskContext cancelTask(String taskName);
 
+    public Map<String, TimerTaskContext> getTasks();
 
     interface TimerTaskContext {
 
         Runnable getRunnable();
+
+        long getStartTimeMillis();
 
         long getIntervalMillis();
     }
--- a/services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
+++ b/services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
@@ -18,13 +18,21 @@
 package org.keycloak.timer.basic;
 
 import org.jboss.logging.Logger;
+import org.keycloak.common.util.Time;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.services.scheduled.ScheduledTaskRunner;
 import org.keycloak.timer.ScheduledTask;
 import org.keycloak.timer.TimerProvider;
 
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.stream.Collectors;
 
 /**
  * @author <a href="mailto:sthorger@redhat.com">Stian Thorgersen</a>
@@ -54,7 +62,7 @@ public class BasicTimerProvider implements TimerProvider {
             }
         };
 
-        TimerTaskContextImpl taskContext = new TimerTaskContextImpl(runnable, task, intervalMillis);
+        TimerTaskContextImpl taskContext = new TimerTaskContextImpl(runnable, task, Time.currentTimeMillis(), intervalMillis);
         TimerTaskContextImpl existingTask = factory.putTask(taskName, taskContext);
         if (existingTask != null) {
             logger.debugf("Existing timer task '%s' found. Cancelling it", taskName);
@@ -87,4 +95,8 @@ public class BasicTimerProvider implements TimerProvider {
         // do nothing
     }
 
+    @Override
+    public Map<String, TimerTaskContext> getTasks() {
+        return Collections.unmodifiableMap(new HashMap<>(factory.getTasks()));
+    }
 }
--- a/services/src/main/java/org/keycloak/timer/basic/BasicTimerProviderFactory.java
+++ b/services/src/main/java/org/keycloak/timer/basic/BasicTimerProviderFactory.java
@@ -23,6 +23,9 @@ import org.keycloak.models.KeycloakSessionFactory;
 import org.keycloak.timer.TimerProvider;
 import org.keycloak.timer.TimerProviderFactory;
 
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 import java.util.Timer;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -75,4 +78,7 @@ public class BasicTimerProviderFactory implements TimerProviderFactory {
         return scheduledTasks.remove(taskName);
     }
 
+    protected Map<String, TimerTaskContextImpl> getTasks(){
+        return scheduledTasks;
+    }
 }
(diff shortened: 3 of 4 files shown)


services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
```diff
@@ -54,7 +57,7 @@ public void run() {
             }
         };
 
-        TimerTaskContextImpl taskContext = new TimerTaskContextImpl(runnable, task, intervalMillis);
+        TimerTaskContextImpl taskContext = new TimerTaskContextImpl(runnable, task, Instant.now().toEpochMilli() , intervalMillis);
```
stianst: Use `org.keycloak.common.util.Time` instead of `Instant`

```suggestion
        TimerTaskContextImpl taskContext = new TimerTaskContextImpl(runnable, task, Time.currentTimeMillis() , intervalMillis);
```

server-spi-private/src/main/java/org/keycloak/timer/TimerProvider.java
```diff
@@ -44,11 +47,14 @@ public default void scheduleTask(ScheduledTask scheduledTask, long intervalMilli
      */
     public TimerTaskContext cancelTask(String taskName);
 
+    public List<TimerTaskContext> getTasks();
 
     interface TimerTaskContext {
 
         Runnable getRunnable();
 
+        long getStartDateMillis();
```
stianst:
```suggestion
        long getStartTimeMillis();
```

server-spi-private/src/main/java/org/keycloak/timer/TimerProvider.java
```diff
@@ -44,11 +47,14 @@ public default void scheduleTask(ScheduledTask scheduledTask, long intervalMilli
      */
     public TimerTaskContext cancelTask(String taskName);
 
+    public List<TimerTaskContext> getTasks();
```
stianst: Should return `Map<String, TimerTaskContext>` or we'd need to add `name` to `TimerTaskContext`; this would be useful if you want to list the tasks and then cancel them as you'd need the name to do that.

stianst (changes requested):
Needs some minor changes, see inline comments

js/pnpm-lock.yaml
stianst, line 1: Please remove the update to this file

js/pnpm-lock.yaml
```diff
@@ -7513,6 +7513,7 @@ snapshots:
       pkg-types: 1.3.1
       ufo: 1.6.1
 
+
```
stianst, line 7516:
```suggestion
```

stianst, line 7516:
```suggestion
```

stianst (commented):
Looks good, just need to remove the changes to `pnpm-lock.yaml` and we're ready to go

js/pnpm-lock.yaml
sashyo: my bad, update has been removed

stianst:
@sashyo changes look good now; but there's something strange about the PR, probably caused by a merge commit as some point. If you see in the PR there are 25 commits, while there should just be the one; simplest fix might just be to reset to main and cherry-pick the one commit you have.

sashyo:
> @sashyo changes look good now; but there's something strange about the PR, probably caused by a merge commit as some point. If you see in the PR there are 25 commits, while there should just be the one; simplest fix might just be to reset to main and cherry-pick the one commit you have.

Hey @stianst , i think i fixed up the PR

stianst (approved):

pedroigor:
@stianst @sashyo I was about to approve the PR. However, I was evaluating some of the issues we have today with the default `TimerProvider`, as well as the capabilities that we might want to add to better address the use cases for other features like Workflows.

Starting with the issues we have today, the default `TimerProvider` is based on `java.util.Timer` and depending on how your `ScheduledTask` tasks are defined, it might end up killing the timer thread and stop the execution of critical server-level tasks, such as those to expire sessions and many others. Even though not a blocker for having this PR merged, I think this is an important concern and something we should try to address as a follow-up. Perhaps we want a different thread model (e.g.: based on `ExecutorService`).

As a capability, I would like to introduce the idea of a task instance and have an ID assigned to it so that we can track multiple `SchedulerTask` executions. 

Another capability would be to track the state of a task. Something like, if a task is waiting, running, cancelled, etc.

Another capability would be to enable tasks to have a state. That could be a simple map that is shared between the task and the server runtime.

Another capability would be to allow setting the task schedule time based on a time-based token similar to https://github.com/keycloak/keycloak/issues/42913.

These are all follow-ups that we can build on top of the changes merged on this PR. Just wanted to let you know some of the things I was trying to make sense of before providing my review and proposing some next steps.

PR #43121, opened 2025-10-01
Add a debug statement when the KeycloakFipsSecurityProvider is created

### LINKED ISSUE #43015: Log FIPS provider using info debug level
### Description

The [documentation about FIPS](https://www.keycloak.org/server/fips#_strict_mode) comments about setting trace level to check if the FIPS provider is being using OK. The debug line is:

```
KC(BCFIPS version 2.0 Approved Mode, FIPS-JVM: enabled) version 1.0 - class org.keycloak.crypto.fips.KeycloakFipsSecurityProvider,
```

The idea of this issue is changing the debug to info to show that line (at least with FIPS mode activated). But currently that line is dumped with all the providers in a long trace paragraph ([here](https://github.com/keycloak/keycloak/blob/26.3.5/common/src/main/java/org/keycloak/common/crypto/CryptoIntegration.java#L41)). Probably the best is just logging the line of the fips provider when fips is enabled.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

rmartinc:
Closes #43015

@mposolda @ahus1 Draft PR for the debug info for the FIPS provider. This was my idea. The info line is executed in the `KeycloakFipsSecurityProvider` and this way it's only displayed when FIPS is configured. Tests modified do not require the extra debug parameters. The only snag that this line is also present for commands like `kcadm`or `kcreg` (but I think we cannot do anything for that), so I needed to also change the related tests when executed in FIPS. Let me know if you prefer the line in another class or you prefer another text. There will be issues in the docs because we need to adjust some links after the 26.4 release (I'll rebase when they are fixed).

### DIFF
--- a/crypto/fips1402/src/main/java/org/keycloak/crypto/fips/KeycloakFipsSecurityProvider.java
+++ b/crypto/fips1402/src/main/java/org/keycloak/crypto/fips/KeycloakFipsSecurityProvider.java
@@ -26,6 +26,7 @@ public class KeycloakFipsSecurityProvider extends Provider {
                 ", FIPS-JVM: " + isSystemFipsEnabled() +
                 ")", 1, "Keycloak pseudo provider");
         this.bcFipsProvider = bcFipsProvider;
+        logger.infof("KeycloakFipsSecurityProvider created: %s", this.toString());
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


ahus1:
>  The only snag that this line is also present for commands like kcadm or kcreg

I assume that people who are running in FIPS mode are delighted  every time they see this message. So having it there is probably fine. 

So I'm good with this change.

ahus1:
> There will be issues in the docs because we need to adjust some links after the 26.4 release (I'll rebase when they are fixed).

If it is the external link checker, this one is not blocking you from merging PRs. It can be fixed in a follow-up PR.

rmartinc:
It seems that are fixed now, they were failing to me in my branch (maybe I have some last commits not applied locally).

rmartinc:
Moving out of draft then. @mposolda Take a look when you have time.

mposolda (approved):
@rmartinc @ahus1 Thanks for the PR and feedback!

PR #43122, opened 2025-10-01
Including OTLP headers for tracing

### LINKED ISSUE #41007: Including OTLP headers for authorization
### Description

Hello,
Is it planned to have quarkus.otel.exporter.otlp.headers (QUARKUS_OTEL_EXPORTER_OTLP_HEADERS) for authorization?

### Discussion

_No response_

### Motivation

Having an authorization option when sending traces to collector.

### Details

_No response_

mabartos:
- Closes #41007

### DIFF
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -100,9 +100,9 @@ spec:
 ----
 NOTE:  The name format of options defined in this way is identical to the key format of options specified in the configuration file.
        For details on various configuration formats, see <@links.server id="configuration"/>.
-       
-NOTE: If you configure mTLS with `https-client-auth` set to `required`, this configuration is inherited by the management interface. 
-If you have not otherwise configured the usage of HTTP for the health endpoints, you will likely want to also set the additionalOption `https-management-client-auth` to `request` or `none` so that a valid client certificate is not required for probe requests. 
+
+NOTE: If you configure mTLS with `https-client-auth` set to `required`, this configuration is inherited by the management interface.
+If you have not otherwise configured the usage of HTTP for the health endpoints, you will likely want to also set the additionalOption `https-management-client-auth` to `request` or `none` so that a valid client certificate is not required for probe requests.
 
 ==== Custom environment variables
 
@@ -297,11 +297,11 @@ If you do not specify a custom topology spread constraint, your Pods will have a
 
 ==== Job Scheduling
 
-Realm import and update Jobs inherit the scheduling of the server Pods if it is specified in the Keycloak CR `spec.scheduling` field. If your server Pod scheduling is highly restrictive this may result in the 
+Realm import and update Jobs inherit the scheduling of the server Pods if it is specified in the Keycloak CR `spec.scheduling` field. If your server Pod scheduling is highly restrictive this may result in the
 Job Pods being unschedulable. In this situation you may override the Job scheduling using the Keycloak CR `spec.update.scheduling` and `spec.import.scheduling` fields.
- 
-For example the following will remove the use of any server Pod scheduling for the update Job: 
- 
+
+For example the following will remove the use of any server Pod scheduling for the update Job:
+
 [source,yaml]
 ----
 apiVersion: k8s.keycloak.org/v2alpha1
@@ -409,6 +409,13 @@ spec:
   additionalOptions:
     - name: tracing-jdbc-enabled
       value: false                            # default 'true'
+    - name: tracing-header-My-Custom-header
+      value: keycloak-custom-header
+    - name: tracing-header-Authorization
+      secret:
+        name: tracing-secret
+        key: token
+
 ----
 
 These fields should reflect 1:1 association with `tracing-*` options that contain more information.
--- a/quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
@@ -19,26 +19,6 @@ public class WildcardOptionsUtil {
      */
     public static final String WILDCARD_END = ">";
 
-        /**
-     * Determines whether the given configuration key represents a wildcard option.
-     * <p>
-     * Examples:
-     * <pre>{@code
-     * isKcWildcardOption("tracing-header-<header>")      → "true"
-     * isKcWildcardOption("tracing-header-<headxxx")      → "false"
-     * isKcWildcardOption("tracing-header-headxxx>")      → "false"
-     * isKcWildcardOption("db-kind-<datasource>")         → "true"
-     * isKcWildcardOption("http-port")                    → "false"
-     * isKcWildcardOption("quarkus.<sth>.end")            → "false"
-     * }</pre>
-     *
-     * @param key the configuration key to check
-     * @return {@code true} if the key represents a Keycloak wildcard option
-     */
-        public static boolean isKcWildcardOption(String key) {
-        return key != null && key.contains(WILDCARD_START) && key.endsWith(WILDCARD_END);
-    }
-
     /**
      * Determines whether the given configuration key represents a wildcard option (contains variable segment)
      * <p>
@@ -61,7 +41,7 @@ public class WildcardOptionsUtil {
 
     /**
      * Extracts the prefix part of a wildcard key.
-     * You should always check the presence of the wildcard via the {@link #isKcWildcardOption(String)}.
+     * You should always check the presence of the wildcard via the {@link #isWildcardOption(String)}.
      * <p>
      * Examples:
      * <pre>{@code
@@ -78,7 +58,7 @@ public class WildcardOptionsUtil {
 
     /**
      * Generates a concrete configuration key by replacing the wildcard placeholder with a specific value.
-     * You should always check the presence of the wildcard via the {@link #isKcWildcardOption(String)}.
+     * You should always check the presence of the wildcard via the {@link #isWildcardOption(String)}.
      * <p>
      * Examples:
      * <pre>{@code
--- a/quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
@@ -1,0 +1,121 @@
+package org.keycloak.config;
+
+/**
+ * Utility class for working with configuration options that use wildcards.
+ * <p>
+ * Wildcard options are configuration keys that contain a variable segment enclosed between {@link #WILDCARD_START} and {@link #WILDCARD_END} characters.
+ * <p>
+ * Wildcard options in Keycloak always <strong>end</strong> with the variable segment.
+ */
+public class WildcardOptionsUtil {
+
+    /**
+     * Marker indicating the start of a wildcard segment in a configuration key.
+     */
+    public static final String WILDCARD_START = "<";
+
+    /**
+     * Marker indicating the end of a wildcard segment in a configuration key.
+     */
+    public static final String WILDCARD_END = ">";
+
+        /**
+     * Determines whether the given configuration key represents a wildcard option.
+     * <p>
+     * Examples:
+     * <pre>{@code
+     * isWildcardOption("tracing-header-<header>")      → "true"
+     * isWildcardOption("tracing-header-<headxxx")      → "false"
+     * isWildcardOption("tracing-header-headxxx>")      → "false"
+     * isWildcardOption("db-kind-<datasource>")         → "true"
+     * isWildcardOption("http-port")                    → "false"
+     * isWildcardOption("quarkus.<sth>.end")            → "false"
+     * }</pre>
+     *
+     * @param key the configuration key to check
+     * @return {@code true} if the key represents a Keycloak wildcard option
+     */
+    public static boolean isWildcardOption(String key) {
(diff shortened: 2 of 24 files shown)


docs/guides/operator/advanced-configuration.adoc
```diff
@@ -406,6 +406,8 @@ spec:
     samplerRatio: 0.01                        # default '1'
     resourceAttributes:
       some.attribute: something
+    headers:
+      Authorization: Bearer ${OTEL_TOKEN}
```
mabartos: It'd be good to navigate users on how to provide the token from secrets into this header. 

@shawkins Do you know what would be the best way to provide a token from the secret in the CR?

shawkins: What secret, and is this is something resolved on the operator side?

mabartos: When exporting traces, usually, it's necessary to include a token/API-key to properly export traces to the tracing backend. 

So, we need to provide some guidance to our users, how they can provide the token on the Operator level - that every KC pod might use the token and be able to export traces with the token.

It means, the natural way is to create a secret and then propagate it to KC pods - so, probably via ENV vars. But it's usually only for a single option. Here, it would be good to have the ability to use the `tracing.headers` map in the CR, where you can reference the token from a secret.

Having something like:
```yaml
headers:
      Authorization: Bearer ${OTEL_TOKEN}
      My-Org-Id: {MY_ORG_API_KEY}
      ...
```
and having `OTEL_TOKEN` and `MY_ORG_API_KEY` obtained from a secrets. But AFAIK, the env var should be created on the Operator side, in order to being able to reference it like that, right? And it would cause some additional security flaws, right? 

It'd be probably better to use `ValueOrSecret` type here for headers.

And all these values would be then handed to KC pods/containers as required key1=val,key2=val2,...

For that purpose, it'd be good to mark the `tracing-headers` as sensitive on the Keycloak level.

WDYT? @shawkins @vmuzikar 

_Thinking out loud and wondering, if you have any other opinion, or better approach for this._

shawkins: > It'd be probably better to use ValueOrSecret type here for headers.

To make that work as a single server side option given:

```
headers:
  - name: Authorization
    secret:
      name: my-secret
      key: my-key
```

You would then have to form a setting like:

KC_TRACING_HEADERS=Authorization: ${TRACING_HEADER_SECRET_X}

Where TRACING_HEADER_SECRET_X is a generated environment variable name that should be unique. Then the add TRACING_HEADER_SECRET_X to the handling of the env list.

The user would be responsible for prepending "Bearer " in the secret variable. 

Alterntively you could stick with what you are showing, and require the user to create the necessary environment variables:

```
...
    headers:
      Authorization: Bearer ${OTEL_TOKEN}
      My-Org-Id: ${MY_ORG_API_KEY}
...
  env:
    - name: OTEL_TOKEN
        secret: my-secret
        key: my-key
```

We may also want to suggest a naming convent / prefix for users to use.

vmuzikar (changes requested):
As discussed offline, let's make this wildcard option: `--tracing-header-<header>`, e.g. `--tracing-header-Authorization`.

For the Operator, we will rely on `additionalOptions` field.

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
```diff
@@ -0,0 +1,121 @@
...
+     * getWildcardName(TracingOptions.TRACING_HEADER, "tracing-header-Authorization") → "Authorization"
+     * getWildcardName(DatabaseOptions.DB_ENABLE, "db-enable-my-store") → "my-store"
+     * getWildcardName(DatabaseOptions.DB_ENABLE, "kc.db-enable-my-store") → "my-store"
+     * }</pre>
+     *
+     * @param option   the option containing a wildcard key
+     * @param namedKey the fully qualified (resolved) configuration key
+     * @return the part of {@code namedKey} that replaces the wildcard in {@code option.getKey()}, otherwise {@code null}
+     */
+    public static String getWildcardName(Option<?> option, String namedKey) {
```
vmuzikar: Shouldn't this be rather `getWildcardValue`? It's a value of the `<wildcard>`, not name (which may imply a key), no?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/TracingPropertyMappers.java
```diff
@@ -127,6 +150,28 @@ private static void validateRatio(String value) {
...
+    private static String transformTracingHeaders(String value, ConfigSourceInterceptorContext ctx) {
+        if (StringUtil.isNotBlank(value)) {
+            log.debug("HTTP headers for the tracing exporter are overridden by the '%s' parent property".formatted(TRACING_HEADERS.getKey()));
+            return value;
+        }
+
+        Map<String, String> headers = new HashMap<>();
+        String prefix = getWildcardPrefix(TRACING_HEADER.getKey());
+
+        Configuration.getPropertyNames().forEach(key -> {
```
vmuzikar, line 162: Iterating through all options is not ideal. Though I don't have a better idea to do this. Maybe @shawkins will.

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
```diff
@@ -0,0 +1,121 @@
...
+     * isExternalWildcardOption("tracing-header-headxxx>")      → "false"
+     * isExternalWildcardOption("db-kind-<datasource>")         → "true"
+     * isExternalWildcardOption("http-port")                    → "false"
+     * isExternalWildcardOption("quarkus.<sth>.end")            → "true"
+     * }</pre>
+     *
+     * @param key the configuration key to check
+     * @return {@code true} if the key represents an external wildcard option
+     */
+    public static boolean isExternalWildcardOption(String key) {
```
vmuzikar: I'd be maybe more explicit in the naming here. Maybe `isKcOrQuarkusWildcardOption`? "External" would suggest we're looking for Quarkus options only.

vmuzikar (commented):
@mabartos Thank you for the PR!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/TracingPropertyMappers.java
shawkins: Unforuntately I don't think there's a better way to do this currently. As long as this is only getting resolved when the tracing-headers quarkus property is requested, it shouldn't be too bad. @mabartos how many time do you see this method hit? Is it erroneously being validated for example.

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
```diff
@@ -0,0 +1,121 @@
+package org.keycloak.config;
+
+/**
+ * Utility class for working with configuration options that use wildcards.
+ * <p>
+ * Wildcard options are configuration keys that contain a variable segment enclosed between {@link #WILDCARD_START} and {@link #WILDCARD_END} characters.
+ * <p>
+ * Wildcard options in Keycloak always <strong>end</strong> with the variable segment.
+ */
+public class WildcardOptionsUtil {
```
shawkins, line 10: If a method is used exclusively internally to WildcardPropertyMapper, could we start with it there? As much as possible I'd like to encapsulate the wildcard logic.

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
mabartos: Works to me. It's up to our consensus.

mabartos: Updated

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
mabartos: Good point. I updated the code and now using `isWildcardOption()` and `isKcWildcardOption()` that should be sufficient for the required use cases.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/TracingPropertyMappers.java
mabartos: I double-checked it and I can confirm it's evaluated only once per startup - when the `quarkus.otel.exporter.otlp.traces.headers` is requested (on the Quarkus side).

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
mabartos: @shawkins Unfortunatelly, it's not used only internally in the WildcardPropertyMapper. I'd like to have some unified approach on how to do the basic ops with wildcard options. So, the util class seems fine to me as it'd be needed in the config-api module. I'd like to also use some parts in the https://github.com/keycloak/keycloak/pull/43542

shawkins: > @shawkins Unfortunatelly, it's not used only internally in the WildcardPropertyMapper.

It just looked like some of the methods weren't used in other places yet. If there are other plans to expand the usage a util class seems fine, but I'd still vote for keeping it to the minimum number of methods.

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
```diff
@@ -0,0 +1,121 @@
...
+     * Determines whether the given configuration key represents a wildcard option.
+     * <p>
+     * Examples:
+     * <pre>{@code
+     * isKcWildcardOption("tracing-header-<header>")      → "true"
+     * isKcWildcardOption("tracing-header-<headxxx")      → "false"
+     * isKcWildcardOption("tracing-header-headxxx>")      → "false"
+     * isKcWildcardOption("db-kind-<datasource>")         → "true"
+     * isKcWildcardOption("http-port")                    → "false"
+     * isKcWildcardOption("quarkus.<sth>.end")            → "false"
```
shawkins: Are you sure we can make this assumption in general?

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
mabartos: Ok, works for me. I'll try to minimize util methods

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
mabartos: @shawkins Why not? We've said the proper KC wildcard option needs to have the wildcard (variable segment) at the end. And the described method shows how is it evaluated as it's only for cases without the resolved name; only for properties with the < and > chars.

Do you have any concern?

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
mabartos: @shawkins I tried to find some reasonable balance on what will be part of the wildcard options util class and I think all these methods belongs together as it's suggested that you first ask whether it's a wildcard option and then do some operations on the wildcard option. 

Is it ok for you to leave it as is, or do you suggest any better approach?

quarkus/config-api/src/main/java/org/keycloak/config/WildcardOptionsUtil.java
shawkins: My concern was specifically with "quarkus.<sth>.end" as that seems to be mixing both qualified and unqualified keys - if we expect quarkus., shouldn't all of them something like kc. or smallrye.?

Then are we sure that quarkus options cannot end with a wildcard, and will always have ".end"?

mabartos: I kept only the `isWildcardOption` method that only check if the property contains the variable segment (< and >)

@shawkins WDYT?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/TracingPropertyMappers.java
shawkins: The only other approach I could imagine would be to add an aggregation routine to validation. That is for propertymappers marked as aggregates, call them with each key value pair during the validation routine, and then some final compute / save method. To have that survive the cutover to the quarkus classloader would probably mean storing it as a system property. 

That would keep the iterations down to a single time, but adds a good bit of complexity. So we'll hold off on doing that until there seems to be an actual performance hit.

mabartos:
@shawkins @vmuzikar Should be ready for review.

shawkins (approved):
LGTM, thank you @mabartos

mabartos:
@vmuzikar Do you wanna review it as well? Thanks

vmuzikar (approved):
Looks good, thanks.

PR #43126, opened 2025-10-01
Filter invalid resources and scopes when processing entries from the cache

pedroigor:
Closes #42907

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/authz/EntitlementAPITest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/authz/EntitlementAPITest.java
@@ -37,6 +37,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/authz/EntitlementAPITest.java
```diff
@@ -2649,6 +2651,54 @@ public void testSameResultRegardlessOPermissionParameterValue() throws Exception
...
+                    authorization.resources().resource(test.get(0).getId()).remove();
+                }
+            }
+        }).start();
+
+        AuthzClient authzClient = getAuthzClient(AUTHZ_CLIENT_CONFIG);
+
+        for (int i = 0; i < 3; i++) {
+            // Thread that will be requesting permissions against the resource being created and deleted
+            new Thread(() -> {
```
pruivo, line 2691: I'm not a big fan of starting multiple threads to reproduce a concurrency issue. 
Wouldn't it be simpler to inject an invalid ID in the cached object? 
Like sending a server task to do `cache.get(cacheKey, queryType).getResources().add("invalid-id");`

pedroigor: I would like to make sure the exact scenario is reproducible and make sure the authorization request, from a client perspective, works as expected under load.

What do you suggest as an alternative? Model test?

pruivo: My initial thought was to do the test in this class. I'm not sure if something like this is possible
```java
        // populate cache
        AuthzClient authzClient = getAuthzClient(AUTHZ_CLIENT_CONFIG);
        authzClient.authorization().authorize();

        testingClient.server().run(session -> {
            var cache = session.getProvider(InfinispanConnectionProvider.class).getCache("??");
            CachedPolicy cachedEntity = cache.get("??");
            // add an invalid id to simulate a resource being removed and the cache has not been invalidated yet
            cachedEntity.getResourcesIds(session, null).add("invalid-id");
            // repeat for other entities?
        });

        // no Exception here!
        authzClient.authorization().authorize();
```

pedroigor: I see. Let me try it out and see how it goes.

pruivo (dismissed):

pruivo (approved):

ahus1 (approved):
Approving based on @pruivo's review

PR #43131, opened 2025-10-01
Prevent users configuring max-count=-1 for caches with a default upper-bound

### LINKED ISSUE #33146: Document the ability to allow unbounded caches with max-count=-1
### Description

Currently it's not possible for users to define an unbounded max-count for caches as passing `--cache-embedded-$\{CACHE_NAME}-max-count=-1` will result in an error due to #33068. Once this limitation has been resolved we should update `caching.adoc` to make it clear that unbounded cache can be explicitly configured using `max-count=-1`.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #33146

### DIFF
--- a/docs/guides/server/caching.adoc
+++ b/docs/guides/server/caching.adoc
@@ -162,7 +162,7 @@ apply the upper bound to. For example, to apply an upper-bound of `1000` to the 
 `--cache-embedded-offline-sessions-max-count=1000`. An upper bound can not be defined on the following caches:
 `actionToken`, `authenticationSessions`, `loginFailures`, `work`.
 
-Setting a maximum cache size for `sessions`, `clientSessions`, `offlineSessions` and `offlineClientSessions` is not supported when volatile sessions are enabled.
+Setting a maximum cache size for `sessions`, `clientSessions` is not supported when volatile sessions are enabled.
 
 === Specify your own cache configuration file
 
--- a/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/DefaultCacheEmbeddedConfigProviderFactory.java
+++ b/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/DefaultCacheEmbeddedConfigProviderFactory.java
@@ -17,6 +17,13 @@
 
 package org.keycloak.spi.infinispan.impl.embedded;
 
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.ALL_CACHES_NAME;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.CLUSTERED_CACHE_NUM_OWNERS;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.CLUSTERED_MAX_COUNT_CACHES;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.LOCAL_CACHE_NAMES;
+import static org.keycloak.connections.infinispan.InfinispanConnectionProvider.LOCAL_MAX_COUNT_CACHES;
+import static org.keycloak.spi.infinispan.impl.embedded.JGroupsConfigurator.createJGroupsProperties;
+
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.nio.file.Paths;
@@ -25,7 +32,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
-import io.micrometer.core.instrument.Metrics;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.cache.StatisticsConfigurationBuilder;
 import org.infinispan.configuration.global.ShutdownHookBehavior;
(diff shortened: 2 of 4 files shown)


pruivo (dismissed):

docs/guides/server/caching.adoc
```diff
@@ -159,8 +159,10 @@ In order to reduce memory usage, it's possible to place an upper bound on the nu
 cache. To specify an upper bound of on a cache, you must provide the following command line argument
 `--cache-embedded-$\{CACHE_NAME}-max-count=`, with `$\{CACHE_NAME}` replaced with the name of the cache you would like to
 apply the upper bound to. For example, to apply an upper-bound of `1000` to the `offlineSessions` cache you would configure
-`--cache-embedded-offline-sessions-max-count=1000`. An upper bound can not be defined on the following caches:
-`actionToken`, `authenticationSessions`, `loginFailures`, `work`.
+`--cache-embedded-offline-sessions-max-count=1000`. Conversely, to disable the upper-bound you would configure a negative value
+`--cache-embedded-offline-sessions-max-count=-1`.
```
ahus1: With @pruivo's recent changes that offline session even when they are not persistent use a bounded cache, I don't think that this is a good example any more. So those caches should always be set up in abounded manner. 

Thinking about it, I now can no longer see a cache that you would configure sometimes bounded, and sometimes unbounded. So maybe this is no longer needed?

ahus1 (changes requested):

docs/guides/server/caching.adoc
pruivo: @ahus1 With small deployments, like we do in our weekly runs, it may make sense to have unbounded caches as everything fits in memory. It saves the cost of keeping track of the used entries for the eviction algorithm. 
If you don't want to explicitly document it, it is fine by me.

ryanemerson: I have updated the docs to use the `users` cache as an example instead, I have also tried to update the docs to reflect the rules currently enforced by `DefaultCacheEmbeddedConfigProviderFactory`.

ahus1: I think it is dangerous if people start with unbounded caches and then miss to make them bounded later when the number of entities grows or usage patterns change and they overshoot on memory usage. It is also not clear to admins how many entries a user takes in the cache, and it can actually depend on the usage and how they are looked up. 

So I'd propose the following behavior: 

"All cache that are currently set by default with an upper bound must not be unbounded by setting the parameter, as this is a sharp edge admins will find difficult to handle."

The docs and the implementation should reflect that.

pruivo: Well, we don't have to document that -1 creates an unbounded cache if it makes everybody happy.

ryanemerson: > "All cache that are currently set by default with an upper bound must not be unbounded by setting the parameter, as this is a sharp edge admins will find difficult to handle."

I have pushed a commit to to enforce ^. At this stage the caches which can actually be unbounded is so small I don't think we should document that this is possible, especially considering we automatically apply this configuration when `persistent-user-sessions` is disabled.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/DefaultCacheEmbeddedConfigProviderFactory.java
```diff
@@ -221,7 +226,8 @@ private static ConfigurationBuilderHolder parseConfiguration(Config.Scope keyclo
     private static void singleSiteConfiguration(Config.Scope config, ConfigurationBuilderHolder holder, KeycloakSessionFactory factory) {
         logger.debug("Configuring Infinispan for single-site deployment");
         CacheConfigurator.checkCachesExist(holder, Arrays.stream(ALL_CACHES_NAME));
-        CacheConfigurator.configureCacheMaxCount(config, holder, Arrays.stream(CLUSTERED_MAX_COUNT_CACHES));
+        CacheConfigurator.configureCacheMaxCount(config, holder, Stream.of(USER_SESSION_CACHE_NAME, CLIENT_SESSION_CACHE_NAME), !Profile.isFeatureEnabled(Profile.Feature.PERSISTENT_USER_SESSIONS));
+        CacheConfigurator.configureCacheMaxCount(config, holder, Stream.of(OFFLINE_USER_SESSION_CACHE_NAME, OFFLINE_CLIENT_SESSION_CACHE_NAME), false);
```
pruivo: This seems unnecessary.  You can apply the logic to `CacheConfigurator#configureSessionsCachesForVolatileSessions` and `CacheConfigurator#configureSessionsCachesForPersistentSessions`

ryanemerson: Makes sense, updated!

ahus1 (approved):
Thank you, looks good to me. Not merging yet to give Pedro the opportunity to review it as well.

ryanemerson:
Commit message and issue updated to reflect the current state of the PR.

pruivo (approved):

PR #43132, opened 2025-10-01
Divide logging guide to sub-guides for every log handler

### LINKED ISSUE #43125: Divide logging guide to sub-guides for every log handler
### Description

Right now, the Logging configuration guide is very extensive, and sometimes it's hard to navigate through it. We should divide it into sub-guides for every log handler, as users might redirect there. 

### Discussion

_No response_

### Motivation

Improve UX about logging configuration

### Details

_No response_

mabartos:
Closes #43125

IMO, the format of the Logging guide is much readable as we separated the log handlers into the sub-guides. 
I've also added some better introduction, and how to enable log handlers. IMHO, it's much better now. 

Old:

https://github.com/user-attachments/assets/8c5b1886-90ac-4782-914a-7ddb3fe7faaa

New:

https://github.com/user-attachments/assets/af6f998e-86a5-4417-a1a3-c46da1113f62

@keycloak/cloud-native Could you please check it? Thanks!

### DIFF
--- a/docs/guides/server/logging.adoc
+++ b/docs/guides/server/logging.adoc
@@ -8,20 +8,40 @@
 title="Configuring logging"
 summary="Configure logging for {project_name}.">
 
-{project_name} uses the JBoss Logging framework.
-The following is a high-level overview for the available log handlers with the common parent log handler `root`:
+Logging is a key mechanism for understanding and operating {project_name}.
+It should help you to monitor {project_name}'s health, debug issues, and maintain an audit trail of important events.
 
-<@opts.expectedValues option="log"/>
-
-== Logging configuration
-
-Logging is done on a per-category basis in {project_name}.
 You can configure logging for the root log level or for more specific categories such as `org.hibernate` or `org.keycloak`.
 It is also possible to tailor log levels for each particular log handler.
 
-This {section} describes how to configure logging.
+In {project_name}, logging goes beyond setting log levels — you can direct output to different handlers, use asynchronous logging for performance, capture HTTP access logs, and more.
+These features make it possible to adapt logging to your operational needs and integrate with observability platforms.
 
-=== Log levels
+{project_name} uses the JBoss Logging framework under the hood.
+
+== Available log handlers
+
+[cols="^1,^1,^1", frame=none, grid=all]
+|===
+| 🖥️ *Console* +
+Logs to standard output
+| 📄 *File* +
+Persists logs to disk
+| 🖧 *Syslog* +
+Sends logs to Syslog server
+|===
+
+To enable log handlers, enter the following command:
+
+<@kc.start parameters="--log=\"<handler1>,<handler2>\""/>
+
+For more details on how to configure specific log handlers, see:
+
+* <@links.server id="logging-console"/>
+* <@links.server id="logging-file"/>
+* <@links.server id="logging-syslog"/>
+
+== Log levels
 
 The following table defines the available log levels.
 
@@ -68,32 +88,6 @@ This example sets the following log levels:
 * The hibernate log level in general is set to debug.
 * To keep SQL abstract syntax trees from creating verbose log output, the specific subcategory `org.hibernate.hql.internal.ast` is set to info. As a result, the SQL abstract syntax trees are omitted instead of appearing at the `debug` level.
 
-=== Adding context for log messages
-
-:tech_feature_name: Log messages with Mapped Diagnostic Context (MDC)
-:tech_feature_id: log-mdc
-
-[NOTE]
-====
-{tech_feature_name} is
-*Preview*
-and is not fully supported. This feature is disabled by default.
-====
-
-You can enable additional context information for each log line like the current realm and client that is executing the request.
-
-Use the option `log-mdc-enabled` to enable it.
-
-.Example configuration
-<@kc.start parameters="--features=log-mdc --log-mdc-enabled=true"/>
-
-.Example output
-----
-2025-06-20 14:13:01,772 {kc.clientId=security-admin-console, kc.realmName=master} INFO ...
-----
-
-Specify which keys to be added by setting the configuration option `log-mdc-keys`.
-
 === Configuring levels as individual options
 When configuring category-specific log levels, you can also set the log levels as individual `log-level-<category>` options instead of using the `log-level` option for that.
 This is useful when you want to set the log levels for selected categories without overwriting the previously set `log-level` option.
@@ -112,17 +106,6 @@ Bear in mind that when using the environmental variables, the category name must
 When using other config sources, the category name must be specified "as is", for example:
 
 <@kc.start parameters="--log-level=\"INFO,org.hibernate:debug\" --log-level-org.keycloak=trace"/>
-
-== Enabling log handlers
-To enable log handlers, enter the following command:
-
-<@kc.start parameters="--log=\"<handler1>,<handler2>\""/>
-
-The available handlers are:
-
-<@opts.expectedValues option="log"/>
-
-The more specific handler configuration mentioned below will only take effect when the handler is added to this comma-separated list.
 
 === Specify log level for each handler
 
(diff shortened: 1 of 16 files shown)


shawkins (dismissed):
LGTM, makes sense to break it down more.

vmuzikar (changes requested):
@mabartos Thank you for the PR.

While I appreciate a clean up of the Logging guide will help, I'm afraid this does not work well for the use case when the guides are displayed as a single one. This is the case for downstream. Here it's displayed as 4 guides, all on the same level instead of the 3 being children of the main Logging guide:
Image
I'd suggest having it as a single guide, attach the sub-guides at the end and leverage anchor links to like to these sub-guides.

Additionally, I do not think we should use emojis in a tech documentation. We risk it won't be displayed correctly (see Syslogging on the screenshot) and probably would not be too appropriate downstream (CC @andymunro).

mabartos:
@vmuzikar Thanks for your input. Yes, the downstream docs are affected. I was wondering if it'd be possible to remove Unicode chars in the downstream docs and have the emojis available at least for upstream? Like with some property? (cc: @ahus1)

 - Removed emojis
 - Changed offset for these sub-guides, so the docs for downstream should look like this:
 
image

@vmuzikar Is it ok for you?

docs/guides/server/logging.adoc
```diff
@@ -295,267 +304,14 @@ You can use regular expressions to exclude them, such as:
...
-
-=== Console
-<@opts.includeOptions includedOptions="log-console-*"/>
-
-=== File
-<@opts.includeOptions includedOptions="log-file log-file-*"/>
-
-=== Syslog
-<@opts.includeOptions includedOptions="log-syslog-*"/>
+<@opts.printRelevantOptions includedOptions="log log-*" excludedOptions="log-console-* log-file log-file-* log-syslog-* log-mdc*">
```
vmuzikar, line 307: I just wonder, should we put some link or hint here that there are more options under the various log handlers? Folks tend to skip chapters and may miss there are log handlers with many more options.

vmuzikar (commented):
Thanks for the update! Overall looks good, added one minor comment.

Additionally, @ahus1 @andymunro do you want to review as well?

docs/guides/server/logging.adoc
mabartos: @vmuzikar That's a good point. I think it makes sense to include all config options related to the logging here.

mabartos: @vmuzikar Updated. IMHO, the duplication of options shown in the general logging guide and in the specific log handler guide for the downstream is not a problem. Users/customers may have link only for the specific log handler and the relevant options are handy to have there. 

Is it ok for you?

docs/guides/server/logging/file.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+A log file named `keycloak.log` is created inside the `data/log` directory of your {project_name} installation.
+
+== Configuring the location and name of the log file
+
+To change where the log file is created and the file name, perform these steps:
+
+. Create a writable directory to store the log file.
++
+If the directory is not writable, {project_name} will start correctly, but it will issue an error and no log file will be created.
```
andymunro:
```suggestion
If the directory is not writable, {project_name} starts correctly, but it issues an error and no log file is created.
```

andymunro: Generally, use present tense, which sounds more active.

docs/guides/server/logging/syslog.adoc
```diff
@@ -0,0 +1,146 @@
...
+<#import "/templates/options.adoc" as opts>
+
+<@tmpl.guide
+title="Syslog logging"
+summary="Send logs to a remote Syslog server"
+tileVisible="false"
+levelOffset=2 >
+
+{project_name} provides the ability to send logs to a remote Syslog server.
+It utilizes the protocol defined in https://datatracker.ietf.org/doc/html/rfc5424[RFC 5424].
```
andymunro:
```suggestion
It uses the protocol defined in https://datatracker.ietf.org/doc/html/rfc5424[RFC 5424].
```

andymunro: We prefer "uses" to "utilizes" in the documentation.

docs/guides/server/logging/syslog.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+When the Syslog handler is enabled, the host is using `localhost` as host value.
+The Default port is `514`.
+
+== Configuring the log level
+Log level for Syslog log handler can be specified by `--log-syslog-level` property as follows:
+
+<@kc.start parameters="--log-syslog-level=warn"/>
+
+For more information, see the section <<Specify log level for each handler>> above.
```
andymunro:
```suggestion
For more information, see <<Specify log level for each handler>>.
```

andymunro: Generally, we just need the link, not "the section" or "above' or below.

docs/guides/server/logging/syslog.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+<@kc.start parameters="--log-syslog-output=json"/>
+
+.Example Log Message
+[source, bash]
+----
+2024-04-05T12:32:20.616+02:00 host keycloak 2788276 io.quarkus - {"timestamp":"2024-04-05T12:32:20.616208533+02:00","sequence":9948,"loggerClassName":"org.jboss.logging.Logger","loggerName":"io.quarkus","level":"INFO","message":"Profile prod activated. ","threadName":"main","threadId":1,"mdc":{},"ndc":"","hostName":"host","processName":"QuarkusEntryPoint","processId":2788276}
+----
+
+When using JSON output, colors are disabled and the format settings set by `--log-syslog-format` will not apply.
```
andymunro:
```suggestion
When using JSON output, colors are disabled and the format settings set by `--log-syslog-format` does not apply.
```

docs/guides/server/logging/syslog.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+<@kc.start parameters="--log-syslog-output=default"/>
+
+.Example Log Message
+[source, bash]
+----
+2024-04-05T12:31:38.473+02:00 host keycloak 2787568 io.quarkus - 2024-04-05 12:31:38,473 INFO  [io.quarkus] (main) Profile prod activated.
+----
+
+As you can see, the timestamp is present twice, so you can amend it correspondingly via the `--log-syslog-format` property.
```
andymunro:
```suggestion
As you can see, the timestamp is present twice, so you can amend it correspondingly by the `--log-syslog-format` property.
```

andymunro: "via" is latin.  The English equivalent is "by."

docs/guides/server/logging/console.adoc
```diff
@@ -0,0 +1,106 @@
...
+<#import "/templates/options.adoc" as opts>
+
+<@tmpl.guide
+title="Console logging"
+summary="Print logs to a console output"
+tileVisible="false"
+levelOffset=2 >
+
+The console log handler is enabled by default, providing unstructured log messages for the console.
+The more specific console handler configuration mentioned below will only take effect when the console logging is enabled (console logging is enabled by default).
```
andymunro:
```suggestion
As shown in the following section, the more specific console handler configuration takes effect only when the console logging is enabled. Console logging is enabled by default.
```

andymunro: avoid using "above" or "below"

docs/guides/server/logging/console.adoc
```diff
@@ -0,0 +1,106 @@
...
+Colored console log output for unstructured logs is disabled by default. Colors may improve readability, but they can cause problems when shipping logs to external log aggregation systems. To enable or disable color-coded console log output, enter following command:
+
+<@kc.start parameters="--log-console-color=<false|true>"/>
+
+== Configuring the console log level
+Log level for console log handler can be specified by `--log-console-level` property as follows:
+
+<@kc.start parameters="--log-console-level=warn"/>
+
+For more information, see the section <<Specify log level for each handler>> above.
```
andymunro:
```suggestion
For more information, see <<Specify log level for each handler>>.
```

andymunro (changes requested):
Very well written, @mabartos. I have a few suggestions.

mabartos:
@andymunro Thanks for the additional suggestions! Some of the text is only divided to separate guides, but with these suggestions, we make the docs better! :) 

@vmuzikar Do you want to review it again? Thanks!

vmuzikar (approved):
@mabartos Sorry for the late response. LGTM, let's get this merged already! :D

PR #43138, opened 2025-10-01
Add CI tests for Azure SQL Database

### LINKED ISSUE #42986: Add CI tests for Azure SQL Database
### Description

This is a follow-up from https://github.com/keycloak/keycloak/issues/42743.

Pepo48:
Closes: #42986

I reused some of the bits from Aurora. The Aurora workflow needs to be run with these changes applied prior to merging this, just to be sure that nothing is broken. The reason why this is not using Ansible even for the Azure VM provisioning is that I could not make the Azure Ansible collection to work. I experienced `Failed to import the required Python library (ansible[azure] (azure >= 2.0.0)`, which seems to be a [known issue](https://github.com/ansible-collections/azure/issues/1463), where some submodules of the collection rely on deprecated parts. Therefore, I opted for scripts using native Azure CLI.

For now the Azure workflow isn't skipped and runs with every push to the branch (for testing purposes). I will adjust it and mimic the behaviour of Aurora once reviewed.

We also need to create a couple of GH Actions secrets in order to make the workflow authenticate successfully (as described in the readme).

Here are the most recent results:
https://github.com/Pepo48/keycloak/actions/runs/18176701245/job/51744192469


mabartos:
@Pepo48 Seems the login does not work in the CI - https://github.com/keycloak/keycloak/actions/runs/18494292935/job/52694874282?pr=43138

Pepo48:
@mabartos it's expected at this point, for the time being you need to look at the run from my fork, the latest one with all of the recent changes above is available [here](https://github.com/Pepo48/keycloak/actions/runs/18497045600/job/52703780846).

.github/workflows/ci.yml
```diff
@@ -581,6 +569,140 @@ jobs:
...
+          AZURE_CLUSTER_NAME="gh-action-$(git rev-parse --short HEAD)-${{ github.run_id }}-${{ github.run_attempt }}"
+          echo "azure-cluster-name=${AZURE_CLUSTER_NAME}" >> $GITHUB_OUTPUT
+          echo "region=${AZURE_REGION}" >> $GITHUB_OUTPUT
+          echo "azure-subscription=${AZURE_SUBSCRIPTION}" >> $GITHUB_OUTPUT
+
+          PROPS=''
+          echo "maven_properties=${PROPS}" >> $GITHUB_OUTPUT
+
+      - name: Login to Azure
+        uses: azure/login@v1
```
vmuzikar: We should use a hardcoded hash instead of floating v1.

.github/workflows/ci.yml
```diff
@@ -581,6 +569,140 @@ jobs:
...
+        run: |
+          AZURE_REGION=westeurope
+          echo "AZURE Region: ${AZURE_REGION}"
+
+          AZURE_SUBSCRIPTION=${{ secrets.AZURE_SUBSCRIPTION_ID }}
+
+          AZURE_CLUSTER_NAME="gh-action-$(git rev-parse --short HEAD)-${{ github.run_id }}-${{ github.run_attempt }}"
+          echo "azure-cluster-name=${AZURE_CLUSTER_NAME}" >> $GITHUB_OUTPUT
+          echo "region=${AZURE_REGION}" >> $GITHUB_OUTPUT
+          echo "azure-subscription=${AZURE_SUBSCRIPTION}" >> $GITHUB_OUTPUT
```
vmuzikar: Should we use the secret directly in steps instead of passing it as a step output? It feels slightly more secure.

.github/workflows/ci.yml
```diff
@@ -581,6 +569,140 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
 
+  azure-integration-tests:
+    name: AzureDB IT
+    needs: conditional
+    if: needs.conditional.outputs.ci == 'true'
```
vmuzikar: Just wanted to highlight this:
> For now the Azure workflow isn't skipped and runs with every push to the branch (for testing purposes). I will adjust it and mimic the behaviour of Aurora once reviewed.

I think we can go ahead and adjust it now so the PR can be merged.

.github/workflows/ci.yml
```diff
@@ -581,6 +569,140 @@ jobs:
...
+          rsync -a $results/* ../../../
+
+      - uses: ./.github/actions/upload-flaky-tests
+        name: Upload flaky tests
+        env:
+          GH_TOKEN: ${{ github.token }}
+        with:
+          job-name: AzureDB IT
+
+      - name: Delete all Azure resources (resource group)
```
vmuzikar, line 737: I believe we're not deleting the Azure DB, just the VM here? Seems like we're not executing `azure_delete.sh` anywhere.

.github/scripts/azure/sql/azure_common.sh
```diff
@@ -0,0 +1,26 @@
...
+}
+
+requiredEnv AZURE_NAME AZURE_REGION AZURE_ADMIN_PASSWORD AZURE_DB_PASSWORD
+
+SCRIPT_DIR=${SCRIPT_DIR:-$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )}
+export AZURE_RG=${AZURE_RG:-"${AZURE_NAME}-rg"}
+export AZURE_ADMIN_USER=${AZURE_ADMIN_USER:-"keycloakadmin"}
+export AZURE_ADMIN_PASSWORD=${AZURE_ADMIN_PASSWORD}
+export AZURE_DB=${AZURE_DB:-"keycloakdb"}
+export AZURE_DB_USER=${AZURE_DB_USER:-"keycloak_user"}
```
vmuzikar: We probably shouldn't have the defaults for usernames either.

vmuzikar (changes requested):
@Pepo48 Sorry for the late reply and thank you for the PR. It's a very nice job!

Added a couple of comments. On top of that we'll need the repo secrets created before merging. Let me know when that's done.

@ahus1 Do you want to give it a look too, as there are changes affecting Aurora workflows as well?

.github/scripts/azure/sql/azure_common.sh
Pepo48: With the consolidation of required repository secrets I reintroduced these defaults. Right now, the only required secret is AZURE_CREDENTIALS, which is used for authentication. For the usernames I used the defaults, the passwords are auto-generated on the fly. There's still possibility to override the behaviour by setting repository vars and secrets respectively.

Alternatively, usernames could be random as well, but I think it's not necessary. SUBSCRIPTION_ID is parsed from AZURE_CREDENTIALS, so it's secret is not needed either.

Results here: https://github.com/Pepo48/keycloak/actions/runs/18987564770/job/54234497195

PTAL. Thanks!

.github/actions/azure-create-database/action.yml
```diff
@@ -0,0 +1,118 @@
...
+    description: 'DB user principal'
+    value: ${{ steps.create.outputs.username }}
+  subscription:
+    description: 'Subscription ID'
+    value: ${{ steps.parse-creds.outputs.subscription }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Parse Azure credentials
```
vmuzikar: Would it be possible to run this step in the CI workflow, not in the action here?

The main reason is I'd like to focus the usage of the secret with creds to a single place, passing it around as little as possible to avoid accidental leaks and to track it more easily where is it used.

Second reason is, parsing the subscription ID is generic, not specific to creating the DB, so would made more sense in a generic place like `ci.yml`, IMHO.

vmuzikar (commented):
@Pepo48 Thank you for the updates!

vmuzikar (approved):
@Pepo48 Thank you for the update. I assume you have tested it with the last commit somewhere.

Approving already. LGTM, overall very nice work, thank you.

@ahus1 I'd like to ask for your review, particularly of the Aurora related changes. Thank you.

vmuzikar:
DCO failed but the sign-off looks fine on the individual commits. I'll fix it when squashing the commits on merge.

Pepo48:
> @Pepo48 Thank you for the update. I assume you have tested it with the last commit somewhere.

https://github.com/Pepo48/keycloak/actions/runs/19344699962/job/55341941039

ahus1 (approved):
Saw a green run on AuroraDB for this branch on my private repo once I added the AWS key. Thanks for the PR! Just now waiting for this to become green.

PR #43170, opened 2025-10-02
Restarting an user session broken for persistent sessions

pruivo:
Fixes #43161

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderTest.java
@@ -526,6 +526,7 @@ public class UserSessionProviderTest extends AbstractTestRealmKeycloakTest {
                 kcSession.getContext().setRealm(r);
                 r.setSsoSessionMaxLifespanRememberMe(r.getSsoSessionMaxLifespan() * 4);
                 r.setSsoSessionIdleTimeoutRememberMe(r.getSsoSessionIdleTimeout() * 4);
+                r.setRememberMe(true);
             });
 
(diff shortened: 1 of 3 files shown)


testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderTest.java
```diff
@@ -526,6 +526,7 @@ public  void testRemoveUserSessionsByExpiredRememberMe(KeycloakSession session)
                 kcSession.getContext().setRealm(r);
                 r.setSsoSessionMaxLifespanRememberMe(r.getSsoSessionMaxLifespan() * 4);
                 r.setSsoSessionIdleTimeoutRememberMe(r.getSsoSessionIdleTimeout() * 4);
+                r.setRememberMe(true);
```
pruivo, line 529: @lhanusov, please apply these changes in the new test, as the test is not 100% correct. 
The method `testRestartSession` is fixed by the other changes.

pruivo:
@ahus1, I know we discussed enabling the `rememberMe` in the realm in the follow-up, but it is simpler enough to do everything in a single PR. Let me know if you agree. The tests pass with these changes.

pruivo:
ps. I'm assuming I didn't break anything else :sweat_smile:

ahus1 (approved):

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderTest.java
lhanusov: thanks @pruivo for the quick fix :-)

PR #43184, opened 2025-10-03
Keycloak CI - MSSQL store failure -> adding startup db configuration

### LINKED ISSUE #42166: [Keycloak CI - Store MSSQL] GroupTest.createMultiDeleteMultiReadMulti:157
### Before reporting an issue

### Area

storage

### Describe the bug

Error in the GroupTest when running MSSQL Store:
* https://github.com/keycloak/keycloak/actions/runs/17243247770/job/48926432742
* https://github.com/keycloak/keycloak/actions/runs/17107583009/job/48520750195
* https://github.com/keycloak/keycloak/actions/runs/17065734133/job/48382642319

```
Error:  Failures: 
Error:    GroupTest.createMultiDeleteMultiReadMulti:157 
Expected: an empty collection
     but: <[jakarta.ws.rs.InternalServerErrorException: HTTP 500 Internal Server Error]>
```

### Version

nightly

### Regression

### Expected behavior

test must pass

### Actual behavior

test is failing 

### How to Reproduce?

CI

### Anything else?

_No response_

lhanusov:
Closes #42166

### DIFF
--- a/test-framework/db-mssql/src/main/java/org/keycloak/testframework/database/MSSQLServerTestDatabase.java
+++ b/test-framework/db-mssql/src/main/java/org/keycloak/testframework/database/MSSQLServerTestDatabase.java
@@ -47,7 +47,11 @@ class MSSQLServerTestDatabase extends AbstractContainerTestDatabase {
 
     @Override
     public List<String> getPostStartCommand() {
-        return List.of("/opt/mssql-tools18/bin/sqlcmd", "-U", "sa", "-P", getPassword(), "-No", "-Q", "CREATE DATABASE " + getDatabase());
+        return List.of("/opt/mssql-tools18/bin/sqlcmd", "-U", "sa", "-P", getPassword(), "-No", "-Q", "CREATE DATABASE " + getDatabase() +
+                " CONTAINMENT = PARTIAL\n" +
+                "GO\n" +
+                "ALTER DATABASE " + getDatabase() + " SET READ_COMMITTED_SNAPSHOT ON\n" +
+                "GO");
     }
 
     @Override


vramik (approved):
@lhanusov thank you for preparing the PR. lgtm

tests/base/src/test/java/org/keycloak/tests/admin/group/GroupTest.java
```diff
@@ -116,6 +117,9 @@ public class GroupTest extends AbstractGroupTest {
 
     
     @Test
+    // Keycloak CI issue #42166 - MSSQL docker image time to time has a deadlock failure,
+    // it's due to a lack of resources of a generic startup on runner.
```
vmuzikar: Would it be feasible to increase the resources rather than disable the test?

ahus1: I think the analysis is not correct. The error message is the following: 

> Uncaught server error: org.hibernate.sql.exec.ExecutionException: A problem occurred in the SQL executor : Error advancing (next) ResultSet position [An error occurred during the current command (Done status 0). Transaction (Process ID 55) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.]

Such deadlocks occur when rows are updated concurrently in different transactions, and MSSQL can't resolve this in the transaction isolation. 

Maybe @sschu can comment on how MSSQL should be configured to prevent such errors on that database.

ahus1 (changes requested):
@sschu - can you please have a look at the issue below?

tests/base/src/test/java/org/keycloak/tests/admin/group/GroupTest.java
sschu: I have never seen this issue before. When we use the MS SQL test container in our integration tests, we do the following configuration to make it behave the same as in Azure SQL:
```
CREATE DATABASE keycloak CONTAINMENT = PARTIAL
GO
ALTER DATABASE keycloak SET READ_COMMITTED_SNAPSHOT ON
GO
```
Would be worth a try to see if the issue goes away.

lhanusov: Thanks @sschu, I will give it a try.

lhanusov: It didn't help.

sschu: What exactly is the test updating? Normally, a deadlock occurs if the same data is updated but in reverse order.

lhanusov: This is the test: https://github.com/keycloak/keycloak/blob/main/tests/base/src/test/java/org/keycloak/tests/admin/group/GroupTest.java#L119

In does concurrently: read and delete

lhanusov: Based on discussion with @stianst I am closing this PR and moving the issue to core-iam team as this seems to be a real bug or a wrongly written test.

sschu:
@lhanusov @ahus1 Ithink the READ_COMMITTED_SNAPSHOT isolation level should be kept if the goal is to support Azure SQL because this is the primary difference between Azure SQL and MS SQL, the first uses that level as default.

PR #43189, opened 2025-10-03
[FGAP] Make additional rest endpoints respect permissions

### LINKED ISSUE #40058: [FGAP] Make additional rest endpoints respect permissions
There are REST endpoint which does not take admin permissions into an account when returning response.

With following permission denying to view a `userC`:
Image

When listing client sessions:

`GET | http://localhost:8080/admin/realms/test/ui-ext/sessions/client?first=0&max=11&type=ALL&clientId=7bb6b97b-67b6-4cc6-9d99-974c73ec9157&search=`

Image

Or when listing users with a certain role:

`GET | http://localhost:8080/admin/realms/test/clients/7bb6b97b-67b6-4cc6-9d99-974c73ec9157/roles/test-role/users?briefRepresentation=true&first=0&max=11`

Image

The lists includes users who should not be visible.

vramik:
Closes #40058

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
@@ -45,6 +45,7 @@ import org.keycloak.models.jpa.entities.UserConsentClientScopeEntity;
 import org.keycloak.models.jpa.entities.UserConsentEntity;
 import org.keycloak.models.jpa.entities.UserEntity;
 import org.keycloak.models.jpa.entities.UserGroupMembershipEntity;
+import org.keycloak.models.jpa.entities.UserRoleMappingEntity;
 import org.keycloak.models.utils.KeycloakModelUtils;
 import org.keycloak.storage.StorageId;
 import org.keycloak.storage.UserStorageProvider;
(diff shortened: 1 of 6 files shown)


integration/admin-client/pom.xml
```diff
@@ -79,6 +79,10 @@
             <artifactId>resteasy-jaxb-provider</artifactId>
             <version>${resteasy.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-rest-admin-ui-ext</artifactId>
```
pedroigor: I'm not sure if we are supposed to have UI-ext exposed. @stianst is it the case?

stianst: Yes, those are not public REST API, and are internal only for the Admin UI

stianst: Perhaps just use http-client directly to test those or something?

pedroigor: Yeah, using the http-client directly is the only way we can go, I think. Similar to org.keycloak.testsuite.user.profile.UIRealmResourceTest.

vramik: thanks for the info, I'll prepare the update

pedroigor (approved):

PR #43193, opened 2025-10-03
fix: adding checks around the hostname path

### LINKED ISSUE #43166: Backend url misses the path with reverse proxy
### Before reporting an issue

### Area

admin/api

### Describe the bug

The Backend URL is not calculated correctly behind a reverse proxy.
The path is missing.
This leads to the situation that our frontend is trying to get a token from the token URL, but querries the base URL

### Version

26.4.0 (docker: keycloak/keycloak:latest)

### Regression

### Expected behavior

The Backend URL matches the base URL (Proto+Host+Path) of the Request URL
Since I am using xForward proxy headers I would expect:
X-Forwarded-Proto + X-Forwarded-Host + X-Forwarded-Prefix 

### Actual behavior

The Backend URL misses the base path
Image

if I acces the server directly like a backend service would everyting is fine
Image

### How to Reproduce?

here is my base example:
run: docker compose up -d --build

.env
``` .env
SERVER_NAME=localhost
DB_PASSWORD=pUIqsDsRUMnq3Lh8
```
certs.sh
``` bash
#!/bin/bash -e
if [ ! -e "/etc/nginx/ssl/cert.pem" ]; then
   echo "Create certs"
   openssl req -x509 -newkey rsa:4096 -keyout /etc/nginx/ssl/key.pem -out /etc/nginx/ssl/cert.pem -sha256 -days 3650 -nodes -subj "/CN=$ENV_SERVER_NAME"
else
   echo "Certs allready created"
fi
```

docker-compose.yaml
``` yaml
version: '3'
volumes:
  pgdata:
  logs:
  plugins:
  certs:
services:
  proxy:
    image: local_proxy
    build: .
    ports:
      - '80:80'
      - '443:443'
    environment:
      ENV_SERVER_NAME: ${SERVER_NAME}
    volumes:
      - type: volume
        source: certs
        target: /etc/nginx/ssl
        read_only: false
    depends_on:
      - keycloak

  keycloak:
    image: keycloak/keycloak
    command: start
    ports:
      - '8090:8080'
    environment:
      KC_BOOTSTRAP_ADMIN_USERNAME: 'admin'
      KC_BOOTSTRAP_ADMIN_PASSWORD: 'admin'
      KC_DB: 'postgres'
      KC_DB_URL: 'jdbc:postgresql://database:5432/keycloak'
      KC_DB_USERNAME: postgres
      KC_DB_PASSWORD: ${DB_PASSWORD}
      KC_HOSTNAME: "https://${SERVER_NAME}/keycloak"
      KC_HOSTNAME_STRICT: false
      KC_HOSTNAME_DEBUG: true
      KC_HTTP_ENABLED: true
      KC_PROXY_HEADERS: xforwarded
      KC_HOSTNAME_BACKCHANNEL_DYNAMIC: true
    depends_on:
      - database

  database:
    image: 'postgres:14'
    ports:
      - '5430:5432'
    volumes:
      - pgdata:/var/lib/postgresql/data
    environment:
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: ${DB_PASSWORD}
      POSTGRES_DB: 'keycloak'    
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER}"]
      interval: 10s
      timeout: 20s
      retries: 5

```

Dockerfile
``` Dockerfile
FROM nginx:alpine

# Add bash
RUN apk add --no-cache bash
RUN apk add --no-cache openssl

RUN touch /etc/nginx/conf.d/proxy.conf
RUN rm /etc/nginx/conf.d/default.conf

COPY ./certs.sh /etc/nginx/
RUN mkdir /etc/nginx/ssl

WORKDIR /etc/nginx/
RUN dos2unix < certs.sh > certs2.sh

COPY ./proxy.conf.template /etc/nginx/

# Default port exposure
EXPOSE 80
EXPOSE 443

# CMD ["tail", "-f", "/dev/null"]
CMD ["sh", "-c", "sh certs2.sh && envsubst '${ENV_SERVER_NAME}' < proxy.conf.template > conf.d/proxy.conf && nginx-debug -g \"daemon off;\""]
```

proxy.conf.template
``` conf
server {
  listen 80;
  server_name _;
  return 301 https://$ENV_SERVER_NAME$request_uri;
}

server {
  listen 443 ssl;
  server_name _;
  client_max_body_size 0;

  ssl_certificate /etc/nginx/ssl/cert.pem;
  ssl_certificate_key /etc/nginx/ssl/key.pem;

  location /keycloak/ {
    rewrite ^/keycloak/(.*)$ /$1 break;
    proxy_pass http://keycloak:8080;
    proxy_set_header X-Forwarded-For        $remote_addr;
    proxy_set_header X-Forwarded-Host       $host;
    proxy_set_header X-Forwarded-Proto      $scheme;
    proxy_set_header X-Forwarded-Scheme     $scheme;
    proxy_set_header X-Forwarded-Prefix     /keycloak;
    proxy_set_header X-Original-Forwarded-For $http_x_forwarded_for;
  }
}

```

### Anything else?

We currently are running this setup with keycloak 24, but since 25 its broken

shawkins:
closes: #43166

@vmuzikar absent support for X-Forwarded-Prefix (which won't work for the Forwarded header regardless) here's a possible way to handle the validation and detection of this case.

There would of course need to be some doc updates here as well. I can understand that you may not want to account for this case in code. The alternative there is to document why it doesn't work and suggest using http-relative-path.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HostnameV2PropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HostnameV2PropertyMappers.java
@@ -11,6 +11,7 @@ import java.util.stream.Stream;
 
 import org.keycloak.common.Profile;
 import org.keycloak.config.HostnameV2Options;
+import org.keycloak.config.HttpOptions;
 import org.keycloak.quarkus.runtime.Environment;
 import org.keycloak.quarkus.runtime.cli.Picocli;
 import org.keycloak.quarkus.runtime.cli.command.AbstractCommand;
@@ -103,10 +104,30 @@ public final class HostnameV2PropertyMappers implements PropertyMapperGrouping {
                 // else might be allowable if HOST is overwritten
             }
 
+            if (proxyHeaders == null && !url.getPath().isEmpty()
+                    && Boolean.valueOf(Configuration.getConfigValue(HostnameV2Options.HOSTNAME_BACKCHANNEL_DYNAMIC).getValue())
+                    && !normalizePath(url.getPath()).equals(
+                            normalizePath(Configuration.getConfigValue(HttpOptions.HTTP_RELATIVE_PATH).getValue()))) {
+                warn.accept("Likely misconfiguration detected. When using a `hostname` that includes a path that does not match the `http-relative-path` you must use `proxy-headers` to properly detect backchannel requests.");
+            }
+
             return true;
         } catch (MalformedURLException e) {
             return false;
         }
+    }
+
+    private static String normalizePath(String path) {
+        if (path == null) {
+            return path;
+        }
+        if (!path.startsWith("/")) {
+            path = "/" + path;
+        }
+        if (path.endsWith("/")) {
+            path = path.substring(0, path.length() - 1);
+        }
+        return path;
     }
 
 }
--- a/services/src/main/java/org/keycloak/url/HostnameV2Provider.java
+++ b/services/src/main/java/org/keycloak/url/HostnameV2Provider.java
@@ -67,7 +67,7 @@ public class HostnameV2Provider implements HostnameProvider {
                 builder.host("localhost");
                 break;
             case BACKEND:
-                builder = backchannelDynamic ? originalUriInfo.getBaseUriBuilder() : getFrontUriBuilder(originalUriInfo);
+                builder = backchannelDynamic && !isFrontendRequest(originalUriInfo) ? originalUriInfo.getBaseUriBuilder() : getFrontUriBuilder(originalUriInfo);
                 break;
             case FRONTEND:
                 builder = getFrontUriBuilder(originalUriInfo);
@@ -76,13 +76,29 @@ public class HostnameV2Provider implements HostnameProvider {
                 throw new IllegalArgumentException("Unknown URL type");
         }
 
+        URI uri = builder.build();
         // sanitize ports
-        URI uriPeak = builder.build();
-        if ((uriPeak.getScheme().equals("http") && uriPeak.getPort() == 80) || (uriPeak.getScheme().equals("https") && uriPeak.getPort() == 443)) {
-            builder.port(-1);
+        int normalizedPort = normalizedPort(uri);
+        if (normalizedPort != uri.getPort()) {
+            builder.port(normalizedPort);
+            uri = builder.build();
         }
 
-        return builder.build();
+        return uri;
+    }
+
+    private int normalizedPort(URI uri) {
+        if ((uri.getScheme().equals("http") && uri.getPort() == 80) || (uri.getScheme().equals("https") && uri.getPort() == 443)) {
+            return -1;
+        }
+        return uri.getPort();
+    }
+
+    private boolean isFrontendRequest(UriInfo originalUriInfo) {
+        URI frontend = getFrontUriBuilder(originalUriInfo).build();
+        return frontend.getScheme().equals(originalUriInfo.getBaseUri().getScheme()) &&
+                frontend.getHost().equals(originalUriInfo.getBaseUri().getHost()) &&
+                frontend.getPort() == normalizedPort(originalUriInfo.getBaseUri());
     }
 
     private UriBuilder getFrontUriBuilder(UriInfo originalUriInfo) {
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/ProxyHostnameV2DistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/ProxyHostnameV2DistTest.java
@@ -104,6 +104,16 @@ public class ProxyHostnameV2DistTest {
     }
 
     @Test
+    @Launch({ "start-dev", "--hostname=https://mykeycloak.org:8443/path", "--proxy-headers=forwarded", "--hostname-backchannel-dynamic=true" })
+    public void testForwardedProxyHeadersWithPathAndDynamicBackchannel(LaunchResult result) {
+        assertFrontEndUrl("https://mykeycloak.org:8443", "https://mykeycloak.org:8443/path/");
+        // a backend url generated via the frontend protocol/host/port should be a front-end url
+        assertBackEndUrl("https://mykeycloak.org:8443", "https://mykeycloak.org:8443/path/");
+        // any other protocol/host/port will be the backend
+        assertBackEndUrl("http://localhost:8080", "http://localhost:8080/");
+    }
+
+    @Test
     @Launch({ "start-dev", "--hostname-strict=false", "--proxy-headers=xforwarded" })
     public void testXForwardedProxyHeaders() {
         assertForwardedHeaderIsIgnored();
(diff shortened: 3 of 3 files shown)


services/src/main/java/org/keycloak/url/HostnameV2Provider.java
```diff
@@ -85,6 +85,13 @@ private URI getUri(UriInfo originalUriInfo, UrlType type) {
         return builder.build();
     }
 
+    private boolean isFrontendRequest(UriInfo originalUriInfo) {
+        URI frontend = getFrontUriBuilder(originalUriInfo).build();
+        String frontendAuthority = frontend.getAuthority();
+        String authority = originalUriInfo.getBaseUri().getAuthority();
+        return frontendAuthority.equals(authority);
```
vmuzikar: I would not go this way. It makes assumptions that might not be true. It means that if the incoming request resembles frontend URL (by hostname and port), we will treat the backend URLs as frontend despite the `hostname-backchannel-dynamic=true` (which means backend is _always_ dynamic). What if someone really uses the same hostname for frontend and backend with only different context roots? It's essentially also a breaking change. Additionally, it introduces additional complexity to quite a fragile area where it might have unexpected side effects.

If you don't feel strongly about it, I would just document the behaviour.

vmuzikar (changes requested):

services/src/main/java/org/keycloak/url/HostnameV2Provider.java
shawkins: > It means that if the incoming request resembles frontend URL (by hostname and port), we will treat the backend URLs

That is correct, if the protocol, host, and port matches we'll assume that the request is via the front-end.

> What if someone really uses the same hostname for frontend and backend with only different context roots?

To elaborate on what that would look like today, it would need to be:

hostname=https://host:port/front-end-path
http-relative-path=/some-other-path
hostname-backchannel-dynamic=true
proxy-headers=forwarded|xforwarded

And you would access the backchannel specifically as https://host:port/some-other-path via the proxy - any other host, port, or protocol will be dynamically handled as the backchannel.

Honestly I would not expect anyone to be using https://host:port/some-other-path via the proxy as the backchannel, but you are correct it does seem possible given our current code.

> It's essentially also a breaking change.

That's not clear to me. Are you also thinking that the setup will have disabled the front-end versions of the generated urls?

vmuzikar: > And you would access the backchannel specifically as https://host:port/some-other-path via the proxy

This could work without proxy as well. The `https://host:port/some-other-path` could be directly resolvable on the internal network while blocked from the outside network.

> That's not clear to me. Are you also thinking that the setup will have disabled the front-end versions of the generated urls?

I meant the use case when the backend will be on a different path (specified by `http-relative-path`) than the frontend.

shawkins: > This could work without proxy as well. The https://host:port/some-other-path could be directly resolvable on the internal network while blocked from the outside network.

Let me elaborate on this a little more. We can also infer that for the protocol to be the same, there shouldn't be an edge setup. For the port to be the same, you are either publicly exposing keycloak at 8443 (or whatever the configured https port is), or have configured keycloak to directly use the 443 priviliged port (which generally is not recommended or is difficult https://github.com/keycloak/keycloak/discussions/42579) 

Internally you then access https://host:port/some-other-path without a proxy, and go directly to the keycloak instance. Regardless of wither you get urls back as https://host:port/some-other-path or https://host:port/front-end-path you will go directly to the keycloak instance, but if a frond-end url is used, then you have the wrong path. Doesn't it seem like there are other ramifications to this setup - as any usage of the public host on the internal network beyond the backchannel is broken?

> I meant the use case when the backend will be on a different path (specified by http-relative-path) than the frontend.

Yes, and it's not very clear why that is breaking. The non-proxied scenario above is extremely nuanced and it's not clear why anyone would do that.

vmuzikar: > Doesn't it seem like there are other ramifications to this setup - as any usage of the public host on the internal network beyond the backchannel is broken?

Wouldn't you be expected to use the proxy for accessing frontend even on internal network?

> The non-proxied scenario above is extremely nuanced and it's not clear why anyone would do that.

Yeah, that's right.. Still, someone might. And IMHO we could say something similar around the use case described in #43166 – either the internal apps use the public proxy or not for accessing both frontend and backchannel from the internal network. If the proxy is used (which they claim they do), they should've configured `backchannel-dynamic=false`. The more I think about #43166, the more it sounds like a misconfig to me, TBH.

shawkins: > Wouldn't you be expected to use the proxy for accessing frontend even on internal network?

I think that is part of my point about why this is a non-working scenario. If the host resolves to an internal keycloak addess then either both those requests are going through a proxy, or both are not. I don't see how internal requests to https://host:port/some-other-path would go directly to keycloak while internal requests to https://host:port/front-end-path would go through a proxy.

> If the proxy is used (which they claim they do), they should've configured backchannel-dynamic=false

backchannel-dynamic=false is not required when using a proxy. Just because you are using the frontend via the proxy doesn't mean that you then should not be able use a dynamic backchannel, or am I missing something?

shawkins:
Updated with a few refinements and a test case.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HostnameV2PropertyMappers.java
```diff
@@ -103,10 +104,30 @@ static boolean validateFullHostname(boolean httpsEnabled, boolean isProd, String
                 // else might be allowable if HOST is overwritten
             }
 
+            if (proxyHeaders == null && !url.getPath().isEmpty()
+                    && Boolean.valueOf(Configuration.getConfigValue(HostnameV2Options.HOSTNAME_BACKCHANNEL_DYNAMIC).getValue())
+                    && !normalizePath(url.getPath()).equals(
+                            normalizePath(Configuration.getConfigValue(HttpOptions.HTTP_RELATIVE_PATH).getValue()))) {
+                warn.accept("Likely misconfiguration detected. When using a `hostname` that includes a path that does not match the `http-relative-path` you must use `proxy-headers` to properly detect backchannel requests.");
```
vmuzikar, line 111: Why would this be the misconfig? This is fixed by the changes in the hostname provider in this PR, no?

shawkins: No, the fix works only if proxy-headers is enabled.

services/src/main/java/org/keycloak/url/HostnameV2Provider.java
vmuzikar: Discussed offline, and agreed to take the proposed path in the end.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HostnameV2PropertyMappers.java
vmuzikar: Discussed offline.

vmuzikar (approved):
LGTM, thank you.

Issue #16375, opened 2023-01-10
Deterministic realm import/export


slieschke:
### Description

Update realm import/export round tripping to be deterministic.

### Motivation

This introduces annoying noise in diffs when storing exported realm configurations are in Git, and changes to the realm configuration are made.

### Details

The current behaviour can be observed by running the following commands from the same directory as [the realm configuration extracted from this archive](https://github.com/keycloak/keycloak/files/10386163/demo-realm-import.json.zip):

```
docker run --volume $(pwd):/opt/keycloak/data quay.io/keycloak/keycloak:20.0.2 import --file /opt/keycloak/data/demo-realm-import.json
docker run --volume $(pwd):/opt/keycloak/data quay.io/keycloak/keycloak:20.0.2 export --file /opt/keycloak/data/demo-realm-export.json --realm demo
```

Diffing the `demo-realm-import.json` and `demo-realm-export.json` files reveals:
* the order of the `allowed-protocol-mapper-types` config for both the anonymous and authenticated allowed prototype mapper types components [have been reordered](https://user-images.githubusercontent.com/742941/211642718-6524331c-c887-4657-a3ca-21b21d6bdbcd.png)
* the IDs of the authentication flows from the original [are changed](https://user-images.githubusercontent.com/742941/211642922-d89a2a3b-8067-4e4a-9a8f-e9aa9964b404.png)

formatq:
The same situation in version 21.0.0

shawkins:
Still occurs in KC 26.

PR #43196, opened 2025-10-03
fix: retaining user creation timestamp when importing

### LINKED ISSUE #43195: Field "Created at" shows "Invalid Date" when created through KeycloakRealmImport
### Before reporting an issue

### Area

operator

### Describe the bug

For non-productive environments, I want to autogenerate users based on the KeycloakRealmImport resource. For that purpose, I create a realm with a simple client and simple user to deploy Keycloak in an automated fashion.

The configuration of the Import for the user is like the following:
```yaml
    users:
      - username: johnDoe
        email: john.doe@example.com
        emailVerified: true
        enabled: true
        credentials:
          - value: johnDoe
```

When Keycloak is deployed and configured, the result in the Admin console is the following:

Image

And if I try to connect to the Realm Console, after logging in and setting the First Name and Last Name, the error that I see is:

Image

However, if I create the user using the Admin console and then try to log in, it works fine. For that reason, I think that the problem is on the Import part.

### Version

26.2.9

### Regression

### Expected behavior

I would expect to be able to log in as I can using a user created using the Admin Console.

### Actual behavior

I cannot access the Realm console due to an unclear error.

### How to Reproduce?

Once the operator is installed, I have the following resources:

```yaml
---
# Source: keycloak-chart/templates/configuration/svc-keycloak-service-trusted.yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.openshift.io/serving-cert-secret-name: keycloak-tls
  labels:
    app: keycloak
  name: keycloak-service-trusted
  namespace: keycloak
spec:
  internalTrafficPolicy: Cluster
  ipFamilies:
    - IPv4
  ipFamilyPolicy: SingleStack
  ports:
    - name: https
      port: 8443
  selector:
    app: keycloak
---
# Source: keycloak-chart/templates/configuration/keycloak-keycloak.yaml
kind: Keycloak
metadata:
  name: keycloak
  namespace: keycloak
  annotations:
    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
spec:
  instances: 1
  bootstrapAdmin:
    user:
      secret: keycloak-admin
  db:
    vendor: postgres
    host: postgresql-db
    usernameSecret:
      name: postgresql-db-credentials
      key: username
    passwordSecret:
      name: postgresql-db-credentials
      key: password
  http:
    tlsSecret: keycloak-tls
  ingress:
    annotations:
      route.openshift.io/destination-ca-certificate-secret: keycloak-tls
      route.openshift.io/termination: reencrypt
    className: openshift-default
  proxy:
    headers: xforwarded # double check your reverse proxy sets and overwrites the X-Forwarded-* headers
---
# Source: keycloak-chart/templates/configuration/keycloakrealmimport.yaml
kind: KeycloakRealmImport
metadata:
  name: example-realm
  namespace: keycloak
  annotations:
    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
spec:
  keycloakCRName: keycloak
  realm:
    id: example-realm
    realm: example-realm
    displayName: Example Realm
    enabled: true
    clients:
      - name: 
        clientId: 
        enabled: true
        redirectUris:
          - "https://oauth-openshift.apps.<changeme>/oauth2callback/keycloak/*"
        directAccessGrantsEnabled: true
    users:
      - username: johnDoe
        email: john.doe@example.com
        emailVerified: true
        enabled: true
        credentials:
          - value: johnDoe
    groups: []
```
For the PostgreSQL, I use a standard configuration.

### Anything else?

_No response_

shawkins:
closes: #43195

I realize that the original intention of nulling out the creation timestamp may have been to keep import / export more symetrical - but we aren't careful about that in other ways #16375, and it seems understandable from a user perspecitve for defaults like this to automatically populate on import and then be present in the export. 

Or would we rather advise users to explicitly set the creation timestamp in the import to avoid displaying "Invalid date", and possibly issues with RLM?

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
@@ -974,7 +974,9 @@ public class DefaultExportImportManager implements ExportImportManager {
         // Import users just to user storage. Don't federate
         UserModel user = UserStoragePrivateUtil.userLocalStorage(session).addUser(newRealm, userRep.getId(), userRep.getUsername(), false, false);
         user.setEnabled(userRep.isEnabled() != null && userRep.isEnabled());
-        user.setCreatedTimestamp(userRep.getCreatedTimestamp());
+        if (userRep.getCreatedTimestamp() != null) {
+            user.setCreatedTimestamp(userRep.getCreatedTimestamp());
+        }
         user.setEmail(userRep.getEmail());
         if (userRep.isEmailVerified() != null) user.setEmailVerified(userRep.isEmailVerified());
         user.setFirstName(userRep.getFirstName());
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/exportimport/ExportImportUtil.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/exportimport/ExportImportUtil.java
@@ -165,8 +165,8 @@ public class ExportImportUtil {
 
         // Test role mappings
         UserRepresentation admin = findByUsername(realmRsc, "admin");
-        // user without creation timestamp in import
-        Assert.assertNull(admin.getCreatedTimestamp());
+
+        Assert.assertNotNull(admin.getCreatedTimestamp());
         Set<RoleRepresentation> allRoles = allRoles(realmRsc, admin);
         Assert.assertEquals(3, allRoles.size());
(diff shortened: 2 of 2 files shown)


vmuzikar (approved):
The change make sense to me, but let's wait for a review from Core IAM (CC @pedroigor).

mabartos (approved):
Makes sense to me as well.

pedroigor (approved):
LGTM.

I cannot think of any reason to null the creation date during import if it is not set in the rep.

PR #43197, opened 2025-10-03
fix: refining activation condition error handling

### LINKED ISSUE #43096: keycloak-operator 26.4.0 missing clusterrole permissions
### Before reporting an issue

### Area

_No response_

### Describe the bug

Getting this after updating to 26.4.0

> 2025-09-30 17:50:07,029 WARN [io.jav.ope.pro.eve.EventProcessor] (ReconcilerExecutor-keycloakcontroller-54) Uncaught error during event processing ExecutionScope{ resource id: ResourceID{name='keycloak', namespace='auth'}, version: 22441437} - but ano ther reconciliation will be attempted because a superseding event has been received or another retry attempt is pending.: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1:443/apis/apiextensions.k8s.io /v1/customresourcedefinitions/servicemonitors.monitoring.coreos.com. Message: customresourcedefinitions.apiextensions.k8s.io "servicemonitors.monitoring.coreos.com" is forbidden: User "system:serviceaccount:auth:keycloak-operator" cannot get resource " customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope. Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=apiextensions.k8s.io, kind=customresourcedefinitions, name=servicemonitors.moni toring.coreos.com, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=customresourcedefinitions.apiextensions.k8s.io "servicemonitors.monitoring.coreos.com" is forbidden: User "system:serviceaccount:auth:keycloak-operator" cannot get resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, statu s=Failure, additionalProperties={}).

Editing `keycloak-operator-clusterrole` to add "list" and "watch" permissions to "customresourcedefinitions" fixes it

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"rbac.authorization.k8s.io/v1","kind":"ClusterRole","metadata":{"annotations":{},"labels":{"app.kubernetes.io/name":"keycloak-operator"},"name":"keycloak-operator-clusterrole"},"rules":[{"apiGroups":["apiextensions.k8s.io"],"resources":["customresourcedefinitions"],"verbs":["get"]},{"apiGroups":["config.openshift.io"],"resources":["ingresses"],"verbs":["get"]}]}
  creationTimestamp: "2025-08-18T15:33:03Z"
  labels:
    app.kubernetes.io/name: keycloak-operator
  name: keycloak-operator-clusterrole
  resourceVersion: "22444256"
  uid: 642d0d98-5e09-405d-a897-89c2dc0cfdaa
rules:
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - config.openshift.io
  resources:
  - ingresses
  verbs:
  - get
```

### Version

26.4.0

### Regression

### Expected behavior

Permissions to work w/o patching

### Actual behavior

Fails with 

> 2025-09-30 17:50:07,029 WARN [io.jav.ope.pro.eve.EventProcessor] (ReconcilerExecutor-keycloakcontroller-54) Uncaught error during event processing ExecutionScope{ resource id: ResourceID{name='keycloak', namespace='auth'}, version: 22441437} - but ano ther reconciliation will be attempted because a superseding event has been received or another retry attempt is pending.: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1:443/apis/apiextensions.k8s.io /v1/customresourcedefinitions/servicemonitors.monitoring.coreos.com. Message: customresourcedefinitions.apiextensions.k8s.io "servicemonitors.monitoring.coreos.com" is forbidden: User "system:serviceaccount:auth:keycloak-operator" cannot get resource " customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope. Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=apiextensions.k8s.io, kind=customresourcedefinitions, name=servicemonitors.moni toring.coreos.com, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=customresourcedefinitions.apiextensions.k8s.io "servicemonitors.monitoring.coreos.com" is forbidden: User "system:serviceaccount:auth:keycloak-operator" cannot get resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, statu s=Failure, additionalProperties={}).

### How to Reproduce?

Installing per https://www.keycloak.org/operator/installation instructions, but into "auth" namespace instead of "keycloak" namespace. 

```bash
kubectl create namespace auth
kubectl -n auth apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/26.4.0/kubernetes/kubernetes.yml
kubectl patch clusterrolebinding keycloak-operator-clusterrole-binding --type='json' -p='[{"op": "replace", "path": "/subjects/0/namespace", "value":"auth"}]'
```

### Anything else?

_No response_

shawkins:
closes: #43096

If we don't want to allow for users to run without the clusterrolbinding setup appropriately, then we can remove the controller code change.

### DIFF
--- a/.github/workflows/operator-ci.yml
+++ b/.github/workflows/operator-ci.yml
@@ -206,6 +206,7 @@ jobs:
         run: |
           kubectl apply -f src/test/resources/service-monitor-crds.yml
           ./scripts/check-crd-installed.sh servicemonitors
+          kubectl delete pod -l name=keycloak-operator
 
       - name: Deploy an example Keycloak with ServiceMonitor
         working-directory: operator/scripts
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
@@ -19,6 +19,7 @@ package org.keycloak.operator.controllers;
 import io.fabric8.kubernetes.api.model.ContainerState;
 import io.fabric8.kubernetes.api.model.ContainerStateWaiting;
 import io.fabric8.kubernetes.api.model.ContainerStatus;
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.PodSpec;
 import io.fabric8.kubernetes.api.model.PodStatus;
 import io.fabric8.kubernetes.api.model.apps.StatefulSet;
@@ -34,6 +35,7 @@ import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 import io.javaoperatorsdk.operator.api.reconciler.Workflow;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.Dependent;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 import io.javaoperatorsdk.operator.processing.dependent.workflow.CRDPresentActivationCondition;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 import io.quarkus.logging.Log;
@@ -51,6 +53,7 @@ import org.keycloak.operator.crds.v2alpha1.deployment.spec.HostnameSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.HostnameSpecBuilder;
 import org.keycloak.operator.update.UpdateLogicFactory;
 
+import java.net.HttpURLConnection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
@@ -67,11 +70,34 @@ import java.util.concurrent.TimeUnit;
         @Dependent(type = KeycloakNetworkPolicyDependentResource.class, reconcilePrecondition = KeycloakNetworkPolicyDependentResource.EnabledCondition.class),
         @Dependent(
               type = KeycloakServiceMonitorDependentResource.class,
-              activationCondition = CRDPresentActivationCondition.class,
+              activationCondition = KeycloakController.ExceptionAwareCRDPresentActivationCondition.class,
               reconcilePrecondition = KeycloakServiceMonitorDependentResource.ReconcilePrecondition.class
         ),
     })
 public class KeycloakController implements Reconciler<Keycloak> {
+
+    public static class ExceptionAwareCRDPresentActivationCondition<R extends HasMetadata, P extends HasMetadata> extends CRDPresentActivationCondition<R, P> {
+
+        boolean logged = false;
+
+        @Override
+        public boolean isMet(DependentResource<R, P> dependentResource, P primary, Context<P> context) {
+            try {
+                return super.isMet(dependentResource, primary, context);
+            } catch (KubernetesClientException e) {
+                if (e.getCode() == HttpURLConnection.HTTP_FORBIDDEN) {
+                    if (!logged) {
+                        logged = true;
+                        Log.info(
+                                "Operator service account lacks permission for %s. If that is added later, you will need to restart the operator."
+                                        .formatted(HasMetadata.getFullResourceName(dependentResource.resourceType())));
+                    }
+                    return false;
+                }
+                throw e; // allow the default handling, but realistically it's better to return false, than allow retries to be exhausted
+            }
+        }
+    }
 
     public static final String OPENSHIFT_DEFAULT = "openshift-default";
 
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
@@ -19,7 +19,6 @@ package org.keycloak.operator.controllers;
 import io.fabric8.kubernetes.api.model.ContainerState;
 import io.fabric8.kubernetes.api.model.ContainerStateWaiting;
 import io.fabric8.kubernetes.api.model.ContainerStatus;
-import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.PodSpec;
 import io.fabric8.kubernetes.api.model.PodStatus;
 import io.fabric8.kubernetes.api.model.apps.StatefulSet;
@@ -35,8 +34,6 @@ import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 import io.javaoperatorsdk.operator.api.reconciler.Workflow;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.Dependent;
-import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
-import io.javaoperatorsdk.operator.processing.dependent.workflow.CRDPresentActivationCondition;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 import io.quarkus.logging.Log;
 import jakarta.inject.Inject;
@@ -53,7 +50,6 @@ import org.keycloak.operator.crds.v2alpha1.deployment.spec.HostnameSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.HostnameSpecBuilder;
 import org.keycloak.operator.update.UpdateLogicFactory;
 
-import java.net.HttpURLConnection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
@@ -70,34 +66,10 @@ import java.util.concurrent.TimeUnit;
         @Dependent(type = KeycloakNetworkPolicyDependentResource.class, reconcilePrecondition = KeycloakNetworkPolicyDependentResource.EnabledCondition.class),
         @Dependent(
               type = KeycloakServiceMonitorDependentResource.class,
-              activationCondition = KeycloakController.ExceptionAwareCRDPresentActivationCondition.class,
-              reconcilePrecondition = KeycloakServiceMonitorDependentResource.ReconcilePrecondition.class
+              activationCondition = KeycloakServiceMonitorDependentResource.ActivationCondition.class
         ),
     })
 public class KeycloakController implements Reconciler<Keycloak> {
-
-    public static class ExceptionAwareCRDPresentActivationCondition<R extends HasMetadata, P extends HasMetadata> extends CRDPresentActivationCondition<R, P> {
-
-        boolean logged = false;
-
-        @Override
-        public boolean isMet(DependentResource<R, P> dependentResource, P primary, Context<P> context) {
-            try {
-                return super.isMet(dependentResource, primary, context);
-            } catch (KubernetesClientException e) {
-                if (e.getCode() == HttpURLConnection.HTTP_FORBIDDEN) {
-                    if (!logged) {
-                        logged = true;
-                        Log.info(
-                                "Operator service account lacks permission for %s. If that is added later, you will need to restart the operator."
-                                        .formatted(HasMetadata.getFullResourceName(dependentResource.resourceType())));
-                    }
-                    return false;
-                }
-                throw e; // allow the default handling, but realistically it's better to return false, than allow retries to be exhausted
-            }
-        }
-    }
 
     public static final String OPENSHIFT_DEFAULT = "openshift-default";
 
@@ -256,6 +228,10 @@ public class KeycloakController implements Reconciler<Keycloak> {
         }
 
         distConfigurator.validateOptions(keycloakCR, status);
+
+        context.managedWorkflowAndDependentResourceContext()
+                .get(KeycloakServiceMonitorDependentResource.SERVICE_MONITOR_WARNING, String.class)
+                .ifPresent(status::addWarningMessage);
     }
 
     public static boolean isRolling(StatefulSet existingDeployment) {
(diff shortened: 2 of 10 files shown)


vmuzikar (commented):
In essence, this looks good to me. A question is whether this is really what we want to do.

A log message is well hidden, even more so if it's logged only once. This can happen from two reasons:
* There's a misconfiguration in the role binding (either caused by the order of how things were applied, or the role is simply missing).
   * In this case Operator restart is required. A log message could be very easily missed, hence users might have trouble spotting it.
* The role is missing on purpose, the user doesn't want to use a cluster role.
   * Do we want to support this use case? We don't document we do.
   * Shouldn't we rather document how do explicitly disable the service monitor, so the role is not even needed when disabled and therefore we would not log it in this case?

+1 for not failing the whole reconciliation. I'd just make it more explicit – maybe a warn is CR status, or even an error?

shawkins:
> Do we want to support this use case? We don't document we do.

We don't document it, but many non-olm non-openshift users may not be aware that the clusterrolebinding is even required.

> +1 for not failing the whole reconciliation. I'd just make it more explicit – maybe a warn is CR status, or even an error?

I think that's fair - if you put the servicemonitor as enabled in a CR, but it isn't supported, then you should see something more than just this initial log.

shawkins:
With the latest changes, the new behavior should be:
- the cluster role is no longer required. Instead at startup we're using a customized activation condition to produce a WARNING message if it fails.
- if you add a servicemonitor spec as enabled to the Keycloak CR, but have not enabled metrics, a warning will be added to the CR status. If desired this could be expanded to also warn about the missing CRD case as well.

vmuzikar:
@shawkins Thanks for the update.

> the cluster role is no longer required. Instead at startup we're using a customized activation condition to produce a WARNING message if it fails.

Should we add a warning to the CR status in this case as well?

> If desired this could be expanded to also warn about the missing CRD case as well.

+1

ryanemerson:
Thanks @shawkins @vmuzikar for taking care of this, I like the approach of handling the missing role gracefully.

> Should we add a warning to the CR status in this case as well?

Would this be too verbose?

With these changes we're allowing the users to use the Operator with limited functionality without the ClusterRole permissions. I'm just thinking if the user knows they do not require a ServiceMonitor to be created by the Operator and always set `spec.serviceMonitor.enabled: false`, it might be strange if their CR(s) still have warnings.

vmuzikar:
> Would this be too verbose?

I prefer verbosity over log message that can be easily missed. We could include a hint in the warning that service monitor needs to be disabled.

ryanemerson:
> We could include a hint in the warning that service monitor needs to be disabled.

+1 if the warning only appears when the ServiceMonitor is enabled, I thought you were proposing always warning on a CR if the permissions were missing.

vmuzikar:
> I thought you were proposing always warning on a CR if the permissions were missing

Nopes, only if it's relevant = the service monitor is enabled, which is the default.

shawkins:
> Nopes, only if it's relevant = the service monitor is enabled, which is the default.

Ok, another refinement to better encapsulate the logic and add the additional warning. The one thing this won't do is keep testing for ServiceMonitor support - you have to restart the operator.

.github/workflows/operator-ci.yml
```diff
@@ -163,6 +163,12 @@ jobs:
         working-directory: operator
         run: ./scripts/install-olm.sh
 
+      - name: Install ServiceMonitor CRD
+        working-directory: operator
+        run: |
+          kubectl apply -f src/test/resources/service-monitor-crds.yml
+          ./scripts/check-crd-installed.sh servicemonitors
```
ryanemerson: This will mean that all  tests are executed with the ServiceMonitor CRD available. The reason for installing this later was to ensure that OLM installations worked as expected with/without the CRD.

We could still test this by installing the ServiceMonitor in the same step location as before, but then do a `kubectl rollout restart ...` like we document before proceeding to deploy the examples?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
```diff
@@ -230,6 +228,10 @@ public void updateStatus(Keycloak keycloakCR, StatefulSet existingDeployment, Ke
         }
 
         distConfigurator.validateOptions(keycloakCR, status);
+
+        context.managedWorkflowAndDependentResourceContext()
+                .get(KeycloakServiceMonitorDependentResource.SERVICE_MONITOR_WARNING, String.class)
+                .ifPresent(warning -> status.addWarningMessage(warning));
```
ryanemerson: `status::addWarningMessage`

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -644,27 +643,34 @@ private static void addUpdateRevisionAnnotation(String revision, StatefulSet toU
...
+
+        if (!legacy && (!health || healthManagementEnabled)) {
             port = HttpManagementSpec.managementPort(keycloakCR);
+            portName = Constants.KEYCLOAK_MANAGEMENT_PORT_NAME;
             if (readConfigurationValue(HTTP_MANAGEMENT_SCHEME, keycloakCR, context).filter("http"::equals).isPresent()) {
                 protocol = "HTTP";
             }
         } else {
             port = tls ? HttpSpec.httpsPort(keycloakCR) : HttpSpec.httpPort(keycloakCR);
+            portName = tls ? Constants.KEYCLOAK_HTTP_PORT_NAME : Constants.KEYCLOAK_HTTP_PORT_NAME;
```
ryanemerson: Identical branches, do you mean:

`portName = tls ? Constants.KEYCLOAK_HTTPS_PORT_NAME : Constants.KEYCLOAK_HTTP_PORT_NAME;` ?

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
```diff
@@ -19,28 +22,74 @@
...
+
+            return true;
+        }
+
+        private boolean isCRDInstalled(DependentResource<ServiceMonitor, Keycloak> dependentResource,
+                Context<Keycloak> context, KeycloakServiceMonitorDependentResource serviceMonitorDependentResource) {
+            if (serviceMonitorDependentResource.crdInstalled != null) {
+                return serviceMonitorDependentResource.crdInstalled;
+            }
+            try (var watch = context.getClient().resources(dependentResource.resourceType()).watch(new Watcher<ServiceMonitor>() {
```
ryanemerson: Nitpick, but this will be easier to read if we break it out e.g.

```java
            var watcher = new Watcher<ServiceMonitor>() {
                @Override
                public void eventReceived(Action action, ServiceMonitor resource) {
                }

                @Override
                public void onClose(WatcherException cause) {
                }
            };

            try (var ignored = context.getClient().resources(dependentResource.resourceType()).watch(watcher)) {
                serviceMonitorDependentResource.crdInstalled = true;
                return true;
            } catch (KubernetesClientException e) {
                if (e.getCode() == HttpURLConnection.HTTP_NOT_FOUND) {
                    Log.warn("ServiceMonitors will not be managed by the operator as the CRD for ServiceMonitors not installed. If the CRD is installed later, the operator will need to be restarted.");
                    serviceMonitorDependentResource.crdInstalled = false;
                    return false;
                }
                throw e;
            }
```

operator/src/test/java/org/keycloak/operator/testsuite/integration/ServiceMonitorTest.java
```diff
@@ -61,14 +69,17 @@ public void testServiceMonitorDisabledExplicitly() {
     }
 
     @Test
-    public void testServiceMonitorDisabledLegacyManagement() {
+    public void testServiceMonitorLegacyManagement() {
```
ryanemerson, line 72: The original implementation intentionally didn't create a ServiceMonitor for `legacy-observability-interface` as it is deprecated. If we're going to support this, should we create a separate issue (happy for it to be part of this PR still) so that the change gets some exposure in the release notes?

shawkins: I didn't see this limitation documented originally wrt to ServiceMonitors, so it doesn't seem like it needs documented now. I'd just like to keep the logic as consistent and centralized as possible.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
shawkins: sounds good

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
shawkins: Good catch, I'll also update the test.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
shawkins: updated

.github/workflows/operator-ci.yml
shawkins: That seems good, updated.

shawkins:
The way the ci test is written the rollout won't happen because we're out of resources, so I switched the ci to delete the operator pod. Using scale down / up is also possible.

operator/src/test/java/org/keycloak/operator/testsuite/integration/ServiceMonitorTest.java
ryanemerson: I'm happy with that, I think I was probably over thinking it :slightly_smiling_face:

ryanemerson (approved):
LGTM, thanks @shawkins

ahus1 (approved):
Thank you for all the work and reasoning about this. 

Approving based on @ryanemerson's review. Not merging yet in case someone from the cloud native team wants to review it.

vmuzikar:
I'd like to take a look today, or early next week.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
```diff
@@ -19,28 +22,75 @@
...
             var opts = configuredOptions(primary);
-            if (Boolean.parseBoolean(opts.get(LEGACY_MANAGEMENT_ENABLED)) || !Boolean.parseBoolean(opts.getOrDefault(METRICS_ENABLED, "false"))) {
+
+            if (!Boolean.parseBoolean(opts.getOrDefault(METRICS_ENABLED, "false"))) {
+                context.managedWorkflowAndDependentResourceContext().put(SERVICE_MONITOR_WARNING, WARN_METRICS_NOT_ENABLED);
                 return false;
             }
 
-            return ServiceMonitorSpec.get(primary).isEnabled();
+            if (!isCRDInstalled(dependentResource, context, (KeycloakServiceMonitorDependentResource)dependentResource)) {
```
vmuzikar, line 54: Should we also add some handling for when the cluster role is not configured properly, i.e. show a warn on the CR status for permissions issues?

operator/src/main/kubernetes/kubernetes.yml
```diff
@@ -93,12 +93,6 @@ metadata:
   labels:
     app.kubernetes.io/name: keycloak-operator
 rules:
-  - apiGroups:
-      - apiextensions.k8s.io
-    resources:
-      - customresourcedefinitions
-    verbs:
-      - get
```
vmuzikar, line 101: Just out of curiosity, we can remove this because we're now using a different logic to detect whether the ServiceMonitor CRD is installed?

shawkins: Yes, we no longer query the general customresourcedefinitions endpoint.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java
shawkins: The clusterrole no longer factors into ServiceMonitor usage.

vmuzikar: Right, we removed it. :D Sorry for the noise.

vmuzikar (approved):
LGTM, thanks.

I'll just rebase the PR to run the new Spotless checks.

PR #43239, opened 2025-10-07
Convert the wizard to create client in progressive to validate every step

### LINKED ISSUE #42971: create clients without required value
### Before reporting an issue

### Area

admin/client-js

### Describe the bug

In manage -> clients -> create clients, I can press 3. login setting button to jump to last setting page and save client without client id filled, which is required value. The new client doesn't have client id and I can't get in it's setting page to do anything.

### Version

26.3.4

### Regression

### Expected behavior

Create client page should stop me from jumping to last setting page or creating an invalid client.

### Actual behavior

I can create a client without client id

### How to Reproduce?

manage -> clients:
1. press create clients
2. press 3. login setting
3. press save

### Anything else?

_No response_

rmartinc:
Closes #42971

Just adding the `isProgressive` to the client wizard. This option does not show the future step links in the left side, only the already completed ones (you can go back but cannot go forward). This way, the administrator needs to pass through the three steps and ensures that all the required fields are filled. I don't see any test failing, so I suppose the links were never tested.

### DIFF
--- a/js/apps/admin-ui/src/clients/add/NewClientForm.tsx
+++ b/js/apps/admin-ui/src/clients/add/NewClientForm.tsx
@@ -107,6 +107,7 @@ export default function NewClientForm() {
             onClose={() => navigate(toClients({ realm }))}
             navAriaLabel={`${title} steps`}
             onSave={save}
+            isProgressive
             footer={<NewClientFooter {...form} />}
           >
             <WizardStep


rmartinc:
@jonkoops @ssilvert This is the PR. It's just a one-liner. :smiley:

js/apps/admin-ui/src/clients/add/NewClientForm.tsx
```diff
@@ -107,6 +107,7 @@ export default function NewClientForm() {
             onClose={() => navigate(toClients({ realm }))}
             navAriaLabel={`${title} steps`}
             onSave={save}
+            isProgressive={true}
```
jonkoops: Code style nit; boolean properties that are `true` can omit their value.
```suggestion
            isProgressive
```

jonkoops (approved):

js/apps/admin-ui/src/clients/add/NewClientForm.tsx
rmartinc: One line and it can be improved :facepalm:. Done now!

jonkoops (approved):

js/apps/admin-ui/src/clients/add/NewClientForm.tsx
jonkoops: To be fair, this is something the linter should have picked up and fixed :P

mposolda (approved):
@rmartinc @jonkoops Thanks!

PR #43245, opened 2025-10-07
Use UserSessionUtil.findValidSessionForAccessToken in revocation endpoint

rmartinc:
Closes #43218

The token revocation endpoint failed for transient tokens created by the TE. The reason is the same than in previous https://github.com/keycloak/keycloak/issues/42565, the session can be a transient online session in which the `sid` claim is present but the client session is transient and does not exist. The PR just re-uses the `UserSessionUtil.findValidSessionForAccessToken` used in other endpoints (user-info or introspection) to retrieve the session and the user. Test added.


graziang (approved):
@rmartinc thanks!

mposolda (approved):
@rmartinc @graziang Thanks for the fix and review!

PR #43276, opened 2025-10-07
fix: adds error handling for common redirect codes

### LINKED ISSUE #31401: Kcadm.sh: (Better) Error messages
### Description

**As a user I need readable and actionable error messages.**

If I want to log in and something is not right, I only get some 302 return messages:
```
$ /opt/keycloak/bin/kcadm.sh create realms -s realm=demorealm -s enabled=true --server=http://${KC_DOMAIN} --user=admin --password=${KC_PW}--realm=master --config /tmp/kcadm.config
Logging into http://rhbk.apps.ocp.ocp-gm.de as user admin of realm master
302 Found
```
The problem is, the server uses https and the redirect is only recognized when called from the browser. The error message seems okay, because it is no 2xx but nothing was processed and no realm got created.

On the other way, if the server url is correct, this is the return message:

```
$ /opt/keycloak/bin/kcadm.sh create realms -s realm=demorealm -s enabled=true --server=https://${KC_DOMAIN} --user=admin --password=${KC_PW}--realm=master --config /tmp/kcadm.config
Logging into https://... as user admin of realm master
Created new realm with id 'demorealm'
```

As a user I would like to get something like: "The server URL redirects to another URL. Please check your server URL. Did you forgot to add https?" Something like that.

### Discussion

_No response_

### Motivation

I've spend two days searching for the reason why my request didn't create anything. "302 Found" just doesn't tell me as a user that something is not right. 

Debugging is a real pain without good and actionable error messages.

### Details

As a user I would like to get something like: "The server URL redirects to another URL. Please check your server URL. Did you forgot to add https?" Something like that.

shawkins:
closes: #31401

### DIFF
--- a/integration/client-cli/admin-cli/src/main/java/org/keycloak/client/cli/util/HttpUtil.java
+++ b/integration/client-cli/admin-cli/src/main/java/org/keycloak/client/cli/util/HttpUtil.java
@@ -42,6 +42,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.HttpURLConnection;
 import java.net.URLEncoder;
 import java.security.KeyManagementException;
 import java.security.KeyStoreException;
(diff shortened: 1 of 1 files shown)


vmuzikar (approved):
LGTM, thanks.

PR #43277, opened 2025-10-07
Prevent root attributes from being configured as multivalued  Closes #42749

adhamahmad:
Prevent root attributes from being configured as multivalued  Closes #42749

### Summary
Fixes issue #42749 where enabling "multivalued" for root attributes such as `email`
or `username` results in 400 errors due to deserialization mismatches.

### Changes
- Added validation in `UPConfigUtils.validateAttribute()` to disallow
  multivalued configuration for root attributes.


adhamahmad:
@pedroigor I’ve added the validation only in `UPConfigUtils.validateAttribute()` for now.

However, I think it would also sense to include similar validation in `UPAttribute.setMultivalued()`

If you agree, shouldn't the `isRootAttribute()` logic from` UserProfileUtil` moved  into a shared location to avoid duplication between the two layers.

This PR intentionally avoids refactoring right now to keep the change small and focused, but I can extend it if that’s the preferred approach.

PR #43282, opened 2025-10-08
Update reverseproxy documentation to point configuration of provider Closes (#43284)

igieon:
Add link to configuration_provider with example how to start spi with simplified spi parameter launch.

Fixes: #43284

### DIFF
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -233,6 +233,7 @@ to load additional certificates from headers `CERT_CHAIN_0` to `CERT_CHAIN_9` if
 | Whether the forwarded certificate is url-encoded or not. In NGINX, this corresponds to the `$ssl_client_cert` and `$ssl_client_escaped_cert` variables. This can also be used for the Traefik PassTlsClientCert middleware, as it sends the client certficate unencoded.
 |===
 
+NOTE: look into <@links.server id="configuration-provider"/>. For example tes parameter like --spi-x509cert-lookup-ngnix-ssl-client-cert=SSL_CLIENT_CERT
 === Configuring the NGINX provider
 
 The NGINX SSL/TLS module does not expose the client certificate chain. {project_name}'s NGINX certificate lookup provider rebuilds it by using the {project_name} truststore.
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -233,7 +233,7 @@ to load additional certificates from headers `CERT_CHAIN_0` to `CERT_CHAIN_9` if
 | Whether the forwarded certificate is url-encoded or not. In NGINX, this corresponds to the `$ssl_client_cert` and `$ssl_client_escaped_cert` variables. This can also be used for the Traefik PassTlsClientCert middleware, as it sends the client certficate unencoded.
 |===
 
-NOTE: look into <@links.server id="configuration-provider"/>. For example tes parameter like --spi-x509cert-lookup-ngnix-ssl-client-cert=SSL_CLIENT_CERT
+NOTE: look into <@links.server id="configuration-provider"/>. For example you can use this parameter <@kc.start parameters="--spi-x509cert-lookup-nginx-ssl-client-cert=SSL_CLIENT_CERT"/>
 === Configuring the NGINX provider
 
 The NGINX SSL/TLS module does not expose the client certificate chain. {project_name}'s NGINX certificate lookup provider rebuilds it by using the {project_name} truststore.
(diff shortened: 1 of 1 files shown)


igieon:
I spend some time to find why x509 certificate authentication doesn't work after searching I found documentation about configuration of provider and then i change config parameter from `spi-x509cert-lookup--nginx--ssl-client-cert` to `spi-x509cert-lookup-nginx-ssl-client-cert` and everything starts work.

docs/guides/server/reverseproxy.adoc
```diff
@@ -233,6 +233,7 @@ to load additional certificates from headers `CERT_CHAIN_0` to `CERT_CHAIN_9` if
 | Whether the forwarded certificate is url-encoded or not. In NGINX, this corresponds to the `$ssl_client_cert` and `$ssl_client_escaped_cert` variables. This can also be used for the Traefik PassTlsClientCert middleware, as it sends the client certficate unencoded.
 |===
 
+NOTE: look into <@links.server id="configuration-provider"/>. For example you can use this parameter <@kc.start parameters="--spi-x509cert-lookup-nginx-ssl-client-cert=SSL_CLIENT_CERT"/>
```
vmuzikar: @igieon Thank you for the PR! Unfortunately, the rendering looks a bit off:
Image

Maybe we could do something like this instead:
```suggestion
[NOTE]
====
Consult the <@links.server id="configuration-provider"/> guide for details on the format of the options. For example you can use:

<@kc.start parameters="--spi-x509cert-lookup-nginx-ssl-client-cert=SSL_CLIENT_CERT"/>
====

```
Which then looks like:
Image

vmuzikar (changes requested):

vmuzikar:
Additionally, could you please open and link an issue as per our [contributing guidelines](https://github.com/keycloak/keycloak/blob/main/CONTRIBUTING.md#commit-messages-and-issue-linking)? That would help us a lot with tracking this.

igieon:
@vmuzikar  Updated

docs/guides/server/reverseproxy.adoc
```diff
@@ -233,6 +233,13 @@ to load additional certificates from headers `CERT_CHAIN_0` to `CERT_CHAIN_9` if
 | Whether the forwarded certificate is url-encoded or not. In NGINX, this corresponds to the `$ssl_client_cert` and `$ssl_client_escaped_cert` variables. This can also be used for the Traefik PassTlsClientCert middleware, as it sends the client certficate unencoded.
 |===
 
+[NOTE]
+====
+Consult the <@links.server id="configuration-provider"/> guide for details on the format of the options. For example you can use:
+
+<@kc.start parameters="--spi-x509cert-lookup-nginx-ssl-client-cert=SSL_CLIENT_CERT"/>
```
vmuzikar: I'm sorry, I realized the issue here is about using `-` instead of `--` as the delimiter in the SPI options. The `--` delimiter is the [preferred one](https://www.keycloak.org/docs/latest/upgrading/index.html#spi-options-separating-the-provider-with-a-single-dash) to prevent ambiguity. I originally misunderstood (and missed your comment) that you want to add a link to the providers config guide to help users understand better the format. I also then missed that in the example there was `-` instead of `--`. So it should be like this:
```suggestion
<@kc.start parameters="--spi-x509cert-lookup--nginx--ssl-client-cert=SSL_CLIENT_CERT"/>
```
You mentioned that didn't work for you?

vmuzikar (changes requested):

docs/guides/server/reverseproxy.adoc
igieon: Keycloak version: 26.2.1
yes with double -- -- doesn't work i find 1 docker compose example https://github.com/CarolinaFernandez/keycloak-mtls/blob/master/docker-compose-proxy.yaml#L80

so i firstli setup by based on enviroment values:
```bash
export KC_SPI_X509CERT_LOOKUP_PROVIDER=nginx
export KC_SPI_X509CERT_LOOKUP_NGINX_SSL_CLIENT_CERT=X-Client-Cert
```
and auth start working. 
Then i pass it without double - as start argument also works previously with 2 not. 
and finally when i put it as in the beginning in keycloak conf and without double -- its start works.

shawkins: We can move this discussion to https://github.com/keycloak/keycloak/issues/43284#issuecomment-3381053430

igieon:
I think we can close this pull request or if you think its valuable link we can keep as it is.

shawkins:
> I think we can close this pull request or if you think its valuable link we can keep as it is.

Given that the issue was already closed, I'll close this as well. We have been adding more version information to our guides, and will hopefully provide a way to access older ones soon. Thank you for taking the time to open the issue / pr.

PR #43288, opened 2025-10-08
Fix sdjwt tests: make all string-byte conversions explicit (UTF-8)

Pepo48:
* this unifies behaviour prior to JDK18 on Windows platform

Closes #43264


rmartinc (approved):
Thanks @Pepo48! I have checked and I think all the uses for `encodeNoPad`  are OK and changed to use UTF-8 now. LGTM!

stianst (approved):

PR #43299, opened 2025-10-08
Automatically dispose of realms created by `createTestBed()`

jonkoops:
Automatically disposes of any realms that are created by `createTestBed()` once a test has completed. This uses [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management) to automatically delete the realm created by the test bed once it is no longer in scope. Since this feature is only available in Node.js 24, I have raised the version on CI, and updated the documentation.

Closes #43298


ssilvert (approved):
Cool stuff.  I like it.

PR #43303, opened 2025-10-08
Security Defenses realm settings lost when switching between Headers …

mposolda:
…and Brute Force Detection tabs

closes #42676

### Analysis of the issue

Here is what it seems to happen (took me some time to figure that with my limited react knowledge...)

1) When the admin clicks on the "Security defenses" tab, it has 2 sub-tabs "Headers" and "Brute force detection" . At this point, method `setupForm` inside `BruteForceDetection.tsx` is already called (even if "Brute force detection" subtab is not rendered yet). The method converts `realm` to "form values" used by brute-force tab.

2) When user does some changes in the "Headers" tab and saves them, the "Brute force" tab still has the link to the realm with previously converted values. There is not new call to `setupForm` inside `BruteForceDetection.tsx` .

3) Now when the user switch to the sub-tab "Brute Force detection" (after he did some changes in the "Headers" tab), the `realm` converted for saving in the DB is still the instance, which was initially converted to the "Form values" in step 1. This instance does not contain the updates from step 2. So saving realm in "Brute force" tab effectively rewrite the updates done by step 2 and those are lost.

### Fix

It seems the easiest fix is to add `realm` as the variable being updated in `useEffects` . When the realm is updated in step 2, the `setupForm` of `BruteForceDetection.tsx` would be still called and it will convert to "Form values" the newly updated `realm` values (including the update from step 2).

Note that this issue does not happen when the opposite order of updated tabs is used (EG. First something is updated in the "Brute force" tab followed by the update in the "Headers" tab) . It seems to me this is due the fact that "Headers" uses the realm instance as it is without converting it to "Form values" and hence the update of the `realm` instance in brute-force tab directly updates the instance used by the "Headers" tab. Hence "Headers" tab has all the updates already done by the "Brute force" tab.

### DIFF
--- a/js/apps/admin-ui/src/realm-settings/security-defences/BruteForceDetection.tsx
+++ b/js/apps/admin-ui/src/realm-settings/security-defences/BruteForceDetection.tsx
@@ -59,7 +59,7 @@ export const BruteForceDetection = ({
     convertToFormValues(realm, setValue);
     setIsBruteForceModeUpdated(false);
   };
-  useEffect(setupForm, []);
+  useEffect(setupForm, [realm]);
 
   const bruteForceMode = (() => {
     if (!form.getValues("bruteForceProtected")) {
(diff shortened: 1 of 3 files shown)


jonkoops (approved):

mposolda:
@jonkoops Thanks!

PR #43310, opened 2025-10-09
Throw and catch UnsupportedOperationException to fix XPathAttributeMapperTest

rmartinc:
Closes #43262

I think that this is a change in jdk between 17 and 21. In 25 it seems that the exception we are throwing in the `XPATH_FACTORY` is caught and re-thrown as `XPathExpressionException` (see [here](https://github.com/openjdk/jdk25u/blob/master/src/java.xml/share/classes/com/sun/org/apache/xpath/internal/jaxp/XPathExpressionImpl.java#L95-L109)). But in 17 the `RuntimeException` is not caught and therefore there is an exception ([here](https://github.com/openjdk/jdk17/blob/master/src/java.xml/share/classes/com/sun/org/apache/xpath/internal/jaxp/XPathExpressionImpl.java#L90-L104)). I'm just changing to throw a  `UnsupportedOperationException` for the XPATH_FACTORY variable and function resolver and catching that specific exception to continue. If you prefer that I add a custom private exception instead of  `UnsupportedOperationException` just let me know. But this is not very problematic as I suspect the change in 25 is because a bug, so the correct behavior is now in place in the jdk.


mposolda:
@rmartinc Thanks!

mposolda (approved):

PR #43325, opened 2025-10-09
openid-connect flow is missing response type on language change

mposolda:
closes #41292

When the authentication flow is configured in a way that OIDC login request does not display the default username/password page, but some custom authenticator page, then the request might not be handled by the `loginFormsProvider.createResponse` method, but rather by `loginFormsProvider.createForm` method. In this case, the URL used for switching the language in the locale combobox is based on the current browser URL ( `UriBuilder.fromUri(baseUri).path(uriInfo.getPath())` ). But in the case when the page with custom authenticator was just displayed, this URL is OIDC login URL (something like `/auth/realms/test/protocol/openid-connect/auth` ), which is incorrect. The correct URL should be something like `/auth/realms/test/login-actions/authenticate` .

The PR does small update to make sure that default URL for the language combobox is based on the `actionUrl` . This seems to be correct behaviour for most of the pages. The corner-case where it is not accurate is `Language expired page` where the URL should be still based on the current browser URL.


rmartinc (approved):
Thanks @mposolda! LGTM!

mposolda:
@rmartinc Thanks for the review!

PR #43326, opened 2025-10-09
Implement workflow update/copy/enable/disable.  Also authorization.

### LINKED ISSUE #43041: Add "view workflow" and "copy workflow" functions to admin UI
### Description

The workflow currently does not have any way to view or edit a workflow that has been already created.  

We really don't have back end support for editing a workflow except for the "enabled" attribute.  So we will implement a `copy` function that allows the user to edit an existing workflow json and save as a new workflow.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

### LINKED ISSUE #43450: Workflows UI needs to implement authorization
### Description

The Workflows UI needs to match whatever role-based and fine-grained authorization is available on the back end.

ssilvert:
Closes #43041
Closes #43450


js/libs/keycloak-admin-client/src/resources/workflows.ts
jonkoops, line 1: Should some tests be added for these new methods?

js/apps/admin-ui/src/workflows/routes/WorkflowDetail.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const WorkflowDetailForm = lazy(() => import("../WorkflowDetailForm"));
+
+export const WorkflowDetailRoute: AppRouteObject = {
+  path: "/:realm/workflows/:mode/:id",
+  element: <WorkflowDetailForm />,
+  breadcrumb: (t) => t("workflowDetails"),
+  handle: {
+    access: "anyone", // TODO: update access when view permission is added
```
jonkoops: Will this be handled as a follow-up issue? Is there an issue logged for this?

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+import { useParams } from "../utils/useParams";
+import {
+  WorkflowDetailParams,
+  toWorkflowDetail,
+} from "./routes/WorkflowDetail";
+import { ViewHeader } from "../components/view-header/ViewHeader";
+
+type AttributeForm = {
+  workflowJSON?: string;
+  [key: string]: any;
```
jonkoops: If possible, providing `unknown` as a type would be preferred over `any`, as it forces a type check at a later point.

For example, if the value is `unkown`:

```ts
value.toString() // won't work, as the type needs to be coerced.

if (typeof value === "number") {
  value.toString() // The compiler knows this is a 'number' so we can now access the method.
}
```

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+      });
+    },
+    (workflow) => {
+      if (!workflow) {
+        return;
+      }
+
+      if (mode === "copy") {
+        delete workflow.id;
+        workflow.name = `${workflow.name} -- ${t("copy")}`;
```
jonkoops, line 69: Ideally this would be interpolated instead, as there might be languages where it only makes sense to add the 'Copy' label at the front, for example:

```ts
// English vs. Japanese:
// 'Copy of Workflow 1'
// 'Workflow 1 のコピー'
workflow.name = t("copyOf", { name: workflow.name ));
```

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const { adminClient } = useAdminClient();
+
+  const { t } = useTranslation();
+  const form = useForm<AttributeForm>({ mode: "onChange" });
+  const { control, handleSubmit } = form;
+  const navigate = useNavigate();
+  const { realm } = useRealm();
+  const { addAlert, addError } = useAlerts();
+  const { mode, id } = useParams<WorkflowDetailParams>();
+  const [workflowJSON, setWorkflowJSON] = useState("");
```
jonkoops, line 51: The state for `workflowJSON` is held here, but also in the form state, it should not be required to have this state in two places, so it should either be in a `useState()` or part of the form state to remove the needs to keep these two states in sync.

jonkoops: (I have some more comments, but I think we should get this big one out-of-the-way first)

js/libs/keycloak-admin-client/src/resources/workflows.ts
ssilvert: Yes.  I'll be doing that separately.  This is just an experimental feature for now so it might change a lot and tests are a lower priority.

js/apps/admin-ui/src/workflows/routes/WorkflowDetail.tsx
ssilvert: I have a list of TODOs but I haven't made issues yet.

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
ssilvert: It is expected that the user will change the name.  And the actual name of the workflow is not meant to be localized.

I considered putting "Copy" at the beginning but that means if the user doesn't change the name then the copy no longer shows up next to the original in the alphabetized list of workflows.

js/libs/keycloak-admin-client/src/resources/workflows.ts
ssilvert: Yes.  I went ahead and created an issue.  https://github.com/keycloak/keycloak/issues/43372

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
ssilvert: Fixed.

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
ssilvert: Fixed.

js/apps/admin-ui/src/workflows/WorkflowDetailForm.tsx
jonkoops: If it doesn't need to be localized, I would keep the translation layer out of it entirely. The English alphabet should not be relied on for implicit sorting, as sorting might be locale-dependent. Either way, I think it should either not be translated at all, or it should be interpolated, enforcing the order in the translation, e.g.:

```
copyOf=${name} -- のコピー
```

js/apps/admin-ui/src/workflows/routes/WorkflowDetail.tsx
jonkoops: Ok, might be good to convert those to issues, or at least tasks on an issue that is logged somewhere.

sguilhen:
LGTM! Is the merge commit intended to be part of the PR?

jonkoops (commented):
The workflow JSON is still being held both in a `useState()` and the form state, this should be using one of these only. I would recommend keeping it in the form state only, and retrieving it when needed.

ssilvert:
> LGTM! Is the merge commit intended to be part of the PR?

@sguilhen No.  I screwed that up.  Closing this PR.  Here is a clean one: https://github.com/keycloak/keycloak/pull/43681

ssilvert:
> The workflow JSON is still being held both in a `useState()` and the form state, this should be using one of these only. I would recommend keeping it in the form state only, and retrieving it when needed.

@jonkoops Created a new PR.  This one should be only using form state.  Please review.
https://github.com/keycloak/keycloak/pull/43681

PR #43333, opened 2025-10-09
fix: adding a single method to get the base uri

### LINKED ISSUE #43330: Optimize hostname provider
### Description

The hostname provider should have an additional method for directly obtaining the whole url or url builder. The most common usage scenario in KeycloakUriInfo results in 4 separate calls to perform that task.

shawkins:
closes: #43330

### DIFF
--- a/server-spi/src/main/java/org/keycloak/urls/HostnameProvider.java
+++ b/server-spi/src/main/java/org/keycloak/urls/HostnameProvider.java
@@ -16,6 +16,8 @@
  */
 package org.keycloak.urls;
 
+import java.net.URI;
+
 import org.keycloak.models.KeycloakContext;
 import org.keycloak.provider.Provider;
 
@@ -117,6 +119,21 @@ public interface HostnameProvider extends Provider {
 
     @Override
     default void close() {
+    }
+
+    /**
+     * Returns the base URI for Keycloak with the scheme, host, port, and context-path set for the given UrlType
+     *
+     * @param originalUriInfo the original URI
+     * @param type type of the request
+     * @return the base URI
+     */
+    default URI getBaseUri(UriInfo delegate, UrlType type) {
+        String scheme = getScheme(delegate, type);
+        String hostname = getHostname(delegate, type);
+        int port = getPort(delegate, type);
+        String contextPath = getContextPath(delegate, type);
+        return delegate.getBaseUriBuilder().scheme(scheme).host(hostname).port(port).replacePath(contextPath).build();
     }
 
 }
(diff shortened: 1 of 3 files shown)


mabartos (dismissed):
LGTM

mabartos:
@vmuzikar These changes are quite straightforward and the maintainer's approval is needed. Could you please handle that? Thanks!

server-spi/src/main/java/org/keycloak/urls/HostnameProvider.java
```diff
@@ -119,4 +121,19 @@ default String getContextPath(UriInfo originalUriInfo) {
...
     }
 
+    /**
+     * Returns the base URI for Keycloak with the scheme, host, port, and context-path set for the given UrlType
+     *
+     * @param originalUriInfo the original URI
+     * @param type type of the request
+     * @return the base URI
+     */
+    default URI getBaseUri(UriInfo delegate, UrlType type) {
```
vmuzikar: Possibly, one nit pick. To be consistent with other methods here, we could have additional method signature as `getBaseUri(UriInfo delegate)` (without the `UrlType`). But it's not blocking, feel free to keep as is.

server-spi/src/main/java/org/keycloak/urls/HostnameProvider.java
```diff
@@ -119,4 +121,19 @@ default String getContextPath(UriInfo originalUriInfo) {
...
+        String scheme = getScheme(delegate, type);
+        String hostname = getHostname(delegate, type);
+        int port = getPort(delegate, type);
+        String contextPath = getContextPath(delegate, type);
+        return delegate.getBaseUriBuilder().scheme(scheme).host(hostname).port(port).replacePath(contextPath).build();
```
vmuzikar: Should we rename the parameter to keep it consistent with the javadoc as well as other methods?
```suggestion
    default URI getBaseUri(UriInfo originalUriInfo, UrlType type) {
        String scheme = getScheme(originalUriInfo, type);
        String hostname = getHostname(originalUriInfo, type);
        int port = getPort(originalUriInfo, type);
        String contextPath = getContextPath(originalUriInfo, type);
        return originalUriInfo.getBaseUriBuilder().scheme(scheme).host(hostname).port(port).replacePath(contextPath).build();
```

vmuzikar (commented):
@shawkins Sorry for the late reply. I wanted to dedicate some time to think about this as I remember attempting similar changes some time ago, but I pulled that back as it had some unexpected consequences. These changes might be different though – I couldn't find any issue with them.

Added two minor comments.

server-spi/src/main/java/org/keycloak/urls/HostnameProvider.java
shawkins: It seems best to just leave that off - having the methods with a type call the ones without by default seems odd.

vmuzikar (approved):
@shawkins LGTM, thanks.

PR #43336, opened 2025-10-09
Prevent using JTA transaction when initializing JDBC_PING

ahus1:
Closes #43335


ahus1:
@pruivo - please review when you have the time. Thanks!

pruivo (approved):

PR #43338, opened 2025-10-09
Register new protocols to avoid exceptions on startup

ahus1:
Closes #43337


ahus1:
@pruivo - please review when you have the time: Thanks!

pruivo (approved):
This is a bit of a waste since the client sessions cannot be shared between KC 26.2 and 26.4.

ahus1:
> This is a bit of a waste since the client sessions cannot be shared between KC 26.2 and 26.4.

At least they will form a cluster ... otherwise I've seen the cluster membership breaking.

pedroigor (approved):

Issue #36514, opened 2025-01-16
The organization claim does not appear if the Organization Membership Mapper is added through a custom client scope


bdovaz:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

organizations

### Describe the bug

The organization claim does not appear if the "Organization Membership Mapper" is added through a custom client scope.

It only appears if the client scope “organization” is added separately in addition to the custom one.

### Version

26.1.0

### Regression

- [ ] The issue is a regression

### Expected behavior

It should be possible to generate the claim by creating a custom client scope with a mapper of the type “Organization Membership Mapper” without being forced to use the client scope “organization” that is already included with Keycloak.

### Actual behavior

The claim does not appear if the client scope “organization” included in Keycloak is not added.

### How to Reproduce?

1. Create a client scope with a single mapper of type “Organization Membership Mapper” with default values.
2. Create a client and add the previous client scope as “Default”.
3. In client scopes go to “Evaluate” and check how the claim does not appear.
4. Add to the client the client scope “organization” (the “official” one of Keycloak).
5. Repeat step 3 and in this case check if the 2 claims appear (the one of our client scope and the one of “organization”).

bdovaz:
> Thank you [@bdovaz](https://github.com/bdovaz) for reporting the issue, may I ask you to provide a little bit more fine grained steps to reproduce the issue? I'd like to be sure I understand the issue correctly. Thank you
> 
> ~missing-info

What exactly is not understood? I can edit the “How to Reproduce?” section, but I don't know what I need to clarify.

pedroigor:
@bdovaz The steps are clear. Can you elaborate on why you want a scope name other than `organization` when using this mapper?

bdovaz:
@pedroigor regardless of how we use the organization client scope or organization membership mapper, what I have found is a reproducible bug.

In our case we add in our own client scopes the organization mapper because depending on the client where we use it we are interested (or it is a requirement for the particularities of the client of the format of the claim that requires) to configure it in different way. For example, enabling or disabling the options to add the organization id.

pedroigor:
@bdovaz @vramik Sent a draft https://github.com/keycloak/keycloak/pull/36568.

ahus1:
@pedroigor - I removed the two release labels and added the backport labels as the backport PRs are not merged yet. The labeler action will update the backport label with the release label once the PR is merged.

PR #43339, opened 2025-10-09
Resolve scope

### LINKED ISSUE #42836: Organization selection changes after token refresh
### Before reporting an issue

### Area

organizations

### Describe the bug

If two or more concurrent tokens are active for the same user/client session (ie: OIDC across two tabs in the same browser for the same client), the selected organization will switch to the most recently authenticated as soon as the refresh token is used to generate a new access token.

Relates to #35935.
- the fix enabled two concurrent auth sessions to track different orgs for the same user/client session, however this does not persist across refresh of the access token.

In practice, the symptom looks like this.
- A user of an SPA authenticates in a browser - selecting orgA.   Starts interacting with the application for tenant A.
- User opens a new tab, upon reauthentication selects orgB.   Starts interacting with the application for tenant B.
- After 5 min (default access token expiry), the first tab will refresh its token (silent refresh).  The application is now in an inconsistent state where the selection has suddenly changed to tenant B.

### Version

26.3.3

### Regression

### Expected behavior

Refreshing an access token should retain the same organization selection that existed before the refresh

The mapper should respect the requested scope (assuming the org still exists, following "Requires User Membership" behavior)
- for `grant_type: refresh_token`, 
  - if the optional `scope` isn't provided, then use the scope from the `refresh_token`
  - if the optional `scope` is provided, use the requested scope instead `organization:XXX`.  the scope of the new tokens should be consistent.

### Actual behavior

Refreshing an access token will switch the organization selection to the most recently selected for the client/user session.    
* The mapper pulls the selection from the clientSession note.

### How to Reproduce?

Having a realm with organizations enabled
- create organization foo and bar 
- create a public client
- add a realm user w/ credentials
- add realm user to both orgs foo and bar

Retrieve two concurrent access tokens for the realm user `protocol/openid-connect/auth`
- only scope differs between the two
1) `openid organization:foo`
2) `openid organization:bar`

Both sets of access token/refresh token will contain both the respective requested scope and org selection
== GOOD

Refresh the token authenticated for foo (1) - `protocol/openid-connect/token` with `grant_type: refresh_token`
- Inspect the access token - scope is retained `organization:foo`, but selected org SWITCHED to bar
```
  "scope": "openid organization:foo",
  "organization": {
    "bar": {
      "id": "719193d1-08d5-42ad-838c-abbf8ddede4f"
    }
  },
``` 
== BAD

### Anything else?

While the repro steps involve using the `organization:xxx` scope to select the organization, using the `organization` scope to prompt organization selection via the browser flow exhibits the same unexpected behavior.

In this scenario, the requested scope cannot be used to infer the selection within the `refresh_token` as it will be simply `organization`.
- one possibility, though not sure this is possible from a mapper, is to rewrite the organization scope `organization` to be the more specific selected scope `organization:foo`
- another possibility might be to persist a new claim to hold the selection in the refresh token and always use this instead

summarizing,
- regardless of  whether the org selection was made directly via scope or during the browser flow, the accessToken and its refreshes (via refreshToken) should persist this selection (assuming it stays valid)

### LINKED ISSUE #42839: UserInfo endpoint returns incorrect organization claim
### Before reporting an issue

### Area

organizations

### Describe the bug

`protocol/openid-connect/userinfo` incorrectly returns the latest org selection and not that associated with the authorizing accessToken.

root cause is the same as #42836

### Version

26.3.3

### Regression

### Expected behavior

The organization claim in the response should match the access token

### Actual behavior

The organization claim in the response does NOT match the access token

### How to Reproduce?

Having a realm with organizations enabled
- create organization foo and bar
- create a public client
- add a realm user w/ credentials
- add realm user to both orgs foo and bar

Retrieve two concurrent access tokens for the realm user `protocol/openid-connect/auth`
- only scope differs between the two
1) openid organization:foo
2) openid organization:bar

use the access token from (1) as BEARER to retrieve userinfo
- GET `protocol/openid-connect/userinfo`

### Anything else?

_No response_

Tungsten78:
Resolves #42836
Resolves #42839

- OrganizationAuthenticator.java -> ANY scope organization selection will be transformed to SINGLE
- OrganizationMembershipMapper.java -> resolve organization from the scope instead of client session note
- OrganizationScopes.java -> update name resolves to account for all possible token refresh scenarios

SINGLE Scope Tests (Previous: organization:foo)
- SINGLE → ANY: organization:foo + request organization → Result: organization:foo
- SINGLE → ALL: organization:foo + request organization:* → Result: organization:foo
- SINGLE → Different SINGLE: organization:foo + request organization:bar → Result: null
- SINGLE → Same SINGLE: Not explicitly shown, but would work via equality check

ALL Scope Tests (Previous: organization:*)
- ALL → SINGLE: organization:* + request organization:foo → Result: organization:foo
- ALL → Different SINGLE: organization:* + request organization:bar → Result: organization:bar
- ALL → ANY: organization:* + request organization → Result: null

Additional SINGLE Tests (Previous: organization:bar)
- SINGLE → ANY: organization:bar + request organization → Result: organization:bar
- SINGLE → ALL: organization:bar + request organization:* → Result: organization:bar
- SINGLE → Different SINGLE: organization:bar + request organization:foo → Result: null

ANY Scope tests
- N/A - the browser flow transforms the scope to SINGLE (see single cases above)

### DIFF
--- a/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
+++ b/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
@@ -27,7 +27,6 @@ import jakarta.ws.rs.core.MultivaluedHashMap;
 import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 
-import org.keycloak.OAuth2Constants;
 import org.keycloak.WebAuthnConstants;
 import org.keycloak.authentication.AuthenticationFlowContext;
 import org.keycloak.authentication.AuthenticationFlowError;
@@ -161,8 +160,6 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
         // make sure the organization is set to the session to make it available to templates
         session.getContext().setOrganization(organization);
 
-        updateUserSelectOrganizationScope(authSession, organization.getAlias());
-
         if (isMembershipRequired(context, organization, user)) {
             return;
         }
@@ -187,20 +184,6 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
             context.success();
         } else {
             attempted(context, username);
-        }
-    }
-
-    private void updateUserSelectOrganizationScope(AuthenticationSessionModel authSession, String orgId) {
-        if (!OrganizationScope.ANY.equals(OrganizationScope.valueOfScope(session))) {
-            // there was no user selection, nothing to do
-            return;
-        }
-
-        String scopeParam = authSession.getClientNote(OAuth2Constants.SCOPE);
-        if (scopeParam != null && scopeParam.contains("organization")) {
-            // Transform generic "organization" to specific "organization:orgId"
-            String transformedScope = scopeParam.replaceAll("\\borganization(?!:)\\b", "organization:" + orgId);
-            authSession.setClientNote(OAuth2Constants.SCOPE, transformedScope);
         }
     }
 
--- a/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
+++ b/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
@@ -27,6 +27,7 @@ import jakarta.ws.rs.core.MultivaluedHashMap;
 import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 
+import org.keycloak.OAuth2Constants;
 import org.keycloak.WebAuthnConstants;
 import org.keycloak.authentication.AuthenticationFlowContext;
 import org.keycloak.authentication.AuthenticationFlowError;
@@ -160,6 +161,8 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
         // make sure the organization is set to the session to make it available to templates
         session.getContext().setOrganization(organization);
 
+        updateUserSelectOrganizationScope(authSession, organization.getAlias());
+
         if (isMembershipRequired(context, organization, user)) {
             return;
         }
@@ -184,6 +187,20 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
             context.success();
         } else {
             attempted(context, username);
+        }
+    }
+
+    private void updateUserSelectOrganizationScope(AuthenticationSessionModel authSession, String orgId) {
+        if (!OrganizationScope.ANY.equals(OrganizationScope.valueOfScope(session))) {
+            // there was no user selection, nothing to do
+            return;
+        }
+
+        String scopeParam = authSession.getClientNote(OAuth2Constants.SCOPE);
+        if (scopeParam != null && scopeParam.contains("organization")) {
+            // Transform generic "organization" to specific "organization:orgId"
+            String transformedScope = scopeParam.replaceAll("\\borganization(?!:)\\b", "organization:" + orgId);
+            authSession.setClientNote(OAuth2Constants.SCOPE, transformedScope);
         }
     }
 
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
@@ -794,6 +794,42 @@ public class OrganizationOIDCProtocolMapperTest extends AbstractOrganizationTest
         assertThat(organizations.contains(orgB.getAlias()), is(true));
     }
 
+    @Test
+    public void testRefreshTokenScopeWithSingleOrganizationAskingForAll() {
+        OrganizationRepresentation orgA = createOrganization("orga", true);
+        MemberRepresentation member = addMember(testRealm().organizations().get(orgA.getId()), "member@" + orgA.getDomains().iterator().next().getName());
+        OrganizationRepresentation orgB = createOrganization("orgb", true);
+        testRealm().organizations().get(orgB.getId()).members().addMember(member.getId()).close();
+        oauth.client("broker-app", "broker-app-secret");
+        String originalScope = "organization:" + orgA.getAlias();
+        oauth.scope(originalScope);
+        loginPage.open(bc.consumerRealmName());
+        loginPage.loginUsername(member.getEmail());
+        loginPage.login(memberPassword);
+        AccessTokenResponse response = assertSuccessfulCodeGrant();
+        assertThat(response.getScope(), containsString(originalScope));
+        AccessToken accessToken = oauth.verifyToken(response.getAccessToken());
+        assertThat(accessToken.getScope(), containsString(originalScope));
+        assertThat(accessToken.getOtherClaims().keySet(), hasItem(OAuth2Constants.ORGANIZATION));
+        List<String> organizations = (List<String>) accessToken.getOtherClaims().get(OAuth2Constants.ORGANIZATION);
+        assertEquals( 1, organizations.toArray().length);
+        assertThat(organizations.contains(orgA.getAlias()), is(true));
+        RefreshToken refreshToken = oauth.parseRefreshToken(response.getRefreshToken());
+        assertThat(refreshToken.getScope(), containsString(originalScope));
+
+        //previous:(SINGLE:orga) -> current:(ALL) == SINGLE:orga
+        String allOrgsScope = "organization:*";
+        oauth.scope(allOrgsScope);
+        response = oauth.doRefreshTokenRequest(response.getRefreshToken());
+        assertThat(response.getScope(), not(containsString(allOrgsScope)));
+        accessToken = oauth.verifyToken(response.getAccessToken());
+        assertThat(accessToken.getScope(), containsString(originalScope));
+        assertThat(accessToken.getOtherClaims().keySet(), hasItem(OAuth2Constants.ORGANIZATION));
+        organizations = (List<String>) accessToken.getOtherClaims().get(OAuth2Constants.ORGANIZATION);
+        assertEquals( 1, organizations.toArray().length);
+        assertThat(organizations.contains(orgA.getAlias()), is(true));
+    }
+
     @SuppressWarnings("unchecked")
     @Test
     public void testPasswordGrantWithAllOrganizationsAndRefresh() throws Exception {
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
@@ -606,7 +606,8 @@ public class OrganizationOIDCProtocolMapperTest extends AbstractOrganizationTest
         orgScope = "organization";
         oauth.scope(orgScope).openid(false);
         response = oauth.doRefreshTokenRequest(response.getRefreshToken());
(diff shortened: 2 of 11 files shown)


Tungsten78:
cc @pedroigor  - looking forward to your review.  thanks

Tungsten78:
@pedroigor, @stianst - is there anything else needed on my end to get this bubbled up for review?  thanks

Tungsten78:
@pedroigor, @stianst - apologies for the bump.   hoping to get status on these pending reviews.  thanks.

vramik (approved):
Thank you @Tungsten78 for preparing the PR and sorry for late response.

Changes LGTM, the PR needs a rebase, can you please do it? Then I'll ask someone to run GHA.

services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
```diff
@@ -186,6 +188,20 @@ private void action(AuthenticationFlowContext context, String username) {
...
     }
 
+    private void updateUserSelectOrganizationScope(AuthenticationSessionModel authSession, String orgId) {
+        if (!OrganizationScope.ANY.equals(OrganizationScope.valueOfScope(session))) {
+            // there was no user selection, nothing to do
+            return;
+        }
+
+        String scopeParam = authSession.getClientNote(OAuth2Constants.SCOPE);
+        if (scopeParam != null && scopeParam.contains("organization")) {
```
pedroigor: We also need to take into account scopes using a name other than `organization`. See #36514.

I'll work on it.

services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
```diff
@@ -186,6 +188,20 @@ private void action(AuthenticationFlowContext context, String username) {
...
+            // Transform generic "organization" to specific "organization:orgId"
+            String transformedScope = scopeParam.replaceAll("\\borganization(?!:)\\b", "organization:" + orgId);
```
pedroigor: Can you elaborate on why we want to change the scope name? It makes more sense to me to keep the same name used by the client. If the application wants to check the selected organization, it can check the  `organizations` claim in the token.

Or perhaps, if really needed, we can update the `scope` parameter string to add the `organization:<alias>`.

Wdyt?

pedroigor (commented):
@Tungsten78 I rebased your PR and changed the following:

* Make sure we the scope name is not hardcoded to `organization`. See https://github.com/keycloak/keycloak/pull/43339#discussion_r2981374433
* Updated docs, in case we really want to proceed with replacing `organization` with `organization:<alias>` when the user selects an org.

I'm not sure about this behavior you are introducing. Can you please check https://github.com/keycloak/keycloak/pull/43339#discussion_r2981697042?

The above is my only concern. The doc update is solely due to this behavior change. The main fix is a great improvement!

vramik:
@pedroigor, I think (and please correct me if I am wrong) there's no per-token storage mechanism in client session model. The scope string is the only per-token data the mapper can access.

  Token refresh flow:                                                                                                                                                                           
    1. Old token has scope "organization" (ANY)
    2. Mapper needs to produce the organization claim
    3. Mapper calls resolveFromRequestedScopes() → scope is ANY
    4. ANY resolution reads kc.org note → gets WRONG org (the last-selected one)
    
 The mapper produces the claim — it can't read back from the claim it's about to generate. During refresh, the only inputs available to the mapper are:                                        
  - the scope string (per-token) is unique per token                                                                                                                                             
  - the shared kc.org client session note - always last selected

Alternatively we could introduce "per-token notes in client session" to keep the original scope name, or perhaps there is better solution.

pedroigor:
@vramik During refresh, you should not be able to switch orgs if you explicitly selected one when authenticating.

vramik:
@pedroigor I think it's not about intentionally swithich orgs during refresh, but about having more than one tab opened.

- if user opens tab1 and logs in with `organization:foo`, `kc.org` will store `{foo.id}`
- if he opens tab2 and logs in with `organization:bar`, `kc.org` will store `{bar.id}`
- when tab1 token expires: refresh occurs and the mapper reads `{bar.id}` instead of `{foo.id}`.

Tungsten78:
Thank you both for your attention on this.

regarding the rewriting of the `organization` scope to `organization:{selected}`.
- @vramik has it correct, during refresh this is how the mapper identifies which organization is being refreshed.

It was called out in the [issue](https://github.com/keycloak/keycloak/issues/42836):

> Anything else?
> While the repro steps involve using the organization:xxx scope to select the organization, using the organization scope to prompt organization selection via the browser flow exhibits the same unexpected behavior.
> 
> In this scenario, the requested scope cannot be used to infer the selection within the refresh_token as it will be simply organization.
> 
> one possibility, though not sure this is possible from a mapper, is to rewrite the organization scope organization to be the more specific selected scope organization:foo
> another possibility might be to persist a new claim to hold the selection in the refresh token and always use this instead

Tungsten78:
> @vramik During refresh, you should not be able to switch orgs if you explicitly selected one when authenticating.

IIRC, this is covered by this test case
SINGLE → Different SINGLE: organization:bar + request organization:foo → Result: null

pedroigor:
@vramik @Tungsten78 I see now the reasoning for rewriting the scope session note. Working on an update.

pedroigor:
@vramik @Tungsten78 Updated the PR to keep track of the selected organization using a claim within the RT. It should allow us to avoid rewriting the scope parameter and keep the scopes the same as requested by clients.

I also introduced some additional validations when refreshing tokens and mapping the claim to make sure the org is enabled and the user is a member.

Can you please review?

vramik (commented):
@pedroigor thanks for the update.

The refresh flow fix LGTM.

I'd like to verify that #42839 (UserInfo) is also fixed. 

It seems the mapper still uses the shared note. 

We could either:
- in `UserInfoEndpoint`, before calling the mapper, resolve the org from the access token's organization claim and set it in context
- in the mapper, try `resolveFromRequestedScopes()` before the `kc.org` note

wdyt?

Also it'd be worth to cover the UserInfo (#42839) in the tests

vramik (changes requested):

pedroigor:
@vramik Updated with the fix for userinfo as well as the introspection endpoint. Can you please re-review?

I think we can improve token transformation (running mappers) by passing to the mappers the bearer token (RT, access token, etc) used to authenticate the user. It would allow us to have the `validateSelectedOrganization` within the organization mapper boundaries and not exposed to `TokenManager. We can do this as a follow-up.

vramik:
@pedroigor thanks for the update. Looking now at `validateSelectedOrganization`. 

I wonder what happens when the token has multiple orgs (`ALL`). Setting org into context is skipped, right?

Should not we then change the order of the resolution in mapper's `setClaim` to check scopes before the `kc.org` note?

>I think we can improve token transformation (running mappers) by passing to the mappers the bearer token (RT, access token, etc) used to authenticate the user. It would allow us to have the validateSelectedOrganization within the organization mapper boundaries and not exposed to `TokenManager. We can do this as a follow-up.

+1, would you create an issue for it please, so we do not loose track of it.

Tungsten78:
@pedroigor - LGTM

thanks.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
```diff
@@ -560,21 +739,217 @@ public void testRefreshTokenScopeAnyAskingSingleOrganization() {
...
+        assertThat(response.getScope(), containsString("organization"));
+        accessToken = TokenVerifier.create(response.getAccessToken(), AccessToken.class).getToken();
+        assertThat(accessToken.getOtherClaims().keySet(), hasItem(OAuth2Constants.ORGANIZATION));
+        organizations = (List<String>) accessToken.getOtherClaims().get(OAuth2Constants.ORGANIZATION);
+        assertThat(organizations, containsInAnyOrder("orga", "orgb"));
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testMultipleTabsWithMixedScopeFormats() {
```
pedroigor, line 871: @vramik Added this test method to cover cross-tab contamination (shared client session) and refreshing tokens with scopes other than the original ones, including checking if the scopes from the last successful refresh are respected if you refresh a token without providing scopes.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
```diff
@@ -560,21 +739,217 @@ public void testRefreshTokenScopeAnyAskingSingleOrganization() {
...
+            tabUtil.switchToTab(0);
+            oauth.scope("organization:" + orgA.getAlias());
+            response = oauth.doRefreshTokenRequest(tab1RefreshToken);
+            assertThat(response.getScope(), containsString("organization:" + orgA.getAlias()));
+            accessToken = oauth.verifyToken(response.getAccessToken());
+            organizations = (List<String>) accessToken.getOtherClaims().get(OAuth2Constants.ORGANIZATION);
+            assertThat(organizations.size(), is(1));
+            assertThat(organizations, containsInAnyOrder(orgA.getAlias()));
+
+            // try to refresh first tab changing scopes SINGLE -> ANY, not allowed
```
pedroigor, line 946: Perhaps we should throw an error?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
```diff
@@ -560,21 +739,217 @@ public void testRefreshTokenScopeAnyAskingSingleOrganization() {
...
+            tabUtil.switchToTab(0);
+            oauth.scope("organization:*");
+            response = oauth.doRefreshTokenRequest(tab1RefreshToken);
+            assertThat(response.getScope(), containsString("organization:*"));
+            accessToken = oauth.verifyToken(response.getAccessToken());
+            organizations = (List<String>) accessToken.getOtherClaims().get(OAuth2Constants.ORGANIZATION);
+            assertThat(organizations.size(), is(2));
+            assertThat(organizations, containsInAnyOrder(orgA.getAlias(), orgB.getAlias()));
+
+            // try to refresh second tab changing scopes ANY -> ALL
```
pedroigor, line 963: Not sure if we need to consider this now .... If the user opted in for a specific organization during authentication, should the client be allowed to step in and pick another org(s)?

The samething if `SINGLE` to `ALL`.

We should narrow scopes ....

vramik (approved):

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/organization/mapper/OrganizationOIDCProtocolMapperTest.java
vramik: I agree, once the scope is narrowed we should deny widening it back. 

Follow-up or are you incorporating it here?

pedroigor: I'm adding here.

pedroigor: @vramik I misread this test method. We are covering switching to `ALL` at `testRefreshTokenScopeWithOrganizationSelectionAskingForAll` and I'm adding another (explicit) one for `SINGLE` -> `ALL`.

pedroigor:
@vramik Updated to throw an error when trying to ask for `ANY` from `ALL`.

pedroigor (approved):

PR #43354, opened 2025-10-10
Make pending email verification attribute removable by admin

### LINKED ISSUE #43351: Make pending email verification attribute removable by admin
### Description

There is a pending email verification attribute in admin ui  rendered in user's details section. Currently it is editable by admin. Admin should be able to only remove the attribute which should cancel the email verification process and invalidate the action token. 
The attribute should be rendered just after email attribute.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

martin-kanis:
Closes #43351


ssilvert (changes requested):
@martin-kanis This is only meant for the admin UI, right?

the `ui-shared` library is meant for code that is common to both admin console and account console.  So I'm a little hesitant to add all this code to the library.  Can you think of a way to do this were the code that only applies to admin console stays in the `admin-ui` code base?

pedroigor:
> the `ui-shared` library is meant for code that is common to both admin console and account console. So I'm a little hesitant to add all this code to the library. Can you think of a way to do this were the code that only applies to admin console stays in the `admin-ui` code base?

The first option was to just allow the admin to clear the value (deleting text) and then click `Save`. The goal here is to allow the admin to clear this field so that the attribute holding the email pending for confirmation is removed.

pedroigor (dismissed):
Thanks, @martin-kanis !

js/apps/admin-ui/src/user/UserForm.tsx
```diff
@@ -240,9 +267,37 @@ export const UserForm = ({
...
               labelIcon={t("emailVerifiedHelp")}
             />
+            {user?.attributes?.["kc.email.pending"] && (
+              <Alert
+                variant={AlertVariant.warning}
+                isInline
+                isPlain
+                title={t("emailPendingVerificationAlertTitle")}
+              >
+                The user has not confirmed the new email address{" "}
```
ssilvert: This needs to be localized.

ssilvert (changes requested):

js/apps/admin-ui/src/user/UserForm.tsx
pedroigor: Fixed. I did not find a way to bold the email in the message property. Is it possible to do that?

pedroigor (approved):

ssilvert (approved):

PR #43359, opened 2025-10-10
Client session may be lost during session restart

### LINKED ISSUE #43349: Client session may be lost during session restart
### Before reporting an issue

### Area

infinispan

### Describe the bug

When a user session is restarted, the client session may also restart. This operation is not saved, and the client session is not persisted. 

Follow-up of https://github.com/keycloak/keycloak/issues/43161

### Version

main

### Regression

### Expected behavior

The user and client sessions are properly restarted and stored. 

### Actual behavior

The user session is working properly due to https://github.com/keycloak/keycloak/issues/43161, but the client session is not. 

### How to Reproduce?

Update the test below by restarting the client session from `test-app`. 

https://github.com/keycloak/keycloak/blob/a19f4d00fcc04b25f923643046f35857dcec52f4/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/model/UserSessionProviderTest.java#L141-L171

### Anything else?

_No response_

pruivo:
Fixes #43349

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/AuthenticatedClientSessionAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/AuthenticatedClientSessionAdapter.java
@@ -18,6 +18,7 @@
 package org.keycloak.models.sessions.infinispan;
 
 import java.util.Map;
+import java.util.Objects;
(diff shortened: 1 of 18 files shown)


pruivo:
@ahus1, when you have the time, can you take a look? The testsuite is finally green :smile:

ahus1 (approved):
Thank you for this PR!

PR #43361, opened 2025-10-10
Add session information to the current context on creation and sessio…

ahus1:
Closes #43360


ahus1:
@pruivo - please review when you have the time. I'm ok to search for a different reviewer if you think you are not familiar with the code. Or I explain it in a short call.

pruivo (approved):

pedroigor (approved):

PR #43363, opened 2025-10-10
Remove extra flush events to increase performance

### LINKED ISSUE #43362: Reduce the number entity manager flushes when creating a realm
### Description

When doing a performance test, I found multiple flushes when creating a realm. 

Those flushes are only needed when they are followed by a clear or when they are removing the entity. In all other cases they are wasting resources. 

They prevent batching of writes to the database, and spend extra CPU cycles inspecting the persistence context. 

### Discussion

_No response_

### Motivation

Faster creation of realms and clients. 

### Details

I'll create a PR

ahus1:
Closes #43362

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java
@@ -1320,24 +1320,27 @@ public class JpaRealmProvider implements RealmProvider, ClientProvider, ClientSc
         Map<String, ClientScopeModel> existingClientScopes = getClientScopes(realm, client, true);
         existingClientScopes.putAll(getClientScopes(realm, client, false));
 
-        clientScopes.stream()
-            .filter(clientScope -> !existingClientScopes.containsKey(clientScope.getName()))
-            .filter(clientScope -> {
-                if (clientScope.getProtocol() == null) {
-                    // set default protocol if not set. Otherwise, we will get a NullPointer
-                    clientScope.setProtocol(OIDCLoginProtocol.LOGIN_PROTOCOL);
-                }
-                return acceptedClientProtocols.contains(clientScope.getProtocol());
-            })
-            .forEach(clientScope -> {
-                ClientScopeClientMappingEntity entity = new ClientScopeClientMappingEntity();
-                entity.setClientScopeId(clientScope.getId());
-                entity.setClientId(client.getId());
-                entity.setDefaultScope(defaultScope);
-                em.persist(entity);
-                em.flush();
-                em.detach(entity);
-            });
+        Set<ClientScopeClientMappingEntity> clientScopeEntities = clientScopes.stream()
+                .filter(clientScope -> !existingClientScopes.containsKey(clientScope.getName()))
+                .filter(clientScope -> {
+                    if (clientScope.getProtocol() == null) {
+                        // set default protocol if not set. Otherwise, we will get a NullPointer
+                        clientScope.setProtocol(OIDCLoginProtocol.LOGIN_PROTOCOL);
+                    }
+                    return acceptedClientProtocols.contains(clientScope.getProtocol());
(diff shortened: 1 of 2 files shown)


ahus1:
@shawkins - may I ask you to review this PR, or should I ask a different team? Thanks!

model/jpa/src/main/java/org/keycloak/models/jpa/RealmAdapter.java
```diff
@@ -1854,7 +1854,6 @@ public RequiredActionProviderModel addRequiredActionProvider(RequiredActionProvi
         action.setPriority(model.getPriority());
         realm.getRequiredActionProviders().add(action);
         em.persist(action);
-        em.flush();
```
shawkins, line 1857: This seems fine. In general there are a lot of flushes in RealmAdapter that seem ad hoc beyond the pattern of persist / flush / detach.

model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java
```diff
@@ -1320,24 +1320,27 @@ public void addClientScopes(RealmModel realm, ClientModel client, Set<ClientScop
...
+                    entity.setClientScopeId(clientScope.getId());
+                    entity.setClientId(client.getId());
+                    entity.setDefaultScope(defaultScope);
+                    em.persist(entity);
+                    return entity;
+                }).collect(Collectors.toSet());
+        if (!clientScopeEntities.isEmpty()) {
+            em.flush();
+            clientScopeEntities.forEach(entity -> em.detach(entity));
+        }
```
shawkins, line 1343: In situations where you have less direct control over the transitive calls, the pattern using the EntityManagers logic looks like:

```suggestion
        EntityManagers.runInBatch(session, () -> {
            clientScopes.stream()
            .filter(clientScope -> !existingClientScopes.containsKey(clientScope.getName()))
            .filter(clientScope -> {
                if (clientScope.getProtocol() == null) {
                    // set default protocol if not set. Otherwise, we will get a NullPointer
                    clientScope.setProtocol(OIDCLoginProtocol.LOGIN_PROTOCOL);
                }
                return acceptedClientProtocols.contains(clientScope.getProtocol());
            })
            .forEach(clientScope -> {
                ClientScopeClientMappingEntity entity = new ClientScopeClientMappingEntity();
                entity.setClientScopeId(clientScope.getId());
                entity.setClientId(client.getId());
                entity.setDefaultScope(defaultScope);
                em.persist(entity);
            });
        }, true);
```

However that will generally result in 2 flushes though, rather than the single flush that is done here.

shawkins (approved):
LGTM, just added a couple of general thoughts. Ideally we can revisit the excess flushing more systematically. If I recall it was roughly doubling the time that it takes to perform an import.

ahus1:
@shawkins - thank you for reviewing; this was just about some small improvements I found while working on something else.

shawkins:
> @shawkins - thank you for reviewing; this was just about some small improvements I found while working on something else.

Yes, that is clear from the issue. Just thought it was worth mentioning how broadly flush calls are being used, and another approach that also prevents the accumulation of entities in the peristence context.

ahus1:
I'll handle this another time when I have more time to dive into this.

ahus1:
@mposolda / @pedroigor / @vramik - as we've briefly discussed it in the chat today: These seem to be some small changes and the CI still passes. So I want to proceed with this changes, but don't introduce any other changes beyond that. 

This is an attempt to make use of some analysis I did to test how Keycloak performs with a lot of realms, so I didn't want to waste it as the result was already there. 

Please let me know if you agree, and if this PR should be merged as is. Any additional work should be tackled in a different issue and should be planned for separately.

vramik (approved):
LGTM, thank you @ahus1 for preparing the PR

pedroigor (approved):

PR #43366, opened 2025-10-10
Mark the reading of admin and user events read-only

ahus1:
This should decrease the memory usage and improve response times

Closes #43365


ahus1:
@shawkins - Would you be willing to review this PR? Thanks!

stianst (approved):

Issue #43165, opened 2025-10-02
Client API v2


vmuzikar:
### Description

The new Admin API v2 will be split into several individual APIs. This issue covers the Client API which will be the first part of Admin API v2 to be implemented.

The new API will be adopted by any first party clients: Operator CR, Java client, CLI, Javascript client, Admin Console, ...

As part of the efforts, a blueprint will be created that will serve as a guideline for creating any following Admin APIs to ensure consistency across the APIs. The blueprint will be updated as the Client API (and its clients) gets implemented.

### Value Proposition

A new interface that will allow managing Clients in a consistent, user friendly and well-documented way, and supports declarative configuration.

Additional value will be a blueprint created alongside the Client API and will be used when designing any future Admin APIs.

### Goals

* New REST API for managing Clients.
* First party clients that consume this API. This includes Java and Javascript clients, Operator Client CR, Terraform Provider.
* Blueprint capturing the generics of the design.

### Non-Goals

* The new REST API to implement all functionality the original one had.

### Discussion

https://github.com/keycloak/keycloak/discussions/categories/api-guidelines

dvob:
@vmuzikar I saw that as part of this feature a MVP Client CR got created. Are there plans to further refine this CR? If yes, where gets this tracked/discussed? Maybe here https://github.com/keycloak/keycloak/issues/45439 ?

We would like to use such an operator and let users provision OIDC Clients via Manifests as part of their deployment manifests.
However, as far as I understand the current implementation, it does not distinguish between creation and updates but use createOrUpdate in both cases. This could lead to the problem that two resources start to manage the same Keycloak client.

I would suggest the following approach:
- Create the client with an API call which fails if the client already exists
- On creation keep track of the owning Kubernetes resource in the attributes of the client (e.g. `operator.owner.namespace`, `operator.owner.name`, `operator.owner.instance` (for the case where you have multiple operators acting on the same Keycloak instance)).
- Then before you start to manage the client (UPDATE, DELETE) you always first check if you own the client

Also it can make sense to keep track of the ownership in the  status of the Kubernetes resource as well. This could work as follows:
- After creation you store the id of the client in the status. Also see https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#representing-allocated-values
- On subsequent reconciles you check whether there already is a stored ID in status. In that case you know that you've created the resource and can start to update right away and don't have to check the ownership with a seperate GET every time.

[...]

PR #43382, opened 2025-10-12
Client Resource must return a single item (#39851)

cherifyazid:
We have such a method org.keycloak.admin.client.resource.ClientsResource.findByClientId(String) that should return a single item, but it returns a List of ClientRepresentation.
 
for more details please check the issue description

Closes #39851

Signed-off-by: cherifyazid <yadzinov@gmail.com>

### DIFF
--- a/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/ClientsResource.java
+++ b/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/ClientsResource.java
@@ -18,10 +18,12 @@
 package org.keycloak.admin.client.resource;
 
 import java.util.List;
+import java.util.Optional;
 
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.PathParam;
@@ -60,9 +62,24 @@ public interface ClientsResource {
                                                  @QueryParam("first") Integer firstResult,
                                                  @QueryParam("max") Integer maxResults);
 
+    /** @deprecated use {@link #findClientByClientId(String)} */
+    @Deprecated
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     List<ClientRepresentation> findByClientId(@QueryParam("clientId") String clientId);
+
+    default Optional<ClientRepresentation> findClientByClientId(String clientId) {
+        List<ClientRepresentation> list = findByClientId(clientId);
+        if (list.isEmpty()) {
+            return Optional.<ClientRepresentation>empty();
+        }
+        return Optional.of(list.get(0));
+    }
+
+    default ClientResource getByClientId(String id) {
+        return get(findClientByClientId(id).map(ClientRepresentation::getId)
+                .orElseThrow(() -> new NotFoundException("client not found " + id)));
+    }
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
--- a/test-framework/examples/tests/src/test/java/org/keycloak/test/examples/RealmSpecificAdminClientTest.java
+++ b/test-framework/examples/tests/src/test/java/org/keycloak/test/examples/RealmSpecificAdminClientTest.java
@@ -47,10 +47,7 @@ public class RealmSpecificAdminClientTest {
     public void testRealmWithClientAndUser() {
         RealmResource realmResource = realmAdminClient.realms().realm(realm.getName());
 
-        List<ClientRepresentation> clients = realmResource.clients().findByClientId("myclient");
-        Assertions.assertEquals(1, clients.size());
-
-        ClientRepresentation client = clients.get(0);
+        ClientRepresentation client = realmResource.clients().findClientByClientId("myclient").orElseThrow();
         Assertions.assertTrue(client.isEnabled());
         Assertions.assertTrue(client.isDirectAccessGrantsEnabled());
         Assertions.assertEquals("mysecret", client.getSecret());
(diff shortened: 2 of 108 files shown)


integration/admin-client/src/main/java/org/keycloak/admin/client/resource/ClientsResource.java
```diff
@@ -61,7 +61,7 @@ List<ClientRepresentation> findAll(@QueryParam("clientId") String clientId,
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    List<ClientRepresentation> findByClientId(@QueryParam("clientId") String clientId);
+    ClientRepresentation findByClientId(@QueryParam("clientId") String clientId);
```
shawkins: Please see https://github.com/keycloak/keycloak-client/issues/155#issuecomment-2804779827 and use the method name findClientByClientId, then deprecate findByClientId.

test-framework/examples/tests/src/test/java/org/keycloak/test/examples/RealmSpecificAdminClientTest.java
```diff
@@ -46,10 +46,8 @@ public void testAdminClientIssuers() throws JWSInputException {
     public void testRealmWithClientAndUser() {
         RealmResource realmResource = realmAdminClient.realms().realm(realm.getName());
 
-        List<ClientRepresentation> clients = realmResource.clients().findByClientId("myclient");
-        Assertions.assertEquals(1, clients.size());
-
-        ClientRepresentation client = clients.get(0);
+        ClientRepresentation client = realmResource.clients().findByClientId("myclient");
+        assertNotNull(client);
```
bmscomp: Keep the same style of assertion, use static method of `Assertions` class instead of static import, just to unify the way asserts are used 

Prefer `Assertions.assertNotNull(client);`

integration/admin-client/src/main/java/org/keycloak/admin/client/resource/ClientsResource.java
cherifyazid: @shawkins  Thanks for the thorough review and helpful suggestions! I’ve addressed the feedback and pushed updates relating to  **#39851** and **#155** :

- Renamed the API to findClientByClientId(String) and deprecated findByClientId(String)

- Kept the old method as a compat shim delegating to the new one (returns empty/singleton list) ClientTest.findByClientListResult()

- Migrated tests to the new API and standardized assertions (assertNotNull, assertNull, etc.);retained one minimal compat test with @SuppressWarnings("deprecation") 

This is my first PR to Keycloak—happy to take any suggestions or guidance to align with project requirements.

shawkins: > This is my first PR to Keycloak—happy to take any suggestions or guidance to align with project requirements.

@cherifyazid thank you for this pr. Everything looks good - can you also add a DCO sign off to your commit?

cherifyazid: @shawkins Done !

cobotrifork:
I might be wrong here, but just stumbled upon this and saw this kind of snippets in the PR:
```java
ClientRepresentation client = realm.admin().clients().findClientByClientId(attachTo);
if (client != null) {
    throw new IllegalStateException("No client found with client id: " + attachTo);
}
```
Shouldn't the condition check that client is null instead? i.e. `if (client == null)`

shawkins:
> Shouldn't the condition check that client is null instead? i.e. if (client == null)

Absolutely those two instances need to be corrected.

Also @cherifyazid are you able to update / rebase this pr?

cherifyazid:
@cobotrifork  nice catch, correction pushed.
@shawkins

shawkins:
@cherifyazid @cobotrifork took a pass over the changes and made several refinements. 

There are now two new default methods on ClientsResource:

- findClientByClientId returning an Optional ClientRepresentation
- and getByClientId that returns the ClientResource - I changed some obvious locations to using this, but there are probably still more places where this could get rid of some boiler-plate.

To all other reviewers if need this pr can be split apart - just the changes to ClientRepresentation could go in by itself, and the changes to the legacy test classes could be omitted.

mabartos:
Nice work! 

Could you please resolve the conflicts?

shawkins:
> Could you please resolve the conflicts?

It's resolved, but unfortunately this will keep coming up as too many files are touched. If we don't see a path to get it in soon whole, then it will be best to break it up.

shawkins:
@mposolda @lhanusov @vmuzikar @cherifyazid based upon more feedback from @keycloak/cloud-native I'll stop updating this PR. If core-clients would entertain these some of these changes, we'll assign the PR to someone in that team. If not, then this PR and the keycloak / keycloak-client issues can be closed.

vmuzikar:
Just to add a bit more context to what @shawkins stated.

The current Admin REST API suffers from numerous issues and inconsistencies. Instead of trying to navigate the possible fixes while managing breaking changes, we decided to take a step back and redesign the API as a version 2, currently focused and scoped to [Client management](https://github.com/keycloak/keycloak/issues/43165). Hence we're not prioritizing enhancements to the current (v1) API as those will come in v2.

cherifyazid:
@shawkins @vmuzikar @mposolda 

Thanks for the additional context — I fully understand the challenges behind evolving the API and the rationale for stepping back to redesign it as v2.
I’m available to help or contribute in any way that could be useful as this work progresses.

vmuzikar:
@cherifyazid Thanks for understanding. Can we close this PR then?

shawkins:
> @cherifyazid Thanks for understanding. Can we close this PR then?

I'd let @mposolda make that call since he was last to triage the keycloak-client issue. If he or those heavily working on the tests like @lhanusov don't want these changes, then definitely this pr and keycloak and keycloak-client issues can be closed.

cherifyazid:
@shawkins yes you can close the PR @vmuzikar

shawkins:
Closing given the lack of additional feedback and how quickly this will continue to go stale.

If someone wants to resurrect this, then just start with the changes for ClientResource https://github.com/keycloak/keycloak/pull/43382/changes#diff-71bd12667deaafa29a3eebb3308d980f6b088732648f5f083bd2d88f57330421 - without all of the test modifications.

PR #43395, opened 2025-10-13
Initial Client API v2 impl

### LINKED ISSUE #43224: First Client API iteration based on protype
### Description

* Move the prototype from [admin-api-v2 branch](https://github.com/keycloak/keycloak/tree/feature/admin-api-v2) to `main`.
* Guard it behind a Feature flag, marked as experimental.

vmuzikar:
Closes #43224


vmuzikar:
Keeping as individual commits for the sake of review. Will squash it before merge.

common/src/main/java/org/keycloak/common/Profile.java
```diff
@@ -59,6 +59,8 @@ public enum Feature {
 
         ADMIN_API("Admin API", Type.DEFAULT),
 
+        CLIENT_ADMIN_API_V2("Client Admin API v2", Type.EXPERIMENTAL, 2, Feature.ADMIN_API),
```
mabartos, line 62:
```suggestion
        ADMIN_API_V2("Admin API v2", Type.EXPERIMENTAL, 2, Feature.ADMIN_API),
```

It'd be better to have the general Admin API v2 as a feature, but I understand that the lifecycle for the initial client API will be different than for other resources like realms.

mabartos (dismissed):
@vmuzikar LGTM. 

When tests passed, I'm ok to merge it. Some polishing and improvement will be done in follow-up tasks. 

Not sure if these commits should be squashed, as there are some contextually different  pieces of code - but leaving the final decision up to you.

common/src/main/java/org/keycloak/common/Profile.java
vmuzikar: Exactly, we want a granular control. Other think to consider is that we can't have two versions of the same feature enabled at the same time AFAIK, and we want Admin v1 to be enabled together with the v2.

mabartos:
@shawkins @Pepo48 Could you please review it to unblock further work? Thanks!

server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
```diff
@@ -211,6 +212,8 @@ default <T> T getAttributeOrDefault(String attribute, T defaultValue) {
      */
     IdentityProviderStorageProvider identityProviders();
 
+    KeycloakServices services();
```
stianst: Makes no sense to me to expose this here; this is adding yet another way to fetch clients, and seems rather to be specific wrapper around clients for admin APIs only, so why add it here?

mabartos: @stianst Yes, we don't have to expose it in the KeycloakSession directly. Can be done in a follow-up task if possible.

vmuzikar: @stianst Thanks for the feedback. +1 for addressing this as a follow-up. The aim of this PR was to move the prototype mostly as is from the feature branch to main. We will be working on it further after it is merged.

Pepo48 (dismissed):

server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
stianst: I don't think that should be handled as a follow-up; as it would then become a blocker to sort it out before the release, which is not a state we want main to be in.

pom.xml
```diff
@@ -591,6 +600,16 @@
...
+            <dependency>
+                <groupId>org.hibernate.validator</groupId>
+                <artifactId>hibernate-validator</artifactId>
+                <version>${hibernate-validator.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.glassfish.expressly</groupId>
+                <artifactId>expressly</artifactId>
+                <version>${expressly.version}</version>
+            </dependency>
```
stianst: This dependency needs approval from the CNCF since it's EPL 2.0. We can merge without approval from the CNCF, but would need to get an exception at some point, ideally before the release.

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelpAll.approved.txt
```diff
@@ -460,6 +460,17 @@ Export:
                      Set the number of users per file. It is used only if 'users' is set to
                        'different_files'. Default: 50.
 
+OpenAPI configuration (Experimental):
```
stianst: Shouldn't these just be features rather than CLI config options?

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
```diff
@@ -0,0 +1,38 @@
...
+    ClientRepresentation getClient();
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation createOrUpdateClient(@Valid ClientRepresentation client,
+                                              @QueryParam("fieldValidation") FieldValidation fieldValidation);
+
+    @PATCH
+    @Consumes({MediaType.APPLICATION_JSON_PATCH_JSON, CONTENT_TYPE_MERGE_PATCH})
```
stianst, line 34: So we support only application/json-patch+json, but not a basic patch? Seems the wrong way around when it comes to priorities

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
stianst, line 1: Please rename to ClientV2Test or something since that's what it's testing. Maybe also worth putting in it's own package to not mix new/old tests

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminRootV2.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Path("")
+    public AdminApi latestAdminApi() {
+        checkApiEnabled();
+        // we could return the latest Admin API if no version is specified
+        return session.getProvider(AdminApi.class);
+    }
+
+    @Path("v2")
+    public AdminApi adminApi() {
```
stianst, line 28: We explicitly don't want versioning on the Admin API as whole; but rather on each distinct API.

server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
stianst: If it needs to be a follow-up then add an issue and mark it as a blocker for 26.5

mabartos: Created PR to Vasek's branch to have it propagated here: 
- https://github.com/vmuzikar/keycloak/pull/18

EDIT: Done.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
mabartos: Created PR to Vasek's branch to have it propagated here: 
- https://github.com/vmuzikar/keycloak/pull/18

EDIT: Done.

pom.xml
mabartos: Created follow-up task: 
- https://github.com/keycloak/keycloak/issues/43569

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelpAll.approved.txt
vmuzikar: Not really IMHO. It's similar to metrics or health endpoints, we do too have config options for those.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
vmuzikar: > but not a basic patch

That's the `application/merge-patch+json`. We'll need to remove `application/json-patch+json` as that's the "smart" PATCH that is out of scope of the MVP. I'll create a separate commit to remove it.

mabartos: cc: @shawkins

server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
vmuzikar: Merged, thanks.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
vmuzikar: Merged, thanks.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminRootV2.java
mabartos: Created follow-up issue:
- https://github.com/keycloak/keycloak/issues/43571

But we need to discuss it within our SIG.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
vmuzikar: > I'll create a separate commit to remove it.

Actually, decided to do it rather as a follow-up. Let's keep this PR as close as possible to the prototype.

vmuzikar: https://github.com/keycloak/keycloak/issues/43572

mabartos (dismissed):
LGTM as the initial implementation -  the main concerns/issues were addressed and for other minor issues, follow-up tasks were created.

core/pom.xml
```diff
@@ -68,6 +68,10 @@
             <groupId>org.eclipse.microprofile.openapi</groupId>
             <artifactId>microprofile-openapi-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+        </dependency>
```
stianst: Core should not have a dependency on `hibernate-validator`.

What's the reason for `CreateClient` and `CreateClientDefault` to be in core? Is it not just used server-side by the Client v2 implementation?

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
stianst, line 1: Should we consider putting representations for Admin v2 into a separate module instead of dumping it alongside everything else in core?

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelpAll.approved.txt
stianst: Don't get why - we have features that can be enabled/disabled, why is there a need for a separate CLI option to do the same?

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+public class BaseRepresentation {
+
+    @JsonIgnore
+    protected Map<String, Object> additionalFields = new LinkedHashMap<String, Object>();
```
stianst, line 15: Doesn't this mean that all representations can have any fields?

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,244 @@
...
+    @JsonPropertyDescription("Whether this client is enabled")
+    private Boolean enabled;
+
+    @URL
+    @JsonPropertyDescription("URL to the application's homepage that is represented by this client")
+    private String appUrl;
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonPropertyDescription("URLs that the browser can redirect to after login")
+    private Set<@NotBlank @URL(message = "Each redirect URL must be valid") String> appRedirectUrls = new LinkedHashSet<String>();
```
stianst, line 42: Why `appRedirectUrls`? Should be `redirectUris`. It's URIs, not URLs, and no need to prefix this randomly with `app`.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonPropertyDescription("Login flows that are enabled for this client")
+    private Set<@NotBlank String> loginFlows = new LinkedHashSet<String>();
```
stianst, line 46: Old ClientRepresentation has authenticationFlowBindingOverrides as a map. Think the way it works is that you can overwrite each flow type with a different flow. What this does I don't know, but doesn't seem compatible with the current implementation.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Valid
+    @JsonPropertyDescription("Authentication configuration for this client")
+    private Auth auth;
```
stianst, line 50: If `auth` is not included, what happens then?

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,244 @@
...
+    public void setServiceAccount(ServiceAccount serviceAccount) {
+        this.serviceAccount = serviceAccount;
+    }
+
+    @JsonInclude(JsonInclude.Include.NON_ABSENT)
+    public static class Auth {
+
+        @NotNull
+        @JsonPropertyDescription("Whether authentication is enabled for this client")
+        private Boolean enabled;
```
stianst, line 172: Shouldn't be needed I think; if auth is null that should just mean it's not using authentication.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
stianst: On second though I guess this is the enabled flows (authz grants, etc.) for the client. This is probably only related to OIDC though; and we'll probably want to have multiple different client representation formats supported. How will we do that?

If the above assumption is correct then it shouldn't be `loginFlows` as not all flows are login; also it's only considered login if it's an OIDC request, not if it's an OAuth request.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,244 @@
...
+    private Set<@NotBlank String> webOrigins = new LinkedHashSet<String>();
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonPropertyDescription("Roles associated with this client")
+    private Set<@NotBlank String> roles = new LinkedHashSet<String>();
+
+    @Valid
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonPropertyDescription("Service account configuration for this client")
+    private ServiceAccount serviceAccount;
```
stianst, line 63: Should just be a flow that is enabled, see above. It's also problematic to mix client definition and permissions.

quarkus/config-api/src/main/java/org/keycloak/config/OptionCategory.java
```diff
@@ -23,6 +23,7 @@ public enum OptionCategory {
     SECURITY("Security", 120, ConfigSupportLevel.SUPPORTED),
     EXPORT("Export", 130, ConfigSupportLevel.SUPPORTED),
     IMPORT("Import", 140, ConfigSupportLevel.SUPPORTED),
+    OPENAPI("OpenAPI configuration", 150, ConfigSupportLevel.EXPERIMENTAL),
```
stianst: If this is experimental it should be wrapped in a feature

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -197,6 +197,7 @@ class KeycloakProcessor {
             JBossJtaTransactionManagerLookup.class,
             DefaultJpaConnectionProviderFactory.class,
             DefaultLiquibaseConnectionProvider.class,
+            //HibernateValidatorProviderFactory.class,
```
stianst: Why is this here, but commented out?

quarkus/runtime/pom.xml
```diff
@@ -394,7 +413,23 @@
...
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
         <!-- Keycloak Dependencies-->
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>zjsonpatch</artifactId>
```
stianst, line 452: Is this what Quarkus uses for patch? I wouldn't think so? APIs are part of Jakarta AFAIK, but not sure what implementation Quarkus uses.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AdminV2DisabledTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static io.restassured.RestAssured.when;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.not;
+
+/**
+ * @author Vaclav Muzikar <vmuzikar@redhat.com>
+ */
+@DistributionTest(keepAlive = true,
+        containerExposedPorts = {8080, 9000})
+public class AdminV2DisabledTest {
```
stianst: Why is this here and not in the new testsuite alongside the other tests?

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminRootV2.java
stianst: This is really not open to discussion; we need to break the massive Admin API into distinct separate and smaller APIs.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
```diff
@@ -0,0 +1,38 @@
...
+    String CONTENT_TYPE_MERGE_PATCH = "application/merge-patch+json";
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation getClient();
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    ClientRepresentation createOrUpdateClient(@Valid ClientRepresentation client,
```
stianst, line 30: Why is FieldValidation a query parameter? I don't get why someone creating a client should be able to bypass validation simply by setting this.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApiSpi.java
```diff
@@ -0,0 +1,35 @@
+package org.keycloak.admin.api.client;
+
+import org.keycloak.common.Profile;
+import org.keycloak.provider.Provider;
+import org.keycloak.provider.ProviderFactory;
+import org.keycloak.provider.Spi;
+
+public class ClientApiSpi implements Spi {
```
stianst, line 8: Why is there a ClientApiSpi?

Shouldn't it just be AdminApi or something? and allow overwriting what handles the `/client` API?

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminApiSpi.java
```diff
@@ -0,0 +1,35 @@
+package org.keycloak.admin.api;
+
+import org.keycloak.common.Profile;
+import org.keycloak.provider.Provider;
+import org.keycloak.provider.ProviderFactory;
+import org.keycloak.provider.Spi;
+
+public class AdminApiSpi implements Spi {
```
stianst, line 8: Don't get why this is an SPI. How/why would you replace the whole Admin API?

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApiSpi.java
stianst: Really don't like there's now an SPI for clients and another for client. IMO it should be rather an AdminApiSpi that has a path; with an option to disable and overwrite built-in providers, and it should be the whole API, not individual parts of it. So there should be something like `ClientAdminApi implements AdminApiProvider`.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/realm/DefaultRealmsApiFactory.java
```diff
@@ -0,0 +1,34 @@
...
+    public String getId() {
+        return PROVIDER_ID;
+    }
+
+    @Override
+    public RealmsApi create(KeycloakSession session) {
+        return new DefaultRealmsApi(session);
+    }
+
+    @Override
```
stianst, line 20: As you'll probably never implement the init/postInit/close methods just add as defaults to the class it implements

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/ClientModelMapper.java
stianst, line 1: Is this mapper stuff going to be used anywhere else than in the Admin APIs? I wouldn't think so, so why not just move it to the admin-api module?

stianst (commented):
If we want to get this in early without wider review; and also in general; there shouldn't be all the changes to `server-spi-private`, `server-spi`, and `services`, but rather this should be moved to separate modules for the Admin API v2.

stianst:
Is there no authentication and authorization here? We really can't have a REST API without that, even if it is experimental.

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelpAll.approved.txt
stianst: This also means these won't be listed on https://www.keycloak.org/server/features; and not consistent with other features/capabilities. What about if we have different versions of this in the future, etc, etc. There are reasons why we have features and want to make this stuff consistent.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -0,0 +1,244 @@
...
+    @JsonProperty(defaultValue = OIDC)
+    @JsonPropertyDescription("The protocol used to communicate with the client")
+    private String protocol;
+
+    @JsonPropertyDescription("Whether this client is enabled")
+    private Boolean enabled;
+
+    @URL
+    @JsonPropertyDescription("URL to the application's homepage that is represented by this client")
+    private String appUrl;
```
stianst, line 38: That is not really the only thing `baseUrl` does. So this mapping between `appUrl` and `baseUrl` is not correct. I would just keep this as baseUrl as previous.

stianst (dismissed):
I'd like to see some changes before we can merge this:

* Would be good to modularise/isolate the code-base so there is a separate module for all interfaces, class, test, etc. (for example rest/api, rest/providers, rest/test). This will simplify a lot for others that want to take a look as all the code will be in one place, and not spread all over the place. We can also give the cloud native team ownership of this module, so you don't need external reviews for it. Obviously some changes will be needed elsewhere, but keep that at a minimum
* We need to have some authn/z here; one idea could be to just have a check at the root that there's an access token with the admin super user role
* OpenAPI should be wrapped in a feature

rest/admin-v2/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    <artifactId>keycloak-admin-v2-parent</artifactId>
+    <name>Keycloak Admin REST API v2 Parent</name>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>rest</module>
+        <module>api</module>
+        <module>providers</module>
+        <module>tests</module>
+    </modules>
```
mabartos, line 21: LGTM

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
```diff
@@ -692,7 +692,7 @@ public void copyProvider(TestProvider provider) {
         Path providerPackagePath = Paths.get(fileUri.getPath());
         JavaArchive providerJar = ShrinkWrap.create(JavaArchive.class, provider.getName() + ".jar")
                 .addClasses(provider.getClasses())
-                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
+                .addAsManifestResource(EmptyAsset.INSTANCE, "META-INF/beans.xml");
```
mabartos: Why is it necessary to make this change now?

mabartos (dismissed):
Latest changes look good to me, even with the submodules. 

We can do some more improvements in a follow-ups.

vmuzikar:
I did some of the requested changes in the last commits (still keeping those separate for clarity, will squash them on merge).
* I reorganized the changes so they are more focused in a single place. I stick to the "Admin v2" naming for now so the context is very clear at the first sight to everyone. We will of course rename later, if needed. There are currently the following modules:
   * `rest/admin-v2/api`: all service interfaces, SPIs, exceptions, etc. – except for SPIs, interfaces and providers related to endpoints
   * `rest/admin-v2/providers`: implementations of the services
   * `rest/admin-v2/rest`: all interfaces, providers, SPIs, ... for endpoints
   * `rest/admin-v2/tests`: integration tests based on the new testsuite
* Added basic auth/authz. It requires the admin role from master realm for now. There will be a follow-up for more granular authz.
* Added a Feature for OpenAPI.
* Reorganized the tests a bit.
* Removed the dependency on Expressly.

I will be creating some issues to capture the follow-ups.

This PR should be ready for another review round.

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
vmuzikar: That was not intentional, thanks.

pom.xml
vmuzikar: The dependency has been removed for now.

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testExportHelpAll.approved.txt
vmuzikar: Resolved, OpenAPI is a feature but I kept the options (which are not marked as experimental any more).

core/pom.xml
vmuzikar: `hibernate-validator` is now dependency only in the admin-v2 modules, services module and Quarkus module.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
vmuzikar: Done.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
vmuzikar: The reasoning is to support scenario where the server receives unknown fields in a request so it doesn't need to drop them but optionally keep them to be used by a custom extension. See https://github.com/keycloak/keycloak/discussions/40261

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
vmuzikar: Captured for a follow-up in https://github.com/keycloak/keycloak/issues/43290#issuecomment-3450448251.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
vmuzikar: Captured for a follow-up in https://github.com/keycloak/keycloak/issues/43290#issuecomment-3450448251.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
vmuzikar: Captured for a follow-up in https://github.com/keycloak/keycloak/issues/43290#issuecomment-3450448251.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
vmuzikar: Then auth should be disabled (as noted in the comment below). Captured for a follow-up in https://github.com/keycloak/keycloak/issues/43290#issuecomment-3450448251.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
vmuzikar: Captured for a follow-up in https://github.com/keycloak/keycloak/issues/43290#issuecomment-3450448251.

quarkus/config-api/src/main/java/org/keycloak/config/OptionCategory.java
vmuzikar: Wrapped in an experimental Feature, but removed the Experimental flag from the options.

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
vmuzikar: Removed.

quarkus/runtime/pom.xml
vmuzikar: That's what Fabric8 (part of Quarkus) uses. CC @shawkins

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AdminV2DisabledTest.java
vmuzikar: Moved to a new Admin v2 testsuite.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
vmuzikar: We'll make it a server config option. See https://github.com/keycloak/keycloak/issues/43728.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/ClientModelMapper.java
vmuzikar: Done.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApiSpi.java
vmuzikar: Captured as https://github.com/keycloak/keycloak/issues/43731

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminApiSpi.java
vmuzikar: Captured as https://github.com/keycloak/keycloak/issues/43731

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/realm/DefaultRealmsApiFactory.java
vmuzikar: Captured in https://github.com/keycloak/keycloak/issues/43731#issuecomment-3450623131

.github/workflows/ci.yml
```diff
@@ -900,6 +901,24 @@ jobs:
       - name: Run tests
         run: ./mvnw package -f tests/pom.xml
 
+  admin-v2-tests:
```
stianst, line 901: Need to also add to checks at the end, otherwise failures will still allow PRs to be merged

vmuzikar: Done. Was not sure if we want to be this strict but since it's run only conditionally, it's safe to do.

pom.xml
vmuzikar: Actually it was not removed as it's a transitive dependency of Hibernate Validator.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientsApi.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Tag(name = KeycloakOpenAPI.Admin.Tags.CLIENTS_V2)
+@Extension(name = KeycloakOpenAPI.Profiles.ADMIN, value = "")
+public interface ClientsApi extends Provider {
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Get all clients", description = "Returns a list of all clients in the realm")
+    Stream<ClientRepresentation> getClients();
```
sschu, line 31: I would just like to point out that this means that all metadata (e.g. pointers to previews/next page) needs to come in the headers. Things like showing if a client is editable for a user as used by the UI also would not work if we directly return representations. I am fine with this, just want to make sure everybody is on board.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/OpenApiPropertyMappers.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static org.keycloak.quarkus.runtime.configuration.Configuration.isTrue;
+import static org.keycloak.quarkus.runtime.configuration.mappers.PropertyMapper.fromOption;
+
+public final class OpenApiPropertyMappers implements PropertyMapperGrouping {
+
+  @Override
+  public List<? extends PropertyMapper<?>> getPropertyMappers() {
+    return List.of(
+        fromOption(OpenApiOptions.OPENAPI_ENABLED)
```
mabartos, line 16: Wouldn't be better to use `fromFeature`? As when the openapi feature is enabled, the whole openapi is enabled as well instead of enabling feature and this property.

WDYT?

Can be done in a follow-up.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.stream.Collectors;
+
+@OpenApiFilter(OpenApiFilter.RunStage.BUILD)
+public class OASModelFilter implements OASFilter {
+
+  @Override
+  public void filterOpenAPI(OpenAPI openAPI) {
+    // Filter Paths that have the '/admin/api/v2' prefix
+    Map<String, PathItem> newPaths = openAPI.getPaths().getPathItems().entrySet().stream()
+        .filter(entry -> entry.getKey().startsWith("/admin/api/v2"))
```
mabartos, line 24: Later, we'd need to be careful about this as we're gonna have versioned several admin APIs.

rest/admin-v2/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+
+    <artifactId>keycloak-admin-v2-parent</artifactId>
+    <name>Keycloak Admin REST API v2 Parent</name>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>rest</module>
+        <module>api</module>
+        <module>providers</module>
+        <module>tests</module>
```
mabartos, line 20: Some afterthought - Do we really need to separate the `rest`, and `providers`? It looks odd to me to have `rest/admin-v2/rest` modules. 

Can we merge the `rest` and `providers` into f.e. 'core'?

services/pom.xml
```diff
@@ -81,7 +81,11 @@
             <groupId>org.twitter4j</groupId>
             <artifactId>twitter4j-core</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
```
mabartos, line 86: We could probably emit this for `services` package as we have the admin v2 separated for now, and the validations are not used in v1, or services itself.

mabartos (dismissed):
LGTM as the base. I'm ok to do additional work in follow-ups.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/OpenApiPropertyMappers.java
vmuzikar: > Wouldn't be better to use `fromFeature`?

I don't think so. The feature will be eventually enabled by default as supported and we still want the spec not to be served by default, and be controlled by a dedicate option.

rest/admin-v2/pom.xml
vmuzikar: I'd like to keep the REST logic separate from the underlying APIs/providers. It's likely those will be moved to some other module later as Admin v2 is more settled.

services/pom.xml
vmuzikar: It's actually required [here](https://github.com/keycloak/keycloak/pull/43395/files#diff-75e0dc0fd3c9a052a6aaa216f1dc01719234cd6048b7095f8c67eda9fa8a473c).

mabartos: Ahhh, right.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/OpenApiPropertyMappers.java
mabartos: ok, works for me

rest/admin-v2/pom.xml
mabartos: Ok, works for me for now.

.github/scripts/find-modules-with-unit-tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 find . -path '**/src/test/java' -type d \
-  | grep -v -E '\./(docs|distribution|misc|operator|tests|testsuite|test-framework|quarkus)/' \
+  | grep -v -E '\./(docs|distribution|misc|operator|((.+/)?tests)|testsuite|test-framework|quarkus)/' \
```
vmuzikar, line 4: This is to exclude the Admin v2 IT from running as UT. Decided to do it in more generic way – I assume if a module is called `tests` it's somewhere executed as IT.

The output before:
```
adapters/saml/core,adapters/saml/wildfly/wildfly-subsystem,authz/client,common,core,crypto/default,crypto/elytron,crypto/fips1402,federation/kerberos,federation/ldap,integration/client-cli/admin-cli,model/infinispan,model/jpa,model/storage-private,saml-core,server-spi,server-spi-private,services%                             
```

After:
```
adapters/saml/core,adapters/saml/wildfly/wildfly-subsystem,authz/client,common,core,crypto/default,crypto/elytron,crypto/fips1402,federation/kerberos,federation/ldap,integration/client-cli/admin-cli,model/infinispan,model/jpa,model/storage-private,rest/admin-v2/tests,saml-core,server-spi,server-spi-private,services%         
```

mabartos (approved):

stianst (approved):

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
robson90: Relates to: https://github.com/keycloak/keycloak/issues/43571

PR #43406, opened 2025-10-13
Determine root distribution directory properly from ZIP file itself.

### LINKED ISSUE #43356: Keycloak tests framework - issue to identify distribution directory inside ZIP file when version string uses suffix
### Before reporting an issue

### Area

testsuite

### Describe the bug

Class RawKeycloakDistribution using version string in filename of the `zip` distribution to construct distribution home directory.
This is not correct when version with suffix is used. 
It prevents tests run at all.

### Version

26.2.9

### Regression

### Expected behavior

Test framework should determine the home directory automatically from the zip file and not based on zip's filename.

### Actual behavior

It fails to start Keycloak Server from described distribution zip file.

### How to Reproduce?

Set version of keycloak sources to `26.2.9-custom-2` and run build and tests.

### Anything else?

_No response_

pskopek:
Closes #43356

### DIFF
--- a/quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
+++ b/quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
@@ -41,6 +41,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
@@ -48,6 +49,9 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.LockSupport;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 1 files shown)


quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
```diff
@@ -494,22 +498,38 @@ private void reset() {
...
+            Optional<? extends ZipEntry> e = zipFile.stream().filter(ZipEntry::isDirectory).findFirst();
+            if (e.isPresent()) {
+                String dirName = e.get().getName();
+                if (dirName.contains("..")) {
+                    throw new RuntimeException("inside zip distribution directory cannot contain relative paths: " + dirName);
+                }
+                return Path.of(dirName);
+            }
+        };
+        throw noDirectories.get();
```
mabartos:
```suggestion
        throw new RuntimeException(String.format("ZIP file '%s' doesn't contain any directories", distPath));
```

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
```diff
@@ -494,22 +498,38 @@ private void reset() {
         keycloak = null;
     }
 
+    private Path inDistZipDirectory(File distFile) throws Exception{
+        Supplier<RuntimeException> noDirectories = () -> new RuntimeException(String.format("ZIP file %s doesn't contain any directories", distPath));
```
mabartos:
```suggestion
      
```

nitpick: I don't think we need the supplier here.

mabartos (commented):
@pskopek Thanks for the PR! I just added a small nitpick - no blocker.

mabartos (approved):

Issue #43394, opened 2025-10-13
SPIFFE client authentication does not work when JWT SVID includes `iss` claim


sabre1041:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Describe the bug

Attempting to obtain credentials using client assertion and SPIFFE Identity Provider and receiving an error obtaining credentials. See _Actual Behavior_ for full details

# Environment Background

* SPIFFE running on a Kubernetes cluster with federation endpoint exposed. Utilizing TLS certificates from public issuer to avoid TLS concerns
* Using inspiration from [this repository](https://github.com/stianst/playground/tree/main/demos/spiffe) for setting up and configuring the scenario

### Version

nightly, 26.4.0

### Regression

- [ ] The issue is a regression

### Expected behavior

Credentials can be obtain using a SPIFFE provided JWT

### Actual behavior

The follow error is reported on the client side

```
{"error":"unauthorized_client","error_description":"Unexpected error when authenticating client"}
```

The following error is returned on the server side

```
2025-10-13 09:48:58,311 ERROR [org.keycloak.services] (executor-thread-25) KC-SERVICES0015: Unexpected error when authenticating client: java.lang.NullPointerException: Cannot invoke "org.keycloak.authentication.FlowStatus.toString()" because "status" is null
	at org.keycloak.authentication.ClientAuthenticationFlow.processResult(ClientAuthenticationFlow.java:145)
	at org.keycloak.authentication.ClientAuthenticationFlow.processFlow(ClientAuthenticationFlow.java:90)
	at org.keycloak.authentication.AuthenticationProcessor.authenticateClient(AuthenticationProcessor.java:964)
	at org.keycloak.protocol.oidc.utils.AuthorizeClientUtil.authorizeClient(AuthorizeClientUtil.java:49)
	at org.keycloak.protocol.oidc.endpoints.TokenEndpoint.checkClient(TokenEndpoint.java:178)
	at org.keycloak.protocol.oidc.endpoints.TokenEndpoint.processGrantRequest(TokenEndpoint.java:134)
	at org.keycloak.protocol.oidc.endpoints.TokenEndpoint$quarkusrestinvoker$processGrantRequest_3903cccf0670c489ab77dc2ba1ba757573ec6d78.invoke(Unknown Source)
	at org.jboss.resteasy.reactive.server.handlers.InvocationHandler.handle(InvocationHandler.java:29)
[...]

PR #43407, opened 2025-10-13
Set auth flow status for FederatedJWTClientAuthenticator

mattbates:
Addresses https://github.com/keycloak/keycloak/issues/43394.

Uses `context.attempted()` to update the Authentication Flow Status for the new FederatedJWTClientAuthenticator.
I've also included debug logging which I found helpful to diagnose the underlying problem (in my case, the SPIFFE identity provider lookup).

(my first Java in years, and my contribution to Keycloak, so apologies in advance if it's not quite right..).


stianst (commented):
If I'm not mistaken context.attempted does nothing in this case. Will need to double check the code.

It's a bit hard to review your PR though due to the amount of irrelevant changes to formatting.

PR #43408, opened 2025-10-13
Separate trust domain and issuer fields in SPIFFE identity provider

mattbates:
Currently, the SPIFFE identity provider (in preview) assumes the issuer (`iss`) in a JWT-SVID token is the SPIFFE trust domain when looking up the identity provider. However, in some cases, it is not feasible for the issuer to be the same as the trust domain; in implementations like SPIRE, it may be necessary for the `iss` to be an HTTPS URL such that it can be used by federated systems for key verification.

This PR proposes separating out the trust domain and the issuer.

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/client/FederatedJWTClientAuthenticator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/client/FederatedJWTClientAuthenticator.java
@@ -1,5 +1,11 @@
 package org.keycloak.authentication.authenticators.client;
 
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import org.jboss.logging.Logger;
 import org.keycloak.Config;
 import org.keycloak.OAuth2Constants;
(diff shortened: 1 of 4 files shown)


stianst (commented):
It's not that it assumes the `iss` claim is the same as trust domain, rather it assumes there is no `iss` claim.

I think the correct approach here is to simply ignore the `iss` claim in the case of SPIFFE SVIDs. I'll get a PR together for that tomorrow probably, and probably close this one as I don't think this is the right approach.

mattbates:
Thanks for the feedback, @stianst; this is my first time looking at Keycloak code! 

>it assumes there is no iss claim

Agreed, and it's noted as being optional. 

I was referring to when the `iss` claim is set, it currently uses that to look up the provider (ie the `iss` claim being the trust domain). That was tripping me up with a POC implementation with SPIRE, causing an NPE (as per https://github.com/keycloak/keycloak/issues/43394).

https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/broker/spiffe/SpiffeIdentityProvider.java#L66
https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/broker/spiffe/SpiffeIdentityProviderConfig.java#L43-L45

>I think the correct approach here is to simply ignore the iss claim in the case of SPIFFE SVIDs

I think that makes sense.

stianst:
Closing as with https://github.com/keycloak/keycloak/pull/43428 we are now ignoring the `iss` claim completely.

PR #43413, opened 2025-10-13
Fix schedule-nightly workflow

stianst:

### DIFF
--- a/.github/workflows/schedule-nightly.yml
+++ b/.github/workflows/schedule-nightly.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - id: latest-release
         run: |
-          branch="$(gh api --paginate repos/keycloak/keycloak/branches -q '.[].name' | grep '^release/' | sort -r -V)"
+          branch="$(gh api --paginate repos/keycloak/keycloak/branches -q '.[].name' | grep '^release/' | sort -r -V | head -n 1)"
           echo "branch=$branch"
           echo "branch=$branch" >> "$GITHUB_OUTPUT"
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/schedule-nightly.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     steps:
       - id: latest-release
         run: |
-          branch="release/$(gh api repos/keycloak/keycloak/branches | jq -r '.[].name' | sort -r | awk -F'/' '/[0-9.]+$/ {print $NF; exit}')"
+          branch="$(gh api --paginate repos/keycloak/keycloak/branches -q '.[].name' | grep '^release/' | sort -r -V)"
```
ahus1: I ran this snippet: 

```
branch="$(gh api --paginate repos/keycloak/keycloak/branches -q '.[].name' | grep '^release/' | sort -r -V)"
```

and it resulted in 

```
$ echo $branch 
release/26.4 release/26.2 release/26.0
```

where it IMHO should return only a single item. 

Maybe it is missing a `head -1` to get just the first one.

ahus1 (changes requested):
I think it needs some more tuning; see below.

.github/workflows/schedule-nightly.yml
stianst:
```suggestion
          branch="$(gh api --paginate repos/keycloak/keycloak/branches -q '.[].name' | grep '^release/' | sort -r -V | head -n 1)"
```

ahus1 (approved):

PR #43412, opened 2025-10-13
Issue #43232: Apply `removeUnusedImports`

Pankraz76:
### Issue #43232: Apply `removeUnusedImports`

enabler for spotless migration:

- https://github.com/keycloak/keycloak/issues/43232
- https://github.com/Pankraz76/keycloak/pull/2

PR #43415, opened 2025-10-13
Issue #43232: Add `removeUnusedImports`

Pankraz76:
failing as expected until fixed:

- https://github.com/Pankraz76/keycloak/pull/2
- https://github.com/keycloak/keycloak/pull/43412

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1529,9 +1529,29 @@
                         </java>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>com.diffplug.spotless</groupId>
+                    <artifactId>spotless-maven-plugin</artifactId>
+                    <version>3.0.0</version>
+                    <configuration>
+                        <java>
+                            <removeUnusedImports />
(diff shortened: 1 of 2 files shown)


stianst:
Already added

Pankraz76:
thanks and sorry for outdate base.

pom.xml
```diff
@@ -1529,9 +1529,29 @@
...
+                        </java>
+                        <!--<pom>-->
+                        <!--    <sortPom/>-->
+                        <!--</pom>-->
+                    </configuration>
+                </plugin>
             </plugins>
         </pluginManagement>
-
+        <plugins>
```
Pankraz76, line 1549: Please consider this again for convenience:

Using the terminal is very costly. Power users prefer using shortcuts to execute the current runnable directly:

Image

Image

I was wondering how this works without being explicitly added. To me, it just looks managed — but it does work fine through the terminal, so it’s not broken.

PR #43416, opened 2025-10-13
Check Admin Roles also in composite roles and groups

### LINKED ISSUE #43405: Allow mapping Admin roles to server administrator only when using composite roles or groups
### Description

Continuation of https://github.com/keycloak/keycloak/issues/39956. We need to extend the changes done for that issue to include admin roles that are assigned using composite roles or groups that contains roles. This was commented in discussion https://github.com/keycloak/keycloak/discussions/42399.

rmartinc:
CLoses #43405

Adding the check for admin roles for composite roles and groups too.

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/AdminRoles.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/AdminRoles.java
@@ -69,5 +69,6 @@ public class AdminRoles {
         ALL_ROLES.add(ADMIN);
         ALL_ROLES.add(CREATE_REALM);
         ALL_ROLES.add(REALM_ADMIN);
+        ALL_ROLES.add(VIEW_SYSTEM);
     }
 }
--- a/services/src/main/java/org/keycloak/services/resources/admin/fgap/GroupPermissionsV2.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/fgap/GroupPermissionsV2.java
@@ -91,9 +91,19 @@ class GroupPermissionsV2 extends GroupPermissions {
         return eval.hasPermission(new GroupModelRecord(group), null, AdminPermissionsSchema.MANAGE_MEMBERS);
     }
 
+    private boolean canMapRolesFromGroup(GroupModel group) {
+        while (group != null) {
+            if (!group.getRoleMappingsStream().allMatch(s -> root.roles().canMapRole(s))) {
+                return false;
+            }
+            group = group.getParent();
+        }
+        return true;
+    }
+
     @Override
     public boolean canManageMembership(GroupModel group) {
-        if (root.hasOneAdminRole(AdminRoles.MANAGE_USERS)) {
+        if (root.hasOneAdminRole(AdminRoles.MANAGE_USERS) && canMapRolesFromGroup(group)) {
             return true;
         }
 
--- a/services/src/main/java/org/keycloak/services/resources/admin/fgap/GroupPermissions.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/fgap/GroupPermissions.java
@@ -354,9 +354,21 @@ class GroupPermissions implements GroupPermissionEvaluator, GroupPermissionManag
         return hasPermission(group, MANAGE_MEMBERS_SCOPE);
     }
 
+    private boolean canMapRolesFromGroup(GroupModel group) {
+        while (group != null) {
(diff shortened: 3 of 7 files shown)


server-spi-private/src/main/java/org/keycloak/models/AdminRoles.java
```diff
@@ -69,5 +69,6 @@ public class AdminRoles {
         ALL_ROLES.add(ADMIN);
         ALL_ROLES.add(CREATE_REALM);
         ALL_ROLES.add(REALM_ADMIN);
+        ALL_ROLES.add(VIEW_SYSTEM);
```
pedroigor, line 72: Is it related? This one is only used to enforce access to the server info endpoint and I guess it is working today without it here?

My concern is that this role is marked for removal in 26.4?

rmartinc: This is needed, although deprecated that role should also be not be assigned because it's an administration role. When removed, we can also remove it from here.

services/src/main/java/org/keycloak/services/resources/admin/fgap/GroupPermissionsV2.java
```diff
@@ -91,9 +91,19 @@ public boolean canManageMembers(GroupModel group) {
...
+            }
+            group = group.getParent();
+        }
+        return true;
+    }
+
     @Override
     public boolean canManageMembership(GroupModel group) {
-        if (root.hasOneAdminRole(AdminRoles.MANAGE_USERS)) {
+        if (root.hasOneAdminRole(AdminRoles.MANAGE_USERS) && canMapRolesFromGroup(group)) {
```
vramik, line 106: @rmartinc Thanks a lot for preparing the PR.

The code itself looks good, but I have one question to fully understand the logic. Can you please elaborate a bit on this change? I'm not seeing the relationship between being able to `manage-membership` of a group and being able to `map-roles` in the group hierarchy.

pedroigor: @rmartinc can confirm that, but it is about checking that the user trying to manage membership has access to all roles assigned to the group  being managed, so that they cannot add members with roles like admin roles as they are enforced by the changes in the role permission types.

rmartinc: @vramik The idea is a group also asigns the roles defined in the group (and parents). So, you only can assign (canManageMembership) the group if you can also assign all the roles associated to that group.

pedroigor: In terms of authz, I prefer to go through the more restricted/constrained path. So yeah, it does make a lot of sense.

I'm wondering, though, a situation where a user is a manager of a group but does not have all the roles assigned to the group hierarchy. Is this a valid scenario?

rmartinc:
@pedroigor @vramik Finally we decided to not implement this one, right? If you agree then I will close the PR and the related issue.

ianwallen:
Any updates on this? I'm not sure why this issue would not want to be resolved.  If this PR is not going to be fixing the issue then is there another PR that will? I have tested with the latest keycloak release and the issue still remains.

Our system currently has 2 groups. realm_admin (which can manage the realm) and realm_user_admin (which can manager users).

All the users who were granted access to realm_user_admin group are able to grant themselves the realm_admin group.  This seems like a pretty big security flaw?

The only workaround that we have at the moment is to ensure that no realm management roles are granted to groups. We need to grant all the roles directly to the users which is not intuitive.  And it is difficult to ensure that realm administrator are not creating groups that have realm management roles assigned to them.

Looking forward to seeing this issue addressed.

Thank you

pedroigor:
@ianwallen It is not a security flaw but a limitation on how you are able to delegate access to realm administrators. I understand your point on being forced to grant admin roles directly to users instead of just having them members of administrative groups and inheriting the admin roles from the groups.

The key point here is that we need to make sure not to potentially impact other valid use cases, where allowing granting admin roles for users in a group should allow them to map roles.

One potentially solution I see for this is to make sure only server admins (granted with `admin` role) and realm admins (granted with `realm-admin`) roles are allowed to grant realm management roles. 

We also need to consider adding more fine-grained control for roles by supporting a role resource type in FGAP

rmartinc:
Then I'm going to close this PR for the moment.

PR #43420, opened 2025-10-13
fix: refining https-protocols documentation

### LINKED ISSUE #43164: Keycloak docs state that only TLSv1.3 is used
### Before reporting an issue

### Area

docs

### Describe the bug

https://www.keycloak.org/server/enabletls#_configuring_tls_protocols states that only TLS 1.3 is used. 

> By default, Keycloak does not enable deprecated TLS protocols. If your client supports only deprecated protocols, consider upgrading the client. However, as a temporary work-around, you can enable deprecated protocols by running the following command:
...
To also allow TLSv1.2, use a command such as the following: kc.sh start --https-protocols=TLSv1.3,TLSv1.2.

But actually since #21912 both TLS 1.3+1.2 are enabled by default. Due to a bug in Quarkus even before both were enabled. 

### Version

main

### Regression

### Expected behavior

The docs should be consistent with the actual defaults.

### Actual behavior

Outdated docs.

### How to Reproduce?

Look at the docs.

### Anything else?

I'll create a PR

shawkins:
closes: #43164

Changed the example to mirror the current quarkus docs, rather than implying that TLSv1.2 is not enabled by default. Also made the option documentation clearer wrt expected values - we may want to log an issue against quarkus to more strongly validate, or we can do this as a breaking change with #43332

### DIFF
--- a/docs/guides/server/enabletls.adoc
+++ b/docs/guides/server/enabletls.adoc
@@ -52,7 +52,7 @@ However, as a temporary work-around, you can enable deprecated protocols by runn
 
 <@kc.start parameters="--https-protocols=<protocol>[,<protocol>]"/>
 
-To also allow TLSv1.2, use a command such as the following: `kc.sh start --https-protocols=TLSv1.3,TLSv1.2`.
+For example to only enable TLSv1.3, use a command such as the following: `kc.sh start --https-protocols=TLSv1.3`.
 
 == Switching the HTTPS port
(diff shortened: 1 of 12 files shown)


shawkins:
@vmuzikar @ahus1 @statist32 the failure makes it look like spotless is running against the entire code base, not just the files touched in the pr. Is this expected?

mabartos (approved):
LGTM

ahus1 (approved):
Thank you for this change. Not merging yet as there are still other reviewers. Ping me if you want this to be merged.

PR #43422, opened 2025-10-13
fix: refines how defaults are shown

shawkins:
closes: #43421

This makes the formatting much closer to what quarkus shows.

### DIFF
--- a/docs/guides/templates/options.adoc
+++ b/docs/guides/templates/options.adoc
@@ -17,9 +17,9 @@
 </#macro>
 
 <#macro list options buildIcon=true anchor=true categoryName="">
-[cols="12a,4",role="options"]
+[cols="12a,2,2",role="options"]
 |===
-| |Value
+| |_Type_ or Values |Default
 
 <#list options as option>
 |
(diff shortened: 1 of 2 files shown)


ahus1 (changes requested):
I'm ok with the change. One thing to fix is that the default value is shown in a smaller font than the type or values. 

Image

shawkins:
> I'm ok with the change. One thing to fix is that the default value is shown in a smaller font than the type or values.

This is coming from the options-default css class. It does render the defaults at a smaller font size even in the current web doc, but it's only noticible when the default is by itself. I'll just remove the usage of that class.

mabartos:
@shawkins Nice improvement. LGTM, thanks!

image

mabartos (dismissed):

shawkins:
Looking at the quarkus docs, using 12,4,4 probably takes away too much width from the description, so I've changed it to 12,2,2 to keep the original percentage for the description.

However there are two things that could still be improved:
- using the convention "value (deprecated)" almost always results in the (deprecated) being wrapped to the next line. I would rather use ~value~ and possibly add a hover explaining it is deprecated.
- We are not using word wrapping on the values / default columns, on smaller screens this will distort the size of the values and especially the default column. An improvement here I think requires a stylesheet change in keycloak-web

~~WDYT of those changes, and should they be done in the context of this issue?~~

Opened as #43458

ahus1 (approved):
Thank you, looks goog to me. 

Image

PR #43431, opened 2025-10-14
Add FIPS suite to the new tests

### LINKED ISSUE #35389: Add fips testsuite to the new tests

lhanusov:
Closes #35389


tests/base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java
```diff
@@ -0,0 +1,56 @@
...
+                .configFile("/org/keycloak/tests/suites/fips/kc.keystore-create.java.security")
+                .configFile("/org/keycloak/tests/suites/fips/keycloak-fips.keystore.pkcs12")
+                .configFile("/org/keycloak/tests/suites/fips/keycloak-fips.truststore.pkcs12")
+                .dependency("org.bouncycastle", "bc-fips")
+                .dependency("org.bouncycastle", "bctls-fips")
+                .dependency("org.bouncycastle", "bcpkix-fips")
+                .dependency("org.bouncycastle", "bcutil-fips")
+                .clientDependency("org.bouncycastle", "bc-fips")
+                .clientDependency("org.bouncycastle", "bctls-fips")
+                .clientDependency("org.bouncycastle", "bcutil-fips");
```
stianst: What's this? According to the docs the deps should only be installed into `providers`

.github/scripts/run-fips-it.sh
```diff
@@ -37,3 +37,7 @@ fi
 
 # Profile app-server-wildfly needs to be explicitly set for FIPS tests
 ./mvnw test -Dsurefire.rerunFailingTestsCount=$SUREFIRE_RERUN_FAILING_COUNT -nsu -B -Pauth-server-quarkus,auth-server-fips140-2,app-server-wildfly -Dcom.redhat.fips=false $STRICT_OPTIONS -Dtest=$TESTS -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
+
+# New Base Tests
+CLASS_NAME=Fips$(if [[ $1 =~ "non" ]] ; then echo $1 | sed 's/\([[:alpha:]]\)/\U\1\4/' ; else echo $1 | sed 's/\([[:alpha:]]\)/\U\1/' | sed 's/-//' ; fi)TestSuite
```
stianst: There's only two right? and if/else would probably read better, I wouldn't know where to start understanding what this expression does ;)

tests/base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java
lhanusov: this is for kcadm.sh and kcreg.sh: https://www.keycloak.org/server/fips#_run_the_cli_on_the_fips_host,
but I will remove it for now

tests/base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java
```diff
@@ -0,0 +1,56 @@
...
+        SuiteSupport.stopSuite();
+    }
+
+    public static class FipsNonStrictServerConfig implements KeycloakServerConfig {
+
+        @Override
+        public KeycloakServerConfigBuilder configure(KeycloakServerConfigBuilder config) {
+            return config.features(Profile.Feature.FIPS)
+                .option("-Dserver.fips.mode", "non-strict")
+                .option("-Dserver.supported.keystore.types", "PKCS12")
```
rmartinc: Non-strict is `PKCS12,BCFKS` for this property. Both are supported here.

tests/base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java
rmartinc: The FIPS TS runs for kcadm and kcreg test classes, don't really know if this is needed or not but take that into account.

tests/base/src/test/java/org/keycloak/tests/suites/FipsStrictTestSuite.java
```diff
@@ -0,0 +1,57 @@
...
+                .configFile("/org/keycloak/tests/suites/fips/kc.keystore-create.java.security")
+                .configFile("/org/keycloak/tests/suites/fips/keycloak-fips.keystore.bcfks")
+                .configFile("/org/keycloak/tests/suites/fips/keycloak-fips.truststore.bcfks")
+                .dependency("org.bouncycastle", "bc-fips")
+                .dependency("org.bouncycastle", "bctls-fips")
+                .dependency("org.bouncycastle", "bcpkix-fips")
+                .dependency("org.bouncycastle", "bcutil-fips")
+                .clientDependency("org.bouncycastle", "bc-fips")
+                .clientDependency("org.bouncycastle", "bctls-fips")
+                .clientDependency("org.bouncycastle", "bcutil-fips");
```
rmartinc: Here I miss the options that are added in AbstractQuarkusDeployableContainer. For example `--spi-password-hashing-pbkdf2-max-padding-length=14`. Take a look if you see problems in the CI when the tests are added.

rmartinc (changes requested):
Thanks @lhanusov! I have added some comments. But you have to check in your tests if the log `KC(BCFIPS version 2.0.102...` is displayed. If you are testing strict the approved mode should be seen. I don't see that in the current output but probably because it is trimmed.

.github/scripts/run-fips-it.sh
```diff
@@ -37,3 +39,6 @@ fi
 
 # Profile app-server-wildfly needs to be explicitly set for FIPS tests
 ./mvnw test -Dsurefire.rerunFailingTestsCount=$SUREFIRE_RERUN_FAILING_COUNT -nsu -B -Pauth-server-quarkus,auth-server-fips140-2,app-server-wildfly -Dcom.redhat.fips=false $STRICT_OPTIONS -Dtest=$TESTS -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
+
+# New Base Tests
+./mvnw package -nsu -B -Dcom.redhat.fips=false $STRICT_OPTIONS -Dtest=$TESTSUITE_NAME -pl tests/base
```
stianst:
```suggestion
./mvnw test -nsu -B -Dcom.redhat.fips=false $STRICT_OPTIONS -Dtest=$TESTSUITE_NAME -pl tests/base
```

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactorySupplier.java
```diff
@@ -15,12 +15,12 @@ public class AdminClientFactorySupplier implements Supplier<AdminClientFactory,
     public AdminClientFactory getValue(InstanceContext<AdminClientFactory, InjectAdminClientFactory> instanceContext) {
         KeycloakServer server = instanceContext.getDependency(KeycloakServer.class);
 
-        if (!server.isTlsEnabled()) {
-            return new AdminClientFactory(server.getBaseUrl());
-        } else {
+        if (server.isTlsEnabled() || server.isFipsEnabled() ) {
```
stianst: I don't like this change - simplest would just be to make the FIPS Suite enable TLS as well (or you can always make the config set TLS when FIPS is enabled).

tests/base/src/test/java/org/keycloak/tests/infinispan/InfinispanXMLBackwardCompatibilityTest.java
```diff
@@ -12,7 +12,7 @@
 @KeycloakIntegrationTest(config = InfinispanXMLBackwardCompatibilityTest.ServerConfigWithCustomInfinispanXML.class)
 public class InfinispanXMLBackwardCompatibilityTest {
 
-    private static final String CONFIG_FILE = "/embedded-infinispan-config/infinispan-xml-kc26.xml";
+    private static final String CONFIG_FILE = "/org/keycloak/tests/infinispan/infinispan-xml-kc26.xml";
```
stianst: Irrelevant change, please remove

tests/base/src/test/resources/org/keycloak/tests/infinispan/infinispan-xml-kc26.xml
stianst: Irrelevant change, please remove

test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java
```diff
@@ -49,10 +49,16 @@ public KeycloakServer getValue(InstanceContext<KeycloakServer, KeycloakIntegrati
         ServerConfigInterceptorHelper interceptor = new ServerConfigInterceptorHelper(instanceContext.getRegistry());
         command = interceptor.intercept(command, instanceContext);
 
-        if (command.tlsEnabled()) {
+        if (command.tlsEnabled() || command.fipsEnabled()) {
```
stianst: Same here; enable TLS in suite, or when fips is enabled in the config.

stianst (changes requested):
Remove KeycloakServer.isFipsEnabled and same for the config; and all the updates to the code that is checking both to see if TLS is enabled or not.

In Fips suites just use tlsEnabled instead of fipsEnabled. It's basically a null-op.

tests/base/src/test/resources/org/keycloak/tests/infinispan/infinispan-xml-kc26.xml
stianst, line 1: Irrelevant change, please remove

test-framework/clustering/src/main/java/org/keycloak/testframework/server/ClusteredKeycloakServer.java
stianst, line 1: What are the changes here? Seems they are to support copying providers and not related to FIPS? If so could you do this as a separate PR instead please?

test-framework/clustering/src/main/java/org/keycloak/testframework/server/ClusteredKeycloakServer.java
```diff
@@ -159,6 +162,11 @@ public boolean isTlsEnabled() {
         return false;
     }
 
+    @Override
+    public boolean isFipsEnabled() {
```
stianst: Remove

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactorySupplier.java
stianst, line 1: Remove changes to this

test-framework/core/src/main/java/org/keycloak/testframework/http/HttpClientSupplier.java
stianst, line 1: Remove changes to this

test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java
stianst: Remove this change

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
stianst, line 1: Remove changes to this

tests/base/src/test/resources/org/keycloak/tests/suites/fips/kc.java.security
stianst, line 1: This seems to do nothing really as there's nothing that updates this (i.e. add `fips.provider.7=XMLDSig`). 

I suggest removing for now, including the ability to set system properties. In the future if we need this we should have something explicit for it (something like `KeycloakServerConfigBuilder#customJavaSecurityFile(Path... p)` or `KeycloakServerConfigBuilder#customJavaSecurityEntries(String... entries)`).

tests/FIPS_104-2.md
```diff
@@ -0,0 +1,27 @@
+## FIPS 140-2 testing
+
+### Integration tests
+
+On the FIPS enabled platform with FIPS enabled OpenJDK 21, you can run this to test against a Keycloak server on Quarkus
```
stianst:
```suggestion
On the FIPS enabled platform with FIPS enabled OpenJDK 21, you can run this to test against a Keycloak server
```

tests/FIPS_104-2.md
```diff
@@ -0,0 +1,27 @@
...
+```
+./mvnw clean package -nsu -B -f tests/base/pom.xml \
+  -Dcom.redhat.fips=false \
+  -Dtest=FipsStrictTestSuite,FipsNonStrictTestSuite
+```
+FIPS test suite configuration (strict and non-strict):
+[FipsStrictTestSuite](./base/src/test/java/org/keycloak/tests/suites/FipsStrictTestSuite.java),
+[FipsNonStrictTestSuite](./base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java)
+
+NOTE 1: The property `com.redhat.fips` is needed so that testsuite itself is executed in the JVM with FIPS disabled. However
```
stianst: Is `com.redhat.fips=false` needed? In embedded mode this system property will be passed on to Keycloak as well (and obviously already set in the JVM running KC).

tests/base/pom.xml
```diff
@@ -130,6 +130,51 @@
...
+                <dependency>
+                    <groupId>org.bouncycastle</groupId>
+                    <artifactId>bcprov-jdk18on</artifactId>
+                    <scope>test</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.bouncycastle</groupId>
+                    <artifactId>bcpkix-jdk18on</artifactId>
+                    <scope>test</scope>
+                </dependency>
```
stianst: It's very strange that these needs to be set here; as they should already be added:
```
$ mvn dependency:tree | grep bouncycastle
[INFO] |     |  |  +- org.bouncycastle:bcprov-jdk18on:jar:1.81:test
[INFO] |     |  |  \- org.bouncycastle:bcpkix-jdk18on:jar:1.81:test
[INFO] |     |  |     \- org.bouncycastle:bcutil-jdk18on:jar:1.81:test
```

I don't think there should be a problem having the BouncyCastle FIPS dependencies added without wrapping in a profile either.

test-framework/core/src/main/java/org/keycloak/testframework/server/EmbeddedKeycloakServer.java
stianst, line 1: Remove changes to this

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServer.java
stianst, line 1: Remove changes to this

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
stianst, line 1: None of the changes to this is needed as far as I can tell. That is after you remove the fipsEnabled stuff

test-framework/core/src/main/java/org/keycloak/testframework/server/RemoteKeycloakServer.java
stianst, line 1: Remove changes

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
```diff
@@ -83,18 +93,15 @@ public void testServerInfo() {
...
         assertNotNull(info.getSystemInfo().getUptime());
 
         Map<String, ProviderRepresentation> jpaProviders = info.getProviders().get("connectionsJpa").getProviders();
         ProviderRepresentation jpaProvider = jpaProviders.values().iterator().next();
-        log.infof("JPA Connections provider info: %s", jpaProvider.getOperationalInfo());
-    }
-
-    @Override
-    public void addTestRealms(List<RealmRepresentation> testRealms) {
+        LOGGER.infof("JPA Connections provider info: %s", jpaProvider.getOperationalInfo());
```
stianst: Not quite relevant to migrating this test, but I'd remove this logging stuff; or change the log statement to an assert.

tests/base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java
```diff
@@ -0,0 +1,60 @@
...
+        SuiteSupport.stopSuite();
+    }
+
+    public static class FipsNonStrictServerConfig implements KeycloakServerConfig {
+
+        @Override
+        public KeycloakServerConfigBuilder configure(KeycloakServerConfigBuilder config) {
+            return config.features(Profile.Feature.FIPS).fipsEnabled(true)
+                .option("fips-mode", "non-strict")
+                .javaOpt("java.security.properties", getFilePathAsString("fips/kc.java.security"))
```
stianst: Remove

tests/base/src/test/java/org/keycloak/tests/suites/FipsNonStrictTestSuite.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static class FipsNonStrictCertificatesConfig implements CertificatesConfig {
+
+        @Override
+        public CertificatesConfigBuilder configure(CertificatesConfigBuilder config) {
+            return config.cryptoProviderOption(CryptoProviderOption.FIPS_NON_STRICT)
+                    .keystoreFormat(KeystoreUtil.KeystoreFormat.PKCS12)
+                    .keystorePassword("passwordpassword".toCharArray())
+                    .truststorePassword("passwordpassword".toCharArray());
```
stianst: Remove; the certificate provider should just handle the passwords; there's no need to make this configurable

tests/base/src/test/java/org/keycloak/tests/suites/FipsStrictTestSuite.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static class FipsStrictCertificatesConfig implements CertificatesConfig {
+
+        @Override
+        public CertificatesConfigBuilder configure(CertificatesConfigBuilder config) {
+            return config.cryptoProviderOption(CryptoProviderOption.FIPS_STRICT)
+                    .keystoreFormat(KeystoreUtil.KeystoreFormat.BCFKS)
+                    .keystorePassword("passwordpassword".toCharArray())
+                    .truststorePassword("passwordpassword".toCharArray());
```
stianst: Remove; the certificate provider should just handle the passwords; there's no need to make this configurable

tests/base/src/test/java/org/keycloak/tests/suites/FipsStrictTestSuite.java
```diff
@@ -0,0 +1,73 @@
...
+    public static class FipsStrictServerConfig implements KeycloakServerConfig {
+
+        @Override
+        public KeycloakServerConfigBuilder configure(KeycloakServerConfigBuilder config) {
+            return config.features(Profile.Feature.FIPS).fipsEnabled(true)
+                .option("fips-mode", "strict")
+                .option("spi-password-hashing-pbkdf2-max-padding-length", "14")
+                .option("spi-password-hashing-pbkdf2-sha256-max-padding-length", "14")
+                .option("spi-password-hashing-pbkdf2-sha512-max-padding-length", "14")
+                .javaOpt("java.security.properties", getFilePathAsString("fips/kc.java.security"))
```
stianst: Remove

test-framework/core/src/main/java/org/keycloak/testframework/https/CertificatesConfigBuilder.java
```diff
@@ -0,0 +1,60 @@
+package org.keycloak.testframework.https;
+
+import org.keycloak.common.util.KeystoreUtil;
+
+public class CertificatesConfigBuilder {
+
+    private CryptoProviderOption cryptoProviderOption = CryptoProviderOption.DEFAULT;
+    private KeystoreUtil.KeystoreFormat keystoreFormat = KeystoreUtil.KeystoreFormat.JKS;
+    private String fileFormatSuffix = ".jks";
```
stianst: I wouldn't deal with the file suffix here (it's not something we need to make configurable); just handle that in the supplier.

test-framework/core/src/main/java/org/keycloak/testframework/https/CertificatesConfigBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+    private char[] keystorePassword = "password".toCharArray();
+    private char[] truststorePassword = "password".toCharArray();
+
+    public CertificatesConfigBuilder () {
+    }
+
+    public CertificatesConfigBuilder cryptoProviderOption(CryptoProviderOption option) {
+        this.cryptoProviderOption = option;
+        return this;
+    }
```
stianst: Setting the crypto provider here is weird; maybe should consider doing a CryptoProviderSupplier?

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
```diff
@@ -56,18 +66,18 @@ public void testServerInfo() {
...
+        Assert.assertNames(info.getThemes().get("login"), "base", "keycloak", "keycloak.v2");
         Assert.assertNames(info.getThemes().get("welcome"), "keycloak");
 
         assertNotNull(info.getEnums());
 
         assertNotNull(info.getMemoryInfo());
         assertNotNull(info.getSystemInfo());
         assertNotNull(info.getCryptoInfo());
-        Assert.assertNames(info.getCryptoInfo().getSupportedKeystoreTypes(), KeystoreUtils.getSupportedKeystoreTypes());
+        Assert.assertNames(info.getCryptoInfo().getSupportedKeystoreTypes(), certificates.getSupportedKeystoreTypes());
```
stianst: I don't like this as it doesn't really test anything; should really be a list specified here that differs if FIPS is set or not. If we did a CryptoProviderSupplier then you could inject that instead of ManagedCertificates to do a different list if FIPS is enabled or not.

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
```diff
@@ -83,18 +93,15 @@ public void testServerInfo() {
                 .stream()
                 .filter(configProp -> Attributes.KEY_SIZE_KEY.equals(configProp.getName()))
                 .findFirst().orElseThrow(() -> new RuntimeException("Not found provider with ID 'rsa-generated'"));
-        Assert.assertNames(keySizeRep.getOptions(), KeyUtils.getExpectedSupportedRsaKeySizes());
+
+        Assert.assertNames(keySizeRep.getOptions(), certificates.getSupportedKeySizes());
```
stianst: Same here really; doesn't assert things properly as it just uses the class used on the server side directly.

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
     public final static String CERT_ENTRY = "cert";
 
 
-    public ManagedCertificates() throws ManagedCertificatesException {
+    public ManagedCertificates(CertificatesConfigBuilder config) throws ManagedCertificatesException {
+        this.config = config;
+
         if (!CryptoIntegration.isInitialised()) {
-            CryptoIntegration.setProvider(new DefaultCryptoProvider());
+            switch (config.getCryptoProviderOption()) {
```
stianst: Shouldn't be handled by ManagedCertificates

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
     private KeyStore clientsTrustStore;
 
     private final static Path KEYSTORES_DIR = Path.of(System.getProperty("java.io.tmpdir"));
-    private final static Path SERVER_KEYSTORE_FILE_PATH = KEYSTORES_DIR.resolve("kc-testing-server-keystore.jks");
-    private final static Path CLIENTS_TRUSTSTORE_FILE_PATH = KEYSTORES_DIR.resolve("kc-testing-clients-truststore.jks");
-
-    private final static char[] PASSWORD = "password".toCharArray();
-
+    private final Path SERVER_KEYSTORE_FILE_PATH;
+    private final Path CLIENTS_TRUSTSTORE_FILE_PATH;
```
stianst: Lower case field names as it's no longer static

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
+
+    public String getClientsTrustStorePassword() {
+        return String.valueOf(config.getTruststorePassword());
+    }
+
+    public String getKeystoreType() {
+        return config.getKeystoreFormat().toString().toLowerCase();
+    }
+
+    public String[] getSupportedKeystoreTypes() {
```
stianst: Doesn't belong here

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
+        return CryptoIntegration.getProvider().getSupportedKeyStoreTypes()
+                .map(obj -> Objects.toString(obj, null)).collect(Collectors.joining(",")).split(",");
+    }
+
+    public String[] getSupportedKeySizes() {
```
stianst: Doesn't belong here

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
+        CLIENTS_TRUSTSTORE_FILE_PATH = KEYSTORES_DIR.resolve("kc-testing-clients-truststore" + config.getFileFormatSuffix());
+
         initServerCerts();
     }
 
     public String getKeycloakServerKeyStorePath() {
         return SERVER_KEYSTORE_FILE_PATH.toString();
     }
 
+    public String getClientsTrustStorePath() {
```
stianst: Is this used anywhere?

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
+        return CLIENTS_TRUSTSTORE_FILE_PATH.toString();
+    }
+
     public String getKeycloakServerKeyStorePassword() {
-        return String.valueOf(PASSWORD);
+        return String.valueOf(config.getKeystorePassword());
+    }
+
+    public String getClientsTrustStorePassword() {
```
stianst: Is this used anywhere?

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -21,40 +22,71 @@
...
+        return String.valueOf(config.getTruststorePassword());
+    }
+
+    public String getKeystoreType() {
```
stianst: Is this used anywhere?

.github/scripts/run-fips-it.sh
lhanusov: package is necessary for correct test execution when potentially using custom provider

test-framework/clustering/src/main/java/org/keycloak/testframework/server/ClusteredKeycloakServer.java
lhanusov: it was due to the javaOpts => removing

tests/base/pom.xml
lhanusov: I've found the root cause it's implemented within 2 libs:
- Bouncycastle library: org.bouncycastle.crypto.CryptoServiceProperties
- Testcontainers library: org.testcontainers.shaded.org.bouncycastle.crypto.CryptoServiceProperties

The conflict in Java is it cannot decide which CryptoServiceProperties interface should use.
https://github.com/testcontainers/testcontainers-java/issues/338
https://github.com/testcontainers/testcontainers-java/issues/839

`[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.240 s <<< FAILURE! -- in org.keycloak.tests.admin.ServerInfoTest
[ERROR] org.keycloak.tests.admin.ServerInfoTest.testServerInfo -- Time elapsed: 0.221 s <<< ERROR!
java.lang.NoSuchMethodError: 'void org.bouncycastle.crypto.CryptoServicesRegistrar.checkConstraints(org.bouncycastle.crypto.CryptoServiceProperties)'
	at org.bouncycastle.jce.provider.BouncyCastleProvider.loadAlgorithms(Unknown Source)
	at org.bouncycastle.jce.provider.BouncyCastleProvider.setup(Unknown Source)
	at org.bouncycastle.jce.provider.BouncyCastleProvider.access$000(Unknown Source)
	at org.bouncycastle.jce.provider.BouncyCastleProvider$1.run(Unknown Source)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:319)
	at org.bouncycastle.jce.provider.BouncyCastleProvider.<init>(Unknown Source)
	at org.keycloak.crypto.def.DefaultCryptoProvider.<init>(DefaultCryptoProvider.java:56)
	at org.keycloak.testframework.https.ManagedCertificates.<init>(ManagedCertificates.java:50)`

test-framework/clustering/src/main/java/org/keycloak/testframework/server/ClusteredKeycloakServer.java
```diff
@@ -117,6 +117,7 @@ private void startContainersWithSameImage(KeycloakServerConfigBuilder configBuil
             containers[i] = container;
 
             copyProvidersAndConfigs(container, configBuilder);
+
```
stianst: Remove - no need for this change; and results in extra files being changed in the PR

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
lhanusov: yup, in the AbstractKeycloakServerSupplier

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
lhanusov: removed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
lhanusov: removed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
stianst: You're mixing two things then; `clientsTrustStore` contains the Keycloak public cert so clients (admin, etc.) can connect to Keycloak.

This should not be configured in Keycloak as its truststore.

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -30,40 +28,45 @@ public class ManagedCertificates {
...
 
-    public String getKeycloakServerKeyStorePassword() {
-        return String.valueOf(PASSWORD);
+    public String getPassword() {
+        return String.valueOf(this.password);
     }
 
     public KeyStore getClientTrustStore() {
-        return clientsTrustStore;
+        return this.clientsTrustStore;
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -30,40 +28,45 @@ public class ManagedCertificates {
...
     }
 
     public X509Certificate getKeycloakServerCertificate() {
         try {
-            return (X509Certificate) serverKeyStore.getCertificate(CERT_ENTRY);
+            return (X509Certificate) this.serverKeyStore.getCertificate(CERT_ENTRY);
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -72,33 +75,33 @@ public X509Certificate getKeycloakServerCertificate() {
     public SSLContext getClientSSLContext() {
         try {
             return SSLContextBuilder.create()
-                    .loadTrustMaterial(clientsTrustStore, null)
+                    .loadTrustMaterial(this.clientsTrustStore, null)
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -121,18 +124,18 @@ private void generateKeystore(KeyStore keyStore, KeyStore trustStore, String sub
 
         keyStore.setCertificateEntry(CERT_ENTRY, cert);
         trustStore.setCertificateEntry(CERT_ENTRY, cert);
-        keyStore.setKeyEntry(PRV_KEY_ENTRY, keyPair.getPrivate(), PASSWORD, new X509Certificate[]{cert});
+        keyStore.setKeyEntry(PRV_KEY_ENTRY, keyPair.getPrivate(), this.password, new X509Certificate[]{cert});
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -121,18 +124,18 @@ private void generateKeystore(KeyStore keyStore, KeyStore trustStore, String sub
...
     }
 
     private KeyPair generateKeyPair() throws NoSuchAlgorithmException, NoSuchProviderException {
-        return cryptoProvider.getKeyPairGen("RSA").generateKeyPair();
+        return this.cryptoProvider.getKeyPairGen("RSA").generateKeyPair();
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -121,18 +124,18 @@ private void generateKeystore(KeyStore keyStore, KeyStore trustStore, String sub
...
     }
 
     private X509Certificate generateX509CertificateCertificate(KeyPair keyPair, String subject) throws Exception {
-        // generate a v1 certificate
-        X509Certificate caCert = cryptoProvider.getCertificateUtils().generateV1SelfSignedCertificate(keyPair, subject);
+        // generate a v1 root certificate authority certificate
+        X509Certificate caCert = this.cryptoProvider.getCertificateUtils().generateV1SelfSignedCertificate(keyPair, subject);
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -121,18 +124,18 @@ private void generateKeystore(KeyStore keyStore, KeyStore trustStore, String sub
...
 
-        // generate a v3 certificate
-        return cryptoProvider.getCertificateUtils().generateV3Certificate(keyPair, keyPair.getPrivate(), caCert, subject);
+        // generate a v3 certificate chain
+        return this.cryptoProvider.getCertificateUtils().generateV3Certificate(keyPair, keyPair.getPrivate(), caCert, subject);
```
stianst: Revert, this change is not needed

test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java
```diff
@@ -52,7 +52,11 @@ public KeycloakServer getValue(InstanceContext<KeycloakServer, KeycloakIntegrati
         if (command.tlsEnabled()) {
             ManagedCertificates managedCert = instanceContext.getDependency(ManagedCertificates.class);
             command.option("https-key-store-file", managedCert.getKeycloakServerKeyStorePath());
-            command.option("https-key-store-password", managedCert.getKeycloakServerKeyStorePassword());
+            command.option("https-key-store-password", managedCert.getPassword());
+            command.option("https-trust-store-file", managedCert.getClientsTrustStorePath());
```
stianst: Don't set Keycloak's truststore to the truststore that should be used by clients; you need a separate truststore for what Keycloak should trust.

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
stianst, line 1: Please remove all unnecessary changes to `this.` - you don't need to use that unless you have both a field and a local variable with the same name.

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
```diff
@@ -56,18 +66,19 @@ public void testServerInfo() {
...
+        Assert.assertNames(info.getThemes().get("login"), "base", "keycloak", "keycloak.v2");
         Assert.assertNames(info.getThemes().get("welcome"), "keycloak");
 
         assertNotNull(info.getEnums());
 
         assertNotNull(info.getMemoryInfo());
         assertNotNull(info.getSystemInfo());
         assertNotNull(info.getCryptoInfo());
-        Assert.assertNames(info.getCryptoInfo().getSupportedKeystoreTypes(), KeystoreUtils.getSupportedKeystoreTypes());
+        Assert.assertNames(info.getCryptoInfo().getSupportedKeystoreTypes(), cryptoProvider.getCryptoProvider().getSupportedKeyStoreTypes()
```
stianst: This doesn't really assert anything; it basically just runs the same code on the client and server-sider, so don't think it serves any value

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
```diff
@@ -83,18 +94,15 @@ public void testServerInfo() {
                 .stream()
                 .filter(configProp -> Attributes.KEY_SIZE_KEY.equals(configProp.getName()))
                 .findFirst().orElseThrow(() -> new RuntimeException("Not found provider with ID 'rsa-generated'"));
-        Assert.assertNames(keySizeRep.getOptions(), KeyUtils.getExpectedSupportedRsaKeySizes());
+
+        Assert.assertNames(keySizeRep.getOptions(), String.join(",", cryptoProvider.getCryptoProvider().getSupportedRsaKeySizes()).split(","));
```
stianst: This doesn't really assert anything; it basically just runs the same code on the client and server-sider, so don't think it serves any value

stianst (commented):
Do we strictly speaking need the whole BC FIPS (+ mvn profile) and changing the crypto provider on the test side? Seems like it should work just fine without it if you update the ServerInfo (guess server could expose an option if FIPS is enabled or not) + have a slightly different way to set the key/truststore format.

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
lhanusov: assertion removed

tests/base/src/test/java/org/keycloak/tests/admin/ServerInfoTest.java
lhanusov: assertion removed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
lhanusov: "this." removed

test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java
lhanusov: removed

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
lhanusov: discussed in chat

quarkus/tests/junit5/src/main/java/org/keycloak/Keycloak.java
```diff
@@ -123,7 +123,7 @@ public Keycloak start(List<String> rawArgs) {
             addOptionIfNotSet(args, HttpOptions.HTTP_PORT);
             addOptionIfNotSet(args, HttpOptions.HTTPS_PORT);
 
-            boolean isFipsEnabled = ofNullable(getOptionValue(args, SecurityOptions.FIPS_MODE)).map(FipsMode::valueOf).orElse(FipsMode.DISABLED).isFipsEnabled();
+            boolean isFipsEnabled = ofNullable(getOptionValue(args, SecurityOptions.FIPS_MODE)).map(FipsMode::valueOfOption).orElse(FipsMode.DISABLED).isFipsEnabled();
```
lhanusov, line 127: @stianst This is a fix for embedded mode when running FIPS: fips-mode=strict or fips-mode=non-strict

However there is this KeycloakProcessor bug: https://github.com/keycloak/keycloak/issues/43975, which fails on nonexistent keycloak-tests-custom-scripts provider when loading keycloak-scripts.json. There is some strange dependency/classloader clash.

test-framework/core/src/main/java/org/keycloak/testframework/server/EmbeddedKeycloakServer.java
```diff
@@ -23,7 +23,9 @@ public void start(KeycloakServerConfigBuilder keycloakServerConfigBuilder) {
         enableTls = keycloakServerConfigBuilder.tlsEnabled();
 
         for(Dependency dependency : keycloakServerConfigBuilder.toDependencies()) {
-            builder.addDependency(dependency.getGroupId(), dependency.getArtifactId(), "");
+            if (!dependency.getGroupId().equals("org.bouncycastle")) {
+                builder.addDependency(dependency.getGroupId(), dependency.getArtifactId(), "");
```
lhanusov: @stianst Embedded server runs Keycloak class, which is already doing the bouncycastle [deployment](https://github.com/keycloak/keycloak/blob/main/quarkus/tests/junit5/src/main/java/org/keycloak/Keycloak.java#L259-L263). This is preventing the classloader to deploy it again.

quarkus/tests/junit5/src/main/java/org/keycloak/Keycloak.java
lhanusov: It can be fixed by removing keycloak-tests-custom-scripts dependency from the tests/base/pom.xml file. But then the PartialImport tests are failing within distribution mode for nonexistent dependency.

stianst (approved):

rmartinc (approved):
Good job @lhanusov and @stianst! LGTM!

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -30,31 +30,35 @@ public class ManagedCertificates {
...
         if (!CryptoIntegration.isInitialised()) {
             CryptoIntegration.setProvider(new DefaultCryptoProvider());
         }
         cryptoProvider = CryptoIntegration.getProvider();
-        initServerCerts();
+
+        serverKeystorePath = KEYSTORES_DIR.resolve("kc-testing-server-keystore" + "." + configBuilder.getKeystoreFormat().getPrimaryExtension());
+        clientsTruststorePath = KEYSTORES_DIR.resolve("kc-testing-clients-truststore" + "." + configBuilder.getKeystoreFormat().getPrimaryExtension());
+
+        password = configBuilder.getKeystoreFormat() == KeystoreUtil.KeystoreFormat.JKS ? "password".toCharArray() : "passwordpassword".toCharArray();
```
vaceksimon: I don't find this necessary. The password could be `"passwordpassword"` in both cases if it's needed for FIPS.

PR #43434, opened 2025-10-14
Issue #43232: Add `formatAnnotations`

Pankraz76:


testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/util/SecondBrowser.java
```diff
@@ -29,6 +29,5 @@
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.PARAMETER})
-@Qualifier
-public @interface SecondBrowser {
+@Qualifier public @interface SecondBrowser {
```
Pankraz76, line 32: is this good for keycloak? Imho this looks kind of odd.

stianst:
As mentioned in https://github.com/keycloak/keycloak/issues/43232 we will be rolling things like this out gradually; Inlining the annotation is also not something we will do; it makes the code less readable + would also require changing pretty much all existing annotations.

PR #43436, opened 2025-10-14
Proposed `licenseHeader`

Pankraz76:
enabler for:

even having multiple versions of the same licence:

image

related to:

- https://github.com/keycloak/keycloak/pull/43432

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1062,7 +1062,7 @@
                 <version>${project.version}</version>
                 <type>zip</type>
             </dependency>
-           <dependency>
+            <dependency>
                 <groupId>org.keycloak</groupId>
                 <artifactId>keycloak-saml-wildfly-subsystem</artifactId>
                 <version>${project.version}</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1523,15 +1523,38 @@
...
+                    </executions>
                     <configuration>
                         <java>
+                            <!--<forbidWildcardImports /> need rewrite to fix codebase.-->
                             <removeUnusedImports />
+                            <!--<importOrder>-->
+                            <!--    <order>java|javax|jakarta,org,org.keycloak,com,\#org.keycloak,\#</order>-->
+                            <!--</importOrder>-->
+                            <!--<licenseHeader>-->
+                            <!--    <file>licence.header</file>-->
```
Pankraz76, line 1542: This kind of change has to be executed by core maintainer as 4k files will change.

Pankraz76 (commented):
The test file is empty and fails the header step then. Could exclude but im wondering if this is still needed ?

stianst:
As mentioned in https://github.com/keycloak/keycloak/issues/43232 we will be rolling this out gradually; and as you've commented in the PR itself this really is a change that we have to do ourselves.

In terms of license header we need to discuss that as strictly speaking we do not actually need to have one in each file; so I'm rather inclined to not require a license header at all.

Pankraz76:
Thanks for the context. Yes, I meant this to be the version you would like and use, and that you would activate by yourself after the merge.

I'm not sure about the header stuff; I don't particularly care about it. All I know is that all the other open-source projects have a license header in each file. This might have some justifiable differences, but I'm not aware of them.

PR #43448, opened 2025-10-14
[quarkus-next] DatasourcesConfigurationTest fails

mabartos:
- Closes #43447

Safer asserts for avoiding any incompatibility is using List instead of the iterator.

Test for the Quarkus-next branch: https://github.com/keycloak/keycloak/pull/43449


vmuzikar (approved):
Thanks for the PR!

PR #43467, opened 2025-10-14
fix: adding type validation and lazily adding cli options

### LINKED ISSUE #42960: KC_VERBOSE is not honored for a fast start
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

Support for KC_VERBOSE works only if picocli fully runs - it's handled as a side effect in the Main command,

### Version

latest

### Regression

### Expected behavior

KC_VERBOSE should still work.

### Actual behavior

KC_VERBOSE does not work.

### How to Reproduce?

KC_VERBOSE=true ./kc.sh ...

### Anything else?

_No response_

### LINKED ISSUE #43466: Picocli refinements
### Description

Picocli logic only performs type validation when using the cli, this means that putting invalid values in the ENV or keycloak.conf will have different, and potentially quarkus specific messages.

Also to increase the performance of picocli we should defer adding all of the options until they are needed.

### Discussion

_No response_

### Motivation

To increase consistency and performance.

### Details

_No response_

shawkins:
closes: #43466

I realize that I tied two things that don't necessary need to go together into this pr. If you want they can be separated.

There's at least one more optimization I could do on top of this, but I think that the performance between using picocli and using fast-start is close enough that I can next resolve https://github.com/keycloak/keycloak/issues/42960 and https://github.com/keycloak/keycloak/issues/38790 by just eliminating fast start. @mabartos @vmuzikar what would be an acceptable delta between "start --optimized --verbose" and "start --optimized"?

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
@@ -31,6 +31,8 @@ import java.util.function.BiFunction;
 import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import org.keycloak.config.DeprecatedMetadata;
@@ -585,7 +587,14 @@ public class PropertyMapper<T> {
                     try {
                         Configuration.getConfig().convert(v, option.getComponentType());
                     } catch (Exception e) {
-                        throw new PropertyException("Invalid value for option %s: %s".formatted(getOptionAndSourceMessage(configValue), e.getMessage()));
+                        // strip the smallrye code if possible
+                        String message = e.getMessage();
+                        Pattern p = Pattern.compile("SRCFG\\d+: (.*)$");
+                        Matcher m = p.matcher(message);
+                        if (m.find()) {
+                            message = m.group(1);
+                        }
+                        throw new PropertyException("Invalid value for option %s: %s".formatted(getOptionAndSourceMessage(configValue), message));
                     }
                 }
 
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
@@ -579,6 +579,9 @@ public class PropertyMapper<T> {
             }
             try {
                 if (option.getComponentType() != String.class && option.getExpectedValues().isEmpty()) {
+                    if (v.isEmpty()) {
+                        throw new PropertyException("Invalid empty value for option %s".formatted(getOptionAndSourceMessage(configValue)));
+                    }
                     try {
                         Configuration.getConfig().convert(v, option.getComponentType());
                     } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
```diff
@@ -187,7 +187,24 @@ public void testInvalidArgumentType() {
         NonRunningPicocli nonRunningPicocli = pseudoLaunch("start-dev", "--http-port=a");
         assertEquals(CommandLine.ExitCode.USAGE, nonRunningPicocli.exitCode);
         assertThat(nonRunningPicocli.getErrString(),
-                containsString("Invalid value for option '--http-port': 'a' is not an int"));
+                containsString("Invalid value for option '--http-port': SRCFG00029: Expected an integer value, got \"a\""));
```
mabartos: It'd be nice to get rid of the converting status: "SRCFG00029"

mabartos (dismissed):
@shawkins Nice work! LGTM as a base for the following work.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
shawkins: It does look a little odd, but it matches the message you see if you currently put http-port=a in your keycloak.conf.

Unfortunately I don't think we have a great way other than string manipulation to eliminate this.

mabartos: As we (probably) know, the converter always includes the code when the converting fails, right? I would not object to remove everything until the first colon. We have quite a comprehensive testsuite, so we could test if everything works as expected. 

WDYT?

shawkins: Added logic to strip the code.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -577,6 +580,24 @@ public void validateValues(ConfigValue configValue, BiConsumer<ConfigValue, Stri
...
+                if (option.getComponentType() != String.class && option.getExpectedValues().isEmpty()) {
+                    if (v.isEmpty()) {
+                        throw new PropertyException("Invalid empty value for option %s".formatted(getOptionAndSourceMessage(configValue)));
+                    }
+                    try {
+                        Configuration.getConfig().convert(v, option.getComponentType());
+                    } catch (Exception e) {
+                        // strip the smallrye code if possible
+                        String message = e.getMessage();
+                        Pattern p = Pattern.compile("SRCFG\\d+: (.*)$");
```
mabartos, line 592:
```suggestion
                        Pattern p = Pattern.compile("^SRCFG\\d+: (.+)$");
```

nitpick

mabartos (approved):
@shawkins Looks good! Just one nitpick, but it's up to you if it will be used or not - I do not insist on that :)

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
shawkins: As I'm not sure that we're guarenteed that message will start with the code, my assumption is that we just wanted what appeared after the colon. 

Would you prefer this hard expectation?

mabartos: If we're not sure, let's keep it as is :)

PR #43487, opened 2025-10-15
Add `sortPom`

Pankraz76:

PR #43488, opened 2025-10-15
Apply `sortPom`

Pankraz76:
PoC for: 

- https://github.com/keycloak/keycloak/pull/43487


stianst:
As I've already mentioned - we are introducing these checks gradually, flooding us with PRs now is not helping.

Pankraz76:
sorry i´ve assumed you meant only spotless. Also this is just some foundation.

PR #43498, opened 2025-10-15
Possible overflow in brute force computation

### LINKED ISSUE #30939: Vulnerability in brute force detection settings
### Before reporting an issue

### Area

authentication

### Describe the bug

The following is a suspected vulnerability in the brute force detection settings in Keycloak version **24.0.1**

Setting the brute force mode to “lockout temporarily,” and then setting the “wait increment” and “max wait” to too high of a value breaks the locking out functionality.

If you configure brute force detection with the following settings, the locking out of an account will not work.

Brute force mode:    Lockout temporarily
Max login failures:       3
Wait increment:           21900 days
Max wait:             21900 days
Failure reset time:    15 minutes

Quick login check milliseconds:     1000
Minimum quick login wait:          1 minute

After 3 failed attempts of logging in, the user should be locked out, though by using the above settings prevents the locking out of the account.

I suspect this may be an integer overflow issue. If days are converted to seconds, at 21900 days, the total seconds approaches the 32 bit signed integer max.

### Version

24.0.1

### Regression

### Expected behavior

For the frontend to not allow for such a high value to be entered into the “wait increment” and “max wait”.
For the backend to protect from such a high value to be set.

### Actual behavior

If you configure brute force detection with the following settings, the locking out of an account will not work.

Brute force mode:    Lockout temporarily
Max login failures:       3
Wait increment:           21900 days
Max wait:             21900 days
Failure reset time:    15 minutes

Quick login check milliseconds:     1000
Minimum quick login wait:          1 minute

After 3 failed attempts of logging in, the user should be locked out, though by using the above settings prevents the locking out of the account.

### How to Reproduce?

If you configure brute force detection with the following settings, the locking out of an account will not work.

Brute force mode:    Lockout temporarily
Max login failures:       3
Wait increment:           21900 days
Max wait:             21900 days
Failure reset time:    15 minutes

Quick login check milliseconds:     1000
Minimum quick login wait:          1 minute

After 3 failed attempts of logging in, the user should be locked out, though by using the above settings prevents the locking out of the account.

### Anything else?

_No response_

mposolda:
closes #30939

The configuration attributes related to brute-force have type `Integer` on `RealmRepresentation` (For example `RealmRepresentation.waitIncrementSeconds` ). And admin console UI as well as server does not allow to configure bigger value than `Integer.MAX_VALUE` . This is correct behavior. Also it is not possible to configure negative values for those attributes, which is also correct.

However the overflow can happen during brute-force calculation as there are various types configured as `Integer` and computation is doing some multiplications, dividing and other operations. In the fix, I am trying to use `long` during calculation on most places to avoid overflow and negative value passed to `userLoginFailure.failedLoginNotBefore` (which could happen in some circumstances before this fix).

### DIFF
--- a/services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
+++ b/services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
@@ -89,12 +89,12 @@ public class DefaultBruteForceProtector implements BruteForceProtector {
         userLoginFailure.incrementFailures();
         logger.debugv("new num failures: {0}", userLoginFailure.getNumFailures());
 
-        int waitSeconds = 0;
-        if(!(realm.isPermanentLockout() && realm.getMaxTemporaryLockouts() == 0)) {
-            if(RealmRepresentation.BruteForceStrategy.MULTIPLE.equals(realm.getBruteForceStrategy())) {
-                waitSeconds = realm.getWaitIncrementSeconds() *  (userLoginFailure.getNumFailures() / realm.getFailureFactor());
+        long waitSeconds = 0L;
+        if (!(realm.isPermanentLockout() && realm.getMaxTemporaryLockouts() == 0)) {
+            if (RealmRepresentation.BruteForceStrategy.MULTIPLE.equals(realm.getBruteForceStrategy())) {
+                waitSeconds = (long) realm.getWaitIncrementSeconds() *  ((long) userLoginFailure.getNumFailures() / realm.getFailureFactor());
             } else {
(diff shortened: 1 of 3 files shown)


ahus1 (commented):
Thank you for the PR. The approach with `long` should work. 

If you want to make it more explicit, there are also the Java `Math.multiplyExact()` and `Math.addExact()` which will throw an `ArithmeticException` on overflow.

This is just a comment - feel free to proceed as you see fit.

mposolda:
@ahus1 Thanks for the suggestion! I am slightly more towards to use the current approach. Also because the aim is to rather "prevent overflow" instead of "throw the exception on overflow" .

graziang (approved):
@mposolda thanks!

mposolda:
@graziang Thanks!

Issue #42091, opened 2025-08-25
[OID4VCI] SD-JWT implementation ignores Keybinding JWT


Captain-P-Goldfish:
### Description

The current SD-JWT implementation is still ignoring the Keybinding JWT if appended to the SD-JWT.

IngridPuppet:
Hello @Captain-P-Goldfish - In the `SdJwtVPVerificationTest` suite, there are multiple tests loading SD-JWTs with Keybinding JWTs. Notice that you use the `SdJwtVP.of()` construct. Check this one for example:

https://github.com/keycloak/keycloak/blob/36cbbbc7b4835b1a091449677a7e2d237892c50f/core/src/test/java/org/keycloak/sdjwt/sdjwtvp/SdJwtVPVerificationTest.java#L62-L72

Doesn't that address the concern behind this ticket?

Captain-P-Goldfish:
I did not meant the verification.
We are currently building a Verifier in Keycloak and thus I tried to create an SD-JWT with Keybinding for unit-testing which proved a little bit nasty, because the Keybinding-JWT was not added to the SD-JWT and it was not easily possible to add a x5c-header to the SD-JWT. The x5c header is currently not part of the specification but it is necessary for the issuer to add a certificate for as long as the trust-infrastructure is not completely defined yet. Regarding these points my unit-test-WalletMock ended up like this with reflections:

```java
VpToken vpToken = new VpToken();
    for ( DcqlCredential dcqlCredential : jwtSecuredAuthRequest.getDcqlQuery().getDcqlCredentials() )
    {
      ObjectNode sdJwtClaims = new ObjectNode(JsonNodeFactory.instance);
      DisclosureSpec.Builder disclosureBuilder = DisclosureSpec.builder();
      for ( DcqlClaim claim : dcqlCredential.getClaims() )
      {
        // TODO nested objects like address are disclosed as complete objects. How to disclose simple nested
        // attributes?
        disclosureBuilder.withUndisclosedClaim(claim.getPath().getFirst());
        sdJwtClaims.set(claim.getPath().getFirst(), getNode(claim.getPath(), sdJwtClaims));
      }

      JWK keybindingKey = JWKBuilder.create().ec(signingKey.getPublicKey());
      ObjectNode keyBindingJwkNode = new ObjectNode(JsonNodeFactory.instance);
      keyBindingJwkNode.set("jwk", JsonSerialization.mapper.convertValue(keybindingKey, ObjectNode.class));
[...]

PR #43506, opened 2025-10-15
[OID4VCI] Tolerate clock skew in SD-JWT time checks

### LINKED ISSUE #43456: Run time comparisons in SD-JWT and SD-JWT VP verification with account of possible clock skew
### Description

Account for clock skew in time checks under the `core/sd-jwt` module.

### Discussion

_No response_

### Motivation

In real-life scenarios, clocks may not be in sync across devices, causing time checks on critical fields such as `iat`, `nbf`, and `exp` to unexpectedly fail. Tolerating few seconds of clock skew generally restores the practicality of the affected feature without infringing on security. This the most common solution to this well-known and discussed issue.

### Details

See similar tickets opened on this repository:
- https://github.com/keycloak/keycloak/issues/34894
- https://github.com/keycloak/keycloak/issues/34805
- https://github.com/keycloak/keycloak/issues/8966

IngridPuppet:
Closes #43456

### DIFF
--- a/core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
+++ b/core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
@@ -24,66 +24,74 @@ package org.keycloak.sdjwt;
  */
 public class TimeClaimVerificationOpts {
 
+    public static final int DEFAULT_LEEWAY_SECONDS = 10;
+
+    // These options configure whether the respective time claims must be present
+    // during validation. They will always be validated if present.
+
+    private final boolean requireIssuedAtClaim;
+    private final boolean requireExpirationClaim;
+    private final boolean requireNotBeforeClaim;
+
     /**
      * Tolerance window to account for clock skew when checking time claims
      */
-    public static final int DEFAULT_LEEWAY_SECONDS = 10;
-
-    private final boolean validateIssuedAtClaim;
-    private final boolean validateExpirationClaim;
-    private final boolean validateNotBeforeClaim;
     private final int leewaySeconds;
 
     protected TimeClaimVerificationOpts(
-            boolean validateIssuedAtClaim,
-            boolean validateExpirationClaim,
+            boolean requireIssuedAtClaim,
+            boolean requireExpirationClaim,
             boolean validateNotBeforeClaim,
             int leewaySeconds
     ) {
-        this.validateIssuedAtClaim = validateIssuedAtClaim;
-        this.validateExpirationClaim = validateExpirationClaim;
-        this.validateNotBeforeClaim = validateNotBeforeClaim;
+        this.requireIssuedAtClaim = requireIssuedAtClaim;
+        this.requireExpirationClaim = requireExpirationClaim;
+        this.requireNotBeforeClaim = validateNotBeforeClaim;
         this.leewaySeconds = leewaySeconds;
     }
 
-    public boolean mustValidateIssuedAtClaim() {
-        return validateIssuedAtClaim;
+    public boolean mustRequireIssuedAtClaim() {
+        return requireIssuedAtClaim;
     }
 
-    public boolean mustValidateExpirationClaim() {
-        return validateExpirationClaim;
+    public boolean mustRequireExpirationClaim() {
+        return requireExpirationClaim;
     }
 
-    public boolean mustValidateNotBeforeClaim() {
-        return validateNotBeforeClaim;
+    public boolean mustRequireNotBeforeClaim() {
+        return requireNotBeforeClaim;
     }
 
     public int getLeewaySeconds() {
         return leewaySeconds;
     }
 
-    protected static class Builder<T extends Builder<T>> {
+    public static <T extends Builder<T>> Builder<T> builder() {
+        return new Builder<>();
+    }
 
-        protected boolean validateIssuedAtClaim = false;
-        protected boolean validateExpirationClaim = true;
-        protected boolean validateNotBeforeClaim = true;
+    public static class Builder<T extends Builder<T>> {
+
+        protected boolean requireIssuedAtClaim = true;
+        protected boolean requireExpirationClaim = true;
+        protected boolean requireNotBeforeClaim = true;
         protected int leewaySeconds = DEFAULT_LEEWAY_SECONDS;
 
         @SuppressWarnings("unchecked")
-        public T withValidateIssuedAtClaim(boolean validateIssuedAtClaim) {
-            this.validateIssuedAtClaim = validateIssuedAtClaim;
+        public T withRequireIssuedAtClaim(boolean requireIssuedAtClaim) {
+            this.requireIssuedAtClaim = requireIssuedAtClaim;
             return (T) this;
         }
 
         @SuppressWarnings("unchecked")
-        public T withValidateExpirationClaim(boolean validateExpirationClaim) {
-            this.validateExpirationClaim = validateExpirationClaim;
+        public T withRequireExpirationClaim(boolean requireExpirationClaim) {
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public static final int DEFAULT_LEEWAY_SECONDS = 10;
+
+    private final boolean validateIssuedAtClaim;
+    private final boolean validateExpirationClaim;
+    private final boolean validateNotBeforeClaim;
+    private final int leewaySeconds;
+
+    protected TimeClaimVerificationOpts(
+            boolean validateIssuedAtClaim,
```
mposolda: Thanks for the PR. I wonder if options `validateIssuedAtClaim` , `validateExpirationClaim` and `validateNotBeforeClaim` are needed at all? Can we just always validate those? Or is it any use-case when validating of the claims should be omitted? This applies for the parent class `IssuerSignedJwtVerificationOpts` .

I suppose it might not be in the scope of this PR to remove those options (as PR is supposed to add clock-skew support), but just taking this PR as an opportunity to discuss this topic :-)

mposolda (commented):
@IngridPuppet I've added comment inline. Could you please check?

core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
IngridPuppet: Thank you for raising this point. The initial reason for adding these options was because the spec used to mandate the presence of the corresponding time claims should it be the verifier's policy to enforce their validity.

https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-11.html#name-selectively-disclosable-val
> Verifiers therefore MUST ensure that all claims they deem necessary for checking the validity of an SD-JWT in the given context are present (or disclosed, respectively) during validation of the SD-JWT. This is implemented in the last step of the verification defined in [Section 8.1](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-11.html#sd_jwt_verification).

In practice, issuers commonly happen not to include these time claims, especially `iat` and `nbf`. The consequence is that enforcing their presence as we validate them basically makes the feature impractical. Providing options enables the verifier to _configure_ its policy.

However, I noticed the spec was clarified in recent versions to allow skipping the validation if the claims are not present should they not be "required validity-controlling claims" as per the verifier. In my opinion, the options will remain (maybe renamed to suggest requirement) but the claims always *validated* if present. Do you suggest I create a ticket for that? Do you have a different understanding or would like to provide further clarification?

https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-22.html#name-verification-of-the-sd-jwt
> Check that the SD-JWT is valid using claims such as nbf, exp, and aud in the processed payload, **if present.**

Captain-P-Goldfish:
maybe we can skip this pull-request. I am currently rebuilding the validation of keybinding-Jwts using predicates as it is done with the TokenVerifier:

example:

in class **TokenVerifier**
```java
public static class ExpCheck implements TokenVerifier.Predicate<JsonWebToken>
    {

        private final int clockSkewSeconds;

        private boolean isOptional;

        public ExpCheck(int clockSkewSeconds) {
            this(clockSkewSeconds, false);
        }

        public ExpCheck(int clockSkewSeconds, boolean isOptional) {
            this.clockSkewSeconds = clockSkewSeconds;
            this.isOptional = isOptional;
        }

        @Override
        public boolean test(JsonWebToken jsonWebToken) throws VerificationException {
            Long expiration = jsonWebToken.getExp();
            if (expiration == null) {
                if (isOptional) {
                    return true;
                }
                else {
                    throw new VerificationException("Missing required claim 'exp'");
                }
            }
            long now = Instant.now().getEpochSecond();

            if (expiration < now - clockSkewSeconds) {
                throw new VerificationException(String.format("Token has expired: now: '%s', exp: '%s'",
                                                              now,
                                                              expiration));
            }
            return true;
        }
    }
```

class **KeyBindingJwtVerificationOpts**
```java
public KeyBindingJwtVerificationOpts build() {
            if (keyBindingRequired && (aud == null || nonce == null || nonce.isEmpty())) {
                throw new IllegalArgumentException(
                        "Missing `nonce` and `aud` claims for replay protection"
                );
            }

            if (useDefaultVerifiers) {
                if (aud != null) {
                    contentVerifiers.add(new TokenVerifier.AudienceCheck(aud));
                }
                if (nonce != null) {
                    contentVerifiers.add(new TokenVerifier.ClaimCheck(IDToken.NONCE, nonce, true));
                }
                contentVerifiers.add(new TokenVerifier.IatLifetimeCheck(clockSkew, allowedMaxAge, true));
                contentVerifiers.add(new TokenVerifier.NbfCheck(clockSkew, true));
                contentVerifiers.add(new TokenVerifier.ExpCheck(clockSkew, true));
                contentVerifiers.add(new TokenVerifier.AudienceCheck(aud));
            }

            return new KeyBindingJwtVerificationOpts(keyBindingRequired,
                                                     allowedMaxAge,
                                                     nonce,
                                                     aud,
                                                     contentVerifiers
            );
        }
```

IngridPuppet:
Hi @Captain-P-Goldfish - The proposal will indeed address the clockSkew issue and enable configuring the verifier's policy regarding whether or not to enforce the presence of time claims prior to their validation. However, you would want to update `IssuerSignedJwtVerificationOpts` so the same logic applies. I'm also a bit concerned about how the snippet handles `nonce` and `aud` claims; they might be specified when building the options but not enforced if no matching TokenVerifier ends up being added to `contentVerifiers`. Is there a PR already, so we close this one in favor that?

Captain-P-Goldfish:
Thansk for your comments @IngridPuppet .
I am still working on it, so there is no PR yet.
The changes are already applied to the `IssuerSignedJwtVerifcationOpts`.
I have managed to get the code compilable again. Now I will take care on the unit-tests that all tests will run successfully again. After that I will address your comments and then I can send the PR. Hopefully I can send it tomorrow, if I am not completely getting occupied by zomm-meetings again ^^°

core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
mposolda: Exactly! I suggest to rename the options for maybe something like `requireValidateIssuedAtClaim` instead of `validateIssuedAtClaim` . Then we can perhaps do:
- In case that claim (like `iat` for example) is present, then it will be always validated (regardless of option value)
- In case that option is `requireValidateIssuedAtClaim` is `true`, then the claim will be required to be always validated. Hence validation fails if the `iat` claim will not be present.

Do we agree on this? If yes, feel free to either update this PR or create follow-up PR for this (whatever you prefer). If you prefer follow-up, please let me know, so I can possibly approve this (if also someone else from the OIDC SIG reviews and approves the PR).

IngridPuppet: Yes, we are on par. Because we agreed on a clear objective and that the change is quite straightforward to implement, I updated this PR directly (https://github.com/keycloak/keycloak/pull/43506/commits/6059dfd209870805fd159529ea6a459c1d346e18). Please could you check?

Also, FYI, @Captain-P-Goldfish plans to send a major update of the SD-JWT API, which will most likely make this PR obsolete. Because the major update might take longer than planned, I guess this PR should not be abandoned yet. The discussion on the update is taking place here in https://github.com/keycloak/keycloak/issues/42091.

mposolda (approved):
@IngridPuppet Thanks!

PR #43513, opened 2025-10-15
Export routes array from Account Console and Admin Console UI libraries

### LINKED ISSUE #43512: Export default routes from npm UI packages
### Description

Extending the Admin Console and Account Console while maintaining the existing routes is quite cumbersome at the moment. The [quickstart](https://github.com/keycloak/keycloak-quickstarts/tree/main/extension/extend-account-console-node) for extending the Account Console [chooses to recreate all of the existing route objects](https://github.com/keycloak/keycloak-quickstarts/blob/main/extension/extend-account-console-node/src/routes.tsx), which isn't too bad since there are only 11 items. However, the existing routes can't fully be recreated because the [Organization page](https://github.com/keycloak/keycloak/blob/main/js/apps/account-ui/src/organizations/Organizations.tsx) is not exported in the [index.ts](https://github.com/keycloak/keycloak/blob/main/js/apps/account-ui/src/index.ts) file. The [quickstart](https://github.com/keycloak/keycloak-quickstarts/tree/main/extension/extend-admin-console-node) for the Admin Console [doesn't even attempt to add the existing routes](https://github.com/keycloak/keycloak-quickstarts/blob/main/extension/extend-admin-console-node/src/routes.tsx) to the custom UI, probably because there are [over 100 unique route objects](https://github.com/keycloak/keycloak/blob/main/js/apps/admin-ui/src/routes.tsx) in that library. A simple fix to this problem would be to export the `routes` array from the `src/routes.ts` file for both libraries. With that array, users can easily customize and add pages to the current React router without having to copy and maintain tens to hundreds of existing route objects.

### Value Proposition

Exporting the Account Console and Admin Console `routes` array will make extending the Keycloak UIs more viable and easier to maintain for developers and administrators.

### Goals

- Export the `routes` array from both the `@keycloak/keycloak-account-ui` and `@keycloak/keycloak-admin-ui` packages
- Export the `Organizations` page component that is currently missing from the `@keycloak/keycloak-account-ui` package entry file

### Non-Goals

This enhancement only affects the npm UI packages and does not require changes to the `keycloak-quickstarts` library. The fix is strictly additive and does not affect current implementations or uses of these libraries.

### Discussion

_No response_

### Notes

_No response_

joeknowsdev:
Closes #43512

### DIFF
--- a/js/apps/account-ui/src/index.ts
+++ b/js/apps/account-ui/src/index.ts
@@ -38,6 +38,7 @@ export { ResourcesTab } from "./resources/ResourcesTab";
 export { ResourceToolbar } from "./resources/ResourceToolbar";
 export { SharedWith } from "./resources/SharedWith";
 export { Oid4Vci } from "./oid4vci/Oid4Vci";
+export { Organizations } from "./organizations/Organizations";
 export { ShareTheResource } from "./resources/ShareTheResource";
 export {
   deleteConsent,
@@ -57,3 +58,4 @@ export type { Environment as AccountEnvironment } from "./environment";
 export { KeycloakProvider, useEnvironment } from "@keycloak/keycloak-ui-shared";
 export { useAccountAlerts } from "./utils/useAccountAlerts";
 export { usePromise } from "./utils/usePromise";
+export { routes } from "./routes";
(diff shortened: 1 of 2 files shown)


ahus1:
@jonkoops / @ssilvert - Maybe you can help here with your expertise?

jonkoops:
@edewit is responsible for the UI libraries. So I'll defer to him, but I guess there is no harm in exporting more from here. Just note, we don't provide any stability guarantees for this API AFAIK, so routes might change between minor versions.

jonkoops (approved):

edewit (approved):

ahus1 (approved):

PR #43526, opened 2025-10-16
Add restart step provider, replacing the recurring config option

### LINKED ISSUE #42910: [RLM] - Restarting a workflow for a resource based on the step chain
### Description

Currently, the workflow definition defines a `recurring` property to restart the workflow for a particular resource after the last step in the step chain is executed.

We can improve the workflow definition and make configuration easier and more intuitive by deciding whether or not a step should be restarted purely based on how the steps are defined in a workflow. Therefore, eliminating yet another property in the workflow definition and making the step chain more intuitive and self-describing.

This improvement is basically about removing the `recurring` property so that the step chain can be configured with a specific step that explicitly restart the workflow for a given resource. In other words, it starts over from the first step after executing the last step in the chain:

```
"steps": [
    {
        "uses": "step-1"
    },
    {
        "uses": "step-2"
    },
    {
        "uses": "restart"
    }
        
]
```

We can also consider introducing a more generic step that mimics the behavior of a "goto" where users can specify a specific step/position in the step chain, being it the first step in the chain or some other in whatever position:

```
"steps": [
    {
        "uses": "step-1"
    },
    {
        "uses": "step-2"
    },
    {
        "uses": "goto",
        "with": {
            "position": 2
         }
    }
        
]
```

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

sguilhen:
Closes #42910

### DIFF
--- a/core/src/main/java/org/keycloak/representations/workflows/WorkflowConstants.java
+++ b/core/src/main/java/org/keycloak/representations/workflows/WorkflowConstants.java
@@ -11,8 +11,6 @@ public final class WorkflowConstants {
     public static final String CONFIG_ON_EVENT = "on";
     public static final String CONFIG_RESET_ON = "reset-on";
     public static final String CONFIG_NAME = "name";
-    public static final String CONFIG_RECURRING = "recurring";
-    public static final String CONFIG_SCHEDULED = "scheduled";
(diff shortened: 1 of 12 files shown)


pedroigor (approved):

vramik (approved):
Thanks @sguilhen

PR #43539, opened 2025-10-16
Restart workflow basen on `concurrency/cancel-if-running` option rath…

### LINKED ISSUE #42911: [RLM] - Canceling workflows for a given resources when a new event is triggered
### Description

Currently, the workflow definition defines a `reset-on` property to define a set of events that will cancel the current execution of a workflow for a resource and start it all over again from the very first action defined in the step chain.

This improvement is basically about removing the `reset-on` property and introducing a `concurrency` property to allow a more descriptive setting on how the workflow should handle concurrent events. In other words, how workflows should react when the same event that triggered it is re-triggered.

Right now, the `concurrency` property will have a single setting:

```
"concurrency": {
    "cancel-if-running": true 
}
```

By enabling `cancel-if-running`, the workflow will behave exactly the same as if defined with `reset-on: ["event"]`, but automatically cancelling and restarting the workflow for a resource based on the event(s) defined in the `on` property (trigger events).

In the future, the `concurrency` property can be expanded to enable additional behaviors, such as defining an alias/group for multiple executions of a workflow.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

vramik:
…er than `reset-on` option

Closes #42911

### DIFF
--- a/core/src/main/java/org/keycloak/representations/workflows/WorkflowConcurrencyRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/workflows/WorkflowConcurrencyRepresentation.java
@@ -1,0 +1,38 @@
+package org.keycloak.representations.workflows;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+import static org.keycloak.representations.workflows.WorkflowConstants.CONFIG_CANCEL_IF_RUNNING;
+
+public class WorkflowConcurrencyRepresentation {
+
+    @JsonProperty(CONFIG_CANCEL_IF_RUNNING)
+    private Boolean cancelIfRunning;
+
+    // A no-argument constructor is needed for Jackson deserialization
+    public WorkflowConcurrencyRepresentation() {}
+
+    public WorkflowConcurrencyRepresentation(Boolean cancelIfRunning) {
+        this.cancelIfRunning = cancelIfRunning;
+    }
+
+    public Boolean isCancelIfRunning() {
+        return cancelIfRunning;
+    }
+
+    @Override
+    public int hashCode() {
+        return cancelIfRunning != null ? cancelIfRunning.hashCode() : 0;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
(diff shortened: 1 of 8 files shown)


vramik:
@sguilhen PTAL, I am aware it will need a rebase once your PR is merged, still wanted to send it sooner than later.

core/src/main/java/org/keycloak/representations/workflows/WorkflowConcurrencyRepresentation.java
```diff
@@ -0,0 +1,36 @@
+package org.keycloak.representations.workflows;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+public class WorkflowConcurrencyRepresentation {
+
+    @JsonProperty("cancel-if-running")
```
pedroigor:
```suggestion
    @JsonProperty(WorkflowConstants.CONFIG_CANCEL_IF_RUNNING)
```

model/jpa/src/main/java/org/keycloak/models/workflow/EventBasedWorkflowProvider.java
```diff
@@ -111,9 +111,7 @@ protected WorkflowsManager getManager() {
         return manager;
     }
 
-    protected boolean isResetEvent(WorkflowEvent event) {
-        return model.getConfig()
-                .getOrDefault(CONFIG_RESET_ON, List.of())
-                .contains(event.getOperation().name());
+    protected boolean isConcurrencyCancelIfRunning() {
```
pedroigor:
```suggestion
    protected boolean isCancelIfRunning() {
```

pedroigor: More aligned with the name at `WorkflowisConcurrencyRepresentation#isCancelIfRunning`

sguilhen (approved):

sguilhen (approved):

pedroigor (approved):

PR #43542, opened 2025-10-16
Ability to enable/disable feature via single property

### LINKED ISSUE #43541: Ability to enable/disable feature via single property
### Description

At this moment, enabling and disabling features can be done only via the configuration options that accept a list of features. It means when a user/customer has enabled multiple preview features (or multiple disabled) and then the state needs to be changed, the user/customer needs to go over the big list and remove the feature from the list. The UX is not very good in this manner and it's also a problem in the testsuite as we need to do multiple workarounds on how to enable/disable features in these lists. 

It'd be better to have a single wildcard option for these features and be possible to change the state of the feature via a single property.

We've already discussed it with @shawkins on some team meetings that would be good to have it. 

### Value Proposition

Improved UX, better management for the operator, and more consistent handling in the testsuite. 

### Goals

- provide the wildcard option
- validate these inputs
- improve the testsuite
- check how it could help in operator

### Non-Goals

n/a

### Discussion

_No response_

### Notes

_No response_

mabartos:
- Closes #43541

@shawkins This is the initial form of how it could work with the single property.

### DIFF
--- a/common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
+++ b/common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
@@ -2,17 +2,25 @@ package org.keycloak.common.profile;
 
 import org.keycloak.common.Profile;
 
+import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
 
+// Features configuration based on the option 'feature-<name>'
 public class SingleProfileConfigResolver implements ProfileConfigResolver {
     private final Map<String, Boolean> features;
 
     public SingleProfileConfigResolver(Map<String, Boolean> features) {
-        this.features = features;
+        this.features = Optional.ofNullable(features).orElseGet(Collections::emptyMap);
     }
 
     @Override
     public Profile.ProfileName getProfileName() {
+        Boolean state = features.get(Profile.ProfileName.PREVIEW.name().toLowerCase());
+        if (state != null && state) {
+            return Profile.ProfileName.PREVIEW;
+        }
+
         return null;
     }
 
--- a/common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
+++ b/common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
@@ -16,11 +16,7 @@ public class SingleProfileConfigResolver implements ProfileConfigResolver {
 
     @Override
     public Profile.ProfileName getProfileName() {
-        Boolean state = features.get(Profile.ProfileName.PREVIEW.name().toLowerCase());
-        if (state != null && state) {
-            return Profile.ProfileName.PREVIEW;
-        }
-
+        // not supporting profiles yet - see https://github.com/keycloak/keycloak/issues/44003
         return null;
     }
 
--- a/common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
+++ b/common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
@@ -1,0 +1,28 @@
+package org.keycloak.common.profile;
+
+import org.keycloak.common.Profile;
+
+import java.util.Map;
+
+public class SingleProfileConfigResolver implements ProfileConfigResolver {
+    private final Map<String, Boolean> features;
+
+    public SingleProfileConfigResolver(Map<String, Boolean> features) {
+        this.features = features;
+    }
+
+    @Override
+    public Profile.ProfileName getProfileName() {
+        return null;
+    }
+
+    @Override
+    public FeatureConfig getFeatureConfig(String feature) {
+        Boolean state = features.get(feature);
+        if (state == null) {
+            return FeatureConfig.UNCONFIGURED;
+        }
+
+        return state ? FeatureConfig.ENABLED : FeatureConfig.DISABLED;
+    }
+}
--- a/docs/guides/server/features.adoc
+++ b/docs/guides/server/features.adoc
@@ -5,13 +5,29 @@
 <@tmpl.guide
 title="Enabling and disabling features"
 summary="Configure {project_name} to use optional features."
-includedOptions="features features-*">
+includedOptions="features features-* feature-*">
 
 {project_name} has packed some functionality in features, including some disabled features, such as Technology Preview and deprecated features. Other features are enabled by default, but you can disable them if they do not apply to your use of {project_name}.
 
 == Enabling features
 
-Some supported features, and all preview features, are disabled by default. To enable a feature, enter this command:
+Some supported features, and all preview features, are disabled by default.
+You can enable feature either via single option or including it into the list of enabled feature.
+
+=== Single option
+
+You can enable the specific feature `<name>` as follows:
+
+<@kc.build parameters="--feature-<name>=enabled|disabled|vX"/>
+
+Possible values are `enabled`, `disabled`, or a specific version of the feature that should be enabled.
+For example, to enable `rolling-updates:v2` and `token-exchange`, enter this command:
+
+<@kc.build parameters="--feature-rolling-updates=v2 --feature-token-exchange=enabled"/>
+
+The single-option mechanism is useful when updating long feature lists is cumbersome or when you want to modify a specific feature without overriding the entire list in a pre-built image.
+
+=== List of enabled features
 
 <@kc.build parameters="--features=\"<name>[,<name>]\""/>
 
@@ -23,6 +39,8 @@ To enable all preview features, enter this command:
 
 <@kc.build parameters="--features=\"preview\""/>
 
+=== Versioning
+
 Enabled feature may be versioned, or unversioned.  If you use a versioned feature name, e.g. feature:v1, that exact feature version will be enabled as long as it still exists in the runtime.  If you instead use an unversioned name, e.g. just feature, the selection of the particular supported feature version may change from release to release according to the following precedence:
 
 . The highest default supported version
@@ -33,7 +51,22 @@ Enabled feature may be versioned, or unversioned.  If you use a versioned featur
 
 == Disabling features
 
-To disable a feature that is enabled by default, enter this command:
+To disable a feature that is enabled by default, you can use a single option or a list of disabled features.
+When a feature is disabled, all versions of that feature are disabled.
+
+=== Single option
+
+You can disable the specific feature `<name>` as follows:
+
+<@kc.build parameters="--feature-<name>=disabled"/>
+
+For example, to disable `dpop` and `recovery-codes`, enter this command:
+
+<@kc.build parameters="--feature-dpop=disabled --feature-recovery-codes=disabled"/>
+
+The single-option mechanism is useful when updating long feature lists is cumbersome or when you want to modify a specific feature without overriding the entire list in a pre-built image.
+
+=== List of disabled features
 
 <@kc.build parameters="--features-disabled=\"<name>[,<name>]\""/>
 
@@ -42,8 +75,6 @@ For example to disable `impersonation`, enter this command:
 <@kc.build parameters="--features-disabled=\"impersonation\""/>
 
 It is not allowed to have a feature in both the `features-disabled` list and the `features` list.
-
-When a feature is disabled all versions of that feature are disabled.
 
 == Supported features
 
--- a/docs/guides/server/features.adoc
+++ b/docs/guides/server/features.adoc
@@ -25,6 +25,10 @@ For example, to enable `rolling-updates:v2` and `token-exchange`, enter this com
 
 <@kc.build parameters="--feature-rolling-updates=v2 --feature-token-exchange=enabled"/>
 
+To enable all preview features, enter this command:
+
+<@kc.build parameters="--feature-preview=enabled"/>
+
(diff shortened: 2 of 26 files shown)


shawkins:
> @shawkins This is the initial form of how it could work with the single property.

Thank you for taking this on. 

What do you think of expressing this as:

`--feature-<name>=vX|enabled|disabled`

stianst:
Is this a real issue? I don't think folks need to enable/disable that many features. If they really do I think we should consider something like also supporting a separate file.

If the argument is that folks will have many features to enable/disable this just makes it even worse IMO as you basically end up with more crud, not less.

stianst:
Is this not also going to bloat the help completely?

You'd have --features that list all features, then also additionally have a separate option for each feature?

shawkins:
> I don't think folks need to enable/disable that many features.

We're starting to add a lot of features. We're also not just talking about people trying things out locally.

Let's say I've been provided a containerized augmented image. Now I want to try out the new serdes optimization feature. 

Without knowing the list of features that may already be configured in that augmented image, I cannot simply try out the feature as I may change the feature configuration in unexpected ways.

> If they really do I think we should consider something like also supporting a separate file.

That will be more difficult to manage from the operator.

> You'd have --features that list all features, then also additionally have a separate option for each feature?

I would vote for eventually getting rid of the list based option.

stianst:
> > I don't think folks need to enable/disable that many features.
> 
> We're starting to add a lot of features. We're also not just talking about people trying things out locally.
> 
> Let's say I've been provided a containerized augmented image. Now I want to try out the new serdes optimization feature.
> 
> Without knowing the list of features that may already be configured in that augmented image, I cannot simply try out the feature as I may change the feature configuration in unexpected ways.

That's a good point, but would mean we'd need to handle what happens if the same feature is listed multiple times; could be `--features-foo=enabled --features-foo=disabled` for example

> 
> > If they really do I think we should consider something like also supporting a separate file.
> 
> That will be more difficult to manage from the operator.
> 
> > You'd have --features that list all features, then also additionally have a separate option for each feature?
> 
> I would vote for eventually getting rid of the list based option.

Please no, that is super convenient and quick to manage if you want to manage enabled/disabled features in one go, which at least personally is what I want

stianst:
Okay, so @shawkins convinced me that we probably should do this; I do like `--feature-<name>=vX|enabled|disabled`, but would we also have a shorthand to enable `--feature-<name>` means it's enabled?

Another thing I'd like to know is how this will look like in the CLI help, and the guide docs needs updating as well.

One proposal would be that help would not list all the `--feature-name` options, as that would just flood it with those; similarly to help for `--features` and `--features-disabled` shows a huge block of text right now (which really has no value has it's so big and messy that I just use the guide instead).

Could we do some separate help command for features? That lists all the features, their description, etc? Something similar to what the guide does?

mabartos:
> What do you think of expressing this as: `--feature-<name>=vX|enabled|disabled`.

Yep, it might be nice. The `vX` may work only if it's sure that we would not have to disable any specific version of a feature. Right now, you can disable only the whole feature, and not particular feature versions, so it's ok for now. 

So, the approach suggested by @shawkins would look like this:
1. Enable Rolling Updates v2: `--feature-rolling-updates=v2`
2. Enable Transient users: `--feature-transient-users=enabled`
3. Disable Device flow: `--feature-device-flow=disabled`

And potentially, as suggested by @stianst, enable the feature like `--feature-transient-users` that could enable the feature based on the [rules](https://www.keycloak.org/server/features#_enabling_features) on the versions. 

-----------------------------------

> Is this not also going to bloat the help completely? ... Another thing I'd like to know is how this will look like in the CLI help, and the guide docs needs updating as well.

The help will be clean. There will be only one option (called wildcard option) that accepts feature names - sth like the `--feature-<name>`, and no existing features will be listed there. In the description, there will be a note on how to list features (see below)

-------------------------

> Could we do some separate help command for features? That lists all the features, their description, etc? Something similar to what the guide does?

Yes, it's on our radar as well as it'd be nice to have the capability of listing features in the distribution itself with some more information + maybe even info whether the feature is enabled or not for already augmented distro. I'll create an issue for this. 

---------------------------

> I would vote for eventually getting rid of the list based option.

I think these 2 approaches can coexist as you can override the feature state by using this new option.

shawkins:
> And potentially, as suggested by @stianst, enable the feature like --feature-transient-users that could enable the feature based on the [rules](https://www.keycloak.org/server/features#_enabling_features) on the versions.

That would need to be done separately. There are several places in our code where we make the assumption that our options must have an argument.

> I think these 2 approaches can coexist as you can override the feature state by using this new option.

Ok, sounds like the concensus is to keep both and just add docs around precedence.

> That's a good point, but would mean we'd need to handle what happens if the same feature is listed multiple times; could be --features-foo=enabled --features-foo=disabled for example

~~The cli will prevent duplicate options.~~ Edit: that is not the case for our propertymapper backed options - this could be captured as another issue. The current behavior is the last one will win.

Another current behavior worth mentioning is that a feature cannot appear in both features and features-disabled - which has usability issue from an overriding perspective because you may need to override both lists.

Otherwise the presedence order for sources in general is:
- cli - so if you use --features, --features-disabled, or --feature-name in your start command, those will be forcing
- env
- keycloak.conf

We'd check first for the value of kc.feature-name, then fallback to the list handling. This makes features and features-disabled more akin to "default" feature handling as once you use feature-name, then features and features-disabled aren't applicible to that feature.

vmuzikar:
> > And potentially, as suggested by @stianst, enable the feature like --feature-transient-users that could enable the feature based on the [rules](https://www.keycloak.org/server/features#_enabling_features) on the versions.
> 
> That would need to be done separately. There are several places in our code where we make the assumption that our options must have an argument.

Exactly. I would not rush this. It is a new concept that we'd allow value-less config options.

> What do you think of expressing this as: `--feature-<name>=vX|enabled|disabled`.

+1 It'd be on par with current `--features` where we allow enabling a specific version.

mabartos:
Created a follow-up task for the old testsuite:
- https://github.com/keycloak/keycloak/issues/43645

Not sure if it's worth the effort though.

common/src/main/java/org/keycloak/common/profile/SingleProfileConfigResolver.java
```diff
@@ -0,0 +1,28 @@
...
+public class SingleProfileConfigResolver implements ProfileConfigResolver {
+    private final Map<String, Boolean> features;
+
+    public SingleProfileConfigResolver(Map<String, Boolean> features) {
+        this.features = features;
+    }
+
+    @Override
+    public Profile.ProfileName getProfileName() {
+        return null;
```
shawkins, line 20: We may want to support feature-preview=enabled

mabartos: Yes, we can support it. Done.

mabartos:
@shawkins @vmuzikar Should be ready for the review.

vmuzikar (commented):
Generally looks good, but I didn't review in depth. @shawkins could you please review? :)

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/FeaturePropertyMappers.java
```diff
@@ -27,10 +31,64 @@ public List<PropertyMapper<?>> getPropertyMappers() {
...
         );
     }
 
+    @Override
+    public void validateConfig(Picocli picocli) {
+        validateSingleFeatureDuplicates(picocli);
+    }
+
+    // Verify a feature specified via '--feature-<name>' is set only once on CLI
+    private static void validateSingleFeatureDuplicates(Picocli picocli) {
```
shawkins: This should be removed. If we want to validate this, then it should be done generally under https://github.com/keycloak/keycloak/issues/43604

It's straight-forward to do this in the relevant config sources if we want to go that route.

shawkins (commented):
Sorry for the delay, I thought I had already submitted a more recent review. It looks good, but I wouldn't introduce a specific duplicate property check.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/FeaturePropertyMappers.java
mabartos: Ok, works to me to do it generally.

mabartos: @shawkins Updated

mabartos:
@shawkins Fixed help tests. Should be ready for review.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/QuarkusProfileConfigResolver.java
```diff
@@ -1,17 +1,73 @@
...
 
-public class QuarkusProfileConfigResolver extends CommaSeparatedListProfileConfigResolver {
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.keycloak.quarkus.runtime.configuration.MicroProfileConfigProvider.NS_KEYCLOAK_PREFIX;
+
+public class QuarkusProfileConfigResolver implements ProfileConfigResolver {
+    private final CommaSeparatedListProfileConfigResolver commaSeparatedResolver;
+    private final SingleProfileConfigResolver singleResolver;
```
shawkins: It might be better to do this is to update the Environment - to add a Quarkus specific SingleProfileConfigResolver ahead of the QuarkusProfileConfigResolver. Especially if the longer-term plan is to refactor the PropertiesProfileConfigResolver so that it is SingleProfileConfigResolver.

mabartos: @shawkins Good point. Updated:  https://github.com/keycloak/keycloak/pull/43542/commits/2eb58560ce56aebe507324cf14f52f71e57277c1

shawkins (dismissed):
Thank you for the updates @mabartos, this LGTM.

mabartos:
@shawkins Thanks! 

@vmuzikar Do you want to do the final review? Thanks

docs/guides/server/features.adoc
```diff
@@ -5,13 +5,33 @@
...
+<@kc.build parameters="--feature-<name>=enabled|disabled|vX"/>
+
+Possible values are `enabled`, `disabled`, or a specific version of the feature that should be enabled.
+For example, to enable `rolling-updates:v2` and `token-exchange`, enter this command:
+
+<@kc.build parameters="--feature-rolling-updates=v2 --feature-token-exchange=enabled"/>
+
+To enable all preview features, enter this command:
+
+<@kc.build parameters="--feature-preview=enabled"/>
```
vmuzikar: Why would we add another way to enable the preview profile? It feels confusing – redundant way to enable a profile and the naming `feature-preview` (feature, not features). We already have `--features=preview`. If a user wants to enable the preview profile with additional features, they can use the combination of `--features=preview` and `--feature-<name>=enabled`.

Sorry if I missed it before in the previous review.

vmuzikar (changes requested):
@mabartos Thank you for the update. Added one comment I'd like to get addressed/discussed before merging

docs/guides/server/features.adoc
shawkins: This is no different than the case for supporting the individual properties in general. If you already have --features set in a provided image, then want to try out preview features, it's much simplier to do that via the feature-preview option.

vmuzikar: IMHO we shouldn't have mixed profiles and features in a single option in the first place – i.e. we shouldn't have done `--features=preview`. It's ambiguous and confusing. That's the main reason I do not want to expand on this by supporting profiles in `--feature-<name>`.

IIRC @stianst had some ideas for introducing "server profiles" in general that would affect not only profile features but also config defaults etc. I'd like to wait for this before we introduce yet another way to toggle feature profiles.

stianst: My idea wasn't `server profiles`, but rather `security profiles` so not relevant to this

One related thing here though is that it would be good to have categories of features, and ability to enable preview for a category.

`dpop` should be `oauth-dpop` and it should be possible to do something like `--features=oauth:preview` to enable all preview oauth features.

vmuzikar: @stianst Thanks for the input.

@mabartos @shawkins We need a wider plan around Features to reflect all the requirements and areas for improvement. Hence I'd vote for removing the support for enabling profiles via `--feature-<profile>`. We will address it as a follow-up based on further planning.

mabartos: > @mabartos @shawkins We need a wider plan around Features to reflect all the requirements and areas for improvement. Hence I'd vote for removing the support for enabling profiles via --feature-<profile>. We will address it as a follow-up based on further planning.

@vmuzikar Works for me; agree.

mabartos: Created follow-up task: 
- https://github.com/keycloak/keycloak/issues/44003

mabartos:
@vmuzikar @shawkins I've updated the code to remove the capability to specify the feature profile via the option `--feature-preview`.

vmuzikar (approved):
Overall LGTM, thanks @mabartos.

But let's wait for @shawkins' review before merging.

shawkins (approved):
LGTM, thanks @mabartos

PR #43549, opened 2025-10-16
fix: noting db support level changes

### LINKED ISSUE #43191: Upgrade guide for 26.4.0 should mention new minimal PostgreSQL server version 13 requirement
### Description

When connecting to a PostgreSQL database, Keycloak 26.4.0 requires at least PostgreSQL server version 13.0.0. This wasn't yet the case with Keycloak 26.3.

I suggest that the upgrade guide be augmented to mention this new minimum PostgreSQL server version requirement.

shawkins:
closes: #43191

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -68,6 +68,24 @@ quarkus.datasource.user-store.username=my-username
 
 Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
 
+=== Supported Databases Versions
+
+The supported version of each database is now shown in link:{databaseguide_link}[{databaseguide_name}].
+
+It is not a supported configuration if the underlying database specific Hibernate dialect allows the use of a version that differs from those shown.
+
+From 26.3 to 26.4 the support changed as follows:
+
+[cols="1,1,1"]
+|===
+|Database |Minimum support |Dialect minimum
+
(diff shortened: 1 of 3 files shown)


ahus1 (dismissed):

ahus1:
@andymunro - please include this changes in the downstream docs for 26.4.x. Thanks!

docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
```diff
@@ -68,6 +68,24 @@ quarkus.datasource.user-store.username=my-username
...
+
+The supported version of each database is now shown in link:{databaseguide_link}[{databaseguide_name}].
+
+It is not a supported configuration if the underlying database specific Hibernate dialect allows the use of a version that differs from those shown.
+
+From 26.3 to 26.4 the support changed as follows:
+
+[cols="1,1,1"]
+|===
+|Database |Minimum support |Dialect minimum
```
vmuzikar, line 81: Do we really want to state dialect minimum? It's not consistent with the guide where we don't state it. It might also be confusing to the users what is "dialect minimum". But I understand the motivation – before, we didn't state the supported versions in the upstream, and users are now surprised by the dialect limitations. So probably ok to keep it.

shawkins: > Do we really want to state dialect minimum?

Unfortunately from a user perspective that is the actual change they may notice, as was the case with https://github.com/keycloak/keycloak/issues/43191 and #43093

I'll go ahead and copy the statement about the dialect allowing for non-supported versions to the db guide.

vmuzikar (approved):
LGTM, thanks.

PR #43560, opened 2025-10-17
Add validation for Workflwow, Condition and Steps fields

### LINKED ISSUE #43559: Add validation for Workflwow, Condition and Steps fields
When creating new workflow in UI. 

* 
```
{}
```
The error message appears: `Could not create the workflow: Workflow name is required.` which is expected

*
```
{
  "name": "my_great_workflow"
}
```
The error is "unknown_error" and in the log there is NPE (the NPE comes from empty steps ... we do use `ofNullable` at some places but all of them).

We should return more user friendly error.

vramik:
Closes #43559

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
@@ -34,6 +34,7 @@ import org.keycloak.representations.workflows.WorkflowConditionRepresentation;
 import org.keycloak.representations.workflows.WorkflowConstants;
 import org.keycloak.representations.workflows.WorkflowRepresentation;
 import org.keycloak.representations.workflows.WorkflowStepRepresentation;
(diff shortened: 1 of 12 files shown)


sguilhen (approved):

pedroigor (approved):

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowsManager.java
```diff
@@ -441,6 +443,7 @@ public Workflow toModel(WorkflowRepresentation rep) {
         List<WorkflowConditionRepresentation> conditions = ofNullable(rep.getConditions()).orElse(List.of());
 
         for (WorkflowConditionRepresentation condition : conditions) {
+            validateField(condition, "uses", condition.getUses());
```
pedroigor, line 446: Better if we use the constant here for properties but we can do it as a follow-up.

pedroigor (approved):

PR #43566, opened 2025-10-17
Add new indices on `offline_client_session`

### LINKED ISSUE #43516: Deleting Client is slow and fails when a lot of client sessions exist
### Before reporting an issue

### Area

core

### Describe the bug

We have more than 10 million active client sessions in our database. When we're now trying to delete a client, it leads to a very expensive SQL statement that doesn't hit an index.
https://github.com/keycloak/keycloak/blob/1c05c8df630bc025d328f4b986798575d3cce7a5/model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java#L37

### Version

26.4.0

### Regression

### Expected behavior

Deletion should be fast

### Actual behavior

Deletion takes sometimes even longer than the lifetime of a token and fails into a 401.

### How to Reproduce?

1. Have a couple million (or even less) client sessions
2. Delete a client

It should be observable that performance degrades the more client session exist.

### Anything else?

Output from
```
EXPLAIN ANALYZE
    DELETE FROM offline_client_session WHERE client_id = '693a0dc3-8e39-4d7f-92af-31a49e1559e8';
```
```
Delete on offline_client_session  (cost=0.56..575418.01 rows=0 width=0) (actual time=297912.707..297912.708 rows=0 loops=1)
  ->  Index Scan using constraint_offl_cl_ses_pk3 on offline_client_session  (cost=0.56..575418.01 rows=1242 width=6) (actual time=297912.705..297912.706 rows=0 loops=1)
        Index Cond: ((client_id)::text = '693a0dc3-8e39-4d7f-92af-31a49e1559e8'::text)
Planning Time: 1.134 ms
Execution Time: 297915.965 ms
```

twobiers:
This PR adds a new indices to the `offline_client_session` table .

Closes #43516

I also thought whether its possible to use a different delete query, but I didn't came up with any. Also, I believe this index makes sense in general as for client sessions we might regular need to filter for a client.

This Index improves the following queries:
https://github.com/twobiers/keycloak/blob/1c05c8df630bc025d328f4b986798575d3cce7a5/model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java#L37
https://github.com/twobiers/keycloak/blob/1c05c8df630bc025d328f4b986798575d3cce7a5/model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java#L47

I decided to document this change explicitly, as the `offline_client_session` table is large and when Keycloak is not executing the migration due to its size, it's easier to spot from the release notes.

### DIFF
--- a/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
+++ b/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
@@ -27,6 +27,15 @@
         </preConditions>
         <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT">
             <column name="CLIENT_ID" type="VARCHAR(255)" />
+            <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet author="keycloak" id="26.5.0-index-offline-css-by-client-storage-provider">
+        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
+            <indexExists tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER" />
+        </preConditions>
+        <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER">
             <column name="CLIENT_STORAGE_PROVIDER" type="VARCHAR(36)" />
             <column name="EXTERNAL_CLIENT_ID" type="VARCHAR(255)"/>
             <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
--- a/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
+++ b/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
@@ -27,15 +27,6 @@
         </preConditions>
         <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT">
             <column name="CLIENT_ID" type="VARCHAR(255)" />
-            <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
-        </createIndex>
-    </changeSet>
-
-    <changeSet author="keycloak" id="26.5.0-index-offline-css-by-client-storage-provider">
-        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
-            <indexExists tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER" />
-        </preConditions>
-        <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER">
             <column name="CLIENT_STORAGE_PROVIDER" type="VARCHAR(36)" />
             <column name="EXTERNAL_CLIENT_ID" type="VARCHAR(255)"/>
             <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
--- a/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
+++ b/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
@@ -32,7 +32,7 @@
             <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
         </createIndex>
         <modifySql dbms="postgresql">
-            <append value="WHERE CLIENT_ID != 'external'" />
+            <append value="WHERE CLIENT_STORAGE_PROVIDER = 'local'" />
         </modifySql>
     </changeSet>
 
@@ -50,18 +50,6 @@
         <modifySql dbms="postgresql">
             <append value="WHERE CLIENT_ID = 'external'" />
         </modifySql>
-    </changeSet>
-
-    <changeSet author="keycloak" id="26.5.0-index-offline-css-by-user-session-id">
-        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
-            <not>
-                <indexExists tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_USER_SESSION" />
-            </not>
-        </preConditions>
-        <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_USER_SESSION">
-            <column name="USER_SESSION_ID" type="VARCHAR(255)"/>
-            <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
-        </createIndex>
     </changeSet>
 
 </databaseChangeLog>
--- a/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
+++ b/model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
@@ -17,6 +17,10 @@
   -->
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
+    <!-- These indexes IDX_OFFLINE_CSS_BY_CLIENT and IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER are used to retrieve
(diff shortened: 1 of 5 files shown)


ahus1:
Thank you for looking into this, and sorry for deleting the index in the first place. I was misguided by the "preloading" in the name of the index, which we stopped doing in KC 25. 

I am surprised that the database suggests to use the index here as it is very expensive. I would have hoped it would rather suggest a full table scan instead. Can you double check with some freshly analyzed indexes? 

Looking at the original index and also the query of how many sessions exist for client, I'd rather like to see the offline flag as part of the index. 

When adding the index, please check that the index does not exist yet - maybe some people want to create the index in advance. 

One more optimization: When you are deleting a lot of client sessions scattered across the  table, this is still very inefficient and can lead to locks. We have procedures in place to skip those entries when we load the user session from the database. 

So while adding the index for the count, I would also suggest to skip removing the client sessions from the table. WDYT?

twobiers:
Thanks for your feedback, I will look into it on Monday in detail, but I think all your suggestions are reasonable.

> I am surprised that the database suggests to use the index here as it is very expensive. I would have hoped it would rather suggest a full table scan instead. Can you double check with some freshly analyzed indexes?

Unfortunately, I'm unable to test this anymore with our dataset as I already created an index manually. I experienced deletion times between 2 and 7 minutes depedening on the current load. Maybe postgres is switching here between index/full table scan based on the expected number of IOPS? I'm by no means an expert in this field - just an assumption. 
I will take a look how it behaves in a smaller deployment on monday.

> One more optimization: When you are deleting a lot of client sessions scattered across the table, this is still very inefficient and can lead to locks. We have procedures in place to skip those entries when we load the user session from the database.
>
>So while adding the index for the count, I would also suggest to skip removing the client sessions from the table. WDYT?

I'm unable to follow, do you mind to elaborate on that again? A few bullet points or maybe a location in the code for me to get the context would be nice.

twobiers:
@ahus1 I took a look at the queries we currently have, the `offline_user_session` table and have created additional indicies to improve the performance of all queries here. Also, they should be consistent with the indices on the `offline_user_session` table.
The thing I'm not sure about is how much it is impacting write performance and if it's actually worth to keep all of them. I will take a closer look at how the queries are used and focus on the ones on a critical path but I would like to get some of your feedback aswell.

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
```diff
@@ -0,0 +1,51 @@
...
+        </createIndex>
+    </changeSet>
+
+    <changeSet author="keycloak" id="26.5.0-index-offline-css-by-user-session-id">
+        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
+            <indexExists tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_USER_SESSION" />
+        </preConditions>
+        <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_USER_SESSION">
+            <column name="USER_SESSION_ID" type="VARCHAR(255)"/>
+            <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
```
ahus1: I'd say we already have the unique key constraint that starts with user_session_id ... can you you name a query that benefits from this index? Should we instead change re-order that other index to have the offline flag more at the start?

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
```diff
@@ -0,0 +1,55 @@
...
+        </createIndex>
+    </changeSet>
```
ahus1: I wonder if the current primary key on the table is good enough - it already has the USER_SESSION_ID as the first column, and there is usually either an online or an offline session. We can still optimize it a bit by changing the PK to start with `USER_SESSION_ID`, `OFFLINE_FLAG`, ...

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
```diff
@@ -0,0 +1,55 @@
...
+    <changeSet author="keycloak" id="26.5.0-index-offline-css-by-client-storage-provider">
+        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
+            <indexExists tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER" />
+        </preConditions>
+        <createIndex tableName="OFFLINE_CLIENT_SESSION" indexName="IDX_OFFLINE_CSS_BY_CLIENT_STORAGE_PROVIDER">
+            <column name="CLIENT_STORAGE_PROVIDER" type="VARCHAR(36)" />
+            <column name="EXTERNAL_CLIENT_ID" type="VARCHAR(255)"/>
+            <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
+        </createIndex>
+    </changeSet>
```
ahus1, line 59: A client session would be either for an external client, or an internal client, but never both. So I'd want to combine those two index into one, which then is `CLIENT_ID`, `CLIENT_STORAGE_PROVIDER`, `EXTERNAL_CLIENT_ID`, `OFFLINE_FLAG`. 

Whenever we then select an external client, we would also pass in `CLIENT_ID` with the value `external`. 

I would hope it would reduce the indexing overhead. WDYT?

ahus1 (commented):
Thank you for the change. I analyzed it a bit more to understand the interdependencies, and I now realize that we really should delete the client sessions once we delete the client, as otherwise the performance will suffer as we would otherwise try to look up clients that no longer exist whenever we load a user session. 

See below for two suggestions, and let me know your thoughts.

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
ahus1: I now realize that this would be difficult due to the columns that are `NULL`, and I am pretty sure that different databases would handle it differently. So it might then be two indexes after all ... on PostgreSQL, I would love to use a partial index instead. Unfortunately Liquibase doesn't support that.

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
twobiers: I like it. It's actually a very reasonable, trivial change which stays maintainable and should improve write performance and storage.
Considering that data is written into the table regulary and is also one of the largest, I would say it's a good change and implemented in [`e448666` (#43566)](https://github.com/keycloak/keycloak/pull/43566/commits/e4486667d439b2ca0fb11eb196485ccd279634d4)

twobiers: Transferring your concern from https://github.com/keycloak/keycloak/pull/43566#discussion_r2449176528 here:
> now realize that this would be difficult due to the columns that are `NULL`, and I am pretty sure that different databases would handle it differently. So it might then be two indexes after all ... on PostgreSQL, I would love to use a partial index instead. Unfortunately Liquibase doesn't support that.

twobiers: I'm by no means a database expert, but It seems you're right about it. 
However, most database engines should be able to use the partial index and this is already the most critical in our case as we use it only for lookups by client id.
So for internal clients we will only use `(client_id)` and for external clients we will use the full index.

Personally, I don't really like it, I would rather revert it then.

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
twobiers: Will compare performance tomrrow in our DB. I believe you're right but want to check before.

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
twobiers: Leaving e4486667d439b2ca0fb11eb196485ccd279634d4 and ca58c2921eeef1c6b19e8ea879a47808c8137dd0 in the history for now and remove them once we have agreed

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
ahus1: @twobiers - what database are you using? It seems I can make partial indexes work for PostgreSQL at least, let's hear what the others say about that.

twobiers: @ahus1 Also using PostgreSQL here.

twobiers: You're right, the index `(USER_SESSION_ID, OFFLINE_FLAG)` is already covered by the primary key `(USER_SESSION_ID, CLIENT_ID, CLIENT_STORAGE_PROVIDER, EXTERNAL_CLIENT_ID, OFFLINE_FLAG)`. 
This new index would **not** have a significant impact on the queries that use `USER_SESSION_ID` and `OFFLINE_FLAG`.
So we can drop it if you agree here.

ahus1 (approved):

PR #43574, opened 2025-10-17
fix: simplify debug handling and remove the 0.0.0.0 default

### LINKED ISSUE #43160: Regression in DEBUG_PORT handling since 26.4.0 – host binding (*:port / 0.0.0.0:port) no longer works
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

Since Keycloak 26.4.0 (after #39924), running Keycloak in Docker with remote debugging enabled fails when using host binding in `DEBUG_PORT` (e.g., `*:8787` or `*:5005`).

In 26.3.5, the same setup worked as expected.

### Version

26.4.0

### Regression

### Expected behavior

Keycloak should start successfully with `DEBUG_PORT="*:5005"` and the debugger should be listening on the specified port.

### Actual behavior

In 26.4.0, Keycloak fails to start with the following error when using host binding in `DEBUG_PORT`:
```
ERROR: transport error 202: getaddrinfo: failed to parse address
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [src/jdk.jdwp.agent/share/native/libjdwp/debugInit.c:700]
```

✅ Works with `DEBUG_PORT=5005` (just the port number, no host binding).
✅ Works with only `DEBUG: "true"` (no DEBUG_PORT), which binds to the default port 8787.
❌ Fails with DEBUG_PORT="*:5005".

### How to Reproduce?

1. Create a docker-compose.yml using Keycloak 26.4.0 with debugging enabled:

```yml
services:
  keycloak-26-4-0:
    image: quay.io/keycloak/keycloak:26.4.0
    container_name: keycloak-26-4-0
    ports:
      - "8080:8080"
      - "5005:5005"
    environment:
      KC_BOOTSTRAP_ADMIN_USERNAME: "admin"
      KC_BOOTSTRAP_ADMIN_PASSWORD: "admin"
      KC_HTTP_PORT: "8080"
      DEBUG: "true"
      DEBUG_PORT: "*:5005"
    command: ["start-dev"]
```

2. Start the container:

`docker compose up`

3. Observe the startup failure with JDWP transport errors.

Repeat with Keycloak 26.3.5 using the same setup — the container starts successfully and the debugger listens on port 5005.

### Anything else?

_No response_

shawkins:
closes: #43160

Changes:
- make sure the cli has precedence over the env
- remove the 0.0.0.0 default
- simplify the logic - supporting just an ip address is possible with just a late check.

@Pepo48 can you please validate the windows logic as well?

### DIFF
--- a/quarkus/dist/src/main/content/bin/kc.bat
+++ b/quarkus/dist/src/main/content/bin/kc.bat
@@ -22,10 +22,33 @@ if "%OS%" == "Windows_NT" (
 set SERVER_OPTS=-Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dquarkus-log-max-startup-records=10000 -Dpicocli.disable.closures=true
 
 set DEBUG_MODE=false
-set DEBUG_PORT_VAR=8787
-set DEBUG_ADDRESS=0.0.0.0:%DEBUG_PORT_VAR%
+set DEBUG_ADDRESS=8787
 set DEBUG_SUSPEND_VAR=n
 set CONFIG_ARGS=
+
+if NOT "x%KC_DEBUG%" == "x" (
+    set DEBUG_MODE=%KC_DEBUG%
+) else (
+    if NOT "x%DEBUG%" == "x" (
+        set DEBUG_MODE=%DEBUG%
+    )
+)
+
+if NOT "x%KC_DEBUG_PORT%" == "x" (
+    set DEBUG_ADDRESS=!KC_DEBUG_PORT!
+) else (
+    if NOT "x%DEBUG_PORT%" == "x" (
+        set DEBUG_ADDRESS=!DEBUG_PORT!
+    )
+)
+
+if NOT "x%KC_DEBUG_SUSPEND%" == "x" (
+    set DEBUG_SUSPEND_VAR=%KC_DEBUG_SUSPEND%
+) else (
+    if NOT "x%DEBUG_SUSPEND%" == "x" (
+        set DEBUG_SUSPEND_VAR=%DEBUG_SUSPEND%
+    )
+)
 
 rem Read command-line args, the ~ removes the quotes from the parameter
 :READ-ARGS
@@ -35,20 +58,9 @@ if "%KEY%" == "" (
 )
 if "%KEY%" == "--debug" (
     set DEBUG_MODE=true
-    if 1%2 EQU +1%2 (
-        rem Plain port
-        set DEBUG_ADDRESS=0.0.0.0:%2
+    (echo %2 | findstr /R "^[0-9][0-9]*$" >nul || echo %2 | findstr /R "^[0-9].*\." >nul || echo %2 | findstr /R "^\[.*:.*\]" >nul) && (
+        set DEBUG_ADDRESS=%2
         shift
-    ) else (
-        rem IPv4 or IPv6 address with optional port
-        (echo %2 | findstr /R "[0-9].*\." >nul || echo %2 | findstr /R "\[.*:.*\]" >nul) && (
-            (echo %2 | findstr /R "]:[0-9][0-9]*" >nul || echo %2 | findstr /R "^[0-9].*:[0-9][0-9]*" >nul) && (
-                set DEBUG_ADDRESS=%2
-            ) || (
-                set DEBUG_ADDRESS=%2:%DEBUG_PORT_VAR%
-            )
-            shift
-        )
     )
     shift
     goto READ-ARGS
@@ -140,36 +152,15 @@ if not "x%JAVA_OPTS_APPEND%" == "x" (
     set JAVA_OPTS=%JAVA_OPTS% %JAVA_OPTS_APPEND%
 )
 
-if NOT "x%KC_DEBUG%" == "x" (
-    set DEBUG_MODE=%KC_DEBUG%
-) else (
-    if NOT "x%DEBUG%" == "x" (
-        set DEBUG_MODE=%DEBUG%
-    )
-)
-
-if NOT "x%KC_DEBUG_PORT%" == "x" (
-    set DEBUG_PORT_VAR=%KC_DEBUG_PORT%
-    set DEBUG_ADDRESS=0.0.0.0:!DEBUG_PORT_VAR!
-) else (
-    if NOT "x%DEBUG_PORT%" == "x" (
-        set DEBUG_PORT_VAR=%DEBUG_PORT%
-        set DEBUG_ADDRESS=0.0.0.0:!DEBUG_PORT_VAR!
-    )
-)
-
-if NOT "x%KC_DEBUG_SUSPEND%" == "x" (
-    set DEBUG_SUSPEND_VAR=%KC_DEBUG_SUSPEND%
-) else (
-    if NOT "x%DEBUG_SUSPEND%" == "x" (
-        set DEBUG_SUSPEND_VAR=%DEBUG_SUSPEND%
-    )
-)
-
 rem Set debug settings if not already set
 if "%DEBUG_MODE%" == "true" (
     echo "%JAVA_OPTS%" | findstr /I "\-agentlib:jdwp" > nul
     if errorlevel == 1 (
+        rem Handle no port
+        (echo %2 | findstr /R "^[0-9][0-9]*" >nul || echo %2 | findstr /R ":[0-9][0-9]*$" >nul) && (
+        ) || (
+            set DEBUG_ADDRESS=%DEBUG_ADDRESS%:8787
+        )
         set JAVA_OPTS=%JAVA_OPTS% -agentlib:jdwp=transport=dt_socket,address=%DEBUG_ADDRESS%,server=y,suspend=%DEBUG_SUSPEND_VAR%
     ) else (
         echo Debug already enabled in JAVA_OPTS, ignoring --debug argument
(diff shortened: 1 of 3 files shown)


shawkins:
@Pepo48 I'm not quite getting the windows changes right. Initially I'm just checking for debug with the next token starting with a digit or [, then later checking if it looks like it needs a port.

quarkus/dist/src/main/content/bin/kc.sh
```diff
@@ -41,10 +41,9 @@ SERVER_OPTS="$SERVER_OPTS -Dpicocli.disable.closures=true"
 SERVER_OPTS="$SERVER_OPTS -Dquarkus-log-max-startup-records=10000"
 CLASSPATH_OPTS="'$(abs_path "../lib/quarkus-run.jar")'"
 
-DEBUG_MODE="${KC_DEBUG:-${DEBUG:-false}}"
```
vmuzikar, line 44: Just wanted to highlight this might be considered a breaking change – we no longer accept `KC_DEBUG` env var. It was not documented, though  – so it should be fine.

quarkus/dist/src/main/content/bin/kc.sh
```diff
@@ -41,10 +41,9 @@ SERVER_OPTS="$SERVER_OPTS -Dpicocli.disable.closures=true"
...
-DEBUG_PORT="${KC_DEBUG_PORT:-${DEBUG_PORT:-8787}}"
+DEBUG_MODE="${KC_DEBUG_MODE:-${DEBUG:-false}}"
+DEBUG_ADDRESS="${KC_DEBUG_PORT:-${DEBUG_PORT:-8787}}"
```
vmuzikar, line 45: I wonder if we should really rely on the default behaviour here with the address. Maybe we could add the explicit `127.0.0.1:8787` as the default instead of just the port?

quarkus/dist/src/main/content/bin/kc.sh
```diff
@@ -148,6 +136,10 @@ fi
 if [ "$DEBUG_MODE" = "true" ]; then
     DEBUG_OPT="$(echo "$JAVA_OPTS" | $GREP "\-agentlib:jdwp")"
     if [ -z "$DEBUG_OPT" ]; then
+        # Handle no port
+        if ! echo "$DEBUG_ADDRESS" | grep -Eq '(^[0-9]+)|(:[0-9]+)$'; then
```
vmuzikar: Should this be only
```suggestion
        if ! echo "$DEBUG_ADDRESS" | grep -Eq ':[0-9]+$'; then
```

quarkus/dist/src/main/content/bin/kc.sh
shawkins: That was not intentional.

quarkus/dist/src/main/content/bin/kc.sh
shawkins: No, we're looking for non-numbers that lack a port.

quarkus/dist/src/main/content/bin/kc.sh
shawkins: I think we can trust the JRE behavior.

vmuzikar: Do you see any downsides if we state the address explicitly? Feels safer to me.

quarkus/dist/src/main/content/bin/kc.sh
vmuzikar: Right, I missed `$` is outside of the bracket and we're matching both `1234` and ending with `:1234`. Maybe should we then put the whole expression inside brackets so it's more clear?
```suggestion
        if ! echo "$DEBUG_ADDRESS" | grep -Eq '((^[0-9]+)|(:[0-9]+))$'; then
```

quarkus/dist/src/main/content/bin/kc.sh
shawkins: From what we have documented there wouldn't be a downside to specify localhost:8787. From a jdwp perspective there could eventually be other ways to affect the default address, which this would prevent.

I'm now thinking that adding the handling for specifying the address only, without the port, was a mistake as it will futher deviates from the jdwp handling and adds a little more complexity to our scripts. If there are no objections, I'll remove that with this pr as well.

shawkins: I'm wondering if we should even rely so much on this kind of debug logic. JAVA_OPTS, JAVA_OPTS_APPEND are perfectly capable of accomodating the options. 

https://docs.oracle.com/en/java/javase/21/docs/specs/jpda/conninv.html - it looks like users may want to control virtual thread listing, the allow list if using a publically accessible address, etc. Maybe we should just update the doc to provide an example of using an environment variable.

There is probably even a case to be made for handling -X arguments in the same way we are handling -D.

@keycloak/cloud-native WDYT?

quarkus/dist/src/main/content/bin/kc.bat
```diff
@@ -22,11 +22,34 @@ if "%OS%" == "Windows_NT" (
...
+    )
+)
+
+if NOT "x%KC_DEBUG_PORT%" == "x" (
+    set DEBUG_ADDRESS=!KC_DEBUG_PORT!
+) else (
+    if NOT "x%DEBUG_PORT%" == "x" (
+        set DEBUG_ADDRESS=!DEBUG_PORT!
+    )
+)
```
Pepo48, line 43:
```suggestion
if NOT "x%KC_DEBUG_PORT%" == "x" (
    set DEBUG_ADDRESS=%KC_DEBUG_PORT%
) else (
    if NOT "x%DEBUG_PORT%" == "x" (
        set DEBUG_ADDRESS=%DEBUG_PORT%
    )
)
```

No need for the delayed expansion here, we're not in the delayed expansion block anyway.

quarkus/dist/src/main/content/bin/kc.bat
```diff
@@ -35,20 +58,9 @@ if "%KEY%" == "" (
...
-            (echo %2 | findstr /R "]:[0-9][0-9]*" >nul || echo %2 | findstr /R "^[0-9].*:[0-9][0-9]*" >nul) && (
-                set DEBUG_ADDRESS=%2
-            ) || (
-                set DEBUG_ADDRESS=%2:%DEBUG_PORT_VAR%
-            )
-            shift
-        )
     )
     shift
     goto READ-ARGS
```
Pepo48, line 68:
```suggestion
if "%KEY%" == "--debug" (
    set DEBUG_MODE=true
    if NOT "x%~2" == "x" (
        echo %~2 | findstr /R "^[0-9[]" >nul && (
            set DEBUG_ADDRESS=%~2
            shift
        )
    )
    shift
    goto READ-ARGS
```

Simplified parsing to just check if next arg starts with digit or `[`

quarkus/dist/src/main/content/bin/kc.bat
```diff
@@ -140,32 +152,6 @@ if not "x%JAVA_OPTS_APPEND%" == "x" (
...
-if NOT "x%KC_DEBUG_SUSPEND%" == "x" (
-    set DEBUG_SUSPEND_VAR=%KC_DEBUG_SUSPEND%
-) else (
-    if NOT "x%DEBUG_SUSPEND%" == "x" (
-        set DEBUG_SUSPEND_VAR=%DEBUG_SUSPEND%
-    )
-)
-
 rem Set debug settings if not already set
 if "%DEBUG_MODE%" == "true" (
```
Pepo48, line 153:
```suggestion
if "%DEBUG_MODE%" == "true" (
    rem If DEBUG_ADDRESS is just a port number, prepend localhost:
    echo %DEBUG_ADDRESS% | findstr /R "^[0-9][0-9]*$" >nul && (
        set DEBUG_ADDRESS=localhost:%DEBUG_ADDRESS%
    )
```

If DEBUG_ADDRESS is just a port number, prepend localhost.

Pepo48 (changes requested):
@shawkins I tried to simplify quite complex IP/port detection logic (hopefully I am not simplifying it too much. I added a short explanation underneath each suggestion. Let me know if everything does make sense and if it really does what you wanted to achieve.

I already did some tests, but very brief ones, I can test it thoroughly tomorrow, based on feedback from you.

Thanks for the PR, I think the provided changes are step into the right direction!

shawkins:
Thank you for the help @Pepo48, this looks ready to go now.

Pepo48 (approved):
LGTM. On Windows I specifically tested the default behaviour (nothing defined), then with custom port, full address and with IPv6 format as well. Everything worked as expected, including the precedence. I could connect a debugger successfully in each case. 

Thanks, @shawkins!

quarkus/dist/src/main/content/bin/kc.sh
shawkins: @vmuzikar are you ok with leaving localhost off, or do you want it on? I can add that, but it will need similar handling as to what was removed for supporting ip only - that is check for a number only after parsing so that it's applied to both the env and the cli cases.

mabartos:
@vmuzikar Do you want to take a look at it?

vmuzikar (approved):
LGTM, thanks.

PR #43580, opened 2025-10-17
fix: allow for --optimized to receive signals

### LINKED ISSUE #43561: Server does not shutdown gracefully when started with --optimized
### Before reporting an issue

### Area

core

### Describe the bug

We were experiencing issues with client timeouts when shutting down single replicas in a Keycloak Docker container cluster.

When drilling down the issue we found out that keycloak does not seem to shutdown on `SIGINT` or `SIGTERM` when running `start` with `--optimized`. When running the same image without the optimized flag the server performs a graceful shutdown and does not abruptly leave the cluster.

### Version

26.4.1

### Regression

### Expected behavior

The keycloak service should perform a graceful shutdown when running with `start --optimized`.

### Actual behavior

The service does not gracefully shutdown leading to client timeouts and warnings about potential dataloss in the remaining members of the keycloak cluster.

### How to Reproduce?

docker-compose.yml
```
version: '3.8'

services:
  keycloak:
    build: .
    container_name: keycloak-local
    ports:
      - '8080:8080'
      - '9000:9000'
    depends_on:
      - postgres
    stop_signal: SIGTERM
    stop_grace_period: 30s
    command: start --optimized
    environment:
      KC_DB: postgres
      KC_DB_USERNAME: postgres
      KC_DB_PASSWORD: postgres
      KC_DB_URL: jdbc:postgresql://postgres-local:5432/keycloak
      KEYCLOAK_ADMIN: admin
      KEYCLOAK_ADMIN_PASSWORD: admin
      KC_HTTP_PORT: 8080
      KC_HTTP_ENABLED: 'true'
      KC_HOSTNAME_STRICT: 'false'

  postgres:
    image: postgres:15
    container_name: postgres-local
    environment:
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: postgres
      POSTGRES_DB: keycloak
    ports:
      - '5435:5432'
    volumes:
      - postgres_data:/var/lib/postgresql/data

volumes:
  postgres_data:

```

Dockerfile
```
WORKDIR /opt/keycloak

# Configure a database vendor
ENV KC_DB=postgres

FROM quay.io/keycloak/keycloak:latest

COPY --from=builder /opt/keycloak/ /opt/keycloak/

# Configure a database vendor
ENV KC_DB=postgres

ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
```

1. Run `docker compose up -d`
2. Wait for keycloak to start
3. Run `docker compose down`
4. Check container logs for graceful shutdown 

---

As a reference when omitting the `--optimized` in the command section of the docker compose and performing the same reproduction steps, the service will gracefully shutdown with the following logs:

```
2025-10-17 09:19:24,751 INFO  [org.infinispan.CLUSTER] (main) ISPN000080: Disconnecting JGroups channel `ISPN`
2025-10-17 09:19:24,762 INFO  [org.keycloak.jgroups.certificates.CertificateReloadManager] (main) Stopping JGroups certificate reload manager
2025-10-17 09:19:24,767 INFO  [com.arjuna.ats.jbossatx] (main) ARJUNA032014: Stopping transaction recovery manager
2025-10-17 09:19:24,782 INFO  [io.quarkus] (main) Keycloak stopped in 0.074s
```

### Anything else?

_No response_

shawkins:
closes: #43561

We have two options:
1. The approach in this fix - keep the use of the background process
2. Go back to only performing the build check with a background process

The reason I want to stick with 1 is that it will keep us moving toward a state we we can more efficiently refine the --optimized concept. This is demonstrated in https://github.com/keycloak/keycloak/pull/43591 - which use the same technique as --optimized the run the command in the background process if it doesn't need reaugmentation.

The approach is to trap signals we are interested in forwarding to the underlying process. Any other signals, such as triggering threaddumps are expected to be sent directly to the java process. Forwarding stdin makes the bootstrap-admin commands work as expected, and the subprocess does recognize that a terminal is attached so the auto-coloring works the same way as well.

### DIFF
--- a/quarkus/dist/src/main/content/bin/kc.sh
+++ b/quarkus/dist/src/main/content/bin/kc.sh
@@ -171,8 +171,14 @@ if [ "$PRINT_ENV" = "true" ]; then
   echo "Using JAVA_RUN_OPTS: $JAVA_RUN_OPTS"
 fi
 
-eval "'$JAVA'" "$JAVA_RUN_OPTS"
-status=$?
+# trap the signals that should be forwarded to the java process
+trap 'status=143; while [ -z "$PID" ]; do sleep 1; done; kill -TERM $PID; wait $pid' TERM INT
+# run the java process in the background using the current stdin
+eval "{ '$JAVA' $JAVA_RUN_OPTS <&3 3<&- & } 3<&0"
+# obtain the pid, and await the exit status
+PID=$!; wait $PID; status=$?
+# remove the trap as signals can be directly handled
+trap - TERM INT
 # only exit code 10 means that implicit reaugmentation occurred and a relaunch is needed 
 if [ $status = 10 ]; then
   JAVA_RUN_OPTS="-Dkc.config.built=true $JAVA_RUN_OPTS"
--- a/quarkus/dist/src/main/content/bin/kc.sh
+++ b/quarkus/dist/src/main/content/bin/kc.sh
@@ -172,7 +172,7 @@ if [ "$PRINT_ENV" = "true" ]; then
 fi
 
 # trap the signals that should be forwarded to the java process
-trap 'status=143; while [ -z "$PID" ]; do sleep 1; done; kill -TERM $PID; wait $pid' TERM INT
+trap 'status=143; while [ -z "$PID" ]; do sleep 0.1; done; kill -TERM $PID; wait $PID' TERM INT
 # run the java process in the background using the current stdin
 eval "{ '$JAVA' $JAVA_RUN_OPTS <&3 3<&- & } 3<&0"
 # obtain the pid, and await the exit status
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/StartCommandDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/StartCommandDistTest.java
@@ -92,6 +92,23 @@ public class StartCommandDistTest {
         cliResult.assertNoError("The following build time options");
     }
 
+    @Test
+    @RawDistOnly(reason = "Containers are immutable")
+    void terminateStartOptimized(KeycloakDistribution dist) {
+        CLIResult cliResult = dist.run("build", "--db=dev-file");
+        cliResult.assertBuild();
+
+        dist.setManualStop(true);
+        cliResult = dist.run("start", "--optimized", "--http-enabled=true", "--hostname-strict=false");
+        cliResult.assertStarted();
+
+        // if the child java process does not clean up, then subsequent start will fail
+        dist.stop();
+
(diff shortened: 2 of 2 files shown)


quarkus/dist/src/main/content/bin/kc.sh
```diff
@@ -171,8 +171,14 @@ if [ "$PRINT_ENV" = "true" ]; then
   echo "Using JAVA_RUN_OPTS: $JAVA_RUN_OPTS"
 fi
 
-eval "'$JAVA'" "$JAVA_RUN_OPTS"
-status=$?
+# trap the signals that should be forwarded to the java process
+trap 'status=143; while [ -z "$PID" ]; do sleep 1; done; kill -TERM $PID; wait $pid' TERM INT
```
mabartos:
```suggestion
trap 'status=143; while [ -z "$PID" ]; do sleep 0.1; done; kill -TERM $PID; wait $PID' TERM INT
```

1. I believe the $PID should be used.
2. Maybe decrease the sleep time?

mabartos (commented):
@shawkins Thanks for the PR!

Overall, it LGTM and I can confirm it resolves the issue after reproducing that on my local machine. 

I'm not a bash expert, but it looks good to me. Added a small comment.

quarkus/dist/src/main/content/bin/kc.sh
shawkins: Looks good to me. The while loop is just for a corner case. It turns out we just need a second wait to restore the prompt on exit, it didn't matter if was using $PID or not so I didn't even notice I had used lower case. Using $PID seems fine as well.

ASzc (approved):
The simplest, most typical way to pass signals and stdio to a child process would be the `exec` command. However the way we have written this script, it must be possible to return from the first call to the JVM, so we can do a second call in the case of reaugmentation. That eliminates `exec`, since it will never return.

The alternative implementation with trap is the 2nd most conventional method, and it can return from the child process. However this is much more code-heavy. So, I am concerned about maintaining our declared POSIX syntax compatibility for this user-facing script, so as not to break less common OSs.

As far as I have been able to determine this doesn't introduce syntax that is a problem in a POSIX context.

If we could reliably anticipate if reaugmentation will happen or not, we could write an `if` block that ends with `exec` in both cases. The less fragile solution before this PR was to let Quarkus decide on reaugmentation itself, and react to that. However, that might have flipped now, but I don't know how hard it is to predetermine reaugmentation?

vmuzikar:
> However, that might have flipped now, but I don't know how hard it is to predetermine reaugmentation?

We can't without running a JVM – which is something we wanted to avoid.

vmuzikar (approved):
LGTM (also  considering other's reviews as I'm not  a shell expert either).

Issue #35040, opened 2024-11-18
Migrate admin package to new test framework


lhanusov:
### Description

Test framework MVP

follows up: #31238

This epic will deliver the whole admin package migration to the new test framework and adding new features that depend on it.

### Discussion

https://github.com/keycloak/keycloak/discussions/34951

### Issues

- #31141 
- #34189 
- #34190
- #34487
- #34488
- #34494
- #34619
- #34859  
- #34955 
- #35029
- #35032
- #35035
- #35082
- #35085
- #35091  
- #35094
- #35173 
- #35315
- #35385
- #35388
- #35389  
- #35501
- #35541
- #35592
- #35995 
- #36608
- #36685 
- #36775
- #36839
- #36974
- #37283 
- #37486   

### Motivation

Providing a new test framework should make it a lot simpler to both write new tests as well as running tests. Additionally, providing a common framework to extension developers will make it simpler to test custom extensions.

PR #43581, opened 2025-10-17
Migrate AdminClientTest

vaceksimon:
Part of #35040 

- `AdminClientTest#noClientSecretWithClientCredentialsAuthSuccess()` is `@Disabled` for now, as it requires mTLS support. This functionality may be added as part of the FIPS integration for the framework (#35389)
- `ManagedUserCleanup` added, but it's not used as I decided to choose another approach for a problem it was supposed to solve. I didn't see a reason to delete it once I added it.


tests/base/src/test/java/org/keycloak/tests/admin/AdminClientTest.java
```diff
@@ -0,0 +1,352 @@
...
+        // Check not possible to invoke anymore
+        try {
+            realm = adminClient.realm(testRealm.getName()).toRepresentation();
+            Assertions.fail("Not expected to successfully get realm");
+        } catch (NotAuthorizedException nae) {
+            // Expected
+        } finally {
+            setClientEnabled(CLIENT_ID, true);
+        }
+    }
```
lhanusov, line 125: @vaceksimon please use Assertions.assertThrows instead

tests/base/src/test/java/org/keycloak/tests/admin/AdminClientTest.java
```diff
@@ -0,0 +1,352 @@
...
+        // Check not possible to invoke anymore
+        try {
+            realm = adminClient.realm(testRealm.getName()).toRepresentation();
+            Assertions.fail("Not expected to successfully get realm");
+        } catch (NotAuthorizedException nae) {
+            // Expected
+        } finally {
+            setClientEnabled(Constants.ADMIN_CLI_CLIENT_ID, true);
+            adminClient.close();
+        }
```
lhanusov, line 173: @vaceksimon please use Assertions.assertThrows instead

tests/base/src/test/java/org/keycloak/tests/admin/AdminClientTest.java
```diff
@@ -0,0 +1,352 @@
...
+        try {
+            realm = adminClientOffline.realm(testRealm.getName()).toRepresentation();
+            Assertions.fail("Not expected to successfully get realm");
+        } catch (NotAuthorizedException nae) {
+            // Expected
+        } finally {
+            setUserEnabled(TEST_USER_USERNAME, true);
+            adminClient.close();
+            adminClientOffline.close();
+        }
```
lhanusov, line 218: @vaceksimon please use Assertions.assertThrows instead, in this case two times

test-framework/core/src/main/java/org/keycloak/testframework/realm/ManagedUserCleanup.java
```diff
@@ -0,0 +1,50 @@
+package org.keycloak.testframework.realm;
+
+
+import org.keycloak.admin.client.resource.UserResource;
+import org.keycloak.representations.idm.UserRepresentation;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class ManagedUserCleanup {
```
lhanusov, line 10: @vaceksimon can you please, show example when can it be used? I understand your point, but it's better to have less functionality at the beginning then later a lot of unused code. I am leaning more towards removing it.

lhanusov (changes requested):

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/ClientScopeBuilder.java
lhanusov, line 1: Rather then moving this file to a shared util module, I propose to create a new ClientScopeConfigBuilder. We already have RoleConfigBuilder, GroupConfigBuilder

lhanusov:
@vaceksimon please, replace the comment message with "Closes" instead of "Part of". This is the last issue, which will finalize the admin package. Check also all remaining old admin package bits in the testsuite scripts or CI. Thank you

tests/base/src/test/java/org/keycloak/tests/admin/AdminClientTest.java
vaceksimon: I believe if I did that, and the assertion failed, I wouldn't be able to call `setClientEnabled(CLIENT_ID, true);` that is in the `finally` block, and revert the operation on line 116

tests/base/src/test/java/org/keycloak/tests/admin/AdminClientTest.java
vaceksimon: See my previous comment

tests/base/src/test/java/org/keycloak/tests/admin/AdminClientTest.java
vaceksimon: See my previous comment

test-framework/core/src/main/java/org/keycloak/testframework/realm/ManagedUserCleanup.java
vaceksimon: For example
```java
managedUser.updateWithCleanup(user -> user.emailVerified(true));
```
As I mentioned in the PR description, I implemented it for a use case in this test, but ultimately decided not to use it. Since both a realm and a client have this functionality, and the user is the only one missing from this "big three," I left it in.

I think sooner or later someone will create it anyway, but yeah now it is unused.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/ClientScopeBuilder.java
vaceksimon: I don't see a significant advantage in doing that. It's a small class used by both test suites, like `CredentialBuilder` or `FederatedIdentityBuilder`.

A new class would probably be put in the `org.keycloak.testframework.realm` package, which is already pretty big and I'd say in need of a restructure.

test-framework/core/src/main/java/org/keycloak/testframework/realm/ManagedUserCleanup.java
lhanusov: Created #44697

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/ClientScopeBuilder.java
lhanusov: Created: #44698

lhanusov:
will be delivered in #44705

PR #43584, opened 2025-10-17
Allow managing realm admin roles if the the realm-admin role is granted

### LINKED ISSUE #43578: "admin" client role now requires server admin user
### Before reporting an issue

### Area

admin/api

### Describe the bug

Hi, 

after upgrading from 26.2 to 26.4.1, the changes made in https://github.com/keycloak/keycloak/issues/39956, also affect client roles named `admin` in non-admin realms.

This means the local realm-management account (service-account of the backend client) can no longer give the client role `admin` (no other roles or mappings associated to it) to other users.

It seems unlikely to me that this is intended behaviour as this role is only an admin role within the app the client is for, and not for the keycloak realm / master realm.

### Version

26.4.1

### Regression

### Expected behavior

Only restrict actual admin (realm) roles.

### Actual behavior

`403 Forbidden`

### How to Reproduce?

* Create a new realm
* Create a new client with service-account and realm-management roles
* Create a new client role `admin`
* Create a dummy user
* Try to have the service-account give that user the `admin` client role

### Anything else?

_No response_

### LINKED ISSUE #43579: 403 Forbidden when assigning realm-management client roles with realm-admin despite FGAP disabled (regression in 26.4.0+)
### Before reporting an issue

### Area

admin/fine-grained-permissions

### Description
In Keycloak 26.4.0 & 26.4.1, the Fine-Grained Admin Permissions (FGAP) code path executes even when FGAP is disabled at both realm and client levels. This causes service accounts with the realm-admin role to receive 403 Forbidden errors when attempting to assign realm-management client roles to groups via the Admin REST API.

This is a regression from 26.3.5, which worked correctly.

###  Environment

  - Keycloak Version: 26.4.0 and 26.4.1
  - Deployment: Kubernetes
  - Working Version: 26.3.5

### Version

26.4.1

### Regression

### Expected behavior

A service account with the realm-admin composite role should be able to assign realm-management client roles (e.g., view-users, manage-users, view-clients) to groups when FGAP is disabled.

### Actual behavior

The operation fails with 403 Forbidden and the error originates from RolePermissions.requireMapRole() in the FGAP code path, despite FGAP being disabled.

### How to Reproduce?

  1. Create a realm with FGAP disabled (adminPermissionsEnabled: false)
  1. Create a service account client and assign it the realm-admin role from the realm-management client
  1. Obtain an access token for the service account
  1. Attempt to assign realm-management client roles to a group via REST API:
  POST /admin/realms/{realm}/groups/{group-id}/role-mappings/clients/{realm-management-client-id}
  Authorization: Bearer {token}
  Content-Type: application/json
```
  [
    {"id": "{role-id}", "name": "view-users"},
    {"id": "{role-id}", "name": "manage-users"}
  ]
```
  1. Observe 403 Forbidden error

####  Realm FGAP Status:
```
  {
    "realm": "my-realm",
    "adminPermissionsEnabled": false,
    "userManagedAccessAllowed": false
  }
```
####  Realm-Management Client FGAP Status:
```
  {
    "error": "Feature not enabled",
    "error_description": "For more on this error consult the server log."
  }
```
####  Service Account Roles:
```
  [
    {
      "id": "db96dcd7-7322-4e63-a45d-f3cb79d2735f",
      "name": "realm-admin",
      "description": "${role_realm-admin}",
      "composite": true,
      "clientRole": true,
      "containerId": "910f78e7-980f-4072-8e80-a9bbbf9c5ca2"
    }
  ]
```
####  Stack Trace
```
  {
    "timestamp": "2025-10-17T13:19:36.878298787Z",
    "loggerName": "org.keycloak.services.error.KeycloakErrorHandler",
    "level": "DEBUG",
    "message": "Error response Forbidden",
    "exception": {
      "exceptionType": "jakarta.ws.rs.ForbiddenException",
      "message": "HTTP 403 Forbidden",
      "frames": [
        {"class": "org.keycloak.services.resources.admin.fgap.RolePermissions", "method": "requireMapRole", "line": 339},
        {"class": "org.keycloak.services.resources.admin.ClientRoleMappingsResource", "method": "addClientRoleMapping", "line": 175}
      ]
    }
  }
```
###  Additional Context

  - This affects both direct client role assignments and composite roles containing client roles
  - The same operation works successfully via the Keycloak Admin Console UI with master realm user
  - This appears to be a regression introduced between 26.3.5 and 26.4.0

###  Impact

This blocks automation tools from managing realm-management role assignments using service accounts, forcing manual intervention through the Admin Console.

pedroigor:
Closes #43579
Closes #43578

### DIFF
--- a/docs/documentation/server_admin/topics/admin-console-permissions/fine-grain-v2.adoc
+++ b/docs/documentation/server_admin/topics/admin-console-permissions/fine-grain-v2.adoc
@@ -8,7 +8,7 @@ Different from the Role-Based Access Control (RBAC) Mechanism provided through t
 <<_master_realm_access_control, Global and Realm specific roles>>, this feature provides a more fine-grained control over
 how realm resources can be accessed and managed based on a well-defined set of operations that can be performed on them.
 
-By relying on a Policy-Based Access Control, server administrators can define permissions to realm resources such as users,
+By relying on a Policy-Based Access Control, realm administrators can define permissions to realm resources such as users,
 groups, and clients, using different policy types, or access control methods, so that a realm administrator is limited to
 access a subset of realm resources and their operations.
 
@@ -17,6 +17,23 @@ not replace it. You are still able to grant administrative roles like `view-user
 to realm administrators but doing so will skip the mechanisms provided by this feature.
 
 Enforcing access to realm resources only applies when managing resources through the administration console or the Admin API.
+
+==== Understanding the different types of realm administrators
+
+There are two different types of realm administrators when managing a realm:
+
+* **Server administrators**: These are users (or service accounts) that have been granted with the `admin` role in the master realm.
+
+* **Realm administrators**: These are users (or service accounts) that have been granted with the `realm-admin` role in a specific realm.
+
+Both types of administrators can manage a realm with full access to all their resources and
+administrative operations.
+
+The main difference between them is that server administrators can manage any realm in the server, while realm administrators can only manage
+the realm they have been granted access to.
+
+Be aware that both server and realm administrators are not affected by the permissions you define when managing access to realm resources through this feature.
+Always make sure to review the users granted with the `admin` or `realm-admin` roles to avoid any potential privilege escalation.
 
 ==== Understanding the Realm Resource Types
 
@@ -210,8 +227,8 @@ properly allow paginating resources as well as avoid an additional overhead on t
 realm resource returned by queries.
 
 Partial evaluation and filtering occurs only if the feature is enabled to a realm, and if the user is not granted
-with view-related administrative roles like `view-users` or `view-clients`. For instance, it will not happen for regular server administrators granted
-with the `admin` role.
+with view-related administrative roles like `view-users` or `view-clients`. For instance, it will not happen for administrators granted
+with the `admin` role at the master realm (server administrators), or realm administrators granted with the `realm-admin` role in a realm other than the master realm.
 
 When querying resources, the partial evaluation mechanism works as follows:
 
@@ -235,7 +252,7 @@ type permissions with the *manage* scope for user `myadmin`).
 .Fine grained permissions search
 image:images/fine-grain-search.png[Fine grained permissions search]
 
-These capabilities allow server administrators to perform queries on their universe of permissions and identify which ones
+These capabilities allow realm administrators to perform queries on their universe of permissions and identify which ones
 are enforcing access to a set of one or more realm resources and their scopes. Combined with the evaluation tool on the
 *Evaluation* tab, they provide a key management tool for managing permissions in a realm. See <<_evaluating-permissions, Evaluating Permissions>>
 for more details.
@@ -310,7 +327,7 @@ themselves.
 ==== Accessing a Realm administration console as a Realm Administrator
 
 Realm administrators can access a dedicated realm-specific administration console that allows them to manage resources within their assigned realm.
-This console is separate from the main {project_name} Admin Console, which is typically used by server administrators.
+This console is separate from the main {project_name} Admin Console, which is typically used by realm administrators.
 
 For more details on dedicated realm administration consoles and available roles, refer to: <<_per_realm_admin_permissions, Dedicated admin consoles>>.
 
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -56,6 +56,18 @@ This prevents problems with client IDs or passwords that contain, for example, a
 
 To revert to the old behavior, change the client authentication to *Client secret sent as HTTP Basic authentication without URL encoding (deprecated)* (`client_secret_basic_unencoded`).
 
+=== Allowing realm administrators granted with the `realm-admin` role to assign admin roles
+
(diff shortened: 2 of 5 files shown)


vramik (changes requested):
Hi @pedroigor, thanks for this PR.

I'm looking at the changes, and I'm a bit confused about the new direction regarding admin role mappings.

My understanding from our previous discussion was that we explicitly decided only server admins should be allowed to map admin roles, and we would not extend this privilege to delegated realm admins.

Based on that agreement, I created:

PR 41145 (https://github.com/keycloak/keycloak/pull/41145) to enforce this restriction.

PR 40271 (https://github.com/keycloak/keycloak/pull/40271) to document this behavior in the release notes.

I also closed (as not planned) a couple of other issues based on that decision.

This PR seems to reverse that decision and allow delegated admins to manage these roles. Could you share the reasoning behind this change?

I just want to make sure I'm on the same page and understand the new context, especially since we'll need to update the documentation if we proceed with this new approach. Also do you plan to backport this change? Thanks!

pedroigor:
> I'm looking at the changes, and I'm a bit confused about the new direction regarding admin role mappings.

I think we missed the difference between a `realm-admin` and a delegated realm admin. An admin explicitly granted with `realm-admin` is a realm admin, similar to a server/master admin, but within the scope of a specific realm. That is why I'm updating `realm-admin` role to also allow granting admin roles.

vramik:
Ah, thanks for that clarification, @pedroigor. That distinction between `realm-admin` and "delegated admin" makes the reasoning for this PR much clearer.

My takeaway from our original discussion was more restrictive than we now want.

The current behavior in 26.3/26.4 is that [only server admins can map these roles](https://www.keycloak.org/docs/latest/upgrading/index.html#restrict-admin-role-mappings-to-server-administrators). This PR will change that behavior for 26.5 by "giving back" this permission to the `realm-admin` role.

We need to make sure we document this change in the 26.5 release notes, as it's a change from the behavior we documented before.

services/src/main/java/org/keycloak/services/resources/admin/fgap/AdminPermissions.java
```diff
@@ -35,27 +36,27 @@ public class AdminPermissions {
 
 
     public static AdminPermissionEvaluator evaluator(KeycloakSession session, RealmModel realm, AdminAuth auth) {
-        if (Profile.isFeatureEnabled(Profile.Feature.ADMIN_FINE_GRAINED_AUTHZ_V2)) {
```
vramik, line 38: Is this change (and other changes in this file) intentional? 

Edit: Please see https://github.com/keycloak/keycloak/issues/40192 for more context. Do we want/need to use `MgmtPermissions` (instead of `MgmtPermissionsV2`) in case admin permissions are not enabled for a realm?

pedroigor: True, I'm removing this. It should always run using v2 if the feature is enabled (even if realm setting tells otherwise).

docs/documentation/server_admin/topics/admin-console-permissions/fine-grain-v2.adoc
```diff
@@ -18,6 +18,23 @@ to realm administrators but doing so will skip the mechanisms provided by this f
...
+
+* **Realm administrators**: These are users (or service accounts) that have been granted with the `realm-admin` role in a specific realm.
+
+Both types of administrators can manage a realm with full access to all their resources and
+administrative operations.
+
+The main difference between them is that server administrators can manage any realm in the server, while realm administrators can only manage
+the realm they have been granted access to.
+
+Be aware that server administrators are not affected by the permissions you define when managing access to realm resources through this feature.
```
vramik:
```suggestion
Be aware that both server and realm administrators are not affected by the permissions you define when managing access to realm resources through this feature.
```

docs/documentation/server_admin/topics/admin-console-permissions/fine-grain-v2.adoc
```diff
@@ -210,8 +227,8 @@ properly allow paginating resources as well as avoid an additional overhead on t
 realm resource returned by queries.
 
 Partial evaluation and filtering occurs only if the feature is enabled to a realm, and if the user is not granted
-with view-related administrative roles like `view-users` or `view-clients`. For instance, it will not happen for regular server administrators granted
-with the `admin` role.
+with view-related administrative roles like `view-users` or `view-clients`. For instance, it will not happen for realm administrators granted
```
vramik:
```suggestion
with view-related administrative roles like `view-users` or `view-clients`. For instance, it will not happen for administrators granted
```

services/src/main/java/org/keycloak/services/resources/admin/fgap/AdminPermissions.java
vramik: Thank you for the confirmation.

vramik (commented):
Thanks @pedroigor, this is great.

- I've suggested two small doc changes to improve clarity. Let me know what you think.
- It also looks like some tests need updating.

Once the tests are addressed, this should be ready to merge.

vramik (approved):

ahus1 (approved):

PR #43590, opened 2025-10-18
Adding a graceful shutdown for the HTTP stack

### LINKED ISSUE #30986: Graceful Shutdown with connection draining
### Description

Quarkus supports special [graceful shutdown features](https://quarkus.io/guides/lifecycle#graceful-shutdown) which are disabled by default. Keycloak could use those to enhance the graceful shutdown mechanism and drain connections before actually shutting down the server. 

### Discussion

_No response_

### Motivation

We're deploying Keycloak in clustered mode with 3 cluster nodes in Kubernetes. For that we use a StatefulSet with a RollingUpdate strategy. When we rollout a new image version, we see that a fraction of requests timeout, and sometimes a pod is not terminated correctly. 
Currently, I assume that this is related to a combination of connections that are not drained, sticky session and an update delay of our ingress controller. However, this is hard to debug. 

As a workaround, we have added a `preStop` lifecycle hook with a 15s delay which seems to work just fine and connections are drained properly.  

### Details

_No response_

### LINKED ISSUE #43589: Gracefully shutting down HTTP stack
### Description

When shutting down Keycloak, any currently in-progress HTTP requests are terminated and do not complete. 

Unfortunately the functionality in 

### Value Proposition

Any in-progress HTTP requests should complete before stopping Keycloak. This helps with graceful rolling restarts and updated.

### Goals

Avoid request errors for callers. With a request aborting, the caller doesn't know if the request completed. 

### Non-Goals

Graceful shutdown of other tasks, for example background jobs.

### Discussion

_No response_

### Notes

While this is not supported in Quarkus yet, let's explore how this can work in Keycloak and then contribute it to Quarkus.

ahus1:
Closes #43589
Closes #30986

TODO: 

* [ ] Make `quarkus.shutdown.delay` configurable via CLI. 
* [ ] Either use the same timeout for the graceful filter, or a different one. Not sure yet.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.quarkus.runtime.filter;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import jakarta.enterprise.inject.spi.CDI;
+
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.quarkus.runtime.shutdown.ShutdownListener;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpVersion;
+import io.vertx.core.http.impl.Http1xServerConnection;
+import io.vertx.core.http.impl.Http2ServerConnection;
+import io.vertx.ext.web.RoutingContext;
+import org.jboss.logging.Logger;
+
+/**
+ * Provide a graceful shutdown. Inspired by {@link io.quarkus.vertx.http.runtime.filters.GracefulShutdownFilter}.
+ * <p>
+ * This explores options while we wait for <a href="https://github.com/keycloak/keycloak/issues/43589">#43589</a>.
+ */
+public class KeycloakGracefulShutdownFilter implements ShutdownListener, Handler<RoutingContext> {
+
+    private static final Logger log = Logger.getLogger(KeycloakGracefulShutdownFilter.class);
+
+    private volatile boolean running = true;
+    private final AtomicInteger currentRequestCount = new AtomicInteger();
+    private final AtomicReference<ShutdownNotification> notification = new AtomicReference<>();
+
+    private final Handler<AsyncResult<Void>> requestDoneHandler = event -> {
+        int count = currentRequestCount.decrementAndGet();
+        if (!running && count == 0) {
+            executeNotification();
+        }
+    };
+    private volatile Long timer;
+
+    private void executeNotification() {
+        ShutdownNotification n = notification.get();
+        if (n != null && notification.compareAndSet(n, null)) {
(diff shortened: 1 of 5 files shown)


ahus1:
@keycloak/cloud-native: It would be great to get your input on this and if you would be willing to add this to 26.5. Thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
```diff
@@ -0,0 +1,142 @@
...
+                ShutdownNotification n = this.notification.get();
+                if (this.notification.compareAndSet(n, null)) {
+                    if (n != null) {
+                        n.done();
+                        log.info("Timeout reached waiting for requests to complete");
+                        if (timer != null) {
+                            getVertx().cancelTimer(timer);
+                        }
+                    }
+                }
```
shawkins: Nit: the original notification could be used.

```suggestion
                if (this.notification.compareAndSet(notification, null)) {
                      notification.done();
                      log.info("Timeout reached waiting for requests to complete");
                      if (timer != null) {
                          getVertx().cancelTimer(timer);
                      }
                }
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private static final Logger log = Logger.getLogger(KeycloakGracefulShutdownFilter.class);
+
+    private volatile boolean running = true;
+    private final AtomicInteger currentRequestCount = new AtomicInteger();
+    private final AtomicReference<ShutdownNotification> notification = new AtomicReference<>();
+
+    private final Handler<AsyncResult<Void>> requestDoneHandler = event -> {
+        int count = currentRequestCount.decrementAndGet();
+        if (!running) {
```
shawkins: Nit: the nested if could be combined.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
```diff
@@ -0,0 +1,142 @@
...
+            if (count == 0) {
+                executeNotification();
+            }
+        }
+    };
+    private Long timer;
+
+    private void executeNotification() {
+        ShutdownNotification n = notification.get();
+        if (n != null) {
```
shawkins: Nit: the nested if could be combined.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
```diff
@@ -0,0 +1,142 @@
...
+    public void preShutdown(ShutdownNotification notification) {
+        this.notification.set(notification);
+        running = false;
+        if (currentRequestCount.get() == 0) {
+            if (this.notification.compareAndSet(notification, null)) {
+                notification.done();
+            }
+        } else {
+            log.info("Waiting for HTTP requests to complete");
+            timer = getVertx().setTimer(TimeUnit.SECONDS.toMillis(10), event -> {
```
shawkins, line 124: This timeout will likely need to be configurable. For Kubernetes it will affect what users can have for their graceful termination seconds - it defaults to 30, but users can specify whatever they want in their unsupported pod template.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
```diff
@@ -0,0 +1,142 @@
...
+    };
+    private Long timer;
+
+    private void executeNotification() {
+        ShutdownNotification n = notification.get();
+        if (n != null) {
+            if (notification.compareAndSet(n, null)) {
+                n.done();
+                log.info("All HTTP requests complete");
+                if (timer != null) {
```
shawkins: Is timer cancellation required? If so, there's a possible race-condition where cancellation is skipped.

shawkins (commented):
This looks good @ahus1, just a couple thoughts / nits.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/filter/KeycloakGracefulShutdownFilter.java
ahus1: No, cancellation is not required in vertx

ahus1:
With the upcoming changes in Quarkus, we'll be just reusing the Quarkus options. See https://github.com/keycloak/keycloak/issues/43589 a discussion

Closing this PR in favor of a different PR that should arrive eventually. 

Please help reviewing the updated specs in #43589

PR #43594, opened 2025-10-19
Add configurable SMTP timeouts

### LINKED ISSUE #14509: Allow configuration for SMTP timeouts via configuration
### Description

Hello,

I noticed that the code that sends emails in keycloak sets 2 timeout properties with the value `10000` in [DefaultEmailSenderProvider](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java) (I checked both branches `main` and `nightly`):

```java
props.setProperty("mail.smtp.timeout", "10000");
props.setProperty("mail.smtp.connectiontimeout", "10000");
```

It would be nice to be able to control these properties either using the already existing property for socket timeout, or by defining a new one for SMTP timeouts. I stumbled upon this because I have a provider who has been flaky and noticed some emails not being delivered correctly.

Do you think this is an useful feature?

Thanks!

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

### LINKED ISSUE #35836: SMTP Timeout Override
### Description

We have a relay that our keycloak sends SMTP to and then chooses a provider. Based on the provider, the timeout can take more than the default found here: https://github.com/keycloak/keycloak/blob/26.0.7/services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java#L111

Can this be configurable?

### Discussion

_No response_

### Motivation

Because the timeout is being hardcoded to 10000 and there may be situations like ours where it should be extended and configurable.

### Details

_No response_

yanxch:
closes #35836 
closes #14509


pedroigor (commented):
Would it be possible to make the new settings available from the Admin UI?

shawkins:
> Would it be possible to make the new settings available from the Admin UI?

Could also be done in a follow up as it doesn't appear that any of the properties are exposed by DefaultEmailSenderProviderFactory.getConfigMetadata

pedroigor:
> > Would it be possible to make the new settings available from the Admin UI?
> 
> Could also be done in a follow up as it doesn't appear that any of the properties are exposed by DefaultEmailSenderProviderFactory.getConfigMetadata

This settings should be defined on a per-realm basis, no? Not sure if it makes sense to define on the server level.

shawkins:
> This settings should be defined on a per-realm basis, no? Not sure if it makes sense to define on the server level.

My mistake, at first glance I thought the properties were coming from the standard Scope config for the provider.

pedroigor (commented):
Updated the admin console. @yanxch Does it look good to you?

pedroigor (approved):

PR #43599, opened 2025-10-20
[OID4VCI] Ensure `openid_credential` is one of `authorization_details_types_supported` on the Authorization Server metadata

IngridPuppet:
I'm sending this PR on behalf of @forkimenjeckayang, on vacation.
Closes #43398 

## Summary

This PR implements the OID4VC specification requirement to include `authorization_details_types_supported` in the OAuth Authorization Server metadata endpoint. This enables credential issuance using `authorization_details` when scope is absent, as required by the OID4VC specification.

## Test Coverage

- **Metadata validation**: Verifies `authorization_details_types_supported` is present and contains `openid_credential`
- **End-to-end flow**: Complete credential issuance using `authorization_details`
- **Cross-reference validation**: Ensures credential issuer metadata points to correct OAuth server
- **Feature flag testing**: Validates behavior when OID4VC is disabled

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/OIDCWellKnownProvider.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/OIDCWellKnownProvider.java
@@ -27,6 +27,7 @@ import org.keycloak.crypto.ClientSignatureVerifierProvider;
 import org.keycloak.crypto.ContentEncryptionProvider;
 import org.keycloak.crypto.CryptoUtils;
 import org.keycloak.crypto.SignatureProvider;
+import org.keycloak.protocol.oid4vc.issuance.OID4VCAuthorizationDetailsProcessor;
 import org.keycloak.jose.jws.Algorithm;
 import org.keycloak.models.CibaConfig;
 import org.keycloak.models.ClientScopeModel;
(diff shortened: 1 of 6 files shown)


services/src/main/java/org/keycloak/protocol/oidc/OIDCWellKnownProvider.java
```diff
@@ -215,6 +216,10 @@ public Object getConfig() {
 
         config.setAuthorizationResponseIssParameterSupported(true);
 
+        if (Profile.isFeatureEnabled(Profile.Feature.OID4VC_VCI) && realm.isVerifiableCredentialsEnabled()) {
```
mposolda: Can we instead retrieve this from available providers? I mean something similar like we're doing for example for retrieve the available signatures? I suppose that something like this may work:
```
session.getKeycloakSessionFactory().getProviderFactoriesStream(AuthorizationDetailsProcessor.class)
                .map(ProviderFactory::getId)
```
This automatically handles the case when `OID4VC_VCI` feature is disabled as in that case the provider factory will be unavailable.

It may also need to instantiate the processor and have new method on the `AuthorizationDetailsProcessor` like `boolean isSupported()` with the possible implementation for oid4vci like:
```
boolean isSupported() {
   return session.getContext().getRealm().isVerifiableCredentialsEnabled();
} 
```

This approach means that provider_id of the `OID4VCAuthorizationDetailsProcessorFactory` will need to be changed to be `OID4VCAuthorizationDetailsProcessor.OPENID_CREDENTIAL_TYPE`, but that is probably OK?

mposolda (commented):
@IngridPuppet Thanks, added inline comment.

mposolda:
@IngridPuppet Thanks for the update!

mposolda (approved):

services/src/main/java/org/keycloak/protocol/oidc/OIDCWellKnownProvider.java
IngridPuppet: Thank you @mposolda for your detailed review comment. I agree with the update for extensibility and updated the code accordingly.

PR #43601, opened 2025-10-20
Add stopSessionPolling when the document is loaded

### LINKED ISSUE #42730: Automatic log-in in another tab is not always working
### Before reporting an issue

### Area

login/ui

### Describe the bug

Sometimes the automatic log-in in another tab is not working.

By briefly checking the code of `authChecker.js` I can guess the reason:

1) The script is run from the `<head>` section of the HTML doc
2) It immediately refers to the list of forms https://github.com/keycloak/keycloak/blob/37a99154a5ac26c7968beea334efb99727305656/themes/src/main/resources/theme/base/login/resources/js/authChecker.js#L4

while the DOM content is not guaranteed to be parsed.
Thus the form may not exist yet and all further logic won't be effective.

Consider using DOMContentLoaded event (https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) or another way to guarantee referring to the elements from the DOM in robust way.

### Version

26.2.5; main

### Regression

### Expected behavior

Auto-login should work

### Actual behavior

Sometimes the login page remains intact.

### How to Reproduce?

Open the needed app in 2 tabs which causes redirects to Keycloak login pages; log-in in one of them.

### Anything else?

It's a random bug.

rmartinc:
Closes #42730

Add the `stopSessionPolling` to forms when `DOMContentLoaded` event is called. I'm not sure this is the real cause for the original issue but it makes no harm.

### DIFF
--- a/themes/src/main/resources/theme/base/login/resources/js/authChecker.js
+++ b/themes/src/main/resources/theme/base/login/resources/js/authChecker.js
@@ -1,14 +1,14 @@
 const SESSION_POLLING_INTERVAL = 2000;
 const AUTH_SESSION_TIMEOUT_MILLISECS = 1000;
 const initialSession = getSession();
-const forms = Array.from(document.forms);
 let timeout;
 
 // Stop polling for a session when a form is submitted to prevent unexpected redirects.
 // This is required as Safari does not support the 'beforeunload' event properly.
 // See: https://bugs.webkit.org/show_bug.cgi?id=219102
-forms.forEach((form) =>
-  form.addEventListener("submit", () => stopSessionPolling()),
+document.addEventListener("DOMContentLoaded", () =>
+  Array.from(document.forms).forEach((form) =>
+    form.addEventListener("submit", () => stopSessionPolling()))
 );
 
 // Stop polling for a session when the page is unloaded to prevent unexpected redirects.


jonkoops (approved):

themes/src/main/resources/theme/base/login/resources/js/authChecker.js
```diff
@@ -1,14 +1,14 @@
...
 const initialSession = getSession();
-const forms = Array.from(document.forms);
 let timeout;
 
 // Stop polling for a session when a form is submitted to prevent unexpected redirects.
 // This is required as Safari does not support the 'beforeunload' event properly.
 // See: https://bugs.webkit.org/show_bug.cgi?id=219102
-forms.forEach((form) =>
-  form.addEventListener("submit", () => stopSessionPolling()),
+document.addEventListener("DOMContentLoaded", () =>
```
jonkoops, line 9: I don't believe this will fix the issue, as this script already has `type=module` attribute, meaning [its execution is deferred by default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#other_differences_between_modules_and_classic_scripts), this cause it to have the [`defer` attributte](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script#defer):

> This Boolean attribute is set to indicate to a browser that the script is meant to be executed after the document has been parsed, but before firing [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) event.

Basically, this means that the DOM should already be fully available at the time it executes. I've approved the PR, but I don't think this change will make a difference.

rmartinc: Then maybe we can close this directly. I'm going to ask if they are using safari. If it's safari we can consider to include this. If not, I think the PR makes no sense.

rmartinc: Moving to draft to not merge it for the moment.

rmartinc:
Closing the PR as we think that the issue is not even relate dto this part. Besides as @jonkoops has commented, module is deferred by default.

Issue #38991, opened 2025-04-15
[Test framework] Embedded server -> dependency download error when no version is specified


lhanusov:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

testsuite

### Describe the bug

there is a bug, when loading e.g. jdbc driver for embedded server: https://github.com/keycloak/keycloak/blob/main/test-framework/db-oracle/src/main/java/org/keycloak/testframework/database/OracleDatabaseSupplier.java#L22

The problem is caused by missing version when downloading a dependency. It should be added then as a provider to Keycloak in an embedded mode.

```
2025-04-15 16:18:27,285 ERROR [io.quarkus.bootstrap.resolver.maven.NonBlockingModelResolutionTaskRunner] (main) 1): io.quarkus.bootstrap.resolver.maven.DeploymentInjectionException: Failed to resolve artifact com.oracle.database.jdbc:ojdbc17:jar:
	at io.quarkus.bootstrap.resolver.maven.ApplicationDependencyResolver.resolve(ApplicationDependencyResolver.java:869)
	at io.quarkus.bootstrap.resolver.maven.ApplicationDependencyResolver$AppDep.getResolvedArtifact(ApplicationDependencyResolver.java:714)
	at io.quarkus.bootstrap.resolver.maven.ApplicationDependencyResolver$AppDep.visitRuntimeDependency(ApplicationDependencyResolver.java:557)
	at io.quarkus.bootstrap.resolver.maven.NonBlockingModelResolutionTaskRunner.lambda$run$0(NonBlockingModelResolutionTaskRunner.java:34)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run$$$capture(CompletableFuture.java:1804)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
	at java.base/java.util.concurrent.ForkJoinTask.doExec$$$capture(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
[...]

PR #43609, opened 2025-10-20
Test Framework support for remotely running Databases

vaceksimon:
Part of #41940

This PR enables testing with a remote database (such as Aurora, MariaDB, ...), instead of a containerized one from the TestContainers library. 

An example of running tests with a remote Postgres db:
```shell
KC_TEST_DATABASE=remote \
  KC_TEST_DATABASE_VENDOR=postgres \
  KC_TEST_DATABASE_USER=testUser \
  KC_TEST_DATABASE_PASSWORD=password \
  KC_TEST_DATABASE_URL=jdbc:postgresql://localhost:5432/keycloak \
  mvn test -Dtest=DatabaseTestSuite
```

All supported databases should work, except when using an Oracle database with an embedded Keycloak server (described in #38991).
Clustered Keycloak server is not tested.

### DIFF
--- a/test-framework/HOW_TO_RUN.md
+++ b/test-framework/HOW_TO_RUN.md
@@ -122,12 +122,43 @@ Valid values:
 | oracle   | Oracle test container                   |
 | postgres | PostgreSQL test container               |
 | tidb     | TiDb test container                     |
+| remote   | Connect to a remotely running database  |
 
 Configuration:
 
 | Value                                               | Description                                                                                                                                                                 |
 |-----------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `kc.test.database.reuse` / `KC_TEST_DATABASE_REUSE` | Set to true to enable reuse of database. Requires [enabling reuse for Testcontainers](https://java.testcontainers.org/features/reuse/) (`TESTCONTAINERS_REUSE_ENABLE=true`) |
+
+#### Remote Database
+
+If connecting to a remotely running database is desired, the following options need to be specified:
+
+Option prefix: `kc.test.database.` / `KC_TEST_DATABASE_`
+
+| Option                              | Description                                                         | Required? |
+|-------------------------------------|---------------------------------------------------------------------|-----------|
+| `vendor`/`VENDOR`                   | Database vendor (valid values mentioned above)                      | yes       |
+| `user`/`USER`                       | Username of the database user                                       | yes       |
+| `password`/`PASSWORD`               | Password of the database user                                       | yes       |
+| `url`/`URL`                         | Full database JDBC URL                                              | yes       |
+| `driver`/`DRIVER`                   | Fully qualified class name of the JDBC driver                       | no        |
+| `driver.artifact`/`DRIVER_ARTIFACT` | Maven artifact containing the driver in format `groupId:artifactId` | no        |
+
+Because some databases may require a special driver, you can specify it to the Keycloak server. To use it, Keycloak needs
+to add it as a dependency, which is why you need to specify the Maven artifact coordinates.
+
+For example, you want to run tests with an Oracle database which needs a specific jdbc driver. You would run:
+```shell
+KC_TEST_DATABASE=remote \
+  KC_TEST_DATABASE_VENDOR=oracle \
+  KC_TEST_DATABASE_USER=testUser \
+  KC_TEST_DATABASE_PASSWORD=password \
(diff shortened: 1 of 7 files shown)


test-framework/core/src/main/java/org/keycloak/testframework/database/RemoteTestDatabase.java
```diff
@@ -0,0 +1,64 @@
...
+        // noop
+    }
+
+    @Override
+    public void stop() {
+        // noop
+    }
+
+    private String getDatabaseVendor() {
+        return getDbConfigOption("db");
```
stianst: We have some direct support for this stuff; it can also be injected into the suppliers.

This should work:

```suggestion
        return Config.getValueTypeFQN(RemoteTestDatabase.class, "db");
```

test-framework/HOW_TO_RUN.md
```diff
@@ -122,13 +122,46 @@ Valid values:
...
+
+| Option                                  | Description                                            | Required? |
+|-----------------------------------------|--------------------------------------------------------|-----------|
+| `db`/`DB`                               | Database vendor (valid values mentioned above)         | yes       |
+| `user`/`USER`                           | Username of the database user                          | yes       |
+| `password`/`PASSWORD`                   | Password of the database user                          | yes       |
+| `url`/`URL`                             | Full database JDBC URL                                 | yes       |
+| `driver`/`DRIVER`                       | Fully qualified class name of the JDBC driver          | no        |
+| `driver.groupid`/`DRIVER_GROUPID`       | GroupId of the maven artifact containing the driver    | no        |
+| `driver.artifactid`/`DRIVER_ARTIFACTID` | ArtifactId of the maven artifact containing the driver | no        |
```
stianst: Should we maybe have a single property for the driver gav instead of two?

test-framework/core/src/main/java/org/keycloak/testframework/database/RemoteTestDatabase.java
vaceksimon: I changed it as per your suggestion, but `Config.getValueTypeFQN(RemoteTestDatabase.class, "db")` returns only the config option, not the value itself (returns `kc.test.database.remote.db`).

I kept the two methods `getDbConfigOption` and `getOptionalDbConfigOption`, which use `Config.get()`, which can throw an exception if the configuration is not found.

test-framework/HOW_TO_RUN.md
```diff
@@ -157,8 +156,7 @@ KC_TEST_DATABASE=remote \
   KC_TEST_DATABASE_REMOTE_PASSWORD=password \
   KC_TEST_DATABASE_REMOTE_URL=jdbc:oracle:thin:@localhost:1521/keycloak \
   KC_TEST_DATABASE_REMOTE_DRIVER=oracle.jdbc.OracleDriver \
-  KC_TEST_DATABASE_REMOTE_DRIVER_GROUPID=com.oracle.database.jdbc \
-  KC_TEST_DATABASE_REMOTE_DRIVER_ARTIFACTID=ojdbc17 \
+  KC_TEST_DATABASE_REMOTE_DRIVER_ARTIFACT=com.oracle.database.jdbc.ojdbc17 \
```
stianst: Should use `:` as delimiter as that's standard in Maven GAVs + an artifactId can contain a `.` ;)

```suggestion
  KC_TEST_DATABASE_REMOTE_DRIVER_ARTIFACT=com.oracle.database.jdbc:ojdbc17 \
```

test-framework/core/src/main/java/org/keycloak/testframework/database/RemoteTestDatabase.java
```diff
@@ -40,11 +38,13 @@ private String getDatabaseDriver() {
     }
 
     static String getDbConfigOption(String option) {
-        return Config.get(CONFIG_PREFIX + option, String.class);
+        String fqn = Config.getValueTypeFQN(RemoteTestDatabase.class, option);
```
stianst: Sorry, I gave you the wrong method. `getValueTypeConfig` is what you want as that gives you the value directly.

test-framework/HOW_TO_RUN.md
vaceksimon: Right! I knew that. I was just testing if you do :zany_face:

stianst (approved):

lhanusov (commented):
LGTM

PR #43627, opened 2025-10-21
Remove dependency on keycloak-server-spi-private from keycloak-model-…

### LINKED ISSUE #43626: Remove dependency on keycloak-server-spi-private from keycloak-model-storage
### Description

Module `keycloak-model-storage` contains the classes of `UserStorageProviderSpi` , which is one of Keycloak officially supported SPIs. Considering this, this module is suitable for using in the Keycloak extensions created by Keycloak customers/users, which are trying to implement their own user-storage .

However the `keycloak-model-storage` contains the dependency on `keycloak-server-spi-private`, which is intended to be the "private" Keycloak module, which people ideally should not use in their own extensions.

### Value Proposition

Current situation means that every customer/user, who wants to consume the "officially supported" `UserStorageProviderSpi` in his Keycloak extension needs to have also dependency on `keycloak-server-spi-private` in his extension.

Can we improve this and make sure that `keycloak-model-storage` does not have dependency on `keycloak-server-spi-private` ?

### Goals

- Make sure that `keycloak-model-storage` does not have dependency on `keycloak-server-spi-private` 
- Make sure that people creating their own user-storage extension does not need to have dependency on `keycloak-server-spi-private` .

### Non-Goals

- Just the task mentioned in the "Goals" are supposed to be the target of this enhancement

### Discussion

_No response_

### Notes

There was the issue #43694 , which was used as "first step" towards this task to give people a chance to migrate to 26.5 and then later to 27 without breaking changes. This task is intended to be done after Keycloak 26.5 release.

Regarding this one, we may need to:

- Remove the "empty" class `org.keycloak.models.UserCredentialManager` . Then rename `org.keycloak.models.SubjectCredentialManager` to `org.keycloak.models.UserCredentialManager` As discussed in the PR https://github.com/keycloak/keycloak/pull/43695#issuecomment-3460918343 .
- Rename `org.keycloak.credential.UserCredentialManager` to `org.keycloak.credential.DefaultUserCredentialManager` (and move it to `keycloak-server-spi-private` )
- Do some additional changes done in the draft PR https://github.com/keycloak/keycloak/pull/43627 in order to make it possible to remove dependency on `keycloak-server-spi-private` from `keycloak-model-storage` . Those other changes are more straightforward
- Update documentation (Upgrading guide, maybe release notes)
- Check quickstarts afterwards. The `user-storage-simple` extension quickstart probably will not need any changes and at the same time, it will not have dependency on `keycloak-server-spi-private` but just on `keycloak-server-spi` and `keycloak-model-storage` .

mposolda:
…storage

closes #43626

The aim is to remove dependency on `keycloak-server-spi-private` from the `keycloak-model-storage` module, so that people don't need to use `keycloak-server-spi-private` from their own user-storage extensions.

If the approach is OK, I can add also a note to the release-notes or upgrading-guide regarding this change.

### Details of the PR

- Moved `UserModelDefaultMethods` to the `keycloak-server-spi` . Also `ExecutorProvider` and it's factory. Those are simple utility class and I don't see issues with moving.

- Moved `UserCredentialManager` from `keycloak-model-storage` to `keycloak-server-spi-private` . This class is referenced from `AbstractUserAdapterFederatedStorage`, which is referenced from our docs and is supposed to be used in Keycloak user extensions. To remove the reference from the class, I've added new method `getUserCredentialManager` to `KeycloakSession` .

I am not so happy with adding new method on `KeycloakSession`, but not sure about better approach? Some alternative approaches, which may not require any changes on `KeycloakSession` are:

1) Add some utility method (for example to `UserStorageUtil`), which will use reflection to reference `UserCredentialManager`. The `AbstractUserAdapterFederatedStorage` can call then this utility method.

2) Convert `SubjectCredentialManager` to be the `Provider` and create some provider factory and SPI for it. Then `AbstractUserAdapterFederatedStorage` could use something like `return session.getProvider(SubjectCredentialManager.class);` .

3) Keep `UserCredentialManager` in the `keycloak-model-storage`. This would require some more dependencies to be declared in `keycloak-model-storage` (like `opentelemetry` and it's dependencies) and much more changes (EG. moving `DatastoreProvider` to `keycloak-model-storage` etc). 

4) Any better alternative?    
    
Another side-effect of moving `UserCredentialManager` is that `keycloak-model-storage` is now dependency of `keycloak-server-spi-private` (as `UserCredentialManager` has some more references on some classes from `keycloak-model-storage`). I don't see any issue with that one.


rmartinc (approved):
LGTM @mposolda! It's not a big change in the end. The only thing is the `getUserCredentialManager` addition is in the `server-spi`, is that a problem? (I don't expect anyone implementing its own session but...)

mposolda:
> LGTM @mposolda! It's not a big change in the end. The only thing is the `getUserCredentialManager` addition is in the `server-spi`, is that a problem? (I don't expect anyone implementing its own session but...)

Yes, I also don't expect this to be a problem. I think we are not expecting people to implement their own instances of things like `KeycloakSession` , `RealmModel` (or any other models) or similar? As otherwise we would not be able to add any new methods there.

IMO adding new method is fine, but updating/removing signature of existing methods is problematic as people may already invoke existing methods from their providers.

mposolda:
Closing for now as this issue will be done in Keycloak 27 to avoid breaking changes between 26.4 and 26.5.

Some changes from this PR were done as part of https://github.com/keycloak/keycloak/issues/43694 .

Issue #35266, opened 2024-11-24
Amazon Identity Provider does not accept scope = openid and Keycloak always sets it


paolotateo:
### Before reporting an issue

- [X] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

identity-brokering

### Describe the bug

Today I tried to setup "Login by Amazon" using the OpenID Connect v1.0 provider, but it didn't work as Amazon responds "An unknown scope was requested". 
Actually I compiled the configuration by inserting "Scopes = profile", which is an admissible value, but Keycloak sends the value "openid+profile". 
In the field tip I read "The scopes to be sent when asking for authorization. It can be a space-separated list of scopes. Defaults to 'openid'."
I guess I can call it a bug, as I should have the option to not send Amazon the value openid.

Does anyone know where the code to fix this is located?

### Version

26.0.4

### Regression

- [ ] The issue is a regression

### Expected behavior

Keycloak should send to the authentication provider only the value (or values) that I enter in the "Scopes" configuration field (ex. profile)

### Actual behavior

Today keycloak also concatenates the openid value (ex. openid+profile)

### How to Reproduce?

Set up a new OpenID identity provider and try to authenticate

pedroigor:
The `openid` scope is mandatory according to the OpenID Connect specification. Otherwise, it is not an OIDC request.

I'm not sure if it makes sense to allow removing the scope from requests when using the OIDC Generic broker but provide either an Amazon IdP compliant broker or an OAuth2 generic provider (perhaps the same as the OIDC but without some OIDC specific settings like the `openid` default scope).

Co-7:
I have the same problem when trying to use the connection with Spotify. It doesn't support oidc, only oAuth 2, and because Keycloak forces the addition of oidc, I can't get Spotify to work. But if I just remove this scope in the url, everything works just fine... It's sad to use keycloak and not be able to use Amazon or Spotify because of it.

Is there any way with any option to remove this addition?

omertahaoztop:
I have the same problem with Yandex, Yandex allows three scopes; login_email, login_info and login_avatar, but since the openid scope is sent, it gives invalid_scope error and prevents me from logging into Keycloak with Yandex.

I will report this to Yandex, but the option to remove the openid scope would be nice.

Thanks for your help.

kominoshja:
[...]

PR #43630, opened 2025-10-21
Correct OAuth 2.0 reference

andymunro:
Closes #43629


ahus1 (changes requested):
When adding Identity providers in Keycloak, it supports several integrations that are based OAuth2 and not on OIDC. This includes Stackoverflow, Bitbucket, GitHub, Instagram, Facebook, Paypal and Microsoft. 

As part of https://github.com/keycloak/keycloak/issues/35266 there is now also a generic OAuth2 identity provider in Keycloak.

So I'd rather keep the statement as is, although it might need some more explanation.

andymunro:
No longer valid.

PR #43634, opened 2025-10-21
win-fix: correct hostname normalization condition for loopback addresses

### LINKED ISSUE #42794: [Windows] Make TrustedHostClientRegistrationPolicyTest Robust to Canonical Hostname localhost

Pepo48:
Closes: #42794

### DIFF
--- a/docs/guides/securing-apps/client-registration.adoc
+++ b/docs/guides/securing-apps/client-registration.adoc
@@ -200,6 +200,9 @@ Currently we have these policy implementations:
 Request sent from some untrusted IP will be rejected. URLs of newly registered client must also use just those trusted hosts or domains. For example it won't be allowed
 to set `Redirect URI` of client pointing to some untrusted host. By default, there is not any whitelisted host, so anonymous client registration is de-facto disabled.
 
+NOTE: {project_name} normalizes loopback addresses to `localhost` to avoid platform-specific
+reverse-DNS differences (for example on Windows a reverse lookup of loopback may return the numeric IP rather than `localhost`). All other trusted domains must be based upon the primary reverse-DNS lookup returned by the platform.
+
 * Consent Required Policy - Newly registered clients will have `Consent Allowed` switch enabled. So after successful authentication, user will always
 see consent screen when he needs to approve permissions (client scopes). It means that client won't have access to any personal
 info or permission of user unless user approves it.
--- a/services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
+++ b/services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
@@ -175,22 +175,19 @@ public class TrustedHostClientRegistrationPolicy implements ClientRegistrationPo
 
             logger.debugf("Trying verify request from address '%s' of host '%s' by domains", hostAddress, hostname);
 
-            // On Windows, reverse lookup for loopback may return the IP (e.g., 127.0.0.1) instead of 'localhost'. Normalize to 'localhost' for consistent domain checks.
-            if (hostname.equals(address.getHostAddress()) && address.isLoopbackAddress()) {
+            // On Windows, reverse lookup for loopback may return the IP (e.g., 127.0.0.1) instead of 'localhost'.
+            // Normalize to 'localhost' for consistent domain checks.
+            if (address.isLoopbackAddress()) {
                 hostname = "localhost";
-            }
-
-            if (hostname.equals(address.getHostAddress())) {
+            } else if (hostname.equals(address.getHostAddress())) {
                 logger.debugf("The hostAddress '%s' was not resolved to a hostname", hostAddress);
                 return null;
             }
 
-            // For loopback, skip strict forward-confirm check after normalizing to 'localhost' to avoid platform differences.
-            if (!address.isLoopbackAddress()) {
-                if (Arrays.stream(InetAddress.getAllByName(hostname)).filter(a -> address.equals(a)).findAny().isEmpty()) {
-                    logger.debugf("The hostAddress '%s' is not among the direct lookups returned resolving '%s'", hostAddress, hostname);
-                    return null;
-                }
+            // For non-loopback addresses, perform a forward-confirmation check: the hostname must resolve back to the same address.
+            if (!address.isLoopbackAddress() && Arrays.stream(InetAddress.getAllByName(hostname)).noneMatch(a -> address.equals(a))) {
+                logger.debugf("The hostAddress '%s' is not among the direct lookups returned resolving '%s'", hostAddress, hostname);
+                return null;
             }
 
             for (String confDomain : trustedDomains) {


services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
```diff
@@ -176,7 +176,7 @@ protected String verifyHostInTrustedDomains(String hostAddress, List<String> tru
             logger.debugf("Trying verify request from address '%s' of host '%s' by domains", hostAddress, hostname);
 
             // On Windows, reverse lookup for loopback may return the IP (e.g., 127.0.0.1) instead of 'localhost'. Normalize to 'localhost' for consistent domain checks.
-            if (hostname.equals(address.getHostAddress()) && address.isLoopbackAddress()) {
+            if (!hostname.equals(address.getHostAddress()) && address.isLoopbackAddress()) {
```
shawkins: To match the existing comment and cover the case of resolving to an ip or non-localhost, you could use:

```suggestion
            if (NetworkUtils.checkForWindows() && address.isLoopbackAddress()) {
```

Alternatively if the hostname cannot return the ip, then the comment needs changed.

Or if we are documenting in some way that localhost is always the cannonical hostname for loopback, then we can also just unconditionally do:

```
if (address.isLoopbackAddress()) {
```

Pepo48: Thanks, Steve!

I mentioned the removal of the first condition during the meeting as the second option. @vmuzikar any preference here?

vmuzikar: +1 for just keeping the second condition.

Pepo48: Thanks guys. Given that we removed the first condition, I rearranged the code a bit and added a note to the Client Registration guide as Steve suggested. Is that ok?

docs/guides/securing-apps/client-registration.adoc
```diff
@@ -200,6 +200,12 @@ Currently we have these policy implementations:
 Request sent from some untrusted IP will be rejected. URLs of newly registered client must also use just those trusted hosts or domains. For example it won't be allowed
 to set `Redirect URI` of client pointing to some untrusted host. By default, there is not any whitelisted host, so anonymous client registration is de-facto disabled.
 
+NOTE: When reverse-DNS returns a hostname for a loopback address, {project_name} normalizes that hostname to `localhost` to avoid platform-specific
+reverse-DNS differences (for example on Windows a reverse lookup of loopback may return the numeric IP rather than `localhost`). If reverse-DNS
+returns only the numeric IP (for example `127.0.0.1`), the host is treated as unresolved and the request will be rejected. Therefore, when
+configuring trusted hosts/domains or validating client URLs, loopback-originating requests that resolve to a hostname are treated as coming from
+`localhost`.
```
shawkins:
```suggestion
NOTE: {project_name} normalizes loopback addresses to `localhost` to avoid platform-specific
reverse-DNS differences (for example on Windows a reverse lookup of loopback may return the numeric IP rather than `localhost`). All other trusted domains must be based upon the primary reverse-DNS lookup returned by the platform.
```

services/src/main/java/org/keycloak/services/clientregistration/policy/impl/TrustedHostClientRegistrationPolicy.java
```diff
@@ -175,22 +175,19 @@ protected String verifyHostInTrustedDomains(String hostAddress, List<String> tru
 
             logger.debugf("Trying verify request from address '%s' of host '%s' by domains", hostAddress, hostname);
 
-            // On Windows, reverse lookup for loopback may return the IP (e.g., 127.0.0.1) instead of 'localhost'. Normalize to 'localhost' for consistent domain checks.
-            if (hostname.equals(address.getHostAddress()) && address.isLoopbackAddress()) {
-                hostname = "localhost";
-            }
-
             if (hostname.equals(address.getHostAddress())) {
```
shawkins: To double-check - this is ok, if the original comment was wrong. That is windows won't resolve 127.0.0.1 to 127.0.0.1, it will always be to a hostname, correct?

Pepo48: Well, as I mentioned previously during our discussions, I am hesitant to say that it is a general rule that I am 100% certain about.

Having said that, I can confirm that's the case in [our CI](https://github.com/Pepo48/keycloak/actions/runs/18727727304) and on my W11 machine - 127.0.0.1 resolves to a hostname on Windows, thus this PR helped to stabilize the failing test. So when it comes to IPv4 environments I would be pretty much certain.

I know that your question was specifically regarding the IPv4, but when it comes to IPv6, that's where I am less certain. 
Because, at least during my tries to mimic IPv6 scenario, `::1`  could be resolved to `0:0:0:0:0:0:0:1`. On the OS level/.NET stack it wasn't so obvious - it returned machine's hostname. But in terms of Java/JVM I noticed that `hostname = "0:0:0:0:0:0:0:1"`and `address.getHostAddress() = "0:0:0:0:0:0:0:1"` - both return literal strings that are equal, which means a null would be returned, even though it’s a legitimate loopback. So there seems to be an inconsistency between OS and JVM.

I am not sure if our goal is to tackle it in this PR, but do you think that a simple reorder would be enough to completely cover it? Something like:
```
if (address.isLoopbackAddress()) {
    hostname = "localhost";
} else if (hostname.equals(address.getHostAddress())) {
    logger.debugf("The hostAddress '%s' was not resolved to a hostname", hostAddress);
    return null;
}
```
Or do we keep going in circles with this? 
Thanks for your time!

Pepo48: Here are some logs I gathered during testing:
```
PS C:\Users\pepo4> @('127.0.0.1','::1','203.0.113.1') | ForEach-Object {
    $ip = $_
    try {
        $entry = [System.Net.Dns]::GetHostEntry([System.Net.IPAddress]::Parse($ip))
        [PSCustomObject]@{
            Query       = $ip
            HostName    = $entry.HostName
        }
    } catch {
        [PSCustomObject]@{
            Query       = $ip
            HostName    = "(error: $($_.Exception.Message))"
            AddressList = ''
        }
    }
} | Format-Table -AutoSize

Query       HostName
-----       --------
127.0.0.1   kubernetes.docker.internal        
::1         Lenovo15ARH7
203.0.113.1 (error: Exception calling "GetHostEntry" with "1" argument(s): "No such host is known")                     

PS C:\Users\pepo4> @'
import java.net.*;

String[] ips = { "127.0.0.1", "::1", "203.0.113.1" };

for (String ip : ips) {
    try {
        InetAddress address = InetAddress.getByName(ip);
        System.out.println("Query: " + ip
            + "  HostName: " + address.getHostName()
            + "  HostAddress: " + address.getHostAddress()
            + "  isLoopbackAddress(): " + address.isLoopbackAddress());
    } catch (Exception e) {
        System.out.println("Query: " + ip + "  Error: " + e.getMessage());
    }
}
/exit
'@ | jshell

|  Welcome to JShell -- Version 21.0.6
|  For an introduction type: /help intro
jshell : 
At line:18 char:6
+ '@ | jshell
+      ~~~~~~
    + CategoryInfo          : NotSpecified: (:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
jshell> 
jshell> 
ips ==> String[3] { "127.0.0.1", "::1", "203.0.113.1" }
jshell> String[] ips = { "127.0.0.1", "::1", "203.0.113.1" }
jshell> 
jshell> for (String ip : ips) {
   ...>         try {
   ...>                 InetAddress address = InetAddress.getByName(ip);
   ...>                 System.out.println("Query: " + ip
   ...>                     + "  HostName: " + address.getHostName()
   ...>                     + "  HostAddress: " + address.getHostAddress()
   ...>                     + "  isLoopbackAddress(): " + address.isLoopbackAddress());
   ...>             }} catch (Exception e) {
   ...>                 System.out.println("Query: " + ip + "  Error: " + e.getMessage());
   ...>             }}
Query: 127.0.0.1  HostName: kubernetes.docker.internal  HostAddress: 127.0.0.1  isLoopbackAddress(): true
Query: ::1  HostName: 0:0:0:0:0:0:0:1  HostAddress: 0:0:0:0:0:0:0:1  isLoopbackAddress(): true
Query: 203.0.113.1  HostName: 203.0.113.1  HostAddress: 203.0.113.1  isLoopbackAddress(): false
   ...>     }
|  Goodbye
```

shawkins: > I am not sure if our goal is to tackle it in this PR, but do you think that a simple reorder would be enough to completely cover it? Something like:

Yes, that is exactly what I would expect and it matches the new doc note.

shawkins (approved):
LGTM, thank you @Pepo48

Pepo48:
@vmuzikar do you want to give it another look as well?

vmuzikar (approved):
Changes LGTM.

(I assume we actually tried to run the test on Win manually this time. :D)

PR #43638, opened 2025-10-22
DPoP replay check should take clockSkew into account

### LINKED ISSUE #43505: DPoP proof replay check doesn't consider clock skew
### Before reporting an issue

### Area

oidc

### Describe the bug

DPoP proofs are checked for `DPoPIsActiveCheck` and `DPoPReplayCheck` conditions in `org.keycloak.services.util.DPoPUtil`.

The first check takes into account both lifetime (10 seconds) and clockSkew (15 seconds), as documented in [Allow a DPoP Proof by considering a clock skew for smooth interoperability](https://github.com/keycloak/keycloak/issues/34894)

The problem is that second check has not been updated accordingly to take clockSkew into account, so there is a 15 seconds window that causes `hashString` to be removed from the `singleUseCache` **because of resulting negative time** in `putIfAbsent`
```java
        public boolean test(DPoP t) throws DPoPVerificationException {
            SingleUseObjectProvider singleUseCache = session.singleUseObjects();
            byte[] hash = HashUtils.hash("SHA1", (t.getId() + "\n" + t.getHttpUri()).getBytes());
            String hashString = Hex.encodeHexString(hash);
            if (!singleUseCache.putIfAbsent(hashString, (int)(t.getIat() + lifetime - Time.currentTime()))) {
                throw new DPoPVerificationException(t, "DPoP proof has already been used");
            }
            return true;
        }
```

### Version

26.4.0

### Regression

### Expected behavior

A 'replayed' DPoP proof must be checked for not being already used for additional clockSkew time

### Actual behavior

Submitting the same DPoP proof multiple times using the appropriate timing can lead to original proof being removed from cache (negative time computed), so attacker can use stolen proof at least one more time to replay.

### How to Reproduce?

I'm using oauth2c to simulate OAuth2 flow with pkce and dpop in the same Keycloak dev machine (resulting clock skew is 0 seconds in this test)
```bash
export AUTHORIZATION_SERVER=my-keycloak-dev-machine
```

```bash
# Simulate with oauth2c the Auth Code Flow and capture Refresh Token from Keycloak response

export REFRESH_TOKEN=`oauth2c "$AUTHORIZATION_SERVER/.well-known/openid-configuration" \
  --client-id pkce-dpop-client \
  --response-types code \
  --response-mode query \
  --grant-type authorization_code \
  --auth-method none \
  --scopes openid,email,offline_access \
  --signing-key jwks.json \
  --pkce \
  --dpop | jq -r .refresh_token`
```

```bash
# Simulate with oauth2c a refresh token request using captured token from previous call, then capture DPoP proof from Keycloak response

export DPOP=`script -q -c 'oauth2c "$AUTHORIZATION_SERVER/.well-known/openid-configuration" \
  --client-id pkce-dpop-client \
  --grant-type refresh_token \
  --auth-method none \
  --refresh-token $REFRESH_TOKEN \
  --signing-key jwks.json \
  --dpop' | sed 's/\x1B\[[0-9;]\{1,\}[A-Za-z]//g' | sed -n 's/.*Dpop: \(.*\)/\1/p'`
```

```bash
# Manually simulate refresh token request using the same refresh token and the same DPoP proof captured from previus call

watch -n 1 'curl -v -k --request POST "$AUTHORIZATION_SERVER/$TOKEN_METHOD_ENDPOINT" \
  --header "Content-Type: application/x-www-form-urlencoded" \
  --header "DPoP:$DPOP" \
  --data-urlencode "grant_type=refresh_token" \
  --data-urlencode "client_id=pkce-dpop-client" \
  --data-urlencode "refresh_token=$REFRESH_TOKEN"'
```

Watch ouput of curl command, first responses are 'DPoP proof has already been used' which is ok, but after 10 seconds the DPoP exists from the cache and attacker can issue a new refresh token

### Anything else?

Fix should be quite simple, just add (at least) `clockSkew` in time computation
```java
            if (!singleUseCache.putIfAbsent(hashString, (int)(t.getIat() + lifetime + clockSkew - Time.currentTime()))) {
                throw new DPoPVerificationException(t, "DPoP proof has already been used");
            }
```

rmartinc:
Closes #43505

The PR just adds the `clockSkew` for the time stored in the single-use cache. If not, the token can be re-used, because its lifespan with the clockSkew can be longer than the time stored in the single-use cache. Test added.

### DIFF
--- a/services/src/main/java/org/keycloak/services/util/DPoPUtil.java
+++ b/services/src/main/java/org/keycloak/services/util/DPoPUtil.java
@@ -210,7 +210,7 @@ public class DPoPUtil {
                 DPoPClaimsCheck.INSTANCE,
                 new DPoPHTTPCheck(uri, method),
                 new DPoPIsActiveCheck(session, lifetime, clockSkew),
-                new DPoPReplayCheck(session, lifetime));
+                new DPoPReplayCheck(session, lifetime + clockSkew));
 
         if (accessToken != null) {
             verifier.withChecks(new DPoPAccessTokenHashCheck(accessToken));
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/DPoPTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/DPoPTest.java
@@ -260,6 +260,36 @@ public class DPoPTest extends AbstractTestRealmKeycloakTest {
     }
 
     @Test
+    public void testDPoPByPublicClientClockSkew() throws Exception {
+        getTestingClient().testing().setTestingInfinispanTimeService();
+        try {
+            sendAuthorizationRequestWithDPoPJkt(null);
+
+            // get a DPoP proof 10 seconds in the future
+            String dpopProofEcEncoded = generateSignedDPoPProof(UUID.randomUUID().toString(), HttpMethod.POST, oauth.getEndpoints().getToken(),
+                    (long) (Time.currentTime() + 10), Algorithm.ES256, jwsEcHeader, ecKeyPair.getPrivate(), null);
+
+            AccessTokenResponse response = successTokenProceduresWithDPoP(dpopProofEcEncoded, jktEc, true, true, false);
+
+            setTimeOffset(25); // 25 <= 10+10+15, proof not expired because clockSkew, detected by replay check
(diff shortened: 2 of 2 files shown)


graziang (approved):
@rmartinc thanks!

mposolda (approved):
@rmartinc @graziang Thanks for the PR and review! I hope to merge once tests are successful (Will re-run failing tests as the test failures probably not related to this PR).

PR #43648, opened 2025-10-22
Ability to define workflows with YAML

### LINKED ISSUE #42687: [RLM] - Ability to define workflows with YAML
### Description

When creating a workflow manually it's cumbersome to use JSON, YAML is a much better format for humans to edit.

We may need to be able to convert to/from json and yaml. For example I can create a workflow using YAML, and it's stored as JSON, then I can retrieve the workflow either as YAML or as JSON.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

vramik:
Closes #42687

This PR adds an ability define workflows in YAML format. It add support for accepting application/yaml content type for 

-  `WorkflowsResource.create`
-  `WorkflowsResource.createAll`

endpoints. The support for admin UI remains to be added as follow-up, as well as potential ability to produce YAML response.

It can be tested by following:

```bash
curl -X POST \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/yaml" \
  -d 'name: workflow' \
  http://localhost:8080/admin/realms/master/workflows

```

or

```bash
curl -X POST \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/yaml"  \
  -d 'name: workflow
      uses: event-based-workflow
      enabled: true                                           
      on:                         
        - USER_LOGIN
      if:                    
        - uses: has-identity-provider-link
          with:
            alias: google
      steps:
        - uses: set-user-required-action
          with:
            action: MY_CUSTOM_ACTION' \
  http://localhost:8080/admin/realms/master/workflows
```
for `create` endpoint.

```bash
curl -X POST \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/yaml" \
  -d '- name: workflow_alpha
      - name: workflow_beta' \
  http://localhost:8080/admin/realms/master/workflows/set
```
for `createAll` endpoint.

NOTE: Without https://github.com/keycloak/keycloak/pull/43560 the requests above fails with NPE.

### DIFF
--- a/tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/WorkflowManagementTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/WorkflowManagementTest.java
@@ -34,6 +34,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.client.WebTarget;
 import jakarta.ws.rs.core.Response;
 import org.hamcrest.Matchers;
 import jakarta.mail.MessagingException;
@@ -42,6 +45,8 @@ import java.io.IOException;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.keycloak.admin.client.Keycloak;
+import org.keycloak.admin.client.resource.BearerAuthFilter;
 import org.keycloak.admin.client.resource.WorkflowsResource;
 import org.keycloak.broker.oidc.KeycloakOIDCIdentityProviderFactory;
 import org.keycloak.common.util.Time;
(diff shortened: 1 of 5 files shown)


integration/admin-client/src/main/java/org/keycloak/admin/client/resource/WorkflowsResource.java
```diff
@@ -30,6 +30,11 @@ public interface WorkflowsResource {
     @Consumes(MediaType.APPLICATION_JSON)
     Response create(WorkflowSetRepresentation representation);
 
+    @Path("set")
+    @POST
+    @Consumes("application/yaml")
+    Response createYaml(WorkflowSetRepresentation representation);
```
sguilhen: is this needed?

vramik: no, I'll be removing it in the next update, thanks

sguilhen (approved):

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/WorkflowManagementTest.java
```diff
@@ -861,6 +876,32 @@ public void testCompleteUserLifecycleWithMultipleNotifications() {
...
+        WebTarget target = httpClient.target(keycloakUrls.getBaseUrl().toString())
+                .path("admin")
+                .path("realms")
+                .path(managedRealm.getName())
+                .path("workflows")
+                .path("set")
+                .register(new BearerAuthFilter(adminClient.tokenManager()));
+
+        Response response = target.request().post(Entity.entity(JsonSerialization.writeValueAsString(expectedWorkflows), "application/yaml"));
+        response.close();
```
pedroigor, line 902: I missed adding assertions to check the expected output, but I see you are doing so as a follow-up.

pedroigor (approved):

PR #43657, opened 2025-10-22
Document infinispan config logging

### LINKED ISSUE #43655: Document debug log settings required to show applied Infinispan configuration
### Description

Document debug log settings required to show applied Infinispan configuration

ryanemerson:
Closes #43655

@ahus1 @pruivo It seems we did already document the logging, but it's positioning in the guide wasn't that obvious. I have made it so we explain the logging in two places now and have updated the log statement to only enable debug logging for the required class, not the package.


ahus1 (approved):
Thank you for this change!

PR #43676, opened 2025-10-23
Information about features in the distribution

mabartos:
WiP

- Closes #43671

Execute:
```
./kc.sh help features
```

Current state:
image

image

https://github.com/user-attachments/assets/4094dd0e-60c4-47e8-94d2-0fef0e355c06

And then, I'd like to also get the info about enabled/disabled features from the persisted config source + maybe add some information about for what command it should be used?

cc: @keycloak/cloud-native

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Help.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Help.java
@@ -1,0 +1,10 @@
+package org.keycloak.quarkus.runtime.cli.command;
+
+import picocli.CommandLine;
+
+@CommandLine.Command(name = Help.NAME, header = Help.HEADER, description = "%n" + Help.HEADER,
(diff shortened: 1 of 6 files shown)


shawkins:
Thank you @mabartos for looking into this, it looks promising.

mabartos:
When executing:
```shell
./kc.sh build --features=workflows,fips --features-disabled=kerberos
```

and then:
```shell
./kc.sh help features
```

FIPS is part of enabled features with showing the type.
Screenshot From 2025-10-24 12-01-00

Worflows are part of enabled features:
Screenshot From 2025-10-24 12-00-54

Kerberos is part of disabled features with the Default type:
Screenshot From 2025-10-24 11-57-53

vmuzikar (changes requested):
Please wait with this until we have a broader plan for enhancements around Features.

shawkins:
Since this is effectively on hold, closing to clear it from the dashboard.

Issue #43573, opened 2025-10-17
Fine-grained policies for authorization grants


mposolda:
### Description

This is follow-up milestone to #43565 .

The aim for this milestone is to provide fine-grained specification of policies for authorization grants for the combination of particular client `foo` and particular identity provider `bar` .

### Goals

- Ability to further clarify what assertion claims are allowed for which client and identity provider combination (for example restriction for `azp`, `client_id`, `scope`, other custom claims like `hd`)
- Support for downscoping

### Non-Goals

- Documentation (There will be other milestone dedicated to it)

### Notes

This might be implemented with client policies, which are natural choice to intercept to the grant request and do some further checks on it.

Client policies have lots of flexibility and have ability to specify that for example:
- All assertions issued by identity provider `bar` are subject to client policy
- All assertions issued by identity provider `bar` when requested by client `foo` are subject to client policy
- All assertions issued by any identity provider to client `foo` are subject to client policy

This may allow to configure policies on the central place without a need to repeat same/similar on multiple places for every identity-provider or client.

Client policies already provide some conditions and executors, which might be helpful to achieve this task. 

Some things, which we may consider to implement as part of this milestone are those (Note that those are just some ideas, but in reality, we can handle it differently than proposed below ) :

### Condition for identity provider

The condition will be applied if specified identity provider is subject of the request. The condition may not be specific only to RFC-7523 authorization grant requests, but can be used by other use-cases/grants where the request is doing something related to the specific identity provider.

Condition might be possibly available just if the authorization-grant feature is enabled

#### Configuration options

- **Identity providers** - List of identity providers to choose from
- **Negate** - If ON, then policy is applied if the identity-provider from authorization grant is NOT in the list of configured providers provided by **Identity providers** option

### Executor for checking further claims

The client policy executor can be used to check claims on the assertion JWT and handle use-cases like:

- Check presence (or not presence) of specified claim
- Check the exact value of specified claim

The executor should be capable of handling use-cases like for instance:

- Claim `iat` must be present (to avoid cases when authorization grant assertion was created 2 months ago for instance)
- Claim `typ` must have value `authz-assertion` (The RFC-7523 does not have any definition what `typ` should look like, but validation on `typ` is natural so that just dedicated assertion-grant tokens would be used instead of for example regular access tokens or ID tokens)
- Claim `azp` must have value `foo` (or must NOT have value `foo`)
- Claim `hd` must have value `urn:my:domain`

[...]

Issue #43568, opened 2025-10-17
Configuration for OIDC identity provider to be able to act as trust-relationship provider for Authorization grants


mposolda:
### Description

Those are new configuration options added for identity providers. Initially maybe only for OIDC Identity provider (and Keycloak OIDC Identity provider, but that might be implicit)

We may initially consider config options like:

- **Support RFC-7523 authorization grant** - Boolean switch to indicate if identity provider is capable of acting as trust-relationship provider, which is able to verify RFC-7523 assertions.

- **Single use authorization grants** - Boolean switch to indicate if assertions should be single-use or not. Requires `jti` claim (which is optional in RFC 7523). ON by default

Maybe other switches will be introduced later once we see the need of them (EG. Maybe Clock Skew)

Note that identity provider may act only as trust-relationship provider, but not be used at all for login of users. We may need follow-up tasks to make sure that it is possible to create Identity provider, which is capable only to do "RFC-7523 assertion validation", but not login.

Task is follow-up of #43444

PR #43680, opened 2025-10-23
Improve JWT Assertion Validation using client validators

### LINKED ISSUE #43642: Improve JWT Assertion Validation
### Description

PR #43444 introduces the `JWTAuthorizationGrantValidationContext` class to perform initial validation of the provided JWT assertion with the authorization grant. Some improvements can be made to the validation process:

* **Reduce code duplication** — in particular, avoid duplicating logic that validates if the token is active. It might be good to reuse `AbstractJWTClientValidator`.
* **Improve handling of maximum exp time and clock skew**, which are currently hardcoded with default values.
* **Add replay check** for assertions by maintaining a set of used `jti` values to prevent JWT replay attacks. 
* **Validate other claims** (`iat`, `nbf`) when present, even though they are not mandatory for the RFC.

rmartinc:
Closes #43642

The PR creates a new intermediate class `AbstractJWTValidator` with part of the methods that previously were in `AbstractJWTClientValidator`. Class `JWTAuthorizationGrantValidationContext` in `server-spi-private` is now an interface (very limited now, but we can add mode methods if needed) and the real code has been moved to `JWTAuthorizationJWTValidator` in `services` which extends the `AbstractJWTValidator` to inherit the old methods. This way the code is shared. The signature validation is performed in the IdP side, and we can move more parts to the IdP if we want. So the current code is something in the middle. The token is validated using the client validator, but the signature is checked by the IdP. Draft for the moment. Questions:

* Do you see this OK? Or do you prefer move signature to the client validator part? Or vice-versa, move token validations to the IdP side? It's a bit repeated in both sides.
* The `nbf` claim is not checked now, I have just detected it. The client validator was not doing it before, so I will add later.
* I have added the options we have currently in the IdP side (clockSkew and algorithm), but there are other options that we can add: token lifespan (default is 5m=300s) and reusePermitted (default to false). Do we want them or is it OK using fixed values for now?

I need to polish this a bit but the I want to hear your comments about the three questions before. :smile:

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
@@ -1069,9 +1069,13 @@ public class OIDCIdentityProvider extends AbstractOAuth2IdentityProvider<OIDCIde
     }
 
     @Override
-    public BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizationGrantValidationContext context) {
+    public BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizationGrantValidationContext context) throws IdentityBrokerException {
 
-        //TODO: proper assertion validation
+        // verify signature
+        if (!verify(context.getJws())) {
+            throw new IdentityBrokerException("Invalid signature");
+        }
+
         BrokeredIdentityContext user = new BrokeredIdentityContext(context.getJWT().getSubject(), getConfig());
         user.setUsername(context.getJWT().getSubject());
         user.setIdp(this);
@@ -1079,4 +1083,8 @@ public class OIDCIdentityProvider extends AbstractOAuth2IdentityProvider<OIDCIde
 
     }
 
+    @Override
+    public int getAllowedClockSkew() {
+        return getConfig().getAllowedClockSkew();
+    }
 }
--- a/server-spi-private/src/main/java/org/keycloak/broker/provider/JWTAuthorizationGrantProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/broker/provider/JWTAuthorizationGrantProvider.java
@@ -18,5 +18,8 @@ package org.keycloak.broker.provider;
 import org.keycloak.protocol.oidc.JWTAuthorizationGrantValidationContext;
 
 public interface JWTAuthorizationGrantProvider {
-    BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizationGrantValidationContext assertion);
+
+    BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizationGrantValidationContext assertion) throws IdentityBrokerException;
+
+    int getAllowedClockSkew();
 }
--- a/server-spi-private/src/main/java/org/keycloak/protocol/oidc/JWTAuthorizationGrantValidationContext.java
+++ b/server-spi-private/src/main/java/org/keycloak/protocol/oidc/JWTAuthorizationGrantValidationContext.java
@@ -1,151 +1,22 @@
 package org.keycloak.protocol.oidc;
 
-import org.keycloak.OAuth2Constants;
-import org.keycloak.common.util.Time;
 import org.keycloak.jose.jws.JWSInput;
-import org.keycloak.models.ClientModel;
-import org.keycloak.representations.IDToken;
 import org.keycloak.representations.JsonWebToken;
 
-import java.util.Collections;
-import java.util.List;
 
-public class JWTAuthorizationGrantValidationContext {
+public interface JWTAuthorizationGrantValidationContext {
 
-    private final String assertion;
+    String getAssertion();
 
-    private final ClientModel client;
+    JsonWebToken getJWT();
 
-    private JsonWebToken jwt;
+    JWSInput getJws();
 
-    private final String expectedAudience;
-
-    private JWSInput jws;
-
-    private final long currentTime;
-
-    public JWTAuthorizationGrantValidationContext(String assertion, ClientModel client, String expectedAudience) {
-        this.assertion = assertion;
-        this.client = client;
-        this.expectedAudience = expectedAudience;
-        this.currentTime = Time.currentTimeMillis();
+    default String getIssuer() {
+        return getJWT().getIssuer();
     }
 
-    public void validateJWTFormat() {
-        try {
-            this.jws = new JWSInput(assertion);
-            this.jwt = jws.readJsonContent(JsonWebToken.class);
-        }
-        catch (Exception e) {
-            failure("The provided assertion is not a valid JWT");
-        }
-    }
-
-    public void validateAssertionParameters() {
-        if (assertion == null) {
-            failure("Missing parameter:" + OAuth2Constants.ASSERTION);
-        }
-    }
-
-    public void validateClient() {
-        if (client.isPublicClient()) {
-            failure("Public client not allowed to use authorization grant");
-        }
-
-        String val = client.getAttribute(OIDCConfigAttributes.JWT_AUTHORIZATION_GRANT_ENABLED);
-        if (!Boolean.parseBoolean(val)) {
-            throw new RuntimeException("JWT Authorization Grant is not supported for the requested client");
-        }
-
-    }
-
-    public void validateTokenActive() {
-        JsonWebToken token = getJWT();
-        int allowedClockSkew = getAllowedClockSkew();
-        int maxExp = getMaximumExpirationTime();
-        long lifespan;
-
-        if (token.getExp() == null) {
-            failure("Token exp claim is required");
-        }
-
-        if (!token.isActive(allowedClockSkew)) {
-            failure("Token is not active");
-        }
-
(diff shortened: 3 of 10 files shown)


rmartinc:
I have changed some class names to be more appropriate. And the `nbf` is already managed in the `isActive` of the `JsonWebToken`. So point 2 is already done, just the first and the third one remain.

graziang:
@rmartinc  Thanks! That’s pretty much how I imagined it, and I’m ok with having the signature validation on the IdP side.
I was wondering if there’s a way to avoid using `ClientAssertionState` in `JWTAuthorizationGrantValidator`, since `ClientAssertionState` was introduced to store a state to be propagated between authenticators and is not needed for the authorization grant. Maybe it could be moved to `AbstractJWTClientValidator` instead of `AbstractBaseJWTValidator`, or handled in some other way? Not sure if that’s possible.

rmartinc:
I think that the `ClientAssertionState` does not make any harm. It just contains the client requesting the assertion and the assertion info (jws, token, assertion and assertionType). So all the properties make sense for the JWT Authorization Grant. It's just another client autentication using an assertion. If we remove it, we need similar methods in the base class (getJWS, getToken,...). So I almost prefer to keep the state but no strong opinion here. @mposolda WDYT?

rmartinc:
Rebased with the last changes added by @graziang. :smile:

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
```diff
@@ -1079,4 +1084,13 @@ public BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizat
...
     }
 
+    @Override
+    public int getAllowedClockSkew() {
+        return getConfig().getAllowedClockSkew();
+    }
+
+    @Override
+    public String getClientAssertionSigningAlg() {
+        return getConfig().getClientAssertionSigningAlg();
```
mposolda: This effectively means that the IDP option `Client assertion signature algorithm` is used for both sending JWT issued by Keycloak for client authentication as well as "expected" algorithm for verification of assertions signed by the client for jwt-bearer grant.

Not 100% sure if this is good thing? Maybe it is OK for most of the cases and it is good to avoid introducing too much options. However for clients, there are different options (EG. ID Token Signature algorithm, User Info Signature algorithm), which is also prescribed in the OIDC specification.

I am like 50/50 whether to use same option or different. If we keep same option, we at least need to update the tooltip IMO.

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
```diff
@@ -1069,7 +1069,12 @@ public boolean verifyClientAssertion(ClientAuthenticationFlowContext context) th
     }
 
     @Override
-    public BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizationGrantValidationContext context) {
+    public BrokeredIdentityContext validateAuthorizationGrantAssertion(JWTAuthorizationGrantValidationContext context) throws IdentityBrokerException {
+
+        // verify signature
+        if (!verify(context.getJws())) {
```
mposolda, line 1075: It seems you addressed also #43644 with this PR :-) I've assigned that issue to yourself due to that.

It may be also good to remove TODO from line 1079 as I am not sure if there is something more to validate, which we are not already validating? Besides some complex stuff, which we may do with client policies (Follow-up issue for that is https://github.com/keycloak/keycloak/issues/43573 ) .

mposolda (commented):
@rmartinc Cool, Thanks!

I have few inline comments (Some of them are more a questions from discussion. Feedback welcome from you and @graziang as well).

Some answers to your questions:


>    Do you see this OK? Or do you prefer move signature to the client validator part? Or vice-versa, move token validations to the IdP side? It's a bit repeated in both sides.

I see this OK as it is now in your PR.

>    The nbf claim is not checked now, I have just detected it. The client validator was not doing it before, so I will add later.

In `AbstractBaseJWTValidator.validateTokenActive` there is a call to `token.isActive` , which checks also "not before" . Is it sufficient or do we need something more?

>    I have added the options we have currently in the IdP side (clockSkew and algorithm), but there are other options that we can add: token lifespan (default is 5m=300s) and reusePermitted (default to false). Do we want them or is it OK using fixed values for now?

IMO ok to hardcode "reusePermitted" to false for now.  There is the other PR https://github.com/keycloak/keycloak/pull/43841/files , which introduces this including proper config.

For the "token lifespan", I am not sure which token you mean? Lifespan of the assertion is handled by the attributes of JWT itself. Lifespan of the issued token is the same for "jwt-bearer" grant like for any other grants AFAIK.

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
rmartinc: Yes, this was not even added by @graziang, I just added because it was in the base class and makes some sense. We can remove it for the moment or change the tooltip and doc to mention that this is the algorithm to sign any client or authorization grant. Whatever you prefer is OK to me.

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
rmartinc: Yes, I did it because using the IdP it was already done. :smile: I will remove the TODO if we accept this way of validating the signature.

rmartinc:
@mposolda The lifespan is the 300 that is hardcoded [here](https://github.com/rmartinc/keycloak/blob/issue-43642/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java#L89C1-L89C58). The `validateTokenActive` method uses that time in case the expiration is too long or when the `iat` is not present to avoid very big times. It's harcoded to 300 now. In clients you can modify that default when you configure JWT to authenticate.

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
rmartinc: I'm going to remove this check for now. If we think we should validate the signing alg for the JWT authorization grant we can add a new prop or change the meaning of the current option. But we can move this forward for the moment.

rmartinc:
I have removed the TODO in the `OIDCIdentityProvider` class and removed the `getClientAssertionSigningAlg` in the `JWTAuthorizationGrantProvider`. With this the only remaining question is if we add more configuration methods in the JWT Authorization Grant:

* The signing algorithm for the JWT authorization grant (same prop or new).
* If reusing the same token is permitted.
* The maximum expiration time that is allowed for the grant.

We can manage those as a follow-up if we agree that any of those options are needed too.

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
rmartinc: Done!

services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
rmartinc: Removed the check and the method in the interface `JWTAuthorizationGrantProvider`.

mposolda: Sounds good for now. IMO we may need to do this, but likely OK as a follow-up. Thanks!

mposolda:
> @mposolda The lifespan is the 300 that is hardcoded [here](https://github.com/rmartinc/keycloak/blob/issue-43642/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java#L89C1-L89C58). The `validateTokenActive` method uses that time in case the expiration is too long or when the `iat` is not present to avoid very big times. It's harcoded to 300 now. In clients you can modify that default when you configure JWT to authenticate.

Ah, ok. Thanks for the clarification!

mposolda:
> I have removed the TODO in the `OIDCIdentityProvider` class and removed the `getClientAssertionSigningAlg` in the `JWTAuthorizationGrantProvider`. With this the only remaining question is if we add more configuration methods in the JWT Authorization Grant:
> 
>     * The signing algorithm for the JWT authorization grant (same prop or new).
> 
>     * If reusing the same token is permitted.
> 
>     * The maximum expiration time that is allowed for the grant.
> 
> 
> We can manage those as a follow-up if we agree that any of those options are needed too.

@rmartinc Thanks for the summary. I agree that those should be possible to configure. I've created a follow-up issue https://github.com/keycloak/keycloak/issues/43873 . The "Re-use" is already added in other subsequent PR by Giuseppe for #43568 . So IMO this PR is ready to go.

mposolda (approved):

mposolda:
@rmartinc Could you please rebase? I think that should help with the failing UI tests (not 100% sure, but I guess it might be related to https://github.com/keycloak/keycloak/commit/0f01444543cf5e755b13a62c233a5c9feef72888 )

graziang (approved):

rmartinc:
Rebased!

PR #43683, opened 2025-10-23
Avoid touching the database layer if no changes are necessary for a user

### LINKED ISSUE #43682: When syncing roles, the database layer can see deadlocks
### Description

When user data is synchronized for example with mappers on SAML login, roles are removed when they don't exist. As the DB code is using excessive locking, this is both slow and leads to deadlocks on concurrent logins.

### Value Proposition

Better performance, and no deadlocks that slow down Keycloak and annoy users. 

### Goals

All user methods that update the user should in the caching layer first check if there is actually a change. Most of the time there is no change when synchronizing users. 

### Non-Goals

Other entities

### Discussion

_No response_

### Notes

I'll create a PR

ahus1:
Closes #43682

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserAdapter.java
@@ -251,7 +251,7 @@ public class UserAdapter implements CachedUserModel {
 
     @Override
     public void addRequiredAction(RequiredAction action) {
-        if (updated == null && getRequiredActionsStream().anyMatch(s -> Objects.equals(s, action.name()))) {
+        if (action == null || updated == null && getCachedRequiredActions().contains(action.name())) {
             return;
         }
         getDelegateForUpdate();
@@ -260,7 +260,7 @@ public class UserAdapter implements CachedUserModel {
 
     @Override
     public void removeRequiredAction(RequiredAction action) {
-        if (updated == null && getRequiredActionsStream().noneMatch(s -> Objects.equals(s, action.name()))) {
+        if (action == null || updated == null && !getCachedRequiredActions().contains(action.name())) {
             return;
         }
         getDelegateForUpdate();
@@ -269,7 +269,7 @@ public class UserAdapter implements CachedUserModel {
 
     @Override
     public void addRequiredAction(String action) {
-        if (updated == null && getRequiredActionsStream().anyMatch(s -> Objects.equals(s, action))) {
+        if (updated == null && getCachedRequiredActions().contains(action)) {
             return;
         }
         getDelegateForUpdate();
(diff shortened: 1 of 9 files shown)


model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserAdapter.java
```diff
@@ -247,30 +251,38 @@ public Stream<String> getRequiredActionsStream() {
...
     public void removeRequiredAction(String action) {
-        if (getRequiredActionsStream().anyMatch(s -> Objects.equals(s, action))) {
-            getDelegateForUpdate();
-            updated.removeRequiredAction(action);
+        if (updated == null && getRequiredActionsStream().noneMatch(s -> Objects.equals(s, action))) {
+            return;
         }
+        getDelegateForUpdate();
+        updated.removeRequiredAction(action);
     }
```
pruivo, line 286: Avoid creating and iterating over streams
```java
    @Override
    public void addRequiredAction(RequiredAction action) {
        if (action == null || updated == null && getCachedRequiredActions().contains(action.name())) {
            return;
        }
        getDelegateForUpdate();
        updated.addRequiredAction(action);
    }

    @Override
    public void removeRequiredAction(RequiredAction action) {
        if (action == null || updated == null && !getCachedRequiredActions().contains(action.name())) {
            return;
        }
        getDelegateForUpdate();
        updated.removeRequiredAction(action);
    }

    @Override
    public void addRequiredAction(String action) {
        if (updated == null && getCachedRequiredActions().contains(action)) {
            return;
        }
        getDelegateForUpdate();
        updated.addRequiredAction(action);
    }

    @Override
    public void removeRequiredAction(String action) {
        if (updated == null && !getCachedRequiredActions().contains(action)) {
            return;
        }
        getDelegateForUpdate();
        updated.removeRequiredAction(action);
    }


    private Set<String> getCachedRequiredActions() {
        return cached.getRequiredActions(keycloakSession, modelSupplier);
    }
```

ahus1: @pruivo - please re-review. Thanks!

pruivo (dismissed):
LGTM, I'm not going to nitpick :)

ahus1:
Let's keep this draft until the backports also work in other branches. In #43697, this run makes me think...

pedroigor (dismissed):

pedroigor (dismissed):

ahus1:
Still some tests failing, needing more time to analyze this.

ahus1:
@pedroigor - can you please review this PR again? It should now be all green with the tests, and all necessary changes in. Thanks!

pruivo (approved):

pedroigor (approved):

Issue #23857, opened 2023-10-10
Use Quarkus extension for picocli


mabartos:
### Description

We should leverage the benefits of Quarkus extensions for our components as there can be made certain optimizations during a build time.  At this moment, we use directly `picocli` dependency and not the Quarkus picocli extension.

mabartos:
As we're migrating the Admin CLI to Picocli, we're already quite tied to it. Unfortunately, we use the direct dependency `info.picocli:picocli`, which does not provide tighter integration with Quarkus and additionally, we need to productize it on our own. 

We'd like to use the official Quarkus Picocli extension, but we are not allowed to do it due to the fact, that Quarkus does not productize it (AFAIK). 

@geoand What do we need to do in order to have a productized Quarkus Picocli extension by Quarkus?

geoand:
> @geoand What do we need to do in order to have a productized Quarkus Picocli extension by Quarkus?

This is a question for @tqvarnst and @maxandersen

mabartos:
@geoand Thanks for the message.

@tqvarnst @maxandersen What are the required steps to do in order to have Quarkus Picocli extension productized by Quarkus?

maxandersen:
@mabartos open Red Hat Build of Quarkus product jira (https://issues.redhat.com/browse/QUARKUS) and ping your PM.

mabartos:
@maxandersen Ok, thanks

stianst:
What is the motivation for this one? The description only says there could be some optimizations, but not if there really is any.

mabartos:
The motivation is to be in sync with Quarkus.

stianst:
That's a rather weak reasoning

PR #43686, opened 2025-10-23
fix: removing the fast start optimization

shawkins:
closes: #38790 #42960

I mentioned before that the difference should be sub-10 millisecond, however that was using an even more aggressive optimization and doing benchmarking that included warmup then a lot of runs. 

With the more concise set of changes in this pr and just comparing cold vm starts, the average time difference on my system seems to around +60 milliseconds.

The additional memory footprint appears to be negliable as the previous changes elimiated most options, and the changes here get rid of extraneous picocli state.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
@@ -170,6 +170,13 @@ public class Picocli {
                 addCommandOptions(cliArgs, cl);
             }
 
+            // ParseResult retain memory. Clear it, so it's not on the stack while the command runs
+            result = null;
+
+            // there's another ParseResult being created under the covers here.
+            // to reuse the previous result either means we need to duplicate the logic in the execute method
+            // or refactor the above logic so that it happens in the command logic
+            // We could also reduce the memory footprint of the ParseResult, but that looks a little hackish
             int exitCode = cmd.execute(argArray);
 
             exit(exitCode);
--- a/docs/guides/server/configuration.adoc
+++ b/docs/guides/server/configuration.adoc
@@ -264,7 +264,7 @@ The `--optimized` parameter tells {project_name} to assume a pre-built, already 
 You can enter all configuration options at startup; these options are the ones in <@links.server id="all-config"/> that are **not** marked with a tool icon.
 
 * If a build option is found at startup with a value that is equal to the value used when entering the `build`, that option gets silently ignored when you use the `--optimized` parameter.
-* If that option has a different value than the value used when a build was entered, a warning appears in the logs and the previously built value is used. For this value to take effect, run a new `build` before starting.
+* If that option has a different value than the value used when a build was entered, an error occurs. For this value to take effect, run a new `build` before starting.
 
 .Create an optimized build
 
@@ -288,8 +288,6 @@ https-certificate-file
 . Start the server with the optimized parameter
 +
     <@kc.start parameters="--optimized"/>
-
-You can achieve most optimizations to startup and runtime behavior by using the `build` command. Also, by using the `keycloak.conf` file as a configuration source, you avoid some steps at startup that would otherwise require command line parameters, such as initializing the CLI itself. As a result, the server starts up even faster.
 
 == Using system variables in the realm configuration
 
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakMain.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakMain.java
@@ -20,7 +20,6 @@ package org.keycloak.quarkus.runtime;
 import static org.keycloak.quarkus.runtime.Environment.getKeycloakModeFromProfile;
 import static org.keycloak.quarkus.runtime.Environment.isNonServerMode;
 import static org.keycloak.quarkus.runtime.Environment.isTestLaunchMode;
-import static org.keycloak.quarkus.runtime.cli.command.AbstractAutoBuildCommand.OPTIMIZED_BUILD_OPTION_LONG;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -45,7 +44,6 @@ import org.keycloak.quarkus.runtime.cli.Picocli;
 import org.keycloak.common.Version;
 import org.keycloak.quarkus.runtime.cli.command.AbstractNonServerCommand;
 import org.keycloak.quarkus.runtime.cli.command.DryRunMixin;
-import org.keycloak.quarkus.runtime.cli.command.Start;
 
 import io.quarkus.runtime.QuarkusApplication;
 import io.quarkus.runtime.annotations.QuarkusMain;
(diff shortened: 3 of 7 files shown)


shawkins:
I should add that the best I can do seems to be about +40 milliseconds - however that requires quite more hack-ish changes.

Until https://github.com/keycloak/keycloak/issues/23857 somehow supports creating the cli via augmentation, I don't think we're going to get much more improvement as the remaining time cost is mostly in the reflective creation of the command / subcommands.

mabartos:
> I should add that the best I can do seems to be about +40 milliseconds - however that requires quite more hack-ish changes.

I'd rather have +40ms than have many hack-ish changes for now. 

> Until https://github.com/keycloak/keycloak/issues/23857 somehow supports creating the cli via augmentation, I don't think we're going to get much more improvement as the remaining time cost is mostly in the reflective creation of the command / subcommands.

Yep, it'd be good to finally include the QUarkus picocli extension and then bring it up on the Quarkus side if it's possible to leverage the augmentation.

mabartos (dismissed):

shawkins:
@vmuzikar do you want to review as well?

vmuzikar:
I'd like to take a look today.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -170,6 +171,11 @@ public void parseAndRun(List<String> cliArgs) {
                 addCommandOptions(cliArgs, cl);
             }
 
+            result = null;
```
vmuzikar, line 174: Just a nitpick. This doesn't seem to be needed. But it's not a blocker.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -170,6 +171,11 @@ public void parseAndRun(List<String> cliArgs) {
...
+            if (Optional.ofNullable(currentCommand).map(AbstractCommand::getName).filter(Start.NAME::equals).isPresent()) {
```
vmuzikar: Couldn't we just test if `cuurentCommand instanceof Start`?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -170,6 +171,11 @@ public void parseAndRun(List<String> cliArgs) {
...
+                reduceMemoryForStart(cmd);
```
vmuzikar: What is the real impact of this optimization? I'd like to avoid hacky optimizations unless they have significant and real impact.

vmuzikar (changes requested):
Thanks for the changes. +1 for removing fast startup as it had very limited impact.

Added a couple of non-blocking comments.

However, what would be good to include in this PR is a docs change here:
https://github.com/keycloak/keycloak/blob/a34b14796f457414a2ab3ede59e1be9b04d7dcb0/docs/guides/server/configuration.adoc?plain=1#L292

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: It clears the memory held in the result object from the stack.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: In the debugger the retained memory with all of the commands was on the order of 400 kB.. Trimming to just the start seems to save over 300 kB.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: That would be fine as long as we don't ever refactor to have something like StartDev be a subclass of Start.

vmuzikar: Jus out of curiosity, why wouldn't `reduceMemoryForStart` work with start-dev?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
vmuzikar: Not sure if 300 kB of memory really matters in the context of whole Keycloak. :) We for sure have a lots of places where we could shave off a couple of 100s kB but it's not something we usually do.

But ok, let's keep it, it's relatively straightforward optimization.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: It could, but the current logic would then remove the StartDev subcommand.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: My goal was simply to get things as close as possible. If 300 kB isn't worth it, I'm fine with removing it.

vmuzikar: I'd say it's not, so my preference would be to just remove it for the simplicity and robustness sake.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
vmuzikar: Thinking about that, could you add some comment there so it doesn't get removed by accident in the future?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: Ok, removed and added comments explaining what could be done next.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: Ok, added.

vmuzikar:
Thanks for the update. Last thing remaining is the docs update.

shawkins:
> Thanks for the update. Last thing remaining is the docs update.

What doc update?

vmuzikar:
Sorry, I meant:

> However, what would be good to include in this PR is a docs change here:
> 
> https://github.com/keycloak/keycloak/blob/a34b14796f457414a2ab3ede59e1be9b04d7dcb0/docs/guides/server/configuration.adoc?plain=1#L292

shawkins:
> Sorry, I meant:
> 
> > However, what would be good to include in this PR is a docs change here:
> > https://github.com/keycloak/keycloak/blob/a34b14796f457414a2ab3ede59e1be9b04d7dcb0/docs/guides/server/configuration.adoc?plain=1#L292

Ah, I missed your earlier comment. Updated now.

vmuzikar (approved):
LGTM, thanks.

PR #43689, opened 2025-10-24
Manage the error when the admin console reads the certificate info

### LINKED ISSUE #43547: Unrecognized field "kty" (class org.keycloak.jose.jwk.JSONWebKeySet), not marked as ignorable (one known property: "keys"])
### Before reporting an issue

### Area

admin/ui

### Describe the bug

I have two keycloak instances running the same version, 26.2.4,.  The instances started I believe around 22.x and upgraded through the intermittent versions.

When I try to access `Clients>Client details` for an existing client I get an error and the keys tab does not display.  The server logs show `Invalid jwks representation!` and `Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "kty" (class org.keycloak.jose.jwk.JSONWebKeySet), not marked as ignorable (one known property: "keys"])` .

Image

[keycloak_keys_error.txt](https://github.com/user-attachments/files/22956182/keycloak_keys_error.txt)

If i try to create a new realm and new client, I can get to the keys tab.  But clicking generate keys gives me a `java.security.KeyStoreException: Cannot overwrite own certificate` as seen in keycloak_create_new_keys_error.txt

Image

[keycloak_create_new_keys_error.txt](https://github.com/user-attachments/files/22956187/keycloak_create_new_keys_error.txt)

### Version

26.2.4

### Regression

### Expected behavior

To display the keys tab without error and then allow me to generate keys

### Actual behavior

I get error accessing the keys tab  and am unable to generate keys.

### How to Reproduce?

# For existing Realm and client
- Goto existng Realm and client with JWKS keys 
- Click the Keys tab under Client Details
- Get error

# For new Realm 
- Create Realm
- Create Client
- Enable Client Authorization
- Click the keys tab
- Click generate keys and get and error

### Anything else?

I don't know when the behavior started, I want to say I didn't see it in 25.x.  But the last time i can recall interacting with this tab was probably in 24.x. If no one else is seeing it , i am guessing something happened during an update that i missed but am unsure how to check or resolve that.

rmartinc:
Closes #43547

After https://github.com/keycloak/keycloak/issues/33820 the certificate endpoint can return errors. The admin console does not expect that. A general error is displayed and the user cannot do anything to fix the situation. With this PR the error is managed, a popup error message is displayed saying the cert cannot be read, and the user can import/generate a new key/certificate normally.

Screenshot From 2025-10-24 08-21-54

Reaching this situation is rare. Because currently the import from the jwks file fails, so the console does not allow this scenario. So the bad json should come from a migration (previously it was a free textarea) or managed directly using `kcadm` or similar. The invalid key never worked, so just an UI error.

### DIFF
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -2173,6 +2173,7 @@ policyEnforcementModeHelp=The policy enforcement mode dictates how policies are 
 selectAUser=Select a user
 groupCreated=Group created
 generateError=Could not generate a new key pair and certificate {{error}}
+getKeyInfoError=Could not read the certificate information {{error}}
 testClusterSuccess=Successfully verified availability for\: {{successNodes}}
 whoWillAppearLinkTextRoles=Who will appear in this user list?
 attestationPreference.not\ specified=Not specified
--- a/js/apps/admin-ui/src/clients/keys/Keys.tsx
+++ b/js/apps/admin-ui/src/clients/keys/Keys.tsx
@@ -69,7 +69,14 @@ export const Keys = ({
   });
 
   useFetch(
-    () => adminClient.clients.getKeyInfo({ id: clientId, attr }),
+    async () => {
+      try {
+        return await adminClient.clients.getKeyInfo({ id: clientId, attr });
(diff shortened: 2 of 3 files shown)


jonkoops (approved):

mposolda (approved):
@rmartinc @jonkoops Thanks for the fix and review!

PR #43695, opened 2025-10-24
Avoid using UserCredentialManager from user storage extensions

### LINKED ISSUE #43694: Avoid using UserCredentialManager from user storage extensions
### Description

The target is to avoid using `UserCredentialManager` from the user-storage extensions, so that class can be moved in future release to `keycloak-server-spi-private` . This will help to make sure that `keycloak-model-storage` will not need dependency on `keycloak-server-spi-private` without breaking custom user-storage extensions. See https://github.com/keycloak/keycloak/issues/43626 for the details.

### Value Proposition

Give developers, which implement Keycloak user-storage extension the time to migrate their extension to avoid using `UserCredentialManager` .

### Goals

- Prepare Keycloak codebase to avoid having `keycloak-model-storage` to be dependency of `keycloak-server-spi-private` .

### Non-Goals

None

### Discussion

_No response_

### Notes

_No response_

mposolda:
closes #43694

There is an issue #43626 for make sure `keycloak-model-storage` does not have dependency on `keycloak-server-spi-private` . 

In the PR for that issue, I took the approach of moving `UserCredentialManager` to the `keycloak-server-spi-private`. But this means that people cannot use the `UserCredentialManager` from their user-storage providers anymore (due the fact that if they use it, they will need dependency on `keycloak-server-spi-private`). However this might be a backwards incompatible change as people may already use `UserCredentialManager` in their providers similarly like for example our quickstart is doing: https://github.com/keycloak/keycloak-quickstarts/blob/main/extension/user-storage-simple/src/main/java/org/keycloak/quickstart/readonly/PropertyFileUserStorageProvider.java#L88 .

So was thinking about doing this in multiple steps:

- In 26.5, make sure that constructor `new UserCredentialManager()` is not recommended for the use, so people have a chance to migrate to avoid using this class directly, but rather access it via the method on `KeycloakSession`
- In 27.0 move the `UserCredentialManager` to `keycloak-server-spi-private` and make sure that `keycloak-model-storage` can remove the dependency on `keycloak-server-spi-private` (will require few other minor things done in .

The alternative approach is to keep `UserCredentialManager` in `keycloak-model-storage`, but this does not seem right to me TBH as this class contains lots of logic, which does not need to be directly accessible to people in their applications. Among other things, it would also require moving few other classes (like `DatastoreProvider` for example) to `keycloak-server-spi` or `keycloak-model-storage` and it would mean that people will need to declare dependency on opentelemetry in their user-storage providers (as `UserCredentialManager` is using opentelemetry). But I can revisit if you think this would be better?

What do you think?

### DIFF
--- a/server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
+++ b/server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
@@ -222,13 +222,6 @@ public interface KeycloakSession extends AutoCloseable {
     UserProvider users();
 
     /**
-     * Default implementation of {@link SubjectCredentialManager} suitable for most of user providers
-     *
-     * @return user credential manager
-     */
-    SubjectCredentialManager getUserCredentialManager(UserModel user);
-
-    /**
      * Key manager
      *
       * @return
--- a/server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
+++ b/server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
@@ -222,6 +222,13 @@ public interface KeycloakSession extends AutoCloseable {
     UserProvider users();
 
     /**
+     * Default implementation of {@link SubjectCredentialManager} suitable for most of user providers
+     *
+     * @return user credential manager
+     */
+    SubjectCredentialManager getUserCredentialManager(UserModel user);
+
+    /**
      * Key manager
      *
       * @return
--- a/docs/documentation/server_development/topics/user-storage/stream-interfaces.adoc
+++ b/docs/documentation/server_development/topics/user-storage/stream-interfaces.adoc
@@ -21,7 +21,7 @@ See this list of the interfaces that offer a `Streams` sub-interface to develope
 | Package | Classes
 ifeval::[{project_community}==true]
 | `org.keycloak.credential` | `CredentialInputUpdater`(*), `UserCredentialStore`
-| `org.keycloak.models` | `GroupModel`, `RoleMapperModel`, `UserCredentialManager`, `UserModel`, `UserProvider`
+| `org.keycloak.models` | `GroupModel`, `RoleMapperModel`, `SubjectCredentialManager`, `UserModel`, `UserProvider`
 | `org.keycloak.models.cache` | `CachedUserModel`, `UserCache`
 | `org.keycloak.storage.federated` | All interfaces
 | `org.keycloak.storage.user` | `UserQueryProvider`(*)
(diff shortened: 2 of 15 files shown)


server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
```diff
@@ -221,6 +221,13 @@ default <T> T getAttributeOrDefault(String attribute, T defaultValue) {
      */
     UserProvider users();
 
+    /**
+     * Default implementation of {@link SubjectCredentialManager} suitable for most of user providers
+     *
+     * @return user credential manager
+     */
+    SubjectCredentialManager getUserCredentialManager(UserModel user);
```
pedroigor: Is it the best place to have this method? It seems a bit weird that we now have this:

```
session.users()
```

and:

```
session.getUserCredentialManager()
```

Instead of something like this:

```
session.users().getCredentialManager(user)
```

pedroigor (commented):
For 27, my proposal would be:

* Rename `org.keycloak.credential.UserCredentialManager` to `org.keycloak.credential.DefaultUserCredentialManager` (and make it private as you are proposing)
* Rename `org.keycloak.models.SubjectCredentialManager` to `org.keycloak.models.UserCredentialManager`

server-spi/src/main/java/org/keycloak/models/KeycloakSession.java
mposolda: That should work as well.

Was thinking about this and originally did not used this just because we have multiple implementations of `UserProvider` and the jpa/infinispan implementations are more towards to have mostly logic related to CRUD of particular storage (EG. `JpaUserProvider` has mostly methods related to CRUD of users to the DB). At the same time, introducing the method to `KeycloakSession` is also not great... So I've updated to use the approach you suggest.

mposolda:
> For 27, my proposal would be:
> 
>     * Rename `org.keycloak.credential.UserCredentialManager` to `org.keycloak.credential.DefaultUserCredentialManager` (and make it private as you are proposing)
> 
>     * Rename `org.keycloak.models.SubjectCredentialManager` to `org.keycloak.models.UserCredentialManager`

This works.

Considering this, I wonder that it is possible to introduce already class `org.keycloak.models.UserCredentialManager`? In this case, people have already the possibility to update using that class instead of having breaking changes later as they can use something like:
```
@Override
public org.keycloak.models.UserCredentialManager credentialManager() {
    return keycloakSession.users().getUserCredentialManager(this);
}
```
in their `UserModel` implementations. This means that if they update their code now, they will not need to update the code again once switching from 26.5 to 27. WDYT?

pedroigor (approved):

mposolda:
@pedroigor Thanks!

PR #43697, opened 2025-10-24
Avoid touching the database layer if no changes are necessary for a user

### LINKED ISSUE #43682: When syncing roles, the database layer can see deadlocks

ahus1:
Closes #43682


ahus1:
We no longer plan to backport this to 26.0

PR #43699, opened 2025-10-24
Role mapper should check if an update is needed for the role

### LINKED ISSUE #43698: Role Mapper is updating the user every time on login
### Before reporting an issue

### Area

identity-brokering

### Describe the bug

Whenever a user logs in, the mappers will update and remove the role every time, even though the user doesn't need updating. 

This is a smaller variant of #43682 which might be more appropriate to backport

### Version

26.4

### Regression

### Expected behavior

The user should not be updated in the database if their role settings match what is expected.

### Actual behavior

The role mapper attempts to delete the role. When doing this, it issues a select-for-update, and too many of them seem to lead to row lock contention and eventually deadlocks. 

In addition to that, the user is invalidated in the cache.

### How to Reproduce?

Look at the code and the logs

### Anything else?

I'll prepare a PR. There is already #43682 with a more general approach, but which might be more risky for backporting.

ahus1:
Closes #43698


martin-kanis (approved):
LGTM

pedroigor (approved):

PR #43702, opened 2025-10-24
Improve SimpleHttp API

### LINKED ISSUE #43701: Improve SimpleHttp API
### Description

- Expose configured HTTP method from SimpleHttpRequest
- Use same ObjectMapper configuration as in JsonSerialization used by the deprecated SimpleHTTP
- Allow to configure the ObjectMapper to use by SimpleHttp
- Expose ObjectMapper factory methods in JsonSerialization

### Value Proposition

The http method should be exposed to analyze SimpleHttpRequest.
The ObjectMapper mapper should be configurable to allow custom JSON generation.

### Goals

Make the ObjectMapper used by SimpleHttp configurable.
The ObjectMapper used by the new SimpleHttp should produce the same Json as the old deprecated SimpleHttp class.

### Non-Goals

-

### Discussion

_No response_

### Notes

_No response_

thomasdarimont:
- Expose configured HTTP method from SimpleHttpRequest - needed by extensions / libraries that wants to enrich the SimpleHTTP call
- Use same ObjectMapper configuration as in JsonSerialization used by the deprecated SimpleHTTP to produce consistent json output.
- Allow to configure the ObjectMapper to use by SimpleHttp - this enables custom JSON formats
- Expose ObjectMapper factory methods in JsonSerialization - allows consistent JSON serialization.

Fixes #43701

### DIFF
--- a/core/src/main/java/org/keycloak/util/JsonSerialization.java
+++ b/core/src/main/java/org/keycloak/util/JsonSerialization.java
@@ -38,17 +38,24 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
  * @version $Revision: 1 $
  */
 public class JsonSerialization {
-    public static final ObjectMapper mapper = new ObjectMapper();
-    public static final ObjectMapper prettyMapper = new ObjectMapper();
+    public static final ObjectMapper mapper = objectMapperWithDefaults();
+    public static final ObjectMapper prettyMapper = prettyObjectMapperWithDefaults();
 
-    static {
+    public static ObjectMapper objectMapperWithDefaults() {
+        ObjectMapper mapper = new ObjectMapper();
         mapper.registerModule(new Jdk8Module());
         mapper.registerModule(new JavaTimeModule());
         mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
         mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        return mapper;
+    }
+
+    public static ObjectMapper prettyObjectMapperWithDefaults() {
+        ObjectMapper prettyMapper = new ObjectMapper();
         prettyMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
         prettyMapper.enable(SerializationFeature.INDENT_OUTPUT);
         prettyMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        return prettyMapper;
     }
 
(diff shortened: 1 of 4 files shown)


thomasdarimont:
@stianst would you mind taking a look at this?

core/src/main/java/org/keycloak/util/JsonSerialization.java
```diff
@@ -38,17 +38,24 @@
  * @version $Revision: 1 $
  */
 public class JsonSerialization {
-    public static final ObjectMapper mapper = new ObjectMapper();
-    public static final ObjectMapper prettyMapper = new ObjectMapper();
+    public static final ObjectMapper mapper = objectMapperWithDefaults();
+    public static final ObjectMapper prettyMapper = prettyObjectMapperWithDefaults();
 
-    static {
+    public static ObjectMapper objectMapperWithDefaults() {
```
stianst, line 44: I'd prefix this with create, and the one below as well. To make it more explicit that it creates an ObjectMapper.

```suggestion
    public static ObjectMapper createObjectMapperWithDefaults() {
```

server-spi-private/src/main/java/org/keycloak/http/simple/SimpleHttpRequest.java
```diff
@@ -120,6 +120,10 @@ public Object getEntity() {
         return entity;
     }
 
+    public SimpleHttpMethod getMethod() {
```
stianst, line 123: Could you explain why this should be exposed? Can't really think why that's useful

stianst:
Closing as out of date. @thomasdarimont feel free to re-open if you still want this

server-spi-private/src/main/java/org/keycloak/http/simple/SimpleHttpRequest.java
thomasdarimont: Sorry I didn't see that comment. If you have generic framework code that prepares a SimpleHttpRequest we need to be able to inspect the configuration of the request.

thomasdarimont:
I rebased the branch and updated the PR with your suggestions. 

I think the `getMethod()` is helpful for generic code and doesn't hurt as the method is read-only and immutable.

I'm not allowed to reopen the PR for some reason.

thomasdarimont:
Seems that I have to create a new PR as the branch does not update anmyore.

thomasdarimont:
@stianst I still think this is helpfull, so I created a new #46171

PR #43727, opened 2025-10-27
Only add the none verifier when attestation conveyance preference is none

### LINKED ISSUE #43723: Only add the none verifier when attestation conveyance preference is none (or default)
### Before reporting an issue

### Area

authentication/webauthn

### Describe the bug

When attestation conveyance preference is  set to direct or indirect we are adding the `none` verifier (see [here](https://github.com/keycloak/keycloak/blob/26.4.2/services/src/main/java/org/keycloak/authentication/requiredactions/WebAuthnRegister.java#L319-L328)). This means that even we are requesting an attestation, the authenticator can return none and we are accepting it. Better if we simply does not include the `none` verifier when the attestation is configured.

### Version

24.0.2

### Regression

### Expected behavior

Only when attestation coneyance is none (or default, empty) the `none` verifier is added and accepted. 

### Actual behavior

The `none` verifier is always accepted even when `direct` attestation is configured.

### How to Reproduce?

1. Star webauthn registration with none allowed in the policy.
2. When the page is presented with 'none' (default), change to 'direct' the policy.
3. Finish the registration, that return no attestation.
4. This is allowed and should be rejected by keycloak.

### Anything else?

_No response_

rmartinc:
Closes #43723

The PR just assigns the none verifier if none is configured in the current policy (or default/empty). This way we avoid allowing none when the attestation conveyance is direct or indirect. A test is added that changes the policy when the registration page is already ready to be sent.

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/requiredactions/WebAuthnRegister.java
+++ b/services/src/main/java/org/keycloak/authentication/requiredactions/WebAuthnRegister.java
@@ -61,6 +61,7 @@ import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.utils.StringUtil;
 
 import com.webauthn4j.converter.util.ObjectConverter;
+import com.webauthn4j.data.AttestationConveyancePreference;
 import com.webauthn4j.data.attestation.authenticator.AttestedCredentialData;
 import com.webauthn4j.data.attestation.statement.AttestationStatement;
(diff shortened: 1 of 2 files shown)


rmartinc:
@tnorimat I have added you as reviewer. This is the PR for the WebAuthn issue we commented some days ago. Thanks!

graziang (approved):
@rmartinc thanks!

tnorimat (approved):
@rmartinc Thank you.

ahus1 (approved):

PR #43740, opened 2025-10-27
Deprecate `org.keycloak.common.util.Base64`

### LINKED ISSUE #43370: Deprecate org.keycloak.common.util.Base64
### Description

Introduced a long time ago when the JDK didn't have a Base64 encoder/decoder, `org.keycloak.common.util.Base64` and `org.keycloak.common.util.Base64Url` can be deprecated for removal, and their usage replaced with the JDK class. 

### Next steps

* Switch all usage to the JDK class
* Add deprecation annotation to the old class

Out of scope: A new performance test of Keycloak

tkyjovsk:
Closes #43370

### DIFF
--- a/services/src/main/java/org/keycloak/social/twitter/TwitterIdentityProvider.java
+++ b/services/src/main/java/org/keycloak/social/twitter/TwitterIdentityProvider.java
@@ -210,7 +210,7 @@ public class TwitterIdentityProvider extends AbstractIdentityProvider<OAuth2Iden
             try (VaultStringSecret vaultStringSecret = session.vault().getStringSecret(providerConfig.getClientSecret())) {
                 String twitterToken = authSession.getAuthNote(TWITTER_TOKEN);
                 RequestToken requestToken;
-                try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(Base64.decode(twitterToken)))) {
+                try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(java.util.Base64.getDecoder().decode(twitterToken)))) {
                     requestToken = (RequestToken) in.readObject();
                 }
 
--- a/services/src/main/java/org/keycloak/social/twitter/TwitterIdentityProvider.java
+++ b/services/src/main/java/org/keycloak/social/twitter/TwitterIdentityProvider.java
@@ -63,8 +63,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
-import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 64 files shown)


services/src/main/java/org/keycloak/social/twitter/TwitterIdentityProvider.java
```diff
@@ -90,14 +95,27 @@ public Object callback(RealmModel realm, AuthenticationCallback callback, EventB
...
+    private String base64EncodeObject(Serializable serializableObject) throws IOException {
+      try (ByteArrayOutputStream baos = new ByteArrayOutputStream(); 
+              ObjectOutputStream oos = new ObjectOutputStream(Base64.getEncoder().wrap(baos))) {
+          oos.writeObject(serializableObject);
+          oos.flush();
+          try {
+              return new String(baos.toByteArray(), "US-ASCII");
+          } catch (UnsupportedEncodingException uue) {
+              return new String(baos.toByteArray());
+          }
```
pruivo: Can be replaced with a single line:
```suggestion
          return baos.toString(StandardCharsets.US_ASCII);
```

services/src/main/java/org/keycloak/social/twitter/TwitterIdentityProvider.java
```diff
@@ -210,7 +228,7 @@ public Response authResponse(@QueryParam("state") String state,
             try (VaultStringSecret vaultStringSecret = session.vault().getStringSecret(providerConfig.getClientSecret())) {
                 String twitterToken = authSession.getAuthNote(TWITTER_TOKEN);
                 RequestToken requestToken;
-                try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(Base64.decode(twitterToken)))) {
+                try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(java.util.Base64.getDecoder().decode(twitterToken)))) {
```
pruivo:
```suggestion
                try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(Base64.getDecoder().decode(twitterToken)))) {
```

pruivo (dismissed):

ahus1:
@pruivo / @tkyjovsk - I reviewed the code and re-introduced some of the caught and wrapped exceptions, please review. 

I also tested the changes to Twitter, and found that the `flush()` needs to be exchanged with a `close()` as some bytes were missing. I think it makes sense, as base64 translates 2 bytes to 3 chars, so any dangling byte would only be written on close, but not at a flush where more bytes could follow. 

Let's see if the CI agrees, and also please let me know if you agree with the changes. Thanks!

pruivo (approved):

ahus1 (approved):

PR #43745, opened 2025-10-27
Check offline scope is still assigned when performing a refresh

### LINKED ISSUE #43734: Refresh token allowed for offline session even the related scope is removed
### Before reporting an issue

### Area

oidc

### Describe the bug

An offline session continues to be valid when the `offline_access` scope is removed from the client. The idea is that the refresh operation should return an error if the scope is not assigned to the client anymore. This can confuse administrators, that think removing the scope is enough and they do not remove the sessions associated. Better if we avod this and return an error, saying invalid scope or similar.

### Version

26.4.2

### Regression

### Expected behavior

The refresh token fails because of the scope issue.

### Actual behavior

The refresh token is valid and a new token is obtained.

### How to Reproduce?

1. Initiate an offline session with a code to token login using a client.
2. In the related client remove the `offline_access` scope.
3. Refresh to token to obtain a new access token.

### Anything else?

_No response_

rmartinc:
Closes #43734

PR to check if `offline_access` is still allowed when doing a refresh token. The PR just returns the error and maintains the session (it follows the same behavior that is used for all the other similar exceptions). If you think The PR should remove the session or similar just let me know. Test added.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
@@ -224,6 +224,10 @@ public class TokenManager {
             throw new OAuthErrorException(OAuthErrorException.INVALID_GRANT, "Stale token");
         }
 
+        if (userSession.isOffline() && !UserSessionUtil.isOfflineAccessGranted(session, clientSession)) {
+            throw new OAuthErrorException(OAuthErrorException.INVALID_GRANT, "Offline session invalid because offline access not granted anymore");
+        }
+
         // Case when offline token is migrated from previous version
         if (oldTokenScope == null && userSession.isOffline()) {
             logger.debugf("Migrating offline token of user '%s' for client '%s' of realm '%s'", user.getUsername(), client.getClientId(), realm.getName());
--- a/services/src/main/java/org/keycloak/protocol/oidc/AccessTokenIntrospectionProvider.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/AccessTokenIntrospectionProvider.java
@@ -205,6 +205,14 @@ public class AccessTokenIntrospectionProvider<T extends AccessToken> implements 
             return false;
         }
 
+        if (userSession.isOffline() && !UserSessionUtil.isOfflineAccessGranted(
+                session, userSession.getAuthenticatedClientSessionByClient(client.getId()))) {
+            logger.debugf("Offline session invalid because offline access not granted anymore");
+            eventBuilder.detail(Details.REASON, "Offline session invalid because offline access not granted anymore");
+            eventBuilder.error(Errors.SESSION_EXPIRED);
+            return false;
+        }
+
         if (!verifyTokenReuse()) {
             return false;
         }
--- a/services/src/main/java/org/keycloak/services/util/UserSessionUtil.java
+++ b/services/src/main/java/org/keycloak/services/util/UserSessionUtil.java
@@ -4,12 +4,14 @@ import java.util.Objects;
 import java.util.function.Consumer;
 
 import org.jboss.logging.Logger;
+import org.keycloak.OAuth2Constants;
 import org.keycloak.common.ClientConnection;
 import org.keycloak.common.Profile;
 import org.keycloak.common.constants.ServiceAccountConstants;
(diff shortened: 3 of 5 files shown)


services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
```diff
@@ -233,6 +233,12 @@ public TokenValidation validateToken(KeycloakSession session, UriInfo uriInfo, C
 
         ClientSessionContext clientSessionCtx = DefaultClientSessionContext.fromClientSessionAndScopeParameter(clientSession, oldTokenScope, session);
 
+        if (oldTokenScope != null && parseScopeParameter(oldTokenScope).anyMatch(s -> OAuth2Constants.OFFLINE_ACCESS.equals(s))
+                && clientSessionCtx.getClientScopesStream().noneMatch(s -> OAuth2Constants.OFFLINE_ACCESS.equals(s.getName()))) {
+            // offline access needed but scope is not allowed
+            throw new OAuthErrorException(OAuthErrorException.INVALID_SCOPE, "Invalid scope: " + OAuth2Constants.OFFLINE_ACCESS);
+        }
```
ahus1: I see that even for lightweight tokens the scope is always present, so I assume there is never the case of `oldTokenScope` being null. 

I was briefly wondering if there would be the the need to handle any other removed scope. When trying this out, I found that any removed scope is then removed on the next token refresh, which is IMHO graceful enough and sufficient.

ahus1 (changes requested):
Thank you for the PR. See some review notes below. 

Could you please add also a check for the token introspection so that an access token for an offline session is considered invalid if the offline scope doesn't exist any more? Thanks!

services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
rmartinc: The normal scopes are filtered when calculating the scopes in the [DefaultClientSessionContext](https://github.com/keycloak/keycloak/blob/26.4.2/services/src/main/java/org/keycloak/services/util/DefaultClientSessionContext.java#L92-L101). The requested scopes are limited to ones the client has (without dynamic scopes [here](https://github.com/keycloak/keycloak/blob/26.4.2/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java#L653-L664)). So, the refresh is allowed but with less scopes (the removed ones are not added, it's the same for the user). So I think that part is OK, the response is just limited to the scopes that are still assigned. I would not change this behavior.

For `offline_access` is a bit different, because the offline session is not valid anymore for this client. Returning an error makes sense IMO. The problem was that we were never checking this. The session was offline, inherited by the new refresh and nothing checked the scope was not present anymore.

I'm going to remove the `oldTokenScope != null`. I think that you are right, it will be empty string in the worst case.

rmartinc: Done! Let's see the tests...

ahus1:
@rmartinc - did you see my comment above, and what's your take on this?

> Could you please add also a check for the token introspection so that an access token for an offline session is considered invalid if the offline scope doesn't exist any more? Thanks!

services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
rmartinc: Sorry! I didn't read the check about introspection. Doing it right now.

rmartinc:
@ahus1 @graziang I have changed this a bit. I created a utility method in `UserSessionUtil` to check if the offline_access is still valid for the client session. This way it can be used for both, introspection and refresh. I have changed also a bit the error messages and I think that now they make more sense in both endpoints. Let's see the CI, I have just run locally the oauth and oidc tests.

graziang (approved):
@rmartinc thanks!

rmartinc:
If you think that it's OK please re-run the failed CI. It should be unrelated.

ahus1 (approved):
Thank you for this change!

ahus1:
@rmartinc - let me know if I should merge this one, or if anyone else from your team should review it.

rmartinc:
@graziang already approved, so I think we are OK to merge it. I'll prepare the backports later.

PR #43760, opened 2025-10-28
Cleaning up threadlocals to prevent (small) memory leak

### LINKED ISSUE #43759: QuarkusKeycloakSession not garbage collected when running Liquibase
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

When Keycloak performs the Liquibase upgrade, it misses to clear a thread local context, so the session is never garbage collected. 

As this is an effort to find all those places, this is a housekeeping task. It will reduce the memory usage of Keycloak slightly in this place, and to a larger extent in other places.

### Version

main

### Regression

### Expected behavior

Thread locals should be removed when no longer used.

### Actual behavior

The thread local is kept

### How to Reproduce?

Start up Keycloak, create a memory dump, and search for QuarkusKeycloakSession

### Anything else?

I'll create a PR.

ahus1:
Closes #43759


ahus1:
@ryanemerson - please review when have the time and feel familiar enough with the code at hand. Thanks!

shawkins (approved):
LGTM

abstractj (approved):

PR #43762, opened 2025-10-28
Don't keep an old session to avoid a stable object and a memory leak

### LINKED ISSUE #43761: QuarkusKeycloakSession kept in memory for each timer
### Before reporting an issue

### Area

core

### Describe the bug

When creating a timer, an instance of the KeycloakSession is kept in memory.

The code only needs the sessionFactory, so this should be avoided. 

### Version

main

### Regression

### Expected behavior

Don't keep the session, instead keep the sessionFactory

### Actual behavior

A session is kept, and never used. It is also problematic as that session is then later already closed and/or stale.

### How to Reproduce?

Start up Keycloak, and create a heap dump. Look for QuarkusKeycloakSession instances and where they are retained.

### Anything else?

I'll prepare a PR

ahus1:
Closes #43761

### DIFF
--- a/services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
+++ b/services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
@@ -51,12 +51,7 @@ public class BasicTimerProvider implements TimerProvider {
 
     @Override
     public void schedule(final Runnable runnable, final long intervalMillis, String taskName) {
-        TimerTask task = new TimerTask() {
-            @Override
-            public void run() {
-                runnable.run();
-            }
-        };
+        TimerTask task = new BasicTimerTask(runnable);
 
         TimerTaskContextImpl taskContext = new TimerTaskContextImpl(runnable, task, Time.currentTimeMillis(), intervalMillis);
         TimerTaskContextImpl existingTask = factory.putTask(taskName, taskContext);
@@ -94,5 +89,22 @@ public class BasicTimerProvider implements TimerProvider {
     @Override
     public Map<String, TimerTaskContext> getTasks() {
         return Collections.unmodifiableMap(new HashMap<>(factory.getTasks()));
+    }
+
+    /**
+     * Using a private static class avoids keeping a reference to {@link BasicTimerProvider} which then fails to be garbage collected,
+     * including its reference to {@link KeycloakSession}.
+     */
+    private static class BasicTimerTask extends TimerTask {
+        private final Runnable runnable;
+
+        public BasicTimerTask(Runnable runnable) {
+            this.runnable = runnable;
+        }
+
+        @Override
+        public void run() {
+            runnable.run();
+        }
     }
 }


ahus1:
@ryanemerson - please review when have the time and feel familiar enough with the code at hand. Thanks!

services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
```diff
@@ -37,13 +38,13 @@ public class BasicTimerProvider implements TimerProvider {
...
 
-    private final KeycloakSession session;
+    private final KeycloakSessionFactory sessionFactory;
     private final Timer timer;
     private final int transactionTimeout;
     private final BasicTimerProviderFactory factory;
 
     public BasicTimerProvider(KeycloakSession session, Timer timer, int transactionTimeout, BasicTimerProviderFactory factory) {
-        this.session = session;
+        this.sessionFactory = session.getKeycloakSessionFactory();
```
ryanemerson: Is the `services` module considered API that we can't break?

If not, I would change the ctor to be `BasicTimerProvider(KeycloakSessionFactory sessionFactory, ...)`, otherwise it's probably not worth deprecating the old one and adding a distinct constructor.

ahus1: Thank you for the review. I went the extra mile to deprecate the old timer provider. It is a private API, still as I plan to backport it to a patch release, I want to go the deprecation route.

shawkins: Rather than the session or sessionFactory member variable, you can use KeycloakModelUtils.getKeycloakSession when creating the ScheduledTaskRunner.

ryanemerson:
@ahus1 I pushed a commit to reduce repetition, feel free to disregard if you prefer the more verbose code.

ryanemerson (dismissed):

services/src/main/java/org/keycloak/timer/basic/BasicTimerProvider.java
ahus1: @shawkins - let's do that as the next step if you are ok with that.

shawkins: @ahus1 looking at this more, I'm unsure of this change in general. It's a common pattern for providers to hold references to the session as they themselves are session scoped.

What exactly is holding a reference to the BasicTimerProvider in this scenario?

ahus1: There is the inner anonymous class that holds a reference, see this screenshot. I'll rework it to use an inner static class instead. 

image

ahus1:
@ryanemerson  or  @shawkins - please re-review when you have the time. Thanks!

ryanemerson (approved):

pedroigor (approved):

shawkins (approved):
LGTM, thanks @ahus1

PR #43764, opened 2025-10-28
Local user can't login when ldap error

### LINKED ISSUE #43639: Local user can't login when ldap error
### Before reporting an issue

### Area

ldap

### Describe the bug

References https://github.com/keycloak/keycloak/issues/29147#issuecomment-3430738378

When Keycloak is configured to allow login with email. The login process attempts to retrieve the user from LDAP if the username contains an '@' character and differs from the stored email. If LDAP is not available, even a local user cannot log in using their username.

### Version

26.4.1

### Regression

### Expected behavior

If a local user exists in the internal store, Keycloak should not attempt any LDAP query, regardless of email lookup.

### Actual behavior

Keycloak attempts to log in by email, querying the LDAP provider and failing — it never reaches the loginByUsername method.

```
at org.keycloak.storage.UserStorageManager.tryResolveFederatedUser(UserStorageManager.java:1142)
        at org.keycloak.storage.UserStorageManager.getUserByAttribute(UserStorageManager.java:1137)
        at org.keycloak.storage.UserStorageManager.getUserByEmail(UserStorageManager.java:541)
        at org.keycloak.models.cache.infinispan.UserCacheSession.getUserByEmail(UserCacheSession.java:438)
        at org.keycloak.models.utils.KeycloakModelUtils.findUserByNameOrEmail(KeycloakModelUtils.java:321)
       at org.keycloak.authentication.authenticators.browser.AbstractUsernameFormAuthenticator.getUserFromForm(
``` 

### How to Reproduce?

1. Configure a user federation with a read-only/unsynced LDAP provider.
2. In realm settings, set "Login with email" to On.
3. Create a local user whose username contains an “@” character.
4. Block communication to the LDAP server.
5. Attempt to log in with the created user (the one with “@” in the username).

### Anything else?

_No response_

martin-kanis:
Closes #43639

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
@@ -1146,9 +1146,18 @@ public class UserStorageManager extends AbstractStorageManager<UserStorageProvid
     }
 
     private UserModel tryResolveFederatedUser(RealmModel realm, Function<UserLookupProvider, UserModel> loader) {
-        return mapEnabledStorageProvidersWithTimeout(realm, UserLookupProvider.class, loader)
-                .findFirst()
-                .orElse(null);
+        return mapEnabledStorageProvidersWithTimeout(realm, UserLookupProvider.class, provider -> {
+            try {
+                return loader.apply(provider);
+            } catch (StorageUnavailableException e) {
+                logger.warnf(e, "User storage provider %s is unavailable. " +
+                             "Continuing with other providers for graceful degradation.",
+                             provider.getClass().getSimpleName());
+                return null;
+            }
+        })
+        .findFirst()
+        .orElse(null);
     }
 
     private boolean isSyncSettingsUpdated(UserStorageProviderModel previous, UserStorageProviderModel actual) {
--- a/common/src/main/java/org/keycloak/common/util/Throwables.java
+++ b/common/src/main/java/org/keycloak/common/util/Throwables.java
@@ -1,0 +1,38 @@
+package org.keycloak.common.util;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * A utility class for working with {@link Throwable} instances.
+ */
+public final class Throwables {
+
+    /**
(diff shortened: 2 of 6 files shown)


model/storage-private/src/main/java/org/keycloak/storage/UserStorageManager.java
```diff
@@ -1140,9 +1140,25 @@ private UserModel getUserByAttribute(RealmModel realm, Function<UserLookupProvid
...
     private UserModel tryResolveFederatedUser(RealmModel realm, Function<UserLookupProvider, UserModel> loader) {
-        return mapEnabledStorageProvidersWithTimeout(realm, UserLookupProvider.class, loader)
-                .findFirst()
-                .orElse(null);
+        return mapEnabledStorageProvidersWithTimeout(realm, UserLookupProvider.class, provider -> {
+            try {
+                return loader.apply(provider);
+            } catch (Exception e) {
+                // Re-throw business logic exceptions that should not be handled gracefully
+                if (e instanceof ModelException || e instanceof IllegalStateException || e instanceof IllegalArgumentException) {
```
pedroigor: Model exceptions might also wrap connectivity issues. For instance, using LDAP.

pedroigor (commented):
I run the same scenario, but using an LDAP provider, as described in the original issue.

The local user is not able to log in if a `ModelException` is thrown due to connectivity issues.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/federation/UserStorageGracefulDegradationTest.java
```diff
@@ -183,4 +203,17 @@ public void testGracefulDegradationOnCountFailure(KeycloakSession session) {
         Assert.assertTrue("User count should be at least 3, but was: " + userCount, userCount >= 3);
     }
 
+    @Test
+    public void testLocalUserLoginWithAtSignSucceedsWhenLDAPFails() {
```
pedroigor: You're testing a provider that does not really mimic the behavior when, for instance, there are connectivity issues with LDAP.

Can you try to use the embedded LDAP server instead and check if your test is still passing?

pedroigor (approved):

PR #43765, opened 2025-10-28
Allow only normalized URLs in requests

### LINKED ISSUE #43763: CVE-2025-10939: Normalizing of Keycloak URLs not documented
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

When processing URLs, Keycloak normalizes URLs based on the Quarkus logic. 

This includes collapsing a double slash `//`, as well as evaluating `..` in the path. 

While this is a convenience feature also partly an RFC3986 as well, it makes URL filtering for Keycloak unnecessarily hard and difficult to document. Also Quarkus adds another normalization by removing the double slash. 

### Version

main

### Regression

### Expected behavior

Allow simple URL filtering for Keycloak rules.

### Actual behavior

Keycloak normalizes URLs, thereby making it hard to filter them.

### How to Reproduce?

Use `curl --path-as-is ...` and use an URL with `..` in it. 

### Anything else?

I'll prepare a PR.

ahus1:
Closes #43763

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
@@ -150,4 +150,12 @@ public class HttpOptions {
             .description("Service level objectives for HTTP server requests. Use this instead of the default histogram, or use it in combination to add additional buckets. " +
                     "Specify a list of comma-separated values defined in milliseconds. Example with buckets from 5ms to 10s: 5,10,25,50,250,500,1000,2500,5000,10000")
             .build();
+
+    public static final Option<Boolean> HTTP_ACCEPT_NON_NORMALIZED_PATHS = new OptionBuilder<>("http-accept-non-normalized-paths", Boolean.class)
+            .category(OptionCategory.HTTP)
+            .description("If the server should accept paths that are not normalized according to RFC3986 or that contain a double slash ('//'). While accepting those requests might be relevant for legacy applications, it is recommended to disable it to allow for more concise URL filtering.")
+            .deprecated()
+            .defaultValue(Boolean.FALSE)
+            .build();
+
 }
--- a/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
+++ b/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
@@ -49,9 +49,12 @@ import io.quarkus.narayana.jta.runtime.TransactionManagerBuildTimeConfig.UnsafeM
 import io.quarkus.resteasy.reactive.server.spi.MethodScannerBuildItem;
 import io.quarkus.resteasy.reactive.server.spi.PreExceptionMapperHandlerBuildItem;
 import io.quarkus.runtime.configuration.ConfigurationException;
+import io.quarkus.vertx.http.deployment.FilterBuildItem;
 import io.quarkus.vertx.http.deployment.HttpRootPathBuildItem;
+import io.quarkus.vertx.http.deployment.ManagementInterfaceFilterBuildItem;
 import io.quarkus.vertx.http.deployment.NonApplicationRootPathBuildItem;
 import io.quarkus.vertx.http.deployment.RouteBuildItem;
+import io.quarkus.vertx.http.runtime.security.SecurityHandlerPriorities;
 import jakarta.persistence.Entity;
 import jakarta.persistence.PersistenceUnitTransactionType;
 import org.eclipse.microprofile.config.spi.ConfigSource;
@@ -270,6 +273,26 @@ class KeycloakProcessor {
                                 .handler(recorder.getRedirectHandler(relativePath))
                                 .build())
                 );
+    }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    @Consume(ConfigBuildItem.class)
+    void filterAllRequests(BuildProducer<FilterBuildItem> filters, KeycloakRecorder recorder) {
+        var filter = recorder.getRejectNonNormalizedPathFilter();
+        if (filter != null) {
+            filters.produce(new FilterBuildItem(filter, SecurityHandlerPriorities.CORS + 1));
+        }
+    }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep(onlyIf = IsManagementEnabled.class)
+    @Consume(ConfigBuildItem.class)
+    void filterAllManagementRequests(BuildProducer<ManagementInterfaceFilterBuildItem> filters, KeycloakRecorder recorder) {
+        var filter = recorder.getRejectNonNormalizedPathFilter();
+        if (filter != null) {
+            filters.produce(new ManagementInterfaceFilterBuildItem(filter, SecurityHandlerPriorities.CORS + 1));
+        }
     }
 
     @Record(ExecutionTime.STATIC_INIT)
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/RejectNonNormalizedPathFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/RejectNonNormalizedPathFilter.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.quarkus.runtime.services;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
+import org.jboss.logging.Logger;
+import org.keycloak.representations.idm.OAuth2ErrorRepresentation;
+import org.keycloak.services.util.ObjectMapperResolver;
+
+import java.util.Objects;
+
+/**
+ * This filter rejects all paths that need normalization as of RFC3986 or that have double slashes.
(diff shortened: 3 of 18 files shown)


quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -272,6 +276,29 @@ void configureRedirectForRootPath(BuildProducer<RouteBuildItem> routes,
                 );
     }
 
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    void filterAllRequests(BuildProducer<FilterBuildItem> filters,
+                           BuildProducer<ManagementInterfaceFilterBuildItem> managementFilters,
```
mabartos:
```suggestion
                           
```
Unused. Or do you need to ensure the order of build step executions?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/RejectNonNormalizedPathFilter.java
```diff
@@ -0,0 +1,58 @@
...
+            LOGGER.debugf("Request with a non-normalized path blocked: %s vs. %s", routingContext.request().path(), routingContext.normalizedPath());
+            OAuth2ErrorRepresentation error = new OAuth2ErrorRepresentation("missingNormalization", "Request path not normalized");
+            routingContext.response().headers().add("Content-Type", "application/json; charset=UTF-8");
+            String jsonString;
+            try {
+                jsonString = MAPPER.writeValueAsString(error);
+            } catch (JsonProcessingException e) {
+                jsonString = "";
+            }
+            routingContext.response().setStatusCode(400).end(jsonString);
```
mabartos, line 52: Wouldn't be better to have it handled by KeycloakErrorHandler? I'd say yes if possible.

mabartos (commented):
@ahus1 Nice work! Just added 2 comments for improvement.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/RejectNonNormalizedPathFilter.java
ahus1: This is happening outside of the RESTEasy stack, so it is not possible.

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
```diff
@@ -150,4 +150,13 @@ public enum ClientAuth {
             .description("Service level objectives for HTTP server requests. Use this instead of the default histogram, or use it in combination to add additional buckets. " +
                     "Specify a list of comma-separated values defined in milliseconds. Example with buckets from 5ms to 10s: 5,10,25,50,250,500,1000,2500,5000,10000")
             .build();
+
+    public static final Option<Boolean> HTTP_ACCEPT_NON_NORMALIZED_PATHS = new OptionBuilder<>("http-accept-non-normalized-paths", Boolean.class)
+            .category(OptionCategory.HTTP)
+            .description("If the server should accept paths that are not normalized according to RFC3986 or that contain a double slash ('//'). While accepting those requests might be relevant for legacy applications, it is recommended to disable it to allow for more concise URL filtering.")
+            .buildTime(true)
+            .deprecated()
```
mabartos, line 157:
```suggestion
            .description("If the server should accept paths that are not normalized according to RFC3986 or that contain a double slash ('//'). While accepting those requests might be relevant for legacy applications, it is recommended to disable it to allow for more concise URL filtering.")
            .deprecated()
```

It might not be necessary to have it as a build-time option.

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -272,6 +276,29 @@ void configureRedirectForRootPath(BuildProducer<RouteBuildItem> routes,
                 );
     }
 
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    void filterAllRequests(BuildProducer<FilterBuildItem> filters,
+                           
+                                      KeycloakRecorder recorder) {
```
mabartos:
```suggestion
    void filterAllRequests(BuildProducer<FilterBuildItem> filters, KeycloakRecorder recorder) {
```

nitpick: Some formatting issue.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/RejectNonNormalizedPathFilter.java
mabartos: +1

mabartos (commented):
Last comments, otherwise LGTM.

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
ahus1: I'm adding the filter to a build time option ... so I'd assume it is build time? Or do you suggest to always add the filter, and then parameterize it?

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -272,6 +276,27 @@ void configureRedirectForRootPath(BuildProducer<RouteBuildItem> routes,
                 );
     }
 
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    void filterAllRequests(BuildProducer<FilterBuildItem> filters, KeycloakRecorder recorder) {
```
mabartos, line 281:
```suggestion
    @BuildStep
    @Consume(ConfigBuildItem.class)
    void filterAllRequests(BuildProducer<FilterBuildItem> filters, KeycloakRecorder recorder) {
```

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -272,6 +276,27 @@ void configureRedirectForRootPath(BuildProducer<RouteBuildItem> routes,
...
+        if (!acceptNonNormalizedPaths()) {
+            filters.produce(new FilterBuildItem(new RejectNonNormalizedPathFilter(), SecurityHandlerPriorities.CORS + 1));
+        }
+    }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep(onlyIf = IsManagementEnabled.class)
+    void filterAllManagementRequests(BuildProducer<ManagementInterfaceFilterBuildItem> filters,
```
mabartos:
```suggestion
    @BuildStep(onlyIf = IsManagementEnabled.class)
    @Consume(ConfigBuildItem.class)
    void filterAllManagementRequests(BuildProducer<ManagementInterfaceFilterBuildItem> filters,
```

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
mabartos: @ahus1 These filters are actually added during the static init phase (`Record(ExecutionTime.STATIC_INIT)`) and not created during the augmentation. 

It's always better to have an option configurable at runtime if it's only some marker of behavior checked in runtime anyway. 

If you remove the `buildTime(true)` here and add the `@Consume(ConfigBuildItem.class)` as described in the additional comments, it should be fine :)

ahus1: I gave it a try, and then the test did no longer pass. I reverted the change.

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
```diff
@@ -150,4 +150,13 @@ public enum ClientAuth {
             .description("Service level objectives for HTTP server requests. Use this instead of the default histogram, or use it in combination to add additional buckets. " +
                     "Specify a list of comma-separated values defined in milliseconds. Example with buckets from 5ms to 10s: 5,10,25,50,250,500,1000,2500,5000,10000")
             .build();
+
+    public static final Option<Boolean> HTTP_ACCEPT_NON_NORMALIZED_PATHS = new OptionBuilder<>("http-accept-non-normalized-paths", Boolean.class)
```
vmuzikar, line 154: Do we really want to make it configurable? It's a breaking change but to fix a security issue. What would be a real and valid use case for the non-normalized paths?

Sorry if this has been already discussed somewhere.

vmuzikar (commented):
@ahus1 Thank you for the PR.

I'm not sure if I'm doing something wrong (probably yes as the tests pass) but if I do something as simple as `curl -v http://localhost:8080/realms/xxx/../master/`, I get 200 response instead of 400.

shawkins:
> @ahus1 Thank you for the PR.
> 
> I'm not sure if I'm doing something wrong (probably yes as the tests pass) but if I do something as simple as `curl -v http://localhost:8080/realms/xxx/../master/`, I get 200 response instead of 400.

curl needs --path-as-is otherwise it will do the normalization.

vmuzikar:
> curl needs --path-as-is otherwise it will do the normalization.

Thanks, sorry for the noise. What confused me is that even browser normalized the path (did not get 400).

quarkus/config-api/src/main/java/org/keycloak/config/HttpOptions.java
ahus1: > What would be a real and valid use case for the non-normalized paths?

I would say adding `..` to the path is not a valid use case, maybe someone used it as a workaround for some issue in a machine-to-machine communication. 

The most common thing I could think of is a double `//` in the URL, where someone might have concatenated URLs, and they end up with two `//`. We would break those setups. And to back out of such a setup, there is this option to revert it. And it is deprecated, so we can remove it again soon - like in 27.0

vmuzikar: I'd say if someone uses `//` it's probably an oversight and a misconfiguration. So having it configurable is more like a convenience feature. But it's not a blocker for me, let's keep it there.

vmuzikar (dismissed):
LGTM, thank you for the fix.

@mabartos Are you ok with the changes as is?

mabartos:
@ahus1 We should ensure the build chain execution order to prevent any issues + would be good to have the property runtime.

Sent a PR to your fork: https://github.com/ahus1/keycloak/pull/434

mabartos (approved):
@ahus1 LGTM, thanks!

ahus1:
@mabartos - thank you for the PR. I merge it to this branch, rebased it, and fixed the help test. Let's wait for the build to be green.

vmuzikar (approved):

pohjanoksajukka:
@ahus1 @mabartos, is this the fix for CVE-2025-10939?

ahus1:
> is this the fix for CVE-2025-10939?

Yes.

PR #43779, opened 2025-10-28
task: testing servicemonitor patching

### LINKED ISSUE #43778: Test ServiceMonitor desired state patching
### Description

When looking at olm v1 support I noticed that the patch verb is not explicitly included in the service monitor rbac. We should test specifically for maintaining desired state, which is done via SSA patch.

shawkins:
closes: #43778

No backport is needed - tested without the RBAC change and things still worked fine. I'm not quite sure then when the patch permission is checked with SSA, but to keep it consistent with the other resources I went ahead and added it.


ryanemerson (approved):
LGTM, thanks @shawkins

ahus1 (approved):

PR #43786, opened 2025-10-28
Resolve session leak in DeclarativeUserProfileProvider

### LINKED ISSUE #43785: QuarkusKeycloakSession leak in DeclarativeUserProfileProvider
### Before reporting an issue

### Area

user-profile

### Describe the bug

Due to the use or anonymous inner classes, the code holds on to the instance and the session. 

Image

### Version

main

### Regression

### Expected behavior

There should be no leak of the KeycloakSession

### Actual behavior

There is a leak

### How to Reproduce?

Start up an instance, log in once to initialize the DeclarativeUserProfileProvider, and take a memory snapshot.

### Anything else?

I'll provide a PR

ahus1:
Closes #43785


ahus1:
@pedroigor - this seems to be the regression of the memory leak introduced in KC 25. Apparently an anonymous class shouldn't be cached. A static nested class is cleaner. Please review and approve. Thanks!

pedroigor (approved):

PR #43805, opened 2025-10-29
fix: considering source ordinality with spi options

### LINKED ISSUE #43793: import does not seem to run db migration
### Before reporting an issue

### Area

import-export

### Describe the bug

We are using keycloak container, and what we do is to always run import to initalise master realm with some settings.

`opt/keycloak/bin/kc.sh import --override=false`

Then we do start using the start command

With latest version, we encountered this : 

```
2025-10-28 18:29:10,327 INFO  [org.keycloak.url.HostnameV2ProviderFactory] (main) If hostname is specified, hostname-strict is effectively ignored
2025-10-28 18:29:35,018 INFO  [org.hibernate.orm.jdbc.batch] (JPA Startup Thread) HHH100501: Automatic JDBC statement batching enabled (maximum batch size 32)
2025-10-28 18:29:46,299 INFO  [com.arjuna.ats.jbossatx] (main) ARJUNA032014: Stopping transaction recovery manager
2025-10-28 18:29:46,620 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) ERROR: Failed to start server in (nonserver) mode
2025-10-28 18:29:46,620 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) ERROR: Database not up-to-date, please enable database migration
2025-10-28 18:29:46,624 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) For more details run the same command passing the '--verbose' option. Also you can use '--help' to see the details about the usage of the particular command.
```

### Version

26.4.2

### Regression

### Expected behavior

I would expect the import command to run database migration before doing the import.

### Actual behavior

Database migration is not done when launching import, even with ` --spi-connections-jpa-quarkus-migration-strategy=update`

### How to Reproduce?

1- Setup keycloak on a previous version
2- Keep the database
3- Change the version and run import, before starting the server

### Anything else?

I'm pretty sure this was working before, we are using keycloak since version 24.0.5, and I don't think we ever had any issues with db upgrade.

shawkins:
closes: #43793

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/MicroProfileConfigProvider.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/MicroProfileConfigProvider.java
@@ -20,13 +20,15 @@ package org.keycloak.quarkus.runtime.configuration;
 import static org.keycloak.quarkus.runtime.configuration.Configuration.OPTION_PART_SEPARATOR;
 import static org.keycloak.quarkus.runtime.configuration.Configuration.toDashCase;
 
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
-import org.eclipse.microprofile.config.ConfigProvider;
-
+import org.eclipse.microprofile.config.ConfigValue;
 import org.keycloak.Config;
 import org.keycloak.Config.Scope;
+
+import io.smallrye.config.SmallRyeConfig;
 
 public class MicroProfileConfigProvider implements Config.ConfigProvider {
 
@@ -36,13 +38,13 @@ public class MicroProfileConfigProvider implements Config.ConfigProvider {
     public static final String NS_QUARKUS = "quarkus";
     public static final String NS_QUARKUS_PREFIX = "quarkus" + ".";
 
-    private final org.eclipse.microprofile.config.Config config;
+    private final SmallRyeConfig config;
 
     public MicroProfileConfigProvider() {
-        this(ConfigProvider.getConfig());
+        this(Configuration.getConfig());
     }
 
-    public MicroProfileConfigProvider(org.eclipse.microprofile.config.Config config) {
+    public MicroProfileConfigProvider(SmallRyeConfig config) {
         this.config = config;
     }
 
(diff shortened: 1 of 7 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/MicroProfileConfigProvider.java
```diff
@@ -137,8 +137,18 @@ private <T> T getValue(String key, Class<T> clazz, T defaultValue) {
...
+            ConfigValue newValue = config.getConfigValue(name);
+            if (newValue.getValue() == null) {
+                name = oldName;
+            } else {
+                ConfigValue oldValue = config.getConfigValue(oldName);
+                if (oldValue.getValue() != null && oldValue.getSourceOrdinal() > newValue.getSourceOrdinal()) {
+                    name = oldName;
+                }
+            }
+            return config.getOptionalValue(name, clazz).orElse(defaultValue);
```
mabartos:
```suggestion
            Converter<T> converter = config.getConverter(clazz).orElseThrow();
            ConfigValue newValue = config.getConfigValue(separatorPrefix.concat(dashCase));
            ConfigValue oldValue = config.getConfigValue(prefix.concat(dashCase));

            String selectedValue = newValue.getValue();
            if (selectedValue == null || (oldValue.getValue() != null && oldValue.getSourceOrdinal() > newValue.getSourceOrdinal())) {
                selectedValue = oldValue.getValue();
            }

            return Optional.ofNullable(selectedValue).map(converter::convert).orElse(defaultValue);
```

IMHO, we could do this improvement to not call the `getConfigValue` and `getOptionalValue` twice for old and new value. @shawkins WDYT?

mabartos (commented):
@shawkins LGTM, just one suggestion for improvement. Thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/MicroProfileConfigProvider.java
shawkins: My initial reasoning for not doing something like this is that getOptionalValue handles the conversion of indexed properties - should we ever add support for that, it would then simply work here.

If we want to ignore that for now, then I would just use SmallryeConfig.convert, rather than dealing with the converter directly.

shawkins: @mabartos updated the logic

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/MicroProfileConfigProvider.java
```diff
@@ -20,14 +20,16 @@
...
 import java.util.stream.StreamSupport;
-import org.eclipse.microprofile.config.ConfigProvider;
-
+import org.eclipse.microprofile.config.ConfigValue;
 import org.keycloak.Config;
 import org.keycloak.Config.Scope;
 
+import io.smallrye.config.SmallRyeConfig;
+
 public class MicroProfileConfigProvider implements Config.ConfigProvider {
```
mabartos, line 33:
```suggestion
public class SmallRyeConfigProvider implements Config.ConfigProvider {
```

So, shouldn't be it SmallRye config provider? :P

mabartos (approved):
@shawkins LGTM, just one small improvement suggestion.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/MicroProfileConfigProvider.java
shawkins: > So, shouldn't be it SmallRye config provider? :P

It could be, but I don't think we need to update the class, factory, and other references, as there's no practical difference to us between MP and SmallRye.

PR #43817, opened 2025-10-29
fix: redoing the watching logic to provide a better status

### LINKED ISSUE #43777: Add missing secret warning status
### Description

The operator tracks if secrets are missing, but this is not directly shown in the Keycloak status.

### Value Proposition

To make the admin aware why a deployment is not progressing.

### Goals

To make the admin aware why a deployment is not progressing.

### Non-Goals

N/A

### Discussion

_No response_

### Notes

_No response_

shawkins:
closes: #43777

At one point I believe the StatefulSet status showed what secrets / configmaps were missing and we would then put that on the keycloak status. This does not seem to be the case anymore. These changes add missing non-optional configmaps and secrets as warnings.

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakController.java
@@ -16,11 +16,13 @@
  */
 package org.keycloak.operator.controllers;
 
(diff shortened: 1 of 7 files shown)


vmuzikar (approved):
LGTM, sorry for the late response.

PR #43823, opened 2025-10-29
Processing workflow events asynchronously

### LINKED ISSUE #42386: [RLM] Review the action execution thread model
### Description

Today, actions are executed sequentially and in a single thread. However, actions are essentially blocking operations that can either update resources in the local database or interact with external services (e.g.: notifications via email).

We need to review how actions are executed based on the following considerations:

* Given their blocking nature, we should consider executing actions in a separate thread when processing a policy.
* * What level of configuration we want to suport in order to change timeouts and parallelism?
* How to handle errors when executing actions and what is the effect on the scheduled policy?
* * Does it make sense a retry mechanism?
* How transactions should be handled? One for each action? A single for all actions?
* * https://github.com/keycloak/keycloak/issues/42110
* Can we leverage executing actions for a set of resources in bulk?
* * https://github.com/keycloak/keycloak/issues/42111

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pedroigor:
Closes #42386


server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    protected void commitImpl() {
+        future = CompletableFuture.runAsync(task, executorService)
+                .orTimeout(timeout, TimeUnit.MILLISECONDS)
+                .whenComplete((result, error) -> {
+                    if (error instanceof TimeoutException) {
+                        LOGGER.warnf("Timeout occurred while processing workflow task: ", task);
+                    }
```
pruivo: If you remove the task from the list here (`executor.removeTask(this);`), you won't need the `WorkflowTaskReaper`.

pedroigor: I appreciate your feedback and help, @pruivo. You have a lot of experience in this area because of ISPN, and I was already planning to ping you for feedback.

On the other hand, removing the reaper we would lose the possibility of monitoring the task queue. One of the key aspects when using a reaper thread.

I'm still evaluating some ideas with the ultimate goal of having complete control over the lifetime and execution of tasks.

pedroigor: @pruivo I'm also trying to see if we can have a single mechanism like that as a replacement for the very limited `TimerProvider` and `ExecutorProvider` SPIs we are using today.

But not as a primary goal now ... Something to consider for the future ...

pruivo: > On the other hand, removing the reaper we would lose the possibility of monitoring the task queue. One of the key aspects when using a reaper thread.

How do you want to monitor the task queue? I would probably add a metric to get the active workflows.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+        ProcessEventTask task = new ProcessEventTask(sessionFactory, session.getContext(), event);
+        transactionManager.enlistAfterCompletion(new WorkflowTask(this, task, eventTimeout, eventExecutor));
+        enlistCleanupTransaction(session, transactionManager);
+    }
+
+    void waitActiveTasks() {
+        synchronized (tasks) {
+            for (WorkflowTask task : tasks) {
+                if (task.isActive()) {
+                    task.await();
```
pruivo: Waiting inside a `synchronized` block will block any other threads from adding/removing tasks. Is it intentional?

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+import org.keycloak.models.KeycloakTransactionManager;
+
+final class WorkflowExecutor {
+
+    private static final AtomicBoolean INITED = new AtomicBoolean(false);
+    private static final String SESSION_ATTRIBUTE_WORKFLOW_TRANSACTION_ENLISTED = "kc.workflow.transaction.enlisted";
+
+    static WorkflowExecutor INSTANCE;
+
+    static WorkflowExecutor getInstance(KeycloakSessionFactory sessionFactory) {
```
pruivo: This class needs to be a provider. How is this going to work in the testsuite? The first test creates a new instance that is going to be used by other tests.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
pedroigor: I haven't thought about exposing metrics for this, but I like the idea. I'll create a follow-up issue.

I was looking at logging only.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowExecutor.java
pedroigor: This is no longer needed. I changed the executor to allow configuring whether it should run in blocking mode.

Right now, this blocking behavior is needed for tests. Or at least while we don't have a nice way of running assertions, considering the async nature of workflow event tasks being introduced here.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowExecutor.java
pedroigor: I don't know where to put this yet. We don't have a specific SPI for workflows in the sense that it allows registering/configuring cross-cutting aspects of the feature, like this executor. We cannot create the executor at runtime when serving requests ... It should be done as part of session factory initialization.

I'll change that before making the PR ready for review and see what others think we should be doing here. For now, it is just unblocking me from moving forward.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
pruivo: Well, you can add the logging here, to the `whenComplete` method, instead of the reaper.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowExecutor.java
pruivo: Keycloak has a nice event system, and the testsuite has options to poll/wait for those. I would use that.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
pedroigor: In addition to logging, also allow administrators to track task execution from the Admin UI.

Based on your idea about exposing metrics, it should help to have a consolidated view of workflow metrics across the different nodes in a cluster. What I'm doing is for each individual node.

pruivo: I don't see any value in having that info in the Admin UI. Not only will you only see the local task execution, but the load balancer gets in the way by connecting to a random Keycloak node.

pedroigor: True, exposing metrics should be better.

I was thinking something in the Admin UI for this, in addition to querying workflows associated with realm resource types.

But still, we need some logging, and I will be creating a follow-up issue for exposing metrics.

sguilhen (approved):

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/ConditionEvaluator.java
```diff
@@ -57,7 +58,7 @@ public Boolean visitAtom(BooleanConditionParser.AtomContext ctx) {
     @Override
     public Boolean visitConditionCall(BooleanConditionParser.ConditionCallContext ctx) {
         String conditionName = ctx.Identifier().getText();
-        WorkflowConditionProvider conditionProvider = manager.getConditionProvider(conditionName, extractParameter(ctx.parameter()));
+        WorkflowConditionProvider conditionProvider = provider.getConditionProvider(conditionName, extractParameter(ctx.parameter()));
         return conditionProvider.evaluate(event);
     }
```
pruivo, line 61: The field `DefaultWorkflowProvider provider` can be removed (it does not use any state from it), and the code here can be replaced with 
```java
        WorkflowConditionProviderFactory<WorkflowConditionProvider> providerFactory = (WorkflowConditionProviderFactory<WorkflowConditionProvider>) session.getKeycloakSessionFactory().getProviderFactory(WorkflowConditionProvider.class, conditionName);

        if (providerFactory == null) {
            throw new WorkflowInvalidStateException("Could not find condition provider: " + conditionName);
        }

        return providerFactory.create(session, extractParameter(ctx.parameter()))
                .evaluate(event);
```

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/PredicateEvaluator.java
```diff
@@ -72,7 +73,7 @@ public Predicate visitAtom(BooleanConditionParser.AtomContext ctx) {
     @Override
     public Predicate visitConditionCall(BooleanConditionParser.ConditionCallContext ctx) {
         String conditionName = ctx.Identifier().getText();
-        WorkflowConditionProvider conditionProvider = manager.getConditionProvider(conditionName, extractParameter(ctx.parameter()));
+        WorkflowConditionProvider conditionProvider = provider.getConditionProvider(conditionName, extractParameter(ctx.parameter()));
```
pruivo: Same here.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+        this.session = session;
+        this.executor = executor;
+        this.sessionFactory = session.getKeycloakSessionFactory();
+        this.stateProvider = sessionFactory.getProviderFactory(WorkflowStateProvider.class).create(session);
+        this.realm = session.getContext().getRealm();
+    }
+
+    @Override
+    public ResourceTypeSelector getResourceTypeSelector(ResourceType type) {
+        Objects.requireNonNull(type, "workflow");
```
pruivo: duplicated, can be removed.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+        Objects.requireNonNull(type, "type");
+
+        if (ResourceType.USERS.equals(type)) {
+            return new UserResourceTypeWorkflowProvider(session);
+        }
+
+        throw new IllegalArgumentException("Unsupported resource type: " + type);
```
pruivo: Using the switch is safer, as the compiler will fail if you add more resource types, rather than failing at runtime. 
```suggestion
                return switch (type) {
            case USERS -> new UserResourceTypeWorkflowProvider(session); 
        };
```

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+        Objects.requireNonNull(workflow, "workflow");
+        ComponentModel component = getWorkflowComponent(workflow.getId());
+        realm.getComponentsStream(workflow.getId(), WorkflowStepProvider.class.getName()).forEach(realm::removeComponent);
+        realm.removeComponent(component);
+        stateProvider.removeByWorkflow(workflow.getId());
+    }
+
+    @Override
+    public Workflow getWorkflow(String id) {
+        return new Workflow(session, getWorkflowComponent(id));
```
pruivo, line 93: Honestly, this seems like a factory method in the `Workflow` class. 
```java
    public static Workflow ofComponent(KeycloakSession session, String componentId) {
        ComponentModel component = realm.getComponent(componentId);

        if (component == null || !WorkflowProvider.class.getName().equals(component.getProviderType())) {
            throw new BadRequestException("Not a valid resource workflow: " + componentId);
        }
        return new Workflow(session, component);
    }
```

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    @Override
+    public Stream<Workflow> getWorkflows() {
+        return realm.getComponentsStream(realm.getId(), WorkflowProvider.class.getName())
```
pruivo, line 98: same here

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+            WorkflowProvider provider = getWorkflowProvider(workflow);
+            ResourceTypeSelector selector = provider.getResourceTypeSelector(ResourceType.USERS);
+            selector.getResourceIds(workflow)
+                    .forEach(resourceId -> processEvent(Stream.of(workflow), new AdhocWorkflowEvent(ResourceType.USERS, resourceId)));
+        }
+    }
+
+    @Override
+    public WorkflowRepresentation toRepresentation(Workflow workflow) {
+        List<WorkflowStepRepresentation> steps = workflow.getSteps().map(this::toRepresentation).toList();
```
pruivo, line 147: This method looks better in `Workflow`

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+        return new WorkflowRepresentation(workflow.getId(), workflow.getName(), workflow.getConfig(), steps);
+    }
+
+    @Override
+    public Workflow toModel(WorkflowRepresentation rep) {
+        validateWorkflow(rep);
```
pruivo, line 153: Same, factory method like `Workflow.fromRepresentation(KeycloakSession, WorkflowRepresentation)`

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+        if (factory == null) {
+            throw new WorkflowInvalidStateException("Step not found: " + step.getProviderId());
+        }
+
+        return factory;
+    }
+
+    private void processEvent(Stream<Workflow> workflows, WorkflowEvent event) {
+        Map<String, ScheduledStep> scheduledSteps = stateProvider.getScheduledStepsByResource(event.getResourceId())
+                .stream().collect(HashMap::new, (m, v) -> m.put(v.workflowId(), v), HashMap::putAll);
```
pruivo:
```suggestion
                .stream().collect(Collectors.toMap(ScheduledStep::workflowId, Function.identity()));
```

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+
+        workflows.forEach(workflow -> {
+            if (!workflow.isEnabled()) {
+                log.debugf("Skipping workflow %s as it is disabled or has no steps", workflow.getName());
+                return;
+            }
```
pruivo, line 218: I've seen this code elsewhere; it could be a `Predicate` used to filter the stream content.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+            } catch (WorkflowInvalidStateException e) {
+                workflow.setEnabled(false);
+                workflow.setError(e.getMessage());
+                updateWorkflowConfig(workflow, workflow.getConfig());
+                log.warnf("Workflow %s was disabled due to: %s", workflow.getId(), e.getMessage());
+            }
+        });
+    }
+
+    private void restartWorkflow(DefaultWorkflowExecutionContext context) {
```
pruivo: only used in a single place, can be removed too.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+
+    private void scheduleWorkflow(WorkflowEvent event, Workflow workflow) {
+        executor.runTask(session, new ScheduleWorkflowTask(new DefaultWorkflowExecutionContext(session, workflow, event)));
+    }
+
+    private void runWorkflow(DefaultWorkflowExecutionContext context) {
+        executor.runTask(session, new RunWorkflowTask(context));
+    }
+
+    private WorkflowStepRepresentation toRepresentation(WorkflowStep step) {
```
pruivo, line 288: should be a method in `WorkflowStep` IMO

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -0,0 +1,360 @@
...
+        return new WorkflowStepRepresentation(step.getId(), step.getProviderId(), step.getConfig());
+    }
+
+    private void validateWorkflow(WorkflowRepresentation rep) {
+        validateField(rep, "name", rep.getName());
+        //TODO: validate event and resource conditions (`on` and `if` properties) using the providers with a custom evaluator that calls validate on
+        // each condition provider used in the expression.
+
+        // if a workflow has a restart step, at least one of the previous steps must be scheduled to prevent an infinite loop of immediate executions
+        List<WorkflowStepRepresentation> steps = ofNullable(rep.getSteps()).orElse(List.of());
```
pruivo, line 298: if `rep.getSteps()` is null or empty, you can return here and avoid further computation.

server-spi-private/src/main/java/org/keycloak/models/workflow/NotifyReaperTransaction.java
```diff
@@ -0,0 +1,30 @@
+package org.keycloak.models.workflow;
+
+import java.util.List;
+
+import org.keycloak.models.AbstractKeycloakTransaction;
+
+public class NotifyReaperTransaction extends AbstractKeycloakTransaction {
```
pruivo: unused

model/jpa/src/main/java/org/keycloak/models/workflow/RestartWorkflowStepProvider.java
pruivo, line 1: It is stateless, can be a singleton.

model/jpa/src/main/java/org/keycloak/models/workflow/RestartWorkflowTask.java
```diff
@@ -0,0 +1,26 @@
...
+        super(context, false);
+        this.context = context;
+    }
+
+    @Override
+    public void run(KeycloakSession session) {
+        Workflow workflow = context.getWorkflow();
+        String resourceId = context.getResourceId();
+        String executionId = context.getExecutionId();
+        context.resetState();
```
pruivo: is it intentional? The class is named "restart", but you're invoking "resetState".

pruivo (changes requested):

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
pedroigor: I'll leave as is because of the log message. Better here than in the predicate.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
pedroigor: I'll leave as is for now.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
pedroigor: I'll leave as is for now.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
pedroigor: I'll leave as is for now.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
pedroigor: I'll leave as is for now.

core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
```diff
@@ -195,8 +191,11 @@ public Builder withConfig(String key, List<String> values) {
         }
 
         public Builder name(String name) {
+            WorkflowRepresentation representation = new WorkflowRepresentation();
```
vramik, line 194: Shouldn't the instantiation be moved to the constructor? Before the changes the `of` was called from line 33 and therefore the representation was always instantiated. 

With new version of the code it seems that if you do not call `name` as first in the line it may cause issues. I can fix that in follow up PR is preferred.

vramik (commented):
I went through the PR only briefly for now, changes are really big. I have one minor comment related to builder. If you want to proceed now, please go ahead, otherwise I will take more closer look tomorrow.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
pedroigor: I'm creating a separate issue for exposing metrics for workflow execution.

core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
pedroigor: Can you check the update?

vramik: Thanks, I wasn't clear enough, sorry. My point is that if you call 

```java
WorkflowRepresentation.create()
                .name("test-workflow")
                .withSteps(
                        WorkflowStepRepresentation.create().of(NotifyUserStepProviderFactory.ID)
                                .after(Duration.ofDays(5))
                                .build(),
                        WorkflowStepRepresentation.create().of(DisableUserStepProviderFactory.ID)
                                .after(Duration.ofDays(10))
                                .build()
                ).build();
```

or 

```java
WorkflowRepresentation.create()
                .withSteps(
                        WorkflowStepRepresentation.create().of(NotifyUserStepProviderFactory.ID)
                                .after(Duration.ofDays(5))
                                .build(),
                        WorkflowStepRepresentation.create().of(DisableUserStepProviderFactory.ID)
                                .after(Duration.ofDays(10))
                                .build()
                )
                .name("test-workflow")
                .build();
```

the result will be different as `name` creates new `WorkflowRepresentation` and new `Builder`. On the other hand all tests we have now does call `name` as first call in the line of fluent api and therefore the result will be expected. Maybe we can add a javadoc to the builder stating that the `name` has to be called first or come up with something better later. As far as I am concerned we can keep it as is for now and improve it in a follow-up.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProviderFactory.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public String getId() {
+        return ID;
+    }
+
+    @Override
+    public DefaultWorkflowProvider create(KeycloakSession session, ComponentModel model) {
+        return new DefaultWorkflowProvider(session, executor);
+    }
```
vramik, line 31: It seems that `WorkflowProviderFactory` should not extend `ComponentFactory` at all. The `ComponentModel` is ignored anyway. Can we extend `ProviderFactory` directly?

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
vramik, line 1: I agree with @pruivo  that we should consider move some of the logic from this class elsewhere, I think we should do it in follow-up task if we want to unblock other work which is waiting for this one to be merged.

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/ConditionEvaluator.java
```diff
@@ -85,4 +85,18 @@ private String unEscapeParameter(String rawText) {
...
+    private WorkflowConditionProviderFactory<WorkflowConditionProvider> getConditionProviderFactory(String providerId) {
+        KeycloakSessionFactory sessionFactory = session.getKeycloakSessionFactory();
+        WorkflowConditionProviderFactory<WorkflowConditionProvider> providerFactory = (WorkflowConditionProviderFactory<WorkflowConditionProvider>) sessionFactory.getProviderFactory(WorkflowConditionProvider.class, providerId);
+
+        if (providerFactory == null) {
+            throw new WorkflowInvalidStateException("Could not find condition provider: " + providerId);
+        }
+
+        return providerFactory;
+    }
```
vramik: This block is pretty much the same as in `ProdicateEvaluator`, we might want to consider to make it generic and move it to its parent, I'm fine to do it as follow-up.

vramik (commented):
Thanks @pedroigor for preparing this PR, it's really big. I've added few comments for now.

services/src/main/java/org/keycloak/models/workflow/WorkflowEventListener.java
```diff
@@ -21,6 +25,7 @@
 public class WorkflowEventListener implements EventListenerProvider, ProviderEventListener {
 
     private final KeycloakSession session;
+    private final Map<String, List<String>> events = new HashMap<>();
```
pruivo: unused, class can be rolled back.

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
```diff
@@ -0,0 +1,104 @@
...
+    private final Runnable task;
+    private final WorkflowExecutor executor;
+    private final String id;
+    private CompletableFuture<Void> future;
+    private long startTime;
+    private AtomicReference<Thread> thread;
+
+    WorkflowTask(WorkflowExecutor executor, Runnable task) {
+        Objects.requireNonNull(executor, "executor");
+        Objects.requireNonNull(executor, "task");
```
pruivo: `Objects.requireNonNull(task, "task");`

server-spi-private/src/main/java/org/keycloak/models/workflow/WorkflowTask.java
```diff
@@ -0,0 +1,104 @@
...
+    public void cancel() {
+        if (future != null) {
+            future.cancel(true);
+            if (thread.get() != null) {
+                thread.get().interrupt();
+            }
+        }
+    }
+
+    void dispose() {
```
pruivo: unused

model/jpa/src/main/java/org/keycloak/models/workflow/ScheduleWorkflowTask.java
```diff
@@ -0,0 +1,58 @@
...
+
+final class ScheduleWorkflowTask extends WorkflowTransactionalTask {
+
+    private static final Logger log = Logger.getLogger(ScheduleWorkflowTask.class);
+
+    private final DefaultWorkflowExecutionContext workflowContext;
+
+    ScheduleWorkflowTask(DefaultWorkflowExecutionContext context) {
+        super(context.getSession());
+        Objects.requireNonNull(context, "Context must not be null when scheduling workflow");
```
pruivo: unnecessary check; if null, it will throw a NPE in the line above.

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/AddRequiredActionTest.java
```diff
@@ -4,45 +4,60 @@
...
                 ).build()).close();
 
-        managedRealm.admin().users().create(UserConfigBuilder.create().username("test").build()).close();
+        managedRealm.admin().users().create(UserConfigBuilder.create().username("myuser").build()).close();
+
+        List< UserRepresentation> users;
+        UserRepresentation userRepresentation;
+
+        int i = 0;
+        while (true) {
```
pruivo: you could use the Awaitility library (change the timeout and poll interval to the desired values)
```java
        Awaitility.await()
                .timeout(Duration.ofSeconds(30))
                .pollInterval(Duration.ofSeconds(1))
                .untilAsserted(() -> {
                    var users = managedRealm.admin().users().search("myuser");
                    assertThat(users, hasSize(1));
                    var userRepresentation = users.get(0);
                    Assertions.assertTrue(userRepresentation.getRequiredActions() != null && !userRepresentation.getRequiredActions().isEmpty());
                });
```

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/BrokeredUserSessionRefreshTimeWorkflowTest.java
```diff
@@ -150,6 +147,23 @@ public void testInvalidateWorkflowOnIdentityProviderRemoval() {
         // create new user - it will trigger an activation event and therefore should disable the workflow
         consumerRealm.admin().users().create(UserConfigBuilder.create().username("test").build()).close();
 
+        int i = 0;
+        while (true) {
```
pruivo: same

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/GroupMembershipJoinWorkflowTest.java
```diff
@@ -149,6 +121,23 @@ public void testRemoveAssociatedGroup() {
         // create new user - it will trigger an activation event and therefore should disable the workflow
         managedRealm.admin().users().create(UserConfigBuilder.create().username("test").build()).close();
 
+        int i = 0;
+        while (true) {
```
pruivo: another :)

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProviderFactory.java
pedroigor: If it's not a component factory, we won't be able to create components due to validations on the provider. The more we move forward, the more it seems we should use specific tables for workflows :)

core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
pedroigor: Now it looks like `WorkflowRepresentation.withName("name")...`. Works for you?

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
pedroigor: It makes sense. I moved quite a lot already. But I would prefer doing more in a separate PR, mainly if we are changing the API to manage workflows directly from the `Workflow` instances.

core/src/main/java/org/keycloak/representations/workflows/WorkflowRepresentation.java
vramik: It does

model/jpa/src/main/java/org/keycloak/models/workflow/conditions/expression/ConditionEvaluator.java
pedroigor: Added a `Workflows` utility class.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProviderFactory.java
vramik: Created https://github.com/keycloak/keycloak/issues/44032

pruivo (approved):

ssilvert (approved):
Approving as a maintainer.  This approval is based on the other approvals and the fact that these changes are low risk due to it being mainly isolated to the experimental workflow feature.

PR #43836, opened 2025-10-30
fix(useHash): correctly extract hash from pushState url

### LINKED ISSUE #43835: useHash hook does not correctly extract hash from pushState URL
### Before reporting an issue

### Area

admin/ui

### Describe the bug

The current implementation of the `useHash` hook assumes that the third argument (`url`) passed to `window.history.pushState` is always a hash string (e.g., `#settings`). However, according to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState), this argument can be a full URL, a path, or a path with a hash (e.g., `/profile#tab2` or `https://example.com/#abc`).

### Version

latest

### Regression

### Expected behavior

The hook should always extract and return only the hash fragment (without the `#`), regardless of the format of the URL passed to `pushState`.

### Actual behavior

When calling `pushState` with `/profile#tab2`, the `useHash` hook sets the hash state to `profile#tab2` (i.e., everything after the first character), instead of just the hash fragment.

For example:
- `pushState({}, '', '/profile#tab2')` → `useHash` returns `profile#tab2`
- `pushState({}, '', 'https://example.com/#abc')` → `useHash` returns `ttps://example.com/#abc`

### How to Reproduce?

1. Call `window.history.pushState({}, '', '/profile#tab2')` in the browser console.
2. Observe that the `useHash` hook sets the state to `profile#tab2` instead of just `tab2`.

## Suggested solution

Use the `URL` constructor to reliably extract the hash fragment from the URL argument.

## Additional context

This can cause bugs in navigation and state management in the admin UI if the URL format changes.

basedest:
Closes #43835

Feedback and suggestions are welcome!

### DIFF
--- a/js/apps/admin-ui/src/context/realm-context/useHash.tsx
+++ b/js/apps/admin-ui/src/context/realm-context/useHash.tsx
@@ -7,7 +7,8 @@ export const useHash = () => {
     const orgPushState = window.history.pushState;
     window.history.pushState = new Proxy(window.history.pushState, {
       apply: (func, target, args) => {
-        setHash(args[2].substring(1));
+        const url = new URL(args[2], window.location.origin);
+        setHash(url.hash.substring(1));
         return Reflect.apply(func, target, args);
       },
     });


mposolda:
@ssilvert @jonkoops Anyone of you available to review this one please?

ssilvert (approved):

jonkoops:
@ssilvert @edewit @basedest a couple of notes:

1. The initial value is incorrect, `location.hash` will always include the `#` if present, whereas subsequent updates do a `.substring(1)` to remove the hash, this should be the same for both.
2. On every call of this hook `decodeURIComponent()` is called over the hash, which should only be done if the hash actually changes value for performance reasons.
3. This hook is monkey patching internal browser APIs every time it is set up, which is bad practice and can lead to unforeseen bugs.


Can we please get rid of this terrible monkey patching of internal browser methods and use something like the [`popstate` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event) instead? I am pretty sure this code can be re-written to (untested):

```js
import { useEffect, useState } from "react";

export const useHash = () => {
  const getHash = () => decodeURIComponent(location.hash.substring(1));
  const [hash, setHash] = useState(getHash);

  useEffect(() => {
    const updateHash = () => setHash(getHash);
    window.addEventListener("popstate", updateHash);
    return () => window.removeEventListener("popstate", updateHash);
  }, []);
 
  return hash;
};
```

edewit:
Unfortunately the `popstate` event is not called when the router is using the hash, that's why we need this monkey patch. And this hook is only called when the hash changes so... but I agree it would be nicer to make a function that calls decodeURIComponent and substring

jonkoops:
Ah! That is unfortunate, from the API docs I was under the impression that `popstate` would also fire for those events. It looks like the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API) could be used here instead. However, it is not yet supported by Firefox and Safari, so it would still be required to keep a fallback around for the time being:

```ts
import { useEffect, useState } from "react";

export const useHash = () => {
  const getHash = () => decodeURIComponent(location.hash.substring(1));
  const [hash, setHash] = useState(getHash);

  useEffect(() => {
    const updateHash = () => setHash(getHash);

    // For browsers that support the Navigation API (https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API)
    if ("navigation" in globalThis && "onnavigatesuccess" in navigation) {
      navigation.addEventListener("navigatesuccess", updateHash);
      return () => navigation.removeEventListener("navigatesuccess", updateHash);
    }

    // For browsers that do not support the Navigation API, can be removed once it is baseline supported.
    const original = history.pushState;

    history.pushState = new Proxy(history.pushState, {
      apply(...args) {
        const result = Reflect.apply(...args);
        updateHash();
        return result;
      },
    });

    return () => {
      history.pushState = original;
    };
  }, []);
 
  return hash;
};
```

This way, the more modern browsers will refrain from patching the built-in APIs, and in the future the fallback can be removed entirely.

PR #43857, opened 2025-10-30
Adding ML-DSA support

### LINKED ISSUE #43684: Add ML-DSA `SignatureProviderFactory` and `ClientSignatureVerifierProviderFactory‎`
### Description

Allow creation and verification of ML-DSA token signatures as in [this RFC draft](https://www.ietf.org/archive/id/draft-ietf-cose-dilithium-10.html)

### Value Proposition

For now, this does not add anything beyond opaque tokens, but the long-term goal is to allow for quantum-secure self-contained tokens. This is an intermediate step towards this goal.

### Goals

Allow users to select ML-DSA-* as "Default Signature Algorithm" under Realm Settings → Tokens

### Non-Goals

Exposing the public key in `./well-known/jwks` is out of scope, as from previous discussions it is my understanding this can not easily be achieved.

### Discussion

https://github.com/keycloak/keycloak/discussions/40496

### Notes

This feature should be considered experimental until the final RFC is published.

Jonaka3385:
Adding providers for ML-DSA and AKP Public JWK. 

Contains Server, Client and Signature Providers; Contains KeyProviders; Contains AKPPublicJWK; Currently works with BouncyCastlePQCProvider and Dilithium, will be changed to BouncyCastleProvider and ML-DSA when unknown error is resolved; Tests will be added; webauthn4j does not support ML-DSA at the moment. 

Closes #43684


stianst (commented):
Nice work, will try to find time to look at this more in detail next week.

I suspect we need to descope the initial PR a bit though, especially since this will need tests added as well.

Honestly I'd probably start with just supporting one ML-DSA variant, let's say 44, not touch SAML, WebAuthN, etc.

That will reduce the size of the initial PR signficantly.

stianst:
Why do you need BouncyCastlePQCProvider? Java 25 onwards support ML-DSA.

Jonaka3385:
> Why do you need BouncyCastlePQCProvider? Java 25 onwards support ML-DSA.

I thought it would be good to have some backwards compatibility. With BouncyCastle, you can use Java 8 or later.

core/src/main/java/org/keycloak/crypto/Algorithm.java
```diff
@@ -42,6 +42,11 @@ public interface Algorithm {
     String Ed25519 = "Ed25519";
     String Ed448 = "Ed448";
 
+    /* ML-DSA */
+    String MLDSA44 = "Dilithium2";
```
stianst:
```suggestion
    String MLDSA44 = "ML-DSA-44";
```

core/src/main/java/org/keycloak/crypto/Algorithm.java
```diff
@@ -42,6 +42,11 @@ public interface Algorithm {
...
+    String MLDSA65 = "Dilithium3";
```
stianst:
```suggestion
    String MLDSA65 = "ML-DSA-65";
```

core/src/main/java/org/keycloak/crypto/Algorithm.java
```diff
@@ -42,6 +42,11 @@ public interface Algorithm {
...
+    String MLDSA87 = "Dilithium5";
```
stianst:
```suggestion
    String MLDSA87 = "ML-DSA-87";
```

core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
```diff
@@ -40,6 +40,11 @@ public class JavaAlgorithm {
     public static final String SHA512 = "SHA-512";
     public static final String SHAKE256 = "SHAKE256";
 
+    /* ML-DSA */
+    public static final String MLDSA44 = "Dilithium2";
```
stianst:
```suggestion
    public static final String MLDSA44 = "ML-DSA-44";
```

core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
```diff
@@ -40,6 +40,11 @@ public class JavaAlgorithm {
...
+    public static final String MLDSA65 = "Dilithium3";
```
stianst:
```suggestion
    public static final String MLDSA65 = "ML-DSA-65";
```

core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
```diff
@@ -40,6 +40,11 @@ public class JavaAlgorithm {
...
+    public static final String MLDSA87 = "Dilithium5";
```
stianst:
```suggestion
    public static final String MLDSA87 = "ML-DSA-87";
```

crypto/default/src/main/java/org/keycloak/crypto/def/DefaultCryptoProvider.java
```diff
@@ -63,6 +64,9 @@ public DefaultCryptoProvider() {
         providers.put(CryptoConstants.ECDH_ES_A128KW, new BCEcdhEsAlgorithmProvider());
         providers.put(CryptoConstants.ECDH_ES_A192KW, new BCEcdhEsAlgorithmProvider());
         providers.put(CryptoConstants.ECDH_ES_A256KW, new BCEcdhEsAlgorithmProvider());
+        providers.put(CryptoConstants.MLDSA44, new BouncyCastlePQCProvider());
+        providers.put(CryptoConstants.MLDSA65, new BouncyCastlePQCProvider());
+        providers.put(CryptoConstants.MLDSA87, new BouncyCastlePQCProvider());
```
stianst: Remove, this is not needed if you change to `ML-DSA-**` as those are available in BouncyCastleProvider

common/src/main/java/org/keycloak/common/crypto/CryptoConstants.java
```diff
@@ -15,6 +15,11 @@ public class CryptoConstants {
     public static final String ECDH_ES_A192KW = "ECDH-ES+A192KW";
     public static final String ECDH_ES_A256KW = "ECDH-ES+A256KW";
 
+    /* ML-DSA */
+    public static final String MLDSA44 = "Dilithium2";
+    public static final String MLDSA65 = "Dilithium3";
+    public static final String MLDSA87 = "Dilithium5";
+
```
stianst: Remove, not needed, see comment in `DefaultCryptoProvider`

core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
```diff
@@ -159,6 +178,10 @@ public static boolean isEddsaJavaAlgorithm(String algorithm) {
         return getJavaAlgorithm(algorithm).contains("Ed");
     }
 
+    public static boolean isMldsaJavaAlgorithm(String algorithm) {
+        return getJavaAlgorithm(algorithm).contains("Dilithium");
```
stianst:
```suggestion
        return getJavaAlgorithm(algorithm).startsWith("ML-DSA-");
```

core/src/main/java/org/keycloak/jose/jws/Algorithm.java
```diff
@@ -42,7 +42,13 @@ public enum Algorithm {
...
     EdDSA(AlgorithmType.EDDSA, null),
     Ed25519(AlgorithmType.EDDSA, null),
-    Ed448(AlgorithmType.EDDSA, null)
+    Ed448(AlgorithmType.EDDSA, null),
+    MLDSA44(AlgorithmType.MLDSA, null),
+    MLDSA65(AlgorithmType.MLDSA, null),
+    MLDSA87(AlgorithmType.MLDSA, null),
+    Dilithium2(AlgorithmType.MLDSA, null),
+    Dilithium3(AlgorithmType.MLDSA, null),
+    Dilithium5(AlgorithmType.MLDSA, null),
```
stianst: Remove Dilithum*

docs/guides/server/fips.adoc
stianst, line 1: Remove changes to this file

js/libs/keycloak-admin-client/package-lock.json
stianst, line 1: Remove

js/package-lock.json
stianst, line 1: Remove

js/themes-vendor/package-lock.json
stianst, line 1: Remove

quarkus/dist/src/main/content/bin/kc.sh
stianst, line 1: Remove

services/src/main/java/org/keycloak/authentication/requiredactions/WebAuthnRegister.java
stianst, line 1: I'd remove changes here; and add when needed and webauthn support for ML-DSA is added

services/src/main/java/org/keycloak/keys/AbstractMldsaKeyProviderFactory.java
```diff
@@ -0,0 +1,68 @@
...
+                .checkBoolean(Attributes.ACTIVE_PROPERTY, false);
+    }
+
+    public static KeyPair generateMldsaKeyPair(String algorithm) {
+        if (!JavaAlgorithm.isMldsaJavaAlgorithm(algorithm)) {
+            throw new IllegalArgumentException(algorithm + " is not supported");
+        }
+        try {
+            KeyPairGenerator keyGen = KeyPairGenerator.getInstance(algorithm);
+            return keyGen.generateKeyPair();
```
stianst, line 56: Should use `new NamedParameterSpec("ML-DSA-44")` etc based on the algorithm to make sure keys are generated with the right properties

stianst (changes requested):
Done a partial review for now; two things we'll need is this needs to be wrapped in a feature and marked as experimental. Second is we need tests.

In general this looks like a very good start.

stianst:
In terms of testing and reviews it may be better to break this up slightly, so we can start getting things merged.

I would probably have broken this down into the following separate PRs:

* JWK support
* Keys
* SignatureProviderFactory
* ClientSignatureVerifierProviderFactory

Let me know what you think about that.

Jonaka3385:
Thanks for the partial review. I’ve now changed Dilithium to ML-DSA, added some tests, and introduced a feature flag that hides ML-DSA in the Token Algorithm dropdown menu (let me know if that’s fine).
There are some tests (testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/ClientAuthMLDSASignedJWTTest.java) that result in an HTTP error. As far as I know, this should be the only issue, but I haven’t figured out how to fix it yet.
Regarding the PR, I would only split it in 3 (JWK, keys, and then all SignatureProviders), but that's up to you.

core/src/main/java/org/keycloak/jose/jwk/JWKBuilder.java
```diff
@@ -169,4 +169,36 @@ public JWK okp(Key key) {
...
         return EdEC_UTILS.okp(kid, algorithm, key, keyUse);
     }
+
+    public JWK akp(Key key) {
+        return akp(key, DEFAULT_PUBLIC_KEY_USE);
+    }
+
+    public JWK akp(Key key, KeyUse keyUse) {
+        AKPPublicJWK k = new AKPPublicJWK();
+        byte[] encodedKey = key.getEncoded();
```
stianst: This is not correct, but have no idea what the correct way or how to do it :/

core/src/main/java/org/keycloak/jose/jwk/JWKParser.java
```diff
@@ -142,8 +145,28 @@ private static PublicKey createRSAPublicKey(JsonNode jwk) {
...
+        if (alg == null) {
+            throw new IllegalArgumentException("Missing 'alg' parameter in AKP-JWK");
+        }
+        String pub = jwk.path(AKPPublicJWK.PUB).asText(null);
+        if (pub == null) {
+            throw new IllegalArgumentException("Missing 'pub' parameter in AKP-JWK");
+        }
+
+        byte[] decodedKey = Base64Url.decode(pub);
+        X509EncodedKeySpec keySpec = new X509EncodedKeySpec(decodedKey);
```
stianst: This is not correct, but have no idea what the correct way or how to do it :/

stianst:
Sent a PR to add AKP support for JWK parsing/building here:

https://github.com/keycloak/keycloak/pull/44203

Would be great if you could take a look; and as a next-step create a separate PR with the keys. One thing to note there is I migrated the keys tests to the new tests suite, see https://github.com/keycloak/keycloak/pull/44130

stianst:
Closing this as it's stale and far from ready

PR #43848, opened 2025-10-30
[quarkus-next] Removed exception escaped OTel attribute

mabartos:
- Closes #43845

Test for quarkus next passed: https://github.com/keycloak/keycloak/pull/43847

PR #43859, opened 2025-10-30
Migrate the OTelProvider test to the new framework

### LINKED ISSUE #43858: Migrate the OTelProvider test to the new framework
### Description

The current test `OTelTracingProviderTest` uses many workarounds to start the server with tracing enabled and it is very inefficient. We should use the new framework.

mabartos:
Closes #43858

It accommodates changes from https://github.com/keycloak/keycloak/pull/43848

### DIFF
--- a/tests/base/src/test/java/org/keycloak/tests/tracing/TracingProviderTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/tracing/TracingProviderTest.java
@@ -1,21 +1,4 @@
-/*
- * Copyright 2024 Red Hat, Inc. and/or its affiliates
- * and other contributors as indicated by the @author tags.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-package org.keycloak.testsuite.tracing;
+package org.keycloak.tests.tracing;
 
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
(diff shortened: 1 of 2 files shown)


mabartos:
@vaceksimon @lhanusov Could you please check it? Thanks

tests/base/src/test/java/org/keycloak/tests/tracing/TracingProviderTest.java
```diff
@@ -1,105 +1,44 @@
...
-            container.restartServer();
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+    public static class ServerConfigWithTracing implements KeycloakServerConfig {
+        @Override
+        public KeycloakServerConfigBuilder configure(KeycloakServerConfigBuilder config) {
+            return config
+                    .option("tracing-enabled", "true")
+                    .option("log-level", "INFO,org.keycloak.quarkus.runtime.tracing:debug");
         }
```
vaceksimon: By convention, we put the class with server config to the bottom of the test class

tests/base/src/test/java/org/keycloak/tests/tracing/TracingProviderTest.java
```diff
@@ -372,7 +311,7 @@ public void traceMultipleAlwaysEnded() {
     }
 
     void runOnServer(TracingConsumer tracing) {
-        getTestingClient().server(TEST_REALM_NAME).run(session -> {
+        runOnServer.run(session -> {
```
vaceksimon: It's confusing that there is a RunOnServerClient instance and a method both named `runOnServer`. `runOnServerTracing(...)` or something may make it clearer, wdyt?

vaceksimon (changes requested):
Only nitpicks, looks good

tests/base/src/test/java/org/keycloak/tests/tracing/TracingProviderTest.java
mabartos: @vaceksimon Yes, I'm kinda aware of that. However, it makes more sense to me to have it visible on the top to see what exact config is used class-wide. Then, when there's 10+ test cases, the config might be deeply hidden (without using IDE f.e.) - just a consideration.

But whatever works for you to have these tests unified - I'll update it.

tests/base/src/test/java/org/keycloak/tests/tracing/TracingProviderTest.java
mabartos: If it's confusing to you, there's a big chance it's confusing to other people. I thought the named lambda param is sufficient (`runOnServer(tracing ->`), but renaming works to me.

mabartos:
@vaceksimon Thanks, updated.

vaceksimon (approved):

mabartos (commented):
@vaceksimon JFYI - I see that my responses to your comments were not submited. But I updated it based on the feedback.

ahus1 (approved):

PR #43860, opened 2025-10-30
Separate HOW_TO_RUN.md file for the new testsuite

mabartos:
IMHO, we should have more visible the guide on how to run tests. 

I'd like to see the HOW_TO_RUN in the `tests` module, not the `test-framework` module.

I think it improves the guidance of the tests execution.

@vaceksimon @lhanusov @stianst WDYT?


tests/HOW_TO_RUN.md
```diff
@@ -0,0 +1,157 @@
...
+* Browser type
+
+There are a few options on how to configure the test framework, with the following ordinal:
+
+* System properties
+* Environment variables
+* `.env.test` file in the project hierarchy
+* A properties file specified with `kc.test.config` system property or `KC_TEST_CONFIG` environment variable
+
+For more details about the test framework, see the [README](../test-framework/README.md).
```
mabartos: Content copied to the HOW_TO_RUN.md and added this little reference to the test framework.

vaceksimon (approved):
I think that's a good idea. Thanks Martin

stianst (changes requested):
I like the idea of separating out the how to run tests instructions into a separate file; but it should still be in the test-framework as we already have multiple test modules + it's intended that folks can use the test-framework for extensions as well.

mabartos:
> I like the idea of separating out the how to run tests instructions into a separate file; but it should still be in the test-framework as we already have multiple test modules + it's intended that folks can use the test-framework for extensions as well.

Works to me as the main intention is to have the HOW_TO_RUN file separated and more visible.

vaceksimon (approved):

mabartos:
@stianst Ready for the review.

stianst (approved):

PR #43861, opened 2025-10-30
fix(admin-ui): correct default value for backchannel logout session

### LINKED ISSUE #43674: Setting "Backchannel logout session required" displayed incorrectly
### Before reporting an issue

### Area

admin/ui

### Describe the bug

The setting "Backchannel logout session required" might be displayed as active while it is actually inactive.

### Version

25.0.6, 26.4.0

### Regression

### Expected behavior

The setting "Backchannel logout session required" should be displayed with the value set.

### Actual behavior

The setting "Backchannel logout session required" is displayed as active after importing a realm, in which for the client in json following line is missing:
"backchannel.logout.session.required" : "true",

### How to Reproduce?

Import a new Realm with a client, that doesn't have the attribute "backchannel.logout.session.required"

### Anything else?

_No response_

M313K:
Fixes #43674

### DIFF
--- a/js/apps/admin-ui/src/clients/add/LogoutPanel.tsx
+++ b/js/apps/admin-ui/src/clients/add/LogoutPanel.tsx
@@ -122,7 +122,7 @@ export const LogoutPanel = ({
               name={convertAttributeNameToForm<FormFields>(
                 "attributes.backchannel.logout.session.required",
               )}
-              defaultValue="true"
+              defaultValue="false"
               control={control}
               render={({ field }) => (
                 <Switch


ahus1:
@keycloak/core-clients - pinging the core-clients team as this PR relates to an help-wanted issue of yours.

rmartinc (approved):
I have checked the default value in code is false ([here](https://github.com/keycloak/keycloak/blob/26.4.2/services/src/main/java/org/keycloak/protocol/oidc/OIDCAdvancedConfigWrapper.java#L375-L378)). So false is the correct default value.

mposolda (approved):
@M313K @rmartinc Thanks for the PR and review!

PR #43880, opened 2025-10-31
Ensure the logout endpoint removes the authentication session (26.4)

### LINKED ISSUE #43853: Ensure the logout endpoint removes the authentication session
### Before reporting an issue

### Area

oidc

### Describe the bug

The current implementation of the logout endpoint, when passing the parameter `id_token_hint` without any session cookie, logs out the user session but does not take into account that an authentication session can be still present. Since the improvements to manage the already logged in issue (https://github.com/keycloak/keycloak/issues/23879), the authentication session can be maintained a few seconds after authentication to allow other tabs to automatically log in. But if, for whatever reason, a tab was closed or does not refresh, the auth session will be present a short period (60s). During those 60s the logout can remove the session but not the associated auth session that can be a problem for a subsequent login.

This is a corner case that only happens in those conditions and for 60s. But we need to also remove the authentication session when present in that logout (like it's done when using refresh_token for example).

### Version

26.4.2

### Regression

### Expected behavior

The authentication session should be removed in that situation.

### Actual behavior

The authentication session remains for 60s by default and can re-use the same ID.

### How to Reproduce?

See description.

### Anything else?

_No response_

rmartinc:
Closes #43853

PR:             https://github.com/keycloak/keycloak/pull/43872
Commit:         https://github.com/keycloak/keycloak/commit/3b3adcf1e4819bf63e08269142459f747c31cb37
PR branch:      backport-43872-26.4
Target branch:  https://github.com/keycloak/keycloak/tree/release/26.4

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/endpoints/LogoutEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/endpoints/LogoutEndpoint.java
@@ -232,28 +232,6 @@ public class LogoutEndpoint {
             }
         }
 
-        AuthenticationSessionModel logoutSession = AuthenticationManager.createOrJoinLogoutSession(session, realm, new AuthenticationSessionManager(session), null, true, true);
-        session.getContext().setAuthenticationSession(logoutSession);
-        if (uiLocales != null) {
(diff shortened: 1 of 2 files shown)


ahus1 (approved):
Approving based on parent PR review

PR #43884, opened 2025-10-31
Allow GET and PUT methods using application/yaml media type

### LINKED ISSUE #42687: [RLM] - Ability to define workflows with YAML

pedroigor:
Closes #42687


ssilvert (dismissed):

pedroigor:
@ssilvert Closing in favor of https://github.com/keycloak/keycloak/pull/43868. The same changes here were applied there.

PR #43890, opened 2025-11-01
Make client and IDP required when using federated client authentication

### LINKED ISSUE #43889: Make client and IDP required when using federated client authentication
### Description

Currently both the IDP alias and the issuer are not shown as required, although they are required.

ahus1:
Closes #43889


ahus1:
@stianst - please review when you have the time. Thanks!

stianst (approved):

PR #43892, opened 2025-11-03
Add version column to user_entity

### LINKED ISSUE #40929: Duplicates in unmanaged attributes
### Before reporting an issue

### Area

admin/api

### Describe the bug

Identity has multiple duplicated unmanaged attributes, same key, same value.

### Version

26.3.0

### Regression

### Expected behavior

When we call REST API for user update many times, where we send attributes, we expect that the user will have only the attributes that we send.

Image

### Actual behavior

When we call REST API for user update many times. Attributes are duplicated.

Image

### How to Reproduce?

- Set up Keycloak 26.3.0 via docker-composue with PostgreDB. Keycloak start with "start-dev", "--debug"
- Enable Unmanaged Attributes in Realm settings.
- Create a new user test-user.
- Obtain a token from Keycloak with manage-realm claim. I use OAuth2.0 authorisation via the admin-cli client and the admin user.
- Set the token to attached _force_attribute_update_KC_, attribute token.
- Set the test-users ID to _force_attribute_update_KC,_ attribute userId.
- Remove suffix .txt from _force_attribute_update_KC.js.txt_
- Run script _k6 run --insecure-skip-tls-verify .\force_attribute_update_KC.js_

[force_attribute_update_KC.js.txt](https://github.com/user-attachments/files/21056861/force_attribute_update_KC.js.txt)

### Anything else?

_No response_

keshavprashantdeshpande:
Closes #40929

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/entities/UserEntity.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/entities/UserEntity.java
@@ -34,6 +34,7 @@ import jakarta.persistence.NamedQuery;
 import jakarta.persistence.OneToMany;
 import jakarta.persistence.Table;
 import jakarta.persistence.UniqueConstraint;
+import jakarta.persistence.Version;
 import java.util.Collection;
 import java.util.LinkedList;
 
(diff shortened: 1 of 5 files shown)


stianst:
We're not going to add a version column; we may consider an updated column though

PR #43918, opened 2025-11-03
Print a warning on duplicate options

### LINKED ISSUE #43604: Warn or error on duplicate options
### Description

The cli and keycloak handling both allow for duplicate keys to be used for kc. options - the last appearance is used, while the previous ones are silent ignored.

### Value Proposition

To warn or error when duplicate options appear.

### Goals

To make users aware of a potential misconfiguration.

### Non-Goals

N/A

### Discussion

_No response_

### Notes

_No response_

mabartos:
- Closes #43604

This can be easily done for CLI, but it might be challenging to do for other config sources - especially properties-based. I tried to find some general approach for all config sources, but AFAIK it's not  possible now. For 'keycloak.conf', or `quarkus.properties`, we can't simply find the duplicated values as the nature of Java's Properties does not support duplicates - it just uses the last value. 

For that case, this approach checks only duplicates on CLI side and the handle method is generalized and prepared for duplications in other config sources. 

I tried to handle duplicates directly in the CLI config source (ConfigArgsConfigSource), but it brings some other problems as printing it multiple times(f.e. during start-dev that requires augmentation, it's printed twice,..).

I think this approach is quite straighforward, and even the static state for duplicates should not cause any issue as the map is always recreated when the config source is created. 

@shawkins @vmuzikar @Pepo48 WDYT?

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
@@ -225,6 +225,7 @@ public class Picocli {
         if (cliArgs.contains(OPTIMIZED_BUILD_OPTION_LONG) && !wasBuildEverRun()) {
             throw new PropertyException(Messages.optimizedUsedForFirstStartup());
         }
+        warnOnDuplicatedOptionsInCli();
 
         IncludeOptions options = getIncludeOptions(cliArgs, abstractCommand, abstractCommand.getName());
 
@@ -950,6 +951,27 @@ public class Picocli {
         Environment.setProfile(profile);
         if (!cliArgs.contains(HelpAllMixin.HELP_ALL_OPTION)) {
             parsedCommand.ifPresent(PropertyMappers::sanitizeDisabledMappers);
+        }
+    }
+
+    // Show warning about duplicated options in CLI
+    public void warnOnDuplicatedOptionsInCli() {
+        var duplicatedOptions = ConfigArgsConfigSource.getDuplicatedArgs();
+        if (!duplicatedOptions.isEmpty()) {
+            Function<String, String> removeNamespace = option -> option.startsWith(NS_KEYCLOAK_PREFIX) ? option.substring(NS_KEYCLOAK_PREFIX.length()) : option;
+
+            var polishedOptionNames = duplicatedOptions.entrySet().stream()
+                    .map(option -> "%s (used value '%s')".formatted(removeNamespace.apply(option.getKey()), option.getValue()))
+                    .toList();
+
+            String duplicatedNames;
+            if (duplicatedOptions.size() > 2) {
+                duplicatedNames = "\n" + String.join("\n", polishedOptionNames.stream().map("- %s"::formatted).toList());
+            } else {
+                duplicatedNames = String.join(", ", polishedOptionNames);
+            }
+
+            warn("Duplicated options present in CLI: %s".formatted(duplicatedNames));
         }
     }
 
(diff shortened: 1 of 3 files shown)


shawkins:
> we can't simply find the duplicated values as the nature of Java's Properties does not support duplicates - it just uses the last value.

That's true. It doesn't seem worth going to lower level logic just to valiate duplicates. I wonder then if we shouldn't worry about this with CLI options either. It is easily understandable in the same way - the last occurrance on the CLI wins.

mabartos:
> That's true. It doesn't seem worth going to lower level logic just to valiate duplicates. 

+1

> I wonder then if we shouldn't worry about this with CLI options either. It is easily understandable in the same way - the last occurrance on the CLI wins.

@shawkins I would say for properties-based config sources, checking the last occurance of the property is much more visible than in CLI. Even with the IDE, it usually tells you that you have some duplication in the file. For CLI, it's a different issue as someone can have a deployment with many many CLI options specified and sometimes, it might be very hard to manage it. 

I would print the warning for CLI as the effort of maintaining it is very low and moreover, we reduce the risk of some unexpected issues in customers deployments (which may in some cases save even the whole business :)).

shawkins:
> I would print the warning for CLI as the effort of maintaining it is very low and moreover, we reduce the risk of some unexpected issues in customers deployments (which may in some cases save even the whole business :)).

Ok, let's keep it. After #43945 I think this logic can just be done inline in Picocli - it relates to the TODO about initializing the configargconfigsource from just the unrecognized args.

shawkins (dismissed):
LGTM

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -953,4 +954,25 @@ public void initConfig(List<String> cliArgs, AbstractCommand command) {
...
     }
 
+    // Show warning about duplicated options in CLI
+    public void warnOnDuplicatedOptionsInCli() {
+        var duplicatedOptions = ConfigArgsConfigSource.getDuplicatedArgs();
+        if (!duplicatedOptions.isEmpty()) {
+            Function<String, String> removeNamespace = option -> option.startsWith(NS_KEYCLOAK_PREFIX) ? option.substring(NS_KEYCLOAK_PREFIX.length()) : option;
+
+            var polishedOptionNames = duplicatedOptions.entrySet().stream()
+                    .map(option -> "%s (used value '%s')".formatted(removeNamespace.apply(option.getKey()), option.getValue()))
```
vmuzikar: Printing the value might be a problem for security sensitive options. I can imagine an attacker could e.g. append some options to print the sensitive ones. Not sure if that's exploitable but still, I'd like to reflect if the option is masked or not here.

vmuzikar (changes requested):

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: Definitely a legitimate hardening, but not exploitable given the machine level access required.

A user would rightfully complain if invoking `--foo=${bar}` were entered on the command line only to see the value of bar in the log.

I would vote for not including the value at all - just track the cli key (not the kc key).

vmuzikar: +1 for not including the value at all, it's always safer.

mabartos: Yep, nice point. I'll remove the value.

mabartos:
@shawkins @vmuzikar I updated the code:
- no values are printed out
- CLI keys are printed out (not only KC options)
- Even duplicated system properties are printed out
- Some optimization made

Please review once you have a time. Thanks!

shawkins (approved):
LGTM, thanks for the changes @mabartos

vmuzikar (approved):
LGTM, thanks @mabartos!

PR #43928, opened 2025-11-04
JWT Authorization grant should not generate refresh and use transient sessions

rmartinc:
Closes #43799

This PR makes the JWT authorization grant use transient sessions and, therefore, no refresh token is allowed. Spec recommends this behavior. I was brave and the PR also refactors the client credentials grant because the code was repeated.

I have done a second change to ensure the access token generated has a lifespan that is limited to the expiration of the initial token grant. The spec also mentions this, saying that the lifetime of the access token should not exceed the validity period of the assertion by a significant period. Let me know if you prefer something different here.

Tests added. I'm sending a draft because I probably need to update the PR when #43841 is merged.


services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
```diff
@@ -106,12 +106,20 @@ public Response process(Context context) {
...
+                    clientConnection.getRemoteHost(), "authorization-grant", false, null, null, UserSessionModel.SessionPersistenceState.TRANSIENT);
+            event.session(userSession);
             ClientSessionContext clientSessionCtx = TokenManager.attachAuthenticationSession(this.session, userSession, authSession);
-            return createTokenResponse(user, userSession, clientSessionCtx, scopeParam, true, null);
+
+            // create the token response
+            TokenManager.AccessTokenResponseBuilder responseBuilder = createTokenResponseBuilder(user, userSession, clientSessionCtx, scopeParam, null);
+            if (authorizationGrantContext.getJWT().getExp() < responseBuilder.getAccessToken().getExp()) {
+                // if the expiration of the assertion is less than default access token lifespan limit to that value
+                responseBuilder.getAccessToken().exp(authorizationGrantContext.getJWT().getExp());
```
mposolda: I think it is good that we have option of `exp` being limited to the JWT assertion expiration. At the same time, I am like 50/50 whether this should be configurable or not...

In theory, I can imagine that this can lead to worse security (maybe without this behaviour people would use very short JWT assertions of 10 seconds or so, however this will enforce them to use assertions with longer lifespan if they want also access-token to have longer lifespan. This might also mean more items in the single-use cache). However maybe I am overthinking this?

@graziang What is your opinion? 

If you both agree that we can keep like it is in the PR, I am fine with the change. We can revisit later if needed...

mposolda (commented):
@rmartinc Thanks! Added inline comment. WDYT? Curious for the feedback from @graziang as well.

services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
rmartinc: I was also thinking about doing this but with a minimum time, like 1 minute or similar. The spec is open, it just says:

> Authorization servers SHOULD NOT issue access tokens with a lifetime that exceeds the validity period of the assertion by a significant period.

And significant period is not very specific. :smile:

We can even remove this part and just use the access code lifespan configured.

graziang: My vote is to keep the default access token lifespan the same as the realm configuration (or the client configuration if defined) to avoid cases when an assertion with too short exp would result in access tokens that are hard to use because of too short exp.

However, I’d like also to add the option to limit the access token lifespan for the authorization grant as a follow-up. I’m not sure about the best approach, if introduce a switch in the IdP config or in the advanced client settings (e.g. “limit access token expiration to assertion expiration for authorization grant”), or use client policies, or others.

mposolda: @graziang @rmartinc Sounds good to keep the default access token lifespan the same in this PR and then have the follow-up issue for limit the access token lifespan for the authorization grant based on the assertion.

+1 that "Significant period" is not very specific :-) Keycloak has short access-token lifespans by default (or at least it recommends to have short lifespan in the docs and has short default access token lifespan), so IMO we are not breaking the specs when we are just using access token lifespan by default..

Also not sure where to put this as a config? Client policies are most flexible as they may allow different settings for various combinations of clients and IDPs, but in general, they are most complex to setup. Either IDP or client policies are OK. We can create follow-up and figure it later where to configure this. WDYT?

rmartinc: Okis, I will remove this part, this also allows me to remove the extra methods in the base class. I'm preparing the change right now.

rmartinc: Done! Let's see the tests. Moving out of draft now.

graziang (approved):
@rmartinc Thanks!

mposolda (approved):

mposolda:
@rmartinc @graziang Thanks!

PR #43934, opened 2025-11-04
Hide the attribute for allowed IdPs when JWT auth capability is not enabled

### LINKED ISSUE #43925: Hide the attribute for allowed IdPs  when JWT auth capability is not enabled
### Description

Currently the option `oauth2.jwt.authorization.grant.idp` that in the admin console is `Allowed Identity Providers for JWT Authorization Grant` is always presented. I think that it makes sense to just show that option when the  `JWT Authorization Grant` capability is enabled. If not the admin is setting an option that is never used.

rmartinc:
Closes #43925

@mposolda @graziang Just a little change in the admin console to not show the IDP select if the JWT auth grant capability is not enabled. It's just a minor thing and I think that is not related with other changes in the queue. Let's see if there are no conflicts later. :smile:


graziang (approved):
@rmartinc Thanks!

mposolda (approved):
@rmartinc @graziang Thanks!

PR #43943, opened 2025-11-04
Create user session expired event

### LINKED ISSUE #43942: Create user session expired event
### Description

Create the user session expiration event to signal a session reaching its lifespan or max-idle time. In this milestone, only the event will be created, and the remaining tasks will be completed in a follow-up issue.

### Goals

Create a user session expiration event similar to other user events.

### Non-Goals

All the remaining parent goals.

### Notes

_No response_

pruivo:
Closes #43942

Missing:

- [x] Test for `JpaUserSessionPersisterProvider#removeExpired()` as I think it is not exercised in the current testsuite.
- [x] Remove fatal log

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
@@ -56,6 +56,7 @@ abstract class BaseUserSessionExpirationListener {
             if (realm == null) {
                 return;
             }
+            session.getContext().setRealm(realm);
             new EventBuilder(realm, session)
                     .session(userSessionId)
                     .user(userId)
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
@@ -21,6 +21,7 @@ import java.lang.invoke.MethodHandles;
 
 import org.infinispan.util.concurrent.BlockingManager;
 import org.jboss.logging.Logger;
+import org.keycloak.events.Details;
 import org.keycloak.events.EventBuilder;
 import org.keycloak.events.EventType;
 import org.keycloak.models.KeycloakSessionFactory;
@@ -58,7 +59,8 @@ abstract class BaseUserSessionExpirationListener {
             new EventBuilder(realm, session)
                     .session(userSessionId)
                     .user(userId)
-                    .event(EventType.USER_SESSION_EXPIRED)
+                    .event(EventType.USER_SESSION_DELETED)
+                    .detail(Details.REASON, Details.EXPIRED_DETAIL)
                     .success();
         });
     }
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.models.sessions.infinispan.listeners;
+
+import java.lang.invoke.MethodHandles;
+
+import org.infinispan.util.concurrent.BlockingManager;
+import org.jboss.logging.Logger;
+import org.keycloak.events.EventBuilder;
+import org.keycloak.events.EventType;
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.models.RealmModel;
+import org.keycloak.models.utils.KeycloakModelUtils;
+
+/**
+ * Base class to handle expired user session.
+ * <p>
+ * It offloads the event creating and sending to a different thread to avoid blocking the caller.
+ */
+abstract class BaseUserSessionExpirationListener {
+
+    protected static final Logger logger = Logger.getLogger(MethodHandles.lookup().lookupClass());
+    private final KeycloakSessionFactory factory;
+    private final BlockingManager blockingManager;
+
+    BaseUserSessionExpirationListener(KeycloakSessionFactory factory, BlockingManager blockingManager) {
+        this.factory = factory;
+        this.blockingManager = blockingManager;
+    }
+
+    protected void sendExpirationEvent(String userSessionId, String userId, String realmId) {
+        blockingManager.runBlocking(() -> doSend(userSessionId, userId, realmId), "expired-" + userSessionId);
+    }
+
+    private void doSend(String userSessionId, String userId, String realmId) {
+        KeycloakModelUtils.runJobInTransaction(factory, session -> {
+            logger.debugf("User session expired. sessionId=%s, userId=%s, realmId=%s", userSessionId, userId, realmId);
+
+            RealmModel realm = session.realms().getRealm(realmId);
+            if (realm == null) {
+                return;
+            }
+            new EventBuilder(realm, session)
+                    .session(userSessionId)
+                    .user(userId)
+                    .event(EventType.USER_SESSION_EXPIRED)
+                    .success();
+        });
+    }
+
+}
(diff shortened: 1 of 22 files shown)


model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/listeners/BaseUserSessionExpirationListener.java
```diff
@@ -0,0 +1,66 @@
...
+        this.blockingManager = blockingManager;
+    }
+
+    protected void sendExpirationEvent(String userSessionId, String userId, String realmId) {
+        blockingManager.runBlocking(() -> doSend(userSessionId, userId, realmId), "expired-" + userSessionId);
+    }
+
+    private void doSend(String userSessionId, String userId, String realmId) {
+        KeycloakModelUtils.runJobInTransaction(factory, session -> {
+            logger.fatalf("User session expired. sessionId=%s, userId=%s, realmId=%s", userSessionId, userId, realmId);
```
ahus1: TODO: Should not be fatal

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
```diff
@@ -542,6 +542,26 @@ UserSessionAdapter<?> wrap(RealmModel realm, UserSessionEntity entity, boolean o
...
+        // * Infinispan cannot update the expiration settings if the realm settings is changed, so we need to check it here :(
+        // * Loading from database means checking twice, in "importUserSession" method and here.
+        // * We should skip checking if the entity already exists in the transaction.
+        UserSessionEntity entity = userSession.getEntity();
+        RealmModel realm = userSession.getRealm();
+        boolean offline = userSession.isOffline();
+
+        if (isUserSessionExpired(realm, entity, offline)) {
+            log.debugf("Session has expired. Do not load from the cache/database. sessionId=%s offline=%s", entity.getId(), offline);
+            sessionTx.getExpirationListener().onSessionExpired(entity);
```
ahus1: IMHO for persistent sessions, we should only send the event when cleaning the entry from the database. Otherwise we might send it multiple times. Deleting an entry from the database when we access it has not proven to be a good choice, as with concurrent accesses it leads to database conflicts.

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
```diff
@@ -257,21 +260,32 @@ private static int calculateOldestSessionTime(RealmModel realm, boolean offline)
...
+        TypedQuery<Object[]> query = em.createNamedQuery("findExpiredUserSessions", Object[].class)
                 .setParameter("realmId", realm.getId())
                 .setParameter("lastSessionRefresh", expired)
+                .setParameter("offline", offlineStr);
+
+        var expiredSessions = query.getResultStream()
+                .map(JpaUserSessionPersisterProvider::userSessionAndUserProjection)
+                .toList();
+
+        KeycloakModelUtils.runJobInTransaction(session.getKeycloakSessionFactory(), s -> sendExpirationEvents(s, realm, expiredSessions));
```
ahus1: Do not create a new transaction, but send/create the events in the same transaction. Otherwise you might end up with the saved events, but the sessions not yet deleted.

model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java
```diff
@@ -39,6 +40,7 @@
         @NamedQuery(name="deleteClientSessionsByClientStorageProvider", query="delete from PersistentClientSessionEntity sess where sess.clientStorageProvider = :clientStorageProvider"),
         @NamedQuery(name="deleteClientSessionsByUser", query="delete from PersistentClientSessionEntity sess where sess.userSessionId IN (select u.userSessionId from PersistentUserSessionEntity u where u.userId = :userId)"),
         @NamedQuery(name="deleteClientSessionsByUserSession", query="delete from PersistentClientSessionEntity sess where sess.userSessionId = :userSessionId and sess.offline = :offline"),
+        @NamedQuery(name="deleteClientSessionsByUserSessions", query="delete from PersistentClientSessionEntity sess where sess.userSessionId in (:userSessionId) and sess.offline = :offline"),
         @NamedQuery(name="deleteExpiredClientSessions", query="delete from PersistentClientSessionEntity sess where sess.offline = :offline AND sess.userSessionId IN (select u.userSessionId from PersistentUserSessionEntity u where u.realmId = :realmId AND u.offline = :offline AND u.lastSessionRefresh < :lastSessionRefresh)"),
```
ahus1, line 45: Remove no longer needed queries

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
```diff
@@ -257,21 +260,32 @@ private static int calculateOldestSessionTime(RealmModel realm, boolean offline)
...
         String offlineStr = offlineToString(offline);
 
-        logger.tracef("Trigger removing expired user sessions for realm '%s'", realm.getName());
+        logger.fatalf("Trigger removing expired user sessions for realm '%s'", realm.getName());
 
-        int cs = em.createNamedQuery("deleteExpiredClientSessions")
+        TypedQuery<Object[]> query = em.createNamedQuery("findExpiredUserSessions", Object[].class)
                 .setParameter("realmId", realm.getId())
                 .setParameter("lastSessionRefresh", expired)
+                .setParameter("offline", offlineStr);
```
ahus1: As a later optimization, you might set `.setHint(AvailableHints.HINT_READ_ONLY, true);` to avoid creating a big persistence context.

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
ahus1, line 253: We would need a separate expiry for remember-me sessions, so ensure that we have a timely removal for them as well. That would require a separate column and possibly updated indexes. I assume you might not want to do this in this PR, so please create a new task for that.

ahus1 (commented):
Thank you for this PR. See below for some comments.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
pruivo: Infinispan triggers the events on access when Keycloak tries to read an expired session. It felt weird that the test was expecting an invalid token event and not a user session expiration event. I have hope that triggering the event multiple times wouldn't be a problem :sweat_smile:

model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java
pruivo: It is intentional, as I'm worried that users might invoke these queries in their custom providers.

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
pruivo: Right, in this PR just want the events in the right place as I described in the sub-issue. Optimization, indexes, and new columns can be in a follow-up. This PR is already large enough (my fault, as I started resolving the warning in the IDE).

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java
ahus1: +1 for having the event when Infinispan is primary store of the sessions (that is when persistent sessions is disabled and it is not an offline session). In all other cases, I want to see only one event from the JPA layer which is the primary store.

model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentClientSessionEntity.java
ahus1: Hm, I would do it differently but don't object. If you want to keep them, please add a deprecated JavaDoc (I think annotations on annotations don't work here?!)

ahus1 (dismissed):
I've added a loop to delete items. This is not an optimization, but a way to ensure that we don't run in to functional errors when running our load tests. 

The other items are just cleanups. 

Let's hope for a green build. If you don't object, I'll continue to merge it once it is green.

server-spi-private/src/main/java/org/keycloak/events/EventType.java
```diff
@@ -185,14 +184,16 @@ public enum EventType implements EnumWithStableIndex {
     REMOVE_CREDENTIAL_ERROR(0x10000 + REMOVE_CREDENTIAL.getStableIndex(), true),
 
     INVITE_ORG(60, true),
-    INVITE_ORG_ERROR(0x10000 + INVITE_ORG.getStableIndex(), true);
+    INVITE_ORG_ERROR(0x10000 + INVITE_ORG.getStableIndex(), true),
+
+    USER_SESSION_DELETED(61, false)
```
ryanemerson: Do we need a corresponding `USER_SESSION_DELETED_ERROR` enum, it seems all other event types have this?

ahus1: lets add it for symmetry reasons.

ahus1 (approved):

PR #43951, opened 2025-11-05
[OID4VCI]: Add backward compatibility for Draft 15 wallets (single proof support)

### LINKED ISSUE #43926: Add backward compatibility for wallets using OID4VCI Draft 15 by supporting single `proof` field and ignoring deprecated `vct` field in credential requests
### Description

In the final OID4VCI specification, credential requests are required to include a `proofs` field — an array of proof objects. However, some wallet implementations that are still based on OID4VCI Draft 15 send requests with a single `proof` object and may also include a deprecated `vct` field, which is no longer part of the specification.

As a result, credential issuance currently fails for these draft 15 wallets when interacting with Keycloak’s OID4VCI issuer implementation.

The goal of this ticket is to introduce backward compatibility for credential requests from wallets that have not yet migrated to the final OID4VCI version, while maintaining full compliance for newer clients.

Specifically, this update:

Accepts both the `proofs` array (as defined in the final specification) and a single `proof` object (for draft 15 compatibility).

Ignores unknown fields such as `vct` during request deserialization.

These changes ensure that Keycloak remains interoperable with legacy wallet implementations (e.g., **Lissi wallet**) and supports a smoother transition path for ecosystem participants upgrading to the final OID4VCI standard.

Awambeng:
This PR adds backward compatibility for wallets using OID4VCI Draft 15 by supporting both a single `proof` (Draft 15) and the `proofs` array (final spec). Single `proof` objects are automatically converted to the array format, and unknown or deprecated fields like `vct` are ignored. This ensures interoperability with legacy wallets while remaining compliant with the final specification.
Additionally, the trailing whitespace was removed from the `ATTESTATION_JWT_TYP` constant.

**Changes**
- Accept both `proof` (Draft 15) and `proofs` array (final spec) in credential requests
- Automatic conversion of single `proof` to `proofs` array
- Ignore unknown or deprecated fields during deserialization
- Removed trailing whitespace from `ATTESTATION_JWT_TYP` constant
- Added tests to verify proof conversion

Closes #43926

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -77,6 +77,7 @@ import org.keycloak.protocol.oid4vc.model.CredentialResponse;
 import org.keycloak.protocol.oid4vc.model.CredentialResponseEncryption;
 import org.keycloak.protocol.oid4vc.model.CredentialResponseEncryptionMetadata;
 import org.keycloak.protocol.oid4vc.model.CredentialsOffer;
+import org.keycloak.protocol.oid4vc.model.Proof;
 import org.keycloak.services.ErrorResponseException;
 import org.keycloak.protocol.oid4vc.model.ErrorResponse;
 import org.keycloak.protocol.oid4vc.model.ErrorType;
@@ -657,7 +658,24 @@ public class OID4VCIssuerEndpoint {
         }
 
         try {
-            return JsonSerialization.mapper.readValue(requestPayload, CredentialRequest.class);
+            CredentialRequest credentialRequest = JsonSerialization.mapper.readValue(requestPayload, CredentialRequest.class);
+
+            // Convert proof to proofs if proof is provided but proofs is not
+            if (credentialRequest.getProof() != null && credentialRequest.getProofs() == null) {
+                LOGGER.debugf("Converting single 'proof' field to 'proofs' array for backward compatibility");
+                Proof singleProof = credentialRequest.getProof();
+                Proofs proofsArray = new Proofs();
+                
+                // Convert single proof to proofs array format
+                if (singleProof.getJwt() != null) {
+                    proofsArray.setJwt(List.of(singleProof.getJwt()));
+                }
+                
+                credentialRequest.setProofs(proofsArray);
+                credentialRequest.setProof(null); // Clear the proof field to avoid confusion
+            }
+            
+            return credentialRequest;
         } catch (JsonProcessingException e) {
             String errorMessage = "Failed to parse JSON request: " + e.getMessage();
             LOGGER.debug(errorMessage);
@@ -737,7 +755,24 @@ public class OID4VCIssuerEndpoint {
 
         // Parse decrypted content to CredentialRequest
         try {
-            return JsonSerialization.mapper.readValue(content, CredentialRequest.class);
+            CredentialRequest credentialRequest = JsonSerialization.mapper.readValue(content, CredentialRequest.class);
+
+            // Convert proof to proofs if proof is provided but proofs is not
+            if (credentialRequest.getProof() != null && credentialRequest.getProofs() == null) {
+                LOGGER.debugf("Converting single 'proof' field to 'proofs' array for backward compatibility");
+                Proof singleProof = credentialRequest.getProof();
+                Proofs proofsArray = new Proofs();
+                
+                // Convert single proof to proofs array format
+                if (singleProof.getJwt() != null) {
+                    proofsArray.setJwt(List.of(singleProof.getJwt()));
+                }
+                
+                credentialRequest.setProofs(proofsArray);
+                credentialRequest.setProof(null); // Clear the proof field to avoid confusion
+            }
+            
+            return credentialRequest;
         } catch (JsonProcessingException e) {
             throw new JWEException("Failed to parse decrypted JWE payload: " + e.getMessage());
         }
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -77,7 +77,7 @@ import org.keycloak.protocol.oid4vc.model.CredentialResponse;
 import org.keycloak.protocol.oid4vc.model.CredentialResponseEncryption;
 import org.keycloak.protocol.oid4vc.model.CredentialResponseEncryptionMetadata;
 import org.keycloak.protocol.oid4vc.model.CredentialsOffer;
-import org.keycloak.protocol.oid4vc.model.Proof;
+import org.keycloak.protocol.oid4vc.model.JwtProof;
 import org.keycloak.services.ErrorResponseException;
 import org.keycloak.protocol.oid4vc.model.ErrorResponse;
 import org.keycloak.protocol.oid4vc.model.ErrorType;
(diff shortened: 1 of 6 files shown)


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -737,7 +755,24 @@ private CredentialRequest decryptCredentialRequest(String jweString, CredentialR
...
+                // Convert single proof to proofs array format
+                if (singleProof.getJwt() != null) {
+                    proofsArray.setJwt(List.of(singleProof.getJwt()));
+                }
+                
+                credentialRequest.setProofs(proofsArray);
+                credentialRequest.setProof(null); // Clear the proof field to avoid confusion
+            }
+            
+            return credentialRequest;
```
IngridPuppet, line 745: This logic is duplicated. Could we refactor to do it just once?

services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialRequest.java
```diff
@@ -46,6 +48,9 @@ public class CredentialRequest {
     @JsonProperty("proofs")
     private Proofs proofs;
 
+    @JsonProperty("proof")
+    private Proof proof;
+
```
IngridPuppet, line 58: Can we explicitly mark this field as deprecated? A comment should state that it is only kept for backward compatibility.

services/src/main/java/org/keycloak/protocol/oid4vc/model/Proof.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Single Proof object for Credential Request in OID4VCI.
+ * This represents a single proof with proof_type and jwt fields.
+ * Used when the request contains a "proof" field (singular) instead of "proofs" (array).
+ *
+ * @see <a href="https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-credential-request">OID4VCI Credential Request</a>
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Proof {
```
IngridPuppet: Can we explicity name this JwtProof and equally mark it as deprecated?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJWTIssuerEndpointTest.java
```diff
@@ -933,4 +934,62 @@ public void testRequestCredentialWhenNoCredentialBuilderForFormat() {
...
+            Proof singleProof = new Proof()
+                    .setJwt("dummy-jwt")
+                    .setProofType("jwt");
+            requestWithProof.setProof(singleProof);
+
+            String requestPayload = JsonSerialization.writeValueAsString(requestWithProof);
+
+            try {
+                // This should work because the conversion happens in validateRequestEncryption
+                issuerEndpoint.requestCredential(requestPayload);
```
IngridPuppet, line 967: You probably mean to add an assert fail.
```suggestion
                // This should work because the conversion happens in validateRequestEncryption
                issuerEndpoint.requestCredential(requestPayload);
                Assert.fail();
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -737,7 +755,24 @@ private CredentialRequest decryptCredentialRequest(String jweString, CredentialR
...
+                Proofs proofsArray = new Proofs();
+                
+                // Convert single proof to proofs array format
+                if (singleProof.getJwt() != null) {
+                    proofsArray.setJwt(List.of(singleProof.getJwt()));
+                }
+                
+                credentialRequest.setProofs(proofsArray);
+                credentialRequest.setProof(null); // Clear the proof field to avoid confusion
+            }
```
Captain-P-Goldfish: I would suggest to remove this duplicate code here. And do it just once in the validateRequestEncryptionMethod

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJWTIssuerEndpointTest.java
```diff
@@ -933,4 +934,62 @@ public void testRequestCredentialWhenNoCredentialBuilderForFormat() {
...
+
+            String bothFieldsPayload = JsonSerialization.writeValueAsString(requestWithBoth);
+
+            try {
+                issuerEndpoint.requestCredential(bothFieldsPayload);
+                Assert.fail("Expected BadRequestException when both proof and proofs are provided");
+            } catch (BadRequestException e) {
+                int statusCode = e.getResponse().getStatus();
+                assertEquals("Expected HTTP 400 Bad Request", 400, statusCode);
+            }
```
IngridPuppet, line 997: Please assert the error message, probably that proof and proofs should not be present at the same time.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -657,7 +658,24 @@ private CredentialRequest validateRequestEncryption(String requestPayload, Crede
         }
 
         try {
-            return JsonSerialization.mapper.readValue(requestPayload, CredentialRequest.class);
+            CredentialRequest credentialRequest = JsonSerialization.mapper.readValue(requestPayload, CredentialRequest.class);
```
Captain-P-Goldfish, line 661: Set the `CredentialRequest` from line 645 into a temporary variable, modify it here and return it. This allows you to remove the duplicate code from the decryptCredentialRequest method.

IngridPuppet (changes requested):
Hello @Awambeng - I checked the PR and have a few comments. Please could you check?

Awambeng:
@IngridPuppet thank you for the review! I've addressed the comments you left. Could you please check again?

Captain-P-Goldfish (approved):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Captain-P-Goldfish: forgot to submit my review...
you can ignore this because you already handled it

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Captain-P-Goldfish: forgot to submit my review...
you can ignore this because you already handled it

Awambeng:
Thank you for your review, @Captain-P-Goldfish

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCJWTIssuerEndpointTest.java
```diff
@@ -989,6 +990,11 @@ public void testProofToProofsConversion() throws Exception {
             } catch (BadRequestException e) {
                 int statusCode = e.getResponse().getStatus();
                 assertEquals("Expected HTTP 400 Bad Request", 400, statusCode);
+                ErrorResponse error = (ErrorResponse) e.getResponse().getEntity();
+                assertEquals(ErrorType.INVALID_CREDENTIAL_REQUEST, error.getError());
+                assertTrue("Error description should mention proof and proofs exclusivity",
+                        error.getErrorDescription() != null && error.getErrorDescription().toLowerCase().contains("proof")
+                                && error.getErrorDescription().toLowerCase().contains("proofs"));
```
IngridPuppet: Please could make this assertion more straightforward?
```suggestion
                assertEquals("Both 'proof' and 'proofs' must not be present at the same time",
                        error.getErrorDescription());
```

IngridPuppet (approved):
Thank you for the update.

mposolda (approved):
@Awambeng @Captain-P-Goldfish @IngridPuppet Thanks for this PR and the reviews!

PR #43982, opened 2025-11-05
task: use client v1 logic for v2 impl

### LINKED ISSUE #43733: Use existing model logic in Client API v2 MVP
### Description

Reuse the existing logic (`ModelToRepresentation` and `RepresentationToModel`) instead of directly hooking into the model layer. This is a temporary solution to make things safer and quicker for the MVP.

Related: https://github.com/keycloak/keycloak/discussions/40385

### LINKED ISSUE #43809: Granular Authz for Client v2
### Description

Make the authz reflect various realm roles for various actions (create, update, delete, view).

shawkins:
closes: #43733
closes: #43731
closes: #43809

I realize the approach here may not be agreeable to all, so I'd like some early feedback.

The changes here are:

- remove the provider framework from the v2 rest layer. This makes it more difficult to pass state, and it's not clear how this would be used by custom logic. If it makes sense, we can introduce it later. If not, then we can combine the interfaces and the default impls.

- At least temporarily break the abstraction of the ClientService layer as it now becomes dependent upon the v1 Resource classes.

- Minor refactorings of v1 logic to expose models. Also captured https://github.com/keycloak/keycloak/issues/43986

- Removes some of the existing duplication of rest logic between v1 and v2. In particular to things like the call to KeycloakContext.setClient (does this really even make sense to keep doing something like this - doesn't it potentially conflate the authenticating client with the client being created / updated?).

By doing the above we are preserving the v1 auth, eventing, defaulting, and validation logic.

What's missing:

~~- The conversion of v2 ClientRepresentation to v1 ClientRepresentation for the createOrUpdate logic~~ - took the approach of creating a detached ClientModel and using the existing model conversion logic. This sidesteps the need to directly convert between v2 and v1.

What would be next:

- Expanding the logic out to any of the secondary resources referenced in the v2 ClientRepresentation.
- Keep refactoring towards a proper service layer.


vmuzikar (commented):
@shawkins Thank you for the draft.

Despite I'm not really keen on the approach of calling the v1 endpoints directly from the v2 logic, I think it's the safest approach for now (if we don't want to make too many changes to v1 to decouple some of the logic like client policies etc. – which we might do later).

What I'd like to do though is to keep the signatures as close to the intended final state as possible – i.e. I'd like to avoid passing around `ClientsResource` and `ClientResource` and instead create them in `ClientService`, so that the changes required later to move away from the dependency on v1 would be limited to the v2 service layer only, not affecting REST layer so much.

vmuzikar:
CC @mabartos This now overlaps with https://github.com/keycloak/keycloak/issues/43731

shawkins:
> What I'd like to do though is to keep the signatures as close to the intended final state as possible – i.e. I'd like to avoid passing around ClientsResource and ClientResource and instead create them in ClientService, so that the changes required later to move away from the dependency on v1 would be limited to the v2 service layer only, not affecting REST layer so much.

I don't think we're there yet for a couple of reasons:
- we will still have to pass around state - starting with the Auth that is created in the DefaultAdminApi. Unless you are thinking that we should recreate it for each service call - which could break down if we have rest calls that end up making multiple service calls.
- I'm also not sure if we even want to imply all of what the resource layer is doing is encapsulated under the service layer -  will it be handling all of auth, CRUD, defaulting, validation, etc. and is it acceptable that it's directly throwing web exceptions? 
- Since it's internal, there's really no need to worry about how it's called at this point. Alternatively we could do away with the formal service layer and revisit it later.

vmuzikar:
> * we will still have to pass around state - starting with the Auth that is created in the DefaultAdminApi. Unless you are thinking that we should recreate it for each service call - which could break down if we have rest calls that end up making multiple service calls.

We could create `Auth` when the ClientService provider is instantiated via the factory, so it would not be re-created with every service call.

> * I'm also not sure if we even want to imply all of what the resource layer is doing is encapsulated under the service layer -  will it be handling all of auth, CRUD, defaulting, validation, etc. and is it acceptable that it's directly throwing web exceptions?

IMHO, it's not completely wrong place that auth would be (optionally) handled in the service layer. It'd be also less error prone if it needs to be reused in different endpoints. We might make it configurable (whether to check the auth) in the future – to support the cases when the auth needs to be loosened/handled in a custom way.

CRUD, defaulting and validations IMHO belong to the service layer. We could add some exceptions mapping later not to throw web exceptions from the service layer.

> * Since it's internal, there's really no need to worry about how it's called at this point. Alternatively we could do away with the formal service layer and revisit it later.

Yeah, true, later some refactoring towards more formal service layer will be needed in any case.

shawkins:
> Yeah, true, later some refactoring towards more formal service layer will be needed in any case.

So should I just go ahead and remove the service layer for now?

vmuzikar:
> So should I just go ahead and remove the service layer for now?

No, quite the opposite. My suggestion is to keep it but to clean up the interfaces (e.g. not passing the state in args but create it there).

shawkins:
> No, quite the opposite. My suggestion is to keep it but to clean up the interfaces (e.g. not passing the state in args but create it there).

I understand that was the initial suggestion, but I think the points against doing that are valid. So the simpliest thing to do would be to just remove the service layer for now.

vmuzikar:
> So the simpliest thing to do would be to just remove the service layer for now.

Ok, let's do that for now. Let's do a proper service layer later than a half-arsed now...

shawkins:
Removed the provider module, but I've left the interface / impl splits for now.

@mabartos @vmuzikar @Pepo48 since this changes conflict with or supercede several of the other open issues, how do you want to proceed with this?

services/src/main/java/org/keycloak/services/resources/admin/ClientsResource.java
```diff
@@ -119,6 +119,20 @@ public Stream<ClientRepresentation> getClients(@Parameter(description = "filter
...
+        return ModelToRepresentation.filterValidRepresentations(
+                getClientModels(clientId, viewableOnly, search, searchQuery, firstResult, maxResults), c -> {
+                    ClientRepresentation representation = ModelToRepresentation.toRepresentation(c, session);
+                    representation.setAccess(auth.clients().getAccess(c));
+                    return representation;
+                });
+    }
+
+    public Stream<ClientModel> getClientModels(String clientId,
+            boolean viewableOnly,
```
vmuzikar, line 131: Seems we're ignoring `viewableOnly`?

vmuzikar (changes requested):

services/src/main/java/org/keycloak/services/resources/admin/ClientsResource.java
vmuzikar: In general, I'd like to avoid touching/refactoring the v1 as much as possible.

vmuzikar (commented):
After some offline discussion yesterday, overall this approach looks good. Just wanted to highlight a few key points of what we're doing here.
* The main focus is on the REST API contract, the actual impl is temporary for now and takes a quick and safe approach.
* We're calling v1 endpoints directly to leverage all the v1 logic (auth, eventing, etc.).
   * This includes granular auth which also resolves https://github.com/keycloak/keycloak/issues/43809.
   * It required some minor changes to the v1 logic to decouple some business logic from REST.
* We're removing the SPIs approach. There were too many of them and most of them weren't necessary. We will selectively reintroduce some of them later.
   * This also resolves https://github.com/keycloak/keycloak/issues/43731.

I will be creating some follow-up issues for post 26.5 to address the tech debt.

services/src/main/java/org/keycloak/services/resources/admin/ClientsResource.java
shawkins: This was captured as https://github.com/keycloak/keycloak/issues/43986 - it's an existing bug that we'll have to account for later.

vmuzikar: Good point, I didn't notice that. Sorry for the noise.

vmuzikar (approved):
@shawkins LGTM, thanks. Let's get this merged already. :)

@mabartos @Pepo48 You should be unblocked now.

PR #43985, opened 2025-11-05
Audience validation according to latest specs proposal

### LINKED ISSUE #43984: Audience validation according to latest specs proposal
### Description

There is this update document, which affects also RFC-7523 - https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rfc7523bis-03 .

For jwt-grant, it mentions that:
```
For the authorization grant, the client is responsible for ensuring that the audience of the
JWT assertion is appropriate for the authorization server to which it is sent. An authorization
server MAY be identified by either its issuer identifier or its token endpoint URL.
```

Considering this, we may allow both issuer and token-endpoint for the audience. At the same time, we should test that there are not multiple audiences to prevent additional audiences being present.

mposolda:
closes #43984

PR also introduces method `getAllowedAudienceForJWTGrant()` on `JWTAuthorizationGrantProvider`. I suppose we may need that later as for example for Google, the audiences might be different (That is for later, but does not hurt to introduce that new method now IMO).

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
@@ -44,7 +44,6 @@ import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.sessions.RootAuthenticationSessionModel;
 
 import jakarta.ws.rs.core.Response;
-import java.util.List;
 
 public class JWTAuthorizationGrantType extends OAuth2GrantTypeBase {
 
@@ -53,7 +52,6 @@ public class JWTAuthorizationGrantType extends OAuth2GrantTypeBase {
         setContext(context);
 
         String assertion = formParams.getFirst(OAuth2Constants.ASSERTION);
-        String expectedAudience = Urls.realmIssuer(session.getContext().getUri().getBaseUri(), realm.getName());
 
         try {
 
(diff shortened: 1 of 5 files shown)


services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
```diff
@@ -64,7 +63,6 @@ public Response process(Context context) {
             authorizationGrantContext.validateClient();
 
             //mandatory claims
-            authorizationGrantContext.validateTokenAudience(List.of(expectedAudience), false);
```
rmartinc, line 67: You have to remove the definition of this `expectedAudience` variable above, right?

rmartinc (changes requested):
Thanks @mposolda! Just a minor thing, the `expectedAudience` is still defined and I think that is not needed now. You can remove that line too.

services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
mposolda: Yes, you are right. Done!

rmartinc (approved):
Thanks @mposolda! LGTM!

graziang (approved):
@mposolda Thanks!

mposolda:
@rmartinc @graziang Thanks!

PR #43988, opened 2025-11-05
AUTH_SESSION_ID cookie has the incorrect route

### LINKED ISSUE #43933: AUTH_SESSION_ID cookie has the incorrect route
### Before reporting an issue

### Area

core

### Describe the bug

Since https://github.com/keycloak/keycloak/issues/34027, the session ID is signed to ensure its integrity. However, when computing the route, the signed value is used, but it should instead use the unsigned session ID, as that's what is stored in the Infinispan cache. See:

https://github.com/keycloak/keycloak/blob/24f67d0c0484955b5c60cd23df2a2ad3bb902fc8/services/src/main/java/org/keycloak/services/managers/AuthenticationSessionManager.java#L108-L116

### Version

main

### Regression

### Expected behavior

The appended route in the cookie to be correct and to be the node where the session lives. 

### Actual behavior

The appended route can be assigned to any node. The probability of being correct is `2/cluster-size`. 

### How to Reproduce?

* Create a cluster with more than 2 nodes.
* Access the admin console login page in one of the nodes and keep track of the node name. 
* In the browser, look at the cookies, and search for AUTH_SESSION_ID. 
* [main branch] If you delete the cookie and refresh the page, it must show the same route every time
* [other version] You can use the metric `vendor_statistics_approximate_entries` to see where the session is stored. It is correct `2/cluster-size` of the time. 

### Anything else?

_No response_

pruivo:
Fixes #43933


testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/cluster/AuthenticationSessionClusterTest.java
```diff
@@ -99,11 +99,21 @@ public void testAuthSessionCookieWithAttachedRoute() {
             String route = authSessionCookie.substring(authSessionCookie.indexOf(".") + 1);
             visitedRoutes.add(route);
 
+            getTestingClientFor(backendNode(0)).server().run(session -> {
```
ryanemerson, line 99: `TopologyInfo#getRouteName` is still being used on L44 in this class. Should this now be using `StickySessionEncoderProvider#SessionIdAndRoute` instead? If so, should we deprecate the `getRouteName` method?

pruivo: The method disables sticky sessions, so the route is always `null`. I can change the test to assert that the route is null if you agree.

I would love to deprecate `TopologyInfo` and remove it. With this PR, it only stores the node name and the site name. I'm considering storing those in a `record NodeInfo(String nodeName, String siteName)`.

ryanemerson: > I can change the test to assert that the route is null if you agree.

Yes I think that would be better.

> I would love to deprecate `TopologyInfo` and remove it.

+1 it would be nice to remove usage and deprecate it, lets create a separate issue for that.

I see that `amIOwner` is currently never used, so in this PR we should replace the usage of `getRouteName` in the test and deprecate the unused public methods.

ryanemerson (dismissed):

ryanemerson:
LGTM, but the build needs to be fixed to play nice with spotless @pruivo

ahus1 (approved):

PR #43998, opened 2025-11-06
Add per-scope consent for optional client scopes

cbcoutinho:
Implements granular OAuth consent allowing users to selectively approve/deny individual optional client scopes while keeping required scopes mandatory. This addresses privacy concerns and gives users more control over their data sharing preferences.

Changes:
- Updated OAuthGrantBean to distinguish required vs optional scopes
- Modified consent screen UI to show checkboxes for optional scopes
- Updated LoginActionsService to process selective scope consent
- Added i18n support for new consent UI labels
- Added comprehensive test coverage

Fixes: #42641

---

_This PR was generated with the help of AI, and reviewed by a Human_


mposolda:
@cbcoutinho Thanks for the PR. The feature makes sense to me. However it should be configurable whether client want to allow this behaviour or not. As right now when client adds `scope=something` to the login URL, client application would expect that such scope is included in the token. However this change would break this expectation.

I would suggest to add some option to the client settings like **Allow user to deselect optional scopes** which will be in the admin console in the client settings below **Consent required** option and would be visible just if **Consent required** is selected. It should be OFF by default. It should likely map to some client attribute, so there is no need to introduce any changes in the DB schema (See for example `OIDCAdvancedConfigWrapper` class for the details).

cbcoutinho:
@mposolda Thank you for the feedback! I've made the per-scope consent feature configurable as requested.

## Changes Made

### 1. Client Configuration
- Added **"Allow user to deselect optional scopes"** client attribute
- Implemented via `OIDCAdvancedConfigWrapper` (no DB schema changes needed)
- **Default: OFF** - maintains backward compatibility

### 2. Admin Console
- Added toggle in client settings, positioned below "Consent required"
- Only visible/enabled when "Consent required" is checked
- Includes help text explaining the feature

### 3. Backend Behavior
- When **disabled** (default): All scopes in the auth session are granted (original behavior)
- When **enabled**: Users can deselect optional scopes; required scopes are always granted
- Updated `LoginActionsService.processConsent()` to check the configuration

### 4. Consent Screen
- Updated both base and keycloak.v2 theme templates
- Checkboxes only appear for optional scopes when the setting is enabled
- Required scopes are clearly marked

This ensures that when a client adds `scope=something` to the login URL, the scope will be included in the token **unless** the client explicitly enables this setting to allow users more granular control.

All core modules compile successfully with these changes.

mposolda:
@cbcoutinho Thanks for the PR and for the additional work!

Few points:

1) After your PR, there are some changes in the consent screen even if the new client switch is disabled. See screenshot before your changes:

Screenshot from 2025-11-14 12-00-33


And after your changes:


Screenshot from 2025-11-14 11-47-17

Could you please make sure that there is not mentioned `Required` as all scopes are actually required if the option `Allow user to deselect optional scopes` is false?

 
2) I've tried to test with the builtin account console client, but it failed for me. Some steps I did:

- In the `master` realm, I've enabled `Consent required` and `Allow to deselect optional scope` to true for the `account-console` client

- In the dedicated browser window, I've opened http://localhost:8080/realms/master/account

- I've manually updated `scope` parameter in the browser URL and added `scope=openid address` to be new value in the URL. Opened that URL again.

- After filling username/password, the consent screen is displayed. I've actually kept the `address` scope selected. But after confirm, user is not able to continue login and there is an error in the server.log:
```
2025-11-14 11:55:24,204 WARN  [org.keycloak.events] (executor-thread-27) type="CODE_TO_TOKEN_ERROR", realmId="d47331db-efa2-4c21-ab84-bfc9f4059204", realmName="master", clientId="account-console", userId="6b7bb9ac-f5a6-499f-aad4-2f5a6cf62ae4", sessionId="Cp0gBCWaCkAl_9icjO-LVxSB", ipAddress="127.0.0.1", error="not_allowed", reason="Client no longer has requested consent from user", grant_type="authorization_code", code_id="Cp0gBCWaCkAl_9icjO-LVxSB", client_auth_method="client-secret"
```
The same error is shown when I select or unselect `address` scope. But when option `Allow to deselect optional scope` is OFF, everything work as expected.

I am not sure what is this error related to (maybe it is related to the keycloak.js adapter), but it will be good to investigate this. Also it will be good if the new automated tests you're added are testing also code-to-token request and make sure that it is possible to obtain the tokens. See for example test `oauthGrantAcceptTest()` in that class `OAuthGrantTest`. Also will be good to validate that if some scopes are deselected, they are really not present in the access-token response.

mposolda:
@cbcoutinho Also it would be needed to rebase your PR on top of latest `main` branch to avoid conflicts as there are some changes in the `import` settings.

PR #44001, opened 2025-11-06
Cleanup UTF-8 charset usage

ahus1:
Closes #44000


ahus1:
The keycloak-core parts still use JDK 8

PR #44011, opened 2025-11-06
fix: consolidating config logic

### LINKED ISSUE #42000: Incorrect logic of getArray() in ComponentModelScope
### Before reporting an issue

### Area

core

### Describe the bug

Redundant check of a known non-null value against the constant null. `String.split()` never returns null so the `origScope` value is unused

https://github.com/keycloak/keycloak/blob/309957033eb0da99860a8da9dad3643cc76f136a/server-spi-private/src/main/java/org/keycloak/component/ComponentModelScope.java#L74-L77

### Version

26.3.3

### Regression

### Expected behavior

Calls to getArray where no config value exists will return the origScope value

### Actual behavior

All calls to getArray would return res, the origScope value is not used.

### How to Reproduce?

Well... 

### Anything else?

Moving the discussion from https://github.com/keycloak/keycloak/pull/40150#discussion_r2276887390 here
Original issue: https://github.com/keycloak/keycloak/issues/40061

shawkins:
closes: #42000

Introduces an AbstractScope and default methods to consolidate the logic as much as possible.

This addresses the inconsistency noted with the componentmodelscope getArray method.

The AbstractScope class is needed only if want to hide the getValue method.

### DIFF
--- a/core/src/main/java/org/keycloak/Config.java
+++ b/core/src/main/java/org/keycloak/Config.java
@@ -20,7 +20,9 @@ package org.keycloak;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.keycloak.common.util.StringPropertyReplacer;
 import org.keycloak.common.util.StringPropertyReplacer.PropertyResolver;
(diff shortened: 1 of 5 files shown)


mabartos (approved):
@shawkins LGTM, thanks!

pedroigor (approved):

Issue #10368, opened 2022-02-22
Investigate how to increase test coverage for WebAuthn tests with Firefox


mabartos:
### Description

After resolving issue #10062, we integrate Firefox browser to our WebAuthn testsuite. However, AFAIK, the browser doesn't have the same capabilities and abilities as Chromium-based browsers and it's not possible to use VirtualAuthenticators (VA) for that, which is a cornerstone of the WebAuthn tests. I wasn't able to execute those tests with the VA as it seems that Gecko Driver doesn't implement the necessary interfaces. So, the only way was to use Soft token for Firefox, but we are not able to programmatically change parameters of the authenticator, so we have only ~50% test coverage despite the Chrome browser. It'd be nice to try to investigate some other approaches, but at this moment, the current one is sufficient.

rmartinc:
Just adding the related issues in selenium and the gecko-driver:

https://github.com/mozilla/geckodriver/issues/2187
https://bugzilla.mozilla.org/show_bug.cgi?id=1957746

PR #44016, opened 2025-11-06
[Test Framework] Migrate initial WebAuthn setup + WebAuthnRegisterAndLoginTest.

### LINKED ISSUE #35395: Add WebAuthN test module
Add webauthn test module to [tests](https://github.com/keycloak/keycloak/tree/main/tests) and execute tests tests as part of [WebAuthn IT](https://github.com/keycloak/keycloak/blob/fda6dde44f676bb151ee89ad2fe3e1bca28e21eb/.github/workflows/ci.yml#L928)

WebAuthN tests should be placed in its own test module rather than a suite as they require specific configuration of the browser, and can not for example run with HtmlUnit. As such it is cleaner to clearly separate these from the base tests.

lhanusov:
Closes #35395


test-framework/core/pom.xml
```diff
@@ -39,6 +39,7 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
+            <version>6.0.0</version>
```
stianst: Don't update to JUnit 6 within this PR

lhanusov: just tested and forgot to remove it, thanks for spotting :-)

mabartos:
@lhanusov There are some failures in the webauthn testsuite. I guess it's related. Did you and the team consider having a separate CI job for the new WebAuthn tests that might be CI optional for some time? Or do you think it's gonna be stable for PRs?

lhanusov:
@mabartos this failure is caused by a different browser version downloaded by CI and supported by selenium, I need to think about, what's the best solution for it, but we don't plan to have it separately out of main CI.

.github/workflows/ci.yml
```diff
@@ -951,6 +951,12 @@ jobs:
           echo "Tests: $TESTS"
           ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Dtest=$TESTS -Dbrowser=${{ matrix.browser }} "-Dwebdriver.chrome.driver=$CHROMEWEBDRIVER/chromedriver" "-Dwebdriver.gecko.driver=$GECKOWEBDRIVER/geckodriver" -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
 
+      - name: Run new WebAuthn IT
+        run: |
+          if [ ${{ matrix.browser }} = "chrome" ] ; then
+              ./mvnw package -f tests/webauthn/pom.xml -Dtest=WebAuthnTestSuite "-Dwebdriver.chrome.driver=$CHROMEWEBDRIVER/chromedriver"
```
stianst: Should be `mvnw test` surely?

.github/workflows/ci.yml
```diff
@@ -951,6 +951,12 @@ jobs:
           echo "Tests: $TESTS"
           ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Dtest=$TESTS -Dbrowser=${{ matrix.browser }} "-Dwebdriver.chrome.driver=$CHROMEWEBDRIVER/chromedriver" "-Dwebdriver.gecko.driver=$GECKOWEBDRIVER/geckodriver" -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
 
+      - name: Run new WebAuthn IT
+        run: |
+          if [ ${{ matrix.browser }} = "chrome" ] ; then
```
stianst: If Firefox is not really being tested in the old testsuite either and the majority of the tests are excluded should we not just drop Firefox from the matrix completely? @mabartos

test-framework/core/src/main/java/org/keycloak/testframework/events/EventAssertion.java
```diff
@@ -37,6 +38,16 @@ public EventAssertion type(EventType type) {
         return this;
     }
 
+    public EventAssertion isSessionId() {
```
stianst: Might be better to use hasSessionId/CodeId than is:
```suggestion
    public EventAssertion hasSessionId() {
```

.github/workflows/ci.yml
stianst: Since it's a separate module you don't need `-Dtest=WebAuthnTestSuite` just run all the tests in the module.

tests/base/src/test/java/org/keycloak/tests/suites/WebAuthnTestSuite.java
stianst, line 1: This isn't needed or used

tests/webauthn/src/test/java/org/keycloak/tests/webauthn/WebAuthnTestSuite.java
stianst, line 1: You don't need this as we should just run all tests within the webauthn module

stianst (commented):
Some minor comments inline

.github/workflows/ci.yml
mabartos: We already had a discussion with @lhanusov about WebAuthn tests and Firefox. Basically, for Firefox, the WebAuthn tests cannot register any Passkey/webauthn-credential, because there's no way at this moment (AFAIK). There's this issue: https://github.com/keycloak/keycloak/issues/10368

WebAuthn tests with Firefox just test if a user is able to go to the pages like register passkey, login with passkey,... so, simply checking the presence of these login screens/forms with Firefox + error pages.

I pointed out to @lhanusov that if some adequate test coverage is achieved for the webauthn screens on Firefox in a different testsuite, it does not have to be tested here. And if not, it depends on the QE decision if it's really worth maintaining the Chrome vs. Firefox discrepancy to just test the WebAuthn screens.

Maybe it'd be better to remove this Firefox support, as the potential approach seems to be far away, and create a simple test class for WebAuthn screens run on Firefox in some general UI Selenium tests?

@stianst @lhanusov WDYT? (cc: @rmartinc)

stianst: What you are explaining around WebAuthn testing for Firefox really doesn't add any value IMO. I can't see any reason why the pages would display on Chrome, but not on Firefox. Without actually doing some real tests you wouldn't know if the JavaScript and anything is actually working or not, which is the parts that are browser dependent.

For now I would just completely drop any WebAuthn testing for Firefox; and revisit this in the future if there ever is support for doing proper testing with WebAuthn and Firefox.

mabartos: > What you are explaining around WebAuthn testing for Firefox really doesn't add any value IMO. I can't see any reason why the pages would display on Chrome, but not on Firefox. Without actually doing some real tests you wouldn't know if the JavaScript and anything is actually working or not, which is the parts that are browser dependent.

Yes, if it's considered as it does not add any value, I'm ok to remove it. 

> For now I would just completely drop any WebAuthn testing for Firefox; and revisit this in the future if there ever is support for doing proper testing with WebAuthn and Firefox.

+1; agree

stianst (dismissed):

.github/workflows/ci.yml
```diff
@@ -943,13 +937,16 @@ jobs:
...
       - name: Run WebAuthn IT
         run: |
           TESTS=`testsuite/integration-arquillian/tests/base/testsuites/suite.sh webauthn`
           echo "Tests: $TESTS"
-          ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Dtest=$TESTS -Dbrowser=${{ matrix.browser }} "-Dwebdriver.chrome.driver=$CHROMEWEBDRIVER/chromedriver" "-Dwebdriver.gecko.driver=$GECKOWEBDRIVER/geckodriver" -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
+          ./mvnw test ${{ env.SUREFIRE_RETRY }} -Pauth-server-quarkus -Dtest=$TESTS -Dbrowser=chrome "-Dwebdriver.chrome.driver=$CHROMEWEBDRIVER/chromedriver" "-Dwebdriver.gecko.driver=$GECKOWEBDRIVER/geckodriver" -pl testsuite/integration-arquillian/tests/base 2>&1 | misc/log/trimmer.sh
+
+      - name: Run new WebAuthn IT
+        run: |
+          ./mvnw test -f tests/webauthn/pom.xml
```
stianst, line 949:
```suggestion
          ./mvnw test -Dkc.test.browser=chrome-headless -f tests/webauthn/pom.xml
```

test-framework/core/src/main/java/org/keycloak/testframework/events/EventAssertion.java
```diff
@@ -37,6 +38,16 @@ public EventAssertion type(EventType type) {
         return this;
     }
 
+    public EventAssertion hasSessionId() {
+        MatcherAssert.assertThat(event.getSessionId(), EventMatchers.isSessionId());
+        return this;
+    }
+
+    public EventAssertion isCodeId() {
```
stianst:
```suggestion
    public EventAssertion hasCodeId() {
```

stianst (approved):

PR #44023, opened 2025-11-07
Ensure delete-user step doesn't trigger removal of federated user from federation provider

### LINKED ISSUE #43538: Ensure delete-user step doesn't trigger removal of federated user from federation provider
### Description

Current implementation of the delete user step is removing the user regardless of them being federated or not. If the workflow is triggered, for example, for a LDAP user, the removal will be propagated all the way to LDAP when the intention might be the removal of the local user only.

We need to check for federation links and trigger local removal for these users. We can also add a config flag to the step to allow propagation of the removal to the storage provider.

### Value Proposition

Prevent accidental removal of users from the storage provider.

### Goals

See description

### Non-Goals

See description

### Discussion

_No response_

### Notes

_No response_

Hathoute:
Closes #43538

### DIFF
--- a/services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
+++ b/services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
@@ -22,15 +22,21 @@ import org.keycloak.component.ComponentModel;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
-
+import org.keycloak.models.cache.UserCache;
+import org.keycloak.storage.UserStoragePrivateUtil;
+import org.keycloak.storage.UserStorageUtil;
 
 public class DeleteUserStepProvider implements WorkflowStepProvider {
 
+    public static final String PROPAGATE_TO_SP = "propagate-to-provider";
+
     private final KeycloakSession session;
+    private final ComponentModel stepModel;
     private final Logger log = Logger.getLogger(DeleteUserStepProvider.class);
 
     public DeleteUserStepProvider(KeycloakSession session, ComponentModel model) {
         this.session = session;
+        this.stepModel = model;
     }
 
     @Override
@@ -46,7 +52,19 @@ public class DeleteUserStepProvider implements WorkflowStepProvider {
             return;
         }
 
-        log.debugv("Deleting user {0} ({1})", user.getUsername(), user.getId());
-        session.users().removeUser(realm, user);
+        if (!user.isFederated() || stepModel.get(PROPAGATE_TO_SP, false)) {
+          log.debugv("Deleting user {0} ({1})", user.getUsername(), user.getId());
+          session.users().removeUser(realm, user);
+          return;
+        }
+
+        // delete the local user only
+        UserStoragePrivateUtil.userLocalStorage(session).removeUser(realm, user);
+        log.debugv("Deleting federated user {0} ({1}) from local storage only", user.getUsername(), user.getId());
+        UserCache userCache = UserStorageUtil.userCache(session);
+        // if cache is enabled, evict the user from cache
+        if (userCache != null) {
+            userCache.evict(realm, user);
+        }
     }
 }
--- a/tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/DeleteUserWorkflowStepTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/DeleteUserWorkflowStepTest.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
```diff
@@ -17,20 +17,27 @@
...
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
-
+import org.keycloak.models.UserProvider;
+import org.keycloak.storage.DatastoreProvider;
+import org.keycloak.storage.datastore.DefaultDatastoreProvider;
 
 public class DeleteUserStepProvider implements WorkflowStepProvider {
 
+    public static final String PROPAGATE_TO_SP = "PROPAGATE_TO_SP";
```
sguilhen: I would suggest using lower case with dashes for these config constants - we've been changing a bit the workflow definition to use lower case constants everywhere as it reads better in the final YAML or JSON definition.

So we are, for example, moving away from `USER_LOGGED_IN` in favor of `user-logged-in`. In this case, I would set the constant to something like `propagate-to-storage-provider` or just `propagate-to-provider`.

services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
```diff
@@ -46,7 +53,17 @@ public void run(WorkflowExecutionContext context) {
...
-        log.debugv("Deleting user {0} ({1})", user.getUsername(), user.getId());
-        session.users().removeUser(realm, user);
+        if (!user.isFederated() || stepModel.get(PROPAGATE_TO_SP, false)) {
+          log.debugv("Deleting user {0} ({1})", user.getUsername(), user.getId());
+          session.users().removeUser(realm, user);
+          return;
+        }
+
+        log.debugv("Deleting federated user {0} ({1}) from local store");
+        localStorage().removeUser(realm, user);
```
sguilhen: You can leverage some utility classes so you don't need to create your own `localStorage` method. This is what I suggest doing:

```
            // delete the local user only
            UserStoragePrivateUtil.userLocalStorage(session).removeUser(realm, user);
            log.debugv("Deleting federated user {0} ({1}) from local storage only", user.getUsername(), user.getId());
            UserCache userCache = UserStorageUtil.userCache(session);
            // if cache is enabled, evict the user from cache
            if (userCache != null) {
                userCache.evict(realm, user);
            }
```

The cache eviction is just to guarantee we are not cashing any deleted user.

services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
```diff
@@ -46,7 +53,17 @@ public void run(WorkflowExecutionContext context) {
...
+    }
+
+    private UserProvider localStorage() {
+        return ((DefaultDatastoreProvider) session.getProvider(DatastoreProvider.class)).userLocalStorage();
```
sguilhen: If you use the utility classes as suggested above, you won't need this method.

sguilhen (changes requested):
Thank you so much for the PR, @Hathoute ! It is really incredible to see help coming from the community for a feature yet to reach supported status.

In general it looks good to, I've added just a few comments that should be easy to fix.

services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
Hathoute: Introducing cache evinction resulted in failure of tests where `propagateToProvider` is false (I was testing against this cache without knowing...)

Now instead of 
```
user = session.users().getUserByUsername(realm, "user1");
``` 
What I'm doing is (I've exposed the dummy users field)
```
DummyUserFederationProviderFactory providerFactory = (DummyUserFederationProviderFactory) session.getKeycloakSessionFactory()
         .getProviderFactory(UserStorageProvider.class, DummyUserFederationProviderFactory.PROVIDER_NAME);
user = providerFactory.getUsers().get("user1");
```
though I'm not sure if there's a better way

services/src/main/java/org/keycloak/models/workflow/DeleteUserStepProvider.java
Hathoute: Using `propagate-to-provider` as it's simpler

sguilhen (commented):
@Hathoute Sorry for missing something in the first review, but would you be able to extract the test into a separate test file? The BrokeredUser test is handling tests involving federated identities (those obtained from identity providers), and I would like to avoid mixing it with federated users.

If it is too much of a hassle we can do it in a follow up - we have a separate task for reviewing the testsuite, so this is not really blocker.

Hathoute:
> @Hathoute Sorry for missing something in the first review, but would you be able to extract the test into a separate test file? The BrokeredUser test is handling tests involving federated identities (those obtained from identity providers), and I would like to avoid mixing it with federated users.
> 
> If it is too much of a hassle we can do it in a follow up - we have a separate task for reviewing the testsuite, so this is not really blocker.

No worries, I've done that while we're at it.

sguilhen:
> > @Hathoute Sorry for missing something in the first review, but would you be able to extract the test into a separate test file? The BrokeredUser test is handling tests involving federated identities (those obtained from identity providers), and I would like to avoid mixing it with federated users.
> > If it is too much of a hassle we can do it in a follow up - we have a separate task for reviewing the testsuite, so this is not really blocker.
> 
> No worries, I've done that while we're at it.

Awesome, thank you!

sguilhen (approved):

sguilhen:
@ahus1 looks good to me now, can you please trigger the CI?

ahus1 (approved):

Issue #43399, opened 2025-10-13
Randomize or round claims containing time-related information for mitigating correlation


Ogenbertrand:
### Description

> Claims containing time-related information, such as issuance or expiration dates, SHOULD be either individually randomized within an appropriate time window (e.g., within the last 24 hours), or rounded (e.g., to the start of the day), to avoid unintended correlation factors. This is not mandated but recommended, so my take is that this should be configurable.

PR #44029, opened 2025-11-07
[OID4VCI] VC of type oid4vc_natural_person has invalid id value

### LINKED ISSUE #43854: OID4VCI credentials have invalid subject id value
### Before reporting an issue

### Area

oid4vc

### Describe the bug

With oid4vc_natural_person  we get credentials like this 

```
{
    "_sd": [
        "57fy_JHyzJS5gt3QrHqnzeq18HVM3NzA1jv2Gz-0DgY",
        "6noKmxXkD9wtoVQq6w6Yy0ldWe6cIgPHrqKzRInh7zA",
        "7qVtLDYk8l5EMxAd8m4q99QzlCqc0ACGThr8-JYzXBY",
        "Ceogod-Tn_vtM_dFWRZsQlqvmZAvcy6UERKI3tAq5Kg",
        "OlwZJoSFht0zVn2--0g5NAoQ0sih7lAl0wgkBV55R2c",
        "OuqThn6OhslgAFYTSxNpY6EZIF0-K4q9YEY4jlx0Ssw",
        "VvgcR5g21HPAi_CzGwuiFR3OqSrK2DNtB9I1rYiwOSM",
        "hY0GlYSCrsOGb63q2h4xLqx4UGjhFHzZI1WGwhilnzk",
        "mUqL3x6r4WJ45D5L8GRoA8pS09M9I_yg4WTbZFl6-68",
        "nH9-5FU1Qzqk8qlBDtPwphUs3tNiidkQJFXXv0yH6MY",
        "siK_io9IWrisfoxs2tNn0ACAR1FKG-YpecGM0j8AmRQ",
        "uJ4bCl6Yu4hr_fVmc5kyQ8oR_BA7Ufs0UFQ6OG7NmbI",
        "v1zy1vJbEvPO3O7xUY1hPe5SD4OQKGl8IdthZLiBPY8"
    ],
    "_sd_alg": "SHA-256",
    "iss": "did:key:z4oJ8dcwF8oDcPcSYbAeioXtoq6vmJZ7o9fAEbPkYncQFnGb71ZaAoSrG1azKQnVawjcJiCgxjEECnpcTYPVHQGEqNoQx",
    "id": "id",
    "cnf": {
        "jwk": {
            "kid": "uBHnFK3vKHYpl6yeaHRH-9eG8zRhpkqbbB-HIMhfyQM",
            "kty": "EC",
            "alg": "ES256",
            "use": "sig",
            "crv": "P-256",
            "x": "H4B9ZHQISPRy7JE39KBKKtccLt3NxkuwUvHJxHtwDK8",
            "y": "QXpXqW_bkVRMBNeFl0E2KsVmGvSc1l7KsUBamkcKaMA"
        }
    }
}
```

The `"id": "id"` is not a valid dc+sd-jwt claim according to [oid4vci-1.0](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#I-D.ietf-oauth-sd-jwt-vc)

The code suggests that this is supposed to be a subject id (OID4VCSubjectIdMapper), which however should be a did and part of the subject claims (instead of a top level jwt claim).

The comment reads _Sets an ID for the credential, either randomly generated or statically configured_, but such a credential id is not to be found in the spec AFAICS.

I suggest to ...

1. Rename that mapper to `OID4VCCredentialIdMapper`
2. Make it work such that it actually creates a random id for every vc of that type
3. Remove it from oid4vc_natural_person for spec alignment

Note, when a credential is issued as a JWT, the credential id should be different from JWT id (jti)

tdiesler:
closes #43854

This PR makes a number of interrelated changes in connection with VC Subject Id.

In the order of file changes, we have ...

* IssuedAt (iat) becomes a disclosed subject claim (not an SD top level claim by spec)
* Improved error messages on invalid disclosures (i.e. name the invalid claim)
* Add default user attributes for --feature=oid4vc-vci (specifically, it adds the User's DID)
* SubjectIdMapper uses the User's DID instead of a hard coded value or a random uuid
* SdJwtCredentialBuilder adds expiry date (exp) as top level claim (by spec)
* SdJwtCredentialBuilder adds subject id (sub) as disclosed subject claim


mposolda:
@tdiesler Thanks for the PR.

For the future, can you please use something like `closes #43854` inside the commit message instead of `See: #43854` ? That will make sure that the comment with `closes #43854` is added in the PR description and that will automatically link the PR with the GH issue. See Keycloak contributors guide and other Keycloak commits for the details.

core/src/test/java/org/keycloak/sdjwt/DisclosureRedListTest.java
```diff
@@ -32,7 +32,7 @@ public void testDefaultRedListedInObjectClaim() {
     public void testDefaultRedListedInArrayClaim() {
         DisclosureSpec.builder()
                 .withUndisclosedClaim("given_name", "2GLC42sKQveCfGfryNRN9w")
-                .withUndisclosedArrayElt("iat", 0, "2GLC42sKQveCfGfryNRN9w")
+                .withUndisclosedArrayElt("exp", 0, "2GLC42sKQveCfGfryNRN9w")
```
Captain-P-Goldfish: This will break with https://github.com/keycloak/keycloak/pull/44227. Duplicate salt values will fail directly on the `build()`-method.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -47,48 +51,53 @@ public SdJwtCredentialBody buildCredentialBody(
...
+        Optional.ofNullable(maybeIssuanceDate).ifPresent(it -> {
+            claims.put(JWT_ISSUANCE_DATE_CLAIM, it.getEpochSecond());
+        });
+        Optional.ofNullable(claims.get(SUBJECT_ID_CLAIM)).ifPresent(it -> {
+            claims.put(JWT_SUBJECT_CLAIM, claims.remove(SUBJECT_ID_CLAIM));
+        });
+
+        // Put inner claims into the disclosure spec, except the one to be kept visible
+        var disclosureSpecBuilder = DisclosureSpec.builder();
+        var outerClaims = credentialBuildConfig.getSdJwtVisibleClaims();
```
Captain-P-Goldfish: Please do not use `var`-keyword

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -31,8 +31,12 @@
...
 public class SdJwtCredentialBuilder implements CredentialBuilder {
 
-    public static final String ISSUER_CLAIM = "iss";
-    public static final String VERIFIABLE_CREDENTIAL_TYPE_CLAIM = "vct";
+    public static final String JWT_ISSUER_CLAIM = "iss";
+    public static final String JWT_ISSUANCE_DATE_CLAIM = "iat";
+    public static final String JWT_EXPIRE_DATE_CLAIM = "exp";
+    public static final String JWT_SUBJECT_CLAIM = "sub";
+    public static final String JWT_CREDENTIAL_TYPE_CLAIM = "vct";
+    public static final String SUBJECT_ID_CLAIM = "id";
```
Captain-P-Goldfish: these constants should be moved to `OID4VCConstants`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -47,48 +51,53 @@ public SdJwtCredentialBody buildCredentialBody(
...
-                        IntStream.range(0, listValue.size())
-                                .forEach(i -> disclosureSpecBuilder
-                                        .withUndisclosedArrayElt(entry.getKey(), i, SdJwtUtils.randomSalt())
-                                );
-                    } else {
-                        disclosureSpecBuilder.withUndisclosedClaim(entry.getKey(), SdJwtUtils.randomSalt());
-                    }
+
+        var maybeIssuanceDate = verifiableCredential.getIssuanceDate();
+        var maybeExpirySeconds = credentialBuildConfig.getExpiryInSeconds();
```
Captain-P-Goldfish: it is somehow irritating why these variables are called `maybe`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -42,61 +44,78 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
 
     public static final String MAPPER_ID = "oid4vc-subject-id-mapper";
 
+    public static final String CLAIM_NAME_ID = "id";
+    public static final String USER_ATTRIBUTE_DID = "did";
```
Captain-P-Goldfish: should be moved to `OID4VCConstants`

Captain-P-Goldfish (changes requested):
not finished with the review yet. But I am once more out of time :-)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
tdiesler: done

tdiesler:
BTW, what is your preferred review workflow? Would you like to resolve each request or should I when I think its done?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -49,54 +54,68 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
...
+                .findFirst()
+                .orElse(null);
+        if (userAttr != null) {
+            claims.put(propertyName, userAttr);
+        }
+    }
+
+    @Override
+    public List<String> getMetadataAttributePath() {
+        return ListUtils.union(getAttributePrefix(), List.of(CLAIM_NAME_SUBJECT_ID));
```
Captain-P-Goldfish: why not

`ListUtils.of(getAttributePrefix(), CLAIM_NAME_SUBJECT_ID)`

?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
```diff
@@ -74,46 +73,45 @@ public class OID4VCUserAttributeMapper extends OID4VCMapper {
...
         String propertyName = attributePath.get(attributePath.size() - 1);
         String userAttribute = mapperModel.getConfig().get(USER_ATTRIBUTE_KEY);
-        boolean aggregateAttributes = Optional.ofNullable(mapperModel.getConfig().get(AGGREGATE_ATTRIBUTES_KEY))
+        boolean aggregate = Optional.ofNullable(mapperModel.getConfig().get(AGGREGATE_ATTRIBUTES_KEY))
                 .map(Boolean::parseBoolean).orElse(false);
-        Collection<String> attributes =
-                KeycloakModelUtils.resolveAttribute(userSessionModel.getUser(), userAttribute,
-                        aggregateAttributes);
-        attributes.removeAll(Collections.singleton(null));
+        var attributes = KeycloakModelUtils.resolveAttribute(userModel, userAttribute, aggregate).stream()
```
Captain-P-Goldfish: please do not use `var`-type

services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialBuildConfig.java
```diff
@@ -85,22 +88,27 @@ public class CredentialBuildConfig {
...
-                                          .setCredentialType(credentialConfiguration.getVct())
-                                          .setTokenJwsType(credentialModel.getTokenJwsType())
-                                          .setNumberOfDecoys(credentialModel.getSdJwtNumberOfDecoys())
-                                          .setSigningKeyId(credentialModel.getSigningKeyId())
-                                          .setSigningAlgorithm(credentialConfiguration.getCredentialSigningAlgValuesSupported()
-                                                                                      .get(0))
-                                          .setHashAlgorithm(credentialModel.getHashAlgorithm())
-                                          .setSdJwtVisibleClaims(credentialModel.getSdJwtVisibleClaims());
+        var signingAlgSupported = credConfig.getCredentialSigningAlgValuesSupported();
+        var buildConfig = new CredentialBuildConfig()
```
Captain-P-Goldfish: `var`-type

services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
```diff
@@ -105,10 +106,67 @@ public class DeclarativeUserProfileProviderFactory implements UserProfileProvide
...
+        //     "validations": {
+        //       "pattern": {
+        //         "pattern": "^did:.+:.+$",
+        //         "error-message": "Value must start with 'did:scheme:'"
+        //       }
+        //     }
+        //   }
+        
+        if (!defaultConfig.hasAttribute("did")) {
+            var attr = new UPAttribute("did");
```
Captain-P-Goldfish: `var`-type

core/src/main/java/org/keycloak/sdjwt/DisclosureSpec.java
```diff
@@ -159,19 +159,25 @@ private void validateRedList() {
...
 
             // Validate undisclosed claims
-            if (redListedClaimNames.containsRedListedClaimNames(undisclosedClaims.keySet())) {
-                throw new IllegalArgumentException("UndisclosedClaims contains red listed claim names");
-            }
+            undisclosedClaims.keySet().forEach( it -> {
+                if (redListedClaimNames.isRedListedClaimName(it)) {
+                    throw new IllegalArgumentException("UndisclosedClaims contains red listed claim name: " + it);
+                }
+            });
```
Captain-P-Goldfish, line 166: the previously called method `containsRedListedClaimNames` is unused now.

server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
```diff
@@ -42,8 +42,10 @@
  */
 public class CredentialScopeModel implements ClientScopeModel {
 
-
-    public static final String SD_JWT_VISIBLE_CLAIMS_DEFAULT = "id,iat,nbf,exp,jti";
+    // Also see DisclosureRedList#redList
+    // https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-11.html#section-3.2.2.2
+    public static final String SD_JWT_VISIBLE_CLAIMS_DEFAULT = "iss,nbf,exp,cnf,vct,status";
```
Captain-P-Goldfish: I think we should keep `id,iat,jti` as default visible claims.
What is your concrete reason for removing them?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -47,47 +55,52 @@ public String getSupportedFormat() {
...
-                        //  Needs further discussion on the wanted behavior.
-
-                        IntStream.range(0, listValue.size())
-                                .forEach(i -> disclosureSpecBuilder
-                                        .withUndisclosedArrayElt(entry.getKey(), i, SdJwtUtils.randomSalt())
-                                );
-                    } else {
-                        disclosureSpecBuilder.withUndisclosedClaim(entry.getKey(), SdJwtUtils.randomSalt());
-                    }
+        List<String> outerClaims = credentialBuildConfig.getSdJwtVisibleClaims();
```
Captain-P-Goldfish: could you rename this variable to `visibleClaims`?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -49,54 +54,68 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
...
-        claims.put(propertyName,
-                   mapperModel.getConfig().getOrDefault(OID4VCMapper.CLAIM_NAME,
-                                                        String.format("urn:uuid:%s", UUID.randomUUID())));
+        var userAttr = KeycloakModelUtils.resolveAttribute(userModel, USER_ATTRIBUTE_NAME_DID,false).stream()
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElse(null);
+        if (userAttr != null) {
+            claims.put(propertyName, userAttr);
+        }
```
Captain-P-Goldfish, line 104: You are reading the subject hardcoded by the `did`-key, and you ignore the configured value under `ID Property Name`.
I have basically no objections to set the subject to null if no value is present, or read it by default from the `did`-attribute, but you SHOULD keep the attributeName configurable as the configuration option is available by the config: 

```java
UserModel userModel = userSessionModel.getUser();
        List<String> attributePath = getMetadataAttributePath();
        String propertyName = attributePath.get(attributePath.size() - 1);
        String userAttributeName = mapperModel.getConfig().get(OID4VCMapper.USER_ATTRIBUTE_KEY);
        String userAttr = KeycloakModelUtils.resolveAttribute(userModel, userAttributeName ,false).stream()
                .filter(Objects::nonNull)
                .findFirst()
                .orElseGet(() -> {
                   return "id".equals(userAttributeName) ? userModel.getId() : null;
                });
```

services/src/main/java/org/keycloak/userprofile/config/UPConfigUtils.java
```diff
@@ -351,7 +355,7 @@ private static UPConfig parseConfig(InputStream is) {
         }
     }
 
-    private static InputStream getSystemDefaultConfig() {
-        return UPConfigUtils.class.getResourceAsStream(SYSTEM_DEFAULT_CONFIG_RESOURCE);
+    private static InputStream getUserProfileConfig(String resource) {
```
Captain-P-Goldfish, line 361: this inputstream is never closed if called by

```java
public static UPConfig parseUserProfileConfig(String resource) {
        return parseConfig(getUserProfileConfig(resource));
    }
```

Captain-P-Goldfish (changes requested):

Captain-P-Goldfish:
I would also like to hear the opinion of the adorsys developers on this PR @IngridPuppet @Ogenbertrand @francis-pouatcha , WDYT?

server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
IngridPuppet: Guard against correlation by default?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -47,47 +55,52 @@ public String getSupportedFormat() {
...
+
+        // Retrieve subject claims
         CredentialSubject credentialSubject = verifiableCredential.getCredentialSubject();
-        Map<String, Object> claimSet = credentialSubject.getClaims();
+        Map<String, Object> claims = new LinkedHashMap<>(credentialSubject.getClaims());
+
+        // Add inner (disclosed) claims iat, sub - the latter being derived from Subject.id
+        Optional.ofNullable(issuanceDate).ifPresent(it -> {
+            claims.put(CLAIM_NAME_IAT, it.getEpochSecond());
+        });
```
IngridPuppet: So far, we're leaving this to be configured by a mapper so one can choose not to include a `iat` claim. Also, there are additional considerations for correlation prevention by rounding/randomizing. You'd maybe want to check this part:

https://github.com/keycloak/keycloak/blob/50179d165cbe241df63a67892b373deadd574148/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java#L1301-L1319

The claims are set there for all supported formats but not necessarily ported to SD-JWTs issued.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -47,47 +55,52 @@ public String getSupportedFormat() {
...
+        // https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-11.html#section-3.2.2.2
 
-        // jti, nbf, iat and exp are all optional. So need to be set by a protocol mapper if needed.
-        // see: https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#name-registered-jwt-claims
+        claims.put(CLAIM_NAME_ISSUER, credentialBuildConfig.getCredentialIssuer());
+        claims.put(CLAIM_NAME_VCT, credentialBuildConfig.getCredentialType());
+        if (issuanceDate != null && expirySeconds != null) {
+            Instant exp = issuanceDate.plus(Duration.ofSeconds(expirySeconds));
+            claims.put(CLAIM_NAME_EXP, exp.getEpochSecond());
+        }
```
IngridPuppet: I guess the goal here is to always include an exp claim, which makes sense to me. However, consider just setting `verifiableCredential.getExpirationDate()` with regard to correlation mitigation. That value should already be rounded/randomized. See https://github.com/keycloak/keycloak/issues/43399

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
IngridPuppet: I agree the source user attribute should be configurable and also tend to believe it was the intent of the PR, given the introduction of the `USER_ATTRIBUTE_KEY` configuration entry.

services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
```diff
@@ -105,10 +106,67 @@ public class DeclarativeUserProfileProviderFactory implements UserProfileProvide
...
+            if (Profile.isFeatureEnabled(OID4VC_VCI)) {
+                addDefaultAttributesForVerifiableCredentials(defaultConfig);
+            }
             PARSED_DEFAULT_RAW_CONFIG = defaultConfig;
         }
     }
 
+    private static void addDefaultAttributesForVerifiableCredentials(UPConfig defaultConfig) {
+
+        // For all intends and purposes a Verifiable Credential (VC) is bound to a Holder's Digital Identity (DID).
```
IngridPuppet:
```suggestion
        // For all intents and purposes, a Verifiable Credential (VC) is bound to a Holder's Digital Identity (DID).
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtIssuingEndpointTest.java
```diff
@@ -551,7 +551,6 @@ protected void handleCredentialResponse(CredentialResponse credentialResponse, C
             JsonWebToken jsonWebToken = TokenVerifier.create(sdJwtVP.getIssuerSignedJWT().getJws(), JsonWebToken.class).getToken();
 
             assertNotNull("A valid credential string should have been responded", jsonWebToken);
-            assertNotNull("The credentials should include the id claim", jsonWebToken.getId());
             assertNotNull("The credentials should be included at the vct-claim.", jsonWebToken.getOtherClaims().get("vct"));
```
IngridPuppet: I guess this is now an outer claim. Can we relocate the assertion accordingly?

services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialBuildConfig.java
IngridPuppet: I'm not on the anti-` var` side, but yes, for non-complex types such as `String`'s, I would agree it does more harm than good 😅

IngridPuppet (commented):
Hello @tdiesler and @Captain-P-Goldfish - I checked the PR and left some comments. Thank you, @tdiesler, for contributing the PR. Please could you check my comments?

server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
tdiesler: The spec says ...

> SD-JWT VCs MAY use any claim registered in the "JSON Web Token Claims" registry as defined in [[RFC7519](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-11.html#RFC7519)].
>
> The following registered JWT claims are used within the SD-JWT component of the SD-JWT VC and MUST NOT be included in the Disclosures, i.e., cannot be selectively disclosed: "iss", "nbf", "exp", "cnf", "vct", "vct#integrity", "status"

i.e. I've taken the list of visible claims from [section 3.2.2.2](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-11.html#section-3.2.2.2) as referenced in the comment

* there is no 'id' claim in rfc7519
* iat, jti, sub are not part of of the 3.2.2.2 list - so I put them into the disclosures

i.e. we are now secret about the iat and the vc subject - exp is supposed to be public

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
tdiesler: ok, will do ...

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
tdiesler: merci

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCSdJwtIssuingEndpointTest.java
tdiesler: we have this now

```
            assertTrue("The credentials should include the id claim.", disclosureMap.containsKey("sub")); // id -> sub
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
tdiesler: now using

```
claims.put(CLAIM_NAME_EXP, verifiableCredential.getExpirationDate());
```

tdiesler:
This is now pending on a decision about the set of visible claims
https://github.com/keycloak/keycloak/pull/44029#discussion_r2585309327

Currently, it would change from 

```
id,iat,nbf,exp,jti
```
to

```
iss,nbf,exp,cnf,vct,status
```

server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
Captain-P-Goldfish: Yes, I’m aware of this section.  
However, from a use-case perspective, I would argue that `id`, `iat`, and especially `jti` are disclosed only in very rare scenarios. In typical usage, these fields should be visible, and since their visibility remains configurable, I believe the default should prioritize convenience for most users.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
Captain-P-Goldfish: the userAttribute cannot be set to "id".We have one use-case at least where this will be necessary.

Captain-P-Goldfish (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
tdiesler: Like this?

```
        Consumer<String> userIdConsumer = (val) -> claims.put(propertyName, val);
        if (UserModel.ID.equals(userAttributeName)) {
            userIdConsumer.accept(userModel.getId());
        } else {
            KeycloakModelUtils.resolveAttribute(userModel, userAttributeName, false).stream()
                    .filter(Objects::nonNull)
                    .findFirst()
                    .ifPresent(userIdConsumer);
        }
```

I like it to be explicit i.e. not as fallback

Captain-P-Goldfish: yes, that is much better than my proposal! :-)
thx.

server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
tdiesler: There are quite some changes in response to your lates comment about id, jti ...

I tried to clarify their meaning here. 

```
    // jti uniquely identifies a SD_JWT instance itself, not the credential, and not the subject
    // If the issuer re-issues another credential later, even with the same underlying claims, it will get a new jti
    // It is useful for:
    //   * replay protection of the SD-JWT
    //   * introspection caches
    //   * deduplication
    //   * credential revocation tracking (optional)
    public static final String CLAIM_NAME_JTI = "jti";

    // id is the credential identifier
    //   * A stable identifier for the VC instance
    //   * May be dereferenceable (URL/URN)
    //   * May be used by verifiers to reference, store, or reason about the credential
    //   * Is independent of the JWT representation
    public static final String CLAIM_NAME_VC_ID = "id";
```

I also changed how the mappers work:

* OID4VCSubjectIdMapper now maps the user did -> jwt.sub
* OID4VCGeneratedIdMapper generates a random vc id -> jwt.id (disclosed) 
* SdJwtCredentialBuilder always generates a new random jti

Note, that OID4VCGeneratedIdMapper now generates a vc claim and not a subject claim any more.

When you run OID4VCSdJwtIssuingEndpointTest.testCredentialIssuance you will see

```
JsonWebToken: {
  "exp" : 1764943371,
  "nbf" : 1764943356,
  "jti" : "urn:uuid:dc0df7d4-4e92-4990-9b72-53ccab6f721c",
  "iss" : "https://localhost:8543/auth/realms/test",
  "_sd" : [ "-3dWj2E8OPMWxjeHtQgT2CCsHspM05fLF-8SjRTMK50", "4CprLCjDBXzcJX77gY3HNJV4Pe3C7p0yUrx5LcMoQbc", "9Jfbfc4oQzXrDB6VVNigft05uTTTga1I0dHLl2uZUVM", "9KwF4Lz4A2J-5BmHT5mJdZwRXfIpeugdYwP2RZcyPtY", "AF60l5m-HD5eTc-EzC3CnKB41rjDmomADkqkalINitw", "BOilfuvXjFSahzZ0Q6UQiDy-suiH075OCNn6Pzai5XQ", "CPhqKosHBj9UQXX0-HKOeVsDPl0M-MCx8x38FcdF7o8", "IOTnrOwA4PzN3mwRaSds_QS-6dKWUJVCvDFEyIxwtVg", "WU6tuzGvN7j5EnuJsOojepTtGmQLc5uT46-2gz4rJ6M", "fnI-3dqpBeGQMTTGHZ_txF3bHS1esQhvMkZf4WLIMK0", "ldWlCzpTJ0rAclkVptVsDnf5E72MG_rDQZyQB5NEY2I", "mNxl7vY0j5HE3v5A7FB92bq_AH7YTlCF1GSyDZCI_Iw", "nlgTv950UqP466pPWjh8mWYa6bGJK7iJLLrCUskVOls", "pc2wSxCwuhDHxQmiMKN5JOMFoUku3_W__A8CdvOePpg", "ta6K2MGOB8loJsqNtYENCg3gYaT-OobTihuo3FauERA", "uXXQfITpvfKHSQ88-tUEcUwvZMJZ5qW5BZALVWYLLnQ", "xEkBoh7hRyXp1ijl0Ow6MkuHdGIxHY1quV7NiQVWK_A" ],
  "_sd_alg" : "SHA-256",
  "vct" : "https://credentials.example.com/SD-JWT-Credential"
}

["GsBa5jpZ8THaF9qghkN1qg","scope-name","sd-jwt-credential"]
["F0bI410ZwyapJRVypiq8pQ","lastName","Doe"]
["q2PhpsojuyiaKn-kS80xAg","firstName","John"]
["9hSvu88HTUnUsmCGjGY7SA","sub","did:key:1234"]
["bpSuV6jas5AfEhAYfwb-uw","iat",1764943356]
["LThe1RqhRFhSRSYJWzzDHg","email","john@email.cz"]
["YO9t7960OJpUZX9w9Iho0Q","id","urn:uuid:bbc4d20c-8463-49d9-9b63-ddc2fff5ce16"]
```

jti is added back to visible claims. 
id and iat are still in the disclosures. Note, that id != jti

If you still think that id + iat should be promoted to visible claims, I can sure do that and we can become more strict later when/if a compatibility test requires it.

The tests are now very explicit in what they expect

```
            // SDJWT have a special format.
            SdJwtVP sdJwtVP = SdJwtVP.of(credentialResponse.getCredentials().get(0).getCredential().toString());
            JsonWebToken jsonWebToken = TokenVerifier.create(sdJwtVP.getIssuerSignedJWT().getJws(), JsonWebToken.class).getToken();
            Map<String, Object> otherClaims = jsonWebToken.getOtherClaims();

            log.infof("JsonWebToken: %s", valueAsPrettyString(jsonWebToken));
            assertNotNull("Expected jti claim", jsonWebToken.getId());
            assertNotNull("Expected exp claim", jsonWebToken.getExp());
            assertNotNull("Expected nbf claim", jsonWebToken.getNbf());
            assertNotNull("Expected iss claim", jsonWebToken.getIssuer());

            assertNull("Unexpected aud claim", jsonWebToken.getAudience());
            assertNull("Unexpected iat claim", jsonWebToken.getIat());

            assertTrue(otherClaims.keySet().containsAll(Set.of(CLAIM_NAME_SD, CLAIM_NAME_SD_HASH_ALGORITHM, CLAIM_NAME_VCT))); // may also contain cnf
            assertEquals("_sd_alg claim mapped correctly", "SHA-256", otherClaims.get("_sd_alg"));
            assertEquals("vct claim mapped correctly", vct, otherClaims.get("vct"));

            Map<String, String> disclosureMap = sdJwtVP.getDisclosures().values().stream()
                    .map(disclosure -> {
                        try {
                            JsonNode jsonNode = JsonSerialization.mapper.readTree(Base64Url.decode(disclosure));
                            return Map.entry(jsonNode.get(1).asText(), jsonNode); // Create a Map.Entry
                        } catch (IOException e) {
                            throw new RuntimeException(e); // Re-throw as unchecked exception
                        }
                    })
                    .peek(en -> log.infof("%s", en.getValue()))
                    .map(en -> Map.entry(en.getKey(), en.getValue().get(2).asText()))
                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));

            assertTrue("Expected id claim", disclosureMap.containsKey("id"));
            assertTrue("Expected iat claim", disclosureMap.containsKey("iat"));
            assertFalse("Unexpected other claim", disclosureMap.containsKey("given_name"));

            assertEquals("sub claim mapped correctly", "did:key:1234", disclosureMap.get("sub"));
            assertEquals("firstName claim mapped correctly", "John", disclosureMap.get("firstName"));
            assertEquals("lastName claim mapped correctly", "Doe", disclosureMap.get("lastName"));
            assertEquals("scope-name claim mapped correctly", clientScope.getName(), disclosureMap.get("scope-name"));
            assertEquals("email claim mapped correctly", "john@email.cz", disclosureMap.get("email"));
```

tdiesler:
This is now failing in some Admin UI test

```
1) [chromium] › test/client-scope/oid4vci-mappers.spec.ts:195:3 › OID4VCI Protocol Mapper Configuration › should work with all OID4VC mapper types 

    Test timeout of 30000ms exceeded.

    Error: locator.click: Test timeout of 30000ms exceeded.
    Call log:
      - waiting for getByText('CredentialSubject ID Mapper', { exact: true })


      23 |
      24 | async function selectMapperType(page: Page, mapperType: string) {
    > 25 |   await page.getByText(mapperType, { exact: true }).click();
         |                                                     ^
      26 |   await page.getByTestId("name").waitFor({ state: "visible" });
      27 | }
      28 |
        at selectMapperType (/home/runner/work/keycloak/keycloak/js/apps/admin-ui/test/client-scope/oid4vci-mappers.spec.ts:25:53)
        at /home/runner/work/keycloak/keycloak/js/apps/admin-ui/test/client-scope/oid4vci-mappers.spec.ts:223:13
```

It may be because of the switch from jti -> id in oid4vc-generated-id-mapper 
or because of the switch from id -> sub in oid4vc-subject-id-mapper

Assuming that this is what we want, I could unfortunately not find how to align the js test with that change.

tdiesler:
Admin UI tests are fixed

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCGeneratedIdMapper.java
```diff
@@ -65,29 +68,33 @@ protected List<ProviderConfigProperty> getIndividualConfigProperties() {
...
-                                       UserSessionModel userSessionModel) {
-        // nothing to do for the mapper.
+    public void setClaimsForCredential(VerifiableCredential verifiableCredential, UserSessionModel userSessionModel) {
+        // Assign a generated ID
+        URI vcId = URI.create(String.format("urn:uuid:%s", UUID.randomUUID()));
+        List<String> attributePath = getMetadataAttributePath();
+        String propertyName = attributePath.get(attributePath.size() - 1);
+        if (CLAIM_NAME_VC_ID.equals(propertyName)) {
+            verifiableCredential.setId(vcId);
+        } else {
```
Captain-P-Goldfish, line 90: I actually thought we should set the id of the user instead of generating a random UUID in this case?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -49,44 +56,51 @@ public String getSupportedFormat() {
...
+        Instant issuanceDate = verifiableCredential.getIssuanceDate();
+        Instant expirationDate = verifiableCredential.getExpirationDate();
+
+        // Retrieve subject claims
         CredentialSubject credentialSubject = verifiableCredential.getCredentialSubject();
-        Map<String, Object> claimSet = credentialSubject.getClaims();
+        Map<String, Object> claims = new LinkedHashMap<>(credentialSubject.getClaims());
+
+        // Always add a new jti - this cannot and should not be mapped
+        claims.put(CLAIM_NAME_JTI, String.format("urn:uuid:%s", UUID.randomUUID()));
```
Captain-P-Goldfish: the jti is used to mitigate replay-attacks. We should use a more unpredictable value than a UUID.
in that case the urn:uuid should also be removed

core/src/main/java/org/keycloak/representations/userprofile/config/UPConfig.java
```diff
@@ -111,6 +111,16 @@ public UPAttribute getAttribute(String name) {
...
 
+    @JsonIgnore
+    public boolean hasAttribute(String name) {
+        for (UPAttribute attribute : getAttributes()) {
+            if (attribute.getName().equals(name)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
Captain-P-Goldfish: method is unused

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -495,23 +489,17 @@ public Builder withExpCheck() {
         }
 
         public Builder withExpCheck(boolean isCheckOptional) {
-            contentVerifiers.removeIf(verifier -> {
-                return verifier instanceof ClaimVerifier.ExpCheck ||
-                    (verifier instanceof ClaimVerifier.ClaimCheck
-                        && ((ClaimCheck) verifier).getClaimName().equalsIgnoreCase(OID4VCConstants.CLAIM_NAME_EXP));
-            });
+            contentVerifiers.removeIf(verifier ->
+                    isTypeOrNamedCheck(verifier, ClaimVerifier.ExpCheck.class, OID4VCConstants.CLAIM_NAME_EXP));
```
Captain-P-Goldfish, line 493: you have remove the fallback checks in case that custom-verifiers were added. Please readd these checks for the methods

Captain-P-Goldfish (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCGeneratedIdMapper.java
tdiesler: This mapper is about the id for the credential itself (vc_id)
The OID4VCSubjectIdMapper is about the id for credentialSubject (sub_id)
In case of SD, there is also the id of the JWT instance (jti), which cannot be mapped.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
tdiesler: done

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
tdiesler: There were a number of withFoo methods that always removed a claim verifier of a given type or of a given name.

I tried to remove duplication of that logic with these two predicate functions. 

```
        private boolean isTypeOrNamedCheck(Predicate<?> it, Class<?> type, String claimName) {
            return type.isInstance(it) || isNamedCheck(it, claimName);
        }

        private boolean isNamedCheck(Predicate<?> it, String claimName) {
            return it instanceof ClaimCheck && ((ClaimCheck) it).getClaimName().equalsIgnoreCase(claimName);
        }
```

This remove logic is supposed to be equivalent to what it was before.

A custom verifier would be removed if withFoo uses the same type or name of that custom verifier. Have I missed something here?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCGeneratedIdMapper.java
Captain-P-Goldfish: fair point. Lets drop this one then.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
Captain-P-Goldfish: yes. Seems I read this part too fast... We can drop this one as well :-)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/JwtCredentialBuilder.java
```diff
@@ -90,12 +93,15 @@ public JwtCredentialBody buildCredentialBody(
...
-                        verifiableCredential
-                                .getCredentialSubject()
-                                .getClaims()
-                                .get(ID_CLAIM_KEY))
+        // Map the subject id claim to 'id'
+        // We can't use claim name 'id' directly because it clashes with vc_id
+        CredentialSubject subject = verifiableCredential.getCredentialSubject();
+        Optional.ofNullable(subject.getClaims().remove(CLAIM_NAME_SUBJECT_ID)).ifPresent( id -> {
+            subject.getClaims().put(ID_CLAIM_KEY, id);
+        });
```
Captain-P-Goldfish: I don’t think this is a good idea.
If someone has intentionally added a mapper for the sub claim that is executed earlier, this code removes it again and writes the value into the id claim instead.
In my opinion, this is unexpected behavior from the perspective of the configuring admin.

Captain-P-Goldfish (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/JwtCredentialBuilder.java
tdiesler: AFAICS, there is currently just one namespace for claims produced by OID4VCGeneratedIdMapper and OID4VCSubjectIdMapper. Both want to create an 'id' claim - the former the CredentialId, the latter the SubjectId.

OID4VCSubjectIdMapper currently uses a hack and puts the subject id into the 'sub' claim, which works for SD VCs but not for the other formats - there it must be called credentialSubject.id.

The above is reasonably safe imho, because it does not make sense to have a 'credentialSubject.sub' claim for these non-sd formats. In any case, we need to find some way to distinguish the vc.id from the subject.id claim.

Perhaps all the subject claim mappers should use some known prefix e.g. 'subject.' as their target claim name. Then , all the credential builders would have to remove that prefix and put the subjects claims in their proper location.

In the future we might want to support hierarchical mappers to map stuff deeper in the claim tree when there are more complex claim layouts.

Anyway, I believe this can be a follow-up issue - for now we only need to make a start to have distinct values for jti, vc.id and vc.credentialSubject.id

PS: I wonder why we even need to have a mapper for the vc.id? This isn't something a user would want to choose, or is it? Perhaps we could get rid of vc.id mapper, which would then free the 'id' claim name and these hacks in (all three) vc builders could go away too.

Captain-P-Goldfish: We have a usecases where we want to use the ids from remote providers. We have IdPs configured with the transient-users feature. It is not possible to set a specific ID for the transient user so we need to use the user-attributes. We want also to preserve the returned IDs from the IdP for the vc.id and the subject.id. In that case this part would be a problem.

tdiesler: Understood. I think I found a way to preserve the mapping to the subject 'id' claim from an arbitrary user attribute while at the same time removing the conflict with vc.id in the claim model.

In the SdJwtCredentialBuilder we now do this ...

```
        // Map subject id => sub
        Optional.ofNullable(claims.remove(CLAIM_NAME_SUBJECT_ID)).ifPresent(it ->
                claims.put(CLAIM_NAME_SUB, it)
        );

        // Add credential id
        Optional.ofNullable(vcId).ifPresent(it ->
                claims.put(CLAIM_NAME_VC_ID, it)
        );
```

The other CredentialBuilders can use the subject 'id' claim without additional mapping hacks. 
Would this work for you?

tdiesler:
This PR has become a bucket of a number of loosely related id issues. Since it is now hopelessly broken due to a large number of merge conflicts, I think its best to address these id issues in separate (much smaller) PRs

PR #44030, opened 2025-11-07
Client policy to enforce only downscoping in Token Exchange

### LINKED ISSUE #43931: Downscoping for standard token exchange
### Description

It should be possible to limit that scopes in the issued token are only subset of the scopes from original `subject_token`.

#### Example scenario

Subject_token issued to client `client1` with the token with `scopes: [ "foo", "bar" ]` .

Then token exchange requested by client `client2` . But `client2` has default client scopes `bar` and `bar2` and additional optional scope `bar3`, which was requested by scope parameter `scope=bar3` . So with downscoping `bar2` and `bar3` cannot be included in the issued token as they are not included in the original subject_token. So how to proceed?:

- Option 1: Reject the request

- Option 2: Remove the scopes and include only scope `bar`

Maybe it should be configurable whether we should reject the request or remove the non-matching scopes (scopes `bar2` and `bar3` from the example above).

#### Scopes not included in token scope

Note that some client scopes may be set with configuration option `Include in token scope=false` and hence they will never be included in the subject token. Such client scopes should be probably considered permitted for the token-exchange (not sure if this should be configurable, but my vote is to always allow them rather than introducing another option whether to permit them or not. At least initially until we see some requirement for the above).

### Goals

- Allow downscoping for the standard token exchange requests to make sure that exchanged token cannot have bigger permissions than the original token

### Non-Goals

- Just the goals mentioned in the description are in scope of this issue

### Discussion

_No response_

### Notes

Should be probably handled by the client policies. There can be dedicated client policy executor with the config option for the above. Hopefully this client policy executor can be re-used also for the use-case of "Downscoping" for authorization grants (See #43573 for the details).

rmartinc:
Closes #43931

The downscope enforcer executor for token exchange that is used to restrict granted scopes to the ones that are present in the subject token. I think that I have followed all that we have been discussing:

* The scopes defined in the client that are default and not included in access token are always included. Those scopes are invisible and are considered compulsory (acr, basic,...). The administrator can configure them too.
* All the rest of scopes (defaults that are included in token or optional that are not included in token) are restricted. In some sense the requester knows about them, or they are present in the subject_token, or they need to be explicitly requested.
* The assignment/calculation of scopes is the same. I mean, default scopes are always added but restricted. So you cannot downscope default scopes (they are restricted to the ones present in the assertion token but you cannot request to not assign them). In theory we can do that just adjusting the `restrictedScopes` variable in the executor, but we need to change the current meaning for the `scope` parameter when this enforcer is used (you would need to include even default scopes in the parameter). I think it makes no sense, the calculation should be the same and the admins can configure default/optional scopes as they want for each client.
* The executor returns an error if any scope that is not present in the access token is requested. If it passes this check, the `restrictedScopes` set is used to ensure that only the original scopes (plus invisible ones) are granted in the token.

It's a draft for the moment. Maybe we need a release note, but I don't know if we want to wait to include this executor for the JWT authorization grant too (we will need some modifications for that).

### DIFF
--- a/services/src/main/java/org/keycloak/services/clientpolicy/executor/DownscopeAssertionGrantEnforcerExecutorFactory.java
+++ b/services/src/main/java/org/keycloak/services/clientpolicy/executor/DownscopeAssertionGrantEnforcerExecutorFactory.java
@@ -59,9 +59,10 @@ public class DownscopeAssertionGrantEnforcerExecutorFactory implements ClientPol
     @Override
     public String getHelpText() {
         return """
-               It ensures that the scopes in the final access token are limited to the ones already present in the assertion passed.
-               For the moment the executor only applies to Standard Token Exchange and the limitation is done over the scopes
-               that are present in the presented subject_token.
+               It ensures that the scopes in the final access token are limited to the ones already present in the JWT assertion passed.
+               For the moment, the executor applies to certain grants where some initial token/assertion is passed (for example
+               subject_token in case of Standard Token Exchange grant). The limitation is done over the scopes that are
+               present in the initial token/assertion, returning an error if any extra scope is requested.
                """;
     }
 
--- a/services/src/main/java/org/keycloak/services/clientpolicy/executor/DownscopeAssertionGrantEnforcerExecutorFactory.java
+++ b/services/src/main/java/org/keycloak/services/clientpolicy/executor/DownscopeAssertionGrantEnforcerExecutorFactory.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.services.clientpolicy.executor;
+
+import java.util.Collections;
+import java.util.List;
+import org.keycloak.Config.Scope;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.provider.ProviderConfigProperty;
+
+/**
+ * <p>Factory that enforces the grant to only use scopes that are already present in
+ * the initial assertion. For the moment it can only be used in the Token Exchange
+ * context.</p>
+ *
+ * @author rmartinc
+ */
+public class DownscopeAssertionGrantEnforcerExecutorFactory implements ClientPolicyExecutorProviderFactory {
+
+    public static final String PROVIDER_ID = "downscope-assertion-grant-enforcer";
+
+    @Override
+    public ClientPolicyExecutorProvider create(KeycloakSession session) {
+        return new DownscopeAssertionGrantEnforcerExecutor(session);
+    }
+
+    @Override
+    public void init(Scope config) {
+    }
+
+    @Override
+    public void postInit(KeycloakSessionFactory factory) {
+    }
+
+    @Override
+    public void close() {
+    }
+
+    @Override
+    public String getId() {
+        return PROVIDER_ID;
+    }
+
+    @Override
+    public String getHelpText() {
+        return """
+               It ensures that the scopes in the final access token are limited to the ones already present in the assertion passed.
+               For the moment the executor only applies to Standard Token Exchange and the limitation is done over the scopes
+               that are present in the presented subject_token.
+               """;
+    }
+
+    @Override
+    public List<ProviderConfigProperty> getConfigProperties() {
+        return Collections.emptyList();
+    }
+}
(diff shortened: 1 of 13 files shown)


services/src/main/java/org/keycloak/services/clientpolicy/executor/DownscopeAssertionGrantEnforcerExecutorFactory.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public String getId() {
+        return PROVIDER_ID;
+    }
+
+    @Override
+    public String getHelpText() {
+        return """
+               It ensures that the scopes in the final access token are limited to the ones already present in the assertion passed.
```
mposolda: My vote is to update this message to something more generic from the beginning. Otherwise, there is a risk that we forgot to update this when adding support for rfc-7523 and this message will become outdated :-)

How about something like:
```
It ensures that the scopes in the final access token are limited to the ones already present in the assertion passed.
For the moment the executor applies to certain grants where some initial token/assertion is passed (For example 
subject_token in case of Standard token exchange grant) and the limitation is done over the scopes that are 
present in the presented initial token/assertion.
```

Probably applies to javadoc of this class as well.

services/src/main/java/org/keycloak/services/util/DefaultClientSessionContext.java
```diff
@@ -245,16 +249,20 @@ public AuthorizationRequestContext getAuthorizationRequestContext() {
...
+                logger.tracef("User '%s' not permitted to have client scope '%s'",
+                        clientSession.getUserSession().getUser().getUsername(), clientScope.getName());
+            }
+            return false;
         }
 
-        if (logger.isTraceEnabled()) {
-            logger.tracef("User '%s' not permitted to have client scope '%s'",
-                    clientSession.getUserSession().getUser().getUsername(), clientScope.getName());
+        if (restrictedScopes != null && !restrictedScopes.contains(clientScope.getName())) {
```
mposolda: 2 nitpicks to this one:

- Maybe worth to add this check before the `isClientScopePermittedForUser` call? As `isClientScopePermittedForUser` could be relatively expensive (including DB calls for retrieve composite roles etc) and hence makes sense to rather avoid expensive call if the scope is restricted?

- The variable name `restrictedScopes` seems to be a bit confusing as those scopes looks rather those, which are not restricted. Especially interesting when the trace message below mentions `Client scope 'foo' is restricted...` just in the case when it is actually not included in the `restrictedScopes` list :-) Not sure about better name, but maybe something like `unrestrictedScopes` or `whitelistedScopes` ? This comment applies to various other method signatures and variable names...

mposolda (commented):
@rmartinc Thanks!

The behaviour is good to me. We will see if we have some feedback requesting more flexibility for this...

I've rather some inline minor comments. Can you check please? Besides that, some docs can be good. At least release notes update can be good. Also this section may nee to list new executor https://www.keycloak.org/docs/latest/server_admin/index.html#executor . Finally will be good to document downscoping in some section in more details (including the behaviour for the scopes with "Include in token scope = false" ) . Not sure if it should be [client scopes section](https://www.keycloak.org/docs/latest/server_admin/index.html#_client_scopes) or rather [token exchange section](https://www.keycloak.org/securing-apps/token-exchange#_standard-token-exchange-scope) . Maybe rather client scopes section with token-exchange section containing only the link to that client scopes docs?

graziang (commented):
@rmartinc Thanks, the final behavior makes sense to me after all the discussion 😄 . I just have a minor comment about the executor’s behavior. If I understood correctly, the executor is used to set the restricted scopes and only validate the scopes requested in the scope parameter of the token exchange request, while the final validation of the allowed scopes (like default scopes) is done later in `DefaultClientSessionContext`. My doubt is whether it would be possible to limit the validation to the executor only, to avoid introducing the `restrictedScopes` in `DefaultClientSessionContext`. I’m not sure if I’m mistaken and it's the only way to do it. WDYT?

rmartinc:
> @rmartinc Thanks, the final behavior makes sense to me after all the discussion 😄 . I just have a minor comment about the executor’s behavior. If I understood correctly, the executor is used to set the restricted scopes and only validate the scopes requested in the scope parameter of the token exchange request, while the final validation of the allowed scopes (like default scopes) is done later in `DefaultClientSessionContext`. My doubt is whether it would be possible to limit the validation to the executor only, to avoid introducing the `restrictedScopes` in `DefaultClientSessionContext`. I’m not sure if I’m mistaken and it's the only way to do it. WDYT?

If we want to also restrict the default scopes I think there is no other way. If we just want to manage the optional scopes yes, it would be possible. But restricting the default ones needs extra code in the `DefaultClientSessionContext`. Mainly because the default scopes are added always no matter the scope parameter we use.

rmartinc:
OK @mposolda! I'll try to add your comments the next week. I'll check what documentation is needed now we think this is more or less OK.

services/src/main/java/org/keycloak/services/clientpolicy/executor/DownscopeAssertionGrantEnforcerExecutorFactory.java
rmartinc: Done!

services/src/main/java/org/keycloak/services/util/DefaultClientSessionContext.java
rmartinc: Done! For the second point I have maintained the `restrictedScopes` name. The idea is that those scopes are restricted or limited, no other than those are allowed. So I have changed the trace log, now it says: `Client scope 'foo' is not among the restricted scopes list and will not be processed`.

rmartinc:
I have done the review comments. Added the documentation and the changes as a second commit. :smile:

mposolda (approved):
@rmartinc Thanks! I hope I can merge this once tests are OK (hopefully re-run helps).

PR #44046, opened 2025-11-07
fix: simplifying addResources

### LINKED ISSUE #44045: Isssues with operator addResources
### Description

The logic is used in muliple places, but we only have 1 set of defaults. We should not emit a debug log if someone has adjusted the update job memory below the main default.

The logic is also modifying the passed in ResourceRequirements which we should avoid doing.

shawkins:
closes: #44045

If we do want something like the previous debug message, it could be added as a warning to the Keycloak CR for the main container only - not the jobs.

Also null checks are generally not needed for limit/request maps, but I left those checks in for now.

One final thought that our defaulting may be a little counterintuative - if someone sets just the limit, the normal expectation is that the request will default to the limit. Here it will instead default to our value.

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/Utils.java
+++ b/operator/src/main/java/org/keycloak/operator/Utils.java
@@ -21,10 +21,10 @@ import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.ResourceRequirements;
+import io.fabric8.kubernetes.api.model.ResourceRequirementsBuilder;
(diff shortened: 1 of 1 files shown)


vmuzikar (approved):

> If we do want something like the previous debug message, it could be added as a warning to the Keycloak CR for the main container only - not the jobs.

That sounds like a good idea. A warn is in place, IMHO. We could do that here, or in a follow-up PR, both work for me – feel free to merge as is. The debug message didn't have much value.

> One final thought that our defaulting may be a little counterintuative - if someone sets just the limit, the normal expectation is that the request will default to the limit. Here it will instead default to our value.

That would be a breaking change.

shawkins:
> That would be a breaking change.

I didn't make a change, just noting that it's not intuitive for Kubernetes users.

PR #44053, opened 2025-11-08
Fix duplicate mapper save error by auto-navigating after creation

### LINKED ISSUE #43948: Adding a client mapper stays on creation screen, with error on second save
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When creating a mapper for a client scope, the UI stays on the "Add mapper" dialog. 

When the user then updates the data and presses "Save" again, there is an error message.

### Version

main

### Regression

### Expected behavior

After creating the mapper, either return to the list of mappers, or switch from create to edit mode

### Actual behavior

The UI stays in the edit mode, and displays the error message: 

> Could not create mapping: 'Protocol mapper exists with same name'

### How to Reproduce?

* Open the Admin UI
* Create a new client "test"
* Go to "Client scopes"
* Open the "test-dedicated" scope
* Add a mapper "Audience"
* Enter a name
* Click "Save"
* Click "Save" -> error message appears

### Anything else?

_No response_

Rathan-Naik:
After successfully creating a new protocol mapper in the Admin UI, the page now automatically navigates back to the mapper list. This prevents users from accidentally clicking Save again, which previously caused a duplicate mapper error.

Changes:
- Auto-navigate to mapper list after successful mapper creation
- Remove clickCancelButton workaround from existing test
- Add new test specifically for Audience mapper auto-navigation

Closes #43948

### DIFF
--- a/js/apps/admin-ui/src/client-scopes/details/MappingDetails.tsx
+++ b/js/apps/admin-ui/src/client-scopes/details/MappingDetails.tsx
@@ -170,6 +170,9 @@ export default function MappingDetails() {
         }
       }
       addAlert(t(`mapping${key}Success`), AlertVariant.success);
+      if (!isUpdating) {
+        navigate(toDetails());
+      }
     } catch (error) {
       addError(`mapping${key}Error`, error);
     }
--- a/js/apps/admin-ui/test/client-scope/mappers.spec.ts
+++ b/js/apps/admin-ui/test/client-scope/mappers.spec.ts
@@ -114,13 +114,18 @@ test.describe("Mappers tab", () => {
     await page.getByRole("button", { name: "Add mapper" }).click();
     await page.getByRole("menuitem", { name: "By configuration" }).click();
 
-    await page.locator('[role="dialog"]').getByText("Audience", { exact: true }).click();
+    await page
+      .locator('[role="dialog"]')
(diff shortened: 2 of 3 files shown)


ahus1:
@Rathan-Naik - thank you for this PR!

@ssilvert - can you please comment on the logic that we would like to see: After creating the mapper, should the UI return to the list of mappers, or switch from create to edit mode of this mapper? I see that in some other places (for example mappers for Identity Providers) switching to the edit mode is the logic I see there. Not sure if this is maybe a design principle in the UI. Thanks!

ssilvert:
@ahus1 @Rathan-Naik We don't have any hard and fast rule on this.  I think the decision should be about what the user will want to do after creation.  If it's likely that the user would want to keep working on the entity after creation then go to edit mode.  Otherwise, go back to the list.

For this, I think going back to the list is fine.

ahus1 (approved):
Thank you for this PR! 

I pushed one more change to fix an eslint problem that made the build fail. I hope the build will be green now.

Rathan-Naik:
> Thank you for this PR!
> 
> I pushed one more change to fix an eslint problem that made the build fail. I hope the build will be green now.

@ahus1  Thank you for the lint fix and merging this PR.

PR #44054, opened 2025-11-09
[rewrite] Apply `RemoveUnusedPrivateMethods`

Pankraz76:
happy CI:
- https://github.com/Pankraz76/keycloak/pull/9

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/jgroups/certificates/CertificateReloadManager.java
+++ b/model/infinispan/src/main/java/org/keycloak/jgroups/certificates/CertificateReloadManager.java
@@ -183,11 +183,6 @@ public class CertificateReloadManager implements Lifecycle {
         }
     }
 
-    private void onInvalidCertificate() {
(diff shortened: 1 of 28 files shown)


model/infinispan/src/main/java/org/keycloak/jgroups/certificates/CertificateReloadManager.java
```diff
@@ -183,11 +183,6 @@ private void bootReload() {
         }
     }
 
-    private void onInvalidCertificate() {
```
Pankraz76, line 186: this seems important but it seems unused.

Pankraz76:
why closing without any comment or hint?

Prone will find much more making it more difficult and risky.

- https://github.com/opensearch-project/OpenSearch/pull/19896

stianst:
We are not interested in contributions around this; and are gradually rolling out one thing at a time. We will consider removing unused private methods in due time.

PR #44058, opened 2025-11-10
added verifiable claim spi

sashyo:

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultUserProfile.java
+++ b/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultUserProfile.java
@@ -49,6 +49,11 @@ import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.storage.ReadOnlyException;
 import org.keycloak.utils.StringUtil;
 
+// >>> Added imports for Verifiable Claims SPI
+import org.keycloak.verifiableclaims.VerifiableClaimProvider;
+import org.keycloak.verifiableclaims.model.UpdateDecision;
+// <<<
+
 /**
  * <p>The default implementation for {@link UserProfile}. Should be reused as much as possible by the different implementations
  * of {@link UserProfileProvider}.
@@ -67,7 +72,7 @@ public final class DefaultUserProfile implements UserProfile {
     private UserModel user;
 
     public DefaultUserProfile(UserProfileMetadata metadata, Attributes attributes, Function<Attributes, UserModel> userCreator, UserModel user,
-            KeycloakSession session) {
+                              KeycloakSession session) {
         this.metadata = metadata;
         this.userSupplier = userCreator;
         this.attributes = attributes;
@@ -120,7 +125,31 @@ public final class DefaultUserProfile implements UserProfile {
         }
 
         try {
-            Map<String, List<String>> writable = new HashMap<>(attributes.getWritable());
+            // >>> Begin Verifiable Claims hook: allow provider to adjust attributes before persistence
+            Attributes effective = this.attributes;
+
+            VerifiableClaimProvider vcProvider = session.getProvider(VerifiableClaimProvider.class);
+            if (vcProvider != null) {
+                RealmModel realm = session.getContext().getRealm();
+                UserProfileContext context = metadata.getContext();
+
+                UpdateDecision decision = vcProvider.onAttributesAboutToUpdate(
+                        session, realm, user, (UserProfile) this, context, effective);
+
+                if (decision != null && decision.getToPersistNow() != null) {
+                    // Wrap the returned Map into DefaultAttributes for this context
+                    effective = new DefaultAttributes(
+                            context,
+                            decision.getToPersistNow(),   // Map<String, List<String>>
+                            user,
+                            metadata,
+                            session
+                    );
+                }
+            }
+            // <<< End Verifiable Claims hook
+
+            Map<String, List<String>> writable = new HashMap<>(effective.getWritable());
 
             for (Map.Entry<String, List<String>> attribute : writable.entrySet().stream().sorted(Map.Entry.comparingByKey()).toList()) {
                 String name = attribute.getKey();
@@ -157,10 +186,10 @@ public final class DefaultUserProfile implements UserProfile {
             if (removeAttributes) {
                 Set<String> attrsToRemove = new HashSet<>(user.getAttributes().keySet());
 
-                attrsToRemove.removeAll(attributes.nameSet());
+                attrsToRemove.removeAll(effective.nameSet());
 
                 for (String name : attrsToRemove) {
-                    if (attributes.isReadOnly(name)) {
+                    if (effective.isReadOnly(name)) {
                         continue;
                     }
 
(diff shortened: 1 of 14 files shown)


server-spi-private/src/main/java/org/keycloak/userprofile/DefaultUserProfile.java
```diff
@@ -226,7 +255,21 @@ public <R extends AbstractUserRepresentation> R toRepresentation() {
...
+        VerifiableClaimProvider vcProvider = session.getProvider(VerifiableClaimProvider.class);
+        if (vcProvider != null) {
+            RealmModel realm = session.getContext().getRealm();
+            UserProfileContext context = metadata.getContext();
+            Attributes enriched = vcProvider.enrichAttributesForRepresentation(session, realm, user, managed, context);
+            if (enriched != null) {
+                managed = enriched;
+            }
+        }
+        // <<< End Verifiable Claims hook
```
stianst, line 270: What's the reasoning behind this part? This is mainly just doing the user representation that is sent to the user/admin.

server-spi-private/src/main/java/org/keycloak/userprofile/DefaultUserProfile.java
```diff
@@ -120,7 +125,31 @@ private UserModel updateInternal(UserModel user, boolean removeAttributes, Attri
         }
 
         try {
-            Map<String, List<String>> writable = new HashMap<>(attributes.getWritable());
+            // >>> Begin Verifiable Claims hook: allow provider to adjust attributes before persistence
```
stianst, line 128: I would have expect this to be more a single line entry here; basically just removing the attributes that need attestation before running the rest as is.

stianst (commented):
I'm okay with using user profile attribute annotations for configuration, one point there though is that won't work for role and group mappings. This is mostly why I suggested the verifiable claims to be a separate configuration.

Using user attributes with suffixes IMO is not going to be a good approach for postponing the updates of attributes. It runs into issues like what happens if two updates happen to the same attribute, we need to make sure these attributes are not returned to rest endpoints, and can't be mapped into tokens, and it is not going to work to delay other types of updates to users like role mappings. It would also be impossible to do some form of batches where  group of updates can be attested as a single update, and would be hard to list outstanding attestations and their state.

A simple end-to-end test-case would be nice to not make sure things are actually working, but would help illustrate how it all fits together. I was honestly struggling a bit to follow the code here, and an example in a test would help a lot from that perspective. It should be relatively easy to create a basic test case in `tests/base` for this purpose.

Finally, this needs to be wrapped in an experimental Feature so it's disabled by default. Add a `Profile.Feature.VERIFIABLE_CLAIMS` or something then have the provider factory implement `EnvironmentDependentProviderFactory` with:

```
    public boolean isSupported(Config.Scope config) {
        return Profile.isFeatureEnabled(Profile.Feature.VERIFIABLE_CLAIMS);
    }
```

server-spi-private/src/main/java/org/keycloak/userprofile/DefaultUserProfile.java
pedroigor: Looks like what we need here is a subclass of `DefaultAttributes` with the necessary code to integrate with the `VerifiableClaimProvider` and enable verifiable claims.

We can consider creating the subclass based on some realm/user-profile setting so that when instantiating a `DefaultUserProfile` object, we check this setting and create a subclass instance instead of a `DefaultAttributes` instance. See https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProvider.java#L107.

We can also choose to add a new `UserProfileProvider`, probably based on the `DeclarativeUserProfileProvider`, to do the same.

pedroigor (commented):
Thanks @sashyo, looks like a great start for me. The integration looks simple and easy to follow. I've only a few suggestions on how we can plug the `VerifiableClaimProvider` so that we can decouple even more the core/default user profile logic and verifiable claims.

pedroigor:
In general, I think I see the path @sashyo is trying to follow, and from here we can start working on making it less user-profile/attribute dependent and make it more generic, so that we can also enable verifiable claims to other things like roles, and groups.

> I'm okay with using user profile attribute annotations for configuration, one point there though is that won't work for role and group mappings. This is mostly why I suggested the verifiable claims to be a separate configuration.

+1. Another argument for it is that annotations are metadata for UIs and not for server-side configuration.

> 
> Using user attributes with suffixes IMO is not going to be a good approach for postponing the updates of attributes. It runs into issues like what happens if two updates happen to the same attribute, we need to make sure these attributes are not returned to rest endpoints, and can't be mapped into tokens, and it is not going to work to delay other types of updates to users like role mappings. It would also be impossible to do some form of batches where group of updates can be attested as a single update, and would be hard to list outstanding attestations and their state.

+1. I separate table for storing verifiable claims is better and more suitable to handle data other than user attributes.

> 
> A simple end-to-end test-case would be nice to not make sure things are actually working, but would help illustrate how it all fits together. I was honestly struggling a bit to follow the code here, and an example in a test would help a lot from that perspective. It should be relatively easy to create a basic test case in `tests/base` for this purpose.
> 
> Finally, this needs to be wrapped in an experimental Feature so it's disabled by default. Add a `Profile.Feature.VERIFIABLE_CLAIMS` or something then have the provider factory implement `EnvironmentDependentProviderFactory` with:
> 
> ```
>     public boolean isSupported(Config.Scope config) {
>         return Profile.isFeatureEnabled(Profile.Feature.VERIFIABLE_CLAIMS);
>     }
> ```

pedroigor:
@stianst @sashyo How are we going to enable/disable verifiable claims to a realm?

server-spi-private/src/main/java/org/keycloak/verifiableclaims/VerifiableClaimProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import org.keycloak.userprofile.UserProfileContext;
+import org.keycloak.verifiableclaims.model.*;
+
+import java.util.Map;
+
+public interface VerifiableClaimProvider extends Provider {
+
+    Map<String, VerifiableAttributeConfig> getVerifiableAttributeConfig(KeycloakSession session, RealmModel realm);
+
+    UpdateDecision onAttributesAboutToUpdate(KeycloakSession session,
```
pedroigor, line 16: What if we have a `VerifiableClaim` type to abstract the different types of verifiable claims: user attributes, roles, groups, etc?

By doing that, we can make this provider more generic and capable of handling different types of verifiable claims. As it stands today, it is very specific to user profile/attributes.

For instance, the `onAttributesAboutToUpdate` would accept a `VerifiableClaim` object that wraps a user attribute, role grant, or group membership. We could also consider changing this method to accept a "update function" so that it will only be invoked (and the data persisted) if the attribute should be "persisted now".

I'm trying to think how to make methods on this provider more generic so that we can:

* Support data other than user profile/attributes
* Make it easier to integrate the `VerifiableClaimProvider` wherever we want to add its capabilities, where updates (or not) are decided and done within the provider

sashyo:
> @stianst @sashyo How are we going to enable/disable verifiable claims to a realm?

We enable the object itself individually. e.g CLIENT ROLE, REALM ROLE, ATTRIBTUE, GROUP or tide's case an entire JWT.
When this is enabled, it is enabled for the entire realm. UI perspective, a toggle and once toggled on, you can see a field to link to an attestor which will be manage separately. We agree with having a VerifiableClaim type abstract, im guessing this is going to be separate factories for each type so the methods align.

The table im thinking of adding
MetaData - Claims that are veriable in realm which links to a attestor in the attestor table
Drafting - look at gist, similar to what tide currently has.
Attestors -  holds all info about attestors e.g. name and json

PR #44061, opened 2025-11-10
Fix Admin Console crash when opening Client'a Authz Permission details

### LINKED ISSUE #44056: Allow only normalized URLs in requests caused a regression in view authz permission details in Admin Consol
### Before reporting an issue

### Area

admin/ui

### Describe the bug

After allowing only normalized URLs in requests (#43765), users can't view/edit permissions in Client Authorization tab in the Admin Console. Once the user clicks on a Permission the UI crashes with this error `missingNormalization`.

### Version

26.4.4

### Regression

### Expected behavior

User should be able to view a client permission.

### Actual behavior

UI crashes with error `missingNormalization`.

### How to Reproduce?

1. In the Admin Console, create a new Client. Enable Authorization.
2. Go to the `Authorization` tab. Then `Permissions` tab.
3. Click on the `Default Permission`.
4. The UI will crash.

Image

### Anything else?

* The issue doesn't happen if you set `http-accept-non-normalized-paths` to `true`.
* The cause of the issue is that the UI code calls the REST API `/auth/admin/realms/{realmId}/clients/{clientId}/authz/resource-server/policy/{policyType}/{policyId}` with empty `policyType` which results in double slashes in the URL.
* After checking the code, in `ResourcesPolicySelect.tsx` calls `findOnePolicy()` with empty `type` [here](https://github.com/keycloak/keycloak/blob/26.4.4/js/apps/admin-ui/src/clients/authorization/ResourcesPolicySelect.tsx#L179).

bahaa:
Closes #44056

### DIFF
--- a/js/apps/admin-ui/src/clients/authorization/ResourcesPolicySelect.tsx
+++ b/js/apps/admin-ui/src/clients/authorization/ResourcesPolicySelect.tsx
@@ -176,7 +176,6 @@ export const ResourcesPolicySelect = ({
         (value || []).map(async (id) =>
           adminClient.clients.findOnePolicy({
             id: clientId,
-            type: "",
             policyId: id,
           }),
         ),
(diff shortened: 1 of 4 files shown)


pedroigor (approved):
@ssilvert LGTM. Can you please review as well?

ssilvert (approved):
LGTM

PR #44073, opened 2025-11-10
Skip FGAP when evaluating permissions for regular clients

### LINKED ISSUE #40712: Authorization -> Evaluate: always returns "No search results"
### Before reporting an issue

### Area

admin/ui

### Describe the bug

I'm trying to set up an application that's only accessible if you're part of a certain realm group, and to test it with the `Evaluate` function in `Clients -> Authorization -> Evaluate`. But this issue happens even on clients where I have an allow-all permission.

In the applications the policies **are enforced**. It's just the admin ui always gives me the same result when running evaluate. I'm not sure what to put in `roles`. It seems to have been made mandatory in some update to Keycloak.

## Allow All

1. Create Policy

Image

Image

2. Create Permission

Image

Image

3. Evaluate

Image

4. Results are "No search results". This page used to tell me whether this user has access or not

Image

## Based on group

Image

Image

Image

Image

Image

Image

### Version

26.2.5

### Regression

### Expected behavior

I expect it to tell me whether this user has access to the client/application. In older versions of Keycloak this was the case. Also "roles" was not mandatory.

### Actual behavior

Always returns "No search results" 

### How to Reproduce?

See screenshots

### Anything else?

_No response_

pedroigor:
Closes #40712

### DIFF
--- a/services/src/main/java/org/keycloak/authorization/admin/PolicyEvaluationService.java
+++ b/services/src/main/java/org/keycloak/authorization/admin/PolicyEvaluationService.java
@@ -150,7 +150,7 @@ public class PolicyEvaluationService {
         EvaluationDecisionCollector decision = new EvaluationDecisionCollector(authorization, resourceServer, request);
 
         if (permissions.isEmpty()) {
-            if (AdminPermissionsSchema.SCHEMA.isAdminPermissionsEnabled(authorization.getRealm())) {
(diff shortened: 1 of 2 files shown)


vramik (commented):
Changes LGTM, thanks. Just wondering, should we also include a test?

pedroigor:
@vramik I'll write a test.

vramik (approved):

ahus1 (approved):

PR #44075, opened 2025-11-10
Efficient row-count on PostgreSQL for indexes

### LINKED ISSUE #42704: SELECT COUNT(*) FROM called multiple times for an index creation
### Before reporting an issue

### Area

core

### Describe the bug

While upgrading Keycloak we saw multiple `SELECT count(*) FROM` on the database logs for the same table and multiple `Following index should be created:` for the same index. 

In `CustomCreateIndexChange` there is a `SELECT count(*) FROM` to check for table size and skip a index creation if a table is too big: 

https://github.com/keycloak/keycloak/blob/aa8321b8373f9ce443eb2851fb37f41a1101e55e/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java#L82

But, this is called inside `SqlStatement[] generateStatements(Database database)` where the javadoc says that this method maybe called multiple times:

>  NOTE: This method may be called multiple times throughout the changelog execution process and may be called in documentation generation and other integration points as well. 
> If this method reads from the current database state or uses any other logic that will be affected by whether previous changeSets have ran or not, you must return true from [generateStatementsVolatile(liquibase.database.Database)](https://javadoc.io/static/org.liquibase/liquibase-core/4.33.0/liquibase/change/Change.html#generateStatementsVolatile(liquibase.database.Database)).

https://javadoc.io/doc/org.liquibase/liquibase-core/latest/liquibase/change/Change.html#generateStatements(liquibase.database.Database)

### Version

26.2.5

### Regression

### Expected behavior

I expect Keycloak to only perform a single `SELECT count(*) FROM` for each index. 

Ideally, if multiple indexes are created against the same table, it could re-use already computed counts. 

Even more ideally, it could use existing statistics provided by the database (like `n_live_tup` in `pg_stat_user_tables` for postgres), but this may be database specific. 

### Actual behavior

It performs the same `SELECT count(*) FROM` for each index 3 times at least. 

### How to Reproduce?

1. Enable logs for all queries in the database
2. Upgrade between version, where a index is to be created on a table with too many rows (default above 300000 rows)
3. Check the database logs to see multiple `SELECT count(*) FROM`  

### Anything else?

_No response_

### LINKED ISSUE #44057: Retrieving row count to evaluate index creation takes a long time on PostgreSQL and big tables
### Before reporting an issue

### Area

core

### Describe the bug

When trying to update a Keycloak instance from 26.3.5 to 26.4.4, `CustomCreateIndexChange` froze the Keycloak boot because of the following SQL query:
```sql
SELECT COUNT(*) FROM public.EVENT_ENTITY
```

The jstack pinpointed https://github.com/keycloak/keycloak/blob/39c4c1ed942a4bdcc0a3c4d68a9b853a082ea9a2/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java#L82-L82 .

In this instance, table `EVENT_ENTITY` is huge with many records. The count request is very slow.
I guess that, ironically, `CustomCreateIndexChange` performs this count to avoid triggering an index creation that would take too much time.

### Version

26.4.4

### Regression

### Expected behavior

Keycloak automatic database update should take a reasonable amount of time.

### Actual behavior

Keycloak automatic database update hangs forever.

### How to Reproduce?

Make sure to have  a huge `EVENT_ENTITY` table on version 26.3.5 then update Keycloak to 26.4.4

### Anything else?

_No response_

ahus1:
Closes #44057
Closes #42704

I tested this manually as follows when testing: 

* Test with PostgreSQL
* In the `migration_model` table, create an entry for 26.4.0 in addition to the 999.0.0 entry
* In the `databasechangelog`, remove `26.5.0-index-offline-css-by-client` and `26.5.0-index-offline-css-by-client-storage-provider`
* Drop the index `idx_offline_css_by_client` and `idx_offline_css_by_client_storage_provider` on `offline_client_session`
* Start Keycloak with `--spi-datastore-legacy-allow-migrate-existing-database-to-snapshot=true` and see that Keycloak performs the migration and adds the indexes

### DIFF
--- a/docs/documentation/upgrading/topics/migrate_db.adoc
+++ b/docs/documentation/upgrading/topics/migrate_db.adoc
@@ -28,6 +28,13 @@ To change the threshold, set the `index-creation-threshold` property, value for 
 kc.[sh|bat] start --spi-connections-liquibase--quarkus--index-creation-threshold=300000
 ----
 
+You can disable this feature by setting it to zero or a negative number:
+
+[source,bash]
+----
+kc.[sh|bat] start --spi-connections-liquibase--quarkus--index-creation-threshold=0
+----
+
 === Manual relational database migration
 
 To enable manual upgrading of the database schema, set the `migration-strategy` property value to "manual" for the
--- a/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
+++ b/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
@@ -31,10 +31,12 @@ import liquibase.change.DatabaseChange;
 import liquibase.change.core.CreateIndexChange;
 import liquibase.database.AbstractJdbcDatabase;
 import liquibase.database.Database;
+import liquibase.database.core.PostgresDatabase;
 import liquibase.exception.DatabaseException;
 import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.exception.ValidationErrors;
 import liquibase.exception.Warnings;
+import liquibase.executor.Executor;
 import liquibase.executor.ExecutorService;
 import liquibase.executor.LoggingExecutor;
 import liquibase.snapshot.InvalidExampleException;
@@ -42,7 +44,7 @@ import liquibase.snapshot.SnapshotGeneratorFactory;
 import liquibase.sqlgenerator.SqlGeneratorFactory;
 import liquibase.statement.SqlStatement;
 import liquibase.statement.core.CreateIndexStatement;
-import liquibase.statement.core.RawSqlStatement;
+import liquibase.statement.core.RawParameterizedSqlStatement;
 import liquibase.structure.core.Schema;
 import liquibase.structure.core.Table;
 import org.jboss.logging.Logger;
@@ -54,19 +56,21 @@ import org.jboss.logging.Logger;
     + 1, appliesTo = "index")
 public class CustomCreateIndexChange extends CreateIndexChange {
     private static final Logger logger = Logger.getLogger(CustomCreateIndexChange.class);
-    private int indexCreationThreshold;
+    private long indexCreationThreshold;
+    private Long entriesInTable = null;
+    private boolean logged;
 
     @Override
     public SqlStatement[] generateStatements(Database database) {
         // This check is for manual migration
-        if (Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor(LiquibaseConstants.JDBC_EXECUTOR, database) instanceof LoggingExecutor)
+        if (getExecutor(database) instanceof LoggingExecutor)
             return super.generateStatements(database);
 
         Object indexCreationThreshold = ((AbstractJdbcDatabase) database)
             .get(DefaultLiquibaseConnectionProvider.INDEX_CREATION_THRESHOLD_PARAM);
 
-        if (indexCreationThreshold instanceof Integer) {
-            this.indexCreationThreshold = (Integer) indexCreationThreshold;
+        if (indexCreationThreshold instanceof Long) {
+            this.indexCreationThreshold = (Long) indexCreationThreshold;
             if (this.indexCreationThreshold <= 0)
                 return super.generateStatements(database);
         } else {
@@ -75,15 +79,18 @@ public class CustomCreateIndexChange extends CreateIndexChange {
         try {
             // To check that the table already exists or not on which the index will be created.
             if (getTableName() == null || !SnapshotGeneratorFactory.getInstance()
-                .has(new Table().setName(getTableName()).setSchema(new Schema(getCatalogName(), getSchemaName())), database))
+                .has(new Table().setName(getTableName()).setSchema(new Schema(getCatalogName(), getSchemaName())), database)) {
                 return super.generateStatements(database);
+            }
 
-            int result = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor(LiquibaseConstants.JDBC_EXECUTOR, database)
-                    .queryForInt(new RawSqlStatement("SELECT COUNT(*) FROM " + getTableNameForSqlSelects(database, getTableName())));
+            Long entriesInTable = computeEntriesInTable(database);
 
-            if (result > this.indexCreationThreshold) {
+            if (entriesInTable > this.indexCreationThreshold) {
                 String loggingString = createLoggingString(database);
-                logger.warnv("Following index should be created: {0}", loggingString);
+                if (!logged) {
+                    logger.warnv("Following index should be created: {0}", loggingString);
+                    logged = true;
+                }
                 getChangeSet().setComments(loggingString);
                 return new SqlStatement[] {};
             }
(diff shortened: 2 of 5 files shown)


ahus1:
@keycloak/cloud-native - looking for a reviewer for this database related change. 

I would like to backport this change, we had some index changes for KC 26.4 / RHBK 26.4, and I those who haven't upgraded yet to benefit from it. 

The build was green once, the latest changes were only about the documentation.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
```diff
@@ -55,6 +56,8 @@
 public class CustomCreateIndexChange extends CreateIndexChange {
     private static final Logger logger = Logger.getLogger(CustomCreateIndexChange.class);
     private int indexCreationThreshold;
```
pruivo: It looks like this needs to be a long too.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
```diff
@@ -55,6 +56,8 @@
...
+    private Long entriesInTable = null;
+    private boolean logged;
```
pruivo, line 61: Are the `CustomCreateIndexChange` instances shared for this to be a field!?

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
```diff
@@ -95,6 +101,38 @@ public SqlStatement[] generateStatements(Database database) {
...
 
+    private Long computeEntriesInTable(Database database) throws DatabaseException {
+        if (entriesInTable != null) {
+            return entriesInTable;
+        }
+
+        if (database instanceof PostgresDatabase) {
+            try {
+                // This avoids locking all rows in the database table to get an exact count and instead takes an estimate
+                entriesInTable = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor(LiquibaseConstants.JDBC_EXECUTOR, database)
```
pruivo: It would be easier to read if `Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor(LiquibaseConstants.JDBC_EXECUTOR, database)` is in a method.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
```diff
@@ -95,6 +101,38 @@ public SqlStatement[] generateStatements(Database database) {
...
+                        .queryForLong(new RawParameterizedSqlStatement("SELECT reltuples::bigint AS estimate FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace WHERE n.nspname = current_schema AND UPPER(c.relname) = UPPER(?)", getTableName()));
+                if (entriesInTable == -1L) {
```
pruivo: Shouldn't it be `if entriesInTable >= 0; then return`? The method will continue and use the "count with limit" query.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
```diff
@@ -95,6 +101,38 @@ public SqlStatement[] generateStatements(Database database) {
...
+                    // no statistics exist for this table, try next approach
+                    entriesInTable = null;
+                }
+            } catch (UnexpectedLiquibaseException e) {
+                logger.warn("No permissions to run SELECT on the pg_class and pg_namespace tables, therefore can't estimate row count. Falling back to slower method to count entries with SELECT COUNT(*).", e);
+            }
+        }
+
+        if (database instanceof PostgresDatabase) {
```
pruivo: Don't need to do `instanceof` again, just move the code inside the if above.

pruivo (changes requested):

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
sguilhen: Shouldn't this test for `entriesInTable == null` instead?

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: Each entry in a changeset item gets its own instance of this class. So when there are two indexes to be created, there are two instances of this class. Due to that, this should is fine - including the caching of the row number.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: It seems that I was too lazy here, added in next commit.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: done in next commit.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: If no-one has created statistics for the table we do not have any information on how many entries there are in the table. I thought it would be better to use he count-with-limit to get a better estimate.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: sure, will do that in the next PR.

ahus1: thanks for spotting this. Changed in next commit.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: Implemented in next commit. It seems I was too lazy to do that.

docs/documentation/upgrading/topics/migrate_db.adoc
```diff
@@ -28,6 +28,13 @@ To change the threshold, set the `index-creation-threshold` property, value for
 kc.[sh|bat] start --spi-connections-liquibase--quarkus--index-creation-threshold=300000
 ----
 
+You can disable this feature by setting it to a negative number:
+
+[source,bash]
+----
+kc.[sh|bat] start --spi-connections-liquibase--quarkus--index-creation-threshold=-1
```
pedroigor: It is not clear why I should disable this setting (perhaps not calling it a feature). Would it make sense to add more information about disabling it in case the database is PostgreSQL, that it will ignore the number of rows, the risk of disabling it to other vendors, and have a link to the guide for more details?

Perhaps it makes sense to prevent disabling this setting if the database is other than PostgreSQL? IMO, I think it is a nice thing to allow admins to disable it for any vendor, as they should know what they are doing.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
pruivo: It is what I said :rofl:

docs/documentation/upgrading/topics/migrate_db.adoc
ahus1: @pedroigor - this sounds like a follow-up discussion to me about this feature in general. I will not consider it for this issue.

ahus1:
@pruivo - can you please re-review my latest changes? Thanks!

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java
ahus1: Hm, maybe it was too late. Adapted as you suggested here and in the other relevant places in this method.

pruivo (dismissed):
LGTM, but someone with more expertise should review it too.

pedroigor (approved):
Nice improvement!

PR #44078, opened 2025-11-10
Fix OpenShift v4 IDP mapper error handling and baseUrl bug

### LINKED ISSUE #43970: Identity Provider Mapping doesn't have good error handling
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When creating OpenShift v4 as an identity provider, and switching to the mappings, there is a the generic error message "unknown_error"  and the user is asked to reload the UI. 

Image

The user is lost and doesn't know what to do next. 

### Version

main

### Regression

### Expected behavior

There should be a descriptive error message and no need to reload the admin UI

### Actual behavior

I need to reload the admin UI, and there is no descriptive error message

### How to Reproduce?

* Create an Identity provider of type OpenShift v4 and save it
* Open the OpenShift v4 IDP
* Switch to the mappers tab -> error appears

### Anything else?

The baseUrl in the UI is optional, but in the backend it is not usedto create the oauthDescriptor. 

The code 

https://github.com/keycloak/keycloak/blob/dbd516f8e6f5152f292bedc1caffa601880f7eb9/services/src/main/java/org/keycloak/social/openshift/OpenshiftV4IdentityProvider.java#L38-L39

should probably read

```
        final String baseUrl = Optional.ofNullable(config.getBaseUrl()).orElse(BASE_URL);
        Map<String, Object> oauthDescriptor = getAuthJson(session, baseUrl);
```

Rathan-Naik:
When accessing the mappers tab for an OpenShift v4 Identity Provider with an invalid or unreachable Base URL, the Admin UI would display a generic error or require reload with no descriptive error message.

**This commit improves the error handling flow:**

1. OpenshiftV4IdentityProvider: Fixed constructor to pass the resolved baseUrl (with fallback to BASE_URL) instead of the potentially null config.getBaseUrl(). This ensures the correct URL is always used when fetching OAuth metadata.

2. IdentityProviderResource: Added try-catch around createIdentityProviderInstance() in getMapperTypes() to catch initialization failures and return a proper HTTP 400 error response with a descriptive error message to the frontend.

3. OpenshiftV4IdentityProviderTest: Updated testHttpClientErrors to use assertThrows for cleaner exception testing.

The error message now includes:
- The identity provider type that failed
- The original exception message from OpenshiftV4IdentityProvider
- Clear indication that mapper types could not be retrieved

Closes #43970

**Below is the screenshot of how latest change looks in the UI**
OPenshift

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/IdentityProviderResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/IdentityProviderResource.java
@@ -290,7 +290,13 @@ public class IdentityProviderResource {
             throw new jakarta.ws.rs.NotFoundException();
         }
 
-        IdentityProvider<?> identityProviderInstance = createIdentityProviderInstance();
+        IdentityProvider<?> identityProviderInstance;
+        try {
+            identityProviderInstance = createIdentityProviderInstance();
+        } catch (Exception e) {
+            throw ErrorResponse.error("Could not get mapper types for identity provider [" + identityProviderModel.getProviderId() + "]: " + e.getMessage(), Response.Status.BAD_REQUEST);
+        }
+
         KeycloakSessionFactory sessionFactory = session.getKeycloakSessionFactory();
         return sessionFactory.getProviderFactoriesStream(IdentityProviderMapper.class)
                 .map(IdentityProviderMapper.class::cast)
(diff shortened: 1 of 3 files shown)


vramik (commented):
Changes LGTM, but I'd like to ask @pedroigor for his review as well, as he is more experienced with this area.

@pedroigor PTAL

PR #44091, opened 2025-11-10
Sessions not removed when user is deleted

### LINKED ISSUE #43323: Sessions not removed when user is deleted
### Before reporting an issue

### Area

infinispan

### Describe the bug

When a user is deleted we are trying to remove the sessions attached to that user. But the session is not really removed. Nevertheless it's quite hard to check the session is still there because the session is auto-removed as soon as it prepared to wrap it.

I think that the reason is that when a user is removed, we are trying to remove the associate sessions:

1. The deletion of the user triggers UserRemovedEvent [here](https://github.com/keycloak/keycloak/blob/26.4.0/server-spi/src/main/java/org/keycloak/models/UserManager.java#L37).
2. The event is processed by the infinispan [here](https://github.com/keycloak/keycloak/blob/26.4.0/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java#L181).
3. That method tries to remove the sessions [here](https://github.com/keycloak/keycloak/blob/26.4.0/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java#L485-L493).
4. Nevertheless the sessions are never retrieved because the [getUserSessionsStream](https://github.com/keycloak/keycloak/blob/26.4.0/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java#L241-L261) method looks up the user to get the sessions and the user is now removed. So it goes to the second if, user is always null when removed, and empty stream is always returned.

So, in the end, no session is really removed. They are removed as soon as they are retrieved (see [this line in the wrap method](https://github.com/keycloak/keycloak/blob/26.4.0/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java#L532-L538)). But for the moment they are in the infinispan cache.

### Version

26.4.0

### Regression

### Expected behavior

Don't really know but I suppose we want them removed as we have calls to delete them.

### Actual behavior

The sessions are maintained and deleted when they are used or by expiration.

### How to Reproduce?

* Create a user in a realm.
* Perform some logins using different browsers to the account console using the same user.
* Check using statistics that the sessions are there.
```
curl -sk https://localhost:9000/metrics | grep "cache=\"sessions\""| grep vendor_statistics_approximate_entries
```
* Remove the user.
* Check using the same curl that the sessions are still there.
* Refresh the browser page to force a re-login.
* Check the sessions disappear now with the curl command.

### Anything else?

As commented it's hard to check the sessions are not removed, but following the code, you will see the sessions are never deleted, because [this line](https://github.com/keycloak/keycloak/blob/26.4.0/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/PersistentUserSessionProvider.java#L395) is never executed.

pruivo:
Fixes #43323

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/marshalling/KeycloakModelSchema.java
+++ b/model/infinispan/src/main/java/org/keycloak/marshalling/KeycloakModelSchema.java
@@ -97,9 +97,11 @@ import org.keycloak.models.sessions.infinispan.events.RealmRemovedSessionEvent;
 import org.keycloak.models.sessions.infinispan.events.RemoveAllUserLoginFailuresEvent;
 import org.keycloak.models.sessions.infinispan.events.RemoveUserSessionsEvent;
 import org.keycloak.models.sessions.infinispan.stream.AuthClientSessionSetMapper;
+import org.keycloak.models.sessions.infinispan.stream.ClientSessionFilterByUser;
 import org.keycloak.models.sessions.infinispan.stream.CollectionToStreamMapper;
 import org.keycloak.models.sessions.infinispan.stream.GroupAndCountCollectorSupplier;
 import org.keycloak.models.sessions.infinispan.stream.MapEntryToKeyMapper;
+import org.keycloak.models.sessions.infinispan.stream.RemoveKeyConsumer;
 import org.keycloak.models.sessions.infinispan.stream.SessionPredicate;
 import org.keycloak.models.sessions.infinispan.stream.SessionUnwrapMapper;
 import org.keycloak.models.sessions.infinispan.stream.SessionWrapperPredicate;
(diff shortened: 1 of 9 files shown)


ahus1 (approved):
Thank you for this fix! I don't think there is a need to backport this to 26.4 - please let me know if you think otherwise. 

If we would choose to backport it, it would be good to first backport the new classes `ClientSessionFilterByUser` and `RemoveKeyConsumer` in one patch release, and the other changes in another patch release. 

I assume that this would only be triggered when a user is deleted which is rare, this would be the extra mile I would like to go.

PR #44095, opened 2025-11-10
fix: performing scale down prior to deletion

### LINKED ISSUE #34868: [Jenkins Operator CI] - Test remote - ClusteringTest on OpenShift
### Before reporting an issue

### Area

ci

### Describe the bug

The following tests are failing constantly when running on OpenShift:

```
2024-11-10 21:55:13  [ERROR] Errors: 
2024-11-10 21:55:13  [ERROR]   ClusteringTest.testKeycloakCacheIsConnected:243 » ConditionTimeout Assertion condition defined as a Lambda expression in org.keycloak.operator.testsuite.integration.ClusteringTest 
2024-11-10 21:55:13  Expecting any elements of:
2024-11-10 21:55:13    [KeycloakRealmImportStatusCondition{type='Done', status=False, message=''},
2024-11-10 21:55:13      KeycloakRealmImportStatusCondition{type='Started', status=False, message=''},
2024-11-10 21:55:13      KeycloakRealmImportStatusCondition{type='HasErrors', status=True, message='Deployment not yet ready, waiting for it to be ready'}]
2024-11-10 21:55:13  to match given predicate but none did within 10 minutes.
2024-11-10 21:55:13  [ERROR]   ClusteringTest.testKeycloakScaleAsExpected:146 » ConditionTimeout Assertion condition defined as a Lambda expression in org.keycloak.operator.testsuite.utils.K8sUtils found status KeycloakStatusCondition{type='Ready', status=False, message='Waiting for more replicas'} and expected status true within 5 minutes.
2024-11-10 21:55:13  [INFO] 
2024-11-10 21:55:13  [ERROR] Tests run: 125, Failures: 0, Errors: 2, Skipped: 5
```
https://keycloak-jenkins/view/OpenShift/job/rhbk-operator-test-pipeline/893/execution/node/69/log/

### Version

latest

### Regression

### Expected behavior

-

### Actual behavior

-

### How to Reproduce?

-

### Anything else?

_No response_

shawkins:
closes: #34868

Try to ensure that sigterm will be received by all keycloak pods during cleanup.

### DIFF
--- a/operator/src/test/java/org/keycloak/operator/testsuite/integration/BaseOperatorTest.java
+++ b/operator/src/test/java/org/keycloak/operator/testsuite/integration/BaseOperatorTest.java
@@ -62,6 +62,7 @@ import org.keycloak.operator.controllers.KeycloakDeploymentDependentResource;
 import org.keycloak.operator.controllers.KeycloakRealmImportController;
 import org.keycloak.operator.controllers.KeycloakUpdateJobDependentResource;
 import org.keycloak.operator.crds.v2alpha1.deployment.Keycloak;
+import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakBuilder;
 import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakSpecBuilder;
 import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakStatus;
 import org.keycloak.operator.crds.v2alpha1.realmimport.KeycloakRealmImport;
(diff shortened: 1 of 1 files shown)


vmuzikar (approved):
From the test perspective, this looks good.

My only concern would be if this means we have a problem with garbage collection of statefulsets. But there's probably not much we can do from the Keycloak side anyway.

shawkins:
> My only concern would be if this means we have a problem with garbage collection of statefulsets. But there's probably not much we can do from the Keycloak side anyway.

It's not clear to me how often something like this happens when deleting keycloaks - it doesn't appear to be very common and it's mostly an issue with the pg locking statement. That same proplem can occur for reasons outside of our control as well.

About all we could do to make this even rarer would be to add a finalizer to the Keycloak CR so that we can perform the scale down first.

Alternatively we could document that users should scale down their Keycloak's before deleting them.

vmuzikar:
> About all we could do to make this even rarer would be to add a finalizer to the Keycloak CR so that we can perform the scale down first.
> 
> Alternatively we could document that users should scale down their Keycloak's before deleting them.

Yes but both feel to me like workarounds to some K8s issue. Not sure... Let's keep as is and we'll see if the issue affects real production deployments as well.

PR #44106, opened 2025-11-11
Sd-Jwt unit tests in the crypto/fips1402 module

### LINKED ISSUE #44104: Sd-Jwt unit tests in the crypto/fips1402 module
### Description

There are some unit tests of sd-jwt, which are tested only in the `crypto/default` and `crypto/elytron` module. It can be good to make sure that those are tested in the `crypto/fips1402` module as well to make sure that sd-jwt SDK works correctly in the FIPS mode.

It should probably work fine as algorithms used do not have any restrictions in the FIPS mode when bc-fips library is used instead of normal non-fips bc library version.

mposolda:
closes #44104

### DIFF
--- a/crypto/fips1402/src/test/java/org/keycloak/crypto/fips/test/sdjwt/FIPS1402JwtVcMetadataTrustedSdJwtIssuerTest.java
+++ b/crypto/fips1402/src/test/java/org/keycloak/crypto/fips/test/sdjwt/FIPS1402JwtVcMetadataTrustedSdJwtIssuerTest.java
@@ -1,0 +1,15 @@
+package org.keycloak.crypto.fips.test.sdjwt;
+
+import org.junit.Assume;
+import org.junit.Before;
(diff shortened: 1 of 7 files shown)


IngridPuppet (commented):
I confirm that we have the expected number of tests and that they extend the intended core tests. However, copyright headers are missing across the new files.

mposolda:
> I confirm that we have the expected number of tests and that they extend the intended core tests. However, copyright headers are missing across the new files.

@IngridPuppet 
Thanks! The copyright headers are actually not needed in the new files (Similarly like `@author` tag is not needed) and it is actually preferred to not use them anymore in Keycloak codebase files. I think it is not officially announced yet anywhere, but it may be probably soon.

IngridPuppet (approved):
Thank you for the clarification on copyright headers and authorship.

mposolda:
@IngridPuppet Thanks!

Issue #44125, opened 2025-11-11
Double-encoding of query parameter values (e.g. acr_values) for version 26.4


jhgojbis:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

identity-brokering

### Describe the bug

**Problem**

uriBuilder.queryParam() already performs URL encoding.

Manually applying URLEncoder.encode(parameter, StandardCharsets.UTF_8) before calling it results in double encoding, where:

space → "+" → "%2B"

Affected method:

uriBuilder.queryParam(forwardParameter, URLEncoder.encode(parameter, StandardCharsets.UTF_8));

Reference (line of code):
https://github.com/keycloak/keycloak/blob/39964befefc41eec3330fcf5c94f5e1ea498d075/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java#L562

Related commit that introduced the issue:
https://github.com/keycloak/keycloak/issues/41755

Take note : It was working in version 26.0

### Version

26.4

### Regression

- [x] The issue is a regression

### Expected behavior

The URI component should only be encoded once.

### Actual behavior

The URI component are encoded twice.

### How to Reproduce?

Create an Identity Provider and add a forwarded query parameter, use a query parameter such as acr_values. You will notice that something like hello world turns into hello%2Bworld when it should be either +or %20.
 

### Anything else?

Reference (line of code):
https://github.com/keycloak/keycloak/blob/39964befefc41eec3330fcf5c94f5e1ea498d075/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java#L562

Related commit that introduced the issue:
https://github.com/keycloak/keycloak/issues/41755

pedroigor:
It does already, but it does not work well for encoding JSON strings.

When I sent that fix, I missed the double-encoding for the string with spaces.

I'll need to find another way to encode JSON strings (e.g.: `claims` parameter) without using the URI builder from Resteasy. It understand `{` as the start of a placeholder for path parameters.

PR #44145, opened 2025-11-12
Bug fix double-encoding for query parameter acr_values

jhgojbis:
Related bug fix in Keycloak version 26.4

space with mutiple values results in → "+" → "%2B"

Reported bug: 
https://github.com/keycloak/keycloak/issues/44125

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
@@ -531,8 +531,6 @@ public abstract class AbstractOAuth2IdentityProvider<C extends OAuth2IdentityPro
             uriBuilder.queryParam(OAuth2Constants.PROMPT, prompt);
         }
 
-        setForwardParameters(authenticationSession, uriBuilder);
-
         if (getConfig().isPkceEnabled()) {
             String codeVerifier = PkceUtils.generateCodeVerifier();
             String codeChallengeMethod = getConfig().getPkceMethod();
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
```diff
@@ -559,7 +559,7 @@ private void setForwardParameters(AuthenticationSessionModel authenticationSessi
             }
 
             if (parameter != null && !parameter.isEmpty()) {
-                uriBuilder.queryParam(forwardParameter, URLEncoder.encode(parameter, StandardCharsets.UTF_8));
+                uriBuilder.queryParam(forwardParameter, parameter);
```
jhgojbis: uriBuilder.queryParam already decode "parameter" we dont need to use URLEncoder.encode

pedroigor: It does encode parameters, but it will not encode parameters with a JSON string as a value. For instance, when forwarding the `claims` parameter.

Supporting the `claims` parameter was the original intent and the fix, as you pointed out, that broke the encoding of forwarded parameters.

I've added a commit to your branch with a different approach so that forwarded parameters are encoded differently, using the `URLEncoder`. It seems to work well for all cases, including values referencing a JSON string.

Please, let me know if it works for you.

jhgojbis: Okej thanks!

What is the plan then?

pedroigor (approved):

services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
pedroigor: Sent https://github.com/keycloak/keycloak/pull/44244. Thanks for the help, and sorry for the confusion.

PR #44146, opened 2025-11-12
Use normal scope parameter checking for the JWT Authorization grant

### LINKED ISSUE #43646: Processing scope
### Description

With the authorization grant, it is possible to use the `scope` parameter to indicate the requested scope.
This task is to check how the scope parameter is handled, improve if needed the processing of the `scope` parameter, as well as adding tests.

rmartinc:
Closes #43646

Just using the inherited method `getRequestedScopes` to check the `scope` parameter is valid for the client, as in any other grant (removing the TODO). I have done an extra change to return the error `invalid_grant` instead of `invalid_request`, to follow spec [RFC 7523 - Authorization Grant Processing](https://datatracker.ietf.org/doc/html/rfc7523#section-3.1).


graziang (approved):
@rmartinc Thanks!

mposolda (approved):

Issue #44161, opened 2025-11-12
Revisit handling of unknown fields in Client v2


vmuzikar:
### Description

We could optionally allow the server to ignore unknown field in the incoming requests. This has been removed in https://github.com/keycloak/keycloak/issues/43728. The aim of this spike is to consider, if this is something we need now.

The main use case is to not break existing clients when we remove a field from a representation without introducing a new API version.

shawkins:
See the comments on #43728 - if we want to allow schema evolution that removes fields, we'll need to 
1. add this parameter back before we consider the feature supported
2. or change (or make configurable) the server handling to IGNORE/WARN, and add the parameter back.

PR #44173, opened 2025-11-12
fix: removing unknown field validation parameter

### LINKED ISSUE #43728: Make unknown fields handling a server side option
### Description

* Remove the `FieldValidation` option.
* Make it a server config option.

See https://github.com/keycloak/keycloak/pull/43395#discussion_r2447106122

shawkins:
closes: #43728

### DIFF
--- a/rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
+++ b/rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
@@ -3,7 +3,6 @@ package org.keycloak.admin.api.client;
 import java.io.IOException;
 import java.util.Objects;
 
-import org.keycloak.admin.api.FieldValidation;
 import org.keycloak.http.HttpResponse;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.KeycloakSession;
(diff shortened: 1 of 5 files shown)


rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
```diff
@@ -105,12 +108,12 @@ public ClientRepresentation patchClient(JsonNode patch, FieldValidation fieldVal
         }
     }
 
-    static void validateUnknownFields(FieldValidation fieldValidation, ClientRepresentation rep, HttpResponse response) {
+    static void validateUnknownFields(ClientRepresentation rep, HttpResponse response) {
         if (!rep.getAdditionalFields().isEmpty()) {
-            if (fieldValidation == null || fieldValidation == FieldValidation.Strict) {
+            if (DEFAULT_FIELD_VALIDATION == null || DEFAULT_FIELD_VALIDATION == FieldValidation.Strict) {
```
vmuzikar: I'd rather prefer we removed also the `FieldValidation` enum. We do not know if we'll even make it configurable (that depends on the result of #44161). We do not need it for now (just like the SPIs we removed in the previous PR).

shawkins: Ok, we'll add it back later.

vmuzikar (approved):
@shawkins Thank you, LGTM.

PR #44182, opened 2025-11-12
Remove remote call when handling events

pruivo:
Closes #44048

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/cluster/infinispan/remote/RemoteInfinispanNotificationManager.java
+++ b/model/infinispan/src/main/java/org/keycloak/cluster/infinispan/remote/RemoteInfinispanNotificationManager.java
@@ -22,7 +22,6 @@ import java.lang.invoke.MethodHandles;
 import java.nio.ByteBuffer;
 import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


pruivo:

ahus1:
Rebased due to conflict which IntelliJ then automatically resolved. 

@ryanemerson - please review when you have the time. Thanks!

model/infinispan/src/main/java/org/keycloak/cluster/infinispan/remote/RemoteInfinispanNotificationManager.java
```diff
@@ -119,28 +124,47 @@ public String getMyNodeName() {
...
+    private String extractKeyFromEvent(ClientCacheEntryCustomEvent<byte[]> event) throws IOException, ClassNotFoundException {
+        byte[] data = event.getEventData();
+        ByteBuffer buffer = ByteBuffer.wrap(data);
+        int length = UnsignedNumeric.readUnsignedInt(buffer);
 
-    @ClientCacheEntryRemoved
-    public void removed(ClientCacheEntryRemovedEvent<String> event) {
-        String key = event.getKey();
-        taskFinished(key);
+        // read the value
```
ryanemerson: This comment is confusing in the context of the method. Maybe "unmarshall the key"?

ryanemerson: We could just inline this method given that it's only ever called from `onEntryRemoved`.

ryanemerson (dismissed):

ahus1 (approved):

PR #44185, opened 2025-11-13
Step up authentication for saml

### LINKED ISSUE #10155: Step-up authentication for SAML clients

rmartinc:
Closes #10155

This is the initial idea for SAML step-up authentication. It's incomplete because I have not done the mapping at realm level. This PR just uses the mapping at client level. In saml we use the authentication context class ref (which is an uri, for example `urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport`) and I suppose it will be different to the OIDC name. So in the client is the same map `{"uri":"loa",...}` but we have to think something for the realm.

What do you prefer?

1. Do more or less like it's in the PR and always override at client level. The realm will use the same OIDC names.
2. Add another mapping at realm level (something like `saml.acr.loa.map` or similar), but it's very convoluted for the admins.
3. Change the current map to an array, for example `[{"acr":"silver","uri":"urn:oasis:names:...","loa":"1"},...]`. But this needs to convert the current realm values on the fly or at upgrade.

Suggestions?

@mposolda The spec says that requesting an incorrect authn context returns an error in saml protocol (second status). So I need to implement a `sendProtocolError` in the bindings that uses the protocol. But I discovered the `inResponseTo` was not properly set for errors. I added this ID in the client data. Is this OK? Or better we just use the `inResponseTo` in the methods that the auth session is available?

Draft for the moment because, we need to manage mapping at realm level before.


rmartinc:
OK, I have implemented the realm map. Finally it is the following:

* The realm has another attribute `acr.uri.map` than contains the map between OIDC acr and SAML uri acr.
* But this is presented in the realm as a table ACR, URI, LoA. There is a new element `RealmLoAMapping` in the UI only for this. The `uri` in that table is optional, so an ACR can have only ACR and LoA and then it is not valid for SAML.

The rest is more or less the same but I have extended testing and polished some corner cases. For me it's OK and it's fully functional (see the tests) and cover everything. The only missing piece is the documentation. Maybe I would manage it as a follow-up.

rmartinc:
Moving to ready for review. The JS CI will be fixed by https://github.com/keycloak/keycloak/pull/44883.

rmartinc:
Rebased again. :smile:

docs/documentation/server_admin/topics/authentication/flows.adoc
```diff
@@ -426,6 +426,85 @@ Note when the login request initiates a request with the `claims` parameter requ
 one of the specified levels. If it is not able to return one of the specified levels (For example if the requested level is unknown or bigger than configured conditions
 in the authentication flow), then {project_name} will throw an error.
 
+[[_step-up-authentication-saml]]
+==== Step-up authentication for SAML
+
+:tech_feature_name: Step-up Authentication for SAML
+:tech_feature_id: step-up-authentication-saml
+include::../../topics/templates/techpreview.adoc[]
```
mposolda, line 434: This adds the "preview" warning in the documentation, however the feature is enabled as "experimental" in this PR. It seems the options are:

1) Mark the feature as "preview" instead of "experimental" . Which IMO is OK as it is small feature and isolated

2) Update this documentation to mark as feature as "Experimental" . It should be either some "hand-written warning" or a template for experimental features (Something like `include::../../topics/templates/techexperimental.adoc` ).

My vote is (1) ATM .

mposolda (commented):
@rmartinc Thanks! Added only one note inline regarding docs. Also there is conflict introduced in the meantime, but hopefully that can be addressed later.

rmartinc:
@mposolda Rebased. Promoted to preview. I have added the release note and also prepared the migration to include the client scope when supported.

docs/documentation/server_admin/topics/authentication/flows.adoc
rmartinc: Moved to preview.

mposolda (approved):
@rmartinc Thanks!

PR #44202, opened 2025-11-13
Deprecate TopologyInfo

### LINKED ISSUE #44047: Deprecate TopologyInfo
### Description

The class `org.keycloak.connections.infinispan.TopologyInfo`  has most of its methods deprecated. The only information in use stored in this class is the node and site names. Those can be stored in a `record`.

pruivo:
Closes #44047


pruivo:

ahus1 (approved):

Issue #44152, opened 2025-11-12
Use the unified constants class for sd-jwt/oid4vc standard data and claims


mposolda:
### Description

In Keycloak, there is a pattern of "Constants" class (EG. `OAuth2Constants`). This makes it easy to have all the related constants in a single place without spread them in various classes as then it makes it harder to figure where is the particular constant defined (if somewhere). This often results in the constant not being used, but rather hardcoded, which is not ideal (for example when we want some refactoring in the future).

It can be good to have dedicated constant class for OID4VC specific stuff.

tdiesler:
Do we really need two ...

* core/src/main/java/org/keycloak/OID4VCConstants.java
* server-spi-private/src/main/java/org/keycloak/constants/Oid4VciConstants.java

PR #44207, opened 2025-11-13
Polishing of key-binding in sd-jwt SDK

### LINKED ISSUE #44206: Polishing of key-binding in sd-jwt SDK
### Description

I found some possible issues when looking at sd-jwt SDK inside `core` module in relation to the key-binding:

- The `KeyBindingJWT` could be more type-safe as it is clearly defined in the specification https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-key-binding-jwt what claims are REQUIRED and must be present. We can possibly use the concrete Java class for more type-safety instead of generic `JsonNode` ?

- The method `withKeyBinding` on `SdJwt.builder()` is effectively ignored and does not add any key binding to the `SdJwt`. It seems that it is not needed as key-binding is supposed to be used in the presentations rather than in "Issuer signed JWT" ? However in that case, it will be good to remove that method to avoid any confusion and make API more clear and secure.

- Per the https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-key-binding-jwt , the `typ` is hardcoded to `kb+jwt` . Considering that, we may not need support for set of any other value of `typ`? Will make it possible to cleanup `KeyBindingJWT` constructor and some related code to remove the `typ` from there?

- The `alg` in the header is required per the specification and MUST not be "none" . This is currently not verified

mposolda:
closes #44206

Marking as draft as it is blocked by the other PR for #44152 (commits from that PR included in this PR as well for now). Just the last commit is the actual changes related to the issue #44206 .


core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingPayload.java
```diff
@@ -0,0 +1,107 @@
...
+    private String nonce;
+
+    @JsonProperty(SD_HASH)
+    private String sdHash;
+
+    @JsonProperty(CLAIM_NAME_EXP)
+    private Long exp;
+
+    @JsonProperty(CLAIM_NAME_NBF)
+    private Long nbf;
```
Captain-P-Goldfish, line 37: I initially wanted to take a similar approach, but I had extended `JsonWebToken`. After discussing it with the developers at adorsys, they preferred to keep it directly on Jackson nodes. I have no objections to going in either direction, so I followed adorsys preference.

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
```diff
@@ -209,6 +222,9 @@ private void validateKeyBindingJwt(
         // Check that the typ of the Key Binding JWT is kb+jwt
         validateKeyBindingJwtTyp();
 
+        // Check that the algorithm is not "none"
+        validateKeyBindingAlgorithm();
```
Captain-P-Goldfish, line 226: if you prefer, I could add this check already into my pull-request #44227.

Captain-P-Goldfish: okay, I have added a new commit that extends #44227 also by a check for the alg-header.

tnorimat:
@mposolda Hello, the blocker PR was merged (https://github.com/keycloak/keycloak/issues/44152). So, I think it might be OK to change the PR from Draft PR to the normal PR.

core/src/main/java/org/keycloak/sdjwt/SdJwtUtils.java
```diff
@@ -113,10 +113,19 @@ public static ArrayNode decodeDisclosureString(String disclosure) throws Verific
         return (ArrayNode) jsonNode;
     }
 
-    public static long readTimeClaim(JsonNode payload, String claimName) throws VerificationException {
+    /**
+     * @param payload
+     * @param claimName
+     * @return Specified time-claim if exists. Could return null if claim does not exists
```
tnorimat, line 119: typos: `exist`

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
```diff
@@ -664,18 +687,18 @@ private void preventKeyBindingJwtReplay(
     private void validateKeyBindingJwtSdHashIntegrity() throws VerificationException {
         Objects.requireNonNull(sdJwtVpString);
 
-        JsonNode sdHash = keyBindingJwt.getPayload().get("sd_hash");
-        if (sdHash == null || !sdHash.isTextual()) {
+        String sdHash = keyBindingJwt.getPayload().getSdHash();
+        if (sdHash == null) {
             throw new VerificationException("Key binding JWT: Claim `sd_hash` missing or not a string");
```
tnorimat, line 692: "or not a string" is no longer needed.

core/src/test/java/org/keycloak/sdjwt/SdJwsTest.java
```diff
@@ -120,7 +118,45 @@ public void testVerifyVctClaim_Negative() {
...
+                try {
+                    return SdJwtUtils.mapper.readTree(jwsInput.getContent());
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        };
+    }
+
+    private SdJws<JsonNode> createSdJws(JsonNode payload, SignatureSignerContext signer, String jwsType) {
```
tnorimat, line 144: This method is `private` and all three callers in this class set the same value `jwt`. It could get rid of `String jwsType`, and hard-code `jwt` inside the method.

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
```diff
@@ -181,9 +142,9 @@ public void testPermissiveVerifierMissingClaims() throws VerificationException {
         ObjectNode payload = SdJwtUtils.mapper.createObjectNode();
```
tnorimat, line 142: this `payload` is no longer used.

tnorimat (commented):
@mposolda Hello, I added some minor comment. I appreciate it if you would check them.

mposolda:
@tnorimat Thanks for your review points! There is other PR, which is very likely to introduce additional conflicts: https://github.com/keycloak/keycloak/pull/44029 . I will possibly look at re-doing this one once the other PR is addressed.

PR #44209, opened 2025-11-14
Add client support to workflows

### LINKED ISSUE #43870: Workflows should support detecting and disabling dormant clients
### Description

It should be possible to configure workflows to detect and disable or delete dormant clients

Hathoute:
Closes #43870

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
@@ -57,6 +57,7 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
 
         return switch (type) {
             case USERS -> new UserResourceTypeWorkflowProvider(session);
+            case CLIENTS -> new ClientResourceTypeWorkflowProvider(session);
         };
     }
 
@@ -64,6 +65,7 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
     public void updateWorkflow(Workflow workflow, WorkflowRepresentation representation) {
         // first step - ensure the updated workflow is valid
         WorkflowValidator.validateWorkflow(session, this, representation);
+        WorkflowValidator.validateWorkflowConditionType(session, representation.getConditions(), workflow.getSupportedType());
 
         // check if there are scheduled steps for this workflow - if there aren't, we can update freely
         if (!stateProvider.hasScheduledSteps(workflow.getId())) {
@@ -76,6 +78,11 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
             WorkflowRepresentation currentRepresentation = toRepresentation(workflow);
             if (!Objects.equals(currentRepresentation.getOn(), representation.getOn())) {
                 throw new ModelValidationException("Cannot update 'on' configuration when there are scheduled resources for the workflow.");
+            }
+
+            // we do not allow changing the workflow type
+            if (representation.getSupports() != null && !Objects.equals(representation.getSupports(), currentRepresentation.getSupports())) {
+                throw new ModelValidationException("Cannot update 'supports' configuration.");
             }
 
             // we also need to guarantee the steps remain the same - that is, in the same order with the same 'uses' property.
@@ -95,6 +102,9 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
                 newStep.setId(currentStep.getId());
                 newStep.setPriority(Long.parseLong(currentStep.getPriority()));
             }
+
+            // set the workflow's type
+            representation.setSupports(currentRepresentation.getSupports());
 
             // finally, update the workflow's config along with the steps' configs
             workflow.updateConfig(representation.getConfig(), newSteps);
@@ -171,7 +181,7 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
             stateProvider.getDueScheduledSteps(workflow).forEach((scheduled) -> {
                 // check if the resource is still passes the workflow's resource conditions
                 DefaultWorkflowExecutionContext context = new DefaultWorkflowExecutionContext(session, workflow, scheduled);
-                EventBasedWorkflow provider = new EventBasedWorkflow(session, getWorkflowComponent(workflow.getId()));
+                EventBasedWorkflow provider = new EventBasedWorkflow(session, workflow.getSupportedType(), getWorkflowComponent(workflow.getId()));
                 if (!provider.validateResourceConditions(context)) {
                     log.debugf("Resource %s is no longer eligible for workflow %s. Cancelling execution of the workflow.",
                             scheduled.resourceId(), scheduled.workflowId());
@@ -192,6 +202,10 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
 
     @Override
     public void activate(Workflow workflow, ResourceType type, String resourceId) {
+        if (type != workflow.getSupportedType()) {
+            throw new BadRequestException("Resource Type '%s' is not supported for this workflow (supports %s)".formatted(type.name(), workflow.getSupportedType()));
+        }
+
         processEvent(Stream.of(workflow), new AdhocWorkflowEvent(type, resourceId));
     }
 
@@ -204,10 +218,11 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
     public void activateForAllEligibleResources(Workflow workflow) {
         if (workflow.isEnabled()) {
             WorkflowProvider provider = getWorkflowProvider(workflow);
-            ResourceTypeSelector selector = provider.getResourceTypeSelector(ResourceType.USERS);
+            ResourceType supportedType = workflow.getSupportedType();
+            ResourceTypeSelector selector = provider.getResourceTypeSelector(supportedType);
             selector.getResourceIds(workflow)
-                    .forEach(resourceId -> processEvent(Stream.of(workflow), new AdhocWorkflowEvent(ResourceType.USERS, resourceId)));
-        }
+                    .forEach(resourceId -> processEvent(Stream.of(workflow), new AdhocWorkflowEvent(supportedType, resourceId)));
+            }
     }
 
     @Override
@@ -230,6 +245,10 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
 
         Workflow workflow = addWorkflow(new Workflow(session, rep.getId(), config));
         workflow.addSteps(rep.getSteps());
+
+        // After adding steps, validate that the condition type is compatible with the computed workflow type.
+        WorkflowValidator.validateWorkflowConditionType(session, workflow.getCondition(), workflow.getSupportedType());
+
         return workflow;
     }
 
@@ -264,7 +283,7 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
                 return;
             }
 
-            EventBasedWorkflow provider = new EventBasedWorkflow(session, getWorkflowComponent(workflow.getId()));
+            EventBasedWorkflow provider = new EventBasedWorkflow(session, workflow.getSupportedType(), getWorkflowComponent(workflow.getId()));
 
             try {
                 if (!provider.supports(event.getResourceType())) {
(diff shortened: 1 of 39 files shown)


model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -298,6 +319,7 @@ private void validateWorkflow(WorkflowRepresentation rep) {
         List<WorkflowStepRepresentation> steps = ofNullable(rep.getSteps()).orElse(List.of());
 
         if (steps.isEmpty()) {
+            // Do we really want to allow workflows with no steps?
```
Hathoute: Does it make sense to support workflows with no steps? Why not throw a validation error?

server-spi-private/src/main/java/org/keycloak/models/workflow/Workflow.java
```diff
@@ -113,6 +116,17 @@ public void addSteps(List<WorkflowStepRepresentation> steps) {
...
+
+            // update allowed types
+            WorkflowStepProviderFactory<WorkflowStepProvider> stepProvider = getStepProviderFactory(step);
+            allowedTypes.retainAll(stepProvider.getTypes());
+        }
+
+        // - Should we allow workflows to have more than one supported type ?
+        // - allowedTypes should probably be saved in config, since we do not allow changing steps,
+        // so that we don't recompute it.
+        if (allowedTypes.isEmpty()) {
```
Hathoute, line 187: Could we discuss the comments above? What do you think?

Hathoute:
Hey @sguilhen , I've rebased the branch on main. Would you please have a look at the comments I've made above?

sguilhen:
Hey @Hathoute, Pedro and I will try to take a look this week - we're wrapping up a sprint, but we plan on checking this right after that. Thanks a lot for the work you've put into it, will get back to you asap.

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
sguilhen: We should probably reject this, doesn't make any sense IMO.

server-spi-private/src/main/java/org/keycloak/models/workflow/Workflow.java
sguilhen: I think we need to persist the resource type of the workflow, and enforce that all conditions and steps operate on the same type.

services/src/main/java/org/keycloak/models/workflow/client/DisableClientStepProvider.java
```diff
@@ -0,0 +1,53 @@
...
+        // nothing to close
+    }
+
+    @Override
+    public void run(WorkflowExecutionContext context) {
+        RealmModel realm = session.getContext().getRealm();
+        ClientModel client = session.clients().getClientById(realm, context.getResourceId());
+
+        if (client != null && client.isEnabled()) {
+            log.debugv("Disabling user {0} ({1})", client.getName(), client.getId());
```
sguilhen: Should be "Disabling client" instead of "Disabling user".

model/jpa/src/main/java/org/keycloak/models/workflow/JpaWorkflowStateProvider.java
```diff
@@ -76,11 +75,11 @@ public void scheduleStep(Workflow workflow, WorkflowStep step, String resourceId
             entity.setWorkflowId(workflow.getId());
             entity.setExecutionId(executionId);
             entity.setScheduledStepId(step.getId());
-            entity.setScheduledStepTimestamp(Instant.now().plus(duration).toEpochMilli());
+            entity.setScheduledStepTimestamp(Time.currentTimeMillis() + duration.toMillis());
             em.persist(entity);
         } else {
             entity.setScheduledStepId(step.getId());
-            entity.setScheduledStepTimestamp(Instant.now().plus(duration).toEpochMilli());
+            entity.setScheduledStepTimestamp(Time.currentTimeMillis() + duration.toMillis());
```
sguilhen: Just curious - any particular reason for this change?

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
```diff
@@ -555,8 +556,9 @@ private void startServer(List<String> arguments) throws Exception {
         ProcessBuilder pb = new ProcessBuilder(getCliArgs(arguments));
         ProcessBuilder builder = pb.directory(distPath.resolve("bin").toFile());
 
-        if (debug) {
+        if (true) {
             builder.environment().put("DEBUG_SUSPEND", "y");
+            builder.environment().put("DEBUG_PORT", "5005");
```
sguilhen: The changes here need probably to be reverted.

server-spi-private/src/main/java/org/keycloak/models/workflow/ResourceType.java
```diff
@@ -37,6 +31,12 @@ public enum ResourceType {
             List.of(OperationType.CREATE),
             List.of(EventType.LOGIN, EventType.REGISTER),
             (session, id) -> session.users().getUserById(session.getContext().getRealm(), id)
+    ),
+    CLIENTS(
+            org.keycloak.events.admin.ResourceType.CLIENT,
+            List.of(OperationType.CREATE),
+            List.of(EventType.CLIENT_LOGIN, EventType.CLIENT_REGISTER),
+            (session, id) -> session.users().getUserById(session.getContext().getRealm(), id)
```
sguilhen: This function is used in the `ResourceTypeSelector.resolveResource` impl that you moved to the AbstractResourceTypeWorkflowProvider, and it needs to resolve to the actual resource - in this case it should be client, not user.

server-spi-private/src/main/java/org/keycloak/models/workflow/ResourceOperationType.java
```diff
@@ -20,47 +21,63 @@
...
+    USER_LOGGED_IN(ResourceType.USERS, List.of(EventType.LOGIN), userLoginPredicate()),
+    USER_FEDERATED_IDENTITY_ADDED(ResourceType.USERS, List.of(FederatedIdentityCreatedEvent.class), fedIdentityPredicate()),
+    USER_FEDERATED_IDENTITY_REMOVED(ResourceType.USERS, List.of(FederatedIdentityRemovedEvent.class), fedIdentityPredicate()),
+    USER_GROUP_MEMBERSHIP_ADDED(ResourceType.USERS, List.of(GroupMemberJoinEvent.class), groupMembershipPredicate()),
+    USER_GROUP_MEMBERSHIP_REMOVED(ResourceType.USERS, List.of(GroupModel.GroupMemberLeaveEvent.class), groupMembershipPredicate()),
+    USER_ROLE_ADDED(ResourceType.USERS, List.of(RoleGrantedEvent.class), roleMembershipPredicate()),
+    USER_ROLE_REMOVED(ResourceType.USERS, List.of(RoleModel.RoleRevokedEvent.class), roleMembershipPredicate()),
+
+    CLIENT_ADDED(ResourceType.CLIENTS, List.of(OperationType.CREATE, ClientCreationEvent.class)),
+    CLIENT_LOGGED_IN(ResourceType.CLIENTS, List.of(EventType.CLIENT_LOGIN)),
```
sguilhen, line 36: I wonder about this - the CLIENT_LOGIN event is, if I recall correctly, only issued in the Client Credentials Grant scenario. However, when trying to identify whether a client is not being used, we should probably also check the user LOGIN event and get the client from the that event as well. In other words, do we want to track only explicit client logins, or do we want to track if the client is being used as part of a user login?

sguilhen (changes requested):

server-spi-private/src/main/java/org/keycloak/models/workflow/ResourceOperationType.java
sguilhen: @pedroigor What do you think?

server-spi-private/src/main/java/org/keycloak/models/workflow/Workflow.java
sguilhen: So if we persist this as part of the workflow, every step needs to support the same type (including the restart step that supports all types and can be used in any workflow).

quarkus/tests/junit5/src/main/java/org/keycloak/it/utils/RawKeycloakDistribution.java
Hathoute: Correct, I forgot to exclude this change

Hathoute:
@sguilhen I actually have another WIP branch in fork where I have implemented what we discussed previously. I'll push everything here (with your latest comments). Do you prefer it being in a new commit or squashed into one?

server-spi-private/src/main/java/org/keycloak/models/workflow/ResourceOperationType.java
Hathoute: You're right, it would be bad to disable a client used for authentication flow. Maybe have a dedicated `ResourceOperationType` ?

sguilhen: tbh I haven't reached a good conclusion yet on how to best handle this ;-) 

We could perhaps make CLIENT_LOGGED_IN also accept the regular LOGIN event type and get the resourceId using that event's clientId. In this case, a single LOGIN event would result in two distinct workflow events - USER_LOGGED_IN and CLIENT_LOGGED_IN. But it mixes CLIENT_LOGIN with LOGIN and I'm not sure this is wanted all the time. Perhaps it is if the intention is to disable clients not used for any authentication in a period of time.

If we mix the two events, then we should perhaps have some conditions in place to allow admins to differentiate the clients that should be affected by the workflow. For example, you might want a particular policy for service accounts (AFAIK they only trigger CLIENT_LOGIN events, not regular LOGIN events), and a different one for the other clients involved in user authentication flows (would trigger only regular LOGIN events), but have them both activated on CLIENT_LOGGED_IN. What do you think?

@vramik @pedroigor Any opinions on this?

sguilhen:
> @sguilhen I actually have another WIP branch in fork where I have implemented what we discussed previously. I'll push everything here (with your latest comments). Do you prefer it being in a new commit or squashed into one?

I think having separate commits is better to review. We can always squash when merging.

server-spi-private/src/main/java/org/keycloak/models/workflow/ResourceOperationType.java
pedroigor: Would it help to introduce a separate event `USER_CLIENT_LOGGED_IN`? Reading it as the client involved during a user logged in event.

model/jpa/src/main/java/org/keycloak/models/workflow/JpaWorkflowStateProvider.java
Hathoute: I've seen that Workflows use the `Time` utils, so I just replaced it here.

server-spi-private/src/main/java/org/keycloak/models/workflow/Workflow.java
```diff
@@ -133,7 +158,23 @@ public void addSteps(List<WorkflowStepRepresentation> steps) {
...
+
+        if (allowedTypes.isEmpty()) {
+            throw new ModelValidationException("Steps provided are not compatible with each other.");
         }
+        else if (allowedTypes.size() > 1) {
+            String formattedTypes = allowedTypes.stream().map(Enum::name).collect(Collectors.joining(", "));
+            throw new ModelValidationException("Steps provided should support a single type, actual: " + formattedTypes);
+        }
+
+        ResourceType supported = allowedTypes.stream().findFirst().orElseThrow();
```
Hathoute, line 195: TBH I am not a fan of detecting the resource type of the workflow this way. I'd rather let the user choose it in a dropdown and then visually display the conditions/steps that are compatible (thinking about a create workflow UI)

Hathoute: @sguilhen what do you think?

pedroigor:
@sguilhen @Hathoute The PR is now rebased with latest main. Can you please check if I missed anything when resolving all the conflicts?

Hathoute:
@pedroigor I'll take a look this week, thank you.

pedroigor (approved):

sguilhen (approved):

sguilhen:
@Hathoute Thanks a lot for the contribution!

Issue #43235, opened 2025-10-07
Add `importOrder` and `forbidWildcardImports` to Spotless


stianst:
### Description

Update Spotless config to enable `importOrder` and `forbidWildcardImports`; and update existing code to pass the check

Pankraz76:
Kindly consider check solution:

- https://github.com/checkstyle/checkstyle/pull/17997

Pankraz76:
could you please tell how you fixed the wildcards? rewrite does not cover them all.

stianst:
> could you please tell how you fixed the wildcards? rewrite does not cover them all.

I used rewrite, but you need to configure a style for it to do all - see https://github.com/keycloak/keycloak/tree/main/misc/rewrite

Pankraz76:
nice, thanks for the update.

Pankraz76:
this would be the follow up and second layer:

- https://github.com/diffplug/spotless/pull/2744

PR #44220, opened 2025-11-14
[spotless] sync `spotless-maven-plugin`

Pankraz76:
imho having to manually sync as current approach is broken by design.

- https://github.com/keycloak/keycloak/issues/43235


ahus1:
As in previous cases, this is a pull request opened without referencing an open issue as outlined in the CONTRIBUTING.md

You will see progress in the use of rewrite and spotless eventually, but from our previous encounters this is an area not suitable for external contributors as it requires a lot of internal coordination. 

Please refrain from creating pull requests related to this topic. You will see improvements over time, although it might take a different route than you might expect.

Pankraz76:
yes understand.

Thanks for the update.

PR #44153, opened 2025-11-12
Use the unified constants class for sd-jwt/oid4vc standard data and c…

mposolda:
…laims

closes #44152

PR #44227, opened 2025-11-14
[OID4VCI] Redesign SDJwt API and handle keybinding JWT

### LINKED ISSUE #42091: [OID4VCI] SD-JWT implementation ignores Keybinding JWT

Captain-P-Goldfish:
fixes #42091

Redesigns the SdJwt API to be usable like this:

```java
KeyBindingJWT keyBindingJWT = KeyBindingJWT.builder()
                                                   /* header */
                                                   .withKid(holderKeybindingKey.getKeyId())
                                                   /* body */
                                                   .withIat(iat)
                                                   .withNbf(nbf)
                                                   .withExp(exp)
                                                   .withNonce(nonce)
                                                   .withAudience(audience)
                                                   .build();

        String disclosurePayload = """
            {
              "given_name": "Carlos",
              "family_name": "Norris"
            }
            """;
        ObjectNode disclosures = JsonSerialization.readValue(disclosurePayload, ObjectNode.class);
        DisclosureSpec disclosureSpec = DisclosureSpec.builder()
                                                      .withUndisclosedClaim("given_name", "123456789")
                                                      .withUndisclosedClaim("family_name", "987654321")
                                                      .build();
        IssuerSignedJWT issuerSignedJWT = IssuerSignedJWT.builder()
                                                         /* header */
                                                         .withKid(issuerJwk.getKeyId())
                                                         /* body */
                                                         .withClaims(disclosures, disclosureSpec)
                                                         .withKeyBinding(holderKeybindingKey)
                                                         .withIat(iat)
                                                         .withNbf(nbf)
                                                         .withExp(exp)
                                                         .build();

        SdJwt sdJwt = SdJwt.builder()
                           .withIssuerSignedJwt(issuerSignedJWT)
                           .withKeybindingJwt(keyBindingJWT)
                           .build(issuerSignerContext, holderSignerContext);
```


mposolda:
@Captain-P-Goldfish Thanks for the PR!

It seems that I was actually working on the same area :-) See PR https://github.com/keycloak/keycloak/pull/44207 . Was not aware that you are working on the same/similar stuff as well. It looks that your changes are quite complex, so I will probably re-work my PR 44207 on top of your stuff once it is merged.

But can you please try to first take a look at my PR with constants, so that we do not have conflicts: https://github.com/keycloak/keycloak/pull/44153 ?

So in summary, I suggest we address the PRs in the order:
- https://github.com/keycloak/keycloak/pull/44153 (my PR for constants)
- This PR https://github.com/keycloak/keycloak/pull/44227 (will need to rebase on top of my PR, which is already rebased on top of latest Keycloak main, which contain big changes related to Java imports (spotless plugin))
- Then I will try to re-work https://github.com/keycloak/keycloak/pull/44207 (or maybe will just close it).

@Captain-P-Goldfish WDYT?

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
```diff
@@ -35,20 +42,22 @@
...
+
+    static class FixedTimeClaimVerifier extends ClaimVerifier {
+
+        public FixedTimeClaimVerifier(int clockSkew, boolean requireClaims) {
+            super(createOptsWithClockSkew(clockSkew, requireClaims)
+                      .getVerifiers()
+                      .stream()
+                      .map(Mockito::spy)
+                      .peek(spy -> {
+                          Mockito.doReturn(Instant.ofEpochSecond(CURRENT_TIMESTAMP)).when(spy).getCurrentTimestamp();
```
mposolda: Can we try to avoid introducing `Mockito` ?

Besides that, I have one point: In Keycloak there is utility class called `Time`, which is used for all internal dealing with time and timestamps instead of direct usage of stuff like `Instant` . Among other things, it allows to call `setTimeOffset`, which allows to "Move" time forward (or backwards) by particular amount of seconds. See also other issue, which I've created for this: https://github.com/keycloak/keycloak/issues/44235 . Can you perhaps try to use it at least for `TimeClaimVerifier.currentTimestamp()` ? This will make it possible to be able to call `setTimeOffset` in tests (like other Keycloak tests use) instead of introducing Mockito.

mposolda (commented):
@Captain-P-Goldfish Your PR looks very good, Thanks! I've added one comment inline for now regarding "Mockito" and time in general. I will try to do more thorough review soon.

Besides that, I suggest that you review my PR https://github.com/keycloak/keycloak/pull/44153 and then rebase yours on top of that once that one is merged (mentioned in previous comment).

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJwtVerificationOpts.java
```diff
@@ -17,36 +17,105 @@
...
+    public static class Builder extends ClaimVerifier.Builder {
+
+        public Builder() {
+        }
+
+        public Builder(Integer clockSkew) {
+            super(clockSkew);
+        }
+
+        @Override
```
mposolda, line 56: Maybe if `ClaimVerifier.Builder` is defined as something like:

```
public class ClaimVerifier

   public static class Builder<T extends Builder> {
        
      @SuppressWarnings("unchecked")
      protected T self() {
          return (T) this;
      }

      public T withIatCheck(Integer allowedMaxAge) {
           ....
            return self();
        }
  ...
}
```

And then there is:
```
public IssuedSignedJwtVerificationOpts
  
... 
   public static class Builder extends ClaimVerifier.Builder<Builder> {
       
      
   }
```

Then there is no need to repeat all these methods on this class just to return `IssuerSignedJwtVerificationOpts.Builder` . As the methods will then automatically return type `IssuerSignedJwtVerificationOtps.Builder` (instead of the superclass `ClaimVerifier.Builder`) when called on the instance of `IssuerSignedJwtVerificationOtps.Builder`.

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
Captain-P-Goldfish: * I changed the currentTimestamp method to use the method `Time.currentTime()`.
* It is actually quite cumbersome to do this without mockito. This would require to refrain from using the builder-methods of the ClaimVerifier and use overridden implementations of the `predicates` which would somehow defeat the purpose of these tests. I just tried to use a delegate-approach but noticed that this also does not work because the `getCurrentTimestamp` is called directly on the delegate. Is it really such a hassle to have mockito in the test-dependencies? Why?

Captain-P-Goldfish:
> @Captain-P-Goldfish Thanks for the PR!
> 
> It seems that I was actually working on the same area :-) See PR #44207 . Was not aware that you are working on the same/similar stuff as well. It looks that your changes are quite complex, so I will probably re-work my PR 44207 on top of your stuff once it is merged.
> 
> But can you please try to first take a look at my PR with constants, so that we do not have conflicts: #44153 ?
> 
> So in summary, I suggest we address the PRs in the order:
> 
>     * [Use the unified constants class for sd-jwt/oid4vc standard data and c… #44153](https://github.com/keycloak/keycloak/pull/44153) (my PR for constants)
> 
>     * This PR [[OID4VCI] Redesign SDJwt API and handle keybinding JWT #44227](https://github.com/keycloak/keycloak/pull/44227) (will need to rebase on top of my PR, which is already rebased on top of latest Keycloak main, which contain big changes related to Java imports (spotless plugin))
> 
>     * Then I will try to re-work [Polishing of key-binding in sd-jwt SDK #44207](https://github.com/keycloak/keycloak/pull/44207) (or maybe will just close it).
> 
> 
> @Captain-P-Goldfish WDYT?


* sorry for not mentioning it. I was only discussing it with our colleagues from adorsys ^^° Also did not expect anyone else to work on it.
* And yes, the order sounds reasonable. Thanks
* I am also thinking about reworking the `SdJwtVP ` implementation. We will probably need a well working parser for one of our custom keycloak plugins, and therefore we need a well working  and usable API on the SdJwt end :-)

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
```diff
@@ -73,61 +75,124 @@ public String getAud() {
...
+            return (Builder) super.withIatCheck(isCheckOptional);
+        }
+
+        @Override
+        public KeyBindingJwtVerificationOpts.Builder withIatCheck(Integer allowedMaxAge, boolean isCheckOptional) {
+            return (Builder) super.withIatCheck(allowedMaxAge, isCheckOptional);
+        }
+
+        @Override
+        public KeyBindingJwtVerificationOpts.Builder withNbfCheck() {
```
mposolda, line 113: I think that my other comment (for the class `IssuerSignedJWTVerificationOpts.builder`) applies here as well. If we use parameterized types for the "builder", we do not need to repeat all these methods.

core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
```diff
@@ -226,9 +226,9 @@ public String present(List<String> disclosureDigests, JsonNode keyBindingClaims,
             return unboundPresentation;
         }
         String sd_hash = SdJwtUtils.hashAndBase64EncodeNoPad(unboundPresentation.getBytes(), getHashAlgorithm());
-        keyBindingClaims = ((ObjectNode) keyBindingClaims).put("sd_hash", sd_hash);
-        KeyBindingJWT keyBindingJWT = KeyBindingJWT.from(keyBindingClaims, holdSignatureSignerContext, jwsType);
-        sb.append(keyBindingJWT.toJws());
+        keyBindingClaims.put("sd_hash", sd_hash);
```
mposolda: Will be good to use constant from `OID4VC` (might be probably done automatically once you rebase this PR).

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
```diff
@@ -175,28 +308,90 @@ public void verifySdHashAlgorithm() throws VerificationException {
...
+                                               false/* TODO do we need this flag? */));
+        setJws(null);
+        setJwsInput(null);
+        Optional.ofNullable(signatureSignerContext).ifPresent(super::sign);
+    }
+
     // SD-JWT Claims
     public static final String CLAIM_NAME_SELECTIVE_DISCLOSURE = "_sd";
     public static final String CLAIM_NAME_SD_HASH_ALGORITHM = "_sd_alg";
+    public static final String CLAIM_NAME_SD_HASH = "sd_hash";
```
mposolda: No need to have this constant here as it would be in `OID4VCConstants` (I guess you probably cleanup this once you rebase).

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
```diff
@@ -218,21 +413,107 @@ public Builder withNestedDisclosures(boolean nestedDisclosures) {
...
+            getClaims().add(new VisibleSdJwtClaim(SdJwtClaimName.of(ClaimVerifier.CLAIM_NAME_EXP), new LongNode(exp)));
+            return this;
+        }
+
+        /**
+         * this method requires the public key to be present in the keybindingJwts header as "jwk" claim
+         */
+        public Builder withKeyBinding(KeyBindingJWT keyBinding) {
+            ObjectNode cnf = JsonNodeFactory.instance.objectNode();
+            Optional.ofNullable(keyBinding.getJwsHeader().getOtherClaims().get("jwk"))
```
mposolda: Could be replaced with constants from `OID4VCConstants` once you rebase (similarly like `cnf` and some other occurences of `jwk` inside this class).

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
```diff
@@ -218,21 +413,107 @@ public Builder withNestedDisclosures(boolean nestedDisclosures) {
...
+        public Builder withClaim(SdJwtClaim sdJwtClaim) {
+            getClaims().add(sdJwtClaim);
             return this;
         }
 
         public IssuerSignedJWT build() {
             // Preinitialize hashAlg to sha-256 if not provided
             hashAlg = hashAlg == null ? "sha-256" : hashAlg;
-            jwsType = jwsType == null ? "dc+sd-jwt" : jwsType;
+            jwsHeader.setType(jwsHeader.getType() == null ? "dc+sd-jwt" : jwsHeader.getType());
```
mposolda: This may also need some update once you rebase to avoid hardcoded constants from here.

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJWT.java
```diff
@@ -16,38 +16,150 @@
...
+        }
+
+        public Builder withPayload(ObjectNode payload) {
+            this.payload = payload;
+            return this;
+        }
+
+        public Builder withIat(long iat)
+        {
+            getPayload().put(ClaimVerifier.CLAIM_NAME_IAT, iat);
```
mposolda: Will need some changes to use the `OID4VCConstants` once you rebase (similarly like some other methods on this class).

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
mposolda: Hmm... Actually I am thinking that when you use something like this in the test:

```
TIMESTAMP2021 = 1609459200L; // Fixed timestamp: 2021-01-01 00:00:00 UTC

int currentTime = Time.currentTime();
Time.setOffset(TIMESTAMP2021 - currentTime);  // Move current time to 2021-01-01

// Some testing here
...

Time.setOffset(0);
```

I suppose it will have same effect like using Mockito to hardcode current time to 2021-01-01 and hence builder methods and everything like that should work the same? But maybe I am missing something? If it is really hard/impossible to test without builder methods, feel free to keep it in this PR. I may try to remove it later as a follow-up...

mposolda (commented):
@Captain-P-Goldfish Thanks again!

I've added some more review comments. But it is mostly minor stuff. I think that SDK is better with this update and vote for including it. Most of the comments are related to rebase this PR to the current main (after "Constants" PR was merged).

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
IngridPuppet: Same comment. There should be a way to avoid repetition.

core/src/main/java/org/keycloak/representations/JsonWebToken.java
```diff
@@ -263,4 +266,43 @@ public void setOtherClaims(String name, Object value) {
...
+        result = 31 * result + Objects.hashCode(subject);
+        result = 31 * result + Objects.hashCode(type);
+        result = 31 * result + Objects.hashCode(issuedFor);
+        result = 31 * result + Objects.hashCode(otherClaims);
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return JsonSerialization.mapper.convertValue(this, JsonNode.class).toPrettyString();
```
IngridPuppet: Do you think using `JsonSerialization.writeValueAsPrettyString(this)` still does the conversion to JsonNode internally? I wonder if we could avoid the double conversion.

Also do we want a "pretty" format? That can be expensive as a default.

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
```diff
@@ -73,61 +75,124 @@ public String getAud() {
...
         }
 
-        public Builder withNonce(String nonce) {
-            this.nonce = nonce;
-            return this;
+        @Override
+        public KeyBindingJwtVerificationOpts.Builder withAudCheck(String expectedAud) {
+            this.aud = expectedAud;
+            return (Builder) super.withAudCheck(expectedAud);
         }
```
IngridPuppet: Could we add another option for supporting Regex Patterns? Wallets don't seem to build this string consistently and we would like to support matching "(.*:)?%s".

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
```diff
@@ -73,61 +75,124 @@ public String getAud() {
...
 
         @Override
         public KeyBindingJwtVerificationOpts build() {
             if (!validateIssuedAtClaim) {
                 throw new IllegalArgumentException(
-                        "Validating `iat` claim cannot be disabled for KB-JWT verification because mandated"
+                    "Validating `iat` claim cannot be disabled for KB-JWT verification because mandated"
                 );
             }
```
IngridPuppet: Are we keeping this attribute? My understanding is that this should be dropped in favor of "withIatCheck".

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,528 @@
...
+import java.util.stream.Collectors;
+
+import org.keycloak.common.VerificationException;
+import org.keycloak.common.util.Time;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Module for checking the validity of JWT time claims. All checks account for a leeway window to accommodate clock
```
IngridPuppet: We're dropping the term leeway to align with the rest of codebase, which uses clock skew.
```suggestion
 * Module for checking the validity of JWT time claims. All checks account for a small window to accommodate clock
```

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,528 @@
...
+ * @author <a href="mailto:Ingrid.Kamga@adorsys.com">Ingrid Kamga</a>
+ */
+public class ClaimVerifier {
+
+    public static final int DEFAULT_CLOCK_SKEW = 10;
+    public static final int DEFAULT_ALLOWED_MAX_AGE = 5 * 60; // 5 minutes
+
+    public static final String CLAIM_NAME_IAT = "iat";
+    public static final String CLAIM_NAME_EXP = "exp";
+    public static final String CLAIM_NAME_NBF = "nbf";
```
IngridPuppet: You'll surely come to notice that these are now constants just to be imported.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,528 @@
...
+        this.contentVerifiers = contentVerifiers;
+    }
+
+    public void verifyClaims(ObjectNode header, ObjectNode body) throws VerificationException {
+        for (Predicate<ObjectNode> verifier : headerVerifiers) {
+            verifier.test(header);
+        }
+        for (Predicate<ObjectNode> verifier : contentVerifiers) {
+            verifier.test(body);
+        }
```
IngridPuppet: What about?
```suggestion
        verifyHeaderClaims(header);
        verifyBodyClaims(body);
```

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
```diff
@@ -73,61 +75,124 @@ public String getAud() {
...
         }
 
-        public Builder withNonce(String nonce) {
-            this.nonce = nonce;
-            return this;
+        @Override
+        public KeyBindingJwtVerificationOpts.Builder withAudCheck(String expectedAud) {
+            this.aud = expectedAud;
+            return (Builder) super.withAudCheck(expectedAud);
         }
```
IngridPuppet, line 95: Is it okay to keep the nonce and aud properties? I fear this creates two sources of truth.

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
```diff
@@ -73,61 +75,124 @@ public String getAud() {
...
+                    "Validating `iat` claim cannot be disabled for KB-JWT verification because mandated"
                 );
             }
 
             if (keyBindingRequired && (aud == null || nonce == null || nonce.isEmpty())) {
                 throw new IllegalArgumentException(
-                        "Missing `nonce` and `aud` claims for replay protection"
+                    "Missing `nonce` and `aud` claims for replay protection"
                 );
             }
```
IngridPuppet: Do you think it will be possible to build these opts without calling "withAudCheck" but a custom "withClaimCheck"?

This is our use case: Wallets don't seem to build the aud string consistently and we would like to support matching the regex "(.*:)?%s" pattern.

I believe this check should be reworked to make it possible.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,521 @@
...
+            this.expectedClaimValue = expectedClaimValue;
+            this.stringComparator = Optional.ofNullable(stringComparator).orElseGet(ClaimCheck::getDefaultComparator);
+            this.invertCheck = invertCheck;
+            this.isOptional = isOptional;
+        }
+
+        protected static BiFunction<String, String, Boolean> getDefaultComparator() {
+            return (s, s2) -> (s == null && s2 == null)
+                || (s != null && s.equals(s2));
+        }
```
IngridPuppet, line 157: Add a Java doc that it checks for equality? Also, is it any different from Objects::equals?

```suggestion
        protected static BiFunction<String, String, Boolean> getDefaultComparator() {
            return Objects::equals;
        }
```

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJwtVerificationOpts.java
Captain-P-Goldfish: The idea is actually pretty good. But I fail to make it work with 2 hierarchy levels. 
`KeyBindingJwtVerificationOpts extends IssuerSignedJwtVerificationOpts`. And this does not work anymore. Have tried quite a few things now to trick it, but without any luck...

core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
Captain-P-Goldfish: already fixed

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,521 @@
...
+        private final String claimName;
+
+        private final String expectedClaimValue;
+
+        private final BiFunction<String, String, Boolean> stringComparator;
+
+        /**
+         * can be used to make sure that the given claim does NOT have the expected given value
+         */
+        private final boolean invertCheck;
```
IngridPuppet: Negating the check of stringComparator? In my opinion, this attribute will not be easy to reason about. And also, because probably rarely used, we should only leave the option to directly implement a negated stringComparator. WDYT?

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,521 @@
...
+                                        .orElse(null);
+            if (claimValue == null && !isOptional) {
+                throw new VerificationException(String.format("Missing claim '%s' in token", claimName));
+            }
+
+            boolean checkSuccessful;
+            if (invertCheck) {
+                checkSuccessful = !stringComparator.apply(expectedClaimValue, claimValue);
+            } else {
+                checkSuccessful = stringComparator.apply(expectedClaimValue, claimValue);
```
IngridPuppet: A case for using the ternary operator.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,521 @@
...
+                    audiences.add(audienceNode.textValue());
+                }
+            }
+            else {
+                audiences.add(audienceArray.textValue());
+            }
+
+            if (audiences.contains(expectedAudience)) {
+                return true;
+            }
```
IngridPuppet, line 407:
```suggestion
            return audiences.contains(expectedAudience);
```

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
```diff
@@ -0,0 +1,521 @@
...
+            return withIatCheck(allowedMaxAge, false);
+        }
+
+        public Builder withIatCheck(boolean isCheckOptional) {
+            return withIatCheck(allowedMaxAge, isCheckOptional);
+        }
+
+        public Builder withIatCheck(Integer allowedMaxAge, boolean isCheckOptional) {
+            this.allowedMaxAge = Optional.ofNullable(allowedMaxAge).orElse(0);
+            contentVerifiers.removeIf(verifier -> verifier instanceof ClaimVerifier.IatLifetimeCheck);
```
IngridPuppet: What if we have a custom ClaimCheck on `iat`? Is it intended behavior that it does not override a default IatCheck?

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
```diff
@@ -26,74 +29,184 @@
...
         super(jwsString);
+        this.disclosureSpec = null;
+        this.disclosureClaims = new ArrayList<>();
+        this.decoyClaims = new ArrayList<>();
+    }
+
+    public IssuerSignedJWT(DisclosureSpec disclosureSpec,
+                           ObjectNode disclosureClaims) {
+        this(disclosureSpec, disclosureClaims, "sha-256");
+    }
```
IngridPuppet: You should reuse the constant for the default hash algorithm.

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
```diff
@@ -133,14 +259,19 @@ private static JsonNode generatePayloadString(List<SdJwtClaim> claims, List<Deco
...
+                          payload.set(claim.getClaimNameAsString(), claim.getVisibleClaimValue(hashAlg));
+                      });
+
 
-        return payload;
+        try {
+            return payload;
+        } catch (Exception ex) {
+            throw new IllegalArgumentException("Got illegal content in claims", ex);
+        }
```
IngridPuppet: Can this ever fail ??? Did you mean to add a check?

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
```diff
@@ -219,21 +413,107 @@ public Builder withNestedDisclosures(boolean nestedDisclosures) {
...
+                this.jwsHeader = new JWSHeader();
+            }
+            this.jwsHeader.setType(jwsType);
+            return this;
+        }
+
+        public Builder withJwsHeader(JWSHeader jwsHeader) {
+            this.jwsHeader = jwsHeader;
+            return this;
+        }
```
IngridPuppet, line 427: We run into a conflict if withJwsType is set before withJwsHeader.

core/src/main/java/org/keycloak/sdjwt/SdJws.java
IngridPuppet, line 1: Did you mean to delete the file at once?

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
```diff
@@ -108,22 +110,22 @@ private Map<String, String> computeDigestDisclosureMap(List<String> disclosureSt
...
         // Validate disclosures.
         JsonNode disclosedPayload = validateDisclosuresDigests();
 
-        // Validate time claims.
+        // Validate time claims and algorithm header claim.
         // Issuers will typically include claims controlling the validity of the SD-JWT in plaintext in the
         // SD-JWT payload, but there is no guarantee they would do so. Therefore, Verifiers cannot reliably
         // depend on that and need to operate as though security-critical claims might be selectively disclosable.
-        validateIssuerSignedJwtTimeClaims(disclosedPayload, issuerSignedJwtVerificationOpts);
+        issuerSignedJwtVerificationOpts.verify(issuerSignedJwt);
```
IngridPuppet, line 128: As mentioned in the comment above (excerpted from the spec), it is actually wanted to validate time claims on the disclosed payload.

core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
IngridPuppet, line 1: The file does not appear deleted as intended.

core/src/main/java/org/keycloak/sdjwt/TimeClaimVerifier.java
IngridPuppet, line 1: The file does not appear deleted as intended.

core/src/test/java/org/keycloak/sdjwt/SdJwtCreationAndSigningTest.java
```diff
@@ -0,0 +1,400 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * @author Pascal Knueppel
+ * @since 13.11.2025
+ */
+public class SdJwtCreationAndSigningTest extends OID4VCIssuerEndpointTest {
+
+    @Test
```
IngridPuppet, line 67: Going through the test suite, it seems to belong more in the core package IMO. Notice that no functionality implemented in the services package is tested. I therefore believe than extending the `OID4VCIssuerEndpointTest` class creates a "false" dependency that can be circumvented. WDYT?

IngridPuppet (commented):
Hello @Captain-P-Goldfish - Thank you for this valuable update of the API. Reviewing the PR, I left a few cosmetic suggestions and asked a couple of questions. Please could you check?

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJwtVerificationOpts.java
mposolda: @Captain-P-Goldfish I've tried something like this for supporting multiple levels of inheritance:

```
    public static class Builder1<T extends Builder1> {

        @SuppressWarnings("unchecked")
        protected T self() {
            return (T) this;
        }

        public T iat() {
            // Some logic here...
            return self();
        }

    }

    public static class Builder2<T extends Builder2> extends Builder1<T> {

    }


    public static class Builder3<T extends Builder3> extends Builder2<T> {

    }


    public static Builder1<Builder1<?>> builder1() {
        return new Builder1<>();
    }

    public static Builder2<Builder2<?>> builder2() {
        return new Builder2<>();
    }

    public static Builder3<Builder3<?>> builder3() {
        return new Builder3<>();
    }

    public void testMe() {
        Builder1<?> b1 = builder1().iat(); // Returns type Builder1
        Builder2<?> b2 = builder2().iat(); // Returns type Builder2
        Builder3<?> b3 = builder3().iat(); // Returns type Builder3
    }
```

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
mposolda: Perhaps something like this might work https://github.com/keycloak/keycloak/pull/44227#discussion_r2539358453 ?

core/src/main/java/org/keycloak/representations/JsonWebToken.java
mposolda: +1 . Is it possible to use the same like `JWSHeader` class?

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJwtVerificationOpts.java
mposolda: However after trying some more testing, I figured this may not work when there are multiple methods (not only `iat`...). Not 100% sure if there is good way, which allows to avoid repetition and at the same time, allow to use inheritance for builders. 
@Captain-P-Goldfish @IngridPuppet maybe repetition is fine if we cannot figure better way?

mposolda: @Captain-P-Goldfish @IngridPuppet Maybe alternative is to use the builder API similar to what infinispan uses. For example something like:
```
import org.infinispan.configuration.cache.Configuration;
import org.infinispan.configuration.cache.ConfigurationBuilder;

....

        Configuration conf = new ConfigurationBuilder()
            .memory()
                .whenFull(EvictionStrategy.REMOVE)
                .maxCount(InfinispanConnectionProvider.KEYS_CACHE_DEFAULT_MAX)
            .statistics().enabled(true)
                .build();
```

Especially see the class `AbstractConfigurationChildBuilder`  from infinispan for seeing more about this approach.

However at the same time, I am not 100% sure if it worth to do it considering that this API is relatively simple (in comparison to infinispan)?

For me, repetition is probably OK as done in the Pascal's PR.

IngridPuppet: Yes, I agree it'd be overkill to have multiple builder levels for this feature. Repetition is finally not that bad.

Captain-P-Goldfish: okay, I leave this part then as it is?

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
Captain-P-Goldfish: I think your idea might work. But I am not a fan of it because the whole class is used with static methods and members.

I can do this change but I still would like to understand why Mockito is a problem in the unit-test section?

core/src/main/java/org/keycloak/representations/JsonWebToken.java
Captain-P-Goldfish: this was an accident. I never meant to use `prettyPrint` :-)
changed the call to `JsonSerialization.writeValueAsPrettyString(this)`

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
Captain-P-Goldfish: good catch. Have overlooked this one.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
Captain-P-Goldfish: already resolved :-)

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
Captain-P-Goldfish: I would actually prefer it like this. This is easier to read instead of:

`boolean checkSuccessful = invertCheck != stringComparator.apply(expectedClaimValue, claimValue);`
or
`boolean checkSuccessful = invertCheck ? !stringComparator.apply(expectedClaimValue, claimValue) : stringComparator.apply(expectedClaimValue, claimValue);`

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
Captain-P-Goldfish: I would also prefer the current state due to the errorMessage. This message will help administrators to find the underlying problem if the check fails.

core/src/main/java/org/keycloak/sdjwt/SdJws.java
Captain-P-Goldfish: I just renamed it to `JwsToken`

core/src/main/java/org/keycloak/sdjwt/TimeClaimVerificationOpts.java
Captain-P-Goldfish: thanks. Merge conflicts I erroneously resolved with IntelliJ.

core/src/test/java/org/keycloak/sdjwt/SdJwtCreationAndSigningTest.java
Captain-P-Goldfish: The reason it is placed here is the security-provider that is not available in the core. The test `testCreateSdJwtWithoutKeybindingAndNoSignature` does not use signing but the test `testCreateSdJwtWithKeybindingJwt` does, and the other test is pretty close to that one why I also have left it here.

Is that okay?

core/src/main/java/org/keycloak/sdjwt/SdJws.java
Captain-P-Goldfish: after reading your other comments I noticed that this file was also not deleted. Fixed

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
Captain-P-Goldfish: I just noticed, that your suggestion has a minor flaw.
If we do this the tests would partially become unreliable due to the call of System.currentTimeMillis(). If we set the offset the value returned from currentTime in the `iatCheck` might differ by a few milliseconds based on the current execution time which might lead to unwanted test-results. Which leads us back to mockito being the more stable option ^^°

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
Captain-P-Goldfish: I think I am not getting where you are pointing at

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
Captain-P-Goldfish: I do not know what you mean

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
Captain-P-Goldfish: So you suggest to check if an aud-value check is present, either as ClaimCheck or as AudCheck?

Captain-P-Goldfish: I tried it like this now:
```java
boolean isAudCheckPresent = contentVerifiers.stream().anyMatch(verifier -> {
                return verifier instanceof AudienceCheck ||
                    (verifier instanceof ClaimCheck && ((ClaimCheck) verifier).getClaimName().equals(JsonWebToken.AUD));
            });
            boolean isNonceCheckPresent = contentVerifiers.stream().anyMatch(verifier -> {
                return verifier instanceof ClaimCheck && ((ClaimCheck) verifier).getClaimName().equals(IDToken.NONCE)
                    && Optional.ofNullable(((ClaimCheck) verifier).getExpectedClaimValue()).map(s -> !s.isEmpty())
                               .orElse(false);
            });
            if (keyBindingRequired && isAudCheckPresent && isNonceCheckPresent) {
                throw new IllegalArgumentException("Missing `nonce` and `aud` claims for replay protection");
            }
```

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
Captain-P-Goldfish: I removed the invert-boolean and added a new class that should work smoothly with the comparison checks since it extends `ClaimCheck`:

```java
public static class NegatedClaimCheck extends ClaimCheck {
        
        public NegatedClaimCheck(String claimName, String expectedClaimValue) {
            super(claimName, expectedClaimValue);
        }

        public NegatedClaimCheck(String claimName, String expectedClaimValue, boolean isOptional) {
            super(claimName, expectedClaimValue, isOptional);
        }

        public NegatedClaimCheck(String claimName,
                                 String expectedClaimValue,
                                 BiFunction<String, String, Boolean> stringComparator) {
            super(claimName, expectedClaimValue, stringComparator);
        }

        public NegatedClaimCheck(String claimName,
                                 String expectedClaimValue,
                                 BiFunction<String, String, Boolean> stringComparator,
                                 boolean isOptional) {
            super(claimName, expectedClaimValue, stringComparator, isOptional);
        }

        @Override
        public boolean test(ObjectNode t) throws VerificationException {
            String claimValue = Optional.ofNullable(t.get(getClaimName())).map(JsonNode::asText).map(String::valueOf)
                                        .orElse(null);
            if (claimValue == null && !isOptional()) {
                throw new VerificationException(String.format("Missing claim '%s' in token", getClaimName()));
            }
            if (claimValue == null && isOptional()) {
                // if optional and not present we do not want to execute the check of the parent.
                return true;
            }
            return !super.test(t);
        }
    }
```

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJwtVerificationOpts.java
mposolda: +1

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
mposolda: Not 100% sure about the details, but I think the tests should be written in a way that few milliseconds should not mean a big difference. For example for checking the expirations etc, it is always better to make sure that there is no test for testing token, which is valid for 1 second only (as then it can accidentally expire before some test logic, which assumes valid token, is executed), but rather make sure that tokens are valid for example for 30 seconds or more to make sure that token will not accidentally expire during the test. Then when testing expiration, you can set time offset for bigger value (EG. 31) to make sure that token is expired.

This approach is used everywhere through Keycloak testsuite and works just fine. Right now, not seeing a reason why it should not work here TBH...

Feel free to keep mockito in this PR. I can try to fix it as a follow-up PR to make sure that this PR is in asap.

core/src/test/java/org/keycloak/sdjwt/SdJwtCreationAndSigningTest.java
mposolda: @Captain-P-Goldfish +1 to this point by @IngridPuppet 

The security provider is actually available if you use the same pattern as used in some other `core` tests. For example take a look at `SdJwtVPVerificationTest`. It is the abstract class, which defines this rule:
```
@ClassRule
public static CryptoInitRule cryptoInitRule = new CryptoInitRule();
```

Actually there are 3 subclasses of it in the various `crypto/*` modules. It seems same pattern could be used for you as well?

Captain-P-Goldfish: have already tried it before and it did not work. I am getting this error in that case:

```
java.lang.IllegalStateException: Not able to load any cryptoProvider with the classLoader: jdk.internal.loader.ClassLoaders$AppClassLoader@76ed5528

	at org.keycloak.common.crypto.CryptoIntegration.detectProvider(CryptoIntegration.java:65)
	at org.keycloak.common.crypto.CryptoIntegration.init(CryptoIntegration.java:33)
	at org.keycloak.rule.CryptoInitRule.before(CryptoInitRule.java:15)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:50)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:237)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)
```

```java

public class SdJwtCreationAndSigningTest {

    @ClassRule
    public static CryptoInitRule cryptoInitRule = new CryptoInitRule();

    @Test
    public void testCreateSdJwtWithoutKeybindingAndNoSignature() throws Exception {
    ...
```

And yes, I moved the class just back to the core-module :-)

core/src/main/java/org/keycloak/sdjwt/IssuerSignedJwtVerificationOpts.java
IngridPuppet: Yes

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
IngridPuppet: I noticed that the `nonce` and `aud` properties on the builder were essentially kept for the null check in the `build()` method. Let's close this point and focus on https://github.com/keycloak/keycloak/pull/44227#discussion_r2544358428, which would naturally resolve this.

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
IngridPuppet: Thank you for the update. I think this addresses my point. It should now be possible not to keep the `aud` and `nonce` properties in the builder.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
IngridPuppet: I tend to agree. Thank you for raising issues of readability.

core/src/main/java/org/keycloak/sdjwt/ClaimVerifier.java
IngridPuppet: Oops, you're right. My suggestion is not identical at all.

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
IngridPuppet: I'm referring to section https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-22.html#name-selectively-disclosable-val

> Issuers will typically include claims controlling the validity of the SD-JWT in plaintext in the SD-JWT payload, but there is no guarantee they would do so. Therefore, Verifiers cannot reliably depend on that and need to operate as though security-critical claims might be selectively disclosable.

In `issuerSignedJwt`, selectively disclosable fields cannot be accessed. Meaning, for example, if `exp` is not in plaintext, this verification will miss it. In `disclosedPayload`, we should have all disclosable fields disclosed.

core/src/test/java/org/keycloak/sdjwt/SdJwtCreationAndSigningTest.java
IngridPuppet: Try making the class **abstract** and providing concrete empty implementations under `crypto/*`. 

Check these files for reference:

_The abstract class testsuite_
core/src/test/java/org/keycloak/sdjwt/sdjwtvp/SdJwtVPVerificationTest.java

_The concrete implementations_
crypto/default/src/test/java/org/keycloak/crypto/def/test/sdjwt/DefaultCryptoSdJwtVPVerificationTest.java
crypto/elytron/src/test/java/org/keycloak/crypto/elytron/test/sdjwt/ElytronCryptoSdJwtVPVerificationTest.java
crypto/fips1402/src/test/java/org/keycloak/crypto/fips/test/sdjwt/ElytronCryptoSdJwtVPVerificationTest.java

Captain-P-Goldfish: thanks @IngridPuppet . Fixed.

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
Captain-P-Goldfish: If I understand you right, you want to make sure that iat, nonce etc. is checked even if not in the body structure of the SDJwt?

I would actually think that this should be done after a stable backwards implementation is present, that is able to create an IssuerSignedJwt with all disclosures readded into the body of the JWS. Meaning resolve disclosures create a normal JSON Body from it and then validate it. What do you think?

This would be out of scope for this commit though^^°

IngridPuppet: > If I understand you right, you want to make sure that iat, nonce etc. is checked even if not in the body structure of the SDJwt?

Yes

> I would actually think that this should be done after a stable backwards implementation is present, that is able to create an IssuerSignedJwt with all disclosures readded into the body of the JWS. Meaning resolve disclosures create a normal JSON Body from it and then validate it. What do you think?

Do you realize the immediate instruction above returns just that as `disclosedPayload`? If you mean the type is problematic, do you think something like this could work in the meantime?

```java
    IssuerSignedJWT disclosedIssuerSignedJWT = new IssuerSignedJWT(
            issuerSignedJwt.getJwsHeader(), (ObjectNode) disclosedPayload);
    issuerSignedJwtVerificationOpts.verify(disclosedIssuerSignedJWT);
```

You'll notice that some tests will precisely fail because the `iat` and `exp` claims were probably selectively disclosable.

core/src/test/java/org/keycloak/sdjwt/TimeClaimVerifierTest.java
Captain-P-Goldfish: I personally like Mockito, especially because it makes testing such specific usecases very easy. :-)

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
Captain-P-Goldfish: > Do you realize the immediate instruction above returns just that as disclosedPayload? If you mean the type is problematic, do you think something like this could work in the meantime?

I just took a look. No I did not check it before in detail...
But I will take a much closer look in the near future, since we will also need this part.

> You'll notice that some tests will precisely fail because the iat and exp claims were probably selectively disclosable.

In that case I am once more not sure what you just expect me to do here now. All tests have passed now. And since this code-snippet is currently only being called from unit-tests I do not see any problem right now. But in the future if the code is called from production we will need to think of a way how to set the `iat`, `ext` and `nbf` checks as optional if necessary. We certainly do not want to make them optional for all credentials.
But currently I would say there is no need to handle it now.

core/src/main/java/org/keycloak/sdjwt/vp/KeyBindingJwtVerificationOpts.java
Captain-P-Goldfish: okay I have removed these as unncessary now.

mposolda (commented):
@Captain-P-Goldfish I've sent PR to your branch to remove mockito and update TimeClaimVerifier to rely on `Time` and time offsets like other Keycloak tests. Also removing fragility to make sure there is not only 1 second difference when testing timestamps to avoid issues when time moves forward by 1 second (or more) during the tests. PR is here: https://github.com/Captain-P-Goldfish/keycloak/pull/1

If you merge it, it should automatically your PR to Keycloak main. Your PR would be fine for me then and ready to merge to main (once also @IngridPuppet approves the PR to make sure that his concerns are addressed).

core/src/main/java/org/keycloak/sdjwt/SdJwtVerificationContext.java
IngridPuppet: Thank you for your response. We can live without this for now indeed.

IngridPuppet (approved):
The PR looks good to me. Thank you for the updates.

Captain-P-Goldfish:
@mposolda thanks, I merged your pull-request. I personally would have still preferred the mockito variant but this is personal preference :-)

Thank you both @mposolda and @IngridPuppet for your great reviews :-)

mposolda (approved):
@Captain-P-Goldfish @IngridPuppet Thanks for all the updates and thorought review!

PR #44230, opened 2025-11-14
Updating the specifics around kubernetes service accounts

### LINKED ISSUE #44064: Kubernetes service account default for federated client authentication
### Description

The subject in the JWT created for Kuberentes service account tokens is `system:serviceaccount:<namespace>:<serviceaccount>`.

If the user didn't assign an explicit service account to the Pod, the service account name is "default" which could lead to setups where the multiple Pods use the same subject identifier, leading to a potentially insecure setup. 

~So we should check that the subject is not the default service account.~ -> As discussed below, this will just be documented. 

This needs to be reflected: 
* the docs

ahus1:
Closes #44064

### DIFF
--- a/docs/documentation/server_admin/topics/identity-broker/kubernetes.adoc
+++ b/docs/documentation/server_admin/topics/identity-broker/kubernetes.adoc
@@ -37,11 +37,21 @@ It depends on the preview feature `client-auth-federated`.
 |===
 +
 . When you create a new realm with the preview feature `client-auth-federated` enabled, the client authentication flow is already configured correctly. For existing realms, add to the client authentication flow the execution of *Signed JWT - Federated* as an alternative step. As built-in flows can not be updated, and if the default flow is your default, you will first need to duplicate the existing clients
-. For each confidential OIDC client that should authenticate via this provider:
+. For each confidential OpenID Connect client that should authenticate via this provider:
 .. Change in the *Credentials* tab the *Client Authenticator* to *Signed JWT - Federated*.
 .. As *Identity provider*, enter the alias of the Kubernetes identity provider added in step 3.
 .. As *Federated Subject*, enter the subject identifier as issued by Kubernetes. This is usually `system:serviceaccount:<namespace>:<serviceaccount>`.
-. For the Pod in Kubernetes add a service account :
++
+[NOTE]
+====
+Each client must have a unique subject identifier within a realm.
+As the subject identifier is built from the namespace and service account name, each client must have its own service account if multiple clients share a namespace.
+
+As a security best practice, do not use the `default` service account in a namespace, as it is shared with all Pods in a namespace.
+Instead, create an individual service account for each client.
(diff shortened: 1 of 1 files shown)


docs/documentation/server_admin/topics/identity-broker/kubernetes.adoc
```diff
@@ -37,11 +37,21 @@ It depends on the preview feature `client-auth-federated`.
...
-. For each confidential OIDC client that should authenticate via this provider:
+. For each confidential OpenID Connect client that should authenticate via this provider:
 .. Change in the *Credentials* tab the *Client Authenticator* to *Signed JWT - Federated*.
 .. As *Identity provider*, enter the alias of the Kubernetes identity provider added in step 3.
 .. As *Federated Subject*, enter the subject identifier as issued by Kubernetes. This is usually `system:serviceaccount:<namespace>:<serviceaccount>`.
-. For the Pod in Kubernetes add a service account :
++
+[NOTE]
+====
+Each client must have a unique subject identifier within a realm.
```
ahus1, line 47: @stianst / @slaskawi - I wonder if we instead should make this "a unique combination of subject identifier and IDP alias, as the same subject identifier can repeat across different IDP aliases. 

If that is the case, the FederatedJWTCLientAuthenticator must first identify the IDP via lookup by the issuer. 

WDYT?

stianst: I'd drop this line and focus more on the Kubernetes side of things; basically what you are explaining below. 

However, it's not about Keycloak clients, but rather about workloads/pods om the Kubernetes side. It's also irrelevant if a workload/pod is configured as a Keycloak client or not from this perspective.

slaskawi: @ahus1 Could you please help me understand what would be the practical use cases that you're explaining here?

stianst (approved):

docs/documentation/server_admin/topics/identity-broker/kubernetes.adoc
ahus1: @slaskawi - the first situation is that you can't have multiple clients connected to the same namespace and service account. We should state this in the docs. 

There is one more thing: If you assume that one Keycloak instance connects to two Kubernetes instances (assuming they are called "dev" and "prod"), and have both the same workloads deployed. If this is the case, they would likely have the same namespaces and service accounts. And then there would be a conflict. 

As the next challenge, each Kubernetes cluster would have an issuer URL if it is a Kubernetes-local URL like `https://kubernetes.default.svc.cluster.local` in the case of Minikube.  So as of today, there is nothing how we could distinguish two of the same subject identifiers coming from two different Kubernetes clusters. 

@stianst and me discussed it briefly earlier today and created issue #44253 to have more of a discussion. Let's continue over there.

PR #44233, opened 2025-11-14
Speed up local container build with default Minikube setup

### LINKED ISSUE #44232: Improve building containers in Minikube
### Description

I tried build the Keycloak images on Minikube which is using docker as described in the Operator's README. 

When doing this, the build was very slow. By changing the ulimit, it was then a lot faster.

ahus1:
Closes #44232

### DIFF
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -1,7 +1,12 @@
 FROM registry.access.redhat.com/ubi9 AS ubi-micro-build
 
 ADD target/ubi-null.sh /tmp/
-RUN bash /tmp/ubi-null.sh java-21-openjdk-headless glibc-langpack-en
+
+# Workaround around slow docker buildx when executing yum/dnf when running for example against minikube
+# https://github.com/keycloak/keycloak/issues/38801#issuecomment-2815350773
+# https://github.com/docker/buildx/issues/379#issuecomment-1052980242
+# As this option is not allowed when running in podman, ignore if this fails
+RUN (ulimit -n 1024000 || true) && bash /tmp/ubi-null.sh java-21-openjdk-headless glibc-langpack-en
 
 FROM registry.access.redhat.com/ubi9-micro
 ENV LANG=en_US.UTF-8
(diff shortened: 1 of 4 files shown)


ahus1:
@shawkins - can you please review this PR, as you have been involved in the ulimit problem before? Thanks!

shawkins:
> can you please review this PR, as you have been involved in the ulimit problem before? Thanks!

I basically found it impossible to change my docker's ulimit on fedora, so I've generally put the workaround in locally if I need to do anything docker specific. I'm not opposed to putting this into the script. To align with our docs it might be better to put the workaround in the Dockerfile where it will also be more visible.

ahus1:
@shawkins - I pushed one more change ... did you have something like this change in mind? 

With this setup I now needed to change two files. 

What is strange: When I run the build via Maven it is slow:

```
mvn clean package -Dquarkus.kubernetes.image-pull-policy=IfNotPresent -Dquarkus.container-image.build=true
```

Unfortunately it doesn't log anything about the docker build on the console, and I don't know where to debug it. It logs that it is using minikube for that:

```
[INFO] Docker host IP address is 192.168.39.165
[INFO] Connected to docker: 
  Server Version: 28.4.0
  API Version: 1.51
  Operating System: Buildroot 2025.02
  Total Memory: 7937 MB
  Labels: 
    provider=kvm2
```

When I run the same docker command it prints out on the console, it is fast. I am also sure it uses minikube, as a `docker ps` shows me all the Kubernetes containers. 

Why's that?

shawkins:
Using `ulimit -n 1024000 &&` works fine, while `(ulimit -n 1024000 || true) &&` does not because I believe the parens are introducing a subshell.

This works:

```
RUN <<EOF
ulimit -n 1024000 || true
/tmp/ubi-null.sh java-21-openjdk-headless glibc-langpack-en
EOF
```

Another approach, which either I missed before or didn't seem to previously work, is to add `quarkus.docker.additional-args=--ulimit,nofile=1024000` to the operator application.properties and update the workaround instructions for anyone hitting this issue to add the --ulimit option to their docker build command https://docs.docker.com/reference/cli/docker/buildx/build/#ulimit

PR #44238, opened 2025-11-14
MLDSA Keys

Jonaka3385:
Adding Key Providers & Algorithm-Identifiers for ML-DSA. As a part of PR [#43857](https://github.com/keycloak/keycloak/pull/43857).

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -159,6 +159,8 @@ public class Profile {
 
         OPENAPI("OpenAPI specification served at runtime", Type.EXPERIMENTAL, CLIENT_ADMIN_API_V2),
 
+        PQC_ML_DSA("Experimental ML-DSA support for JWKs and Signatures", Type.EXPERIMENTAL),
+
         /**
          * @see <a href="https://github.com/keycloak/keycloak/issues/37967">Deprecate for removal the Instagram social broker</a>.
          */
--- a/core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
+++ b/core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
@@ -40,6 +40,10 @@ public class JavaAlgorithm {
     public static final String SHA512 = "SHA-512";
     public static final String SHAKE256 = "SHAKE256";
 
+    public static final String ML_DSA_44 = "ML-DSA-44";
+    public static final String ML_DSA_65 = "ML-DSA-65";
+    public static final String ML_DSA_87 = "ML-DSA-87";
+
     public static String getJavaAlgorithm(String algorithm) {
         return getJavaAlgorithm(algorithm, null);
     }
@@ -70,6 +74,12 @@ public class JavaAlgorithm {
                 return PS384;
             case Algorithm.PS512:
                 return PS512;
+            case Algorithm.ML_DSA_44:
+                return ML_DSA_44;
+            case Algorithm.ML_DSA_65:
+                return ML_DSA_65;
+            case Algorithm.ML_DSA_87:
+                return ML_DSA_87;
             case Algorithm.EdDSA:
                 if (curve != null) {
                     return curve;
@@ -142,6 +152,10 @@ public class JavaAlgorithm {
             case Algorithm.Ed448:
             case Algorithm.Ed25519:
                 return KeyType.OKP;
+            case Algorithm.ML_DSA_44:
+            case Algorithm.ML_DSA_65:
+            case Algorithm.ML_DSA_87:
+                return KeyType.AKP;
             default:
                 return KeyType.OCT;
         }
@@ -157,6 +171,10 @@ public class JavaAlgorithm {
 
     public static boolean isEddsaJavaAlgorithm(String algorithm) {
         return getJavaAlgorithm(algorithm).contains("Ed");
+    }
+
+    public static boolean isMldsaJavaAlgorithm(String algorithm) {
+        return getJavaAlgorithm(algorithm).startsWith("ML-DSA-");
     }
 
     public static boolean isHMACJavaAlgorithm(String algorithm) {
--- a/services/src/main/java/org/keycloak/keys/AbstractMldsaKeyProvider.java
+++ b/services/src/main/java/org/keycloak/keys/AbstractMldsaKeyProvider.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.keys;
+
+import org.keycloak.common.util.KeyUtils;
+import org.keycloak.component.ComponentModel;
+import org.keycloak.crypto.KeyStatus;
+import org.keycloak.crypto.KeyType;
+import org.keycloak.crypto.KeyUse;
+import org.keycloak.crypto.KeyWrapper;
+import org.keycloak.models.RealmModel;
+
+import java.security.KeyPair;
+import java.util.stream.Stream;
+
+public abstract class AbstractMldsaKeyProvider implements KeyProvider {
+
+    private final KeyStatus status;
+
+    private final ComponentModel model;
+
+    private final KeyWrapper key;
+
+    protected final String algorithm;
+
+    public AbstractMldsaKeyProvider(RealmModel realm, ComponentModel model) {
+        this.model = model;
+        this.status = KeyStatus.from(model.get(Attributes.ACTIVE_KEY, true), model.get(Attributes.ENABLED_KEY, true));
+        this.algorithm = model.get(Attributes.ALGORITHM_KEY);
+
+        if (model.hasNote(KeyWrapper.class.getName())) {
+            key = model.getNote(KeyWrapper.class.getName());
+        } else {
+            key = loadKey(realm, model);
+            model.setNote(KeyWrapper.class.getName(), key);
+        }
+    }
+
+    protected abstract KeyWrapper loadKey(RealmModel realm, ComponentModel model);
+
+    @Override
+    public Stream<KeyWrapper> getKeysStream() {
+        return Stream.of(key);
+    }
+
+    protected KeyWrapper createKeyWrapper(KeyPair keyPair) {
+        KeyWrapper key = new KeyWrapper();
+
+        key.setProviderId(model.getId());
+        key.setProviderPriority(model.get("priority", 0l));
+
+        key.setKid(KeyUtils.createKeyId(keyPair.getPublic()));
+        key.setUse(KeyUse.SIG);
+        key.setType(KeyType.AKP);
+        key.setAlgorithm(algorithm);
+        key.setStatus(status);
+        key.setPrivateKey(keyPair.getPrivate());
+        key.setPublicKey(keyPair.getPublic());
+
+        return key;
+    }
+}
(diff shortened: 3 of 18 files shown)


common/src/main/java/org/keycloak/common/crypto/CryptoConstants.java
```diff
@@ -15,6 +15,11 @@ public class CryptoConstants {
     public static final String ECDH_ES_A192KW = "ECDH-ES+A192KW";
     public static final String ECDH_ES_A256KW = "ECDH-ES+A256KW";
 
+    /* ML-DSA */
+    public static final String ML_DSA_44 = "ML-DSA-44";
+    public static final String ML_DSA_65 = "ML-DSA-65";
+    public static final String ML_DSA_87 = "ML-DSA-87";
+
```
stianst: Remove, these are not needed/used
```suggestion
```

core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
```diff
@@ -124,6 +135,10 @@ public static String getJavaAlgorithmForHash(String algorithm, String curve) {
                     }
                 }
                 return SHA512;
+            case ML_DSA_44:
+            case ML_DSA_65:
+            case ML_DSA_87:
+                return SHAKE256;
```
stianst: Where did you discover that this should be SHAKE256?

core/src/main/java/org/keycloak/jose/jws/Algorithm.java
stianst, line 1: Remove changes to this file, it's deprecated and we should be using `org.keycloak.crypto.Algorithm`

services/src/main/java/org/keycloak/keys/GeneratedMldsaKeyProviderFactory.java
stianst, line 1: Should implement EnvironmentDependentProviderFactory, with:

```
    public boolean isSupported(Config.Scope config) {
        return Profile.isFeatureEnabled(Profile.Feature.PQC_MLDSA);
    }
```

services/pom.xml
```diff
@@ -283,6 +283,10 @@
         	<groupId>org.keycloak</groupId>
         	<artifactId>keycloak-config-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk18on</artifactId>
+        </dependency>
```
stianst: Services should not have a direct dependency on BouncyCastle

services/src/main/java/org/keycloak/keys/JavaKeystoreKeyProviderFactory.java
```diff
@@ -159,15 +160,29 @@ private static void validateCertificateChain(List<X509Certificate> certificates)
...
-        List<String> algorithms = Stream.of(
-                        List.of(Algorithm.AES, Algorithm.EdDSA),
-                        List.of(Algorithm.ES256, Algorithm.ES384, Algorithm.ES512),
-                        Attributes.HS_ALGORITHM_PROPERTY.getOptions(),
-                        Attributes.RS_ALGORITHM_PROPERTY.getOptions(),
-                        Attributes.RS_ENC_ALGORITHM_PROPERTY.getOptions(),
-                        GeneratedEcdhKeyProviderFactory.ECDH_ALGORITHM_PROPERTY.getOptions())
-                .flatMap(Collection::stream)
-                .toList();
+        List<String> algorithms;
```
stianst: Don't duplicate current lists of lists, but rather add a condition to append to the original list of Feature.PQC_MLDSA feature is enabled. Using `Stream.of` or just creating a new linked list and adding current + ML-DSA to it.

services/src/main/java/org/keycloak/keys/loader/HardcodedPublicKeyLoader.java
stianst, line 1: Remove changes to this as it's irrelevant to providing ML-DSA keys for Keycloak

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/util/TokenSignatureUtil.java
stianst, line 1: Remove changes to this as it's not used to test generated keys

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/util/KeyUtils.java
stianst, line 1: Remove changes to this as it's not needed for this PR

tests/utils/src/main/java/org/keycloak/tests/utils/KeyUtils.java
```diff
@@ -67,6 +67,14 @@ public static KeyPair generateEdDSAKey(String curve) throws NoSuchAlgorithmExcep
         return kpg.generateKeyPair();
     }
 
+    public static KeyPair generateMLDSAKey(String algorithm) throws NoSuchAlgorithmException, NoSuchProviderException {
+        if (!JavaAlgorithm.isMldsaJavaAlgorithm(algorithm)) {
+            throw new NoSuchAlgorithmException("Not a ML-DSA Algorithm");
+        }
```
stianst: No need for this check; just remove it

services/src/main/java/org/keycloak/keys/AbstractMldsaKeyProviderFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.spec.NamedParameterSpec;
+
+import static org.keycloak.provider.ProviderConfigProperty.LIST_TYPE;
+
+public abstract class AbstractMldsaKeyProviderFactory implements KeyProviderFactory {
+
+    protected static final String MLDSA_PRIVATE_KEY_KEY = "mldsaPrivateKey";
+    protected static final String MLDSA_PUBLIC_KEY_KEY = "mldsaPublicKey";
```
stianst: Maybe we should use Attributes.privateKey and Attributes.certificate here for consistency?

services/src/main/java/org/keycloak/keys/GeneratedMldsaKeyProvider.java
stianst, line 1: For storing the keys I'd just use key.getEncoded directly. We can store the full X.509 representation and don't need to strip it as it's done for JWKs. That should remove the need to add BouncyCastle dependency directly in services module

stianst (changes requested):
See a few comments inline. Some additional things:

* Don't include merge commits; use rebase
* Squash to a single commit linking to https://github.com/keycloak/keycloak/issues/44142
* You need to do the DCO sign-of (see contribution guide for instructions)
* We need more tests (verifying generated keys + fallback keys)

core/src/main/java/org/keycloak/crypto/JavaAlgorithm.java
stianst: This can be postponed, and maybe is a non-issue though as this is only used for `at_hash` and `c_hash` for ID tokens using hybrid flows.

I'd remove this change from here for now as it's irrelevant to adding keys to the server. More an issue when testing/supporting different flows with ML-DSA

common/src/main/java/org/keycloak/common/Profile.java
```diff
@@ -157,6 +157,8 @@ public enum Feature {
 
         OPENAPI("OpenAPI specification served at runtime", Type.EXPERIMENTAL, CLIENT_ADMIN_API_V2),
 
+        ML_DSA("Experimental ML-DSA support for JWKs and Signatures", Type.EXPERIMENTAL),
```
stianst: I'd prefix this with PQC
```suggestion
        PQC_ML_DSA("Experimental ML-DSA support for JWKs and Signatures", Type.EXPERIMENTAL),
```

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
stianst, line 1: Not sure why you are updating this file?

js/apps/admin-ui/src/utils/useIsFeatureEnabled.ts
stianst, line 1: Not sure why you are updating this file?

services/src/test/java/org/keycloak/jose/jwk/ServerJWKTest.java
stianst, line 1: This is not adding any additional testing, and not sure why this exists at all; as we are already testing JWK in https://github.com/keycloak/keycloak/tree/main/crypto/default/src/test/java/org/keycloak/crypto/def/test

stianst (commented):
Take a look at https://github.com/keycloak/keycloak/blob/main/tests/base/src/test/java/org/keycloak/tests/keys/GeneratedRsaKeyProviderTest.java

You can also use RunOnServer to get the keys directly from the manager, to check Key Wrappers directly to make sure the Java representations are correct.

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
Jonaka3385: I'm using this to hide ML-DSA in the frontend (DropDown menu for Default Signature Algorithm). Imo it's the easiest way for a Feature-Flag.

stianst: This PR doesn't include any signature providers; and in a follow-up PR when it does they should be environment dependent so won't be loaded at all on the server side as long as the feature is not enabled. So this should not be needed at all.

tests/base/src/test/java/org/keycloak/tests/keys/GeneratedMldsaKeyProviderTest.java
stianst, line 1: This is not properly testing the key provider; it's merely testing the generateMldsaKeyPair method, so this is a unit test, not a intergration test. Please see how other tests are using admin endpoints to create new keys; you can also use RunOnServer in addition to validate the keys themselves if you want.

Jonaka3385: Reworked it again, it should be correct now.

services/src/main/java/org/keycloak/keys/JavaKeystoreKeyProvider.java
```diff
@@ -195,6 +198,21 @@ private KeyWrapper loadEdDSAKey(KeyStore keyStore, String keyAlias, KeyUse keyUs
...
 
+    private KeyWrapper loadMLDSAKey(KeyStore keyStore, String keyAlias, KeyUse keyUse, String algorithm) throws GeneralSecurityException {
+        KeyStore.PrivateKeyEntry privateKeyEntry = checkKeyEntry(keyStore, keyAlias, KeyStore.PrivateKeyEntry.class, keyUse);
+        PrivateKey privateKey = checkKey(privateKeyEntry.getPrivateKey(), keyAlias, PrivateKey.class, null);
+        X509Certificate x509Cert = checkCertificate(privateKeyEntry.getCertificate());
+        try {
+            JavaAlgorithm.getJavaAlgorithmForHash(algorithm);
+        } catch (RuntimeException e) {
+            throw new UnrecoverableKeyException("Invalid params.");
+        }
```
stianst: Remove as this block does nothing. It will already be caught by the switch in `loadKey`.

services/src/main/java/org/keycloak/keys/JavaKeystoreKeyProvider.java
stianst, line 1: Should not allow loading ML-DSA keys if the `PQC_ML_DSA` feature is not enabled. This will have some implications on tests; as we'd need to have a separate test for it that enables the feature.

services/src/main/java/org/keycloak/keys/AbstractMldsaKeyProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final KeyWrapper key;
+
+    protected final String algorithm;
+
+    public AbstractMldsaKeyProvider(RealmModel realm, ComponentModel model) {
+        this.model = model;
+        this.status = KeyStatus.from(model.get(Attributes.ACTIVE_KEY, true), model.get(Attributes.ENABLED_KEY, true));
+
+        this.algorithm = realm.getDefaultSignatureAlgorithm();
```
stianst: This makes no sense; the applicable algorithm should be loaded from the model, and not use the realm default signature algorithm.

stianst (changes requested):

stianst (changes requested):
Have you run the tests locally? They don't pass here and can't as there's a number of problems in the tests (not enabling the feature for one), it still also uses `realm.getDefaultSignatureAlgorithm()` some places, tests are not setting algorithm in the component, etc.

Jonaka3385:
When I run the tests locally, I get a http error. But I don't really know why (maybe the SSLSocket that's used).

PR #44255, opened 2025-11-17
[OID4VCI] Credential Offer must be created by Issuer not Holder

tdiesler:
closes #44116

This PR introduces CredentialOfferState that is decoupled from the any login session. It is acted upon getCredentialOffer and (later) upon getCredential. To create a credential offer, one must have the credential-offer-create role i.e. this is an issuer admin task and not something anybody can do. The credential offer is potentially bound to a target clientId and/or a target userId. Other constraints could also be enforced through this mechanism.

Once, the credential-offer-uri has been created successfully and (somehow) delivered to the wallet. There is no pre-requisite of any existing login session to redeem the credential-offer. 

Credential Offer Validity Matrix
| pre-auth | clientId | userId | Valid | Notes                                               |
|----------|----------|--------|-------|-----------------------------------------------------|
| no       | no       | no     | yes   | Generic offer; any logged-in user may redeem.       |
| no       | no       | yes    | yes   | Offer restricted to a specific user.                |
| no       | yes      | no     | yes   | Bound to client; user determined at login.          |
| no       | yes      | yes    | yes   | Bound to both client and user.                      |
|----------|----------|--------|-------|-----------------------------------------------------|
| yes      | no       | no     | no    | Pre-auth requires a user subject; missing userId.   |
| yes      | yes      | no     | no    | Same as above; userId required.                     |
| yes      | no       | yes    | yes   | Pre-auth for a specific user; client unconstrained. |
| yes      | yes      | yes    | yes   | Fully constrained: user + client.                   |

In the pre-auth offer case, the AccessTokenResponse does contain one AuthorizationDetail with both properties (i.e. credential_identifiers, credential_configuration_id). This is now natively supported in AccessTokenResponse.

The spec talks about what the wallet has to use in the CredentialRequest. It must use a  credential identifier in the pre-auth case - this is now enforced in the endpoint.

Using the credential_configuration_id in the CredentialRequest is still supported.
Using the credential scope as credential identifier is (imho) a hack that was used by many test cases - support for that has been removed and tests have been migrated to credential_configuration_id.

The credential endpoint now does this ...

```
        // When the CredentialRequest contains a credential identifier the caller must have gone through the
        // CredentialOffer process or otherwise have set up a valid CredentialOfferState

        if (credentialRequestVO.getCredentialIdentifier() != null) {
            var credId = credentialRequestVO.getCredentialIdentifier();

            // Retrieve the associated credential offer state
            //
            var offerStorage = session.getProvider(CredentialOfferStorage.class);
            var offerState = offerStorage.findOfferStateByCredentialId(session, credId);
            if (offerState == null) {
                var errorMessage = "No credential offer state for credential id: " + credId;
                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_IDENTIFIER, errorMessage));
            }

            var authDetails = offerState.getAuthorizationDetails();
            String credConfigId = (String) authDetails.getOtherClaims().get("credential_configuration_id");

            if (credConfigId == null) {
                var errorMessage = "Cannot credential configuration mapping for: " + credId;
                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
            }
```

In short, a credential identifier can (currently) only be obtained through a pre-auth credential offer.

The tests that have a credential identifier, now also use it.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -30,7 +30,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.UUID;
 import java.util.stream.Collectors;
 
 import jakarta.ws.rs.BadRequestException;
@@ -324,7 +323,7 @@ public class OID4VCIssuerEndpoint {
     }
 
     /**
-     * Creates a Credential Offer Uri that is not bound to the calling target user.
+     * Creates a Credential Offer Uri that is bound to the calling user.
      */
     public Response getCredentialOfferURI(String credConfigId) {
         UserSessionModel userSession = getAuthenticatedClientSession().getUserSession();
@@ -332,7 +331,7 @@ public class OID4VCIssuerEndpoint {
     }
 
     /**
-     * Creates a Credential Offer Uri that is bound to a specific target user.
+     * Creates a Credential Offer Uri that is bound to a specific user.
      */
     public Response getCredentialOfferURI(String credConfigId, boolean preAuthorized, String targetUser) {
         return getCredentialOfferURI(credConfigId, preAuthorized, null, targetUser, OfferUriType.URI, 0, 0);
@@ -390,27 +389,15 @@ public class OID4VCIssuerEndpoint {
 
         // Check required role to create a credential offer
         //
-        boolean hasCredentialOfferRole = userModel.getRoleMappingsStream().anyMatch(rm -> rm.getName().equals(CREDENTIAL_OFFER_CREATE.getName()));
+        boolean hasCredentialOfferRole = userModel.getRoleMappingsStream()
+                .anyMatch(rm -> rm.getName().equals(CREDENTIAL_OFFER_CREATE.getName()));
         if (!hasCredentialOfferRole) {
             var errorMessage = "Credential offer creation requires role: " + CREDENTIAL_OFFER_CREATE.getName();
             throw new CorsErrorResponseException(cors,
-                    INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
+                    INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.FORBIDDEN);
         }
 
         LOGGER.debugf("Get an offer for %s", credConfigId);
-
-        // Check whether the credential configuration exists in available client scopes
-        //
-        List<String> availableInClientScopes = session.clientScopes()
-                .getClientScopesByProtocol(realmModel, OID4VC_PROTOCOL)
-                .map(it -> it.getAttribute(CredentialScopeModel.CONFIGURATION_ID))
-                .toList();
-        if (!availableInClientScopes.contains(credConfigId)) {
-            var errorMessage = "Invalid credential configuration id: " + credConfigId;
-            LOGGER.debugf("%s not found in supported credential config ids: %s", credConfigId, availableInClientScopes);
-            throw new CorsErrorResponseException(cors,
-                    INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
-        }
 
         // Check whether given client/user ids actually exist
         //
@@ -437,6 +424,19 @@ public class OID4VCIssuerEndpoint {
             }
         }
 
+        // Check whether the credential configuration exists in available client scopes
+        //
+        List<String> availableInClientScopes = session.clientScopes()
+                .getClientScopesByProtocol(realmModel, OID4VC_PROTOCOL)
+                .map(it -> it.getAttribute(CredentialScopeModel.CONFIGURATION_ID))
+                .toList();
+        if (!availableInClientScopes.contains(credConfigId)) {
+            var errorMessage = "Invalid credential configuration id: " + credConfigId;
+            LOGGER.debugf("%s not found in supported credential config ids: %s", credConfigId, availableInClientScopes);
+            throw new CorsErrorResponseException(cors,
+                    INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
+        }
+
         CredentialsOffer credOffer = new CredentialsOffer()
                 .setCredentialIssuer(OID4VCIssuerWellKnownProvider.getIssuer(session.getContext()))
                 .setCredentialConfigurationIds(List.of(credConfigId));
@@ -445,7 +445,7 @@ public class OID4VCIssuerEndpoint {
         CredentialOfferState offerState = new CredentialOfferState(credOffer, appClientId, appUserId, expiration);
 
         if (preAuthorized) {
-            String code = "urn:oid4vci:code:" + UUID.randomUUID();
+            String code = "urn:oid4vci:code:" + SecretGenerator.getInstance().randomString(64);
             credOffer.setGrants(new PreAuthorizedGrant().setPreAuthorizedCode(
                     new PreAuthorizedCode().setPreAuthorizedCode(code)));
         }
@@ -665,12 +665,13 @@ public class OID4VCIssuerEndpoint {
         // Both credential_configuration_id and credential_identifier are optional.
         // If the credential_configuration_id is present, credential_identifier can't be present.
         // But this implementation will tolerate the presence of both, waiting for clarity in specifications.
-        // This implementation will privilege the presence of the credential_configuration_id.
-        String requestedCredentialConfigurationId = credentialRequestVO.getCredentialConfigurationId();
-        String requestedCredentialIdentifier = credentialRequestVO.getCredentialIdentifier();
+        // This implementation will privilege the presence of credential_identifier.
+
+        String credentialIdentifier = credentialRequestVO.getCredentialIdentifier();
+        String credentialConfigurationId = credentialRequestVO.getCredentialConfigurationId();
 
         // Check if at least one of both is available.
-        if (requestedCredentialConfigurationId == null && requestedCredentialIdentifier == null) {
+        if (credentialIdentifier == null && credentialConfigurationId == null) {
             LOGGER.debugf("Missing both credential_configuration_id and credential_identifier. At least one must be specified.");
             throw new BadRequestException(getErrorResponse(ErrorType.MISSING_CREDENTIAL_IDENTIFIER_AND_CONFIGURATION_ID));
         }
@@ -680,15 +681,14 @@ public class OID4VCIssuerEndpoint {
         // When the CredentialRequest contains a credential identifier the caller must have gone through the
         // CredentialOffer process or otherwise have set up a valid CredentialOfferState
 
-        if (credentialRequestVO.getCredentialIdentifier() != null) {
-            String credId = credentialRequestVO.getCredentialIdentifier();
+        if (credentialIdentifier != null) {
 
             // Retrieve the associated credential offer state
             //
             CredentialOfferStorage offerStorage = session.getProvider(CredentialOfferStorage.class);
-            CredentialOfferState offerState = offerStorage.findOfferStateByCredentialId(session, credId);
+            CredentialOfferState offerState = offerStorage.findOfferStateByCredentialId(session, credentialIdentifier);
             if (offerState == null) {
-                var errorMessage = "No credential offer state for credential id: " + credId;
+                var errorMessage = "No credential offer state for credential id: " + credentialIdentifier;
                 throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_IDENTIFIER, errorMessage));
             }
 
@@ -705,8 +705,8 @@ public class OID4VCIssuerEndpoint {
             //
             SupportedCredentialConfiguration credConfig = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session).get(credConfigId);
             if (credConfig == null) {
-                LOGGER.errorf("Mapped credential configuration not found: %s", credConfigId);
-                throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
+                var errorMessage = "Mapped credential configuration not found: " + credConfigId;
+                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
             }
 
             // Verify the user login session
@@ -714,34 +714,36 @@ public class OID4VCIssuerEndpoint {
             UserSessionModel userSession = authResult.session();
             UserModel userModel = userSession.getUser();
             if (!userModel.getUsername().equals(offerState.getUserId())) {
-                LOGGER.errorf("Unexpected login user: %s != %s", userModel.getUsername(), offerState.getUserId());
-                throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST));
+                var errorMessage = "Unexpected login user: " + userModel.getUsername();
+                LOGGER.errorf(errorMessage + " != %s", offerState.getUserId());
+                throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST, errorMessage));
             }
 
             // Verify the login client
             //
             ClientModel clientModel = session.getContext().getClient();
             if (offerState.getClientId() != null && !clientModel.getClientId().equals(offerState.getClientId())) {
-                LOGGER.errorf("Unexpected login client: %s != %s", clientModel.getClientId(), offerState.getClientId());
-                throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST));
+                var errorMessage = "Unexpected login client: " + clientModel.getClientId();
+                LOGGER.errorf(errorMessage + " != %s", offerState.getClientId());
+                throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST, errorMessage));
             }
 
             // Find the configured scope in the login client
             //
             ClientScopeModel clientScope = clientModel.getClientScopes(false).get(credConfig.getScope());
             if (clientScope == null) {
-                LOGGER.errorf("Client scope not found: %s", credConfig.getScope());
-                throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
+                var errorMessage = String.format("Client scope not found: %s", credConfig.getScope());
+                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
             }
 
             requestedCredential = new CredentialScopeModel(clientScope);
-            LOGGER.debugf("Successfully mapped credential identifier %s to scope %s", credId, clientScope.getName());
+            LOGGER.debugf("Successfully mapped credential identifier %s to scope %s", credentialIdentifier, clientScope.getName());
 
-        } else if (credentialRequestVO.getCredentialConfigurationId() != null) {
+        } else if (credentialConfigurationId != null) {
             // Use credential_configuration_id for direct lookup
             requestedCredential = credentialRequestVO.findCredentialScope(session).orElseThrow(() -> {
-                LOGGER.errorf("Credential scope not found for configuration ID: %s", credentialRequestVO.getCredentialConfigurationId());
-                return new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
+                var errorMessage = "Credential scope not found for configuration id: " + credentialConfigurationId;
+                return new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
             });
         } else {
             // Neither provided - this should not happen due to earlier validation
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -108,6 +108,7 @@ import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.protocol.oid4vc.utils.ClaimsPathPointer;
 import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantType;
 import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantTypeFactory;
+import org.keycloak.protocol.oidc.rar.AuthorizationDetailsResponse;
 import org.keycloak.representations.AccessToken;
 import org.keycloak.representations.dpop.DPoP;
 import org.keycloak.saml.processing.api.util.DeflateUtil;
@@ -389,7 +390,7 @@ public class OID4VCIssuerEndpoint {
 
         // Check required role to create a credential offer
         //
-        var hasCredentialOfferRole = userModel.getRoleMappingsStream().anyMatch(rm -> rm.getName().equals(CREDENTIAL_OFFER_CREATE.getName()));
+        boolean hasCredentialOfferRole = userModel.getRoleMappingsStream().anyMatch(rm -> rm.getName().equals(CREDENTIAL_OFFER_CREATE.getName()));
         if (!hasCredentialOfferRole) {
             var errorMessage = "Credential offer creation requires role: " + CREDENTIAL_OFFER_CREATE.getName();
             throw new CorsErrorResponseException(cors,
@@ -400,7 +401,7 @@ public class OID4VCIssuerEndpoint {
 
         // Check whether the credential configuration exists in available client scopes
         //
-        var availableInClientScopes = session.clientScopes()
+        List<String> availableInClientScopes = session.clientScopes()
                 .getClientScopesByProtocol(realmModel, OID4VC_PROTOCOL)
                 .map(it -> it.getAttribute(CredentialScopeModel.CONFIGURATION_ID))
                 .toList();
@@ -441,7 +442,7 @@ public class OID4VCIssuerEndpoint {
                 .setCredentialConfigurationIds(List.of(credConfigId));
 
         int expiration = timeProvider.currentTimeSeconds() + preAuthorizedCodeLifeSpan;
-        var offerState = new CredentialOfferState(credOffer, appClientId, appUserId, expiration);
+        CredentialOfferState offerState = new CredentialOfferState(credOffer, appClientId, appUserId, expiration);
 
         if (preAuthorized) {
             String code = "urn:oid4vci:code:" + UUID.randomUUID();
@@ -449,7 +450,7 @@ public class OID4VCIssuerEndpoint {
                     new PreAuthorizedCode().setPreAuthorizedCode(code)));
         }
 
-        var offerStorage = session.getProvider(CredentialOfferStorage.class);
+        CredentialOfferStorage offerStorage = session.getProvider(CredentialOfferStorage.class);
         offerStorage.putOfferState(session, offerState);
 
         LOGGER.debugf("Stored credential offer state: [ids=%s, cid=%s, uid=%s, nonce=%s]",
@@ -536,8 +537,8 @@ public class OID4VCIssuerEndpoint {
 
         // Retrieve the associated credential offer state
         //
-        var offerStorage = session.getProvider(CredentialOfferStorage.class);
-        var offerState = offerStorage.findOfferStateByNonce(session, nonce);
+        CredentialOfferStorage offerStorage = session.getProvider(CredentialOfferStorage.class);
+        CredentialOfferState offerState = offerStorage.findOfferStateByNonce(session, nonce);
         if (offerState == null) {
             var errorMessage = "No credential offer state for nonce: " + nonce;
             eventBuilder.event(INTROSPECT_TOKEN_ERROR).detail(Details.REASON, errorMessage).error(Errors.INVALID_REQUEST);
@@ -547,7 +548,7 @@ public class OID4VCIssuerEndpoint {
         // We treat the credential offer URI as an unprotected capability URL and rely solely on the later authorization step
         // i.e. an authenticated client/user session is not required nor checked against the offer state
 
-        var credOffer = offerState.getCredentialsOffer();
+        CredentialsOffer credOffer = offerState.getCredentialsOffer();
         LOGGER.debugf("Found credential offer state: [ids=%s, cid=%s, uid=%s, nonce=%s]",
                 credOffer.getCredentialConfigurationIds(), offerState.getClientId(), offerState.getUserId(), offerState.getNonce());
 
@@ -680,12 +681,12 @@ public class OID4VCIssuerEndpoint {
         // CredentialOffer process or otherwise have set up a valid CredentialOfferState
 
         if (credentialRequestVO.getCredentialIdentifier() != null) {
-            var credId = credentialRequestVO.getCredentialIdentifier();
+            String credId = credentialRequestVO.getCredentialIdentifier();
 
             // Retrieve the associated credential offer state
             //
-            var offerStorage = session.getProvider(CredentialOfferStorage.class);
-            var offerState = offerStorage.findOfferStateByCredentialId(session, credId);
+            CredentialOfferStorage offerStorage = session.getProvider(CredentialOfferStorage.class);
+            CredentialOfferState offerState = offerStorage.findOfferStateByCredentialId(session, credId);
             if (offerState == null) {
                 var errorMessage = "No credential offer state for credential id: " + credId;
                 throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_IDENTIFIER, errorMessage));
@@ -693,7 +694,7 @@ public class OID4VCIssuerEndpoint {
 
             // Get the credential_configuration_id from AuthorizationDetails
             //
-            var authDetails = offerState.getAuthorizationDetails();
+            AuthorizationDetailsResponse authDetails = offerState.getAuthorizationDetails();
             String credConfigId = (String) authDetails.getOtherClaims().get("credential_configuration_id");
             if (credConfigId == null) {
                 var errorMessage = "No credential_configuration_id in AuthorizationDetails";
@@ -702,7 +703,7 @@ public class OID4VCIssuerEndpoint {
 
             // Find the credential configuration in the Issuer's metadata
             //
-            var credConfig = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session).get(credConfigId);
+            SupportedCredentialConfiguration credConfig = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session).get(credConfigId);
             if (credConfig == null) {
                 LOGGER.errorf("Mapped credential configuration not found: %s", credConfigId);
                 throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
@@ -710,8 +711,8 @@ public class OID4VCIssuerEndpoint {
 
             // Verify the user login session
             //
-            var userSession = authResult.session();
-            var userModel = userSession.getUser();
+            UserSessionModel userSession = authResult.session();
+            UserModel userModel = userSession.getUser();
             if (!userModel.getUsername().equals(offerState.getUserId())) {
                 LOGGER.errorf("Unexpected login user: %s != %s", userModel.getUsername(), offerState.getUserId());
                 throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST));
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -107,7 +107,6 @@ import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.protocol.oid4vc.utils.ClaimsPathPointer;
 import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantType;
 import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantTypeFactory;
-import org.keycloak.protocol.oidc.rar.AuthorizationDetailsResponse;
 import org.keycloak.representations.AccessToken;
 import org.keycloak.representations.dpop.DPoP;
 import org.keycloak.saml.processing.api.util.DeflateUtil;
@@ -694,8 +693,8 @@ public class OID4VCIssuerEndpoint {
 
             // Get the credential_configuration_id from AuthorizationDetails
             //
-            AuthorizationDetailsResponse authDetails = offerState.getAuthorizationDetails();
-            String credConfigId = (String) authDetails.getOtherClaims().get("credential_configuration_id");
+            OID4VCAuthorizationDetailsResponse authDetails = offerState.getAuthorizationDetails();
+            String credConfigId = authDetails.getCredentialConfigurationId();
             if (credConfigId == null) {
                 var errorMessage = "No credential_configuration_id in AuthorizationDetails";
                 throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -691,33 +691,51 @@ public class OID4VCIssuerEndpoint {
                 throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_IDENTIFIER, errorMessage));
             }
 
+            // Get the credential_configuration_id from AuthorizationDetails
+            //
             var authDetails = offerState.getAuthorizationDetails();
             String credConfigId = (String) authDetails.getOtherClaims().get("credential_configuration_id");
-
             if (credConfigId == null) {
-                var errorMessage = "Cannot credential configuration mapping for: " + credId;
+                var errorMessage = "No credential_configuration_id in AuthorizationDetails";
                 throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
             }
 
-            // Use the mapped credential configuration ID to find the credential scope
-            Map<String, SupportedCredentialConfiguration> supportedCredentials = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session);
-            if (supportedCredentials.containsKey(credConfigId)) {
-                SupportedCredentialConfiguration config = supportedCredentials.get(credConfigId);
-                ClientModel client = session.getContext().getClient();
-                Map<String, ClientScopeModel> clientScopes = client.getClientScopes(false);
-                ClientScopeModel clientScope = clientScopes.get(config.getScope());
-                if (clientScope != null) {
-                    requestedCredential = new CredentialScopeModel(clientScope);
-                    LOGGER.debugf("Successfully mapped credential identifier %s to configuration %s",
-                            credentialRequestVO.getCredentialIdentifier(), credConfigId);
-                } else {
-                    LOGGER.errorf("Client scope not found for mapped credential configuration: %s", config.getScope());
-                    throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
-                }
-            } else {
+            // Find the credential configuration in the Issuer's metadata
+            //
+            var credConfig = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session).get(credConfigId);
+            if (credConfig == null) {
                 LOGGER.errorf("Mapped credential configuration not found: %s", credConfigId);
                 throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
             }
+
+            // Verify the user login session
+            //
+            var userSession = authResult.session();
+            var userModel = userSession.getUser();
+            if (!userModel.getUsername().equals(offerState.getUserId())) {
+                LOGGER.errorf("Unexpected login user: %s != %s", userModel.getUsername(), offerState.getUserId());
+                throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST));
+            }
+
+            // Verify the login client
+            //
+            ClientModel clientModel = session.getContext().getClient();
+            if (offerState.getClientId() != null && !clientModel.getClientId().equals(offerState.getClientId())) {
+                LOGGER.errorf("Unexpected login client: %s != %s", clientModel.getClientId(), offerState.getClientId());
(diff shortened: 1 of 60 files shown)


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -434,19 +517,37 @@ public Response getCredentialOfferPreflight(@PathParam("sessionCode") String ses
...
-                .entity(credentialsOffer));
+        var offerStorage = session.getProvider(CredentialOfferStorage.class);
+        var offerEntry = offerStorage.findOfferStateByNonce(nonce);
+        if (offerEntry == null) {
+            var errorMessage = "No offer entry for nonce: " + nonce;
+            eventBuilder.event(INTROSPECT_TOKEN_ERROR).detail(Details.REASON, errorMessage).error(Errors.INVALID_REQUEST);
+            throw new BadRequestException(getErrorResponse(ErrorType.INVALID_TOKEN));
+        }
+
+        var credOffer = offerEntry.credentialsOffer;
```
Captain-P-Goldfish: I am not sure how the other developers think. But I actually like strong typing and not using the `var`-keyword.  It blurs the readability of the code in my opinion.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorage.java
```diff
@@ -0,0 +1,41 @@
+package org.keycloak.protocol.oid4vc.issuance.credentialoffer;
+
+import java.util.HashSet;
+import java.util.Set;
+
+class InMemoryCredentialOfferStorage implements CredentialOfferStorage {
+
+    private final Set<CredentialOfferState> offerStorage = new HashSet<>();
```
Captain-P-Goldfish: MemoryLeak.
You never call the `removeOfferState` method and this map will start to fill itself indefinitely.
Rather use `keycloaksession.singleUseObjects`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +324,159 @@ public Response getCredentialOfferURIPreflight() {
...
+     * | no       | no        | no      | yes     | Generic offer; any logged-in user may redeem.       |
+     * | no       | no        | yes     | yes     | Offer restricted to a specific user.                |
+     * | no       | yes       | no      | yes     | Bound to client; user determined at login.          |
+     * | no       | yes       | yes     | yes     | Bound to both client and user.                      |
+     * +----------+-----------+---------+---------+-----------------------------------------------------+
+     * | yes      | no        | no      | no      | Pre-auth requires a user subject; missing userId.   |
+     * | yes      | yes       | no      | no      | Same as above; userId required.                     |
+     * | yes      | no        | yes     | yes     | Pre-auth for a specific user; client unconstrained. |
+     * | yes      | yes       | yes     | yes     | Fully constrained: user + client.                   |
+     * +----------+-----------+---------+---------+-----------------------------------------------------+
```
Captain-P-Goldfish, line 356: a little note that this matrix is referencing to the request-params would be helpful.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +324,159 @@ public Response getCredentialOfferURIPreflight() {
...
+        var hasCredentialOfferRole = userModel.getRoleMappingsStream().anyMatch(rm -> rm.getName().equals(CREDENTIAL_OFFER_CREATE.getName()));
+        if (!hasCredentialOfferRole) {
+            var errorMessage = "Credential offer creation requires role: " + CREDENTIAL_OFFER_CREATE.getName();
+            throw new CorsErrorResponseException(cors,
+                    ErrorType.INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
+        }
+
+        // Check whether the credential configuration exists in the issuer metadata
+        //
         Map<String, SupportedCredentialConfiguration> credentialsMap = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session);
```
Captain-P-Goldfish: since you are not using this map for anything else it would probably be much more performant to use `keycloaksession.realms().getClientScopesByAttributes(realm, Map.of(CredentialScopeModel.CONFIGURATION_ID, credConfigId), false)`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -569,7 +670,10 @@ public Response requestCredential(String requestPayload) {
 
             // First try to get the client session and look for the mapping there
             UserSessionModel userSession = authResult.session();
-            AuthenticatedClientSessionModel clientSession = userSession.getAuthenticatedClientSessionByClient(authResult.client().getId());
+            String userId = userSession.getUser().getUsername();
+            ClientModel clientModel = authResult.client();
+            String clientId = clientModel.getClientId();
+            AuthenticatedClientSessionModel clientSession = userSession.getAuthenticatedClientSessionByClient(clientModel.getId());
```
Captain-P-Goldfish: userId and clientId are never used here

Captain-P-Goldfish (changes requested):
I am not yet finished with the review but I do not have time anymore ^^°

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: For me it reads better this way :-) but never mind I can do both if preferred.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: Done

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorage.java
tdiesler: remove is called in 

* PreAuthorizedCodeGrantType.process(...)
* OID4VCIssuerEndpoint.getCredentialOffer(...)

In a follow up issue, we need to look at how the credential request can enforce the credential offer constraints (i.e. the offer state must out live the above)

The spec also mentions that a credential offer can be re-used multiple times to fetch the credential. To cover that scenario, we would need some sort of expiry reaper that removes expired offers. This is also another issue, I'd like to look at.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: There is RealmCacheSession.getClientScopesByAttributes() but not RealmProvider.getClientScopesByAttributes()

Performance is probably not an issue, for this manual admin task. If there is a better way to check whether a given configId exists, we can of course use that.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: In the credential-offer endpoint we now have this comment

> We treat the credential offer URI as an unprotected capability URL and rely solely on the later authorization step
> i.e. an authenticated client/user session is not required nor checked against the offer state

As a result, anyone who gets hold of the offer uri (qr code) can fetch the credential offer, but will later fail to retrieve  the actual credential because that would be protected by the offer state.

tdiesler:
@Captain-P-Goldfish Thanks Pascal, for reviewing this

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Captain-P-Goldfish: ups my mistake.
It is: `keycloakSession.clientScopes().getClientScopesByAttributes(...)`
But I agree that it will probably not be necessary since we have an authentication and a new authorization check. But it could become necessary in a future version if a lot of requests are trying to generate QR-Codes. We ourselves will maybe have potential usecases that might profit here from the direct search.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorage.java
Captain-P-Goldfish: Before I pulled just again my IntelliJ claimed that this method is never called :-)

But it is still a potential MemoryLeak:
* The `addOffer` method is called in `getCredentialOfferURI`
* The `removeOffer` method is called in  `getCredentialOffer` and `PreAuthorizedCodeGrantType`.

So it might happen very very slowly. But for long running Keycloak instances there is a potential threat. I would really recommend to use `keycloakSession.singleUseObjects()`. In this case it will also work in a Keycloak cluster.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Captain-P-Goldfish: I have troubles following you.
What I meant was simply this:
grafik
unused variables.

testsuite/integration-arquillian/servers/auth-server/services/testsuite-providers/src/main/java/org/keycloak/testsuite/rest/TestingResourceProvider.java
```diff
@@ -1121,18 +1126,30 @@ public Response getBlankPage() {
...
         RealmModel realm = getRealmByName(realmName);
-        AuthenticatedClientSessionModel ascm = session.sessions()
-                .getUserSession(realm, userSessionId)
+        UserSessionModel userSession = session.sessions().getUserSession(realm, userSessionId);
+        AuthenticatedClientSessionModel ascm = userSession
                 .getAuthenticatedClientSessions()
                 .values()
                 .stream().filter(acsm -> acsm.getClient().getClientId().equals(clientId))
                 .findFirst()
                 .orElseThrow(() -> new RuntimeException("No authenticatedClientSession found."));
```
Captain-P-Goldfish: `AuthenticatedClientSessionModel ascm` is never used

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/CredentialOfferStorage.java
```diff
@@ -0,0 +1,56 @@
...
+        public final String userId;
+        public final String nonce;
+        public final int expiration;
+
+        public CredentialOfferState(CredentialsOffer credOffer, String clientId, String userId, int expiration) {
+            this.credentialsOffer = credOffer;
+            this.clientId = clientId;
+            this.userId = userId;
+            this.expiration = expiration;
+            this.nonce = UUID.randomUUID() + "." + UUID.randomUUID();
```
Captain-P-Goldfish: why two UUIDs? is one not enough? if you just want to increase the entropy you can also use: `RandomSecret.createRandomSecret(50)`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +325,160 @@ public Response getCredentialOfferURIPreflight() {
...
+     */
+    public Response getCredentialOfferURI(String credConfigId, boolean preAuthorized, String targetUser) {
+        return getCredentialOfferURI(credConfigId, preAuthorized, null, targetUser, OfferUriType.URI, 0, 0);
+    }
+
+    /**
+     * Creates a Credential Offer Uri that can be pre-authorized and hence bound to a specific client/user id.
+     * <p>
+     * Credential Offer Validity Matrix for preAuthorized, appClientId, appUserId combinations.
+     * <p>
```
Captain-P-Goldfish: two opening paragraphs `<p>`

Also reading it for the first-time I would still not immediately get that the combinations refer to the request-parameters. I just want to make life easier for the next one to read it :-)

Maybe:
```html
<p>
  Credential Offer Validity Matrix for the supported request parameters "pre_authorized", "client_id", "user_id" combinations.
</p>
```
?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/CredentialOfferStorage.java
```diff
@@ -0,0 +1,56 @@
...
+
+public interface CredentialOfferStorage extends Provider {
+
+    class CredentialOfferState {
+
+        public final CredentialsOffer credentialsOffer;
+        public final String clientId;
+        public final String userId;
+        public final String nonce;
+        public final int expiration;
```
Captain-P-Goldfish: Making these public and not defining getter-methods somehow conflicts with the rest of the Keycloak implementation that uses getters and setters for members.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: I'm sorry - this is the 'credential' endpoint and yes, those should not be there (just yet)

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -524,46 +529,78 @@ public void testCompleteFlowWithCredentialOfferBasedAuthorizationDetails() throw
...
-            OID4VCAuthorizationDetailsResponse authDetailResponse = authDetailsResponse.get(0);
+            authDetailResponse = authDetailsResponse.get(0);
             assertNotNull("Credential identifiers should be present", authDetailResponse.getCredentialIdentifiers());
             assertEquals(1, authDetailResponse.getCredentialIdentifiers().size());
 
             credentialIdentifier = authDetailResponse.getCredentialIdentifiers().get(0);
             assertNotNull("Credential identifier should not be null", credentialIdentifier);
+
+            credentialConfigurationId = authDetailResponse.getCredentialConfigurationId();
+            assertNotNull("Credential configuration id should not be null", credentialConfigurationId);
```
Captain-P-Goldfish, line 540: When looking at the spec it says:

```
* credential_identifier: REQUIRED when an Authorization Details of type openid_credential was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the credential_configuration_id MUST NOT be present. 

* credential_configuration_id: REQUIRED if a credential_identifiers parameter was not returned from the Token Response as part of the authorization_details parameter. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the credential_configurations_supported Credential Issuer metadata. The corresponding object in the credential_configurations_supported map MUST contain one of the value(s) used in the scope parameter in the Authorization Request. When this parameter is used, the credential_identifier MUST NOT be present.
```

So in this specific case the correct check should rather be `assertNull`

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -524,46 +529,78 @@ public void testCompleteFlowWithCredentialOfferBasedAuthorizationDetails() throw
...
+
+            String requestBody = JsonSerialization.writeValueAsString(credentialRequest);
+            postCredential.setEntity(new StringEntity(requestBody, StandardCharsets.UTF_8));
+
+            try (CloseableHttpResponse credentialResponse = httpClient.execute(postCredential)) {
+                assertEquals(HttpStatus.SC_BAD_REQUEST, credentialResponse.getStatusLine().getStatusCode());
+                String responseBody = IOUtils.toString(credentialResponse.getEntity().getContent(), StandardCharsets.UTF_8);
+                assertTrue(responseBody.contains("UNKNOWN_CREDENTIAL_IDENTIFIER"));
+            }
+        }
```
Captain-P-Goldfish, line 584: spec chapter 3.3.4:

```
When the Wallet uses scope parameter in the Authorization Request, the scope value(s) are used to identify requested Credential Configurations. In this case, the Authorization Server has two options. If the Authorization Server supports returning credential_identifiers parameter in the Token Response, it MAY do so, in which case the Wallet uses those credential_identifier values in the Credential Request. If the Authorization Server does not support returning an authorization_details parameter containing the credential_identifiers parameter in the Token Response, the Wallet uses credential_configuration_id parameter in the Credential Request.
```

So if the `credential_identifier`  is returned in the TokenResponse the `credential_identifier` must be used in the CredentialRequest

Captain-P-Goldfish (changes requested):

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: How about this ...

```
// Check whether the credential configuration exists in available client scopes
//
var availableInClientScopes = session.clientScopes()
        .getClientScopesByProtocol(realmModel, OID4VC_PROTOCOL)
        .map(it -> it.getAttribute(CredentialScopeModel.CONFIGURATION_ID))
        .toList();
if (!availableInClientScopes.contains(credConfigId)) {
    var errorMessage = "Invalid credential configuration id: " + credConfigId;
    LOGGER.debugf("%s not found in supported credential config ids: %s", credConfigId, availableInClientScopes);
    throw new CorsErrorResponseException(cors,
            INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
}
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorage.java
tdiesler: ok, I'll look into that

tdiesler: done

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
tdiesler: This needs more work as the offer state would have to maintain the credential id. Could we postpone this until I look into enforcing the constraints in the 'credential' endpoint?

tdiesler: In the pre-auth offer case, the AccessTokenResponse does contain one AuthorizationDetail with both properties (i.e. credential_identifiers, credential_configuration_id). This is now natively supported in AccessTokenResponse.

The spec talks about what the wallet has to use in the CredentialRequest. So yes, it must use a  credential identifier in the pre-auth case - this is now enforced in the endpoint.

Using the credential_configuration_id in the CredentialRequest is still supported.
Using the credential scope as credential identifier is (imho) a hack that was used by many test cases - support for that has been removed and tests have been migrated to credential_configuration_id.

The credential endpoint now does this ...

```
        // When the CredentialRequest contains a credential identifier the caller must have gone through the
        // CredentialOffer process or otherwise have set up a valid CredentialOfferState

        if (credentialRequestVO.getCredentialIdentifier() != null) {
            var credId = credentialRequestVO.getCredentialIdentifier();

            // Retrieve the associated credential offer state
            //
            var offerStorage = session.getProvider(CredentialOfferStorage.class);
            var offerState = offerStorage.findOfferStateByCredentialId(session, credId);
            if (offerState == null) {
                var errorMessage = "No credential offer state for credential id: " + credId;
                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_IDENTIFIER, errorMessage));
            }

            var authDetails = offerState.getAuthorizationDetails();
            String credConfigId = (String) authDetails.getOtherClaims().get("credential_configuration_id");

            if (credConfigId == null) {
                var errorMessage = "Cannot credential configuration mapping for: " + credId;
                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_CONFIGURATION, errorMessage));
            }
```

In short, a credential identifier can (currently) only be obtained through a pre-auth credential offer.

The tests that have a credential identifier, now also use it.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
tdiesler: Our AccessTokenResponse does contain one AuthorizationDetail with both properties (i.e. credential_identifiers, credential_configuration_id). This is now natively supported in AccessTokenResponse.

The Wallet can then use either property in the CredentialRequest, but not both.

We currently have (unfortunately) no way of rejecting a CredentialRequest when the Wallet used a credential_configuration_id when instead they should have been using a credential identifier. In that case we would have a CredentialOfferState available but we don't know how to associate that with CredentialRequest.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
mposolda: +1 , There is not using of `var` inside Keycloak codebase. It is not strictly restricted, but I think it would be in the near future as AFAIK there might be some more checks for the code format introduced.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
Captain-P-Goldfish: looks good to me

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
Captain-P-Goldfish: From my point of view this seems actually to be a good way to handle this. I just noticed this when reading the unit-test ^^

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
Captain-P-Goldfish: I have basically no objections. This will probably also make the endpoints more fail-proof for some wallet-implementations.

Captain-P-Goldfish:
by the way. It would be easire for us reviewers if you would add your changes in a new commit instead of amending and force-pushing it. This branch will be squashed anyway before it is merged.

tdiesler:
@Captain-P-Goldfish good to know, will do

tdiesler:
Is there still anything for me to do here? (e.g. do you want me to replace the var ... with explicit types)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -691,33 +691,51 @@ public Response requestCredential(String requestPayload) {
...
+            var credConfig = OID4VCIssuerWellKnownProvider.getSupportedCredentials(session).get(credConfigId);
+            if (credConfig == null) {
                 LOGGER.errorf("Mapped credential configuration not found: %s", credConfigId);
                 throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_CONFIGURATION));
             }
+
+            // Verify the user login session
+            //
+            var userSession = authResult.session();
+            var userModel = userSession.getUser();
```
Captain-P-Goldfish: Marek also gave his preference and a good reason for not using the `var` keyword. Marek said he expects validations in the future that will suppress the usage of the `var`-keyword in keycloak. So it will probably be best to remove it.

Captain-P-Goldfish (changes requested):

tdiesler:
It seems that "var ..." definitions are used all over the place already

```
$ git grep "var " | grep '.java' | wc -l
    1843
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: done

tdiesler:
imho, its failing unrelated

```
Error: OR] Failures: 
Error: OR] org.keycloak.testsuite.oauth.hok.HoKTest.postLogoutByHoKRefreshTokenWithClientCertificate
Error: OR]   Run 1: HoKTest.postLogoutByHoKRefreshTokenWithClientCertificate:575
Error: OR]   Run 2: HoKTest.postLogoutByHoKRefreshTokenWithClientCertificate:575
Error: OR]   Run 3: HoKTest.postLogoutByHoKRefreshTokenWithClientCertificate:575
```

Captain-P-Goldfish:
there are still test failures

tdiesler:
I don't think they are mine

Captain-P-Goldfish:
> I don't think they are mine

looks reasonable. Did not check before which tests failed. Just did not expect. In that case. Approve.

Captain-P-Goldfish (approved):

tdiesler:
Is there a way to trigger re-testing other than rebase and forced push?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorage.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public CredentialOfferState findOfferStateByCredentialId(KeycloakSession session, String credId) {
+        if (session.singleUseObjects().contains(credId)) {
+            String entryJson = session.singleUseObjects().get(credId).get(ENTRY_KEY);
+            return JsonSerialization.valueFromString(entryJson, CredentialOfferState.class);
+        }
+        return null;
+    }
```
IngridPuppet, line 68: Do these non-state-mutating methods really benefit from using the `synchronized` keyword? If you mean it, then you probably also meant to add it to the last method.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorageFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
+
+public class InMemoryCredentialOfferStorageFactory implements CredentialOfferStorageFactory {
+
+    private static CredentialOfferStorage INSTANCE;
+
+    @Override
+    public CredentialOfferStorage create(KeycloakSession session) {
+        if (INSTANCE==null) {
```
IngridPuppet: Would you mind auto-formatting the file?

```suggestion
        if (INSTANCE == null) {
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +324,164 @@ public Response getCredentialOfferURIPreflight() {
     }
 
     /**
-     * Provides the URI to the OID4VCI compliant credentials offer
+     * Creates a Credential Offer Uri that is not bound to the calling target user.
```
IngridPuppet: According to the implementation, did you mean the following instead instead?
```suggestion
     * Creates a Credential Offer Uri that is bound to the calling target user.
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +324,164 @@ public Response getCredentialOfferURIPreflight() {
...
-                    ErrorType.INVALID_CREDENTIAL_REQUEST.toString(),
-                    "Invalid credential configuration ID",
-                    Response.Status.BAD_REQUEST);
+        // Check required role to create a credential offer
+        //
+        boolean hasCredentialOfferRole = userModel.getRoleMappingsStream().anyMatch(rm -> rm.getName().equals(CREDENTIAL_OFFER_CREATE.getName()));
+        if (!hasCredentialOfferRole) {
+            var errorMessage = "Credential offer creation requires role: " + CREDENTIAL_OFFER_CREATE.getName();
+            throw new CorsErrorResponseException(cors,
+                    INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
```
IngridPuppet: A 403 appears more suitable.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +324,164 @@ public Response getCredentialOfferURIPreflight() {
...
+        if (preAuthorized) {
+            if (appClientId == null) {
+                appClientId = clientModel.getClientId();
+                LOGGER.warnf("Using fallback client id for credential offer: %s", appClientId);
+            }
+            if (appUserId == null) {
+                var errorMessage = "Pre-Authorized credential offer requires a target user";
+                throw new CorsErrorResponseException(cors,
+                        INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
+            }
```
IngridPuppet, line 423: Can we do this check even before looking up the matching credential scope for performance?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -318,82 +324,164 @@ public Response getCredentialOfferURIPreflight() {
...
-                    "Failed to process credential offer",
-                    Response.Status.BAD_REQUEST);
+                .setCredentialConfigurationIds(List.of(credConfigId));
+
+        int expiration = timeProvider.currentTimeSeconds() + preAuthorizedCodeLifeSpan;
+        CredentialOfferState offerState = new CredentialOfferState(credOffer, appClientId, appUserId, expiration);
+
+        if (preAuthorized) {
+            String code = "urn:oid4vci:code:" + UUID.randomUUID();
+            credOffer.setGrants(new PreAuthorizedGrant().setPreAuthorizedCode(
```
IngridPuppet: Because sensitive, could we use a secure random code with higher entropy for the pre-authorized code? Using class `SecretGenerator` for example?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -563,58 +677,66 @@ public Response requestCredential(String requestPayload) {
...
+            CredentialOfferState offerState = offerStorage.findOfferStateByCredentialId(session, credId);
+            if (offerState == null) {
+                var errorMessage = "No credential offer state for credential id: " + credId;
+                throw new BadRequestException(getErrorResponse(UNKNOWN_CREDENTIAL_IDENTIFIER, errorMessage));
+            }
+
+            // Get the credential_configuration_id from AuthorizationDetails
+            //
+            AuthorizationDetailsResponse authDetails = offerState.getAuthorizationDetails();
+            String credConfigId = (String) authDetails.getOtherClaims().get("credential_configuration_id");
```
IngridPuppet: Introduce a constant to avoid the magic constant here? Also, could you check if you could type the details as `OID4VCAuthorizationDetailsResponse` directly? The model already exists.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -563,58 +677,66 @@ public Response requestCredential(String requestPayload) {
...
-
-            if (clientSession != null) {
-                mappedCredentialConfigurationId = clientSession.getNote(mappingKey);
-                if (mappedCredentialConfigurationId != null) {
-                    LOGGER.debugf("Found credential configuration ID mapping in client session for identifier %s: %s",
-                            credentialRequestVO.getCredentialIdentifier(), mappedCredentialConfigurationId);
-                }
+            UserModel userModel = userSession.getUser();
+            if (!userModel.getUsername().equals(offerState.getUserId())) {
+                LOGGER.errorf("Unexpected login user: %s != %s", userModel.getUsername(), offerState.getUserId());
```
IngridPuppet: Did you mean to run this check only when `offerState.getUserId()` is not null? My understanding is that Authorization Code flows may not have the user constrained in the credential offer.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -563,58 +677,66 @@ public Response requestCredential(String requestPayload) {
...
-                } catch (Exception e) {
-                    LOGGER.errorf(e, "Failed to find credential scope for identifier: %s", credentialRequestVO.getCredentialIdentifier());
-                    throw new BadRequestException(getErrorResponse(ErrorType.UNKNOWN_CREDENTIAL_IDENTIFIER));
-                }
+            // Verify the login client
+            //
+            ClientModel clientModel = session.getContext().getClient();
+            if (offerState.getClientId() != null && !clientModel.getClientId().equals(offerState.getClientId())) {
+                LOGGER.errorf("Unexpected login client: %s != %s", clientModel.getClientId(), offerState.getClientId());
+                throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST));
```
IngridPuppet: Add a message as well to the exception?

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -63,70 +64,112 @@ public Response process(Context context) {
...
+                    "Code is expired", Response.Status.BAD_REQUEST);
         }
-        AuthenticatedClientSessionModel clientSession = result.getClientSession();
-        ClientSessionContext sessionContext = DefaultClientSessionContext.fromClientSessionAndScopeParameter(clientSession,
-                OAuth2Constants.SCOPE_OPENID, session);
+        var credOffer = offerState.getCredentialsOffer();
+
+        var appUserId = offerState.getUserId();
+        var userModel = session.users().getUserByUsername(realm, appUserId);
+        if (userModel == null ) {
```
IngridPuppet: Would you mind autoformatting your changes in the file?
```suggestion
        if (userModel == null) {
```

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -63,70 +64,112 @@ public Response process(Context context) {
...
+        }
+
+        var appClientId = offerState.getClientId();
+        ClientModel clientModel = realm.getClientByClientId(appClientId);
+        if (clientModel == null ) {
+            var errorMessage = "No client model for: " + appClientId;
+            event.detail(Details.REASON, errorMessage).error(Errors.INVALID_CODE);
+            throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_REQUEST,
+                    errorMessage, Response.Status.BAD_REQUEST);
+        }
```
IngridPuppet, line 106: Do you mean to add this check? The matrix in the description of the PR makes Pre-Authorized Code flows possible without a constrained client.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
```diff
@@ -667,28 +721,34 @@ public void testCompleteFlowWithClaimsValidation() throws Exception {
...
 
-        // Step 2: Request the actual credential using the identifier
+        // Step 2: Request the actual credential using the identifier and config id
         HttpPost postCredential = new HttpPost(ctx.credentialIssuer.getCredentialEndpoint());
         postCredential.addHeader(HttpHeaders.AUTHORIZATION, "Bearer " + token);
         postCredential.addHeader(HttpHeaders.CONTENT_TYPE, "application/json");
 
         CredentialRequest credentialRequest = new CredentialRequest();
         credentialRequest.setCredentialIdentifier(credentialIdentifier);
+        credentialRequest.setCredentialConfigurationId(credentialConfigurationId);
```
IngridPuppet: Setting both fields likely does not align with the spec.

```
credential_identifier: [...] When this parameter is used, the credential_configuration_id MUST NOT be present.
```

Any counter-arguments?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/OID4VCICredentialOfferMatrixTest.java
```diff
@@ -0,0 +1,460 @@
+/*
+ * Copyright 2024 Red Hat, Inc. and/or its affiliates
```
IngridPuppet:
```suggestion
 * Copyright 2025 Red Hat, Inc. and/or its affiliates
```

IngridPuppet (commented):
Hello @tdiesler - Thank you for the substantial update to the OpenID4VCI flow. I checked the PR and left a few comments for consideration. Please could you check them? 

My main concern would be that support for retrieving credentials with only the `credential_identifier` parameter at the Credential Endpoint was seemingly dropped, judging from the changes in tests.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/InMemoryCredentialOfferStorage.java
tdiesler: No, that's a leftover from a previous impl - thanks

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: Yes, it should read: "Creates a Credential Offer Uri that is bound to the calling user"

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: ok

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: done

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: done

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: done, reviewed others

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: We do this on CredentialOfferState creation. So yes, a pre-auth offer must always be targeted to a user

```
        if (preAuthorized) {
            if (appClientId == null) {
                appClientId = clientModel.getClientId();
                LOGGER.warnf("Using fallback client id for credential offer: %s", appClientId);
            }
            if (appUserId == null) {
                var errorMessage = "Pre-Authorized credential offer requires a target user";
                throw new CorsErrorResponseException(cors,
                        INVALID_CREDENTIAL_OFFER_REQUEST.toString(), errorMessage, Response.Status.BAD_REQUEST);
            }
        }
```

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
tdiesler: On creation of the CredentialOfferState, we fall back to the caller clientId. 
i.e. The user with the "credential-offer-create" role pre-authorizes offers to be used with the same client_id as herself.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: done

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java
tdiesler: removed

tdiesler:
@IngridPuppet thank you so much for your detailed review - great suggestions, comments and lots of TLC for detail :-)

> My main concern would be that support for retrieving credentials with only the `credential_identifier` parameter at the Credential Endpoint was seemingly dropped, judging from the changes in tests.

Using the credential scope as credential identifier is (imho) a hack that was used by many test cases - support for that has been removed and tests have been migrated to credential_configuration_id.

Requesting a credential with only `credential_identifier` is actually the preferred way of doing things. We have a comment like this in the OID4VCIssuerEndpoint endpoint ...

```
        // When the CredentialRequest contains a credential identifier the caller must have gone through the
        // CredentialOffer process or otherwise have set up a valid CredentialOfferState
```

IngridPuppet (approved):
Thank you for addressing my concerns and providing clarification. The PR looks good to me.

testsuite/integration-arquillian/tests/base/pom.xml
```diff
@@ -93,6 +93,11 @@
             <artifactId>junit</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
```
mposolda: Can you please rather remove this dependency and use the same `org.junit.Assert` stuff used by other tests? It's maybe ok, but I don't see any reason of including this and not 100% sure about the side-effect of this.

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -63,70 +65,112 @@ public Response process(Context context) {
...
+        var appClientId = offerState.getClientId();
+        ClientModel clientModel = realm.getClientByClientId(appClientId);
+        if (clientModel == null) {
+            var errorMessage = "No client model for: " + appClientId;
+            event.detail(Details.REASON, errorMessage).error(Errors.INVALID_CODE);
+            throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_REQUEST,
+                    errorMessage, Response.Status.BAD_REQUEST);
+        }
+
+        UserSessionModel userSession = session.sessions().createUserSession(null, realm, userModel, userModel.getUsername(),
```
mposolda, line 108: Question: Is it strictly required to create new userSession (and new clientSession) for this use-case? For example for some similar use-cases (like for example OAuth2 client credentials grant implemented in `ClientCredentialsGrantType`), we are using "transient" user session.

This means that tokens are not attached to the real userSession, which is persisted into the DB. But tokens can be still used to be sent to Keycloak endpoints (like introspection-endpoint or admin REST API or anything else. I believe that oid4VCI credentials endpoint should work as well). 

In short: Real user session is needed just if token refreshes are supposed to work. But it seems that tokens issued from "pre-authorized" grant doesn't need to be refreshed?

mposolda (commented):
@tdiesler Thanks for the PR and sorry for the late review.

@IngridPuppet @Captain-P-Goldfish Thanks for your reviews!

I've added 2 inline comments. Can you please check them? Also it seems that `HoKTest` failure might be a regression of your PR as it doesn't fail for any other PRs sent to Keycloak main. In your PR, you did some changes unrelated to `oid4vci` (like EG. marking one of the clients to have direct-grants enabled by default or renaming endpoints on `TestingResource`) and in theory, some of these can cause the test failure. Maybe just rebase on top of latest keycloak `main` helps.

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
tdiesler: The client exchanges the pre-auth code for an AccessToken it then calls the /credential endpoint with that AccessToken. For the TRANSIENT case authentication fails with invalid token here ...

```
                UserSessionUtil.UserSessionValidationResult validationResult = UserSessionUtil.findValidSessionForAccessToken(session, realm, token, client, invalidUserSessionCallback);
                if (validationResult.getError() != null) {
                    return null;
                }
```

It would be expected to have no valid user session, because ...

* the Issuer created the the Pre-Auth CredentialOffer for a given target user (which might not have a user session)
* the target user then exchanges the pre-auth code for an AccessToken (i.e. no other login required)
* the issuer is expected to provide the pre-auth credential without the user ever needing to login

For clarity, I added the AuthorizationDetails to the pre-auth AccessToken as well (and not just to the AccessTokenResponse)

I left the the persistent user session in place (for now)

```
        UserSessionModel userSession = session.sessions().createUserSession(null, realm, userModel, userModel.getUsername(),
                null, "pre-authorized-code", false, null,
                null, UserSessionModel.SessionPersistenceState.PERSISTENT);
```

Perhaps there is a way to work with a TRANSIENT user session. This could then be an improvement for  later imho.

tdiesler:
retest this please

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
mposolda: @tdiesler Yes, I think we can make it working with userSession not being created, but agree that this could be a follow-up. Created https://github.com/keycloak/keycloak/issues/44534 for that. Will try to take a look when I have some time.

Thanks again for all the work on the PR!

mposolda (approved):

PR #44258, opened 2025-11-17
Fix for WebAuthnSigningInTest WebAuthn test

### LINKED ISSUE #43477: Flaky test: org.keycloak.testsuite.webauthn.account.WebAuthnSigningInTest#setUpLinksTest
## org.keycloak.testsuite.webauthn.account.WebAuthnSigningInTest#setUpLinksTest
[Keycloak CI - WebAuthn IT (firefox)](https://github.com/keycloak/keycloak/actions/runs/18513736372/job/52760258937)

```
java.lang.AssertionError: 
Set up link for "webauthn-passwordless" is not visible
Expected: is <true>
     but: was <false>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
...
```

rmartinc:
Closes #43477

In `firefox` the test was flaky (not in chrome). There are several improvements, although I think that the first one is the main fix:

1. Required actions for webauthn are searched and disabled/enabled instead of creating a new one. This was the main problem for the tests, it seems that two required actions were present with the same ID and that created issues sometimes.
2. The refresh page is replaced by clicking to another account page and click back to the signing in page (device activity and then back to signing in). The problem with the refresh is that all the oidc dance is performed (redirects) and it's much more problematic.
3. In the method `navigateBeforeTest` I'm waiting to the page to be fully translated. Very rarely I detected the test failed because the page was loaded but the texts were still not translated (I suppose waiting for the messages to be retrieved and applied).
4.  The `sidebar` is removed because it's outdated (it didn't work for me and I need to re-implement the `navigateToUsingSidebar`). So I deleted this obsolete code.

The full `firefox` WebAuthn CI ran OK 100 times (10x10). See https://github.com/rmartinc/keycloak/actions/runs/19422528841.


mposolda (approved):
@rmartinc Thanks!

PR #44259, opened 2025-11-17
In ClientRegistrationAuth, EventBuilder#error invocations fail with IllegalStateException

reda-alaoui:
Closes #44263

Invoking `EventBuilder#error` before `EventBuilder#event` leads to an `IllegalStateException` because of:

https://github.com/keycloak/keycloak/blob/22deeda2c09c55a60979f9a1bc6cc55c0739a0c7/server-spi-private/src/main/java/org/keycloak/events/EventBuilder.java#L227-L230

Because of that, each error happening in `ClientRegistrationAuth`:
- ends up with an HTTP status code 500
- is never traced in the event log

This PR prepends every invocation of `EventBuilder#event` in `ClientRegistrationAuth` with an invocation to `EventBuilder#event`.

### DIFF
--- a/services/src/main/java/org/keycloak/services/clientregistration/ClientRegistrationAuth.java
+++ b/services/src/main/java/org/keycloak/services/clientregistration/ClientRegistrationAuth.java
@@ -34,6 +34,7 @@ import org.keycloak.common.util.Time;
 import org.keycloak.events.Details;
 import org.keycloak.events.Errors;
 import org.keycloak.events.EventBuilder;
+import org.keycloak.events.EventType;
 import org.keycloak.models.AdminRoles;
 import org.keycloak.models.ClientInitialAccessModel;
 import org.keycloak.models.ClientModel;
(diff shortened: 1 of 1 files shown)


ahus1 (changes requested):
Thank you for submitting this PR. Please first create a GitHub issue and connect it to this PR as outlined in the `CONTRIBUTING.md` file.

reda-alaoui:
@ahus1 issue linked

ahus1 (changes requested):
For the error to be created correctly, it is sufficient if `event.event()` is called with the regular event, not the error, any time before `event.error()` is called. 

In the example of the added `event.event(EventType.CLIENT_REGISTER_ERROR);`, the method `requireCreate()` is called from `AbstractClientRegistrationProvider.create()` which already calls `event.event(EventType.CLIENT_REGISTER)`. 

The places where a missing `event.event()` exists might be revealed when you share the stack trace of the 500 error.

Please add a full stack trace in the parent issue to show what is happening in your scenario.

reda-alaoui:
> For the error to be created correctly, it is sufficient if `event.event()` is called with the regular event, not the error, any time before `event.error()` is called.
> 
> In the example of the added `event.event(EventType.CLIENT_REGISTER_ERROR);`, the method `requireCreate()` is called from `AbstractClientRegistrationProvider.create()` which already calls `event.event(EventType.CLIENT_REGISTER)`.
> 
> The places where a missing `event.event()` exists might be revealed when you share the stack trace of the 500 error.
> 
> Please add a full stack trace in the parent issue to show what is happening in your scenario.

@ahus1 Ok I see. The issue happens when I invoke `requireView` from my Keycloak plugin in class extending AbstractClientRegistrationProvider. So I guess, my plugin is faulty.
Sorry for all the noise.

PR #44271, opened 2025-11-17
Prevent slash duplication in request URLs

### LINKED ISSUE #44269: Admin Client creates malformed paths for requests
### Before reporting an issue

### Area

admin/client-js

### Describe the bug

When the admin client is constructed with a URL that contains a trailing slash, paths are joined with multiple slashes, causing failures. This is compounded by recent changes that enforce URLs do not contain multiple slashes (see #43765)

### Version

main

### Regression

### Expected behavior

The admin client does not add double slashes to base URLs that already end in a slash.

### Actual behavior

The admin client adds double slashes to base URLs that already end in a slash.

### How to Reproduce?

```js
const adminClient = new KeycloakAdminClient({
  baseUrl: "http://localhost:8080/"
});
```

### Anything else?

_No response_

jonkoops:
Prevents the admin client from duplicating slashes in request URLs, which can make the request URL invalid.

Closes #44269

### DIFF
--- a/js/libs/keycloak-admin-client/package.json
+++ b/js/libs/keycloak-admin-client/package.json
@@ -40,7 +40,6 @@
   },
   "dependencies": {
     "camelize-ts": "^3.0.0",
-    "url-join": "^5.0.0",
     "url-template": "^3.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


rmartinc (approved):
Thanks @jonkoops! LGTM! This does not fix if the user passes some weird URL like `https.//localhost//auth`, but I don't think we should really fix that URL. So more than enough for me.

jonkoops:
Agreed, this is more about simple path joining, which the admin client should be able to handle, if the user provides a malformed URL, that is 'garbage in, garbage out', and in their realm of responsibility.

@mposolda can I ask you for a merge?

mposolda:
@jonkoops @rmartinc Thanks for the PR and review!

mposolda (approved):

PR #44275, opened 2025-11-17
Make sure LDAP sync runs in a single cluster node and respecting the configured period

### LINKED ISSUE #43752: LDAP synchronization happens too often in a cluster and always clears the cache
### Before reporting an issue

### Area

ldap

### Describe the bug

When analyzing a production setup that had multiple LDAP federations configured, I noticed some scheduling problems and also some oddities that lead to situation where synchronizations didn't happen, or happens with a performance impact.

### Version

main

### Regression

### Expected behavior

Given a Keycloak cluster with 4 nodes:
 
* When setting a synchronization interval for changed users, it should synchronize once per that interval 
* When setting a synchronization interval for all users (for the same provider), it should synchronize once per that interval
* The realm cache should not be cleared at the end of the synchronization

### Actual behavior

Given a Keycloak cluster with 4 nodes:
 
* I've seen a lot more synchronizations happening. It seems that any scheduling doesn't see previous runs of other nodes, as it is holding on to the instance of the UserStorageProviderModel in https://github.com/keycloak/keycloak/blob/4dc4de7c1218bd129b4ae6853a74b7a2954b6eec/model/storage-private/src/main/java/org/keycloak/storage/managers/UserStorageSyncManager.java#L244 - instead, it should look up the provider by its ID again to fetch the latest data to see synchronizations performed by other providers. Due to this, and due to the logic `timeSinceLastSync * 2 > period` the synchronization runs now four times (once per node) within the specified period.
* To do any calculation of the last sync time, there is only one timestamp to track both the full and the update synchronization. Assuming a update synchronization of every hour, and a full synchronization of every 24 hours, the full synchronization will never happen (to be verified). 
* When the synchronization is complete, it writes the updated time into the provider's component configuration. When doing so, this automatically triggers a clear of the full realm cache. I've seen the clearing of the realm cache in the logs. 
* Both the update and the full synchronization use the same lock mechanism. If a full synchronization would start during an update synchronization, the full synchronization would be skipped, which is not what users would expect. Instead, the system should check after an update synchronization if there is also a full synchronization due, and then execute it, so it is never lost. 

### How to Reproduce?

Reviewed in the customer's log files.

### Anything else?

This user would love to see a cron based scheduling so they could run the full sync during the night instead of running it on an offset after the server has started, which is quite arbitrary, and not a good fit for environments that dynamically scale up and down instances. 

When running with a CRON schedule, it might not be so important to store the latest time it has run in the database. Maybe storing the time in the work cache and doing the locking in the work cache might be sufficient. 

There might be some low-hanging fixes here, and other items might be moved to a new feature. While IMHO a lot of people would welcome the cron syntax, it would need to take some thought to prepare an automatic migration from the old configurations.

pedroigor:
Closes #43752


model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -1242,7 +1242,6 @@ public static ComponentModel convertFedProviderToComponent(String realmId, UserF
         model.setFullSyncPeriod(fedModel.getFullSyncPeriod());
         model.setPriority(fedModel.getPriority());
         model.setChangedSyncPeriod(fedModel.getChangedSyncPeriod());
-        model.setLastSync(fedModel.getLastSync());
```
ahus1, line 1245: I don't know when this is actually called ... is it called when converting some very old realm imports, or is it still in use?

ahus1 (commented):
Thank you for this PR! I reviewed it and also manually tested it. It looks good to me. 

One some question remains, see below.

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
pedroigor: I'm also not sure but it does not make much sense to me to set the last sync. It is an internal state.

ahus1 (approved):

byn-florian-posch:
FYI, the [sync mode change](https://github.com/keycloak/keycloak/pull/44275/changes#diff-016c561bf64875bf5bd2c485708d5c046b5e28c555a43d90df226d12a7f47e2a) got us some regression with high-volume user realms with external user providers in some of our environments.

* _lastSync_  is not evaluated anymore (now using lastSync_CHANGED/FULL)
* the next differential change (for example, create a user and do a "CHANGED" sync) will reload your full userbase from the provider

I'm not sure if this was intended, but imo, there should have been a database migration to write the current value of _lastSync_ in _lastSync_FULL_ (or _CHANGED)

byn-florian-posch:
> FYI, the [sync mode change](https://github.com/keycloak/keycloak/pull/44275/changes#diff-016c561bf64875bf5bd2c485708d5c046b5e28c555a43d90df226d12a7f47e2a) got us some regression with high-volume user realms with external user providers in some of our environments.
> 
> * _lastSync_  is not evaluated anymore (now using lastSync_CHANGED/FULL)
> * the next differential change (for example, create a user and do a "CHANGED" sync) will reload your full userbase from the provider
> 
> I'm not sure if this was intended, but imo, there should have been a database migration to write the current value of _lastSync_ in _lastSync_FULL_ (or _CHANGED)

Never mind, I found this in the [discussion](https://github.com/keycloak/keycloak/issues/46367#issuecomment-4007330888) of the actual change, so it is on the radar.

PR #44281, opened 2025-11-17
More accurate user session expiration logic

pruivo:
Closes #44204

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentUserSessionEntity.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentUserSessionEntity.java
@@ -67,22 +67,34 @@ import org.hibernate.annotations.DynamicUpdate;
                 " FROM PersistentClientSessionEntity clientSess INNER JOIN PersistentUserSessionEntity sess ON clientSess.userSessionId = sess.userSessionId AND sess.offline = clientSess.offline" +
                 " WHERE sess.offline = :offline AND sess.realmId = :realmId AND sess.lastSessionRefresh >= :lastSessionRefresh" +
                 " GROUP BY clientSess.clientId, clientSess.externalClientId, clientSess.clientStorageProvider"),
-        @NamedQuery(name = "findUserSessionAndDataWithNullRememberMe",
+        @NamedQuery(name = "findUserSessionAndDataWithNullRememberMeLastRefresh",
                 query = "SELECT sess.userSessionId, sess.data" +
                         " FROM PersistentUserSessionEntity sess" +
-                        " WHERE sess.realmId = :realmId AND sess.offline = '0' AND sess.rememberMe IS NULL AND (sess.lastSessionRefresh < :lastSessionRefresh OR sess.createdOn < :createdOn)"),
+                        " WHERE sess.realmId = :realmId AND sess.offline = '0' AND sess.rememberMe IS NULL AND sess.lastSessionRefresh < :lastSessionRefresh"),
+        @NamedQuery(name = "findUserSessionAndDataWithNullRememberMeCreatedOn",
+                query = "SELECT sess.userSessionId, sess.data" +
+                        " FROM PersistentUserSessionEntity sess" +
+                        " WHERE sess.realmId = :realmId AND sess.offline = '0' AND sess.rememberMe IS NULL AND sess.createdOn < :createdOn"),
         @NamedQuery(name = "updateUserSessionRememberMeColumn",
                 query = "UPDATE PersistentUserSessionEntity sess" +
                         " SET sess.rememberMe = :rememberMe" +
                         " WHERE sess.userSessionId IN (:userSessionIds)"),
-        @NamedQuery(name = "findExpiredOfflineUserSessions",
+        @NamedQuery(name = "findExpiredOfflineUserSessionsLastRefresh",
                 query = "SELECT sess.userSessionId, sess.userId" +
                         " FROM PersistentUserSessionEntity sess" +
-                        " WHERE sess.realmId = :realmId AND sess.offline = '1' AND (sess.lastSessionRefresh < :lastSessionRefresh OR sess.createdOn < :createdOn)"),
-        @NamedQuery(name = "findExpiredRegularUserSessions",
+                        " WHERE sess.realmId = :realmId AND sess.offline = '1' AND sess.lastSessionRefresh < :lastSessionRefresh"),
+        @NamedQuery(name = "findExpiredOfflineUserSessionsCreatedOn",
                 query = "SELECT sess.userSessionId, sess.userId" +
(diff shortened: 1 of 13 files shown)


model/jpa/src/main/java/org/keycloak/models/jpa/session/Util.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.models.jpa.session;
+
+import org.keycloak.models.session.PersistentAuthenticatedClientSessionAdapter;
+import org.keycloak.storage.StorageId;
+
+public final class Util {
```
ahus1: Could you please name it `JpaSessionUtil` as the current name is not very specific. Thanks!

model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentUserSessionEntity.java
```diff
@@ -81,7 +98,7 @@ public class PersistentUserSessionEntity {
     @Column(name = "REALM_ID", length = 36)
     protected String realmId;
 
-    @Column(name="USER_ID", length = 255)
+    @Column(name="USER_ID")
```
ahus1, line 116: Please restore the length attribute of the annotation.

model/jpa/src/main/java/org/keycloak/models/jpa/session/UserSessionExpirationLogic.java
```diff
@@ -0,0 +1,276 @@
...
+
+        long duration = System.nanoTime() - start;
+        logger.debugf("Regular user session expiration task completed for realm '%s'. Took %dms", realm.getName(), TimeUnit.NANOSECONDS.toMillis(duration));
+    }
+
+    /**
+     * Migrates the remember me flag into to its own column, for an efficient query.
+     * <p>
+     * It only affects regular user sessions since offline sessions do not have remember me, and only migrates sessions
+     * closer to the expiration time.
```
ahus1:
```suggestion
     * close to the expiration time to avoid concurrency issues on existing sessions.
```

model/jpa/src/main/java/org/keycloak/models/jpa/session/UserSessionIdAndRememberMe.java
```diff
@@ -0,0 +1,48 @@
...
+
+    static UserSessionIdAndRememberMe fromQueryProjection(Object[] projection) {
+        assert projection.length == 2;
+        assert projection[0] != null;
+        assert projection[1] != null;
+        try {
+            String id = String.valueOf(projection[0]);
+            String data = String.valueOf(projection[1]);
+            Map<?, ?> values = JsonSerialization.readValue(data, Map.class);
+            // TODO should we make PersistentUserSessionData public?
```
ahus1, line 40: Not sure here ... maybe it is OK to leave it as is, as we would want to have the migration code rather stable and self-contained?

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
```diff
@@ -97,13 +97,19 @@
...
             <column name="REALM_ID" type="VARCHAR(36)"/>
             <column name="OFFLINE_FLAG" type="VARCHAR(4)"/>
             <column name="REMEMBER_ME" type="BOOLEAN"/>
             <column name="LAST_SESSION_REFRESH" type="INT"/>
+            <column name="USER_SESSION_ID" type="VARCHAR(36)"/>
+            <column name="USER_ID" type="VARCHAR(255)"/>
         </createIndex>
+        <!-- replaced by the 2 indexes above -->
+        <dropIndex tableName="OFFLINE_USER_SESSION" indexName="IDX_OFFLINE_USS_BY_LAST_SESSION_REFRESH"/>
     </changeSet>
```
ahus1, line 137: Each of these DDLs is affecting a very large table. Sometimes the migration might time out, and then it is "stuck in the middle" as Liquibase will try to re-run it from the beginning. 

So a good practice is to split this changeset to one-DDL-per-changeset.

ahus1 (commented):
Thank you for this PR. Some pre-review comments below.

model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentUserSessionEntity.java
pruivo: The default length is 255.

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
pruivo: If I understand your comment, you want 4 changesets?
1. add column
2. add index
3. add index
3. remove index

ahus1: Yes.

model/jpa/src/main/java/org/keycloak/models/jpa/session/PersistentUserSessionEntity.java
ahus1: OK, thanks, then keep it as is. Sorry for the noise.

pruivo:
rebased to fix the build :crossed_fingers:

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
```diff
@@ -89,21 +89,51 @@
...
             <column name="REMEMBER_ME" type="BOOLEAN"/>
             <column name="LAST_SESSION_REFRESH" type="INT"/>
+            <column name="USER_SESSION_ID" type="VARCHAR(36)"/>
+            <column name="USER_ID" type="VARCHAR(255)"/>
         </createIndex>
     </changeSet>
+    <changeSet id="26.5.0-drop-sess-refresh-idx" author="keycloak">
+        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
+            <indexExists tableName="OFFLINE_USER_SESSION" indexName="IDX_OFFLINE_USS_BY_LAST_SESSION_REFRESH"/>
+        </preConditions>
```
ahus1, line 134: The pre-conditions to check if an index exists are important as an index might not be created due to the number of entries in the table. 

For the other changesets, this is not necessary. At the same time, it will not hurt.

ahus1 (approved):
Thank you for this pull request! One more step towards efficient cleaning of sessions!

model/jpa/src/main/resources/META-INF/jpa-changelog-26.5.0.xml
pruivo: I've seen it used in other changesets and I decided to copy the pattern :rofl:

PR #44298, opened 2025-11-18
Avoid un-escaped strings in the login templates for HTML entities

### LINKED ISSUE #44296: Avoid un-escaped strings in the login templates for HTML entities
### Description

There are several places where items are not escaped due to HTML entities in the messages. 

As we allow UTF-8 encoded messages for quite a while now, this should no longer be necessary. 

### Value Proposition

More secure-by-default themes. Also the remaining `?no_esc` should be simpler to review.

### Goals

Secure-by-default themes.

### Non-Goals

Changing other `no_esc` items.

### Discussion

_No response_

### Notes

I'll prepare a PR

ahus1:
Closes #44296


ahus1:
@keycloak/core-clients - as a hardening exercise, this removes some of the `?no_esc` from the template. Please let me know your thoughts. 

Consider this a first round of changes. Once we agree on these hardenings, there might be more. 

TL;DR: `no_esc` should be the exception, and simple to review.

rmartinc (approved):
For me this is OK. We are using other UTF-8 symbols already in the messages, so I don't think we need the HTML entities anymore. And I suppose this makes easier to remove the `no_esc` in the themes. It's OK to me. Approved.

pedroigor (dismissed):

pedroigor (approved):

PR #44307, opened 2025-11-18
Add support for cloudflare turnstile

SferaDev:
This PR adds native support for Cloudflare Turnstile as a CAPTCHA provider in Keycloak, providing a privacy-friendly alternative to reCAPTCHA. Turnstile can protect registration forms, login flows, and password reset pages with customizable themes and sizes.

Closes #44305

### New Authentication flows
- **RegistrationTurnstile**: Form action for protecting registration flows
- **UsernamePasswordFormWithTurnstile**: Authenticator for login forms with Turnstile protection
- **ResetCredentialChooseUserWithTurnstile**: Authenticator for password reset flows

### DIFF
--- a/docs/documentation/release_notes/topics/26_5_0.adoc
+++ b/docs/documentation/release_notes/topics/26_5_0.adoc
@@ -39,3 +39,11 @@ This provides a more fine-grained way to manage features and eliminates the need
 The `feature-<name>` option takes precedence over both `features` and `features-disabled`.
 
 For more details, see the https://www.keycloak.org/server/features[Enabling and disabling features] guide.
+
+= Cloudflare Turnstile support
+
+{project_name} has integration with Cloudflare Turnstile as a CAPTCHA provider.
+
+Turnstile can be configured to protect registration forms, login flows, and password reset pages.
+
+For more details, see the link:{adminguide_link}#proc-enabling-turnstile_server_administration_guide[{adminguide_name}].
--- a/docs/documentation/server_admin/topics/assembly-managing-users.adoc
+++ b/docs/documentation/server_admin/topics/assembly-managing-users.adoc
@@ -30,6 +30,7 @@ include::users/proc-allow-user-to-delete-account.adoc[leveloffset=+2]
 
 include::users/con-user-impersonation.adoc[leveloffset=+2]
 include::users/proc-enabling-recaptcha.adoc[leveloffset=+2]
+include::users/proc-enabling-turnstile.adoc[leveloffset=+2]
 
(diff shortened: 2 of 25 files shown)


stianst (changes requested):
We won't be able to accept this PR in its current shape I'm afraid. One thing is the duplication of authenticators.

What we really need is to refactor captcha support in Keycloak to be more pluggable by introducing a CaptchaProvider that can optionally be enabled/disabled in various forms.

Another issue is in terms of testing. We really need to have some automated testing to be able to accept Turnstile. 

There are quite a lot of captcha solutions, and everyone has their own favourite. So, in the end it may be better to provide a proper CaptchaSPI/Provider integration to make it simple for folks to integrate with what they want, or to provide extensions.

SferaDev:
Because keycloak doesn't natively support turnstile we were forced to use a [third-party plugin](https://github.com/panpaul/keycloak-turnstile) that has ended up being unmaintained and newer features broke the compatibility (ie. passwordless passkeys).

I understand the points for an abstraction and trying to accomodate more captcha solutions, I'll take a look to see how feasible it is for me to do the change.

I'd guess we would mark the existing reCAPTCHA and reCAPTCHA enterprise as deprecated (to be removed in v27) and recommend using the new abstraction from now on.

> Another issue is in terms of testing. We really need to have some automated testing to be able to accept Turnstile.

I agree, I added some integration tests for it and I was trying to make them pass in CI now, but I'll stop for now. The good thing is that they offer debug keys that don't require set-up. IIRC the existing reCAPTCHA implementation doesn't have a test suite.

stianst:
> Because keycloak doesn't natively support turnstile we were forced to use a [third-party plugin](https://github.com/panpaul/keycloak-turnstile) that has ended up being unmaintained and newer features broke the compatibility (ie. passwordless passkeys).
> 
> I understand the points for an abstraction and trying to accomodate more captcha solutions, I'll take a look to see how feasible it is for me to do the change.
> 
> I'd guess we would mark the existing reCAPTCHA and reCAPTCHA enterprise as deprecated (to be removed in v27) and recommend using the new abstraction from now on.

We shouldn't need to deprecate reCaptcha, but rather refactor it.

> 
> > Another issue is in terms of testing. We really need to have some automated testing to be able to accept Turnstile.
> 
> I agree, I added some integration tests for it and I was trying to make them pass in CI now, but I'll stop for now. The good thing is that they offer debug keys that don't require set-up. IIRC the existing reCAPTCHA implementation doesn't have a test suite.

The lack of tests around reCaptcha has been an issue quite a few times where it ended up being broken either by changes we did or Google did, but we didn't detect it until bugs where reported.

PR #44332, opened 2025-11-19
Operator Update Logic: add hash based comparison

### LINKED ISSUE #44280: Unsupported Pod template can cause AUTO logic to keep StatefulSet at 0 replicas
### Before reporting an issue

### Area

operator

### Describe the bug

Relates to #38846 - Env values for the existing statefulset and the desired state may not be identical if the apiversion is missing on some transitive resources like field references.

### Version

26.2

### Regression

### Expected behavior

Non-meaningful differences should not trigger upgrading behavior.

### Actual behavior

A missing apiVersion in the unsupported pod template will cause AUTO upgrading logic to continually be engaged - which can result in the StatefulSet being kept at 0 replicas.

### How to Reproduce?

Add an env entry to the unsupported pod template spec similar to:           
```
          - name: MY_NAME
            valueFrom:
              fieldRef:
                fieldPath: metadata.name
```
Along with having the update strategy set to AUTO.

### Anything else?

_No response_

pruivo:
Fixes #44280


shawkins:
@pruivo I tried to take a pass over things to simplify were possible.

pruivo:
@shawkins thanks, they look fine to me.

shawkins (approved):
@pruivo great, then I think we should be good. Thank you for the quick turn around on this.

shawkins:
@vmuzikar @ahus1 this should be a narrow enough set of changes to also backport to 26.4.

PR #44335, opened 2025-11-19
Delete operation for Client v2

### LINKED ISSUE #43291: Delete operation for Client v2
### Description

Implement delete operation.

Pepo48:
Closes: #43291

### DIFF
--- a/rest/admin-v2/api/src/main/java/org/keycloak/services/client/ClientService.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/services/client/ClientService.java
@@ -33,8 +33,6 @@ public interface ClientService extends Service {
 
     Stream<ClientRepresentation> getClients(ClientsResource clientsResource, RealmModel realm, ClientProjectionOptions projectionOptions, ClientSearchOptions searchOptions, ClientSortAndSliceOptions sortAndSliceOptions);
 
-    ClientRepresentation deleteClient(RealmModel realm, String clientId);
-
     Stream<ClientRepresentation> deleteClients(RealmModel realm, ClientSearchOptions searchOptions);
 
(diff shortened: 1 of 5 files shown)


shawkins (approved):
This looks good. Might be worth mentioning:

- we may eventually want to return something from the delete endpoint. Either the resource or a status if deferred deletions will be supported. 
- similarly we may eventually need parameter(s) on the delete operation.

Both of these could use the keycloak UID if it were exposed via the v2 representation.

cc @mabartos @vmuzikar

vmuzikar (approved):
LGTM, thank you.

vmuzikar:
> Either the resource

+1 but it's a potential follow-up.

> or a status if deferred deletions will be supported

What would be the use case to defer deletion?

> similarly we may eventually need parameter(s) on the delete operation

What would be the use case? Deletion based on querying?

shawkins:
> What would be the use case to defer deletion?

If we lack the concept of cascading delete, or it's not applicable, and there are resources that reference the client. Or you want to implement graceful termination of sessions using the client.

> What would be the use case? Deletion based on querying?

No, querying will be a clients level endpoint. General delete options in kubernetes include dry run, locking based upon resourceVersion or uid, and propogation control.

vmuzikar:
> If we lack the concept of cascading delete, or it's not applicable, and there are resources that reference the client. Or you want to implement graceful termination of sessions using the client.

I think all child resources needs to be deleted (or un-referenced) when a Client is deleted, deferred deletion does not sound to me like something Keycloak should do. That includes sessions.

> General delete options in kubernetes include dry run, locking based upon resourceVersion or uid, and propogation control.

That would be something more generic, not specific to deletion.

shawkins:
> That would be something more generic, not specific to deletion.

Those are all options on a common DeleteOption payload. For the other operations you can lock on resourceVersion by including that in the resource payload, locking on uid and propogation aren't applicable, and dry run has a different parameter.

PR #44345, opened 2025-11-19
Enhancement: normalize FilesPlaintextVaultProvider secret paths to pr…

Pepo48:
…event false positives in CSAs

### DIFF
--- a/services/src/main/java/org/keycloak/vault/FilesPlainTextVaultProvider.java
+++ b/services/src/main/java/org/keycloak/vault/FilesPlainTextVaultProvider.java
@@ -50,7 +50,7 @@ public class FilesPlainTextVaultProvider extends AbstractVaultProvider {
 
     @Override
     protected VaultRawSecret obtainSecretInternal(String vaultSecretId) {
-        Path secretPath = vaultPath.resolve(vaultSecretId);
+        Path secretPath = vaultPath.resolve(vaultSecretId).normalize();
         if (!Files.exists(secretPath)) {
             logger.warnf("Cannot find secret %s in %s", vaultSecretId, secretPath);
             return DefaultVaultRawSecret.forBuffer(Optional.empty());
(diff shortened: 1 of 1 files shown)


vmuzikar (commented):
@Pepo48 Thank you for the PR.

Could you please open an issue for this and link it to the PR?

Did you try if it resolves the false positive reports from the scanners?

Just a note for other reviewers: this is not fixing any actual issue, we already had sufficient validations in place to prevent the path traversal. Still, it triggers security scanners but it is false positive.

Pepo48:
The reason I opted for such a "fix" for now is that I remember that we had a similar (if not the same) issue with CodeQL false positives and the keystore vault implementation. Back then I solved it with the `toAbsolutePath().normalize()` path transformation. 

@abstractj how can I be certain that this adjustment really fixed the problem? What I did prior to creating the PR was that I ran the CI and the CodeQL stage in particular, within my fork. At the same time I tried to observe whether an alert would appear and whether it would be linked to the existing CSA reports. Nothing like that happened. Is there anything else that I should do?

Thanks.

shawkins (approved):
I'm fine with the changes assuming they satisfy the check.

abstractj:
> @abstractj how can I be certain that this adjustment really fixed the problem? What I did prior to creating the PR was that I ran the CI and the CodeQL stage in particular, within my fork. At the same time I tried to observe whether an alert would appear and whether it would be linked to the existing CSA reports. Nothing like that happened. Is there anything else that I should do?
> 
> Thanks.

@Pepo48 don't worry about satisfying the checks from CodeQL, if you and @shawkins believe that the solution proposed solves the issue, that's pretty much fine. Whenever there's a false positive, please dismiss it adding the reasoning behind it, or if you are unable to dismiss, please provide the justification on the same ticket and I will do that.

vmuzikar (approved):

PR #44361, opened 2025-11-20
Use PrivateKey directly when decrypting SAML

### LINKED ISSUE #44289: Failure when decrypting SAML Assertions with HSM provided key
### Before reporting an issue

### Area

saml

### Describe the bug

We have a HSM box from LUNA and have stored our private key in there.

We have made a two org.keycloak.keys.KeyProviderFactory, creating keywrappers on with use SIG and one with use ENC

Signing works fine, but encryption fails in org.keycloak.common.util.DerUtils#decodePrivateKey(byte[]) from org.keycloak.protocol.saml.SAMLDecryptionKeysLocator#getKeys

This is as far as I understand, since what is in encoded is not private key, because that can of course not be extracted from a HSM

When just skipping the encoding and decoding, and casting the key from KeyWrapper to a private key, it seems to work with our hsm module.

So I image a fix could be something like this:

` 
        // Map keys to PrivateKey
        return keysStream
                .map(this::toPrivateKey)
                .filter(Objects::nonNull)
                .collect(Collectors.toList());
` 

`  private PrivateKey toPrivateKey(KeyWrapper keyWrapper) {
    if (keyWrapper.isUsePrivateKeyDirectly()) {
      return (PrivateKey) keyWrapper.getPrivateKey();
    }
    if (keyWrapper.getPrivateKey() != null) {
      byte[] encoded = keyWrapper.getPrivateKey().getEncoded();
      try {
        return DerUtils.decodePrivateKey(encoded);
      } catch (Exception e) {
        throw new RuntimeException("Could not decode private key.", e);
      }
    }
    return null;
  }`

where isUsePrivateKeyDirectly should be a new field on KeyWrapper, which is something KeyProviderFactory can
 set.

### Version

26.4.5

### Regression

### Expected behavior

Keycloak will successfully decrypt the assertions from the identity provider

### Actual behavior

It throws         throw new InvalidKeySpecException("Unable to decode the private key with supported algorithms: " + String.join(", ", algorithms));
from org.keycloak.common.util.DerUtils#decodePrivateKey(byte[])

### How to Reproduce?

Probably difficult to produce without a HSM, but generally:

1. Setup keyprovider for encryption that uses a HSM-module
2. Setup saml identity provider that uses Wants Encryption
3. Login

### Anything else?

_No response_

rmartinc:
Closes #44289

I'm just removing the decoding part and directly casting to `PrivateKey`. In the `KeyWrapper` the `getPrivateKey` should return just that, and we are doing the cast directly in other parts already. Besides, the decryption checks that the encryption algorithm is understood by keycloak (`rsa-oaep`, `rsa-oaep-mgf1p` or `rsa-1_5`) and it is translated to the correct internal algorithm (`RSA_OAEP` or `RSA1_5`). So I expect that any `KeyProvider` is going to return a `PrivateKey` here. I tested the CI is OK with the change.

The reporter seems to be implementing their own keys provider to use an HSM, and therefore the encoded part is not available for the hardware key.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/saml/SAMLDecryptionKeysLocator.java
+++ b/services/src/main/java/org/keycloak/protocol/saml/SAMLDecryptionKeysLocator.java
@@ -17,9 +17,6 @@
 
 package org.keycloak.protocol.saml;
 
-
-
-import java.security.Key;
 import java.security.PrivateKey;
 import java.util.LinkedList;
 import java.util.List;
@@ -28,7 +25,6 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.keycloak.common.util.DerUtils;
 import org.keycloak.crypto.KeyUse;
 import org.keycloak.crypto.KeyWrapper;
 import org.keycloak.models.KeycloakSession;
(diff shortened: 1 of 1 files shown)


mposolda (approved):
@rmartinc Thanks

PR #44370, opened 2025-11-20
Make sd-jwt key binding verification work with EdDSA keys

### LINKED ISSUE #44369: sd-jwt key binding verification does not work with EdDSA keys
### Description

SD-JWT key binding verification does not work with EdDSA keys as those are not supported in `JwkParsingUtils` .

mposolda:
closes #44369 

Summary of the changes:

- Make sure that sd-jwt VP key-binding verification supports also EdDSA like other Keycloak functionality

- Involves small update of `JwkParsingUtils` and update duplication for retrieve default `algorithm` for EC keys in case not provided as it is already implemented in `KeyWrapper.getAlgorithmOrDefault()` (Also allowed to remove class `ECCurve` )

- Renamed method `withKeyBinding` on `IssuerSignedJWT.Builder` to `withKeyBindingKey` to avoid some confusion as we're providing rather the key-binding key used inside `cnf` claim of issuer-signed JWT rather than the key-binding (The key-binding itself is rather provided in `SdJwt.withKeyBindingJwt`). Also removed the unused methods `withKeyBinding` on `IssuerSignedJWT.Builder` to make only single signature `withKeyBindingKey(JWK keyBinding)` 

- Added `SdJwtKeyBindingTest` for more thorough testing for the key-binding creation and verification with various algorithms (EC, EdDSA, RSA) to make sure both parts (signing as well as verification) work fine.

- Did some refactoring and added some stuff to `KeyUtils` to generate EdDSA keys and EC keys to avoid code duplications in Keycloak codebase. Also added `KeyWrapperUtils` in `core` module to avoid some duplications.


mposolda:
@tnorimat @Captain-P-Goldfish @IngridPuppet @Awambeng @Ogenbertrand @forkimenjeckayang @francis-pouatcha Anyone available to review the PR please?

core/src/main/java/org/keycloak/crypto/KeyWrapper.java
```diff
@@ -91,15 +97,21 @@ public String getAlgorithmOrDefault() {
                     if (curve != null) {
                         switch (curve) {
                             case "P-256":
+                            case EC_KEY_SECP256R1:
                                 return Algorithm.ES256;
                             case "P-384":
+                            case EC_KEY_SECP384R1:
                                 return Algorithm.ES384;
                             case "P-512":
```
tnorimat, line 105: It seems that "P-512" is not used from anywhere, so how about removing "P-512"?

Captain-P-Goldfish (approved):
nice and clean code. Easy to read and to follow. Good Work!
And sorry for the late review. Simply forgot it...

tnorimat (commented):
I added one minor comment. Could you check that?

core/src/main/java/org/keycloak/crypto/KeyWrapper.java
Captain-P-Goldfish: I would prefer to keep it. It might very well be that we are going to use it in one of our custom plugins

mposolda: @tnorimat @Captain-P-Goldfish  Thanks. I also vote for keeping it. Also for the backwards compatibility corner-cases.
@tnorimat Is it ok for you to keep it then?

tnorimat: @mposolda @Captain-P-Goldfish Yes I agree with that.

mposolda:
@Captain-P-Goldfish @tnorimat Thanks!

PR #44383, opened 2025-11-21
fix: handle localized date formatting in message format

### LINKED ISSUE #44377: Unlocalized date format in (email messages)
### Before reporting an issue

### Area

translations

### Describe the bug

When using date parameters in the `msg` formatter, they are not formatted according to the user locale.

Freemarker wraps the Java `Date` object in [`SimpleData`](https://freemarker.apache.org/docs/api/freemarker/template/SimpleDate.html) object and the `MessageFormatterMethod` considers it a `SimpleScalar`, which is converted to a `String`.

see: https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/theme/beans/MessageFormatterMethod.java#L72-L73

While Freemarker would use the locale to format a date, this is no longer possible with the stringified date.
A simple fix could look like:

```
	if (item instanceof SimpleDate) {
		item = ((SimpleDate) item).getAsDate();
	} else if (item instanceof SimpleScalar) {
		item = ((SimpleScalar) item).getAsString();
	}
```

### Version

26.4.2

### Regression

### Expected behavior

Emails contain a correctly formatted date according the users locale. In the example case, the format should be something like `21.11.25, 09:46`.

### Actual behavior

Receiving a mail containing a date, it is not correctly formatted to the users locale.

An example with a German locale setting:

Image

### How to Reproduce?

- Activate "email" event listener in realm settings -> Events
- Configure realm email settings so you can receive emails
- Configure (additionally) German as a supported realm language
- Configure a user with an email address which you can access
- Login as a realm user into account console
- Set user locale to German
- Update password ("Anmeldung" -> "Aktualisieren")
- An email was sent to the user which contains the formatted date, but with a non-German format

### Anything else?

I am planning to provide a pull request for this.

resah:
Closes #44377

Handles `SimpleDate` (like `SimpleNumber`) in message formatter.
Added unit tests for locale-specific formatting.

### DIFF
--- a/services/src/test/java/org/keycloak/theme/beans/MessageFormatterMethodTest.java
+++ b/services/src/test/java/org/keycloak/theme/beans/MessageFormatterMethodTest.java
@@ -18,9 +18,12 @@
 package org.keycloak.theme.beans;
 
 import java.util.Arrays;
+import java.util.Date;
 import java.util.Locale;
 import java.util.Properties;
+import java.util.TimeZone;
 
+import freemarker.template.SimpleDate;
 import freemarker.template.SimpleNumber;
 import freemarker.template.SimpleScalar;
 import freemarker.template.TemplateModelException;
@@ -43,6 +46,7 @@ public class MessageFormatterMethodTest {
         properties.setProperty("client_admin-console", "Admin Console");
         properties.setProperty("realm_example-realm", "Example Realm");
         properties.setProperty("key", "foo {0,choice,0#foo|1#bar|1<{0} foobar} bar");
+        properties.setProperty("simpleKey", "{0}");
 
 
         MessageFormatterMethod fmt = new MessageFormatterMethod(locale, properties);
@@ -64,6 +68,27 @@ public class MessageFormatterMethodTest {
 
         msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("key"),new SimpleNumber(2)));
         Assert.assertEquals("foo 2 foobar bar", msg);
+
+        msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("simpleKey"),new SimpleNumber(2.5)));
+        Assert.assertEquals("2.5", msg);
+
+        TimeZone defaultTimeZone = TimeZone.getDefault();
+        try {
+            TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
+
+            msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("simpleKey"), new SimpleDate(new Date(1763721018000L), 2)));
+            Assert.assertTrue(msg.matches("11/21/25, 10:30.AM"));
+
+            MessageFormatterMethod germanFmt = new MessageFormatterMethod(Locale.GERMANY, properties);
+
+            msg = (String) germanFmt.exec(Arrays.asList(new SimpleScalar("simpleKey"), new SimpleNumber(2.5)));
+            Assert.assertEquals("2,5", msg);
+
+            msg = (String) germanFmt.exec(Arrays.asList(new SimpleScalar("simpleKey"), new SimpleDate(new Date(1763721018000L), 2)));
+            Assert.assertEquals("21.11.25, 10:30", msg);
+        } finally {
+            TimeZone.setDefault(defaultTimeZone);
+        }
     }
 
 }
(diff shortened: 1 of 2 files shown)


ahus1 (changes requested):
@resah - thank you for your PR. I've tested it and it good for merging. 

We recently implemented spotless to ensure a consistent ordering of imports. I've created a PR https://github.com/bare-id/keycloak/pull/1 to fix that. Please merge this PR to your branch. 

Thanks!

services/src/test/java/org/keycloak/theme/beans/MessageFormatterMethodTest.java
```diff
@@ -64,6 +67,20 @@ public void test() throws TemplateModelException {
 
         msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("key"),new SimpleNumber(2)));
         Assert.assertEquals("foo 2 foobar bar", msg);
+
+        msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("simpleKey"),new SimpleNumber(2.5)));
+        Assert.assertEquals("2.5", msg);
+
+        msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("simpleKey"),new SimpleDate(new Date(1763721018000L), 2)));
+        Assert.assertEquals("11/21/25, 11:30 AM", msg);
```
ahus1: This test fails, as the file contains a narrow-non-breaking-space UTF-8 character. But Java uses a simple space. Please update your PR

```suggestion
        Assert.assertEquals("11/21/25, 11:30 AM", msg);
```

ahus1 (changes requested):
Thank you for merging the import changes. Unfortunately the new unit test fails for me. See below for a suggested change.

services/src/test/java/org/keycloak/theme/beans/MessageFormatterMethodTest.java
```diff
@@ -64,6 +67,20 @@ public void test() throws TemplateModelException {
...
+        msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("simpleKey"),new SimpleDate(new Date(1763721018000L), 2)));
+        Assert.assertEquals("11/21/25, 11:30 AM", msg);
+
+        MessageFormatterMethod germanFmt = new MessageFormatterMethod(Locale.GERMANY, properties);
+
+        msg = (String) germanFmt.exec(Arrays.asList(new SimpleScalar("simpleKey"),new SimpleNumber(2.5)));
+        Assert.assertEquals("2,5", msg);
+
+        msg = (String) germanFmt.exec(Arrays.asList(new SimpleScalar("simpleKey"),new SimpleDate(new Date(1763721018000L), 2)));
+        Assert.assertEquals("21.11.25, 11:30", msg);
```
ahus1: It seems that your test is dependent on the time zone of the server. See below for a suggestion to normalize the time zone for the duration of the test:

```suggestion
        TimeZone defaultTimeZone = TimeZone.getDefault();
        try {
            TimeZone.setDefault(TimeZone.getTimeZone("UTC"));

            msg = (String) fmt.exec(Arrays.asList(new SimpleScalar("simpleKey"), new SimpleDate(new Date(1763721018000L), 2)));
            Assert.assertEquals("11/21/25, 10:30 AM", msg);

            MessageFormatterMethod germanFmt = new MessageFormatterMethod(Locale.GERMANY, properties);

            msg = (String) germanFmt.exec(Arrays.asList(new SimpleScalar("simpleKey"), new SimpleNumber(2.5)));
            Assert.assertEquals("2,5", msg);

            msg = (String) germanFmt.exec(Arrays.asList(new SimpleScalar("simpleKey"), new SimpleDate(new Date(1763721018000L), 2)));
            Assert.assertEquals("21.11.25, 10:30", msg);
        } finally {
            TimeZone.setDefault(defaultTimeZone);
        }
```

ahus1 (changes requested):
The unit test now passes on my local machine, but it fails on GitHub actions due to a different time zone. See below for a suggested change.

ahus1:
@resah - thank you. The build now fails as it is missing imports. Please have a look.

resah:
@ahus1 I fixed the imports, sqashed the commits and also added a unicode character for a narrow non-breaking space for US time formats, since this test failed for me when using a normal whitespace character.

ahus1:
@resah - it actually seems to be a bit more complicated with that nonbreaking space: Some JDK versions use a classic space, and newer ones a non-breaking one. See https://bugs.openjdk.org/browse/JDK-8324308

As we're testing with both JDK17 and JDK25, you would need to consider possibly a regex for verifying the US date format.

resah:
@ahus1 Thank you for bearing with me, I changed it to an regular expression.

ahus1 (approved):
Thank you for this contribution!

PR #44395, opened 2025-11-21
Escape passkeys descriptions and labels depending on the context

### LINKED ISSUE #44387: Description of Passkey is not escaped on the login screen
### Before reporting an issue

### Area

login/ui

### Describe the bug

The 

### Version

26.4

### Regression

### Expected behavior

The HTML should be escaped

### Actual behavior

It is not escaped:

Image

### How to Reproduce?

* Create a passkey as a second factor and provide it a label with HTML. Here a screenshot from the admin UI:

Image

### Anything else?

I'll create a PR

ahus1:
Closes #44387

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/webauthn/registration/PolicyJsInjectionTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/webauthn/registration/PolicyJsInjectionTest.java
@@ -160,7 +160,7 @@ public class PolicyJsInjectionTest extends AbstractWebAuthnVirtualTest {
             assertThat(authenticators, notNullValue());
             assertThat(authenticators.getItems(), not(Matchers.empty()));
 
-            assertThat(authenticators.getLabels().get(0), is("label`;window.prompt(\"another\");"));
+            assertThat(authenticators.getLabels().get(0), is(originalLabel));
         }
     }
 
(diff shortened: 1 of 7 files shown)


mposolda (commented):
The idea of the PR looks good. Few questions:

- Is this backwards compatible? For example if some user already has DB with the label of the passkey credential set to some strange value, would it be still OK for him to authenticate after migration ? I suppose it is OK if user (or administrator) cannot set some strange values anymore, but just doublecheck if everything fine if already set. But it seems it should be OK as the PR is mainly abou sanitization. But just asking :-)

- Is this related also to OTP credential? That one also has label, which can be set by administrator (or user) and is eventually displayed on the login screen (especially if user has multiple OTP credentials as then he can choose from them, which he prefers). Should this be added here as well (if it is not already there) for consistency?

@rmartinc Can you please review as well?

ahus1:
> Is this backwards compatible? For example if some user already has DB with the label of the passkey credential set to some strange value, would it be still OK for him to authenticate after migration ?

Yes, it is backwards compatible. It is only about showing the value. If they have some strange values in description, those strange values will appear 1:1 as they entered them, and will not be rendered as HTML. 

Please note: This is not about preventing setting "strang"e values, it is about showing those "strange" values 1:1 and safely as the user entered them. 

> Is this related also to OTP credential?

I see your point, but I didn't check this in this PR. I'd like to handle this in a follow-up issue. It took me a while already to get it all working for Passkeys. :smile:

rmartinc (approved):
I didn't know you can do the `<#outputformat "JavaScript">` in the ftl. :smile: 
Good job @ahus1! LGTM!
I think that the labels in OTP are not using the `no_esc`, we can check it later just in case.

ahus1:
@rmartinc - about `<#outputformat "JavaScript">` -- just be aware to use the `?c` for all items in that block, otherwise they will not be escaped at all.

mposolda (approved):
@ahus1 @rmartinc Thanks! +1 to check OTP as a follow-up if needed.

Issue #39033, opened 2025-04-16
Organization with Identity Provider allowing login with password


williancolognesitrimble:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

organizations

### Describe the bug

When using Keycloak with Organizations enabled and an Identity Provider configured for users within a specific domain, users are still able to log in with their Keycloak username and password instead of being redirected to the external Identity Provider. This occurs if the user has any credentials registered in Keycloak.

It's unclear whether this behavior is intentional, as the documentation doesn't mention it. However, it seems that users should be redirected to the external Identity Provider regardless of any existing credentials in Keycloak.

### Version

26.1.4

### Regression

- [ ] The issue is a regression

### Expected behavior

The expected behavior for item 7 would be redirect the user "test@example.com" to identity provider tied to it, and it's forwarding the user to username/password keycloak's page.

I believe this might be even a security issue if we rely that the user should only authenticate through identity provider.

### Actual behavior

It's asking for the user to login with username/password instead of forward to identity provider when the flag `Redirect when email domain matches` is enabled.

### How to Reproduce?

1. Enable Organization
2. Create a new Organization and add a "Domain", let's say "example.com"
3. Create a new Identity Provider
4. Go to Organization -> Identity Providers -> Link Identity Provider -> Add the domain "example.com" and enable "Redirect when email domain matches" to On.
5. Go to Users -> Create new user with username "test@example.com"
6. Go to user credentials tab and set a password.
7. Go to login page and fill the username

williancolognesitrimble:
> Perhaps we should not be opinionated about the behavior and allow choosing to force users to redirect when credentials are set.

Make sense, there is a reason why the behavior is different when the credentials is set? It's unclear in the docs that it would be possible.

pedroigor:
The assumption is that if the user has credentials set, we should favor authenticating using their credential. I understand that might not be the case all the time because:

* The user can set a credential via the account console if allowed to do so
* If forgot password is enabled, the user can set a credential

We did not update the docs when this behavior was introduced.

pedroigor:
Thanks, @williancolognesitrimble. For me, it makes sense. Let us see if others can share their interest in this enhancement.

BJSummerfield:
I would like this feature or a work around.  
[...]

PR #44410, opened 2025-11-23
Align organization broker redirect after OTP setup

h0xerz2:
## Summary
The following configuration is assumed:
1. Organization created and linked to IDP
2. IDP is configured with - *Hide on login page*
2. IDP is configured with - *Redirect when email domain matches*
3. IDP links *Post login flow* enabling OTP

Problem:  user is not  redirected to IDP after first broker login, instead password is requested from the user.

The solution is to check for first factor credentials within `OrganizationAuthenticator#tryRedirectBroker` method. First factor credentials are defined with a help of `SubjectCredentialManager#getFirstFactorCredentialsStream` method.
## Issue
Closes #40510

## Testing
- [ ] `./mvnw -f testsuite/integration-arquillian/tests/base/pom.xml -Dtest=OrganizationPostBrokerLoginTest#testAutoRedirectAfterPostBrokerOtpSetup verify`

## Documentation
- [ ] Documentation update is not required

## Checklist
- [ ] Rebased on `main`
- [ ] Single signed-off commit
- [ ] Spotless check passes (`./mvnw spotless:check`)
- [ ] Added simple test - `OrganizationPostBrokerLoginTest.java`

### DIFF
--- a/docs/documentation/server_admin/topics/organizations/authenticating-members.adoc
+++ b/docs/documentation/server_admin/topics/organizations/authenticating-members.adoc
@@ -35,6 +35,7 @@ In addition to identifying the user once the username is provided, the identity-
 * Matching an email domain to an organization.
 * Deciding if the authentication flow should continue or not if an account already exists for the username provided
 * Deciding how the user should be authenticated depending on how the domains and the identity providers are configured to an organization
+and the set of credentials configured to the user account.
 * Seamlessly authenticating users through an identity provider associated with an organization if the email domain matches the domain set to the identity provider
 
 The identity-first login provides the same capabilities that are provided by the usual login page with the username and
@@ -65,7 +66,9 @@ For more details, see <<_managing_identity_provider_,Managing identity providers
 
 In a similar situation to the previous section, the organization may have an identity provider set with one of the organization domains.
 In this situation, the user is redirected to the identity provider if that user's email matches a specific domain from the organization.
-Once the flow completes, an account is created and the user joins the organization.
+Once the flow completes, an account is created and the user joins the organization. In case the user has any first-factor credentials
+configured (e.g.: password, passwordless, kerberos) to the account, the user is not automatically redirected to the identity provider but asked to authenticate using their
+credentials.
 
 == Configuring existing authentication flows
 
--- a/model/storage/src/main/java/org/keycloak/credential/UserCredentialManager.java
+++ b/model/storage/src/main/java/org/keycloak/credential/UserCredentialManager.java
@@ -44,22 +44,24 @@ import io.opentelemetry.api.trace.StatusCode;
  *
  * @author Alexander Schwartz
  */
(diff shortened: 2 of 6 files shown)


ahus1:
@h0xerz2 - thank you for the PR. Unfortunately there is a merge conflict. Please rebase your PR. Thanks!

ahus1:
Pinging @keycloak/core-iam  for a review.

pedroigor:
Hi @h0xerz2, we are interested in having your changes. Sorry for the delay.

I've added a commit with a few adjustments. Please let me know if they work for you. My goal was to keep changes to a minimum as well as update the docs.

docs/documentation/server_admin/topics/organizations/authenticating-members.adoc
```diff
@@ -65,7 +66,9 @@ For more details, see <<_managing_identity_provider_,Managing identity providers
 
 In a similar situation to the previous section, the organization may have an identity provider set with one of the organization domains.
 In this situation, the user is redirected to the identity provider if that user's email matches a specific domain from the organization.
-Once the flow completes, an account is created and the user joins the organization.
+Once the flow completes, an account is created and the user joins the organization. In casse the user has any first-factor credentials
```
dasniko: small typo: `case` not `casse`

dasniko (changes requested):
Thanks for this PR.
I've just tested it and can confirm it works. 👍

dasniko:
@pedroigor regarding your comment here: https://github.com/keycloak/keycloak/issues/45456#issuecomment-3754451985
> I'm also wondering if we should add an option to allow organization administrators to choose whether they want to or not to authenticate using first-factor credentials if they exist for a user. We can discuss that in the PR.

I'd like to have such a feature, also I know several customers who would like to have such an option.
But perhaps it would also make (more?) sense to have an option to disallow org users from setting a local password (or first-factor cred) at all. Most use-cases I'm aware of, have the org users to authenticate through external Idp only, at least for first factor, and only have a local (either optional or mandatory or depending on the `amr` claim send from the IdP, if possible) 2FA credential. Currently I've implemented an approach in a custom override of the `UpdatePassword` action.

Perhaps it's worth to discuss this in a separate/new issue!?

pedroigor:
> But perhaps it would also make (more?) sense to have an option to disallow org users from setting a local password (or first-factor cred) at all.

It does make more sense but we'll require more changes. Let's hold this idea a bit and plan delivering users from setting up credentials ....

pedroigor:
@dasniko If you want to send your changes to disable credential management for users, please feel free to do so. 

The corresponding issue is https://github.com/keycloak/keycloak/issues/39033#issuecomment-3276034621. There are two or more solutions being discussed there, but I think most people agree that enabling/disabling credential management for users federated from an specific organization broker is what we want.

pedroigor (approved):

pedroigor:
Thanks @h0xerz2!

PR #44431, opened 2025-11-24
Fix NPE when importing SAML EntityDescriptor without SPSSODescriptor

### LINKED ISSUE #44430: NPE when importing SAML EntityDescriptor without SPSSODescriptor
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

When importing a SAML EntityDescriptor via the SAML Entity Descriptor Client Registration endpoint, Keycloak throws a NullPointerException if the uploaded metadata does not contain an SPSSODescriptor.

### Version

main

### Regression

### Expected behavior

Error handling

### Actual behavior

Keycloak returns server error:
```
Uncaught server error: java.lang.NullPointerException: Cannot invoke "org.keycloak.dom.saml.v2.metadata.SPSSODescriptorType.isWantAssertionsSigned()" because "spDescriptorType" is null
        at org.keycloak.protocol.saml.EntityDescriptorDescriptionConverter.loadEntityDescriptors(EntityDescriptorDescriptionConverter.java:193)
        at org.keycloak.protocol.saml.EntityDescriptorDescriptionConverter.convertToInternal(EntityDescriptorDescriptionConverter.java:73)
        at org.keycloak.protocol.saml.clientregistration.EntityDescriptorClientRegistrationProvider.createSaml(EntityDescriptorClientRegistrationProvider.java:47)
        at org.keycloak.protocol.saml.clientregistration.EntityDescriptorClientRegistrationProvider$quarkusrestinvoker$createSaml_7b43b364c4f60a385ea6d147f84361448f568d5a.invoke(Unknown Source)
        at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:183)
        at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:645)
        at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)
        at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)
        at java.base/java.lang.Thread.run(Thread.java:840)
```

### How to Reproduce?

 1. Build and run Keycloak:
```
./mvnw -pl quarkus/dist -am clean install
./kc.sh start-dev
```
 2. POST SAML metadata without an SPSSODescriptor:
```
curl -X POST \
  -H "Content-Type: application/xml" \
  --data-binary @keycloak/saml-core/src/test/resources/org/keycloak/saml/processing/core/parsers/saml/KEYCLOAK-4809-IdPMetadata_test.xml \
  http://localhost:8080/realms/master/clients-registrations/saml2-entity-descriptor
```

### Anything else?

_No response_

PavlNekrasov:
resolves #44430

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/saml/EntityDescriptorDescriptionConverter.java
+++ b/services/src/main/java/org/keycloak/protocol/saml/EntityDescriptorDescriptionConverter.java
@@ -27,6 +27,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
+import jakarta.ws.rs.BadRequestException;
+
 import org.keycloak.Config;
 import org.keycloak.dom.saml.v2.metadata.EndpointType;
 import org.keycloak.dom.saml.v2.metadata.EntitiesDescriptorType;
@@ -153,7 +155,7 @@ public class EntityDescriptorDescriptionConverter implements ClientDescriptionCo
         try {
             metadata = SAMLParser.getInstance().parse(is);
         } catch (ParsingException e) {
-            throw new RuntimeException(e);
+            throw new BadRequestException(e);
         }
         EntitiesDescriptorType entities;
 
@@ -165,7 +167,7 @@ public class EntityDescriptorDescriptionConverter implements ClientDescriptionCo
         }
 
         if (entities.getEntityDescriptor().size() != 1) {
-            throw new RuntimeException("Expected one entity descriptor");
+            throw new BadRequestException("Expected one entity descriptor");
         }
 
         EntityDescriptorType entity = (EntityDescriptorType) entities.getEntityDescriptor().get(0);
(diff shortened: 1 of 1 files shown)


mposolda:
@rmartinc Do you please have a chance to review this PR?

services/src/main/java/org/keycloak/protocol/saml/EntityDescriptorDescriptionConverter.java
```diff
@@ -187,6 +187,7 @@ private static ClientRepresentation loadEntityDescriptors(InputStream is) {
         attributes.put(SamlConfigAttributes.SAML_SIGNATURE_ALGORITHM, SignatureAlgorithm.RSA_SHA256.toString());
         attributes.put(SamlConfigAttributes.SAML_AUTHNSTATEMENT, SamlProtocol.ATTRIBUTE_TRUE_VALUE);
         SPSSODescriptorType spDescriptorType = getSPDescriptor(entity);
+        if (spDescriptorType == null) return app;
```
rmartinc: This method is to create a SAML client via the  `SPSSODescriptor`. So, if the descriptor is not for an SP, maybe it's better to throw an exception. Because it makes no sense returning something that will be completely empty or filled with default values. Similar to the case when there is no entity descriptor a few lines before. Something like:

```java
if (spDescriptorType == null) {
    throw new RuntimeException("No SPSSODescriptorType defined in the entity descriptor file");
}
```

If you prefer you can return a `BadRequestException`. Because, just curiosity, what are you trying to import here? A IDPSSODescriptor or another thing?

rmartinc (changes requested):
I think that throwing an exception would be better than returning an empty client. WDYT?

PavlNekrasov:
I was fuzzing `convertToInternal()`.
I’d suggest using a **BadRequestException** in `loadEntityDescriptors` everywhere, because `RuntimeException` isn’t handled in `createSaml`, making `BadRequestException` a controlled exception.

rmartinc (approved):
Thanks @PavlNekrasov!

mposolda (approved):
@PavlNekrasov @rmartinc Thanks for the PR and review!

PR #44439, opened 2025-11-24
[OID4VCI] Conformance Test Fixes 

### LINKED ISSUE #44659: [OID4VCI] Conformance Test Fixes
### Description

There are few updates, which needs to be done to improve Keycloak passing with the OIDF conformance testsuite. See related PR description for the details: https://github.com/keycloak/keycloak/pull/44439

forkimenjeckayang:
closes #44659

## Summary

This PR fixes multiple issues from the OID4VC conformance tests, ensuring all tests now pass for both Pre-Authorized and Authorization Code flows.

---

#### 1. Missing exp Claim in Credentials
- **Issue:** Credentials missing expiration claim
- **Fix:** Added automatic expiration date from credential configuration
- **Result:** Test passes

#### 2. DPoP Proof Nonce Reuse
- **Issue:** DPoP nonce could be reused within the same request
- **Fix:** Implemented request-scoped caching to prevent reuse
- **Result:** Test passes

#### 3. Missing x5c Certificate Header in SD-JWT Credentials
- **Issue:** SD-JWT credentials missing required x5c certificate header (required by HAIP-6.1.1)
- **Fix:** Added x5c certificate chain to SD-JWT credential headers by retrieving the certificate chain from the `SignatureSignerContext` and encoding it as base64. This ensures the certificate matches the signing key used for the credential.
- **Result:** HAIP-6.1.1 requirement for conformance test passes

#### 4. Authorization Code Flow Failure
- **Issue:** Authorization code flow failed with HTTP 500 errors
- **Root Cause:** Authentication was re-validated multiple times, bypassing DPoP cache
- **Fix:** Updated authentication flow to use cached result within request lifecycle
- **Result:** Authorization Code Flow tests now pass

## Final Outcome

image


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -78,7 +79,16 @@ public SdJwtCredentialBody buildCredentialBody(VerifiableCredential verifiableCr
         claimSet.put(ISSUER_CLAIM, credentialBuildConfig.getCredentialIssuer());
         claimSet.put(VERIFIABLE_CREDENTIAL_TYPE_CLAIM, credentialBuildConfig.getCredentialType());
 
-        // jti, nbf, iat and exp are all optional. So need to be set by a protocol mapper if needed.
+        // Set exp claim from verifiable credential expiration date
+        // expiry is optional, but should be set if available to comply with HAIP
+        // see: https://openid.github.io/OpenID4VC-HAIP/openid4vc-high-assurance-interoperability-profile-wg-draft.html#section-6.1
+        // Only set if not already set by a protocol mapper
+        if (!claimSet.containsKey("exp")) {
```
mposolda: Minor: Can be replaced with `CLAIM_NAME_EXP` from `OID4VCConstants` . Same on line 88.

mposolda (commented):
@forkimenjeckayang Thanks for the PR. Added one minor comment inline.

Would be good to have also review from other members (CC @Captain-P-Goldfish @tnorimat @wistefan )

forkimenjeckayang:
Thank you for spotting @mposolda 
Changes addressed

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -768,8 +772,24 @@ public Response requestCredential(String requestPayload) {
...
             Proofs originalProofs = credentialRequestVO.getProofs();
+            // Determine the proof type from the original proofs
+            String proofType = null;
+            if (originalProofs != null) {
+                if (originalProofs.getJwt() != null && !originalProofs.getJwt().isEmpty()) {
+                    proofType = ProofType.JWT;
+                } else if (originalProofs.getAttestation() != null && !originalProofs.getAttestation().isEmpty()) {
+                    proofType = ProofType.ATTESTATION;
+                }
+            }
```
Captain-P-Goldfish: could you add this logic in a getter-method in the class `Proofs`?

```java
public String getProofType()
{
if (originalProofs.getJwt() != null && !originalProofs.getJwt().isEmpty()) {
                    return ProofType.JWT;
                } else if (originalProofs.getAttestation() != null && !originalProofs.getAttestation().isEmpty()) {
                    return ProofType.ATTESTATION;
                }
}
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -768,8 +772,24 @@ public Response requestCredential(String requestPayload) {
...
+
             for (String currentProof : allProofs) {
-                credentialRequestVO.setProofs(new Proofs().setJwt(List.of(currentProof)));
+                Proofs proofForIteration = new Proofs();
+                if (ProofType.JWT.equals(proofType)) {
+                    proofForIteration.setJwt(List.of(currentProof));
+                } else if (ProofType.ATTESTATION.equals(proofType)) {
+                    proofForIteration.setAttestation(List.of(currentProof));
+                }
```
Captain-P-Goldfish: this logic could also be added in a setter-method within the `Proofs`-class

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -567,7 +569,9 @@ private void checkScope(CredentialScopeModel requestedCredential) {
         String vcIssuanceFlow = clientSession.getNote(PreAuthorizedCodeGrantType.VC_ISSUANCE_FLOW);
 
         if (vcIssuanceFlow == null || !vcIssuanceFlow.equals(PreAuthorizedCodeGrantTypeFactory.GRANT_TYPE)) {
-            AccessToken accessToken = bearerTokenAuthenticator.authenticate().token();
+            // Use getAuthResult() instead of bearerTokenAuthenticator.authenticate() directly
+            // This ensures we benefit from the cachedAuthResult caching that prevents DPoP proof reuse
+            AccessToken accessToken = getAuthResult().getToken();
```
Captain-P-Goldfish: `getToken()` is deprecated

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -1000,15 +1024,35 @@ private List<String> getAllProofs(CredentialRequest credentialRequestVO) {
...
+        // Validation already happened in normalizeProofFields, so we can safely extract proofs
+        if (hasProofEntries(proofs.getJwt())) {
+            allProofs.addAll(proofs.getJwt());
+        } else if (hasProofEntries(proofs.getAttestation())) {
+            allProofs.addAll(proofs.getAttestation());
         }
 
-        allProofs.addAll(proofs.getJwt());
         return allProofs;
     }
```
Captain-P-Goldfish, line 1023: wouldn't it be better to move this logic also into the `Proofs`-class?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -768,8 +772,24 @@ public Response requestCredential(String requestPayload) {
...
-                credentialRequestVO.setProofs(new Proofs().setJwt(List.of(currentProof)));
+                Proofs proofForIteration = new Proofs();
+                if (ProofType.JWT.equals(proofType)) {
+                    proofForIteration.setJwt(List.of(currentProof));
+                } else if (ProofType.ATTESTATION.equals(proofType)) {
+                    proofForIteration.setAttestation(List.of(currentProof));
+                }
+                credentialRequestVO.setProofs(proofForIteration);
                 Object theCredential = getCredential(authResult, supportedCredential, credentialRequestVO);
                 responseVO.addCredential(theCredential);
```
Captain-P-Goldfish, line 791: A little comment is missing in my opinion:

```java 
// Creating credential with keybinding to the current proof 
```

In the first moment I was irritated by call to the `setProofs` method

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -963,7 +983,11 @@ private void normalizeProofFields(CredentialRequest credentialRequest) {
             }
             credentialRequest.setProofs(proofsArray);
             credentialRequest.setProof(null);
+            validateProofTypes(proofsArray);
+            return;
         }
+
+        validateProofTypes(credentialRequest.getProofs());
     }
```
Captain-P-Goldfish, line 987: simply:

```java
        if (credentialRequest.getProof() != null) {
            LOGGER.debugf("Converting single 'proof' field to 'proofs' array for backward compatibility");
            JwtProof singleProof = credentialRequest.getProof();
            Proofs proofsArray = new Proofs();
            if (singleProof.getJwt() != null) {
                proofsArray.setJwt(List.of(singleProof.getJwt()));
            }
            credentialRequest.setProofs(proofsArray);
            credentialRequest.setProof(null);
        }

        validateProofTypes(credentialRequest.getProofs());
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -166,6 +166,8 @@ public class OID4VCIssuerEndpoint {
 
     private Cors cors;
 
+    private AuthenticationManager.AuthResult cachedAuthResult;
```
Captain-P-Goldfish, line 176: a little bit of javadoc would be helpful

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -1201,16 +1249,18 @@ private AuthenticationManager.AuthResult getAuthResult() {
...
-        AuthenticationManager.AuthResult authResult = bearerTokenAuthenticator.authenticate();
-        if (authResult == null) {
+        try {
+            return getAuthResult();
+        } catch (BadRequestException e) {
+            LOGGER.debugf(e, "Authentication failed, throwing error response");
             throw errorResponse;
         }
-        return authResult;
     }
```
Captain-P-Goldfish: method is not used anymore

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -1358,6 +1409,11 @@ private void enforceKeyBindingIfProofProvided(VCIssuanceContext vcIssuanceContex
         if (proofs.getJwt() != null && !proofs.getJwt().isEmpty()) {
             validateProofs(vcIssuanceContext, ProofType.JWT);
         }
+
+        // Validate each attestation proof if present
+        if (proofs.getAttestation() != null && !proofs.getAttestation().isEmpty()) {
+            validateProofs(vcIssuanceContext, ProofType.ATTESTATION);
+        }
```
Captain-P-Goldfish, line 1389: I would say it makes sense to also move this into the class `Proofs`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/keybinding/AttestationProofValidatorFactory.java
```diff
@@ -17,28 +17,97 @@
...
+ * Factory for creating AttestationProofValidator instances with configurable trusted keys.
+ * Trusted keys are configured via realm attribute 'oid4vc.attestation.trusted_keys'.
+ *
  * @author <a href="mailto:Rodrick.Awambeng@adorsys.com">Rodrick Awambeng</a>
  */
 public class AttestationProofValidatorFactory implements ProofValidatorFactory {
 
+    private static final Logger logger = Logger.getLogger(AttestationProofValidatorFactory.class);
+
+    public static final String TRUSTED_KEYS_REALM_ATTR = "oid4vc.attestation.trusted_keys";
```
Captain-P-Goldfish: it would make sense to move this constant into `OID4VCIConstants`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/keybinding/AttestationProofValidatorFactory.java
```diff
@@ -17,28 +17,97 @@
...
+
+    /**
+     * Loads trusted keys from realm attributes.
+     */
+    private Map<String, JWK> loadTrustedKeysFromRealm(KeycloakSession session) {
+        RealmModel realm = session.getContext().getRealm();
+        if (realm == null) {
+            return Map.of();
+        }
+
```
Captain-P-Goldfish, line 93: this is only called from `AttestationValidatorUtil` and results eventually in a `VerificationException`. It is not quite easy to follow this logic in this case since you return an empty map here which will cause an exception eventually.

Maybe using the keys from the realm under `keycloakSession.keys()` as default?  Or merging these keys with the keys stored in the realm-attribute?

By the way. realm-attribute-fields are very limited in length. So we can store 2 or 3 keys at maximum in the attribute-field. If we need more keys this is certain to cause problems.

Captain-P-Goldfish (changes requested):
Nice changes.
I added some comments and thoughts to your changes.

forkimenjeckayang:
Ran tests again after updating code

oid4vci-funal

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/keybinding/AttestationProofValidatorFactory.java
```diff
@@ -22,28 +22,38 @@
...
 import com.fasterxml.jackson.core.type.TypeReference;
 import org.jboss.logging.Logger;
 
 /**
  * Factory for creating AttestationProofValidator instances with configurable trusted keys.
- * Trusted keys are configured via realm attribute 'oid4vc.attestation.trusted_keys'.
+ * Trusted keys are loaded from two sources:
+ * <ul>
+ *   <li>Realm session keys (default): All enabled keys from the realm's key providers</li>
+ *   <li>Realm attribute keys (override): Keys configured via realm attribute 'oid4vc.attestation.trusted_keys'</li>
```
Captain-P-Goldfish: I would even go one step further here because we have some implications here:
If we use the enabled keys the keys will also be provided via the .well-known/openid-configuration endpoint. I would say we do not want this in some usecases. So we should add an additional property into the realm-attributes that allows us to directly reference the keys  to use by their keyId regardless if enabled or not. WDYT?

server-spi-private/src/main/java/org/keycloak/constants/OID4VCIConstants.java
```diff
@@ -39,6 +39,7 @@ public final class OID4VCIConstants {
     // --- Keybinding/Credential Builder ---
     public static final String SOURCE_ENDPOINT = "source_endpoint";
     public static final String BATCH_CREDENTIAL_ISSUANCE_BATCH_SIZE = "batch_credential_issuance.batch_size";
+    public static final String TRUSTED_KEYS_REALM_ATTR = "oid4vc.attestation.trusted_keys";
```
Captain-P-Goldfish, line 42: This attribute should also be listed in the documentation `vc-issuer-configuration.adoc`

Captain-P-Goldfish (changes requested):

Captain-P-Goldfish (approved):
thanks for adressing all of my concerns. Looks good to me.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/signing/SdJwtCredentialSigner.java
```diff
@@ -46,6 +53,52 @@ public String signCredential(CredentialBody credentialBody, CredentialBuildConfi
...
+     *
+     * @param sdJwtCredentialBody The SD-JWT credential body to add x5c to
+     * @param signer              The signer context containing the certificate(s) for the signing key
+     */
+    private void addX5cHeader(SdJwtCredentialBody sdJwtCredentialBody, SignatureSignerContext signer) {
+        List<X509Certificate> certificateChain = signer.getCertificateChain();
+
+        if (certificateChain != null && !certificateChain.isEmpty()) {
+            List<String> x5cList = certificateChain.stream()
+                    .filter(cert -> cert != null)
```
tnorimat: This lambda can be written as follows. It is up to you to accept that.
```
                    .filter(Objects::nonNull)
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -81,9 +88,77 @@ public List<AuthorizationDetailsResponse> process(UserSessionModel userSession,
...
+        ClientModel client = clientSession != null ? clientSession.getClient() : null;
+        UserModel user = userSession != null ? userSession.getUser() : null;
+
+        if (client == null || user == null) {
+            return;
+        }
+
+        // Skip if we're in pre-authorized code flow (it already has an offer state that will be updated)
+        // Pre-authorized flow sets VC_ISSUANCE_FLOW note on the client session
+        if (clientSession != null) {
```
tnorimat: This condition is always true.
In Line 106, `client == null` if `clientSession == null`.
In Line 109, the condition is true, so we does not reach Line 115.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessor.java
```diff
@@ -81,9 +88,77 @@ public List<AuthorizationDetailsResponse> process(UserSessionModel userSession,
...
+                return;
+            }
+        }
+
+        CredentialOfferStorage offerStorage = session.getProvider(CredentialOfferStorage.class);
+
+        // Process all OID4VC authorization details to create offer states for each credential
+        for (AuthorizationDetailsResponse authDetail : authDetailsResponse) {
+            if (authDetail instanceof OID4VCAuthorizationDetailsResponse) {
+                OID4VCAuthorizationDetailsResponse oid4vcDetail = (OID4VCAuthorizationDetailsResponse) authDetail;
```
tnorimat: These lines can be written as follows:
```
            if (authDetail instanceof OID4VCAuthorizationDetailsResponse oid4vcDetail) {
```

tnorimat (commented):
@forkimenjeckayang Hello, I added some minor comments. Could you check them?

tnorimat (approved):
LGTM

VinodAnandan:
@mposolda Could you please review the this PR, both @tnorimat  and @Captain-P-Goldfish  approved it ?

mposolda (approved):
@forkimenjeckayang @tnorimat @Captain-P-Goldfish Thanks for all the updates and reviews!

PR #44441, opened 2025-11-24
Reorder translation keys for zh_Hans

jasonqsong:
Reorder the translation keys in zh_Hans messages to keep the key order align with en messages.

Closes #44440


ahus1 (approved):

ahus1:
I've merged this PR, but only found out too late that there was a pending Weblate translation PR pending. There are now merge requests that I'll need to resolve.

I suggest to avoid any reordering in the future, as this is likely to cause conflicts.

jasonqsong:
@ahus1 I'm sorry for all the chaos I've made. You can drop any zh_Hans changes from weblate. I have backup in my worktree.

ahus1:
@jasonqsong - no worries, I should be more careful. 

I found a button in Weblate to write the latest changes from Weblate to the repository... I'll try that one once the PR  #44372 is merged. I'll keep you posted. Weblate switched automatically to a read-only mode to keep the confusion at bay.

jasonqsong:
I looked into the reason why the conflict occurs. I tried to keep zh_Hans clean on Weblate this week and prepared for the reorder. The en messages removed two keys this week and the changes were brought to Weblate main for every languages. I should do a sanity check before opening this PR.

ahus1:
@jasonqsong - the conflict is now resolved, mostly by rolling back the changes in the file that lead to the conflicts. Weblate is also unlocked again since yesterday. 

Only when that happend I noticed that several English texts were added to the zh-hans files. This might be confusing for Weblate, and also not necessary as Keycloak will fall back to English automatically. It might also be a problem if we would define different fallbacks in the future based on the user's language preference. 

With that, I would recommend not to add any English texts to the zh-hans files.

jasonqsong:
@ahus1 Thank you again for solving the mess.

The reason is that I want to keep the key order in the message file. I'm afraid that those keys will be added to the end of file if I don't keep a placeholder. I expect these placeholders will be translated by the end of this month.

We will finish the rest of zh-hans translations with Weblate.

ahus1:
> The reason is that I want to keep the key order in the message file.

I hear you, still what you are asking for should likely be an add-on for Weblate. 

There is already an addon to sort the keys in the translation file alphabetically (although this is not what we would want here): https://docs.weblate.org/en/latest/admin/addons.html#format-the-java-properties-file

Maybe you can search in their GitHub project for similar issues, or create a new issue to ask for advice. It is a Python project. I haven't contributed to that project yet, but maybe you or someone else would have some Python skills for that. 

We're using the Weblate hosted edition. So once the plugin would exist, I can enable it. 

In the meantime, you could sort the files locally when necessary to do the comparisons on those temporary files. Note that you can also upload translation files, still I assume it will no honor the ordering, and will just import the translation.

PR #44446, opened 2025-11-25
Cache expression EvaluatorContext in the workflow component model's n…

### LINKED ISSUE #42961: [RLM] - Cache expressions using as a component note
### Description

Today, expressions are cached on the provider level, and that is not as efficient as caching at the component level, where entries are cached at the realm cache so that entries are kept across multiple invocations of the provider.

If we do not change our plans to rely on the component model of the workflow to store conditions and their config, we should use the notes in the workflow component.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

sguilhen:
…otes

Closes #42961


vramik (approved):
lgtm, thanks

pedroigor (dismissed):

pedroigor (dismissed):

pedroigor (approved):

PR #44454, opened 2025-11-25
Fix race condition in SAML DocumentBuilderFactory creation

### LINKED ISSUE #44438: Intermittent ConcurrentModificationException during SAML initialization causing status code 400 for clients
### Before reporting an issue

### Area

saml

### Describe the bug

When several requests trigger SAML-related initialization concurrently, a ConcurrentModificationException is occasionally thrown from the SAML module. The stack traces lead back to calls made during the creation of ThreadLocal `DocumentBuilder` instances.

### Version

latest

### Regression

### Expected behavior

Keycloak should start consistently without intermittent exceptions, even when many threads initialize their ThreadLocal `DocumentBuilder` instances at the same time.

### Actual behavior

Keycloak startup sometimes fails with a `ConcurrentModificationException` originating from the XML parsing components used by the SAML module. When this occurs, the client attempting to authenticate receives an `HTTP 400` response. The issue is intermittent and typically appears only during multi-threaded startup conditions.

### How to Reproduce?

A JUnit test that spawns multiple threads and repeatedly triggers ThreadLocal initialization of the XML `DocumentBuilder` can expose the issue. The test does not fail on every run—the bug is timing-dependent—but repeatedly running the test will at some point expose the bug where the `ConcurrentModificationException` is thrown, demonstrating that the problem exists.

### Anything else?

I will provide a PR with the failing test and a proposed fix.

msoderstr:
Closes #44438

### DIFF
--- a/saml-core/src/main/java/org/keycloak/saml/common/util/DocumentUtil.java
+++ b/saml-core/src/main/java/org/keycloak/saml/common/util/DocumentUtil.java
@@ -394,7 +394,7 @@ public class DocumentUtil {
      *
      * @return
      */
-    private static DocumentBuilderFactory getDocumentBuilderFactory() {
+    private synchronized static DocumentBuilderFactory getDocumentBuilderFactory() {
         boolean tccl_jaxp = SystemPropertiesUtil.getSystemProperty(GeneralConstants.TCCL_JAXP, "false")
                 .equalsIgnoreCase("true");
         ClassLoader prevTCCL = SecurityActions.getTCCL();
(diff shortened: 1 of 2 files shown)


ahus1 (approved):
Thank you for this fix and the test case. I slightly updated it to use double-locking with a volatile. 

Looking at the exception, I also found why this was happening, as the instance variable was filled too early after it wasn't yet fully initialized.

ahus1:
cc: @keycloak/core-clients for visibility - I'll take care of a backport.

rmartinc (approved):
Thanks @msoderstr for the PR! LGTM!

PR #44457, opened 2025-11-25
Fix ClassCastException on mixing AddressMapper with ClaimsMapper

### LINKED ISSUE #44455: ClassCastException on mixing AddressMapper with ClaimMapper
### Before reporting an issue

### Area

oidc

### Describe the bug

If the Keycloak internal AddressMapper is used in conjunction with a ClaimMapper that tries to add an additional attribute to the address-structure we get a ClassCastException here:

```java
public class OIDCAttributeMapperHelper {

private static void mapClaim(List<String> split, Object attributeValue, Map<String, Object> jsonObject, boolean isMultivalued) {
          ....
          else {
                @SuppressWarnings("unchecked")
                Map<String, Object> nested = (Map<String, Object>) jsonObject.get(component);

                if (nested == null) {
                    nested = new HashMap<>();
                    jsonObject.put(component, nested);
                }

                jsonObject = nested;
            }
        }
}
...
}
```

The reason is that the AddessMapper creates an object of `AddressClaimSet` that is added into the token, while the ClaimAttributeMapper is working with a `Map<String, Object>`.

```
2025-11-25 07:54:17,716 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-33) Uncaught server error: java.lang.ClassCastException: class org.keycloak.representations.AddressClaimSet cannot be cast to class java.util.Map (org.keycloak.representations.AddressClaimSet is in unnamed module of loader io.quarkus.bootstrap.runner.RunnerClassLoader @4bec1f0c; java.util.Map is in module java.base of loader 'bootstrap')
	at org.keycloak.protocol.oidc.mappers.OIDCAttributeMapperHelper.mapClaim(OIDCAttributeMapperHelper.java:345)
	at org.keycloak.protocol.oidc.mappers.OIDCAttributeMapperHelper.mapClaim(OIDCAttributeMapperHelper.java:317)
	at org.keycloak.protocol.oidc.mappers.OIDCAttributeMapperHelper.mapClaim(OIDCAttributeMapperHelper.java:278)
	at org.keycloak.protocol.oidc.mappers.UserAttributeMapper.setClaim(UserAttributeMapper.java:103)
	at org.keycloak.protocol.oidc.mappers.AbstractOIDCProtocolMapper.setClaim(AbstractOIDCProtocolMapper.java:159)
	at org.keycloak.protocol.oidc.mappers.AbstractOIDCProtocolMapper.transformUserInfoToken(AbstractOIDCProtocolMapper.java:80)
	at org.keycloak.protocol.oidc.TokenManager$3.applyMapper(TokenManager.java:821)
	at org.keycloak.protocol.oidc.TokenManager$3.applyMapper(TokenManager.java:818)
```

### Version

26.4.5 and before

### Regression

### Expected behavior

Both ClaimMappers should work together and create an extended address-object.
In our specific case we want to add a `type`-claim into the `address`-claim:

```json
"address": {
  ...,
  "type": "STRUCTURED"
}
```

### Actual behavior

An exception occurs and we get a 500 internal server error

### How to Reproduce?

see description

### Anything else?

_No response_

Captain-P-Goldfish:
fixes #44455

### DIFF
--- a/core/src/main/java/org/keycloak/representations/IDToken.java
+++ b/core/src/main/java/org/keycloak/representations/IDToken.java
@@ -332,14 +332,12 @@ public class IDToken extends JsonWebToken {
         this.phoneNumberVerified = phoneNumberVerified;
     }
 
+    @JsonProperty("address")
     public Map<String, Object> getAddressClaimsMap() {
         return address;
     }
 
-    public void setAddressClaimsMap(Map<String, Object> address) {
-        this.address = address;
-    }
-
+    @JsonIgnore
     public AddressClaimSet getAddress() {
         return Optional.ofNullable(address).map(a -> {
                            return JsonSerialization.mapper.convertValue(a, AddressClaimSet.class);
@@ -349,6 +347,13 @@ public class IDToken extends JsonWebToken {
 
     public void setAddress(Map<String, Object> address) {
         this.address = address;
+    }
+
+    @JsonIgnore
+    public void setAddress(AddressClaimSet address) {
+        this.address = Optional.ofNullable(address)
+                               .map(a -> JsonSerialization.mapper.convertValue(a, Map.class))
+                               .orElse(null);
     }
 
     public Long getUpdatedAt() {
--- a/core/src/main/java/org/keycloak/representations/IDToken.java
+++ b/core/src/main/java/org/keycloak/representations/IDToken.java
@@ -17,13 +17,14 @@
 
 package org.keycloak.representations;
 
+import java.util.Map;
+import java.util.Optional;
+
 import org.keycloak.TokenCategory;
+import org.keycloak.util.JsonSerialization;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import org.keycloak.util.JsonSerialization;
-
-import java.util.Map;
 
 /**
  * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/keycloak/representations/UserInfo.java
```diff
@@ -277,11 +277,11 @@ public void setPhoneNumberVerified(Boolean phoneNumberVerified) {
         this.phoneNumberVerified = phoneNumberVerified;
     }
 
-    public AddressClaimSet getAddress() {
+    public Map<String, Object> getAddress() {
```
mposolda: Is it possible to make sure that getter/setter are backwards compatible? Something similar like you did for `IDToken` perhaps might work here as well?

core/src/main/java/org/keycloak/representations/IDToken.java
```diff
@@ -328,11 +332,22 @@ public void setPhoneNumberVerified(Boolean phoneNumberVerified) {
...
+    }
+
+    public AddressClaimSet getAddress() {
+        return Optional.ofNullable(address).map(a -> {
+                           return JsonSerialization.mapper.convertValue(a, AddressClaimSet.class);
+                       })
+                       .orElse(null);
+    }
+
+    public void setAddress(Map<String, Object> address) {
```
mposolda, line 348: For backwards compatibility, it will be good to keep `setAddress(AddressClaimSet address)` as you did for the `getAddress()` . I guess it should be possible to make sure that setter can convert the `AddressClaimSet` to `Map` and then set to `address` variable?

Question: Should rather the `getAddressClaimsMap` getter method have the `@JsonIgnore` annotation? It can be great to avoid potential issues with `addressClaimsMap` claim to accidentally appear in the IDToken JSON somehow :-)

mposolda (commented):
@Captain-P-Goldfish Thanks for the PR! Added few comments inline.

core/src/main/java/org/keycloak/representations/IDToken.java
Captain-P-Goldfish: I have added the `@JsonIgnore` to the methods with the `AddressClaimSet` parameters because the native structure should be preferred as default. What do you think?

core/src/main/java/org/keycloak/representations/UserInfo.java
Captain-P-Goldfish: done

mposolda (approved):
@Captain-P-Goldfish That works as well, Thanks!

Issue #44433, opened 2025-11-24
Configure what headers and cookies should be masked in HTTP access log


mabartos:
### Description

It'd be good if users have the ability to configure what additional headers and cookies should be masked in HTTP access log. After closing this issue https://github.com/quarkusio/quarkus/pull/50672, we should be in a state where certain Keycloak-specific headers or cookies are masked by default and should not be overridden.

mabartos:
Duplicate of https://github.com/keycloak/keycloak/issues/43811

PR #44475, opened 2025-11-25
Mask certain HTTP headers in the HTTP access log

### LINKED ISSUE #43811: Mask certain HTTP headers in the HTTP access log
### Description

We support 3 log patterns/formats for printing out HTTP access logs - common(default), combined, or long.

The long pattern prints out even all headers with sensitive information like Authorization, Cookie,...

These headers should be masked by default. 

Related to: 
- https://github.com/keycloak/keycloak/issues/43156

mabartos:
- Closes #43811
-------------
- Maintain the `CustomAllRequestHeadersAttribute` on our own until Quarkus has the fix included (changes necessary for Keycloak are marked)
- It automatically omits the information about the `Authorization` HTTP header value, such as tokens.
- Certain Keycloak cookies values are also omitted from the log: https://github.com/keycloak/keycloak/pull/44475/files#diff-84b8e8a0ac3d785229a1f97fc214aa466ebfd3df27891b3313213b7dfb5d2ca0R27
- Follow-up task to provide the ability to specify additional masked headers and cookies: https://github.com/keycloak/keycloak/issues/44433
- If we accept to go this way, I'll create a follow-up issue to remove this workaround.

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/HttpAccessLogOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/HttpAccessLogOptions.java
@@ -1,5 +1,7 @@
 package org.keycloak.config;
 
+import java.util.List;
+
 public class HttpAccessLogOptions {
 
     public static final Option<Boolean> HTTP_ACCESS_LOG_ENABLED = new OptionBuilder<>("http-access-log-enabled", Boolean.class)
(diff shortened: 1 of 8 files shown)


quarkus/config-api/src/main/java/org/keycloak/config/HttpAccessLogOptions.java
```diff
@@ -20,4 +22,22 @@ public class HttpAccessLogOptions {
...
+            "KEYCLOAK_SESSION",
+            "AUTH_SESSION_ID_LEGACY",
+            "KEYCLOAK_IDENTITY_LEGACY",
+            "KEYCLOAK_SESSION_LEGACY"
+    );
+
+    public static final Option<List<String>> HTTP_ACCESS_LOG_MASKED_COOKIES = OptionBuilder.listOptionBuilder("http-access-log-masked-cookies", String.class)
+            .category(OptionCategory.HTTP_ACCESS_LOG)
+            .description("Set of HTTP Cookie headers whose values must be masked when the 'long' pattern or '%{ALL_REQUEST_HEADERS}' format is enabled with the 'http-access-log-pattern' option.")
+            .hidden() // hidden for now as we do not have the full Quarkus support for this yet
```
shawkins, line 40: Does it matter if there's quarkus support whether this is hidden?

Will this ever be set by a user? If so, we could consider making it additive to the default.

mabartos: It was rather for purposes to be on a safe side, but yes, it might not be hidden.

It's expected that some additives will be there: https://github.com/keycloak/keycloak/issues/44433

mabartos:
Put on hold, as we probably wait for Quarkus to have it included (needs to be discussed with them).

This PR can hang in here for cases it's not included in Quarkus in the foreseeable future.

shawkins:
> This PR can hang in here for cases it's not included in Quarkus in the foreseeable future.

Makes sense. One more thought - it would be safer to specify what headers / cookies can be shown vs. what should be masked.

mabartos:
Closing in favor of https://github.com/keycloak/keycloak/pull/45400

PR #44477, opened 2025-11-25
Add new option to schedule user session expiration

### LINKED ISSUE #44068: Add new option to schedule user session expiration
### Description

Add a new option to control how frequently expired entries should be checked for removal. This allows users to control the database space and how fast the notification should be sent. 

### Goals

Delete a user sessions `X` seconds after the user session timed out or reached its maximum lifetime

### Non-Goals

See parent task.

### Notes

_No response_

pruivo:
Closes #44068

- [x] Mention it in Notable Changes

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -41,7 +41,7 @@ As this has led to a hard-to-configure URL filtering, for example, in reverse pr
 
 To analyze rejected requests in the server log, enable debug logging for `org.keycloak.quarkus.runtime.services.RejectNonNormalizedPathFilter`.
 
-To revert to the previoius behavior and to accept non-normalized URLs, set the option `http-accept-non-normalized-paths` to `true`. With this configuration, enable and review the HTTP access log to identify problematic requests.
+To revert to the previous behavior and to accept non-normalized URLs, set the option `http-accept-non-normalized-paths` to `true`. With this configuration, enable and review the HTTP access log to identify problematic requests.
 
 // ------------------------ Notable changes ------------------------ //
 == Notable changes
@@ -115,7 +115,7 @@ This was only possible by granting the `admin` role to a master realm user, maki
 In this release, realm administrators with the `realm-admin` role can assign admin roles to users in their realm, allowing them to delegate administrative tasks without needing server admin privileges.
 
 If you are using FGAP to delegate administration to users in a realm other than the master realm,
-make sure the users granted with the `realm-admin` role are expected to have this role to avoid privilege scalation.
+make sure the users granted with the `realm-admin` role are expected to have this role to avoid privilege escalation.
 
 The documentation is also updated with additional information about the different types of realm administrators.
 For more information, see link:{adminguide_link}#_fine_grained_permissions[Delegating realm administration using permissions].
@@ -141,6 +141,15 @@ See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to 
 
 When enabling authorization on a new client (creating a Resource Server), Keycloak no longer automatically creates a "Default Resource", a "Default Policy," and a "Default Permission."
 
+=== New logic to delete expired sessions from the database
+
+New logic has been implemented to delete expired sessions from the database.
+Previously, a cluster-aware scheduled task triggered the removal, where only a single Keycloak instance would check and remove expired entries.
+
+In the new version, each Keycloak instance handles different realms.
+Additionally, a new SPI option has been added to control the frequency of the internal deletion task: `--spi-user-sessions--infinispan--session-expiration-period` (value in seconds).
+
+
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
 
@@ -162,6 +171,24 @@ The getter methods like `getSession()` have been deprecated in favor of the `ses
 The option `http-accept-non-normalized-paths` was introduced to restore the previous behavior where {project_name} accepted non-normalized URLs.
 
 As this behavior can be problematic for URL filtering, it is deprecated and will be removed in a future release.
+
+=== Deprecation of methods for removing expired authentication sessions from `AuthenticationSessionProvider`
+
+The methods `removeAllExpired()` and `removeExpired(RealmModel realm)` are annotated with the `@Deprecated` annotation.
+They have been deprecated since Keycloak 19, as the built-in implementations now use their own internal cleanup mechanisms.
+If you are providing a custom implementation of this provider, implement an internal mechanism to delete expired sessions.
+
+=== Deprecation of methods for removing expired user sessions from `UserSessionProvider`
+
+The methods `removeAllExpired()` and `removeExpired(RealmModel realm)` are deprecated.
+A new internal cleanup mechanism has been implemented to automatically remove expired sessions from the database.
+If you are providing a custom implementation of this provider, implement an internal mechanism to delete expired sessions.
+
+=== Deprecation of cluster scheduled task `ClearExpiredUserSessions`
+
+As `AuthenticationSessionProvider` and `UserSessionProvider` now have an internal mechanism to delete expired entries, the scheduled task `ClearExpiredUserSessions` has been deprecated.
+It is still triggered in this Keycloak version, but will be removed in a future release.
+
 
 // ------------------------ Removed features ------------------------ //
 == Removed features
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -139,16 +139,25 @@ See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to 
 
 === Default authorization policies and resources are no longer auto-created
 
-When enabling authorization on a new client (creating a Resource Server), Keycloak no longer automatically creates a "Default Resource", a "Default Policy," and a "Default Permission."
+When enabling authorization on a new client (creating a Resource Server), {project_name} no longer automatically creates a "Default Resource", a "Default Policy," and a "Default Permission."
 
-=== New logic to delete expired sessions from the database
+=== Deleting expired sessions from the database
 
-New logic has been implemented to delete expired sessions from the database.
-Previously, a cluster-aware scheduled task triggered the removal, where only a single Keycloak instance would check and remove expired entries.
+Previously, a cluster-aware scheduled task triggered the removal, where only a single instance would check and remove expired entries.
+Each {project_name} node triggered this by default every 15 minutes.
+This led to a situation that the more instances a cluster had, the more often this ran, although it would never run concurrently on two nodes at the same time.
 
-In the new version removal is handled by all Keycloak instances, with each instance handling the expiration of sessions in different realms
-Additionally, a new SPI option has been added to control the frequency of the internal deletion task: `--spi-user-sessions--infinispan--session-expiration-period` (value in seconds).
+Starting from this version, removal is handled by all {project_name} instances, with each instance handling the expiration of sessions in different realms.
+Additionally, a new SPI option has been added to control the frequency of the internal deletion task: `--spi-user-sessions--infinispan--session-expiration-period` (value in seconds). By default, this runs every three minutes.
+If you now add more instances to a clusters, the expiration is still run at the same interval.
 
+Previously, {project_name} deleted sessions from the database only once the session idle time including the remember-me idle time had passed.
+
+With the new setup and persistent user sessions enabled, sessions are deleted more timely based on the realm settings on session lifetime and idle time, including the remember me settings. For this, the table `OFFLINE_USER_SESSION` now has a new column `REMEMBER_ME`, that is filled for new sessions and updated incrementally for existing sessions.
+The deletion is triggered currently three minutes after the session has expired to allow for clock skew between instances. Note that this interval can change in future release as we optimize this functionality.
+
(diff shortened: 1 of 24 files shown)


model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
```diff
@@ -86,12 +89,16 @@ public class InfinispanUserSessionProviderFactory implements UserSessionProvider
     private static final boolean DEFAULT_USE_CACHES = true;
     public static final String CONFIG_USE_BATCHES = "useBatches";
     private static final boolean DEFAULT_USE_BATCHES = false;
+    public static final String CONFIG_EXPIRATION_PERIOD = "sessionExpirationPeriod";
+    private static final int DEFAULT_EXPIRATION_PERIOD_SECONDS = 180;
```
pruivo, line 93: The default in 26.4 is 15 min; 3 min seems a bit low to me. Opinions?

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -141,6 +141,15 @@ See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to
 
 When enabling authorization on a new client (creating a Resource Server), Keycloak no longer automatically creates a "Default Resource", a "Default Policy," and a "Default Permission."
 
+=== New logic to delete expired sessions from the database
+
+New logic has been implemented to delete expired sessions from the database.
+Previously, a cluster-aware scheduled task triggered the removal, where only a single Keycloak instance would check and remove expired entries.
+
+In the new version, each Keycloak instance handles different realms.
```
ryanemerson: Maybe we can make this a bit more detailed, e.g.:

"In the new version removal is handled by all Keycloak instances, with each instance handling the expiration of sessions in different realms".

model/storage-private/src/main/java/org/keycloak/services/scheduled/ClearExpiredUserSessions.java
```diff
@@ -25,7 +25,9 @@
 
 /**
  * @author <a href="mailto:sthorger@redhat.com">Stian Thorgersen</a>
+ * @deprecated to be removed without replacement. The providers are response to purge the expired entries themselves.
```
ryanemerson: "The providers are responsible for purging the expired entries themselves."

server-spi/src/main/java/org/keycloak/models/UserSessionProvider.java
```diff
@@ -161,17 +161,26 @@ UserSessionModel createUserSession(String id, RealmModel realm, UserModel user,
 
     /**
      * Remove expired user sessions and client sessions in all the realms
+     *
+     * @deprecated to be removed without replacement. The providers are response to purge the expired entries
```
ryanemerson: "The providers are responsible for purging the expired entries themselves."

server-spi/src/main/java/org/keycloak/models/UserSessionProvider.java
```diff
@@ -161,17 +161,26 @@ UserSessionModel createUserSession(String id, RealmModel realm, UserModel user,
...
     /**
-     * Removes expired user sessions owned by this realm from this provider.
-     * If this `UserSessionProvider` uses `UserSessionPersister`, the removal of the expired
-     * {@link UserSessionModel user sessions} is also propagated to relevant `UserSessionPersister`.
+     * Removes expired user sessions owned by this realm from this provider. If this `UserSessionProvider` uses
+     * `UserSessionPersister`, the removal of the expired {@link UserSessionModel user sessions} is also propagated to
+     * relevant `UserSessionPersister`.
      *
      * @param realm {@link RealmModel} Realm where all the expired user sessions to be removed from.
+     * @deprecated to be removed without replacement. The providers are response to purge the expired entries
```
ryanemerson: "The providers are responsible for purging the expired entries themselves."

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
ryanemerson: Any reason to change from 15?

ryanemerson (commented):
A few comments around the docs. I'm going to take a proper look at the code tomorrow with a fresh brain.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
pruivo: No specific reason, just a gut feeling that the database might become too large or that each expiration round could be slow since it has to go through more sessions to remove. With 3 minutes, I'm concerned it would be too small and hit the database too often, causing a performance impact.

ahus1 (dismissed):
Thank you for this extensive change. I especially the idea about distributing the tasks with consistent hashing: With this setup, we would also avoid the situation where the lock in the work cache expires, and another node would then run the task in parallel. 

I want this feature also for the regular clustered task runner and the LDAP sync 😄

ahus1:
When writing the docs, I realized that people when building logic on top of the `USER_SESSION_DELETED` event would really want it to fire every time. With that, I created #44513 as a follow-up to fire it also for invalid remember-me sessions.

pruivo:
> Thank you for this extensive change. I especially the idea about distributing the tasks with consistent hashing: With this setup, we would also avoid the situation where the lock in the work cache expires, and another node would then run the task in parallel.
> 
> I want this feature also for the regular clustered task runner and the LDAP sync 😄

You're welcome. I like the hashing approach too. If you want, I can convert the hashing code to a provider, so it can be used anywhere in the Keycloak code.


> When writing the docs, I realized that people when building logic on top of the `USER_SESSION_DELETED` event would really want it to fire every time. With that, I created #44513 as a follow-up to fire it also for invalid remember-me sessions.

Ah, ok. I commented on it as I read it first before reaching this message :sweat_smile:

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserSessionProviderFactory.java
ryanemerson: How about we start with a higher value and then re-asses based upon user feedback. Maybe 5 or 10 minutes?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/expiration/BaseExpirationTask.java
```diff
@@ -0,0 +1,151 @@
...
+        } catch (Throwable t) {
+            logUnexpectedErrorDuringDeletion(t);
+        } finally {
+            long duration = System.nanoTime() - start;
+            onTaskExecuted.accept(Duration.of(duration, ChronoUnit.NANOS));
+            log.debugf("PurgeExpired tasks completed in %s seconds", TimeUnit.NANOSECONDS.toSeconds(duration));
+        }
+    }
+
+    final int expiration() {
```
ryanemerson: Method never used

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/expiration/ExpirationTaskFactory.java
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Manually trigger the expiration task, bypassing any scheduling.
+     * <p>
+     * Testing purpose only! Do not invoke in production.
+     */
+    public static void manualTriggerTask(KeycloakSession session) {
+        getEventTask(session).ifPresent(BaseExpirationTask::purgeExpired);
+    }
+
+    public static int membersSize(KeycloakSession session) {
```
ryanemerson, line 119: JavaDoc is missing. The only current usage is in a test, do we want to flag this as for testing only or are we happy for users to rely on it?

ahus1 (dismissed):

ryanemerson:
I have disabled auto-merge, an additional commit and comments are still pending from me.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
```diff
@@ -362,10 +370,16 @@ public static void configureNumOwners(Config.Scope keycloakConfig, Configuration
      */
     public static ConfigurationBuilder getRemoteCacheConfiguration(String cacheName, Config.Scope config, String[] sites) {
         return switch (cacheName) {
-            case CLIENT_SESSION_CACHE_NAME, OFFLINE_CLIENT_SESSION_CACHE_NAME ->
-                    remoteCacheConfigurationBuilder(cacheName, config, sites, RemoteAuthenticatedClientSessionEntity.class);
-            case USER_SESSION_CACHE_NAME, OFFLINE_USER_SESSION_CACHE_NAME ->
-                    remoteCacheConfigurationBuilder(cacheName, config, sites, RemoteUserSessionEntity.class);
+            case CLIENT_SESSION_CACHE_NAME, OFFLINE_CLIENT_SESSION_CACHE_NAME -> {
+                var builder = remoteCacheConfigurationBuilder(cacheName, config, sites, RemoteAuthenticatedClientSessionEntity.class);
```
ryanemerson: Nitpick, but we can move the `configureSessionExpirationReader` call to inside `remoteCacheConfigurationBuilder` and add a boolean parameter to trigger this. That way we can keep the cases as single line statements without the need for `yield`.

testsuite/model/src/test/java/org/keycloak/testsuite/model/session/UserSessionPersisterProviderTest.java
```diff
@@ -1047,6 +1052,130 @@ public void testSessionExpirationBatch() {
...
+                barrier.await();
+
+                if (primary) {
+                    // persist sessions
+                    createSessions(sessionCount, value -> false);
+                }
+
+                barrier.await();
+                assertEquals(sessionCount, countPersistedSessionsForUser1());
+                if (InfinispanUtils.isRemoteInfinispan()) {
```
ryanemerson, line 1079: Is this test ever executed with a remote ISPN cluster or was this for local testing during dev?

ryanemerson (commented):
Thanks @pruivo! Overall LGTM, just a couple of minor comments. I have also pushed a new commit to update the JavaDocs.

testsuite/model/src/test/java/org/keycloak/testsuite/model/session/UserSessionPersisterProviderTest.java
pruivo: Yes, tests are run with an external ISPN cluster, with profiles
https://github.com/keycloak/keycloak/blob/18eeef7b261e3fa34074acbdb06bef2b6c2f770b/testsuite/model/pom.xml#L245
https://github.com/keycloak/keycloak/blob/18eeef7b261e3fa34074acbdb06bef2b6c2f770b/testsuite/model/pom.xml#L265

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/expiration/BaseExpirationTask.java
pruivo: Nice catch, I usually rewrite my code a couple of times until I'm happy with it, and this was a leftover.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/expiration/ExpirationTaskFactory.java
pruivo: Yes, this is for testing only. I added the Javadoc.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
pruivo: A `boolean` is too strict and does not allow to have different values for different caches. There is a chance to have persistence configured for the remaining unbound cache to prevent OOME. If it goes forward, I need different values for different caches. 
Let me see what I can do...

pruivo: I've added another commit; please let me know what you think.

ryanemerson: With the new changes the expiration reaper is enabled for all caches, whereas previously it was only applied to `CLIENT_SESSION_CACHE_NAME, OFFLINE_CLIENT_SESSION_CACHE_NAME, USER_SESSION_CACHE_NAME, OFFLINE_USER_SESSION_CACHE_NAME`.

Do we want the reaper on the other caches, if not should we wrap the expiration config in a `if (expirationWakeupPeriodMillis > 0)` check?

pruivo: The reaper is enabled by default, don't give me a heart attack :sweat_smile: 

https://github.com/infinispan/infinispan/blob/fc71f27d2b6f8fd251e14d918b0520e7464af708/core/src/main/java/org/infinispan/configuration/cache/ExpirationConfiguration.java#L17

ryanemerson (approved):

ahus1 (approved):

PR #44482, opened 2025-11-25
When joining a group, don't rely on cached values if user has already been updated

### LINKED ISSUE #44480: Wrong persistent group permissions when multiple group membership changes happen in the same request
### Before reporting an issue

### Area

core

### Describe the bug

When two IDP group mappers assign the same group and one decides to map the group and the other doesnt, there is currently a condition where sometimes the user is in the group and sometimes it is not. This is due to the following change: https://github.com/keycloak/keycloak/commit/bb9015a1f26ada048610e8b89f34c0f3959ac9a9 / #43682
Where `joinGroup()` is missing the condition `&& updated == null` causing it to miss other updates to the user done in the same request.

### Version

26.4.5

### Regression

### Expected behavior

If two group mappers map the same group and one wants to join and one wants to leave, the join always wins.

### Actual behavior

Race condition, sometimes user joins, sometimes he doesn't, depending on mapper execution order.

### How to Reproduce?

Skipping as discussed with @ahus1 

### Anything else?

_No response_

ahus1:
Closes #44480

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/UserAdapter.java
@@ -485,7 +485,7 @@ public class UserAdapter implements CachedUserModel {
 
     @Override
     public void joinGroup(GroupModel group) {
-        if (group.getType() == Type.REALM && cached.getGroups(keycloakSession, modelSupplier).contains(group.getId())) {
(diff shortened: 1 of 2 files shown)


ahus1:
@pedroigor  / @pedroigor - can one of you please review this change? It is unfortunately a regression that I added in 26.4.4. Thanks!

pedroigor (approved):

PR #44485, opened 2025-11-26
Better events for jwt-bearer and check all details in the tests

### LINKED ISSUE #44137: Events for jwt-bearer grant
### Description

Make sure that:
- There are proper events triggered during invoking of jwt-bearer grant. Especially make sure there is single event for a request, which is either success or failure
- Make sure that all relevant data are available in the event. See standard token-exchange event for the inspiration. In addition to standard-token-exchange, this event might have some additional data related to identity providers (EG. identity-provier used, identity-provider subject from the token, maybe scopes from the assertion JWT etc)
- In case of error, make sure that error event has all the sensible data (EG. reason of the error in case of failed validations).
- Make sure events are tested in the related automated tests

rmartinc:
Closes #44137

PR that adds some details to the events in the `jwt-bearer` grant. Reading the issue, I just missed the scopes in the assertion but they are not added in any other place (if you think that it is important just let me know). The details added are the issuer and the subject of the assertion when the JWT is parsed, and the IdP alias once located by keycloak. The events were already checked in the tests, but the details are better checked now.

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/events/Details.java
+++ b/server-spi-private/src/main/java/org/keycloak/events/Details.java
@@ -39,7 +39,9 @@ public interface Details {
     String AUTH_TYPE = "auth_type";
     String AUTH_METHOD = "auth_method";
     String IDENTITY_PROVIDER = "identity_provider";
+    String IDENTITY_PROVIDER_ISSUER = "identity_provider_issuer";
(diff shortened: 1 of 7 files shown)


graziang (approved):
@rmartinc thanks!

mposolda (approved):
@rmartinc @graziang Thanks for the PR and additional review!

PR #44487, opened 2025-11-26
Use central method to create a DocumentBuilder for SAML

### LINKED ISSUE #44486: Use central method to create a DocumentBuilder for SAML
### Description

For SAML, we're using a central instance for the `DocumentBuilder`. 

We should use it in all places to avoid false-positives about missing XML safety parameters - see #44476

ahus1:
Closes #44486


rmartinc (approved):
Thanks @ahus1! I have checked that all the rest of usages are in tests. So we are OK with this PR.

mposolda (approved):
Approving on behalf of @rmartinc review.
@ahus1 @rmartinc Thanks!

PR #44496, opened 2025-11-26
feat: Keycloak as a Windows service

### LINKED ISSUE #37704: Support for running Keycloak as a Windows Service
### Description

Add support for running Keycloak natively as a Windows service. Focus on a good OOTB experience with minimal manual steps.

### Value Proposition

Leverage all the benefits of Windows services (start at boot, lifecycle management etc.) natively with Keycloak.

### Goals

* Easy to install OOTB experience to run Keycloak as a Windows service.
* Possibility to provide Keycloak configuration to the service in a user friendly way that is aligned with current configuration scheme.

### Non-Goals

* Changes to Keycloak codebase. The service should be just a wrapper around existing Keycloak distribution.

### Discussion

_No response_

### Notes

_No response_

Pepo48:
Closes: https://github.com/keycloak/keycloak/issues/37704

Signed-off-by: Peter Zaoral <pepo48@gmail.com>


Pepo48:
A couple of points that I would like to discuss before this PR gets out of the draft state:
* **A way of `prunsrv.exe` distribution?** (currently the dist does not contain the binary that is needed for OOTB experience; a user needs to download it first, as it's described in the readme.
* **A new guide?** I think this was briefly mentioned on one of our meetings - do we want to make this part of our guides or the current readme is enough?
*  **A cost of testing the service lifecycle** - I was able to cut it down, yet the test execution in its current form still takes around 40 seconds on my Windows machine (originally it was around 2 minutes). For now the test has the `WIN` tag, but if the cost will be too high, we can certainly remove it.

shawkins:
Is there anyway to just reuse kc.bat more? It seems like if / when we want to do something like allowing kc home dir to be set, or other top level changes, we'll need to do that in two .bat scripts instead of one.

Pepo48:
> Is there anyway to just reuse kc.bat more?

Should be doable. I had similar idea initially, but given the complexity, I kept it separate.

quarkus/dist/src/main/README.md
```diff
@@ -25,4 +25,134 @@ on Windows:
 
 After the server boots, open http://localhost:8080 in your web browser. The welcome page will indicate that the server is running.
 
-To get started, check out the [configuration guides](https://www.keycloak.org/guides#server).
\ No newline at end of file
+To get started, check out the [configuration guides](https://www.keycloak.org/guides#server).
+
+## Keycloak as a Windows Service
```
vmuzikar: Instead of a README file, we should create a new guide. Incl. instructions to download `prunsrv.exe`.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Service.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.keycloak.quarkus.runtime.cli.command;
+
+import picocli.CommandLine.Command;
+
+@Command(name = Service.NAME,
+        description = "Manage Keycloak as a Windows service.",
+        subcommands = {ServiceInstall.class, ServiceUninstall.class})
+public class Service {
```
vmuzikar: Could we hide the `kc.bat service` command on non-Windows hosts?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/ServiceInstall.java
```diff
@@ -0,0 +1,304 @@
...
+            description = "The user account the service should run as. Defaults to LocalSystem.")
+    String serviceUser;
+
+    @Option(names = "--service-password",
+            description = "The password for the service user account.")
+    String servicePassword;
+
+    @Option(names = "--classpath-jar",
+            description = "Path to quarkus-run.jar if auto-detection fails.")
+    String classpathJar;
```
vmuzikar: Do we need this?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/WindowsServiceInstall.java
```diff
@@ -0,0 +1,304 @@
...
+            description = "Maximum Java heap size in MB.")
+    Integer jvmMx;
+
+    @Option(names = "--service-user",
+            description = "The user account the service should run as. Defaults to LocalSystem.")
+    String serviceUser;
+
+    @Option(names = "--service-password",
+            description = "The password for the service user account.")
+    String servicePassword;
```
vmuzikar, line 92: We should provide also alternative ways to provide the password. At least an env var.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/ServiceInstall.java
```diff
@@ -0,0 +1,304 @@
...
+            "For production deployments, run 'kc.bat build' first to optimize the server."
+        },
+        footerHeading = "Examples:",
+        footer = { "  Install with default settings:%n%n"
+                + "      $ ${PARENT-COMMAND-FULL-NAME:-$PARENTCOMMAND} ${COMMAND-NAME}%n%n"
+                + "  Install for production mode:%n%n"
+                + "      $ ${PARENT-COMMAND-FULL-NAME:-$PARENTCOMMAND} ${COMMAND-NAME} --kc-args=\"start --http-port=8080\"%n%n"
+                + "  Install with custom service name and memory settings:%n%n"
+                + "      $ ${PARENT-COMMAND-FULL-NAME:-$PARENTCOMMAND} ${COMMAND-NAME} --name=my-keycloak --jvm-ms=512 --jvm-mx=1024%n"})
+public class ServiceInstall extends AbstractCommand {
```
vmuzikar: A potential follow-up might be a wizard mode, i.e. the command would prompt the user for any args that they didn't specify as an arg.

vmuzikar (commented):
@Pepo48 Thank you for the PR. I think it's a good start.

vmuzikar:
> * **A way of `prunsrv.exe` distribution?** (currently the dist does not contain the binary that is needed for OOTB experience; a user needs to download it first, as it's described in the readme.

We will document where to download it.

> * **A cost of testing the service lifecycle** - I was able to cut it down, yet the test execution in its current form still takes around 40 seconds on my Windows machine (originally it was around 2 minutes). For now the test has the `WIN` tag, but if the cost will be too high, we can certainly remove it.

40 seconds is fine.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Main.java
```diff
@@ -67,6 +67,7 @@
                 Import.class,
                 ShowConfig.class,
                 Tools.class,
+                Service.class,
```
shawkins: Unless there's a reason for users to still be able to call the service command when isWindows returns false, we can also consider removing or adding conditionally in Picocli.createCommandLine

```
if (!windows) {
  spec.subcommands().remove("service");
}
```

or not even include the Service class in the Main annotation, then add it conditionally. This is inline with a small optimization that may be done in the future to only initially add subcommands if their command name appears in the argument list.

Pepo48:
@vmuzikar @shawkins as I missed this week's weekly, let me say a couple of words here.

After the discussion and feedback I got from you, I decided to revisit prunsrv's exe mode that is supposed to address limitations of the standard Java (the native communication with Windows Service Control Manager is not possible, [here's a test and explanation](https://github.com/keycloak/keycloak/issues/37704#issuecomment-2894145577)). I tried the Prunsrv's exe mode in the past, but I experienced some issues with the service lifecycle management, therefore I stuck with JVM mode initially.

The exe mode Starts an external process - it does not embed or host anything itself. It then monitors that external process (for start, stop, exit codes, etc.) just as WSCM requires from a service. That could solve all the obstacles we discussed (respecting all the kc.bat options, env vars, keycloak.conf, auto-augmentation mechanism and reduce the complex logic around JVM that initial implementation had (in the JVM mode prunsrv loads the JVM inside its own process via the jvm.dll and executes everything inside it). So I gave the exe mode one more shot.

I simplified the logic and made it to work in the end. All the things I mentioned above are now available as it acts almost like a generic Windows service. 

There are still some specifics, e.g. with env vars - a user ideally needs to declare vars as system-wide ones, since Windows services do not inherit user variables. I described all the limitations I've found and in the troubleshooting section of the new guide.

I still have couple of questions/challenges that I would like to discuss with you:
* Do we want to give users option to run the service in the development mode? For now, the production mode is the hardcoded default.
* The lifecycle test is currently failing in CI because of missing prunsrv.exe binary. Should the test download it automatically or can I place the binary to the test resources?
* I also tried to address all the suggestion from Vašek's review. That includes hiding platform specific commands, which required an addition to Picocli and HelpCommandDistTest, which now tries to detect, whether a command is hidden and if so, it changes the expected help template structure on the fly. I wanted to make it generic without a need for a separate set of templates. Hopefully I haven't overengineered it. 

In any case, please, let me know :). Thank you!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Main.java
Pepo48: Thanks Steve. I missed your suggestion before posting the summary below. I introduced a mechanism for hiding the commands (see hidePlatformSpecificCommands in Picocli), but as I already mentioned, there is a high chance that it's needlessly overengineered as I wanted to make it generic.

Regardless of the approach, HelpCommandDistTest would still require some changes, right? (unless we want to introduce a separate set of templates)

shawkins: > Regardless of the approach, HelpCommandDistTest would still require some changes, right? (unless we want to introduce a separate set of templates)

That is correct - we could also consider not even running them on windows.

shawkins: If a system property or environment variable also controlled the presense of the service command the HelpCommandDistTest could selectively turn it off.

Alternatively if it's easy to just pass all command options from one .bat to another, we could have a kc-service.bat script set a  system property or environment variable to then call kc.bat - then normal usage of kc.bat won't expose the service command.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/WindowsServiceUninstall.java
vmuzikar, line 1: Potential issue with this is that it can delete services that were not created by the `kc.bat`. But that should be fine, if the user has permissions to execute `prunsrv.exe` through `kc.bat`, they can do that directly as well.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Service.java
```diff
@@ -0,0 +1,38 @@
...
+package org.keycloak.quarkus.runtime.cli.command;
+
+import org.keycloak.quarkus.runtime.Environment;
+
+import picocli.CommandLine.Command;
+
+@Command(name = Service.NAME,
+        description = "Manage Keycloak as a Windows service.",
+        subcommands = {ServiceCreate.class, ServiceDelete.class})
+public class Service {
```
vmuzikar: One thing that I'm wondering now is whether we should create one more subcommand `windows`, so it would be `kc.bat service windows create/delete`. That would allow us to add stuff like `kc.sh service systemd` etc. if needed, or even multiple service types on a single OS if that was ever supported.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -627,9 +628,24 @@ public <K> K create(Class<K> cls) throws Exception {
...
+     * Hides platform specific commands that should not be shown on certain platforms
+     */
+    private void hidePlatformSpecificCommands(CommandLine cmd) {
+        if (Service.shouldBeHidden()) {
+            CommandLine serviceCmd = cmd.getSubcommands().get(Service.NAME);
+            if (serviceCmd != null) {
+                serviceCmd.getCommandSpec().usageMessage().hidden(true);
+            }
+        }
+    }
```
vmuzikar: I wonder if we could make this generic. E.g. move `shouldBeHidden` to some abstract parent command class?

WDYT @shawkins?

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
```diff
@@ -223,6 +227,60 @@ private void assertHelp(CLIResult cliResult) {
...
+                }
+                for (CommandLine subSub : sub.getSubcommands().values()) {
+                    for (String name : subSub.getCommandSpec().names()) {
+                        regexes.add("(?m)^    " + Pattern.quote(name) + ".*\\R");
+                    }
+                }
+            }
+        }
+        return regexes;
+    }
```
vmuzikar: This might be prone to errors. We're basically checking if expected commands are present based on if they are not hidden in the code. I.e. if we hid a command by accident, the tests would pass.

Maybe we could take a different approach. What if we prefixed all the OS specific lines in the expected help txts with some reserved word (like `OS:WIN`?). Then based on the OS we'd either remove those lines, or remove the prefix, and the do the comparison actual vs expected. WDYT?

quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/dist/approvals/cli/help/HelpCommandDistTest.testDefaultToHelp.approved.txt
```diff
@@ -27,6 +27,9 @@ Commands:
   show-config             Print out the current configuration.
   tools                   Utilities for use and interaction with the server.
     completion            Generate bash/zsh completion script for kc.sh.
+  service                 Manage Keycloak as a Windows service.
+    create                Create a Keycloak Windows service.
+    delete                Delete Keycloak Windows service.
```
vmuzikar: This would be something like:
```suggestion
OS:WIN  service                 Manage Keycloak as a Windows service.
OS:WIN    create                Create a Keycloak Windows service.
OS:WIN    delete                Delete Keycloak Windows service.
```

vmuzikar (commented):
@Pepo48 Thank you for the update!

vmuzikar:
> * Do we want to give users option to run the service in the development mode? For now, the production mode is the hardcoded default.

I'd say no. Service implies prod mode.

> * The lifecycle test is currently failing in CI because of missing prunsrv.exe binary. Should the test download it automatically or can I place the binary to the test resources?

I'd download it and cache it in GHA. That way we'd be always testing with the latest prunsrv.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: See https://github.com/keycloak/keycloak/pull/44496#discussion_r2611728386 for some additional thoughts on hiding vs simply not having the command.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Service.java
shawkins: @vmuzikar can you explain more about what a systemd or unixd service call would do? Are you thinking it would spit out a systemd file somewhere? https://github.com/keycloak/keycloak/discussions/10180#discussioncomment-2499174

vmuzikar: > Are you thinking it would spit out a systemd file somewhere?

Yes, exactly. :)

shawkins: I'm not familar with that being a common feature.

vmuzikar: Well, in any case I'd like to design the command in a way that we can add other types of services in the future if needed.

shawkins: > Well, in any case I'd like to design the command in a way that we can add other types of services in the future if needed.

Which of the following makes the most sense:

`service type ...`

`win-service ...`

Or we can use the existing `tools` command, which currently only has completion subcommand:

`tools win-service ...`

vmuzikar: > Or we can use the existing `tools` command, which currently only has completion subcommand:
> 
> `tools win-service ...`

That sounds interesting, I like that. If we ever needed that, we could add something like `tools systemd` or whatever. @Pepo48 WDYT?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/WindowsServiceUninstall.java
Pepo48: Moreover, for the service management a user need the elevated rights (a precaution required by the system itself). Therefore it is sort of expected that user knows their stuff.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/Service.java
Pepo48: Addressed per suggestion.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
Pepo48: Replaced this just with:
```
  output = output.replaceAll("(?m)^ {4}win-service\\s+.*\\R", "");
  output = output.replaceAll("(?m)^ {6}create\\s+Create a Keycloak Windows service\\.\\R", "");
  output = output.replaceAll("(?m)^ {6}delete\\s+Delete Keycloak Windows service\\.\\R", "");
```
no reflection needed, it's part of the assertHelp, among other Windows related adjustments following the same regex pattern.

Is it more acceptable?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
Pepo48: My final suggestion differs a bit from what I mentioned during the meeting, but basically there is no hiding logic, I reverted the filtering changes in SubCommandListRenderer. The command either exists or it doesn't, based on the platform. This way the command tree is cleaner, the commands can't be accidentally hidden and the test logic is simpler (see comment below); inspired by Steve's first suggestion.

Pepo48:
> I'd download it and cache it in GHA. That way we'd be always testing with the latest prunsrv.

Addressed per suggestion

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
shawkins: Thank you @Pepo48, those changes look good.

docs/guides/server/windows-service.adoc
```diff
@@ -0,0 +1,174 @@
...
+[source,bash]
+----
+bin\kc.bat tools win-service create --startup=manual --stop-timeout=60
+----
+
+Delayed auto-start:
+
+[source,bash]
+----
+bin\kc.bat tools win-service create --startup=delayed
```
shawkins: The current warning for unnecessary cli options is limited to runtime variables only. However for the tools commands that is not correct.

If you do something like: `kc.bat tools win-service create --startup=delayed --db=oracle`

There won't be a warning that --db=oracle is not used.

We should either handle that warning or for commands that don't accept either buildtime or runtime options, just error when unrecognized args are provided.

This of course can be done in a follow-up.

shawkins (commented):
This is looking fantastic @Pepo48 !

Just wanted to understand better, where does the user set environment variables that will be picked up by the start? This will be the only way to manipulate things like the jvm settings right?

Pepo48:
> where does the user set environment variables that will be picked up by the start?

On the system level - as a system wide properties (the reason is described in the guide):
image
Of course, you can achieve the same via Command-line or PowerShell.

> This will be the only way to manipulate things like the jvm settings right?

That's correct, based on our initial discussions I ditched the idea of exposing `--env` param, where a user could pass vars directly during the service creation. Procrun itself offers `Environment` parameter that could be exposed in a similar fashion like `DependsOn` is. You may see that in the first commit I used more of these, but we agreed to keep it rather simpler, so I kept only a small subset of available params (the complete list can be find [here](https://commons.apache.org/proper/commons-daemon/procrun.html)).

shawkins:
> That's correct, based on our initial discussions I ditched the idea of exposing `--env` param, where a user could pass vars directly during the service creation. Procrun itself offers `Environment` parameter that could be exposed in a similar fashion like `DependsOn` is. You may see that in the first commit I used more of these, but we agreed to keep it rather simpler, so I kept only a small subset of available params (the complete list can be find [here](https://commons.apache.org/proper/commons-daemon/procrun.html)).

Makes perfect sense, thank you.

docs/guides/server/windows-service.adoc
Pepo48: @shawkins if I read the code the right way, it happens because it checks if PropertyMappers.getMapperByCliKey("--db") returns a mapper and since `--db` is a valid option, it is removed from unrecognized args?

In that case should I add something like this to Picocli?
image
This way such options should not be removed from unrecognized args. But just an idea, I haven't tried it yet.

Anyway, it probably touches more that I would want to, at this time of the year :). So you are right with the follow-up.

shawkins: Roughly like that. In double-checking the 26.4 behavior - it's a little different than I thought. completion for example by-passes the unrecognized arg check altogether. As does show-config, because it skips the validateConfig call. 

26.5 has already changed that a little as show-config args will get the intial validation.

 > Anyway, it probably touches more that I would want to, at this time of the year :). So you are right with the follow-up.

Yes, I'll open an issue.

shawkins: https://github.com/keycloak/keycloak/issues/44975

stianst (commented):
I'd would change the command to `windows-service` instead of the shorter `win-service`. It's less ambiguous and more clear to the user. It's also not something that is super frequently used, so doesn't matter if it's a bit longer.

You'd also wouldn't want to shorten `systemd-service` when/if that is added so would be more consistent.

Finally, I'd argue that you are installing/uninstalling the service, not creating and deleting. So, think `tools windows-service install` is more clear than `tools windows-service create`.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -627,9 +629,27 @@ public <K> K create(Class<K> cls) throws Exception {
...
+
+        removePlatformSpecificCommands(cmd);
+
         return cmd;
     }
 
+    /**
+     * Removes platform-specific commands on non-applicable platforms
+     */
+    private void removePlatformSpecificCommands(CommandLine cmd) {
```
vmuzikar, line 641: This is quite specific to that single command but considering it's currently just a single command, we don't need anything more generic.

.github/actions/prunsrv-setup/action.yml
```diff
@@ -0,0 +1,54 @@
+name: Setup Apache Commons Daemon (Procrun)
+description: Download and cache Apache Commons Daemon prunsrv.exe for Windows service tests
+
+inputs:
+  version:
+    description: Apache Commons Daemon version
+    required: false
+    default: "1.5.1"
```
vmuzikar: Just a nit. Might be good to resolve this dynamically and use the latest, so we know if something breaks. But we can tackle that as a follow-up. Note that we don't tell the users which version to install, which implicitly says "latest" I believe.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
```diff
@@ -213,16 +212,23 @@ private void assertHelp(CLIResult cliResult) {
...
             // On Windows, all output should have at least one "kc.bat" in it.
             MatcherAssert.assertThat(output, Matchers.containsString("kc.bat"));
             output = output.replaceAll("kc.bat", "kc.sh");
             output = output.replaceAll(Pattern.quote("data\\log\\"), "data/log/");
             // line wrap which looks differently due to ".bat" vs. ".sh"
             output = output.replaceAll("including\nbuild ", "including build\n");
+            // win-service command only present on Windows, remove for cross-platform comparison
+            output = output.replaceAll("(?m)^ {4}win-service\\s+.*\\R", "");
+            output = output.replaceAll("(?m)^ {6}create\\s+Create a Keycloak Windows service\\.\\R", "");
+            output = output.replaceAll("(?m)^ {6}delete\\s+Delete Keycloak Windows service\\.\\R", "");
```
vmuzikar: Should we update the help test txt files too? Tests pass, it seems we don't run `HelpCommandDistTest` on Win?

docs/guides/server/windows-service.adoc
```diff
@@ -0,0 +1,174 @@
...
+<#import "/templates/links.adoc" as links>
+
+<@tmpl.guide
+    title="Run {project_name} as a Windows Service"
+    summary="Install and run {project_name} as a Windows service using Apache Commons Daemon.">
+
+This guide explains how to install and run {project_name} as a Windows service using Apache Commons Daemon. The service runs `kc.bat` in "exe" mode so behavior matches running `kc.bat start` manually. The service runs in "exe" mode, where Procrun executes `kc.bat start` as an external process. Environment variables, such as `KC_*`, along with `conf/keycloak.conf`, are respected. The `kc.bat` script handles augmentation and build logic, ensuring the service behaves exactly like a manual start.
+
+==  Apache Commons Daemon Setup
+To run {project_name} as a Windows service, you need the Apache Commons Daemon Procrun binary (`prunsrv.exe`). Download it for your platform: https://downloads.apache.org/commons/daemon/binaries/windows/
```
vmuzikar, line 11: Maybe we could say something like:
```suggestion
To run {project_name} as a Windows service, you need the Apache Commons Daemon Procrun binary (`prunsrv.exe`). Download version 1.5.1 for your platform: https://downloads.apache.org/commons/daemon/binaries/windows/
```

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
Pepo48: Well, I opted out for not changing the templates and just strip out the output on Windows. There is a case, when service could accidentally get lost and we wouldn't catch it. In the latest commit, I just flipped the logic - edited the templates, which are then stripped out on the non-Windows platforms. This is possibly a better approach, that should cover some edge cases.

Is this where you was heading with this?

.github/actions/prunsrv-setup/action.yml
Pepo48: I tried to come up with something - PTAL.

vmuzikar: Awesome, thanks!

Pepo48:
> Finally, I'd argue that you are installing/uninstalling the service, not creating and deleting. So, think `tools windows-service install` is more clear than `tools windows-service create`.

@stianst the nomenclature have been discussed (in the original PoC I called it `service install/uninstall`), but then we decided to change it to create/delete and I did not oppose the change, simply because when it comes to Windows services, it's a [standard way of naming things](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/sc-create). Just to explain the final choice.

In reality, I don't have a strong opinion, so I just renamed it back, as per your suggestion.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
Pepo48: @vmuzikar @shawkins I revamped quite a bit. With the original setup, I didn’t have full control over what was being read and compared unless I used the hacks I initially tried, which we later agreed were too fragile.

Assuming that ideally we want to retain the ApprovalTests library in place, there are two options I see:
* to use our own `Approvals.verify(ApprovalApprover approver)`
* to use our own `Approvals.verify(String, Options.withComparator(...)) ` [1]

I went for the latter - it seems to suit our use case better and, based on the examples, is also easier to implement. The only thing I’m slightly concerned about is that the custom comparator functionality is available starting with version 18.5.0 (Oct 2022), while we are currently using version 14.0.0 (Feb 2022). Can we upgrade without concern? As you can see in the last commit, I already tried it with the latest version (26.1.0) - it was fairly easy to use and implement and the code looks a bit cleaner to me.

I guess it is also a good foundation for the upcoming systemd service.

One more thing - I removed `REPLACE_EXPECTED` constant as they provide their own mechanisms (I added links to quarkus/README).

I already tested it on Mac and Windows, the tests were green, the outputs looked as expected. Let me know what you think.

Thank you!

[1] https://github.com/approvals/ApprovalTests.Java/blob/master/approvaltests/docs/reference/Options.md#custom-comparators

Pepo48: I noticed that bunch of test failed in CI due to a trailing space after the "Usage: " sequence. This may be a long standing (most likely CI) bug that has been hidden from us, because the `HelpCommandDistTest` haven't been enabled for GH Windows runners. At least I don't see anything like that on my local Windows machine, the test is passing. ApprovalsTests seems to ignore line endings by default.

Let me investigate if there's a connection with the changes.
Edit: probably not, the same failures appeared earlier today, even before I implemented a custom comparator.

Pepo48: Unfortunately, I could not solve this, for now I reverted the change that enables HelpCommandDistTest on the Windows runners. It's out of the scope of this PR anyway, so we will sort it out in a follow-up.

Pepo48: https://github.com/keycloak/keycloak/issues/45044

vmuzikar:
@shawkins Do you want to give it another look too?

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
```diff
@@ -39,22 +36,21 @@
...
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.OS;
 
 import static org.keycloak.quarkus.runtime.cli.command.AbstractAutoBuildCommand.OPTIMIZED_BUILD_OPTION_LONG;
 
 @DistributionTest
 @RawDistOnly(reason = "Verifying the help message output doesn't need long spin-up of docker dist tests.")
 public class HelpCommandDistTest {
 
-    public static final String REPLACE_EXPECTED = "KEYCLOAK_REPLACE_EXPECTED";
```
vmuzikar, line 56: Just wanted to highlight we got rid of this dev feature that allowed replacing the expected help output with the actual output.

This works for me as there's no good way to do that in an automated way anyway, now that we have Windows specific lines there. We can perhaps improve on that to accommodate for Windows and reintroduce `KEYCLOAK_REPLACE_EXPECTED` as a follow-up.

shawkins (approved):
Seems good to me

vmuzikar (approved):
@Pepo48 LGTM, thank you. Very nice work!

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
shawkins: > reintroduce KEYCLOAK_REPLACE_EXPECTED as a follow-up.

Yes, I'd vote for reintroducing it. I wouldn't mind adding handling in picocli for an environment variable (could just be KEYCLOAK_REPLACE_EXPECTED) that exposes all commands regardless of the OS. Then we can separate targeted tests that just ensure the os specific commands aren't there for the wrong OS.

vmuzikar: Captured in https://github.com/keycloak/keycloak/issues/45044#issuecomment-3675906877 which will likely have to capture the OS specifics of the help test.

PR #44499, opened 2025-11-26
Be more explicit when executing individual tests for Admin UI tests

mabartos:
It'll help users to be able to run specific Admin UI tests from the CLI.


stianst (commented):
You don't need to specify the full path to the test; just the filename is sufficient.

mabartos:
@stianst Thanks, updated.

stianst (approved):

PR #44504, opened 2025-11-26
[client-api-v2] Use embedded mode for tests in IDE

mabartos:
Fixing tests or experimenting with the new functionality for Client API v2 is quite hard as after every change made to the feature, you need to rebuild the distribution in order to see the changes. 

I'd propose to start the embedded Quarkus server when executing tests in IDE, and full distribution (default) when executing via Maven. 

@shawkins @vmuzikar @Pepo48 Could you please check it? Thanks!

@vaceksimon @stianst @lhanusov Could you please check it? And would you accept these changes as default behavior even for your framework?

### DIFF
--- a/rest/admin-v2/tests/pom.xml
+++ b/rest/admin-v2/tests/pom.xml
@@ -45,6 +45,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <systemPropertyVariables>
+                        <kc.test.server>distribution</kc.test.server>
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                         <java.util.concurrent.ForkJoinPool.common.threadFactory>io.quarkus.bootstrap.forkjoin.QuarkusForkJoinWorkerThreadFactory</java.util.concurrent.ForkJoinPool.common.threadFactory>
                     </systemPropertyVariables>
--- a/rest/admin-v2/tests/src/test/resources/keycloak-test.properties
+++ b/rest/admin-v2/tests/src/test/resources/keycloak-test.properties
@@ -1,5 +1,8 @@
 kc.test.log.level=WARN
 
+# Start embedded mode in IDE and distribution when running with Maven
+kc.test.server=embedded
+
 kc.test.log.filter=true
 
 kc.test.log.category."org.keycloak.tests".level=INFO


shawkins:
The only issue may be https://github.com/keycloak/keycloak/issues/44099 - which is likely due to somehow retaining old classloaders in memory.

If you run the whole test suite in the ide, do you see the OOM, or is there forking happening somewhere to prevent that?

stianst (changes requested):
Makes no sense to me; just add .env.test locally with `KC_TEST_SERVER=embedded`

stianst:
> The only issue may be #44099 - which is likely due to somehow retaining old classloaders in memory.
> 
> If you run the whole test suite in the ide, do you see the OOM, or is there forking happening somewhere to prevent that?

If you run 4-5 tests that have different server config causing server restarts the OOM happens

mabartos:
@stianst Why doesn't it make sense to you? :) I'd say it's very useful for contributors that every change is propagated and tests are aware of that. Instead, contributors now need to rebuild the whole distribution to see their changes, which is quite PITA :grin: 

However, as there are some problems already with the embedded, I can close for now. I've found another issue today as I still have the system property hanging somewhere - even when removed from the surefire plugin. Not sure how JUnit handles that, as even rebuilding didn't help :/ 

> just add .env.test locally with KC_TEST_SERVER=embedded

Ok, at least some approach to improve the DevEx. It'd be nice to have it documented. Thanks!

stianst:
> @stianst Why doesn't it make sense to you? :) I'd say it's very useful for contributors that every change is propagated and tests are aware of that. Instead, contributors now need to rebuild the whole distribution to see their changes, which is quite PITA 😁

Because it's easy to use the mode that you want to use; and not enforce your choice on others. Personally I primarily use remote mode when writing tests; as that saves a lot of time not having to wait for the server to start. I use `kcw dev-build` whenever I re-build the server which uses Maven caches so is usually pretty fast (< 30 seconds depending on what's changed). I primarily only use the embedded mode when I'm fixing something that is broken in the tests or debugging, and when I do I just quickly change to that.

That's basically the whole point with `.env.test` to allow you to put your own personal config for the tests. What server, browser, logging, etc, etc. 

> 
> However, as there are some problems already with the embedded, I can close for now. I've found another issue today as I still have the system property hanging somewhere - even when removed from the surefire plugin. Not sure how JUnit handles that, as even rebuilding didn't help :/
> 
> > just add .env.test locally with KC_TEST_SERVER=embedded
> 
> Ok, at least some approach to improve the DevEx. It'd be nice to have it documented. Thanks!

It is - https://github.com/keycloak/keycloak/blob/main/test-framework/HOW_TO_RUN.md

mabartos:
Ok, thanks for the additional info :)

Issue #40902, opened 2025-07-03
More fully document operator upgrade scenarios, in particular with custom images


shawkins:
Officially, we do not support using older KC image with newer Operator (or vice versa). That said, we do not have a good documented process how to upgrade when using custom images. So we're trying to avoid doing breaking changes in the Operator when not needed. And the current breaking changes policy in the server helps us with that: we're not doing any in minor releases. That enables us to use newer server version with older operator. Hence the approach @shawkins makes sense to me to document:

For minor and micro releases I would want the sequence to be:

- update your CRs to new images - I know it's not considered supported, but there are few circumstances where this won't work, so we should push for this instead.
  - for micro, and even some validated minor releases, the expectation is that this should be rolled in. For everything else a recreate would happen.
  - the admin is in control of the blast radius of this change. It does not require all Keycloaks have a coordinated outage window

- Upgrade the operator
  - there is still a potential for a wide blast radius as all managed Keycloaks will be updated simultaneously, but for those with a custom image hopefully we've limited that to just rolling restarts

That said, this only works for minor releases. I think we need to make it clear that for major upgrades (where we will not even consider supporting zero-downtime upgrade), the user must scale down their deployment before upgrading the Operator.

This might change in the future if we make the operator aware of the server version to support using different server versions with different operator versions. But for now, clear docs is IMHO the way to go.

_Originally posted by @vmuzikar in https://github.com/keycloak/keycloak/pull/40690#discussion_r2182457035_

shawkins:
>  I think we need to make it clear that for major upgrades (where we will not even consider supporting zero-downtime upgrade), the user must scale down their deployment before upgrading the Operator.

There are two scenarios here:

1. The instances or the CRD has incompatible changes that are not backwards compatible. Honestly if we're following best practices this should be an extremely rare case. If it's a structural issue with the CR, then OLM will prevent the upgrade whether the instances are scaled down or not. Otherwise as long as you handle deprecations prior to needing the upgrade, generally this should be expected to work.

2. The CRD is compatible. The only deviation from the the normal procedure from above is that we would strongly caution that using the Explicit update-strategy to force rolling is not supported.

vmuzikar:
Thinking about it more and also considering the future scenarios with cluster-wide Operator, I wonder if it would make sense to rather focus on the reversed scenario: update Operator first and then the image. There are couple of reasons I can see that support this:
[...]

Issue #32470, opened 2024-08-28
Remove Hostname v1 support from Operator


shawkins:
### Description

As a follow-up to #27731 we should remove / restrict handling for hostname v1 in the operator.

shawkins:
To be re-evaluated in 27. The final consideration will be on what is the oldest product custom image that will be supported by a KC 27 release.

PR #44506, opened 2025-11-26
Moved adding Kubernetes/OpenShift CAs to the server

slaskawi:
## Summary

Fixes https://github.com/keycloak/keycloak/issues/42900


operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -472,17 +472,8 @@ private void addEnvVars(StatefulSet baseDeployment, Keycloak keycloakCR, Watched
         LinkedHashMap<String, EnvVar> varMap = Stream.concat(Stream.concat(unsupportedEnv.stream(), firstClasssEnvVars.stream()), Stream.concat(additionalEnvVars.stream(), env))
                 .collect(Collectors.toMap(EnvVar::getName, Function.identity(), (e1, e2) -> e1, LinkedHashMap::new));
 
-
-        if (!Boolean.FALSE.equals(keycloakCR.getSpec().getAutomountServiceAccountToken())) {
```
slaskawi, line 479: Deprecate the CR option(s)

operator/src/test/java/org/keycloak/operator/testsuite/unit/PodTemplateTest.java
```diff
@@ -624,23 +624,6 @@ public void testCacheConfigFileMount() {
         Mockito.verify(this.watchedResources).annotateDeployment(Mockito.eq(Watched.of("cm")), Mockito.eq(ConfigMap.class), Mockito.any(), Mockito.any());
     }
 
-    @Test
```
slaskawi, line 659: We need some test for this. Maybe print it out in the logs and then parse here?

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
```diff
@@ -253,4 +244,71 @@ private static void mergeTrustStore(KeyStore truststore, String file, KeyStore a
                     "Failed to initialize truststore, could not merge: " + new File(file).getAbsolutePath(), e);
         }
     }
+
+    static File saveTruststore(KeyStore truststore, String dataDir, char[] password) {
```
slaskawi: Move up and compare with the upstream

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/KeycloakRecorder.java
```diff
@@ -94,8 +94,6 @@ public void configureTruststore() {
 
         if (truststoresDir != null && truststoresDir.exists() && Optional.ofNullable(truststoresDir.list()).map(a -> a.length).orElse(0) > 0) {
             truststores = Stream.concat(Stream.of(truststoresDir.getAbsolutePath()), Stream.of(truststores)).toArray(String[]::new);
-        } else if (truststores.length == 0) {
-            return; // nothing to configure, we'll just use the system default
```
slaskawi, line 138: If this still causes FIPS failures, we'll need to use a different approach

slaskawi:
Verified, the documentation failure is not related - it's also on main branch: https://github.com/keycloak/keycloak/actions/runs/19846799925

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -187,6 +187,24 @@ The option `http-accept-non-normalized-paths` was introduced to restore the prev
...
+* `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (OpenShift service CA)
+
+This behavior is enabled by default and can be controlled with the new server options:
+
+* `--truststore-kubernetes-enabled=true|false` (default: `true`)
+* `--truststore-openshift-enabled=true|false` (default: `true`)
+
+No changes are required for most deployments. If you previously relied on the Operator to manage these truststore entries, the server now performs the same function directly. For details, see the server guide section "Cluster CA certificates (Kubernetes and OpenShift)" in the Truststore guide.
+
+=== Deprecation of `automountServiceAccountToken` field in Keycloak CRD
```
shawkins: I don't think this should be deprecated. automountServiceAccountToken allows you to control this behavior without using the unsupported pod template. It is a hardening applicable beyond whether the keycloak vm picks up the certs or not - if a malicious provider jar for example is installed and uses the mounted certs to access the api server.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -187,6 +187,24 @@ The option `http-accept-non-normalized-paths` was introduced to restore the prev
...
+==== Truststore initialization moved to the server; Operator CRD field deprecated
+
+Keycloak now automatically discovers and trusts cluster certificate authorities when running on Kubernetes or OpenShift, without relying on the Operator to preconfigure the truststore. Specifically, if present in the container filesystem, the following files are added to the system truststore at runtime:
+
+* `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (Kubernetes service account CA)
+* `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (OpenShift service CA)
+
+This behavior is enabled by default and can be controlled with the new server options:
+
+* `--truststore-kubernetes-enabled=true|false` (default: `true`)
```
shawkins: Could we start with having just a single property? Or do see scenarios where you want one as false and the other as true?

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
slaskawi: Interesting approach. Perhaps, we should preserve it and if set to `false` it should disable the Kube CA scanning in the server. This way it will be consistent and backwards compatible. 

If you agree - I'll implement it this way tomorrow.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
slaskawi: Sounds good to me!

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
shawkins: If it doesn't matter whether the files are present, then I wouldn't worry about setting the scanning property explicitly to false.

slaskawi: No, actually the implementation cares about it. So we need to explicitly turn it off. Shouldn't be a big deal.

shawkins:
Just want to double check on generally defaulting to true. Are there any situations where users would expect the current behavior - not automatically picking up these certs - to remain the default?

cc @vmuzikar

ahus1:
> Just want to double check on generally defaulting to true. Are there any situations where users would expect the current behavior - not automatically picking up these certs - to remain the default?

IMHO this can be cured by adding it also to the notable changes and not only to the deprecations in the migration guide.

vmuzikar:
> IMHO this can be cured by adding it also to the notable changes and not only to the deprecations in the migration guide.

+1

Making it default makes sense to me.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
```diff
@@ -472,16 +474,9 @@ private void addEnvVars(StatefulSet baseDeployment, Keycloak keycloakCR, Watched
...
-            String truststores = SERVICE_ACCOUNT_DIR + "ca.crt";
```
shawkins, line 480: In general we don't handle this situation well.

We recommend when upgrading the operator to first upgrade to the newer keycloak images. However if someone does it the other way around, then removing this logic can cause something to break when the new operator is used with an older Keycloak image.

What opinions do we have here - just another thing to call out in a guide, or retaint the old logic for now and remove it later? cc @vmuzikar @ahus1

slaskawi: Could you please tell us more about this case @shawkins?

I might be totally wrong but I would expect Keycloak Operator to take the following steps:

1. Upgrade the Operator image
2. Keycloak Operator reconciles Keycloak StatefulSet and upgrades the Keycloak image
3. New Pods are being rolled out

If the above understanding is correct - users should not notice anything. If however, between steps 1 and 2 the Operator rolls out new StatefulSet with the pre-upgrade version of Keycloak - than indeed we need to call this out in the upgrade docs as users might notice interruptions.

slaskawi:
> Just want to double check on generally defaulting to true. Are there any situations where users would expect the current behavior - not automatically picking up these certs - to remain the default?

It's probably worth calling out that @stianst was also onboard with adding Kube CAs by default. Here's the relevant quote from the discussion (https://github.com/keycloak/keycloak/discussions/42852#discussioncomment-14484256):

> I'd say it should include the Kube CA irrespective if the user has specified certs/truststore themselves.
>
> There should be an option to disable it though (--cert-disable-kube-ca or something like that).

slaskawi:
@ahus1 @shawkins @vmuzikar I moved the documentation part to the Notable Changes section as suggested. 

Depending on the outcome of the discussion on upgrades with @shawkins  (https://github.com/keycloak/keycloak/pull/44506#discussion_r2597231354), I may add one of two sentences on expected interruptions in some corner cases.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakDeploymentDependentResource.java
shawkins: > If the above understanding is correct - users should not notice anything

We actually recommend against this, and tell users to specify a custom image. However they also then need to follow the recommended upgrade path (custom image first), or they would hit into a compatibility issue with the removed code.

slaskawi: Thanks for clarification @shawkins! I'll add more words to the docs to ensure users are aware of this.

slaskawi: @shawkins Added more words on the upgrade procedure. Resolving the thread, please reopen it if you think it needs further discussion.

slaskawi:
@shawkins Added a bit more explanation to the upgrade procedure. The PR is ready to re-review.

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -375,10 +375,24 @@ stringData:
...
+kind: Keycloak
+metadata:
+  name: example-kc
+spec:
+  additionalOptions:
+    - name: truststore-kubernetes-enabled
+      value: "false"
+----
+
+Deprecated: The `automountServiceAccountToken` field in the Keycloak CR is deprecated and has no effect on truststore configuration. If your security policies require not mounting the service account token, consider adjusting your Pod/ServiceAccount settings accordingly, but use the server options above to control whether cluster CAs are trusted by the server.
```
shawkins:
```suggestion
```

This should be removed or reworded. We are not deprecating behavior, rather introducing what could be a breaking change if you follow the wrong upgrade order.

Also we don't want people to use the unsupported pod template, the automountServiceAccountToken field would remain the supported mechanism to disable mounting.

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
```diff
@@ -48,9 +49,17 @@ public class TruststoreBuilder {
     public static final String DUMMY_PASSWORD = "keycloakchangeit"; // fips length compliant dummy password
     static final String PKCS12 = "PKCS12";
 
+    private static final String DEFAULT_KUBERNETES_CA_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt";
+    private static final String DEFAULT_SERVICE_CA_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt";
+    // Testability and flexibility: allow overriding CA paths via system properties
```
shawkins: I don't see this being used by tests. As for flexibility, wouldn't users just add a truststore path, rather than set a system property to change the default kubernetes locations?

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
```diff
@@ -106,10 +144,8 @@ private static void mergeFiles(String[] truststores, KeyStore truststore, boolea
                 var format = KeystoreUtil.getKeystoreFormat(file).orElse(null);
                 if (format == KeystoreFormat.PKCS12) {
                     mergeTrustStore(truststore, file, loadStore(file, PKCS12, null));
-                    if (!topLevel) {
-                        discoveredFiles.add(f.getAbsolutePath());
-                    }
-                } else if (mergePemFile(truststore, file, topLevel) && !topLevel) {
+                    discoveredFiles.add(f.getAbsolutePath());
```
shawkins, line 143: The notion of what is discovered was changed to all files, not just those under directories specified in the path list. However the message was not updated.

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
```diff
@@ -62,6 +71,35 @@ public static void setSystemTruststore(String[] truststores, boolean trustStoreI
...
+     * @return the updated truststore paths
+     */
+    public static String[] includeKubernetesTrustStorePaths(String[] truststores, boolean includeKubernetesCa) {
+        final List<String> truststoreList = new ArrayList<>(Arrays.asList(truststores));
+
+        if (includeKubernetesCa) {
+            String kubernetesCaPath = System.getProperty(KUBERNETES_CA_PATH_PROPERTY, DEFAULT_KUBERNETES_CA_PATH);
+            File kubernetesCA = new File(kubernetesCaPath);
+            if (kubernetesCA.exists() && kubernetesCA.isFile()) {
+                LOGGER.infof("Adding trusted Kubernetes CA from %s", kubernetesCaPath);
```
shawkins: This and the other info message are probably not needed if you have expanded the notion of discovered files to all files.

shawkins (commented):
Thanks @slaskawi just a couple of thoughts.

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
slaskawi: Good point. Fixed.

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
slaskawi: Good callout! Fixed

services/src/main/java/org/keycloak/truststore/TruststoreBuilder.java
slaskawi: It's a leftover. Removing.

shawkins:
@slaskawi updated the conflict to get another build running.

@vmuzikar @stianst do you want this targeting 26.5 or put on hold?

stianst:
@shawkins I'm 50/50 - if you and @vmuzikar are confident about it not introducing any issues then let's go ahead, but if we're more uncertain let's postpone it

shawkins (dismissed):
Thank you @slaskawi for the recent changes. LGTM

> @shawkins I'm 50/50 - if you and @vmuzikar are confident about it not introducing any issues then let's go ahead, but if we're more uncertain let's postpone it

Other than the possible issues, which are already noted in the docs, I don't forsee additional risk.

slaskawi:
> @slaskawi updated the conflict to get another build running.

@shawkins Thank you!

> @shawkins I'm 50/50 - if you and @vmuzikar are confident about it not introducing any issues then let's go ahead, but if we're more uncertain let's postpone it

@stianst I'm voting for including it. Even if something unpredictable happens, the blast radius shouldn't be big as we have a switch to disable this behavior.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -175,6 +175,21 @@ See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to
 For each expired user session there is a new user event `USER_SESSION_DELETED` fired.
 As part of this change, the process now deletes rows from the table in small batches, instead of issuing a delete statements that affects the whole table. This should allow for better response times when there are a lot of sessions in the table.
 
+=== Truststore initialization moved to the server
```
vmuzikar: Just a nit. We should also have some small note of this in the release notes. It's basically a new feature.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
```diff
@@ -175,6 +175,21 @@ See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to
...
+* `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (Kubernetes service account CA)
+* `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (OpenShift service CA)
+
+This behavior is enabled by default and can be controlled with the server option:
+
+* `--truststore-kubernetes-enabled=true|false` (default: `true`)
+
+No changes are required for most deployments. If you previously relied on the Operator to manage these truststore entries, the server now performs the same function directly. For details, see the server guide section "Cluster CA certificates (Kubernetes and OpenShift)" in the Truststore guide.
+
+If you are deploying {project_name} by using the {project_name} Operator and your deployment relies on trusting the Kubernetes CA, upgrade the Keycloak image first. This ensures that the Kubernetes CAs remain trusted for the entire duration of the upgrade.
```
vmuzikar: Folks having automatic Operator upgrades enabled in OLM with custom KC images might run into troubles with this but I assume it's fine as we officially support only 1:1 version mapping with server and the operator.

vmuzikar (commented):
Seems safe to include in 26.5.

@slaskawi Thank you, nice work. LGTM, added just some non-blocking comments.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
shawkins: We cover recommeding against automatic upgrades in general here: https://www.keycloak.org/operator/installation#_configuring_manual_approval_for_olm_upgrades

I thought we had added a blurb about updating custom images first to the docs already - however I was wrong. That is covered in this issue https://github.com/keycloak/keycloak/issues/40902 - but was never added to the docs.

Since we weren't already explicit about this upgrade procedure,  I think we need to add back the operator logic that is explicitly specifying the paths.

vmuzikar: > I think we need to add back the operator logic that is explicitly specifying the paths.

IMHO no need. Lt's just make sure we document the trustore related change more explicitly in the upgrading guide.

shawkins: Can you help clarify on what / when we consider something a breaking change then for the operator? For example we defered https://github.com/keycloak/keycloak/issues/32470 to KC 27 on the chance that someone would be using the 26.x operator with a 25.x image that still relied upon hostname v1.

vmuzikar: #32470 was affecting CRDs – a cluster-wide resource. Doing it, we'd effectively prevent users from using older operator versions in the cluster. That was the main argument, as far as I understand.

shawkins: >  Doing it, we'd effectively prevent users from using older operator versions in the cluster.

I don't think that was that entirely because the global management of the CRD is problematic regardless - which we do note in the guide. And that problem goes both ways - if you use older operators with the newer CRD, then you also can specify stuff in the CR that won't work with older operators. 

As far as I recall it was based upon the feature still being available in KC 25, and that in theory it was possible for downstream users to continue to use older images, even temporarily, with the newer operator.

In this case a working Keycloak using a 26.4 custom image could be broken by upgrading the operator to 26.5.

The only rationale for that not be be a breaking change based upon the old reasoning is that there is a workaround - affected Keycloaks with older images will require the CR to be updated to specify these truststore paths.

vmuzikar: > The only rationale for that not be be a breaking change based upon the old reasoning is that there is a workaround - affected Keycloaks with older images will require the CR to be updated to specify these truststore paths.

Yes, that would be good enough, IMHO.

shawkins: Ok, so should the workaround be documented explicitly? Or in case are we requiring the images to be upgraded?

vmuzikar: > so should the workaround be documented explicitly?

I'd say yes, in this specific case. I'd include that in the upgrading guide.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
slaskawi: Added a note about this.

docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
slaskawi: Added more in-depth notes

shawkins:
> Seems safe to include in 26.5.
> 
> @slaskawi Thank you, nice work. LGTM, added just some non-blocking comments.

I would still prefer just leaving the old handling in the operator to remove a chance this is a breaking change - that seems more user friendly than relying on doc notes - but if everyone else prefers this approach I won't stand in the way.

vmuzikar:
> I would still prefer just leaving the old handling in the operator to remove a chance this is a breaking change - that seems more user friendly than relying on doc notes - but if everyone else prefers this approach I won't stand in the way.

Either way works for me. I slightly more prefer just the docs note so we avoid creating a follow-up for future clean-up of the  Operator logic.

shawkins:
We should consider this again for 26.6. https://github.com/keycloak/keycloak/pull/46380 tries to provide some general upgrade instructions, which would be compatible with what is already in this PR.

slaskawi:
This sounds good to me! I will allocate a bit of time this week to update this PR.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -120,6 +120,36 @@ It does not affect any explicitly configured clock skew for example in identity
...
+
+This behavior is enabled by default and can be controlled with the server option:
+
+* `--truststore-kubernetes-enabled=true|false` (default: `true`)
+
+No changes are required for most deployments. If you previously relied on the Operator to manage these truststore entries, the server now performs the same function directly. For details, see the server guide section "Cluster CA certificates (Kubernetes and OpenShift)" in the Truststore guide.
+
+If you are deploying {project_name} by using the {project_name} Operator and your deployment relies on trusting the Kubernetes CA, upgrade the Keycloak image first. This ensures that the Kubernetes CAs remain trusted for the entire duration of the upgrade.
+
+If you cannot upgrade the image first, it is recommended to explicitly configure the truststore path in the Keycloak Custom Resource prior to changing the Operator version. This ensures that older pods continue to trust the cluster CA during the upgrade procedure:
```
vmuzikar: Should we update these instructions now that https://github.com/keycloak/keycloak/pull/46380 was merged?

shawkins: Expanded on them a little.

slaskawi: LGTM

slaskawi:
Thank you for all the changes @shawkins. I highly appreciate all the hard work on this item!

docs/guides/operator/advanced-configuration.adoc
```diff
@@ -375,10 +375,22 @@ stringData:
     ...
 ------
 
-When running on a Kubernetes or OpenShift environment well-known locations of trusted certificates are included automatically.
-This includes `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` and the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` when present.
-In order to not include `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` in Keycloak pods, set the `automountServiceAccountToken` field in the spec to `false`
```
vmuzikar, line 382: We should keep `automountServiceAccountToken` documented in some form. The option still exists and is useful in some use cases.

quarkus/config-api/src/main/java/org/keycloak/config/TruststoreOptions.java
```diff
@@ -11,6 +11,12 @@ public class TruststoreOptions {
             .description("List of pkcs12 (p12, pfx, or pkcs12 file extensions), PEM files, or directories containing those files that will be used as a system truststore.")
             .build();
 
+    public static final Option<Boolean> TRUSTSTORE_KUBERNETES_CA_ENABLED = new OptionBuilder<>("truststore-kubernetes-enabled", Boolean.class)
+            .category(OptionCategory.TRUSTSTORE)
+            .description("If enabled, the server will automatically include the default Kubernetes service account CA certificate from \"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt\" and the OpenShift service CA certificate from \"/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt\" when running in a container environment.")
+            .defaultValue(true)
```
vmuzikar, line 17: I still wonder if setting it to true might be viewed as a breaking change and potentially loosening security. The scenario I'm worried about is: someone deploying KC in K8s without the Operator (quite a common use case), doesn't read upgrading notes thoroughly and upgrades to 26.6; KC suddenly trusts the K8s certs which might potentially be used by an attacker.

docs/guides/server/keycloak-truststore.adoc
```diff
@@ -16,14 +16,31 @@ The existing Java default truststore certs will always be trusted.  If you need
...
+When running {project_name} on Kubernetes or OpenShift, the server will automatically include well-known cluster certificate authorities in the system truststore if they are present on the filesystem:
+
+* Kubernetes service account CA: `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`
+* OpenShift service CA: `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`
+
+This behavior is enabled by default and does not require the Operator. It can be controlled with the following option:
+
+* `--truststore-kubernetes-enabled=true|false` (default: `true`)
+
+When `truststore-kubernetes-enabled` is set to `true`, {project_name} will also discover and include the OpenShift Service CA if it is present on the filesystem. To disable automatically trusting the cluster CAs entirely:
```
vmuzikar: Do we need the explicit mention of OpenShift here? We already mention it above.
```suggestion
To disable automatically trusting the cluster CAs entirely:
```

quarkus/config-api/src/main/java/org/keycloak/config/TruststoreOptions.java
```diff
@@ -11,6 +11,12 @@ public class TruststoreOptions {
             .description("List of pkcs12 (p12, pfx, or pkcs12 file extensions), PEM files, or directories containing those files that will be used as a system truststore.")
             .build();
 
+    public static final Option<Boolean> TRUSTSTORE_KUBERNETES_CA_ENABLED = new OptionBuilder<>("truststore-kubernetes-enabled", Boolean.class)
```
vmuzikar, line 14: Could you also please add a server side test for this? We're testing this through the Operator but it ccould get easily forgotten there, and we should not test server features through Operator only.

vmuzikar (changes requested):

quarkus/config-api/src/main/java/org/keycloak/config/TruststoreOptions.java
shawkins: Let's elaborate on the potential attacks.

- out-going https, ldap, and smtp requests. The malicious actor has to also control the KC instance at an admin level to configure the hostnames for those endpoints - or do you see another attack here?

- in-coming mtls is problematic if you don't set the https-truststore-* options as we currently lack other controls on what client certs are valid. We call that out in the docs, but if you are ok with what is trusted by the JRE, then you may not have set it. Here we're expanding that trust to the kubernetes CA. So if we want to be safer, we should only default to true if mtls is disabled or https-truststore-* is set.

docs/guides/operator/advanced-configuration.adoc
slaskawi: Sure, this definitely makes sense!

quarkus/config-api/src/main/java/org/keycloak/config/TruststoreOptions.java
slaskawi: Sure! Added

quarkus/config-api/src/main/java/org/keycloak/config/TruststoreOptions.java
slaskawi: @shawkins has a point - this PR expands the mTLS trust boundary and the user hasn't opted in for it. 

I'm not sure if we should include the Kube CA conditionally on enabled/disabled https-truststore-*. This seems confusing from the configuration standpoint. 

Until we figure this out, I'll just add a small comment to the docs about it.

vmuzikar: Yes, we discussed this further with @shawkins offline and we agreed to keep as is.

vmuzikar (commented):
Could you also please add a brief Quarkus test for completeness?

slaskawi:
> Could you also please add a brief Quarkus test for completeness?

@vmuzikar Unfortunately adding a Quarkus-based test is not an easy task. Both paths (Kube CA and Service CA) are hardcoded to avoid someone trying to spoof them and I don't expect them to be writable in our CI system. So there's no  elegant way to access it.

With the above in mind, I decided to go for a unit test (to protect against accidental changes) and an e2e test in the Operator. Both should give us the layered-type of protection we need here.

vmuzikar (dismissed):
> Both paths (Kube CA and Service CA) are hardcoded to avoid someone trying to spoof them and I don't expect them to be writable in our CI system.

Right, didn't realize this.

LGTM then, thank you @slaskawi.

@shawkins Do you want to give it another look too?

shawkins:
> > Both paths (Kube CA and Service CA) are hardcoded to avoid someone trying to spoof them and I don't expect them to be writable in our CI system.
> 
> Right, didn't realize this.
> 
> LGTM then, thank you @slaskawi.
> 
> @shawkins Do you want to give it another look too?

Seems good to me based upon everything that has been discussed.

To recap:

- It will not be breaking to operator based installs if they follow the recommended upgrade instructions.
- There is a warning about the expansion of trust in the main docs - I think it would be good to call that out directly in the upgrading guide.

Could we get another signoff from @stianst or @ahus1 on the above as well?

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -120,6 +120,36 @@ It does not affect any explicitly configured clock skew for example in identity
...
+Keycloak now automatically discovers and trusts cluster certificate authorities when running on Kubernetes or OpenShift, without relying on the Operator to preconfigure the truststore. Specifically, if present in the container filesystem, the following files are added to the system truststore at runtime:
+
+* `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (Kubernetes service account CA)
+* `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (OpenShift service CA)
+
+This behavior is enabled by default and can be controlled with the server option:
+
+* `--truststore-kubernetes-enabled=true|false` (default: `true`)
+
+No changes are required for most deployments. If you previously relied on the Operator to manage these truststore entries, the server now performs the same function directly. For details, see the server guide section "Cluster CA certificates (Kubernetes and OpenShift)" in the Truststore guide.
```
shawkins:
```suggestion
No changes are required for most deployments. If you previously relied on the Operator to manage these truststore entries, the server now performs the same function directly. 

WARNING: when `https-client-auth` is set to `required` or `request` without an explicit `https-trust-store-file`, mTLS client certificate validation falls back to the system truststore. With `truststore-kubernetes-enabled=true`, this means certificates signed by the Kubernetes cluster CA will be accepted as valid client certificates. If this is not desired, either set `https-trust-store-file` explicitly or disable `truststore-kubernetes-enabled`.

For details, see the server guide section "Cluster CA certificates (Kubernetes and OpenShift)" in the Truststore guide.
```

shawkins:
@vmuzikar I've put the warning in the upgrade notes as well. Based upon our meeting today it seems like you feel good about merged without additional approvals, is that correct?

shawkins (approved):
LGTM based upon signoff from @vmuzikar

slaskawi:
Merging `upstream/main` to see if the CI failures persist (they are probably not related)

vmuzikar (approved):

PR #44510, opened 2025-11-26
Correct OAuth reference for 26.2

andymunro:
Closes #44507


stianst:
OAuth2 is supported as IdP

PR #44516, opened 2025-11-27
Make sure that signature validation possible to configure for OIDC id…

mposolda:
…entity provider when JWT authz grant or clientAssertions authentication enabled

closes #44473

- Switch `Validate signature` of OIDC identity provider is used just to indicate that tokens from OIDC authentication scenarios should be validated (or external-internal token exchange for token-exchange:v1 as it was before). However the signatures are always required to be validated when JWT-Authorization-grant or client-federated-authentication is enabled

- Possible to setup keys (JWKS URL or hardcoded key) in OIDC identity provider UI when "validate signatore" or "jwt authorization grant" or "client assertions" are enabled

- Some minor refactoring (EG. renaming boolean getter methods on `JWTAuthorizationGrantConfig` to use `is*` instead of `get*` . Also fixed some missing UI tooltips to be shown


js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -2479,7 +2479,7 @@ targetContextAttributes=Target Context Attributes
 targetContextAttributesHelp=Defines the evaluation of context attributes (claims) instead of identity attributes
 filteredByClaim=Verify essential claim
 rowCancelBtnAriaLabel=Cancel edits for {{messageBundle}}
-validateSignatureHelp=Enable/disable signature validation of external IDP signatures. For Federated Client Authentication and JWT Authorization Grant the signature validation must be enabled.
+validateSignatureHelp=Enable/disable signature validation of external IDP tokens. When enabled, Keycloak will validate JWT tokens retrieved from the Identity provider during scenarios related to user authentication like for example when Keycloak obtains IDToken or accessToken after completion of the OIDC/OAuth2 flow with Identity provider. For Federated Client Authentication and JWT Authorization Grant the signature validation is always used regardless of the value of this switch.
```
rmartinc:
```suggestion
validateSignatureHelp=Enable/disable signature validation of external IDP tokens. When enabled, Keycloak will validate JWT tokens retrieved from the Identity provider during scenarios related to user authentication, for example when Keycloak obtains IDToken or accessToken after completion of the OIDC/OAuth2 flow with Identity provider. For Federated Client Authentication and JWT Authorization Grant, the signature validation is always used regardless of the value of this switch.
```

rmartinc (approved):
Thanks @mposolda! Just a minor comment for the label. But it's minor just apply if you think it's better. Approving anyway.

graziang (approved):
@mposolda thanks!

mposolda:
@rmartinc graziang Thanks for the review! Applied the suggestion for label @rmartinc suggested.

rmartinc (approved):
Thanks @mposolda! Approved!

ahus1 (approved):

mposolda:
@rmartinc @graziang @ahus1 Thanks!

PR #44519, opened 2025-11-27
Login failure cache: Evict entries after the configured failure reset time

### LINKED ISSUE #44801: Infinispan: LoginFailures entries should expire
### Description

The entries in the login failures cache currently do not expire.

Depending on the brute force strategy, the entries can have a lifetime, which can significantly reduce the caches. 

### Value Proposition

Smaller memory footprint, better behavior on rolling updates when due to less state transfer

From the new docs: 

> Previous entries in the `loginFailures` cache never expired, and entries accumulated as users entered wrong credentials, increasing the memory consumption.

> Starting with this release, entries will expire based on the "Failure reset time" configured in the "Brute force detection" for the modes "Lockout temporarily" and "Lockout permanently after temporary lockout". For "Lockout permanently", entries will not expire as before, as this mode does not have a "Failure reset time".

### Goals

* Be more scalable when there are lot of users

### Non-Goals

* Making the cache truly bounded / with a restricted memory usage. This is part of #32080

### Discussion

_No response_

### Notes

_No response_

chehrhar:
Closes #44801

Set the lifespan for cache entries in the login failures cache to match the "failure reset time" (default 12 hours) configured for brute force protection.

After this period, Keycloak would reset the failure counter, anyway, so the corresponding cache entry can be safely evicted.


ahus1:
@chehrhar  - thank you for the pull request. I've been writing a test for this change, and the test is green. Once test was failing, and this is due to a scenario where there are already old entries in the login failures cache, and then you either change the policy or the duration. 

The docs should now reflect the new behavior, still I think this is a bit too much to explain here. The code should improve to simpler to handle, and that we need fewer docs.  I'll think about it a bit more probably next week. We might need to iterate over all entries in the cache align their lifetimes.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +152,57 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
     public void close() {
 
     }
+
+    @Override
+    public void migrate(String modelVersion) {
```
pruivo: It is unnecessary to migrate volatile data. The operator will shutdown the cluster when updating the Keycloak version.

ahus1: I know that the Operator will shut down the cluster, but not all the custom helm charts and deployments out there. There was a point where I was considering backporting it. With the amount of changes it accumulated, maybe not any more. Maybe I'll backport a lighter variant. Let's see.

pruivo: Even with those scenarios, I'm still going to vote against any kind of migration.

If they are doing rolling updates in an unsupported way, and they haven't complained about the login failures cache performance, there is no need to increase the cluster load with the migration; it causes conflicts/delays during the logins in progress.

The cached data will be migrated on a successful or failed login anyway, and delaying startup is unnecessary. 

Also, there is a second problem; if there is a 26.4 node running in the cluster, it will "revert" the migration on access anyway. 

Just my 2c.

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmAdapter.java
```diff
@@ -259,9 +262,43 @@ public boolean isBruteForceProtected() {
...
+                @Override
+                protected void rollbackImpl() {
+                    updateBruteForceSettings = false;
+                }
+            });
+        }
+    }
+
+    private static void runUpdateOfLoginFailureProvider(KeycloakSessionFactory keycloakSessionFactory, String realmId) {
+        KeycloakModelUtils.runJobInTransaction(keycloakSessionFactory,
```
pruivo, line 295: I really don't like this. Can't we use provider events or any local event? We need to come up with a solution to work with other realm settings too. (yes, I'm thinking about the user sessions lifespan/max-idle with volatile data)

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +152,57 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
...
+            CompletionStages.join(stage.freeze());
+        }
+    }
+
+    @Override
+    public void updateWithLatestRealmSettings(RealmModel realm) {
+        Cache<LoginFailureKey, SessionEntityWrapper<LoginFailureEntity>> cache = loginFailuresTx.getCache();
+        AggregateCompletionStage<Void> stage = CompletionStages.aggregateCompletionStage();
+        if (!realm.isBruteForceProtected()) {
+            cache.getAdvancedCache().entrySet().stream()
```
pruivo: don't need to reinvent the wheel, just invoke `removeAllUserLoginFailures(RealmModel realm)`

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +152,57 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
...
+                    .filter(entry -> Objects.equals(entry.getKey().realmId(), realm.getId()))
+                    .forEach(entry -> stage.dependsOn(removeKeyFromCache(cache, entry.getKey())));
+        } else {
+            cache.getAdvancedCache().entrySet().stream()
```
pruivo: Will brought to my attention that this pattern will trigger the operations for all keys really quickly, which may overload the server. Use instead:

```java
CompletionStages.performConcurrently(<stream here>, <number of concurrent ops / batch size>, Scheduler.from(new WithinThreadExecutor()), <consumer here>);
```

See comment here:
https://github.com/infinispan/infinispan/issues/16245#issuecomment-3577850168

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/remote/RemoteUserLoginFailureProvider.java
```diff
@@ -77,6 +87,58 @@ public void removeAllUserLoginFailures(RealmModel realm) {
...
+    @Override
+    public void updateWithLatestRealmSettings(RealmModel realm) {
+        RemoteCache<LoginFailureKey, LoginFailureEntity> cache = transaction.getCache();
+        AggregateCompletionStage<Void> stage = CompletionStages.aggregateCompletionStage();
+        if (!realm.isBruteForceProtected()) {
+            cache.entrySet().stream()
+                    .filter(entry -> Objects.equals(entry.getKey().realmId(), realm.getId()))
+                    .forEach(entry -> stage.dependsOn(removeKeyFromCache(cache, entry.getKey())));
+        } else {
+            cache.entrySet().stream()
```
pruivo: You can have server-side filtering with the method `retrieveEntriesByQuery()`. 
The `realmId` is indexed, so it should be relatively fast to retrieve the data.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +152,57 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
...
+                    .filter(entry -> Objects.equals(entry.getKey().realmId(), realm.getId()))
```
pruivo:
```suggestion
                    .filter(SessionWrapperPredicate.create(realmId))
```

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +152,57 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
...
+                    });
+        }
+        CompletionStages.join(stage.freeze());
+    }
+
+    private static void updateLifetimeOfCacheEntry(Map.Entry<LoginFailureKey, SessionEntityWrapper<LoginFailureEntity>> entry, RealmModel realm, AggregateCompletionStage<Void> stage, Cache<LoginFailureKey, SessionEntityWrapper<LoginFailureEntity>> cache) {
+        long lifespanMs = SessionTimeouts.getLoginFailuresLifespanMs(realm, null, entry.getValue().getEntity());
+        long maxIdleMs = SessionTimeouts.getLoginFailuresMaxIdleMs(realm, null, entry.getValue().getEntity());
+        stage.dependsOn(
+                cache.getAdvancedCache().computeIfPresentAsync(entry.getKey(),
```
pruivo: Infinispan has an identity function: `StreamMarshalling.identity()`

pruivo (changes requested):

ahus1:
@pruivo - thank you for all the comments. Let's do more review not this week, but next week. Thanks!

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/remote/RemoteUserLoginFailureProvider.java
```diff
@@ -77,6 +90,34 @@ public void removeAllUserLoginFailures(RealmModel realm) {
...
+        long maxIdleMs = SessionTimeouts.getLoginFailuresMaxIdleMs(realm, null, value);
+        return cache.computeIfPresentAsync(key,
+                // Keep the original value - this should only update the lifespan and idle time
+                (loginFailureKey, loginFailureEntitySessionEntityWrapper) -> loginFailureEntitySessionEntityWrapper,
+                lifespanMs, TimeUnit.MILLISECONDS, maxIdleMs, TimeUnit.MILLISECONDS);
+    }
+
+    private static CompletableFuture<LoginFailureEntity> removeKeyFromCache(RemoteCache<LoginFailureKey, LoginFailureEntity> cache, LoginFailureKey key) {
+        return cache.removeAsync(key);
+    }
```
pruivo, line 115: Using `QueryHelper.streamAll()` to avoid downloading "all" the cache, as it uses pagination. 

``` suggestion
            final long maxDeltaTimeMillis = realm.getMaxDeltaTimeSeconds() * 1000L;
            final boolean isPermanentLockout = realm.isPermanentLockout();
            final int maxTemporaryLockouts = realm.getMaxTemporaryLockouts();
            Query<LoginFailureEntity> query = LoginFailureQueries.searchByRealmId(cache, realm.getId());
            CompletionStages.performConcurrently(
                    QueryHelper.streamAll(query, 20, Function.identity()),
                    20,
                    Schedulers.from(new WithinThreadExecutor()),
                    entry -> updateLifetimeOfCacheEntry(entry, cache, isPermanentLockout, maxTemporaryLockouts, maxDeltaTimeMillis));
        }
    }

    private static CompletionStage<?> updateLifetimeOfCacheEntry(LoginFailureEntity entry, RemoteCache<LoginFailureKey, LoginFailureEntity> cache, boolean isPermanentLockout, int maxTemporaryLockouts, long maxDeltaTimeMillis) {
        long lifespan = SessionTimeouts.getLoginFailuresLifespanMs(isPermanentLockout, maxTemporaryLockouts, maxDeltaTimeMillis, entry);
        return cache.computeIfPresentAsync(new LoginFailureKey(entry.getRealmId(), entry.getUserId()),
                // Keep the original value - this should only update the lifespan and idle time
                (loginFailureKey, loginFailureEntitySessionEntityWrapper) -> loginFailureEntitySessionEntityWrapper,
                lifespan, TimeUnit.MILLISECONDS);
    }
```

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +150,32 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
...
+            cache.entrySet().stream()
+                    .filter(SessionWrapperPredicate.create(realm.getId()))
+                    .forEach(RemoveKeyConsumer.getInstance());
+        } else {
+            final long maxDeltaTimeMillis = realm.getMaxDeltaTimeSeconds() * 1000L;
+            final boolean isPermanentLockout = realm.isPermanentLockout();
+            final int maxTemporaryLockouts = realm.getMaxTemporaryLockouts();
+            cache.entrySet().stream()
+                    .filter(SessionWrapperPredicate.create(realm.getId()))
+                    .<LoginFailureKey, LoginFailureEntity>forEach((c, entry) -> {
```
pruivo: Fixes 
```
[kc-server] Caused by: java.lang.ClassCastException: class org.keycloak.models.sessions.infinispan.changes.SessionEntityWrapper cannot be cast to class org.keycloak.models.sessions.infinispan.entities.LoginFailureEntity (org.keycloak.models.sessions.infinispan.changes.SessionEntityWrapper and org.keycloak.models.sessions.infinispan.entities.LoginFailureEntity are in unnamed module of loader io.quarkus.bootstrap.runner.RunnerClassLoader @2890c451)
```

```suggestion
                    .<LoginFailureKey, SessionEntityWrapper<LoginFailureEntity>>forEach((c, entry) -> {
```

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/remote/RemoteUserLoginFailureProvider.java
```diff
@@ -36,9 +48,11 @@ public class RemoteUserLoginFailureProvider implements UserLoginFailureProvider
     private static final Logger log = Logger.getLogger(MethodHandles.lookup().lookupClass());
 
     private final LoginFailureChangeLogTransaction transaction;
+    private final KeycloakSession session;
```
pruivo: unused

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/InfinispanUserLoginFailureProvider.java
```diff
@@ -145,4 +150,32 @@ InfinispanChangelogBasedTransaction<LoginFailureKey, LoginFailureEntity> getLogi
...
+                        var entity = entry.getValue().getEntity();
+                        long lifespan = SessionTimeouts.getLoginFailuresLifespanMs(isPermanentLockout, maxTemporaryLockouts, maxDeltaTimeMillis, entity);
+                        c.getAdvancedCache()
+                                .withFlags(Flag.ZERO_LOCK_ACQUISITION_TIMEOUT,Flag.FAIL_SILENTLY, Flag.IGNORE_RETURN_VALUES)
+                                .computeIfPresent(entry.getKey(), (key, value) -> value, lifespan, TimeUnit.MILLISECONDS);
```
pruivo: Fixes
```
[kc-server] Caused by: java.lang.NoSuchMethodException: org.keycloak.models.sessions.infinispan.InfinispanUserLoginFailureProvider$$Lambda/0x00007fa17413fab8.writeReplace()
```

```suggestion
                                .computeIfPresent(entry.getKey(), (SerializableBiFunction<? super LoginFailureKey, ? super SessionEntityWrapper<LoginFailureEntity>, ? extends SessionEntityWrapper<LoginFailureEntity>>) (key, value) -> value, lifespan, TimeUnit.MILLISECONDS);
```

pruivo (changes requested):
A couple of fixes for my mistakes below. 

I would convert the lambdas to concrete classes and annotate them with ProtoStream annotation: less error prone, and the lambda logic can be evolved.
 I can do that in a follow-up PR if you like.

ahus1:
> I would convert the lambdas to concrete classes and annotate them with ProtoStream annotation: less error prone, and the lambda logic can be evolved.
I can do that in a follow-up PR if you like.

Thank you for the review, I applied the changes - let's wait for the build to complete.

It would be great if you could apply the ProtoStream magic in a follow-up PR.

testsuite/model/src/test/java/org/keycloak/testsuite/model/infinispan/RetryAndBackOffTest.java
```diff
@@ -97,6 +97,7 @@ public void testRetryWithPut() {
 
         inComittedTransaction(session -> {
             var realm = session.realms().getRealm(realmId);
+            session.getContext().setRealm(realm);
```
pruivo, line 100: There are 3 `inCommittedTransaction` missing the `session.getContext().setRealm(realm);`. 
```diff
diff --git a/testsuite/model/src/test/java/org/keycloak/testsuite/model/infinispan/RetryAndBackOffTest.java b/testsuite/model/src/test/java/org/keycloak/testsuite/model/infinispan/RetryAndBackOffTest.java
index eee9ba87a2e..421a6b2ecf6 100644
--- a/testsuite/model/src/test/java/org/keycloak/testsuite/model/infinispan/RetryAndBackOffTest.java
+++ b/testsuite/model/src/test/java/org/keycloak/testsuite/model/infinispan/RetryAndBackOffTest.java
@@ -118,6 +118,7 @@ public class RetryAndBackOffTest extends KeycloakModelTest {
 
         inComittedTransaction(session -> {
             var realm = session.realms().getRealm(realmId);
+            session.getContext().setRealm(realm);
             var loginFailures = session.loginFailures().getUserLoginFailure(realm, userId);
             loginFailures.incrementFailures();
         });
@@ -138,6 +139,7 @@ public class RetryAndBackOffTest extends KeycloakModelTest {
 
         inComittedTransaction(session -> {
             var realm = session.realms().getRealm(realmId);
+            session.getContext().setRealm(realm);
             session.loginFailures().removeUserLoginFailure(realm, userId);
         });
 
@@ -160,6 +162,7 @@ public class RetryAndBackOffTest extends KeycloakModelTest {
 
         inComittedTransaction(session -> {
             var realm = session.realms().getRealm(realmId);
+            session.getContext().setRealm(realm);
             var loginFailures = session.loginFailures().getUserLoginFailure(realm, userId);
             loginFailures.incrementFailures();
         });

```

pruivo (approved):

ahus1 (approved):
Thank you for the contribution. It grew larger once we added some of the polishing to update existing entries when admins change the expiry time. There were also some corner cases as the realm was not accessible from when using the remote Infinispan.

With that, this change is unfortunately too large to backport it fully to KC 26.4. Maybe we can do a partial backport.

Issue #32955, opened 2024-09-16
[OID4VCI] Simplify SD-JWT API usage


Captain-P-Goldfish:
### Description

In my opinion the current SD-JWT implementation is a bit to stiff and accepts too many parameters. I would suggest to simplify the approach to make it easier usable. I got an idea of an easier implementation by the SD-JWT implementation from `authlete` and I used it here in our reference implementation:

- https://gitlab.opencode.de/governikus/eudi-wallet-pid-provider-governikus-kg/-/blob/main/eudi-wallet-pid-provider/src/main/java/de/governikus/autent/eudiwallet/keycloak/provider/credentialbuilder/SdJwtCredentialBuilderProvider.java?ref_type=heads

The corresponding signature-service implementation can be found here:

https://gitlab.opencode.de/governikus/eudi-wallet-pid-provider-governikus-kg/-/blob/main/eudi-wallet-pid-provider/src/main/java/de/governikus/autent/eudiwallet/keycloak/provider/signingservices/SdJwtSigningProvider.java?ref_type=heads

Awambeng:
I am currently working on this

PR #44526, opened 2025-11-27
Removing SdJwtFacade

### LINKED ISSUE #44525: Do we need SdJwtFacade?
### Description

There is class `SdJwtFacade` in the `core` module as part of sd-jwt SDK. However I wonder if we really need this class? It is not used anywhere and it seems that using of this class does not provide any real advantage/simplification as of now. Removing it can reduce the complexity of the SDK.

mposolda:
closes #44525

@francis-pouatcha @Awambeng  @IngridPuppet @forkimenjeckayang @Ogenbertrand @Captain-P-Goldfish I am curious for the feedback if the class is needed for something? If you think that this class have some good use, I can close the PR and the issue.


Captain-P-Goldfish:
We are not using it. Haven't even seen it actually until now.

IngridPuppet (approved):
The class was introduced to address ticket https://github.com/keycloak/keycloak/issues/32955. Since we're moving toward a new API overall, I would agree it becomes less relevant.

mposolda:
@IngridPuppet @Captain-P-Goldfish Thanks!

Issue #14623, opened 2022-09-27
Ability to specify URL alias for realms


talasjanos:
### Description

https://issues.redhat.com/plugins/servlet/mobile#issue/KEYCLOAK-12297

### Details

What:
Add ability to map a complete realm URL to a custom URL.
e.g. 
instead of: https://sso.example.com/auth/realms/awesome-realm
to just : https://sso.example.com/
Currently with version 8.0.0 it is only possible to achieve with basic configuration
https://sso.example.com/realms/awesome-realm
Even though you hide and rewrite path(s) behind a reverse proxy, one needs to rewrite response body (ui, oidc json body etc.) , headers (redirects), cookie paths and so on. 
Even after achieving rewriting all paths it is not possible to change the issuer information inside a JWT token, which will always have the path /realms/<realm> attached to the end of it.
Why:
1 - realms/<realm> path instantly suggests that the system is using keycloak.
a - It is unlikely but this may be a vulnerability for a system which using a keycloak, i.e. if it is known from outside world that a system uses keycloak, and if keycloak has a known vulnerability.
b - Sharing tech stack with outside world for a system is not preferred generally.
2 - Shorter/ fully customizable URLs can be provided and with custom themes on the UI the product is fully customizable for consumers.
How:
1 - There is already a configuration in the realm settings to set front-end URL, with a checkbox one can configure if this URL should override the entire realm/issuer URL.
2 - If the checkbox not selected (default) the system behaves as it is behaving now.
3 - If the checkbox is selected then system overrides all URLs of the realm with the provided realm URL including back-end URLs.
4 - Keycloak server can identify a realm for incomming requests with the unique Frontend URL or domain.
As an alternative
1 - In realm settings there might be a field to set custom unique realm path.

ssilvert:
Whether or not to implement this doesn't depend on the UI team.

mot256:
This suggestion will also help us do an inline replacement of an existing oauth provider/tech (Identity Server 4) without the need to reconfigure systems to authenticate against another issuer URL. Any plans to implement this request any time soon?

shawkins:
Marked an older issue as a duplicate, which had been upvoted 3 times. Also discuessed here: https://github.com/keycloak/keycloak/discussions/31505

@ssilvert @vmuzikar it doesn't seem like cloud-native will have sole ownership of this either. It spans UI, cloud-native, and core. Not sure if there's concensus that this is even achievable with a reasonable amount of effort.

[...]

PR #44527, opened 2025-11-27
[admin-api-v2] Every distinct Admin API should be versioned

### LINKED ISSUE #43571: Every distinct Admin API should be versioned
### Description

As mentioned in the [comment](https://github.com/keycloak/keycloak/pull/43395#discussion_r2435158147), we might consider not to have the whole Admin API versioned, but potentially have versioned sub-APIs as Clients API, etc.

It'd need some discussion.

mabartos:
- Closes #43571
---------------------------
- Added HTTP method annotation to impl classes for better visibility
- Path changed `admin/api/v2/realms/master/clients/*` -->  `admin/api/api/master/clients/v2/*`
- No need for the RealmApis
- No default version handling


shawkins:
> When the version is not present, the default is used -> /clients -> /clients/v2

Is this for the dev experience? We shouldn't provide a endpoint with a schema that may change. Alternatively with additional metadata facilties you could perhaps advertise the default api version.

> Or do you wanna be more explicit, and the version needs to be specified in the path always?

Everything should be versioned for consistency. Here again metadata facilities could be needed to advertise the versions.

This will start to resemble the way kubernetes versions apis - /group/version/[namespace/name]/resource

Do we want to consider a group concept, or is the thinking that every root resource is separately versioned - even if there are relationships between them?

mabartos:
> Is this for the dev experience? We shouldn't provide a endpoint with a schema that may change. Alternatively with additional metadata facilties you could perhaps advertise the default api version.

Yes, the intention is mostly for dev experience - having the same convention as for features. It means, that we recommended to our users to use the versioned keys for features, but we accept also the unversioned feature name to be enabled. 

However, as we start this whole new thing, we can do some changes right away. It means that we can decide to be always on the consistent right way - requiring the version to be always specified. 

To be fully consistent, as mentioned, we should have the API path for clients as follows:
```
/admin/api/realms/<realm-api-version>/<realm-name>/clients/<client-api-version>/*
```
So f.e:
```
/admin/api/realms/v1/master/clients/v2/
```

I'm ok to have it like this as it's an API and the dev experience does not have to be so "strict" as we have the OpenAPI and swagger anyway.

@vmuzikar @keycloak/cloud-native-maintainers WDYT?

----------------------------------

> Alternatively with additional metadata facilties you could perhaps advertise the default api version.

Do you have any brief example/use-case how it could look like and how it could be used? Thanks

----------------------------------

> This will start to resemble the way kubernetes versions apis - /group/version/[namespace/name]/resource. Do we want to consider a group concept, or is the thinking that every root resource is separately versioned - even if there are relationships between them?

@shawkins Not sure how it'd look like. Do you have any Keycloak specific example how we could leverage groups?

shawkins:
> Yes, the intention is mostly for dev experience - having the same convention as for features.

I think the belief for features is that most version changes won't have incompatiblity concerns.

I don't think the assumption holds here as for example the difference between v1 and v2 is that they are completely incompatible. It would be best not to offer this style of endpoint from the start.

> To be fully consistent, as mentioned, we should have the API path for clients as follows:

This is part of the issue I meant about relationships between the entities. I'm not sure something like `/admin/api/realms/v1/master/clients/v2/` makes sense.

Unless we now know the relationship structure for all api versions moving forward, then we probably don't want this kind of cross-version relationship. There could for example be invalid combinations moving forward that is a realm v3 might only support foo-clients, not clients. Then do we start the versioning of foo-clients at v1 or skip directly to v3?

I realize this is a contrived example, but hopefully it conveys the problem.

> Do you have any brief example/use-case how it could look like and how it could be used? Thanks

For example https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#definitions - they have preferred group version defined. Clients could fetch the metadata, then know what default version is.

> @shawkins Not sure how it'd look like. Do you have any Keycloak specific example how we could leverage groups?

If there are distinct subgraphs of resources under realms, they could be grouped.

mabartos:
> It would be best not to offer this style of endpoint from the start.

@shawkins Thanks. +1 

> Unless we now know the relationship structure for all api versions moving forward, then we probably don't want this kind of cross-version relationship.

Agree.

> If there are distinct subgraphs of resources under realms, they could be grouped.

Let's start simple, with not complicating things for Client API v2 for now, and solves the grouping of other potential APIs later - those would have a different path anyway, so there should not be any clash with that what we introduce now.

mabartos:
@keycloak/cloud-native @shawkins Updated (also PR description).

vmuzikar:
I don't like mixing the schemes (i.e. `/admin/api/realms/<realm-api-version>/<realm-name>/clients/<client-api-version>/*`) either. What if we decoupled the realms from the version like this: `/admin/api/clients/<client-api-version>/<realm-name>/*`. I.e. having the realm represented only as a path param _inside_ the Clients API.

vmuzikar:
CC @stianst

mabartos:
> What if we decoupled the realms from the version like this: /admin/api/clients/<client-api-version>/<realm-name>/*. I.e. having the realm represented only as a path param inside the Clients API.

@vmuzikar Thanks for the comment. I don't think it's a proper resource-based approach, as there should be a hierarchy of these resources (clients are part of the realm). 

------------------

TL;DR I think the `/admin/api/realms/<realm>/clients/v2` is ok (Realm API v2 is quite a far future IMHO).

Having only the `/admin/api/clients/v2` path requires having the realm identified somehow. It means either via path/query param, or have it include it in the payload as "realm" attribute of the client rep. The path/query param does not really make sense to me as people would be confused on how to even use it. Having it in the payload is not so bad in this case, but I'd prefer to have the separation on the resource level (path level) for better isolation.

vmuzikar:
> as there should be a hierarchy of these resources

I guess I was trying to say that this ^^ kinda contradicts the request for having distinct versions of the individual APIs. Either it's a single API with proper structure, or it's separate APIs with just references to cross-API resources

> Realm API v2 is quite a far future IMHO

But what then? Are we ok with `/admin/api/realms/v2/<realm>/clients/v2`?

vmuzikar:
To summarize an offline discussion. We agreed on the following two alternatives:
1. `/admin/api/<api-group>/<api-version>/<realm-name>/*`
   * E.g. `GET /admin/api/clients/v2/my-realm/clients/my-client-id`
2. `/admin/api/<api-group>/<api-version>/realms/<realm-name>/*`
   * E.g. `GET /admin/api/clients/v2/realms/my-realm/clients/my-client-id`

Option 1 doesn't leave our door open if we ever needed global resources that are not tied to any realm, option 2 does. Another alternative would be to use a reserved word in place of a realm name should a resource be considered global. E.g. `GET /admin/api/clients/v2/GLOBAL/clients/my-client-id` would represent a hypothetical client that is not belonging to any realm.

I'd vote for option 1 as it doesn't imply the realm is a resource – having `/realms/<realm-name>` in a path suggests I can do e.g. `GET /admin/api/clients/v2/realms/my-realm` which is invalid for getting the realm resource as that's the Client API group (`/admin/api/clients`).

@stianst What are your thoughts? Should we even consider the possibility for global resources?

shawkins:
> I'd vote for option 1 as it doesn't imply the realm is a resource – having /realms/<realm-name> in a path suggests I can do e.g. GET /admin/api/clients/v2/realms/my-realm which is invalid for getting the realm resource as that's the Client API group (/admin/api/clients).

Realm is a built-in concept of containment. Here we are using exactly the same form of URL as Kubernetes - `/group/version/namespaces/<namespace-name>/resource` - and it's well understood that you aren't able to retreive the namespace via `/group/version/namespaces/<namespace-name>`, nor is it suggested you can in any of the api documentation or spec. Also most of the time users won't be hand-crafting the urls - it will be done via a client / tooling.

mabartos:
@shawkins _btw. I added the code formatting for your examples, because the <> content was not visible._ 

---------------

After yesterday's discussion, I would opt for option 2 and basically follow the Kubernetes approach. Yesterday, I had concerns about the intuitiveness of the path for users, but as was mentioned, it's an API that will be used by the generated clients. We'll have a nice, polished OpenAPI spec, with nice endpoint visibility.

I'd say the namespace|realm should be clear from the first look and the `/realms/` should be there the same as there's the `/namespaces/` in k8s. Having the reserved word for the realm name is very error-prone. With the option 2, we can be prepared more for the future, and we do not risk any issue on our users/customers side when potentially they have the realm named like that (even this small risk is there).

@shawkins @vmuzikar So, I'm for Option 2 as there are more benefits than tradeoffs.

mabartos:
@vmuzikar @shawkins Updated based on the Option 2 now. Please review, and we can further discuss whether Option 1 or Option 2.

stianst:
I see realm as the tenant, so should be the base of the URL, something like:

```
<my realm URL>/admin/clients/v2
```

This allows us to in the future to have things like virtual hosts that points to a realm; it also makes it easier to use different API groups for the same realm, as the base URL doesn't change.

Also, adding that in case you want to move a realm to a different cluster, today that can be done by changing the core protocols URLs (`/realms/myrealm`) and the core admin API ('/admin/api/myrealm`), but if you add the realm name into each API group, then that becomes more or less impossible to do.

mabartos:
@stianst Having the realm defined in the beginning looks good to me. 

So by default, it might look like this:

```
localhost:8080/realms/test/admin/clients/v2
``` 

Is it what was intended, right? 

@vmuzikar @shawkins WDYT?

shawkins:
> This allows us to in the future to have things like virtual hosts that points to a realm; 

Are you suggesting configuring a hostname-admin per realm?

> Also, adding that in case you want to move a realm to a different cluster, today that can be done by changing the core protocols URLs (/realms/myrealm) and the core admin API ('/admin/api/myrealm`), but if you add the realm name into each API group, then that becomes more or less impossible to do.

Can you expand on what is impossible here?

> Is it what was intended, right?

If I understand the proposal correctly this gives us:

list clients - /realms/test/admin/clients/v2/clients
get the realm - /realms/test/admin/v2 (which doesn't seem consistent)
public stuff - /realms/test/... 
/serverinfo

This also complicates the convention we have about what is an admin url as that determination is not based upon the root of the path, but on something in the middle. And that would sit along side the existing stuff including:

/admin/test/console
/admin/serverinfo
/admin/realms/test ...
/realms/...

stianst:
> > This allows us to in the future to have things like virtual hosts that points to a realm;
> 
> Are you suggesting configuring a hostname-admin per realm?
> 
> > Also, adding that in case you want to move a realm to a different cluster, today that can be done by changing the core protocols URLs (/realms/myrealm) and the core admin API ('/admin/api/myrealm`), but if you add the realm name into each API group, then that becomes more or less impossible to do.
> 
> Can you expand on what is impossible here?

The more API groups you have the more rules you'll need in a reverse proxy to route one realm to one cluster and another realm to another cluster.

> 
> > Is it what was intended, right?
> 
> If I understand the proposal correctly this gives us:
> 
> list clients - /realms/test/admin/clients/v2/clients get the realm - /realms/test/admin/v2 (which doesn't seem consistent) public stuff - /realms/test/... /serverinfo
> 
> This also complicates the convention we have about what is an admin url as that determination is not based upon the root of the path, but on something in the middle. And that would sit along side the existing stuff including:
> 
> /admin/test/console /admin/serverinfo /admin/realms/test ... /realms/...

stianst:
My main point here is that `realm` should be before the API group name, not after. Not quite sure how it should look like, but think we'd probably still want a separate URL for admin endpoints.

Admin v1 endpoints are on `/admin/realms/<realm name>`, while non-admin endpoints are at `/realms/<realm name>`. I also never really liked the fact that we have `/realms/` in there.

We could perhaps do something like the following for the new Admin APIs:

* `/admin/<myrealm>/clients`  -- default version for clients API, which would be v2 for now
* `/admin/<myrealm>/clients/v2|v3`  -- specific version
* `/admin/<myrealm>/realm` -- default version for realms API, again v2 would be the default here

An alternative would be to go head and just start a new and better resource paths in general, not moving other things for now, but thinking how the paths could look like in the future:

* `<realm base url>/admin`
* `<realm base url>/protocols/oidc`

Where realm base URL by default is `<keycloak base url>/<realm name>`, but we would eventually also support things like lookup realm from other places in the URL. For example `https://myrealm.mycompany.com` would point to `myrealm`, while `https://myother.mycompany.com` would point to `myother`. Perhaps with the ability to also specify aliases/virtual hosts for realms, for example `https://something.something.org` has a mapping to `mysomethingrealm` in server config.

One thing that needs some special handling is the ability to list realms for the `master` realm. That may be a special endpoint only available to the master realm (something like `/admin/master/realms` with no sub-resources, just that one single listing realms endpoint.

mabartos:
So, as a summary of the conversation above: 
- [x] Realm should be specified before the API group in the path
- [x] We should distinguish between admin and non-admin endpoints
- [x] Would be good to comply with the convention of using the `/admin/*` prefix
- [x] Possibly remove the `/realms` keyword from the path

It means that something like `/admin/<myrealm>/clients/v2|v3` should comply with all these requirements.

Maybe it'd be good to move a little bit further and follow the Admin console example:
The `/admin/<realm>/console` accesses the admin console. In this example, the `console` might be some kind of admin app for the realm.

Wouldn't be nice to have the `api` as another admin app for the realm? 
It means that we could have the `/admin/<realm>/api` for API groups.
- For the new Clients API group, the path would be `/admin/<realm>/api/clients/v2`.
- For other API groups (like users), the path would be `/admin/<realm>/api/users/v2|v3`, etc.

These API groups would be separated from the other Admin stuff.

WDYT? @stianst @shawkins @vmuzikar

stianst:
> So, as a summary of the conversation above:
> 
> * [x]  Realm should be specified before the API group in the path
> * [x]  We should distinguish between admin and non-admin endpoints
> * [x]  Would be good to comply with the convention of using the `/admin/*` prefix
> * [x]  Possibly remove the `/realms` keyword from the path
> 
> It means that something like `/admin/<myrealm>/clients/v2|v3` should comply with all these requirements.
> 
> Maybe it'd be good to move a little bit further and follow the Admin console example: The `/admin/<realm>/console` accesses the admin console. In this example, the `console` might be some kind of admin app for the realm.
> 
> Wouldn't be nice to have the `api` as another admin app for the realm? It means that we could have the `/admin/<realm>/api` for API groups.
> 
> * For the new Clients API group, the path would be `/admin/<realm>/api/clients/v2`.
> * For other API groups (like users), the path would be `/admin/<realm>/api/users/v2|v3`, etc.
> 
> These API groups would be separated from the other Admin stuff.
> 
> WDYT? @stianst @shawkins @vmuzikar

+1

Can we also have a default version available on `/admin/<realm>/api/clients`? Each major release would remain on the same default version for minor releases, and it could be updated to the latest version on a new major release.

mabartos:
> Can we also have a default version available on /admin/<realm>/api/clients? Each major release would remain on the same default version for minor releases, and it could be updated to the latest version on a new major release.

@stianst I got a similar idea, but as @shawkins mentioned (and as we discussed in the CN team), it'll be better to require the version for consistency. The API will be primarily consumed by the generated clients and not crafted by hand, so we should not be so strict about the "dev" experience (some discussion above is related to this).

Even though the major version release is relevant for quite a long time, we should be on the safe side and not change the behavior "automagically" just by stating it in the docs (@shawkins talked about the potential metadata that could advertise the default version,...). If somebody wants to use the new version of the API, the version must be explicitly set. Otherwise, we might face some issues in the future. 

cc: @vmuzikar @shawkins

stianst:
> > Can we also have a default version available on /admin//api/clients? Each major release would remain on the same default version for minor releases, and it could be updated to the latest version on a new major release.
> 
> @stianst I got a similar idea, but as @shawkins mentioned (and as we discussed in the CN team), it'll be better to require the version for consistency. The API will be primarily consumed by the generated clients and not crafted by hand, so we should not be so strict about the "dev" experience (some discussion above is related to this).
> 
> Even though the major version release is relevant for quite a long time, we should be on the safe side and not change the behavior "automagically" just by stating it in the docs (@shawkins talked about the potential metadata that could advertise the default version,...). If somebody wants to use the new version of the API, the version must be explicitly set. Otherwise, we might face some issues in the future.
> 
> cc: @vmuzikar @shawkins

I don't agree with that - sure a client library or the CLI should use an explicit version, but that doesn't mean some will just want to use the latest and greatest. From what I've seen around REST API versioning there is typically a default version, and specifying/using the version number is optional.

shawkins:
> From what I've seen around REST API versioning there is typically a default version, and specifying/using the version number is optional.

The concern is that such a mode is really only valid for ad hoc usage. The parameters, in-coming and out-going payload structure, etc. are effectively unknown to the caller. The only thing they have correctly guessed is a path.

We either need to:
- Prominantly warn that usage of these endpoints is only for dev / ad hoc scenarios - otherwise users could consider any default version change breaking.
- Document the supported way to use them, including describing when / why the version changes, and add a metadata endpoint to determine the default version

If we don't do the latter, it would be best if these endpoints don't show up in any of our api contracts.

vmuzikar:
> * Prominantly warn that usage of these endpoints is only for dev / ad hoc scenarios - otherwise users could consider any default version change breaking.
> * Document the supported way to use them, including describing when / why the version changes, and add a metadata endpoint to determine the default version

+1 to both. Being clear in docs and provide some metadata won't hurt and will provide the necessary clarity.

stianst:
Wonder if a version header would be better than a different path?

Makes it simpler to switch version as you only need to make the changes not change the path.

And more importantly it may be easier for us to handle different versions of the API that way?

vmuzikar:
> Makes it simpler to switch version as you only need to make the changes not change the path.

In case we needed to bump the API version, it means we needed to do some radical braking changes. Changing the version in the path then will be the least of your concerns.

mabartos:
### Summary

- Path changed `admin/api/v2/realms/master/clients/*` -->  `admin/master/api/clients/v2/clients/*`
- No need for the RealmApis
- Created the ClientsApiGroup 
- When no version is specified, the default is used -> `admin/master/api/clients/` = `admin/master/api/clients/v2`
- Warned about not using the version for the API group in Javadoc
- Added HTTP method annotation to impl classes for better visibility
- Then, there might be some other client-related resource like `admin/master/api/clients/v2/scopes`,...

@vmuzikar @shawkins @stianst WDYT? We can continue with improvements in follow-up tasks.

shawkins:
Just wanted to add more from our v2 meeting. What I'm basically advocated for is along the lines of front-end url == virtural host, which is discussed in https://github.com/keycloak/keycloak/issues/14623 and related issues.

To support something like this we need an additional realm option to omit realms/name in paths we construct.

Just as today when you set a front-end url at a proxy level you'd still need the mapping:

`front-end -> kc-host / realms / name`

But instead of accessing keycloak via front-end / realms / name, users would use just front-end /

Admin access by default looks like:

`front-end / admin -> kc-host / realms / name / admin`

However URLs for server scoped resources - realms, server-info, etc. - don't make much sense. At worst every server scoped resource would need added to a rule in the form of - front-end / resource -> kc-host / admin / resource - and we'd have to ensure there could never be a conflict between global resource types and top level paths available from a public realm.

Users are likely to need an admin specific mapping anyway to restrict access. An alternative is to only have realms/name omitted for front-end urls, and require an additional admin mapping of:

`front-end / admin -> kc-host / admin`

Now you have:

`front-end / admin / realms / name -> kc-host / admin / realms / name`
`front-end / admin / server-info -> kc-host / admin / server-info`

@vmuzikar can the requirement to have node level specific routing of admin requests using urls like `kc-host / realms / name / admin` be elaborated on?

And it was brought up in the meeting we could support both conventions. That is regardless of how they look, just serve requests such as:

kc-host / realms / name / admin / server-info
kc-host / realms / name / admin / realms-api / v2 / realms / name
kc-host / realms / name / admin / realms-api / v2 / realms 

I'm not entirely sure, but I think to do this would need some additional realm level configuration option to indicate what admin paths should be formed.

cc @keycloak/cloud-native @sschu 

Also this doesn't have to be based just upon the front-end url - you can have realm specific admin urls as well.

vmuzikar:
@shawkins Virtual hosts for realm identification is definitely something we should look into in the future but I don't see it as  something we need to figure out now, it's not really conflicting with our current work around designing the realm-based multi-tenancy. We will need to have a good story for putting the realm name in the path regardless – not everyone will use virtual hosts.

---

To summarize today's discussion on the WG sync, we need to consider the following:
* Multi-tenancy support. For this, we want to have the realm name as sort of prefix/namespace, see [the previous comment](https://github.com/keycloak/keycloak/pull/44527#issuecomment-3631300771).
   * The main question is whether to place the realm name before or after `/admin/`. Placing it before has the pro is that all endpoints (admin, frontend) use a single prefix which helps with routing based on tenant/realm name. The con is more difficult filtering of admin endpoints as they don't start with `/admin`.
      * Before: `/realms/my-realm/admin/api/clients/v2/...`.
      * After:  `/admin/realms/my-realm/api/clients/v2/...`.
* Future theoretical Realms API. If the realm name is a prefix in the path (as specified in the previous point) this might get confusing very quickly.
   * Seems the only correct way is to make the Realms API a global resource. I.e. instead of having something like `/realms/my-realm/admin/api/realms/v2/realms/my-realm`, we'd have only `/admin/api/realms/v2/realms/my-realm`.
   * To avoid ambiguity, we'll need include in the path `/realms/<realm-name>` instead of just `/<realm-name>`.
      * An alternative to this would be some reserved word, something like `global` but major problem would be a migration path if someone already uses that realm name.
* `/api/` segment in the path. We might not need it. Instead, we'd suffix the API groups with `-api`. This would also differentiate the group names from the resources names.
   * E.g. instead of `/realms/my-realm/admin/api/clients/v2/clients/my-client`, we'd have `/realms/my-realm/admin/clients-api/v2/clients/my-client`.

**Overall proposal**
* Realms: `/admin/realms-api/v2/realms/my-realm`.
* Any other (realm scoped) resource: `/realms/my-realm/admin/clients-api/v2/clients/my-client-id`.
   * Alternatively: `/admin/realms/my-realm/clients-api/v2/clients/my-client-id`

shawkins:
> @shawkins Virtual hosts for realm identification is definitely something we should look into in the future

To clarify I'm referening virutal hosts because that's what @stianst was referencing above. However I'm trying, and probably failing, to relate this functionality to our realm-based front-end URL support and how I think users would actually want their full front-end urls to look. Let me see if I can make this less confusing.

> but I don't see it as something we need to figure out now, it's not really conflicting with our current work around designing the realm-based multi-tenancy. 

I see a strong overlap between the multi-tennancy concerns that are being brought up here and what URLs will be used externally and how that maps to internal URLs.

Let's say I have realm-a on cluster 1 and realm-x on cluster 2. Now I'm tasked with moving realm-a from cluster 1 to cluster 2.

Will realm-a and realm-x use the same hostname? Most likely not - which means this transition will involve setting a front-end url on realm-x. Or am I missing something here - are we considering using hostname-strict=false on cluster 2?

With the realm level front-end url set, I completely agree it is straight-forward to change how the proxy handles `front-end / realms / realm-a` from `cluster-1-host / realms / realm-a` to `cluster-2-host / realms / realm-a`.

A concern expressed above is that it will be difficult to handle admin access unless we have URLs with relative paths such as `/ realms / name / admin`. 

Again unless I'm missing something this is not generally possible today because we lack a per-realm ability to set the admin hostname url.

So let's assume that feature, or at least a feature that treats the realm front-end url as the admin url, has been added. If you are already handling `front-end / realms / realm-a`, I agree that also handles admin requests of this form as well. It was also mentioned that node specific routing may be needed for admin requests - or did I misunderstand something?

However using `/ realms / name / admin` makes server scoped resource URLs messier (realms, server-info), and more makes it more complicated to generally restrict admin access.

Is that a fair assessment?

> We will need to have a good story for putting the realm name in the path regardless – not everyone will use virtual hosts.

I'm not suggesting that they will - realm specific front-end URL differentiation could be based upon path, not just hostname.

stianst:
> > Makes it simpler to switch version as you only need to make the changes not change the path.
> 
> In case we needed to bump the API version, it means we needed to do some radical braking changes. Changing the version in the path then will be the least of your concerns.

We? Or the user of the API? That all depends on what they are using or not. I could for instance just be doing a CURL and the endpoints I'm using hasn't changed.

stianst:
We don't need to figure out the full virtual hosts story as part of this; there's a number of complications around that. Some may want a single virtual host that points to both the frontend and the admin endpoints. Some may want to expose it on different virtual hosts, or only use a virtual host for the frontend URLs, but not have a virtual host for the admin endpoints. Quite a lot needs changing to support this, and figuring out, hence why we haven't done that yet. My main point here was that the realm name must be before the API group, not after. Having it after made no sense anyways.

My suggestion for new endpoints are:

`/admin/<realm name>/api/<api group>/<version>`

Examples:

`GET /admin/myrealm/api/clients/v2` - returns list of clients
`POST /admin/myrealm/api/clients/v2` - creates a client
`GET /admin/myrealm/api/clients/v2/myclient` - returns the client with `clientId=myclient`

`GET /admin/myrealm/api/realm/v2` - returns the realm config
`PATCH /admin/myrealm/api/realm/v2` - partial update of the realm config
`POST /admin/myrealm/api/realm/v2` - complete update of the realm config

`GET /admin/master/api/realms/v2` - list all realms (only available on `master` realm)
`POST /admin/master/api/realms/v2` - create a new realm
`DELETE /admin/master/api/realms/myrealm/v2` - delete a realm

In all example URLs the version can be omitted, in which case the default version is used. I still would suggest using a header for version instead of path though.

mabartos:
EDIT: I see that @stianst  is suggesting something similar as described below.

> The main question is whether to place the realm name before or after /admin/. Placing it before has the pro is that all endpoints (admin, frontend) use a single prefix which helps with routing based on tenant/realm name. The con is more difficult filtering of admin endpoints as they don't start with /admin.

The situation around this is little bit different than using a single prefix. 

## Current Keycloak endpoints
We have Admin and non-Admin endpoints, like AdminRealmsResource and RealmsResource. 

### 1. Realms resource - `/realms/{realm}/*`
For protocol handling(like OIDC), brokering, login actions,... Ops that are executed by non-admin users. 

### 2. Admin Realms resource - `/admin/realms/{realm}/*`
We have this path for Admin realm resources (no server scoped resources).
It means that users/customers might have filtered the Admin API endpoints based on the prefix `/admin/*`.
When there's the `/realms/{realm}/admin`, the filtering will indeed be worse, as they would need to filter also the additional endpoints. 

### Summary 
If we're able to reuse the `/admin` root without ambiguity, why not do it? As we should distinguish between the admin and non-admin endpoints anyway, on the proxy level, there should not be a big difference in how we map the `front-end / realms / realm-a` and the `front-end / admin / realms / realm-a`, right?

Maybe I'm missing something, but I don't see a reason why we can't use this suggestion (with RESTEasy, no ambiguity should be present due to the literal segments preference,..):
- `/admin/realms-api/v2/realms/my-realm` - Realms API group (server scoped)
- `/admin/{realm}/clients-api/v2/` - Clients API group (realm-scoped) 

As for the multi-tenancy and proxy-level mapping, users/customers have to already deal with the current `/admin` endpoints anyway, right? @shawkins I probably don't understand all your points, but can you elaborate on why the suggestion above might be problematic?

vmuzikar:
@shawkins I'm not sure I entirely understand the concerns. Could you please elaborate a bit more?

> Now I'm tasked with moving realm-a from cluster 1 to cluster 2.

Could you elaborate what you mean here? It's be a single Keycloak instance, we wouldn't be moving realms. Do you mean swapping hostnames between two realms?

> Again unless I'm missing something this is not generally possible today because we lack a per-realm ability to set the admin hostname url.

Yes, we do not support per realm admin URLs. That's something we'd need to tackle especially with virtual hosts but it's not something we need to find a solution for now, IMHO.

> It was also mentioned that node specific routing may be needed for admin requests - or did I misunderstand something?

Yes, that's my understanding too. In multi-tenancy scenarios, some customers might be more heavy on the admin API and might have e.g. dedicated nodes.

> However using `/ realms / name / admin` makes server scoped resource URLs messier (realms, server-info), and more makes it more complicated to generally restrict admin access.
> 
> Is that a fair assessment?

I think yes.

---

@stianst 

> We? Or the user of the API?

The API user.

> `GET /admin/myrealm/api/clients/v2` - returns list of clients
> `POST /admin/myrealm/api/clients/v2` - creates a client

I don't really like having the root resource of the API the client list. What if I want to manage client scopes under the client API? IMHO the root resource should be "nothing", i.e. to get clients, you'd do `GET /admin/myrealm/api/clients/v2/clients`, for client scopes you'd do `GET /admin/myrealm/api/clients/v2/scopes`, etc.

> `GET /admin/myrealm/api/realm/v2` - returns the realm config

> `GET /admin/master/api/realms/v2` - list all realms (only available on `master` realm)

This IMHO feels quite confusing. Realms API vs Realm API. Realms accessible only via master realm. It's not really consistent.

Do you have any concerns with server scoped resources (that do not belong to any realm)?

---

@mabartos If I understand your proposal correctly, you suggest to have `/admin` before the realm identifier, right? I think I tend to agree with you that it has more advantages.

mabartos:
> @mabartos If I understand your proposal correctly, you suggest to have /admin before the realm identifier, right? I think I tend to agree with you that it has more advantages.

@vmuzikar Yes. It'd also be consistent with our current Admin endpoints.

shawkins:
> Could you elaborate what you mean here? It's be a single Keycloak instance, we wouldn't be moving realms. Do you mean swapping hostnames between two realms?

At one point I believe a justification for having all urls rooted by `/ realms / name` was about how easy it would be to manipluate at the proxy level - including the ability to have a realm on one keycloak cluster migrated to another keycloak cluster. I was trying to expand upon that consideration, but perhaps my interpretation was not what was intended.

> Yes, that's my understanding too. In multi-tenancy scenarios, some customers might be more heavy on the admin API and might have e.g. dedicated nodes.

This is something I'd like to better understand. Is this done currently? And what is the basis for the dedication - are for example the cluser nodes somehow hetrogeneous?  Or is this seen as an extension of session affinity?

> /admin/{realm}/clients-api/v2/ - Clients API group (realm-scoped)

@mabartos this is missing the realms term in the path, it should be - /admin/realms/{realm}/clients-api/v2/ - Clients API group (realm-scoped)

> @shawkins I probably don't understand all your points, but can you elaborate on why the suggestion above might be problematic?

@stianst 's point from https://github.com/keycloak/keycloak/pull/44527#issuecomment-3659771731is that he wants the realm name before the api group. That mostly works, but is conceptually confusing for realms themselves and other server scoped admin resources, which @vmuzikar notes in https://github.com/keycloak/keycloak/pull/44527#issuecomment-3660174457

It might help to walk through more examples and to clearly separate the thinking about external and internal urls.

Let's say we know that my-realm lives on kc-host. If the realm name is not in some simple location at the root of **every** external URL path, then we can only do proxy based logic if there is a 1-1 mapping between the admin-url and kc-host. Otherwise:

| External Request  | Internal Request / Notes |
| ------------- | ------------- | 
| front-end/realms/my-realm  | kc-host/realms/my-realm - Nothing new here, straight-forward to handle and works as expected |
| admin-url/admin/realms/my-realm/clients-api/v2/clients  | kc-host/admin/realms/my-realm/clients-api/v2/clients - Similarly straight-forward, but does require separate handling that then front-end
| admin-url/admin/realms-api/v2/realms/my-realm  | I can determine the realm / host, but it's not straight-forward |
| admin-url/admin/server-info  | I don't know what host this should go to because there is no realm |

When would the admin-url not be 1-1 with the kc-host? Unless I'm missing something, this only comes into play when there are multiple keycloak clusters behind a single proxy - which relates to the case mentioned before about moving a realm from being hosted on one keycloak cluster to another.

The less exotic scenario here is wanting the proxy to do realm specific access restrictions. 

Finally yet another scenario that was broached involves per realm mappings to specific cluster nodes - I don't have a good grasp on when and why this is being done yet, but that could also influence what seems like an acceptable solution. 

We can make the per realm manipulation straight-forward if instead we do:

| External Request  | Internal Request / Notes |
| ------------- | ------------- | 
| admin-url/admin/realms/my-realm/realms-api/v2/realms/my-realm | kc-host/admin/realms/my-realm/realms-api/v2/realms/my-realm - I can easly determine the realm, but both internal and extenral URLs are ugly |
| admin-url/admin/realms/my-realm/server-info | kc-host/admin/realms/my-realm/server-info - "

Where do we go from here:

1. Use @stianst 's suggestion to create additional special case handling for realm URLs in particular
2. Just accept the ugliness and move on for now.
3. I believe we have ways to improve upon these URLs, but only if we are willing to expand on constructs like the realm specific front-end url. I understand that is even more new functionality and may only be confusing things more at this point, so bringing this up is just to further tha case that we could accept the ugliness and move on, then revisit this later.

mabartos:
@shawkins Thanks for the clarification. 

I've just updated the PR a little to include the `/realms` segment.

However, I had to add also the `/api` segment to distinguish between the current Admin API and the new one to prevent any ambiguity. 

So, the path looks like this: `/admin/realms/{realm-ame}/api/<api-group>/` now. 

@shawkins @vmuzikar Unfortunately, I can't allocate more time to this before Christmas, so feel free to update this PR (or create a new one) based on your discussion. 

Thanks!

stianst:
I'm a little bit lost in this conversation by now, and not really clear on what everyone is suggestion.

What's wrong with what I've been suggesting from the start `/admin/myrealm/api/clients/v2`?

* `/realms/` is just unnecessary and provides no value to have it there
* `GET ../clients/v2/clients` to list clients is just weird, and `scopes` will be on it's own API, not under client API. That's literally the whole point of API groups
* `/clients-api` is weird, it should be `/api/clients/`

sschu:
The assumption was that an api group subsumes different types of resources, that's why it would be `/admin/myrealm/api/clients/v2/clients` and not `/admin/myrealm/api/clients/v2`. So we could also have `/admin/myrealm/api/clients/v2/client-scopes`. And because of this doubling `...clients/v2/clients` the idea was to use `clients-api` instead of `api/clients`.
Does everybody agree that api groups have different types of resources?

vmuzikar:
After some offline discussion with @stianst, there are the following considerations:
* API groups will be encapsulating just a single resource. So Clients API will be for clients management only, not scopes etc. As a result it is ok to have e.g. `GET /admin/api/my-realm/clients/v2` for fetching all clients, and `GET /admin/api/my-realm/clients/v2/my-client-id` for fetching a single client.
* To support the future hostname to realm name mapping, all resources and APIs will need to be associated with a realm. I.e. there will be no real "global resources". We want to avoid the need of artificial hostname for global resources. As a result we don't need `/realms/` in the path.
   * Any resource that could be considered global will be either associated with the master realm (e.g. realms management), or with an arbitrary realm for given tenant (e.g. server info endpoint).
* Admin URLs will need to be per realm in the future. This is to allow use cases like:
   * `admin.my-realm.my-kc.org` mapping to `my-kc.local/admin/api/my-realm`
   * `my-realm.my-kc.org` mapping to `my-kc.local/my-realm`
* We won't be placing `/admin/` after the realm name, it will be at the beginning of path.

So the proposal is:
* Get clients: `GET /admin/api/my-realm/clients/v2`
* Get single client: `GET /admin/api/my-realm/clients/v2/my-client-id`
* Get realms: `GET /admin/api/master/realms/v2` (future – out of scope of Client API efforts)
* Get single realm: `GET /admin/api/my-realm/realm/v2` (future)
* Get server info: `GET /admin/api/my-realm/server-info` (future)

@stianst Does that accurately capture the discussion we had earlier?

shawkins:
> API groups will be encapsulating just a single resource. So Clients API will be for clients management only, not scopes etc. As a result it is ok to have e.g. GET /admin/api/my-realm/clients/v2 for fetching all clients, and GET /admin/api/my-realm/clients/v2/my-client-id for fetching a single client.

While this adds more flexibility over a global version, I would add:
- we should avoid like the plague adding additional versions. Whenever possible we should evolve, not immediately break the schema.
- it could be more confusing than with api groups to introduce a new resource type. Does its version relate to the other existing versions - will it start at v1, v2, etc.?
- you may still end up needing to coordianate breaking changes across highly related resource types

Also it would be good to clarify with any non-global versioning scheme what the relationship is to the feature version. In the prototype we have just a single feature version for the entire admin api. 

Related to this how, or should, admins be able to configure which serving versions are available?

> We want to avoid the need of artificial hostname for global resources. As a result we don't need /realms/ in the path.

Having api in the path serves a similar purpose - it's a sibling to the v1 /admin/realms

> So the proposal is:

From a client perspective needing to have specialized URL construction handling for a single realm isn't ideal, but it's also not that difficult to manage.

For the future it could be good to version server-info as well.

Finally just like with other proposals, there's still a potential v1 conflict, this time with admin/api/console - but as of now we haven't proposed to actually have anything served of that endpoint so it's probably ok to ignore this.

sschu:
Do we need the `/api`? Is there any case where we wouldn't have `/api` after `/admin`? I am wondering if that could interfere with current URLs like if somebody calls a realm `api` and the path to its admin console would be `/admin/api/console`? If we leave the `api` out, we would always have the realm name after `/admin`.

vmuzikar:
@shawkins As for the versioning concerns. I see each API version also a Feature version, i.e. you can selectively enable/disable the API versions. To do that we'll need to add the ability to enable multiple Feature versions simultaniously. But that's for the future, right now we have a single Client API feature.

Any subresources will be part of the parent resource API group. Adjacent resources (e.g. users and roles) will use separate API groups with independent versioning, but can use key references to the other resource as we don't expect keys would ever be affected by version bumps.

@shawkins @sschu The potential conflict with `/admin/api/console` can be solved by the `Content-Type` header.

shawkins:
>  I see each API version also a Feature version, i.e. you can selectively enable/disable the API versions.

Ok, so we'll have an admin api v1 feature, then a v(1+n) feature for each root resource type, and there is no presumed relationship between the same versions of different resources. If we add a new resource type, it will likely start at v2 to avoid confusion with the global v1 designation.

>  To do that we'll need to add the ability to enable multiple Feature versions simultaniously.

In general this reinforces the need to minimize additional versions - adding support for version ranges, min versions, etc. will add even more complexity on top of adding a lot of features.

> but can use key references to the other resource as we don't expect keys would ever be affected by version bumps.

This is true, but can still be confusing. Again with the case of adding a new type - if you need to reference that from an existing type, having to independently manage those versions is less than optimal.

> @shawkins @sschu The potential conflict with /admin/api/console can be solved by the Content-Type header.

It's probably good to note that v1 already has degenerate cases - a realm named realms, serverinfo, or console.

If for example you create a realm called realms, then issue GET /admin/realms/console - the resteasy path matching logic will resolve that to AdminRoot.getRealmsAdmin, not getAdminConsole. Content-Type nor Accept headers come into play as neither operation is looking for them, and those do seem to be one of the last considerations in general.

We all agree the v2 proposal adds api to the problematic set of realm names. In addition to the html expected from GET /admin/api/console, subpaths under console expect and produce different media types. While I don't see any conflict yet with v2 resources, it seems like it will be more complicated to address than adding produces / consumes to v1 / v2 methods and could further complicate things down the road. It's not ideal but if worst comes to worst we can just say that api was an unfortunate naming choice for a realm - just as realms, serverinfo, and console currently are.

vmuzikar:
> Ok, so we'll have an admin api v1 feature, then a v(1+n) feature for each root resource type, and there is no presumed relationship between the same versions of different resources. If we add a new resource type, it will likely start at v2 to avoid confusion with the global v1 designation.

Yes, correct.

> In general this reinforces the need to minimize additional versions

Yeah, introducing a new API version should be a rare event.

> if you need to reference that from an existing type, having to independently manage those versions is less than optimal.

Could you elaborate on that?

> We all agree the v2 proposal adds api to the problematic set of realm names. In addition to the html expected from GET /admin/api/console, subpaths under console expect and produce different media types. While I don't see any conflict yet with v2 resources, it seems like it will be more complicated to address than adding produces / consumes to v1 / v2 methods and could further complicate things down the road. It's not ideal but if worst comes to worst we can just say that api was an unfortunate naming choice for a realm - just as realms, serverinfo, and console currently are.

I think we can't fully avoid the risk of conflicts– with or without `/api/` in the path.

shawkins:
> Could you elaborate on that?

With api groups you shouldn't have references in one resource to other resources in the group that don't exist in that version.

Let's say we have an api group with version v1 and v2. v2 adds a new resource and that is referenced in a v2 resource that has both a v1 and a v2 version. Everything in my highly related api group is logcally consistent at v2. Fetching the v1 form of the exisitng resource shouldn't even have that new reference present - because it wouldn't be consistent.

When you start developing against our fine-grained version management you'll basically be creating a version lock file:

realms=v2
clients=v3
roles=v2
users=v2
...

Based upon the latest versions supported by the Keycloak you initially target for development against.

From there the developer will be responsible for individually updating these versions. And if you for example want to start using new resource v2, that may also require you to update to users=v3 - not an obvious relationship.

mabartos:
> So the proposal is:
> 
> * Get clients: `GET /admin/api/my-realm/clients/v2`
> * Get single client: `GET /admin/api/my-realm/clients/v2/my-client-id`
> * Get realms: `GET /admin/api/master/realms/v2` (future – out of scope of Client API efforts)
> * Get single realm: `GET /admin/api/my-realm/realm/v2` (future)
> * Get server info: `GET /admin/api/my-realm/server-info` (future)

Updated based on the proposal here: https://github.com/keycloak/keycloak/pull/44527#issuecomment-3665780347

@vmuzikar @shawkins Is it allright? Or do you have any more suggestions that we could use for this specific Client Admin API?

It seems there's a bug not related to this PR: https://github.com/keycloak/keycloak/issues/45215

shawkins:
> @vmuzikar @shawkins Is it allright? Or do you have any more suggestions that we could use for this specific Client Admin API?

No further suggestions from me. My last couple of comments summarized any remaining thoughts.

shawkins:
> > @vmuzikar @shawkins Is it allright? Or do you have any more suggestions that we could use for this specific Client Admin API?
> 
> No further suggestions from me. My last couple of comments summarized any remaining thoughts.

@vmuzikar it may not have been clear in the meeting, or previous comments, so let me state it this way - I'm most in favor of versioned api groups, a global version, or fine-grained versioning - in that order. It seems like we're committed to fine-grained versioning and the associated feature management, so we can just keep moving forward at this point.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminRootV2.java
```diff
@@ -19,18 +21,19 @@ public class AdminRootV2 {
...
-    public AdminApi adminApi() {
+    @Path("{realmName}")
+    public AdminApi adminApi(@PathParam("realmName") String realmName) {
         checkApiEnabled();
-        return new DefaultAdminApi(session);
+        return new DefaultAdminApi(session, realmName);
     }
 
-    @Path("{any:.*}")
+    // TODO Fix preflights
```
vmuzikar, line 30: @mabartos Do we have a follow-up issue for the preflights fix?

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/DefaultAdminApi.java
```diff
@@ -27,12 +31,15 @@ public DefaultAdminApi(KeycloakSession session) {
             throw new NotAuthorizedException("Wrong permissions");
         }
         this.realmsAdminResource = new RealmsAdminResource(session, auth, new TokenManager());
+        this.realmAdminResource = realmsAdminResource.getRealmAdmin(realmName);
     }
 
-    @Path("realms")
+    @Path("clients/{version:v\\d+}")
```
vmuzikar, line 37: Just wanted to highlight one caveat with this approach: we don't have a default version. This is fine by me but I know @stianst had some concerns with that. In any case, we could address that as a follow-up if needed.

vmuzikar (dismissed):
LGTM, thank you.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/DefaultAdminApi.java
mabartos: @vmuzikar I've added the default handling for now, but then, we will create a possible conflict with the client ID and the version. 

It means that when a client is named `v3`, it cannot be managed as it's considered as a Client API group version and not the client. 

Or when the client is named `v2`, and this call is executed `GET /admin/api/master/clients/v2` we can't really determine the intention - whether to get all clients (the v2 is a version), or get the client named `v2` (the v2 is a clientID).

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/AdminRootV2.java
mabartos: https://github.com/keycloak/keycloak/issues/45402

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/DefaultAdminApi.java
shawkins: We shouldn't add in default handling initially. This is only useful for those manually typing in URLs, requires additional documentation, etc. The ambiguity issue should be highlighted on any follow-up, and may necessitate using a different convention - clients/v - to mean latest.

mabartos: @shawkins Agree. Done

mabartos:
@shawkins @vmuzikar Ready for review.

vmuzikar (approved):

PR #44530, opened 2025-11-27
User session deleted events for invalid sessions

### LINKED ISSUE #44513: Deleting invalid sessions when remember-me is disabled does not trigger events
### Description

The `UserSessionExpirationLogic.deleteInvalidSessions()` deletes the sessions from the database, but does not trigger a session deletion event. 

Maybe this can be worked-around by having a lifespan or idle time of `0` when remember me is disabled, so we can reuse the existing logic.

pruivo:
Closes #44513

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/session/UserSessionExpirationLogic.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/session/UserSessionExpirationLogic.java
@@ -305,7 +305,7 @@ final class UserSessionExpirationLogic {
     private static void runInBatches(KeycloakSessionFactory sessionFactory, KeycloakSessionTaskWithResult<Boolean> task, Runnable afterBatchAction) {
         boolean hasMore;
         do {
-            hasMore = runJobInTransactionWithResult(sessionFactory, task);
+            hasMore = runJobInTransactionWithResult(sessionFactory, null, task, "session-expiration-task");
             afterBatchAction.run();
         } while (hasMore);
     }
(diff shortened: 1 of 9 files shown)


model/jpa/src/main/java/org/keycloak/models/jpa/session/UserSessionExpirationLogic.java
```diff
@@ -350,14 +302,12 @@ private static int updateRememberMeColumn(EntityManager em, boolean rememberMe,
...
-            return 0;
-        }
-        return em.createNamedQuery("deleteUserSessions")
-                .setParameter("offline", offlineStr)
-                .setParameter("userSessionIds", sessionIds)
-                .executeUpdate();
+    private static void runInBatches(KeycloakSessionFactory sessionFactory, KeycloakSessionTaskWithResult<Boolean> task, Runnable afterBatchAction) {
+        boolean hasMore;
+        do {
+            hasMore = runJobInTransactionWithResult(sessionFactory, task);
```
ryanemerson: Should we call the version of this method that allows a `String taskName` to be provided, that way we can have something more meaningful e.g. `session-expiration-task` instead of "Non-HTTP task"

pruivo: done

ryanemerson (approved):

ahus1 (approved):
Approving based on @ryanemerson's review

Issue #44159, opened 2025-11-12
Enhanced ObjectMapper instantiation for Client v2


vmuzikar:
### Description

Revisit how we get and configure the ObjectMapper. Consider using the Quarkus one.

shawkins:
The mapper is currently being handled specifically for rest: https://github.com/keycloak/keycloak/blob/fa3e964df7c9c8741f889c9b8705e6c5f5c05b2a/services/src/main/java/org/keycloak/services/util/ObjectMapperResolver.java

Quarkus offers https://quarkus.io/guides/rest-json#jackson

PR #44531, opened 2025-11-27
[admin-api-v2] Remove the unnecessary ObjectMapper from the Client API v2

mabartos:
There should be no need to use the workaround for the object mapper. We can reuse the existing one. 


shawkins:
The change to use this mapper was intentional, the follow up is #44159

mabartos:
@shawkins Could you briefly remind me why it's intentional to use a different object mapper?

shawkins:
> @shawkins Could you briefly remind me why it's intentional to use a different object mapper?

This is the mapper which is used by the v1 api, and the follow up issue to customize it and make it available in a better way.

mabartos:
@shawkins Ok, thanks. Closing now.

PR #44533, opened 2025-11-27
Polishing of sd-jwt SDK builder related methods

### LINKED ISSUE #44532: Polishing of sd-jwt SDK builder related methods
### Description

Few minor concerns with sd-jwt sdk:

- Some classes have `Builder` pattern, but despite of this, they use `public` constructors with many arguments for various combinations. This is typically not needed when using `Builder` pattern. Some examples are `KeyBindingJWT` or especially `IssuerSignedJWT` .

- There is no need to have lots of overriden `build` methods with various arguments, but usually one `build()` method with the arguments sent as separate methods to the `Builder` . AFAIK that is the main intent of the `Builder` pattern proper usage. Some examples are `KeyBindingJWT` or `SdJwt` classes.

mposolda:
closes #44532

@francis-pouatcha @Awambeng @IngridPuppet @forkimenjeckayang @Ogenbertrand @Captain-P-Goldfish Can you please review if those changes are OK with you? Thanks

### DIFF
--- a/core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
+++ b/core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
@@ -230,7 +230,10 @@ public class SdJwtVP {
         }
         String sd_hash = SdJwtUtils.hashAndBase64EncodeNoPad(unboundPresentation.getBytes(), getHashAlgorithm());
         keyBindingClaims.put(SD_HASH, sd_hash);
-        KeyBindingJWT keyBindingJWT = new KeyBindingJWT(keyBindingClaims, holdSignatureSignerContext);
+        KeyBindingJWT keyBindingJWT = KeyBindingJWT.builder()
+                .withPayload(keyBindingClaims)
+                .withSignerContext(holdSignatureSignerContext)
+                .build();
         sb.append(keyBindingJWT.getJws());
         return sb.toString();
     }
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
```diff
@@ -230,7 +230,10 @@ public String present(List<String> disclosureDigests,
         }
         String sd_hash = SdJwtUtils.hashAndBase64EncodeNoPad(unboundPresentation.getBytes(), getHashAlgorithm());
         keyBindingClaims.put(SD_HASH, sd_hash);
-        KeyBindingJWT keyBindingJWT = new KeyBindingJWT(keyBindingClaims, holdSignatureSignerContext);
+        KeyBindingJWT keyBindingJWT = KeyBindingJWT.builder()
+                .withPayload(keyBindingClaims)
+                .withSignerContext(holdSignatureSignerContext)
+                .build();
```
Captain-P-Goldfish, line 236: I haven't checked the `SDJwtVP` implementation yet in detail, but aren't we missing the header of the keybinding-jwt here?

Captain-P-Goldfish (approved):
except for the question that must not necessarily be handled in this PR everything looks good. And I like your change with the decoyClaims boolean :-)

core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
mposolda: @Captain-P-Goldfish We are not missing it as we're adding that in the `KeyBindingJWT` constructor. The `typ` claim is set to `kb+jwt` exactly as required by this specs https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-key-binding-jwt . But thanks for checking this!

PR #44536, opened 2025-11-27
[admin-api-v2] SPIs for ModelMapper and JakartaValidator

### LINKED ISSUE #44535: SPIs for ModelMapper and JakartaValidator
### Description

In the Admin API v2, we use the ModelMapper and JakartaValidator components. The purpose of the ModelMapper is to map entities/models to DTOs and the other way around. 

The purpose of JakartaValidator is to provide general validation mechanism for the API. It helps us to validate the incoming objects. 

These components can be used multiple times during one request and we should avoid multiple instantiation. 

When we provide the SPI for them, we'd have benefits like:
- Cache ModelMapper for the application lifecycle (initialize only once at server startup and then reuse)
- Cache JakartaValidator for the request lifecycle (initialize at every request, and subsequent calls to the validator are coming from the cached instance) (we might consider have them application scoped, but the init is done via CDI right now, so there might be issues... we can start like this for now)
- Have an abstract layer for the validation and model mapping
- Ability to potentially override the implementation of them if necessary

mabartos:
- Closes https://github.com/keycloak/keycloak/issues/44535
- The MapStruct and HibernateValidator should not lie in the `api` module
- Created the proposed `providers` module, where we can put the implementation
- More reasoning about benefits in the parent issue
- SPIs are not enabled based on the ClientAPI v2 feature - I think we can have them enabled by default + after merging https://github.com/keycloak/keycloak/pull/44527, we can potentially use the `DefaultAdminApi.isApiEnabled()`

@vmuzikar @shawkins @Pepo48 Could you please check it? Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1147,6 +1147,11 @@
             </dependency>
             <dependency>
                 <groupId>org.keycloak</groupId>
+                <artifactId>keycloak-admin-v2-providers</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.keycloak</groupId>
                 <artifactId>keycloak-admin-v2-rest</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/quarkus/runtime/pom.xml
+++ b/quarkus/runtime/pom.xml
@@ -426,6 +426,17 @@
 
         <dependency>
             <groupId>org.keycloak</groupId>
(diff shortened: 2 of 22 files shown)


shawkins:
Let's put any spi creation on hold until we're fully done with the functionality of the prototype.

mabartos:
@shawkins Sure, np.

shawkins:
Closing as out-of-date.

PR #44539, opened 2025-11-27
[admin-api-v2] Provide functionality on how to map from the v2 to v1 representation

### LINKED ISSUE #44538: [admin-api-v2] Provide functionality on how to map from the v2 to v1 representation
### Description

At this moment, we have a workaround in the `DefaultClientService` that creates some simple client in the DB, converts from the model to representation, remove it from the DB and then creates the specific client with the old DTO ClientRepresentation.

```java
            // dummy add/remove to obtain a detached model
            model = realm.addClient(client.getClientId());
            realm.removeClient(model.getId());

            mapper.toModel(model, client, realm);
            var rep = ModelToRepresentation.toRepresentation(model, session);
            model = clientsResource.createClientModel(rep);
```

https://github.com/keycloak/keycloak/blob/main/rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java#L66

We should avoid that and map directly the new representation to the old one (when we use the API v1 underhood)

mabartos:
- Closes #https://github.com/keycloak/keycloak/issues/44538
- Built on top of https://github.com/keycloak/keycloak/issues/44535
-------------------------------------------------

- Removes the workaround of the artificial client creation in the DB (more info in the issue)
- Maps directly V2 -> V1 representation
- Adds `equals()`+`hashCode()` for the equality check
- Maps better the attributes than before even for the model -> representation

There are some issues/concerns about the new ClientRep attributes (even for issue https://github.com/keycloak/keycloak/issues/43290):

1. `loginFlows` - It probably reffers to the flow overrides for the browser flow? I think we should still have the Map to be able to override any other flow. I think this might be useless from my understanding
2. Should we have the `auth.certificate` here? Isn't it better to have the "secret" of the used client authentication type (plain secret, JWT, cert) in the `auth.secret` field?
3. Roles -  So, do we want to have the roles included here? Even the service account roles? We potentially can, but as these roles are separate entities, it's harder to obtain them without some more code (which might not be a problem).
4. Additional fields - right now, we have only the logic that check the "allow list" of possible fields - we don't know the future of the additional fields anyway

But IMO, solving this is out of scope of this PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1147,6 +1147,11 @@
             </dependency>
             <dependency>
                 <groupId>org.keycloak</groupId>
+                <artifactId>keycloak-admin-v2-providers</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.keycloak</groupId>
                 <artifactId>keycloak-admin-v2-rest</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/quarkus/runtime/pom.xml
+++ b/quarkus/runtime/pom.xml
@@ -426,6 +426,17 @@
 
         <dependency>
             <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-admin-v2-providers</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.keycloak</groupId>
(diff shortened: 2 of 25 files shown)


vmuzikar (changes requested):
@mabartos Thank you for the PR. This won't be eventually needed. We currently leverage the v1 reps as a temporary workaround before we have a proper service layer. We should keep the temp logic at minimum as it will go away after 26.5.

mabartos:
Parent issue closed due to irrelevance. See the comment https://github.com/keycloak/keycloak/issues/44538#issuecomment-3595818526

PR #44541, opened 2025-11-27
[admin-api-v2] Create client does not return 201 status code

### LINKED ISSUE #44540: [admin-api-v2] Create client does not return 201 status code
### Before reporting an issue

### Area

admin/api

### Describe the bug

In the new Admin Client API v2, when you make a HTTP POST request to create a client, the 200 is returned, and not 201.

We should fix it. 

### Version

main

### Regression

### Expected behavior

Response status code is 201

### Actual behavior

Response status code is 200

### How to Reproduce?

Execute the POST request on that endpoint

### Anything else?

_No response_

mabartos:
- Closes #44540
- The POST endpoint now returns 201 response code
- For the `createOrUpdate()` PUT request, it also returns 201 when the client does not exist -> Is it ok for the operator and idempotency? As these response codes might be different
- As we're talking about polymorphic representation, we could use the `Response` type as the return type. WDYT?

### DIFF
--- a/rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
+++ b/rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
@@ -8,6 +8,7 @@ import jakarta.ws.rs.PATCH;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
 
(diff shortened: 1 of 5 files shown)


shawkins (approved):
LGTM, thanks @mabartos

sschu:
@mabartos I don't think idempotency requires to always get the same response code, only that the state of resources stays the same no matter how many times you send the same request.

mabartos:
@vmuzikar Do you want to review, or can we merge it? Thanks!

mabartos:
Merging now, as it's quite a straightforward fix.

PR #44549, opened 2025-11-28
Add MariaDB to MySQL description and specify SQL to support UTF-8

### LINKED ISSUE #44548: Add MariaDB to MySQL description and specify SQL to support UTF-8
### Description

MySQL was mentioned in [the public doc](https://www.keycloak.org/server/db#_configuring_unicode_support_for_the_database) regarding UTF-8 support, but there was no explicit reference to MariaDB.
The doc should mention it on MariaDB as well. Also, no specific SQL statement to support UTF-8 is described.

### Value Proposition

It provides customers with a clear and concrete method for UTF-8 support.

### Goals

Add specific instructions to the documentation.

### Non-Goals

Code enhancement.

### Discussion

_No response_

### Notes

_No response_

hokuda:
Closes #44548

### DIFF
--- a/docs/guides/server/db.adoc
+++ b/docs/guides/server/db.adoc
@@ -236,13 +236,18 @@ Unicode characters are supported only for the special fields for a Microsoft SQL
 The `sendStringParametersAsUnicode` property of JDBC driver should be set to `false` to significantly improve performance. Without this parameter,
 the Microsoft SQL Server might be unable to use indexes.
 
-=== Configuring Unicode support for a MySQL database
+=== Configuring Unicode support for a MySQL/MariaDB database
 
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
Thank you very much for contributing to our documentation! Your changes will be included in the upcoming 26.5 version in January.

ahus1:
@hokuda - thank you again for contributing this to the docs. 

I got one question: Could we make Keycloak simpler to use if we would check at startup that the database uses a UTF-8 character set? This would present situation where people do not read the docs. 

A potential behavior could be: 

* If the database has already been set up, and the charset is not UTF-8, print a warning at startup
* If the database is empty, show an error. People can then proceed by maybe adding another CLI option to force the creation of the database

If you think this makes sense, please create an enhancement issue. Ideally it would state the SQL statement for MariaDB/MySQL DB that you consider best to detect that. Once you have created the issue, post the issue number here so the two are connected. 

Thanks!

PR #44579, opened 2025-12-01
fix: refining arg handling

### LINKED ISSUE #44578: Further refine argument handling
### Description

We are manipulating the command line args in quite a few places. This should be minimized so that it's done in as few places as possible. It should also be based upon the picocli parsing, so that we're not attempting to duplicate the rules / handling in another place.

shawkins:
relying on picocli parsing as much as possible

Also changes the message about ensuring a build is used to just adding --optimized.

closes: #44578


shawkins:
Let me know if you want this broken up more.

I believe these changes also open up the possibility to reusing reaugmentations across Keycloak and CLITest invocations should we want to refine the logic more - cc @pedroigor 

The only non-distribution CLITests left are under org.keycloak.it.storage.database - most of the distribution related annotatations are not applicable, thus the change to the oracle test. Do we even want to keep these tests as they are, or could we create a follow-up to remove this style (either going full distribution, or moving over to the new test suite).

Another follow-up change: https://github.com/keycloak/keycloak/discussions/38894#discussioncomment-15124256

shawkins:
Marked with a hold as this is not needed for 26.5

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigArgsConfigSourceTest.java
```diff
@@ -30,23 +30,23 @@ public class ConfigArgsConfigSourceTest {
     @Test
     public void testParseArgs() {
         List<String> values = new ArrayList<>();
-        ConfigArgsConfigSource.parseConfigArgs(Arrays.asList("--key=value", "-cf", "file", "command", "arg", "--db", "value"), (key, value) -> values.add(key+'='+value), values::add);
```
mabartos, line 33: Does it mean that using the `-cf` is not possible?

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/ShowConfigCommandDistTest.java
```diff
@@ -108,7 +108,7 @@ void testConfigSourceNames(KeycloakDistribution distribution) {
         distribution.setEnvVar("KC_LOG", "file");
         distribution.copyOrReplaceFile(Paths.get("src/test/resources/ShowConfigCommandTest/quarkus.properties"), Path.of("conf", "quarkus.properties"));
 
-        result = distribution.run(String.format("%s=%s", CONFIG_FILE_LONG_NAME, Paths.get("src/test/resources/ShowConfigCommandTest/keycloak-keystore.conf").toAbsolutePath().normalize()), ShowConfig.NAME, "all");
```
mabartos, line 111: Is this change necessary?

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
```diff
@@ -1017,11 +1017,6 @@ public void duplicatedCliOptions() {
         assertThat(nonRunningPicocli.getErrString(), containsString("Unknown option: '--non-existing'"));
         onAfter();
 
-        nonRunningPicocli = pseudoLaunch("start-dev", "-Dsome.property=123", "-Dsome.property=456");
-        assertEquals(CommandLine.ExitCode.OK, nonRunningPicocli.exitCode);
-        assertThat(nonRunningPicocli.getOutString(), containsString("WARNING: Duplicated options present in CLI: -Dsome.property"));
```
mabartos, line 1022: So, we are not warned about system props duplications on CLI?

mabartos (commented):
@shawkins Nice work! Just a few questions.

mabartos:
> The only non-distribution CLITests left are under org.keycloak.it.storage.database - most of the distribution related annotatations are not applicable, thus the change to the oracle test. Do we even want to keep these tests as they are, or could we create a follow-up to remove this style (either going full distribution, or moving over to the new test suite).

@shawkins IMHO, It'd better to move them to the new testsuite as it's more prepared for the additional dependencies.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/cli/PicocliTest.java
shawkins: System properties won't make it to that point in the code. They are not passed as program arguments, and are consumed by the java launch.

quarkus/runtime/src/test/java/org/keycloak/quarkus/runtime/configuration/ConfigArgsConfigSourceTest.java
shawkins: As an argument to parseConfigArgs, yes. That routine is now only for --kc.prop=value options.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/ShowConfigCommandDistTest.java
shawkins: It is so the assertion about finding a cli property still works.

mabartos (approved):
LGTM

mabartos:
@vmuzikar @Pepo48 @pedroigor Anyone else want to review as you're in the list?

Pepo48 (approved):
LGTM. Thanks, Steve!

PR #44587, opened 2025-12-01
[admin-api-v2] Incorrect DTO/DAO mapping

### LINKED ISSUE #44586: [admin-api-v2] Incorrect DTO/DAO mapping
### Before reporting an issue

### Area

admin/api

### Describe the bug

For the Client API v2, when the required client representation is sent via PUT/POST, the resulting client is not properly set based on these attributes. We need to fix it. 

### Version

999

### Regression

### Expected behavior

All client properties are properly set

### Actual behavior

Only a few client properties are set

### How to Reproduce?

Create a ClientRepresentation v2 object, and send it via PUT/POST to the `/admin/api/v2/realms/master/clients` and check if the client has all the required properties/attributes. 

### Anything else?

_No response_

mabartos:
- Closes #44586
- Fixes `Representation --> Model` mapping
- Fixes `Model --> Representation` mapping
- The model mapping layer will be used even for the new polymorphic representations
- Declarative approach for roles (exactly the required for the client)
- Creates a service account when `serviceAccount.enabled` is true
- Declarative approach for service account roles (exactly the required for the client) - assign/unassign existing roles
- Adds `equals()` and `hashCode()` methods for better testing (would be nice to have Lombok for this :P)
- Simplified contract for ClientService (it is an internal thing that we use the Resources objects, so it does not have to be part of the service contract)
- Adds additional test cases to verify the proper update/create of a client + declarative approach

@vmuzikar @shawkins @Pepo48 Could you please check it? Thanks!

### DIFF
--- a/rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
@@ -1,12 +1,9 @@
 package org.keycloak.models.mapper;
 
 import java.util.Collections;
-import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
-import jakarta.ws.rs.core.Response;
 
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.KeycloakSession;
@@ -14,10 +11,7 @@ import org.keycloak.models.RealmModel;
 import org.keycloak.models.RoleModel;
 import org.keycloak.representations.admin.v2.ClientRepresentation;
 import org.keycloak.services.ServiceException;
-import org.keycloak.services.managers.ClientManager;
-import org.keycloak.services.managers.RealmManager;
 
-import org.mapstruct.AfterMapping;
 import org.mapstruct.Context;
 import org.mapstruct.Mapper;
 import org.mapstruct.Mapping;
@@ -78,30 +72,6 @@ public interface MapStructClientModelMapper extends ClientModelMapper {
         var model = realm.addClient(rep.getClientId());
         realm.removeClient(model.getId());
         return model;
-    }
-
-    @AfterMapping
-    default void addRoles(@MappingTarget ClientModel model, ClientRepresentation rep, @Context RealmModel realm, @Context KeycloakSession session) {
-        Optional.ofNullable(rep.getRoles())
-                .orElse(Collections.emptySet())
-                .stream()
-                .filter(role -> model.getRole(role) == null)
-                .forEach(model::addRole);
-
-        // Service Account roles
-        var serviceAccount = rep.getServiceAccount();
-        if (serviceAccount != null && serviceAccount.getEnabled() && !serviceAccount.getRoles().isEmpty()) {
-            new ClientManager(new RealmManager(session)).enableServiceAccount(model);
-            var sa = session.users().getServiceAccount(model);
-
-            serviceAccount.getRoles().forEach(role -> {
-                var foundRole = realm.getRole(role);
-                if (foundRole == null) {
-                    throw new ServiceException("Cannot assign role to the service account (field 'serviceAccount.roles') as it does not exist", Response.Status.BAD_REQUEST);
-                }
-                sa.grantRole(foundRole);
-            });
-        }
     }
 
     @Named("isPublicClientPrimitive")
--- a/rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
@@ -1,36 +1,119 @@
 package org.keycloak.models.mapper;
 
+import java.util.Collections;
+import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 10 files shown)


rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
```diff
@@ -1,46 +1,133 @@
...
-    @Override
-    void toModel(@MappingTarget ClientModel model, ClientRepresentation rep, @Context RealmModel realm);
+    /*-------------------------------------*
+     *          HELPER METHODS             *
+     *-------------------------------------*/
+    @ObjectFactory
+    default ClientModel createClientModel(@Context RealmModel realm, ClientRepresentation rep) {
+        // dummy add/remove to obtain a detached model
+        var model = realm.addClient(rep.getClientId());
+        realm.removeClient(model.getId());
```
vmuzikar, line 73: I don't think mapping logic should be doing this but it's ok for now as a way to avoid model layer refactoring, and we did that before as well.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
```diff
@@ -1,46 +1,133 @@
...
+    default void addRoles(@MappingTarget ClientModel model, ClientRepresentation rep, @Context RealmModel realm, @Context KeycloakSession session) {
+        Optional.ofNullable(rep.getRoles())
+                .orElse(Collections.emptySet())
+                .stream()
+                .filter(role -> model.getRole(role) == null)
+                .forEach(model::addRole);
+
+        // Service Account roles
+        var serviceAccount = rep.getServiceAccount();
+        if (serviceAccount != null && serviceAccount.getEnabled() && !serviceAccount.getRoles().isEmpty()) {
```
vmuzikar: I assume this won't remove roles and disable the service account on update, right? For now, we should probably just leverage the v1 logic here...

vmuzikar (changes requested):
@mabartos Thank you for the PR!

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
mabartos: Yes, good point. I'll try to reuse the v1 logic.

mabartos: Done.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/MapStructClientModelMapper.java
mabartos: I'd say it's better fit than the service itself as the mapping logic transforms the rep to model here. But we can improve it more later.

mabartos:
@vmuzikar Updated. Right now, the logic for roles, and service account roles should be correct and reuse the v1. We can do any other improvements in the follow-ups. 

Updated the PR description.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
```diff
@@ -211,6 +212,21 @@ public String getCertificate() {
         public void setCertificate(String certificate) {
             this.certificate = certificate;
         }
+
+        @Override
+        public boolean equals(Object o) {
```
shawkins, line 217: Ideally we could use lombok for default equals and hashcode implementations.

@vmuzikar is there anything preventing us from doing that?

shawkins (approved):
LGTM, thanks @mabartos

vmuzikar (approved):
LGTM, thanks.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/ClientRepresentation.java
vmuzikar: We can do as a follow-up but not sure about introducing this dependency (if it's not there already as a transitive one).

mabartos: Yes, I had the same idea :) 

I'll try to check it.

mabartos: @vmuzikar @shawkins The Lombok dependency is not included, not even as a transitive one. Lombok does not have the Apache 2.0 license, and [no license exception](https://github.com/cncf/foundation/blob/main/license-exceptions/CNCF-licensing-exceptions.csv) on the CNCF side either.

So, probably not a way forward.

shawkins: The way we'd use Lombok is as a provided dependency only for source processing. It will not be included in the runtime.

mabartos: @shawkins Ok, that's right. AFAIK, CNCF might probably not restrict this.

PR #44592, opened 2025-12-01
fix: allowing settable connection request timeout

### LINKED ISSUE #44500: HttpClient provider should allow connectionRequestTimeout to be settable
### Description

#44408 showed that http requests which are not completing in a timely fashion will then also cause attempts to get pooled connections to block indefinitely. The connectionRequestTimeout should be configurable and/or have a reasonable default. If a default is chosen, we can consider appliying something similar to connectTimeout ( establishConnectionTimeout).

### Value Proposition

To prevent the keycloak server from becoming unresponsive when HttpClient requests are not completing in a timely manner.

### Goals

To fail faster.

### Non-Goals

N/A.

### Discussion

_No response_

### Notes

connectionRequestTimeout was a top level field to set on the old SimpleHttp client.

shawkins:
This timeout applies to obtaining the connection - whether pooled or not - so it mostly obviates the need to set the connection timeout.

Setting the default to 5000 seems to be reasonable - it's inline with the default ldap pooled connection timeout, and matches the default socket timeout. Or this could be seen as breaking, and we can instead default to unlimited. I'll add documentation based upon what seems best.

closes: #44500

### DIFF
--- a/docs/guides/server/outgoinghttp.adoc
+++ b/docs/guides/server/outgoinghttp.adoc
@@ -24,6 +24,9 @@ The following are the command options:
 *establish-connection-timeout-millis*::
 Maximum time in milliseconds until establishing a connection times out. Default: Not set.
 
+*connection-request-timeout-millis*::
+Maximum time in milliseconds to obtain a connection for a request. Default: 5000ms
+
 *socket-timeout-millis*::
 Maximum time of inactivity between two data packets until a socket connection times out, in milliseconds. Default: 5000ms
 
--- a/services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
+++ b/services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
@@ -77,10 +77,12 @@ public class DefaultHttpClientFactory implements HttpClientFactory {
 
     private static class InputStreamResponseHandler extends AbstractResponseHandler<InputStream> {
 
+        @Override
         public InputStream handleEntity(HttpEntity entity) throws IOException {
             return entity.getContent();
         }
 
+        @Override
         public InputStream handleResponse(HttpResponse response) throws IOException {
             return super.handleResponse(response);
         }
@@ -173,6 +175,7 @@ public class DefaultHttpClientFactory implements HttpClientFactory {
                 if (httpClient == null) {
                     long socketTimeout = config.getLong("socket-timeout-millis", 5000L);
                     long establishConnectionTimeout = config.getLong("establish-connection-timeout-millis", -1L);
+                    long connectionRequestTimeout = config.getLong("connection-request-timeout-millis", HttpClientBuilder.DEFAULT_CONNECTION_REQUEST_TIMEOUT_MILLIS);
                     int maxPooledPerRoute = config.getInt("max-pooled-per-route", 64);
                     int connectionPoolSize = config.getInt("connection-pool-size", 128);
                     long connectionTTL = config.getLong("connection-ttl-millis", -1L);
@@ -206,6 +209,7 @@ public class DefaultHttpClientFactory implements HttpClientFactory {
 
                     builder.socketTimeout(socketTimeout, TimeUnit.MILLISECONDS)
                             .establishConnectionTimeout(establishConnectionTimeout, TimeUnit.MILLISECONDS)
+                            .connectionRequestTimeout(connectionRequestTimeout, TimeUnit.MILLISECONDS)
                             .maxPooledPerRoute(maxPooledPerRoute)
                             .connectionPoolSize(connectionPoolSize)
                             .connectionTTL(connectionTTL, TimeUnit.MILLISECONDS)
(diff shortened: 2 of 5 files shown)


shawkins:
@vmuzikar added an note in the changes.

docs/guides/server/outgoinghttp.adoc
```diff
@@ -24,6 +24,9 @@ The following are the command options:
 *establish-connection-timeout-millis*::
 Maximum time in milliseconds until establishing a connection times out. Default: Not set.
 
+*connection-request-timeout-millis*::
+Maximum time in milliseconds to obtain a connection for a request. Default: 5000ms
```
mabartos, line 28:
```suggestion
Maximum time in milliseconds to obtain an existing connection from the connection pool for a request. Default: 5000ms
```

@shawkins It'd be nice to be more explicit. WDYT?

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -324,6 +328,12 @@ public List<ProviderConfigProperty> getConfigMetadata() {
                 .defaultValue(-1L)
                 .add()
                 .property()
+                .name("connection-request-timeout-millis")
+                .type("long")
+                .helpText("When trying to obtain a connection, what is the timeout?")
```
mabartos:
```suggestion
                .helpText("When trying to obtain an existing connection, what is the timeout?")
```

@shawkins It'd be nice to be more explicit. WDYT?

services/src/main/java/org/keycloak/connections/httpclient/HttpClientBuilder.java
```diff
@@ -123,6 +127,20 @@ public HttpClientBuilder establishConnectionTimeout(long timeout, TimeUnit unit)
         return this;
     }
 
+    /**
+     * When trying to obtain a connection, what is the timeout?
```
mabartos, line 131:
```suggestion
     * When trying to obtain an existing connection, what is the timeout?
```

mabartos (commented):
@shawkins Looks good to me. It'd be nice to be more explicit and improve the docs. 

Moreover, the release note might be nice as well. 

@shawkins WDYT?

docs/guides/server/outgoinghttp.adoc
shawkins: It's applicable to obtaining any connection. It doesn't have to be an existing connection.

mabartos: @shawkins So, in that case, would it be possible to share the difference between obtaining a connection and establishing a connection?

shawkins: See the original pr comment, this timeout mostly overshadows the establish connection timeout. You can have an establish connection timeout that is shorter than the connection request timeout, however longer is not effective.

mabartos: Ahh, I see. 

So it means that:
- connection does NOT exist -> connection time(out) = establish connection time + get connection from the pool time
- connection exists -> connection time(out) =  get connection from the pool time

Is that correct?

I still see the need to be more explicit in this and explain things more to our community (I don't believe I'm not the only one who finds this confusing :stuck_out_tongue_closed_eyes:).

@shawkins Or do you think it does not require any more relation explanation?

cc: @vmuzikar

shawkins: > Is that correct?

No, it is that the establish connection timeout applies concurrently or logically within the connection request timeout.

Their code is basically:

connectionManager.get().wait(connection request timeout) - regardless of what happens in the get method, the connection request timeout is honored.

> @shawkins Or do you think it does not require any more relation explanation?

Honestly I was hoping this would be such a rare consideration that it didn't warrent much additional explanation - for fear of making it even more confusing.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
vmuzikar: Just a nitpick to make the help msg more consistent with others.
```suggestion
                .helpText("Timeout when trying to obtain a connection.")
```

vmuzikar (commented):

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
shawkins: That text came directly from the establish connection timeout. Do you want to change that one as well?

shawkins: @mabartos @vmuzikar updated both of the descriptions.

shawkins:
Not sure what the javascript test is failing - seems to happen with our without the change to the default setting.

vmuzikar (approved):
LGTM, thanks. @mabartos Do you want to give it another look too?

mabartos (approved):
LGTM, thanks

PR #44594, opened 2025-12-01
Session limit error code

### LINKED ISSUE #44474: Keycloak Maximum Concurrent Session Limit is incorrectly registered as a login failure, triggering Keycloak's Brute Force protection
### Before reporting an issue

### Area

authentication

### Describe the bug

The custom authentication flow for limiting concurrent sessions is incorrectly classifying “Access Denied due to Session Limit” events as Login Failures. Because of this misclassification, Keycloak’s Brute Force Protection is triggered, causing legitimate users to be temporarily locked out when they attempt to log in after reaching their maximum allowed sessions.

Image

### Version

keycloak 26.4.4, keycloak 26.3.1

### Regression

### Expected behavior

The user should get denied only, but not get temporarily blocked due to brute force

### Actual behavior

The user gets temporarily blocked due to brute force

### How to Reproduce?

Steps to reproduce the issue:

1. Create a realm named Demo.
2. Duplicate the browser flow and name it User Session Limiter.
3. Set this new flow as the bound Browser Flow.
4. In the User Session Limiter flow, add the User session count limiter execution and mark it as Required.
- Set Maximum concurrent sessions per user = 2
- Set Behavior when the limit is exceeded = Deny new session
5. Go to Realm Settings → Security Defenses → Brute Force Detection and set Max Login Failures = 2.

After this configuration, create two active sessions for a user in the Demo realm. Any additional login attempt should simply be denied because the session limit has been reached. Instead, Keycloak registers this denial as a Login Failure, which increments the brute-force counter and results in the user being temporarily locked out even though they used the correct credentials.

### Anything else?

_No response_

keshavprashantdeshpande:
Closes #44474

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
@@ -173,8 +173,10 @@ public class UserSessionLimitsAuthenticator implements Authenticator {
                         .orElse(SESSION_LIMIT_EXCEEDED);
 
                 context.getEvent().error(Errors.ACCESS_DENIED);
+                context.getEvent().detail("detail", eventDetails);
                 Response challenge = context.form().setError(errorMessage).createErrorPage(Response.Status.FORBIDDEN);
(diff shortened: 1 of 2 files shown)


ahus1:
Closing for inactivity.

PR #44600, opened 2025-12-02
Improve Public Key Management for JWTAuthorizationGrant identity provider

### LINKED ISSUE #44243: Improve Public Key Management for JWTAuthorizationGrant identity provider
### Description

This task is to improve the management of public keys for the `OIDCIdentityProvider` and `JWTAuthorizationGrantIdentityProvider`.

Currently  in `JWTAuthorizationGrantIdentityProvider` it is only possible to set the `jwksUrl` and the keys are loaded using the `JWTAuthorizationGrantJWKSEndpointLoader`.

For `OIDCIdentityProvider`, it is possible to use either the `jwksUrl` or import keys manually, but the manual import is limited to RSA and RS256.

An example to consider is the keys section in the client setting, you can upload keys in different formats.

### LINKED ISSUE #44523: Simple UI test for jwt-authorization-grant identity provider
### Description

Can we have some basic UI test for the jwt-authorization-grant identity provider? Maybe something similar to `spiffe.spec.ts` is sufficient. We can maybe also test that creation/update of the identity provider fails if some needed fields are missing etc.

rmartinc:
Closes #44243
closes #44523 

Adding the option of using a public key for JWT Authorization Grant instead of the JWKS url. The PR shares the same components with the OIDC identity provider. It also adds an import button in the admin console to allow importing the same formats we are using in other places for the key. Added tests for OIDC and JWT Authorization Grant. Two little tests are added to the admin UI. They test the JWT provider creation and imports two files. The PR ended up being quite long. Sorry.


mposolda (approved):
@rmartinc Nice, Thanks!

graziang (approved):
@rmartinc thanks!

PR #44601, opened 2025-12-02
Added validation for client session timeout post comparing the realm session timeouts

### LINKED ISSUE #41019: Validate client session session timeout and lifetime settings on edit
### Description

A client session lifetime and timeout settings can never be longer than the realm settings for the user session aka SSO settings and if enabled the remember me timeout, as the user session would expire first, and then the client session will expire as well. 

We already have a test for this in RefreshTokenTest#refreshTokenUserClientMaxLifespanGreaterThanSession where we return the correct refresh token duration duration calculated taking the shorter realm settings into account.

### Motivation

Let users know that they will need to change the realm session settings to make this work.

### Details

For now, the validation would be done when a client is edited. No validation would be done when the realm settings are changed to validate all client configurations.

ruchikajha95:
session timeouts

Closes #41019

This PR adds validation for client-level  session timeout to ensure they do not exceed the realm-level SSOconfiguration.

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -1,11 +1,25 @@
 // ------------------------ Breaking changes ------------------------  //
 == Breaking changes
 
-= Loopback Hostname Verification on Windows
+Breaking changes are identified as those that might require changes for existing users to their configurations or applications.
+In minor or patch releases, {project_name} will only introduce breaking changes to fix bugs.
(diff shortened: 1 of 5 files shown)


ahus1 (approved):
Thank you for contributing this change to allow admins to safely change client settings!

ahus1:
@keycloak/core-clients - FYI, this PR is adding a validation of the client settings for timeouts and max lifetime based on previous user complaints and mishaps. @ruchikajha95 is looking into additional validations in follow-up pull requests.

PR #44612, opened 2025-12-02
Added section on recommended isolation level to db guides

### LINKED ISSUE #44611: Document recommended transaction isolation level for MS SQL in guides and warn if wrong level is detected on startup
### Description

The default isolation level for MS SQL is `READ_COMMITTED` which can lead to deadlocks under high load. On MS SQL, the recommended isolation level is `READ_COMMITTED_SNAPSHOT`. This should be documented accordingly and also a warning should be issued on Keycloak boot  when the wrong level is detected.

### Value Proposition

Customers use the correct isolation level on MS SQL.

### Goals

- document recommended isolation level
- warn if wrong level is detected on startup

### Non-Goals

- setting the correct isolation level on startup

### Discussion

_No response_

### Notes

_No response_

sschu:
Also added warning if wrong level is detected on startup

Closed #44611

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -54,6 +54,18 @@ The SPIFFE Identity Provider preview feature now uses the `trustDomain` configur
 was done to make it more explicit that a `trustDomain` is configured and not the `iss` of the token, which is
 usually not set in SPIFFE JWT SVIDs.
 
+=== Running on Microsoft SQL Server now recommends READ_COMMITTED_SNAPSHOT
+
+On MS SQL Server, the default transaction isolation level is `READ_COMMITTED`, which can lead to deadlocks during high load. Therefore, the recommended isolation level for {project_name} is `READ_COMMITTED_SNAPSHOT`.
+
+Use the following statement to changes this for your database:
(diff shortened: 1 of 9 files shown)


ahus1:
Thank you for the PR. I expanded it a little so that the new testsuite runs with the newly configured database. 

Unfortunatly I still get the error message that this is not correctly configured. 

https://github.com/boschglobal/keycloak/pull/1

I found some references to an isolation level of 4096: https://stackoverflow.com/questions/13929532/ms-sql-snapshot-isolation-and-hibernate-behaviour

But I was unable to get that to work with Quarkus. While using for example the agroal JDBC connection pool properties, I saw some exceptions that the level was not a legal level. 

Any ideas? Is it working for you?

sschu:
@ahus1 Tests are failing because of problems with repo.maven.org. We also saw them internally but they seem to be gone now so a rerun should help.

ahus1:
@sschu - we'll need for https://github.com/keycloak/keycloak/issues/44700 as the model tests are broken in main.

ahus1 (approved):

ahus1:
FYI @keycloak/cloud-native - this now adds a recommendation for MSSQL isolation level and updates the docs and the tests. Thank you for @sschu for the input on the topic and the PR!

PR #44615, opened 2025-12-02
[OID4VCI] Expose advanced realm-level OID4VCI settings in the Admin UI

### LINKED ISSUE #43900: Expose advanced realm-level OID4VCI settings in the Admin UI
### Description

Several realm-level OID4VCI settings are not exposed in the UI. Extend Realm Settings → Tokens → OID4VCI attributes to include:

- Signed Issuer Metadata (enable, lifespan, signing algorithm)
- Encryption Policy (require encryption, supported compression algorithms)
- Batch Issuance Size
- Time-based Correlation Mitigation (strategy, randomization window, rounding unit) etc...

Ogenbertrand:
Extend Realm Settings → Tokens → OID4VCI attributes to include:

- Signed Issuer Metadata (enable, lifespan, signing algorithm)
- Encryption Policy (require encryption, supported compression algorithms)
Batch Issuance Size
- Time-based Correlation Mitigation (strategy, randomization window, rounding unit) etc...

- Screenshoots:

image

image

closes: https://github.com/keycloak/keycloak/issues/43900


Captain-P-Goldfish:
Hi Bertrand,
could you maybe take this occasion and add the new attributes from here (https://github.com/keycloak/keycloak/pull/44439) too?

```java
    public static final String TRUSTED_KEYS_REALM_ATTR = "oid4vc.attestation.trusted_keys";
    public static final String TRUSTED_KEY_IDS_REALM_ATTR = "oid4vc.attestation.trusted_key_ids";
```

Ogenbertrand:
> Hi Bertrand, could you maybe take this occasion and add the new attributes from here (#44439) too?
> 
> ```java
>     public static final String TRUSTED_KEYS_REALM_ATTR = "oid4vc.attestation.trusted_keys";
>     public static final String TRUSTED_KEY_IDS_REALM_ATTR = "oid4vc.attestation.trusted_key_ids";
> ```

I have to wait for that PR to be merge, so this is a blocker: https://github.com/keycloak/keycloak/pull/44439.

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -3604,6 +3604,29 @@ oid4vciNonceLifetimeHelp=The lifetime of the OID4VCI nonce in seconds.
 preAuthorizedCodeLifespan=Pre-Authorized Code Lifespan
 preAuthorizedCodeLifespanHelp=The lifespan of the pre-authorized code in seconds.
 oid4vciFormValidationError=Please ensure the OID4VCI attribute fields are filled with values 30 seconds or greater.
+signedIssuerMetadata=Signed Issuer Metadata
+signedIssuerMetadataHelp=Enable signing of the issuer metadata. When enabled, the issuer metadata will be signed using the configured signing algorithm.
+signedMetadataLifespan=Signed Metadata Lifespan
+signedMetadataLifespanHelp=The lifetime of the signed metadata in seconds. After this time, the signed metadata will expire.
```
mposolda:
```suggestion
signedMetadataLifespanHelp=The lifetime of the signed metadata. After this time, the signed metadata will expire.
```

Is it possible to omit `in seconds` from this help text? The reason is that admin console displays also time units, so it is clear to the administrator what time unit is used based on that and using the term `in seconds` in the tooltip makes it rather more confusing though... Besides that, other tooltips in the "Tokens" tab are not using `in seconds` as well.


Same change can be good to do for tooltips of already existing attributes "OID4VCI Nonce Lifetime" and "Pre-Authorized Code Lifespan" (Just noticed it during the review of this PR. Feel free to do in this PR as well).

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
```diff
@@ -674,6 +697,136 @@ export const RealmSettingsTokensTab = ({
...
+              <SelectControl
+                name={convertAttributeNameToForm(
+                  "attributes.oid4vci.signed_metadata.alg",
+                )}
+                label={t("signedMetadataSigningAlgorithm")}
+                labelIcon={t("signedMetadataSigningAlgorithmHelp")}
+                controller={{
+                  defaultValue: "RS256",
+                }}
+                options={defaultSigAlgOptions.map((p) => ({
```
mposolda: I can see that this contains also symmetric algorithms like `HS256` . It seems that symmetric algorithms should be omitted though per the specification https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-12.2.3 and also it does not makes sense to use them as their keys cannot be retrieved from the realm JWKS endpoint and does not provide a way for consumer to verify the signature (They are used mostly for the use-cases when Keycloak is signing something, which is supposed to be later verified by Keycloak itself). Is it possible to omit symmetric algorithms? There are some existing options in the admin console, which use only asymetric algorithms (For example in the `keys` tab of the OIDC client where you can select the algorithm for private_key_jwt client authentication).

Also we may need to check this on server-side as well and make sure that symmetric algorithms are omitted (but probably not in the scope of this PR).

mposolda (commented):
@Ogenbertrand Thanks for the PR!

Captain-P-Goldfish (approved):
LGTM

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
```diff
@@ -59,6 +63,10 @@ export const RealmSettingsTokensTab = ({
     serverInfo.providers!["signature"].providers,
   );
 
+  const asymmetricSigAlgOptions = defaultSigAlgOptions.filter(
+    (p) => !p.startsWith("HS"),
```
mposolda: Nitpick: This is bit error-prone and will break if we introduce any symetric algorithm not starting with `HS` into Keycloak.

There are already some other UI options in Keycloak using asymmetric algorithms and those can be directly obtained from server-info. For example see https://github.com/keycloak/keycloak/blob/main/js/apps/admin-ui/src/clients/credentials/SignedJWT.tsx#L19 .

mposolda (commented):
@Ogenbertrand @Captain-P-Goldfish Thanks for the PR and review!

I am not yet confident with all the options introduced into UI, but this PR is great addition and vote for merging it.

Added just one minor nitpick inline. Can you please check?

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
Ogenbertrand: Dear @mposolda !
Thanks for pointing that out and for the existing solution. your comment was indeed considered and addressed.
Do you mind checking again !!!

mposolda (approved):

js/apps/admin-ui/src/realm-settings/TokensTab.tsx
mposolda: @Ogenbertrand Thanks!

mposolda:
@Ogenbertrand Can you please rebase with latest Keycloak main? There was a fix for the admin UI workflows tests and for Account UI tests, which were recently pushed to Keycloak main, which should fix UI tests in this PR as well.

mposolda:
@Ogenbertrand Thanks!

PR #44628, opened 2025-12-03
[admin-api-v2] Remove GlassFish Expressly dependency for Hibernate Validator

### LINKED ISSUE #43569: Handle GlassFish Expressly dependency for Hibernate Validator
### Description

As mentioned in the [comment](https://github.com/keycloak/keycloak/pull/43395#discussion_r2435126289), the dependency uses EPL 2.0 license. It'd be good to revisit on the Hibernate Validator site, if there's an alternative way on how to proceed without the need to include the dependency. Otherwise, we'd need an exception from the CNCF.

mabartos:
- Closes #43569

In the end, we can exclude the `expressly` dependency as we do not use the Expression Language (EL) for the Jakarta validation :tada: 

The default interpolator in Hibernate Validator uses EL to resolve things like `${validatedValue}`, `#{someBean.property}`, `${min}`, `${max}` in messages. `ParameterMessageInterpolator` ignores EL entirely - only simple parameter placeholders like ${min}, ${max}, are replaced. No EL dependency is required on the classpath. 

It is also recommended to use this interpolator in the Quarkus error message.

I think it's a very suitable approach for us, as the license exception is not very clear https://github.com/cncf/foundation/issues/1177

@keycloak/cloud-native Could you please check it? 

The dependency is not part of the distribution JARs:
image

### DIFF
--- a/quarkus/runtime/pom.xml
+++ b/quarkus/runtime/pom.xml
@@ -145,6 +145,12 @@
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-hibernate-validator</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.glassfish.expressly</groupId>
+                    <artifactId>expressly</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <!-- SmallRye -->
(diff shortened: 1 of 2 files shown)


mabartos:
Compilation error in CI is fixed by https://github.com/keycloak/keycloak/pull/44627

vmuzikar (dismissed):
LGTM, thanks.

ahus1:
@mabartos - spotless is failing

mabartos:
@ahus1 @vmuzikar Fixed, thanks.

ahus1 (approved):

PR #44655, opened 2025-12-04
Remove log4j 1.x from testsuite/model and drop log4j.properties

mohta1:
- Deleted log4j dependency and log4j.properties file from the module.
- Logging remains via org.jboss.logging.Logger, so no change to logger calls.

Closes #44554


stianst (changes requested):
We can't just get rid of logging configuration; log4j.properties needs to be converted to logging.properties

stianst:
PR including logging.properties here https://github.com/keycloak/keycloak/pull/44809

Issue #44492, opened 2025-11-26
Promote jwt-authorization-grant to preview feature


mposolda:
### Description

This task is about promote existing jwt-authorization-grant to preview feature in the codebase and release notes and documentation. Might be done after other tasks from #43573 and #43575 are done (especially documentation).

**WARNING:** It is needed to uncomment a note about "preview" feature for JWT Authorization grant docs introduced in the docs PR https://github.com/keycloak/keycloak/pull/44666

ahus1:
@mposolda - without this change, the feature will not have any marker in the docs, therefore people might consider it supported instead when reading the docs. They will only find out that it is experimental once they try using it and find out the options are not available until they enable them...

PR #44666, opened 2025-12-04
Documentation for JWT Authorization Grant

### LINKED ISSUE #44136: Documentation for rfc-7523 grant
### Description

Proper documentation for rfc-7523 grant (including release notes, proper description of the grant, related client policies etc).

Documentation might be added as a dedicated guide in the "Securing apps" section: https://www.keycloak.org/guides#securing-apps . Probably can be added after the "Token exchange" guide.

Token-exchange guide may be also slightly updated - the existing section of legacy token exchange external-internal can have some note/warning at the beginning that preferred way of doing this is to use that new jwt-bearer assertion grant.

This might be possibly divided into subtasks if it is easier as docs might be relatively bigger task.

rmartinc:
Closes #44136

Documentation for JWT Authorization Grant. It's placed after Token Exchanged and before the specification guide. We have to remember to add the tech-preview note when moving the feature out of experimental. The RFC 7521/7523 are already in the list because of the client authentication. It was not so easy explaining this new feature. :smile:


docs/guides/securing-apps/jwt-authorization-grant.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+<@tmpl.guide
+title="JWT Authorization Grant"
+priority=130
+summary="Guide for the JWT Authorization Grant specification RFC 7521 / 7523.">
+
+// TODO: add tech preview when moved out experimental
+//< @ features.techpreview feature="jwt-authorization-grant"/>
+
+This guide defines how a JWT Bearer Token can be used in {project_name} as an authorization grant. This feature allows clients to send a JWT assertion to request an access token when the client wants to use an existing trust relationship, expressed through the semantics of the JWT, without a direct user-approval step at the authorization server. The trust relationship usually refers to another Identity Provider server (another OIDC server), and allows to obtain a cross-domain or cross-realm access token. In this sense, it is similar to the external to internal request in token exchange V1 (see <@links.securingapps id="token-exchange" anchor="_external-token-to-internal-token-exchange" /> for more information).
```
graziang: I would make this sentence a bit simpler as a presentation sentence, but I'm not sure how to do it, maybe by simplifying the trust relationship part.

graziang (approved):
@rmartinc thanks! Looks great, just a small minor comment.

docs/guides/securing-apps/jwt-authorization-grant.adoc
rmartinc: That phrase is from the spec, it was not mine for sure :smile:. I'll try to remove the part between the commas and add that part as a second sentence saying that the JWT auth grant just depends on the JWT claims.

rmartinc: Check if you think it's better now. :smile:

docs/guides/securing-apps/jwt-authorization-grant.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Besides the previous specific options, both identity provider types need some basic configuration related to assertion and signature validation.
+
+* **Issuer**: Issuer for the assertion. Required.
+* **Use JWKS URL**: Whether a JWKS endpoint URL is used to obtain the keys that will validate the assertion signature. If disabled, the keys should be provider manually by the administrator. The recommended value is On.
+* **JWKS URL**: The URL for downloading the signing keys. Required if **Use JWKS URL** is enabled.
+* **Validating public key id**: Fixed `kid` for validating assertion signatures. This option can be left empty to just validate the signature with the configured public key. This option can only be specified if **Use JWKS URL** is disabled and the validating key is defined as a fixed key in PEM format.
+* **Validating public key**: The public key in PEM or JWKS format that must be used to verify external IdP signatures. Required if **Use JWKS URL** is disabled.
+
+WARNING: The JWT Authorization Grant feature needs the OIDC provider to validate signatures (the option **Validate Signatures** should be enabled).
```
mposolda:
```suggestion
WARNING: When JWT Authorization Grant is configured with the OIDC Identity provider, the signatures on the JWT tokens sent to the token endpoint are always validated. The OIDC identity provider option **Validate Signatures** is ignored for the JWT Authorization Grant as it is used just for validation of the signatures on the tokens retrieved from the OIDC identity provider during authentication of the users with this OIDC identity provider.
```
I think this WARNING is not accurate with the latest changes done in https://github.com/keycloak/keycloak/issues/44473 . Adding small suggestions to clarify it.

docs/guides/securing-apps/jwt-authorization-grant.adoc
```diff
@@ -0,0 +1,213 @@
+<#import "/templates/guide.adoc" as tmpl>
+<#import "/templates/links.adoc" as links>
+<#import "/templates/features.adoc" as features>
+
+<@tmpl.guide
+title="JWT Authorization Grant"
+priority=130
+summary="Guide for the JWT Authorization Grant specification RFC 7521 / 7523.">
+
+// TODO: add tech preview when moved out experimental
```
mposolda, line 10: FYI. Added warning about this to https://github.com/keycloak/keycloak/issues/44492

docs/guides/securing-apps/jwt-authorization-grant.adoc
```diff
@@ -0,0 +1,213 @@
...
+{
+  "access_token":"eyJhbG[...redacted...].eyJleH[...redacted...].RFnNEv[...redacted...]",
+  "expires_in":300,
+  "refresh_expires_in":0,
+  "token_type":"Bearer",
+  "not-before-policy":0,
+  "scope":"email profile"
+}
+----
+
```
mposolda, line 129: Maybe it worth to add some additional warning/note/paragraph that this grant does not support issuing refresh tokens or ID tokens, but only the access token as shown in the response above? Also that user session is not created. And that the newly issued access token can be used in Keycloak introspection endpoint, user-info endpoint or any other endpoints until it is not expired or explicitly revoked by the revocation endpoint.

WDYT?

docs/guides/securing-apps/token-exchange.adoc
```diff
@@ -675,6 +675,8 @@ These types of changes required a configured identity provider in the Admin Cons
 
 NOTE:  SAML identity providers are not supported at this time.  Twitter tokens cannot be exchanged either.
 
+WARNING: When two {project_name} servers or realms are involved, external to internal Token Exchange will be replaced by <<_standard-token-exchange,Standard Token Exchange V2>> and <@links.securingapps id="jwt-authorization-grant" />. Following the idea presented in the current draft specification link:https://datatracker.ietf.org/doc/html/draft-ietf-oauth-identity-chaining-06[OAuth Identity and Authorization Chaining Across Domains], the final access token will be obtained performing two requests: a standard token exchange in the external realm (internal to internal) to get a valid JWT assertion; a JWT Authorization Grant to finally request an access token in the internal realm.
```
mposolda:
```suggestion
WARNING: External to internal Token Exchange will be replaced by <<_standard-token-exchange,Standard Token Exchange V2>> and <@links.securingapps id="jwt-authorization-grant" />. Following the idea presented in the current draft specification link:https://datatracker.ietf.org/doc/html/draft-ietf-oauth-identity-chaining-06[OAuth Identity and Authorization Chaining Across Domains], the final access token will be obtained performing two requests: a standard token exchange in the external identity server to get a valid JWT assertion; a JWT Authorization Grant to finally request an access token in the internal {project_name} realm. The JWT Authorization grant can be eventually used directly to exchange your external tokens as long as they are JWT tokens, which conforms to all the requirements specified for the JWT Authorization Grant.
```

AFAIK The external-internal token exchange would be removed entirely at some point and is not limited to replace only the Keycloak-Keycloak deployment, but any other 3rd party identity provider as well. Maybe something like this can be used?

mposolda (commented):
@rmartinc Nice docs! Thanks!

Added few comments inline.

@andymunro Can you please review whenever you have a chance?

docs/guides/securing-apps/jwt-authorization-grant.adoc
rmartinc: Yep, thanks!

docs/guides/securing-apps/token-exchange.adoc
rmartinc: Okis

docs/guides/securing-apps/jwt-authorization-grant.adoc
rmartinc: But the ID token is generated if `oidc` scope is present, right? I will just comment about the rest.

rmartinc: Done!

rmartinc:
I added your suggestions and another infor note talking about refresh tokend and transient sessions. So it's done until @andymunro review.

mposolda (approved):
@rmartinc Thanks for the updates!

andymunro (approved):
Looks good, @mposolda

mposolda:
@andymunro Thanks for the additional review!

PR #44680, opened 2025-12-04
Improve enumeration of features in Admin UI

mabartos:
- Closes #44677
- When you hover over them, you see the description
- Suggestion on how we could improve the UX around features

## OLD
Screenshot From 2025-12-04 14-56-24

## NEW

https://github.com/user-attachments/assets/9435f42c-5658-4f73-93ad-952ac11e82bb

#### Smaller boxes for enabled features 
Usually, there's not so many enabled features from categories Preview, Deprecated, Experimental. So, they're wrapped together in small categories aside the supported features

Screenshot From 2025-11-21 14-07-35

#### Multiple columns for enabled features
When the other categories (preview, deprecated, experimental) contains >3 features, there are more columns for enabled features to have better layout. 

Screenshot From 2025-11-21 14-13-08


stianst (changes requested):
I disagree with this change. It looks worse, and I also think it's better to keep the features listed alphabetically rather than grouping by level.

Additionally I think we should consider guides, CLI output, and admin console collectively and have a consistent view across these.

PR #44705, opened 2025-12-05
Move AdminClientTest to the new testsuite

### LINKED ISSUE #35040: Migrate admin package to new test framework

lhanusov:
Closes #35040


test-framework/core/src/main/java/org/keycloak/testframework/annotations/InjectAdminClientFactory.java
```diff
@@ -15,4 +15,6 @@
     String ref() default "";
 
     LifeCycle lifecycle() default LifeCycle.CLASS;
+
+    boolean mTls() default false;
```
stianst: I'd remove this; should just be automatic when the certificates are configured with mTLs for clients

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactorySupplier.java
```diff
@@ -15,7 +15,11 @@ public class AdminClientFactorySupplier implements Supplier<AdminClientFactory,
     public AdminClientFactory getValue(InstanceContext<AdminClientFactory, InjectAdminClientFactory> instanceContext) {
         KeycloakServer server = instanceContext.getDependency(KeycloakServer.class);
 
-        if (!server.isTlsEnabled()) {
+        if (instanceContext.getAnnotation().mTls()) {
```
stianst: No need to handle if mTLS is enabled or not here; just handle it within ManagedCertificates and have it return an SSLContext with or without key material.

test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientFactorySupplier.java
```diff
@@ -15,7 +15,11 @@ public class AdminClientFactorySupplier implements Supplier<AdminClientFactory,
...
+            ManagedCertificates managedCert = instanceContext.getDependency(ManagedCertificates.class);
+            SSLContext sslContext = managedCert.getMTlsClientSSLContext();
+            return new AdminClientFactory(server.getBaseUrl(), sslContext);
+        } else if (!server.isTlsEnabled()) {
```
stianst: We may want to also get rid of server.isTlsEnabled, and just handle that within the managed certificates.

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -54,11 +56,38 @@ public ManagedCertificates(CertificatesConfigBuilder configBuilder) throws Manag
...
+        try {
+            return SSLContextBuilder.create()
+                    .loadTrustMaterial(clientsTrustStore, null)
+                    .build();
+        } catch (KeyManagementException | NoSuchAlgorithmException | KeyStoreException e) {
+            throw new ManagedCertificatesException(e);
+        }
+    }
+
+    public SSLContext getMTlsClientSSLContext() {
```
stianst: I'd combine these into a single method; and have it return the SSLContext as needed. We could also move the TLS enabled to within managed certificates; and it it's not enabled then we just return null here; then the admin client can just getSSLContext from ManagedCertificates and if not null then use it.

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
```diff
@@ -116,6 +117,15 @@ public KeycloakServerConfigBuilder tlsEnabled(boolean enabled) {
     public boolean tlsEnabled() {
         return tlsEnabled ;
     }
+
+    public KeycloakServerConfigBuilder mTlsEnabled(boolean enabled) {
```
stianst: We shouldn't need this; just make it configurable through ManagedCertificates

stianst (commented):
I think we can simplify this stuff a lot. Basically what I'm thinking is that we always have a ManagedCertificates dependency in server and clients; then it will just return null if nothing is available.

That means you have a single thing to configure to handle TLS and mTLS, which is just the ManagedCertificates. The rest is just wired up by getting a dependency on ManagedCertificates, and the applicable annotation for configure that. 

I think this would make the implementation simpler, configuration easier, and the dependency lifecycles should work more smoothly this way.

test-framework/core/src/main/java/org/keycloak/testframework/https/Certificate.java
```diff
@@ -0,0 +1,60 @@
+package org.keycloak.testframework.https;
+
+import java.nio.file.Path;
+import java.security.KeyStore;
+
+public class Certificate {
```
stianst: Should be package protected; could also be a record class

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -29,104 +30,167 @@ public class ManagedCertificates {
...
+            client.setCertificateEntry("client-certificate");
+        }
+    }
+
+    private Path resolvePath(String fileName) {
+        return KEYSTORES_DIR.resolve(fileName + "." + keystoreFormat.getPrimaryExtension());
+    }
 
-        password = configBuilder.getKeystoreFormat() == KeystoreUtil.KeystoreFormat.JKS ? "password".toCharArray() : "passwordpassword".toCharArray();
+    public SSLContext getSSLContext() {
```
stianst:
```suggestion
    public SSLContext getClientSSLContext() {
```

test-framework/core/src/main/java/org/keycloak/testframework/https/ManagedCertificates.java
```diff
@@ -29,104 +30,167 @@ public class ManagedCertificates {
...
     }
 
-    public String getKeycloakServerKeyStorePath() {
-        return serverKeystorePath.toString();
+    public String getServerTrustStorePath() {
+        return server.getTruststorePath().toString();
     }
 
-    public String getKeycloakServerKeyStorePassword() {
+    public String getKeyAndTrustStorePassword() {
```
stianst: I'd split this into 2; `getServerKeyStorePassword` and `getServerTrustStorePassword`. It's fine they are using the same password now, but could be different in the future, or for a different supplier.

stianst (dismissed):

stianst (dismissed):

lhanusov (commented):
LGTM

stianst (approved):

Issue #42694, opened 2025-09-17
Workflows: review test coverage


sguilhen:
### Description

Current test coverage needs reviewing and a better organization of the tests based on the feature being tested. Currently many unrelated tests are grouped together in the `WorkflowManagementTest` , so we need to review and possibly extract some tests into their own test classes.

Lynch65:
Hey, 

@sguilhen  may you assign me the issue , I'm interested .

sguilhen:
@Lynch65 Done!

This might turn to be a bit of a work as I think we should probably have specific tests for each condition/step implementation. Currently we're testing them as part of a workflow, but to me it makes sense to have  a unit test approach for them, and them separate tests that integrates conditions and steps in a workflow.

Also, the `WorkflowManagementTest` is bloated right now. For example, a bunch of email notification tests are there, along with some other tests for random capabilities. Probably this test should focus on the management of workflows (CRUD operations basically).

Feel free to chime in with ideas on how to best approach this and thanks for volunteering!

Lynch65:
Hey, I went through src/test/java/org/keycloak/tests/model/workflow/ and noticed that while many tests are modular (ExpressionConditionWorkflowTest.java, RoleWorkflowConditionTest.java, WorkflowStepManagementTest.java), WorkflowManagementTest.java is doing too much (CRUD, steps, conditions, emails…).

I suggest:
Split tests in WorkflowManagementTest.java by responsibility (CRUD, conditions, steps, emails).
Keep isolated tests as unit tests and full workflow flows as integration tests.
Proposed structure:

```
workflow/
 ├── unit/
 │    ├── WorkflowStepTest.java
 │    ├── WorkflowConditionTest.java
 │    └── EmailNotificationTest.java
 ├── integration/
 │    ├── AdhocWorkflowTest.java
 │    ├── UserSessionRefreshTimeWorkflowTest.java
 │    └── UserCreationTimeWorkflowTest.java
 └── management/
      └── WorkflowManagementTest.java
```

This would clarify responsibilities and reduce coupling.

PR #44710, opened 2025-12-05
Improvements to the notify step

pedroigor:
Closes #44708

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowExecutionContext.java
+++ b/model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowExecutionContext.java
@@ -77,6 +77,11 @@ final class DefaultWorkflowExecutionContext implements WorkflowExecutionContext 
         return event;
     }
 
+    @Override
+    public WorkflowStep getNextStep() {
+        return workflow.getSteps(currentStep.getId()).skip(1).findFirst().orElse(null);
+    }
+
     String getExecutionId() {
         return this.executionId;
     }
--- a/tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/NotificationStepTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/NotificationStepTest.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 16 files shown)


model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowExecutionContext.java
```diff
@@ -77,6 +78,22 @@ public WorkflowEvent getEvent() {
...
+        if (nextStepIndex == -1) {
+            return null;
+        }
+
+        if (++nextStepIndex < steps.size()) {
+            return steps.get(nextStepIndex);
+        }
+
+        return null;
+    }
```
sguilhen, line 83: There's a getSteps(String stepId) in Workflow.java that gives you a stream of steps starting at the specified step... perhaps you could do then

```
workflow.getSteps(currentStep.getId()).skip(1).findFirst();
```

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/NotificationStepTest.java
```diff
@@ -0,0 +1,105 @@
...
+            for (String content : List.of(body.getText(), body.getHtml())) {
+                assertTrue(content.contains("Dear Bob Doe,"));
+                assertTrue(content.contains("Welcome to " + managedRealm.getName() + "!"));
+                assertTrue(content.contains("The next step is scheduled to 7 days."));
+            }
+        } finally {
+            mailServer.runCleanup();
+        }
+    }
+}
```
sguilhen, line 105: Can you also move the notification-related tests from `WorkflowManagementTest` to this class? I think it makes sense to have all notification tests in the same place

themes/src/main/resources/theme/base/email/html/workflow-notification.ftl
```diff
@@ -2,22 +2,23 @@
...
+        <p><strong>Time remaining: ${daysRemaining} day<#if daysRemaining != 1>s</#if></strong></p>
+    </#if>
+
+    <p>If you have questions, please contact your ${realmName} administrator.</p>
+
+    <p>
+        Best regards,<br>
+        ${realmName} Administration
+    </p>
+</#if>
```
sguilhen, line 22: I recall Stian suggesting that the e-mail should be more direct and make it clear that it is coming from an automated workflow. So no "Dear subject, "  or "Best regards," in the message. I don't recall exactly where we discussed this though.

pedroigor: Not sure, this is the same text as before. If we want to change, we can have a follow-up.

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/NotificationStepTest.java
pedroigor: See https://github.com/keycloak/keycloak/issues/42694. I would rather do it separately as it will require moving not only methods but some methods to base classes, etc.

themes/src/main/resources/theme/base/email/html/workflow-notification.ftl
pedroigor: Also note that the content is something that the realm admin should decide how it should look. These changes are now allowing this without any need to customize templates.

tests/base/src/test/java/org/keycloak/tests/admin/model/workflow/NotificationStepTest.java
sguilhen: Good point. Let's keep this one simpler then

sguilhen (approved):

ahus1 (approved):

PR #44715, opened 2025-12-05
Credential offer endpoint has parameter user_id, but expects username

mposolda:
closes #44642

PR also checks if user is enabled (This was not done beforehand).

The user reference in CredentialOffer is still user_id as it is internal and lookup by ID is a bit more efficient than lookup by username.

PR #44716, opened 2025-12-05
Account console QR code generation for OID4VCI credentials is broken

### LINKED ISSUE #44623: Account console QR code generation for OID4VCI credentials is broken
### Before reporting an issue

### Area

oid4vc

### Describe the bug

It seems that current functionality of the account console for generation of QR codes is broken.

### Version

nightly (from 2025-12-03)

### Regression

### Expected behavior

QR code generation should work in the account console

### Actual behavior

When trying to generate QR code of OID4VCI verifiable credential, it fails currently.

### How to Reproduce?

- Start keycloak with `oid4vci` feature
- Create realm and enable switch `Verifiable credentials` in the realm settings
- Create some user and password for him and assign him role `credential-offer-create`
- Login to account console and go to `Verifiable credentials` tab
- Try to generate some QR code. It currently fails with 403 forbidden code

### Reason

- The `account-console` client does not have switch `oid4vci.enabled` enabled. This causes `OID4VCIssuerEndpoint.checkClientEnabled()` to fail. We should probably either make sure that the `account-console` have this enabled by default or eventually hide the `Verifiable credentials` tab for the account console if it does not have this switch enabled? Not yet sure what is better approach...
- The parameter `user_id` is not sent from the account console, which causes request to fail. This is probably a regression of the change in https://github.com/keycloak/keycloak/commit/54bf9206 .

### Anything else?

It can be great if account console has a bit better usability. For example:
- It can add some introduction text to the user that he can use the `Verifiable credentials` page with his wallet and copy the QR code to his wallet
- It can be great if it is possible to copy URL in addition to the QR code. See for example experience from the walt.id in the screenshot.

Image

mposolda:
closes #44623 

Marked as draft for now.

I've also removed warnings in the server.log (which were triggered by invoking deprecated OID4VCI well-known endpoint) and updated the size of QR code to 300x300, which is the size used also on setup OTP screen, so there is consistency in various Keycloak screens.

Screenshot for the reference:

Screenshot from 2025-12-05 19-37-49

### DIFF
--- a/js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
+++ b/js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
@@ -14,7 +14,7 @@ import {
 } from "@patternfly/react-core";
 import { useEffect, useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
-import { getIssuer, requestVCOffer } from "../api";
+import { getVCIssuer, requestVCOffer } from "../api";
 import { CredentialsIssuer } from "../api/representations";
 import { Page } from "../components/page/Page";
 import { usePromise } from "../utils/usePromise";
@@ -33,7 +33,7 @@ export const Oid4Vci = () => {
   const [credentialsIssuer, setCredentialsIssuer] =
     useState<CredentialsIssuer>();
 
-  usePromise(() => getIssuer(context), setCredentialsIssuer);
+  usePromise(() => getVCIssuer(context), setCredentialsIssuer);
 
   const selectOptions = useMemo(() => {
     if (typeof credentialsIssuer !== "undefined") {
@@ -119,8 +119,8 @@ export const Oid4Vci = () => {
               {offerQRVisible && (
                 <ActionListItem>
                   <img
-                    width="500"
-                    height="500"
+                    width="300"
+                    height="300"
                     src={qrCode}
                     data-testid="qr-code"
                   />
--- a/js/apps/account-ui/src/api.ts
+++ b/js/apps/account-ui/src/api.ts
@@ -73,21 +73,19 @@ function checkResponse<T>(response: T) {
   return response;
 }
 
-export async function getIssuer(context: KeycloakContext<BaseEnvironment>) {
+export async function getVCIssuer(context: KeycloakContext<BaseEnvironment>) {
   const response = await request(
     joinPath(
-      "/realms/",
+      "/.well-known/openid-credential-issuer/realms/",
       context.environment.realm,
-      "/.well-known/openid-credential-issuer",
     ),
     context,
     {},
     new URL(
       joinPath(
         context.environment.serverBaseUrl,
-        "/realms/",
+        "/.well-known/openid-credential-issuer/realms/",
         context.environment.realm,
-        "/.well-known/openid-credential-issuer",
       ),
     ),
   );
@@ -105,9 +103,11 @@ export async function requestVCOffer(
     {
       searchParams: {
         credential_configuration_id: supportedCredentialConfiguration.id,
+        client_id: context.environment.clientId,
+        username: context.keycloak.tokenParsed?.preferred_username,
         type: "qr-code",
-        width: "500",
-        height: "500",
+        width: "300",
+        height: "300",
       },
     },
     new URL(


mposolda:
@IngridPuppet @Ogenbertrand @francis-pouatcha @forkimenjeckayang @Awambeng @Captain-P-Goldfish @tdiesler @wistefan Hello, can I please ask for the review of the PR (There is other PR https://github.com/keycloak/keycloak/pull/44715 , which this one depends on, so please feel free to review only the last commit with the UI changes in the account console if they look OK to you).

js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
IngridPuppet, line 1: My suggestions:
1. Align elements on the UI vertically (so it doesn't look broken)
2. Rename "for import" to "to import"

image

js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
```diff
@@ -33,7 +33,7 @@ export const Oid4Vci = () => {
   const [credentialsIssuer, setCredentialsIssuer] =
     useState<CredentialsIssuer>();
 
-  usePromise(() => getIssuer(context), setCredentialsIssuer);
+  usePromise(() => getVCIssuer(context), setCredentialsIssuer);
```
IngridPuppet, line 37: My understanding of #44116 is that users self-requesting credential offers is no longer wanted. This call will precisely fail for users not having the "credential-offer-create" role. Do we still need this view altogether?

IngridPuppet (commented):
Hello @mposolda - I reviewed the last commit on the PR and left a few comments. Please could you check?

Awambeng (commented):
@mposolda I tried testing this on my side, but I still get a `403` response even though the request for retrieving the offer seems correct. I’m not sure if I’m missing something. @IngridPuppet could you please check whether this works on your side as well?

Image

I was also wondering whether it would make sense to show the **Verifiable Credentials** tab only when the client supports OID4VCI (i.e., when `oid4vci.enabled` is true). What do you think?

js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
mposolda: @IngridPuppet That is great question. When figuring some more, I am also not really sure if we still need this view in the account console. Moreover, I am not very sure about the custom Keycloak endpoint for credential-offer-uri creation TBH...

I've started the discussion https://github.com/keycloak/keycloak/discussions/44764 . Feedback welcome!

mposolda:
> @mposolda I tried testing this on my side, but I still get a `403` response even though the request for retrieving the offer seems correct. I’m not sure if I’m missing something. @IngridPuppet could you please check whether this works on your side as well?
> Image
> 
> I was also wondering whether it would make sense to show the **Verifiable Credentials** tab only when the client supports OID4VCI (i.e., when `oid4vci.enabled` is true). What do you think?

@Awambeng Yes, you are right. There are some pre-requisites, which need to be met to be able to see the QR code. Especially:

- The realm attribute `Verifiable credentials` must be enabled. Otherwise the `Verifiable credentials` view is entirely missing from account console

- The user logged in account console must be in `credential-offer-create` role.

- The `account-console` client must have `Enable OID4VCI` switch enabled (Switch in the `Advanced` tab of that client in the admin console

I agree it would likely makes sense to show the tab just if `account-console` client has the switch enabled and user is in the `credential-offer-create` role.

However at the same time, I am not 100% sure if this view in the account console makes sense at all :-) See this comment (and linked GH discussion) for the details: https://github.com/keycloak/keycloak/pull/44716#discussion_r2601754766 .

@Awambeng @IngridPuppet Right now, I am keeping the PR open, but not 100% sure about rather closing it or rewriting it to remove the OID4VCI view from the account console entirely? But may require further discussion...

js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
tdiesler: It'd be good to have that view for whoever creates credential offers for registered users. In the context of a University for example, it would be someone from administration who holds the 'credential-offer-create' role. That person would want to see not only the offerUri but also the associated QR code, which can then be passed on to the student's wallet (either out-of-band or pushed to the endpoint that receives such offers on a a spec compliant wallet)

This brings me to why the current getCredentialOfferURI(...) response is inadequate because it requires two calls: one for the uri and one for the qr code. Instead, it should work such that the uri is always returned and the qr code can be optionally returned in the same payload as well - it should just be another representation for the uri.

Currently, two calls to that endpoint create two distinct credential offers.

Awambeng:
> > @mposolda I tried testing this on my side, but I still get a `403` response even though the request for retrieving the offer seems correct. I’m not sure if I’m missing something. @IngridPuppet could you please check whether this works on your side as well?
> > Image
> > I was also wondering whether it would make sense to show the **Verifiable Credentials** tab only when the client supports OID4VCI (i.e., when `oid4vci.enabled` is true). What do you think?
> 
> @Awambeng Yes, you are right. There are some pre-requisites, which need to be met to be able to see the QR code. Especially:
> 
> * The realm attribute `Verifiable credentials` must be enabled. Otherwise the `Verifiable credentials` view is entirely missing from account console
> * The user logged in account console must be in `credential-offer-create` role.
> * The `account-console` client must have `Enable OID4VCI` switch enabled (Switch in the `Advanced` tab of that client in the admin console
> 
> I agree it would likely makes sense to show the tab just if `account-console` client has the switch enabled and user is in the `credential-offer-create` role.
> 
> However at the same time, I am not 100% sure if this view in the account console makes sense at all :-) See this comment (and linked GH discussion) for the details: [#44716 (comment)](https://github.com/keycloak/keycloak/pull/44716#discussion_r2601754766) .
> 
> @Awambeng @IngridPuppet Right now, I am keeping the PR open, but not 100% sure about rather closing it or rewriting it to remove the OID4VCI view from the account console entirely? But may require further discussion...

@mposolda Thank you for these clarifications. I had all the prerequisites you mentioned done, but I still wonder what could be causing the issue.
Anyway, since we are still not sure about these changes, maybe we should just focus on the discussion you started, and then we will surely know what needs to be done next.

IngridPuppet:
> @Awambeng Yes, you are right. There are some pre-requisites, which need to be met to be able to see the QR code. Especially:
> 
> * The realm attribute `Verifiable credentials` must be enabled. Otherwise the `Verifiable credentials` view is entirely missing from account console
> * The user logged in account console must be in `credential-offer-create` role.
> * The `account-console` client must have `Enable OID4VCI` switch enabled (Switch in the `Advanced` tab of that client in the admin console

Hello @mposolda - Thank you for making these precisions. We checked with @Awambeng, and the issue he faced was due to a browser extension causing security restrictions. Both of us could get the QR code on the view eventually. Additionally, we are a bit surprised that it is possible to get a credential offer without the client scope configuring the credential type assigned to the client. We'll dig into why this possible as of now; we believe it shouldn't.

> I agree it would likely makes sense to show the tab just if `account-console` client has the switch enabled and user is in the `credential-offer-create` role.

Certainly. These are basically the conditions for retrieving a credential offer.

> @Awambeng @IngridPuppet Right now, I am keeping the PR open, but not 100% sure about rather closing it or rewriting it to remove the OID4VCI view from the account console entirely? But may require further discussion...

This PR primarily resolves an existing bug on the view, so I think it should be merged in the meantime all is settled regarding whether it should exist and how it should be rewritten to meet its purpose.

js/apps/account-ui/src/oid4vci/Oid4Vci.tsx
IngridPuppet: That would mean rewriting the view for a potentially larger set of responsibilities... because right now the credential offer retrieved is basically for self-issuance, which makes the benefits of the view very limited given the 'credential-offer-create' role restriction. Anyway, I think it is not a good idea to supercharge this PR so I suggest @mposolda marks this point as resolved while discussion is taking place in #44764.

mposolda:
I am closing this PR. As mentioned in the comments above, current `Verifiable credentials` view in account console is broken as:

- It displays all the verifiable credentials (when some of those user should not be able to request self-issuance)
- It works only for users in the role `credential-offer-create`
- Every view of the credential displays the credential offer in the UI. But every "display" of the QR code calls REST endpoint `credential-offer-uri` for creating new credential offer. Hence there is new credential offer always created when user switches the credentials in the combo-box. This should be rather rewritten to use AIA, so that user will be redirected to dedicated Keycloak page from where the credential-offer can be created (Similarly like other Keycloak functionality, for example the functionality of updating user's own credentials or updating creating identity provider links). Related issue: #46197 .

Related issue for how the account console UI should ideally looks like: https://github.com/keycloak/keycloak/issues/46282

PR #44728, opened 2025-12-07
feat(admin-ui): use BrowserRouter instead of HashRouter

### LINKED ISSUE #32979: Use browser router instead of hash router for admin ui
### Description

Right now we use the browser router for account, we should use the same for admin ui

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

Afshal-Hassan:
### Summary

_Replace the HashRouter used in the Admin UI with BrowserRouter._

Closes #32979


edewit:
there is more to it then just this we also need to fix the routable tabs closing this one in favor of #51160

PR #44733, opened 2025-12-07
Sort realms alphabetically in manage realms list

### LINKED ISSUE #44667: Manage Realm list is not sorted alphabetically / cannot be sorted
### Before reporting an issue

### Area

welcome/ui

### Describe the bug

Manage Realm list is not sorted alphabetically / cannot be sorted --> /auth/admin/master/console/#/master/realms

### Version

26.4.6

### Regression

### Expected behavior

list be sorted alphabetically

### Actual behavior

sorted by creation date (I guess)

### How to Reproduce?

create different realms (different naming f.e.  d... / u... / c...)

### Anything else?

_No response_

amrelnaggar:
Closes #44667.

This makes realms list behaves in a consistent way with other lists in Clients, Users, Scopes, etc...

### Tests
```
cd js/apps/admin-ui
pnpm exec playwright install
pnpm test:integration -- test/masthead/realm.spec.ts
```

### DIFF
--- a/js/apps/admin-ui/test/masthead/realm.spec.ts
+++ b/js/apps/admin-ui/test/masthead/realm.spec.ts
@@ -1,4 +1,4 @@
-import { test } from "@playwright/test";
+import { test, expect } from "@playwright/test";
 import { v4 as uuid } from "uuid";
 import adminClient from "../utils/AdminClient.ts";
 import { DEFAULT_REALM } from "../utils/constants.ts";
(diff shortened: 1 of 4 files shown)


edewit (dismissed):

ahus1 (changes requested):
Thank you for the pull request. This change will sort all realm in memory, and for that it would retrieve the complete list of realms from the database which is not efficent. The sorting should be done on the database level instead. 

So this would be a better place to do this: 

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/model/jpa/src/main/java/org/keycloak/models/jpa/entities/RealmEntity.java#L52-L55

PR #44758, opened 2025-12-08
Additional fields for the Welcome Resource

### LINKED ISSUE #41425: Add more fields to the Welcome Page
### Description

As a follow-up on https://github.com/keycloak/keycloak/pull/41416#issuecomment-3116964423, the goal of this ticket is to add more fields to the Welcome Page:

* pre-populate account name with "admin"
* Add First and Last Names
* Email

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

slaskawi:
## Summary

This Pull Request introduces additional fields (First Name, Last Name and Email) to the Welcome Page.

## Screenshot

image

## Tickets

Fixes https://github.com/keycloak/keycloak/issues/41425


VinodAnandan:
@slaskawi Thank you for the PR! Would it be possible to allow an admin (privileged user) to register using a strong authentication or passwordless method, such as exclusively using passkeys?

slaskawi:
> @slaskawi Thank you for the PR! Would it be possible to allow an admin (privileged user) to register using a strong authentication or passwordless method, such as exclusively using passkeys?

@VinodAnandan The Welcome Screen creates an initial user for managing Keycloak. What you described sounds like a day-2 administration work.

themes/src/main/resources/theme/keycloak/welcome/index.ftl
```diff
@@ -101,6 +101,42 @@
...
+                      <div class="pf-v5-c-form__group-label">
+                        <label class="pf-v5-c-form__label" for="firstName">
+                          <span class="pf-v5-c-form__label-text">First name</span>
+                        </label>
+                      </div>
+                      <div class="pf-v5-c-form__group-control">
+                        <span class="pf-v5-c-form-control">
+                          <input id="firstName" type="text" name="firstName" autocomplete="given-name">
+                        </span>
+                      </div>
```
slaskawi, line 114: A note to myself - align the field order with the self-registration screen.

slaskawi:
@stianst Ready for review

stianst (approved):

stianst:
Thanks, since it may be a little bit controversial I'm leaving this one open for a few days before merging.

ahus1:
Thank you for raising this. I think this was triggerd when this admin is no longer a temporary admin account, but a permanent one.

Some thoughts / concerns: 

* The fields rendered on the welcome screen don't use the user profile settings about validation and required fields. 
* That admin account might not be a real user, but a technical user, so it doesn't have a firstname and a lastname. 
* What is the benefit of configuring an email address for an admin? You don't know yet if you connect the system to an email server anyway. As it might be a technical account, it wouldn't even have an email address?

stianst:
> Thank you for raising this. I think this was triggerd when this admin is no longer a temporary admin account, but a permanent one.
> 
> Some thoughts / concerns:
> 
> * The fields rendered on the welcome screen don't use the user profile settings about validation and required fields.

It matches the default user profile configuration, and an admin needs to be created as a first step before updating that. So don't see how that is relevant?

> * That admin account might not be a real user, but a technical user, so it doesn't have a firstname and a lastname.

An admin account should be a real user - what do you mean by a technical user?

> * What is the benefit of configuring an email address for an admin? You don't know yet if you connect the system to an email server anyway. As it might be a technical account, it wouldn't even have an email address?

It's optional - and for a real admin it would be best practice to set a first/last and email; which helps identifying who owns the account.

ahus1:
> An admin account should be a real user - what do you mean by a technical user?

I was thinking that this user might be call "Keycloak Admin" that is then possibly an account that is shared among multiple real people. It is not a good practice, but still quite common. 

> It matches the default user profile configuration

OK, so it is never trigger at a later point in time? I assume it is also not used for account recovery? If this is the case, I'm ok to proceed. The scenario where someone provisioned a master realm via a JSON with a specific user profile but without an admin account, that is then only a niche?

stianst:
> > An admin account should be a real user - what do you mean by a technical user?
> 
> I was thinking that this user might be call "Keycloak Admin" that is then possibly an account that is shared among multiple real people. It is not a good practice, but still quite common.

Shared users is really a no-no, and is actually one of the reasons why this PR is a good thing. Asking for a name and email makes it clear that this is intended to be a "personal" account.

> 
> > It matches the default user profile configuration
> 
> OK, so it is never trigger at a later point in time? I assume it is also not used for account recovery? If this is the case, I'm ok to proceed. The scenario where someone provisioned a master realm via a JSON with a specific user profile but without an admin account, that is then only a niche?

You could in theory log-in as this admin; then change the user profile; then delete all users in the realm, which would then allow creating a new admin user, but doubt anyone has ever done that. If we where worried about that, maybe we should then increase the checks to really only allow the welcome page to be used a single time, and not just when there are 0 users in the master realm.

ahus1 (approved):
Thank you for the additional import, approving.

Issue #44699, opened 2025-12-05
Not able to find key for credential signature if client scope was saved from admin console


mposolda:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

oid4vc

### Describe the bug

Assume following scenario:

- Import some realm with some OID4VCI client scopes
- Go to some client scope (EG. `sd-jwt-credential`) and save it without any changes. Especially leave the option *Signing Key ID* with the value *Use default (Realm's active signing key)* . At this point, the SigningKeyID is saved with empty string value instead of null
- Now try to send credentialRequest for the credential corresponding to this client scope. It fails due the `AbstractCredentialSigner.getKey` currently fallbacks to the realm active key just if signingKey is null .

It seems we should either make sure that signer is saved from admin console with value null or some different placeholder value (EG. `realm_active_key`) or `AbstractCredentialSigner.getKey` allows also empty strings (or the placeholder value if we use this option).

### Version

nightly (from 2025-12-05)

### Regression

- [ ] The issue is a regression

### Expected behavior

See description for the details.

### Actual behavior

See description for the details.

### How to Reproduce?

See description for the details.

### Anything else?

I am not 100% sure if there is a need for this option at all? Should it perhaps rather always use current realm active key for the specified algorithm similarly like all other Keycloak functionality is doing? Can we perhaps remove it?

Proposal to removing the option of `Signing Key ID` and instead keep only the option for selected signature algorithm (single algorithm like used for example in OIDC clients for stuff like `ID Token signature algorithm`): https://github.com/keycloak/keycloak/pull/44765#issuecomment-3647303529 . If we go this approach, it should fix this bug as well.

mposolda:
For the reference: Related discussion about possibility to remove the option for signingKey: https://github.com/keycloak/keycloak/pull/44765 . It seems Adorsys team is OK with that.

mposolda:
Update: In the meeting before Christmas, it was discussed that we keep only "Signing key ID" for now. There might be some use-case for it. However it is not 100% clear what the use-case is. I may try to revisit the discussion in the future...

PR #44765, opened 2025-12-09
[OID4VCI] Realign naming of attribute configuring algorithms for credential

### LINKED ISSUE #44621: [OID4VCI] Realign naming of attribute configuring algorithms for credential signing
### Description

The comment on the property aligns with the implementation (i.e., configures metadata entry `credential_signing_alg_values_supported`), but naming it `vc.proof_signing_alg_values_supported` could mislead into thinking the attribute rather configures metadata entry `proof_signing_alg_values_supported`. Consider renaming it `vc.credential_signing_alg_values_supported` for consistency.

https://github.com/keycloak/keycloak/blob/ae7e7ba084db880b7013df6b3adfade794a4179d/server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java#L74-L78

Make sure that the renaming is well propagated, including affected comments, tests, UI controls, and documentation files.

Also, I think it makes sense to discard symmetric algorithms from the default values of the configuration entry.

### Value Proposition

Improves the intelligibility of the configuration attributes.

### Goals

- Renaming the attribute
- Improve its default values

### Non-Goals

The ticket is not about making it possible to configure `proof_signing_alg_values_supported`. IMO, we shouldn't bother to.

### Discussion

N/A

### Notes

CC: @Captain-P-Goldfish

Vitalisn4:
This PR renames the credential signing algorithm attribute for clarity and updates the default filter settings to match the new naming. No functional changes beyond renaming and filtering adjustments.

Closes #44621

### DIFF
--- a/docs/documentation/server_admin/topics/oid4vci/vc-issuer-configuration.adoc
+++ b/docs/documentation/server_admin/topics/oid4vci/vc-issuer-configuration.adoc
@@ -248,7 +248,7 @@ Create a JSON file (e.g., `client-scopes.json`) with the following content:
     "vc.verifiable_credential_type": "my-vct",
     "vc.supported_credential_types": "credential-type-1,credential-type-2",
     "vc.credential_contexts": "context-1,context-2",
-    "vc.credential_signing_alg_values_supported": "ES256",
+    "vc.credential_signing_alg": "ES256",
     "vc.cryptographic_binding_methods_supported": "jwk",
     "vc.signing_key_id": "key-id-123456",
     "vc.display": "[{\"name\": \"IdentityCredential\", \"logo\": {\"uri\": \"https://university.example.edu/public/logo.png\", \"alt_text\": \"a square logo of a university\"}, \"locale\": \"en-US\", \"background_color\": \"#12107c\", \"text_color\": \"#FFFFFF\"}]",
@@ -358,9 +358,9 @@ _Default_: `$\{name}+`
 | The context values of the Verifiable Credential Type. +
 _Default_: `$\{name}+`
 
-| `vc.credential_signing_alg_values_supported`
+| `vc.credential_signing_alg`
 | optional
-| Supported signature algorithms for this credential. +
+| Supported signature algorithm for this credential. +
 _Default_: All asymmetric signing algorithms backed by realm keys.
 
 | `vc.cryptographic_binding_methods_supported`
--- a/docs/documentation/server_admin/topics/oid4vci/vc-issuer-configuration.adoc
+++ b/docs/documentation/server_admin/topics/oid4vci/vc-issuer-configuration.adoc
@@ -248,7 +248,7 @@ Create a JSON file (e.g., `client-scopes.json`) with the following content:
     "vc.verifiable_credential_type": "my-vct",
     "vc.supported_credential_types": "credential-type-1,credential-type-2",
     "vc.credential_contexts": "context-1,context-2",
-    "vc.proof_signing_alg_values_supported": "ES256",
+    "vc.credential_signing_alg_values_supported": "ES256",
     "vc.cryptographic_binding_methods_supported": "jwk",
     "vc.signing_key_id": "key-id-123456",
     "vc.display": "[{\"name\": \"IdentityCredential\", \"logo\": {\"uri\": \"https://university.example.edu/public/logo.png\", \"alt_text\": \"a square logo of a university\"}, \"locale\": \"en-US\", \"background_color\": \"#12107c\", \"text_color\": \"#FFFFFF\"}]",
@@ -358,10 +358,10 @@ _Default_: `$\{name}+`
 | The context values of the Verifiable Credential Type. +
 _Default_: `$\{name}+`
 
-| `vc.proof_signing_alg_values_supported`
+| `vc.credential_signing_alg_values_supported`
 | optional
 | Supported signature algorithms for this credential. +
-_Default_: All present keys supporting JWS algorithms in the realm.
+_Default_: All asymmetric signing algorithms backed by realm keys.
 
 | `vc.cryptographic_binding_methods_supported`
 | optional
--- a/server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/oid4vci/CredentialScopeModel.java
@@ -72,10 +72,9 @@ public class CredentialScopeModel implements ClientScopeModel {
     public static final String CONTEXTS = "vc.credential_contexts";
 
     /**
-     * If the credential is only meant for specific signing algorithms the global default list can be overridden here.
-     * The global default list is retrieved from the available keys in the realm.
+     * The credential signature algorithm. If it is not configured, then the realm active key is used to sign the verifiable credential
      */
-    public static final String SIGNING_ALG_VALUES_SUPPORTED = "vc.credential_signing_alg_values_supported";
+    public static final String SIGNING_ALG = "vc.credential_signing_alg";
 
     /**
      * if the credential is only meant for specific cryptographic binding algorithms the global default list can be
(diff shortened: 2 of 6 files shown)


IngridPuppet (approved):
@Vitalisn4 is with adorsys, and the changes were reviewed internally.

mposolda (dismissed):
@Vitalisn4 @IngridPuppet Thanks for this PR and +1 for only asymmetric algorithms.

I hope to merge if automated tests are OK.

mposolda:
@Vitalisn4 The failing test seems related to your changes. Could you please update the test to expect only asymmetric algorithms?

Vitalisn4:
Hello @mposolda , please I updated the logic for the tests to to pass, can you please check!

mposolda:
Thanks for the updates.

When looking more into this, I would like to doublecheck why it is needed that `credential_signing_alg_values_supported` from the metadata should contain just the algorithm supported by the current realm keys? Should not it contain all the algorithms supported by Keycloak? I did not found anything in the OID4VCI specification about limiting this, but maybe I am wrong?

- For example OIDC well-known endpoint returns all the algorithms in the metadata fields like `id_signing_alg_values_supported` . Even those not listed in the realm keys. Any reason why this one should be different and limit only to realm keys?


- With this PR, when I delete all the RSA keys from the realm `keys` tab (in particular RSA keys), I have the exception like:
```
2025-12-12 16:11:43,052 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-31) Uncaught server error: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:385)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at java.base/java.util.Collections$UnmodifiableList.get(Collections.java:1502)
	at org.keycloak.protocol.oid4vc.model.CredentialBuildConfig.parse(CredentialBuildConfig.java:101)
	at org.keycloak.protocol.oid4vc.model.SupportedCredentialConfiguration.parse(SupportedCredentialConfiguration.java:140)
	at org.keycloak.protocol.oid4vc.issuance.OID4VCIssuerWellKnownProvider.lambda$getSupportedCredentials$6(OID4VCIssuerWellKnownProvider.java:463)

```

- Even when for example RSA key is not available in realm keys, it does not mean that Keycloak realm does not support RS256 algorithm. For example when I delete all the RSA realm keys and I login to account-console as some user, login still works and token is signed with RS256 key. And you can see that there is new key "automatically" created inside realm keys, which is marked as a "fallback key" . In other words, realm supports RS256 (all other algorithms) even if there is not active RSA key.


Also I wonder why there is a need to support multiple signature algorithms for any verifiable credential? For example in case of normal OIDC client, there is only one algorithm, which you can select for signing the ID token (configured as client attribute `id_token_signature_attribute` and configurable in the admin console as well). So I wonder why we support multiple algorithms for the verifiable credential considering that there is no way in `authorization_details` parameter or in credential-request to select which of those algorithms you prefer? Isn't it sufficient to support only single preferred algorithm?

When I was digging more in this, I ended up with the updates done in the follow-up PR sent to your branch https://github.com/adorsys/keycloak-oid4vc/pull/212 . Are you ok with those changes? If yes and you merge that PR, it should automatically update this PR AFAIK.

Also as a follow-up, I was thinking about making credential signature to use same approach like OIDC client (when for example signature algorithm for IDToken or USerInfo is going to be chosen). So in particular:

- Remove option `vc.signing_key_id` from both configuration and admin console and instead add the `vc.credential_signing_alg` to the UI of client scope (currently it is missing from the admin console UI) . As I am not sure if there is use-case for select specific `kid` ? Specific `kid` doesn't work well with the key rotations.

- The key selected for signing the credential would be based either on the configured `vc.credential_signing_alg` or alternatively on the realm default key (RSA256 specified by `Constants.DEFAULT_SIGNATURE_ALGORITHM`) . It will always use "active-key" for the specified algorithm (instead of using specific `kid`). 

Would Adorsys team be OK with this approach?

IngridPuppet:
Hi @mposolda - Thank you for the insightful contribution.

> When looking more into this, I would like to doublecheck why it is needed that `credential_signing_alg_values_supported` from the metadata should contain just the algorithm supported by the current realm keys? Should not it contain all the algorithms supported by Keycloak? I did not found anything in the OID4VCI specification about limiting this, but maybe I am wrong?

This is a per-implementation decision. Maybe because there is not much need to declare an algorithm supported while we know in advance we don't have a key to sign credentials on that algorithm; which is quite different from just verifying signatures. I admit to not being aware of the automatic key generation mechanism you later mention. Does that apply to all signing algorithms globally supported by Keycloak?

> For example OIDC well-known endpoint returns all the algorithms in the metadata fields like `id_signing_alg_values_supported` . Even those not listed in the realm keys. Any reason why this one should be different and limit only to realm keys?

If same logic and that has always worked so far, then **I agree it makes sense to remain consistent.**

> Also I wonder why there is a need to support multiple signature algorithms for any verifiable credential? For example in case of normal OIDC client, there is only one algorithm, which you can select for signing the ID token (configured as client attribute `id_token_signature_attribute` and configurable in the admin console as well). So I wonder why we support multiple algorithms for the verifiable credential considering that there is no way in `authorization_details` parameter or in credential-request to select which of those algorithms you prefer? Isn't it sufficient to support only single preferred algorithm?

I recall a mechanism whereby an Issuer could issue a batch of credentials signed with different algorithms, so a wallet could present one or the other given the cryptographic capabilities of a verifier. But looks like that was dropped as I am unable to find a reference in the spec. I'll open a new issue if the need arises in the future. For now, **a single preferred algorithm should be okay;** the current implementation just uses the first algorithm in the list actually. Nevertheless, I find the move somehow too radical. There should be some valid current or future reasons why the spec is keeping the property plural.

> * Remove option `vc.signing_key_id` from both configuration and admin console and instead add the `vc.credential_signing_alg` to the UI of client scope (currently it is missing from the admin console UI) . As I am not sure if there is use-case for select specific `kid` ? Specific `kid` doesn't work well with the key rotations.

We'll be sending a PR for supporting `vc.credential_signing_alg` on the UI shortly (https://github.com/adorsys/keycloak-oid4vc/pull/205). As for removing configuration support for `vc.signing_key_id`, why not simply leave that to the user in total transparence of implications regarding limited key rotation? It might be useful to be specific in case there are multiple keys available for the same algorithm.

> Would Adorsys team be OK with this approach?

Provided minor updates on our side, this won't break any of our existing flows. So, yes. Thank you.

mposolda (approved):
@Vitalisn4 @IngridPuppet Thank you very much for the updates! I hope we can merge the PR if tests are OK. Maybe rebase of this PR with Keycloak main would be needed, but not 100% sure yet. We will see once tests are finished.

mposolda:
@IngridPuppet Answering some of your other questions inline.

> This is a per-implementation decision. Maybe because there is not much need to declare an algorithm supported while we know in advance we don't have a key to sign credentials on that algorithm; which is quite different from just verifying signatures. I admit to not being aware of the automatic key generation mechanism you later mention. Does that apply to all signing algorithms globally supported by Keycloak?

Yes, AFAIK this is supported by all signing algorithms supported by Keycloak

> > * Remove option `vc.signing_key_id` from both configuration and admin console and instead add the `vc.credential_signing_alg` to the UI of client scope (currently it is missing from the admin console UI) . As I am not sure if there is use-case for select specific `kid` ? Specific `kid` doesn't work well with the key rotations.
> 
> We'll be sending a PR for supporting `vc.credential_signing_alg` on the UI shortly ([adorsys#205](https://github.com/adorsys/keycloak-oid4vc/pull/205)). As for removing configuration support for `vc.signing_key_id`, why not simply leave that to the user in total transparence of implications regarding limited key rotation? It might be useful to be specific in case there are multiple keys available for the same algorithm.

@IngridPuppet I am thinking about removing `vc.signing_key_id` because of:

- Consistency with other places in Keycloak. It is not supported by any other Keycloak functionality to select specific `kid` . Especially for classic OIDC flows, you can select preferred algorithm for the client for ID-token signature, user-info signature or any other. You never select specified `kid` . The active key for the specified algorithm (key with "biggest priority") is always chosen.

- Ability to choose specified `kid` is another source of complexity and another possible source of bugs. In particular this one causes #44699
 
- Specified `kid` may not work well with key rotation. Administrator will always need to re-configure if he want to rotate keys

So my preferred way would be to remove this option `vs.signing_key_id`. However if adorsys team has a use-case for keeping, it is fine with me to keep it (if we document the impact on the key rotation and if it is empty by default, which is already the case). WDYT?

mposolda:
@Vitalisn4 @IngridPuppet Is it possible if you rebase the branch to latest Keycloak main? I hope that will help with fixing UI test failures. Those are unrelated to this PR and were fixed in Keycloak main recently.

IngridPuppet:
> So my preferred way would be to remove this option `vs.signing_key_id`. However if adorsys team has a use-case for keeping, it is fine with me to keep it (if we document the impact on the key rotation and if it is empty by default, which is already the case). WDYT?

Hi @mposolda - There is no use case for the option that I can recall at the moment. However, after creating the ticket, I suggest to confirm with other members of the SIG if they are okay as well with the removal.

mposolda:
@IngridPuppet Thanks! Yes, ok. I will check with rest of the team as well. Maybe will send draft PR to quickstart the discussion for this...

Issue #44266, opened 2025-11-17
Provide additional benchmarks for more than 100k users


ahus1:
### Description

Gather performance numbers for the following setup to show that the performance of Keycloak doesn't depend on the number of users

* 30M users in the database w/ 10k unique passwords
* 250 logins per second
* 3k token refreshes per second
* 3M sessions in the database at the end of the test (would mean 3.3 hours test run), or improvement to the dataset provider - create sessions with the dataset provider. Run at least 30 minutes so the scheduled tasks do some work. Adjust the session timeout to keep a level 3 millions session during the test. Maybe adjust the last refresh time and creation time; also ensure to create online sessions, not offline sessions.
* Variants 0/10/20 ms extra delay - ensure that we only set up the data once
* Version to test: 26.4.x
* CPU / Memory: use as much as needed, same for DB instance type
* Number of Pods: Start with 3 and see if it is enough. Scale to 6 if needed
* Scale up the OpenShift ingress controllers if needed
* Multi-AZ single-cluster setup, ARM (our default)
* Caching as default unless proven otherwise. 

### TODOs

* [x] Enhance the dataset generation + Gatling: Generate users with a limited number of unique passwords 
* [x] Enhance the dataset generation: Generate regular sessions, with staged expiration times
* [x] Run the test with 0 ms extra delay and lock the version (26.4.5 at the time of writing). Expect 99th pct to be less than 250 ms. Document response times, CPU and memory, DB instance. 

### Motivation

We didn't see a limitation in Keycloak around the 100k users, still it is hard to explain again and again that it should not make a difference for 1M or 10M users. 

### Details

This is a one-off test to show that the number of users doesn't matter. 

As a follow-up, we might change our nightly runs to regularly use 1M or 10M users.

pruivo:
### Environment

**Openshift**
* Version 4.17 deployed across three availability zones in eu-west-1.
* AWS VM Type: c8g.8xlarge

**Amazon Aurora PostgreSQL**
* Version 17.5 deployed across three availability zones in eu-west-1.
* VM Type: db.r8g.2xlarge

**Keycloak**
* CPU requests/limits: 24 vCPU
* Memory requests/limits: 4GB/8GB
* Database Connection Pool: 96

**Data**
* 30 million users
* 3 million sessions

### Summary

The table below summarizes the 99th percentiles observed by Gatling.

| RTT Latency | 26.3* | 26.4* | 26.4.7 |
|--------|--------|--------|--------|
[...]

PR #44788, opened 2025-12-09
Avoid flushing user information in batch mode

### LINKED ISSUE #44787: Avoid flushing user information in batch mode
### Description

When creating new users in keycloak benchmark, the persistence context is flushed for each user, joined group and added role. This is slow. 

### Value Proposition

By avoiding the flush when in batch mode, we expect to be faster in creating the users. We are aiming for 30M users.

### Goals

Faster setup of new users (would work for dataset generation, but also for importing)

### Non-Goals

Get rid of all other "flush" in the JPA layer.

### Discussion

_No response_

### Notes

I'll create a PR

ahus1:
Closes #44787

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
@@ -44,6 +44,7 @@ import jakarta.persistence.criteria.Root;
 import org.keycloak.authorization.fgap.AdminPermissionsSchema;
 import org.keycloak.common.util.Time;
 import org.keycloak.component.ComponentModel;
+import org.keycloak.connections.jpa.support.EntityManagers;
 import org.keycloak.credential.CredentialModel;
 import org.keycloak.credential.UserCredentialStore;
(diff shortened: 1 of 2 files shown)


ahus1:
@shawkins - this is trying to build on top of your previous optimizations at least to the extend that we need it to create a lot of users (see #44266). 

Could you please comment on this approach? I know it is not complete, but it might be good enough for us and a step in the right direction?

cc: @pruivo

shawkins:
> @shawkins - this is trying to build on top of your previous optimizations at least to the extend that we need it to create a lot of users (see https://github.com/keycloak/keycloak/issues/44266).

Yes, it would be great to start addressing more of the explicit flushes. From what I recall before and limited testing, supressing all flushes via the proxy during batching cut the import times for large numbers of simple users by half. What kind of performance improvement are you seeing with this?

ahus1:
For the creation of users in a local H2 database, creating 20_000 users is down from 55 seconds to 35 seconds, so it is a significant change IMHO.

shawkins (approved):
LGTM

stianst (approved):

PR #44792, opened 2025-12-09
Avoid lookup of existing workflow instances when not needed

### LINKED ISSUE #44791: Workflows look up entries from the database on each event
### Before reporting an issue

### Area

_No response_

### Describe the bug

In `DefaultWorkflowProvider#processEvent()`, each event will call `stateProvider.getScheduledStepsByResource` which touches the database and is therefore slow.

### Version

main

### Regression

### Expected behavior

Only touch the database when there are workflow, and a relevant entity is at hand.

### Actual behavior

It touches the database even when there is no workflow for the realm configured.

### How to Reproduce?

Enable SQL logging

### Anything else?

I'll create a PR

ahus1:
Closes #44791

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
@@ -246,12 +246,12 @@ public class DefaultWorkflowProvider implements WorkflowProvider {
             EventBasedWorkflow provider = new EventBasedWorkflow(session, getWorkflowComponent(workflow.getId()));
 
             try {
-                DefaultWorkflowExecutionContext context = new DefaultWorkflowExecutionContext(session, workflow, event);
-
                 if (!provider.supports(event.getResourceType())) {
                     // Prevents loading of scheduled steps when this resource type is not supported for the workflow
                     return;
                 }
+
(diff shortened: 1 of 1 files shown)


pedroigor (dismissed):

model/jpa/src/main/java/org/keycloak/models/workflow/DefaultWorkflowProvider.java
```diff
@@ -247,9 +246,20 @@ private void processEvent(Stream<Workflow> workflows, WorkflowEvent event) {
             EventBasedWorkflow provider = new EventBasedWorkflow(session, getWorkflowComponent(workflow.getId()));
 
             try {
-                ScheduledStep scheduledStep = scheduledSteps.get(workflow.getId());
                 DefaultWorkflowExecutionContext context = new DefaultWorkflowExecutionContext(session, workflow, event);
 
+                if (!provider.supports(event.getResourceType())) {
```
pruivo, line 249: The `DefaultWorkflowExecutionContext` constructor invokes `realm.getComponentsStream(getId(), WorkflowStepProvider.class.getName())` in one code path. You may want to move this check above line 249.

pruivo (approved):

sguilhen (approved):

pedroigor (approved):

PR #44794, opened 2025-12-10
CredentialRequest with credentialIdentifier does not work when creden…

### LINKED ISSUE #44793: CredentialRequest with credentialIdentifier does not work when credential-offer was created by OAuth2 authorization_code grant
### Before reporting an issue

### Area

oid4vc

### Describe the bug

In case that there is credentialRequest sent to the OID4VCI credential endpoint with the `credential_identifier` parameter, which relates to the parameter returned in the OAuth2 token response (in case of OAuth2 `authorization_code` grant), then the request would fail with the error like:
```
2025-12-10 08:44:01,390 ERROR [org.keycloak.protocol.oid4vc.issuance.OID4VCIssuerEndpoint] (executor-thread-12) Unexpected login user: john != john
```

### Cause

This update https://github.com/keycloak/keycloak/issues/44642 broke this functionality as now stored credentialOffer is supposed to store the `user_id` , however during creation of credential-offer in case of OAuth2 authorization_code grant ( done in `OID4VCAuthorizationDetailsProcessor.process`) , there is still `username` stored inside the credential-offer instead of user ID

### Version

nightly (from 2025-12-10)

### Regression

### Expected behavior

Credential request should work when sent with `credential_identifier` and credential-offer was created by OAuth2 authz code flow

### Actual behavior

Credential request fails when sent with `credential_identifier` and credential-offer was created by OAuth2 authz code flow

### How to Reproduce?

See details

### Anything else?

_No response_

mposolda:
…tial-offer was created by OAuth2 authorization_code grant

closes #44793

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BiFunction;
 
 import jakarta.ws.rs.core.HttpHeaders;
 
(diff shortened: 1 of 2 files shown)


testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
IngridPuppet, line 1: Not related to your changes, but the IDE is issuing multiple warnings in this file, including unused parameters, deprecated syntax (Arrays.asList), and unnecessary template typing. It'd be great if you could clean up the file at once.

IngridPuppet (approved):
LGTM

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
mposolda: @IngridPuppet Agree, that cleanup of that test class is never bad. I've added separate commit to this PR with some cleanups of this class. Now I don't see any warnings in the IDE (at least Intellij Idea). WDYT?

IngridPuppet: Exactly what I was pointing to. Thank you.

IngridPuppet (approved):

mposolda:
@IngridPuppet Thanks!

PR #44822, opened 2025-12-10
fix: moving nonserver defaults out of application.properties

shawkins:
closes: #42332

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
@@ -10,6 +10,7 @@ import java.util.function.BooleanSupplier;
 
 import org.keycloak.common.Profile;
 import org.keycloak.config.CachingOptions;
+import org.keycloak.config.CachingOptions.Mechanism;
 import org.keycloak.config.Option;
 import org.keycloak.infinispan.util.InfinispanUtils;
 import org.keycloak.quarkus.runtime.Environment;
@@ -35,6 +36,10 @@ final class CachingPropertyMappers implements PropertyMapperGrouping {
     public List<PropertyMapper<?>> getPropertyMappers() {
         List<PropertyMapper<?>> staticMappers = List.of(
                 fromOption(CachingOptions.CACHE)
+                        .transformer(
+                                (value, context) -> org.keycloak.common.util.Environment.isNonServerMode()
+                                        ? Mechanism.local.name()
+                                        : Optional.ofNullable(value).orElse(Mechanism.ispn.name()))
                         .paramLabel("type")
                         .build(),
                 fromOption(CachingOptions.CACHE_STACK)
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
@@ -292,7 +292,7 @@ public class PropertyMapper<T> {
         // fall back to the transformer when no mapper is explicitly specified in .mapFrom()
         var theMapper = parentValue && parentMapper != null ? this.parentMapper : this.mapper;
         // since our mapping logic assumes fully resolved values, we cannot reliably map if Expressions are disabled
-        if (Expressions.isEnabled() && theMapper != null && (!name.equals(getFrom()) || parentValue)) {
+        if (Expressions.isEnabled() && theMapper != null && (name.equals(getTo()) || parentValue)) {
             mappedValue = theMapper.map(getNamedProperty().orElse(null), value, context);
             mapped = true;
         }
--- a/services/src/main/java/org/keycloak/url/HostnameV2ProviderFactory.java
+++ b/services/src/main/java/org/keycloak/url/HostnameV2ProviderFactory.java
@@ -23,6 +23,7 @@ import java.util.Optional;
 
 import org.keycloak.Config;
 import org.keycloak.common.Profile;
+import org.keycloak.common.util.Environment;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.provider.EnvironmentDependentProviderFactory;
 import org.keycloak.urls.HostnameProvider;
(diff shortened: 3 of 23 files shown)


vmuzikar:
Let's merge after 26.5

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
```diff
@@ -292,7 +292,7 @@ private ConfigValue transformValue(String name, ConfigValue configValue, ConfigS
         // fall back to the transformer when no mapper is explicitly specified in .mapFrom()
         var theMapper = parentValue && parentMapper != null ? this.parentMapper : this.mapper;
         // since our mapping logic assumes fully resolved values, we cannot reliably map if Expressions are disabled
-        if (Expressions.isEnabled() && theMapper != null && (!name.equals(getFrom()) || parentValue)) {
+        if (Expressions.isEnabled() && theMapper != null && (name.equals(getTo()) || parentValue)) {
```
mabartos, line 295: @shawkins Can you elaborate why we need this change? What does it solve? So, the `mapFrom` transformation still works properly, right?

services/src/main/java/org/keycloak/url/HostnameV2ProviderFactory.java
```diff
@@ -45,6 +46,9 @@ public class HostnameV2ProviderFactory implements HostnameProviderFactory, Envir
 
     @Override
     public void init(Config.Scope config) {
+        if (Environment.isNonServerMode()) {
```
mabartos, line 49: @shawkins There might be even other factories that do not have to be initialized, right?

mabartos (commented):
@shawkins Thanks! I've just added some comments.

services/src/main/java/org/keycloak/url/HostnameV2ProviderFactory.java
shawkins: Yes, in a team meeting we briefly covered that this could be generalized. Since optimizing the start of non-server commands isn't important yet, we'll probably do this only when more of them become problematic wrt configuration (or initialization time).

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
mabartos: @shawkins Could you please briefly comment why is this necessary? Thanks

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
```diff
@@ -35,6 +36,10 @@ final class CachingPropertyMappers implements PropertyMapperGrouping {
     public List<PropertyMapper<?>> getPropertyMappers() {
         List<PropertyMapper<?>> staticMappers = List.of(
                 fromOption(CachingOptions.CACHE)
+                        .transformer(
+                                (value, context) -> org.keycloak.common.util.Environment.isNonServerMode()
+                                        ? Mechanism.local.name()
```
vmuzikar, line 41: Should we also consider the dev profile here?

vmuzikar (commented):
@shawkins Thank you for the PR and sorry for the late response.

After giving it more thought, I'm not sure if the previous approach with `application.properties` wasn't more clear.
* We still need to configure a bunch of profile scoped options in `application.properties`,  e.g. raw Quarkus options, SPI options.
* Dev profile scope options are still present in `application.properties`, so it's not really consistent with the nonserver profile.
* When using `application.properties`, all profile scope overrides are at single place, not scattered across the code base.

shawkins:
> @shawkins Thank you for the PR and sorry for the late response.
> 
> After giving it more thought, I'm not sure if the previous approach with `application.properties` wasn't more clear.

It's clearer, but also not really effective.

> * We still need to configure a bunch of profile scoped options in `application.properties`,  e.g. raw Quarkus options, SPI options.

For the most part we only need things in there that don't have options with defaults. 

> * Dev profile scope options are still present in `application.properties`, so it's not really consistent with the nonserver profile.

Unfortunately dev vs prod is somewhat  orthogonal to server vs nonserver. We build as dev vs prod, then at runtime switch to nonserver. Beyond that any concept we have of configuration profiles is pretty broken.

> * When using `application.properties`, all profile scope overrides are at single place, not scattered across the code base.

They already are - we have other transformer logic that is profile dependent.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMapper.java
shawkins: We need the mapper applied in more circumstances.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
shawkins: If you mean force the cache to be local in dev, I don't think that's currently forbidden or a hard requirement. It is certainly something else we could force, but I'd prefer a follow-up.

vmuzikar:
> but also not really effective

Do you mean it did not work properly?

> They already are - we have other transformer logic that is profile dependent.

I know but I'd rather see getting rid of those where we can rather than adding more.

shawkins:
> > but also not really effective
> 
> Do you mean it did not work properly?

Not intuitively, no. If you for example put cache=ispn in the keycloak.conf, then run an import command, I would not expect it to join the cache cluster, but it currently does. We actually have a workaround for this in the operator.
 
> > They already are - we have other transformer logic that is profile dependent.
> 
> I know but I'd rather see getting rid of those where we can rather than adding more.

I don't believe that is always possible or advisable if it results in more counterintuitive behavior.

vmuzikar:
> Not intuitively, no. If you for example put cache=ispn in the keycloak.conf, then run an import command, I would not expect it to join the cache cluster, but it currently does. We actually have a workaround for this in the operator.

Right...

> I don't believe that is always possible or advisable if it results in more counterintuitive behavior.

Ok, thinking about that more I agree.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/CachingPropertyMappers.java
vmuzikar: Ok, let's leave as is.

vmuzikar: I did not realize before that we have `%dev.kc.cache=local` in `application.property` which makes it non-null, so it doesn't default to `ispn`.

vmuzikar (approved):
LGTM.

@mabartos Do you want to give it another look too?

mabartos (approved):

mabartos (approved):
Still LGTM.

PR #44828, opened 2025-12-11
[OID4VCI] Add mapper for mapping unmanaged attributes

### LINKED ISSUE #44780: [OID4VCI] Missing User Attribute Mapper
### Description

For OpenID4VCI there is currently a mapper called `User Attribute Mapper` which is actually just a `User Property Mapper` that has a selection field of user-properties to be mapped:

Image

What we still need is an actual `User Attribute Mapper` that will allow us to map `Unmanaged Attributes` into the credentials.

Captain-P-Goldfish:
closes #44780


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserPropertyMapper.java
```diff
@@ -0,0 +1,141 @@
+/*
+ * Copyright 2024 Red Hat, Inc. and/or its affiliates
```
IngridPuppet: The copyright headers are actually no longer needed. See https://github.com/keycloak/keycloak/pull/44106#issuecomment-3521908655.
```suggestion
 * Copyright 2025 Red Hat, Inc. and/or its affiliates
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserPropertyMapper.java
```diff
@@ -0,0 +1,141 @@
...
+import org.keycloak.provider.ProviderConfigProperty;
+
+/**
+ * Allows adding user properties to the credential subject
+ *
+ * @author <a href="https://github.com/wistefan">Stefan Wiedemann</a>
+ */
+public class OID4VCUserPropertyMapper extends OID4VCMapper {
+
+    public static final String MAPPER_ID = "oid4vc-user-attribute-mapper";
```
IngridPuppet: Call it `oid4vc-user-property-mapper`? I'm trying to avoid sources of confusion.
```suggestion
public class OID4VCUserPropertyMapper extends OID4VCMapper {

    public static final String MAPPER_ID = "oid4vc-user-property-mapper";
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserPropertyMapper.java
```diff
@@ -0,0 +1,141 @@
...
+    public static final String AGGREGATE_ATTRIBUTES_KEY = "aggregateAttributes";
+    public static final List<String> SELECTABLE_PROPERTIES = List.of(UserModel.USERNAME,
+                                                                     UserModel.LOCALE,
+                                                                     UserModel.FIRST_NAME,
+                                                                     UserModel.LAST_NAME,
+                                                                     UserModel.DISABLED_REASON,
+                                                                     UserModel.EMAIL,
+                                                                     UserModel.EMAIL_VERIFIED);
```
IngridPuppet: Do you confirm this should be mappable? `UserModel.EMAIL_VERIFIED`

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
```diff
@@ -58,12 +47,9 @@ public class OID4VCUserAttributeMapper extends OID4VCMapper {
...
         ProviderConfigProperty userAttributeConfig = new ProviderConfigProperty();
         userAttributeConfig.setName(USER_ATTRIBUTE_KEY);
-        userAttributeConfig.setLabel("User attribute");
-        userAttributeConfig.setHelpText("The user attribute to be added to the credential subject.");
-        userAttributeConfig.setType(ProviderConfigProperty.LIST_TYPE);
-        userAttributeConfig.setOptions(
-                List.of(UserModel.USERNAME, UserModel.LOCALE, UserModel.FIRST_NAME, UserModel.LAST_NAME,
-                        UserModel.DISABLED_REASON, UserModel.EMAIL, UserModel.EMAIL_VERIFIED));
+        userAttributeConfig.setLabel("User property");
+        userAttributeConfig.setHelpText("The user attributes to be added to the credential subject.");
```
IngridPuppet: It seems you meant "User attribute" here.
```suggestion
        userAttributeConfig.setLabel("User attribute");
        userAttributeConfig.setHelpText("The user attribute to be added to the credential subject.");
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
```diff
@@ -18,31 +18,20 @@
...
 import org.keycloak.provider.ProviderConfigProperty;
 
 /**
- * Allows to add user attributes to the credential subject
+ * Allows adding user attributes to the credential subject
  *
  * @author <a href="https://github.com/wistefan">Stefan Wiedemann</a>
  */
-public class OID4VCUserAttributeMapper extends OID4VCMapper {
+public class OID4VCUserAttributeMapper extends OID4VCUserPropertyMapper {
```
IngridPuppet, line 45: Having two mappers for relatively the same goal can be confusing. I guess the reason was to make it possible to select known user attributes on the admin console with one mapper, while allowing custom, "unmanaged" attributes to be mappable with the other mapper. 

Is there some technical limitation regarding the possibility of a dropdown with predefined attributes while allowing the input of custom values? It seems this property type is meant for that:

```java
property.setType(ProviderConfigProperty.USER_PROFILE_ATTRIBUTE_LIST_TYPE);
```

Are there other motivations for keeping two mappers that you would kindly share?

IngridPuppet (commented):
Hello @Captain-P-Goldfish - Thank you for implementing this flexibility regarding the mapping of user attributes/properties on the admin console. I reviewed the PR and left some comments. Please could you check?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserPropertyMapper.java
Captain-P-Goldfish: this is actually the original implementation. github is not showing it correctly.

I know what you mean but I actually would say that I don't know. I can surely imagine that there are some usecases in the future where email-verified might be of interest for very short-living credentials.
But the disabled_reason makes no sense at all for me. I think we can remove this one.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserPropertyMapper.java
Captain-P-Goldfish: I agree I was just thinking of backwards compatibility, so I am really happy that you are mentioning this one :-)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
Captain-P-Goldfish: I just did not check properly...
Thanks a lot for pointing that out I will change it.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
```diff
@@ -36,34 +36,37 @@
...
         ProviderConfigProperty userAttributeConfig = new ProviderConfigProperty();
         userAttributeConfig.setName(USER_ATTRIBUTE_KEY);
         userAttributeConfig.setLabel("User attribute");
         userAttributeConfig.setHelpText("The user attribute to be added to the credential subject.");
-        userAttributeConfig.setType(ProviderConfigProperty.LIST_TYPE);
-        userAttributeConfig.setOptions(
-                List.of(UserModel.USERNAME, UserModel.LOCALE, UserModel.FIRST_NAME, UserModel.LAST_NAME,
-                        UserModel.DISABLED_REASON, UserModel.EMAIL, UserModel.EMAIL_VERIFIED));
+        userAttributeConfig.setType(ProviderConfigProperty.USER_PROFILE_ATTRIBUTE_LIST_TYPE);
+        userAttributeConfig.setOptions(SELECTABLE_PROPERTIES);
```
IngridPuppet: Setting options does not appear necessary for this property type. Could you double check?
```suggestion
        userAttributeConfig.setType(ProviderConfigProperty.USER_PROFILE_ATTRIBUTE_LIST_TYPE);
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
```diff
@@ -36,34 +36,37 @@
...
 
     private static final List<ProviderConfigProperty> CONFIG_PROPERTIES = new ArrayList<>();
 
     static {
         ProviderConfigProperty subjectPropertyNameConfig = new ProviderConfigProperty();
         subjectPropertyNameConfig.setName(CLAIM_NAME);
         subjectPropertyNameConfig.setLabel("Claim Name");
-        subjectPropertyNameConfig.setHelpText("The name of the claim added to the credential subject that is extracted " +
-                                                      "from the user attributes.");
+        subjectPropertyNameConfig.setHelpText("The user attributes to be added to the credential subject.");
```
IngridPuppet:
```suggestion
        subjectPropertyNameConfig.setHelpText("The user attribute to be added to the credential subject.");
```

IngridPuppet (approved):
Thank you for the update. The new changes address my concerns, but could you check the minor comments below?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
Captain-P-Goldfish: you are right. Removed the options.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCUserAttributeMapper.java
Captain-P-Goldfish: @IngridPuppet  I checked the explanations from the original OIDCMappers which are a lot better and I tried to use these now.

**User Attribute:** Name of stored user attribute which is the name of an attribute within the UserModel.attribute map.
**Token Claim Name:** Name of the claim to insert into the token. This can be a fully qualified name such as 'address.street'. In this case, a nested JSON object is created. To prevent nesting and use dot literally, escape the dot with a backslash (\.).

mposolda (approved):
@Captain-P-Goldfish @IngridPuppet Thanks for the PR and review! Agree that this capability would be useful in Keycloak. Came around this and was also thinking about creating an issue, so I am glad that we already have the issue and PR :-) 

I hope to merge if tests are OK. I've rebased the PR on top of latest main

PR #44840, opened 2025-12-11
[OID4VCI] Make natural_person configuration available in all formats

tdiesler:
closes #44833


tdiesler:
As we discussed yesterday, I added commits for ...

- set scope default = false - so that these are not automatically added to new clients
- remove natural_person_ld from preinstalled scopes - as this is broken anyway

See #44875

core/src/main/java/org/keycloak/OID4VCConstants.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
```
IngridPuppet: It turns out copyright headers are not longer prefered. See https://github.com/keycloak/keycloak/pull/44106#issuecomment-3521908655.

```suggestion
/*
 * Copyright 2026 Red Hat, Inc. and/or its affiliates
 * and other contributors as indicated by the @author tags.
```

core/src/main/java/org/keycloak/VCFormat.java
```diff
@@ -15,34 +15,61 @@
...
+            }
+        }
+        throw new IllegalArgumentException("Unknown VC format: " + value);
+    }
+
+    public static VCFormat fromScopeName(String name) {
+        if (name.endsWith("_jwt")) return JWT_VC;
+        if (name.endsWith("_ld")) return LDP_VC;
+        return SD_JWT_VC;
+    }
```
IngridPuppet: Could you please add a comment clarifying why the flexibility with the `endsWith` calls are needed? Can we have an explicit line for SD_JWT_VC for consistency?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/JwtCredentialBuilder.java
```diff
@@ -60,7 +60,7 @@ public JwtCredentialBuilder(TimeProvider timeProvider, UnaryOperator<Instant> is
 
     @Override
     public String getSupportedFormat() {
-        return Format.JWT_VC;
+        return VCFormat.JWT_VC.getValue();
     }
```
IngridPuppet, line 67: The signature of the interface method is controlled by the OID4VCI code, and I think it makes sense to change the return type there to the new `VCFormat` enum. WDYT?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCMapper.java
```diff
@@ -118,10 +118,10 @@ public List<String> getMetadataAttributePath() {
...
 
     protected List<String> getAttributePrefix() {
-        return switch (Optional.ofNullable(format).orElse("")) {
-            case Format.JWT_VC, Format.LDP_VC -> List.of(CREDENTIAL_SUBJECT);
-            default -> Collections.emptyList();
-        };
+        if (VCFormat.SD_JWT_VC.getValue().equals(format))
+            return Collections.emptyList();
+        else
+            return List.of(CREDENTIAL_SUBJECT);
```
IngridPuppet: Do you mind adding curly braces?  The current formatting gives Python vibes.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/signing/JwtCredentialSignerFactory.java
```diff
@@ -17,14 +17,14 @@
...
 import org.keycloak.models.KeycloakSession;
-import org.keycloak.protocol.oid4vc.model.Format;
 
 public class JwtCredentialSignerFactory implements CredentialSignerFactory {
 
     @Override
     public String getSupportedFormat() {
-        return Format.JWT_VC;
+        return VCFormat.JWT_VC.getValue();
     }
```
IngridPuppet, line 28: Same remark here. We could update the return type in the interface. WDYT?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -122,18 +125,27 @@ public Object createProtocolEndpoint(KeycloakSession keycloakSession, EventBuild
...
-            naturalPersonScope = newRealm.addClientScope(String.format("%s_%s", OID4VC_PROTOCOL, "natural_person"));
-            naturalPersonScope.setDescription("OID4VCI Scope, that adds properties required for a natural person.");
-            naturalPersonScope.setProtocol(OID4VC_PROTOCOL);
-            naturalPersonScope.addProtocolMapper(builtins.get(SUBJECT_ID_MAPPER));
-            naturalPersonScope.addProtocolMapper(builtins.get(EMAIL_MAPPER));
-            naturalPersonScope.addProtocolMapper(builtins.get(FIRST_NAME_MAPPER));
-            naturalPersonScope.addProtocolMapper(builtins.get(LAST_NAME_MAPPER));
-            addClientScopeDefaults(naturalPersonScope);
-            newRealm.addDefaultClientScope(naturalPersonScope, true);
+        String credentialType = "natural_person";
```
IngridPuppet: Could you turn this into a class constant?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -122,18 +125,27 @@ public Object createProtocolEndpoint(KeycloakSession keycloakSession, EventBuild
...
+
+        // [TODO #44875] Proof Type null is not supported for format ldp_vc
+        // https://github.com/keycloak/keycloak/issues/44875
+        List<VCFormat> supportedFormats = List.of(JWT_VC, SD_JWT_VC);
+
+        for (VCFormat format : supportedFormats) {
+            String scopeName = credentialType + format.getSuffix();
+            ClientScopeModel clientScope = KeycloakModelUtils.getClientScopeByName(newRealm, scopeName);
+            if (clientScope == null) {
+                LOGGER.debug("Add natural person scope");
```
IngridPuppet: Add information about the format to the debug message?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -122,18 +125,27 @@ public Object createProtocolEndpoint(KeycloakSession keycloakSession, EventBuild
...
+                clientScope = newRealm.addClientScope(String.format("%s_%s", OID4VC_PROTOCOL, scopeName));
+                clientScope.setDescription("OID4VCI Scope, that adds properties required for a natural person.");
```
IngridPuppet: Add information about the specific format to the description?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -144,25 +156,36 @@ public void setupClientDefaults(ClientRepresentation rep, ClientModel newClient)
 
     @Override
     public void addClientScopeDefaults(ClientScopeRepresentation clientScope) {
+        String scopeName = clientScope.getName();
+
+        clientScope.getAttributes().computeIfAbsent(FORMAT, k -> VCFormat.fromScopeName(scopeName).getValue());
+        VCFormat format = VCFormat.fromValue(clientScope.getAttributes().get(FORMAT));
+
+        int idx = scopeName.lastIndexOf(format.getSuffix());
+        String credentialType = idx > 0 ? scopeName.substring(0, idx) : scopeName;
```
IngridPuppet, line 164: I suggest to inject the FORMAT attribute into the client scope in `createDefaultClientScopes` directly, so as to avoid this string parsing logic. WDYT?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -144,25 +156,36 @@ public void setupClientDefaults(ClientRepresentation rep, ClientModel newClient)
...
         clientScope.getAttributes().putIfAbsent(INCLUDE_IN_METADATA, "true");
-        clientScope.getAttributes().computeIfAbsent(CONFIGURATION_ID, k -> clientScope.getName());
-        clientScope.getAttributes().computeIfAbsent(CREDENTIAL_IDENTIFIER, k -> clientScope.getName());
-        clientScope.getAttributes().computeIfAbsent(TYPES, k -> clientScope.getName());
-        clientScope.getAttributes().computeIfAbsent(CONTEXTS, k -> clientScope.getName());
-        clientScope.getAttributes().computeIfAbsent(VCT, k -> clientScope.getName());
-        clientScope.getAttributes().computeIfAbsent(FORMAT, k -> FORMAT_DEFAULT);
+        clientScope.getAttributes().computeIfAbsent(CONFIGURATION_ID, k -> scopeName);
+        clientScope.getAttributes().computeIfAbsent(CREDENTIAL_IDENTIFIER, k -> credentialType);
+        clientScope.getAttributes().computeIfAbsent(FORMAT, k -> format.getValue());
```
IngridPuppet: Setting the FORMAT here again looks like a double operation.

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -144,25 +156,36 @@ public void setupClientDefaults(ClientRepresentation rep, ClientModel newClient)
...
+        clientScope.getAttributes().computeIfAbsent(TYPES, k -> credentialType);
+        clientScope.getAttributes().computeIfAbsent(CONTEXTS, k -> credentialType);
+        clientScope.getAttributes().computeIfAbsent(VCT, k -> credentialType);
```
IngridPuppet, line 176: Strictly speaking, setting the VCT should also go under `if (format == SD_JWT_VC) {`.

testsuite/integration-arquillian/servers/auth-server/services/testsuite-providers/src/main/java/org/keycloak/testsuite/rest/TestingResourceProvider.java
IngridPuppet, line 1: Judging from the updates under the testsuite package, it does not look like a test path checks if the natural_person VC is properly loaded for the JWT_VC format. Do you mind adding one?

IngridPuppet (commented):
Hello @tdiesler - I reviewed the PR and left a series of minor comments. Please could you check?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: By (our) convention, ... 

* credential_configuration_id == credential.scope
* credential.scope = credential.type + format_suffix

when there are more than one supported formats for a given credential.type 

There is actually no requirement that the credential_configuration_id or scope indicate the format in any way (i.e. the user is free to name the them as they like) 

Whenever a new client scope is is created, we call `addClientScopeDefaults` and fill in default values for potentially missing attributes. The logic that derives the format from a scope name is there for others to follow the same convention (if they like)

```
        clientScope.getAttributes().computeIfAbsent(FORMAT, k -> getVCFormatFromScope(scopeName).getValue());
        VCFormat format = VCFormat.fromValue(clientScope.getAttributes().get(FORMAT));
```

Unless I miss-understand, I believe its ok as is. Please reopen, if not resolved.

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: yes

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: I was wondering about that property - it shows up in the well-known config and should not be there

Also, I noticed that it is being used in other places as synonym for the credential type (with incorrect default to scope) - not only for SD credentials

```
                    assertEquals("The jwt_vc-credential should offer vct",
                            verifiableCredentialType,
                            jwtVcConfig.getVct());
```

I renamed it to VC_TYPE

testsuite/integration-arquillian/servers/auth-server/services/testsuite-providers/src/main/java/org/keycloak/testsuite/rest/TestingResourceProvider.java
tdiesler: done

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
IngridPuppet: The property is only meant for SD-JWT credentials.

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-6

It's indeed not intended it leaks to other credential types.

IngridPuppet (approved):
Thank you for the update.

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: Created this follow issue #45552

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -126,18 +131,25 @@ public Object createProtocolEndpoint(KeycloakSession keycloakSession, EventBuild
...
+                LOGGER.debugf("Add client scope: %s", scopeName);
+                clientScope = newRealm.addClientScope(String.format("%s_%s", OID4VC_PROTOCOL, scopeName));
+                clientScope.setDescription("OID4VCI credential scope that represents a natural person in format: " + format.getValue());
+                clientScope.setProtocol(OID4VC_PROTOCOL);
+                clientScope.addProtocolMapper(builtins.get(SUBJECT_ID_MAPPER));
+                clientScope.addProtocolMapper(builtins.get(EMAIL_MAPPER));
+                clientScope.addProtocolMapper(builtins.get(FIRST_NAME_MAPPER));
+                clientScope.addProtocolMapper(builtins.get(LAST_NAME_MAPPER));
+                addClientScopeDefaults(clientScope);
+                newRealm.addDefaultClientScope(clientScope, false);
```
mposolda, line 146: Since this is client scope for example purposes, is it please possible to make sure that the client scope is not added as default client-scope to newly created clients? AFAIK just removing this line might do the trick.

core/src/main/java/org/keycloak/VCFormat.java
```diff
@@ -15,34 +15,55 @@
...
      * LD-Credentials {@see https://www.w3.org/TR/vc-data-model/}
      */
-    public static final String LDP_VC = "ldp_vc";
+    LDP_VC("ldp_vc"),
 
     /**
      * JWT-Credentials {@see https://identity.foundation/jwt-vc-presentation-profile/}
      */
-    public static final String JWT_VC = "jwt_vc";
+    JWT_VC("jwt_vc"),
```
mposolda: Just for the reference, there is related issue about `jwt_vc` not being format supported by the specification #45486 . Issue already WiP by adorsys team. There would be probably some conflicts with your PR and that PR. IMHO my preference is included the other PR first and then update your PR accordingly with correct formats according to specs. Is it ok to you?

mposolda (commented):
@tdiesler Thanks, added some comments inline.

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: I checked that these client scopes exist, but are not part of the master.account client.
I also created a new client and checked that they are not added to that new client either.
Instead they can be added explicitly with "Add client scope"

It is the boolean param that drives this

```
    /**
     * Adds given client scope among default/optional client scopes of this realm.
     * The scope will be assigned to each new client.
     * @param clientScope to be added
     * @param defaultScope if {@code true} the scope will be added among default client scopes,
     * if {@code false} it will be added among optional client scopes
     */
    void addDefaultClientScope(ClientScopeModel clientScope, boolean defaultScope);
```

Should work as wanted already.

core/src/main/java/org/keycloak/VCFormat.java
tdiesler: Oh dear - you are totally right.
I updated the enum value and links to the spec for all

core/src/main/java/org/keycloak/VCFormat.java
```diff
@@ -0,0 +1,69 @@
...
+     * JWT-Credentials {@see https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-w3c-verifiable-credentials}
+     */
+    JWT_VC("jwt_vc_json"),
+
+    /**
+     * SD-JWT-Credentials {@see https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-ietf-sd-jwt-vc}
+     */
+    SD_JWT_VC("dc+sd-jwt");
+
+    private final String value;
```
Captain-P-Goldfish: Please don't!!!!

such an enum existed previously and was removed on purpose. Introducing an enum here limits the number of actually supported formats and breaks extensibility. Please do not use enums.

Captain-P-Goldfish (changes requested):

core/src/main/java/org/keycloak/VCFormat.java
tdiesler: No problem - that only changes about every file :-)

tdiesler:
Successfully rebased

tdiesler:
Includes #45913

tdiesler:
This needs to be split up to find why stuff is leaking from one test realm to another

PR #44844, opened 2025-12-11
[OID4VCI] Remove notification ID handling and obsolete endpoint

### LINKED ISSUE #44802: Notification_id returned from the credential response
### Before reporting an issue

### Area

oid4vc

### Describe the bug

There is claim `notification_id` returned from the successful credential-response. However Keycloak doesn't have any support for further receiving of notifications (no support for `notification_endpoint` described in the [OID4VCI specification](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-notification-endpoint) and this claim is not used anywhere for anything .

Returning `notification_id`, which is not supported anywhere seems to be confusing. Would be probably good to remove the `notification_id` from the credential response (preferred way ATM) or add proper support for the `notification_id` (would be probably more work and increases scope of OID4VCI feature and hence not preferred IMO).

### Version

nightly (from 2025-12-10)

### Regression

### Expected behavior

Not returning `notification_id` from the credential-response. It is confusing to return it considering that Keycloak doesn't support it further.

### Actual behavior

The `notification_id` is returned

### How to Reproduce?

See details

### Anything else?

_No response_

Awambeng:
This PR removes the notification ID handling from the OID4VCIssuerEndpoint and all associated classes.
It also removes the related obsolete endpoint.

Closes #44802


mposolda (approved):
@Awambeng Thanks!

PR #44866, opened 2025-12-11
Allow running scheduled workflows

### LINKED ISSUE #44865: Allow running scheduled workflows
### Description

The use case is about running a workflow for existing realm resources (e.g.: users) in a realm. Not necessarily when an event is fired.

Such workflows can be considered as "preventive" workflows, responsible for keeping a realm in pace with security policies and recurring maintenance tasks.
One of the key concrete use cases here is how we allow administrators to run workflows to track inactive/dormant users in a realm, without necessarily having to wait to them to authenticate in order to be able to associated them with a workflow. Some users might not even authenticate after the workflow is created.

Today we have a activation mechanism, a.k.a the election mechanism, that checks of elegible resources and run a workflow for them. This operation is triggered manually via Workflow API and it will query the database, elect resources, and create a workflow execution for each resource. This operation does not make much sense and it does not scale well. It does not make sense because the ultimate goal of a "preventive" workflow is to continuously track the realm state for updates, and eventually perform the necessary steps to update the realm resources. The ultimate goal being to eventually reach a point where your realm is in a state where all your security policies and constraints are being enforced and you don't have more resources to go through a workflow.

That said, I think we should add a scheduling mechanism in Workflows. It can be, for now, as simple as defining the following settings in the workflow definition:

```
 schedule:
   after: 60d
   batch-size: 1000
```

By doing this, we are going to run a workflow every 60 days where the realm resources are processed in batches of 1000.

This new configuration should also work if you want a workflow to listen to events and be scheduled. Or you can still have a workflow that only listen for events:

```
name: Track inactive users
on: user_authenticated
schedule:
   after: 60d
   batch-size: 1000 
```

By defining this, for instance, you are saying "This workflow should listen for these events (on) as well as run every 60d.

### Value Proposition

Support use cases where workflows should run on existing realm resources without the need for listent for events. Also, allow the creation of continuous workflows.

### Goals

Support use cases where workflows should run on existing realm resources without the need for listent for events. Also, allow the creation of continuous workflows.

### Non-Goals

N/A

### Discussion

_No response_

### Notes

_No response_

pedroigor:
Closes #44865


tests/base/src/test/java/org/keycloak/tests/workflow/GroupMembershipJoinWorkflowTest.java
sguilhen, line 1: This empty file is odd

sguilhen (changes requested):
@pedroigor LGTM, thanks! Only thing I've noticed is the addition of an empty file - probably a mistake.

tests/base/src/test/java/org/keycloak/tests/workflow/GroupMembershipJoinWorkflowTest.java
pedroigor: Removed. Leftover after resolving conflicts when rebasing with the testsuite refactoring.

sguilhen (approved):

stianst (approved):

ssilvert:
See https://github.com/keycloak/keycloak/issues/44865#issuecomment-3710980374

PR #44871, opened 2025-12-12
[OID4VCI]: Use Keycloak time utility for OID4VC related timestamps

### LINKED ISSUE #44235: Use the Keycloak Time.currentTime utility instead of Instant.
### Description

There are some utilities in the OID4VC related code, which uses `Instant.now()` for retrieve current timestamp. Can we try to use `Time.currentTime()` instead of it to make it consistent with other Keycloak code?

Some examples are `TimeClaimVerifier`, `JwtCNonceHandler`, `OID4VCIssuedTimeClaimMapper` and few more (especially some unit tests).

This is used by other Keycloak code. Among other things, it also allows to test timestamps as there is `Time.setTimeOffset`, which allows to "move time forward" or "move time backward", which is useful to simular various scenarios related to timeouts and expirations. The main arquillian testsuite has method `setTimeOffset`, which is automatically "restarted" at the end of each test. The tests in the `core` needs to manually reset time offset when it is used to test something.

forkimenjeckayang:
## Summary
- Use Keycloak’s shared time helper for OID4VC and related tests
- Align timestamp checks with existing offset-based time handling

closes #44235


mposolda (approved):
@forkimenjeckayang Thanks!

PR #44924, opened 2025-12-15
Use a custom serializer to write policy config values as JSON array o…

### LINKED ISSUE #44573: Authorization policy endpoint returns stringified JSON arrays in config fields instead of proper arrays
### Before reporting an issue

### Area

admin/api

### Describe the bug

The authorization policy REST API endpoint returns policy configuration values as stringified JSON arrays instead of native JSON arrays, violating REST API best practices and causing type safety issues for API consumers.
Endpoint in question: `GET /admin/realms/{realm}/clients/{client-id}/authz/resource-server/policy`

### Version

26.4.6

### Regression

### Expected behavior

The `config.clients` field should be returned as a native JSON array:
```json
{
    "id": "140d0d32-10ea-4604-9327-c716eee22389",
    "name": "foo",
    "type": "client",
    "logic": "POSITIVE",
    "decisionStrategy": "UNANIMOUS",
    "config": {
        "clients": ["bc8b367f-2b67-45ec-8843-16877c65293f"]
    }
}
```

### Actual behavior

The `config.clients` field is returned as a stringified JSON array:
```json
{
    "id": "140d0d32-10ea-4604-9327-c716eee22389",
    "name": "foo",
    "type": "client",
    "logic": "POSITIVE",
    "decisionStrategy": "UNANIMOUS",
    "config": {
        "clients": "[\"bc8b367f-2b67-45ec-8843-16877c65293f\"]"
    }
}
```

### How to Reproduce?

1. Have Admin Permissions enabled
2. Create a client
3. Create a client-type policy with one or more clients configured
4. Call `GET /admin/realms/{realm}/clients/{client-id}/authz/resource-server/policy?first=0&max=11&permission=false`
5. Observe that `config.clients` is a string containing JSON rather than a proper array

### Anything else?

_No response_

sguilhen:
…bjects instead of a stringfied array

Closes #44573

### DIFF
--- a/core/src/main/java/org/keycloak/representations/idm/authorization/PolicyConfigDeserializer.java
+++ b/core/src/main/java/org/keycloak/representations/idm/authorization/PolicyConfigDeserializer.java
@@ -1,0 +1,62 @@
+package org.keycloak.representations.idm.authorization;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+
+public class PolicyConfigDeserializer extends StdDeserializer<Map<String,String>> {
+
+    public PolicyConfigDeserializer() {
+        this(null);
+    }
+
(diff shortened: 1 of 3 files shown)


vramik (approved):

pedroigor (changes requested):
The change makes sense, and it is better to return the field using a JSON array format.

However, we are changing the contract and potentially breaking existing clients relying on the string format.

I'm OK to change the format if we add release/upgrade notes and make this change in a major release. Probably 27? 

Ideally, managing client, role, group, users, etc, policies should be done through their respective endpoints so that you have a payload that is specific for that policy type. Instead of a generic representation like the `PolicyRepresentation`.

PR #44928, opened 2025-12-16
Improve grammar on keys and remove unused

### LINKED ISSUE #44927: Improve gramma on Tooltips
### Description

As suggested from Weblate comments.
https://hosted.weblate.org/translate/keycloak/-/en/?q=+context%3A%3DclientUpdaterTrustedHostsTooltip+OR+context%3A%3DlogoutConfirmationHelp+&sort_by=-priority%2Cposition&checksum=

Key: logoutConfirmationHelp
Present text: 
`After logout of the user (OIDC RP-Initiated logout), there will be additional confirmation info page displayed to the user with the text like 'You are logged out' before redirecting a user to a post-logout landing page. On this info page, user needs to confirm that he want to be redirected to the post-logout landing page.`
New text:
`After the user logs out (OIDC RP-Initiated Logout), an additional confirmation page will be displayed showing a message such as 'You are logged out' before redirecting the user to the post-logout landing page. On this page, the user must confirm that they want to be redirected to the post-logout landing page.`

Key: clientUpdaterTrustedHostsTooltip
Present text: 
`List of Hosts, which are trusted. If that client registration or update request comes from the host/domain specified in this configuration, the condition evaluates to true. You can use hostnames or IP addresses. If you use star at the beginning (for example '*.example.com'), the whole domain example.com is trusted.`
New text: 
`List of hosts that are trusted. If a client registration or update request comes from a host or domain specified in this configuration, the condition evaluates to true. You can use hostnames or IP addresses. If you use an asterisk at the beginning (for example, '*.example.com'), the entire domain example.com will be trusted.`

robson90:
As it seems the key "clientUpdaterTrustedHostsTooltip" is not used anymore, so I removed it.

I updated the text for key "logoutConfirmationHelp".

@ahus1 please double check, if key "clientUpdaterTrustedHostsTooltip" is really unused.

Improvements came from: https://hosted.weblate.org/user/till.reymann/

Closes #44927


ahus1:
I confirm that the key is no longer used, and the admin UI uses `trusted-hosts.tooltip` instead. 

I'll push one cleanup where I revert the changes to all translation files. Instead, I would like to rely on Weblate to remove the keys - this avoid one more occation for merge conflicts. I know that this will not fix those that are not managed by Weblate, but that's the smaller of the two problems.

robson90:
Alright. Works for me. 
Will keep it in mind for the next changes.

ahus1 (approved):

PR #44931, opened 2025-12-16
[OID4VCI] Use consistent parameter name `nonce` in credential offer preflight endpoint

forkimenjeckayang:
## Summary

Fixes parameter name inconsistency in `OID4VCIssuerEndpoint.getCredentialOfferPreflight` method.

### Issue
The `getCredentialOfferPreflight` method was using `sessionCode` as the path parameter name, while the actual `getCredentialOffer` endpoint uses `nonce`. T

Closes #44742


mposolda (approved):
@forkimenjeckayang Thanks

PR #44932, opened 2025-12-16
Add timeout option for keycloak-admin-client

rmartinc:
Closes #42644

PR that adds the `timeout` to the `ConnectionConfig`. This is later transformed in the `signal` with the abort. The PR adds two little tests to check the timeout is applied at `fetch` level.

### DIFF
--- a/js/libs/keycloak-admin-client/src/resources/agent.ts
+++ b/js/libs/keycloak-admin-client/src/resources/agent.ts
@@ -248,6 +248,9 @@ export class Agent {
         ...requestOptions,
         headers: requestHeaders,
         method,
+        ...(this.#client.timeout
+          ? { signal: AbortSignal.timeout(this.#client.timeout) }
+          : {}),
       });
 
       // now we get the response of the http request
--- a/js/libs/keycloak-admin-client/test/timeout.spec.ts
+++ b/js/libs/keycloak-admin-client/test/timeout.spec.ts
@@ -1,0 +1,53 @@
+import { expect } from "chai";
+import { KeycloakAdminClient } from "../src/client.js";
+import { credentials } from "./constants.js";
+import { Server, createServer } from "node:http";
+
+describe("Timeout", () => {
+  let server: Server;
+
+  before(async () => {
+    server = createServer((req, res) => {
+      res.writeHead(200, { "Content-Type": "text/plain" });
+      setTimeout(() => res.end("Hello, world!\n"), 1500);
+    });
+    server.listen(8888, "localhost");
+  });
+
+  after(async () => {
+    await server[Symbol.asyncDispose]();
+  });
+
+  void it("create without timeout", async () => {
+    const client = new KeycloakAdminClient({
+      baseUrl: "http://localhost:8888",
+    });
+
+    try {
+      await client.auth(credentials);
+    } catch (error) {
+      expect(error).to.be.an("Error");
+      expect((error as Error).message).to.contain("Unexpected token 'H'");
+      return;
+    }
+    expect.fail(null, null, "auth did not fail");
+  });
+
+  void it("create with timeout", async () => {
+    const client = new KeycloakAdminClient({
+      baseUrl: "http://localhost:8888",
+      timeout: 1000,
+    });
+
+    try {
+      await client.auth(credentials);
+    } catch (error) {
+      expect(error).to.be.an("DOMException");
+      expect((error as DOMException).message).to.contain(
+        "The operation was aborted due to timeout",
+      );
+      return;
+    }
+    expect.fail(null, null, "auth did not fail");
+  });
+});
--- a/js/libs/keycloak-admin-client/src/client.ts
+++ b/js/libs/keycloak-admin-client/src/client.ts
@@ -19,6 +19,8 @@ import { WhoAmI } from "./resources/whoAmI.js";
 import { Credentials, getToken } from "./utils/auth.js";
 import { defaultBaseUrl, defaultRealm } from "./utils/constants.js";
 
+export type RequestOptions = Omit<RequestInit, "signal">;
+
 export interface TokenProvider {
   getAccessToken: () => Promise<string | undefined>;
 }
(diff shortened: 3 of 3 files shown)


js/libs/keycloak-admin-client/test/timeout.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { expect } from "chai";
+import { KeycloakAdminClient } from "../src/client.js";
+import { credentials } from "./constants.js";
+import { Server, createServer } from "http";
```
jonkoops: When importing from the Node.js standard library, it is recommended to use the `node:` prefix to avoid conflicts with installed NPM packages.

```suggestion
import { Server, createServer } from "node:http";
```

js/libs/keycloak-admin-client/src/resources/agent.ts
```diff
@@ -248,6 +248,9 @@ export class Agent {
         ...requestOptions,
         headers: requestHeaders,
         method,
+        ...(this.#client.timeout
+          ? { signal: AbortSignal.timeout(this.#client.timeout) }
```
jonkoops, line 252: Note that `AbortSignal.timeout()` is considered 'newly available' to baseline, this means that older browsers might not support it. Looking at the compatibility table, I think this is acceptable and there is no need to wait for 'widely available'.

See:
- https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static
- https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility

js/libs/keycloak-admin-client/test/timeout.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    server = createServer((req, res) => {
+      res.writeHead(200, { "Content-Type": "text/plain" });
+      setTimeout(() => res.end("Hello, world!\n"), 1500);
+    });
+    server.listen(8888, "localhost");
+  });
+
+  after(async () => {
+    server.close();
+  });
```
jonkoops, line 19: Looking at [the documentation](https://nodejs.org/api/net.html#serverclosecallback), it seems that `server.close()` takes a callback as the first argument that is called when the server actually closes. Ideally, the async function only resolves when this callback is triggered, whereas now the function resolves immediately, as there is nothing `await`ed.

It seems that the server implements [`Symbol.asyncDispose`](https://nodejs.org/api/http.html#serversymbolasyncdispose) ([see MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncDispose) for more details), which would allow us to clean up the server using an asynchronous Promise-based API (untested):

 
```suggestion
  after(() => server[Symbol.asyncDispose]());
```

js/libs/keycloak-admin-client/test/timeout.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  void it("create without timeout", async () => {
```
jonkoops, line 21: Is there a particular reason this needs to be `void`ed?

js/libs/keycloak-admin-client/test/timeout.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    const client = new KeycloakAdminClient({
+      baseUrl: "http://localhost:8888",
+    });
+
+    try {
+      await client.auth(credentials);
+    } catch (error) {
+      expect(error).to.be.an("Error");
+      expect((error as Error).message).to.contain("Unexpected token 'H'");
```
jonkoops, line 30: This could likely be a non-error case if the server returned JSON, but I guess this also works for proving the point of the test.

js/libs/keycloak-admin-client/test/timeout.spec.ts
rmartinc: Done!

js/libs/keycloak-admin-client/src/resources/agent.ts
rmartinc: We can add our own implementation if needed, but doing nothing as you suggested.

js/libs/keycloak-admin-client/test/timeout.spec.ts
rmartinc: I have done the following:

```js
  after(async () => {
    await server[Symbol.asyncDispose]();
  });
```

js/libs/keycloak-admin-client/test/timeout.spec.ts
rmartinc: The reason is that lint complains. :smile:
I can remove it, but then I need to not check the commit with lint (the hook at commit time). It happens to me in all these tests.

```
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.eslint[@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises)
```

js/libs/keycloak-admin-client/test/timeout.spec.ts
rmartinc: I have changed a bit the test, to fail if the `auth` executes OK.

js/libs/keycloak-admin-client/test/timeout.spec.ts
jonkoops: That works as well, since it is a Promise the `async` and `await` are redundant, but perhaps it is more legible/extensible, so LGTM.

js/libs/keycloak-admin-client/test/timeout.spec.ts
jonkoops: I see, probably when using `async` the return value of `it()` also becomes a Promise. This seems appropriate to mitigate using `void` 👍

jonkoops (approved):
LGTM!

mposolda (approved):
@rmartinc @jonkoops Thanks for the PR and review!

PR #44948, opened 2025-12-16
Remote Infinispan should return count per client only for the current…

### LINKED ISSUE #44577: Remote Infinispan should return count per client only for the current realm
### Before reporting an issue

### Area

infinispan

### Describe the bug

From https://github.com/keycloak/keycloak/issues/44424#issuecomment-3596034811: 

The query for the external Infinispan is missing the `realmId`. The javadoc does not say it is only the sessions belonging to a realm 😅

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/query/ClientSessionQueries.java#L38

### Version

main

### Regression

### Expected behavior

Only the counts of the clients in the current realm should be returned.

### Actual behavior

It doesn't filter by the realm.

### How to Reproduce?

See query

### Anything else?

_No response_

pruivo:
… realm

Closes #44577


ryanemerson (approved):
LGTM, thanks @pruivo

stianst (approved):

PR #44950, opened 2025-12-16
Create a LocalCacheProvider SPI

ryanemerson:
Closes #42223

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/cache/LocalCacheConfiguration.java
+++ b/server-spi-private/src/main/java/org/keycloak/cache/LocalCacheConfiguration.java
@@ -1,0 +1,72 @@
+package org.keycloak.cache;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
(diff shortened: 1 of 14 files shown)


server-spi-private/src/main/java/org/keycloak/cache/LocalCacheConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+            return this;
+        }
+
+        public Builder<K, V> loader(Function<K, V> loader) {
+            this.loader = loader;
+            return this;
+        }
+
+        public LocalCacheConfiguration<K, V> build() {
+            Objects.requireNonNull(name, "A cache name must be configured");
```
pruivo: I would do this check in the constructor, just in case the builder is not used.

```java
    public LocalCacheConfiguration {
        Objects.requireNonNull(name, "A cache name must be configured");
    }
```

ryanemerson: Good point, updated.

pruivo (approved):

stianst (approved):

PR #44954, opened 2025-12-16
[OID4VCI] Mandatory claims are not enforced for OID4VCI

### LINKED ISSUE #44796: Claims configued by OID4VCI protocol mappers as mandatory are not enforced to be mandatory
### Before reporting an issue

### Area

oid4vc

### Describe the bug

ATM When there is OID4VCI token request sent with `authorization_details` , which contains some "mandatory" claims, Keycloak properly enforces that the claim is really mandatory.

For example in case of request like:
```
[{"type":"openid_credential","credential_configuration_id":"sd-jwt-credential-config-id","locations":["https://localhost:8543/auth/realms/test"],"claims":[{"path":["lastName"],"mandatory":true}]}]
```
When such request is used, but user doesn't have `lastName` (and hence `lastName` claim missing from the credential), then the OID4VCI credential request fails. This is expected IMO.

However when the claim is configured as mandatory by OID4VCI protocol mapper (and subsequently announced as mandatory in the OID4VCI metadata well-known response for the particular credential), then it is not enforced that the claim is present and credential-response is successfully issued even without that mandatory claim -> KO

### Version

nightly (from 2025-12-10)

### Regression

### Expected behavior

The credential-response is issued just if mandatory claims enforced in both `authorization_details` parameter as well as in protocol mappers (and subsequently OID4VC metadata) are present

### Actual behavior

The credential-response is issued even if mandatory claims configured by protocol mappers (and subsequently OID4VC metadata) are not present

### How to Reproduce?

See details

### Anything else?

The related code is `OID4VCIssuerEndpoint.validateRequestedClaimsArePresent` , but checks just the mandatory claims requested in the `authorization_details`, but not by the config.

It worth to mention that there are multiple other issues with the "mandatory" claims, which maybe worth to be fixed.

- **Issue 2:** Mandatory claims seems to be enforced just when credentialScopeId is the same as name of the `scope` .
The `OID4VCIssuerEndpoint.validateRequestedClaimsArePresent` uses `String claimsKey = AUTHORIZATION_DETAILS_CLAIMS_PREFIX + scope;` . However the `OID4VCAuthorizationDetailsProcessor.buildAuthorizationDetailResponse` uses `String claimsKey = OID4VCIssuerEndpoint.AUTHORIZATION_DETAILS_CLAIMS_PREFIX + credentialConfigurationId;`

- **Issue 3:** - Even claims configured as mandatory are probably not enforced right now. There is `BadRequestException` correctly thrown [here](https://github.com/keycloak/keycloak/blob/921b10ee804224ff60396c0c70af5f97b612ba10/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java#L1472) , but that Exception is caught [here](https://github.com/keycloak/keycloak/blob/921b10ee804224ff60396c0c70af5f97b612ba10/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java#L1479) causing the flow to successfully continue even without mandatory claims (Just some errors in the server.log are added).

- **Issue 4:** Assuming that there is subsequent OAuth2 login request for `authorization_code` grant (Regular login request, not any use of OID4VCI expected), which happens in the same browser session where some other OAuth2 login request with some OID4VCI `authorization_details` previously happened. Then the `authorization_details` is still returned in the token-response and it seems that credential-offer is also created (not 100% sure. To be doublechecked...). 
The subsequent login requests should not return `authorization_details` and should not create any credential offers.
ATM It seems that saving authorization_details as a note to the userSession may not be the best approach. Might need to double-check, but probably worth to store them to the token instead (maybe both access-token and refresh-token) instead of storing as a note inside userSession. I've created dedicated issue for this one: https://github.com/keycloak/keycloak/issues/44961

mposolda:
closes #44796

Summary of the changes:

- PR fixes the issues 1-3 mentioned in the description of #44796 . I've created separate issue #44961 for the "Issue 4" from the #44796 description and hope to address that as a follow-up

- When the claim is configured as "Mandatory" either in the protocol-mapper or in the `authorization_details` parameter, it is enforced that the claim is really present in the credential response.

- The check for mandatory claims was already implemented in `OID4VCIssuerEndpoint.validateRequestedClaimsArePresent` , however it did not work (due the lots of nested try/catch blocks, which caused thrown exception to be incorrectly caught and not propagated to the caller). Added automated tests to make sure that this is not broken anymore

- It works for both `sd-jwt` credentials as well as for `jwt_vc` verifiable credentials. In case of `jwt_vc` verifiable credentials, there was one tricky thing as the `authorization_details` mandatory claim has the full path (including `credentialSubject`) but the claims returned by protocol mappers method `setClaimsForSubject` returns claims in the "flat map" without `credentialSubject` . I've introduced another method `setClaimsWithMetadataPrefix` to the `OID4VCMapper`, which creates the map with whole "metadata attribute path" (in case of `jwt_vc` credential, the returned map contains single `credentialSubject` claim with the other claims as sub-claims of it). That works with checking the mandatory claim for `jwt_vc` credential. The original map returned by `setClaimsForSubject` is still kept and used in the subsequent steps for credential building (I did not want to refactor that part as it might be quite a lot of changes, so instead introduced that new method `setClaimsWithMetadataPrefix` ).

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -25,6 +25,7 @@ import java.security.PublicKey;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -1369,8 +1370,13 @@ public class OID4VCIssuerEndpoint {
         Map<String, Object> subjectClaims = new HashMap<>();
         protocolMappers.forEach(mapper -> mapper.setClaim(subjectClaims, authResult.session()));
 
+        Map<String, Object> subjectClaimsWithMetadataPrefix = new HashMap<>();
+        protocolMappers
+                .forEach(mapper -> mapper.setClaimWithMetadataPrefix(subjectClaims, subjectClaimsWithMetadataPrefix));
+
         // Validate that requested claims from authorization_details are present
-        validateRequestedClaimsArePresent(subjectClaims, authResult.session(), credentialConfig.getScope());
+        String credentialConfigId = credentialConfig.getId();
+        validateRequestedClaimsArePresent(subjectClaimsWithMetadataPrefix, credentialConfig, authResult.session(), credentialConfigId);
 
         // Include all available claims
         subjectClaims.forEach((key, value) -> vc.getCredentialSubject().setClaims(key, value));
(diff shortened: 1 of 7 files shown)


mposolda:
@IngridPuppet @Captain-P-Goldfish @Awambeng @forkimenjeckayang @Ogenbertrand @wistefan @pskopek Anyone available to review the PR please?

tnorimat (approved):
LGTM.

mposolda:
@tnorimat Thanks!

PR #44965, opened 2025-12-17
Disable state transfer for session caches when persistent sessions are enabled

### LINKED ISSUE #44518: Disable state transfer for session caches when persistent sessions are enabled
### Description

When persistent sessions are enabled, the data about sessions is available in the database. As state transfer puts a high load on the cluster when a node starts up or shuts down, this should be disabled by default.

### Value Proposition

Have less state transfer on new nodes joining and leaving the cluster. 

Note that this affects all the nodes, not only the joining or leaving node. 

The tradeoff is then higher database usage after a node restart. 

### Goals

Lower latencies on node restarts. Allowing for a larger cache of sessions eventually.  

### Non-Goals

?

### Discussion

_No response_

### Notes

_No response_

ryanemerson:
Closes #44518

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
+++ b/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
@@ -253,6 +253,8 @@ public final class CacheConfigurator {
             builder.clustering().hash().numOwners(1);
             if (sessionCaches.contains(name)) {
                 configureSessionExpirationReaper(builder);
+                // Disable state-transfer to reduce the overhead of new nodes joining
+                builder.clustering().stateTransfer().fetchInMemoryState(false);
             }
         }
     }
@@ -304,6 +306,8 @@ public final class CacheConfigurator {
                 builder.clustering().hash().numOwners(1);
             }
             configureSessionExpirationReaper(builder);
+            // Disable state-transfer to reduce the overhead of new nodes joining
+            builder.clustering().stateTransfer().fetchInMemoryState(false);
         }
     }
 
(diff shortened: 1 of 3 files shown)


model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
```diff
@@ -507,6 +508,12 @@ public static ConfigurationBuilder getCacheConfiguration(String cacheName, boole
             case OFFLINE_USER_SESSION_CACHE_NAME:
                 if (clustered) {
                     builder.clustering().cacheMode(CacheMode.DIST_SYNC).hash().numOwners(1);
+
```
pruivo: I would disable the state transfer in the `configureSessionsCachesForVolatileSessions` and `configureSessionsCachesForPersistentSessions` instead. 

After these lines:
https://github.com/keycloak/keycloak/blob/ca617d971156ea3d457b6bbfacf3191030a003c5/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java#L255
https://github.com/keycloak/keycloak/blob/ca617d971156ea3d457b6bbfacf3191030a003c5/model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java#L306

ryanemerson: Thanks, that's much nicer and I hadn't concerned the offline caches when !persistent-user-sessions.

model/infinispan/src/main/java/org/keycloak/spi/infinispan/impl/embedded/CacheConfigurator.java
```diff
@@ -302,6 +305,8 @@ public static void configureSessionsCachesForVolatileSessions(Config.Scope keycl
             ) {
                 logger.infof("Setting a memory limit implies to have exactly one owner. Setting num_owners=1 to avoid data loss.", name);
                 builder.clustering().hash().numOwners(1);
+                // Disable state-transfer to reduce the overhead of new nodes joining
+                builder.clustering().stateTransfer().fetchInMemoryState(false);
```
pruivo: @ryanemerson IMO, you can move this outside the `if` block. Even when a shared store is enabled, the join/leaver can fetch the data from there.

ryanemerson: Yeah definitely, that was a mistake, fixed!

pruivo (dismissed):

ryanemerson:
@pruivo Spotless failed as I forgot to remove an unused import, can you please re-approve. Thanks

ahus1 (approved):

PR #44968, opened 2025-12-17
Validate client session timeout and lifetime settings on realm settings edit

### LINKED ISSUE #44910: Validate client session timeout and lifetime settings on realm settings edit
### Description

- Follows up on #41019 

The Client Session Idle and Client Session Max values configurable under Realm Settings Sessions must never exceed the effective realm SSO session settings (SSO Session Idle / Max, and Remember Me settings if enabled).

Since client sessions are always bounded by the user session (SSO), configuring client session values longer than the realm session settings is misleading.
Currently,  no validation exists when editing a client,  when these client session defaults are edited at the realm level.

**Motivation** 

Adding validation on this screen:

- Prevents misleading configurations
- Makes the dependency between realm SSO sessions and client sessions explicit
- Aligns behavior with existing validation on the Client Settings page

**Details**

Add validation when editing Client Session Idle and Client Session Max under Realm Settings → Sessions.

The configured values must not exceed:

- SSO Session Idle
- SSO Session Max
- SSO Session Idle Remember Me / Max Remember Me, if Remember Me is enabled

If a client session value exceeds the effective realm session value, show a validation error explaining that:

- Client sessions are limited by realm SSO sessions
- Validation should occur only when saving the realm settings

Aligns behavior with existing validation on the Client Settings page.

ruchikajha95:
Validate client session timeout and lifetime settings on realm settings edit  and testcase for the same.

Closes #44910

### DIFF
--- a/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
+++ b/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
@@ -900,6 +900,7 @@ public class DefaultExportImportManager implements ExportImportManager {
 
         updateCibaSettings(rep, realm);
         updateParSettings(rep, realm);
+        validateClientAndRealmTimeouts(realm);
         session.clientPolicy().updateRealmModelFromRepresentation(realm, rep);
 
         if (rep.getSmtpServer() != null) {
@@ -1730,6 +1731,26 @@ public class DefaultExportImportManager implements ExportImportManager {
                         provider.addMember(orgModel, m);
                     }
                 }
+            }
+        }
+    }
+
+    private void validateClientAndRealmTimeouts(RealmModel realm) {
+        if (realm.isRememberMe()) {
+            if (realm.getClientSessionIdleTimeout() > Math.max(realm.getSsoSessionIdleTimeout(), realm.getSsoSessionIdleTimeoutRememberMe())) {
+                throw new ModelException("Client session idle timeout cannot exceed realm SSO session idle timeout and RememberMe idle timeout.");
+            }
+
+            if (realm.getClientSessionMaxLifespan() > Math.max(realm.getSsoSessionMaxLifespan(), realm.getSsoSessionMaxLifespanRememberMe())) {
+                throw new ModelException("Client session max lifespan cannot exceed realm SSO session max lifespan and RememberMe Max span.");
+            }
+        } else {
+            if (realm.getClientSessionIdleTimeout() > realm.getSsoSessionIdleTimeout()) {
+                throw new ModelException("Client Session Idle Timeout cannot be greater than Realm SSO Idle Timeout.");
+            }
+
+            if (realm.getClientSessionMaxLifespan() > realm.getSsoSessionMaxLifespan()) {
+                throw new ModelException("Client session max lifespan cannot exceed realm SSO session max lifespan.");
             }
         }
     }
--- a/tests/base/src/test/java/org/keycloak/tests/admin/realm/RealmUpdateTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/admin/realm/RealmUpdateTest.java
@@ -288,4 +288,63 @@ public class RealmUpdateTest extends AbstractRealmTest {
         assertEquals(expected.isAdminEventsEnabled(), actual.isAdminEventsEnabled());
         assertEquals(expected.isAdminEventsDetailsEnabled(), actual.isAdminEventsDetailsEnabled());
     }
+
+    @Test
+    public void testRealmClientSessionTimeoutValidation() {
+        RealmRepresentation rep = managedRealm.admin().toRepresentation();
+        // Remember-Me Disabled
+        rep.setRememberMe(false);
+        rep.setSsoSessionIdleTimeout(300);
+        rep.setSsoSessionMaxLifespan(600);
+
+        // Invalid: client idle > realm idle
+        rep.setClientSessionIdleTimeout(400);
+        rep.setClientSessionMaxLifespan(500);
+
+        try {
+            managedRealm.admin().update(rep);
+            Assertions.fail("Expected validation error for client idle timeout");
+        } catch (Exception e) {
+            assertEquals("HTTP 400 Bad Request", e.getMessage());
+        }
+
+        // Fix idle, break max lifespan
+        rep.setClientSessionIdleTimeout(200);
+        rep.setClientSessionMaxLifespan(700);
+
+        try {
+            managedRealm.admin().update(rep);
+            Assertions.fail("Expected validation error for client max lifespan");
+        } catch (Exception e) {
+            assertEquals("HTTP 400 Bad Request", e.getMessage());
+        }
+        // Remember-Me Enabled
+        rep = managedRealm.admin().toRepresentation();
+        rep.setRememberMe(true);
+        rep.setSsoSessionIdleTimeout(300);
+        rep.setSsoSessionIdleTimeoutRememberMe(500);
+        rep.setSsoSessionMaxLifespan(600);
+        rep.setSsoSessionMaxLifespanRememberMe(900);
+
+        // Invalid: exceeds allowed remember-me idle
+        rep.setClientSessionIdleTimeout(550);
+        rep.setClientSessionMaxLifespan(800);
+
+        try {
+            managedRealm.admin().update(rep);
+            Assertions.fail("Expected validation error for remember-me idle timeout");
+        } catch (Exception e) {
+            assertEquals("HTTP 400 Bad Request", e.getMessage());
+        }
+        // Fix idle, break max lifespan
+        rep.setClientSessionIdleTimeout(300);
+        rep.setClientSessionMaxLifespan(950);
+
+        try {
+            managedRealm.admin().update(rep);
+            Assertions.fail("Expected validation error for remember-me max lifespan");
+        } catch (Exception e) {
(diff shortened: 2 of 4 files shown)


model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
...
+
+        int realmRememberIdle = rep.getSsoSessionIdleTimeoutRememberMe() != null
+                ? rep.getSsoSessionIdleTimeoutRememberMe()
+                : realm.getSsoSessionIdleTimeoutRememberMe();
+
+        int realmRememberMax = rep.getSsoSessionMaxLifespanRememberMe() != null
+                ? rep.getSsoSessionMaxLifespanRememberMe()
+                : realm.getSsoSessionMaxLifespanRememberMe();
+
+        boolean rememberMeEnabled = rep.isRememberMe() != null
```
ryanemerson: We can simplify the initialisation of all these variables by using `Objects.requireNonNullElse`, for example:

```
int realmIdle = Objects.requireNonNullElse(rep.getSsoSessionIdleTimeout(), realm.getSsoSessionIdleTimeout());
```

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
         if (rep.getOfflineSessionMaxLifespanEnabled() != null) realm.setOfflineSessionMaxLifespanEnabled(rep.getOfflineSessionMaxLifespanEnabled());
         if (rep.getOfflineSessionMaxLifespan() != null)
             realm.setOfflineSessionMaxLifespan(rep.getOfflineSessionMaxLifespan());
+        // ===== REALM CLIENT SESSION VALIDATION =====
+         // Values coming from UI (new values)
```
ryanemerson: I would remove these comments as they are implied.

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
...
+        Integer clientIdle = rep.getClientSessionIdleTimeout();
+        Integer clientMax  = rep.getClientSessionMaxLifespan();
+
+        // Resolve realm SSO values (new OR existing)
```
ryanemerson: Can be removed.

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
...
+                : realm.isRememberMe();
+        if(!rememberMeEnabled) {
+           if(clientIdle != null && clientIdle > realmIdle) {
+               throw new ModelException("Client Session Idle Timeout cannot be greater than Realm SSO Idle Timeout.");
+           }
+            if (clientMax != null && clientMax > realmMax) {
+                throw new ModelException("Client session max lifespan cannot exceed realm SSO session max lifespan.");
+            }
+        }
+        else {
```
ryanemerson: Formatting should be updated so that the `}` and `else {` are on the same line

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
...
+            int allowedIdle = Math.max(realmIdle, realmRememberIdle);
+            int allowedMax  = Math.max(realmMax, realmRememberMax);
+            if (clientIdle != null && clientIdle > allowedIdle) {
+                throw new ModelException("Client session idle timeout cannot exceed realm SSO or Remember Me idle timeout."
+                );
+            }
+            if (clientMax != null && clientMax > allowedMax) {
+                throw new ModelException("Client session max lifespan cannot exceed realm SSO or Remember Me max lifespan."
+                );
```
ryanemerson: `);` should just be included in the previous line

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
...
+            }
+        }
+        else {
+            int allowedIdle = Math.max(realmIdle, realmRememberIdle);
+            int allowedMax  = Math.max(realmMax, realmRememberMax);
+            if (clientIdle != null && clientIdle > allowedIdle) {
+                throw new ModelException("Client session idle timeout cannot exceed realm SSO or Remember Me idle timeout."
+                );
+            }
+            if (clientMax != null && clientMax > allowedMax) {
```
ryanemerson: This logic is repeated four times, we can reduce this repetition my making use of a method that we then call for each validation:

```java

    private void validateClientTimeout(Integer clientVal, Integer realmVal, String message) {
        if (clientVal != null && clientVal > realmVal) {
            throw new ModelException(message);
        }
    }
```

we can then call this each time e.g.

```java
validateClientTimeout(
  rep.getClientSessionIdleTimeout(),
  Math.max(realmIdle, realmRememberIdle),
  "Client session idle timeout cannot exceed realm SSO or Remember Me idle timeout."
);
```

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -858,6 +858,51 @@ public void updateRealm(RealmRepresentation rep, RealmModel realm) {
...
+
+        int realmMax = rep.getSsoSessionMaxLifespan() != null
+                ? rep.getSsoSessionMaxLifespan()
+                : realm.getSsoSessionMaxLifespan();
+
+        int realmRememberIdle = rep.getSsoSessionIdleTimeoutRememberMe() != null
+                ? rep.getSsoSessionIdleTimeoutRememberMe()
+                : realm.getSsoSessionIdleTimeoutRememberMe();
+
+        int realmRememberMax = rep.getSsoSessionMaxLifespanRememberMe() != null
```
ryanemerson: `realmRemberIdle` and `realmRememberMax` are only used in the else branch of the if statement, so we can define them there in order to limit there visibility to that scope.

tests/base/src/test/java/org/keycloak/tests/admin/realm/RealmUpdateTest.java
```diff
@@ -288,4 +288,62 @@ private void checkRealmEventsConfigRepresentation(RealmEventsConfigRepresentatio
         assertEquals(expected.isAdminEventsEnabled(), actual.isAdminEventsEnabled());
         assertEquals(expected.isAdminEventsDetailsEnabled(), actual.isAdminEventsDetailsEnabled());
     }
+    @Test
```
ryanemerson, line 292: We should add a new line here to separate this test from the previous one.

ryanemerson (changes requested):

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
```diff
@@ -1733,4 +1735,38 @@ private void importOrganizations(RealmRepresentation rep, RealmModel newRealm) {
             }
         }
     }
+
+    private void validateClientAndRealmTimeouts(RealmModel realm) {
+        if (realm.isRememberMe()) {
+            validateClientTimeout(
+                  realm.getClientSessionIdleTimeout(),
+                  Math.min(realm.getSsoSessionIdleTimeout(), realm.getSsoSessionIdleTimeoutRememberMe()),
```
ryanemerson: @ruchikajha95 I changed this from `Math.min` to `Math.max` to reflect the error message, which indicates that the client timeout cannot be greater than either of these SSO values not just the larger of the two. Is this what you originally intended?

ryanemerson (changes requested):
I have pushed a couple of commits to simplify things, please see the commit message for an explanation.

We also still need to update `docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc` to reflect that we now validate the client settings.

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
ryanemerson: I don't believe these checks are necessary after all, as we always set the realm value to equal that of the `rep` one if it is not null, so if we perform the validation after this initialisation is complete, we can simplify the code and just use the `realm` value.

model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultExportImportManager.java
ryanemerson: Logic reverted back to Max and Exception message updated to reflect AND logic.

ryanemerson (dismissed):

ahus1 (approved):
Thank you for the change! I've updated the migration guide a bit to also include the realm settings that are now validated.

PR #44972, opened 2025-12-17
[OID4VCI] DefaultCryptoSdJwsTest.shouldValidateAgeSinceIssued_IfJwtIsTooOld() s…

### LINKED ISSUE #44971: DefaultCryptoSdJwsTest.shouldValidateAgeSinceIssued_IfJwtIsTooOld() sometimes fails in CI
### Before reporting an issue

### Area

ci

### Describe the bug

For example see https://github.com/keycloak/keycloak/actions/runs/20303914020/job/58316657008?pr=44828 .

### Version

nightly (from 2025-12-17)

### Regression

### Expected behavior

Test always working

### Actual behavior

Test sometimes fails

### How to Reproduce?

Issue is hard to reproduce as it happens just sometimes that error message is not as expected when the time moves by 1 second forward during test in progress.

But it is possible to reproduce it with debugger and debug the test and wait a little bit (1 second or more) after the `now` variable is created in the test, but before the token is issued. 

### Anything else?

_No response_

mposolda:
…ometimes fails in CI

closes #44971

Using same pattern as in `TimeClaimVerifierTest` (Checking exact time may fail if the current time moves by 1 second during the test running, which causes that test sometimes fail in the CI).

### DIFF
--- a/core/src/test/java/org/keycloak/sdjwt/SdJwsTest.java
+++ b/core/src/test/java/org/keycloak/sdjwt/SdJwsTest.java
@@ -35,6 +35,7 @@ import org.junit.Test;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
 public abstract class SdJwsTest {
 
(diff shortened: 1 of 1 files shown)


wistefan (approved):
LGTM

mposolda:
@wistefan Thanks!

PR #44977, opened 2025-12-17
[OID4VCI] More capabilities in SdJwtVP API when creating presentations

### LINKED ISSUE #44976: More capabilities in SdJwtVP API when creating presentations
### Description

I have the possible use-cases like

- Create the Sd-JWT presentation with disclosing specified claims. The use-case is for example `Create presentation with only given_name and family_name disclosed` 

- Create the Sd-JWT presentation without any claims disclosed

Current `SdJwtVP.present` method from Sd-JWT API has limitations that:
- It only allows to specify hashes (Disclosure digests) of claims to disclose. If I want to present specified claims (EG. `given_name`) it is not easily possible with the current API
- It does not allow to create presentation without any claims disclosed. Because when parameter `disclosureDigests` is null or empty, the presentation is considered to disclose everything

mposolda:
closes #44976

- Adding new parameter `discloseAllClaims` to existing method `present` . It allows to disclose all claims of this presentation when true. When false and the given `disclosureDigests` is empty, it means that nothing would be disclosed. The possibility of "disclose nothing" was missing from the current API
- Adding new method `presentWithSpecifiedClaims`, which allows to specify names of the claims to disclose (not digests, but actual names like `given_name` or `family_name` for example)

### DIFF
--- a/core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
+++ b/core/src/main/java/org/keycloak/sdjwt/vp/SdJwtVP.java
@@ -209,19 +209,32 @@ public class SdJwtVP {
         return issuerSignedJWT.getCnfClaim().orElse(null);
     }
 
+    /**
+     * Create new Sd-JWT presentation from this Sd-JWT
+     *
+     * @param disclosureDigests Disclosure digests (hashes) of the claims to disclose.
+     * @param discloseAllClaims When the parameter is true, then disclosureDigests parameter is ignored and everything is presented. When false, then only claims specified
+     *                         by disclosureDigests are presented
+     * @param keyBindingClaims Key binding claims. When omitted, created presentation may not contain key-binding
+     * @param holdSignatureSignerContext Useful for signing the key-binding JWT
+     * @return String with new Sd-JWT presentation with added key-binding and selected disclosed claims
+     */
     public String present(List<String> disclosureDigests,
+                          boolean discloseAllClaims,
                           ObjectNode keyBindingClaims,
                           SignatureSignerContext holdSignatureSignerContext) {
         StringBuilder sb = new StringBuilder();
-        if (disclosureDigests == null || disclosureDigests.isEmpty()) {
+        if (discloseAllClaims) {
             // disclose everything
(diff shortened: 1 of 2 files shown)


wistefan (approved):
LGTM

Just for my understanding, whats the use case of an SD_JWT when disclosing everything? Why not use a plain JWT in this case?

mposolda:
@wistefan Thanks! The ability to disclose everything was already available before this PR. This PR adds ability to "disclose nothing" from the undisclosed claims. I can certainly see use-cases for both like:

- Ability to disclose everything: The VC is available to the holder. Holder may want to send presentation to specific verifier `verifier1` and disclose everything with him and also include the key-binding of holder himself. However for some other verifier `verifier2`, holder may want to disclose only set of specific claims.

- Ability to disclose nothing: The Holder may want to send presentation to specific verifier, but only include the original "Issuer signed JWT" with the claims undisclosed directly in this issuer-signed JWT. Holder may still need to include key-binding. Verifier would be able to verify key-binding and see only the claims available in the "Issuer signed JWT", but no any other claims available in the `_sd` field of the "Issuer signed JWT" .

wistefan:
> @wistefan Thanks! The ability to disclose everything was already available before this PR. This PR adds ability to "disclose nothing" from the undisclosed claims. I can certainly see use-cases for both like:
> 
>     * Ability to disclose everything: The VC is available to the holder. Holder may want to send presentation to specific verifier `verifier1` and disclose everything with him and also include the key-binding of holder himself. However for some other verifier `verifier2`, holder may want to disclose only set of specific claims.
> 
>     * Ability to disclose nothing: The Holder may want to send presentation to specific verifier, but only include the original "Issuer signed JWT" with the claims undisclosed directly in this issuer-signed JWT. Holder may still need to include key-binding. Verifier would be able to verify key-binding and see only the claims available in the "Issuer signed JWT", but no any other claims available in the `_sd` field of the "Issuer signed JWT" .

Hi, thanks for the answer.
I might misunderstand something here, but doesn´t "disclose everything" mean that everything is put into the credential in plain, instead of allowing the holder to decide about the disclosure later?

PR #44984, opened 2025-12-18
Being possible to accept terms and conditions befure User is saved in…

### LINKED ISSUE #28714: Require user to accept terms and conditions, in first-broker-login flow
### Description

When a new user registers/signs in, using the first-broker-login flow personal data is already collected, stored and processed before the necessary consent is obtained from the user. This is not GDPR compliant within Europe, as explicit consent, through accepting terms and conditions, must be done, before any collection and processing, and storage can be done.

Adding a terms and condition subflow, into the first-broker-login flow, results in an error saying userid not found, suggesting keycloak doesn't support creating/login of users, in that specific flow. 

Someone else mentioned the same issue, and that was fixed in the registration flow. Since a user can technically also be created, if they log in through an external idp, using the first-broker-login flow, this might have been overlooked. 

Does anyone have a fix for this? 

I could try and fix it myself, but I have no idea where to start. So if someone could help me get started, that would be appreciated. 

### Discussion

See https://github.com/keycloak/keycloak/issues/8750

### Motivation

Following European data protection laws.

### Details

_No response_

cgeorgilakis:
… Keycloak during first broker login. Follow GDPR.

Closes #28714

User profile makes implementation more easy, as only editable attributes are shown and validation rules exists. Terms and condition must be enabled as required actions and set to true in `Review Profile config` flow in order federated user during first broker login asked to accept _Terms and Conditions_.

gdpr_update_profile_upstream

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/forms/login/LoginFormsProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/forms/login/LoginFormsProvider.java
@@ -38,6 +38,7 @@ import org.keycloak.sessions.AuthenticationSessionModel;
 public interface LoginFormsProvider extends Provider {
 
     String UPDATE_PROFILE_CONTEXT_ATTR = "updateProfileCtx";
+    String TERMS_ACCEPTANCE_REQUIRED = "termsAcceptanceRequired";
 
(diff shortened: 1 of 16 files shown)


Cyb4Black:
Any progress on this one?

PR #44995, opened 2025-12-18
[OID4VCI] Disable OID4VCI functionality when Verified Credentials switch is off

### LINKED ISSUE #44622: OID4VCI functionality should be disabled for the realm when "Verified Credentials" switch is disabled
### Before reporting an issue

### Area

oid4vc

### Describe the bug

We have switch `Verified Credentials` in the realm settings in the admin console, which is visible when `oid4vci` feature is enabled.

When this is disabled, the OID4VCI functionality should be probably disabled, but this is not the case.

### Version

nightly (from 2025-12-03)

### Regression

### Expected behavior

When the switch `Verified Credentials` in the realm settings in the admin console is disabled, it can be good if there are not any references to the OID4VCI in the admin console and account console UI and the related providers and endpoints are disabled as well.

### Actual behavior

- The link to the "OpenID4VCI Credential Issuer Metadata" is shown just when switch enabled: OK behaviour
- The tab "Verifiable credentials" in the account console is shown just when the switch is enabled: OK behaviour
- The various endpoints and grants seems to be still enabled even if this switch is disabled: KO
- Various options in the admin console are still visible when the switch is disabled: KO
- It is possible to create client of `OID4VCI` protocol even if the switch is disabled: KO (should be handled both server-side and admin console UI IMO)
- The client scopes of `oid4vci` are still visible in the client scopes tab: KO (But I think they should be visible when assigning/unassigning them for the clients, so it is possible to eventually unassign them from existing clients if already assigned to them)

### How to Reproduce?

See description and other sections for more details

### Anything else?

_No response_

forkimenjeckayang:
## Sumarry

This PR ensures that OID4VCI (OpenID for Verifiable Credential Issuance) functionality is properly disabled when the "`Verified Credentials`" switch is turned off in realm settings.

**Changes:**
- All OID4VCI endpoints return 403 Forbidden when the realm setting is disabled
- The OID4VCI discovery endpoint `(.well-known/openid-credential-issuer)` returns 404 Not Found
- Pre-authorized code grant is rejected when the feature is disabled
- OID4VCI protocol clients cannot be created via Admin API or UI when disabled
- OID4VCI-related UI sections are hidden in Admin Console and Account Console when disabled
- OID4VCI client scopes remain visible for management purposes (expected behavior)

**Testing:**
- Added backend tests to verify endpoint and grant behavior when disabled
- Added frontend tests to verify UI visibility and protocol option filtering
- Fixed test configuration to enable verifiable credentials by default for OID4VCI tests

closes https://github.com/keycloak/keycloak/issues/44622


testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oauth/PreAuthorizedGrantTest.java
```diff
@@ -97,6 +97,33 @@ public void testPreAuthorizedGrantNoCode() throws Exception {
...
+            assertEquals("Pre-authorized grant should be forbidden when verifiable credentials are disabled.",
+                    HttpStatus.SC_FORBIDDEN, accessTokenResponse.getStatusCode());
+        } finally {
+            // Re-enable verifiable credentials so other tests see the default behavior
+            RealmRepresentation realmRepReset = adminClient.realm(TEST_REALM_NAME).toRepresentation();
+            realmRepReset.setVerifiableCredentialsEnabled(true);
+            adminClient.realm(TEST_REALM_NAME).update(realmRepReset);
+        }
+    }
+
```
Captain-P-Goldfish, line 126: Just as question. How would the AuthorizationCodeFlow be handled if a scope value is present that relates to an OID4VCI Scope?
in my opinion the scope should be treated as unknown scope and should be ignored as defined in the oidc spec:

```
Scope values used that are not understood by an implementation SHOULD be ignored
```
from
```
scope
    REQUIRED. OpenID Connect requests MUST contain the openid scope value. If the openid scope value is not present, the behavior is entirely unspecified. Other scope values MAY be present. Scope values used that are not understood by an implementation SHOULD be ignored. See Sections [5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims) and [11](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess) for additional scope values defined by this specification.
``` 

But since Keycloak is not applying to the `SHOULD`-suggestion anyway we could also simply abort with an OAuth2 error because the scope cannot be handled.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/client/ClientRegistrationTest.java
```diff
@@ -206,6 +207,34 @@ public void registerClientAsAdminWithCreateOnly() throws ClientRegistrationExcep
...
+            assertEquals("Creating an OID4VC client should be rejected when verifiable credentials are disabled.",
+                    400, response.getStatus());
+        } finally {
+            // Re-enable verifiable credentials so other tests see the default behavior
+            RealmRepresentation resetRealm = adminClient.realm(REALM_NAME).toRepresentation();
+            resetRealm.setVerifiableCredentialsEnabled(true);
+            adminClient.realm(REALM_NAME).update(resetRealm);
+        }
+    }
+
```
Captain-P-Goldfish, line 224: as far as I know clients do not support the value `oid4vc`  as proctol anymore. This was still from the old approach when we tried to implement OID4VCI as a new protocol and not as an OAuth2-extension.

The protocol value `oid4vc` can be set for ClientScopes though.

services/src/main/java/org/keycloak/services/resources/admin/ClientsResource.java
```diff
@@ -206,6 +207,14 @@ public Response createClient(final ClientRepresentation rep) {
...
         auth.clients().requireManage();
 
+        // Validate that OID4VC protocol cannot be used when Verifiable Credentials is disabled for the realm
+        if (OID4VCLoginProtocolFactory.PROTOCOL_ID.equals(rep.getProtocol()) && !realm.isVerifiableCredentialsEnabled()) {
+            throw new ErrorResponseException(
+                    Errors.INVALID_INPUT,
+                    "OID4VC protocol cannot be used when Verifiable Credentials is disabled for this realm",
+                    Response.Status.BAD_REQUEST);
+        }
+
```
Captain-P-Goldfish: should be unnecessary, see other comment on unit-test

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -247,6 +247,28 @@ private Map<String, CredentialBuilder> loadCredentialBuilders(KeycloakSession ke
...
+    /**
+     * Validates whether OID4VCI functionality is enabled for the realm.
+     * <p>
+     * If the realm setting is disabled, this method logs the status and throws a
+     * {@link CorsErrorResponseException} with an appropriate error message.
+     * </p>
+     *
+     * @throws CorsErrorResponseException if OID4VCI is disabled for the realm.
+     */
+    private void checkRealmEnabled() {
```
Captain-P-Goldfish: I think the name of the method is misleading. I would suggest to name it `checkIsOid4vciEnabled`

Captain-P-Goldfish (changes requested):

Captain-P-Goldfish (changes requested):
Much better but now it is still possible (is I am not mistaken) to create ClientScopes of type oid4vc if the button in the UI is disabled. Am I wrong?

forkimenjeckayang:
> Much better but now it is still possible (is I am not mistaken) to create ClientScopes of type oid4vc if the button in the UI is disabled. Am I wrong?

On that, yes.
When the button is disabled, we cannot create a client scope of type oid4vc

Captain-P-Goldfish:
> > Much better but now it is still possible (is I am not mistaken) to create ClientScopes of type oid4vc if the button in the UI is disabled. Am I wrong?
> 
> On that, yes. When the button is disabled, we cannot create a client scope of type oid4vc

I am asking because I do not find a check that prevents us from creating ClientScopes with protocolType `oid4vc`. The protocolType for ClientScopes is checked here:

```java
@POST
    @Consumes(MediaType.APPLICATION_JSON)
    @NoCache
    @Tag(name = KeycloakOpenAPI.Admin.Tags.CLIENT_SCOPES)
    @Operation(summary = "Create a new client scope Client Scope’s name must be unique!")
    @APIResponses(value = {
        @APIResponse(responseCode = "201", description = "Created"),
        @APIResponse(responseCode = "403", description = "Forbidden"),
        @APIResponse(responseCode = "409", description = "Conflict")
    })
    public Response createClientScope(ClientScopeRepresentation rep) {
        auth.clients().requireManageClientScopes();
        ClientScopeResource.validateClientScopeName(rep.getName());
 -----> ClientScopeResource.validateClientScopeProtocol(session, rep.getProtocol());
        ClientScopeResource.validateDynamicClientScope(rep);
       ...
    }
   ...
```
```java
public static void validateClientScopeProtocol(KeycloakSession session, String protocol)
            throws ErrorResponseException {
        KeycloakSessionFactory sessionFactory = session.getKeycloakSessionFactory();
        Set<String> acceptedProtocols = sessionFactory.getProviderFactoriesStream(LoginProtocol.class)
                                                      .map(type -> (LoginProtocolFactory) type)
                                                      .map(LoginProtocolFactory::getId)
                                                      .collect(Collectors.toSet());
        // the OID4VC protocol is not registered to prevent it from being displayed in the client-details ui
        acceptedProtocols.add(OID4VCLoginProtocolFactory.PROTOCOL_ID);

        if (protocol == null || !acceptedProtocols.contains(protocol)) {
            throw ErrorResponse.error("Unexpected protocol", Response.Status.BAD_REQUEST);
        }
    }
```

and there is no such validation to check for the realms configuration yet.

forkimenjeckayang:
> > > Much better but now it is still possible (is I am not mistaken) to create ClientScopes of type oid4vc if the button in the UI is disabled. Am I wrong?
> > 
> > 
> > On that, yes. When the button is disabled, we cannot create a client scope of type oid4vc
> 
> I am asking because I do not find a check that prevents us from creating ClientScopes with protocolType `oid4vc`. The protocolType for ClientScopes is checked here:
> 
> ```java
> @POST
>     @Consumes(MediaType.APPLICATION_JSON)
>     @NoCache
>     @Tag(name = KeycloakOpenAPI.Admin.Tags.CLIENT_SCOPES)
>     @Operation(summary = "Create a new client scope Client Scope’s name must be unique!")
>     @APIResponses(value = {
>         @APIResponse(responseCode = "201", description = "Created"),
>         @APIResponse(responseCode = "403", description = "Forbidden"),
>         @APIResponse(responseCode = "409", description = "Conflict")
>     })
>     public Response createClientScope(ClientScopeRepresentation rep) {
>         auth.clients().requireManageClientScopes();
>         ClientScopeResource.validateClientScopeName(rep.getName());
>  -----> ClientScopeResource.validateClientScopeProtocol(session, rep.getProtocol());
>         ClientScopeResource.validateDynamicClientScope(rep);
>        ...
>     }
>    ...
> ```
> 
> ```java
> public static void validateClientScopeProtocol(KeycloakSession session, String protocol)
>             throws ErrorResponseException {
>         KeycloakSessionFactory sessionFactory = session.getKeycloakSessionFactory();
>         Set<String> acceptedProtocols = sessionFactory.getProviderFactoriesStream(LoginProtocol.class)
>                                                       .map(type -> (LoginProtocolFactory) type)
>                                                       .map(LoginProtocolFactory::getId)
>                                                       .collect(Collectors.toSet());
>         // the OID4VC protocol is not registered to prevent it from being displayed in the client-details ui
>         acceptedProtocols.add(OID4VCLoginProtocolFactory.PROTOCOL_ID);
> 
>         if (protocol == null || !acceptedProtocols.contains(protocol)) {
>             throw ErrorResponse.error("Unexpected protocol", Response.Status.BAD_REQUEST);
>         }
>     }
> ```
> 
> and there is no such validation to check for the realms configuration yet.

I get your point @Captain-P-Goldfish  and thanks for pointing that out. On the UI side we cannot , but on the admin REST API part, it's still possible regardless if the switch veriable credentials is disabled.
I've adjusted that..

tests/base/src/test/java/org/keycloak/tests/admin/client/ClientScopeTest.java
```diff
@@ -744,7 +744,7 @@ public void createClientScopeWithoutProtocol() {
 
     @DisplayName("Create ClientScope with protocol:")
     @ParameterizedTest
-    @ValueSource(strings = {"openid-connect", "saml", "oid4vc"})
+    @ValueSource(strings = {"openid-connect", "saml"})
     public void createClientScopeWithOpenIdProtocol(String protocol) {
         createClientScope(protocol);
     }
```
Captain-P-Goldfish, line 750: You basically just removed the test that the creation of an oid4vc ClientScope succeeds and as far as I can see there is no replacement for it. Please readd this test or at least add it somewhere else if its about the verifiableCredentialsEnabled flag.

Captain-P-Goldfish (changes requested):

tests/base/src/test/java/org/keycloak/tests/admin/client/ClientScopeTest.java
forkimenjeckayang: I did so because we have a dedicated class for testing client scopes for oid4vc `ClientScopeTestOid4Vci.java`
And since we require the switch for verifibale credentials that was enbaled only for that test file. Suggestion coud be included here instead `ClientScopeTestOid4Vci.java`

Captain-P-Goldfish: I see, in that case I am happy :-)
thanks for clarifying that

Captain-P-Goldfish (approved):
LGTM

mposolda:
@forkimenjeckayang @Captain-P-Goldfish Thanks!

I've did additional review as well and changes looks good. Besides (as usual), I propose that we add provider specific hooks instead of `if/else` branches specific to OID4VCI protocol within core codebase classes. Considering that I've sent PR with additional changes to your branch https://github.com/adorsys/keycloak-oid4vc/pull/232 . If you are OK with the changes, you can merge it to your branch and it should update this PR automatically as well.

mposolda:
@forkimenjeckayang Thanks for the updates! I've sent additional PR to your branch, which should hopefully fix the tests: https://github.com/adorsys/keycloak-oid4vc/pull/233 . Hopefully tests are fixed in this PR once that one is merged into your branch.

Could you please also rebase your branch afterwards to make sure that this PR is aligned with latest Keycloak main? Thanks!

forkimenjeckayang:
> @forkimenjeckayang Thanks for the updates! I've sent additional PR to your branch, which should hopefully fix the tests: [adorsys#233](https://github.com/adorsys/keycloak-oid4vc/pull/233) . Hopefully tests are fixed in this PR once that one is merged into your branch.
> 
> Could you please also rebase your branch afterwards to make sure that this PR is aligned with latest Keycloak main? Thanks!

Hello @mposolda 
Let me do that, if tests are fixed I'll rebase . If not i'll try to fix befoe rebasing.
Thank you

mposolda:
@forkimenjeckayang Thanks a lot for all the work on this!

PR #45004, opened 2025-12-18
OID4VCI credentials have invalid subject id value

### LINKED ISSUE #43854: OID4VCI credentials have invalid subject id value

tdiesler:
closes #43854

This is the second attempt to fix the subject id issue. Unlike the previous attempt it is not concerned with SD JWT claim visibility nor does it introduce an additional 'did' user attribute.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
@@ -21,20 +21,23 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.UUID;
+import java.util.Objects;
+import java.util.function.Consumer;
 
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.ProtocolMapperModel;
+import org.keycloak.models.UserModel;
 import org.keycloak.models.UserSessionModel;
+import org.keycloak.models.utils.KeycloakModelUtils;
 import org.keycloak.protocol.ProtocolMapper;
 import org.keycloak.protocol.oid4vc.OID4VCLoginProtocolFactory;
 import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.provider.ProviderConfigProperty;
 
-import org.apache.commons.collections4.ListUtils;
+import static org.keycloak.OID4VCConstants.CLAIM_NAME_SUBJECT_ID;
 
 /**
- * Sets an ID for the credential, either randomly generated or statically configured
+ * Sets an ID for the credential subject, either by attribute mapping or statically configured
  *
  * @author <a href="https://github.com/wistefan">Stefan Wiedemann</a>
  */
@@ -48,10 +51,31 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
         ProviderConfigProperty idPropertyNameConfig = new ProviderConfigProperty();
         idPropertyNameConfig.setName(CLAIM_NAME);
         idPropertyNameConfig.setLabel("ID Property Name");
-        idPropertyNameConfig.setHelpText("Name of the property to contain the id.");
-        idPropertyNameConfig.setDefaultValue("id");
+        idPropertyNameConfig.setHelpText("Name of the property to contain the subject id.");
         idPropertyNameConfig.setType(ProviderConfigProperty.STRING_TYPE);
+        idPropertyNameConfig.setDefaultValue(CLAIM_NAME_SUBJECT_ID);
         CONFIG_PROPERTIES.add(idPropertyNameConfig);
+
+        ProviderConfigProperty userAttributeConfig = new ProviderConfigProperty();
+        userAttributeConfig.setName(USER_ATTRIBUTE_KEY);
+        userAttributeConfig.setLabel("User attribute");
+        userAttributeConfig.setHelpText("The user attribute that maps to the subject id.");
+        userAttributeConfig.setType(ProviderConfigProperty.LIST_TYPE);
+        userAttributeConfig.setOptions(List.of(UserModel.USERNAME, UserModel.EMAIL, UserModel.ID));
+        userAttributeConfig.setDefaultValue(UserModel.ID);
+        CONFIG_PROPERTIES.add(userAttributeConfig);
+    }
+
+    public static ProtocolMapperModel create(String name, String claimName, String userAttribute) {
+        var mapperModel = new ProtocolMapperModel();
+        mapperModel.setName(name);
+        Map<String, String> configMap = new HashMap<>();
+        configMap.put(CLAIM_NAME, claimName);
+        configMap.put(USER_ATTRIBUTE_KEY, userAttribute);
+        mapperModel.setConfig(configMap);
+        mapperModel.setProtocol(OID4VCLoginProtocolFactory.PROTOCOL_ID);
+        mapperModel.setProtocolMapper(MAPPER_ID);
+        return mapperModel;
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -48,45 +51,57 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
         ProviderConfigProperty idPropertyNameConfig = new ProviderConfigProperty();
         idPropertyNameConfig.setName(CLAIM_NAME);
         idPropertyNameConfig.setLabel("ID Property Name");
-        idPropertyNameConfig.setHelpText("Name of the property to contain the id.");
-        idPropertyNameConfig.setDefaultValue("id");
+        idPropertyNameConfig.setHelpText("Name of the property to contain the subject id.");
```
mposolda: AFAIK This configuration property refers to the name of the claim, which will be in the OID4VC credential. IMO the label "ID Property Name" is confusing and tooltip is confusing as well and also it is not consistent with the naming used by "User Attribute Mapper". So I wonder if for consistency it should just simply use something like:
```
idPropertyNameConfig.setLabel(OIDCAttributeMapperHelper.TOKEN_CLAIM_NAME_LABEL);
idPropertyNameConfig.setHelpText("Name of the claim to insert into the token. This can be a fully qualified name such as 'address.street'. In case that 'id' is used as a value of this configuration property, it would be mapped into sd-jwt credential as claim 'sub'.");
```
in same way like "User Attribute Mapper" is using? Note that I've used also slightly different note for help-text to inform that "id" is mapped into "sub" in case of sd-jwt credential (if I understand correctly the code snippet you added in `SdJwtCredentialBuilder`).

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -48,45 +51,57 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
...
-        return CONFIG_PROPERTIES;
-    }
 
-    @Override
-    public List<String> getMetadataAttributePath() {
-        return ListUtils.union(getAttributePrefix(), List.of("id"));
+        ProviderConfigProperty userAttributeConfig = new ProviderConfigProperty();
+        userAttributeConfig.setName(USER_ATTRIBUTE_KEY);
+        userAttributeConfig.setLabel("User attribute");
+        userAttributeConfig.setHelpText("The user attribute that maps to the subject id.");
```
mposolda:
```suggestion
        userAttributeConfig.setHelpText("The name of the user attribute that maps to the subject id.");
```
Just a nitpick to make this tooltip a little bit cleaner.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -21,20 +21,23 @@
...
 import org.keycloak.protocol.oid4vc.OID4VCLoginProtocolFactory;
 import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.provider.ProviderConfigProperty;
 
-import org.apache.commons.collections4.ListUtils;
+import static org.keycloak.OID4VCConstants.CLAIM_NAME_SUBJECT_ID;
 
 /**
- * Sets an ID for the credential, either randomly generated or statically configured
+ * Sets an ID for the credential subject, either by attribute mapping or statically configured
```
mposolda:
```suggestion
 * Sets an ID for the credential subject, either from User ID or by attribute mapping
```
Small nitpick to this javadoc as AFAIK the "statically configured" is not supported anymore with your changes.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -96,7 +111,7 @@ public String getDisplayType() {
 
     @Override
     public String getHelpText() {
-        return "Assigns a subject ID to the credentials subject. If no specific id is configured, a randomly generated one is used.";
+        return "Assigns a subject ID to the credentials subject. The default is the user's id, others can be selected.";
```
mposolda:
```suggestion
        return "Sets an ID for the credential subject, either from User ID or by attribute mapping.";
```
Another small nitpick: I like a bite more the formulation used in the javadoc and hence proposed here to update "help text" to use the same text. It sounds a little bit more clear what this mapper is doing. But I am not strong on this to be used as help text.

mposolda (commented):
@tdiesler Thanks for the PR! I've added few review comments in the PR. Can you check them please?

@Captain-P-Goldfish @IngridPuppet @Awambeng @forkimenjeckayang @Ogenbertrand Do someone of you prefer to review as well?

mposolda (approved):
@tdiesler Thanks! I am merging since this PR blocks the other PR.

If someone else wants to review, feel free to create follow-up issues / PRs.

Issue #44574, opened 2025-12-01
Unable to find contextual data of type: org.keycloak.models.KeycloakSession


tdiesler:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

testsuite

### Describe the bug

Most test in oid4vci fail in some after test logic (i.e. AuthServerTestEnricher.removeTestRealms)

```
08:59:02,594 INFO  [org.keycloak.testsuite.arquillian.AuthServerTestEnricher] removing test realms after test class
08:59:02,675 ERROR [io.undertow.request] UT005023: Exception handling request to /auth/admin/realms/test
org.jboss.resteasy.spi.LoggableFailure: RESTEASY003880: Unable to find contextual data of type: org.keycloak.models.KeycloakSession
	at org.jboss.resteasy.core.ContextParameterInjector$GenericDelegatingProxy.invoke(ContextParameterInjector.java:137)
	at jdk.proxy2/jdk.proxy2.$Proxy139.getKeycloakSessionFactory(Unknown Source)
	at org.keycloak.storage.StoreSyncEvent.fire(StoreSyncEvent.java:47)
	at org.keycloak.services.managers.RealmManager$1.commitImpl(RealmManager.java:310)
	at org.keycloak.models.AbstractKeycloakTransaction.commit(AbstractKeycloakTransaction.java:46)
	at org.keycloak.services.DefaultKeycloakTransactionManager.lambda$commitWithTracing$1(DefaultKeycloakTransactionManager.java:171)
	at org.keycloak.tracing.NoopTracingProvider.trace(NoopTracingProvider.java:59)
	at org.keycloak.tracing.NoopTracingProvider.trace(NoopTracingProvider.java:69)
	at org.keycloak.services.DefaultKeycloakTransactionManager.commitWithTracing(DefaultKeycloakTransactionManager.java:170)
	at org.keycloak.services.DefaultKeycloakTransactionManager.lambda$commit$0(DefaultKeycloakTransactionManager.java:147)
	at org.keycloak.tracing.NoopTracingProvider.trace(NoopTracingProvider.java:59)
	at org.keycloak.tracing.NoopTracingProvider.trace(NoopTracingProvider.java:69)
	at org.keycloak.services.DefaultKeycloakTransactionManager.commit(DefaultKeycloakTransactionManager.java:122)
[...]

PR #45008, opened 2025-12-18
[OID4VCI] Add support for user did as subject id

### LINKED ISSUE #45006: [OID4VCI] Add support for user did as subject id
### Description

For most intents and purposes, it is the Holder's DID that becomes the `vc.credentialSubject.id` and `sub` claim in case of a VC JWT

Currently we have a OID4VCSubjectIdMapper that allows User ...

* id (default)
* username
* email
* other unmanaged attributes

to be mapped to the subject id. 

In the context of this issue, we would need to add a DID attribute to the User's profile and make that the default. This would also be a prerequisite for EBSI compliance.

In practice, a verifiable credential is issued to one of the Holder's DIDs. A Verifier can resolve the subject id (did) to a DID Document  where it can find the public key material associated with that DID. It can then request proof of possession from the Holder and such verify that the given VC was indeed issued to the Holder that presents the VC.

### Value Proposition

Make the User's DID an attribute that is available when oid4vci is enabled for a given realm.

### Goals

see above

### Non-Goals

For now, we can limit ourselves to did:key:... i.e. obtain the public key from the did itself. DID resolution for other did schemes is out of scope

### Discussion

_No response_

### Notes

_No response_

tdiesler:
closes #45006


tdiesler:
Depends on #45004

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
```diff
@@ -62,8 +63,8 @@ public class OID4VCSubjectIdMapper extends OID4VCMapper {
         userAttributeConfig.setLabel("User attribute");
         userAttributeConfig.setHelpText("The name of the user attribute that maps to the subject id.");
         userAttributeConfig.setType(ProviderConfigProperty.LIST_TYPE);
-        userAttributeConfig.setOptions(List.of(UserModel.USERNAME, UserModel.EMAIL, UserModel.ID));
-        userAttributeConfig.setDefaultValue(UserModel.ID);
+        userAttributeConfig.setOptions(List.of(USER_ATTRIBUTE_NAME_DID, UserModel.USERNAME, UserModel.EMAIL, UserModel.ID));
+        userAttributeConfig.setDefaultValue(USER_ATTRIBUTE_NAME_DID);
         CONFIG_PROPERTIES.add(userAttributeConfig);
     }
```
IngridPuppet: The current implementation is definitely not DID-first. For that reason, I suggest to keep the previous default value, i.e., `UserModel.ID`, as the name of the mapper prompts to. What is your take?

services/src/main/resources/org/keycloak/userprofile/config/keycloak-oid4vci-user-profile.json
```diff
@@ -0,0 +1,78 @@
...
+          "view": [ "admin", "user" ],
+          "edit": [ "admin", "user" ]
+        },
+        "validations": {
+          "pattern": {
+            "pattern": "^did:.+:.+$",
+            "error-message": "Value must start with 'did:scheme:'"
+          }
+        }
+      }
```
IngridPuppet: Indentation misalignment. Autoformat the file?

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/AuthServerTestEnricher.java
```diff
@@ -728,6 +728,8 @@ public static void removeTestRealms(TestContext testContext, Keycloak adminClien
                     realms.append(testRealm.getRealm()).append(", ");
                 } catch (NotFoundException e) {
                     // Ignore
+                } catch (Exception e) {
+                    // Ignore
                 }
```
IngridPuppet, line 731: Remove the "NotFoundException" block?

IngridPuppet (commented):
Hello @tdiesler - I reviewed the PR and left minor comments. Please could you check?

Also, I recall @Captain-P-Goldfish had some specific concerns regarding updates to `OID4VCSubjectIdMapper`. It'd be nice he checks the PR too.

services/src/main/resources/org/keycloak/userprofile/config/keycloak-oid4vci-user-profile.json
tdiesler: I reformatted both files - so that the delta between them becomes visible again

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/AuthServerTestEnricher.java
tdiesler: That leaked in and is a workaround for #44574

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/mappers/OID4VCSubjectIdMapper.java
tdiesler: I gradually would like to move it to "did first". Issuing a credential to a subject that is identified by some keycloak internal user id doesn't make much sense IMHO. What would a verifier say about that ;-)

One major step to make Keycloak a proper oid4vci Issuer has started here #45544

Anyway, we can leave it UserModel.ID until we feel ready to switch the default.

IngridPuppet (approved):
Thank you for the update.

tdiesler:
@Captain-P-Goldfish this is part of an old PR that you reviewed some ago - nothing much has changed wrt to user.did since then. Could you perhaps have another quick look, its been sitting here for quite a while.

services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
```diff
@@ -521,21 +526,63 @@ protected Map<UserProfileContext, UserProfileMetadata> getContextualMetadataRegi
...
         // If no user-defined configuration is set, the system default configuration must have been set
         // In Quarkus, the system default configuration is set at build time for optimization purposes
-        UPConfig defaultConfig = ofNullable(config.get("configFile"))
+        UPConfig parsedConfig = ofNullable(config.get("configFile"))
                 .map(Paths::get)
                 .map(UPConfigUtils::parseConfig)
-                .orElse(PARSED_DEFAULT_RAW_CONFIG);
+                .orElse(null);
+
+        // For all intents and purposes, a Verifiable Credential (VC) is bound to a Holder's Digital Identity (DID).
```
mposolda: Minor: Is it possible to remove this long comment from the `DeclarativeUserProfileProviderFactory`? That class has logic specific to user-profile and very long comment related to the OID4VCI feature is not the right place for that. Maybe there is better place in some OID4VCI related utility class?

mposolda (commented):
@tdiesler Thanks for the PR and sorry for late review.

I've added some minor comment inline.

Besides that, it would be good to avoid introducing another file `keycloak-oid4vci-user-profile.json` configuration and have 2 JSON files with the user profile configuration within Keycloak codebase. This has various issues (EG. especially risk that there are more places to maintain and introduced risk of one of the files being out-of-sync with other etc).

The approach typically used is, that some attributes are added to user profile "in runtime" based on some features being enabled or disabled etc. This is for example attribute `locale`, which is added into profile when "Localization" is enabled for the realm. See for example `DeclarativeUSerProfileProviderFactory` and how `UserModel.LOCALE` is being added to profile types for "admin" and for "user" etc.

This would also allow that `did` attribute is present just when OID4VCI-VCI feature is enabled and also OID4VCI is enabled for the realm (as it should be likely omitted when feature is enabled, but switch is disabled for particular realm).

mposolda:
@tdiesler Other minor comment: It would be good if the newly added attribute is also "localized" . The displayName should refer to some value, which is referenced from the localization file https://github.com/keycloak/keycloak/blob/main/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties (this one is the only one required. Other languages does not need to be updated unless you know the corresponding language :-)

services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
tdiesler: done

tdiesler:
I removed the extra `keycloak-oid4vci-user-profile.json` config file and now add it to the default config like this ...

```
        if (config.getAttribute(UserModel.DID) == null) {
            UPAttribute attr = new UPAttribute(UserModel.DID);
            attr.setDisplayName("${did}");
            attr.setPermissions(new UPAttributePermissions(Set.of(ROLE_ADMIN, ROLE_USER), Set.of(ROLE_ADMIN, ROLE_USER)));
            attr.setValidations(Map.of(PatternValidator.ID, Map.of(
                    "pattern", "^did:.+:.+$",
                    "error-message", "Value must start with 'did:scheme:'")));
            config.addOrReplaceAttribute(attr);
            PARSED_DEFAULT_RAW_CONFIG = null;
        }
```

This is driven by the feature at start time and not at the realm level when oid4vci gets enabled/disabled. Also, there is another duplicate configuration file `testrealm-oid4vci.json`. It'd be great if we could remove that too.

services/src/main/java/org/keycloak/userprofile/config/UPConfigUtils.java
```diff
@@ -59,7 +59,7 @@
  */
 public class UPConfigUtils {
 
-    private static final String SYSTEM_DEFAULT_CONFIG_RESOURCE = "keycloak-default-user-profile.json";
+    public static final String SYSTEM_DEFAULT_CONFIG_RESOURCE = "keycloak-default-user-profile.json";
```
mposolda: Is it possible to remove the changes from the class `UPConfigUtils` from this PR?

The PR now makes `SYSTEM_DEFAULT_CONFIG_RESOURCE` to be `public` (which is not needed) and introduces another `public static` method etc, but it seems none of those changes are needed after all.

mposolda (commented):
@tdiesler Thanks for the latest updates!

Added some additional minor comment inline.

services/src/main/java/org/keycloak/userprofile/config/UPConfigUtils.java
tdiesler: sure

mposolda: Sorry to bother and thanks for updating `SYSTEM_DEFAULT_CONFIG_RESOURCE`, but is it possible to remove all the updates from `UPConfigUtils` class from this PR as I've suggested above? As you don't need them. At least would be good to remove newly added public method as it is not needed and definitely doesn't need to be public ( `public static UPConfig parseUserProfileConfig(String resource)`) .

tdiesler: Yes, that can now be reverted since we don't need to load alternative configs any more

mposolda (approved):
@tdiesler Thanks!

tdiesler:
merci

PR #36161, opened 2024-12-29
Client cert lookup provider compliant to RFC 9440

seiferma:
Closes #20761

This pull request addresses the issue brought up in #20761. I decided to build a certificate lookup provider that is compliant to RFC 9440 instead of just building a caddy-compliant provider. This is the solution that has been suggested in the issue.

Design decisions /  implementation details:
- All configuration options are optional because the RFC already provides reasonable default values
- The length option for certificate chains defines the amount of certificates that are processed and exceeding the limit is treated as an error
- I created unit tests for the actual lookup but none for the factory because I could not find examples of tests for the other providers
- Test data is the data provided in RFC 9440

Checklist from the contribution guidelines:
- [ ] A discussion around the change (https://github.com/keycloak/keycloak/discussions/categories/ideas)
- [x] A GitHub Issue with a good description associated with the PR
- [x] One feature/change per PR
- [x] One commit per PR
- [x] PR rebased on main (git rebase, not git pull)
- [x] [Good descriptive commit message, with link to issue](https://github.com/keycloak/keycloak/blob/main/CONTRIBUTING.md#commit-messages-and-issue-linking)
- [x] No changes to code not directly related to your PR
- [x] Includes functional/integration test
- [x] Includes documentation

PR #45010, opened 2025-12-18
fix: making cert lookup spi public

### LINKED ISSUE #33818: Request for Enhancement: Make x509cert-lookup SPI public
### Description

Please consider changing the status of `x509cert-lookup` SPI from private to public.

### Discussion

_No response_

### Motivation

The project promotes the development of client certificate lookups for various proxies as extensions. See discussions in PRs #33159 and #31778.

To support the extension development, it would be great for Keycloak to ensure the stability of this SPI.

### Details

Currently, the SPI is marked as internal https://github.com/keycloak/keycloak/blob/2f1307a16230ed90f03f8ee19c8a5c6c019d7b8a/services/src/main/java/org/keycloak/services/x509/X509ClientCertificateLookupSpi.java#L33-L36

This results in a warning when Keycloak is started with an extension enabled:

```
2024-10-11 09:15:29,052 WARN  [org.key.services] (build-13) KC-SERVICES0047: envoy (io.github.nordix.keycloak.services.x509.EnvoyProxySslClientCertificateLookupFactory) is implementing the internal SPI x509cert-lookup. This SPI is internal and may change without notice
```

shawkins:
Rationale for this change:
- the api has been stable for years
- we have several examples of users creating their own cert providers, but that is not considered supported and they see warnings about this being an internal api.
- external contribution of new providers seems to take a long time #36161 

On timing, this is optional for 26.5.

closes: #33818

### DIFF
--- a/services/src/main/java/org/keycloak/services/x509/X509ClientCertificateLookup.java
+++ b/services/src/main/java/org/keycloak/services/x509/X509ClientCertificateLookup.java
@@ -35,6 +35,10 @@ public interface X509ClientCertificateLookup extends Provider {
     /**
      * Returns a client certificate, and optionally any certificates
      * in the certificate chain.
+     * <p>
+     * IMPORTANT: implementations should ensure that the source of the certificate is trusted.
+     * See for example the {@link HttpRequest#isProxyTrusted()} method.
+     *
      * @return
      */
     X509Certificate[] getCertificateChain(HttpRequest httpRequest) throws GeneralSecurityException;
--- a/services/src/main/java/org/keycloak/services/x509/X509ClientCertificateLookupSpi.java
+++ b/services/src/main/java/org/keycloak/services/x509/X509ClientCertificateLookupSpi.java
@@ -32,7 +32,7 @@ public class X509ClientCertificateLookupSpi implements Spi {
 
     @Override
     public boolean isInternal() {
-        return true;
+        return false;
     }
 
     @Override


shawkins:
There appears to be an unrelated CRDTest expectation change - which doesn't seem to occur for me locally. If needed that can be separated to a different pr.

EDIT: moved the other commit to its own pr.

vmuzikar (commented):
We would probably need some docs to officially consider it supported.

shawkins:
> We would probably need some docs to officially consider it supported.

That seemed optional as there appear to be quite a few non-internal spis not covered by https://www.keycloak.org/docs/latest/server_development/index.html

Or is there another place where this should be documented?

vmuzikar:
> That seemed optional as there appear to be quite a few non-internal spis not covered by https://www.keycloak.org/docs/latest/server_development/index.html

AFAIU, those not mentioned in the docs are not officially supported.

> Or is there another place where this should be documented?

No, this would be it.

I think we have some grey area though. We have public SPIs, yet we don't document them, and hence don't support them. I do not know then why we even mark SPIs as public.

Adding a new supported SPI is a bigger step and would require broader team involvement, IMHO. We could potentially just mark the cert lookup SPI as public and not mention it in the release notes not to give any implication it's supported. Not sure if that makes much difference though from a private SPI.

CC @stianst

shawkins:
Updated after more discussion to a compromise state: just mark as non-internal, and create follow-up issues on rationalizing isInternal and the notion of what is supported.

vmuzikar (approved):

PR #45011, opened 2025-12-18
[OID4VCI] Fix OID4VCI credential requests to restrict Default client scopes

### LINKED ISSUE #44737: CredentialRequest requires that client scope is assigned as 'Optional'
### Before reporting an issue

### Area

oid4vc

### Describe the bug

Let's assume that I send credential request for the client scope requesting credential related to the client scope `sd-jwt-credential` . As long as client scope is assigned to the client as "Optional" client scope, this works OK. However if the client scope is assigned to the client as "Default" client scope, it doesn't work.

Hint: As of today, there are some places in the `OID4VCIssuerEndpoint` class calling something like : `ClientScopeModel clientScope = clientModel.getClientScopes(false)...` 

This considers only "optional" client scopes, but not "default" client scopes assigned to the client.

### Version

nightly (from 2025-12-08)

### Regression

### Expected behavior

All client scopes assigned to the client should be considered during credentialRequest

### Actual behavior

Only "optional" client scopes assigned to the client are considered during credentialRequest

### How to Reproduce?

See details.

### Anything else?

_No response_

Awambeng:
This PR implements backend validation to restrict OID4VCI client scope assignments as default:

Changes:
- OID4VCI scopes can only be assigned as Optional scopes (not Default)
- OID4VCI scopes cannot be assigned when the Verifiable Credentials feature is disabled at the realm level
- The pre-created `oid4vc_natural_person` scope is now assigned as Optional by default

Closes #44737

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
@@ -137,7 +137,7 @@ public class OID4VCLoginProtocolFactory implements LoginProtocolFactory, OID4VCE
             naturalPersonScope.addProtocolMapper(builtins.get(FIRST_NAME_MAPPER));
             naturalPersonScope.addProtocolMapper(builtins.get(LAST_NAME_MAPPER));
             addClientScopeDefaults(naturalPersonScope);
-            newRealm.addDefaultClientScope(naturalPersonScope, true);
+            newRealm.addDefaultClientScope(naturalPersonScope, false);
         }
     }
 
@@ -202,6 +202,25 @@ public class OID4VCLoginProtocolFactory implements LoginProtocolFactory, OID4VCE
     @Override
     public int order() {
         return OIDCLoginProtocolFactory.UI_ORDER - 20;
+    }
+
+    @Override
+    public void validateClientScopeAssignment(KeycloakSession session, ClientScopeModel clientScope, boolean defaultScope, RealmModel realm) {
+        if (!OID4VC_PROTOCOL.equals(clientScope.getProtocol())) {
+            return;
+        }
+
+        if (!realm.isVerifiableCredentialsEnabled()) {
+            throw new ErrorResponseException("invalid_request",
+                    "OID4VCI client scopes cannot be assigned when Verifiable Credentials is disabled for the realm",
+                    Response.Status.BAD_REQUEST);
+        }
+
+        if (defaultScope) {
+            throw new ErrorResponseException("invalid_request",
+                    "OID4VCI client scopes cannot be assigned as Default scopes. Only Optional scope assignment is supported.",
+                    Response.Status.BAD_REQUEST);
+        }
     }
 
     // Private ---------------------------------------------------------------------------------------------------------
--- a/server-spi-private/src/main/java/org/keycloak/protocol/LoginProtocolFactory.java
+++ b/server-spi-private/src/main/java/org/keycloak/protocol/LoginProtocolFactory.java
@@ -19,6 +19,7 @@ package org.keycloak.protocol;
 
 import java.util.Map;
 
+import jakarta.ws.rs.BadRequestException;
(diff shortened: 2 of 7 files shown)


Captain-P-Goldfish (approved):
LGTM

mposolda:
@Awambeng Sorry for late response. Added comment on the related issue https://github.com/keycloak/keycloak/issues/44737#issuecomment-3777664371 . Is it OK for you to update this PR according to that?

Awambeng:
> @Awambeng Sorry for late response. Added comment on the related issue [#44737 (comment)](https://github.com/keycloak/keycloak/issues/44737#issuecomment-3777664371) . Is it OK for you to update this PR according to that?

@mposolda It's fine. I'll take a look.

Awambeng:
@mposolda @Captain-P-Goldfish I've reworked the implementation taking into consideration the discussions in #44737. Could you please have a look when you get a chance?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -137,7 +137,7 @@ public void createDefaultClientScopes(RealmModel newRealm, boolean addScopesToEx
             naturalPersonScope.addProtocolMapper(builtins.get(FIRST_NAME_MAPPER));
             naturalPersonScope.addProtocolMapper(builtins.get(LAST_NAME_MAPPER));
             addClientScopeDefaults(naturalPersonScope);
-            newRealm.addDefaultClientScope(naturalPersonScope, true);
+            newRealm.addDefaultClientScope(naturalPersonScope, false);
```
mposolda, line 140: Probably better to not assign this at all and just keep this example client scope as "none" and make sure that administrator assign this to clients when they need. But this is not a blocker for this PR as there is other PR https://github.com/keycloak/keycloak/pull/44840 related to this.

Also it seems that "Realm optional client scopes" of OID4VCI protocol are not assigned as "optional" to newly created clients (both OIDC clients and OID4VCI clients). But this is separate issue (maybe not issue at all and is intended to work like this...?). Anyway, not blocker for this PR IMO.

mposolda (approved):
@Awambeng Thanks! I've tested this and works nice IMO.

Added one comment inline, but not a blocker for this PR.

@Captain-P-Goldfish Do you want to re-review as well?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
Awambeng: > Probably better to not assign this at all and just keep this example client scope as "none" and make sure that administrator assign this to clients when they need. But this is not a blocker for this PR as there is other PR #44840 related to this.
> 
You’re right, at the moment `oid4vc_natural_person` is created and assigned as Optional to behave like a “default” scope and make the OID4VCI setup work out of the box. I’m fine changing this to leave the scope unassigned (type `none`) so that administrators explicitly attach it to clients when needed. I can either do that here or we can align it with the changes in PR #44840.

> Also it seems that "Realm optional client scopes" of OID4VCI protocol are not assigned as "optional" to newly created clients (both OIDC clients and OID4VCI clients). But this is separate issue (maybe not issue at all and is intended to work like this...?). Anyway, not blocker for this PR IMO.
>
I don’t see this as a major issue, as admins should explicitly decide which scopes to assign per client. I’d prefer to stick with the current approach.

Captain-P-Goldfish:
> @Awambeng Thanks! I've tested this and works nice IMO.
> 
> Added one comment inline, but not a blocker for this PR.
> 
> @Captain-P-Goldfish Do you want to re-review as well?

I will check if I find some time soon. Currently it is pretty difficult to find open slots.

Captain-P-Goldfish (approved):
LGTM

Awambeng:
@mposolda I’ve rebased on the latest main. Could you please review and merge?

mposolda (approved):
@Awambeng @Captain-P-Goldfish Thanks!

PR #45013, opened 2025-12-18
fix: updating the health check docs

### LINKED ISSUE #44634: Enhance health check docs
### Description

Link to just https://microprofile.io/specifications/microprofile-health/ (linked by Quarkus as https://github.com/microprofile/microprofile-health) instead of the quarkus documentation as their docs don't provide good descriptions of the functionality and any of the configuration and customizations shown would be unsupported.

The docs should describe how to appropriately use the checks for containers rather than just discussing readiness.

### Value Proposition

To help clarify the roles of the given health checks. See also the original linked issue comment.

### Goals

So that custom usage of the health endpoints is appropriate.

### Non-Goals

N/A

### Discussion

_No response_

### Notes

_No response_

shawkins:
closes: #44634

- directly added the meaning of each endpoint
- changed the link for more information to the wire format, not the quarkus docs
- added a note to highlight the function of the container healthcheck

cc @kodebach

### DIFF
--- a/docs/guides/observability/health.adoc
+++ b/docs/guides/observability/health.adoc
@@ -18,12 +18,10 @@ When this option is `false` you should block unwanted external traffic to `/heal
 
 {project_name} exposes 4 health endpoints:
 
-* `/health/live`
-* `/health/ready`
-* `/health/started`
-* `/health`
-
-See the https://quarkus.io/guides/smallrye-health#running-the-health-check[Quarkus SmallRye Health docs] for information on the meaning of each endpoint.
+* `/health/started` - allows startup probes that are used for initial verification of the application before the liveness probe takes over. This is useful for applications which require additional startup time on their first initialization.
+* `/health/live` - if the application is running. If this procedure fails the application can be discarded (terminated, shutdown).
+* `/health/ready` - if the application is ready to process requests or not. If this procedure fails the application may be made unavailable to serve new requests.
+* `/health` - accumulates all health check procedures in the application.
 
 These endpoints respond with HTTP status `200 OK` on success or `503 Service Unavailable` on failure, and a JSON object like the following:
 
@@ -53,6 +51,8 @@ These endpoints respond with HTTP status `200 OK` on success or `503 Service Una
     ]
 }
 ----
+
+See the https://github.com/microprofile/microprofile-health/blob/main/spec/src/main/asciidoc/protocol-wireformat.asciidoc[wire format] for a more detailed description.
 
 == Enabling the health checks
 It is possible to enable the health checks using the build time option `health-enabled`:
@@ -96,6 +96,8 @@ The Containerfile `+HEALTHCHECK+` instruction defines a command that will be per
 ----
 
 The above code depends on the values of the {project_name} options such as `http-relative-path` (`http-management-relative-path`) and `http-management-port`. In case those are changed the code needs to be modified accordingly.
+
+NOTE: the `+HEALTHCHECK+` shown is based upon the ready endpoint. This is appropriate for informational purposes or to determine if the container should receive requests. It should not be used to determine if the container needs restarted.
 
 If you enable TLS as shown in <@links.server id="enabletls" />, the management interface will also use TLS. Depending upon how the management interface endpoints are intended to be used you can still have plain HTTP health checks if:
 
--- a/docs/guides/observability/health.adoc
+++ b/docs/guides/observability/health.adoc
@@ -19,7 +19,7 @@ When this option is `false` you should block unwanted external traffic to `/heal
 {project_name} exposes 4 health endpoints:
 
 * `/health/started` - Startup probe used for initial startup of {project_name} before the liveness probe takes over.
-* `/health/live` - Check if {project_name} is running. If this fails the application should discarded (terminated, shutdown) and started again.
+* `/health/live` - Check if {project_name} is running. If this fails the application should be discarded (terminated, shutdown) and started again.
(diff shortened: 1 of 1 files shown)


docs/guides/observability/health.adoc
```diff
@@ -18,12 +18,10 @@ When this option is `false` you should block unwanted external traffic to `/heal
 
 {project_name} exposes 4 health endpoints:
 
-* `/health/live`
-* `/health/ready`
-* `/health/started`
-* `/health`
-
-See the https://quarkus.io/guides/smallrye-health#running-the-health-check[Quarkus SmallRye Health docs] for information on the meaning of each endpoint.
+* `/health/started` - allows startup probes that are used for initial verification of the application before the liveness probe takes over. This is useful for applications which require additional startup time on their first initialization.
```
ahus1:
```suggestion
* `/health/started` - Startup probe used for initial startup of {project_name} before the liveness probe takes over.
```

docs/guides/observability/health.adoc
```diff
@@ -18,12 +18,10 @@ When this option is `false` you should block unwanted external traffic to `/heal
...
+* `/health/live` - if the application is running. If this procedure fails the application can be discarded (terminated, shutdown).
```
ahus1:
```suggestion
* `/health/live` - Check if {project_name} is running. If this fails the application should discarded (terminated, shutdown) and started again.
```

docs/guides/observability/health.adoc
```diff
@@ -18,12 +18,10 @@ When this option is `false` you should block unwanted external traffic to `/heal
...
+* `/health/ready` - if the application is ready to process requests or not. If this procedure fails the application may be made unavailable to serve new requests.
```
ahus1:
```suggestion
* `/health/ready` - Checks if {project_name} is ready to process requests or not. Only direct traffic to a newly started instance once the readiness probe succeeds. If this procedure fails, the application is temporarily unavailable to serve new requests, and traffic should be directed elsewhere.
```

docs/guides/observability/health.adoc
```diff
@@ -97,6 +97,8 @@ The Containerfile `+HEALTHCHECK+` instruction defines a command that will be per
 
 The above code depends on the values of the {project_name} options such as `http-relative-path` (`http-management-relative-path`) and `http-management-port`. In case those are changed the code needs to be modified accordingly.
 
+NOTE: the `+HEALTHCHECK+` shown is based upon the ready endpoint. This is appropriate for informational purposes or to determine if the container should receive requests. It should not be used to determine if the container needs restarted.
```
ahus1, line 100: Do we have information on how people use HEALTHCHECK? I tried the official Docker docs, but they state that a container is eventually considered "unhealthy", but didn't describe any actions.

ahus1 (dismissed):
Thank you for the documentation updates. Reading the health check description, it seems to take more the perspective of an application developer, while it IMHO should take the perspective of the person who wants to call them. And it could be a bit more specific to Keycloak. 

See below for some comments.

shawkins:
>  And it could be a bit more specific to Keycloak.

That all looks good. I mostly just did a light edit from the smallrye docs to start with, so I can understand wanting to make it sound more Keycloak centric.

docs/guides/observability/health.adoc
```diff
@@ -18,12 +18,10 @@ When this option is `false` you should block unwanted external traffic to `/heal
...
 {project_name} exposes 4 health endpoints:
 
-* `/health/live`
-* `/health/ready`
-* `/health/started`
-* `/health`
-
-See the https://quarkus.io/guides/smallrye-health#running-the-health-check[Quarkus SmallRye Health docs] for information on the meaning of each endpoint.
+* `/health/started` - Startup probe used for initial startup of {project_name} before the liveness probe takes over.
+* `/health/live` - Check if {project_name} is running. If this fails the application should discarded (terminated, shutdown) and started again.
```
kodebach:
```suggestion
* `/health/live` - Check if {project_name} is running. If this fails the application should be discarded (terminated, shutdown) and started again.
```

docs/guides/observability/health.adoc
shawkins: In plain Docker, it's just informational. Docker Compose seems to use them for service ordering [1]. Docker Swarm looks to go beyond and can be configured for restarts as well. The user who instigated the issue indicated that other orchestration platforms are configurable as well.

[1] From the docs with condition: service_healthy - "Compose waits for healthchecks to pass on dependencies marked with service_healthy. In the following example, db is expected to be "healthy" before web is created."

kodebach (commented):
LGTM much easier to understand now, if you just Google "Keycloak healthcheck" and find this page

docs/guides/observability/health.adoc
kodebach: > Docker Compose seems to use them for routing. 

Docker Compose doesn't do routing by itself. It's also just informal. But one could use [Traefik](https://doc.traefik.io/traefik/reference/install-configuration/providers/docker/) as a reverse proxy. That would respect the healthcheck for routing requests to a container.

There is also https://github.com/willfarrell/docker-autoheal. To use that correctly, the `HEALTHCHECK` should use the liveness probe.

vmuzikar (approved):
LGTM, thanks.

docs/guides/observability/health.adoc
shawkins: Thanks @kodebach updated the bit about Docker Compose in the comment above - the search summary led me astray.

shawkins:
@ahus1 do you want to review again based upon the clarification about how healthchecks are used?

ahus1:
@shawkins - I'm good with these changes, thanks!

PR #45016, opened 2025-12-18
Implement asynchronous logging when called from nonblocking threads

### LINKED ISSUE #43443: Keycloak should warn when ISPN or JGROUPS is running in debug level logging
### Description

When looking into #43367, we found several situations where debug logging for `org.jgroups` can deadlock the virtual threads.

There should be a warning in the docs, and a warning at Keycloak startup when that logging is enabled.

### Discussion

_No response_

### Motivation

Avoid deadlocks for users. 

### Details

One would probably need to parse `log-level` and `log-level-<category>` to see if there are problematic entries. 

Loggers are not initialized eagerly. Due to that, iterating over all loggers will not help.

### LINKED ISSUE #45015: Deadlock in Infinispan virtual threads
### Before reporting an issue

### Area

_No response_

### Describe the bug

As reported in https://github.com/keycloak/keycloak/issues/43508#issuecomment-3670695086 there can be a deadlock when there is a state transfer

### Version

26.4

### Regression

### Expected behavior

There should be no deadlocks

### Actual behavior

There are deadlocks. 

See https://github.com/keycloak/keycloak/issues/43508#issuecomment-3671935155 and https://github.com/keycloak/keycloak/issues/43508#issuecomment-3672199467 for the thread dumps and logs.

### How to Reproduce?

Issue frequent state transfers

### Anything else?

I'll attempt to create a workaround

ahus1:
Closes #45015
Closes #43443

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
@@ -31,6 +31,7 @@ import org.keycloak.connections.infinispan.InfinispanConnectionProvider;
 
 import io.quarkus.bootstrap.logging.InitialConfigurator;
 import io.quarkus.logging.LoggingFilter;
+import org.infinispan.commons.jdkspecific.ThreadCreator;
 
 /**
  * @author Alexander Schwartz
@@ -63,7 +64,7 @@ public final class KeycloakLogFilter implements Filter {
             }
         }
 
-        if (Thread.currentThread().getName().startsWith("non-blocking-thread")) {
+        if (ThreadCreator.isVirtual(Thread.currentThread())) {
             executor.submit(new RecordLogger(record));
             return false;
         }
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
@@ -45,7 +45,16 @@ public final class KeycloakLogFilter implements Filter {
 
     // Use this thread pool to asynchronously log from non-blocking threads, which could otherwise be pinned and lead to deadlocks.
     // A single thread ensures that all log entries appear in the correct order.
-    private final ExecutorService executor = Executors.newSingleThreadExecutor();
+    private final ExecutorService executor;
+
+    public KeycloakLogFilter() {
+        // The class ThreadCreator needs to be called and initialized here as when we do this in isLoggable() we'll have a recursive logging
+        if (ThreadCreator.useVirtualThreads()) {
+            executor = Executors.newSingleThreadExecutor();
+        } else {
+            executor = null;
+        }
+    }
 
     @Override
     public boolean isLoggable(LogRecord record) {
@@ -64,12 +73,9 @@ public final class KeycloakLogFilter implements Filter {
             }
         }
 
-        if (Boolean.getBoolean("org.infinispan.threads.virtual")) {
-            // Do not call ThreadCreator as it initializes the INSTANCE only once, and the code might set the system property only later.
-            if (ThreadCreator.isVirtual(Thread.currentThread())) {
-                executor.submit(new RecordLogger(record));
-                return false;
-            }
+        if (ThreadCreator.isVirtual(Thread.currentThread())) {
+            executor.submit(new RecordLogger(record));
+            return false;
         }
 
         return true;
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
@@ -27,7 +27,9 @@ import java.util.logging.LogRecord;
 import java.util.regex.Pattern;
 
 import org.keycloak.common.util.MultiSiteUtils;
+import org.keycloak.config.LoggingOptions;
 import org.keycloak.connections.infinispan.InfinispanConnectionProvider;
+import org.keycloak.quarkus.runtime.configuration.Configuration;
 
 import io.quarkus.bootstrap.logging.InitialConfigurator;
 import io.quarkus.logging.LoggingFilter;
@@ -36,8 +38,7 @@ import org.infinispan.commons.jdkspecific.ThreadCreator;
 /**
  * @author Alexander Schwartz
  */
-@LoggingFilter(name = "keycloak-filter")
-public final class KeycloakLogFilter implements Filter {
+public abstract class KeycloakLogFilter implements Filter {
 
     // avoid logging ISPN000312 for sessions, offlineSessions, clientSessions and offlineClientSessions caches only.
     private static final Pattern ISPN000312_PATTERN = Pattern.compile(
@@ -49,12 +50,22 @@ public final class KeycloakLogFilter implements Filter {
 
     public KeycloakLogFilter() {
         // The class ThreadCreator needs to be called and initialized here as when we do this in isLoggable() we'll have a recursive logging
-        if (ThreadCreator.useVirtualThreads()) {
+        if (ThreadCreator.useVirtualThreads() && isHandlerEnabled() && !isAsyncLoggingEnabled()) {
             executor = Executors.newSingleThreadExecutor();
         } else {
             executor = null;
         }
     }
+
+    /**
+     * Whether the logging is enabled for specific handler
+     */
+    public abstract boolean isHandlerEnabled();
+
+    /**
+     * Whether the async logging is enabled for specific handler
+     */
+    public abstract boolean isAsyncLoggingEnabled();
 
     @Override
     public boolean isLoggable(LogRecord record) {
@@ -73,7 +84,7 @@ public final class KeycloakLogFilter implements Filter {
             }
         }
 
-        if (ThreadCreator.isVirtual(Thread.currentThread())) {
+        if (executor != null && ThreadCreator.isVirtual(Thread.currentThread())) {
             executor.submit(new RecordLogger(record));
             return false;
         }
@@ -90,4 +101,42 @@ public final class KeycloakLogFilter implements Filter {
         }
     }
 
+    @LoggingFilter(name = "keycloak-filter-console")
+    private static final class KeycloakConsoleLogFilter extends KeycloakLogFilter {
+        @Override
+        public boolean isHandlerEnabled() {
+            return Configuration.isTrue(LoggingOptions.LOG_CONSOLE_ENABLED);
+        }
+
+        @Override
+        public boolean isAsyncLoggingEnabled() {
+            return Configuration.isTrue(LoggingOptions.LOG_CONSOLE_ASYNC);
+        }
+    }
+
+    @LoggingFilter(name = "keycloak-filter-file")
+    private static final class KeycloakFileLogFilter extends KeycloakLogFilter {
+        @Override
+        public boolean isHandlerEnabled() {
+            return Configuration.isTrue(LoggingOptions.LOG_FILE_ENABLED);
+        }
+
+        @Override
+        public boolean isAsyncLoggingEnabled() {
+            return Configuration.isTrue(LoggingOptions.LOG_FILE_ASYNC);
+        }
+    }
+
+    @LoggingFilter(name = "keycloak-filter-syslog")
+    private static final class KeycloakSyslogLogFilter extends KeycloakLogFilter {
+        @Override
+        public boolean isHandlerEnabled() {
+            return Configuration.isTrue(LoggingOptions.LOG_SYSLOG_ENABLED);
+        }
+
+        @Override
+        public boolean isAsyncLoggingEnabled() {
+            return Configuration.isTrue(LoggingOptions.LOG_SYSLOG_ASYNC);
+        }
+    }
 }
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
@@ -64,9 +64,12 @@ public final class KeycloakLogFilter implements Filter {
             }
         }
 
-        if (ThreadCreator.isVirtual(Thread.currentThread())) {
-            executor.submit(new RecordLogger(record));
-            return false;
+        if (Boolean.getBoolean("org.infinispan.threads.virtual")) {
+            // Do not call ThreadCreator as it initializes the INSTANCE only once, and the code might set the system property only later.
+            if (ThreadCreator.isVirtual(Thread.currentThread())) {
+                executor.submit(new RecordLogger(record));
+                return false;
+            }
         }
 
         return true;
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
@@ -18,7 +18,10 @@
 package org.keycloak.quarkus.runtime.logging;
 
 import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.logging.Filter;
+import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.regex.Pattern;
@@ -26,6 +29,7 @@ import java.util.regex.Pattern;
 import org.keycloak.common.util.MultiSiteUtils;
 import org.keycloak.connections.infinispan.InfinispanConnectionProvider;
 
+import io.quarkus.bootstrap.logging.InitialConfigurator;
 import io.quarkus.logging.LoggingFilter;
 
 /**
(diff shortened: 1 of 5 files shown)


ahus1:
@pruivo - please have a look when you have the time. Thanks!

shawkins:
@ahus1 the virtual thread issue was one of the reasons we added general support for async logging. Would you consider generally defaulting it to enabled when virtual threads are in use?

ahus1:
> Would you consider generally defaulting it to enabled when virtual threads are in use?

Maybe, maybe not. Async logging might leak into some other parts, still we could try to default to it once we are confident it works ok.

One thing that I'm concerned with is that the queue length is limited to 512 entries. Once that is reached, it will block, so I'd rather keep this suggested fix as it will not have a max queue length. 

I hope we can soon move to JDK 25 where this will no longer be a problem, as it doesn't have pinned threads. So once we use JDK 25, we wouldn't need to make async logging the new default. 

So TL;DR I'd be leaning towards not making async logging the default, at least not to fix the issue at hand. There might be other good reasons to still enable it.

shawkins:
> One thing that I'm concerned with is that the queue length is limited to 512 entries. Once that is reached, it will block, so I'd rather keep this suggested fix as it will not have a max queue length.

If you don't agree with the general default, changing that seems like a better approach and than adding special purpose async logging.

cc @mabartos

shawkins:
> One thing that I'm concerned with is that the queue length is limited to 512 entries. Once that is reached, it will block

Keep in mind that if it does block, that won't typically lead to starvation. The logging is being done on a dedicated thread, so as long as the queue can be worked down at all, then processing will resume.

ahus1:
> If you don't agree with the general default, changing that seems like a better approach and than adding special purpose async logging.

Well, for asynchronous logging it is generally a good idea have a fixed buffer, as otherwise you can run out of memory when you log faster than you can write to your log sinks. Still with our delicate pinned threads, we never want to get pinned, and therefore an unbounded queue for those entries might be the smaller of two evils here. 

> Keep in mind that if it does block, that won't typically lead to starvation. 

Let's hope for that, though I'm not sure about that yet. 

Maybe we can make async logging the default for an upcoming minor release. I assume async logging can then still be disabled?  

IMHO we still need a fix that can be backported to 26.2/26.4, as those are affected by this as well, and I don't think we would switch those to async logging by default in a patch release.

ahus1:
@shawkins - it seems that `CRDTest.testUpdateSpecValidation:110` is failing, which I think is unrelated. Any ideas?

shawkins:
> Let's hope for that, though I'm not sure about that yet.

The issue with these deadlocks / starvations is that a virtual thread is responsible for doing the logging work. If all virtual threads become pinned, the logging that needs to happen clear the lock isn't guarenteed.

With a dedicated thread logging work can always proceed.

> Maybe we can make async logging the default for an upcoming minor release. I assume async logging can then still be disabled?

Yes, an explicit configuration can override the default.

> IMHO we still need a fix that can be backported to 26.2/26.4, as those are affected by this as well, and I don't think we would switch those to async logging by default in a patch release.

No, but you can certainly direct users to enable log-async in these situations.

shawkins:
> @shawkins - it seems that `CRDTest.testUpdateSpecValidation:110` is failing, which I think is unrelated. Any ideas?

It is unrelated. It looks like we're somehow running a different api server version in CI, so the error message is slightly different. I put a change as a second commit in this pr https://github.com/keycloak/keycloak/pull/45010 - but it can put into a separate pr if needed.

ahus1:
> No, but you can certainly direct users to enable log-async in these situations.

I previously assumed it would happen only when DEBUG is enabled, but now we're seeing happening with INFO level logging, which concerns me. It IMHO means it could happen anytime, and "directing users to enable log-async in these situations" wouldn't be enough. 

We could choose to not use the code in this PR if async logging is enabled, but enable it when async logging is not enabled, virtual threads are enabled and JDK 25 is not used. 🙄

ahus1:
> I put a change as a second commit in this pr https://github.com/keycloak/keycloak/pull/45010 - but it can put into a separate pr if needed.

The failing CI might block other PRs as well. As long as #45010  is merged rather soon, I don't mind either way.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
```diff
@@ -55,6 +63,21 @@ public boolean isLoggable(LogRecord record) {
             }
         }
 
+        if (Thread.currentThread().getName().startsWith("non-blocking-thread")) {
```
pruivo: Most virtual threads do not have names, and JGroups vthreads do not start with `non-blocking-thread`. 
You probably want to use `org.infinispan.commons.jdkspecific.ThreadCreator#isVirtual(Thread)`

shawkins:
> > I put a change as a second commit in this pr #45010 - but it can put into a separate pr if needed.
> 
> The failing CI might block other PRs as well. As long as #45010 is merged rather soon, I don't mind either way.

Opened https://github.com/keycloak/keycloak/pull/45018

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
ahus1: Thank you, I was hoping someone would have implemented something like that to make the detection available independent of the JDK version! Applied in next commit - ready for one more review.

pruivo:
@ahus1, after sleeping over the subject, I have to give a :-1: to the PR. It makes more sense to disable vthreads if the core count is less than 8 or 16. With such a low core count, Keycloak does not generate enough load to overload the JGroups thread pool. We have been recommending that users disable it, and we haven't received any reports about JGroups thread pools or queues.

ahus1:
I would still like to give it a chance. Let's discuss next year.

mabartos (commented):
I'm not really sure about this current approach. It can bring some other problems to solve this one.
 
### 1. Async logging + this basic async logging enabled
So, when you have the async logging enabled, does it mean that there would be this new thread calling the AsyncHandlers? So, the execution would look like this: virt thread -> separated this new thread -> separated console thread? I don't know if we should allow this kind of hierarchy to workaround that problem. It'd be better to understand more about what's going on underneath -> in section 2. 

I'm not sure if that's what you intended based on the comments, but it should work like this now. If the intention was different, we could talk also about the separated queues for every log handler using async logging and the separation/order of those...

### 2. Why the deadlock?

Secondly, it'd be good to analyze what exactly is causing the deadlock - in what situations. In this [issue](https://github.com/keycloak/keycloak/issues/45015#issuecomment-3672358667), you mentioned:
 _"Assumed root cause: The logging is blocked in a virtual thread, and is postponed. Eventually some other threads that are pinned want to make progress on the same resource/logging, but they can't, so it is deadlocked."_

Assuming sync logging (default) and using virtual threads from the ISPN/JGroups side, publishing logs to slow handlers (like file, or syslog) should park the virtual thread and schedule the continuation, as you mentioned. So, the carrier thread (worker/OS/platform thread) should not be really blocked. It's blocked only when the virtual thread is pinned, so the whole worker thread is blocked - due to waiting for the inner reentrant lock release, or some hidden synchronized block on the ISPN/JGRoups side? So, are there still some places using the synchronized blocks that we're aware of on the ISPN/JGroups side? (I hope the assumption is correct)

> Eventually some other threads that are pinned want to make progress on the same resource/logging, but they can't, so it is deadlocked.

@ahus1 I'd like to understand it more. Can you elaborate on the situation why they can't make progress? AFAIK, the reentrant lock should be owned by the virtual thread, so the scenario might look like this: wants to log to slow handler -> scheduled continuation (postponed) -> other log wants to be written to the handler -> it cannot because of the handler lock -> once the first log is completed, the lock should be released. How does the deadlock occur?  

------------

> Would you consider generally defaulting it (async logging) to enabled when virtual threads are in use?

So, async logging would be turned on for all our handlers? Async logging also brings some tradeoffs, and enabling it by default might break users/customers assumptions - as some logs can be lost due to ungraceful shutdown, etc. 

> I hope we can soon move to JDK 25 where this will no longer be a problem, as it doesn't have pinned threads. So once we use JDK 25, we wouldn't need to make async logging the new default.

I'm not sure if it'll be really soon - AFAIK, there are still many issues on the Quarkus side that should be completed. And they're targeting the latest Quarkus versions, so I'm not sure if we have the Java 25 support near for the 3.27.x LTS. 

------------------------------------------
### Summary

TBH, maybe it'd be just better to have the warnings in the log detecting the ISPN/JGroups logging is set to track DEBUG, TRACE,...  (if that's the problem only for these fine-grained levels). And recommend to use the async logging to have enabled (in CLI, docs for HA,...)

But I don't have a really strong preference. 

cc: @ahus1 @shawkins @pruivo

ahus1:
Current progress on JDK 25 in upstream Quarkus, see https://github.com/quarkusio/quarkus/pull/51677

pruivo:
Hey @mabartos 

There are numerous `synchronized` blocks in Infinispan (and presumably in JGroups as well). I didn't consider that logging might cause deadlocks, and the following is happening:

1. A platform thread is logging and has the lock acquired
2. One or more virtual threads want to log something outside any `synchronized` block; they are unmounted and queued.
3. `numCores ` virtual threads arrive to log something inside a `synchronized` block; they will get pinned in the carrier thread. 
4. The platform thread releases the lock. 
5. If the lock is acquired by one of the virtual threads in (2.), the JVM will try to mount it, but no carrier thread is available; deadlock.

For transparency, I also suggested increasing the minimum number of cores to 4 or 8, here:
https://github.com/keycloak/keycloak/blob/706390adddc0a407591d6793603b1bf7dae66b38/model/infinispan/src/main/java/org/keycloak/infinispan/util/InfinispanUtils.java#L105-L109

shawkins:
> I didn't consider that logging might cause deadlocks

It was one of the drivers to add the general async logging support - https://github.com/keycloak/keycloak/issues/37266#issuecomment-2654344494 https://github.com/keycloak/keycloak/issues/37266#issuecomment-2714642293

My main point to @ahus1 here was that I'd rather refine that support instead of adding another async mechanism. 

However in reviewing what @pruivo wrote, I do agree with @ahus1's concern about any blocking by async logging can still be a problem. While the default async logging makes the situation far less likely, if all the carrier threads do become pinned then deadlock can still occur.

> For transparency, I also suggested increasing the minimum number of cores to 4 or 8, here:

Can you elaborate on why Keycloak is unable to generate this level of parallelism?

pruivo:
> Can you elaborate on why Keycloak is unable to generate this level of parallelism?

I'm sorry @shawkins, but I'm not sure if I understand your question. 

If you are asking about the `getParallelism()` method, it reads a JVM system property that configures the number of carrier threads (informally speaking). It defaults to the number of available cores, and Keycloak enables virtual threads when there are 2 or more cores.

shawkins:
> I'm sorry @shawkins, but I'm not sure if I understand your question.

No problem, I'm trying to understand the reasoning why 4 or 8 threads would matter to the deadlock situation. Does it become less likely at higher thread counts that those threads won't be simultaneously pinned, or are you trying to limit when virtual threads are used, or something else?

pruivo:
> Does it become less likely at higher thread counts that those threads won't be simultaneously pinned

Exactly, I'm playing with the odds here. The place where it happens, reported by the user, is when the state transfer completes; it logs it inside a synchronized block. The probability of 2 caches finishing at the same time is much higher than 4 or all 7 caches finishing the state transfer at the same time.

ahus1:
I'm switching this PR to "ready for review". Thank you for everyone who already commented on this. 

I consider this change a minimal fix that can be backported to 26.4/26.5 (possibly to 26.2 as well) with minimal changes to current users, and no additional docs on when virtual threads are used and when not, and how to monitor for example the  JGroup thread pool.  

Going forward, I agree that this is not ideal, and we should get rid of this change as soon as possible, and I would create a follow-up issue for that: 
* JDK 25 would be an option which would be available possibly in 26.6, and then we wouldn't need that change.
* Having async logging by default is IMHO an option even for 26.6 if we could argue it is a notable change (we never "promised" synchronous logging in our docs, didn't we :smiling_imp:?)

Please let me know your thoughts. Thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
```diff
@@ -55,6 +73,21 @@ public boolean isLoggable(LogRecord record) {
             }
         }
 
+        if (ThreadCreator.isVirtual(Thread.currentThread())) {
```
mabartos: @ahus1 Based on my findings, the block is unreachable based on the condition `ThreadCreator.isVirtual()`. So, that means for the ISPN logs, the new async logging is not used. 

In the log, you can find the thread names like `non-blocking-thread-*` (which should be the virtual threads) from where the calls to the LogFilter are made:

Image

Look at these watches: 

Image

You can see the `isVirtual()` does not consider the thread as virtual, and the block is unreachable. 

Even no additional thread is created based on this executor (the `pool-6-thread-1` is a thread created by the AsyncHandler for console -> `--log-console-async=enabled`):  

Image

Therefore, I don't think this approach currently solves the issue. Would it be possible to determine the thread based on other attributes, like f.e. name? Not sure how fragile it might be.

mabartos (commented):
@ahus1 I've just added one comment to this. How did you verify it works properly? I also have some other feedback, but I'll write it down in the next review comment.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
pruivo: The thread pool `non-blocking-thread` does not have virtual threads in Infinispan 15. You can see it on the watch; it says `Thread` and not `VirtualThread`. 

Edit:
Screenshot From 2026-01-07 11-18-13

mabartos: @pruivo Yeah, but AFAIK, we still use ISPN 15, right? So, you're debugging it with ISPN 16? Or can you please elaborate more?

mabartos:
### Problem
@ahus1 This pseudo-async logging does not really fit into scenarios on how we use the general async logging. The difference is not only in not using a queue, but also in how the additional logging thread is used. ATM, we have the 3 default handlers (console, file, syslog). When the `--log-async=true` is set, an additional 3 threads for logging for every handler are created. That's fine, because some handlers/sinks might be slower, and they should be independent of each other. The queue helps to create a backpressure on the calling thread. When no queue is present, it can run OOM, like for some connection issue with the remote syslog server. 

### Your approach
In this PR, there's only ONE thread created for all handlers as the KeycloakLogFilter is instantiated once and shared across handlers (general filter). It means that only this one thread takes care of all 3 handlers (figure below), so if there's some slow connection(or whatever) on the remote syslog server, the ISPN/JGroups logs are not written even to the console or file due to the slow connection on syslog. Even though the normal worker threads are free to make another work due to the virtual threads parking, the risk is that the additional logging thread runs OOM due to no backpressure and you can lost these logs. 

```
+----------+       +--------------------+
|  *main*  | ----> | *new-async-thread* |
+----------+       +--------------------+
                         |
                         |----> [ console handler ]
                         |
                         |----> [ file handler ]
                         |
                         |----> [ syslog handler ]
```

Another problem is with our global async logging for handlers. When the `--log-async=true` is set, these 3 additional threads for these handlers are created. It means, that the _main_ thread delegates the logging to the `new-async-thread` (the one with the executor here), which delegates the logging to the other thread for specific handler. I don't think we need this. There's also a probability (smaller than with sync logging) that the `new-async-thread` is swamped due to backpressure from these additional threads.

For illustration:
```
+----------+       +--------------------+
|  *main*  | ----> | *new-async-thread* |
+----------+       +--------------------+
                         |
                         |----> [ ASYNC console handler ]       ----> [ *CONSOLE-async-thread* ]
                         |
                         |----> [ ASYNC file handler ]          ----> [ *FILE-async-thread* ]
                         |
                         |----> [ ASYNC syslog handler ]        ----> [ *SYSLOG-async-thread* ]
```

### Suggestion
The best might be to use the async logging for everything and do not handle it on our own, but as we'd like to backport these changes, we need to consider it. 

We should consider if the async logging is enabled for specific handlers, and if yes, do not use your approach. 

It means that my current suggestion would be: 
- if the `async-enabled=true`, do not use your approach
- if the async logging is enabled for a specific handler, do not use your approach for the handler
- Create a separate thread for handling ISPN/JGroups for every handler (more info below)

```
--log=console,file,syslog --log-console-async=true

+----------+     
|  *main*  |
+----------+       
   |
   |----> [ ASYNC console handler ]       ----> [ *CONSOLE-async-thread* ]
   |
   |----> [ *new-FILE-async-thread* ] (ISPN/JGROUPS)
   |
   |----> [ *new-SYSLOG-async-thread* ] (ISPN/JGROUPS)
```

I'd say we should be close to the async logging principles for handlers and accept these separated threads for the ISPN/JGroups logs instead of having it managed by one thread for all handlers (but if these slow sinks/backpressure is not a problem, we can discuss)

Suggestions:  https://github.com/ahus1/keycloak/pull/435

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/logging/KeycloakLogFilter.java
pruivo: No, I'm debugging with ISPN 15 and this PR.
I've started a second node, clustered. Only JGroups uses virtual threads, and the screenshot is JGroups delivering a message.

mabartos: @pruivo Ahh, alright. Thanks for the explanation. Resolving my comment.

ahus1:
> This pseudo-async logging does not really fit into scenarios on how we use the general async logging.

@mabartos - I'd like to defer this discussion to a follow-up PR. This is about scenarios when there an occasional message from JGroups, and I don't want to make it bigger than essentially necessary. There might be some additional messages once someone decides to enable debug logging for JGroups, still I don't expect that to stall the system to an OOM.

ahus1:
@pruivo - I've updated the PR to include @mabartos's suggestions. Please re-review - Thanks!

ahus1:
The PR is finally green. @pruivo finally convinced me to increase the core limit to four as well, as with ISPN 15 one of the virtual threads is always pinned.

Please review when you have the time. Thanks!

pruivo (approved):
LGTM, thanks @ahus1

pedroigor (approved):

PR #45018, opened 2025-12-18
fix: updating test crdtest expectation

### LINKED ISSUE #45017: CRDTest failing in CI
### Description

The expected message for a CRDTest is slightly different due to picking up a different version of the api server.

shawkins:
closes: #45017

@ahus1 separated this change, so that there's no delay in getting it in.


ahus1 (approved):
Thank you for this fix!

PR #45043, opened 2025-12-19
[OID4VCI] CredentialBuildConfig ignores vc.credential_signing_alg

### LINKED ISSUE #45042: [OID4VCI] CredentialBuildConfig ignores vc.credential_signing_alg
### Before reporting an issue

### Area

oid4vc

### Describe the bug

Client scope attribute 'vc.credential_signing_alg' is ignored by CredentialBuildConfig, which always chooses the first available in the config. As a result, neither the holder wallet nor the issuer config can influence what signing algorithm is used.

### Version

latest

### Regression

### Expected behavior

respect the config

### Actual behavior

--

### How to Reproduce?

Set this attribute in the config and verify issued vc jwt header

### Anything else?

_No response_

tdiesler:
closes #45042

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialBuildConfig.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/model/CredentialBuildConfig.java
@@ -88,17 +88,24 @@ public class CredentialBuildConfig {
     public static CredentialBuildConfig parse(KeycloakSession keycloakSession,
                                               SupportedCredentialConfiguration credentialConfiguration,
                                               CredentialScopeModel credentialModel) {
-        final String credentialIssuer = Optional.ofNullable(credentialModel.getIssuerDid()).orElse(
-                OID4VCIssuerWellKnownProvider.getIssuer(keycloakSession.getContext()));
+
+        String modelIssuerDid = credentialModel.getIssuerDid();
+        String modelSigningKey = credentialModel.getSigningKeyId();
+        String modelSigningAlg = credentialModel.getSigningAlg();
+
+        String credentialIssuer = Optional.ofNullable(modelIssuerDid)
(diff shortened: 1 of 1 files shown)


IngridPuppet (commented):
Hello @tdiesler - Thank you for the fix. Could you add a test or an assertion to one of the existing tests to confirm this now works as expected?

tdiesler:
Sure, let me see if I can do that. For now, we are dealing only with this single line code change

```
Optional.ofNullable(modelSigningAlg)
                .orElse(credentialConfiguration.getCredentialSigningAlgValuesSupported().get(0))
```

mposolda:
+1 for the test (or assertion to an existing test).

One related concern: What happens when both "Signing key ID" is chosen "Signature algorithm", but the algorithm is different than the one of chosen "Signing Key ID" ? I guess "Signing key ID" should have a preference in case of this scenario? If yes, we may likely update the tooltip in the admin console to mention this.

At the same time, it is the question if "Signing key ID" option is really needed? Other Keycloak functionality (For example signining of regular OIDC ID-tokens or user-info responses) use only "Signature algorithm" option and always choose the first active key of the chosen algorithm. There is no option of "Signing key ID" anywhere else besides OID4VCI.

@IngridPuppet Do you recall the use-case for keeping "Signing key ID" option in the OID4VCI configuration?

tdiesler:
AFAICS, it would be ok to have a 1:N relationship like this: Issuer => Realm => Key[alg]
No need to have multiple keys for the same algorithm in a given realm (for now).

How the issuer would rotate keys is a different topic.

mposolda:
FYI. I've sent dedicated PR https://github.com/keycloak/keycloak/pull/45445 where I've included changes from this PR as well (not exactly same code, but similar) and it also includes automated tests.

mposolda:
Closing as this was incorporated by the other PR #45445

PR #45050, opened 2025-12-19
fix: preventing args to service commands

### LINKED ISSUE #44975: tools completion accepts any arguments
### Description

For example `./kc.sh tools completion --foo=dev-mem` does not produce an error. The command should actually not accept any property mapper / spi arguments, let alone someing completely invalid like --foo.

Also relates to new commands being added for windows services - they similarly don't need to accept any arguments not defined on the command.

See  https://github.com/keycloak/keycloak/pull/44496#discussion_r2627649298

shawkins:
also outputting all commands for help files

closes #44975

### DIFF
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
@@ -17,11 +17,14 @@
 
 package org.keycloak.it.cli.dist;
 
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 
 import org.keycloak.it.junit5.extension.CLIResult;
 import org.keycloak.it.junit5.extension.DistributionTest;
 import org.keycloak.it.junit5.extension.RawDistOnly;
+import org.keycloak.it.junit5.extension.WithEnvVars;
 import org.keycloak.it.utils.KeycloakDistribution;
 import org.keycloak.quarkus.runtime.Environment;
 import org.keycloak.quarkus.runtime.cli.command.BootstrapAdmin;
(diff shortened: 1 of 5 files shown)


shawkins:
@Pepo48 this reestablishes the replacement logic, but I see that more is still needed to enable the tests on windows.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
```diff
@@ -36,21 +39,24 @@
...
-import org.approvaltests.core.Options;
-import org.approvaltests.core.VerifyResult;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import static org.keycloak.quarkus.runtime.cli.command.AbstractAutoBuildCommand.OPTIMIZED_BUILD_OPTION_LONG;
 
+@WithEnvVars({"KEYCLOAK_ALL_COMMANDS", "true"})
```
vmuzikar: With this approach, we won't know the hiding of the OS-specific cocmmands actually works.

shawkins: Sorry, I meant to add a Linux specific test showing that these commands won't work there.

The main thought here is that we don't need helpcommanddisttest to be OS dependent.

I'd also be fine to extend this approach to a way to set the OS for picocli, rather than just relying upon the Environment method.

Pepo48:
> @Pepo48 this reestablishes the replacement logic, but I see that more is still needed to enable the tests on windows.

@shawkins correct, as I mentioned in the Windows service PR, this issue most likely has nothing to do with the changes introduced by the service feature. If the test had been enabled, it probably would have failed a long time ago. It’s a tricky one, because I couldn’t reproduce it locally. The raw GitHub logs show just a single trailing space (you can try searching the logs for "Usage: "), and I can’t directly download the received .txt files to check whether there’s anything more.

I tried to implement a [scrubber](https://github.com/approvals/ApprovalTests.Java/blob/master/approvaltests/docs/Scrubbers.md) using various regexes, but again - what worked locally didn’t work in CI. Therefore, I suspect there must be something else at play, possibly an environmental difference or a Picocli discrepancy.

As a result, simply adding the tag did not resolve the issue for me either, and it requires deeper analysis.

I’m now wondering whether someone noticed this a long time ago and whether this was the actual reason the tests were never enabled on Windows runners. 😄.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
vmuzikar: > The main thought here is that we don't need helpcommanddisttest to be OS dependent.

Not sure about that. How do we know the OS dependent logic for hiding/removing commands also works correctly for the help command?

shawkins: @vmuzikar updated the branch to show this more completely. We also have the option of moving the helpdisttest entirely out of the integration tests if we don't want to expose an environment variable for testing purposes.

vmuzikar: @shawkins Thanks for the update. I understand the reasoning and the push for unit test-like approach. It's alright, I think we can keep this approach for the helpdisttest. Still, I'd like to include at least a small integration smoke test that would verify our logic for excluding commands work on unsupported OSs. I'd feel better when merging if we had something like that here. :)

shawkins:
> I'd feel better when merging if we had something like that here. :)

Added a non-windows test to make sure that the command is not recognized.

> I’m now wondering whether someone noticed this a long time ago and whether this was the actual reason the tests were never enabled on Windows runners.

Right, I'm giving up on that aspect for this test. There isn't an obvious difference in the log with these changes to the help dist test, so maybe it can be reexamined under the other issue.

vmuzikar (approved):
Looks good, thanks @shawkins!

vmuzikar:
> so maybe it can be reexamined under the other issue.

+1, it'd be good to have the helpdisttest running on Windows but it's a separate issue.

shawkins:
@Pepo48 do you want to give a final review?

Pepo48 (approved):
LGTM, thanks @shawkins!

PR #45062, opened 2025-12-22
Add logout event to SessionResource

### LINKED ISSUE #44842: Publish an event for logging out a single session or all sessions or all sessions via the Account Console
IMHO the following should be a regular logout event, as it is also triggered by the user from account console:
   https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/services/src/main/java/org/keycloak/services/resources/account/SessionResource.java#L147-L154

   https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/services/src/main/java/org/keycloak/services/resources/account/SessionResource.java#L128-L135

robson90:
Closes: https://github.com/keycloak/keycloak/issues/44842

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/account/SessionResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/account/SessionResource.java
@@ -33,6 +33,8 @@ import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
 import org.keycloak.device.DeviceActivityManager;
+import org.keycloak.events.EventBuilder;
+import org.keycloak.events.EventType;
 import org.keycloak.models.AccountRoles;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.KeycloakSession;
@@ -56,12 +58,14 @@ public class SessionResource {
     private final Auth auth;
     private final RealmModel realm;
     private final UserModel user;
+    private final EventBuilder event;
 
-    public SessionResource(KeycloakSession session, Auth auth) {
+    public SessionResource(KeycloakSession session, Auth auth, EventBuilder event) {
         this.session = session;
         this.auth = auth;
         this.realm = auth.getRealm();
         this.user = auth.getUser();
+        this.event = event;
     }
 
     /**
@@ -129,8 +133,14 @@ public class SessionResource {
         auth.require(AccountRoles.MANAGE_ACCOUNT);
         session.sessions().getUserSessionsStream(realm, user).filter(s -> removeCurrent || !isCurrentSession(s))
                 .collect(Collectors.toList()) // collect to avoid concurrent modification as backchannelLogout removes the user sessions.
-                .forEach(s -> AuthenticationManager.backchannelLogout(session, s, true));
-
+                .forEach(s -> {
+                    AuthenticationManager.backchannelLogout(session, s, true);
+                    event.clone()
+                        .event(EventType.LOGOUT)
+                        .user(user)
+                        .session(s)
+                        .success();
+                });
         return Response.noContent().build();
     }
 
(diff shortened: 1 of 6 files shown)


services/src/main/java/org/keycloak/services/resources/account/SessionResource.java
```diff
@@ -129,8 +133,13 @@ public Response logout(@QueryParam("current") boolean removeCurrent) {
...
                 .collect(Collectors.toList()) // collect to avoid concurrent modification as backchannelLogout removes the user sessions.
-                .forEach(s -> AuthenticationManager.backchannelLogout(session, s, true));
-
+                .forEach(s -> {
+                    AuthenticationManager.backchannelLogout(session, s, true);
+                    event.event(EventType.LOGOUT)
+                        .user(user)
+                        .session(s)
+                        .success();
+                });
```
robson90, line 143: While implementing, I thought why not sending an event per logged out session. 

@ahus1 do you think that's the correct approach ?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/account/AccountRestServiceTest.java
```diff
@@ -1144,6 +1145,35 @@ public void testDeleteSessions() throws IOException {
         assertEquals(1, sessions.size());
     }
 
+
+    @Test
+    public void testDeletionOfAllUserSessionsWillFireEvents() throws IOException {
```
robson90:
```suggestion
    public void testDeletionOfAllUserSessionsWillFireLogoutEvents() throws IOException {
```

@ahus1 wdyt?

ahus1: +1 for `testDeletionOfAllUserSessionsWillFireLogoutEvents`

services/src/main/java/org/keycloak/services/resources/account/SessionResource.java
ahus1: +1 for having one event per logged out session, as an event is connected to a session. This way you can filter by session ID, and eventually see when this one session was logged out. 

One more thing here: the `.success()` should only be called once on an event. If you want to send multiple events, you should first call `.clone()`. 

BTW, in a follow-up issue, we might want to enforce that `send()` is called only once, and either use an assert there (which only fails in tests), or log a warning that the API is misused.

services/src/main/java/org/keycloak/services/resources/account/SessionResource.java
ahus1, line 161: I know this is still in draft mode. Still, don't forget about this one to also throw an event. Thanks!

robson90: Yeah got you. Just wanted to verify that it's the correct approach. 

Updated.

services/src/main/java/org/keycloak/services/resources/account/SessionResource.java
```diff
@@ -129,8 +133,14 @@ public Response logout(@QueryParam("current") boolean removeCurrent) {
...
-                .forEach(s -> AuthenticationManager.backchannelLogout(session, s, true));
-
+                .forEach(s -> {
+                    AuthenticationManager.backchannelLogout(session, s, true);
+                    event.clone()
+                        .event(EventType.LOGOUT)
+                        .user(user)
+                        .session(s)
+                        .success();
+                });
```
robson90, line 143: > One more thing here: the `.success()` should only be called once on an event. If you want to send multiple events, you should first call `.clone()`.

Like this ?

ahus1: Yes, exactly. There might be a grey area where the clone keeps the event type, still I'd keep it like this. Thanks!

robson90:
Hey @ahus1, I think its ready for review.

ahus1 (approved):

ahus1:
Thank you for the PR. It is approved and should auto-merge soon. Things can take a bit longer during PTO season :D

PR #45067, opened 2025-12-23
fix typo

### LINKED ISSUE #45066: Typo in default authentication flows
### Description

The description of "Account verification options" has a typo. It should read "Method with which to verify the existing account" rather than "Method with which to verity the existing account"

llamington:
Fix account verification options typo

Closes #45066


ahus1 (approved):
Thank you for this fix!

PR #45071, opened 2025-12-23
Extend documentation on ServiceMonitor creation

### LINKED ISSUE #45070: ServiceMonitor is not created by keycloak operator
### Before reporting an issue

### Area

operator

### Describe the bug

When creating a keycloak instance, no service-monitor is created although explicitly enabled (even though this is default anyway, and the documentation only describes how to disable it).

Digging into the code, I realized that the service monitor will only be created once `metrics-enabled` is configured as keycloak option.

### Version

v26.4.7

### Regression

### Expected behavior

After applying the information from the documentation, the service monitor is created.

### Actual behavior

Service monitor not created due to missing documentation.

### How to Reproduce?

Install keycloak through operator without setting `metrics-enabled`. Check prometheus.

### Anything else?

_No response_

JensErat:
The documentation so far only mentions how to disable the ServiceMonitor, leading to the assumption that it is enabled by default.

Getting a service monitor created also requires enabling metrics in keycloak, though:

https://github.com/keycloak/keycloak/blob/658faf210f2ff91ee254ff9fcc62ddfbcbcd9e42/operator/src/main/java/org/keycloak/operator/controllers/KeycloakServiceMonitorDependentResource.java#L50

While the missing setting creates a warning, the administrator should already have this information before applying the configuration.

Closes #45070

### DIFF
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -544,6 +544,21 @@ The {project_name} Operator automatically generates a `ServiceMonitor` resource 
 `ServiceMonitor` resource to be created, the `monitoring.coreos.com/v1:ServiceMonitor` Custom Resource Definition (CRD)
 must be installed on your {kubernetes} cluster.
 
+A ServiceMonitor is only created if the `metrics-enabled` option is set.
+
+.Enable Keycloak metrics
+[source,yaml]
+----
+apiVersion: k8s.keycloak.org/v2alpha1
+kind: Keycloak
+metadata:
+  name: example-kc
+spec:
+  additionalOptions:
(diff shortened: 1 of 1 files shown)


shawkins (approved):
Thanks for the PR @JensErat - this looks good from a cloud native perspective.

vmuzikar (approved):

ahus1:
cc: @ryanemerson - FYI

vmuzikar:
@ahus1 Sorry, I just noticed you self-assigned this PR. Let me know if you have any concerns about it and we can address them as a follow-up.

ahus1:
@vmuzikar - no worries, this is fine.

PR #45091, opened 2025-12-28
Implement automatic key rotation for realm keys

### LINKED ISSUE #11693: Automatic realm key rotation
### Description

The server administration guide describes how to rotate signing keys for tokens: https://www.keycloak.org/docs/latest/server_admin/#rotating-keys

It should be possible to automate this steps configure an interval for automatic key rotation.

Transferred from [KEYCLOAK-11607](https://issues.redhat.com/browse/KEYCLOAK-11607)

### Discussion

_No response_

### Motivation

Keys are usually not rotated if it is a manual process. Applications might not rely on JWKS resource and break,, when keys are rotated frequently.

### Details

_No response_

volck:
Adds scheduled task that automatically rotates realm signing and encryption keys based on configurable rotation periods. This addresses the security concern of long-lived keys and reduces operational burden.

Key features:
- Configurable rotation period per key provider
- Automatic expiration of passive keys after grace period
- Optional automatic deletion of disabled keys
- Prometheus metrics for monitoring rotation events
- Admin events for audit trail
- Support for RSA, ECDSA, HMAC, and AES key providers

The rotation task runs as part of the existing scheduled tasks infrastructure and is cluster-aware to prevent race conditions.

Closes #11693


volck:
@stianst I dont know who to tag, but is some one going to review this soon?

ahus1:
@volck - thank you for this PR. I see some emoji, it would be great if someone of those people would help to assess this PR. 

Some things I notice: 

* When setting an deletion time, keys need to be around so that they exceed the maximum idle time for any of the session types in the realm. Assuming that offline sessions could be idle for several days this should be validated. Maybe the deletion time could be derived from those settings instead leaving it to the user to configure it manually. 
* Please post some screenshots of your UIs as they are presented the user, this helps with the review. 
* There might be blocker as key rotation might not be work yet in all places. This needs at least a review - who can help? First possible issue identified: #46350

volck:
@ahus1 thanks for getting back to me! Please see attached screenshots:
this is adding a new provider:
image

 - this is from the "edit provider" in realmkeys:
image
...when a key is set to be rotated:

in console we get the following messages:

`2026-02-16 09:03:55,937 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation task: 1 providers with auto-rotation enabled, rotated=0, expired=0, deleted=0 keys in 3 ms, next run in 900 seconds
2026-02-16 09:18:55,981 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) ROTATING KEY for provider 'rsa-enc-generated' (id=f82ee2b3-120b-4a08-a3ac-51a15b8d9640) in realm 'master'
2026-02-16 09:18:59,088 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation activated: Created and configured new key provider 'rsa-enc-generated-1771229936000' (id=5fd75630-8a93-4777-8b4f-b0a730241b57) for realm 'master'
2026-02-16 09:18:59,090 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation task: 1 providers with auto-rotation enabled, rotated=1, expired=0, deleted=0 keys in 3154 ms, next run in 900 seconds
2026-02-16 09:33:56,012 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Disabled expired passive key provider 'rsa-enc-generated' in realm 'master' at 1771230836004
2026-02-16 09:33:56,021 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) ROTATING KEY for provider 'rsa-enc-generated-1771229936000' (id=5fd75630-8a93-4777-8b4f-b0a730241b57) in realm 'master'
2026-02-16 09:33:58,777 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation activated: Created and configured new key provider 'rsa-enc-generated-1771230836034' (id=b86572c1-a1f1-4a5b-bd7c-94d0014fc130) for realm 'master'
2026-02-16 09:33:58,781 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation task: 2 providers with auto-rotation enabled, rotated=1, expired=1, deleted=0 keys in 2843 ms, next run in 900 seconds
2026-02-16 09:48:55,938 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) ROTATING KEY for provider 'rsa-enc-generated-1771230836034' (id=b86572c1-a1f1-4a5b-bd7c-94d0014fc130) in realm 'master'
2026-02-16 09:48:56,851 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation activated: Created and configured new key provider 'rsa-enc-generated-1771231735955' (id=ffc54da8-77a0-4c4d-9a1e-97f33c828178) for realm 'master'
2026-02-16 09:48:56,863 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Disabled expired passive key provider 'rsa-enc-generated-1771229936000' in realm 'master' at 1771231736854
2026-02-16 09:48:56,865 INFO  [org.keycloak.services.scheduled.AutomaticKeyRotationTask] (Timer-0) Automatic key rotation task: 3 providers with auto-rotation enabled, rotated=1, expired=1, deleted=0 keys in 927 ms, next run in 900 seconds`


image

For the deletion/retention period: Ican try to derive the minimum passive key retention as max(ssoSessionMaxLifespan, ssoSessionMaxLifespanRememberMe, offlineSessionMaxLifespan, clientOfflineSessionMaxLifespan) plus a configurable safety margin, and validate user-configured values against this floor. Is this a good first stab?

As for #46350, could we expire the cookie by default when a key is outside of both deletion periods?

volck:
Added https://github.com/keycloak/keycloak/pull/45091/commits/b60f7ee675332f3407efaac0be2bf615d84a4a59  which derives the minimum passive key retention period from the realm's session timeout settings (max of all session types + a safety margin of 10%, minimum 1 hour). This prevents keys from being expired before long-lived sessions (e.g. offline sessions) have a chance to verify tokens signed with them. Added tests for the guardrail. If someone could verify I'd appreciate it.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,566 @@
...
+import org.keycloak.timer.ScheduledTask;
+
+import org.jboss.logging.Logger;
+
+/**
+ * Scheduled task for automatic key rotation.
+ * This task checks all key providers across all realms and rotates keys that are due for rotation.
+ * 
+ */
+public class AutomaticKeyRotationTask implements ScheduledTask {
```
slaskawi, line 62: Keycloak recently introduced Workflows (see `org.keycloak.models.workflow.WorkflowProvider` and dependent classes). I wonder using it here wouldn't be a better abstraction? It would enable users to customize the rotation to their own needs and it would eliminate handling the timer from this code. I suspect, the resulting implementation would be smaller and simpler. 

cc @stianst @pedroigor @sguilhen

slaskawi (commented):
I'm much in favor of this feature! Rotating keys becomes more and more important in highly secured production environments.

The UI part looks great and the code implementation looks fine too. 

I added one question, whether or not we'd like to implement this as a Workflow?

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
volck: Thanks for taking the time to review, much appreciated. I will look into using the workflow provider if that is easier and more flexible for the enduser. I dont know if the maintainers have any preference?

volck:
Thanks for taking the time to review, much appreciated. I will look into using the workflow provider if that is easier and more flexible for the enduser. I dont know if the maintainers have any preference?

volck:
@stianst @ahus1 anyone? :)

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,566 @@
...
+            
+        } catch (Exception e) {
+            logger.infov(e, "Failed to update time until rotation metric for provider '{0}'", provider.getName());
+        }
+    }
+
+    /**
+     * Checks if a key should be rotated based on the last rotation time and rotation period.
+     */
+    private boolean shouldRotateKey(ComponentModel provider, RealmModel realm) {
```
gh-baptiste-delafaire: `shouldRotateKey` has a hidden write side-effect
`shouldRotateKey` looks like a pure predicate, but it mutates persistent state when `lastRotationTime` is missing:

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,566 @@
...
+        long rotationPeriodSeconds = rotationPeriodStr != null ? Long.parseLong(rotationPeriodStr) : 7776000L; // Default: 90 days in seconds
+        long rotationPeriodMillis = java.util.concurrent.TimeUnit.SECONDS.toMillis(rotationPeriodSeconds);
+
+        String lastRotationTimeStr = provider.get(LAST_ROTATION_TIME_KEY);
+        if (lastRotationTimeStr == null || lastRotationTimeStr.trim().isEmpty()) {
+            // No rotation has happened yet - initialize the timestamp but don't rotate immediately
+            MultivaluedHashMap<String, String> config = new MultivaluedHashMap<>(provider.getConfig());
+            config.putSingle(LAST_ROTATION_TIME_KEY, String.valueOf(Time.currentTimeMillis()));
+            provider.setConfig(config);
+            realm.updateComponent(provider);
```
gh-baptiste-delafaire: Does `ClusterAwareScheduledTaskRunner`  protect against concurrent writes coming from other paths/pods ?  Do we have an optimistic lock here ?

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,566 @@
...
+    private static final String KEY_USE = "keyUse";
+    private static final String ALGORITHM_KEY = "algorithm";
+    private static final String KEY_SIZE_KEY = "keySize";
+    private static final String PASSIVE_KEY_EXPIRATION_KEY = "passiveKeyExpiration";
+    private static final String AUTO_DELETE_DISABLED_KEYS_KEY = "autoDeleteDisabledKeys";
+    private static final String DELETION_GRACE_PERIOD_KEY = "deletionGracePeriod";
+    private static final String DISABLED_TIME_KEY = "disabledTime";
+
+    @Override
+    public void run(KeycloakSession session) {
```
gh-baptiste-delafaire, line 122: This method looks a bit weird for me as the responsabilities are not clear inside each component, IMO we should consider splitting each component, for exemple `updateLastRotationMetric` and `shouldRotateKey` both read `LAST_ROTATION_TIME_KEY` and the second one mutates it. both share the same implicit state. 

Shouldn't we consider splitting this code in small steps with one concern. 
1 - We could create a small immutable DTO :
```java
record ProviderRotationState(
        String realmId,
        String realmName,
        String providerId,
        Long lastRotationTimeMillis, // null when missing
        long rotationPeriodSeconds,
        long passiveExpirationSeconds,
        [...]
) {
    static ProviderRotationState from(ComponentModel c, RealmModel r) { parse once, with try/catch }
}
```
2 - Make the predicate and the metric pure

```java
static boolean isDueForRotation(ProviderRotationState s, long now) {
    if (!s.autoRotationEnabled() || !s.active()) return false;
    if (s.lastRotationTimeMillis() == null) return false;
    long elapsedMs = now - s.lastRotationTimeMillis();
    return elapsedMs >= TimeUnit.SECONDS.toMillis(s.rotationPeriodSeconds());
}
private void publishLastRotationGauge(ProviderRotationState s) {
    if (!s.active() || s.lastRotationTimeMillis() == null) {
        removeGauge(s);
        return;
    }
    gaugeFor(s).set(s.lastRotationTimeMillis() / 1000);
}
```
Both functions are now side-effect free with respect to persistent state. They cannot drift from each other because they read from the same snapshot.

3 - Move the "init lastRotationTime" out of the hot loop

```java
private void initializeMissingRotationTimestamps(KeycloakSessionFactory factory, RealmModel realm) {
    realm.getComponentsStream(realm.getId(), KeyProvider.class.getName())
         .filter(c -> "true".equalsIgnoreCase(c.get(AUTO_ROTATION_ENABLED_KEY)))
         .filter(c -> isBlank(c.get(LAST_ROTATION_TIME_KEY)))
         .forEach(c -> KeycloakModelUtils.runJobInTransaction(factory, txSession -> {
             RealmModel txRealm = txSession.realms().getRealm(realm.getId());
             ComponentModel fresh = txRealm.getComponent(c.getId());
             if (fresh == null || !isBlank(fresh.get(LAST_ROTATION_TIME_KEY))) return; // re-check in tx
             MultivaluedHashMap<String, String> cfg = new MultivaluedHashMap<>(fresh.getConfig());
             cfg.putSingle(LAST_ROTATION_TIME_KEY, String.valueOf(Time.currentTimeMillis()));
             fresh.setConfig(cfg);
             txRealm.updateComponent(fresh);
         }));
}
```
The re-read inside the transaction provides the compare-and-set guarantee the original code was missing. Each init is its own short transaction, so a failure on one provider doesn't roll back the others.

4 - The new run() becomes a clean pipeline

```java
public void run(KeycloakSession session) {
    long now = Time.currentTimeMillis();
    long start = now;
    session.realms().getRealmsStream().forEach(realm -> {
        try {
            initializeMissingRotationTimestamps(session.getKeycloakSessionFactory(), realm);
            List<ProviderRotationState> snapshots = realm
                    .getComponentsStream(realm.getId(), KeyProvider.class.getName())
                    .map(c -> ProviderRotationState.from(c, realm))
                    .toList();
            // Pure: publish observability from the snapshot
            snapshots.forEach(this::publishLastRotationGauge);
            // Decide, then act. Each action runs in its own short transaction.
            for (ProviderRotationState s : snapshots) {
                if (s.autoRotationEnabled() && isDueForRotation(s, now)) {
                    rotateKeyTx(session.getKeycloakSessionFactory(), s);
                }
                if (!s.active() && shouldExpire(s, now, realm)) {
                    expirePassiveKeyTx(session.getKeycloakSessionFactory(), s);
                }
                if (shouldDelete(s, now)) {
                    deleteDisabledKeyTx(session.getKeycloakSessionFactory(), s);
                }
            }
        } catch (Exception e) {
            logger.errorv(e, "Automatic key rotation failed for realm '{0}'", realm.getName());
        }
    });
    log(start);
```

So the improvements are : 
- single source of truth per iteration, metric and decision both read identical data.
- no write inside a predicate
- Init, rotation, expiration and deletion each own their transaction, with re-check inside, removing the silent last-write-wins issue and shortening lock duration.
- A bad provider (e.g. malformed lastRotationTime) fails its own transaction and is logged; it doesn't poison the entire realm scan.
- The orchestration in run() is ~20 lines and reads as a sentence: snapshot → publish → decide → act.

gh-baptiste-delafaire (approved):
That a very important and very nice improvement, I can't wait to see this in Keycloak ! 
THANKS 🙏

gh-baptiste-delafaire:
Hi guys, do we have any ETA about a possible release ? It will be a very great change for keycloak

volck:
Thanks for your feedback @gh-baptiste-delafaire, very much appreciated. I'd like to address them but I'm a bit unsure what to do about it given maintainers have not said what they want. Once they do, I'll happily address them.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
slaskawi: I talked with @ahus1 about this PR and the workflows recently. We agreed that workflows might be more suitable for user-facing things (e.g. expiring accounts). This however, is a core-type of functionality. So using Scheduled Task is actually OK. 

I believe this tread can be resolved.

volck: I'll mark it as resolved. Thanks! 👍

eicki:
Is there a time window in which a new signing key will be available in passive state before rotation happens, some kind of pre-activation phase?
If clients are smart enough to fetch keys from the jwks endpoint on demand that is no issue, but if clients only fetch keys based on a cron job or even manual action is required, a pre-activation phase is needed.

volck:
> Is there a time window in which a new signing key will be available in passive state before rotation happens, some kind of pre-activation phase? If clients are smart enough to fetch keys from the jwks endpoint on demand that is no issue, but if clients only fetch keys based on a cron job or even manual action is required, a pre-activation phase is needed.

Seems like a great addition. The timewindow should probably be configurable? or should it be derived based on something else?

volck:
> Thanks for your feedback @gh-baptiste-delafaire, very much appreciated. I'd like to address them but I'm a bit unsure what to do about it given maintainers have not said what they want. Once they do, I'll happily address them.

got bored of waiting and (had some time) so I decided to try to address the issues you pointed out @gh-baptiste-delafaire, see if these makes sense. I think having separated concerns makes a lot of sense here.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
volck: Fixed in the refactor (`738a053`): the init write moved out of the predicate into `initializeMissingRotationTimestamps`, so `isDueForRotation` is now a pure, side-effect-free predicate.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
volck: Addressed in `738a053`: each mutation (init/rotate/expire/delete) now runs in its own short transaction that re-reads the component and re-checks the predicate before writing — the compare-and-set you suggested. The cluster runner (`executeIfNotExecuted`) already prevents concurrent task runs across nodes; this adds in-transaction CAS against concurrent admin edits. It's application-level CAS rather than a DB optimistic lock.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
volck: Done in `738a053`: introduced an immutable `ProviderRotationState` parsed once via `from()`. The metric (`publishLastRotationGauge`) and the decision (`isDueForRotation`) now read from the same snapshot so they can't drift, and `run()` reads as snapshot → publish → decide → act.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,700 @@
...
+        
+        // Copy algorithm if present
+        if (currentProvider.contains(ALGORITHM_KEY)) {
+            config.putSingle(ALGORITHM_KEY, currentProvider.get(ALGORITHM_KEY));
+        }
+        
+        // Copy key size if present
+        if (currentProvider.contains(KEY_SIZE_KEY)) {
+            config.putSingle(KEY_SIZE_KEY, currentProvider.get(KEY_SIZE_KEY));
+        }
```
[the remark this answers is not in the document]
volck: Fixed in `945322b`: rotation now copies all generation parameters via a `GENERATION_PARAM_KEYS` allowlist (`secretSize`, `ecdsa`/`ecdh`/`eddsaEllipticCurveKey`, `ecGenerateCertificate`, …). Added regression tests for HMAC secret size and ECDSA curve (P-521 no longer downgrades to P-256).

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,700 @@
...
+            for (ProviderRotationState state : snapshots) {
+                if (isDueForRotation(state, now) && rotateKeyTx(factory, state, now)) {
+                    counters.rotated++;
+                }
+                if (shouldExpirePassiveKey(state, now) && expirePassiveKeyTx(factory, state, now)) {
+                    counters.expired++;
+                }
+                if (shouldDeleteDisabledKey(state, now) && deleteDisabledKeyTx(factory, state, now)) {
+                    counters.deleted++;
+                }
```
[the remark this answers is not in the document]
volck: Fixed in `945322b`: each provider's actions — and the timestamp init — now run inside their own `try/catch` in `processRealm`, so one provider's failure is logged and no longer skips the rest of the realm.

services/src/main/java/org/keycloak/keys/Attributes.java
```diff
@@ -92,4 +92,27 @@ public interface Attributes {
...
             JWEConstants.RSA1_5, JWEConstants.RSA_OAEP, JWEConstants.RSA_OAEP_256);
 
+    String AUTO_ROTATION_ENABLED_KEY = "autoRotationEnabled";
+    ProviderConfigProperty AUTO_ROTATION_ENABLED_PROPERTY = new ProviderConfigProperty(AUTO_ROTATION_ENABLED_KEY, "Enable Automatic Rotation", "Enable automatic key rotation for this provider", BOOLEAN_TYPE, "false");
+
+    String ROTATION_PERIOD_KEY = "rotationPeriod";
+    ProviderConfigProperty ROTATION_PERIOD_PROPERTY = new ProviderConfigProperty(ROTATION_PERIOD_KEY, "Rotation Period (seconds)", "Number of seconds between automatic key rotations", STRING_TYPE, "7776000");
+
+    String PASSIVE_KEY_EXPIRATION_KEY = "passiveKeyExpiration";
+    ProviderConfigProperty PASSIVE_KEY_EXPIRATION_PROPERTY = new ProviderConfigProperty(PASSIVE_KEY_EXPIRATION_KEY, "Passive Key Expiration (seconds)", "Number of seconds to keep passive keys before they are disabled", STRING_TYPE, "2592000");
```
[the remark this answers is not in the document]
volck: Hardened at runtime in `945322b`: a non-positive `rotationPeriod` now falls back to the 90-day default (no more rotate-every-run), and a malformed value causes the provider to be skipped (logged) rather than mis-rotated. I kept this in the task rather than adding factory-level input validation — happy to add `validateConfiguration` checks to the key-provider factories as well if you'd prefer to reject bad values on create/update.

services/src/main/java/org/keycloak/keys/Attributes.java
```diff
@@ -92,4 +92,27 @@ public interface Attributes {
...
+
+    String LAST_ROTATION_TIME_KEY = "lastRotationTime";
+    // Note: This is typically set automatically by the rotation task, not manually configured
+    ProviderConfigProperty LAST_ROTATION_TIME_PROPERTY = new ProviderConfigProperty(LAST_ROTATION_TIME_KEY, "Last Rotation Time", "Timestamp (milliseconds) of the last automatic key rotation - set automatically", STRING_TYPE, null);
+
+    String AUTO_DELETE_DISABLED_KEYS_KEY = "autoDeleteDisabledKeys";
+    ProviderConfigProperty AUTO_DELETE_DISABLED_KEYS_PROPERTY = new ProviderConfigProperty(AUTO_DELETE_DISABLED_KEYS_KEY, "Auto-Delete Disabled Keys", "Automatically delete disabled keys after grace period expires", BOOLEAN_TYPE, "false");
+
+    String DELETION_GRACE_PERIOD_KEY = "deletionGracePeriod";
+    ProviderConfigProperty DELETION_GRACE_PERIOD_PROPERTY = new ProviderConfigProperty(DELETION_GRACE_PERIOD_KEY, "Deletion Grace Period (seconds)", "Number of seconds to wait after disabling a key before deleting it (default 3600 = 1 hour)", STRING_TYPE, "3600");
```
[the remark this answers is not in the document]
volck: The data-loss path is closed in `945322b`: a negative `deletionGracePeriod` never triggers deletion (`shouldDeleteDisabledKey` returns false), with a regression test. As with the rotation period, this is a runtime guard rather than factory-level validation — I can add the latter too if preferred.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,700 @@
...
+        // "Remember me" variants (0 means "use regular SSO value", so skip if 0)
+        if (realm.getSsoSessionMaxLifespanRememberMe() > 0) {
+            maxLifespanSeconds = Math.max(maxLifespanSeconds, realm.getSsoSessionMaxLifespanRememberMe());
+        }
+        if (realm.getSsoSessionIdleTimeoutRememberMe() > 0) {
+            maxLifespanSeconds = Math.max(maxLifespanSeconds, realm.getSsoSessionIdleTimeoutRememberMe());
+        }
+
+        // Offline session idle timeout — typically the largest (default 30 days)
+        maxLifespanSeconds = Math.max(maxLifespanSeconds, realm.getOfflineSessionIdleTimeout());
```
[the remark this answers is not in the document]
volck: Addressed in `9dffd35`: `computeMinimumPassiveKeyRetention` now also includes access-token and action-token lifespans (admin + user, including per-action overrides), not just session timeouts. Added a test that a passive key isn't expired while the action-token floor still covers it.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,700 @@
...
+                
+                // Serialize details as representation
+                try {
+                    adminEvent.setRepresentation(org.keycloak.util.JsonSerialization.writeValueAsString(details));
+                } catch (Exception e) {
+                    logger.infov(e, "Failed to serialize event details");
+                }
+                
+                // Always save with full details for system-generated security events
+                eventStore.onEvent(adminEvent, true);
```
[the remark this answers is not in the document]
volck: Fixed in `9dffd35`: `fireAdminEvent` now dispatches to the realm's configured and global `EventListenerProvider`s in addition to the store, and honors `isAdminEventsDetailsEnabled` — mirroring the normal admin-event path. Added a test using `@InjectAdminEvents`.

tests/base/src/test/java/org/keycloak/tests/keys/AutomaticKeyRotationTest.java
```diff
@@ -0,0 +1,671 @@
...
+            new AutomaticKeyRotationTask().run(session);
+            System.out.println("=== AFTER ROTATION TASK ===");
+        });
+
+        // Verify that NO new provider was created (rotation didn't happen)
+        List<ComponentRepresentation> afterProviders = realm.admin().components()
+                .query(realm.getId(), KeyProvider.class.getName())
+                .stream()
+                .filter(p -> "rsa-generated".equals(p.getProviderId()))
+                .filter(p -> p.getName().startsWith(keyProvider.getName())) // Check for our provider or rotated versions
```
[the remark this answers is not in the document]
volck: Tightened in `945322b`/`aec3269`: `testKeyRotationNotDueYet` now explicitly asserts no `rsa-generated-<timestamp>` replacement was created, so an unexpected rotation is detected.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,700 @@
...
+                // Register the gauge with Micrometer
+                List<Tag> tags = new ArrayList<>();
+                tags.add(Tag.of("realm", state.realmName()));
+                tags.add(Tag.of("provider", state.providerType()));
+                tags.add(Tag.of("name", state.providerName()));
+
+                Gauge.builder(KEY_LAST_ROTATED_METER_NAME, value, AtomicLong::get)
+                    .description("Unix timestamp (seconds) when key was last rotated")
+                    .tags(tags)
+                    .register(Metrics.globalRegistry);
```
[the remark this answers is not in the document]
volck: Fixed in `9dffd35`: gauges are tracked with their meter (`GaugeHandle`) and a `reconcileStaleGauges` step each scan removes meters for providers/realms that no longer exist, preventing unbounded stale cardinality. Added a test that deletes a provider and asserts the meter is removed.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,700 @@
...
+                    adminEvent.setRepresentation(org.keycloak.util.JsonSerialization.writeValueAsString(details));
+                } catch (Exception e) {
+                    logger.infov(e, "Failed to serialize event details");
+                }
+                
+                // Always save with full details for system-generated security events
+                eventStore.onEvent(adminEvent, true);
+            }
+        } catch (Exception e) {
+            logger.warnv(e, "Failed to fire admin event for key rotation in realm '%s'", realm.getName());
```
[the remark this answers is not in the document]
volck: Fixed in `945322b`: changed the `warnv` placeholder from `%s` to `{0}` so the realm name is included.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,735 @@
...
+        }
+    }
+
+    /**
+     * Pure predicate: is an active, auto-rotation-enabled key past its rotation period?
+     */
+    static boolean isDueForRotation(ProviderRotationState state, long now) {
+        if (!state.autoRotationEnabled() || !state.active()) {
+            return false;
+        }
```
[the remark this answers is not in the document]
volck: Fixed in `702f052`: `isDueForRotation` now also requires `enabled`, so an administratively disabled (`enabled=false`) provider is never rotated and can't be silently re-enabled. Added a regression test.

gh-baptiste-delafaire:
Thanks a lot, I hope that owner will approve this pr very soon ! looking foward to see it in production

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,1129 @@
...
+     * overwrites concurrent changes from another node.
+     */
+    private void processRealm(KeycloakSessionFactory factory, RealmModel realm, long now, RotationCounters counters) {
+        try {
+            // Compute the session/token-derived retention floor once per realm (it iterates the
+            // realm's clients) rather than once per provider, then reuse it for every snapshot.
+            long minimumPassiveRetentionSeconds = computeMinimumPassiveKeyRetention(realm);
+
+            // Snapshot: parse each provider's configuration exactly once into an immutable state object.
+            List<ProviderRotationState> snapshots = realm
```
slaskawi, line 167: This approach seems sub-optimal. We collect all the snapshots before checking if the auto-rotation is turned on. I think it would be better to flip the logic here - first check if it's turned on and **then** collect the snapshots.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,1129 @@
...
+     * Whether Keycloak metrics are enabled ({@code --metrics-enabled=true}). When disabled, this
+     * task must not create counters or gauges, matching the behavior of other metric producers such
+     * as {@code ExpirationTaskFactory} and {@code PasswordCredentialProviderFactory}.
+     */
+    private static boolean isMetricsEnabled() {
+        return Config.scope().root().getBoolean(METRICS_ENABLED_CONFIG_KEY, Boolean.FALSE);
+    }
+
+    // -------------------------------------------------------------------------
+    // Immutable snapshot + pure decision helpers
```
slaskawi, line 342: nit: Personally, I'm not a big fan of these comments. They strongly suggest that something should go to another file but for some reason didn't. Could we either remove it or ensure this class is split properly?

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,1129 @@
...
+     * A provider that an administrator has disabled (enabled=false) is never rotated, so rotation
+     * cannot silently re-enable disabled key material. A key that is itself a not-yet-promoted
+     * pre-activation key, or one that already has a pre-activation successor pending promotion, is
+     * also skipped so a single rotation does not spawn multiple replacements.
+     */
+    static boolean isDueForRotation(ProviderRotationState state, long now) {
+        if (!state.autoRotationEnabled() || !state.active() || !state.enabled()) {
+            return false;
+        }
+        if (state.activationTimeMillis() != null || state.preActivationPending()) {
```
slaskawi, line 455: Pre-activation marks the active predecessor as pending. If an administrator deletes the pending replacement before activation, no task path clears that predecessor flag, so this predicate skips the still-active predecessor forever.

model/storage-private/src/main/java/org/keycloak/services/scheduled/AutomaticKeyRotationTask.java
```diff
@@ -0,0 +1,1129 @@
...
+        return KeycloakModelUtils.runJobInTransactionWithResult(factory, txSession -> {
+            RealmModel realm = txSession.realms().getRealm(state.realmId());
+            if (realm == null) {
+                return false;
+            }
+            ComponentModel component = realm.getComponent(state.componentId());
+            if (component == null) {
+                return false;
+            }
+            ProviderRotationState fresh = ProviderRotationState.from(component, realm, computeMinimumPassiveKeyRetention(realm));
```
slaskawi, line 595: The task lock expires after the configured interval and is not renewed. If this run exceeds that TTL, two nodes can both pass this re-check against the same due provider and each add a replacement. Please add provider-level serialization (or a renewed lease) and cover an overlapping run.

slaskawi (commented):
I'd be definitely in favor of getting this feature into the main codebase! Automatic Realm Key rotation is a very robust method for satisfying various compliance requirements. 

The code looks generally OK to me (though I found a few things that could be improved). 

The biggest missing piece at the moment is documentation. We'd definitely need put information about this feature into the Server Admin Guide as well as Release Notes.

mabartos:
@sguilhen @pedroigor Workflows feature could be reused for this, right?

volck:
> @sguilhen @pedroigor Workflows feature could be reused for this, right?

according to https://github.com/keycloak/keycloak/pull/45091#discussion_r3137089133 this seems to conclude that it might be a scheduled task.

sguilhen:
@mabartos In theory yes, Workflows already has the pieces this would need (periodic condition-based rescan, and the `restart` step for a validated infinite loop). But it's not a clean fit today: `ResourceType` is a closed enum (only `USERS`/`CLIENTS`), and rotation's core action (spawning a new key component and handing off the "active" lifecycle to it) doesn't map to the existing step model, where a step always mutates the same resource it's tracking. Making that work would mean extending Workflows with a new resource type, event/schedule wiring, and solving the "step spawns a new resource" case, which is real design work, not a quick swap.

So for this case I'd suggest going with the scheduled task approach (same pattern as `ClearExpiredEvents), and revisit this if/when Workflows support other resource types.

PR #45099, opened 2025-12-30
Add logout event to UserSessionLimitsAuthenticator

### LINKED ISSUE #44843: Publish an event when the UserSessionLImitsAuthenticator terminates an older session
I think this should be a normal logout event as it is triggered by the user.

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java#L207-L209

robson90:
Closes: https://github.com/keycloak/keycloak/issues/44843

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
@@ -14,6 +14,8 @@ import org.keycloak.authentication.AuthenticationFlowError;
 import org.keycloak.authentication.AuthenticationFlowException;
 import org.keycloak.authentication.Authenticator;
 import org.keycloak.events.Errors;
+import org.keycloak.events.EventBuilder;
+import org.keycloak.events.EventType;
 import org.keycloak.models.AuthenticatorConfigModel;
 import org.keycloak.models.ClientModel;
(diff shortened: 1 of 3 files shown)


services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
```diff
@@ -179,7 +181,7 @@ private List<UserSessionModel> handleLimitExceeded(AuthenticationFlowContext con
 
             case UserSessionLimitsAuthenticatorFactory.TERMINATE_OLDEST_SESSION:
                 logger.info("Terminating oldest session");
-                var removedSessions = logoutOldestSessions(userSessions, limit);
+                var removedSessions = logoutOldestSessions(userSessions, limit, context.getEvent());
```
robson90, line 184: @ahus1 Im unsure if there is any better solution here.
We can also just use the context itself. 

Wdyt?

ahus1: This is a private method, so I'm ok to leave it as is, and change it later once we need a change.

ahus1 (approved):
Thank you for this change!

PR #45112, opened 2026-01-02
Add ability to delete offline sessions via account console

### LINKED ISSUE #15502: Ability to remove offline_access tokens from the Account Management client and/or Admin interface
### Description

I have found that offline_access refresh tokens (or their sessions rather) are not removed even if you sign out all sessions from the admin UI or from the user's Account management interface.

This seems like a gap in functionality: If a user signs in to a device/app to their account so that that device/app can have access to their account, it should be possible for that user to revoke that device's access from their management interface like they can with online sessions.

Perhaps I missed it?  Is there some permission I need to grant to the Account client?  I'd suppose I could, except not even admins can revoke offline sessions from the admin interface.  

Worse, I'm finding that non-expiring offline sessions just build up in the offline_user_session and offline_client_session tables.  (Devices/apps can be discarded, leaving no way for the user even to sign out from there.)

### Discussion

_No response_

### Motivation

Proper user and/or admin control over offline sessions
Cleanable cruft in the db.

Plenty of online services allow you to sign out devices under account management and don't require going to the physical device (which may not be in their possession any longer).

### Details

_No response_

robson90:
Closes: https://github.com/keycloak/keycloak/issues/15502

UI will look like this: 
Screenshot 2026-01-02 at 18 23 30
Screenshot 2026-01-02 at 18 23 34

CC: @ahus1

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_0.adoc
@@ -59,6 +59,11 @@ Additionally, both `IdentityProviderModel` and `IdentityProviderRepresentation` 
 configuration like `isHideOnLogin` to be null in order to not include these in Identity Provider types that do
 not need these configurations.
 
+=== Account UI shows and manages offline sessions
+
+The Account UI and its REST API now shows offline sessions of a user together with regular sessions in the *Device Activity* tab.
+A user can log out of those sessions like any other session, including the option *Sign out all devices*.
+
 === Realm default locale attempted for translation fallbacks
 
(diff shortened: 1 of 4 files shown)


ahus1:
@robson90 - thank you for this PR. I had a look and applied some refinement for a case when there is both an offline and a regular session with the same ID. Please have a look. 

If there are no objections from you, I'll go ahead and merge it for the upcoming 26.5 release.

Thanks!

ahus1 (approved):

robson90:
Hey @ahus1, works for me. Ive overlooked the part with "isCurrentSession". Good catch !

Thanks for merging!

PR #45124, opened 2026-01-04
Cleanup orphaned broker links when duplicate federated identities found

### LINKED ISSUE #44933: Broker links are not cleaned up for deleted users
### Before reporting an issue

### Area

identity-brokering

### Describe the bug

I'm not sure if this is also a problem when using internal user storage, but when using a federated user provider (e.g. LDAP) and identity providers mapped on those users a BrokerLinkEntity is created. These entities will remain there for ever. 

When the user is deleted (or user id is changed) in the federated user provider but a new user can be mapped from the IdP, then a new BrokerLinkEntity is made for the same broker_user_id. The first time this happens the user can successfully log in. The second time the user will get an error because there are 2 BrokerLinkEntity records for the same broker_user_id. This problem cannot be fixed without going into the database to remove the old BrokerLinkEntity.

### Version

26.3.3

### Regression

### Expected behavior

Expected behavior is that the user can also log in the second (and third) time after a new BrokerLinkEntity was registered. As the previous broker link entity no longer resolves to a user, it should be considered as gone.

### Actual behavior

Assume the following realm setup.
Federated user storage using LDAP.
An OAuth2 IdP, where users are matched based on the email address.

1. User1 with email address foo@example.org logs in via the IdP, a broker link entity is registered for broker_user_id foo
2. User1 is deleted from the LDAP
3. User2 is created in the LDAP with email address foo@example.org
4. Now the user logs in again via the IdP
5. UserStorageManager will try to lookup the user by federated identity in getUserByFederatedIdentity
6. it will find and user ID based on the existing BrokerLinkEntity. However... the getUserById will return null as the user does not exist anymore in LDAP
7. so a new new BrokerLinkEntity will (eventually) be created linking the broker_user_id with the new LDAP user
8. the second time the user logs in getUserByFederatedIdentity in JpaUserFederatedStorageProvider will fail as there are now 2 BrokerLinkEntity records

As User1 can no longer be found as a user you cannot delink the user from the previously registered IdP link.

### How to Reproduce?

See the actual behavior above.

### Anything else?

_No response_

Rathan-Naik:
When a federated user (e.g., from LDAP) is deleted from the external source, the broker link record in Keycloak is not cleaned up. If a new user with the same IdP mapping logs in later, a second broker link is created. On subsequent logins, the query returns multiple results and throws IllegalStateException.

### This fix:
In both JpaUserFederatedStorageProvider and JpaUserProvider, when multiple results are found:

- Iterate through each result
- Check if the user still exists using session.users().getUserById(realm, userId)
- If the user no longer exists (returns null), remove the orphaned broker link
- If multiple valid users exist, throw the IllegalStateException (this is a real data integrity issue)
- Return the single valid user

This approach cleans up stale data automatically while preserving the error for genuine duplicate scenarios.

Closes #44933

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
@@ -558,12 +558,37 @@ public class JpaUserProvider implements UserProvider, UserCredentialStore, JpaUs
         if (results.isEmpty()) {
             return null;
         } else if (results.size() > 1) {
-            throw new IllegalStateException("More results found for identityProvider=" + identity.getIdentityProvider() +
-                    ", userId=" + identity.getUserId() + ", results=" + results);
+            return resolveMultipleFederatedIdentities(realm, identity, results);
         } else {
             UserEntity user = results.get(0);
             return new UserAdapter(session, realm, em, user);
         }
+    }
+
+    private UserModel resolveMultipleFederatedIdentities(RealmModel realm, FederatedIdentityModel identity, List<UserEntity> results) {
+        UserEntity validUser = null;
+        for (UserEntity userEntity : results) {
+            if (userEntity.getFederationLink() != null) {
+                UserModel resolvedUser = session.users().getUserById(realm, userEntity.getId());
+                if (resolvedUser != null) {
+                    if (validUser != null) {
+                        throw new IllegalStateException("More results found for identityProvider=" + identity.getIdentityProvider() +
+                                ", userId=" + identity.getUserId() + ", results=" + results);
+                    }
+                    validUser = userEntity;
+                } else {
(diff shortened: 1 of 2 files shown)


vramik (approved):
@Rathan-Naik thank you for preparing the PR. LGTM.

@pedroigor since you're more familiar with the context, can you please also take a look?

ahus1:
@Rathan-Naik - Please avoid merging the main branch into the PR or rebasing the PR unless there are merge conflicts reported by GitHub or a reviewer asks you for a rebase. This prevents spamming reviewers with notifications and saves minutes on the GitHub actions CI. Thanks!

Rathan-Naik:
> @Rathan-Naik - Please avoid merging the main branch into the PR or rebasing the PR unless there are merge conflicts reported by GitHub or a reviewer asks you for a rebase. This prevents spamming reviewers with notifications and saves minutes on the GitHub actions CI. Thanks!

Noted.
Thank You

PR #45138, opened 2026-01-05
[OID4VCI] Make OID4VC credential offer storage cluster-aware and rename provider

### LINKED ISSUE #44674: InMemoryCredentialOfferStorage is not cluster-aware
### Description

There is in-memory implementation `InMemoryCredentialOfferStorage`, which works probably in single cluster node. This limitation should be mitigated to make sure that OID4VCI properly works in cluster similarly like any other Keycloak functionality.

Awambeng:
This PR renames and improves `InMemoryCredentialOfferStorage` to reflect its true cluster-aware behavior. 

## Changes:

- Replace `InMemoryCredentialOfferStorage` with `DefaultCredentialOfferStorage` backed by singleUseObjects (Infinispan)
- Fix expiration handling by converting absolute expiration timestamps to lifespan seconds and skipping already-expired entries
- Rename factory to `DefaultCredentialOfferStorageFactory` and change provider id to default
- Update SPI service registration and add JavaDoc clarifying cluster/cross-DC behavior

Closes #44674

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/DefaultCredentialOfferStorage.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/DefaultCredentialOfferStorage.java
@@ -19,23 +19,59 @@ package org.keycloak.protocol.oid4vc.issuance.credentialoffer;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


mposolda (approved):
@Awambeng Thanks for renaming the class and also for catching/fixing the lifespans!

PR #45171, opened 2026-01-06
Admin UI: slow response time listing second user page

### LINKED ISSUE #44860: Admin UI: slow response time listing second user page
### Before reporting an issue

### Area

admin/ui

### Describe the bug

I'm running Keycloak Benchmark with 30 million users in the database (Aurora PostgreSQL) (edit: all users belong to the same realm).
On the admin UI page, listing the first page of users is fast, independently of the number of users per page, but navigating to the second page is quite slow (28 seconds, see screenshot)

Image

### Version

26.4.7

### Regression

### Expected behavior

A more reasonable response time.

### Actual behavior

A very long response time.

### How to Reproduce?

Navigate to the user listing page and click on the next page.

### Anything else?

Aurora TOP SQL query at the time. It may help track the issue.

Image

Possible related to https://github.com/keycloak/keycloak/issues/39646, but I'm not using LDAP.

pruivo:
Fixes #44860

### Performance in local environment

I'm using a single Keycloak instance with PostgreSQL 17 running in Podman. I have a single realm and 1 million users. The time it takes to list the users is computed from the browser. 

**Baseline - main branch**

About 1.6 seconds.

Screenshot From 2026-01-06 11-32-19

**Only the `JpaUserProvider` optimization**

The reasoning behind this change is that `count distinct` is an expensive operation for a database and can be skipped if the query does not involve any joins with other tables. I got a 5x faster response time.

Screenshot From 2026-01-06 11-35-11

**This Pull Request**

I went a step further and modified `UserStorageManager` to minimize executing the count where possible. In my deployment, the count is skipped altogether. If LDAP or any other external user storage is available, the count is still invoked, but I "optimized" by avoiding invoking count on the last storage provider from the list :crossed_fingers:. 

It is around 50x faster :partying_face: 

Screenshot From 2026-01-06 11-39-35

### DIFF
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/JpaUserProvider.java
@@ -1062,7 +1062,6 @@ public class JpaUserProvider implements UserProvider, UserCredentialStore, JpaUs
                     break;
                 case UserModel.EXACT:
                     break;
-                // All unknown attributes will be assumed as custom attributes
                 case UserModel.INCLUDE_SERVICE_ACCOUNT: {
                     if (!attributes.containsKey(UserModel.INCLUDE_SERVICE_ACCOUNT)
                             || !Boolean.parseBoolean(attributes.get(UserModel.INCLUDE_SERVICE_ACCOUNT))) {
@@ -1071,6 +1070,7 @@ public class JpaUserProvider implements UserProvider, UserCredentialStore, JpaUs
                     break;
                 }
                 default:
+                    // All unknown attributes will be assumed as custom attributes
                     Join<UserEntity, UserAttributeEntity> attributesJoin = root.join("attributes", JoinType.LEFT);
                     if (value.length() > 255) {
                         customLongValueSearchAttributes.put(key, value);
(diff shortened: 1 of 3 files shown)


ahus1:
@pruivo - great work, and also great documentation to show the results. 

The comment from @sguilhen reminded me that there might be one more possible optimization: If you do a `count(...)` (but not a `count(distinct ...)` you can apply first a LIMIT as shown here: 

https://github.com/keycloak/keycloak/blob/bb971dc6fca50fbb82b14d2f33331144fd7ced0f/model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/CustomCreateIndexChange.java#L123-L125

So if one would rephrase the question from "give me the number of users" to "give me the number of users, but don't count more than X entries",  

At the same time, I think the results of this PR are good enough, and I'll be happy to fully review and merge it once the build is green. At the moment, it seems to be blocked by Maven central not returning artifacts reliably.

pruivo:
@ahus1 unfortunately, you cannot do `COUNT(*)` with the JPA query builder. I've not found a way to use it, but it reduces the query time by half.

In any case, that trick is slower with my dataset (am I doing it right?). Using the PgAdmin explain feature, I get the following response times (the first one is the query that Hibernate creates)

* ~120ms: `select count(u.id) from public.user_entity as u where u.service_account_client_link is null and u.realm_id = 'realm-0'`
* ~50ms: `select count(*) from (SELECT 1 from public.user_entity as u where u.service_account_client_link is null and u.realm_id = 'realm-0' limit 10)`
* ~45ms: `select count(*) from public.user_entity as u where u.service_account_client_link is null and u.realm_id = 'realm-0'`

ahus1:
@pruivo - thank you for the additional input. The SQL looks good, but explain vs the real execution might still differ. It might not make much of a different for 1 M users, maybe only when we go to 30 M users. 

Let's assume the current changes are good enough and proceed with PR as you originally suggested.

ahus1 (approved):
Hello Pedro thank you for the PR. I've reviewed it, and just moved one old comment and added one more comment about the "why" for the count. 

This PR looks good to me and ready to be merged. If you agree and mark it as ready-to-review, I'll be happy to merge it. Thanks!

PR #45183, opened 2026-01-06
Add indexes to BROKER_LINK table

### LINKED ISSUE #45009: Performance improvement: Missing indexes on BROKER_LINK table columns
### Description

Currently, the `BROKER_LINK` table lacks specific indexes required to optimize common lookups performed by the Keycloak JPA provider. Specifically, there are no indexes on the `user_id` column nor a composite index on `realm_id`, `identity_provider`, and `broker_user_id`.

As data grows, operations that link or look up users via Identity Providers (IdP) experience significant latency due to full table scans (Sequential Scans) in PostgreSQL and other relational databases.

### Value Proposition

Adding these indexes will:

- **Reduce latency** during login flows involving external Identity Providers.

- **Improve performance** for the User Storage SPI and Identity Brokering operations.

- **Decrease CPU and I/O usage** on the database server by avoiding full table scans on large `BROKER_LINK` tables.

### Goals

* Add an index on the `user_id` column to optimize `findBrokerLinkByUser`.

* Add a composite index on (`realm_id`, `identity_provider`, `broker_user_id`) to optimize `findUserByBrokerLinkAndRealm`.

* Ensure these indexes are included in the Liquibase changelogs for future releases.

### Non-Goals

* This enhancement does not aim to change the primary key structure of the table.

* It does not involve refactoring the JPA entities themselves, only the underlying schema performance.

### Discussion

_No response_

### Notes

The need for these indexes is evidenced by the following `@NamedQuery` definitions in the Keycloak source code, which currently lack backing indexes beyond the Primary Key:

1. **For user lookups**: `@NamedQuery(name="findBrokerLinkByUser", query="select link from BrokerLinkEntity link where link.userId = :userId")` Current state: Causes a scan if the `user_id` is not the first part of the PK (in some DBs) or if the table is large.

2. **For Brokering resolution**: `@NamedQuery(name="findUserByBrokerLinkAndRealm", query="select link.userId from BrokerLinkEntity link where link.realmId = :realmId and link.identityProvider = :identityProvider and link.brokerUserId = :brokerUserId") ` Current state: Requires a full scan to match the broker-specific identifiers.

**Proposed SQL (PostgreSQL example):**
```sql
CREATE INDEX IDX_BROKER_LINK_USER ON BROKER_LINK (USER_ID);
CREATE INDEX IDX_BROKER_LINK_FEDERATION ON BROKER_LINK (REALM_ID, IDENTITY_PROVIDER, BROKER_USER_ID);
``` 

In our environment (Keycloak 26.3.3 on PostgreSQL), we observed significant performance degradation during concurrent insertions and lookups that were resolved by manually adding these indexes.

ryanemerson:
Closes #45009

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -14,16 +14,10 @@ It also lists significant changes to internal APIs.
 
 === New database indexes on the `BROKER_LINK` table
 
-The `BROKER_LINK` table now contains two additional indexes `IDX_EVENT_BROKER_LINK_USER_ID` and `IDX_EVENT_BROKER_LINK_USER_FEDERATION`.
+The `BROKER_LINK` table now contains two additional indexes `IDX_BROKER_LINK_USER_ID` and `IDX_BROKER_LINK_IDENTITY_PROVIDER` to improve performance.
 
-`IDX_EVENT_BROKER_LINK_USER_ID` improves performance when querying Identity Providers associated with a given user.
-Whilst `IDX_EVENT_BROKER_LINK_USER_FEDERATION` improves performance when retrieving the {project_name} identity associated
-with the external Identity Provider identity.
-
-By default, neither of these indexes will be created automatically during migration
-if the number of entries in the `BROKER_LINK` table exceeds `300000`. See the link:{upgradingguide_link}[{upgradingguide_name}]
-for more details.
-
+If the table contains more than 300000 entries, {project_name} will skip the index creation by default during the automatic schema migration and instead log the SQL statement on the console during migration to be applied manually after {project_name}'s startup.
+See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to configure a different limit.
 
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -15,9 +15,14 @@ It also lists significant changes to internal APIs.
 === New database indexes on the `BROKER_LINK` table
 
 The `BROKER_LINK` table now contains two additional indexes `IDX_EVENT_BROKER_LINK_USER_ID` and `IDX_EVENT_BROKER_LINK_USER_FEDERATION`.
+
 `IDX_EVENT_BROKER_LINK_USER_ID` improves performance when querying Identity Providers associated with a given user.
 Whilst `IDX_EVENT_BROKER_LINK_USER_FEDERATION` improves performance when retrieving the {project_name} identity associated
 with the external Identity Provider identity.
+
+By default, neither of these indexes will be created automatically during migration
+if the number of entries in the `BROKER_LINK` table exceeds `300000`. See the link:{upgradingguide_link}[{upgradingguide_name}]
+for more details.
 
 
 // ------------------------ Deprecated features ------------------------ //
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -1,0 +1,36 @@
+// ------------------------ Breaking changes ------------------------  //
+== Breaking changes
+
+Breaking changes are identified as those that might require changes for existing users to their configurations or applications.
+In minor or patch releases, {project_name} will only introduce breaking changes to fix bugs.
+
+=== <TODO>
+
+// ------------------------ Notable changes ------------------------ //
+== Notable changes
+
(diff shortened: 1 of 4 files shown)


ahus1:
@ruchikajha95 - can you please review this PR? The usual rule of thumb applies: If you schedule a 1:1 with the author, you can review any pull request.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -0,0 +1,36 @@
...
+== Notable changes
+
+Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
+It also lists significant changes to internal APIs.
+
+=== New database indexes on the `BROKER_LINK` table
+
+The `BROKER_LINK` table now contains two additional indexes `IDX_EVENT_BROKER_LINK_USER_ID` and `IDX_EVENT_BROKER_LINK_USER_FEDERATION`.
+`IDX_EVENT_BROKER_LINK_USER_ID` improves performance when querying Identity Providers associated with a given user.
+Whilst `IDX_EVENT_BROKER_LINK_USER_FEDERATION` improves performance when retrieving the {project_name} identity associated
```
ahus1: It would be good to add the usual paragraph about "the index is not created automatically if it has more than X entries".

model/jpa/src/main/resources/META-INF/jpa-changelog-26.6.0.xml
```diff
@@ -0,0 +1,32 @@
...
+    <changeSet author="keycloak" id="26.6.0-45009">
+        <createIndex tableName="BROKER_LINK" indexName="IDX_EVENT_BROKER_LINK_USER_ID">
+            <column name="USER_ID" type="VARCHAR(255)" />
+        </createIndex>
+
+        <createIndex tableName="BROKER_LINK" indexName="IDX_EVENT_BROKER_LINK_USER_FEDERATION">
+            <column name="REALM_ID" type="VARCHAR(36)" />
+            <column name="IDENTITY_PROVIDER" type="VARCHAR(255)" />
+            <column name="BROKER_USER_ID" type="VARCHAR(255)" />
+        </createIndex>
```
ahus1: While not required, it is a good practice to split the DDLs to two distinct changesets as this would make them re-startable. Otherwise if the first index creation succeeds, and the second one fails, Liquibase will on a rerun fail with the first index creation as it already exists. An index creation can fail for example if the table is busy, or when it times out.

ahus1 (dismissed):
Hello Ryan, see below for some suggested changes. Thanks!

model/jpa/src/main/resources/META-INF/jpa-changelog-26.6.0.xml
```diff
@@ -0,0 +1,34 @@
...
+        <createIndex tableName="BROKER_LINK" indexName="IDX_EVENT_BROKER_LINK_USER_ID">
+            <column name="USER_ID" type="VARCHAR(255)" />
+        </createIndex>
+    </changeSet>
+
+    <changeSet author="keycloak" id="26.6.0-45009-user-federation">
+        <createIndex tableName="BROKER_LINK" indexName="IDX_EVENT_BROKER_LINK_USER_FEDERATION">
+            <column name="REALM_ID" type="VARCHAR(36)" />
+            <column name="IDENTITY_PROVIDER" type="VARCHAR(255)" />
+            <column name="BROKER_USER_ID" type="VARCHAR(255)" />
```
pruivo, line 40: @ahus1, shouldn't the `BROKER_USER_ID` be the first column of the index? 

The PostgreSQL documentation says `the index is most efficient when there are constraints on the leading (leftmost) columns` (https://www.postgresql.org/docs/17/indexes-multicolumn.html)

ahus1: Looking at the docs, it seems you're citing from "query conditions that involve any subset of the index's columns". This index has been created for a query that uses all columns of the query, so IMHO this wouldn't apply here. 

At the same time, for some databases they might be more efficient when you put a the column with the highest number of different values first, as it might avoid less balanced indexes, and "hot" areas. Still this is an area of a lot of folklore and superstition, and it is hard to get it right across all databases.

ahus1:
@ryanemerson - I noticed that "EVENT" slipped somehow in the index name, maybe a copy-and-past thing. I also remembered a good practice to see if the index not already exists, as sometimes people want to create indexes in advance (like in earlier releases / ahead of the patch release). 

The text now includes less of the internal behavior, and I copied it 1:1 from an earlier release.

@pruivo - looking at the different delete queries in BrokerLinkEntity, there are some deletes for realm and realm + identityProvider. For those, the indexes in this change are in the correct order, that is: realm first, and identityProvider second. 

This PR is now good to be merged. Thank you for all the reviews on this one!

ahus1 (approved):

Issue #31705, opened 2024-07-29
Organizations: Assign Identity Provider to more than one organization.


osajdapawel:
### Description

Add possibility to link Identity Provider to multiple organization. 
At the moment Identity Provider can be linked only to one organization. It would be useful to link for example global Microsoft IdP to many organizations and filter the users based on company domain name.

### Motivation

At the moment if we want to use for instance Microsoft SSO within application and split users into organizations and allows user to authenticate with M365 account we need to create multiple the same IdProviders (one per organization). Then we assign IdProvider to organization and choose domain for org.

### Details

Would be interesting to discuss when we have possibility to link IdProvider to many organizations:

- show on login page
- redirect when domain matches (filter by domain: if we want more email domains team1.contoso.com, team2.contoso.com, IMO we could assing multiple times the same IdProvider to org and specify separate domains)

ggomarighetti:
In a B2B SaaS application using the Organizations feature to separate customers, if we use GitHub or Google as login methods, we must configure a separate Identity Provider (IdP) for each customer.

This is not only relevant to internal applications or workforce-related use cases but is also critical for end users of the service. This aligns with the direction outlined in issue [#30180](https://github.com/keycloak/keycloak/issues/30180) to evolve Keycloak towards a "Customer Identity and Access Management (CIAM) with a focus on Business-to-Business (B2B) use cases."

FredrikPAC:
Hi, great feature!
Just wanted to give my upvote.
We are a B2B SaaS looking at keycloak as a solution. We have umbrella organizations wanting to connect to our service, where the high level organization provide the idp while the sub level organizations are treated as tenants in our application based on email domain. Excited to follow the development on this.

wschomburg:
Hi, we have the same use case where we have multiple organizations using the same IDP.

bcmedeiros:
Is this currently planned?

wschomburg:
I started to create dedicated IDPs for each organization and with Terraform it worked quite well. Just a few lines of config and a new organization with a dedicated IDP were configured in Keycloak. They all use the same Azure multi-tenant app configuration. Then I realized that Azure doesn't support wildcards in the redirect URLs and I have to add every IDP redirect URL in Azure, and it is also limited to 255 URLs per Azure application. 
So, having separate IDPs with the same configuration for each organization is also a limited solution. I'll go for only one organization and IDP for all customers using our multi-tenant Azure application for login for now.

[...]

PR #45190, opened 2026-01-06
Add Wildcard Subdomain Matching Support for Organization Domains

### LINKED ISSUE #40026: Support subdomain matching for Organizations in IdP redirect logic
### Description

When using the Organizations feature in Keycloak 26.2.2, domain-based redirection to the correct Identity Provider (IdP) only works if the email domain matches exactly the configured domain in the organization settings.

This leads to problems in real-world scenarios where users log in with email addresses that include subdomains (e.g. user@sub.example.com) while the organization is configured for example.com. The domain lookup fails and Keycloak either shows a login error or throws a NullPointerException.

Support for subdomain matching (e.g. *.example.com) or suffix-based resolution (e.g. endsWith("onmicrosoft.com")) is currently not available.

### Discussion

_No response_

### Motivation

Subdomain matching would improve usability and make the Organizations feature suitable for:

Microsoft 365 tenants that use subdomains like tenant.onmicrosoft.com
Multi-tenant environments where customers use shared top-level domains
Cases where departments or subsidiaries have their own subdomain
Avoiding hard-to-maintain lists of every possible subdomain

This would also reduce login failures and improve user experience in diverse enterprise environments.

### Details

A potential solution could include:

Supporting wildcards in organization domain configuration (*.example.com)
Adding a fallback mechanism for partial matches (e.g., endsWith())
Allowing a checkbox in the UI to disable automatic domain-to-organization resolution for generic domains like gmail.com
We’re happy to contribute ideas or testing support if this feature is considered for a future release.

shaidar:
## Problem Statement

Closes #40026

This addresses this [issue](https://github.com/keycloak/keycloak/issues/40026). When using the Organizations feature in Keycloak, domain-based redirection to Identity Providers (IdPs) only works with exact domain matches. This creates problems in real-world scenarios where users authenticate with email addresses containing subdomains (e.g., `user@sub.example.com`) while the organization is configured with the base domain (`example.com`).

**Common use cases that were previously unsupported:**
- Microsoft 365 tenant subdomains: `user@contoso.onmicrosoft.com` when org is configured for `onmicrosoft.com`
- University departments: `user@cs.university.edu` when org uses `university.edu`
- Multi-tenant SaaS environments with customer-specific subdomains
- Corporate environments with regional or departmental subdomains

This resulted in authentication failures, NullPointerExceptions, or users being unable to access their organization.

## Solution

Added a new `matchSubdomains` boolean flag to organization domains that enables hierarchical subdomain matching. When enabled for a domain like `example.com`, it will match:
- ✅ `example.com` (exact match still works)
- ✅ `sub.example.com`
- ✅ `deep.sub.example.com`
- ✅ Any level of subdomain nesting

The feature is **opt-in per domain** and **backward compatible** - all existing domains default to exact matching.

### Architecture

The implementation follows a simple, performant approach:

```java
// Core matching logic
if (email.equals(domain)) return true;  // Exact match
if (domain.matchSubdomains) {
    return email.endsWith("." + domain);  // Subdomain match
}
return false;
```

### Files Modified

**1. Data Model Layer (3 files)**
- `server-spi/src/main/java/org/keycloak/models/OrganizationDomainModel.java`
  - Added `matchSubdomains` boolean field with default value `false`
  - Added new constructor overload maintaining backward compatibility
  - Added `isMatchSubdomains()` getter

- `core/src/main/java/org/keycloak/representations/idm/OrganizationDomainRepresentation.java`
  - Added `matchSubdomains` field to REST API representation
  - Added getter/setter methods

- `model/jpa/src/main/java/org/keycloak/models/jpa/entities/OrganizationDomainEntity.java`
  - Added `MATCH_SUBDOMAINS` column mapping
  - Added getter/setter for JPA entity

**2. Database Layer (2 files)**
- `model/jpa/src/main/resources/META-INF/jpa-changelog-27.0.0.xml` *(new)*
  - Liquibase changeset to add `MATCH_SUBDOMAINS` column with default `false`
  
- `model/jpa/src/main/resources/META-INF/jpa-changelog-master.xml`
  - Included new changelog

**3. Data Mapping Layer (3 files)**
- `model/jpa/src/main/java/org/keycloak/organization/jpa/OrganizationAdapter.java`
  - Updated `toModel()` to include `matchSubdomains` when converting entity to model
  - Updated `setDomains()` to persist `matchSubdomains` flag

- `model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java`
  - Enhanced `getByDomainName()` with hierarchical lookup:
    1. Try exact match first (fast path)
    2. Parse email domain into parent domains
    3. Check each parent domain for wildcard enabled
  - Example: `deep.sub.example.com` checks `example.com`, `sub.example.com`

- `server-spi-private/src/main/java/org/keycloak/models/utils/RepresentationToModel.java`
  - Updated `toModel()` conversion to include `matchSubdomains` field

**4. Business Logic Layer (3 files)**
- `services/src/main/java/org/keycloak/organization/utils/Organizations.java`
  - Added `domainMatches(String emailDomain, OrganizationDomainModel orgDomain)` helper method
  - Centralized matching logic used throughout the codebase
  - Updated `matchesOrganizationDomain()` to use new helper

- `services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java`
  - Updated `redirect()` method to use wildcard-aware domain matching
  - Enables automatic IdP redirect for subdomain emails

- `services/src/main/java/org/keycloak/organization/validator/OrganizationMemberValidator.java`
  - Updated `validateEmailDomain()` to support wildcard matching
  - Added `validateEmailDomainMatch()` helper method
  - Ensures subdomain emails pass validation when wildcard is enabled

### Database Schema

```sql
ALTER TABLE ORG_DOMAIN 
ADD COLUMN MATCH_SUBDOMAINS BOOLEAN DEFAULT FALSE NOT NULL;
```

### Unit Tests (1 new file)
Created `OrganizationDomainMatchingTest.java` with 8 comprehensive unit tests:
- Exact domain matching (default behavior)
- Wildcard subdomain matching
- Case insensitivity
- Real-world scenarios (Microsoft 365, universities)
- Null/empty value handling
- Edge cases (special characters, deep nesting)
- Backward compatibility verification

### Integration Tests (10 new tests)
Added to existing test classes:

**OrganizationTest.java:**
- `testDomainWithWildcardSubdomains()` - CRUD operations for wildcard flag

**AbstractBrokerSelfRegistrationTest.java:**
- `testRedirectToIdentityProviderWithWildcardSubdomainMatching()` - Subdomain redirect
- `testRedirectToIdentityProviderWithDeepSubdomain()` - Multi-level subdomain redirect
- `testNoRedirectWithoutWildcardSubdomainMatching()` - Verify disabled wildcard blocks subdomains
- `testExactDomainStillWorksWithWildcardEnabled()` - Exact match with wildcard enabled
- `testEmailDomainValidationWithWildcardSubdomains()` - Member creation with subdomain
- `testEmailDomainValidationRejectsSubdomainWithoutWildcard()` - Rejection when disabled
- `testDeepSubdomainValidationWithWildcard()` - Multi-level validation

All tests pass and follow Keycloak testing patterns.

### REST API

**Create organization with wildcard subdomain matching:**
```json
POST /admin/realms/{realm}/organizations
{
  "name": "Example Corp",
  "domains": [
    {
      "name": "example.com",
      "verified": true,
      "matchSubdomains": true
    }
  ]
}
```

**Update existing domain:**
```json
PUT /admin/realms/{realm}/organizations/{id}
{
  "domains": [
    {
      "name": "example.com",
      "verified": true,
      "matchSubdomains": true
    }
  ]
}
```

### Admin Console
The `matchSubdomains` field will be available in the organization domain configuration UI (requires separate UI changes).

### Behavior

**With `matchSubdomains: false` (default):**
- User `john@example.com` → Matches ✅
- User `john@sub.example.com` → Does not match ❌

**With `matchSubdomains: true`:**
- User `john@example.com` → Matches ✅
- User `john@sub.example.com` → Matches ✅
- User `john@team.dev.example.com` → Matches ✅

## Backward Compatibility

✅ **Fully backward compatible:**
- Existing domains automatically default to `matchSubdomains: false`
- No behavior changes for existing deployments
- Database migration adds column with safe default value
- Old API calls without `matchSubdomains` field continue to work
- Existing constructors maintained with new overloads

## Migration Notes

**For administrators:**
- Existing organizations are unaffected - all domains use exact matching
- To enable subdomain matching, update domain configuration via REST API or Admin Console
- Consider enabling for domains like `onmicrosoft.com` or organizational base domains
- Test thoroughly before enabling in production

**For developers:**
- New `OrganizationDomainModel` constructor available with `matchSubdomains` parameter
- Old constructors still work (default to `false`)
- Use `Organizations.domainMatches()` helper for custom domain matching logic

## Performance Considerations

- **Exact match**: Single database query (no performance impact)
- **Wildcard match**: Additional queries only when exact match fails
  - Parses domain into parent domains (`sub.example.com` → `example.com`)
  - Queries each parent domain sequentially
  - Short-circuits on first match
- **Optimization**: Most common case (exact match) remains fast
- **Caching**: Existing organization cache applies

## Security Considerations

✅ **Safe by default** - wildcard matching must be explicitly enabled
✅ **Domain validation** - existing domain validation rules still apply
✅ **No new attack vectors** - only matches legitimate DNS subdomain hierarchy
✅ **Verification required** - verified domains should be verified before enabling wildcards

## Related Issues

Resolves issues with:
- Microsoft 365 multi-tenant scenarios
- University/enterprise subdomain structures
- SaaS platforms with customer-specific subdomains
- Any environment where email addresses use subdomains of a managed domain

## Checklist

- [x] Code changes implemented
- [x] Unit tests added
- [x] Integration tests added
- [x] Database migration created
- [x] Backward compatibility maintained
- [x] No breaking changes
- [x] Documentation in code comments
- [x] All tests passing

### DIFF
--- a/docs/documentation/server_admin/topics/organizations/managing-identity-providers.adoc
+++ b/docs/documentation/server_admin/topics/organizations/managing-identity-providers.adoc
@@ -39,6 +39,9 @@ The identity provider you want to link to the organization. An identity provider
 Domain::
 The domain from the organization that you want to link with the identity provider.
 
+Excluded domains::
+A comma-separated list of domains to skip automatic redirection. You can use wildcard domains like `*.example.com` to exclude all subdomains of `example.com`.
+
 Hide on login page::
 If this identity provider should be hidden in login pages when the user is authenticating in the scope of the organization.
 
--- a/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
@@ -77,6 +77,8 @@ import static org.keycloak.models.UserModel.LAST_NAME;
 import static org.keycloak.models.UserModel.USERNAME;
 import static org.keycloak.models.jpa.PaginationUtils.paginateQuery;
 import static org.keycloak.organization.utils.Organizations.isReadOnlyOrganizationMember;
+import static org.keycloak.organization.utils.Organizations.resolveByDomain;
+import static org.keycloak.organization.utils.Organizations.validateDomain;
 import static org.keycloak.utils.StreamsUtil.closing;
 
 public class JpaOrganizationProvider implements OrganizationProvider {
@@ -233,18 +235,51 @@ public class JpaOrganizationProvider implements OrganizationProvider {
 
     @Override
     public OrganizationModel getByDomainName(String domain) {
-        TypedQuery<OrganizationEntity> query = em.createNamedQuery("getByDomainName", OrganizationEntity.class);
+        if (domain == null) {
+            return null;
+        }
+
+        String emailDomain = domain.toLowerCase();
+
+        validateDomain(emailDomain);
+
         RealmModel realm = getRealm();
+        TypedQuery<OrganizationEntity> query = em.createNamedQuery("getByDomainName", OrganizationEntity.class);
+
         query.setParameter("realmId", realm.getId());
-        query.setParameter("name", domain.toLowerCase());
+
+        List<String> domainPatterns = new ArrayList<>();
+
+        // Add exact match
+        domainPatterns.add(emailDomain);
+
+        query.setParameter("names", domainPatterns);
+
         try {
             OrganizationEntity entity = query.getSingleResult();
             return new OrganizationAdapter(session, realm, entity, this);
-        } catch (NoResultException nre) {
-            return null;
+        } catch (NoResultException ignore) {
         }
-    }
 
+        // Strip subdomains to check for parent wildcard domains
+        String[] parts = emailDomain.split("\\.");
+
+        // Also check for wildcard at the current level
+        domainPatterns.add("*." + emailDomain);
+
+        for (int i = 1; i < parts.length - 1; i++) {
+            String parentDomain = String.join(".", java.util.Arrays.copyOfRange(parts, i, parts.length));
+            // Check for both exact parent and wildcard parent
+            domainPatterns.add(parentDomain);
+            domainPatterns.add("*." + parentDomain);
+        }
+
+        query.setParameter("names", domainPatterns);
+
+        return resolveByDomain(query.getResultList().stream()
+                .map(entity -> getById(entity.getId())).filter(Objects::nonNull).toList(), emailDomain);
+    }
+    
     @Override
     public Stream<OrganizationModel> getAllStream(String search, Boolean exact, Integer first, Integer max) {
         CriteriaBuilder builder = em.getCriteriaBuilder();
--- a/core/src/main/java/org/keycloak/representations/idm/OrganizationDomainRepresentation.java
+++ b/core/src/main/java/org/keycloak/representations/idm/OrganizationDomainRepresentation.java
@@ -19,6 +19,12 @@ package org.keycloak.representations.idm;
 
 /**
  * Representation implementation of an organization internet domain.
+ * 
+ * <p>Supports pattern-based domain matching:
+ * <ul>
+ *   <li><code>example.com</code> - exact match only</li>
+ *   <li><code>*.example.com</code> - matches example.com and all subdomains</li>
+ * </ul>
  *
  * @author <a href="mailto:sguilhen@redhat.com">Stefan Guilhen</a>
  */
--- a/docs/documentation/server_admin/topics/assembly-managing-organizations.adoc
+++ b/docs/documentation/server_admin/topics/assembly-managing-organizations.adoc
@@ -4,6 +4,7 @@
 
 include::organizations/intro.adoc[leveloffset=+2]
 include::organizations/managing-organization.adoc[leveloffset=+2]
+include::organizations/managing-domains.adoc[leveloffset=+2]
 include::organizations/managing-attributes.adoc[leveloffset=+2]
 include::organizations/managing-members.adoc[leveloffset=+2]
 include::organizations/managing-groups.adoc[leveloffset=+2]
--- a/docs/documentation/server_admin/topics/organizations/managing-domains.adoc
+++ b/docs/documentation/server_admin/topics/organizations/managing-domains.adoc
@@ -1,0 +1,68 @@
+[id="managing-organization-domains_{context}"]
+
+= Managing organization domains
+
+The domain associated with an organization plays an important role in how
+organization members authenticate to a realm and how their profiles are validated.
+
+One of the key roles of a domain is to help to identify the organizations where a user is a member. By looking at their
+email address, {project_name} will match a corresponding organization using the same domain and eventually change the
+authentication flow based on the organization requirements.
+
+The domain also allows organizations to enforce that users are not allowed to use a domain in their emails
+other than those associated with an organization. This restriction is especially useful when users, and their identities,
+are federated from identity providers associated with an organization and you want to force a specific email domain for their email addresses.
+
+== Assigning domains
+
+Domains are assigned through the *Domains* setting when creating or editing an organization. An organization can have
+zero, one, or multiple domains.
+
+When no domain is assigned, members are not validated against domain restrictions and email-based redirection to
+identity providers is unavailable.
+
+== Domain types
+
+Exact::
+Matches only the specified value. `example.com` matches `user@example.com` but not `user@sub.example.com`.
+
+Wildcard::
+Prefix the base domain with `*.` to match the base domain and all its subdomains at any depth.
+`*.example.com` matches `user@example.com`, `user@sub.example.com`, and `user@deep.sub.example.com`.
+
+== Mapping email addresses to organizations
+
+{project_name} extracts the domain from the email address, normalizes it to lowercase, and selects the best-matching
+organization using the following rules:
+
+. Matching is case-insensitive.
+. The most specific match wins, measured by the number of domain parts. For example, when resolving `user@deep.sub.example.com`:
+  * `deep.sub.example.com` (4 parts, exact) wins over `*.sub.example.com` (4 parts, wildcard) and `*.example.com` (3 parts, wildcard).
+  * `*.sub.example.com` (4 parts) wins over `*.example.com` (3 parts).
+. When two domains have the same number of parts, an exact match wins over a wildcard.
+
+The same rules determine which domain is reported in token claims such as the `domain` claim from the
+organization membership mapper.
+
+== Validation rules
+
+* A domain cannot be shared across organizations in the same realm.
+* Domains are stored and compared in lowercase.
+* A domain must have at least 2 parts (e.g. `example.com`). Single-part domains such as `com`, `org`, or `*.com` are rejected.
+* A domain cannot have more than 10 parts.
+* The `*.` wildcard prefix may appear only once and only at the start of the value (e.g. `sub.*.example.com` is rejected).
+
+== Considerations
+
+Prefer the most specific value::
+Use an exact domain for a single, well-known address. Use a wildcard when new subdomains may appear over time and should
+all map to the same organization.
+
+Overlapping domains across organizations are allowed but resolved deterministically::
+An organization may own `*.example.com` while another owns `sub.example.com`. Only one organization is returned per lookup,
+always the most specific match, with exact winning over wildcard at equal specificity. Verify the intended mapping
+before relying on overlapping configurations.
+
+Use exact domains to model exceptions::
+An organization can hold both `*.example.com` and `vip.example.com`. Addresses under `vip.example.com` resolve to the
+exact entry; all other subdomains fall back to the wildcard.
--- a/docs/documentation/server_admin/topics/organizations/managing-organization.adoc
+++ b/docs/documentation/server_admin/topics/organizations/managing-organization.adoc
@@ -59,16 +59,6 @@ Once you create an organization, you can manage the additional settings that are
 * <<_managing_groups_,Manage groups>>
 * <<_managing_identity_provider_,Manage identity providers>>
 
-== Understanding organization domains
-
-When managing an organization, the domain associated with an organization plays an important role in how
-organization members authenticate to a realm and how their profiles are validated.
-
-One of the key roles of a domain is to help to identify the organizations where a user is a member. By looking at their email address, {project_name} will match a corresponding organization using the same domain and eventually change the authentication flow based on the organization requirements.
-
-The domain also allows organizations to enforce that users are not allowed to use a domain in their emails
-other than those associated with an organization. This restriction is especially useful when users, and their identities, are federated from identity providers associated with an organization and you want to force a specific email domain for their email addresses.
-
 == Disabling an organization
 
 To disable an organization, toggle *Enabled* to *Off*.
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -3829,3 +3829,5 @@ role_offline-access=Offline access
 role_uma_authorization=Obtain permissions
 scimApiEnabled=SCIM API
 scimApiEnabledHelp=If enabled, exposes realm resources through an API based on the System for Cross-domain Identity Management (SCIM) specification, namely RFC7643 and RFC7644.
+excludedDomains=Excluded domains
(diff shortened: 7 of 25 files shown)


ahus1:
@shaidar - thank you for the PR. Can you please update the PR contain only contain your commits? This can probably be done by rebasing it. 

Once you have done that, there is no need for further rebases until a reviewer asks for you for a rebase, or GitHub reports merge conflicts.

ahus1:
@keycloak/core-iam - can you please have a look at this PR? Thanks!

model/jpa/src/main/resources/META-INF/jpa-changelog-27.0.0.xml
```diff
@@ -0,0 +1,30 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.2.xsd">
+
+    <changeSet author="keycloak" id="27.0.0-org-domain-wildcard">
```
sguilhen: I would just change this to 26.6.0 (along with the filename), as Keycloak 27 won't be released in the near future.

sguilhen:
@ahus1 changes look ok to me, apart from the jpa changeset version that I would switch from 27.0.0 to 26.6.0. We can trigger the CI to see if there's anything else that needs to be taken care of

sguilhen:
@shaidar looks like there's a compilation failure - can you take a look and update the PR?

pedroigor (commented):
@shaidar As you mentioned, we need the setting available from the UI. I can work on it if you are not doing it already.

sguilhen:
@shaidar build is failing due to wrong import order in one of the files. Can you run `mvn spotless:apply` to fix the imports and update the PR? Thanks!

pedroigor:
Added a commit to fix the code format errors and changes to the liquibase changelog.

pedroigor:
@shaidar Sent more changes:

* Fixing issue with caching domains as they were not invalidated when matching a subdomain
* Improving querying domains so that we have a single query to match orgs by domain
* Adding validations to domains so that we prevent anorg from using a subdomain, as a domain, from another org. I'm not sure if what I did is enough.
* Fixing tests

ssilvert:
> @shaidar As you mentioned, we need the setting available from the UI. I can work on it if you are not doing it already.

@shaidar @pedroigor 

For the UI, the simplest approach might be to add a new tab for domains.  The current MultiLineInput component does not support anything other than an array of strings.

Also, I think I discovered a nasty bug.  From the code, it looks like we always save the domains with the verified flag set to `false`.  So if a domain was verified and someone adds a new domain to the organization,  saving will send a `PUT` with all domains setting verified to `false`.  I haven't tested this, so maybe the back end ignores what the UI is sending?

Adding a new tab for domains would likely fix this problem.

shaidar:
@ssilvert Thanks for the input and feedback. Any chance you have the cycles to make that setting available from the UI? I'm not very familiar with that and would most likely take a bit to make that happen.

ssilvert:
> @ssilvert Thanks for the input and feedback. Any chance you have the cycles to make that setting available from the UI? I'm not very familiar with that and would most likely take a bit to make that happen.

I think we can schedule some time for me to do it.  If I'm right about the bug it will be imperative that we take care of this anyway.

@pedroigor WDYT?

pedroigor:
@ssilvert I confirm the issue with the `verified` property when updating the organization. Not so critical, though, because we do not have any functionality around domain verification yet. Looks like the UI is always sending `false`.

Regarding the UI for the new setting, we can talk about it. It will probably mean changing how we are managing domains in the UI ...

shaidar:
@ssilvert @pedroigor The other thing that might be worth adding to this as it's a use case that applies to us is the ability to exclude some subdomains. For example, use the wildcard for the domain, but be able to add a few subdomains to an exclusion list.

pedroigor:
@shaidar I was wondering if we should consider a different approach for subdomain matching where we just use an `*` before the domain name. For instance:

```
*.mydomain.org
```

If a domain starts with `*`, subdomain matching is implicit. It looks simpler (no changes to the database) and aligned with how we handle other types of URIs such as redirect URIs for clients.

> For example, use the wildcard for the domain, but be able to add a few subdomains to an exclusion list.

With the approach above, we could parse a domain `-.sub.mydomain.org` as an organization domain but exclude it from redirection?

shaidar:
@pedroigor you want to take a look and see if the changes make sense to you as I think that's my understanding of what you've proposed as far as handling exclusions.

pedroigor:
@shaidar Sorry for the delay. 

> @pedroigor you want to take a look and see if the changes make sense to you as I think that's my understanding of what you've proposed as far as handling exclusions.

If you agree about this alternative to handle subdomain mathing and exclusion, I vote for doing it.

As I mentioned, this approach is aligned with how we handle redirect uris for clients where you have special characters such as `*`, `+`, etc, to indicate how they are handled during validation.

It will also provide a much simpler solution to the original problem without require changes to both UI and database.

I would appreciate it if you could make these changes, we are really busy with planned tasks for 26.6. I think this is a great addition to organizations, if you cannot make it, I will try my best to do it and have it in 26.6.

pedroigor:
@shaidar Thanks! I'm rebasing your PR, adding more tests, docs, and fixing spotless errors.

I'm also wondering if we should handle exclusion domains differently, as they are mainly used to disable automatic redirection for certain domains. We keep supporting wildcard domains as you did, but for exclusion domains, we use instead an additional configuration option to the broker settings (in the scope of the org) to set the exclusion domain, using the `-.` prefix.

Does it make sense to you? It is a bit odd to set a domain like `-.sub.example.com` where this domain actually maps to the organization but it just disables auto-redirect.

docs/documentation/server_admin/topics/organizations/managing-identity-providers.adoc
```diff
@@ -39,6 +39,9 @@ The identity provider you want to link to the organization. An identity provider
 Domain::
 The domain from the organization that you want to link with the identity provider.
 
+Excluded domains::
```
pedroigor, line 42: Added a new option to allow setting a comma-separated list of domains to exclude from auto-redirect.

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
```diff
@@ -231,18 +231,64 @@ public OrganizationModel getById(String id) {
...
+        List<String> domainPatterns = new ArrayList<>();
+
+        // Add exact match
+        domainPatterns.add(emailDomain);
+
+        query.setParameter("names", domainPatterns);
+
         try {
             OrganizationEntity entity = query.getSingleResult();
             return new OrganizationAdapter(session, realm, entity, this);
```
pedroigor, line 260: We first try to find an exact match, as it has precedence over wildcard domains.

This is because you can have an organization with a `sales.example.com` domain and another organization with `*.example.org`. In this case, we should resolve the former and not the latter.

pedroigor (commented):
@shaidar Added another commit with the following changes:

* Excluded domains in broker settings. For now, excluded domains are only about skiping auto-redirect to specific subdomains within an org domain. In the future, we can also allow defining exclusion domains at the organization level so that these domains will force to not match to the organization
* Updated docs and added more tests
* There is one open point related to how to handle situations where a organization has a `sales.example.com` domain and another organization is created using `*.example.com`. In this case, should we fail the creation of the second org or we allow it so that domain matching will favor exact match over wildcard domains?

pedroigor:
The last commit is changing:

* Exclusion domains can only be set to skip auto-redirect based on a blacklist of domains. The UI was updated to allow setting the new broker setting. Exclusion domains are no longer valid when managing organization domains.
* Introduces a new mapper option to add a  `domain` claim to tokens with the domain (if any) that matched the user email domain during authentication. The idea is to provide more context for applications so that they know the exact domain from the organization that is mapped to the user. I think this should allow addressing use cases in the future where a domain can also change how users authenticate and how they are managed within an org.
* More tests

sguilhen (approved):

pedroigor (dismissed):

pedroigor:
@shaidar Now that 26.6 was released, we can resume discussions and merge this new capability.

One question I have after re-reviewing the PR is why not continue checking for conflicts based on an exact match when validating domains (not when fetching orgs by domain, this one is fine)?

If you have a OrgA with `example.com` and another OrgB with `*.example.com`, when a user authenticates using `user@example.com` we always resolve to OrgA because of the exact-matching-first policy. By doing that, we are going to allow administrators to map organizations to a specific sub-domain if they want to have users from that sub-domain in a different tenant.

Isn't it simpler and more flexible if we do it this way? I'm not sure if we should be so strict about sub-domains and base domains. If so, we are disallowing the use case above because a wildcard can make an organization the owner of a domain and all its sub-domains, with no option to have a separate organization for a specific sub-domain.

The use case I have in mind is a company where you have a base domain `*.example.com` and subsidiaries that share the same base domain such as  `s1.example.com` and `s2.example.com`. I want users from `s1.example.com` authenticating using specific policies for `s1`. But others, such as those from `s2.example.com` or `example.com`, I want to use the same policies.

Wdyt?

sguilhen:
I agree with @pedroigor 's reasoning. I think the use-case described is realistic and we should be on the side of flexibility here.

vramik (approved):
I agree with @pedroigor and @sguilhen — we should allow overlapping domains (e.g., OrgA with sales.example.com alongside OrgB with *.example.com) and rely on the exact-match-first policy to disambiguate. Restricting overlapping domains would force all authentication policies, brokers, etc. into a single org for an entire domain space, which doesn't reflect real-world organizational structures like subsidiaries or departments.                   

  This approach also sets the right foundation for two related features we'll need down the road:                                                       

- Multi-org IDP (#31705): When a single IdP (e.g., shared EntraID) is linked to multiple organizations, domain matching becomes one of the key disambiguation mechanisms. The exact-match-first policy established here works naturally with that — a user from sales.example.com routes to the subsidiary org, while *.example.com catches everyone else, regardless of whether they share the same IdP.                                             

- Multi-org domains: In some B2B scenarios (e.g., maintenance providers working across multiple customers), a domain may legitimately relate to multiple organizations. Today `getByDomainName()` returns a single org, which is fine for now. But the fact that this PR already handles priority/disambiguation (exact > wildcard) means the mental model is compatible with eventually returning multiple matches and letting other signals (requested scope, user membership, org selection) resolve the final context.

ssilvert (changes requested):
Reviewing only the UI parts, (messages_en.properties and LinkIdentityProviderModal.tsx):

Issues
1. Type mismatch — LinkRepresentation not updated
The existing type at [LinkIdentityProviderModal.tsx:30-36](LinkIdentityProviderModal.tsx#L30-L36) only declares config: { "kc.org.domain": string }. The PR adds a field for kc.org.excluded.domain but doesn't extend this type. It won't cause a runtime error (the value flows through convertFormValuesToObject), but it's a TypeScript inconsistency worth fixing for correctness.

2. Attribute key name mismatch — singular vs. plural
The i18n key is excludedDomains (plural), the label says "Excluded domains" (plural), but the config attribute is kc.org.excluded.domain (singular). The attribute key should probably be kc.org.excluded.domains for consistency — or the backend key needs clarification.

3. Backticks in help text won't render as code

excludedDomainsHelp=...You can use wildcard domains like `*.example.com`...
The labelIcon prop in TextControl renders as a PF popover with plain text — backticks will appear literally. Consider using plain prose instead: "...like *.example.com..." or just omit the formatting.

4. No input validation
The field accepts any string with no pattern validation. A comma-separated domain list is a structured input — consider adding a regex pattern rule (or at minimum a placeholder like *.example.com, sub.example.com) so users understand the expected format.

pedroigor:
@shaidar, would you mind reviewing and trying out the latest changes to check if it is still aligned with your expectations?

ssilvert (dismissed):

PR #45196, opened 2026-01-06
Allow view-users role to see Groups->Role Mapping tab.

### LINKED ISSUE #45093: Enable visibility of Role Mapping tab for users with view-users role
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Currently "Role Mapping" tab in the group section is only visible to users who have "manage-users".
The Admin API already allows viewing these mappings with the view-users role, so the UI should be updated to reflect this capability.
This was only a frontend change.

### Version

26.4.7

### Regression

### Expected behavior

Users with the view-users role should be able to see and click on the "role mapping" tab but not update role-mappings.

### Actual behavior

The "Role Mapping" tab is completely hidden for any user who does not have the manage-users role, even if they have view-users.

### How to Reproduce?

1. Create a new user
2. Give the user query-groups and view-users role.
3. Log in as the user in the admin console
4. Go to groups, click on a group.
5. Role tabbing is not visible.

### Anything else?

_No response_

ssilvert:
Fixes #45093

### DIFF
--- a/js/apps/admin-ui/src/groups/GroupsSection.tsx
+++ b/js/apps/admin-ui/src/groups/GroupsSection.tsx
@@ -62,14 +62,14 @@ export default function GroupsSection() {
   const [key, setKey] = useState(0);
   const refresh = () => setKey(key + 1);
 
-  const { hasAccess } = useAccess();
+  const { hasAccess, hasSomeAccess } = useAccess();
   const isFeatureEnabled = useIsFeatureEnabled();
   const canViewPermissions =
     isFeatureEnabled(Feature.AdminFineGrainedAuthz) &&
     hasAccess("manage-authorization", "manage-users", "manage-clients");
   const canManageGroup =
-    hasAccess("manage-users") || currentGroup()?.access?.manage;
-  const canManageRoles = hasAccess("manage-users");
+    hasAccess("manage-users") || currentGroup()?.access?.manage || false;
+  const canViewRoles = hasSomeAccess("view-users", "manage-users");
   const canViewDetails =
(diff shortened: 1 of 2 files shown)


js/apps/admin-ui/src/groups/GroupsSection.tsx
```diff
@@ -62,14 +62,14 @@ export default function GroupsSection() {
...
-  const { hasAccess } = useAccess();
+  const { hasAccess, hasSomeAccess } = useAccess();
   const isFeatureEnabled = useIsFeatureEnabled();
   const canViewPermissions =
     isFeatureEnabled(Feature.AdminFineGrainedAuthz) &&
     hasAccess("manage-authorization", "manage-users", "manage-clients");
   const canManageGroup =
-    hasAccess("manage-users") || currentGroup()?.access?.manage;
-  const canManageRoles = hasAccess("manage-users");
+    hasAccess("manage-users") || currentGroup()?.access?.manage || false;
```
Abassion, line 71: is `|| false` here necessary?

ssilvert: Yes.  Otherwise, this predicate could be undefined.

Abassion: Ah I see! This might be subjective, but I personally find the double negation better in terms of readability:

`const canManageGroup = !!hasAccess("manage-users") || !!currentGroup()?.access?.manage;
`

Or maybe using the `Boolean()` constructor function.

pedroigor (approved):

PR #45197, opened 2026-01-06
Enable HelpCommandDistTest on Windows runners

Pepo48:
Closes: #45044

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
@@ -626,7 +626,7 @@ public class Picocli {
                 }
                 return result;
             }
-        }).name(Environment.getCommand());
+        }).name(getCommandNameForHelp());
         updateSpecHelpAndUnmatched(spec, unrecognizedArgs);
 
         CommandLine cmd = new CommandLine(spec);
@@ -638,6 +638,7 @@ public class Picocli {
         cmd.getHelpSectionMap().put(SECTION_KEY_COMMAND_LIST, new SubCommandListRenderer());
         cmd.setErr(getErrWriter());
         cmd.setOut(getOutWriter());
+        configureUsageHelpWidth(cmd);
 
         removePlatformSpecificCommands(cmd);
 
@@ -669,6 +670,25 @@ public class Picocli {
         // not an official option, just a way for integration tests to produce the same output regardless of OS
         return Optional.ofNullable(System.getenv("KEYCLOAK_COMMAND_MODE")).map(CommandMode::valueOf)
                 .orElse(Environment.isWindows() ? CommandMode.WIN : CommandMode.UNIX);
+    }
+
+    private String getCommandNameForHelp() {
+        // enforce kc.sh for ALL mode to ensure consistent line wrapping
+        if (getCommandMode() == CommandMode.ALL) {
+            return "kc.sh";
+        }
+        return Environment.getCommand();
+    }
+
+    private void configureUsageHelpWidth(CommandLine cmd) {
+        // not an official option, just a way to make help wrapping configurable
+        Optional.ofNullable(System.getenv("KEYCLOAK_HELP_WIDTH"))
+                .map(Integer::parseInt)
+                .filter(width -> width > 0)
+                .ifPresent(width -> {
+                    cmd.setUsageHelpAutoWidth(false);
+                    cmd.setUsageHelpWidth(width);
+                });
     }
 
     public PrintWriter getErrWriter() {
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
@@ -42,15 +42,15 @@ import org.keycloak.quarkus.runtime.cli.command.UpdateCompatibilityMetadata;
 import io.quarkus.test.junit.main.Launch;
 import org.apache.commons.io.FileUtils;
 import org.approvaltests.Approvals;
-import org.hamcrest.MatcherAssert;
-import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import static org.keycloak.quarkus.runtime.cli.command.AbstractAutoBuildCommand.OPTIMIZED_BUILD_OPTION_LONG;
 
-@WithEnvVars({"KEYCLOAK_COMMAND_MODE", "ALL"})
+@WithEnvVars({"KEYCLOAK_COMMAND_MODE", "ALL", "KEYCLOAK_HELP_WIDTH", "80"})
 @DistributionTest
 @RawDistOnly(reason = "Verifying the help message output doesn't need long spin-up of docker dist tests.")
+@Tag(DistributionTest.WIN)
 public class HelpCommandDistTest {
 
     public static final String REPLACE_EXPECTED = "KEYCLOAK_REPLACE_EXPECTED";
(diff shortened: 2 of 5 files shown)


Pepo48:
This bit helped me to debug the issue:
```
- name: Upload ApprovalTests (HelpCommandDistTest) artifacts
  if: failure()
  uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
  with:
    name: quarkus-approvals-${{ matrix.os }}-${{ matrix.suite }}
    path: |
      quarkus/tests/integration/src/test/resources/**/approvals/**/*.approved.txt
      quarkus/tests/integration/target/test-classes/**/approvals/**/*.received.txt
      quarkus/tests/integration/target/surefire-reports/**
```
At first I wanted to include it into the PR, but then I realized that it is job-scoped, so the artiffacts would be uploaded with every job failure. In the current CI codebase I don't see an example of detecting a specific test failure, so not sure if a conditional upload mechanism is something we actually would appreciate.

Nevertheless, I'm leaving it here as a discussion point and possible follow-up.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
```diff
@@ -221,8 +227,21 @@ private void assertHelp(CLIResult cliResult) {
                     .replace("\r\n", "\n");
         }
 
+        // normalize whitespace via custom comparator to prevent text wrapping differences
+        Options options = new Options().withComparator((received, approved) -> {
+            try {
+                String receivedContent = FileUtils.readFileToString(received, StandardCharsets.UTF_8);
+                String approvedContent = FileUtils.readFileToString(approved, StandardCharsets.UTF_8);
+                String normalizedReceived = receivedContent.replaceAll("\\s+", " ").trim();
+                String normalizedApproved = approvedContent.replaceAll("\\s+", " ").trim();
```
Pepo48: This collapses every occurrence of whitespace into a single space. Is it acceptable approach?

The reason for this was that two tests were failing on Windows due to a different wrapping, likely caused by a different terminal width.

(The rest of the non visible failures were caused by ANSI escape sequences).

vmuzikar: From my perspective, this is fine.

vmuzikar:
> At first I wanted to include it into the PR, but then I realized that it is job-scoped, so the artiffacts would be uploaded with every job failure.

I would not overcomplicate it at the moment. It was the first instance we needed something like that. It's ok to leave that out for now, from my perspective.

vmuzikar (dismissed):
@Pepo48 LGTM, thank you for looking into this further.

@shawkins Do you want to give this a look too?

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
shawkins: I would consider doing these normalizations only when on windows, and require the output to match on unix.

The only downside is that would make it difficult for windows users to update the expected results - but that may be for the best because if someone were to checkin expected windows files, that would contain superficial differences. Alternatively you could consider applying the normalizations during the "update expected" logic, so that we create an OS agnostic expected file.

vmuzikar: > Alternatively you could consider applying the normalizations during the "update expected" logic, so that we create an OS agnostic expected file.

I'd keep the expected file as is. It's useful during the reviews to see the actual command structure without needing to compile and run it locally. :D

vmuzikar (dismissed):

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HelpCommandDistTest.java
Pepo48: @shawkins @vmuzikar alternatively, if we would like to tackle this at its core, we can add to `Picocli.java`:
```
CommandLine cmd = new CommandLine(spec);
...
cmd.setUsageHelpAutoWidth(false);
cmd.setUsageHelpWidth(100);
```
and then revert the platform wrapping normalization stuff in `HelpCommandDistTest.java`.

shawkins: @Pepo48 I'm fine with doing something like that - would it be driven from a config / environment variable to keep the default auto behavior the same otherwise?

Pepo48: Yes, my plan was to make this configurable, but I wanted to keep it internal, just for the tests, similar to `KEYCLOAK_COMMAND_MODE` pattern. Something like:
```
@WithEnvVars({"KEYCLOAK_COMMAND_MODE", "ALL", "KEYCLOAK_HELP_WIDTH", "100"})
```
Frankly, I didn't plan to expose it publicly. Is it fine like that or do you prefer to give a user the option to configure it?

shawkins: No need to expose it.

vmuzikar: > alternatively, if we would like to tackle this at its core

+1 I like that

vmuzikar (dismissed):
LGTM, thanks.

vmuzikar:
@Pepo48 Could you  please check the failing test?

shawkins (dismissed):
LGTM, thanks @Pepo48

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
```diff
@@ -671,6 +672,25 @@ protected CommandMode getCommandMode() {
                 .orElse(Environment.isWindows() ? CommandMode.WIN : CommandMode.UNIX);
     }
 
+    private String getCommandNameForHelp() {
+        // enforce kc.sh for ALL mode to ensure consistent line wrapping
+        if (getCommandMode() == CommandMode.ALL) {
+            return "kc.sh";
+        }
+        return Environment.getCommand();
+    }
```
Pepo48, line 681: @vmuzikar @shawkins this is how I ended up solving the line wrapping inconsistencies. It's basically same approach as before, just moved from the test-level ` .replace("kc.bat", "kc.sh")` to its current form in Picocli. The test-level text replacement was not enough, because it changed the text but not the linebreak position. In other words, the linebreak was determined earlier than the text replacement execution. 

Just a reminder from the last week - the core of the issue is that kc.bat is one character longer (6 chars) than kc.sh (5 chars), which causes Picocli to wrap text at different positions. 

Sorry for the postponement, I kinda forgot to wrap this up.

Pepo48: That being said, `KEYCLOAK_HELP_WIDTH` var does not solve anything - the Picocli autowrap seems to be consistent over the platforms. I left it in the PR though, as we may still find it useful.

shawkins: That's unfortunate. There's no harm in leaving it in for now.

shawkins (approved):
LGTM, but just wanted to ask again whether we should only do the normalization on windows and expect linux formatting otherwise.

Pepo48:
@shawkins with the earlier implementation we put everything into single line, which worked and solved all of the underlying discrepancies as a byproduct, but was quite drastic. Keeping such workaround just for Windows made sense. The current implementation tackles these issues one by one at its core:
* stripping trailing whitespace (just before EOL) on every line and in the heading (heading was needed to be done separately),
* stripping ANSI escape sequences
* replacing kc.bat with kc.sh a step earlier

These are cosmetic changes and are not even specific to Windows. When I run the tests locally, the only adjustment I needed to make was the last one from the list. The CI is the place that required the rest - for some reason the environment there does not ignore what's normally ignored locally on both platforms. So rather cosmetic nature was the reason why I decided that doing this just specifically for Windows is not necessary. Anyway, I’m willing to make any improvements you suggest here, even though it was auto-merged after your approval. Let me know. Thanks!

shawkins:
> Anyway, I’m willing to make any improvements you suggest here, even though it was auto-merged after your approval. Let me know. Thanks!

I don't know that we have to do anything additional. Just wanted to make sure we were ok with always doing the normalizations, rather than putting them further down into the isWindows check - and essentially requiring, as we did previously, that the output would match exactly on unix.

PR #45200, opened 2026-01-07
Add SAML client model mapper for admin-v2 API

### LINKED ISSUE #44853: Client v2 SAML representation
### Description

This is a follow-up from https://github.com/keycloak/keycloak/pull/44727.

* Create MVP SAML Client representation.
* Create a test for getting list of Clients that should mix both OIDC and SAML clients.

edewit:
Add SAML client model mapper for admin-v2 API

Implements mapper and factory for converting between SAMLClientModel
and SAMLClientRepresentation, including support for SAML-specific
attributes like signature algorithms, name ID formats, and certificates.

Fixes #44853

### DIFF
--- a/rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/SAMLClientRepresentation.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/SAMLClientRepresentation.java
@@ -1,16 +1,195 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.keycloak.representations.admin.v2;
 
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 /**
+ * Representation of a SAML client.
  * @author Vaclav Muzikar <vmuzikar@redhat.com>
  */
-@Schema
+@Schema(description = "SAML Client configuration")
 public class SAMLClientRepresentation extends BaseClientRepresentation {
     public static final String PROTOCOL = "saml";
+
+    @JsonPropertyDescription("Name ID format to use for the subject (e.g., 'username', 'email', 'transient', 'persistent')")
+    private String nameIdFormat;
+
+    @JsonPropertyDescription("Force the specified Name ID format even if the client requests a different one")
+    private Boolean forceNameIdFormat;
+
+    @JsonPropertyDescription("Include AuthnStatement in the SAML response")
+    private Boolean includeAuthnStatement;
+
+    @JsonPropertyDescription("Sign SAML documents on the server side")
+    private Boolean signDocuments;
+
+    @JsonPropertyDescription("Sign SAML assertions")
+    private Boolean signAssertions;
+
+    @JsonPropertyDescription("Require client to sign SAML requests")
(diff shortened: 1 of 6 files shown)


rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
```diff
@@ -224,6 +228,86 @@ public void deleteClient() throws Exception {
         }
     }
 
+    @Test
+    public void getClientsMixedProtocols() throws Exception {
```
vmuzikar, line 232: Could we set and verify some OIDC and SAML specific fields in the test? That would ensure Keycloak really interprets those as different protocols.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
```diff
@@ -224,6 +228,86 @@ public void deleteClient() throws Exception {
...
+        setAuthHeader(deleteOidc);
+        try (var response = client.execute(deleteOidc)) {
+            assertEquals(204, response.getStatusLine().getStatusCode());
+        }
+
+        HttpDelete deleteSaml = new HttpDelete(HOSTNAME_LOCAL_ADMIN + "/realms/master/clients/mixed-test-saml");
+        setAuthHeader(deleteSaml);
+        try (var response = client.execute(deleteSaml)) {
+            assertEquals(204, response.getStatusLine().getStatusCode());
+        }
```
vmuzikar, line 351: It would be nice to have the cleanup in a `finally` block. But since we lack a nice generic cleanup mechanism in this test in general, and other test methods don't cleanup at all, it's ok as is at the moment.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
vmuzikar, line 1: Again, more a generic note rather than something specific to this PR: we lack tests for the mapping logic, which includes the OIDC client too. Created a follow-up issue: https://github.com/keycloak/keycloak/issues/45277.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/SAMLClientRepresentation.java
vmuzikar, line 1: I think the fields selection is a good start. We are missing some (like encryption keys config, signature key name, metadata descriptor URL, etc.) but since this is an MVP, I think it's good enough for now. We might add some after a wider Keycloak team review.

vmuzikar (changes requested):
@edewit Thank you for the PR.

Added some comments, mostly just observations. The only thing I'd like to get addressed before merging is having some specific OIDC and SAML fields in `getClientsMixedProtocols`.

vmuzikar (dismissed):
LGTM, thank you.

vmuzikar:
@edewit I noticed the new test failed. Could you please check that?

rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -84,10 +87,17 @@ public CreateOrUpdateResult createOrUpdate(RealmModel realm, BaseClientRepresent
...
+            // We can't use mapper.toModel(client) directly for creation because the "detached model"
+            var basicRep = new ClientRepresentation();
+            basicRep.setClientId(client.getClientId());
+            basicRep.setProtocol(client.getProtocol());
+
+            // Create the client in the database
+            model = clientsResource.createClientModel(basicRep);
             clientResource = clientsResource.getClient(model.getId());
+
+            mapper.toModel(client, model);
```
vmuzikar, line 100: So it looks like the creation of Clients didn't really work. :D Thanks for the fix.

To complete that, we should also clean up the mapping logic and contract to prohibit calls with `null` models.

mabartos (approved):
@edewit LGTM, thanks!

Merging based on the previous review of @vmuzikar

vmuzikar:
@mabartos We were missing the clean up mentioned in https://github.com/keycloak/keycloak/pull/45200#discussion_r2682965544. :) But we can do that as a follow-up.

mabartos:
> @mabartos We were missing the clean up mentioned in https://github.com/keycloak/keycloak/pull/45200#discussion_r2682965544. :) But we can do that as a follow-up.

@vmuzikar Oops, sorry. Yes, we should do it in a follow-up now :P 

cc: @edewit

PR #45206, opened 2026-01-07
[OID4VCI] Fix `authorization_details` being returned in subsequent SSO logins after OID4VCI flow

### LINKED ISSUE #44961: Authorization_details added to token-response even when should not be
### Before reporting an issue

### Area

oid4vc

### Describe the bug

Consider this scenario:

- OIDC Authentication login request sent with some scope corresponding to OID4VCI (EG. `scope=oid4vc_natural_person`) and with the `authorization_details` parameter asking for creating credential-offer of some OID4VCI credential
- Credential-offer is created and `authorization_details` are added to the token response -> OK
- Now assume that there is 2nd OIDC authentication login request to the same client in the same browser. This is normal SSO login request, which does not do anything related to OID4VCI . However in the token-response, the `authorization_details` are still returned after this 2nd OIDC authentication flow is finished  -> KO
- Also I think that the access-token returned from this 2nd-request is possible to be used in the credential-request as that 2nd request created OID4VCI credential-offer -> KO

### Version

nightly (from 2025-12-17)

### Regression

### Expected behavior

Only the 1st token-response, which was triggered from the OIDC flow asking for OID4VCI credential, is supposed to return `authorization_details` and create credential-offer. The subsequent OIDC login requests in same browser session should not do that (especially if those requests are regular SSO login requests not asking for any OID4VCI scope or authorization_details)

### Actual behavior

The subsequent token-responses (2nd and others) are adding `authorization_details` even if they should not

### How to Reproduce?

See details

### Anything else?

I think we can add `authorization_details` as a claim into access-token (and maybe also probably refresh-token if issued) instead of adding them as a note in userSession. Storing them inside token instead of userSession can make sure that credential-request can be successful just if the sent token contains the relevant `authorization_details` . That should also help with fixing the issue https://github.com/keycloak/keycloak/issues/44670 and probably also https://github.com/keycloak/keycloak/issues/44534 as userSession may not be needed anymore when pre-authorization flow is used.

forkimenjeckayang:
## Problem

When a user performs an OIDC login with OID4VCI scope and `authorization_details`, the token correctly includes `authorization_details`. However, if the same user performs a subsequent regular SSO login (without OID4VCI scope or authorization_details), the token response incorrectly included `authorization_details` from the previous login.

This occurred because `authorization_details` were stored in the user session, which persisted across multiple authentication flows in the same browser session.

## Solution

- Store `authorization_details` as claims in access tokens and refresh tokens instead of in the user session
- Only generate `authorization_details` from credential offers when:
  - No `authorization_details` were processed from the request, AND
  - A credential offer note exists in the client session (indicating an OID4VCI flow)

This ensures that `authorization_details` are only included in tokens that were explicitly issued with OID4VCI scope and `authorization_details` parameter.

closes #44961

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -109,6 +109,7 @@ import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.protocol.oid4vc.utils.ClaimsPathPointer;
 import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantType;
 import org.keycloak.protocol.oidc.grants.PreAuthorizedCodeGrantTypeFactory;
+import org.keycloak.protocol.oidc.rar.AuthorizationDetailsResponse;
 import org.keycloak.representations.AccessToken;
 import org.keycloak.representations.dpop.DPoP;
 import org.keycloak.saml.processing.api.util.DeflateUtil;
@@ -775,8 +776,8 @@ public class OID4VCIssuerEndpoint {
 
         if (credentialIdentifier != null) {
 
-            // Retrieve the associated credential offer state
-            //
+            // First check if the credential identifier exists
+            // This allows proper error reporting for unknown identifiers
             CredentialOfferStorage offerStorage = session.getProvider(CredentialOfferStorage.class);
             CredentialOfferState offerState = offerStorage.findOfferStateByCredentialId(session, credentialIdentifier);
             if (offerState == null) {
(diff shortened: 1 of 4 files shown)


tdiesler (approved):
Works for now - ultimately there should not be a distinction between an OAuth2Code and a pre-authorized code from a credential offer. In case of Authorization Code Flow, the Wallet would set the `authorization_details` in the AuthorizationRequest. From there, they find their way into the auth code and from there into the AccessToken.

For the pre-authorized credential offer, it would be the Issuer who associated appropriate `authorization_details` with the code. From there they find their way into the AccessToken.

Both types of Authorization Code types can be treated equivalently.

PS: we likely need to rework how pre-authorized codes are created wrt `authorization_details`

Anyhow, the PR looks good (enough) for me

forkimenjeckayang:
> Works for now - ultimately there should not be a distinction between an OAuth2Code and a pre-authorized code from a credential offer. In case of Authorization Code Flow, the Wallet would set the `authorization_details` in the AuthorizationRequest. From there, they find their way into the auth code and from there into the AccessToken.
> 
> For the pre-authorized credential offer, it would be the Issuer who associated appropriate `authorization_details` with the code. From there they find their way into the AccessToken.
> 
You're right @tdiesler , thanks for pointing that out. In upcomming tasks we'll need to rework the implementation to ajdust this

mposolda:
Thanks for the PR. I've did some testing and figured few more things:

- The response from refresh-token does not have `authorization_details` included as response parameter

- The credential-endpoint verifies `authorization_details` from the access-token, however it does not use those in the `OID4VCIssuerEndpoint.getClaimsFromAuthzDetails`, but still uses those from userSession. This can cause that incorrect `authorization_details` are possibly used in some scenario (See `OID4VCAuthorizationCodeFlowTestBase.testCorrectAccessTokenUsed()`, which I've added here: https://github.com/keycloak/keycloak/pull/45622

- Figured that `AuthorizationDetailsProcessor` for processing `authorization_details` parameter is considered only based on the priority of the provider. However in case that there are multiple different `type` of `authorization_details`, then this behaviour is likely not correct. It may be rather desired to make sure that `AuthorizationDetailsProcessor` is chosen based on the `type` of the `authorization_details` entry, so that different types of authorization_details can be possibly processed by appropriate processors (once we support multiple processors in Keycloak)

Ended with the other PR, which I've sent #45622 . It contains your commit together with some additional changes. See the description of the PR for the details. Would you be able to review that please and check if it is OK for you?

mposolda:
Closing as those changes were incorporated into other PR https://github.com/keycloak/keycloak/pull/45622 .

PR #44751, opened 2025-12-08
fix: changes testcontainers artifact names

shawkins:
closes: #44750

@vmuzikar @Pepo48 @mabartos how do we usually handle something like this?

Summary of changes:
- h2 version change
- testcontainers version change
- we need to manage microprofile-metrics-api https://github.com/quarkusio/quarkus/pull/49735

Issue #44679, opened 2025-12-04
Make sure events are properly used in OID4VCI endpoints


mposolda:
### Description

It seems there is not use of events in the credential-request (class `OID4VCIssuerEndpoint.requestCredential`) or other endpoints related to OID4VCI.

The endpoints should properly use events and error-events similarly like all other supported Keycloak functionality. Will be good if events are tested as well in some OID4VCI automated tests.

forkimenjeckayang:
will work on this

PR #45214, opened 2026-01-07
fix: removing references to shaded testcontainer classes

### LINKED ISSUE #45188: Upgrade to quarkus 3.30.5
### Description

Upgrade to quarkus 3.30.5

### Value Proposition

To align to the latest quarkus release.

### Goals

N/A

### Non-Goals

N/A

### Discussion

_No response_

### Notes

_No response_

shawkins:
Pulling some changes over from #44751

closes: #45188

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/EventOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/EventOptions.java
@@ -23,22 +23,7 @@ public class EventOptions {
             .category(OptionCategory.EVENTS)
             .description("Comma-separated list of events to be collected for user event metrics. This option can be used to reduce the number of metrics created as by default all user events create a metric.")
             .buildTime(false)
-            .expectedValues(sortedListOfEvents())
             .deprecatedMetadata(DeprecatedMetadata.deprecateValues("Use `remove_credential` instead of `remove_totp`, and `update_credential` instead of `update_totp` and `update_password`.", "remove_totp", "update_totp", "update_password"))
             .build();
-
-    private static List<String> sortedListOfEvents() {
-        List<String> events = new java.util.ArrayList<>(List.of("register", "login", "code_to_token", "logout", "client_login", "refresh_token", "introspect_token", "federated_identity_link",
-                "remove_federated_identity", "update_email", "update_profile",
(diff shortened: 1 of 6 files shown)


mabartos (dismissed):

vmuzikar (dismissed):

shawkins:
@vmuzikar @mabartos the EventOptionsTest failure is unrelated to these changes. ~It looks like that test should have been failing even in 26.5, so there must be something else about the change to quarkus latest where we're now running that test - but it was not triggered the initial PR.~

Edit: it was https://github.com/keycloak/keycloak/issues/44679 that caused the regression. cc @mposolda 

I'll see if I can update it as part of this PR.

shawkins:
We can also consider removing the logic from helpcommanddisttest that removes the event options, so that it's clear when they change.

mabartos:
The test should be already fixed: https://github.com/keycloak/keycloak/commit/cc3ab865441010ac7ec37c510c067b17fee913e4

@shawkins I like this approach more than having the hardcoded sorted events on the EventOptions class. However, it's quite unrelated to this issue, so it would be nice to have it done in a separate issue for better addressability.

shawkins:
> @shawkins I like this approach more than having the hardcoded sorted events on the EventOptions class. However, it's quite unrelated to this issue, so it would be nice to have it done in a separate issue for better addressability.

In this pr it's a separate commit associated with #44679. Since that was committed to main only, I would prefer to not open additional issues to just address the test refinement - it seems fine to associate the change either with #44679 or the test fix issue.

At this point it's not a large change so a separate PR doesn't seem necessary. If we want to change the help command dist logic to stop omitting changes in events, then I'll definitely separate the PRs.

mabartos:
> In this pr it's a separate commit associated with https://github.com/keycloak/keycloak/issues/44679. Since that was committed to main only, I would prefer to not open additional issues to just address the test refinement - it seems fine to associate the change either with https://github.com/keycloak/keycloak/issues/44679 or the test fix issue.

Ahh, I didn't notice that these 2 commits closes different issues. Works to me to have it as is.

mabartos (approved):

mabartos:
@vmuzikar @mposolda I see you're in a reviewer list. Do you plan to look at it, or can we merge?

mabartos:
Merging now. 

If there's more feedback, it can be done in follow-ups.

vmuzikar (commented):
Sorry, the notification slipped between the cracks. No concerns  from me, looks good.

PR #45228, opened 2026-01-07
Login[v2]: Improve a11y for authenticator selector

fbieler:
Clickable elements should be focusable and have interactive semantics. See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Keyboard?utm_source=devtools&utm_medium=a11y-panel-checks-keyboard#Interactive_elements_must_be_focusable

Closes #45227


mposolda (approved):
@fbieler Thanks!

PR #45229, opened 2026-01-07
Align user count to user search

### LINKED ISSUE #45219: User REST Admin API - count and search returns different amount of users
### Before reporting an issue

### Area

account/api

### Describe the bug

Problem: via REST API, when I use `GET /admin/realms/{realm}/users/count?exact=true` endpoint, Keycloak returns all users except this service account user. But when I call `GET /admin/realms/{realm}/users?exact=true`, it returns the same list of users plus this service user as well.

The misalignment was already partially fixed in https://github.com/keycloak/keycloak/pull/41980, but missed the `exact` check in the if statement of the count function (see below) 

### Version

26.4

### Regression

### Expected behavior

`/users/count?exact=true` and `users?exact=true` behaviour should be aligned and return the same number of users

### Actual behavior

`users?exact=true` returns also the service account users, while the `count?exact=true` does not

### How to Reproduce?

- create a new realm
- create a client of type service account in the realm
- invoke the `users/count?exact=true`, the result is 0
- invoke the `users/count?exact=true`, it returns the service account (1 user)

### Anything else?

https://github.com/keycloak/keycloak/blob/release/26.4/services/src/main/java/org/keycloak/services/resources/admin/UsersResource.java#L323 the `if` statements in the getUsers method contains an `|| exact != null`. While the same if statement in the getUsersCount does not https://github.com/keycloak/keycloak/blob/release/26.4/services/src/main/java/org/keycloak/services/resources/admin/UsersResource.java#L452.

It also seems that the not null condition on `idpAlias` and `idpUserId` are not aligned.

ecostanzi:
Closes #45219

Aligns the behaviour of the user count and user search admin REST API. Null checks on `exact`, `idpAlias` and `idpUserId` are consistents between the search and the count API.

Linked to https://github.com/keycloak/keycloak/pull/41980

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/UsersResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/UsersResource.java
@@ -450,7 +450,8 @@ public class UsersResource {
                     return session.users().getUsersCount(realm, parameters);
                 }
             }
-        } else if (last != null || first != null || email != null || username != null || emailVerified != null || enabled != null || !searchAttributes.isEmpty()) {
+        } else if (last != null || first != null || email != null || username != null || emailVerified != null
+            || idpAlias != null || idpUserId != null || enabled != null || exact != null || !searchAttributes.isEmpty()) {
             Map<String, String> parameters = new HashMap<>();
             if (last != null) {
                 parameters.put(UserModel.LAST_NAME, last);


vramik:
Thank you @ecostanzi for the contribution! 

Really appreciated.

I can see that it was fixed in the meantime by https://github.com/keycloak/keycloak/pull/45421, closing this one.

PR #45251, opened 2026-01-08
Set initial default value and pass required for MultivaluedString

### LINKED ISSUE #43949: MultivaluedString in mappers is not saved to backend with shown default value
### Before reporting an issue

### Area

_No response_

### Describe the bug

When using a MultivaluedString in a mapper in a client scope, the shown default values are not saved to the backend.

This occurred to me when implementing https://github.com/keycloak/keycloak/pull/42984

### Version

main

### Regression

### Expected behavior

The default values in a MultivaluedString should be saved to the backend

### Actual behavior

The MultivaluedString values are only saved to the backend when the value is modified

### How to Reproduce?

* Open the Admin UI
* Create a new client "test" of type SAML
* Go to "Client scopes"
* Open the "test-dedicated" scope
* Add a mapper "Authenticator reference type to AuthnContext"
* Click "Save" -> network tab shows that the field "authenticatorReferenceTypes" is missing 

image

Image

### Anything else?

A workaround is to in the backend to use the default when the value is not set.

rmartinc:
Closes #43949

The default value is never set to the attribute. I have passed also the `required` and added the required message if empty.

### DIFF
--- a/js/apps/admin-ui/src/components/dynamic/MultivaluedStringComponent.tsx
+++ b/js/apps/admin-ui/src/components/dynamic/MultivaluedStringComponent.tsx
@@ -32,6 +32,7 @@ export const MultiValuedStringComponent = ({
         aria-label={t(label!)}
         name={fieldName}
         isDisabled={isDisabled}
+        isRequired={required}
         defaultValue={convertDefaultValue(defaultValue)}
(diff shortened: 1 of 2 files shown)


mposolda (approved):
@rmartinc Thanks!

PR #45264, opened 2026-01-08
Wrap required indicators with "required" span

tagliala:
Some asterisks indicating required fields in `user-profile-commons.ftl`
and `register.ftl` are not wrapped in a `<span class="required">` tag.

This caused incorrect styling in custom themes that inherit the base theme.

Update the markup to wrap required field indicators with
`<span class="required">`, matching the rest of the theme and ensuring
consistent styling.

Closes #45058

Signed-off-by: Geremia Taglialatela <tagliala.dev@gmail.com>

---

### Reference

The current implementation in config-totp template:

image

### Before (Register)

image

### After (Register)

image

### Note

the "laquo" issue appears even when I eject the default template without changes, so it should not be related to this change

### DIFF
--- a/themes/src/main/resources/theme/base/login/register.ftl
+++ b/themes/src/main/resources/theme/base/login/register.ftl
@@ -17,7 +17,8 @@
                     <#if passwordRequired?? && (attribute.name == 'username' || (attribute.name == 'email' && realm.registrationEmailAsUsername))>
                         <div class="${properties.kcFormGroupClass!}">
                             <div class="${properties.kcLabelWrapperClass!}">
-                                <label for="password" class="${properties.kcLabelClass!}">${msg("password")}</label> *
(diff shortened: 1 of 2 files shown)


ahus1 (approved):

ahus1:
Thank you for the change. The "laquo" text should not appear in the most recent version of the themes. Maybe your setup was using an old message bundle. This was changed in 2210b1ed505e81ced15a3ad6a87138629bf08ff9

tagliala:
> The "laquo" text should not appear in the most recent version of the themes. Maybe your setup was using an old message bundle. This was changed in https://github.com/keycloak/keycloak/commit/2210b1ed505e81ced15a3ad6a87138629bf08ff9

Thanks, I was actually using templates from 26.4.7

I now see 26.5.0 being released, I'll take a look

PR #45270, opened 2026-01-08
PLEASE IGNORE AND DELETE ME, was Set inital status in context to avoid subsequent NPE #45003

### LINKED ISSUE #45003: Bug in JWTClientAuthenticator and JWTClientSecretAuthenticator causes NPE
### Before reporting an issue

### Area

_No response_

### Describe the bug

In the classes `JWTClientAuthenticator` and `JWTClientSecretAuthenticator` there exists a "short-circuit" check in the `authenticateClient()` method that aborts the current flow, but without setting the proper context state. That code is triggered if the client has sent a JWT via the `client_assertion` request parameter.

Now think of a situation where the issuer and subject of the token do not match for some reason.

The critical code is in `JWTClientAuthenticator`:
> 
            // Ignore for client assertions signed by third-parties
            if (!Objects.equals(jwt.getIssuer(), jwt.getSubject())) {
                return;
            }

The method is called by `ClientAuthenticationFlow.processFlow()`. Subsequently it calls the` processResult()` method:
>
                // Check if client authentication matches
                if (factory.getId().equals(expectedClientAuthType)) {
                    Response response = processResult(context);
                    if (response != null) return response;

That method in turn tries to read the current state of the context and log it:

>
    protected Response processResult(AuthenticationProcessor.Result result) {
        AuthenticationExecutionModel execution = result.getExecution();
        FlowStatus status = result.getStatus();

        logger.debugv("client authenticator {0}: {1}", status.toString(), execution.getAuthenticator());

Alas, this causes a `NullPointerException` in the given setting, because the `status` field of the context is `null`.

### Version

26.4.7

### Regression

### Expected behavior

The code has to ensure the proper state is set and not just abort method execution.

This may look like this, eg.

> 
       final String expectedClientAuthType = context.getClient().getClientAuthenticatorType();

      // Ignore for client assertions signed by third-parties
      if (!Objects.equals(jwt.getIssuer(), jwt.getSubject())) {
        if (PROVIDER_ID.equals(expectedClientAuthType)) {
          final Response challengeResponse = ClientAuthUtil.errorResponse(
              Response.Status.BAD_REQUEST.getStatusCode(),
              OAuthErrorException.INVALID_CLIENT,
              "Client authentication with signed JWT failed: subject and issuer do not match"
          );

          context.failure(AuthenticationFlowError.INVALID_CLIENT_CREDENTIALS, challengeResponse);
        }

        return;
      }

### Actual behavior

A `NullPointerException` is raised in the given setting, because the `status` field of the context is `null`.

### How to Reproduce?

Send a token request including a `client_assertion` JWT where issuer and subject do not match.

### Anything else?

_No response_

mdahm:
Set inital status in context to avoid subsequent NPE in ClientAuthenticationFlow

Closes https://github.com/keycloak/keycloak/issues/45003

### DIFF
--- a/themes/src/main/resources/theme/base/login/login-config-totp.ftl
+++ b/themes/src/main/resources/theme/base/login/login-config-totp.ftl
@@ -52,7 +52,7 @@
 
         <form action="${url.loginAction}" class="${properties.kcFormClass!}" id="kc-totp-settings-form" method="post">
             <div class="${properties.kcFormGroupClass!}">
-                <div class="${properties.kcInputWrapperClass!}">
+                <div class="${properties.kcLabelWrapperClass!}">
                     <label for="totp" class="control-label">${msg("authenticatorCode")}</label> <span class="required">*</span>
                 </div>
                 <div class="${properties.kcInputWrapperClass!}">
@@ -74,7 +74,7 @@
             </div>
 
             <div class="${properties.kcFormGroupClass!}">
-                <div class="${properties.kcInputWrapperClass!}">
+                <div class="${properties.kcLabelWrapperClass!}">
                     <label for="userLabel" class="control-label">${msg("loginTotpDeviceName")}</label> <#if totp.otpCredentials?size gte 1><span class="required">*</span></#if>
                 </div>
 


stianst (commented):
See my comment here:
https://github.com/keycloak/keycloak/issues/45003#issuecomment-3724470270

Not really convinced about this, as I wasn't convinced about it being added the federated client authenticator either. Seems more like a patch/work-around than fixing things properly.

rmartinc (changes requested):
Thanks @mdahm for the PR!

You need to sign the commit (option `-s` or `--signoff` when doing the commit, you can do `git commit -s --amend` to sign  the last commit). If not, the DCO test fails.

Regarding the fix, I also prefer removing the `toString()` if it's enough. My feeling is that this can happen again in the future and in custom authenticators implemented by others. So better if we don't trust the status is filled.

mdahm:
Thanks for the hint!

I removed the toString() call as discussed in https://github.com/keycloak/keycloak/issues/45003
and amended to the last commit with -s.

Hope, it works now and may be integrated.

PS Please don't mind for the whitespace mess, the actual changes are really small in fact.

mdahm:
Sorry I messed up my fork and had to redo everything 🥹

The **REAL** Pull request is now here: https://github.com/keycloak/keycloak/pull/45320

PR #45276, opened 2026-01-08
Fix: Persist refreshed external IDP tokens in DB

Bademeischta:
When retrieving an external IDP token via `retrieveToken`, if the token is expired and successfully refreshed, the new token was updated in the `FederatedIdentityModel` in memory but not persisted to the database. This caused subsequent requests to fail or use old tokens.

This change ensures that `updateFederatedIdentity` is called to persist the refreshed token if `storeToken` is enabled configuration.

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
@@ -264,6 +264,12 @@ public abstract class AbstractOAuth2IdentityProvider<C extends OAuth2IdentityPro
                         newResponse.setAccessTokenExpiration(accessTokenExpiration);
                     }
                     identity.setToken(JsonSerialization.writeValueAsString(newResponse));
+
+                    if (getConfig().isStoreToken()) {
+                        RealmModel realm = session.getContext().getRealm();
+                        UserModel user = session.users().getUserById(realm, identity.getUserId());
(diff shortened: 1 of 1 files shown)


ahus1:
@Bademeischta - thank you for analyzing this and the fix. Can you please create a parent bug issue and then link it here? This way we can also provide a backport. 

See CONTRIBUTING.md for the details of the process. Ideally, we would like to see a test case for that. I can help with that if you like.

Bademeischta:
@ahus1 I will do another pr dw

ahus1:
@Bademeischta - no need for a new pr. Once you provide me the issue number I can link them. No worries.

Bademeischta:
@ahus1 The issue number is: #45280 Is that okay?

ahus1:
Thank you for the issue, the issue looks good. Now that I look more closely to this PR, I see that the PR was closed because the cloned repository was deleted. So unfortunately you'll need to create a new pull request as this PR can not be reopened, not even from me. The code changes are still visible in the "Files changed" tab.

Bademeischta:
Yes ill send u the new pr tag right here then

Bademeischta:
@ahus1 #45282

Bademeischta:
@ahus1 I added an test case, maybe that will work too

Bademeischta:
I gave the test case a shot based on existing tests. If the CI fails or if I missed a convention, let me know and I'm happy to adjust it.

PR #45279, opened 2026-01-08
Use MIME decoder instead of the default one to replace deprecated Base64 class

### LINKED ISSUE #45226: Failure when decrypting SAML Response since 26.5.0
### Before reporting an issue

### Area

saml

### Describe the bug

Since we have moved to 26.5.0, Keycloaks fails to decrypt the SAML response coming from one of our customers' identity provider.
We are getting that error log

`2026-01-07 13:58:26,911 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-6) Uncaught server error: java.lang.IllegalArgumentException: base64 decode failed: Illegal base64 character d`

I can provide the encrypted SAML response privately.

### Version

26.5.0

### Regression

### Expected behavior

The SAML response is decrypted as it used to be in the previous versions

### Actual behavior

`2026-01-07 13:58:26,911 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-6) Uncaught server error: java.lang.IllegalArgumentException: base64 decode failed: Illegal base64 character d`

### How to Reproduce?

I have the SAML response coming from the customer IDP, I can provide it privately

### Anything else?

_No response_

rmartinc:
Closes #45226

This is a fix for a regression caused by https://github.com/keycloak/keycloak/issues/43370. The old Base64 class skips some characters (`\n`, `\t` and `\r` and space itself are -5 in [_STANDARD_DECODABET](https://github.com/keycloak/keycloak/blob/26.4.7/common/src/main/java/org/keycloak/common/util/Base64.java#L245)), that are just not considered for decoding (see [here](https://github.com/keycloak/keycloak/blob/26.4.7/common/src/main/java/org/keycloak/common/util/Base64.java#L1184-L1195)). So the Base64 class is more like the mime instead of the normal decoder. This PR changes the the calls that where changed in the previous PR for `getDecoder`, this time for `getMimeDecoder`. I have added a unit test to check a MIME encoded saml response is also parsed successfully.

### DIFF
--- a/adapters/saml/core/src/main/java/org/keycloak/adapters/saml/config/PemUtils.java
+++ b/adapters/saml/core/src/main/java/org/keycloak/adapters/saml/config/PemUtils.java
@@ -125,7 +125,7 @@ public class PemUtils {
     private static byte[] pemToDer(String pem) {
         try {
             pem = removeBeginEnd(pem);
-            return Base64.getDecoder().decode(pem);
+            return Base64.getMimeDecoder().decode(pem);
         } catch (IllegalArgumentException e) {
             throw new PemException(e);
         }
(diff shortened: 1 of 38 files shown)


ahus1:
@rmartinc: I'm sorry for this regression. I also reviewed the original PR. Thanks for fixing it! 

cc: @pruivo, @tkyjovsk

mposolda (approved):
@rmartinc Thanks a lot for fixing this, LGTM!

@ahus1 @tkyjovsk @pruivo Are you OK with merging this PR, or do you want to review as well?

ahus1:
@mposolda - I'm reviewing it just now and I'm ok to merge it.

PR #45282, opened 2026-01-08
External IDP tokens are not refreshed automatically for OAuth2 & OIDC IDPs

### LINKED ISSUE #45280: External IDP tokens are not refreshed automatically for OAuth2 & OIDC IDPs
### Before reporting an issue

### Area

identity-brokering

### Describe the bug

External IDP tokens are not refreshed automatically for OAuth2 & OIDC IDPs when retrieving the external token.
The fix involves updating the `AbstractOAuth2IdentityProvider` to persist the refreshed token to the database.
Previously, the token was only updated in memory.
This ensures that subsequent requests use the valid, refreshed token.

### Version

v26.5.0
### Regression

### Expected behavior

Keycloak should check the expiration of the external token. If the token is expired, it should attempt to use the associated refresh token to fetch a new access token from the Identity Provider before returning it to the client.

### Actual behavior

Keycloak returns the existing external token stored in the database/session without checking if it is expired or attempting a refresh.

### How to Reproduce?

1. Configure an OIDC Identity Provider (e.g., Google or Keycloak) with "Store Tokens" enabled.
2. Login via this Identity Provider.
3. Wait for the external access token to expire (or manually revoke it/shorten the lifespan in the IDP).
4. Attempt to retrieve the external token.
5. Observe that the returned token is the old, expired one.

### Anything else?

This issue is created to track the fix provided in Pull Request #45282

Bademeischta:
Closes https://github.com/keycloak/keycloak/issues/45280

Store the token in the user's federated identity if configured.


ahus1:
Thank you for the PR, I'll have a look in the upcoming days.

pedroigor (commented):
Without your changes, your test is also passing. As @ahus1 mentioned in the issue, the token should be updated whenever retrieving it via API and when the token is actually refreshed.

Bademeischta:
@pedroigor Thanks for pointing this out. If the test passes without my changes, then my test case was indeed not strictly verifying the persistence to the database (it might have been reading from the session cache or checking the immediate API response only).

I will adjust the test case to ensure it strictly validates that the refreshed token is persisted to the DB (likely by forcing a new transaction/session for the check).

Once I have a test that fails without the fix and passes with the fix, I will update the PR.

Bademeischta:
@pedroigor I have updated the test case significantly based on your feedback.

Changes in KcOidcBrokerTokenRefreshTest:

Strict DB Verification: The test now compares the token stored in the DB before and after the refresh. It asserts that they are different.

Execution Context: I fixed the NullPointerException issues in RunOnServer by explicitly loading the realm context, ensuring the test runs reliably.

Scope: It explicitly tests the full flow: Login -> Expire Token (Time Travel) -> Trigger Refresh via API -> Verify Persistence.

I verified locally that this test fails without the fix (token in DB remains unchanged) and passes with the fix included.

Ready for another look!

pedroigor:
@Bademeischta Thanks for the updates. However, the test is still passing without your changes.

Running locally, I can see the refresh token request being issued when retrieving the token, and the new token set and persisted to the federated identity.

I'm not sure what might be causing a different behavior in your environment, but the behavior described by @ahus1 is what is happening.

ahus1:
Closing the PR as the parent issue was closed due to no response for two weeks.

edamato:
Also being affected by this. IDP provides the old token even past its `exp` which is clearly defective behaviour. 

1. Are there any workarounds at the moment that would allow the stored token to be removed? Would a forced logout via the API for the given user force the stored token to be removed?
2. @pedroigor the ticket here is closed. Are there any plans to merge this patch?
3. The fix seems to be to re-issue the token based on offline/refresh tokens (please correct me if I misunderstood the patch). What happens to IDPs that do not issue `offline` or `refresh` tokens?

Thanks

Bademeischta:
@edamato Thank you for confirming this! It validates that this is indeed a generic issue in the current codebase and not isolated to my environment.

Regarding your question: Yes, my fix explicitly handles the persistence of the token after a successful refresh. 
If the IDP does not issue offline/refresh tokens: Then Keycloak cannot refresh the session anyway (standard OAuth2 behavior), so this fix wouldn't apply there.
But (your case): Since you are seeing an "old token past its exp", it implies Keycloak has a session and a refresh token, but is failing to update the DB with the fresh access token that it fetched internally. That is exactly what this PR fixes.

@ahus1 @pedroigor Since we now have independent confirmation from another user that this bug affects production environments (returning expired tokens), could you please reopen this PR and the associated Issue?

I am still ready to assist with the test case adjustments if the previous NPEs in the test suite were the blocker, but the fix itself seems necessary for the community.

ahus1:
@edamato / @Bademeischta - in order to proceed, we need a reproducer that can later be turned into a test case. 

Given @pedroigor's latest comment, the test didn't relate to the code changes: "the test is still passing without your changes."

If @edamato has a reproducer for this, please create a new GitHub bug issue.

edamato:
@ahus1 could you please clarify what a reproducer requires from your point of view? 

Are you talking about me providing you logs and other data or an actual end to end reproducer that would allow you to run an e2e process that catches the problem?


I have two systems where this is happening: 

- The target external IDP is MS EntraID. 
- I have not tried the same procedure against a different IDP. 
- brokering is using authz code as normal. 
- refresh tokens are issued as part of the initial token. 
- the broker is OIDC v1 configured to store tokens and set stored tokens to readable. 
- the broker is configured to disable user info due to requirements from the EntraID client and scopes are very standard. 

What I see:

- initial token works well
- after exp keycloak keeps on providing the expired token to my requests
- even if it doesn't give me a properly refreshed token it should not give me an expired token 
- I have not found any way of flushing this expired token apart from re establishing the external IDP link again via the full user authz code flow. 


Edit:
- as an afterthought maybe it gives me the expired token because the refresh token is part of it but nonetheless I'd expect a refreshed token on my behalf and not having to test and go directly to EntraID to refresh the token. 
- if only this keycloak IDP is allowed to do external connections then serving the old token wouldn't help the client app.

edamato:
@Bademeischta thank you for your reply. Does your patch also allows for us to see if the actual refresh request to the upstream idp fails? Or do we just enable tracing for that?

edamato:
Opened new ticket

https://github.com/keycloak/keycloak/issues/46373

PR #45285, opened 2026-01-09
CIMD - Persistent CIMD

tnorimat:
closes #45284

## Class structure

Mainly there are two components:
- Processing CIMD/MCP specification
  - `AbstractClientIdMetadataDocumentExecutor` / `AbstractClientIdMetadataDocumentExecutorFactory` : can provide basic features regarding specifications.
  - `ClientIdMetadataDocumentExecutor` / `ClientIdMetadataDocumentExecutorFactory` : can extends the basic abstract class.
- Storing and managing client metadata
  - `ClientIdMetadataDocumentProviderSpi` / `ClientIdMetadataDocumentProvide` / `ClientIdMetadataDocumentProvideFactory` : SPI
  - `AbstractPersistentClientIdMetadataDocumentProvider` / `AbstractPersistentClientIdMetadataDocumentProviderFactory`: can provide basic basic features regarding persisting client metadata.
  - `PersistentClientIdMetadataDocumentProvider` / `PersistentClientIdMetadataDocumentProviderFactory`: can extends the basic abstract class.

The size of the PR is large, but about 1.2k of the PR codes are test codes and the codes includes a lot of code comments.

## Implementation patterns
Both adopt Factory pattern to make it easy to custom/extend the features.

### Processing CIMD/MCP specification component
Client policies are used. Its advantage is that we can construct client policy profile for each version of MCP. To apply the profile, the following condition is also added:
- `ClientIdUriSchemeCondition` / `ClientIdUriSchemeConditionFactory` : filter the request whose `client_id` starts with http/https.

### Regarding Storing and managing a client metadata component
To make it easy to implement non-persistenting client metadata, new SPI is introduced. 

### Displaying information on client metadata on the grant screen
Augmentation of client metadata in `PersistentClientIdMetadataDocumentProvider`, the information of fetched client metadata is shown on the existing grant screen.

### Integration tests
`TestingOIDCEndpointsApplicationResource` simulates the server hosting client metadata.

## Enchancement and customization:
- If we want to add keycloak-specific policies for Client ID validation and/or Client Metadata: modify `ClientIdMetadataDocumentExecutor` / `ClientIdMetadataDocumentExecutorFactory`.
- If keycloak users want to add their own Client ID validation and/or Client Metadata: extend `ClientIdMetadataDocumentExecutor` / `ClientIdMetadataDocumentExecutorFactory`.
- If we want to add keycloak-specific features/augmentation to client metadata: modify  `PersistentClientIdMetadataDocumentProvider` / `PersistentClientIdMetadataDocumentProviderFactory`
- If keycloak users want to add their own features/augmentation to client metadata: extend `PersistentClientIdMetadataDocumentProvider` / `PersistentClientIdMetadataDocumentProviderFactory`
- If we want to support non-persisting client metadata: implement `ClientIdMetadataDocumentExecutor` / `ClientIdMetadataDocumentExecutorFactory`

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
@@ -4,6 +4,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import org.keycloak.OAuth2Constants;
 import org.keycloak.models.KeycloakSession;
@@ -41,12 +42,23 @@ public class ClientIdUriSchemeCondition extends AbstractClientPolicyConditionPro
         @JsonProperty(ClientIdUriSchemeConditionFactory.CLIENT_ID_URI_SCHEME)
         protected List<String> clientIdUriSchemes = Collections.emptyList();
 
+        @JsonProperty(ClientIdUriSchemeConditionFactory.TRUSTED_DOMAINS)
+        protected List<String> trustedDomains = null;
+
         public List<String> getClientIdUriSchemes() {
             return clientIdUriSchemes;
         }
 
         public void setClientIdUriSchemes(List<String> clientIdUriSchemes) {
             this.clientIdUriSchemes = clientIdUriSchemes;
+        }
+
+        public List<String> getTrustedDomains() {
+            return trustedDomains;
+        }
+
+        public void setTrustedDomains(List<String> permittedDomains) {
+            this.trustedDomains = permittedDomains;
         }
     }
 
@@ -61,7 +73,7 @@ public class ClientIdUriSchemeCondition extends AbstractClientPolicyConditionPro
             case PRE_AUTHORIZATION_REQUEST:
                 PreAuthorizationRequestContext paContext = (PreAuthorizationRequestContext) context;
                 String clientId = ((PreAuthorizationRequestContext) context).getRequestParameters().getFirst(OAuth2Constants.CLIENT_ID);
-                if (isUriSchemeMatched(clientId)) return ClientPolicyVote.YES;
+                if (isUriSchemeMatched(clientId) && isTrustedDomainMatched(clientId)) return ClientPolicyVote.YES;
                 return ClientPolicyVote.NO;
             default:
                 return ClientPolicyVote.ABSTAIN;
@@ -82,5 +94,49 @@ public class ClientIdUriSchemeCondition extends AbstractClientPolicyConditionPro
         }
 
         return configuration.getClientIdUriSchemes().stream().anyMatch(i->i.equals(uri.getScheme()));
+    }
+
+    private boolean isTrustedDomainMatched(String clientId) {
+        List<String> trustedDomains = convertContentFilledList(configuration.getTrustedDomains());
+        if (trustedDomains == null || trustedDomains.isEmpty()) {
+            logger.debug("trusted domain list is vacant.");
+            return false;
+        }
+
+        final URI uri;
+        try {
+            uri = new URI(clientId);
+        } catch (URISyntaxException e) {
+            logger.warnv("Malformed URL: {0}", clientId);
+            return false;
+        }
+
+        if (uri.getHost() == null) {
+            logger.warnv("not trusted domain: host = {0}", uri.getHost());
+            return false;
+        }
+
+        if (trustedDomains.stream().noneMatch(i->checkTrustedDomain(uri.getHost(), i))) {
+            logger.warnv("not trusted domain: host = {0}", uri.getHost());
+            return false;
+        }
+
+        return true;
+    }
+
+    private List<String> convertContentFilledList(List<String> list) {
+        if (list == null) {
+            return Collections.emptyList();
+        }
+        return list.stream().filter(Objects::nonNull).filter(i->!i.isBlank()).distinct().toList();
+    }
+
+    // apply the same logic in TrustedHostClientRegistrationPolicy.
+    private boolean checkTrustedDomain(String hostname, String trustedDomain) {
+        if (trustedDomain.startsWith("*.")) {
+            String domain = trustedDomain.substring(2);
+            return hostname.equals(domain) || hostname.endsWith("." + domain);
+        }
+        return hostname.equals(trustedDomain);
     }
 }
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
@@ -1,0 +1,86 @@
+package org.keycloak.protocol.oauth2.cimd.clientpolicy.condition;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Collections;
+import java.util.List;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.representations.idm.ClientPolicyConditionConfigurationRepresentation;
+import org.keycloak.services.clientpolicy.ClientPolicyContext;
+import org.keycloak.services.clientpolicy.ClientPolicyException;
+import org.keycloak.services.clientpolicy.ClientPolicyVote;
+import org.keycloak.services.clientpolicy.condition.AbstractClientPolicyConditionProvider;
+import org.keycloak.services.clientpolicy.context.PreAuthorizationRequestContext;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jboss.logging.Logger;
+
+/**
+ * The class is a condition of client policies. On {@code PRE_AUTHORIZATION_REQUEST} event,
+ * it checks if the value of {@code client_id} parameter is URI and
+ * the scheme part of the URI is the one defined in its configuration.
+ *
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public class ClientIdUriSchemeCondition extends AbstractClientPolicyConditionProvider<ClientIdUriSchemeCondition.Configuration> {
+
+    private static final Logger logger = Logger.getLogger(ClientIdUriSchemeCondition.class);
+
+    public ClientIdUriSchemeCondition(KeycloakSession session) {
+        super(session);
+    }
+
+    @Override
+    public Class<ClientIdUriSchemeCondition.Configuration> getConditionConfigurationClass() {
+        return ClientIdUriSchemeCondition.Configuration.class;
+    }
+
+    public static class Configuration extends ClientPolicyConditionConfigurationRepresentation {
+        @JsonProperty(ClientIdUriSchemeConditionFactory.CLIENT_ID_URI_SCHEME)
+        protected List<String> clientIdUriSchemes = Collections.emptyList();
+
+        public List<String> getClientIdUriSchemes() {
+            return clientIdUriSchemes;
+        }
+
+        public void setClientIdUriSchemes(List<String> clientIdUriSchemes) {
+            this.clientIdUriSchemes = clientIdUriSchemes;
+        }
+    }
+
+    @Override
+    public String getProviderId() {
+        return ClientIdUriSchemeConditionFactory.PROVIDER_ID;
+    }
+
+    @Override
+    public ClientPolicyVote applyPolicy(ClientPolicyContext context) throws ClientPolicyException {
+        switch (context.getEvent()) {
+            case PRE_AUTHORIZATION_REQUEST:
+                PreAuthorizationRequestContext paContext = (PreAuthorizationRequestContext) context;
+                String clientId = ((PreAuthorizationRequestContext) context).getRequestParameters().getFirst(OAuth2Constants.CLIENT_ID);
+                if (isUriSchemeMatched(clientId)) return ClientPolicyVote.YES;
+                return ClientPolicyVote.NO;
+            default:
+                return ClientPolicyVote.ABSTAIN;
+        }
+    }
+
+    private boolean isUriSchemeMatched(String clientId) {
+        if (clientId == null || configuration.getClientIdUriSchemes() == null || configuration.getClientIdUriSchemes().isEmpty()) {
+            return false;
+        }
+
+        final URI uri;
+        try {
+            uri = new URI(clientId);
+        } catch (URISyntaxException e) {
+            logger.debugv("not URL: clientId = {0}", clientId);
+            return false;
+        }
+
+        return configuration.getClientIdUriSchemes().stream().anyMatch(uri.getScheme()::equals);
+    }
+}
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
@@ -81,6 +81,6 @@ public class ClientIdUriSchemeCondition extends AbstractClientPolicyConditionPro
             return false;
         }
 
-        return configuration.getClientIdUriSchemes().stream().anyMatch(uri.getScheme()::equals);
+        return configuration.getClientIdUriSchemes().stream().anyMatch(i->i.equals(uri.getScheme()));
     }
 }
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
@@ -10,7 +10,7 @@ import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProvide
 
 /**
  * The class is the factory class of {@link ClientIdUriSchemeCondition}.
- * It provides one configuration: Scheme part of the URI.
+ * It provides two configuration: Scheme part and host part of the URI.
  *
  * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
  */
@@ -19,6 +19,7 @@ public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyCondi
     public static final String PROVIDER_ID = "client-id-uri";
 
     public static final String CLIENT_ID_URI_SCHEME = "client-id-uri-scheme";
+    public static final String TRUSTED_DOMAINS = "client-id-uri-allow-permitted-domains";
 
     private static final List<ProviderConfigProperty> configProperties = new ArrayList<>();
 
@@ -31,6 +32,20 @@ public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyCondi
                 "URI scheme",
                 "Scheme part of the URI",
                 ProviderConfigProperty.MULTIVALUED_STRING_TYPE, null);
+        configProperties.add(property);
+
+        property = new ProviderConfigProperty(
+                TRUSTED_DOMAINS,
+                "Trusted domains",
+                "If some domains are filled, The condition evaluates to true " +
+                        "if the host part of client_id parameter in an authorization request matches one of the filled domains. " +
+                        "Otherwise, the condition evaluates to false " +
+                        "The domains are checked by using regex. " +
+                        "If the domains not filled, the condition evaluate to false regardless of the client_id parameter value. " +
+                        "For example, use pattern like this '(.*)\\.example\\.org' if you want to accept the parameter / property whose domain is 'example.org'." +
+                        "Don't forget to use escaping of special characters like dots as otherwise dot is interpreted as any character in regex!",
+                ProviderConfigProperty.MULTIVALUED_STRING_TYPE,
+                null);
         configProperties.add(property);
     }
 
@@ -46,7 +61,8 @@ public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyCondi
 
     @Override
     public String getHelpText() {
-        return "The condition checks whether client_id is URI and its scheme.).";
+        return "The condition checks that the scheme part of client_id parameter matches one of the filled ones " +
+               "and the host part of the client_id matches one of the filled domains.";
     }
 
     @Override
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
@@ -1,0 +1,56 @@
+package org.keycloak.protocol.oauth2.cimd.clientpolicy.condition;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.services.clientpolicy.condition.AbstractClientPolicyConditionProviderFactory;
+import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProvider;
+
+/**
+ * The class is the factory class of {@link ClientIdUriSchemeCondition}.
+ * It provides one configuration: Scheme part of the URI.
+ *
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyConditionProviderFactory {
+
+    public static final String PROVIDER_ID = "client-id-uri";
+
+    public static final String CLIENT_ID_URI_SCHEME = "client-id-uri-scheme";
+
+    private static final List<ProviderConfigProperty> configProperties = new ArrayList<>();
+
+    static {
+        addCommonConfigProperties(configProperties);
+
+        ProviderConfigProperty property;
+        property = new ProviderConfigProperty(
+                CLIENT_ID_URI_SCHEME,
+                "URI scheme",
+                "Scheme part of the URI",
+                ProviderConfigProperty.MULTIVALUED_STRING_TYPE, null);
+        configProperties.add(property);
+    }
+
+    @Override
+    public ClientPolicyConditionProvider create(KeycloakSession session) {
+        return new ClientIdUriSchemeCondition(session);
+    }
+
+    @Override
+    public String getId() {
+        return PROVIDER_ID;
+    }
+
+    @Override
+    public String getHelpText() {
+        return "The condition checks whether client_id is URI and its scheme.).";
+    }
+
+    @Override
+    public List<ProviderConfigProperty> getConfigProperties() {
+        return configProperties;
+    }
+}
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
@@ -3,7 +3,10 @@ package org.keycloak.protocol.oauth2.cimd.clientpolicy.condition;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.keycloak.Config;
+import org.keycloak.common.Profile;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.provider.EnvironmentDependentProviderFactory;
 import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.services.clientpolicy.condition.AbstractClientPolicyConditionProviderFactory;
 import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProvider;
@@ -14,7 +17,8 @@ import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProvide
  *
  * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
  */
-public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyConditionProviderFactory {
+public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyConditionProviderFactory
+        implements EnvironmentDependentProviderFactory {
 
     public static final String PROVIDER_ID = "client-id-uri";
 
@@ -68,5 +72,10 @@ public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyCondi
     @Override
     public List<ProviderConfigProperty> getConfigProperties() {
         return configProperties;
+    }
+
+    @Override
+    public boolean isSupported(Config.Scope config) {
+        return Profile.isFeatureEnabled(Profile.Feature.CIMD);
     }
 }
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
@@ -12,6 +12,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.Response;
@@ -413,6 +414,8 @@ public abstract class AbstractClientIdMetadataDocumentExecutor<CONFIG extends Ab
     // Client Metadata Validation Errors
     public static final String ERR_METADATA_URIS_SAMEDOMAIN = "Invalid Client Metadata: client_id parameter, redirect_uri parameter and at least one of redirect_uris properties in client metadata should be under the same domain.";
     public static final String ERR_METADATA_NO_REQUIRED_PROPERTIES = "Invalid Client Metadata: it does not include all required properties.";
+    public static final String ERR_METADATA_NO_ALL_URIS_SAMEDOMAIN = "Invalid Client Metadata: some uri property is not under the same permitted domain";
+
 
     // Implementation
     // Fetch Client Metadata
@@ -744,10 +747,6 @@ public abstract class AbstractClientIdMetadataDocumentExecutor<CONFIG extends Ab
             return;
         }
 
-        // NOTE: java.net.URI does not follow RFC 3986, so it cannot parse the following IPv6 address, resulting getHost() == null
-        //       It follows RFC 2396 and RFC 2732.
-        // ex.  0:0:0:0:0:0:0:1  -> getHost() == null
-        //     [0:0:0:0:0:0:0:1] -> getHost() == [0:0:0:0:0:0:0:1]
         if (uri.getHost() == null) {
             getLogger().warnv("not trusted domain: host = {0}", uri.getHost());
             throw invalidClientIdMetadata(ERR_HOST_UNRESOLVED);
@@ -795,7 +794,7 @@ public abstract class AbstractClientIdMetadataDocumentExecutor<CONFIG extends Ab
         // An authorization server MAY impose restrictions or relationships
         // between the redirect_uris and the client_id or client_uri properties
 
-        // same domain policy
+        // same domain policy: client_id, redirect_uri, redirect_uris, client_uri, logo_uri, tos_uri,, policy_uri, jwks_uri
         List<String> trustedDomains = convertContentFilledList(getConfiguration().getTrustedDomains());
         if (getConfiguration().isRestrictSameDomain() && trustedDomains != null && !trustedDomains.isEmpty()) {
             // Client Metadata verification ensures that
@@ -805,6 +804,21 @@ public abstract class AbstractClientIdMetadataDocumentExecutor<CONFIG extends Ab
             if (trustedDomains.stream().noneMatch(i->checkTrustedDomain(clientIdURI.getHost(), i) && checkTrustedDomain(redirectUriURI.getHost(), i))) {
                 getLogger().warnv("client_id and redirect_uri domain not match: client_id host part = {0}, redirect_uri host part = {1}", clientIdURI.getHost(), redirectUriURI.getHost());
                 throw invalidClientIdMetadata(ERR_METADATA_URIS_SAMEDOMAIN);
+            }
+
+            List<String> l = Stream.of(clientOIDC.getClientId(), clientOIDC.getClientUri(), clientOIDC.getLogoUri(), clientOIDC.getTosUri(), clientOIDC.getPolicyUri(), clientOIDC.getJwksUri())
+                    .filter(Objects::nonNull).toList();
+            try {
+                for (String s : l) {
+                    URI u = new URI(s);
+                    if (trustedDomains.stream().filter(i->!i.isBlank()).noneMatch(i -> checkTrustedDomain(u.getHost(), i) && checkTrustedDomain(redirectUriURI.getHost(), i))) {
+                        getLogger().warnv("not under the same domain = {0}", u.getHost());
+                        throw invalidClientIdMetadata(ERR_METADATA_NO_ALL_URIS_SAMEDOMAIN);
+                    }
+                }
+            } catch (URISyntaxException e) {
+                getLogger().warnv("URI not resolved {0}}", e);
+                throw invalidClientIdMetadata(ERR_METADATA_NO_ALL_URIS_SAMEDOMAIN);
             }
         }
 
--- a/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
+++ b/services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
@@ -105,10 +105,15 @@ public abstract class AbstractClientIdMetadataDocumentExecutor<CONFIG extends Ab
     protected CONFIG configuration;
     protected ClientIdMetadataDocumentProvider<CONFIG> provider;
 
+    // Factory Global Setting:
+    // Client ID Metadata Document Provider Name
+    protected String cimdProviderName; // non-null
+
     protected abstract Logger getLogger();
 
-    protected AbstractClientIdMetadataDocumentExecutor(KeycloakSession session) {
+    protected AbstractClientIdMetadataDocumentExecutor(KeycloakSession session, String cimdProviderName) {
         this.session = session;
+        this.cimdProviderName = cimdProviderName;
     }
 
     protected ClientIdMetadataDocumentProvider<CONFIG> getProvider() {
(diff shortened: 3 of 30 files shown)


mposolda:
@tnorimat Thanks for the PR. @rmartinc will take a look at your PR in more details. Just one quick thing I've noticed that flaky test reported on this PR ( `org.keycloak.testsuite.client.policies.ClientIdMetadataDocumentTest#testClientIdMetadataDocumentExecutor` ) seems like caused by this PR?

tnorimat:
@mposolda Thank you for the comment. I will resolve the error.
The PR's code side is large, so I will add code comments to make the PR easily understandable. After that, I will swich the draft PR to the review-ready PR and let you know.

For the test of the PR, I wrote the integration test. Moreover, I prepated [the server hosting client metadata](https://github.com/tnorimat/cimd-server) and check the PR manually.

rmartinc:
Thanks @tnorimat! I need to also read the standards, so I'll try to do it before you move the PR to ready for review.

tnorimat:
@mposolda @rmartinc Hello, the PR is ready. I completed writing code comments, enriching [comments](https://github.com/keycloak/keycloak/issues/45284#issue-3795186526) on the issue regarding the PR's specification and adding [comments](https://github.com/keycloak/keycloak/pull/45285#issue-3795419610) on the PR regarding its implementation.
I summarized the requirements of CIMD and MCP in [the google document sheet](https://docs.google.com/spreadsheets/d/1nCej9mro0ZLirE0FBus6lWL2UoBdGFXYTdqwA7FeoPY/edit?gid=546261892#gid=546261892).
I would appreciated it if you could review the PR. Many thanks.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
```diff
@@ -0,0 +1,86 @@
...
+
+        final URI uri;
+        try {
+            uri = new URI(clientId);
+        } catch (URISyntaxException e) {
+            logger.debugv("not URL: clientId = {0}", clientId);
+            return false;
+        }
+
+        return configuration.getClientIdUriSchemes().stream().anyMatch(uri.getScheme()::equals);
```
rmartinc: The scheme can be null. I prefer to change to change the condition, but if we use the scheme, it can be null, for example if you use any common client_id like `test-app`, the scheme will be null.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
```diff
@@ -0,0 +1,865 @@
...
+        @JsonProperty(AbstractClientIdMetadataDocumentExecutorFactory.REQUIRED_PROPERTIES)
+        protected List<String> requiredProperties = null;
+        @JsonProperty(AbstractClientIdMetadataDocumentExecutorFactory.CONSENT_REQUIRED)
+        protected boolean consentRequired = true;
+        @JsonProperty(AbstractClientIdMetadataDocumentExecutorFactory.FULL_SCOPE_DISABLED)
+        protected boolean fullScopeDisabled = true;
+
+        // Client ID Metadata Document Provider Name
+        @JsonProperty(AbstractClientIdMetadataDocumentExecutorFactory.CIMD_PROVIDER_NAME)
+        protected String cimdProviderName = PersistentClientIdMetadataDocumentProviderFactory.PROVIDER_ID;
```
rmartinc: The CIMD executor is very dangerous. Note that, if we do not limit what domains are allowed, any attacker can add a metadata in an URL and it can create a client in keycloak with its own configuration. I like the idea of using client policy because then CIMD is not allowed by default and it should be explicitly added. But I also think we should limit the behavior by default.

The `allowPermittedDomains` or similar should be compulsory and it should be at least one. I mean, by default no domain is allowed if you don't explicitly add one. I even prefer using a similar syntax to redirect_uris (maybe we can re-use some code here) that can be called Allowed or Valid URIs . Something like: `https://app.sample.com/app/*` or exact `https://localhost:8080/metadata.json`.  The spec mentions some restrictions like https, path, no query or fragment,... I don't know if we should be strict or allow them with a relax option.

In my opinion there are too many options and we should simplify this a lot.

* Options like allow private, allow http scheme, restrict domain,... If using URIs or valid domains you can explicitly allow them adding `localhost`, `127.0.0.1` or whatever you need. I would not add options for this that also add a lot of DNS lookups. I think they can be just allowed using the previous more generic option (using domains or uris, whatever we decide).
* The restrict domain makes no sense, because we are always going to restrict because of security.
* I would move the provider name to a config property for the factory. It would be a global configuration but I think that it's enough.
* With this, I think we can just manage allowedDomains or allowedUris if we prefer the URI syntax, requiredProperties, consentRequired and fullScopeDisabled. We can add a switch to relax or strict the behavior (strict by default) and allow http and query string or other should restrictions in the spec.

More things I have been thinking reading the spec:
* Probably we need to a add min and max cache time. The min cache at least should be defined by default (at least a few minutes, for example 5min). I would use a factory option like I suggested for the provider (global setting).
* The spec mentions a limit of 5KB of returned data. The SimpleHttp has a limit but I think it's larger (around 10MB). We can also limit at factory level and use the option in the SimpleHttp to establish a different size limit if wanted.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
```diff
@@ -0,0 +1,865 @@
...
+        }
+
+        // Client identifier URLs MUST contain a path component.
+        if (uri.getPath() == null || uri.getPath().isEmpty()) {
+            getLogger().warn("Empty path:");
+            throw invalidClientId(ERR_CLIENTID_EMPTY_PATH);
+        }
+
+        // Client identifier URLs MUST NOT contain single-dot or double-dot path segments.
+        if (uri.getRawPath().contains("/./") || uri.getRawPath().contains("/../")) {
```
rmartinc: Take care with encoded characters, take a look to [RedirectUtils](https://github.com/keycloak/keycloak/blob/26.5.2/services/src/main/java/org/keycloak/protocol/oidc/utils/RedirectUtils.java#L151).

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
```diff
@@ -0,0 +1,352 @@
...
+            }
+        }
+        getLogger().debugv("client need to create: clientId = {0}", clientId);
+        return AbstractClientIdMetadataDocumentExecutor.FetchOperation.CREATE;
+    }
+
+    @Override
+    public ClientModel createClientMetadata(AbstractClientIdMetadataDocumentExecutor.OIDCClientRepresentationWithCacheControl clientOIDCWithCacheControl) throws ClientPolicyException {
+        // do the same thing as in dynamic client registration except for:
+        //   - not set client registration token
```
rmartinc, line 136: Is it possible to share the code instead of duplicating it? Or is this not exactly the same?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
```diff
@@ -0,0 +1,352 @@
...
+        } catch (Exception e) {
+            getLogger().warnv("Exception: {0}", e);
+            throw invalidClientMetadata("invalid request");
+        }
+    }
+
+    @Override
+    public ClientModel updateClientMetadata(AbstractClientIdMetadataDocumentExecutor.OIDCClientRepresentationWithCacheControl clientOIDCWithCacheControl) throws ClientPolicyException {
+        // do the same thing as in dynamic client registration except for:
+        //   - not set client registration token
```
rmartinc, line 205: Same here. if we can re-use the code, it's much better.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public Class<ClientIdUriSchemeCondition.Configuration> getConditionConfigurationClass() {
+        return ClientIdUriSchemeCondition.Configuration.class;
+    }
+
+    public static class Configuration extends ClientPolicyConditionConfigurationRepresentation {
+        @JsonProperty(ClientIdUriSchemeConditionFactory.CLIENT_ID_URI_SCHEME)
+        protected List<String> clientIdUriSchemes = Collections.emptyList();
```
rmartinc, line 43: I think that this is condition is too broad. My feeling is that we need to add the same restriction we add to the executor. If we decide to use the domain, the valid domains, if we decide to use a kind of valid URIs, the URIs.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/client/policies/ClientIdMetadataDocumentTest.java
```diff
@@ -0,0 +1,1171 @@
...
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * This test class is for testing an executor of client policies.
+ * 
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+@EnableFeature(value = Profile.Feature.CIMD)
+public class ClientIdMetadataDocumentTest extends AbstractClientPoliciesTest {
```
rmartinc, line 73: We are using now the new test framework (in folder tests/base). Check if you think it's viable to use that. Not strictly necessary, but a good nice to have we can. There are some tests that are already using client policies to test JWT authorization grants for example. You can take a look. But if you see it's complicated, we can continue with  the old way.

rmartinc (changes requested):
Thanks @tnorimat!

The PR is promising but I think we should improve the security of the solution. The CIMD should be restricted to some domains/URIs by default for sure. I prefer to use valid URIs (similar to redirect URIs solution). We can also reduce the number of configuration properties. My proposal is moving some of them to factory (global) config and remove some of them in favor of valid domains/URIs (whatever we decide to use). WDYT?

I have done some other minor comments.

Probably we need a few round-trips for this PR. So, once we decide what configuration we use, we can start checking the details.

Thanks again!

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/ClientIdMetadataDocumentExecutor.java
```diff
@@ -0,0 +1,144 @@
...
+        return new ClientIdMetadataDocumentExecutor.Configuration();
+    }
+
+    public static class Configuration extends AbstractClientIdMetadataDocumentExecutor.Configuration {
+        // additional settings
+        // Client Metadata Validation
+        @JsonProperty(ClientIdMetadataDocumentExecutorFactory.ONLY_ALLOW_CONFIDENTIAL_CLIENT)
+        protected boolean onlyAllowConfidentialClient = false;
+        @JsonProperty(ClientIdMetadataDocumentExecutorFactory.ALL_URIS_RESTRICT_SAME_DOMAIN)
+        protected boolean allUrisRestrictSameDomain = false;
```
rmartinc: I would also check `allUrisRestrictSameDomain` all the time (I mean this property should be always true and cannot be disabled). I would not allow any URI outside the valid domains or URIs.

tnorimat:
@rmartinc Thank you for the review. I will check your review comments.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
tnorimat: Thank you. I will use `UNSAFE_PATH_PATTERN`.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/ClientIdMetadataDocumentExecutor.java
tnorimat: Yes, I agree on that. I will set `true` to the property as default.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/client/policies/ClientIdMetadataDocumentTest.java
tnorimat: I'd love to try out a new testing framework, but I'd rather prioritize other things. Is that OK?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
tnorimat: I designed the condition that it can be applied to not only CIMD but also other specification like SPIFFE.

I am afraid that if we added the same restriction of the CIMD executor to the condition, the check of the restriction whould duplicatets (by condition and executor).

I think that the condition plays the role of filtering `client_id` complies with CIMD and the executor plays the role of verifing and validating the `client_id`. What do you think about this idea?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
tnorimat: As the comments said, there are slightly difference between the codes for creating/updating a client by Dynamic Client Registration (DCR) by `OIDCClientRegistrationProvider` and `AbstractClientRegistrationProvider` and CIMD.

For example, the former issues `client_secret` while the latter does not.

In fact, both can share the same codes. To do that, we need to modify the codes of `OIDCClientRegistrationProvider` and `AbstractClientRegistrationProvider`. I was aware of this, but was hesitant to change existing, working code.

Is it better to modify the codes of  `OIDCClientRegistrationProvider` and `AbstractClientRegistrationProvider` to shere the codes with the CIMD? If so, I will do that.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
tnorimat: the same above.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
tnorimat: Yes, the line 84 shouldbe
```
        return configuration.getClientIdUriSchemes().stream().anyMatch(i->i.equals(uri.getScheme()));
```

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
tnorimat: I agreed to the following points:

- The configurations of the CIMD executor:
  - Reducing the number of the configurations
  - Moving some of them to the configuration of the factory (global settings): the following can be moved:
    - CIMD provider name: I think it should be global setting because if we changed the CIMD provider during Keycloak is running (ex. persistent CIMD provier to non-persistent CIMD provider), the Keycloak might not work well.

- Allow Permitted Domain policy in Client ID Metadata Validation:
  - Secure default setting:  the policy should be activated by default.
  - Secure default setting: the permitted domain list is vacant by defaut.
  - Secure default operatoin: any `client_id` is denied if the permitted domain list is vacant or has only null values
  - Using the same syntax to `redirect_uris`
 
- Client Metadata Caching:
  - Min and max time settings: these are included in the factory configuration, namely global configuration. The min time default value is 5 min. The max time default value is TBD.
  - The upper limit of the metadata: it is included in the factory configuration, namely global configuration. Its default value is 5KB.

The follwing is my proposal:

- The configurations of the CIMD executor:
  - Removing the configurations: `consentRequired` and `fullScopeDisabled` can be removed. Instead, the client's `consentRequired` and `fullScopeDisabled` setting should be `true` for security. Accodring to the CIMD spec, it requres an authorization server to show an end-user the hostname of `client_id` on the consent screen.
  - Performance - Avoid DNS lookups as possible: at least one lookup is needed because the CIMD spec does not allow private address that is resolved from `client_id` and any URI-formatted properties in Client Metadata like `logo_uri`. To reduce the chance of the lookup, In Client ID verification phase, this lookup is executed after all verification for `client_id` are done. In Client Metadata verification phase, this lookup is executed after all verification for any URI-formatted properties in Client Metadata are done.

The questions about your commets is as follows:
- Same Domain policy `restrictSameDomain` in Client ID Validation: do you mean that it is not needed if Allow Permitted Domain policy is activated as default? I think that the following case is skeptical where `client_id` and `redirect_uri` are under the permitted domain but the domain of `client_id` and `redirect_uri` is different.
- Same Domain policy `allUrisRestrictSameDomain` in Client Metadata Validation: the same as above.

Looking forward to hearing your view. Also please let me know if I missed the points you commented.

tnorimat:
@rmartinc Hello, thank you for the first-round review. I replied your comments. I would appreciated it if you would check them. 
Regarding the points I agreed, I will revise the code of the PR to incopororate the points.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
rmartinc: Thanks @tnorimat! I think that your proposal is OK and in the line I was thinking.

About `retrictSameDomain`, I didn't get that this means that all the client URL options should be in the same domain. You are right and it's OK then to be maintained. Regarding `allUrisRestrictSameDomain`, I would do something similar to `restrictSameDomain`, by default all URLs in the client should be in the allowed ones (this cannot be avoided by configuration). But if `restricSameDomain` is true, then all the other URLs in the metadata should be in the same domain too. I would just use one option for this restriction. My intention here was to avoid any issue (XSS or whatever) accessing a non-validated URL (logo, jwks or whatever). WDYT?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
rmartinc: Yep, I understand that the condition duplicates the check, but IMHO the problem is that if you want to setup two different policies for two different types of CIMD clients. For example `https://www.app1.com/*` and `https://www.app2.com/*` you cannot do that. With this condition you need to setup just one policy for all your CIMD clients (always). Because if you setup two, the executor in the first one will allow or reject and the second policy will never be executed. I think that you need to have the same condition although the configuration can be broader than in the executor. Note also that you can have a client (not CIMD, just a normal one) that can have URI type clientId (for example `https://www.internal.com/myapp`) that will go to the CIMD executor and you cannot avoid that because there is no condition that allows using the URL as a pattern. WDYT?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
rmartinc: :+1:

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
rmartinc: :+1:

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/ClientIdMetadataDocumentExecutor.java
rmartinc: As I commented before, I would use one `restrictSameDomain`, you can use true or false as default (whatever you prefer). But at least I would check that all URLs are inside the valid ones. This way no URL in the metadata is not checked, all of them are validated.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
rmartinc: This was just a suggestion. I prefer to re-use the code but if you think that it's going to be problematic or convoluted you can continue doing it this way.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
rmartinc: Same, just a suggestion.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/client/policies/ClientIdMetadataDocumentTest.java
rmartinc: Yes, no problem at all. I understand that this is not a priority.

rmartinc (changes requested):
Thanks @tnorimat for considering my suggestions. I think that I have gone through all your questions, if I missed something or you have some doubt, please just let me know.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
tnorimat: I fixed the point.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
tnorimat: I fixed the point by using the regex parttern:
```
    // any access to parent folder /../ or current /./ is unsafe with or without encoding
    private final static Pattern UNSAFE_PATH_PATTERN = Pattern.compile(
            "(/|%2[fF]|%5[cC]|\\\\)(%2[eE]|\\.){1,2}(/|%2[fF]|%5[cC]|\\\\)|(/|%2[fF]|%5[cC]|\\\\)(%2[eE]|\\.){1,2}$");
```

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
tnorimat: I might consider re-using the code by refactoring `OIDCClientRegistrationProvider` and `AbstractClientRegistrationProvider`, but I'd rather prioritize other things.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
tnorimat: I might consider re-using the code by refactoring `OIDCClientRegistrationProvider` and `AbstractClientRegistrationProvider`, but I'd rather prioritize other things.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/client/policies/ClientIdMetadataDocumentTest.java
tnorimat: Thank you for your understanding.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
tnorimat: Yes, I agree with you. So I would like to add the check the same as Allow Permitted Domains of the CIMD executor.
By doing so, we can filter an authorization request based on `client_id`'s domain part. 
Are there any other kinds of checks to filter requests?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/ClientIdMetadataDocumentExecutor.java
tnorimat: Thank you. I will do the following:

Parameters and properties whose values are URI:
- Authorization Request: `client_id`, `redirect_uri`
- Client Metadata: `client_id`, `redirect_uris`, `logo_uri`, `tos_uri`, `policy_uri`, `client_uri`

The executer checks if they belongs to one of Permitted Domains (`allowPermittedDomains`). The check is always carried out.
The executer checks if they belongs to the same doain. The check is carried out only if its setting `restrictSameDomain` is `ON`.

tnorimat:
@rmartinc Thank you for the reply.

The points that have been agreed upon in the discussions so far are listed below.  I would appreciated it if you could check them.

* The configurations of the CIMD executor:
  - Reducing the number of the configurations
  - Moving some configurations to the factory (global settings):
    * CIMD provider name
  - Changing the meanings of the configurations:
    * Allow Permitted Domains:
      If the list of the permitted domains null, vacant or only includes null entries, then the value of the following URI-formatted parameter is denied:
      - Authorization Request parameter: `client_id`, `redirect_uris`
      - Client Metadata properties: `client_id`, `redirect_uris`, `jwks_uri`, `logo_uri`, `policy_uri`, `tos_uri`, `client_uri`
    * Restrict Same Domain:
      If the configuration is ON, then the executer checks if all the value of the above URI-formatted parameters belong to the same doain.
  - Removing some configurations:
    * All URIs Restrict Same Domain
    * Consent Required
    * Full Scope Allowed

* Using the same terms and syntax for both `client_id` and `redirect_uri`

* Performance
  - Avoid DNS lookups as possible: 
  At least one lookup is needed because the CIMD spec does not allow private address that is resolved from `client_id` and any URI-formatted properties in Client Metadata like `logo_uri`. To reduce the chance of the lookup, In Client ID verification phase, this lookup is executed after all verification for `client_id` are done. In Client Metadata verification phase, this lookup is executed after all verification for any URI-formatted properties in Client Metadata are done.

* The configuration of the factory of the CIMD executor as global settings:
  - Min Cache Time
  - Max Cache Time
  - Upper Limit of the Client Metadata
  - CIMD Provider Name (moved from the configuratin of the CIMD executor)

* The condition for applying the CIMD executor
  - Adding allow permitted domain check the same as the CIMD executor

* Low priority tasks
  - Refactoring and sharing the same codes for craeting/updating a client by DCR and CIMD
  - Testing by new test framework

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeCondition.java
rmartinc: I would say that no. Just using the URL in a similar way than in the executor.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/ClientIdMetadataDocumentExecutor.java
rmartinc: :+1:

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
rmartinc: :+1:

services/src/main/java/org/keycloak/protocol/oauth2/cimd/provider/AbstractPersistentClientIdMetadataDocumentProvider.java
rmartinc: :+1:

rmartinc (changes requested):
Thanks @tnorimat!

rmartinc:
Hey @tnorimat! Please ping me here when you think that this PR is ready. I see the condition still checks only the scheme so I think you are still working on it. You can also move to draft while working on it, whatever you prefer.
Thanks!

tnorimat:
@rmartinc Hello. The remaining fix point is only `ClientIdUriSchemeCondition`. I will add the commit for that today or tomorrow. After adding the commit, I will ping you. Thank you.

tnorimat:
@rmartinc Hello, I finished and pushed the final commit and the PR is ready for re-review. Could you re-review the PR?

tnorimat:
@rmartinc VS Code supported CIMD as an MCP client side, so I used it to check if Keycloak works well in the real CIMD usecase. It seems that it works well.

keycloak

vscode

accesstoken

tnorimat:
@rmartinc I tried to write test codes by using the new test framework (junit5) but gave it up.
The reason is that the PR uses `testsuite-providers` for hosting client metadata and jwks keys but it is not yet ported to the new test framework. It takes a lot of codes and time to port it to the the new test framework in the PR. It might be better to handle it by other PR.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutorFactory.java
```diff
@@ -0,0 +1,137 @@
...
+
+        property = new ProviderConfigProperty(
+                REQUIRED_PROPERTIES,
+                "Required properties",
+                "If client metadata does not include all the properties, the executor does not accept the client metadata.",
+                ProviderConfigProperty.MULTIVALUED_STRING_TYPE,
+                null);
+        configProperties.add(property);
+    }
+
```
rmartinc, line 118: @tnorimat Here you can add your factory properties with the method `getConfigMetadata` . This way they are automatically added to the list of options that you can add at startup. Something like:

```java
    @Override
    public List<ProviderConfigProperty> getConfigMetadata() {
        return ProviderConfigurationBuilder.create()
                .property()
                .name(CONFIG_CIMD_PROVIDER_NAME)
                .type("string")
                .helpText("Provider to use for the CIMD.")
                .defaultValue(ClientIdMetadataDocumentExecutorFactoryProviderConfig.DEFAULT_CONFIG_CIMD_PROVIDER_NAME)
                .add()
                ...
                .build();
```

You can check the `all-provider-config.adoc` is properly updated when you add the options here.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
```diff
@@ -0,0 +1,880 @@
...
+        }
+
+        public final String getNormalizedCacheControlHeaderValue() {
+            return normalizedCacheControlHeaderValue;
+        }
+
+        public int getCacheExpiryTimeInSec() {
+            if (isNoCache() || isNoStore()) {
+                return Time.currentTime();
+            }
```
rmartinc, line 272: This means that `nocache` and `noStore` are not cached at all. I would only allow that if the minCache in the factory is set to 0 or -1 or similar. WDYT?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
```diff
@@ -0,0 +1,880 @@
...
+    public CONFIG getConfiguration() {
+        return configuration;
+    }
+
+    public static class Configuration extends ClientPolicyExecutorConfigurationRepresentation {
+        // Client ID Verification
+        @JsonProperty(AbstractClientIdMetadataDocumentExecutorFactory.ALLOW_PRIVATE_ADDRESS)
+        protected boolean allowPrivateAddress = false;
+        @JsonProperty(AbstractClientIdMetadataDocumentExecutorFactory.ALLOW_LOOPBACK_ADDRESS)
+        protected boolean allowLoopbackAddress = false;
```
rmartinc: I still think that `allowPrivateAddress` and `allowLoopbackaddress` is like duplicated if we use allow-list configuration for the domain.  If the admin allows domains `localhost` or `127.0.0.1`, he is allowing loopback, the same for the private addresses. WDYT? Do we remove this two options and related checks?

rmartinc (changes requested):
Thanks @tnorimat! Just three comments for your consideration. You can resolve all the previous threads.

Chatting with Marek and Stian we probably create two follow-up issues after merging this PR:

1. One to move the test to the new test-suite. It seems that we are going to force the new testsuite from now on. So everything new should use the new testsuite. The tests are OK in terms of completeness.
2. Instead of allowed domains I prefer to use allowed URIs (same syntax that valid redirect_uris for OIDC clients). I will file a new issue to see if we change the domain for an uri or at least to consider that idea.

But we will work on that after merging this one. So those two points are just for your information. Just the new three comments are for your consideration.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
tnorimat: Regarding `no-cache` directive, I agree to your idea. Accodring to [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.4), this directive does not means that the data must be cached in any case.

While regarding `no-store` directive, I think the current PR's handling of the directive might be good to comply with RFC 9111. According to [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.5), caching the data must not be allowed.

What do you think about the idea?

rmartinc: The [cimd spec](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00#section-4.4) says:

> The authorization server SHOULD respect HTTP cache headers [[RFC9111](https://www.rfc-editor.org/rfc/rfc9111)] when caching client metadata, but MAY define its own upper and/or lower bounds on an acceptable cache lifetime as well.

So I think we can enforce a min level of caching to ensure proper functionality and avoid too many requests. If you think that `nostore` should never cache, at least explain that in the min level prop of the factory.

tnorimat: I see. I will treat both `no-cache` and `no-store` directives the same, namely caching the data in the value of `minCache` configuration as you suggested.

I will priotize the directives as follows:
`no-cache`, `no-store` > `s-maxage` > `max-age`

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
tnorimat: Yes, I fixed that point.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
tnorimat: I fixed that point.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutorFactory.java
tnorimat: I fixed the point.

tnorimat:
@rmartinc Hello, I fixed all the three point you've mentioned. Could you check them?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
```diff
@@ -0,0 +1,834 @@
...
+
+        // The client_secret and client_secret_expires_at properties MUST NOT be used.
+        if (clientOIDC.getClientSecret() != null || clientOIDC.getClientSecretExpiresAt() != null) {
+            getLogger().warn("client metadata includes client_secret or client_secret_expires_at.");
+            throw invalidClientIdMetadata(ERR_METADATA_CLIENTSECRET);
+        }
+
+        // An authorization server MUST validate redirect URIs presented in an authorization request
+        // against those in the metadata document.
+        if (clientOIDC.getRedirectUris() == null || !clientOIDC.getRedirectUris().contains(redirectUri)) {
```
rmartinc: Probably should check the redirect URI like in the rest of cases:

```java
        if (clientOIDC.getRedirectUris() == null || RedirectUtils.verifyRedirectUri(session, clientOIDC.getClientUri(), redirectUri, Set.copyOf(clientOIDC.getRedirectUris()), true) == null) {
``` 
 Sorry, I just detected now.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
rmartinc: :+1: You can resolve this.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutorFactory.java
rmartinc: :+1: You can resolve this.

rmartinc (changes requested):
Thanks @tnorimat! It's almost Ok. Just one last request, rebase the PR to fix the Ci issues resolved in https://github.com/keycloak/keycloak/issues/46312. I have added another minor change to check the redirect uris like in the rest of clients (I have just seen it today and makes sense). With this, when the CI is OK, I'll approve it. Thanks a lot for the effort!

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutor.java
rmartinc: :+1: You can resolve this.

tnorimat:
@rmartinc Hello, I fixed the final point.

rmartinc (approved):
Thanks @tnorimat! Approved!

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
```diff
@@ -0,0 +1,72 @@
...
+import org.keycloak.services.clientpolicy.condition.AbstractClientPolicyConditionProviderFactory;
+import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProvider;
+
+/**
+ * The class is the factory class of {@link ClientIdUriSchemeCondition}.
+ * It provides two configuration: Scheme part and host part of the URI.
+ *
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public class ClientIdUriSchemeConditionFactory extends AbstractClientPolicyConditionProviderFactory {
```
mposolda: Since this condition is used primarily for the purpose of CIMD feature, would it makes sense to also mark it as `EnvironmentDependentFactory` enabled just if CIMD feature is enabled?

That will allow us some more flexibility as otherwise, we would need to preserve backwards compatibility, which may cause complications if we want to update this condition implementation later and possibly add new configuration options or rename existing configuration options etc (or eventually remove condition entirely).

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutorFactory.java
```diff
@@ -0,0 +1,149 @@
...
+ *     <li>Client Metadata Validation</li>
+ *     <ul>
+ *         <li>Restrict same domain: only allow {client_id} and {redirect_uri} parameter of an authorization request whose hostname is under the one of the permitted domain (wildcard * can be used)</li>
+ *         <li>Required properties: only allow a client metadata that includes all required properties</li>
+ *     </ul>
+ * </ul>
+ *
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public abstract class AbstractClientIdMetadataDocumentExecutorFactory implements ClientPolicyExecutorProviderFactory {
```
mposolda: Should this be rather EnvironmentDependentFactory enabled just if the CIMD feature is enabled? See my similar comment like for `ClientIdUriSchemeConditionFactory` .

mposolda (commented):
@tnorimat @rmartinc Thanks for the PR and review! Added 2 comments inline. Could you check please?

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/condition/ClientIdUriSchemeConditionFactory.java
tnorimat: Yes, I fixed the point as you suggested.

services/src/main/java/org/keycloak/protocol/oauth2/cimd/clientpolicy/executor/AbstractClientIdMetadataDocumentExecutorFactory.java
tnorimat: Yes, I fixed the point as you suggested.

tnorimat:
@mposolda Thank you for the review. I fixed the point you mentioned. Could you check the points?

rmartinc (approved):

mposolda (approved):
@tnorimat @rmartinc Thanks!

tnorimat:
@mposolda  Hello, as follow-up tasks, are updating [MCP guide](https://www.keycloak.org/securing-apps/mcp-authz-server) and adding CIMD documentation to Admin Document or Guide needed? If so, I will create the issues for them.

mposolda:
@tnorimat Hello, I think that guide would be better for CIMD. But right now, it is experimental feature, so it doesn't strictly need to be documented if setup/integration is still subject to change (considering other PRs opened for this etc). So up to you if you prefer to document this now or rather later when it is more polished.

tnorimat:
@mposolda Thank you for the comment. So, I will only update [MCP Guide](https://www.keycloak.org/securing-apps/mcp-authz-server) for keycloak 26.6.

PR #45305, opened 2026-01-09
Handle fragment when accessing the admin console

### LINKED ISSUE #45116: Realm-level admininistrators can no longer use Admin Console since 26.3.0 (UI fails to render)
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Starting with Keycloak 26.3.0, a realm-level administrator (user with the `realm-admin` role in a non-master realm) can no longer use the realm-specific Admin Console.

After authenticating directly against **the realm-level console URL** (e.g. /admin/_\<realm\>_/console/#/), the UI becomes largely unusable: most components do not render correctly and the web console is effectively non-functional.
Image

This workflow worked as expected in 26.2.5 and earlier versions: realm-level administrators were able to access the realm-specific admin console to perform maintenance tasks.

From 26.3.0 through 26.4.7 (latest at time of reporting), realm-level administration appears to be possible only by signing into the master realm Admin Console with a server administrator account and switching to the target realm, which is a functional regression for delegated realm administration.

### Version

26.4.7

### Regression

### Expected behavior

A user with realm-level administrative permissions (e.g., `realm-admin` in realm Test) should be able to:
1. Navigate to the realm-specific Admin Console endpoint: http://localhost:8080/admin/Test/console/#/
2. Authenticate successfully
3. Use a fully rendered, functional Admin Console UI for that realm (as in 26.2.5 and earlier)

### Actual behavior

After logging in to the realm-level Admin Console (/admin/_\<realm\>_/console/#/) as the realm admin user:

- The Admin Console UI is broken / mostly not rendered
- Navigation and UI components are not usable
- The console cannot be used to perform administrative tasks from the realm-level entry point

### How to Reproduce?

1. Install Keycloak 26.3.0+
2. Create a realm named `Test`
3. Create a new user in realm `Test` and grant them the `realm-admin` role
4. Open the realm-level Admin Console URL, for example: http://localhost:8080/admin/Test/console/#/
6. Log in as the realm-level administrator created in step 3

**Observe:** the Admin Console UI does not render correctly and is not usable.

### Anything else?

_No response_

pedroigor:
Closes #45116

### DIFF
--- a/js/apps/admin-ui/src/App.tsx
+++ b/js/apps/admin-ui/src/App.tsx
@@ -50,12 +50,17 @@ export const App = () => {
   const [adminClient, setAdminClient] = useState<KeycloakAdminClient>();
 
   useEffect(() => {
+    const fragment = "#/";
+    if (window.location.href.endsWith(fragment)) {
+      const newPath = window.location.pathname.replace(fragment, "");
+      window.history.replaceState(null, "", newPath);
+    }
     const init = async () => {
       const client = await initAdminClient(keycloak, environment);
       setAdminClient(client);
     };
     init().catch(console.error);
-  }, []);
+  }, [environment, keycloak]);
 
   if (!adminClient) return <KeycloakSpinner />;
   return (


ssilvert:
@pedroigor I have verified that your fix works.  Just need to fix the tests.

ssilvert:
@pedroigor Let's also see if we can get rid of this warning by adding the dependencies:
image

ssilvert (approved):

PR #45309, opened 2026-01-09
DCR endpoint ignores client's requested token_endpoint_auth_method in…

### LINKED ISSUE #44403: DCR endpoint ignores client's requested token_endpoint_auth_method
### Before reporting an issue

### Area

oidc

## Summary

Keycloak's Dynamic Client Registration (DCR) endpoint ignores the client's requested `token_endpoint_auth_method` and always returns `client_secret_basic`, even when clients explicitly request `client_secret_post`.

## Impact

This breaks compatibility with OAuth clients that require `client_secret_post`, including:
- Model Context Protocol (MCP) clients (per RFC 9110 requirement)
- Clients that cannot support HTTP Basic authentication
- Environments with security policies requiring POST-based authentication

### Version

26.4.5

### Regression

### Expected behavior

Per [RFC 7591 Section 2](https://datatracker.ietf.org/doc/html/rfc7591#section-2), the `token_endpoint_auth_method` field indicates authentication method for the token endpoint requested by the client. Since `client_secret_post` is a standard method supported by Keycloak, the server should honor this request.

### Actual behavior

Keycloak always returns `client_secret_basic` regardless of the client's request, effectively ignoring the parameter.

### Setup Keycloak

Use the preconfigured test realm from FastMCP (includes DCR enabled and trusted hosts configured):

```bash
git clone https://github.com/stephaneberle9/fastmcp.git
cd fastmcp
git checkout feat/keycloak-auth-provider
cd examples/auth/keycloak_auth/keycloak
./start-keycloak.sh  # Imports realm with DCR enabled
```

Setup details: [README.md](https://github.com/stephaneberle9/fastmcp/blob/feat/keycloak-auth-provider/examples/auth/keycloak_auth/keycloak/README.md)

### Test with cURL

```bash
# DCR request explicitly requesting client_secret_post
curl -X POST http://localhost:8080/realms/fastmcp/clients-registrations/openid-connect \
  -H "Content-Type: application/json" \
  -d '{
    "client_name": "Test Client",
    "redirect_uris": ["http://localhost:8000/callback"],
    "token_endpoint_auth_method": "client_secret_post"
  }'
```

**Expected:** Response includes `"token_endpoint_auth_method": "client_secret_post"`
**Actual:** Response includes `"token_endpoint_auth_method": "client_secret_basic"`

## Test Resources

Complete reproduction environment available in FastMCP repo (branch `feat/keycloak-auth-provider`):
- [realm-fastmcp.json](https://github.com/stephaneberle9/fastmcp/blob/feat/keycloak-auth-provider/examples/auth/keycloak_auth/keycloak/realm-fastmcp.json) - Test realm configuration
- [README.md](https://github.com/stephaneberle9/fastmcp/blob/feat/keycloak-auth-provider/examples/auth/keycloak_auth/keycloak/README.md) - Setup instructions

### Anything else?

_No response_

mposolda:
… case it is client_secret_post

closes #44403

### Previous behaviour

The OIDC client registration has parameter `token_endpoint_auth_method`, which can be set either to `client_secret_basic` or `client_secret_post` in the client registration request. We historically implement both these client authentication methods by single class `ClientIdAndSecretAuthenticator`. When the client registration used `token_endpoint_auth_method` with either of these values, we set client authenticator to `ClientIdAndSecretAuthenticator` and we allowed client authentication with both methods. The OIDC client registration response sent back always `client_secret_basic` (even if `client_secret_post` was used as `token_endpoint_auth_method`).

### Updates in this PR

- PR introduces new client attribute `client.secret.authentication.allowed.method`. It allowed to specify that only one of the methods is allowed (either "Basic" or "POST"). By default, it is still set to "Any", which means that both client authentication methods are supported for client authentication with `ClientIdAndSecretAuthenticator` .

- When OIDC client registration request (or OIDC client update request) is sent, it uses only the method specified in the `token_endpoint_auth_method` and set it as a value of the `client.secret.authentication.allowed.method`, so the correct method is returned from the OIDC client registration response.

- I've added the documentation to the "Upgrading guide" as strictly speaking, this is little bit backwards incompatible (as previously when client used `token_endpoint_auth_method=client_secret_basic`, it allowed to authenticate client with both BASIC and POST). However clients migrated from previous versions are not affected and they have new attribute set to "Any" .

- I was considering some other approaches, but did not find any way (without introducing new client attribute). Introducing dedicated client authenticators (for `client_secret_basic` and `client_secret_post`) would be even more intrusive with impact on backwards compatibility.


rmartinc (approved):
Thanks @mposolda! LGTM!

mposolda:
@rmartinc Thanks!

PR #45312, opened 2026-01-09
Wrap form buttons in `kcFormButtonsClass` div

### LINKED ISSUE #45164: Base theme: `login-config-totp`, buttons are not wrapped in a `kcFormGroupClass`
### Before reporting an issue

### Area

login/ui

### Describe the bug

Usually buttons are wrapped in a `<div class="${properties.kcFormGroupClass!}">`

Examples:

https://github.com/keycloak/keycloak/blob/e6b2f0dc60d349560569fac7f775d27e2efd83c1/themes/src/main/resources/theme/base/login/login-update-password.ftl#L62-L73

https://github.com/keycloak/keycloak/blob/e6b2f0dc60d349560569fac7f775d27e2efd83c1/themes/src/main/resources/theme/base/login/login-password.ftl#L43-L45

This is not the case for config-totp:

https://github.com/keycloak/keycloak/blob/e6b2f0dc60d349560569fac7f775d27e2efd83c1/themes/src/main/resources/theme/base/login/login-config-totp.ftl#L98-L112

### Version

main branch

### Regression

### Expected behavior

Buttons to be wrapped in a div with `kcFormGroupClass` class

### Actual behavior

buttons are not wrapped, causing `margin` inconsistencies across template

### How to Reproduce?

https://github.com/keycloak/keycloak/blob/e6b2f0dc60d349560569fac7f775d27e2efd83c1/themes/src/main/resources/theme/base/login/login-config-totp.ftl#L98-L112

### Anything else?

As a workaround, I've extracted the template and customized

tagliala:
- Closes #45164

Previously, the buttons in `login-config-totp.ftl` were not enclosed in a `<div>` with the `${properties.kcFormButtonsClass!}` class.

This resulted in inconsistent margins and alignment compared to other base theme templates, especially affecting button groups.

Update the markup to wrap the buttons in a `kc-form-buttons` div with the appropriate class. This aligns the template with the styling standards used elsewhere and ensures margin consistency across the UI.

Related to: #45164

---

### Notes for review

1. I've preserved `id="kc-form-buttons"` for uniformity with other templates
2. This may affect other templates
3. The implementation is mimicking `login-update-password.ftl`: 
    https://github.com/keycloak/keycloak/blob/1aa1621eaa950d957d64f5f7c5aa6392329f83cb/themes/src/main/resources/theme/base/login/login-update-password.ftl#L65-L72
5. Please review with "hide whitespace" turned on, because this is a very simple change but default diff is bigger than it should be
    image

### DIFF
--- a/themes/src/main/resources/theme/base/login/login-config-totp.ftl
+++ b/themes/src/main/resources/theme/base/login/login-config-totp.ftl
@@ -93,23 +93,25 @@
 
             <div class="${properties.kcFormGroupClass!}">
                 <@passwordCommons.logoutOtherSessions/>
-            </div>
 
-            <#if isAppInitiatedAction??>
-                <input type="submit"
-                       class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonLargeClass!}"
-                       id="saveTOTPBtn" value="${msg("doSubmit")}"
-                />
-                <button type="submit"
-                        class="${properties.kcButtonClass!} ${properties.kcButtonDefaultClass!} ${properties.kcButtonLargeClass!} ${properties.kcButtonLargeClass!}"
-                        id="cancelTOTPBtn" name="cancel-aia" value="true" />${msg("doCancel")}
-                </button>
-            <#else>
-                <input type="submit"
-                       class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonBlockClass!} ${properties.kcButtonLargeClass!}"
-                       id="saveTOTPBtn" value="${msg("doSubmit")}"
-                />
-            </#if>
+                <div id="kc-form-buttons" class="${properties.kcFormButtonsClass!}">
+                    <#if isAppInitiatedAction??>
+                        <input type="submit"
+                               class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonLargeClass!}"
+                               id="saveTOTPBtn" value="${msg("doSubmit")}"
+                        />
+                        <button type="submit"
+                                class="${properties.kcButtonClass!} ${properties.kcButtonDefaultClass!} ${properties.kcButtonLargeClass!} ${properties.kcButtonLargeClass!}"
+                                id="cancelTOTPBtn" name="cancel-aia" value="true" />${msg("doCancel")}
+                        </button>
+                    <#else>
+                        <input type="submit"
+                               class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonBlockClass!} ${properties.kcButtonLargeClass!}"
+                               id="saveTOTPBtn" value="${msg("doSubmit")}"
+                        />
+                    </#if>
+                </div>
+            </div>
         </form>
     </#if>
 </@layout.registrationLayout>


tagliala:
Hello, this PR is now ready for review

mabartos (commented):
@tagliala Thanks for the PR, but this is a change for the Login V1, which is deprecated now, and you should use the Login V2. AFAIK, we might really not accept additional changes to the Login V1 (cc: @keycloak/core is that correct?)

But it seems the `login-config-totp.ftl` inside the `theme/keycloak.v2/login` has the same issue.

tagliala:
Thanks.

I did not understand that `base` = `v1`

is there any risk of base theme being removed from the source?

I can see that `v2` is based on base as well:

https://github.com/keycloak/keycloak/blob/0223b6823b1c58d3b54f93ab99f323cb292ac6cc/themes/src/main/resources/theme/keycloak.v2/login/theme.properties#L1

mabartos:
> I did not understand that base = v1

Not really; the base should be both for v1 and v2, but if the file is already in the keycloak.v2, that one is used instead of the base version. It means that in this case of the OTP config, your changes will be applied only for the Login V1 and not for the Login V2.

@tagliala But yeah, we can also have your changes to the base. So, please update the version for Login V2 as well, thanks!

@keycloak/core-maintainers Chime in if you think otherwise.

rmartinc:
Thanks @tagliala and @mabartos! Yes, `keycloak` is v1 and `keycloak.v2` is v2. As @mbartos commented if `keycloack.v2` is the same, send also the changes for v2 and we can fix in both.

tagliala:
Ok thanks, so this is not fixing the deprecated theme `keycloak` but fixing `base`, which both themes inherit from.

I will make the changes to `v2` as well and comment again

tagliala:
@rmartinc @mabartos, sorry I do not understand the changes that should be applied to v2

This change only adds a wrapper around buttons where missing.

`keycloack.v2` already has wrappers.

If the problem are hardcoded class, there is no 1:1 substitution.

At the best of my understanding, it is:
- `pf-v5-c-form__group pf-m-action` => `${properties.kcFormGroupClass!} pf-m-action`
- `pf-v5-c-form__actions` => `${kcFormActionGroupClass!}`, which also adds `pf-v5-u-pt-xs`

But I'm not confident to make the change.

Feel free to push on this PR, it allows edit by maintainers

rmartinc:
@tagliala I think that your change is not OK. The buttons are out of frame:

Screenshot From 2026-01-21 17-06-54

I think you have to add the new div inside the previous div (the one with the `logoutOtherSessions` like it is in [update-password.ftl](https://github.com/keycloak/keycloak/blob/26.5.1/themes/src/main/resources/theme/base/login/login-update-password.ftl#L62-L73).

Don't care about the v2 version. I think that in that we need to use the `buttons.ftl` as it is in the `login-update-password.ftl`.

tagliala:
Thanks.

Rebased and fixed according to the suggestion

### Before

image

### After

image

There is more bottom margin after: 45px vs 30px. 45px should be the correct one, since it matches `Update Password` template and others

> [!IMPORTANT]
> This is a fix for the `base` theme, not the deprecated `keycloak` theme. The screenshot is for presentation only

rmartinc (approved):
Thanks @tagliala! LGTM now!

mposolda (approved):
@tagliala @rmartinc Thanks for the PR and review!

tagliala:
Thank you all

Issue #13102, opened 2022-07-14
Add support for specifying `client.secret` using vault


mhajas:
### Description

Currently, only LDAP/SMTP/IDP passwords support providing value using the vault. This issue should add support also for client secrets.

hecsalazar:
We are also interested in this feature. The motivation is that we deploy keycloak using the operator and we want to pull the secrets from Hashicorp vault which is our corporate source of secrets.

We are using Helm deployments with the Keycloak Operator CRDs

Our current workaround is to deploy KeycloakClient CR with the secrets in plaintext. Our CICD has code to fetch secrets and inject them at deployment time to avoid checking in the actual secrets values in source control.

hecsalazar:
Another approach would be to support secret references in this property to make it more cloud native

mhajas:
@hecsalazar I see there is a lot of interest in this feature. Currently, we don't have any near plans for working on this inside the Keycloak team but a contribution would be welcomed. I would be happy to help answer any questions. 

The addition itself should be straightforward, we just need to find all places where the client's secret is used and replace it with a vault call. Testing could be more challenging though. 

Previous work for other vault-enabled fields:
- https://github.com/keycloak/keycloak/pull/6285
- https://github.com/keycloak/keycloak/pull/6297
- https://github.com/keycloak/keycloak/pull/6293

robson90:
@hecsalazar Keep in mind, that your solution will show all your credentials in your manifest view(for .e.g on Google Cloud Platform)

@mhajas What are the chances that the extension will find its way into the project in a timely manner if it is developed in the same way as the other three paths? (LDAP bindCredentials, SMTP PW, IDP)

Also is there any need in supporting something like [Bitnami Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets)?

isaac-mercieca:
We are also interested in this feature because we deploy Keycloak and need to retrieve secrets from AWS Secrets Manager/Azure Key Vault, our corporate secrets repository.

I've explored the Vault SPI and am eager to contribute to developing this capability. However, since I'm new to Keycloak, could you give me a high-level overview of how to approach this issue? Would implementing this as an extension be a viable solution?

ahus1:
As there has been some interested in this issue from the community and  there being a pull request, it is also a rather big one. I see that @rmartinc has commented on it. In the meantime, I'm adding it to the planning board of core-clients.

PR #45316, opened 2026-01-09
Operator logic for clients in admin api v2

### LINKED ISSUE #45047: Consider having the explicit discriminator field in Client v2 representations
If having the protocol field not be part of the schema is causing it to be parsed into the additionalFields, perhaps we should just have it as a proper field.

_Originally posted by @shawkins in https://github.com/keycloak/keycloak/pull/44727#discussion_r2620226990_

### LINKED ISSUE #46022: Initial Client CR implementation
### Description

Create an initial Operator implementation for Client v2.

shawkins:
Closes #46022

- Creates base controller logic that is really just performing puts or deletes.
   - Delete handling is enforced via a finalizer
   - These operations are in response to CR changes only - the state of the server side client is not actively monitored.
   - It is not clear yet how the notion of ownership / control factors in here. Since we always are just performing a PUT we aren't differentiating the case of whether the client already exists - if it already exists, are we somehow asserting ownership of it?
   
- Differences between the CRD and admin api resources are handled via subclasses that add targeted JsonIgnores and new fields. There may be a way to do this Jackson views, but none of the serialization logic is view aware.
  - The OIDC Auth.secret field is not present in the CRD. Instead there is a "secretRef" field. The naming can of course change - e.g. secretSecret?

- Once set, the CRD enforces that you cannot change the Keycloak instance nor the realm name in the client CRs because these should effectively trigger cleanups, which we're not guaranteed to get the event for.

- Assumes that a "keycloak cr name"-admin Secret will contain a clientId and clientSecret. This is just a placeholder - it is likely that it will get refined, such as the operator invoking bootstrap admin to create the necessary service account.

- Puts only a minimal amount of information in the status. If the returned representation (because of defaulting or normalizations) is of interest, then we'll need to add a sanitized version of that to the status.

- Getting the admin client is hacked in. 
  - I'm temporarily using reflection to get at the client WebTarget, then use that along with our rest interfaces to make the remote calls. 
  - There's no optimization yet for pooling the client
  - https trust has been added, and as discussed below - hostname verification is disabled, rather than introducing hostname overrides and additional logic.
  - mTLS handling is TDB - we can add an new admin.tlsSecret or even http.adminTlsSecret (if we want it close to the existing http tlsSecret).

- No attempt has been made to separate the client crds from the existing ones.

- We are always attempting to use the service hostname and http if available. This may not always be desirable. A user has even requested targeting arbitrary Keycloak instances: https://github.com/keycloak/keycloak/issues/44613

- Do we need a validation that the client-admin-api:v2 feature is enabled?
  - How does this relate to resource ownership - if I create some client CRs, then toggle client-admin-api:v2 from enabled to disabled, does that effectively lock the state of those clients on the server-side and prevent the deletion of the CRs?

Let me know your early thoughts.


shawkins:
Had to tweak things further to get the CRD definition as expected.

@keycloak/cloud-native just wanted to check more on how we want to access the servers. http/edge scenarios are not a problem, so my initial thinking is that we'd go directly to the service endpoints. 

For pass-through that's a problem as we'll likely fail any hostname validation checks.
For reencrypt there is probably a path where service serving certficiates could be used. 

Also this approach doesn't work for testing in local mode as the operator is not running in cluster - I've hacked in a workaround with access via a NodePort. 

I can either:
- relax hostname checks
- instead try to determine the appropriate hostname or hostname-admin. However assuming that the user has specified hostname-admin to restrict ingress, they will need to ensure that access via the operator is allowed.

vmuzikar:
> I can either:
> 
> * relax hostname checks
> * instead try to determine the appropriate hostname or hostname-admin. However assuming that the user has specified hostname-admin to restrict ingress, they will need to ensure that access via the operator is allowed.

I'm not sure about this. Relaxing hostname checks feels kinda insecure... for a secure product like an IAM solution. Using `hostname-admin` could lead to some unexpected routing issues as that's supposed to be a public host. Still, feels like a more proper solution, a lesser evil.

shawkins:
> I'm not sure about this. Relaxing hostname checks feels kinda insecure... for a secure product like an IAM solution. 

I realize that it's not great, but I was wondering if this situation already comes up - if someone configures keycloak for tls passthrough, and does not enable http, does that mean dynamic back channel requests:
- aren't really a supported case because your using tls either way, so you might as well go through the public hostname
- require the cert to expose some internal host
- must disable hostname verification

> Using hostname-admin could lead to some unexpected routing issues as that's supposed to be a public host. Still, feels like a more proper solution, a lesser evil.

I'd like to say this only needs done in the passthrough case - but we can't reliably differentiate between passthrough and reencrypt. So I think the logic would need to be that if tls is configured, but http is not enabled, then we'll have to access via hostname / hostname-admin.

Thinking about it more and looking at customizability of the client, I think we should be able to still access internally, but validate against the expected external hostname - although it would be better to cleanly know when we in a passthrough vs a reencrypt scenario.

vmuzikar:
> if someone configures keycloak for tls passthrough, and does not enable http, does that mean dynamic back channel requests

That's a good point – that the Operator is in the same position as any app doing a backchannel request through the internal network.

Thinking about it more, I believe we'll need to document that passthrough TLS (without HTTP enabled) is generally incompatible with the backchannel requests. Exposing internal hostnames in the cert might be considered a security vulnerability, IMHO. Similarly to relaxing hostname checks.

For now, let's make the Operator use hostname/hostname-admin. If we ever see issues with that, we can add some explicit setting for telling the Operator whether passthrough or reencrypt is used, and based on that we'll decide whether to use internal host.

shawkins:
@vmuzikar what do you think of the approach in the last commit of validating against the hostname admin and the service hostname?

The biggest concern I can think of is whether the reencrypt hostname will actually be the service hostname.

Another corner case - if you setup the instance with hostname-strict=false and don't enable http. I realize this is generally not something we recommend or support, but in passthrough mode, we may not have any idea what hostname to use.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,373 @@
...
+                // TODO: validate this approach
+
+                // if we're configured for passthrough, and adminHostname is present
+                if (adminHostname.isPresent() && verifier.verify(adminHostname.get(), session)) {
+                    return true;
+                }
+                // otherwise we expect the cert to be valid against the svc hostname - which
+                // should happen with service serving certificates
+                return verifier.verify(hostname, session);
+            }
```
vmuzikar: I like this idea of "extending the trustore with the admin hostname". We could also do this for the regular hostname, just in case.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,373 @@
...
+        String address = System.getProperty(KEYCLOAK_TEST_ADDRESS);
+
+        int port = https?HttpSpec.httpsPort(keycloak):HttpSpec.httpPort(keycloak);
+
+        if (address == null) {
+            // uses the service host - TODO: assumes the operator and the keycloak instance are in the same cluster
+            // this may not eventually hold if we are flexible about where the kube client can target
+            address = String.format("%s.%s.svc:%s", KeycloakServiceDependentResource.getServiceName(keycloak),
+                    keycloak.getMetadata().getNamespace(), port);
+        }
```
vmuzikar, line 389: Will we use the hostname from the CR spec?

shawkins: My initial intent was to see if we could reasonably use the service / back channel.

If that proves too problematic, then we'll switch to hostname-admin / hostname (and remove the custom hostname validation). In addition to the concern about users needing to ensure access from the operator to keycloak, you could even have dns issues - prior to this, there was no hard requirement that hostname-admin / hostname were resolvable by the operator.

shawkins:
> Thinking about it more, I believe we'll need to document that passthrough TLS (without HTTP enabled) is generally incompatible with the backchannel requests. 

Captured as https://github.com/keycloak/keycloak/issues/45606

shawkins:
@keycloak/cloud-native I know we talked about this extensively in our meeting, but after more reflection I believe we should simply disable hostname verification.

Since we are only trusting only the server certificate, which we could even enforce that it's not allowed to be a ca certificate, the tlsSecret would have to be compromised for any kind of attack.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseOIDCClientRepresentation.java
```diff
@@ -0,0 +1,151 @@
...
+
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotBlank;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema
+public class BaseOIDCClientRepresentation<T extends BaseOIDCClientRepresentation.Auth> extends BaseClientRepresentation {
```
vmuzikar: Can't we use schema swap instead of making auth pluggable for this specific use case? Using generics does not scale really well – there might be more security sensitive fields. Also the base representation should not include changes just for the Operator.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,350 @@
...
+
+public abstract class KeycloakClientBaseController<R extends CustomResource<? extends KeycloakClientSpec<S>, KeycloakClientStatus>, T extends BaseClientRepresentation, S extends BaseClientRepresentation>
+        implements Reconciler<R>, Cleaner<R> {
+
+    private static final String CLIENT_API_VERSION = "v2";
+    private static final String HTTPS = "https";
+    // TODO allows for local testing - could be promoted in some way to the CR if needed
+    public static final String KEYCLOAK_TEST_ADDRESS = "KEYCLOAK_TEST_ADDRESS";
+
+    static class KeycloakClientStatusAggregator {
```
vmuzikar, line 101: It would be good to have this as a separate class similarly to the `KeycloakStatusAggregator`.

vmuzikar (commented):
@shawkins I am sorry for the late response. I went through the changes and thought about it more as well. IMHO the approach is good as is.

> No attempt has been made to separate the client crds from the existing ones.

Do you mean CRDs from the old Operator? I think it's enough the API group and version differ.

> We are always attempting to use the service hostname and http if available. This may not always be desirable.

I think that's ok. If we keep the control of the address on the Operator which always uses an internal hostname (without the user option to override it), it should be safe to also disable the hostname validation.

> Do we need a validation that the client-admin-api:v2 feature is enabled?

Yes, we should. The controller should be noop when this feature is disabled.

> How does this relate to resource ownership - if I create some client CRs, then toggle client-admin-api:v2 from enabled to disabled, does that effectively lock the state of those clients on the server-side and prevent the deletion of the CRs?

We don't have resource ownership just not. Later I reckon any resource can be managed only by its owner, or a master realm admin. But that's for a later discussion.

---

The follow-ups then would be:
* mTLS support
* Usage of proper Admin Client (#45364)
* Admin Client pooling
* Other auth mechanism (K8s service accounts)

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+        return restEasyClient;
+    }
+
+    private static String getAdminUrl(Keycloak keycloak, KubernetesClient client) {
+        boolean httpEnabled = KeycloakServiceDependentResource.isHttpEnabled(keycloak);
+        // for now preferring to use http if available
+        boolean https = isTlsConfigured(keycloak) && !httpEnabled;
+        String protocol = https?HTTPS:"http";
+        String address = System.getProperty(KEYCLOAK_TEST_ADDRESS);
```
vmuzikar: Theoretically, this might be a security vulnerability where an attacker could target their own KC instance and read CRs with Secrets that they otherwise don't have access to. But since it's just a sys prop and not an env var, we should be safe.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakOIDCClientRepresentation.java
```diff
@@ -0,0 +1,50 @@
...
+
+        @JsonPropertyDescription("Secret containing the client secret")
+        public SecretKeySelector getSecretRef() {
+            return secretRef;
+        }
+
+        public void setSecretRef(SecretKeySelector secretRef) {
+            this.secretRef = secretRef;
+        }
+
```
vmuzikar, line 30: Thinking about this more, I'm afraid we can't do that. The Keycloak admin and Client admin are two different personas. This would allow a Client admin to read any Secret in the namespace, incl. e.g. the DB Secret, or admin bootstrap Secret.

I wonder if we should wait for https://github.com/keycloak/keycloak/issues/13102 to rather fully leverage Vault with security sensitive values. On the other hand, we could potentially end up in a similar situation – Client admin reading any secret from the Vault.

We'll probably need to lock this down. Maybe annotate Secrets with the CR name that they can be referenced in?

CC @stianst

vmuzikar (changes requested):

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakOIDCClientRepresentation.java
vmuzikar: Potential solution would be if we allowed referencing KC instances in other namespaces, so that we'd allow referencing Secrets only in the same namespace.

shawkins:
> @shawkins I am sorry for the late response. I went through the changes and thought about it more as well. IMHO the approach is good as is.
> 
> > No attempt has been made to separate the client crds from the existing ones.
> 
> Do you mean CRDs from the old Operator? I think it's enough the API group and version differ.

No, the existing CRDs from the new operator. Alternatively we can come back to this if we get much closer to a release and don't feel good about the stability or usability of the new CRDs.

> Yes, we should. The controller should be noop when this feature is disabled.

This relates to some of long-standing configuration related issues. Validating build-time options will generally mean that users will need to use unoptimized images, or to replicate the necessary build-time options on the CR. 

I'm generally leary of the complexity that features per api and version are bringing into this.

> > How does this relate to resource ownership - if I create some client CRs, then toggle client-admin-api:v2 from enabled to disabled, does that effectively lock the state of those clients on the server-side and prevent the deletion of the CRs?
> 
> We don't have resource ownership just not. Later I reckon any resource can be managed only by its owner, or a master realm admin. But that's for a later discussion.

Until that is implemented, we'd have to document that the operator can assume ownership of existing clients, and that client modifications not done by the operator will be undone - but only based upon CR changes.

> The follow-ups then would be:
> 
> * mTLS support

This can be done in this PR.

> * Usage of proper Admin Client (#45364)
> * Admin Client pooling
> * Other auth mechanism (K8s service accounts)

* Additional status information

vmuzikar:
> No, the existing CRDs from the new operator. Alternatively we can come back to this if we get much closer to a release and don't feel good about the stability or usability of the new CRDs.

The controller should be disabled by the feature flag. Hence the CRs effectively ignored.

> This relates to some of long-standing configuration related issues. Validating build-time options will generally mean that users will need to use unoptimized images, or to replicate the necessary build-time options on the CR.
> 
> I'm generally leary of the complexity that features per api and version are bringing into this.

It should be ideally determined from the server, i.e. fetch the features configuration from a live instance. But that's out of scope. For now, I'd be ok with some nasty solution like having to include features in the CR.

> Until that is implemented, we'd have to document that the operator can assume ownership of existing clients, and that client modifications not done by the operator will be undone - but only based upon CR changes.

Yes.

> This can be done in this PR.

Sounds good.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakOIDCClientRepresentation.java
shawkins: > The Keycloak admin and Client admin are two different personas. This would allow a Client admin to read any Secret in the namespace, incl. e.g. the DB Secret, or admin bootstrap Secret.

They are different personas, but is that needing the be inforced a the Kubernetes level? Aren't we mostly targetting gitops flow such that there would be controls already on who can modify the client and other admin api style resources?

If you want this to be enforced a the Kubernetes level, then we'll need to probably think of this like a separate operator, which functions more like what is being requested in #44613 - but we need to then have some other notion of how to find the Keycloak cluster other than a reference to the CR.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
shawkins: It doesn't really matter if it's an ENV or sys prop - someone with rights to manipulate the operator Deployment could influence this in some way. However someone with that level of access can do far more than that configuration-wise.

If we don't end up changing how we access the keycloak server, we can certianly revisit whether we even want to allow local mode testing.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseOIDCClientRepresentation.java
shawkins: > Can't we use schema swap instead of making auth pluggable for this specific use case?

Sort of, but that is not without a good bit of messiness. You still need to introduce the extended Auth type, then swap it in.

However that refined type won't actually then be used for deserialization - anything customization of the schema would have to accounted for via the base type additionalFields.

> Also the base representation should not include changes just for the Operator.

You can consider doing this, introducing parallel classes, or other messy hacks.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakOIDCClientRepresentation.java
shawkins: As discussed off-line there are two conserns here. A Client admin persona directly administering clients in the Keycloak namespace being given read permission to secrets in that namespace, and the Client admin persona being able to reference and exfiltrate Secrets they shouldn't have access to.

The proper solution to both is to introduce the ability to use separate namespaces per Client admin. Since we are reluctant to let the operator install in cluster scoped mode, this will lead to an even greater proliforation of operator instances - unless we consider splitting this into a separate operator.

For now that seems to be large shift in approach, so we'll continue how this PR currently is. At worst case this will mean initially documenting that those with the ability to create / edit Client CRs must also be namespace level admins.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseOIDCClientRepresentation.java
shawkins: Added a commit switching to schemaswap - this didn't end up affecting the test logic because I was already not using the fluent for this - I think in some even earlier incarnation the buildable attribute would be needed in the base representation classes. 

The upside:
- no changes to the base representation

The downsides:
- the operator will start with the base representation, not the refined type
- it will be impossible for the fluent/builder to understand this convention

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,350 @@
...
+import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
+import io.javaoperatorsdk.operator.api.reconciler.ErrorStatusUpdateControl;
+import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+import io.quarkus.logging.Log;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
+
+import static org.keycloak.operator.crds.v2alpha1.CRDUtils.isTlsConfigured;
+
+public abstract class KeycloakClientBaseController<R extends CustomResource<? extends KeycloakClientSpec<S>, KeycloakClientStatus>, T extends BaseClientRepresentation, S extends BaseClientRepresentation>
```
vmuzikar, line 95: Might be good to add some comment with describing the generic types here as there are quite a few of them and it'd make it easier to maintain.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakOIDCClient.java
```diff
@@ -0,0 +1,50 @@
...
+import io.fabric8.kubernetes.api.model.Namespaced;
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Version;
+import io.quarkiverse.operatorsdk.annotations.CSVMetadata;
+import io.sundr.builder.annotations.Buildable;
+import io.sundr.builder.annotations.BuildableReference;
+
+@CSVMetadata(
+        description="Represents a Keycloak OIDC Client",
```
vmuzikar, line 32: We might want to be more explicit here.
```suggestion
        description="EXPERIMENTAL: Represents a Keycloak OIDC Client",
```

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakSAMLClient.java
```diff
@@ -0,0 +1,50 @@
...
+import io.fabric8.kubernetes.api.model.Namespaced;
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Version;
+import io.quarkiverse.operatorsdk.annotations.CSVMetadata;
+import io.sundr.builder.annotations.Buildable;
+import io.sundr.builder.annotations.BuildableReference;
+
+@CSVMetadata(
+        description="Represents a Keycloak SAML Client",
```
vmuzikar, line 32:
```suggestion
        description="EXPERIMENTAL: Represents a Keycloak SAML Client",
```

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
```diff
@@ -0,0 +1,158 @@
...
+        initCustomBootstrapAdminServiceAccount(kc);
+        var deploymentName = kc.getMetadata().getName();
+        deployKeycloak(k8sclient, kc, true);
+
+        Map<String, String> labels = Utils.allInstanceLabels(kc);
+        labels.put("app.kubernetes.io/component", "server");
+
+        var nodeport = new ServiceBuilder().withNewMetadata().withName(NODEPORT_SERVICE).endMetadata().withNewSpec()
+                .withType("NodePort").addToSelector(labels).addNewPort().withPort(https?Constants.KEYCLOAK_HTTPS_PORT:Constants.KEYCLOAK_HTTP_PORT)
+                .endPort().endSpec().build();
```
vmuzikar, line 131: As discussed offline, we might want to try port forwarding.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,350 @@
...
+            return existingStatus;
+        }
+
+    }
+
+    @Inject
+    Config config;
+
+    @Override
+    public UpdateControl<R> reconcile(R resource, Context<R> context) throws Exception {
```
vmuzikar, line 145: Just not to forget what was discussed before. We should reconcile only if the `client-admin-api:v2` feature is enabled in the CR explicitly.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseOIDCClientRepresentation.java
vmuzikar: Thank you for the update. I think it's not super ideal but good enough. It'd be ok to go ahead with this approach for now.

Though it might be worth it to also try generating the class with the modification using something like Byte Buddy.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/client/KeycloakOIDCClient.java
shawkins: Captured https://github.com/keycloak/keycloak/issues/45794 and https://github.com/keycloak/keycloak/issues/45795 to deal with this in a more comprehensive way.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
shawkins: Just to remove any initial concern, I've switched the logic to use an instance variable.

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
shawkins: That can help with not having to specify a different port than what in on the Keycloak CR, but it doesn't help substancially vs using a nodeport for accessing when outside the cluster.

shawkins:
> Though it might be worth it to also try generating the class with the modification using something like Byte Buddy.

Okay if / when parsing into the non-refined type causes us an issue, we can look into that more.

shawkins:
I'm capturing the follow-ups to this PR under the MVP issue. The admin client pooling isn't necessary for the MVP, so it will be created as a stand-alone issue for later.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
shawkins: I'll capture this as a follow-up as it will need documentation as well.

Related to this is the consideration of checking the server state vs the CR state. You can set the feature in the CR and that not actually represent the state of the server for a while (in particular if some other invalid change is paired with it). 

I'm fine with chalking this up to an eventual consistency issue and just letting the operator fail at the admin operation until the cluster is up-to-date. There is also no check currently of whether the server is in a ready state in general - we'll simply try the admin operation, record the failure in the status, and try again in 10 seconds.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
shawkins: Added

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
vmuzikar: The comment above is still valid but it's not blocking.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
vmuzikar: > I'll capture this as a follow-up as it will need documentation as well.

jFTR – this is tracked under #45810.

> Related to this is the consideration of checking the server state vs the CR state. You can set the feature in the CR and that not actually represent the state of the server for a while (in particular if some other invalid change is paired with it).
> 
> I'm fine with chalking this up to an eventual consistency issue and just letting the operator fail at the admin operation until the cluster is up-to-date. There is also no check currently of whether the server is in a ready state in general - we'll simply try the admin operation, record the failure in the status, and try again in 10 seconds.

That is fine. More concerning is that we ignore the actual config that might be baked in a custom image, but that's a broader issue and not specific to Client v2 and features.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -0,0 +1,399 @@
...
+                .inNamespace(keycloak.getMetadata().getNamespace())
+                .withName(keycloak.getMetadata().getName() + "-admin").require();
+
+        String adminUrl = getAdminUrl(keycloak, client, addressOverride);
+
+        Client restEasyClient = null;
+
+        // create a custom client if using https
+        if (adminUrl.startsWith(HTTPS)) {
+            // TODO: support for mtls
```
vmuzikar: Just a nitpick. We already support mTLS.
```suggestion
        // create a custom client if using https/mtls
        if (adminUrl.startsWith(HTTPS)) {
```

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
```diff
@@ -170,6 +171,10 @@ public class KeycloakSpec {
     @JsonPropertyDescription("Set this to to false to disable automounting the default ServiceAccount Token and Service CA. This is enabled by default.")
     private Boolean automountServiceAccountToken;
 
+    @JsonProperty("admin")
+    @JsonPropertyDescription("In this section you can find all properties related to making admin connections from the operator to the server. These settings are not used by the server.")
+    private AdminSpec adminSpec;
```
vmuzikar, line 176: Thinking about this more, I wonder if we're not jumping the gun here with introducing the adminSpec. We might later introduce a dedicated mTLS field that would also configure the server. But I reckon ithe adminSpec will have other uses too.

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
vmuzikar, line 1: Would be also good to have a SAML CR test but we can do that as a follow-up.

vmuzikar (dismissed):
@shawkins Sorry for the late review.

Added a few comments but nothing blocking. Approving, leaving up to you if you want to merge it and address those in follow-ups.

Also, created and linked #46022.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
shawkins: > Thinking about this more, I wonder if we're not jumping the gun here with introducing the adminSpec.

We can use an annotation instead of spec if we're having qualms about this.

> We might later introduce a dedicated mTLS field that would also configure the server.

I'm not sure what you mean - why whould you need to include a client private cert with the server configuration in general?

> We might later introduce a dedicated mTLS field that would also configure the server.

Honestly I'm not sure if it will have other uses - or if we'd want to go the other direction given the secret discussion and introduce a whole other CR that configures Keycloak instance we're administering.

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
shawkins: I'm fine with a follow-up.

vmuzikar: Ok, created #46042.

operator/src/main/java/org/keycloak/operator/crds/v2alpha1/deployment/KeycloakSpec.java
vmuzikar: > We can use an annotation instead of spec if we're having qualms about this.

You mean annotation directly on the CR? That would feel a bit weird and inconsistent.

> I'm not sure what you mean - why whould you need to include a client private cert with the server configuration in general?

Sorry, right, server needs only truststore, that's different config from what admin client needs.

> Honestly I'm not sure if it will have other uses

Auth method might be one of the additional uses.

> or if we'd want to go the other direction given the secret discussion and introduce a whole other CR that configures Keycloak instance we're administering.

Thinking about it further, I don't think we should go this direction. Except for auth credentials (for different tenants), I don't think there would be differences in admin client settings for a single KC instance.

vmuzikar (approved):

mabartos:
Just information from my end - I built it locally, and it looks very good. I think the missing parts were already mentioned here - like there should be improvements in the status area for sure. Then the Client CR name is used as the clientId of the client (which is nice), but would it be possible to warn users that changing the name of the CR does not do anything?

But overall, very good job!

PR #45317, opened 2026-01-09
Add Metrics for http-client requests

### LINKED ISSUE #44547: Add Metrics for http-client requests
### Description

It would be great to enhance the class `HttpClientProvider` with metrics, so that we can see metrics about http requests issued by keycloak and the corresponding pool.

**Http Requests**
Labels (maybe not all can be always populated - depends on the scenario):
* name (e.g. Backchannellogout)
* authority (schema://host:port)
* path
* method
* statuscode

Metrics:
* http_client_requests_seconds_count
* http_client_requests_seconds_max
* http_client_requests_seconds_sum
* http_client_requests_seconds_bucket (p50, p75, p90, p99) -> maybe configurable

**Connection Pool**
Probably further metrics describing the connection-pool would make sense.

Labels
* name (e.g. Backchannellogout)
* authority (schema://host:port)

Metrics:
* http_client_pool_available_count
* http_client_pool_leased_count
* http_client_pool_pending_count
* http_client_pool_max_count

Maybe we can use an existing component as foundation and enhance it:
https://docs.micrometer.io/micrometer/reference/reference/httpcomponents.html

### Value Proposition

Better observability

### Goals

Add metrics for outbound http requests

### Non-Goals

-

### Discussion

_No response_

### Notes

_No response_

ryanemerson:
Closes #44547


services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -207,7 +209,11 @@ private void lazyInit(KeycloakSession session) {
 
                     HttpClientBuilder builder = newHttpClientBuilder();
 
+                    boolean enableMetrics = ConfigProvider.getConfig().getOptionalValue("quarkus.micrometer.enabled", Boolean.class).orElse(false) &&
```
ryanemerson: Is there a better/more correct way to test that whether the micrometers metrics are enabled here?

mabartos:
```suggestion
                    boolean enableMetrics = config.root().get(MetricsOptions.METRICS_ENABLED.getKey(), "false") &&
```

@ryanemerson I think this should be sufficient.

ryanemerson: Much nicer, thanks!

services/src/main/java/org/keycloak/connections/httpclient/HttpClientBuilder.java
```diff
@@ -266,18 +279,40 @@ public CloseableHttpClient build() {
...
+                    .setConnectionManager(connectionManager)
+                    .setDefaultRequestConfig(requestConfig);
+
+            if (metrics) {
+                new PoolingHttpClientConnectionManagerMetricsBinder(connectionManager, "default")
+                      .bindTo(Metrics.globalRegistry);
+
+                builder.setRequestExecutor(
+                      MicrometerHttpRequestExecutor.builder(Metrics.globalRegistry)
+                            .build()
```
ahus1, line 334: This will log a warning during startup, and we don't want to log this warning only log warnings that are actionable to the Keycloak administrator. 

> 2026-01-12 21:27:08,634 WARN  [io.micrometer.core.instrument.binder.httpcomponents.MicrometerHttpRequestExecutor] (executor-thread-1) This class has been deprecated. Please use ObservationExecChainHandler for Apache HTTP client 5 support instead.

You can consider setting a specific log level like we do it here, or we inline this class and also change its tags and metric name - see comment about the filter. 

https://github.com/keycloak/keycloak/blob/ed69f332af370cb9b4762f5611bd4c22364bff63/quarkus/runtime/src/main/resources/application.properties#L30-L34

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/metrics/HttpClientMeterFilter.java
```diff
@@ -0,0 +1,18 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.MeterFilter;
+
+@Singleton
+public class HttpClientMeterFilter implements MeterFilter {
+    @Override
+    public Meter.Id map(Meter.Id id) {
+        if (id.getName().startsWith("httpcomponents.httpclient.")) {
+            // Replace the httpcomponents.httpclient prefix with one that is implementation independent
+            return id.withName(id.getName().replace("httpcomponents.httpclient.", "http.client."));
```
ahus1: There are semantic conventions on how http client request metrics should be recorded, see https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientrequestduration

This is about metric names and their labels, and if we show any metrics in this space, we should stick to the specs, or IMHO not map at all.

I would argue the following: 

* Let's keep all httpcomponents.httpclient metrics as is, except for maybe the `httpcomponents.httpclient.request` (see following item)
* EITHER Replace `MicrometerHttpRequestExecutor` with our own implementation that follows the semantic conventions, choosing the right metric name and the expected tags OR keep the metric name so people don't expect it to be semantically correct
* Independent of what way we would go, we then wouldn't need to the mapper any more. 

I'm happy to discuss this.

docs/guides/observability/metrics-for-troubleshooting-http.adoc
```diff
@@ -77,6 +77,30 @@ m| http_server_bytes_read_sum
...
 
+=== Client
+
+The metrics below helps to monitor the performance of the client used by the {project_name} server to make HTTP requests.
+
+|===
+|Metric |Description
+
+m| http_client_request_seconds
+| Duration of HttpClient request execution
```
ahus1: The current metrics have in my tests always the `uri="UNKNOWN"`. With this setup I can't distinguish different target hosts, which makes this metric difficult to use to monitor the latency of a specific endpoint. 

At the same time the tag should be low cardinality, so I'd suggest to use only the hostname and port there, maybe the path, but never any query parameters.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -29,6 +29,12 @@ it is required to authenticate the client really with the client secret sent as
 
 It is still possible to make the OIDC client to allow both methods and clients migrated from previous versions are set by default to allow authentication with both methods.
 
+=== HTTP client metrics
+
+The HTTP client used by {project_name} now exposes metrics which allow request latency and connection pool status to be monitored.
```
ahus1, line 40: To be able to monitor latencies, it would be good if I could enable histograms as well like for the HTTP server metrics.

ahus1 (changes requested):
Thank you for the pull request. See below for some comments. Feel free to schedule a 1:1.

docs/guides/observability/metrics-for-troubleshooting-http.adoc
ryanemerson: We can map the URI using:

```java
                      MicrometerHttpRequestExecutor.builder(Metrics.globalRegistry)
                            .uriMapper(request -> {
                                // TODO
                                return "";
                            })
                            .build()
```

It's simple enough to record only the hostname and port, however including the path is trickier as this needs to be defined on a case-by-case basis. A workaround is to make use of a custom header and then use this in the uri mapper. For example:

```
// Request code
HttpGet request = new HttpGet("http://service/api/v1/users/123");
request.addHeader("X-Metrics-Template", "/api/v1/users/{id}");

// Mapper
.uriMapper(request -> {
    Header header = request.getFirstHeader("X-Metrics-Template");
    return (header != null) ? header.getValue() : "UNKNOWN"; // This could also default to just the host+port
})
```

The downside to this is that we will need to define the `X-Metrics-Template` throughout our code.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/metrics/HttpClientMeterFilter.java
ryanemerson: `MicrometerHttpRequestExecutor` and `PoolingHttpClientConnectionManagerMetricsBinder` are actually deprecated due to them being associated with the 4.5 client. Replacing these does not seem overly complex, so maybe it is preferable to provide our implementations that conforms to the OTel conventions.

ryanemerson: > This is about metric names and their labels, and if we show any metrics in this space, we should stick to the specs, or IMHO not map at all.

Just one thing to note. Currently we don't follow the OTel guidelines for HTTP server metrics either

docs/guides/observability/metrics-for-troubleshooting-http.adoc
ahus1: Adding a template is possibly unrealistic, and IMHO shouldn't be done as part of this PR. Let's go with hostname and port, and still add some prevention for a maximum number of tags. 

The Internet claims that there is some `MeterFilter#maximumAllowableTags` but I wouldn't know how to apply it.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/metrics/HttpClientMeterFilter.java
ahus1: Maybe we would keep the original metric names, assuming someone might already  have a httpcommons dashboard? Hm....

ahus1: > MicrometerHttpRequestExecutor and PoolingHttpClientConnectionManagerMetricsBinder are actually deprecated due to them being associated with the 4.5 client. Replacing these does not seem overly complex, so maybe it is preferable to provide our implementations that conforms to the OTel conventions.

Replacing it might have a some ripple effects, and I fear that. Even when changing that, I would be surprised we we would suddenly be OTel compliant.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
ryanemerson: Should we reuse the existing CLI options for the HTTP server metrics here, e.g. `http-metrics-histograms-enabled`, or do we want dedicated SPI/CLI options for the client?

ahus1: I'd argue to have separate ones. I suggest the names `http-client-metrics-histograms-enabled` and `http-client-metrics-slos` for that. WDYT?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/services/metrics/HttpClientMeterFilter.java
ryanemerson: Ack. I have reverted to the original metric names.

ahus1 (dismissed):
Thank you for the PR. I've added a limiter for the distinct number of tags. For now it is not configurable to have more than 100 distinct tags. 

Let me know what you think, not merging yet.

services/src/main/java/org/keycloak/connections/httpclient/HttpClientBuilder.java
```diff
@@ -310,19 +310,14 @@ public CloseableHttpClient build() {
                 new PoolingHttpClientConnectionManagerMetricsBinder(connectionManager, "default")
                       .bindTo(Metrics.globalRegistry);
 
+                String meterName = "httpcomponents.httpclient.request";
+                limitNumberOfTagsForMeter(meterName, "target.host");
+                limitNumberOfTagsForMeter(meterName, "target.port");
+                limitNumberOfTagsForMeter(meterName, "uri");
```
ryanemerson, line 325: Do we really want to limit the number of host and port tags? I don't think it would be common to need more than 100, but I would imagine those metrics should always be desired instead of some hosts being arbitrarily ignored.

For the uri tag, unless a `uriMapper` is defined, the `uri` is always `UNKNOWN`, which is why the MetricsDistTest is failing.

If we want the 100 limit to apply to all `uri` fields then we would need to do:

`.uriMapper(request -> request.getRequestLine().getUri());`

However, if I understand the `MeterFilter` correctly,  I think that would limit the total number of unique `uri` tags across all hosts to 100, which doesn't make sense if you consider that the client might be reaching out to multiple different hosts each of which have endpoints like `/user/{id}` where `id` changes each call.

My suggestion would be that we remove the tag limit and define the following mapper:

```
.uriMapper(request -> {
    Header header = request.getFirstHeader("X-Metrics-Template");
    return (header != null) ? header.getValue() : "UNKNOWN"; // This could also default to just the host+port
})
```

Users of the client can then add the `X-Metrics-Template`  if desired. We could also have a follow-up task to add `X-Metrics-Template` throughout the code.

services/src/main/java/org/keycloak/connections/httpclient/HttpClientBuilder.java
```diff
@@ -310,19 +310,14 @@ public CloseableHttpClient build() {
...
-                                URI uri = URI.create(request.getRequestLine().getUri());
-                                String scheme = Objects.requireNonNullElse(uri.getScheme(), "");
-                                String host = Objects.requireNonNullElse(uri.getHost(), "");
-                                StringBuilder sb = new StringBuilder(scheme).append(host);
-                                if (uri.getPort() != -1) {
-                                    sb.append(":").append(uri.getPort());
-                                }
-                                return sb.toString();
-                            })
+                            .exportTagsForRoute(true)
```
ryanemerson, line 329: :+1: thanks for the change

We may still need a `UriMapper` though, see https://github.com/keycloak/keycloak/pull/45317/commits/e9ba1b91ea62d238736177e2017397b0d8e75594#r2695215811

docs/guides/observability/metrics-for-troubleshooting-http.adoc
```diff
@@ -77,6 +77,45 @@ m| http_server_bytes_read_sum
...
+m| http_client_request_seconds_count
+| The total number HttpClient requests for a given URI
+
+m| http_client_request_seconds_max
+| Duration of the longest HttpClient request for a given URI
+
+m| http_client_request_seconds_sum
+| Duration of all HttpClient requests for a given URI
+
+m| http_client_pool_total_connections
```
pruivo: Are these metric names correct? The test is asserting different names
```java
.body(containsString("TYPE httpcomponents_httpclient_pool_total_connections"))
```

ryanemerson: Good spot, that was a leftover from the initial implementation. Updated.

quarkus/tests/integration/src/test-providers/java/org/keycloak/it/resource/realm/TestRealmResource.java
```diff
@@ -33,6 +26,18 @@
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
+import org.keycloak.connections.httpclient.HttpClientProvider;
```
pruivo: unused import (maybe Spotless does not check the test class?)

services/src/main/java/org/keycloak/connections/httpclient/HttpClientBuilder.java
ahus1: The URI Mapper is not doing what you think it is doing. When I tested it, the request line didn't include a host.

pruivo (dismissed):
You have an unused import in the test class, but it LGTM!

services/src/main/java/org/keycloak/connections/httpclient/HttpClientBuilder.java
ahus1: This was supposed to be a small change, but the effort we're putting into this is getting out of hand. 

There is AFAIK a default URI mapper in place that applies the template. We're not using the template, maybe a 3rd party library does. I don't have any intentions to appliy the X-Metrics-Template in our code. 

Limiting is a recommended practice, otherwise you can run out of memory. We don't know how many different URLs are called - it can become a lot if there are a lot of realms. 

If we think it is not useful as it is, we could also divert to not implementing it at all: If you can't distinguish the targets, you can't make sense of the metrics. At the moment you can't even distinguish the sources (which realm, which for example client configuration). 

So would we skip this one as we think it will not be very useful? 

Happy to jump on a call tomorrow.

ryanemerson: Following up from our call earlier, I have implemented a default `uriMapper` which uses the value of the `X-Metrics-Template` header if present for the `uri` tag, otherwise the value will be "UNKNOWN".

I have updated the various identity providers to set the `X-Metrics-Template` where possible. In the case of user configurable URLs, I have used the url path as the template value when the URL is not expected to contain specific resource identifiers in the path.

I have also made it so that the maximum number of tags can be defined via the CLI option `http-client-metrics-tag-limit`, which currently defaults to 100.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
```diff
@@ -158,6 +158,20 @@ public List<PropertyMapper<?>> getPropertyMappers() {
                         .to("quarkus.rest.jackson.optimization.enable-reflection-free-serializers")
                         .build(),
                 fromOption(HttpOptions.HTTP_ACCEPT_NON_NORMALIZED_PATHS)
+                        .build(),
+                fromOption(HttpOptions.HTTP_CLIENT_METRICS_ENABLED)
+                        .isEnabled(MetricsPropertyMappers::metricsEnabled, MetricsPropertyMappers.METRICS_ENABLED_MSG)
+                        .build(),
+                fromOption(HttpOptions.HTTP_CLIENT_METRICS_HISTOGRAMS_ENABLED)
+                        .isEnabled(MetricsPropertyMappers::metricsEnabled, MetricsPropertyMappers.METRICS_ENABLED_MSG)
```
mabartos, line 166:
```suggestion
                        .isEnabled(MetricsPropertyMappers::httpClientMetricsEnabled, MetricsPropertyMappers.HTTP_CLIENT_METRICS_ENABLED_MSG)
```

@ryanemerson I'd suggest to check here whether the `http-client-metrics-enabled` is true for the HTTP client metrics related options. As these options should not be enabled when the `http-client-metrics-enabled=false`.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
```diff
@@ -158,6 +158,20 @@ public List<PropertyMapper<?>> getPropertyMappers() {
...
+                        .build(),
+                fromOption(HttpOptions.HTTP_CLIENT_METRICS_SLOS)
+                        .isEnabled(MetricsPropertyMappers::metricsEnabled, MetricsPropertyMappers.METRICS_ENABLED_MSG)
+                        .paramLabel("list of buckets")
+                        .build(),
+               fromOption(HttpOptions.HTTP_CLIENT_METRICS_TAG_LIMIT)
+                        .isEnabled(MetricsPropertyMappers::metricsEnabled, MetricsPropertyMappers.METRICS_ENABLED_MSG)
+                        .to("kc.spi-connections-http-client--default--metrics-tag-limit")
```
mabartos, line 174:
```suggestion
                        .paramLabel("limit")
                        .to("kc.spi-connections-http-client--default--metrics-tag-limit")
```

nitpick: It should be better than having the <PARAM> in the help message.

federation/ipatuura/src/main/java/org/keycloak/ipatuura_user_spi/Ipatuura.java
```diff
@@ -81,10 +83,11 @@ public Integer csrfAuthLogin() {
         String password = model.getConfig().getFirst("loginpassword");
 
         /* Execute GET to get initial csrftoken */
-        String url = String.format("https://%s%s", server, "/admin/login/");
+        String loginPath = "/admin/login";
+        String url = String.format("https://%s%s", server, loginPath);
 
         try {
-            response = SimpleHttp.create(session).doGet(url).asResponse();
+            response = SimpleHttp.create(session).doGet(url).header(METRICS_URI_TEMPLATE_HEADER, loginPath).asResponse();
```
mabartos, line 90: @ryanemerson Wouldn't it be better to add the header by default when the `http-client-metrics-enabled=true` for the SimpleHttp(which is a Keycloak thing)? As we should know the path of the request, we can automatically set the path for the header, right?

mabartos (commented):
@ryanemerson I just added a few tiny comments.

federation/ipatuura/src/main/java/org/keycloak/ipatuura_user_spi/Ipatuura.java
ahus1: The purpose is here that the URL might templated, and in that case we need the source of the template, otherwise this doesn't work. 

FYI: The the effort to make this useful is currently IMHO out of hand. Let's try to make it simple and useful.

services/src/main/java/org/keycloak/connections/httpclient/DefaultHttpClientFactory.java
```diff
@@ -66,6 +68,8 @@ public class DefaultHttpClientFactory implements HttpClientFactory {
     private static final String HTTP_PROXY = "http_proxy";
     private static final String NO_PROXY = "no_proxy";
     public static final String MAX_CONSUMED_RESPONSE_SIZE = "max-consumed-response-size";
+    public static final String METRICS_ENABLED = "metrics-enabled";
+    public static final String METRICS_URI_TEMPLATE_HEADER = "X-Metrics-Template";
```
ahus1, line 72: IMHO asking callers to set this as a header is exposing the API in way that is hard to understand. I am also confused to understand why micrometer defines `DefaultUriMapper#URI_PATTERN_HEADER` which is "URI_PATTERN", and this one is defining something different. 

When it comes to SimpleHttpRequest, IMHO there should be a separate method like `uriTemplate()` to provide this as a parameter. Unfortunately this might not work in all the cases, as some places use the the interface directly IMHO, but I need to check if this is mostly in tests only.

I also see that this header is not only handled locally within Keycloak, but also transmitted over the wire. Ideally it would never be transmitted over the wire but filtered out before sending it. And with HTTP client metrics not enabled, it should definitely not be transmitted. 

In its current state I would rather revert the URI handling part as I'm not 100% sure how to fix it in a reasonable time in a meaningful way. If people see any errors or slowness, I suggest they should get more information from tracing instead.

ahus1 (changes requested):
Thank you for the pull request. I have some concerns about the URI handling, see below.

shawkins:
Based upon the age and conflicts, I'm going to close this for now to clear it off of the open PR list for cloud-native.

PR #45334, opened 2026-01-12
Make target configurable in workflow notify user step

### LINKED ISSUE #44676: Workflows: allow specification of the target (either email or username) of the notify-user step
### Description

The notify user step implementation always sends the e-mail to the user associated with the workflow. However, in some instances, it might be desired to send a notification email to a different user (e.g. an admin user) or to a mailing list (e.g. `admins@mycompany.com`). We need to either make the target configurable in the standard notify user step, or add a new step with these capabilities.

ksushant881:
Closes #44676

### DIFF
--- a/docs/documentation/server_admin/topics/workflows/defining-steps.adoc
+++ b/docs/documentation/server_admin/topics/workflows/defining-steps.adoc
@@ -68,6 +68,7 @@ on the realm resource associated with the event, so that each realm resource typ
 | `notify-user` | Notify the user by email a|
 - `subject`: The email subject
 - `message`: The email message in plain text or HTML format
+- `to`: The recipient email address. If not provided, the user's email address will be used
 | `set-user-attribute` | Set an attribute to the user. Allows providing multiple `<name>`/`<value>` pairs a|
 - `<name>`: The attribute name
 - `<value>`: The value of the attribute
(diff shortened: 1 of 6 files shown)


ahus1:
@pedroigor / @sguilhen - maybe you can have a look at this one. Thanks!

sguilhen (approved):
LGTM - @martin-kanis @pedroigor would you mind reviewing as well?

pedroigor (commented):
I would change the name of the property from `send_to` to just `to`. Also, we are using slashes instead as a separator.

Otherwise, LGTM.

ksushant881:
> I would change the name of the property from `send_to` to just `to`. Also, we are using slashes instead as a separator.
> 
> Otherwise, LGTM.

Will change it to `to`.
Do you mean `send_to` should have been `send/to` when you say `we are using slashes instead as a separator`?

pedroigor:
> Do you mean send_to should have been send/to when you say we are using slashes instead as a separator?

@ksushant881 Sorry, I meant "dashes". But yeah, the property can be just `to`.

pedroigor (commented):
Thanks, @ksushant881! @sguilhen LGTM.

pedroigor:
Added a commit with an update to the documentation.

pedroigor (approved):

PR #45356, opened 2026-01-12
Feature Search realms by "Display name" in the "Manage realms" view

### LINKED ISSUE #45292: Search realms by "Display name" in the "Manage realms" view
### Description

It's not easily to find realms if there are many realms and realm name is not human readable as it's not possible to search by realm display name

### Value Proposition 

Searching through display name as well as name will make it easier to find realms if names are not human readable

### Goals

* Should be possible to search for realms in the manage realms view by display name

tre2man:
Closes #45292

https://github.com/user-attachments/assets/92efae66-8097-404b-9c84-d5137ca90c2c

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_3_0.adoc
@@ -19,6 +19,11 @@ user profile configuration where too much information was returned in the past.
 
 Notable changes where an internal behavior changed to prevent common misconfigurations, fix bugs or simplify running {project_name}.
 
+=== OpenTelemetry span attributes location changed
+
+In previous releases, certain OpenTelemetry span attributes such as `kc.realmName` and `kc.clientId` appeared on parent HTTP request spans.
+Span attributes are now set only on child spans where the information is actually known. Monitoring tools that relied on filtering parent HTTP spans by these attributes will need to query the appropriate child spans instead.
+
 === Different credentials of a user need to have different names
 
 When adding an OTP, WebAuthn or any other 2FA credentials, the name the user assigns to this credential needs to be unique for the given user.


ahus1:
@tre2man - thank you for the PR. Please add a sign-off to your commit as outlined in `CONTRIBUTING.md`. 

@stianst - as you opened the original issue, maybe you can provide feedback to this PR, or involve the core-iam team.

tre2man:
> @tre2man - thank you for the PR. Please add a sign-off to your commit as outlined in `CONTRIBUTING.md`.
> 
> @stianst - as you opened the original issue, maybe you can provide feedback to this PR, or involve the core-iam team.

Thanks for reply. I added sign-off to commit.

pedroigor (commented):
Thanks!

Ideally, the display name should be a top-level realm property. However, I'm afraid we do not want to go through the migration path to change this now.

Would it be so bad in terms of migration if we introduce a column to the realm table instead? Not sure how that left join will play out for those running multiple realms and it would be much cleaner to just run queries without any type of join ....

tre2man:
> Thanks!
> 
> Ideally, the display name should be a top-level realm property. However, I'm afraid we do not want to go through the migration path to change this now.
> 
> Would it be so bad in terms of migration if we introduce a column to the realm table instead? Not sure how that left join will play out for those running multiple realms and it would be much cleaner to just run queries without any type of join ....

If realm and displayName are in a 1:1 logical relationship, it looks good to have a display name column in the realm table in the long run. It seems necessary to check the impact of moving the displayName to the realm table as a whole and reflect it in the test code.

However, I am not sure how to write migration code when I upgrade the version from keycloak. Is there an open source contribution guide related to migration?

pedroigor:
@tre2man There is no such information in our contribution guide. But you can follow an approach similar to one of the last issues we fixed that required not only changes to the schema but also running a DML to migrate data. Can you please check https://github.com/keycloak/keycloak/pull/45379?

Basically, we need a Liquibase change set to:

* Add the `DISPLAY_NAME` column to the realm table
* Execute a `customChange` (https://github.com/keycloak/keycloak/pull/45379/changes#diff-2363bc705ea2e82bca752fdd70cf6c063aa66e7a10b117830170721e812d4988R151) to execute an `UPDATE` to the realm table to set the `DISPLAY_NAME` value based on the value from the real attribute table. Something like `UPDATE REALM R SET DISPLAY_NAME = (SELECT VALUE FROM REALM_ATTRIBUTE RA WHERE RA.REALM_ID = R.ID AND RA.NAME = 'displayName')`
* Execute a `DELETE` to delete the display name entry from the realm attribute table

@vramik Anything else to add or change?

tre2man:
> @tre2man There is no such information in our contribution guide. But you can follow an approach similar to one of the last issues we fixed that required not only changes to the schema but also running a DML to migrate data. Can you please check #45379?
> 
> Basically, we need a Liquibase change set to:
> 
> * Add the `DISPLAY_NAME` column to the realm table
> * Execute a `customChange` (https://github.com/keycloak/keycloak/pull/45379/changes#diff-2363bc705ea2e82bca752fdd70cf6c063aa66e7a10b117830170721e812d4988R151) to execute an `UPDATE` to the realm table to set the `DISPLAY_NAME` value based on the value from the real attribute table. Something like `UPDATE REALM R SET DISPLAY_NAME = (SELECT VALUE FROM REALM_ATTRIBUTE RA WHERE RA.REALM_ID = R.ID AND RA.NAME = 'displayName')`
> * Execute a `DELETE` to delete the display name entry from the realm attribute table
> 
> @vramik Anything else to add or change?

Thank you for your pr suggestion. If there are no additional changes, I think i can work on it. 

Can I upload the commit (new work) as force push?

ahus1:
> Can I upload the commit (new work) as force push?

Yes, this is the preferred way.

pedroigor:
Thanks, @tre2man. Please let me know if you need anything else.

tre2man:
> Thanks, @tre2man. Please let me know if you need anything else.

Thanks to the detailed description, I was able to complete the task.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/JpaUpdate26_6_0_MigrateRealmDisplayName.java
```diff
@@ -0,0 +1,71 @@
...
+ * Custom SQL change to migrate the displayName from the REALM_ATTRIBUTE table to the REALM table.
+ * See: <a href="https://github.com/keycloak/keycloak/issues/45356">keycloak#45356</a>
+ * @author tre2man
+ */
+public class JpaUpdate26_6_0_MigrateRealmDisplayName extends CustomKeycloakTask {
+
+    @Override
+    protected void generateStatementsImpl() throws CustomChangeException {
+        // Move displayName from REALM_ATTRIBUTE to REALM table
+        try (PreparedStatement ps = connection.prepareStatement(
```
pedroigor: We should probably run a single `UPDATE` statement to migrate values to the new column. I'll provide an update to this PR if @vramik agrees we should be doing so.

pedroigor (commented):
Thanks, @tre2man!

LGTM, only a few comments. I think @vramik also has something to comment about indexes.

I can update your branch with any missing changes.

model/jpa/src/main/java/org/keycloak/connections/jpa/updater/liquibase/custom/JpaUpdate26_6_0_MigrateRealmDisplayName.java
tre2man: Just a question, is it because of rollback that we treat as a single `update` statement?

pedroigor: More because we avoid a `select` using a single `update` to update all realms using a single statement.

tre2man: Thank you for your advice. 

I changed migration to single statements and processed branching to reflect for each rdb.

pedroigor:
@tre2man Thanks. Can you elaborate why not just use the standard SQL for all databases? Did you encounter any issues when running the migration for any of those databases when using the standard SQL?

Other than that, thanks a lot and LGTM.

I'm just waiting for a second review from others in @keycloak/core-iam team about the point above.

tre2man:
@pedroigor 

The reason why i didn't use standard sql is for speed optimization.

GenerateUpdateQueryUsingStandardSQL makes it universal, but subquery has a negative impact on performance. So i created a custom query for each rdb.

When we used standard sql, I didn't see any problems running the migration.

martin-kanis:
@pedroigor @tre2man I would not add a custom query for each major DB. Small performance optimization is simply not worth the hassle and extra code. The migration runs only once and the realm table is small. 

The second point is about localization and realm overrides. If those are used the search functionality won't work.  I know we have this problem already in admin console, but it is much less exposed as here (i.e search by role description). So one could say that if we are adding this enhancement, it should work properly for all cases.

tre2man:
In my mistake, pull request is closed. tre2man:main synced with keycloak:main. Can i create new pull request?

@martin-kanis I change migration sql to standard. As it was my first contribution, I was so concerned about optimization because I wanted to do well.

And Regarding localization and financial issues, I don't really understand the case where the search function doesn't work. Can i get any related issues or PR?

martin-kanis:
@tre2man To reproduce the issue you need to:

1. go to the realm settings and set `${key}` as a display name. 
2. go to the `Localization` tab and `Realm overrides` subtab
3. Click on `Add translation` and add `key` key and `value` value.
4. Go to `Manage realms` and see the Display name, it says `value`
5. Now try the search. It it possible to search only by `key` key which is stored in database. Not by actual overridden value `value`. 

This is how it "works" with some other searches in admin console, so it is not introduced by this PR. But, If we are adding this "feature" the ultimate goal in my opinion should be that it works for everyone and not only for those who do not use translations / realm overrides. 
Unfortunately, a fix to this problem is not easy and straightforward.

martin-kanis:
@tre2man We are not able to re-open your PR. Can you please send a new one? We discussed it with @pedroigor and we think that it can be merged even if it is not working with translations as it is the same as in other places in the admin console.

tre2man:
> @tre2man We are not able to re-open your PR. Can you please send a new one? We discussed it with @pedroigor and we think that it can be merged even if it is not working with translations as it is the same as in other places in the admin console.

I opened new PR. I'v reflected all the things mentioned earlier.

https://github.com/keycloak/keycloak/pull/47568

PR #45376, opened 2026-01-13
Add tests for mappers and representation for the admin v2

### LINKED ISSUE #45277: Review test coverage for Client v2
### Description

We have some gaps in test coverage, e.g. in rep/model mappers. The result of this issue should be either a PR with added tests, or list of follow-up issues.

edewit:
Fixes: #45277


vmuzikar (changes requested):
@edewit Thank you for the PR. We're usually trying to avoid the usage of Mockito in Keycloak unless it has some strong benefit. The benefit is that this is basically a unit test – fast and lightweight. However, it also comes with a cost of maintenance burden when setting the mocks and maintaining them.

I'd suggest to rewrite this as an integration test. This would also provide more real life and accurate results that the model layer is interacting with the mapper as expected. We could use `RunOnServerClient` or take some inspiration from the model tests.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientModelMapperSpiTest.java
vmuzikar, line 1: We don't need to test static SPIs config like that. I'd suggest removing this test class.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/OIDCClientModelMapperFactoryTest.java
vmuzikar, line 1: Similarly to the previous comment, we don't need to test the factory itself like that. AFAIK, we don't do it for other SPIs and factories either.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/OIDCClientModelMapperTest.java
```diff
@@ -0,0 +1,449 @@
...
+      clientModel.setBaseUrl("http://localhost:8080");
+      clientModel.setRedirectUris(Set.of("http://localhost:8080/callback"));
+      clientModel.setPublicClient(true);
+      clientModel.setStandardFlowEnabled(false);
+      clientModel.setImplicitFlowEnabled(false);
+      clientModel.setDirectAccessGrantsEnabled(false);
+      clientModel.setServiceAccountsEnabled(false);
+      clientModel.setWebOrigins(Set.of());
+
+      OIDCClientModelMapper mapper = new OIDCClientModelMapper(session);
```
vmuzikar: Could we obtain it through the session instead of directly instantiating it? This applies to other test methods too.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/OIDCClientModelMapperTest.java
```diff
@@ -0,0 +1,449 @@
...
+
+@KeycloakIntegrationTest(config = OIDCClientModelMapperTest.ServerConfig.class)
+public class OIDCClientModelMapperTest {
+
+  @InjectRealm
+  ManagedRealm realm;
+
+  @TestOnServer
+  public void fromModel_mapsBasicFields(KeycloakSession session) {
+    RealmModel realm = session.realms().getRealmByName("master");
```
vmuzikar: Just a non-blocking nitpick. Instead of doing this in each test method, we could do something like:
https://github.com/keycloak/keycloak/blob/c284f9ae66830df5ff869dd6f48453a992c7f340/tests/base/src/test/java/org/keycloak/tests/admin/AbstractPermissionsTest.java#L39-L40

We could then use stuff like:
https://github.com/keycloak/keycloak/blob/c284f9ae66830df5ff869dd6f48453a992c7f340/tests/base/src/test/java/org/keycloak/tests/admin/AbstractPermissionsTest.java#L61
instead of using try-finally.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/OIDCClientModelMapperTest.java
vmuzikar, line 1: Could we add `SAMLClientModelMapperTest` too, or do you want to do it as a follow-up?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/representations/BaseClientRepresentationTest.java
```diff
@@ -0,0 +1,496 @@
...
+        assertThat(rep1, not(equalTo(rep2)));
+    }
+
+    @Test
+    void testHashCodeConsistency() {
+        OIDCClientRepresentation rep1 = createRepWithAllFields();
+        OIDCClientRepresentation rep2 = createRepWithAllFields();
+        
+        assertThat(rep1.hashCode(), is(rep2.hashCode()));
+    }
```
vmuzikar: Unlike validations, I don't think we need to test getters/setters like this. It's just a POJO.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/representations/OIDCClientRepresentationTest.java
vmuzikar, line 1: As mentioned above, IMHO we need to test POJOs like this.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/representations/SAMLClientRepresentationTest.java
vmuzikar, line 1: As mentioned above, IMHO we need to test POJOs like this.

vmuzikar (commented):
@edewit Thank you for the update! I think it looks good, added some suggestions.

rest/admin-v2/tests/pom.xml
```diff
@@ -37,6 +41,51 @@
             <groupId>org.keycloak</groupId>
             <artifactId>keycloak-admin-v2-rest</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-admin-v2-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
mabartos:
```suggestion
```

It's a transitive dependency of the `keycloak-admin-v2-api`.

rest/admin-v2/tests/pom.xml
```diff
@@ -37,6 +41,51 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
```
mabartos:
```suggestion
```

These should be already included in the test framework, right?

rest/admin-v2/tests/pom.xml
```diff
@@ -37,6 +41,51 @@
...
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
```
mabartos:
```suggestion
```

We don't use the mockito anywhere, right?

rest/admin-v2/tests/pom.xml
```diff
@@ -37,6 +41,51 @@
...
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.expressly</groupId>
+            <artifactId>expressly</artifactId>
+            <scope>test</scope>
+        </dependency>
```
mabartos:
```suggestion
```
Should be already present as transitive dependencies. Correct?

mabartos (changes requested):
@edewit Nice work! The bug fix and tests themselves LGTM. 

Only a few comments related to the test `pom.xml`.

edewit:
@mabartos yeah I had some mockito tests in here before... removed them, but not the dependencies.. thanks for that

mabartos (approved):
@edewit LGTM, thanks!

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/OIDCClientModelMapperTest.java
vmuzikar: This is still valid, but it's not blocking. We can improve as a follow-up.

vmuzikar (approved):
Looks good, thank you @edewit.

Issue #44320, opened 2025-11-19
[OID4VCI] Requires credential scope although already given in AuthorizationDetails


tdiesler:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

oid4vc

### Describe the bug

```json
{
  "clientId": "oid4vci-client",
  "redirectUri": "urn:ietf:wg:oauth:2.0:oob",
  "scope": [
    "openid",
    "oid4vc_identity_credential"
  ],
  "authorizationDetails": [
    {
      "type": "openid_credential",
      "locations": [
        "https://auth.localtest.me/realms/oid4vci"
      ],
      "credential_configuration_id": "oid4vc_identity_credential"
    }
  ]
  "clientMetadata": {
    "authorization_endpoint": "urn:ietf:wg:oauth:2.0:oob"
  }
}
```

### Version

latest

### Regression

- [ ] The issue is a regression

### Expected behavior

It should be possible to use scope "openid" alone in the AuthorizationRequest. Additional credential scopes should be derived from the authorization_details

### Actual behavior

Credential scope is required in AuthorizationRequest scopes

### How to Reproduce?

Try with authorization_details and without explicit credential scope

### Anything else?

CrossRef: https://github.com/tdiesler/nessus-identity/issues/264

tdiesler:
You can assign this to me

mposolda:
+1 that this would be good to fix. The specification mentions in the [section 5](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-5.1) something like:

```
There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The 
first method involves using the authorization_details request parameter, as defined in [RFC9396], containing one or more 
authorization details of type openid_credential, as specified in Section 5.1.1. The second method utilizes scopes, as outlined 
in Section 5.1.2.
```

I can still reproduce this with latest Keycloak main as of today. When sending the OIDC authentication request (`authorization_code` flow) with only `scope=openid` and with `authorization_details` corresponding to my credential, for example this:

```
[...]

PR #45391, opened 2026-01-13
[OID4VCI] Add comprehensive tests for OID4VC authorization code flow

### LINKED ISSUE #44795: OID4VCI test coverage review
### Description

This task is to review the OID4VCI test coverage.

It seems lots of scenarios are tested properly (including succes/failure scenarios), but at the same time, there might be some scenarios, which are not tested very thoroughly. For example as of today (2025-12-10), there is only single unit test for OID4VCI credential issuance with `authorization_code` grant (The `OID4VCAuthorizationCodeFlowTestBase` having single unit test tested by 2 subclasses for different credential formats).

Awambeng:
This PR introduces a set of tests to strengthen the OID4VC authorization code flow implementation. It covers the following scenarios:

- Reuse of authorization codes results in `invalid_grant` errors.
- Invalid or malformed authorization codes are rejected.
- Token exchange without authorization details succeeds.
- Mismatched or unknown `credential_configuration_id` or `credential_identifier` triggers errors.
- Missing required parameters (`redirect_uri`, `client_id`) produce appropriate error responses.
- Malformed `authorization_details` JSON is rejected.

These tests enhance the security and robustness of the authorization code flow.

Closes #44795


forkimenjeckayang (approved):
This was reviewed internally aswell and will like others to review.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
```diff
@@ -412,6 +410,445 @@ public void testCompleteFlowWithSigningAlgorithmAndKeyIdConfigured() throws Exce
...
+        try (CloseableHttpResponse tokenHttpResponse = httpClient.execute(postToken)) {
+            assertEquals(HttpStatus.SC_BAD_REQUEST, tokenHttpResponse.getStatusLine().getStatusCode());
+            String responseBody = IOUtils.toString(tokenHttpResponse.getEntity().getContent(), StandardCharsets.UTF_8);
+            OAuth2ErrorRepresentation error = JsonSerialization.readValue(responseBody, OAuth2ErrorRepresentation.class);
+            assertEquals("invalid_request", error.getError());
+            assertTrue("Error description should indicate authorization_details processing error",
+                    error.getErrorDescription() != null && error.getErrorDescription().contains("Error when processing authorization_details"));
+        }
+    }
+
```
Captain-P-Goldfish, line 852: This test is confusing me.
The `authorization_details` parameter is meant either for the AuthorizationRequest `performAuthorizationCodeLogin` or for the TokenRequest in case of `Pre Authorized Code Flow`, but here we are using the `Authorization Code Flow` and sending the `authorization_details` parameter in the token request of type `auth_code`.
Correct me if I’m wrong, but when checking the OID4VCI spec chapter **6.1 Token Request** clearly states:
```
The following are the extension parameters to the Token Request used in the Pre-Authorized Code Flow defined in Section 3.5:
```
and defines the `authorization_details` four paragraphs later. 


So from my understanding in the `Authorization Code Flow` the `authorization_details` parameter must already be sent in the `AuthorizationRequest`.
And in the `Pre Authroized Code Flow` the `authorization_details` parameter must be sent in the `TokenRequest`

tdiesler: Yes, `authorization_details` are part of the AuthorizationRequest, become part of the OAuth2Code and when the code is redeemed, they should show up in the AccessToken. In that way, the AccessToken documents what has originally be requested. At least, that is my understanding.

Note, for the pre-authorized offer code, it would be the Issuer who associates appropriate `authorization_details` with the code. Pre-authorized code redemption works as above.

tdiesler: Yes, authorization_details are part of the AuthorizationRequest. They should become part of the OAuth2Code and when the code is redeemed, they should show up in the AccessToken. In that way, the AccessToken documents what has originally be requested. At least, that is my understanding.

Note, for the pre-authorized offer code, it would be the Issuer who associates appropriate authorization_details with the code. Pre-authorized code redemption works as above.

Awambeng: Thanks for the explanations @Captain-P-Goldfish and @tdiesler, this really helped clarify things.

My initial understanding was based on the `MAY` language in the spec, which made it seem possible to send `authorization_details` either in the Authorization Request or in the Token Request for the Authorization Code Flow, and in the Token Request for the Pre-Authorized Code Flow. Based on that, I went with the token request approach in this test.

That said, I agree with you that, technically and architecturally, it makes more sense to send `authorization_details` in the Authorization Request for the Authorization Code Flow rather than in the Token Request.

I’m happy to move this parameter to the Authorization Request for this test.

One additional concern I have is that the current implementation seems quite permissive and allows `authorization_details` to be sent in either the Authorization Request or the Token Request for the Authorization Code Flow. Do you think it would make sense to also restrict this and reject `authorization_details` in the Token Request for this flow?

tdiesler: I'd say, it is ok to allow `authorization_details` in the Token Request

> The Wallet may send the authorization_details parameter in the Token Request even when the parameter has been previously sent in the [Authorization Request](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#credential-authz-request) as described in Section 6.1 of [[RFC9396](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#RFC9396)]. It allows the AS to make a decision based on whether the Wallet is asking for more or less access than the previous request.

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-request-credential-issuance

We should reject Token Requests that ask for more than was granted.

Awambeng:
Hi @Captain-P-Goldfish @tdiesler, I’ve updated the authorization code flow handling and tests:

- `authorization_details` is now sent in the authorization request instead of the token request for auth code flow tests.
- `AuthorizationCodeGrantType` now rejects token requests that send `authorization_details` differing from what was granted and bound to the authorization code.

Please have another look when you get a chance.

tdiesler (approved):
Does it also cover 

* [#44320] Requires credential scope although already given in AuthorizationDetails
* redirect_uri=urn:ietf:wg:oauth:2.0:oob

The latter is not strictly oid4vci and probably covered elsewhere
The former is just a question, not a request

Awambeng:
> Does it also cover
> 
> * [#44320] Requires credential scope although already given in AuthorizationDetails
> * redirect_uri=urn:ietf:wg:oauth:2.0:oob
> 
> The latter is not strictly oid4vci and probably covered elsewhere The former is just a question, not a request

Thanks for the question @tdiesler.

This PR does not explicitly address #44320. The changes here focus on improving OID4VCI test coverage and on validating that the `authorization_details` used in the token request do not exceed what was granted during the authorization request.

The behavior around requiring a credential scope when `authorization_details` are already present remains unchanged and would need to be handled separately if required.

Agree, the `redirect_uri` aspect is not OID4VCI-specific and is out of scope for this PR.

mposolda:
@Awambeng @tdiesler @forkimenjeckayang Thanks for the PR and Thanks everyone for the review!

@Awambeng I hope to also look at it soon, but it seems this might introduce conflicts with https://github.com/keycloak/keycloak/pull/45784 . I hope to make sure that other PR is merged first (as pretty much any change in OID4VCI Keycloak codebase break that other PR, so it would never be possible to complete it if other things are done in the meantime...). Is it ok to you?

Awambeng:
> @Awambeng @tdiesler @forkimenjeckayang Thanks for the PR and Thanks everyone for the review!
> 
> @Awambeng I hope to also look at it soon, but it seems this might introduce conflicts with #45784 . I hope to make sure that other PR is merged first (as pretty much any change in OID4VCI Keycloak codebase break that other PR, so it would never be possible to complete it if other things are done in the meantime...). Is it ok to you?

@mposolda Thanks for the clarification. That sounds fine to me. It makes sense to merge PR #45784 first to avoid conflicts. I am okay with waiting.

Awambeng:
@tdiesler @Captain-P-Goldfish @mposolda @forkimenjeckayang I've migrated the tests to use the OAuth client utilities. Please check when you get a chance.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/oid4vc/Oid4vcCredentialRequest.java
```diff
@@ -38,6 +39,21 @@ public Oid4vcCredentialRequest proofs(Proofs proofs) {
...
+
+    /**
+     * Set a custom raw body string for the request.
+     * This is useful for testing edge cases like malformed JSON.
+     * When set, this takes precedence over the structured body.
+     *
+     * @param body the custom body string to send
+     * @return this request instance for method chaining
+     */
+    public Oid4vcCredentialRequest customBody(String body) {
```
tdiesler: I would prefer this API to stay clean. Invalid requests could be sent via the generic base class.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
```diff
@@ -58,11 +60,66 @@ public AccessTokenRequest param(String name, String value) {
...
     }
 
+    /**
+     * Omit the redirect_uri parameter from the token request.
+     * This is useful for testing edge cases where redirect_uri should not be included.
+     *
+     * @return this request instance for method chaining
+     */
+    public AccessTokenRequest omitRedirectUri() {
+        this.omitRedirectUri = true;
```
tdiesler: Again, invalid parameter configurations should not leak into this API. In contrary, invalid parameter configurations can be handled in one dedicated test case, with all other usages of this API documenting it's valid use.

tdiesler (approved):
@Awambeng much better - thanks :-)

forkimenjeckayang (approved):

mposolda (approved):
@Awambeng @tdiesler @forkimenjeckayang Thanks for the work on this PR and for the review!

PR #45407, opened 2026-01-13
Fix SearchDropdown clear button and URI search in Authorization tabs

### LINKED ISSUE #45406: SearchDropdown clear button doesn't reset form fields, URI search broken
### Before reporting an issue

### Area

admin/ui

### Describe the bug

 The search dropdown in the Authorization section (Resources, Policies, and Permissions tabs) has two issues:

  1. **Clear button doesn't visually reset form fields** - When clicking the "Clear" button, the search state is reset in the background, but the form inputs retain their values visually.

  2. **URI search doesn't work (Resources tab)** - The form uses `name="uris"` but the API expects `uri` parameter, so searching by URI never returns results.

  **Affected Areas:**
  - Clients → Authorization → **Resources** → "Search resource" dropdown
  - Clients → Authorization → **Policies** → "Search policy" dropdown
  - Clients → Authorization → **Permissions** → "Search permission" dropdown

  **Root Cause** (in `js/apps/admin-ui/src/clients/authorization/SearchDropdown.tsx`):
  1. `useForm` hook doesn't have `defaultValues` defined. When `reset({})` is called, react-hook-form sets fields to `undefined`, causing TextInput to switch to uncontrolled mode and retain DOM values.
  2. The TextControl uses `name="uris"` but the API expects `uri`.

### Version

latest

### Regression

### Expected behavior

  1. When clicking "Clear" button, all form fields should be visually empty
  2. When searching by URI, resources with matching URI should be found

### Actual behavior

  1. When clicking "Clear" button, Search button becomes disabled (state cleared in background), but form fields still show the entered values
  2. When searching by URI, no results are returned even if matching resources exist

### How to Reproduce?

 **Clear button issue:**
  1. Navigate to Admin Console → Clients → [Any client with Authorization enabled] → Authorization → Resources (or Policies/Permissions)
  2. Click the search dropdown
  3. Enter values in the form fields
  4. Close the dropdown without clicking Search or Clear
  5. Reopen the dropdown
  6. Notice the values are still there
  7. Click "Clear" button
  8. Observe: Search button becomes disabled but form fields still show the entered values

  **URI search issue (Resources tab only):**
  1. Open "Search resource" dropdown
  2. Enter any URI that exists in one of your resources and click "Search"
  3. Observe: No results found

### Anything else?

_No response_

https://github.com/user-attachments/assets/acd83e6a-e63a-4c1e-bc1a-2f9ed5296763

https://github.com/user-attachments/assets/44fa5c42-428a-4353-a510-5adbe21842c9

erenkan:
## Summary

Fixes two issues in the SearchDropdown component used in Authorization tabs (Resources, Policies, Permissions):

1. **Clear button doesn't visually reset form fields** - Form inputs retain their values after clicking Clear
2. **URI search doesn't work** - Form used `name="uris"` but API expects `uri` parameter

## Changes

- Add `defaultValues` to `useForm` for proper form reset
- Call `reset(defaultValues)` in Clear button handler
- Fix form field name from `uris` to `uri` to match API parameter
- Add missing `reset` dependency to useEffect

## Test plan

- [ ] Navigate to Clients → Authorization → Resources/Policies/Permissions
- [ ] Open search dropdown, enter values, close and reopen - values should persist (existing behavior)
- [ ] Click Clear button - all form fields should be empty
- [ ] Search by URI in Resources tab - matching resources should be found

Closes #45406

https://github.com/user-attachments/assets/67227289-47c8-412d-ab70-a7033917f8a1

### DIFF
--- a/js/apps/admin-ui/src/clients/authorization/SearchDropdown.tsx
+++ b/js/apps/admin-ui/src/clients/authorization/SearchDropdown.tsx
@@ -30,6 +30,15 @@ type SearchDropdownProps = {
   type: "resource" | "policy" | "permission";
 };
 
+const defaultValues: SearchForm = {
+  name: "",
+  type: "",
+  uri: "",
+  owner: "",
+  scope: "",
+  resource: "",
+};
+
 export const SearchDropdown = ({
   types,
   search,
@@ -37,7 +46,7 @@ export const SearchDropdown = ({
   type,
 }: SearchDropdownProps) => {
   const { t } = useTranslation();
-  const form = useForm<SearchForm>({ mode: "onChange" });
+  const form = useForm<SearchForm>({ mode: "onChange", defaultValues });
   const {
     reset,
     formState: { isDirty },
@@ -51,7 +60,7 @@ export const SearchDropdown = ({
     onSearch(form);
   };
 
-  useEffect(() => reset(search), [search]);
+  useEffect(() => reset(search), [search, reset]);
 
   return (
     <Dropdown
@@ -80,7 +89,7 @@ export const SearchDropdown = ({
           {type === "resource" && (
             <>
               <TextControl name="type" label={t("type")} />
-              <TextControl name="uris" label={t("uris")} />
+              <TextControl name="uri" label={t("uris")} />
               <TextControl name="owner" label={t("owner")} />
             </>
           )}
@@ -116,7 +125,10 @@ export const SearchDropdown = ({
             <Button
               variant="link"
               data-testid="revert-btn"
-              onClick={() => onSearch({})}
+              onClick={() => {
+                reset(defaultValues);
+                onSearch({});
+              }}
             >
               {t("clear")}
             </Button>


muhammedogz (approved):

js/apps/admin-ui/src/clients/authorization/SearchDropdown.tsx
```diff
@@ -80,7 +89,7 @@ export const SearchDropdown = ({
           {type === "resource" && (
             <>
               <TextControl name="type" label={t("type")} />
-              <TextControl name="uris" label={t("uris")} />
+              <TextControl name="uri" label={t("uris")} />
```
pedroigor, line 92: +1. Fixing a bug where the `uris` parameter is not expected by the Admin API to query resources by URI.

pedroigor (commented):
Thanks, @erenkan.

I still see a weird behavior when clearing the search parameters.

If you open the search dialog for the first time, set values for each field, and then click `Clear`, the values are cleared as expected.

However, if you open the search dialog for the first time, set values for each field, then close the dialog, and open it again, clicking `Clear` will not clear the values.

The same is true as above if you set values to only some of the fields and then close the dialog. The fields you set won't be cleared.

erenkan:
> Thanks, @erenkan.
> 
> I still see a weird behavior when clearing the search parameters.
> 
> If you open the search dialog for the first time, set values for each field, and then click `Clear`, the values are cleared as expected.
> 
> However, if you open the search dialog for the first time, set values for each field, then close the dialog, and open it again, clicking `Clear` will not clear the values.
> 
> The same is true as above if you set values to only some of the fields and then close the dialog. The fields you set won't be cleared.

Thanks for the review @pedroigor!

I couldn't reproduce the behavior you mentioned. I've tested it on my end and attached a video showing the results. Would you mind sharing a quick video of the issue so I can see exactly what's happening?

[field-clearing-steps.webm](https://github.com/user-attachments/assets/52e1a927-f1d7-4cd4-8544-4bf3a3a57e9d)

js/apps/admin-ui/src/clients/authorization/SearchDropdown.tsx
```diff
@@ -116,7 +125,10 @@ export const SearchDropdown = ({
             <Button
               variant="link"
               data-testid="revert-btn"
-              onClick={() => onSearch({})}
+              onClick={() => {
+                reset(defaultValues);
+                onSearch({});
+              }}
```
edewit, line 131:
```suggestion
              onClick={() => onSearch(defaultValues)}
```
As this will change the search and that will trigger a reset, otherwise the reset is executed twice, small nitpick

erenkan: I tested your suggestion. It works initially, but it fails in a specific scenario:

Open dialog, fill fields, click Clear -> Works.

Fill fields again (without closing), click Clear -> Fails.

Reason: In step 2, the search prop in the parent is already effectively 'empty' (from step 1). Calling onSearch(defaultValues) sends the same state again. Since the prop doesn't change, the useEffect doesn't trigger, and the form remains dirty.

Keeping the explicit reset(defaultValues) inside the click handler fixes this issue and ensures the form clears reliably every time.

[field-clear-suggestion.webm](https://github.com/user-attachments/assets/19794372-69be-432a-aa2f-6e74d3c0218e)

pedroigor (approved):
Thanks, @erenkan ! Works well now.

Let's see if @edewit has something to add so we can have it in.

PR #45427, opened 2026-01-14
45417 fixed unmanaged attributes to not allow writing when only admin can view policy is enabled

### LINKED ISSUE #45417: Unmanaged Attributes Type (Only administrators can view) allows admin API to set Unmanaged Attributes
### Before reporting an issue

### Area

user-profile

### Describe the bug

When the Unmanaged Attributes is set to "Only administrators can view" it is still possible for the admin to write unmanaged attributes. 

### Version

26.4.7

### Regression

### Expected behavior

Unmanaged attributes cannot be written when the unmanaged attribute policy is set to "Only administrators can view".

### Actual behavior

Unmanaged attributes can be written when the unmanaged attribute policy is set to "Only administrators can view".

### How to Reproduce?

Set unmanaged attribute policy to "Only Administrators can view" and do a `curl --location 'http://localhost:8081/admin/realms/school/users' \ --header 'Content-Type: application/json' \ --header 'Authorization: Bearer {{admin_access_token}}' \ --data-raw '{ "username": "user10", "email": "p@test.com", "enabled": true, "attributes": {     "className": "V" }, "credentials": [ { "type": "password", "value": "1234", "temporary": false } ] }'` and see the attribute `className` being written.

### Anything else?

_No response_

### LINKED ISSUE #45873: CVE-2025-13881 keycloak: Keycloak: Limited administrator can retrieve sensitive user attributes via Admin API
## Description

Keycloak: Limited administrator can retrieve sensitive user attributes via Admin API

The Keycloak Admin API (/unmanagedAttributes) endpoint fails to respect the visibility configuration defined in the User Profile settings. This allows an administrator with limited privileges (specifically the view-users role) to retrieve sensitive custom attributes—such as phone numbers or personal addresses—that are explicitly configured to be hidden from both users and administrators in the GUI.

---
*This issue was originally tracked in the private repository. Migrated by @stianst.*

sschu:
Fixes #45417

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultAttributes.java
+++ b/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultAttributes.java
@@ -261,6 +261,12 @@ public class DefaultAttributes extends HashMap<String, List<String>> implements 
         Map<String, List<String>> attributes = new HashMap<>(this);
 
         for (String name : nameSet()) {
+            RealmModel realm = session.getContext().getRealm();
+
+            if ((UserModel.USERNAME.equals(name) && realm.isRegistrationEmailAsUsername())) {
+                continue;
+            }
+
             if (isReadOnly(name)) {
                 attributes.remove(name);
             }
--- a/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultAttributes.java
+++ b/server-spi-private/src/main/java/org/keycloak/userprofile/DefaultAttributes.java
@@ -261,20 +261,10 @@ public class DefaultAttributes extends HashMap<String, List<String>> implements 
         Map<String, List<String>> attributes = new HashMap<>(this);
 
         for (String name : nameSet()) {
-            AttributeMetadata metadata = getMetadata(name);
-            RealmModel realm = session.getContext().getRealm();
-
-            if ((UserModel.USERNAME.equals(name) && realm.isRegistrationEmailAsUsername())
-                || isReadableOrWritableDuringRegistration(name)
-                || !isManagedAttribute(name)) {
-                continue;
-            }
-
-            if (metadata == null || !metadata.canEdit(createAttributeContext(metadata))) {
+            if (isReadOnly(name)) {
                 attributes.remove(name);
             }
         }
-
         return attributes;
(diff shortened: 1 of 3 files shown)


server-spi-private/src/main/java/org/keycloak/userprofile/DefaultAttributes.java
```diff
@@ -261,20 +261,10 @@ public Map<String, List<String>> getWritable() {
...
-            if ((UserModel.USERNAME.equals(name) && realm.isRegistrationEmailAsUsername())
-                || isReadableOrWritableDuringRegistration(name)
-                || !isManagedAttribute(name)) {
-                continue;
-            }
-
-            if (metadata == null || !metadata.canEdit(createAttributeContext(metadata))) {
+            if (isReadOnly(name)) {
                 attributes.remove(name);
             }
```
pedroigor, line 272: Can you try ?

```
for (String name : nameSet()) {

            if ((UserModel.USERNAME.equals(name) && realm.isRegistrationEmailAsUsername())) {
                continue;
            }

            if (isReadOnly(name)) {
                attributes.remove(name);
            }
}
```

The check for the username + email as username is a special case where the username is automatically set with the email. I think that is the what is causing failures.

But yeah, I think we can simplify that code a bit ...

sschu: Done. I am wondering if it would also be possible to put this condition inside the `isReadOnly` method?

pedroigor: We can try ...

sschu: There were test failures so I rolled it back.

sschu: @pedroigor Everything looks green now.

pedroigor (approved):

pedroigor:
Thanks, @sschu!

sschu:
@pedroigor Should I create the backport PRs?

Issue #43577, opened 2025-10-17
Authorization grant for Google


mposolda:
### Description

Ability to use Google ID-token to be sent as an JWT assertion compatible with RFC-7523 to the Keycloak token-endpoint .

Exchanging Google access-token is out-of-scope as Google access-token is not JWT and hence cannot be used as RFC-7523 compatible assertion.

For Google, we can consider allowing request just if `aud` from Google ID-token matches with the `Client ID` configured for Google identity-provider in the admin console.

It may allow further configuration of the Google identity-provider in order to allow the RFC-7523 requests to success (At least some On/Off switch on Google IDP to allow this).

mposolda:
Closing as all subtasks are done

PR #45443, opened 2026-01-14
Allow client_id as an audience in the JWT Authorization Grant and Client Assertions

### LINKED ISSUE #45178: Allow client_id as an audience in the JWT Authorization grant assertion
### Description

The JWT authorization grant currently requires that audience (claim `aud`) of the JWT assertion is either the issuer or token-endpoint URL of the Keycloak server. This is per [specs](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rfc7523bis-03#section-4).

However we may need an option on OIDC Identity provider that audience might be also client_id of the client configured as "client ID" option on the OIDC Identity provider. So for example if OIDC Identity provider has Client ID configured as `my_keycloak_client`, we will allow the JWT assertion with the `aud` of the value of `my_keycloak_client` . This is safe as on the domain-a side (on the 3rd party OIDC identity provider side), the client `my_keycloak_client` represents the Keycloak server. Hence it is ok to assume that when `aud` claim of the token issued by 3rd party IDP is `my_keycloak_client`, then this is JWT assertion, which is supposed to be used by Keycloak server.

This option may need to be available on demand if some switch on the OIDC Identity provider (EG. switch name might be "Allow client_id as an audience of JWT assertion") is explicitly enabled as it is not specs compliant (See [Point A of the chapter 4 of the specs](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rfc7523bis-03#section-4) )

### Additional notes

- The same option may be needed for federated client authentication as well. We can possibly re-use same option for both JWT authorization grant and federated client authentication (When Identity provider doesn't have flexibility to use issuer URL or token-endpoint URL, it will likely miss this flexibility for both federated-client authentication and JWT authorization grant. So in case that OIDC IDP is used for both JWT authz grant and federated-client authentication, we can hopefully make sure to re-use same thing for both)
- We may need to make the tooltip clear and mention security implications of use this option with care as it is not specs compliant
- Even if this option is enabled, it should be still requirement that there must be only single audience in the JWT assertion (or in the token used for federated-client authentication).

rmartinc:
Closes #45178

The PR adds a boolean option in the OIDC IdP to allow the configured Client ID as a valid audience for JWT Authorization Grant and Client Assertions (the Client ID is a valid `aud` added to the ones allowed by the respective specification). It follows the idea presented in the issue #45178. I was thinking if this is enough or we should directly allow custom audiences configured somehow in keycloak. For example, I don't know if social providers (like google in #43577) will need a more detailed configuration.

Some ideas that can extend the current PR:

1. Define a list of extra valid audiences for the IdP (all clients uses the same audiences as extra valid ones).
2. Define a list of extra valid audiences but defined per client. For example we could define one or more attributes in the IdP with the information of extra audiences per client (for example `allowedAudience.client1=aud1##aud2` and `allowedAudience.client2=aud1`. This complicates a lot the admin console part but would be more generic and valid for all the IdPs (not only OIDC).

It's a draft for the moment just in case we want to extend the functionality commented in the issue.


rmartinc:
I have added `isOIDC` in the UI part just in case.

services/src/main/java/org/keycloak/authentication/authenticators/client/FederatedJWTClientValidator.java
```diff
@@ -27,7 +31,10 @@ protected String getExpectedTokenIssuer() {
 
     @Override
     protected List<String> getExpectedAudiences() {
-        return Collections.singletonList(Urls.realmIssuer(context.getUriInfo().getBaseUri(), realm.getName()));
+        List<String> expectedAudiences = new ArrayList<>();
```
mposolda: Maybe a detail, but when the new option "allow client id as an audience" is enabled, my vote is to use client-id **instead of** token-url + issuer-url rather than **in addition to** token-url + issuer-url.

My thinking is, if IDP requires to use client-id as an audience, it does not use token-endpoint or issuer-url and hence there is no need to support it? I don't see a reason why IDP would need both client-id together with token-url and issuer-url .

@stianst @graziang WDYT?

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -3287,6 +3287,8 @@ caseSensitiveOriginalUsername=Case-sensitive username
 caseSensitiveOriginalUsernameHelp=If enabled, the original username from the identity provider is kept as is when federating users. Otherwise, the username from the identity provider is lower-cased and might not match the original value if it is case-sensitive. This setting only affects the username associated with the federated identity as usernames in the server are always in lower-case.
 supportsClientAssertions=Supports client assertions
 supportsClientAssertionReuse=Allows client assertions to be re-used
+allowClientIdAsAudience=Allow Client ID as audience for assertions
+allowClientIdAsAudienceHelp=If enabled, the Client ID configured in the Identity Provider is also valid as audience for Client Authentication and Authorization Grants assertions (Client Assertions and JWT Authorization Grant). Note this behavior is not covered by the respective specifications.
```
mposolda:
```suggestion
allowClientIdAsAudienceHelp=If enabled, the Client ID configured in the Identity Provider is also valid as audience for Federated client authentication and for JWT Authorization Grants assertions (Client Assertions and JWT Authorization Grant). Note this behavior is not covered by the respective specifications as they allow just token-url and issuer-url to be used as valid audience.
```
A small suggestion to make the tooltip little bit clearer. If we allow client-id **instead of** token-endpoint + issuer rather than **in addition to**, then the tooltip might be even a bit different.

mposolda (commented):
Thanks!

Added some inline comments and will welcome feedback.

Besides that, it seems we may need to cover this switch in the documentation? For JWT authorization grant, it is not ideal that switch is in the bottom of the page (outside of "Authorization grant settings" section), however not sure how to do it better considering that switch is applicable for both "Authorization grant settings" and "Federated client authentication".

I wonder if it is good for usability to have the same switch duplicated in UI in both sections? That would be probably better for usability, but not 100% sure about the side-effects and potential bugs (especially considering if the switch is enabled in one section, but disabled in other section etc, so it will need some clever mirroring if we go this path though and not sure whether to do this as part of this PR...)

mposolda:
@graziang Your ideas for extending the PR and cover Google makes sense. There are separate tasks for Google, so likely can be done in scope of them rather than in this PR.

services/src/main/java/org/keycloak/authentication/authenticators/client/FederatedJWTClientValidator.java
rmartinc: Okis, it makes sense. I will wait for more opinions then.

rmartinc:
@mposolda I think we can duplicate the switch (if you enable one the other will also be enabled, if react works as I think it works). But I would maintain it this way, as it's a weird option out of spec I think it's OK in the advanced settings at the bottom.

services/src/main/java/org/keycloak/authentication/authenticators/client/FederatedJWTClientValidator.java
graziang: yes at the moment it makes sense to me as well. The only doubt is that if in the future we add the ability to configure a custom audience this switch should be disabled as it would allow having only the client-id in the audience

rmartinc:
> @graziang Your ideas for extending the PR and cover Google makes sense. There are separate tasks for Google, so likely can be done in scope of them rather than in this PR.

My point here is that, if we are going to do something different to google, I would do the same for all IdPs. I think that makes no sense to do two different things, when the google one is more generic and disabled by default. Just my opinion though.

For the moment, as @graziang agrees too, I'm going to add the suggestions commented.

graziang:
@mposolda I’m ok with using the switch. We can consider adding a text field for a custom audience in the future if a use comes up. 

Regarding where to place the switch, I agree with @rmartinc that it should be in the Advanced Settings. In the “Authorization Grant Settings” section we also have the subtitle with the references to RFC 7523, but by enabling the switch we would go outside the scope of the RFC.

> @mposolda I think we can duplicate the switch (if you enable one the other will also be enabled, if react works as I think it works). But I would maintain it this way, as it's a weird option out of spec I think it's OK in the advanced settings at the bottom.

mposolda:
> @mposolda I’m ok with using the switch. We can consider adding a text field for a custom audience in the future if a use comes up.
> 
> Regarding where to place the switch, I agree with @rmartinc that it should be in the Advanced Settings. In the “Authorization Grant Settings” section we also have the subtitle with the references to RFC 7523, but by enabling the switch we would go outside the scope of the RFC.
> 
> > @mposolda I think we can duplicate the switch (if you enable one the other will also be enabled, if react works as I think it works). But I would maintain it this way, as it's a weird option out of spec I think it's OK in the advanced settings at the bottom.

@rmartinc @graziang Makes sense. Thanks! So we should be fine with the docs at least.

mposolda:
> > @graziang Your ideas for extending the PR and cover Google makes sense. There are separate tasks for Google, so likely can be done in scope of them rather than in this PR.
> 
> My point here is that, if we are going to do something different to google, I would do the same for all IdPs. I think that makes no sense to do two different things, when the google one is more generic and disabled by default. Just my opinion though.
> 
> For the moment, as @graziang agrees too, I'm going to add the suggestions commented.

IMO this switch "Allow Client ID as an audience" does not makes sense for Google in particular as for Google, the behaviour of this switch should be always enabled. The ID tokens from Google always have the client-id as an audience and never have the token-endpoint or issuer-endpoint as an audience.

Hence if "Authorization grant" is enabled for Google, it should use client-id by default IMO. In general, for Google, the behaviour would be probably slightly different than for generic OIDC IDP and hence ATM it seems better to me as different PR...

services/src/main/java/org/keycloak/authentication/authenticators/client/FederatedJWTClientValidator.java
rmartinc: Done!

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
rmartinc: I have also changed the description to use one or the other current behavior.

rmartinc:
I have done the suggestions and modified the documentation to include the new option. I have also changed the tests to test teh previous setting does not work after using the ClientID. If the tests run OK I will move out of draft.

mposolda (approved):
@rmartinc Thanks!

Leaving PR open for a bit in case @graziang or @stianst want to review and have more feedback. Otherwise hope to merge EOD tomorrow at the latest.

graziang (approved):
@rmartinc thanks!

PR #45445, opened 2026-01-14
[OID4VCI] Not able to find key for credential signature if client scope was sav…

### LINKED ISSUE #44699: Not able to find key for credential signature if client scope was saved from admin console

mposolda:
…ed from admin console

closes #44699

- Make it possible that `keyId` is considered just if it is not null or empty string (as admin-console can save this option as an empty string)

- Added automated tests for some combinations of "Signing key algorithm" and "Signing key ID" .

- Some changes done in the `OID4VCIssuerEndpoint` regarding exception handling, so that proper error is returned to the user and proper error event is triggered in case of some error (EG. Signing-key of configured KID is not available. Previously this was not catched and was just propagated as HTTP error 500 and no error event was triggered)

- I've also removed the `Details.ERROR_TYPE` in this PR and instead I've updated events to use `error` property instead. It seems to me there is no reason to have dedicated detail `error_type` as we can use proper type "error" for directly throwing the error event type.

### DIFF
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -3665,7 +3665,7 @@ tokenJwsTypeHelp=The type value written into the typ header of the JWT. Defaults
 visibleClaims=Visible Claims
 visibleClaimsHelp=Comma-separated list of claims that are always disclosed in the SD-JWT body (e.g., "id,iat,nbf,exp,jti,given_name"). Defaults to "id,iat,nbf,exp,jti". Only applicable for SD-JWT format.
 signingKeyId=Signing Key ID
-signingKeyIdHelp=Optional. The ID of the realm key used to sign the credential. If not specified, the realm's active signing key will be used automatically.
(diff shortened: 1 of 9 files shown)


wistefan (approved):
LGTM

mposolda:
@wistefan Thanks!

PR #45447, opened 2026-01-14
Avoid breaking DB changes during patch releases

### LINKED ISSUE #38888: Avoid breaking DB changes during patch releases
### Description

Breaking database changes can prevent a rolling update for patch releases. 

### Discussion

_No response_

### Motivation

_No response_

### Details

We would agree with the core teams that we wouldn't do breaking DB changes.

For that, we would add a test that would allow in any patch release JPA changelog only an allow-list of operations (for example createIndex). In the beginning, we could even ensure that no DB upgrades happen.

ryanemerson:
Closes #38888

### DIFF
--- a/model/jpa/src/test/java/org/keycloak/connections/jpa/DatabasePatchUpdateTest.java
+++ b/model/jpa/src/test/java/org/keycloak/connections/jpa/DatabasePatchUpdateTest.java
@@ -1,0 +1,150 @@
+package org.keycloak.connections.jpa;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamConstants;
+import javax.xml.stream.XMLStreamReader;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * A test to ensure that database changes are not added to the changelog of a Keycloak patch release without considering
+ * the impact on zero-downtime rolling upgrades.
+ * <p>
+ * Before a database modification is added to a patch release, it is crucial that the author of the changes verifies that
+ * the update will not lock the entire database. Updates that lock the entire database are not compatible with zero-downtime
+ * rollouts as when the lock is acquired it will prevent the existing Keycloak nodes from functioning.
+ * <p>
+ * If a database update is possible in a patch release without causing problematic locking, then the id of the liquibase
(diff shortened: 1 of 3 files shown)


model/jpa/src/test/java/org/keycloak/connections/jpa/DatabasePatchUpdateTest.java
```diff
@@ -0,0 +1,148 @@
...
+ * rollouts as when the lock is acquired it will prevent the existing Keycloak nodes from functioning.
+ * <p>
+ * If a database update is possible in a patch release without causing problematic locking, then the id of the liquibase
+ * changeset should be added to the {@link DatabasePatchUpdateTest#ALLOWED_PATCHES} {@link Set}.
+ * <p>
+ * If a database update is essential, i.e. to fix a CVE or significant performance regression, and database wide locking
+ * cannot be avoided then three steps are required:
+ * <ol>
+ * <li>Inform the wider Keycloak team that the database change is unavoidable</li>
+ * <li>Update the logic in {@link org.keycloak.compatibility.KeycloakCompatibilityMetadataProvider} so that a rolling upgrade is avoided when upgrading to the patch version that introduces the databse update</li>
```
pruivo:
```suggestion
 * <li>Update the logic in {@link org.keycloak.compatibility.KeycloakCompatibilityMetadataProvider} so that a rolling upgrade is avoided when upgrading to the patch version that introduces the database update</li>
```

model/jpa/src/test/java/org/keycloak/connections/jpa/DatabasePatchUpdateTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+                    // 1. Handle Root Element
+                    if (tagName.equals("databaseChangeLog")) {
+                        continue;
+                    }
+
+                    // 2. Process changeSet
+                    if (tagName.equals("changeSet")) {
+                        String id = reader.getAttributeValue(null, "id");
+                        Assertions.assertTrue(ALLOWED_PATCHES.contains(id), "Unexpected change '%s' in patch stream".formatted(id));
```
pruivo: I would add the `release` parameter value in the error message.

model/jpa/src/test/java/org/keycloak/connections/jpa/DatabasePatchUpdateTest.java
```diff
@@ -0,0 +1,148 @@
...
+                            if (event == XMLStreamConstants.START_ELEMENT) {
+                                depth++;
+                            } else if (event == XMLStreamConstants.END_ELEMENT) {
+                                depth--;
+                            }
+                        }
+                        continue;
+                    }
+                    // 3. Strict Requirement: Throw exception for any other element
+                    throw new RuntimeException("Unexpected element found: <%s>. Only <changeSet> is permitted within <databaseChangeLog>".formatted(tagName));
```
pruivo: nitpick, I would rather have `fail(...)` instead of throwing a runtime exception, and add the `release` parameter value in the error message

pruivo (changes requested):

pruivo (approved):

ahus1:
We've analyzed recent PR changes and found that this suggested approach doesn't handle current backporting practices. 

When backporting changes, the file name, author and the change ID need to stay constant. Sometimes we use the 26.6.0 in upstream and then backport it as in https://github.com/keycloak/keycloak/pull/45318 with the same filename 26.6.0 - and this wouldn't be detected as a breaking change here. 

There have been also some other crazy changes in those PRs:

* Changing Indexes in existing migration file
https://github.com/keycloak/keycloak/pull/45579

* A very custom change that was fixed 
https://github.com/keycloak/keycloak/pull/45559

* Fixing another changeset
https://github.com/keycloak/keycloak/pull/45266

So we discussed a different approach to keep a list of all known changes in a branch at release time. I'll update the parent issue with the latest discussion.

PR #45450, opened 2026-01-14
When creating or updating a Kubernetes IDP, check if issuer URL is unique

### LINKED ISSUE #45449: When creating or updating a Kubernetes IDP, check if issuer URL is unique
### Description

When doing this, this helps admins to avoid misconfigurations by showing an error message in the admin UI.

ahus1:
Closes #45449

### DIFF
--- a/services/src/main/java/org/keycloak/broker/kubernetes/KubernetesIdentityProviderConfig.java
+++ b/services/src/main/java/org/keycloak/broker/kubernetes/KubernetesIdentityProviderConfig.java
@@ -1,9 +1,14 @@
 package org.keycloak.broker.kubernetes;
 
+import java.util.Objects;
+
 import org.keycloak.broker.oidc.OIDCIdentityProviderConfig;
+import org.keycloak.cache.AlternativeLookupProvider;
 import org.keycloak.models.IdentityProviderModel;
+import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
 import org.keycloak.util.Strings;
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/keycloak/broker/kubernetes/KubernetesIdentityProviderConfig.java
```diff
@@ -49,5 +54,15 @@ public void validate(RealmModel realm) {
...
         checkUrl(realm.getSslRequired(), issuer, ISSUER);
+
+        KeycloakSession session = KeycloakSessionUtil.getKeycloakSession();
+        if (Objects.equals(session.getContext().getRealm().getId(), realm.getId())) {
+            AlternativeLookupProvider lookupProvider = session.getProvider(AlternativeLookupProvider.class);
+            IdentityProviderModel existingIdp = lookupProvider.lookupIdentityProviderFromIssuer(session, getIssuer());
+            if (existingIdp != null && (getInternalId() == null || !Objects.equals(existingIdp.getInternalId(), getInternalId()))) {
+                throw new IllegalArgumentException("Issuer URL already used for IDP '" + existingIdp.getAlias() + "'");
+            }
+        }
```
ahus1, line 63: @stianst - IMHO it is important to validate this for the admin UI to deliver a good UX. Still I don't have access to the session here, and the lookup is depending on the current realm of the session. 

Any thoughts on how to validate this in a different way? Thanks!

stianst (approved):

PR #45454, opened 2026-01-14
Check if two IDPs with the same issuer URL exist before caching them

### LINKED ISSUE #45453: Validate DB consistency that no two IDPs have the same issuer
### Description

The DefaultAlternativeLookupProvider will today will return the first match, but will not show an error if there are more IDPs with the same issuer URL. 

This makes it hard to spot inconsistencies. It should have the same handling as for clients, which throws an exception when there are two entries.

ahus1:
Closes #45453

### DIFF
--- a/services/src/main/java/org/keycloak/cache/DefaultAlternativeLookupProvider.java
+++ b/services/src/main/java/org/keycloak/cache/DefaultAlternativeLookupProvider.java
@@ -29,12 +29,20 @@ public class DefaultAlternativeLookupProvider implements AlternativeLookupProvid
             }
         }
 
-        IdentityProviderModel idp = session.identityProviders().getAllStream(IdentityProviderQuery.any())
+        List<IdentityProviderModel> idps = session.identityProviders().getAllStream(IdentityProviderQuery.any())
                 .filter(i -> issuerUrl.equals(i.getConfig().get(IdentityProviderModel.ISSUER)))
-                .findFirst().orElse(null);
-        if (idp != null && idp.getAlias() != null) {
-            lookupCache.put(alternativeKey, idp.getAlias());
+                .limit(2)
+                .toList();
+        IdentityProviderModel idp = null;
+        if (idps.size() == 1) {
+            idp = idps.get(0);
+            if (idp.getAlias() != null) {
+                lookupCache.put(alternativeKey, idp.getAlias());
+            }
+        } else if (idps.size() > 1) {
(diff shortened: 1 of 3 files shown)


ahus1:
@stianst - this test is now green and aligned with the other PR around the Kubernetes duplicate issuer testing. Please have a look when you have the time.

I still think that explicitly failing is a better thing than fetching the first IDP which is quite implicit, AFAIK not documented, and can change when new IDPs are added or removed. 

There is no validate-dulicate-issuer on safe similar to #45449, as this code is not specific to Kubernetes IDPs. Should there be one, and for which of them should we add it - for JWTAuthorizationGrantIdentityProviderConfig?

ahus1:
@stianst - in today's discussion we came to the conclusion that it needs a validation on editing entries with a more thorough constraint checking possibly backed by the database. See #45747

For the time being, I'd suggest to merge this PR.

services/src/main/java/org/keycloak/cache/DefaultAlternativeLookupProvider.java
```diff
@@ -29,12 +29,20 @@ public IdentityProviderModel lookupIdentityProviderFromIssuer(KeycloakSession se
             }
         }
 
-        IdentityProviderModel idp = session.identityProviders().getAllStream(IdentityProviderQuery.any())
+        List<IdentityProviderModel> idps = session.identityProviders().getAllStream(IdentityProviderQuery.any())
```
slaskawi, line 32: Is it possible to check this condition at the IdP creation time? This way a user would get feedback earlier.

ahus1: This is discussed in #45747, and I think we need both.

slaskawi: Ok it makes sense.

slaskawi (approved):

ahus1:
Marking this as a draft: It might be better to have a full database constraint to avoid those kind of situations in the first place. See #45747 for a discussion.

mposolda (approved):
Approving this PR.

IMO checking that there is unique issuer at the IDP lookup time (in `DefaultAlternativeLookupProvider.lookupIdentityProviderFromIssuer`) is always a good thing and we should do that. Not doing that might have bad implications including security implications (as request could be processed by different IDP than what admin expected).

At the same time, agree that we should ensure issuer uniqueness also at the "Write" time (CRUD of IDP). Perhaps can be done as a follow-up.  There is already another draft PR for that: https://github.com/keycloak/keycloak/pull/46024

PR #45460, opened 2026-01-15
Added typescript based module for the client admin v2

edewit:
Based on the new openapi client admin api this module can be generated
based on the defenition.

Had to change the schema a bit as the BaseRepresentation was empty and hiding it would create an invalid reference

fixes: #45365


vmuzikar (commented):
We should consider integrating the new stuff to the existing REST client, instead of creating a new client.

js/libs/keycloak-admin-client/api.yml
vmuzikar, line 1: We should get this from the server build, not store it. There might be some circular dependency but we'll need to figure that out.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
```diff
@@ -10,7 +10,7 @@
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
-@Schema
+@Schema(hidden = true)
```
vmuzikar: Just wanted to highlight that based on @Pepo48's research we might actually need to keep the BaseRepresentation. But we'll see later, we can update this if needed.

js/libs/keycloak-admin-client/src/resources/clients.ts
```diff
@@ -53,6 +54,11 @@ export interface PolicyQuery extends PaginatedQuery {
 }
 
 export class Clients extends Resource<{ realm?: string }> {
+  /**
+   * Clients v2 API - New versioned API with OpenAPI-generated client.
+   */
+  #v2: ClientsV2;
```
vmuzikar, line 60: I was thinking about one last thing – whether we should isolate this until v2 is at least a Preview Feature. I.e. somehow exclude it from the client unless explicitly enabled. Not sure what are the best practices around this in TS.

vmuzikar (commented):
@edewit Thanks for the PR and sorry for the late review. I think it's a good start, just added some last comments/thoughts.

Also, could you please check the tests? Seem to be failing.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
michalvavrik: > Had to change the schema a bit as the BaseRepresentation was empty and hiding it would create an invalid reference

I have exactly the same issue with Go client generation, which is logical as it is the same generator. I work around this with:

```
sed '/^    BaseRepresentation:$/,/^    [A-Z]/ {                                                                      
    /^      type: object$/ {                                                                                           
      a\                                                                                                               
        properties:\                                                                                                   
          id:\                                                                                                         
            type: string\                                                                                              
            description: Resource identifier\                                                                          
        additionalProperties: true
    }
  }' /home/mvavrik/sources/keycloak/quarkus/server/target/openapi.yaml > openapi-patched.yaml
```

but if it could be fixed in the definition, that would be better. I didn't find a workaround in kiota.

js/libs/keycloak-admin-client/src/resources/clients.ts
edewit: We could exclude it by putting it in a separate bundle then you need to import it from a another location `@keycloak/keycloak-admin-client/experimental`, but then it wouldn't be part of this client anymore. 

Another idea would be to have a config option to enable it and throw a error when you user it without this something like this:
```typescript
// In ConnectionConfig
  export interface ConnectionConfig {
    // ... existing fields
    enableExperimentalApis?: boolean;
  }
  // In Clients class
  v2() {
    if (!this.#experimentalEnabled) {
      throw new Error(
        "The v2 API is experimental. To use it, set `enableExperimentalApis: true` in the client
  configuration."
      );
    }
    return this.#v2.api();
  }
```

This this is a simple way to do it but would mean that the code is still there

vmuzikar: I like the `enableExperimentalApis` switch, IMHO that's a reasonable compromise.

js/libs/keycloak-admin-client/api.yml
vmuzikar: This comment is still valid. @edewit Is there any way we can omit storing the spec like this?

vmuzikar (commented):
@edewit Thank you for the update. LGTM except for https://github.com/keycloak/keycloak/pull/45460#discussion_r2763559557.

js/libs/keycloak-admin-client/api.yml
edewit: I don't see how you need the server to run to build the server

vmuzikar: I see. Created a follow-up: https://github.com/keycloak/keycloak/issues/46388. Let me know if it makes sense.

vmuzikar (approved):
LGTM, thank you.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
```diff
@@ -7,10 +7,8 @@
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
-import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
-@Schema
```
michalvavrik, line 13: @Pepo48 is dealing with this, please remove it for now, or let's put this PR on hold until that is done. Thanks

js/apps/keycloak-server/scripts/start-server.js
```diff
@@ -60,7 +60,8 @@ async function startServer() {
     path.join(SERVER_DIR, `bin/kc${SCRIPT_EXTENSION}`),
     [
       "start-dev",
-      `--features="transient-users,oid4vc-vci,declarative-ui,quick-theme,spiffe,kubernetes-service-accounts,workflows,client-auth-federated,jwt-authorization-grant"`,
+      `--features="transient-users,oid4vc-vci,declarative-ui,quick-theme,spiffe,kubernetes-service-accounts,workflows,client-auth-federated,openapi,client-admin-api:v2"`,
+      "--openapi-enabled=true",
```
michalvavrik, line 64: I may be off with this one as I don't have your context:

Why is it desirable to enable openapi by default and should it be documented? this `start-server.js` is not in a test, so I am not sure if it has significance. If it is used in production, maybe we should let users to decide? I think it could be used to infer what keycloak version is used based on available endpoints etc. I also remember some users don't want to indicate that they are using Keycloak at all (though I may be confusing this with Quarkus, don't remember where I read it 😀 )

michalvavrik (commented):
Maybe that is given, but if not, please squash the commits to one with nice commit message. I rather like the commit message of `Added typescript based module for the client admin v2` :-) Thanks

js/apps/keycloak-server/scripts/start-server.js
michalvavrik: I see in README `This app allows you to run a local development version of the Keycloak server.`, in this case forget that comment please thanks. (I can't close the comment thread myself).

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
Pepo48: I made an update after the today's sync that should address this https://github.com/keycloak/keycloak/pull/46180/changes/b670a20e0f5fc00c5590b8bcaf86727e70c69a75.

Regardless, my PR should not be in conflict with this one, I believe the base rep changes are exactly the same.

michalvavrik: As you are fine with it, it doesn't matter who makes the change. Thanks

vmuzikar:
> Maybe that is given, but if not, please squash the commits to one with nice commit message. I rather like the commit message of `Added typescript based module for the client admin v2` :-) Thanks

It'll be squashed on merge. :)

Pepo48 (approved):

michalvavrik:
I think at least JavaScript failures are related? Should I have a look?

michalvavrik:
As I need this in for my task, I'll try to reproduce it locally.

vmuzikar:
> Should I have a look?

@michalvavrik Yes, please do as Erik is away.

michalvavrik:
> > Should I have a look?
> 
> @michalvavrik Yes, please do as Erik is away.

@vmuzikar I cannot push into this PR, could you pick up https://github.com/michalvavrik/keycloak/pull/1/changes/de4f775a571a631c95c2922a364e74b5cb1a59a2 and push it, please? I tested it here https://github.com/michalvavrik/keycloak/actions/runs/22136259018/job/63988664900?pr=1 which i think is the same workflow.

vmuzikar:
@michalvavrik Feel free to open a new PR which will supersede this one.

michalvavrik:
I opened my PR, thanks. But now that I am looking into Windows failures, they are related as well I think:

```

2026-02-17T16:37:34.3851590Z [INFO] 
2026-02-17T16:37:34.3857776Z [INFO] �?� [libs/keycloak-admin-client:generate:openapi] exited with exit code 1. Output:
2026-02-17T16:37:34.3859352Z [INFO] 
2026-02-17T16:37:34.3860302Z [INFO] 'OPENAPI_FILE' is not recognized as an internal or external command,
2026-02-17T16:37:34.3862010Z [INFO] operable program or batch file.
2026-02-17T16:37:34.3862644Z [INFO]   0% [0 / 6] [2 running] themes-vendor:build
```

so I have to look at it as well.

vmuzikar:
Superseded by #46440

PR #45478, opened 2026-01-15
Use virtual threads when there are at least four cores to avoid deadlocks

### LINKED ISSUE #45015: Deadlock in Infinispan virtual threads

ahus1:
Closes #45015

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_2_13.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_2_13.adoc
@@ -1,0 +1,10 @@
+// ------------------------ Notable changes ------------------------ //
+== Notable changes
+
+Notable changes where an internal behavior changed to prevent common misconfigurations, fix bugs or simplify running {project_name}.
+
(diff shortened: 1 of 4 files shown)


ahus1:
@pruivo - please review when you have the time. Thanks!

pruivo (approved):

pedroigor (approved):

PR #45487, opened 2026-01-15
Domain matching case-insensitive in organization identity provider

### LINKED ISSUE #43613: case insensitive match on organization identity provider domain - In case 'ANY' option is chosen
### Before reporting an issue

### Area

organizations

### Describe the bug

A great fix was introduced in case #40258 to handle a miss idp identification in domain due to case sensitivity.

The issue was fixed for a case where a specific domain is chosen, but need to add the same fix for an **ANY** scenario.

Need to add this to this line as well:

https://github.com/keycloak/keycloak/blob/26.4.0/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java#L304

` idp = organization.getIdentityProviders().filter(IdentityProviderRedirectMode.EMAIL_MATCH::isSet) .filter(broker -> ANY_DOMAIN.equals(broker.getConfig().get(OrganizationModel.ORGANIZATION_DOMAIN_ATTRIBUTE))) .filter(broker -> organization.getDomains().map(OrganizationDomainModel::getName).anyMatch(domain::equals)) .findFirst().orElse(null);`

### Version

26.4.1

### Regression

### Expected behavior

We should excpect that in a case where ANY is chosen we will still be able to ignore case when comapring domain names.

Image

This will fail:

Image

But it should succeed like this case:

Image

### Actual behavior

This is failing:

Image

### How to Reproduce?

1. Register and organization with domain in lower case. For example: test.com
2. Configure IDP to this organization to automatically direct to the domain. Use the ANY option instead of choosing a specific domain.
3. Add a user to this organization from this domain. For example: [username@test.com](mailto:username@test.com).
4. Try to log in with a user from this domain, but use Upper case domain. For example: [username@TEST.com](mailto:username@TEST.com)
5. You will see keycloak does not automatically directs to this domain.

### Anything else?

Need to add the fix to this line as well:

https://github.com/keycloak/keycloak/blob/26.4.0/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java#L304

` idp = organization.getIdentityProviders().filter(IdentityProviderRedirectMode.EMAIL_MATCH::isSet) .filter(broker -> ANY_DOMAIN.equals(broker.getConfig().get(OrganizationModel.ORGANIZATION_DOMAIN_ATTRIBUTE))) .filter(broker -> organization.getDomains().map(OrganizationDomainModel::getName).anyMatch(domain::equals)) .findFirst().orElse(null);`

ornis1612:
Support Domain matching case-insensitive in organization identity provider
closes #43613

### DIFF
--- a/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
+++ b/services/src/main/java/org/keycloak/organization/authentication/authenticators/browser/OrganizationAuthenticator.java
@@ -290,10 +290,10 @@ public class OrganizationAuthenticator extends IdentityProviderAuthenticator {
             return true;
         }
 
-        // look for an idp that can match any of the org domains
+        // look for an idp that can match any of the org domains (case-insensitive)
         idp = organization.getIdentityProviders().filter(IdentityProviderRedirectMode.EMAIL_MATCH::isSet)
                 .filter(broker -> ANY_DOMAIN.equals(broker.getConfig().get(OrganizationModel.ORGANIZATION_DOMAIN_ATTRIBUTE)))
-                .filter(broker -> organization.getDomains().map(OrganizationDomainModel::getName).anyMatch(domain::equals))
+                .filter(broker -> organization.getDomains().map(OrganizationDomainModel::getName).anyMatch(domain::equalsIgnoreCase))
                 .findFirst().orElse(null);
 
         if (idp != null) {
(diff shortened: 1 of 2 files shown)


mabartos:
@pedroigor Hi! As you have assigned the parent issue, could you please review this PR? Thanks!
 
 cc: @keycloak/core-iam

martin-kanis (changes requested):
@ornis1612 Thank you for the PR and sorry about the late response. 
The fix looks good. I would just add one test `testRedirectToIdentityProviderAssociatedWithOrganizationDomainUsingAnyMatchCaseInsensitive` and remove the other two as they seems redundant. Thanks!

ahus1:
Should be resolved by https://github.com/keycloak/keycloak/pull/47425

martin-kanis:
@ornis1612 Thanks for the changes. We wanted to include it into the 26.6 release which is due soon, so I've sent a new PR with your commit and a new commit that removes the tests.

PR #45490, opened 2026-01-15
Optimize deletion of composite roles

### LINKED ISSUE #45065: Client deletion timeout due to large number of client roles
### Before reporting an issue

### Area

core

### Describe the bug

We are seeing increased timeouts since upgrading from 25.0.6 to 26.4.0 when calling delete on clients that have ~30 client roles attached. This is happening only in an environment with a large number of clients and users, so it seems the latency increase scales with database size: pushing latencies over the timeout window for large environments. We are seeing large numbers of DB queries issued from this operation, with pretty substantial time gaps between some of them (probably ispn syncing across our distributed nodes, though we haven't gotten trace data for that).

We tracked down the source of the issue to the [removeRole](https://github.com/keycloak/keycloak/blob/1e48f9e512feb4795fdbba23910b3288692e64e7/model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java#L456) method, which changed in the upgrade.

### Version

26.4.0

### Regression

### Expected behavior

Client deletion does not timeout in large environments

### Actual behavior

Client deletion times out in large environments

### How to Reproduce?

Given a telemetry agent (like datadog) is connected:
1. Run keycloak version 25
2. Delete a client
3. Run keycloak version 26
4. Delete an identical client
5. Observe a larger number of DB queries in version 26

### Anything else?

_No response_

ahus1:
Closes #45065

With this change and ~1000 clients, a delete is fast again, and should complete within 100ms with a local database running PostgreSQL: 

```
$ node clientRoleDeletionDemo.js 
deleted client 69 in 0.082 seconds
deleted client 62 in 0.085 seconds
deleted client 269 in 0.077 seconds
```

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
@@ -44,6 +44,10 @@ public class CachedGroup extends AbstractRevisioned implements InRealm {
     private final String parentId;
     private final LazyLoader<GroupModel, MultivaluedHashMap<String, String>> attributes;
     private final LazyLoader<GroupModel, Set<String>> roleMappings;
+    /**
+     * Use this so the cache invalidation can retrieve any previously cached role mappings to determine if this
+     * items should be evicted.
+     */
     private Set<String> cachedRoleMappings = new HashSet<>();
     private final LazyLoader<GroupModel, Set<String>> subGroups;
     private final Type type;
@@ -74,6 +78,10 @@ public class CachedGroup extends AbstractRevisioned implements InRealm {
         return cachedRoleMappings;
     }
 
+    /**
+     * Use this so the cache invalidation can retrieve any previously cached role mappings to determine if this
+     * items should be evicted. Will return an empty list if it hasn't been cached yet (and then no invalidation is necessary)
+     */
     public Set<String> getCachedRoleMappings() {
         return cachedRoleMappings;
     }
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
@@ -18,6 +18,7 @@
 package org.keycloak.models.cache.infinispan.entities;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -43,6 +44,7 @@ public class CachedGroup extends AbstractRevisioned implements InRealm {
     private final String parentId;
     private final LazyLoader<GroupModel, MultivaluedHashMap<String, String>> attributes;
     private final LazyLoader<GroupModel, Set<String>> roleMappings;
+    private Set<String> cachedRoleMappings = new HashSet<>();
     private final LazyLoader<GroupModel, Set<String>> subGroups;
     private final Type type;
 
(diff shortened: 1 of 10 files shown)


ahus1:
@mposolda / @vmuzikar / @mabartos - I think the performance of this eroded in multiple changes, and the final killing bit was then the `roleEntity.getParentRoles().forEach(...)` which was the parent roles with all of its children. 

I've now made `ComposeRole` its own standalone entity to avoid the previous problems, and the tests are green.

Can you please have a comment on this approach? I also had a challenge to flush the role in the right moment as Hibernate now doesn't know about the relation between CompositeRoleEntity and RoleEntity ... AFAIK I can't use RoleEntity as a primary key in ComposeRoleEntity.

Thanks!

ahus1:
@pedroigor - it should now be ready for the final review. Thanks!

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
```diff
@@ -43,6 +44,7 @@ public class CachedGroup extends AbstractRevisioned implements InRealm {
     private final String parentId;
     private final LazyLoader<GroupModel, MultivaluedHashMap<String, String>> attributes;
     private final LazyLoader<GroupModel, Set<String>> roleMappings;
+    private Set<String> cachedRoleMappings = new HashSet<>();
```
pedroigor, line 51: Isn't enough to rely on `roleMappings`? Caching is implicit when using a `LazyLoader`.

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
```diff
@@ -68,11 +70,12 @@ public MultivaluedHashMap<String, String> getAttributes(KeycloakSession session,
...
-        if (group == null) {
-            return Collections.emptySet();
-        }
-        return roleMappings.get(session, group);
+        cachedRoleMappings = roleMappings.get(session, group);
+        return cachedRoleMappings;
+    }
+
+    public Set<String> getCachedRoleMappings() {
+        return cachedRoleMappings;
```
pedroigor, line 86: Is it expected to eventually return an empty set if `getRoleMappings` wasn't called before?

ahus1: For `getCachedRoleMappings()` yes - this is the contract. I'll add some more javadoc for this.

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
ahus1: Updated in next commit.

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/entities/CachedGroup.java
ahus1: Updated in next commit.

davoustp:
Hi @ahus1, thanks for updating the initial related PR.

It looks to me that switching from an id-oriented relation table (no entity here, so Hibernate has no clue) to an entity-oriented one (there Hibernate includes it in its entities relationship graph) like you did will allow proper low-scale join selects but will actually be detrimental to the other parts of Hibernate performance, which is the dirty checking bit when looking at large scale selects: you may have a very large number of entities in Hibernate's session cache to dirty check at flush time...

My 0.02 €.

ahus1:
@davoustp - thank you, noted. Yes, there will be more entities in the context when updating entities. When reading the composite roles for a roles with `RoleAdapter.getChildRoles()`, it should should AFAIK only load the `RoleEntity`.

pedroigor (approved):

PR #45491, opened 2026-01-15
Display scheduled workflows for a user in admin console

### LINKED ISSUE #44955: Display scheduled workflows for a user in admin console
### Description

Add a new tab to the user section to display scheduled workflows for that user.

### Value Proposition

Admins can see the workflows.

### Goals

Admin can see the workflows.

### Non-Goals

Should be read-only, nothing else at this time.

### Discussion

_No response_

### Notes

_No response_

ssilvert:
Resolves #44955

Once this is approved and merged, Playwright tests will be done in a subsequent PR.


js/apps/admin-ui/src/user/UserWorkflows.tsx
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  return (
+    <KeycloakDataTable
+      key={key}
+      loader={workflowsLoader}
+      ariaLabelKey="titleWorkflows"
+      columns={[
+        { name: "name", displayKey: "name" },
+        { name: "on", displayKey: "on" },
```
pedroigor: IIRC, we decided to now show `on` because you are not forced to set an event to a workflow. This is the case when you have a scheduled workflow. See https://www.keycloak.org/docs/latest/server_admin/#_scheduling_workflows_.

Instead, we should show here the current step the user is at. The next pending step. If the last step is completed, we can perhaps show the name of the last step or `Completed`.

js/apps/admin-ui/src/user/UserWorkflows.tsx
```diff
@@ -0,0 +1,160 @@
...
+        {
+          name: "steps",
+          displayKey: "steps",
+          cellRenderer: (row) => row.steps!.length.toString(),
```
pedroigor, line 152: Is this useful information? IMO, it is more useful to show the current step the user is at, with the option to see all the steps and their status, as you did.

ssilvert:
@pedroigor @sguilhen Made the changes we discussed. Here are screenshots:

image

image

pedroigor (approved):

sguilhen (approved):

PR #45495, opened 2026-01-16
fix(scroll-form): update active section on click and scroll

### LINKED ISSUE #45494: ScrollForm navigation does not update active section on click or scroll
### Before reporting an issue

### Area

admin/ui

### Describe the bug

 The `ScrollForm` component's navigation (JumpLinks) does not update the active section indicator when:                                          
  1. A section is clicked in the navigation
  2. A new section comes into view while scrolling

#### Video

https://github.com/user-attachments/assets/1e716d33-0d6f-4d61-826e-4dca0e0fb921

### Version

latest

### Regression

### Expected behavior

- When clicking a section in the JumpLinks navigation, the active indicator should immediately highlight the clicked section
- When scrolling through sections, the active indicator should update to reflect which section is currently in view

### Actual behavior

- Clicking a section scrolls to it, but the active indicator does not update
- Scrolling does not update the active section indicator

### How to Reproduce?

1. Navigate to a page using ScrollForm (e.g., Clients → select a client → Settings tab → scroll down)
2. Click on a section in the right-side navigation
3. Observe the active indicator does not update
4. Scroll through the sections manually
5. Observe that the active section in the navigation remains unchanged as you scroll through different sections

### Anything else?

_No response_

BurakYildrm:
Closes #45494 

# Description

Fixed ScrollForm to properly update the active section in JumpLinks navigation on click and scroll. 

# Changes

1. `js/libs/ui-shared/src/scroll-form/ScrollForm.tsx`

    - Added custom scroll detection to track active section position
    - Added click handler with smooth scrolling and conflict prevention during scroll animation
    - Added bottom detection to activate last section when scrolled to end 
    - Added `SECTION_ACTIVATION_OFFSET` and `BOTTOM_SCROLL_TOLERANCE` constants for tuning

# Fix Result

[kc-scrollform-update-section-fix.webm](https://github.com/user-attachments/assets/a23a7aa4-c5b7-42f7-be00-febafb9b236c)

### DIFF
--- a/js/libs/ui-shared/src/scroll-form/ScrollForm.tsx
+++ b/js/libs/ui-shared/src/scroll-form/ScrollForm.tsx
@@ -6,24 +6,13 @@ import {
   JumpLinksItem,
   PageSection,
 } from "@patternfly/react-core";
-import {
-  Fragment,
-  ReactNode,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import { Fragment, ReactNode, useMemo } from "react";
 import { FormPanel } from "./FormPanel";
 import { ScrollPanel } from "./ScrollPanel";
 
 import style from "./scroll-form.module.css";
 
 export const mainPageContentId = "kc-main-content-page-container";
-/** Distance in pixels from top of container at which a section becomes active */
-const SECTION_ACTIVATION_OFFSET = 80;
-/** Tolerance in pixels for detecting when scrolled to bottom */
-const BOTTOM_SCROLL_TOLERANCE = 10;
 
 type ScrollSection = {
   title: string;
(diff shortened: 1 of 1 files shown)


ssilvert (commented):
I consulted with Claude Code on this and it suggested a better fix:

```
               return (
                 <JumpLinksItem
                   key={title}
-                  onClick={() => {
-                    const element = document.getElementById(scrollId);
-                    if (element) {
-                      element.scrollIntoView({
-                        behavior: "smooth",
-                        block: "start",
-                      });
-                    }
-                  }}
+                  node={`#${scrollId}`}
                   data-testid={`jump-link-${scrollId}`}
```

The `onClick` handler isn't needed with `node=` or `href=`.  It was only there as a workaround.

Here is what it had to say about using `node=` instead of `href=`


> Both props tell PF's JumpLinks where to find the section DOM elements for scroll-spy tracking (getScrollItems reads child.props.node || child.props.href). The difference is what each prop does beyond that:
> 
> href={#${scrollId}} (the PR's approach)
> 
> PF renders href="#login-settings" on the inner <a> element
> Clicking the link makes the browser follow the anchor: it changes window.location.hash to #login-settings and fires a popstate event
> Keycloak Admin UI uses createHashRouter from React Router, which listens for popstate to drive navigation
> React Router sees the new hash #login-settings, treats it as a route path, finds no matching route, and navigates away from the client settings page — the page disappears
> node={#${scrollId}} (the better approach)
> 
> PF uses node only internally to locate the section element; it is not rendered onto the <a> tag
> The inner <a> has no href, so clicking it fires PF's own onClick handler (which calls setActiveIndex and scrolls the container) but triggers no browser navigation
> No popstate fires, React Router stays on the current route, the page stays intact
> In short: href has a dual purpose in HTML (both a link destination and a hook for JS libraries), and that dual purpose causes a conflict with hash routing. node is a PF-private prop that satisfies the library's need to find the DOM element without the side-effect of making the <a> a navigable link.

Would you like to change this to use Claude's suggestion?

ssilvert:
@BurakYildrm Sorry, I didn't realize this PR was so old.  I guess you didn't see my recent comment.  I'm going to close this PR and just open a new one to fix this.  Thanks for the contribution.

ssilvert:
Actually, Claude's solution breaks one of the tests.  I'm going to use yours as long as the tests still pass.

ssilvert:
Your code makes the same tests fail.  Closing again.

PR #45521, opened 2026-01-16
Test for authorization chaining across domains

### LINKED ISSUE #45468: Review tests for authorization chaining across domains
### Description

Review if we have some automated tests for the scenario outlined in the authorization chaining across domains documentation: https://github.com/keycloak/keycloak/issues/45466 .

It is possible that we have automated tests for this already. But maybe not (maybe we have tests for testing issuing access-tokens for specific audience and in different place tests for rfc-7523 authz grant). We can maybe add some simple automated test class for testing both parts (domain-a, domain-b) to make sure that we test the use-case, which we document.

rmartinc:
Closes #45468

Just adding a test to do the authorization chaining across domains. The only difference with the demo is that the audience mapper is added in the dedicated scope of the client (adding a client scope is complicated because you need all the default scopes too). The PR just adds one test. If you think we should check more scenarios just let me know.

### DIFF
--- a/tests/base/src/test/java/org/keycloak/tests/oauth/AuthChainingAcrossDomainsTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/oauth/AuthChainingAcrossDomainsTest.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2026 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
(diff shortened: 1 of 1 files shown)


graziang (approved):
@rmartinc Thanks! I think this test is enough for now as we already have all the dedicated tests for JWT Authorization Grant and Standard Token Exchange, If any other use cases come up we can add more tests in the future.

mposolda (approved):

PR #45534, opened 2026-01-16
Keycloak should not allow matrix parameters in URLs as we don't use them

### LINKED ISSUE #45533: Keycloak should not allow matrix parameters in URLs as we don't use them
### Before reporting an issue

### Area

_No response_

### Describe the bug

As part of the JAX-RS features each path element allows RFC-compliant matrix parameters, although Keycloak is not using them anywhere. They are basically ignored.

It would be good to not allow any matrix parameter to harden Keycloak until we actually make use of them.

### Version

main

### Regression

### Expected behavior

Return a 400 response when there is a semicolon in the URL

### Actual behavior

Matrix parameters are silently ignored.

### How to Reproduce?

Use a semicolon in a URL

### Anything else?

I'll prepare a PR

ahus1:
Closes #45533

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -39,6 +39,17 @@ This change should prevent deadlocks due to pinned virtual threads.
 
 Specific sensitive information is omitted from the HTTP Access log, such as the value of the `Authorization` HTTP header. Moreover, values of specific sensitive {project_name} cookies, such as `KEYCLOAK_SESSION`, `KEYCLOAK_IDENTITY`, or `AUTH_SESSION_ID`, are also omitted.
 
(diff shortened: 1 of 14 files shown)


ahus1:
@mabartos - as you recently reviewed a pull request similar to this one, I would be happy if you could review this one as well. Thanks!

mabartos (dismissed):
@ahus1 LGTM, thanks.

I'm just wondering if this might be a breaking change for deployments that use the `RealmResourceProvider`s and `AdminRealmResourceProvider`s. But, I don't expect that the majority of these users/deployers use matrix parameters; quite the opposite. Moreover, these SPIs are internal, so we should be fine.

stianst (dismissed):

mabartos:
@ahus1 JFYI there's a conflict.

mabartos (approved):

pedroigor (approved):

PR #45537, opened 2026-01-16
Add backchannel logout on user deletion

### LINKED ISSUE #45120: OIDC BackChannel logout does not get called when the user gets deleted
### Before reporting an issue

### Area

oidc

### Describe the bug

When I delete a user out of keycloak the oidc backchannel logout url does not get called for this user.
This can lead to users still have access to services even if they are deleted.

### Version

26.4.7

### Regression

### Expected behavior

If I delete a user from keycloak the user all configured backchannel logout urls should get called so that the user can be logged out of all services correctly.

### Actual behavior

OIDC Backchannel url gets ignored.

### How to Reproduce?

1. Configure BackChannel logout for an OIDC client
2. create a user and authroize it to the client
3. delete the user
4. no back channel url gets called

### Anything else?

_No response_

Rathan-Naik:
Centralize logout logic in UserSessionUtil.logoutAllUserSessions(). Add setNotBeforeForUser to invalidate tokens before deletion. Skip setNotBeforeForUser for lightweight users (they don't persist). Trigger backchannel logout for all user sessions (online and offline). Apply to all three deletion paths:
- Admin API deletion (UserResource.deleteUser)
- Self-service deletion (DeleteAccount.processAction)
- Workflow deletion (DeleteUserStepProvider.run)

This ensures that when a user is deleted, all clients with backchannel logout URLs configured are notified via backchannel logout requests.

Closes #45120

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/requiredactions/DeleteAccount.java
+++ b/services/src/main/java/org/keycloak/authentication/requiredactions/DeleteAccount.java
@@ -46,6 +46,7 @@ import org.keycloak.provider.ProviderConfigProperty;
 import org.keycloak.services.managers.AuthenticationManager;
 import org.keycloak.services.managers.AuthenticationSessionManager;
 import org.keycloak.services.messages.Messages;
+import org.keycloak.services.util.UserSessionUtil;
 import org.keycloak.sessions.AuthenticationSessionModel;
 
 import org.jboss.logging.Logger;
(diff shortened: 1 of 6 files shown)


ahus1:
@graziang - there is now this PR for the bug that triaged to be "help wanted". Can you or someone from your team please have a look?

graziang:
Thanks @Rathan-Naik, and sorry for the delay.

Note that the linked issue #45120 is closed as not planned, so it should be reopened before this can move forward.
To continue with this PR we need to rebase with the main branch and check the behavior is still correct.
Are you still interested to continue the work on this PR? If we don't hear back, we can close this for now and it can be reopened later.

PR #45541, opened 2026-01-16
Fix: Prevent brute force counter increment on session limit denial

### LINKED ISSUE #44474: Keycloak Maximum Concurrent Session Limit is incorrectly registered as a login failure, triggering Keycloak's Brute Force protection

sabeer02:
### Root Cause
 
Users with valid credentials were being locked out when hitting session limits because the system treated session denials as authentication failures, incrementing the brute force counter.

### Solution
 
Changed DENY_NEW_SESSION handling from context.failure() to context.forceChallenge()
Updated error event from GENERIC_AUTHENTICATION_ERROR to ACCESS_DENIED
Updated tests to match new error behavior
 
### Impact
 
Session limit denials no longer increment brute force counters, preventing legitimate users from being locked out due to session limits.
 
Fixes: #44474

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/sessionlimits/UserSessionLimitsAuthenticator.java
@@ -174,9 +174,9 @@ public class UserSessionLimitsAuthenticator implements Authenticator {
                         .map(f -> f.get(UserSessionLimitsAuthenticatorFactory.ERROR_MESSAGE))
                         .orElse(SESSION_LIMIT_EXCEEDED);
 
-                context.getEvent().error(Errors.GENERIC_AUTHENTICATION_ERROR);
+                context.getEvent().error(Errors.ACCESS_DENIED);
                 Response challenge = context.form().setError(errorMessage).createErrorPage(Response.Status.FORBIDDEN);
-                context.failure(AuthenticationFlowError.GENERIC_AUTHENTICATION_ERROR, challenge, eventDetails, errorMessage);
+                context.forceChallenge(challenge);
                 return Collections.emptyList();
 
             case UserSessionLimitsAuthenticatorFactory.TERMINATE_OLDEST_SESSION:
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/sessionlimits/UserSessionLimitsTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/sessionlimits/UserSessionLimitsTest.java
@@ -146,7 +146,7 @@ public class UserSessionLimitsTest extends AbstractTestRealmKeycloakTest {
         // Login the same user again and verify the configured error message is shown
         loginPage.open();
         loginPage.login("test-user@localhost", "password");
-        events.expect(EventType.LOGIN_ERROR).user((String) null).error(Errors.GENERIC_AUTHENTICATION_ERROR).assertEvent();
+        events.expect(EventType.LOGIN_ERROR).user((String) null).error(Errors.ACCESS_DENIED).assertEvent();
         errorPage.assertCurrent();
         assertEquals(ERROR_TO_DISPLAY, errorPage.getError());
     }
@@ -211,7 +211,7 @@ public class UserSessionLimitsTest extends AbstractTestRealmKeycloakTest {
             // Login the same user again and verify the configured error message is shown
             loginPage.open();
             loginPage.login("test-user@localhost", "password");
-            events.expect(EventType.LOGIN_ERROR).user((String) null).error(Errors.GENERIC_AUTHENTICATION_ERROR).assertEvent();
+            events.expect(EventType.LOGIN_ERROR).user((String) null).error(Errors.ACCESS_DENIED).assertEvent();
             errorPage.assertCurrent();
             assertEquals(ERROR_TO_DISPLAY, errorPage.getError());
         } finally {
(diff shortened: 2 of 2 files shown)


sabeer02:
Flaky test reported: https://github.com/keycloak/keycloak/issues/45546

sabeer02:
@ahus1 PR #45541 CI passed successfully after re-trigger, Ready for review !

CI_passed_after_retrigger_45541

ahus1:
@mposolda - as marked the parent PR as help-wanted, can you or someone from your team please have a look?

mabartos:
@mposolda Even though this solves the issue of not incrementing the counter of the brute force protector, Keycloak should have some more fine-grained control over the protector. 

But it's out of scope of this PR, I just wanted to share my 2 cents.

To be more specific - IMO, not all the login failures should be counted by the protector. The `logFailure()` execution provokes incrementing the counter, and based on the current usage in Keycloak, the user can be locked down when the auth execution is not found (admin misconfiguration, deployment issues,...), or when the user might not have a specific attribute (like when using the DenyAccessAuthenticator), missing permissions for the authenticated user, missing OIDC scope for userinfo endpoint,... it needs to be executed multiple times, but is it really a brute force attack? 

Maybe it'd be good to consider the use cases for the brute force protector and having protected Keycloak only from the brute force attacks. But as there's not so many complaining about user lock-downs, this can be considered as an enhancement for the protector. Or do you have any other opinion? Thanks!

sabeer02:
Thanks @mabartos for the review!

@mposolda -  Could you please share your review and thoughts.

Would it be possible for the team to consider merging this in time for the next patch release.

mabartos:
@mposolda Adding a little bit more context to my previous message -> The brute force protector we have is now the only DoS protection. It means that when I was referring to not incrementing the counter for the brute force protector when some actions happen, they'd need to be handled by some other mechanism. I don't want to loosen the brute force protector radius without the next mechanism, for which you could specify how to treat these users with a potential DoS attack. But it's rather a future vision/idea. 

Btw. IMO, it might be ok to not increment the counter for this session limit check.

sabeer02:
@mposolda - Could you please review and approve this PR when you have a moment.

mposolda:
@pskopek Could you please take a look at this PR? Is it aligned with the work you're doing for brute-force protector?

sabeer02:
Hi @mposolda and @pskopek,
I wanted to follow up on this PR to see if there are any updates or if there is anything specific you'd like me to adjust, Please let me know. 
Thanks

pskopek:
Hi @sabeer02 , my new [PR regarding slightly different](https://github.com/keycloak/keycloak/pull/46168) problem is fix this problem. 
Can you check it, please?

sabeer02:
Hi @pskopek, I cherry-picked the commits from your PR and tested with sample values on my end to verify the behavior. However, it looks like the session limit issue persists.

Observations:
- Brute Force is still triggered/incremented when the session is denied, leading to a user lockout.
- The event error remains generic_authentication_error. In my PR, I updated this to ACCESS_DENIED to make the rejection clearer.

User Locked out:
Screenshot from 2026-02-11 19-30-40

Logs for reference:

> 
> 2026-02-11 19:23:24,771 INFO  [org.keycloak.authentication.authenticators.sessionlimits.UserSessionLimitsAuthenticator] (executor-thread-7) Denying new session
> 2026-02-11 19:23:24,772 WARN  [org.keycloak.events] (executor-thread-7) type="LOGIN_ERROR", realmId="fce36c65-d206-4cfb-bfc3-7dd0f63c65fb", realmName="master", clientId="security-admin-console", userId="null", ipAddress="127.0.0.1", error="generic_authentication_error", auth_method="openid-connect", auth_type="code", redirect_uri="http://localhost:8080/admin/master/console/", code_id="jqwZGnJ_ZhTecZjnKDIltuVW"


**Please let me know if I am missing anything ? Also, if you have a moment, please check if you're able to reproduce the problem.**

 In my case, once the session limit is hit, the subsequent "denied" attempts are still being counted toward the brute force threshold.


Other Screenshots for reference:
Screenshot from 2026-02-11 19-29-46

Screenshot from 2026-02-11 19-32-02
Screenshot from 2026-02-11 19-32-39
Screenshot from 2026-02-11 19-33-11

sabeer02:
Hi @pskopek and @mposolda,

IMHO, I believe both PR #46168 and PR #45541 solve related but different brute force scenarios. PR #46168 addresses OTP-based brute force attacks, while PR #45541 is purely about session limit denials.

Would it be possible to proceed with this PR #45541 to close issue #44474? Thanks !

sabeer02:
@pskopek and @mposolda ,

could someone please take another look?

sabeer02:
@pskopek and @mposolda  - Any comments

pskopek:
Hi @sabeer02 ! Thanks for the PR. With the latest changes to the brute force protector we can fix the issue using proper of `UserSessionLimitsAuthenticatorFactory:getReferenceCategory`.
It should return non-null string different from categories sensitive to brute force protector. E.g. `"not-brute-force-protected"`. 

I suggest to close this PR and open the new one with the fix above.

mposolda:
Closing as suggested by @pskopek

PR #45547, opened 2026-01-17
Fix OIDC IdP logout when session already terminated via backchannel

Rathan-Naik:
When using an OIDC Identity Provider with backchannel logout configured, a race condition can occur during browser-based logout:

1. Consumer redirects to Provider for logout
2. Provider sends backchannel logout to Consumer (session deleted)
3. Provider redirects browser back to Consumer's logout_response endpoint
4. Consumer fails to find session and shows error page

This fix handles the case where the session is already gone gracefully:
- Log a warning instead of error (session was legitimately logged out)
- Record success event instead of error
- Clear identity cookies to ensure clean state
- Redirect to account page instead of showing error

Closes #45536

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
@@ -17,6 +17,7 @@
 package org.keycloak.broker.oidc;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
```diff
@@ -385,11 +386,17 @@ public Response logoutResponse(@QueryParam("state") String state) {
             }
             UserSessionModel userSession = session.sessions().getUserSession(realm, state);
             if (userSession == null) {
-                logger.error("no valid user session");
+                logger.warnf("User session not found for state %s - likely already logged out via backchannel", state);
```
mposolda, line 389: It seems that if userSession is not found, it should try to lookup the "offline" user session if I understand correctly the description of #45536 . See for example `UserSessionManager.indOfflineUserSession` .

I suppose the fallback might be needed just if also offline session not found.

mposolda (commented):
@Rathan-Naik Thanks for the PR. Added comment inline. Could you please also try to add automated test(s) for this scenario?

mposolda:
Closing due the lack of feedback. Please re-open (or rather open new PR) if this is still an issue. Please incorporate the changes from the above.

PR #45556, opened 2026-01-19
Hot deploy custom providers from module to test server

### LINKED ISSUE #34188: Deploy custom providers from module to test server
This is mainly for extension developers, so they can easily deploy and test a provider, ideally without having to do a Maven build.

For a module using the test framework it should be possible to optionally deploy the module itself to Keycloak.

An example includes a Maven module with `src/main/` containing a custom realm resource provider, as well as various other classes and resources. In `src/test/java` there is a test that is testing the custom provider, this would be specified by a server config that deploys the module itself:

```
KeycloakTestServerConfig {
    default boolean includeCustomProviders() {
        return true;
    }
}
```

A basic implementation could just deploy the Maven artifact, and require building the module first with Maven. This would not have the best experience as it would be required to build with Maven first. A better approach would be to deploy from sources built by the IDE so there is no need for a separate build.

vaceksimon:
Closes #34188 

Adds the `KC_TEST_SERVER_HOT_DEPLOY=[true/false]` config option. It affects only the `distribution` Keycloak server.

If enabled, dependencies and providers marked as `hotDeployable` will be packaged into a JAR from the most recent compiled sources and deployed.

This eliminates the need to compile this dependency before running tests with the `distribution` Keycloak server.

Also adds the option to deploy a provider from the current module with `KeycloakServerConfigBuilder#dependencyCurrentProject()`

### DIFF
--- a/test-framework/core/pom.xml
+++ b/test-framework/core/pom.xml
@@ -84,12 +84,10 @@
         <dependency>
             <groupId>org.jboss.shrinkwrap</groupId>
             <artifactId>shrinkwrap-api</artifactId>
-            <version>${version.shrinkwrap}</version>
         </dependency>
         <dependency>
             <groupId>org.jboss.shrinkwrap</groupId>
             <artifactId>shrinkwrap-impl-base</artifactId>
-            <version>${version.shrinkwrap}</version>
         </dependency>
     </dependencies>
 
--- a/test-framework/core/pom.xml
+++ b/test-framework/core/pom.xml
@@ -81,6 +81,16 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-bootstrap-maven-resolver</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-api</artifactId>
+            <version>${version.shrinkwrap}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-impl-base</artifactId>
+            <version>${version.shrinkwrap}</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
@@ -35,11 +35,11 @@ import org.keycloak.it.utils.Maven;
 import org.keycloak.quarkus.runtime.Environment;
 import org.keycloak.representations.info.ServerInfoRepresentation;
 import org.keycloak.testframework.util.FileUtils;
+import org.keycloak.testframework.util.JarUtil;
 import org.keycloak.testframework.util.ProcessUtils;
 import org.keycloak.testframework.util.TmpDir;
 
 import io.quarkus.fs.util.ZipUtils;
-import io.quarkus.maven.dependency.Dependency;
 import org.jboss.logging.Logger;
 import org.jetbrains.annotations.NotNull;
 
@@ -69,7 +69,7 @@ public class DistributionKeycloakServer implements KeycloakServer {
         this.tlsEnabled = tlsEnabled;
 
         List<String> args = keycloakServerConfigBuilder.toArgs();
-        Set<Dependency> dependencies = keycloakServerConfigBuilder.toDependencies();
+        Set<KeycloakDependency> dependencies = keycloakServerConfigBuilder.toDependencies();
 
         try {
             boolean installationCreated = createInstallation();
@@ -87,7 +87,7 @@ public class DistributionKeycloakServer implements KeycloakServer {
                 String startedWithArgs = startupArgsFile.isFile() ? FileUtils.readStringFromFile(startupArgsFile) : null;
                 String requestedArgs = String.join(" ", args);
 
-                Set<String> requestedDependencies = dependencies.stream().map(d -> d.getGroupId() + "__" + d.getArtifactId() + ".jar").collect(Collectors.toSet());
+                Set<String> requestedDependencies = dependencies.stream().map(DistributionKeycloakServer::getProviderJarName).collect(Collectors.toSet());
                 Set<String> startedWithDependencies = existingProviders.stream().map(File::getName).collect(Collectors.toSet());
 
                 if (requestedArgs.equals(startedWithArgs) && setEquals(requestedDependencies, startedWithDependencies)) {
@@ -174,7 +174,11 @@ public class DistributionKeycloakServer implements KeycloakServer {
         return outputHandler;
     }
 
-    private static void updateProviders(List<File> existingProviders, Set<Dependency> dependencies, File providersDir) throws IOException {
+    private static String getProviderJarName(KeycloakDependency dependency) {
+        return dependency.getGroupId() + "__" + dependency.getArtifactId() + ".jar";
+    }
+
+    private static void updateProviders(List<File> existingProviders, Set<KeycloakDependency> dependencies, File providersDir) throws IOException {
         existingProviders.stream()
                 .filter(f -> f.getName().endsWith(".jar"))
                 .filter(f -> {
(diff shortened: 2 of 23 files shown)


test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServer.java
```diff
@@ -10,4 +12,8 @@ public interface KeycloakServer {
 
     String getManagementBaseUrl();
 
+    static boolean getDependencyHotDeployEnabled() {
+        return Boolean.parseBoolean(Config.getValueTypeConfig(KeycloakServer.class, "hot.deploy", "false", String.class));
+    }
+
```
vaceksimon, line 18: If `KC_TEST_SERVER_HOT_DEPLOY=true`, all dependencies with `hotDeployable()` will be looked up in the project and packaged in a JAR. Otherwise, all dependencies are treated the same

test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakDependency.java
vaceksimon, line 1: A wrapper that only adds the `hotDeployable` attribute to a dependency. Used to differentiate dependencies during deployment.

Initially, I used two sets of dependencies, one of normal dependencies and one for hotDeployable dependencies. It led to code duplication, especially in the `EmbeddedKeycloakServer`, which doesn't care about hot deployment.

vaceksimon:
I made some adjustments to the `Maven` util too. I believe that's maintained by the cloud native team. Can you take a look @mabartos?

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
```diff
@@ -188,23 +192,41 @@ private static void updateProviders(List<File> existingProviders, Set<Dependency
                     FileUtils.delete(new File(f.getAbsolutePath() + ".lastModified"));
                 });
 
+        boolean hotDeployEnabled = KeycloakServer.getDependencyHotDeployEnabled();
         Path providersPath = providersDir.toPath();
-        for (Dependency d : dependencies) {
-            Path dependencyPath = Maven.resolveArtifact(d.getGroupId(), d.getArtifactId());
+
+        for (KeycloakDependency d : dependencies) {
+            boolean shouldBuildFromSources = hotDeployEnabled && d.isHotDeployable();
```
stianst: buildFromSources is not accurate, it packages already built sources, and doesn't build

stianst (commented):
I tried this out with an extension, but getting an error:
```
ERROR:     FAILED - io.stianst.test.MyProviderTest / myTest

java.lang.RuntimeException: Failed to find keycloak-parent module.

	at org.keycloak.it.utils.Maven.getRootModuleProject(Maven.java:160)
	at org.keycloak.it.utils.Maven.findLocalModule(Maven.java:168)
	at org.keycloak.testframework.server.DistributionKeycloakServer.getDependencyPath(DistributionKeycloakServer.java:224)
	at org.keycloak.testframework.server.DistributionKeycloakServer.updateProviders(DistributionKeycloakServer.java:202)
	at org.keycloak.testframework.server.DistributionKeycloakServer.start(DistributionKeycloakServer.java:105)
	at org.keycloak.testframework.server.AbstractKeycloakServerSupplier.getValue(AbstractKeycloakServerSupplier.java:82)
	at org.keycloak.testframework.server.AbstractKeycloakServerSupplier.getValue(AbstractKeycloakServerSupplier.java:23)
	at org.keycloak.testframework.injection.Registry.deployRequestedInstances(Registry.java:215)
	at org.keycloak.testframework.injection.Registry.beforeEach(Registry.java:119)
	at org.keycloak.testframework.KeycloakIntegrationTestExtension.beforeEach(KeycloakIntegrationTestExtension.java:32)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

Seems currently it will only work for extensions that are in the Keycloak project, and not custom extensions.

The example I used was https://github.com/stianst/playground/tree/main/kc-ext/test-extension

stianst:
How does this not affected the embedded server? Doesn't that also just add a dependency on the built jar rather than package classes built by the IDEA?

mabartos:
> How does this not affected the embedded server? Doesn't that also just add a dependency on the built jar rather than package classes built by the IDEA?

@stianst Embedded server does not use the Maven utility at all and has its own maven context resolver for these dependencies. AFAIK, it should not build anything automagically, and it just puts the required deps JARs on the classpath - artifacts are resolved, not built. If you've built the JAR locally, the mvn context should take it from the mvn cache, otherwise trying to obtain it from the maven servers (mvn central).  

So, that's why the embedded server is not affected by these changes.

stianst:
> > How does this not affected the embedded server? Doesn't that also just add a dependency on the built jar rather than package classes built by the IDEA?
> 
> @stianst Embedded server does not use the Maven utility at all and has its own maven context resolver for these dependencies. AFAIK, it should not build anything automagically, and it just puts the required deps JARs on the classpath - artifacts are resolved, not built. If you've built the JAR locally, the mvn context should take it from the mvn cache, otherwise trying to obtain it from the maven servers (mvn central).
> 
> So, that's why the embedded server is not affected by these changes.

The whole point of this is to use the classes built by the IDE for providers/extensions, not by Maven. I should basically be able to update a provider in IntelliJ then run a test without requiring a Maven build. So, not quite following your comment.

mabartos:
> The whole point of this is to use the classes built by the IDE for providers/extensions, not by Maven. I should basically be able to update a provider in IntelliJ then run a test without requiring a Maven build. So, not quite following your comment.

@stianst My comment was only intended to show the difference between the embedded and distribution server for the Maven.java util class. I'm not familiar with the problem you're trying to solve in this PR yet. I'll try to find some time to look at this if necessary, but if you confirm it works, it may be fine :)

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
vaceksimon: renamed to `shouldPackageClasses`

vaceksimon:
- fixed it for extensions
- added the `dependencyCurrentProject` method, mainly for extensions
- moved the utility methods from `Maven` to the framework's `MavenProjectUtil`

@stianst @lhanusov

test-framework/core/pom.xml
```diff
@@ -95,6 +95,16 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-bootstrap-maven-resolver</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-api</artifactId>
+            <version>${version.shrinkwrap}</version>
```
stianst: Remove version, and add to dependencyManagement in /test-framework/pom.xml

test-framework/core/pom.xml
```diff
@@ -95,6 +95,16 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-impl-base</artifactId>
+            <version>${version.shrinkwrap}</version>
```
stianst: Remove version, and add to dependencyManagement in /test-framework/pom.xml

test-framework/examples/providers/src/test/java/org/keycloak/providers/example/MyCustomProviderWithinSameModuleTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+
+/**
+ *
+ * @see org.keycloak.providers.example.MyCustomRealmResourceProvider
+ * @see org.keycloak.test.examples.MyCustomProviderTest
+ * @author <a href="mailto:svacek@redhat.com">Simon Vacek</a>
+ */
+@KeycloakIntegrationTest(config = MyCustomProviderWithinSameModuleTest.ServerConfig.class)
+public class MyCustomProviderWithinSameModuleTest {
```
stianst, line 26: Use SimpleHttp instead of creating your own HttpClient:

https://github.com/stianst/playground/blob/main/kc-ext/test-extension/src/test/java/io/stianst/test/MyProviderTest.java

test-framework/examples/providers/pom.xml
```diff
@@ -36,6 +36,12 @@
             <groupId>jakarta.ws.rs</groupId>
             <artifactId>jakarta.ws.rs-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.keycloak.testframework</groupId>
+            <artifactId>keycloak-test-framework-core</artifactId>
+            <version>${version}</version>
```
stianst:
```suggestion
            <version>${project.version}</version>
```

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
stianst, line 1: This is getting a bit messy; could do with a bit of cleanup, perhaps extracting some stuff into a utility class or something?

stianst (commented):
Nice, tested it out and seems to work perfectly :)

Didn't get a chance to review it fully, but have a few comments. In general it could do with a little bit of cleanup/refactoring as some of the code is a bit messy/complex.

lhanusov:
There should be a documentation around it, it's not trivial and people may lost when trying to get things done.
Please, create a follow-up ticket or make it part of this PR, I am fine with both. :-)

Good job @vaceksimon :+1:

stianst (approved):

PR #45558, opened 2026-01-19
[OID4VCI] Add replay protection for credential offers by reference

### LINKED ISSUE #44660: Credential offer by reference can be triggered multiple times
### Description

When the credential-offer is sent by reference, the wallet is supposed to invoke the credential offer URL, which typically look something like this (the last part of the path is unique "nonce"): https://as.keycloak-fapi.org:8443/realms/test/protocol/oid4vc/credential-offer/WvGzjQdCaW3oJBimFWDBClcsqnXs_nzK2IB3eOaVjm-XxDbnZfC2PZapZYtYSA9hwnhvzuL6ytGiCr2jejGjPw .

This URL can be triggered multiple times right now. This means that also pre-authorized code could be potentially obtained multiple times. The specification does not seem to have anything about this URL needs to be unique in this section https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-sending-credential-offer-by- . It has only something about make sure that pre-authorized code should be unique https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-replay-prevention .

This issue is to make sure that we either allow only single request for obtain the pre-authorized code from the credential-offer URL for given "nonce" , or we ensure that pre-authorized code is unique to be used in the token-request of pre-authorized grant.

IMO the replay protection should work even if `tx_code` is not used (I consider `tx_code` handling out of scope of this issue).

Awambeng:
- Add consumed flag to CredentialOfferState to track offer access
- Implement checks in getCredentialOffer endpoint to ensure offers can only be consumed once
- Mark offers as consumed after first access and throw error on subsequent attempts
- Add testCredentialOfferReplayProtection to verify same nonce fails on replay
- Add testCredentialOfferDifferentNoncesIndependent to verify different nonces work independently

This prevents replay attacks where the same credential offer URL could be accessed multiple times to obtain the same pre-authorized code multiple times.

Closes #44660


tdiesler (changes requested):
A `CredentialOfferState` is put into storage like this ...

```
        String entryJson = JsonSerialization.valueAsString(entry);
        session.singleUseObjects().put(entry.getNonce(), lifespanSeconds, Map.of(ENTRY_KEY, entryJson));
        entry.getPreAuthorizedCode().ifPresent(it -> {
            session.singleUseObjects().put(it, lifespanSeconds, Map.of(ENTRY_KEY, entryJson));
        });
        Optional.ofNullable(entry.getAuthorizationDetails()).ifPresent(it -> {
            it.getCredentialIdentifiers().forEach( cid -> {
                session.singleUseObjects().put(cid, lifespanSeconds, Map.of(ENTRY_KEY, entryJson));
            });
        });

```

Hence, the same state can be discovered by 

* pre-auth code
* each `credential_identifier`

When an offer expires or when it gets consumed it should be removed from storage (i.e. to prevent memory leak)

The use of a `consumed` flag is questionable because it somehow suggests that it can still be read after the offer has been consumed. In fact, it seems that `removeOfferState` is never called (any more) - this would likely be incorrect.

How about removing the offer from storage when it has successfully been consumed or when it has expired? In which case we would loose the ability to distinguish between ...

* already expired
* already consumed

It would simply be "not found", which I think is ok. If the offer is not there (anymore) we can't say why - unless we maintain state about the evicted offer, which in turn creates an eviction challenge related to that state.

Having said that, we could probably resurrect "already expired" (at least for pre-auth offers) when the code becomes JWT with a corresponding `exp` attribute.

mposolda:
> The use of a `consumed` flag is questionable because it somehow suggests that it can still be read after the offer has been consumed. In fact, it seems that `removeOfferState` is never called (any more) - this would likely be incorrect.

+1 . Can we use `removeOfferState` instead of "consumed" flag or is it any corner-case where it will not work as expected?

FYI. Created https://github.com/keycloak/keycloak/issues/45931 as a related issue (related to the comment made by @tdiesler [here](https://github.com/keycloak/keycloak/issues/44660#issuecomment-3823185687) )

Awambeng:
> When an offer expires or when it gets consumed it should be removed from storage (i.e. to prevent memory leak)
> 
> The use of a `consumed` flag is questionable because it somehow suggests that it can still be read after the offer has been consumed. In fact, it seems that `removeOfferState` is never called (any more) - this would likely be incorrect.
> 
> How about removing the offer from storage when it has successfully been consumed or when it has expired? In which case we would loose the ability to distinguish between ...
> 
> * already expired
> * already consumed
> 
> It would simply be "not found", which I think is ok. If the offer is not there (anymore) we can't say why - unless we maintain state about the evicted offer, which in turn creates an eviction challenge related to that state.
> 
Thank you @tdiesler for the feedback. I agree that `removeOfferState()` is necessary to prevent memory leaks and that the current `consumed` flag is questionable. However, calling `removeOfferState()` as soon as the offer URL is accessed introduces a timing issue in the flow.

The pre-authorized code must remain valid after the credential offer URL is accessed. The flow consists of two phases:

1. **Offer URL access (via nonce)** → returns the offer JSON containing the pre-authorized code  
2. **Token request (via pre-authorized code)** → looks up the offer state using that code

If `removeOfferState()` is called during the first phase, this removes all associated entries (nonce, pre-authorized code, and credential identifiers), which would break the second phase when the token request is made.

Instead of removing all entries at offer URL access, we could:

1. **Remove only the nonce entry** when the offer URL is accessed using  
   `singleUseObjects().remove(nonce)`  
   - Provides atomic, single-use replay protection  
   - Keeps the pre-authorized code available for the token exchange  
   - Prevents memory leaks for the nonce entry

2. **Remove the remaining entries after token exchange (optional):**  
   - After a successful token exchange, call `removeOfferState()` to clean up the pre-authorized code and credential identifier entries  
   - Fully prevents memory leaks

With this approach, the `consumed` flag becomes unnecessary while preserving the intended behavior.

@tdiesler what do you think?

Regarding the storage optimization mentioned in #45931, I suggest handling that in a separate PR, depending on the direction we take here.  
@mposolda would that be acceptable?

tdiesler:
> > When an offer expires or when it gets consumed it should be removed from storage (i.e. to prevent memory leak)
> > The use of a `consumed` flag is questionable because it somehow suggests that it can still be read after the offer has been consumed. In fact, it seems that `removeOfferState` is never called (any more) - this would likely be incorrect.
> > How about removing the offer from storage when it has successfully been consumed or when it has expired? In which case we would loose the ability to distinguish between ...
> > 
> > * already expired
> > * already consumed
> > 
> > It would simply be "not found", which I think is ok. If the offer is not there (anymore) we can't say why - unless we maintain state about the evicted offer, which in turn creates an eviction challenge related to that state.
> 
> Thank you @tdiesler for the feedback. I agree that `removeOfferState()` is necessary to prevent memory leaks and that the current `consumed` flag is questionable. However, calling `removeOfferState()` as soon as the offer URL is accessed introduces a timing issue in the flow.
> 
> The pre-authorized code must remain valid after the credential offer URL is accessed. The flow consists of two phases:
> 
> 1. **Offer URL access (via nonce)** → returns the offer JSON containing the pre-authorized code
> 2. **Token request (via pre-authorized code)** → looks up the offer state using that code
> 
> If `removeOfferState()` is called during the first phase, this removes all associated entries (nonce, pre-authorized code, and credential identifiers), which would break the second phase when the token request is made.
> 
> Instead of removing all entries at offer URL access, we could:
> 
> 1. **Remove only the nonce entry** when the offer URL is accessed using
>    `singleUseObjects().remove(nonce)`
>    
>    * Provides atomic, single-use replay protection
>    * Keeps the pre-authorized code available for the token exchange
>    * Prevents memory leaks for the nonce entry
> 2. **Remove the remaining entries after token exchange (optional):**
>    
>    * After a successful token exchange, call `removeOfferState()` to clean up the pre-authorized code and credential identifier entries
>    * Fully prevents memory leaks
> 
> With this approach, the `consumed` flag becomes unnecessary while preserving the intended behavior.
> 
> @tdiesler what do you think?
> 
> Regarding the storage optimization mentioned in #45931, I suggest handling that in a separate PR, depending on the direction we take here. @mposolda would that be acceptable?

Yes, we need to retain the `CredentialOfferState` for as long as it's needed. Removal should happen on ...

* Credential Request
* CredentialOffer expired

IIRC, the AccessToken request is too early if we want to verify whether a requested `credential_identifier` was actually offered.

I guess in the end it boils down to what we can do with `KeycloakSession.singleUseObjects()`. Is it possible to maintain a set of "active" credential offers, that we can search for a uri-nonce, pre-auth code or other properties? Registering the same `CredentialOfferState` under different keys was but the first attempt of storing these things - maybe there is a better way.

Awambeng:
@tdiesler I've implemented the changes based on your feedback:

- Removed the `consumed` flag entirely.
- For replay protection, the credential offer endpoint now removes only the nonce entry using `session.singleUseObjects().remove(nonce)`, keeping the pre-authorized code and credential identifier entries available for token and credential requests.
- After successful credential issuance, `issueCredential()` calls `CredentialOfferStorage.removeOfferState(...)` to clean up the remaining entries.

The existing integration tests for replay protection and the pre-authorized code flow have been updated to reflect the new behavior.

Regarding #45931: given that `singleUseObjects()` does not support searching and we now explicitly clean up all registered keys, my impression is that the storage optimization may no longer be strictly necessary, but I'm happy to align with your preference there.

tdiesler:
> Regarding #45931: given that `singleUseObjects()` does not support searching and we now explicitly clean up all registered keys, my impression is that the storage optimization may no longer be strictly necessary, but I'm happy to align with your preference there.

`CredentialOfferState` storage is already well encapsulated in `CredentialOfferStorage`. Other APIs do have to juggle SingleUseObjects for CRUD and find operations. In future we may want to store offers in DB - for now SingleUseObjects is fine as long as `CredentialOfferStorage` is the only place where access to `CredentialOfferState` happens.

One more thing ... is `CredentialOfferState` expiry already taken care of? If not could you perhaps create an issue about that?

Thanks & well done :-)

Awambeng:
> One more thing ... is `CredentialOfferState` expiry already taken care of? If not could you perhaps create an issue about that?
> 
@tdiesler CredentialOfferState expiry is wired end-to-end:

When the state is created, we set an absolute expiration timestamp (pre-authorized offers use `currentTimeSeconds + preAuthorizedCodeLifeSpan`,  auth-code offers use `currentTimeSeconds + 3600`).

`DefaultCredentialOfferStorage` derives the TTL from this timestamp and passes it to `singleUseObjects().put(...)`, so expired offers are automatically evicted by the underlying cache.

There is also an integration test covering this behavior (`OID4VCCredentialOfferCorsTest.testCredentialOfferWithExpiredNonce`).

mposolda:
@Awambeng @tdiesler Thanks for the update and discussion! I am approving and hope we can address remaining things in a follow-up.

Few additional points:

> Regarding the storage optimization mentioned in https://github.com/keycloak/keycloak/issues/45931, I suggest handling that in a separate PR, depending on the direction we take here.
> @mposolda would that be acceptable?

Yes, I think this works. I've added "Additional note" to the description of https://github.com/keycloak/keycloak/issues/45931 . We might be able to have single entry in the cache and possibly update it once needed (EG. add some flag to the entry when the "nonce" was consumed, so that it is not possible to call "nonce" endpoint another time). 

> I guess in the end it boils down to what we can do with KeycloakSession.singleUseObjects(). Is it possible to maintain a set of "active" credential offers, that we can search for a uri-nonce, pre-auth code or other properties? Registering the same CredentialOfferState under different keys was but the first attempt of storing these things - maybe there is a better way.

Yeah, I hope there is slightly better way and we can use single entry (might help to avoid amount of writes to the cache, also helps with the size of the cache etc as both could be issues in the environments with large amount of cache entries). Hope it could be done in a follow-up #45931 .

> One more thing ... is CredentialOfferState expiry already taken care of? If not could you perhaps create an issue about that?

Not 100% sure what you mean by `CredentialOfferState` expiry, but "singleUse()" is implemented by infinispan under the covers and it should take care of the expiration of evicted entries automatically. In other words, when you call for example something like:
```
session.singleUseObjects().put(it, 120, Map.of(ENTRY_KEY, entryJson));
```
the entry should be automatically removed from infinispan cache in 120 seconds. In other words, no need to care explicitly about manual expiration of entries etc as this should be already taken care of.

mposolda (approved):

Awambeng:
> Yes, I think this works. I've added "Additional note" to the description of #45931 . We might be able to have single entry in the cache and possibly update it once needed (EG. add some flag to the entry when the "nonce" was consumed, so that it is not possible to call "nonce" endpoint another time).
> 
@mposolda Thanks for the clarification and for updating #45931.

>Not 100% sure what you mean by CredentialOfferState expiry, but "singleUse()" is implemented by infinispan under the covers and it should take care of the expiration of evicted entries automatically. In other words, when you call for example something like:
>
`session.singleUseObjects().put(it, 120, Map.of(ENTRY_KEY, entryJson));`
>
>the entry should be automatically removed from infinispan cache in 120 seconds. In other words, no need to care explicitly about manual expiration of entries etc as this should be already taken care of.
>
If i'm not wrong `CredentialOfferState` relies on a TTL-based eviction which passes it to `singleUseObjects()` for expiry cases

PR #45562, opened 2026-01-19
Organization Groups Core Backend & API

### LINKED ISSUE #45505: Organization Groups Core Backend & API
### Description

Implement the baseline functionality for Organization Groups based on the POC. This involves the database schema changes, the internal group structure, and the REST API endpoints.

**Scope & Notes**

- **Internal Structure**: Use the "Internal Group" pattern where the organization member is explicitly a member of the internal root group.
- **Membership Model**: Implement "Dual Explicit Membership". A user is a member of the internal org group and optionally a member of the specific subgroup.
- **Isolation**: Ensure Realm groups and Org groups cannot be mixed in the hierarchy. Org groups cannot be default groups.
- **Database**: GroupEntity has a relationship to OrganizationEntity.
- **API**: Implement CRUD operations under the organization path context.

vramik:
Closes #45505

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/GroupAdapter.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/GroupAdapter.java
@@ -27,6 +27,7 @@ import java.util.stream.Stream;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.GroupModel;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.OrganizationModel;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.RoleModel;
 import org.keycloak.models.cache.infinispan.entities.CachedGroup;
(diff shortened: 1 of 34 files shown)


model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/GroupAdapter.java
```diff
@@ -309,4 +310,11 @@ public Type getType() {
         if (isUpdated()) return updated.getType();
         return cached.getType();
     }
+
+    @Override
+    public OrganizationModel getOrganization() {
```
pedroigor: I would add a `OrganizationGroup` interface with this method instead. And make sure org groups instances are returned as this type.

tests/base/src/test/java/org/keycloak/tests/admin/group/GroupSearchTest.java
```diff
@@ -108,7 +108,7 @@ public void querySubGroups() {
         }
 
         // search for subgroups filtering by name - all groups with 'gro' in the name.
-        subGroups = parentGroupResource.getSubGroups("gro", false, 0, 10, true);
+        subGroups = parentGroupResource.getSubGroups("*gro", false, 0, 10, true);
```
pedroigor: Are we changing behavior when searching subgroups by name?

pedroigor (approved):

PR #45575, opened 2026-01-19
Additional restrictions when to issue a redirect to the caller on rolling updates

### LINKED ISSUE #45574: Hardening of redirect of obsolete version resource URLs
### Description

When creating a redirect on rollingUpdates:v2, verify that we never redirect to a not trusted URL. 

With that, we shouldn't provide any query parameters where not necessary, and rely on parsed values instead of values passed in as parameters. We can also check if the file that is about to be fetched actually exists.

ahus1:
Closes #45574

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/ThemeResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/ThemeResource.java
@@ -116,29 +116,43 @@ public class ThemeResource {
                     return Response.status(Response.Status.NOT_FOUND).build();
                 }
 
-                if (!version.equals(Version.RESOURCES_VERSION) && !hasContentHash) {
+                // Only enter here if the requested version is different, doesn't have a content hash in the URL,
+                // and we didn't default to the default theme as the theme is unknown.
+                if (!version.equals(Version.RESOURCES_VERSION) && !hasContentHash && Objects.equals(theme.getName(), themeName)) {
                     // If it is not the right version, and it does not have a content hash, redirect.
                     // If it is not the right version, but it has a content hash, continue to see if it exists.
 
                     // A simpler way to check for encoded URL characters would be to retrieve the raw values.
                     // Unfortunately, RESTEasy doesn't support this, and UrlInfo will throw an IllegalArgumentException.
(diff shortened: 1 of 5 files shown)


ahus1:
@pruivo - this is the hardening of the redirects that I had in mind. Please have a look, and we can discuss this in more depth tomorrow. Thanks!

services/src/main/java/org/keycloak/services/resources/ThemeResource.java
```diff
@@ -116,30 +116,44 @@ public Response getResource(@PathParam("version") String version, @PathParam("th
...
+                            .path("/{version}/{themeType}/{themeName}/{path}").build(version, theme.getType().toString().toLowerCase(), theme.getName(), path).getPath())) {
                         // This prevents half-open redirects
                         log.debugf("No URL encoding should be necessary for the path, returning a 404: %s", uriInfo.getRequestUri().getPath());
                         return Response.status(Response.Status.NOT_FOUND).build();
                     }
 
+                    boolean resourceExists;
+                    try (InputStream is = theme.getResourceAsStream(path)) {
+                        resourceExists = is != null;
+                    }
```
pruivo: I would create a method in `Theme`, `boolean hasResource(String path)`, using this logic as a default implementation. 
Both `ClassLoaderTheme` and `FolderTheme` can check if the file exists without having to open and close the file for reading.

ahus1: @pruivo - I've addressed your comments. Please re-review. Thanks!

services/src/main/java/org/keycloak/theme/FolderTheme.java
```diff
@@ -88,6 +88,11 @@ public URL getTemplate(String name) throws IOException {
         return file.isFile() ? file.toURI().toURL() : null;
     }
 
+    @Override
+    public boolean hasResource(String path) throws IOException {
+        return ResourceLoader.getFile(resourcesDir, path) != null;
```
pruivo:
```suggestion
        var file = ResourceLoader.getFile(resourcesDir, path);
        return file != null && file.isFile();
```

pruivo (changes requested):

pruivo (approved):

pedroigor (approved):

PR #45600, opened 2026-01-20
Upgrade command rolling updates for patch releases / step 3: Infinispan/JGroups.

### LINKED ISSUE #38884: Upgrade command rolling updates for patch releases / step 3: Infinispan/JGroups support
### Description

There is already the rolling update functionality for the Operator in KC 26.2 as built in #36305 

once the smoke tests are in place for ISPN/JGroups, do the following

### Motivation

The Operator should handle rolling updates for patch releases of Infinispan.

This is pending the upgrade to ISPN 16 which is planned for KC 26.6

### Details

* Allow for rolling updates for patch releases for JGroups / ISPN patch releases
* Future ISPN updates would require a manual review of changes and might be added to a deny-list when there are breaking changes for example due to bug fixes

ruchikajha95:
Changes for Upgrade command rolling updates for patch releases / step 3: Infinispan/JGroups.

Closes #38884

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
+++ b/model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
@@ -4,6 +4,7 @@ import java.util.Map;
 import java.util.stream.Stream;
 
 import org.keycloak.Config;
+import org.keycloak.common.Profile;
 import org.keycloak.compatibility.AbstractCompatibilityMetadataProvider;
 import org.keycloak.infinispan.util.InfinispanUtils;
 import org.keycloak.spi.infinispan.CacheEmbeddedConfigProviderSpi;
@@ -24,14 +25,28 @@ public class CachingEmbeddedMetadataProvider extends AbstractCompatibilityMetada
 
     @Override
     public Map<String, String> customMeta() {
+        String rawIspnVersion = Version.getVersion();
+        String rawJgroupsVersion = org.jgroups.Version.printVersion();
+        if (Profile.isAnyVersionOfFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
+            rawIspnVersion = majorMinorOf(rawIspnVersion);
+            rawJgroupsVersion = majorMinorOf(rawJgroupsVersion);
+        }
         return Map.of(
-              "version", Version.getVersion(),
-              "jgroupsVersion", org.jgroups.Version.printVersion()
+                "version", rawIspnVersion,
+                "jgroupsVersion", rawJgroupsVersion
         );
     }
-
     @Override
     public Stream<String> configKeys() {
         return Stream.of(DefaultCacheEmbeddedConfigProviderFactory.CONFIG, DefaultCacheEmbeddedConfigProviderFactory.STACK);
+    }
+
+    private String majorMinorOf(String version) {
+        if (version == null || version.isEmpty()) {
+            return version;
+        }
+        // Pattern to grab only the "Major.Minor" (e.g., 16.0)
+        java.util.regex.Matcher matcher = java.util.regex.Pattern.compile("^(\\d+\\.\\d+)").matcher(version);
+        return matcher.find() ? matcher.group(1) : version;
     }
 }
--- a/model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
+++ b/model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
@@ -1,6 +1,8 @@
 package org.keycloak.infinispan.compatibility;
 
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import org.keycloak.Config;
@@ -27,7 +29,7 @@ public class CachingEmbeddedMetadataProvider extends AbstractCompatibilityMetada
     public Map<String, String> customMeta() {
         String rawIspnVersion = Version.getVersion();
         String rawJgroupsVersion = org.jgroups.Version.printVersion();
-        if (Profile.isAnyVersionOfFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
+        if (Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
             rawIspnVersion = majorMinorOf(rawIspnVersion);
             rawJgroupsVersion = majorMinorOf(rawJgroupsVersion);
         }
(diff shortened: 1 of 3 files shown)


model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
```diff
@@ -24,14 +25,28 @@ protected boolean isEnabled(Config.Scope scope) {
...
+                "jgroupsVersion", rawJgroupsVersion
         );
     }
-
     @Override
     public Stream<String> configKeys() {
         return Stream.of(DefaultCacheEmbeddedConfigProviderFactory.CONFIG, DefaultCacheEmbeddedConfigProviderFactory.STACK);
     }
+
+    private String majorMinorOf(String version) {
```
pruivo:
```suggestion
    private static String majorMinorOf(String version) {
```

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
```diff
@@ -24,14 +25,28 @@ protected boolean isEnabled(Config.Scope scope) {
...
+        if (version == null || version.isEmpty()) {
+            return version;
+        }
+        // Pattern to grab only the "Major.Minor" (e.g., 16.0)
+        java.util.regex.Matcher matcher = java.util.regex.Pattern.compile("^(\\d+\\.\\d+)").matcher(version);
```
pruivo: Replace with imports
```suggestion
        Matcher matcher = Pattern.compile("^(\\d+\\.\\d+)").matcher(version);
```

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
```diff
@@ -24,14 +25,28 @@ protected boolean isEnabled(Config.Scope scope) {
 
     @Override
     public Map<String, String> customMeta() {
+        String rawIspnVersion = Version.getVersion();
+        String rawJgroupsVersion = org.jgroups.Version.printVersion();
+        if (Profile.isAnyVersionOfFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
```
pruivo: Wrong method used; we do not want "any version of the feature" :) 
```suggestion
        if (Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
```

ahus1:
@ruchikajha95 - I see that the test fails, please have a look. 

The PR also needs a docs change. The following sentence is no longer correct, as we now allow differences in the patch versions of Infinispan and JGroups. 
As we don't plan to change the minor version of Infinispan and JGroups in patch releases, please discuss with @pruivo if this sentence can be removed completely. 

https://github.com/keycloak/keycloak/blob/3c3915556c30adbd778331b50280e43f95ded034/docs/guides/server/update-compatibility.adoc?plain=1#L228

pruivo:
@ahus1 @ruchikajha95 yes, it can be removed.

ruchikajha95:
@ahus1 @pruivo  Thanks for the review . i have fixed all the comments and all the test failures as well. Please let me know if any further changes required.

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
```diff
@@ -24,14 +27,28 @@ protected boolean isEnabled(Config.Scope scope) {
 
     @Override
     public Map<String, String> customMeta() {
+        String rawIspnVersion = Version.getVersion();
+        String rawJgroupsVersion = org.jgroups.Version.printVersion();
+        if (Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
+            rawIspnVersion = majorMinorOf(rawIspnVersion);
+            rawJgroupsVersion = majorMinorOf(rawJgroupsVersion);
+        }
```
ahus1, line 35: With this behavior, if you disable the ROLLING_UPDATES_V2 feature for your setup, you'll see a non-rolling update as the patch information is added to JSON. 

There is an alternative approach where one wouldn't store different data into the JSON, but instead have a different mode of comparison. For that, there is `isCompatible()`.

So if the `ROLLING_UPDATES_V2` is enabled, I would rather suggest to store the same information as before, but tweak the comparison to no longer compare the strings as is, but only the major/minor versions.

ahus1 (changes requested):
Thank you for this PR. While this would work, see below for a suggestion on moving the changes to the comparison and not the data stored in the JSON. 

This could provide a slightly better user experience. Storing the full detail version could also help us to have a better exception handling in case we would have breaking changes in a patch release.

model/infinispan/src/main/java/org/keycloak/infinispan/compatibility/CachingEmbeddedMetadataProvider.java
pruivo: -1 for adding complexity.
Why would a user disable `ROLLING_UPDATES_V2`? This should be an essential feature, to be honest.

ahus1: >  Why would a user disable ROLLING_UPDATES_V2? 

Maybe because we broke something? Sure, we make it an essential feature soon. And we'll have it enabled by default. 

How do you think about "Storing the full detail version could also help us to have a better exception handling in case we would have breaking changes in a patch release."?

pruivo: If we break something, we have to blacklist/denylist the release in one of the `CompatibilityMetadataProvider` implementations.
You won't tell the users "hey, when creating your optimized image for version x.y.z, please disable the feature `ROLLING_UPDATES_V2`". It seems very painful and complex. 
And the operator has manual control over the upgrade process in the worst-case scenario.

> How do you think about "Storing the full detail version could also help us to have a better exception handling in case we would have breaking changes in a patch release."?

I would say not useful, as you can add your custom messages.
```java
    @Override
    public CompatibilityResult isCompatible(Map<String, String> other) {
        if (kcVersionIsBroken) {
            return new CompatibilityResult() {
                @Override
                public int exitCode() {
                    return ExitCode.RECREATE.value();
                }

                @Override
                public Optional<String> errorMessage() {
                    return Optional.of("We break things because of X, Y and Z. See Git Hub Issue #0000");
                }
            };
        }
        return super.isCompatible(other);
    }
```

ahus1 (approved):
Thank you for the explanation!

pruivo (approved):

Issue #40125, opened 2025-06-02
Stop using jboss-parent


stianst:
### Description

Keycloak should not depend on `jboss-parent`

PR #45616, opened 2026-01-20
Bump jboss-parent version to 49

### LINKED ISSUE #45615: Bump jboss-parent version to 49
### Description

As we don't override the Maven Assembly Plugin version in our parent POM, the effective version is inherited from the JBoss parent, which currently pins it to 3.1.1.

Version 3.1.1 (released January 2019) has suboptimal performance on Windows.

Upgrading org.jboss:jboss-parent from 39 to **49** will make the effective maven-assembly-plugin 3.7.1. This results to ~70% faster distribution build in some cases.

Pepo48:
Closes: #45615

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.jboss</groupId>
         <artifactId>jboss-parent</artifactId>
-        <version>39</version>
+        <version>49</version>
     </parent>
 
     <name>Keycloak</name>


Pepo48:
@ASzc I wanted to ask you for your opinion - is this fine from your/continuous delivery perspective?

Here's the pom: https://repo1.maven.org/maven2/org/jboss/jboss-parent/49/jboss-parent-49.pom

Thanks!

ASzc:
A productized version is available. I don't foresee any issues right now, but I will monitor the product's nightly build after this merges. If anything breaks then we can follow up

stianst:
Need to check what has changed in jboss-parent as it can bring problems with the Keycloak release, which it has in the past. In general the plan is to get [rid of jboss-parent](https://github.com/keycloak/keycloak/issues/40125) so we can control things ourselves, as it does bring a bit of oddities that we don't really want.

Pepo48:
@stianst just yesterday, when we were discussing this in the meeting, we talked about getting rid of `jboss-parent`. We were wondering whether [`quarkus-platform-bom`](https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus/bom/pom.xml) could serve as a substitute, but for example the `maven-assembly-plugin` (which is the primary reason for this PR) is not part of it. They just [consume](https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L5) `jboss-parent` the same way as we do.

As an alternative, I can simply override the `maven-assembly-plugin` version without touching the `jboss-parent` version. 

In any case, this is not urgent by any means.

stianst:
> @stianst just yesterday, when we were discussing this in the meeting, we talked about getting rid of `jboss-parent`. We were wondering whether [`quarkus-platform-bom`](https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus/bom/pom.xml) could serve as a substitute, but for example the `maven-assembly-plugin` (which is the primary reason for this PR) is not part of it. They just [consume](https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L5) `jboss-parent` the same way as we do.
> 
> As an alternative, I can simply override the `maven-assembly-plugin` version without touching the `jboss-parent` version.
> 
> In any case, this is not urgent by any means.

quarkus-bom can serve for dependencies, but not as a parent.

Quarkus has their [own parent](https://github.com/quarkusio/quarkus/blob/main/independent-projects/parent/pom.xml) and doesn't look like it's inheriting jboss-parent?

stianst (changes requested):
Don't merge as it breaks the release. Had a look at the diff from 39 to 49 and there is a lot that is updated.

Pepo48:
> Quarkus has their [own parent](https://github.com/quarkusio/quarkus/blob/main/independent-projects/parent/pom.xml) and doesn't look like it's inheriting jboss-parent?

You’re correct, I linked their Quarkus Platform Configuration POM, not their parent POM, where they don’t explicitly set a parent.

Earlier I also did some more investigation. I was wondering, what is the minimum we need to do in case we would like to solve this by removing jboss-parent. For now, the scope was to stabilize the project build and GH CI (while preserving the Windows build time improvement).

In terms of the maven-assembly-plugin, we don't need to do much. The version can be automatically determined by Maven from its [super POM](https://maven.apache.org/ref/3.9.8/maven-model-builder/super-pom.html) (3.9.8, as used by our Maven wrapper). This is the same approach that Quarkus uses, except their docs module, where they specifically set the version.

As for the project build, the only thing that seems to be needed is to set `version.enforcer.plugin` in the parent pom. I tested it via `mvn validate` with profiles specified.

Then to stabilize new Keycloak base testsuite I added to the parent pom:
```
<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
```

The last thing was fixing some UT/IT tests - all of the failures seems to be related to maven-surefire-plugin configuration in jboss-parent-39:
```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-surefire-plugin</artifactId>
  <version>${version.surefire.plugin}</version>
  <configuration>
    <trimStackTrace>false</trimStackTrace>
    <systemPropertyVariables>
      <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
    </systemPropertyVariables>
  </configuration>
</plugin>
```
which sets `java.io.tmpdir` inside the project. Without this being set some of the relative paths of the resources broke. I tried to fix it per test level (the code may be dirty, I just wanted to test the feasibility).

Only the Keycloak CI was affected, other workflows were green. Also, there might be some issues beyond my initial scope, so take the analysis with a grain of salt.

Here are the Keycloak CI results: https://github.com/Pepo48/keycloak/actions/runs/21231356205 (a test branch in my fork, for now). 
Here is the commit with above described changes:  https://github.com/keycloak/keycloak/compare/main...Pepo48:keycloak:test

Is this the intended direction?

vmuzikar:
> Is this the intended direction?

I think it looks good. @stianst WDYT?

stianst:
Since updating jboss-parent breaks things; and contains a rather large number of changes. I'd rather drop this PR and aim to remove jboss-parent dependency completely.

It's not just local builds and GA testing that needs to be considered, we also rely on some stuff from jboss-parent for the releases (how Maven repository is set, gpg keys, apidocs, etc.). Honestly, jboss-parent has been a pain from that perspective and we've had to do some hacks/workarounds due to it.

For the release aspect first verify with `-Pdistribution`, if that works then you can use the release testing organization to do some full releases and nightly releases to make sure that stuff works. We will probably need to make some tweaks to the release workflows though, as for one we'd want to get rid of `-Pjboss-release`. Let me know and I can get you access to trying out the release part.

Pepo48:
Closing in favour of #40125

PR #45618, opened 2026-01-20
use type of resource to create the link

### LINKED ISSUE #45428: Admin UI: Wrong redirect for permissions    accessed via resource details
### Before reporting an issue

### Area

admin/ui

## Description

   When clicking on a permission from the **Resources** tab in the client authorization panel, the navigation always
   treats it as a \"resource\" type permission, even if it's a \"scope\" type permission.

   ### Video

https://github.com/user-attachments/assets/0d4d9225-1922-4a7d-9d31-b7178019a2bc

### Version

latest

### Regression

## Expected Behavior

   The link should navigate to the correct permission details page based on the actual permission type (\`resource\`
   or \`scope\`).

### Actual behavior

   All permission links navigate to \`/authorization/permission/resource/{id}\` regardless of the actual permission
   type, causing scope-based permissions to fail to load correctly.

### How to Reproduce?

   1. Go to a client with authorization enabled
   2. Navigate to the **Authorization** tab → **Resources** sub-tab
   3. Expand a resource to see its associated permissions
   4. Click on a scope-based permission link

edewit:
instead of hard coded "resource" we should use the type of permission

fixes: #45428

### DIFF
--- a/js/apps/admin-ui/src/clients/authorization/DetailCell.tsx
+++ b/js/apps/admin-ui/src/clients/authorization/DetailCell.tsx
@@ -70,7 +70,7 @@ export const DetailCell = ({ id, clientId, uris }: DetailCellProps) => {
             id: clientId,
             realm,
             permissionId: permission.id!,
-            permissionType: "resource",
+            permissionType: permission.type,
           })
         }
       />
--- a/js/libs/keycloak-admin-client/src/defs/resourceServerRepresentation.ts
+++ b/js/libs/keycloak-admin-client/src/defs/resourceServerRepresentation.ts
@@ -9,6 +9,7 @@ export default interface ResourceServerRepresentation {
   id?: string;
   clientId?: string;
   name?: string;
+  type?: string;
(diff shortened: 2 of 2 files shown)


pedroigor:
Thanks, @edewit. Failures seem related?

muhammedogz:
Hello @edewit, Thanks for working on this solution, I appreciate the approach.

I had also opened a [PR](https://github.com/keycloak/keycloak/pull/45429) targeting the same behavior, but with a slightly different direction. In my version, I tried to reuse the existing type, whereas this approach introduces an additional field. Both seem viable, so I'm happy to adapt my changes if we want to converge on a single solution.

As a small type-safety improvement, we could also consider using `type?: "resource" | "scope"` instead of a plain `string`.

edewit:
closed in favor of #45429

edewit:
@muhammedogz closed this one in favor of your PR, sorry didn't notice yours I think it's fine like you have it. We could make the types nicer, but atm they are generated from the java model classes and in the future they will contain more information

PR #45622, opened 2026-01-20
[OID4VCI] Make authorizationDetails processing more generic and not tightly coupled to OID4VCI. Fixes

### LINKED ISSUE #44961: Authorization_details added to token-response even when should not be

mposolda:
closes #44961

PR is a follow-up of the other PR https://github.com/keycloak/keycloak/pull/45206 with additional commit added, which addresses limitations mentioned in [the comment](https://github.com/keycloak/keycloak/pull/45206#issuecomment-3783785117) .

The most changes are coming from the `AuthorizationDetailsProcessor` refactoring and related changes. In nutshell:

- Trying to make `AuthorizationDetailsProcessor` processing more generic and prepare it to be able to handle various types of `authorization_details` parameter. Idea is that `AuthorizationDetailsProcessor` implementation can support processing of specific `type` and hence correct processor should be chosen for the particular `type` of `AuthorizationDetailsProcessor` based on the "type" (not just based on the "priority" of the provider).

- Introduced `AuthorizationDetailsProcessorManager` for dealing with authorization details processing

- Renamed existing class `AuthorizationDetail` for OID4VCI specific request parameter to `OID4VCAuthorizationDetail` and make sure it inherits from already existing class `AuthorizationDetailsJSONRepresentation`, which is generic RAR representation of `authorization_details` .

- I've also made the `AuthorizationDetailsResponse` to inherit from `AuthorizationDetailsJSONRepresentation` to be able to re-use those constructs. Added `AuthorizationDetailsResponse.asSubtype()` to be able to cast it to correct subtype effectively (without a need to do full JSON parsing again, which is not performant)

- Added the `AuthorizationDetails` as 1st class citizen into `AccessToken` and into `AccessTokenResponse` as it is generic construct mentioned in the RAR specification (not specific to OID4VCI).

- When the error is during authorizationRequest or tokenRequest during authorization_details processing, then the error returned is `invalid_authorization_details` as suggested by RAR specification https://datatracker.ietf.org/doc/html/rfc9396#name-token-request . Thanks @thiagogcm for [pointing this ](https://github.com/keycloak/keycloak/pull/45622#issuecomment-3775532942) !
@forkimenjeckayang The OID4VCI specification does not explicitly mention `invalid_authorization_details` in [this section](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-token-error-response)  . However per my understanding, the section of OID4VCI specification just provides clarification for some error scenarios, but does not discourage use of other error codes. It explicitly does not mention invalid format of `authorization_details` parameter in any of those clarification scenarios, which suggests to me that it is appropriate to return `invalid_authorization_details` in those scenarios as well when `authorization_details` parameter is incorrect. WDYT?

### Possible follow-up

The RAR specification does not differ for `authorization_details` type for request and response. However OID4VCI specification differs between them (it introduces `credential_identifiers` as another field, which is supposed to be added to the response). I was thinking about removing `AuthorizationDetailsResponse` class entirely and instead make `AuthorizationDetailsJSONRepresentation` as only generic field for both request and responses. In that case, the inheritance of JAva classes would be possible like `AuthorizationDetailsJSONRepresentation` -> `OID4VCAuthorizationDetail` -> OID4VCAuthorizationDetailsResponse` . In the current PR, I did not make `OID4VCAuthorizationDetailsResponse` to inherit from `OID4VCAuthorizationDetail` as it already inherits from `AuthorizationDetailsResponse` and multiple types not supported for inheritance. But I am thinking about this as a follow-up to avoid making this PR too big (it is already relatively big). WDYT? Would it be good to do those changes?


thiagogcm:
@mposolda I am building a Grant Management extension as per the [Grant Management Spec](https://openid.net/specs/fapi-grant-management.html) and this patch helps a lot.

Just a minor detail I noticed from the initial implementation of the AuthorizationDetailsProcessor:
RFC 9396 specifies the error code as [`invalid_authorization_details`](https://www.rfc-editor.org/rfc/rfc9396#section-5)

mposolda:
> @mposolda I am building a Grant Management extension as per the [Grant Management Spec](https://openid.net/specs/fapi-grant-management.html) and this patch helps a lot.
> 
> Just a minor detail I noticed from the initial implementation of the AuthorizationDetailsProcessor: RFC 9396 specifies the error code as [`invalid_authorization_details`](https://www.rfc-editor.org/rfc/rfc9396#section-5)

@thiagogcm Thanks for the feedback! I've addressed the comment and used `invalid_authorization_details` as the error code. See the PR description for some additional details.

forkimenjeckayang:
Thanks @mposolda for the thorough work on this PR! This is a solid fix that addresses critical issues 
and significantly improves the architecture. I've reviewed carefully and have feedback/clarifications...

- `handleMissingAuthorizationDetails()` in `AuthorizationDetailsProcessorManager.java` still Priority-Based. This method still uses priority-based processor selection unlike `processAuthorizationDetails()` and `processStoredAuthorizationDetails()` which use type-based routing. Please clarify if this refactoring was intentional?
- could you please add a Javadoc clarifying when/how the `registerParser()` should be called ?
- Given that this is a new file now `AuthorizationDetailsResponse.java` in `core/src/main/java/org/keycloak/representations` should the copyright and author tag still be used ?

> * When the error is during authorizationRequest or tokenRequest during authorization_details processing, then the error returned is `invalid_authorization_details` as suggested by RAR specification https://datatracker.ietf.org/doc/html/rfc9396#name-token-request . Thanks @thiagogcm for [pointing this ](https://github.com/keycloak/keycloak/pull/45622#issuecomment-3775532942) !
>   @forkimenjeckayang The OID4VCI specification does not explicitly mention `invalid_authorization_details` in [this section](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-token-error-response)  . However per my understanding, the section of OID4VCI specification just provides clarification for some error scenarios, but does not discourage use of other error codes. It explicitly does not mention invalid format of `authorization_details` parameter in any of those clarification scenarios, which suggests to me that it is appropriate to return `invalid_authorization_details` in those scenarios as well when `authorization_details` parameter is incorrect. WDYT?
> 
Your reasoning is correct. While OID4VCI spec doesn't explicitly mention `invalid_authorization_details`, it only provides "additional clarifications" for RFC 6749 error codes , it doesn't restrict or forbid other error codes. Using `invalid_authorization_details` for authorization_details format/validation errors is  the right choice. It follows the parent spec (RAR) and is more semantic than generic errors.

> ### Possible follow-up
> The RAR specification does not differ for `authorization_details` type for request and response. However OID4VCI specification differs between them (it introduces `credential_identifiers` as another field, which is supposed to be added to the response). I was thinking about removing `AuthorizationDetailsResponse` class entirely and instead make `AuthorizationDetailsJSONRepresentation` as only generic field for both request and responses. In that case, the inheritance of JAva classes would be possible like `AuthorizationDetailsJSONRepresentation` -> `OID4VCAuthorizationDetail` -> OID4VCAuthorizationDetailsResponse`. In the current PR, I did not make`OID4VCAuthorizationDetailsResponse`to inherit from`OID4VCAuthorizationDetail`as it already inherits from`AuthorizationDetailsResponse` and multiple types not supported for inheritance. But I am thinking about this as a follow-up to avoid making this PR too big (it is already relatively big). WDYT? Would it be good to do those changes?

RAR spec Section 4 explicitly states: "This specification does not define extensions  to the authorization response." This means authorization_details responses are  implementation-specific, not generic. So your idea is correct, however, IMO i suggest you address in a follow up PR, this is already substantial with critical fixes.

One last thing, about the flaky tests which were reported by the bot from OID4VCI, I guess that can be addressed seperately?

mposolda:
@forkimenjeckayang Thanks for the feedback! I've addressed your concerns in additional commit. Could you please re-review? Some details below:

> Thanks @mposolda for the thorough work on this PR! This is a solid fix that addresses critical issues and significantly improves the architecture. I've reviewed carefully and have feedback/clarifications...
> 
>     * `handleMissingAuthorizationDetails()` in `AuthorizationDetailsProcessorManager.java` still Priority-Based. This method still uses priority-based processor selection unlike `processAuthorizationDetails()` and `processStoredAuthorizationDetails()` which use type-based routing. Please clarify if this refactoring was intentional?

The `handleMissingAuthorizationDetails` is a bit different than other methods as it does not have any "input" authorizationDetails to process. However it is probably better if all processors have possibility to "handle missing details" and hence I've updated the PR to make sure that manager returns "union" list of all authorizationDetails returned by all processors rather than returning the list just from very 1st processor . WDYT?
 
>     * could you please add a Javadoc clarifying when/how the `registerParser()` should be called ?

Yes, added Javadoc

> 
>     * Given that this is a new file now `AuthorizationDetailsResponse.java` in `core/src/main/java/org/keycloak/representations` should the copyright and author tag still be used ?

From what I know, we do not need anymore to add copyright and author tags, so I stopped to adding them in new files. Feel free to still add them in your PRs if you prefer, but it is not needed.

> 
> > * When the error is during authorizationRequest or tokenRequest during authorization_details processing, then the error returned is `invalid_authorization_details` as suggested by RAR specification https://datatracker.ietf.org/doc/html/rfc9396#name-token-request . Thanks @thiagogcm for [pointing this ](https://github.com/keycloak/keycloak/pull/45622#issuecomment-3775532942) !
> >   @forkimenjeckayang The OID4VCI specification does not explicitly mention `invalid_authorization_details` in [this section](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-token-error-response)  . However per my understanding, the section of OID4VCI specification just provides clarification for some error scenarios, but does not discourage use of other error codes. It explicitly does not mention invalid format of `authorization_details` parameter in any of those clarification scenarios, which suggests to me that it is appropriate to return `invalid_authorization_details` in those scenarios as well when `authorization_details` parameter is incorrect. WDYT?
> 
> Your reasoning is correct. While OID4VCI spec doesn't explicitly mention `invalid_authorization_details`, it only provides "additional clarifications" for RFC 6749 error codes , it doesn't restrict or forbid other error codes. Using `invalid_authorization_details` for authorization_details format/validation errors is the right choice. It follows the parent spec (RAR) and is more semantic than generic errors.
> 
> > ### Possible follow-up
> > The RAR specification does not differ for `authorization_details` type for request and response. However OID4VCI specification differs between them (it introduces `credential_identifiers` as another field, which is supposed to be added to the response). I was thinking about removing `AuthorizationDetailsResponse` class entirely and instead make `AuthorizationDetailsJSONRepresentation` as only generic field for both request and responses. In that case, the inheritance of JAva classes would be possible like `AuthorizationDetailsJSONRepresentation` -> `OID4VCAuthorizationDetail` -> OID4VCAuthorizationDetailsResponse`. In the current PR, I did not make`OID4VCAuthorizationDetailsResponse`to inherit from`OID4VCAuthorizationDetail`as it already inherits from`AuthorizationDetailsResponse` and multiple types not supported for inheritance. But I am thinking about this as a follow-up to avoid making this PR too big (it is already relatively big). WDYT? Would it be good to do those changes?
> 
> RAR spec Section 4 explicitly states: "This specification does not define extensions to the authorization response." This means authorization_details responses are implementation-specific, not generic. So your idea is correct, however, IMO i suggest you address in a follow up PR, this is already substantial with critical fixes.

Thanks, created https://github.com/keycloak/keycloak/issues/45706 as a follow-up. Hope to work on this soon in a follow-up PR, which will probably be much smaller.
 
> One last thing, about the flaky tests which were reported by the bot from OID4VCI, I guess that can be addressed seperately?

I am not 100% sure if flaky OID4VCI tests in FIPS mode were caused by this PR or by previous changes, but I've fixed them in this PR.

forkimenjeckayang (approved):
Thank you @mposolda for clarifying my doubts. It LGTM now.
When this is merged and closed, this https://github.com/keycloak/keycloak/pull/45206 should be closed aswell

mposolda:
@forkimenjeckayang Thanks! Closing https://github.com/keycloak/keycloak/pull/45206 as well. I am planning to send later today PR for https://github.com/keycloak/keycloak/issues/45706 as a follow-up.

mposolda:
@forkimenjeckayang I've sent follow-up PR here: https://github.com/keycloak/keycloak/pull/45742 . Review would be welcome

forkimenjeckayang:
> @forkimenjeckayang I've sent follow-up PR here: #45742 . Review would be welcome
Thank you @mposolda 
I'll definitely take a look and leave my feedback before the end of today

PR #45627, opened 2026-01-21
Add abstract property for themes and do not display base for selection

### LINKED ISSUE #41924: Internal server error after changing Admin UI theme to "base" - An old, persisted problem
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Regarding issue [#18884](https://github.com/keycloak/keycloak/issues/18884), this problem has persisted since 2023.
I'm running Keycloak via Docker (latest version, 26.3.2), and in the Realm Settings → Themes section, when changing the Admin UI theme to "base" via the dropdown menu, the system crashes unexpectedly. The following error message appears:

>  "An internal server error has occurred"

This behavior is consistent and reproducible. Please plan a fix.

Image

### Version

26.3.2

### Regression

### Expected behavior

Disappearing "base" from dropdown or moving to a valid state after changing theme to "base"

### Actual behavior

Application crashes

### How to Reproduce?

Setup the latest version of keycloak via docker, go to "Realm Settings", go to "Themes" tab, for "Admin UI" select the theme "base".

### Anything else?

_No response_

rmartinc:
Closes #41924

The PR adds an `abstract` property  to themes that avoids adding that theme to the server-info response. All the base themes are tagged as abstract now. Two comments:

1. I have not validated the data when the realm is updated/created. It can bring some issues and I prefer this way.
2. The base theme is OK for email and login (in email base and keycloak are the same, and in login base is just the plain html without any styling), only admin and account are really broken. But the PR adds the property to all the base themes (account, admin, login and email). If you think that only admin and account should be tagged as abstract, just let me know.

### DIFF
--- a/docs/guides/ui-customization/themes.adoc
+++ b/docs/guides/ui-customization/themes.adoc
@@ -130,6 +130,7 @@ Theme properties are set in the file `<THEME TYPE>/theme.properties` in the them
 
 * parent - Parent theme to extend
 * import - Import resources from another theme
+* abstract - Boolean that, when present and `true`, marks the theme is only used as base for other extensions and is not displayed to be selected in the admin console on the realm themes page. Examples of this are the `base` themes described on this guide.
 * common - Override the common resource path. The default value is `common/keycloak` when not specified. This value would be used as value of suffix of ${r"`${url.resourcesCommonPath}`"}, which is used typically in freemarker templates (prefix of ${r"`${url.resoucesCommonPath}`"} value is theme root uri).
 * styles - Space-separated list of styles to include
 * locales - Comma-separated list of supported locales
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -50,6 +50,10 @@ To analyze rejected requests in the server log, enable debug logging for `org.ke
 To revert to the previous behavior and to accept matrix parameters set the option `http-accept-non-normalized-paths` to `true`.
 With this configuration, enable and review the HTTP access log to identify problematic requests.
 
+=== The base themes are now abstract
+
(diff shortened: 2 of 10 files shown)


docs/guides/ui-customization/themes.adoc
```diff
@@ -130,6 +130,7 @@ Theme properties are set in the file `<THEME TYPE>/theme.properties` in the them
 
 * parent - Parent theme to extend
 * import - Import resources from another theme
+* abstract - Boolean that, when present and `true`, marks the theme is only used as base for other extensions and is not displayed to be selected in the realm page.
```
mposolda:
```suggestion
* abstract - Boolean that, when present and `true`, marks the theme is only used as base for other extensions and is not displayed to be selected in the admin console on the realm themes page. Example of this are `base` themes described on this page.
```
Just a small suggestion (but don't think that it is a blocker for this PR).

mposolda (commented):
@rmartinc Thanks! Added only minor comment inline (which I don't think it is a blocker for this PR).

Besides that, I also think that it is OK that `login` and `email` base themes are also abstract. Just wondering if we want to document in upgrading guide (maybe in "noticeable changes") just to be on safe side for the case when someone already uses `base` as the login theme for example?

For example I wonder about the corner-case scenario like this:
- Admin already has `base` theme configured as login theme from previous Keycloak version
- Now he migrates to latest version
- Then he wants to update `email` theme (not `login` theme) from the admin console. Will he be able to update just the `email` theme and still keep `login` theme as `base` ? If not (and he is forced to update `login` theme as well) then IMO it would be really good to document this.

rmartinc:
@mposolda I did a quick test, if you have `base` now in any theme in the admin console it appears empty.

Screenshot From 2026-01-21 11-58-56

And you can change it or maintain it with the wrong (not available). You can save the realm with another theme changed and the old value `base` is maintained. But I'm going to add a little upgrading note just in case.

rmartinc:
@mposolda Upgrading note added.

mposolda (approved):
@rmartinc Thanks for the update and checking the scenario!

PR #45630, opened 2026-01-21
Avoiding that Xpath Attribute Importer breaks login process

### LINKED ISSUE #45560: Xpath Attribute Importer breaks login
### Before reporting an issue

### Area

saml

### Describe the bug

XPath Attribute Importer breaks login if XML from provider has an unescaped character.

### Version

26.5.1

### Regression

### Expected behavior

Parts of xml that could not be parsed should be ignored, no matter the reason.

### Actual behavior

If the SAML response  XML has a unescaped character like '&' the login fails and server shows the error bellow:

` Could not parse xml element                                                                                                                                ││ acesso     at org.keycloak.broker.saml.mappers.XPathAttributeMapper.lambda$applyXPath$4(XPathAttributeMapper.java:221)  `

And the login process fail.

### How to Reproduce?

Add a SAML identity provider, and create a Xpath Attribute Importer.
The XML that identity provider responds must have a unescaped special character.
like: <root>test&</root>

### Anything else?

_No response_

daniloferreira:
Closes #45560

### DIFF
--- a/services/src/main/java/org/keycloak/broker/saml/mappers/XPathAttributeMapper.java
+++ b/services/src/main/java/org/keycloak/broker/saml/mappers/XPathAttributeMapper.java
@@ -40,6 +40,7 @@ import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.protocol.saml.mappers.SamlMetadataDescriptorUpdater;
 import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.saml.common.exceptions.ParsingException;
 import org.keycloak.saml.common.util.DocumentUtil;
 import org.keycloak.saml.common.util.StringUtil;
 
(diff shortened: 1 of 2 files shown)


ahus1:
@graziang - as you've triaged the parent issue, can you please have a look? 

I checked that we are already handling some of the exceptions and mapping them to an empty string. 

Can you verify if this changed behavior would require a note in the migration guide, as some might rely on this feature? I didn't find documentation about the mappers, so I updated the help text. Let me know what you think about this approach. Thanks!

graziang:
Thanks @daniloferreira, and sorry for the delay.

@ahus1 I noticed that we already added an exception handling for another exception in the same place last year, I don't think it will cause issues.

Note that the linked issue #45560 is closed as not planned, so it should be reopened before this can move forward.
To continue with this PR we need to rebase with the main branch and check the behavior is still correct.
Are you still interested to continue the work on this PR? If we don't hear back, we can close this for now and it can be reopened later.

ahus1 (changes requested):
@graziang - given all the security discussion we had lately around this: The XPAth exception happens because and admin mistyped something. This might be an OK fallback, but still problematic. 

The ParsingException happens when the XML that is provided is malformed. With that, I'd say it is not safe to continue the login process, as the follow-up logic might assume that the field is filled, and on that either approve or deny the the login. 

With that, I'd reject the PR. The correct way is to fix the XML.

PR #45631, opened 2026-01-21
use helptext field from condition instead of message bundle

edewit:
Also removed some duplictate state

fixes: #45586


ahus1:
@edewit - thank you for this fix! I had a look and I found one place in the dropdown that could benefit from the help texts. Can you please have a look? 

I also tried to remove those extra message texts that were duplicated in the UI message bundle for the provider keys. WDYT?

js/apps/admin-ui/src/realm-settings/NewClientPolicyCondition.tsx
```diff
@@ -202,7 +201,7 @@ export default function NewClientPolicyCondition() {
             fieldId="conditionType"
             labelIcon={
               <HelpItem
-                helpText={t(condition?.helpText || "conditionsHelp")}
+                helpText={condition?.helpText || t("conditionsHelp")}
```
edewit, line 204: but if you change it to this it means it can no longer be translated

edewit:
:+1:

js/apps/admin-ui/src/realm-settings/NewClientPolicyCondition.tsx
ahus1: AFAIK translations work only when the backend would provide a message key here. 

Instead, the backend provides a full text as per convention of the Configured Provider API:

https://github.com/keycloak/keycloak/blob/ca205272ba6360bc808d19b5f8e2af119fa37c5a/services/src/main/java/org/keycloak/services/clientpolicy/condition/GrantTypeConditionFactory.java#L49-L51

When such a text is passed to the `t()` function, it would in the best of worlds do nothing to it, and if it goes sideways do strange things to it. 

I would love to see those being translated, still I think this would be outside of the scope of this PR. 

The placed were I removed the keys, there were similar texts in the Admin UI and the Java backend, which is a duplication that is likely to cause out-of-date information. 

WDYT?

edewit: The key can be the entire message, but yeah

ahus1 (approved):
Thank you for this fix and the cleanup!

PR #45632, opened 2026-01-21
Possible mismatch of charset/collation between columns on mysql/mariadb

### LINKED ISSUE #45597: Possible mismatch of collation between columns on mysql/mariadb
### Before reporting an issue

### Area

organizations

### Describe the bug

Similar to https://github.com/keycloak/keycloak/issues/45239 there is another possible mismatch in collation between `ORG.ID` and `IDENTITY_PROVIDER.ORGANIZATION_ID`.

### Version

26.5.1

### Regression

### Expected behavior

see description

### Actual behavior

see description

### How to Reproduce?

see description

### Anything else?

_No response_

vramik:
Closes #45597


sguilhen (approved):
I think it is ok as long as we're sure the usage of ${database.defaultSchemaName} in the changeset condition doesn't end up generating a different checksum for every schema name.

vramik:
> I think it is ok as long as we're sure the usage of ${database.defaultSchemaName} in the changeset condition doesn't end up generating a different checksum for every schema name.

I've tested `keycloak` and `keycloak2` schema names and in both cases the checksum for the changeset is the same: `9:322cb11fc03181903dcd67a54f8b3cf0`.

pedroigor (approved):

PR #45675, opened 2026-01-22
remove /admin/api from openapi definition

### LINKED ISSUE #45674: make `admin/api` part of the base path
### Description

Make `admin/api` part of the base path so that generated code will not create / methods / functions based on that prefix

### Value Proposition

For instance kiota will create client that will look like:
```java
  client.admin().api().byRealmName("myrealm").clients().byVersion("v2").get()
```
  This changes to:
```java
  client.byRealmName("myrealm").clients().byVersion("v2").get()
```

add:
```yaml
 servers:                                                                                                                                   │
     - url: /admin/api     
```
to the openapi yaml

edewit:
making generated api based on this definition more consise

fixes: #45674

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
@@ -44,12 +44,17 @@ public class OASModelFilter implements OASFilter {
         });
     }
 
+    private static final String ADMIN_API_PREFIX = "/admin/api";
+
     @Override
     public void filterOpenAPI(OpenAPI openAPI) {
-        // Sort Paths
+        // Set the server URL since we're stripping the /admin/api prefix from paths
+        openAPI.setServers(java.util.List.of(OASFactory.createServer().url(ADMIN_API_PREFIX)));
+
+        // Sort Paths and strip /admin/api prefix
         Map<String, PathItem> newPaths = openAPI.getPaths().getPathItems().entrySet().stream()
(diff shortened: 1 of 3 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
vmuzikar, line 1: I wonder why we need to do the changes on the filter level. Isn't the `@OpenAPIDefinition(servers = @Server(url = "/admin/api")` annotation enough?

edewit: no, that's not enough it needs this filter to work

mabartos:
@edewit The CI failures should be related to this PR. Can you check them?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
vmuzikar: The we don't need it in the annotation form and we could just keep this:
```
openAPI.setServers(java.util.List.of(OASFactory.createServer().url(ADMIN_API_PREFIX)));
```
Or am I missing something?

mabartos: @edewit Could you check it please? Thanks!

edewit:
Solved this by wrapping the generated code, no longer need this, sorry

Issue #45662, opened 2026-01-21
Increase in startup memory consumption in post 26.5 versions


rnett:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

dist/quarkus

### Describe the bug

Forked from https://github.com/keycloak/keycloak/issues/45459#issuecomment-3774931316.

We run keycloak in our CI pipelines as a JAR (not in a container) using the Keycloak start script.  We had inadvertently set a max heap of 256m via JAVA_OPTS globally on CI, but keycloak functions fine under this limit on pre-25.6 versions.  After upgrading to 26.5.1, Keycloak started OOMing dugin startup.  This was resolved by increasing the memory limit to 512m, so it was not a significant increase in absolute terms.

PS: I saw mention of #45498, but trying to disable `swagger-ui` via feature toggles did not work and threw errors about not recognizing that feature.  What is the correct way to do this?

### Version

26.5.1

### Regression

- [x] The issue is a regression

### Expected behavior

Keycloak's memory consumption during startup is not significantly increased.

### Actual behavior

Keycloak's memory consumption increased enough we had to bump the max heap from 256m to 512m.

### How to Reproduce?

Run the 26.5.1 JAR via the start script with a max heap of 256m.  We do have some custom plugins so I'm not 100% it will reproduce.

### Anything else?

Our installed plugin string is `agroal, cdi, hibernate-orm, hibernate-validator, jdbc-postgresql, keycloak, narayana-jta, opentelemetry, reactive-routes, rest, rest-jackson, smallrye-context-propagation, smallrye-openapi, swagger-ui, vertx`.

shawkins:
@rnett using the configuration provided on the other issue, I couldn't cause an OOM during augmentation until going down to about 140m. Can you provide show-config output and more details of anything in the provider directory that may be contributing here?

We have seen reports of other augmentation changes, such as #44310, and it will may be difficult for us and the quarkus team to figure out what is going on without a full reproducer.

rnett:
To be honest, considering we use some private/paid plugins, I doubt we'll be able to produce a realistic reproducer. If you've checked memory consumption on your side, which it sounds like you have, I think it's fine to close this issue. 

I'd still like to know how to disable the swagger ui, or if there's anything I need to do there, though 🙂

mabartos:
[...]

PR #45681, opened 2026-01-22
Hibernate Validator is enabled by default when not used

### LINKED ISSUE #45677: Hibernate Validator is enabled by default when not used
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

The Hibernate Validator is enabled by default, even when it's not used currently. The Hibernate Validator is only used for the new Client API v2 right now. Enabling Hibernate Validator is quite resource-consuming for startup due to expensive initialization. That's also a difference between 26.4.x and 26.5.x, so it's considered as regression. 

Startup time:

Image

Startup memory allocations (against 26.4.x) (~10MB with other related procedures):
 
Image

### Version

26.5.1

### Regression

### Expected behavior

Hibernate Validator is not enabled when not needed

### Actual behavior

Hibernate Validator is always enabled

### How to Reproduce?

Start the server and see enabled Quarkus extensions

### Anything else?

_No response_

mabartos:
- Closes #45677
- Do not enable expensive Hibernate Validator when it's not needed
- Extension does not do any execution + no Quarkus extension is listed in the log
- Might contribute to https://github.com/keycloak/keycloak/issues/45662

Screenshot From 2026-01-22 13-10-29

### DIFF
--- a/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
+++ b/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
@@ -131,6 +131,7 @@ import io.quarkus.bootstrap.logging.InitialConfigurator;
 import io.quarkus.datasource.deployment.spi.DevServicesDatasourceResultBuildItem;
 import io.quarkus.datasource.runtime.DataSourcesBuildTimeConfig;
 import io.quarkus.deployment.IsDevelopment;
+import io.quarkus.deployment.IsTest;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.Consume;
@@ -861,7 +862,7 @@ class KeycloakProcessor {
         removeBeans.produce(new BuildTimeConditionBuildItem(disabledBean.asClass(), false));
     }
 
-    @BuildStep
+    @BuildStep(onlyIfNot = IsTest.class) // needed for embedded Keycloak
     void disableHibernateValidatorCustomizer(BuildProducer<BuildTimeConditionBuildItem> removeBeans, CombinedIndexBuildItem index) {
         if (!Profile.isFeatureEnabled(Profile.Feature.CLIENT_ADMIN_API_V2)) {
             // disables the filter
--- a/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
+++ b/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
@@ -100,6 +100,7 @@ import org.keycloak.quarkus.runtime.services.health.KeycloakClusterReadyHealthCh
 import org.keycloak.quarkus.runtime.services.health.KeycloakReadyHealthCheck;
 import org.keycloak.quarkus.runtime.storage.database.jpa.NamedJpaConnectionProviderFactory;
 import org.keycloak.quarkus.runtime.themes.FlatClasspathThemeResourceProviderFactory;
+import org.keycloak.quarkus.runtime.validation.HibernateValidatorFactoryCustomizer;
 import org.keycloak.representations.provider.ScriptProviderDescriptor;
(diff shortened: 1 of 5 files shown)


services/src/main/java/org/keycloak/services/error/KeycloakErrorHandler.java
```diff
@@ -102,8 +101,6 @@ public static Response getResponse(KeycloakSession session, Throwable throwable)
                 error.setErrorDescription("Cannot parse the JSON");
             } else if (isServerError) {
                 error.setErrorDescription("For more on this error consult the server log.");
-            } else if (throwable instanceof ValidationException) {
```
mabartos, line 105: Not needed as we have the ValidationExceptionHandler.

vmuzikar (dismissed):
Looks good, thanks @mabartos.

mabartos:
@vmuzikar Thanks! I've just added one small change to fix Quarkus UT tests (embedded).

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -860,6 +862,16 @@ private static void disableReadyHealthCheck(BuildProducer<BuildTimeConditionBuil
         removeBeans.produce(new BuildTimeConditionBuildItem(disabledBean.asClass(), false));
     }
 
+    @BuildStep(onlyIfNot = IsTest.class) // needed for embedded Keycloak
```
vmuzikar, line 865: @mabartos Could you please share more context why this is needed?

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
```diff
@@ -860,6 +862,16 @@ private static void disableReadyHealthCheck(BuildProducer<BuildTimeConditionBuil
...
+    void disableHibernateValidatorCustomizer(BuildProducer<BuildTimeConditionBuildItem> removeBeans, CombinedIndexBuildItem index) {
+        if (!Profile.isFeatureEnabled(Profile.Feature.CLIENT_ADMIN_API_V2)) {
```
shawkins, line 867: Not specific to this pr, but just wanted to highlight that moving forward this type of check is going to get messy. There will be a set of features and versions where hibernate validation might be used.

shawkins (approved):
LGTM

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
mabartos: @shawkins Once we adapt the Hibernate Validator keycloak-codebase-wide, we remove all these checks and the Hibernate Validator will be enabled by default. Right now, as it's not needed (only for experimental feature), we should not have it enabled by default as it's causing perf regressions for 26.5.x.

quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
mabartos: For the Quarkus embedded server, we don't have a really nice approach on how to disable specific artifacts. At this moment, we just don't care about the ignored artifacts for the embedded server and we just have them enabled all by default (except for explicit FIPS handling). We could improve it for sure, but not sure if it worth the effort. 

That means that the Hibernate validator is enabled by default for embedded server (as other features), and in order to have it initialized, we need the validator factory customizer that do not include the Expression Language (EL) functionality that needs the expressly dependency (that we don't have in project). It means, that for UT tests, we just say that we don't want the EL. 

It's ok, because integration tests are executed with a distribution server - the officially created distribution that excludes the bean for the validator customizer (this step is executed).

So I think we're safe to do this :) 

@vmuzikar Is it ok to you, or should I explain more?

vmuzikar: Thank you, works for me.

vmuzikar (approved):

PR #45686, opened 2026-01-22
Covering hiding username/email when brute force is enabled during identity-first login

### LINKED ISSUE #45685: Covering hiding username/email when brute force is enabled during identity-first login
### Description

It is related to https://github.com/keycloak/keycloak/issues/44892.

The fix for https://github.com/keycloak/keycloak/issues/43091 did not cover the scenario where the user exists, it is not a member of org, and try to authenticate using an email that maps an organization.

We need to introduce coverage for this scenario.

pedroigor:
Closes #45685

### DIFF
--- a/testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/pages/LoginPage.java
+++ b/testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/pages/LoginPage.java
@@ -155,6 +155,10 @@ public class LoginPage extends LanguageComboboxAwarePage {
         return !driver.findElements(By.id("username")).isEmpty();
     }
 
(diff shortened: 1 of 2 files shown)


msdaly200 (approved):
Reviewed test and it is covering the issue where if there is an extra field with username or email and the attempted field is there, the test will fail.  Approving assuming failed checks will pass.

ahus1 (approved):

@pedroigor - Before merging: Can you please double-check that the flaky test that appeared during development of this PR has been addressed? For now I assumed "yes" as there was one more commit after that.

pedroigor:
> Can you please double-check that the flaky test that appeared during development of this PR has been addressed? 

It should. @msdaly200 commented on that too and, as you said, I sent an update to fix the flakiness.

PR #45687, opened 2026-01-22
Feat/45668/Add-Csv-Pdf-Export

### LINKED ISSUE #45668: Add CSV and PDF export to Admin Console list views
### Description

Allow exporting Admin Console list views as CSV or PDF. The export should include all items matching the current filters and sorting (across pages).

### Value Proposition

This request originates from an ongoing public-sector procurement and is a mandatory requirement.

### Goals

Add export functionality to the Admin Console list views, supporting CSV and PDF.
This is especially relevant for Clients, Client Scopes, Realm Roles, Users, Sessions, and Events.

The export must honor the active filters and sorting and include all matching records across pages.

For CSV exports, the delimiter should be configurable (default: semicolon ';'). CSV should include a header row and use UTF-8.

Exports should respect existing view permissions and only include fields the user is allowed to see.

### Non-Goals

No additional export formats (e.g., XLSX, JSON, XML, ODS, HTML) are requested.
No custom report templates or advanced PDF layout; no signing, encryption, or watermarking of PDFs.
No persistence of export presets or user-defined templates.

### Discussion

_No response_

### Notes

The exact control type (two buttons vs. split button vs. overflow menu) is flexible; the mock-up is only meant to clarify the idea.

Image

Goal is discoverability and consistency across list views; any equivalent placement that meets these goals is fine.
Labels/icons are indicative; neutral styling is preferred. Tooltips like “Exports all filtered results across pages” help clarify behavior.

madelyngamble2:
Closes #45668


stianst:
This would require more interest before we consider adding it; we'll leave the issue open to see if it gets any upvotes.

PR #45690, opened 2026-01-22
Optimize user joining group or role

### LINKED ISSUE #45689: When a user joins a role or group, it should not read all existing roles and groups from the database
### Description

The current implementation reads a stream of all entries back from the database, and sometimes even looks them up via the realm and the cache. This is not necessary and should be optimized.

### Value Proposition

Faster and near-constant time when joining a group or role. 

### Goals

Better performance

### Non-Goals

Analyzing similar problems

### Discussion

_No response_

### Notes

This came up when analyzing a different performance problem, so I'll submit my findings as a pull request.

ahus1:
Closes #45689


ahus1:
@keycloak/core-iam - looking for a reviewer on this PR. Thanks!

sguilhen (approved):

stianst (approved):

PR #45691, opened 2026-01-22
fix: aligns our dev http-host default behavior with that of quarkus

### LINKED ISSUE #42876: dev mode should bind only to localhost if possible
### Description

We should follow the default quarkus behavior such that dev mode binds to localhost only when possible.

### Discussion

_No response_

### Motivation

 To prevent inadvertant exposure of dev instances.

### Details

_No response_

shawkins:
closes: #42876

Matches the quarkus behavior for dev mode binding to localhost. To further prevent this from being seen as breaking, it also checks if kc_run_in_container is true - however I wasn't sure if this needs mentioning in the doc notes.


mabartos (commented):
@shawkins In general, LGTM.

But I'm still able to access the dev instance from the other instance.

I started Keycloak in dev mode with your changes. 

Image

And when I access the 192.168.0.181:8080 on my phone, I can see the Keycloak login page. 

Are these Quarkus changes in the release that we currently use, or will it be available after the 3.31.0?

mabartos (changes requested):
The check on the Quarkus side is probably needed.

shawkins:
> The check on the Quarkus side is probably needed.

Yes, I had launched the Keycloak class and saw that it appropriately picked up the localhost default - but we're not when using start-dev from the cli. The likely problem is that for the latter quarkus doesn't think that is actually a test / dev launch.

shawkins:
@mabartos should be better now. Unfortunately it makes this leakier into our code than I would have liked, but it's a much bigger change to have start-dev actually be in the quarkus dev or test launch mode.

vmuzikar:
@Pepo48 Could you please review for the Windows bits?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
```diff
@@ -66,6 +89,13 @@ public List<PropertyMapper<?>> getPropertyMappers() {
                         .build(),
                 fromOption(HttpOptions.HTTP_HOST)
                         .to("quarkus.http.host")
+                        .mapFrom(new OptionBuilder<>("run-in-container", Boolean.class).build(), (v, c) -> {
```
vmuzikar: Can't we simply check for the `KC_RUN_IN_CONTAINER` env var in the transformer?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
```diff
@@ -56,6 +58,27 @@ private static void setCustomExceptionTransformer() {
         });
     }
 
+    // taken from VertxConfigBuilder
+    private static boolean isWSL() {
+        var sysEnv = System.getenv();
+        return sysEnv.containsKey("IS_WSL") || sysEnv.containsKey("WSL_DISTRO_NAME");
+    }
```
vmuzikar, line 64: Why this doesn't work in WSL?

vmuzikar (commented):
@shawkins Thanks for the PR. It'd be good to have more tests for  this, though.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
shawkins: Right, that would be simplier :)

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
shawkins: That's a limitation documented from quarkus - it's probably like starting in a container with localhost only.

vmuzikar (dismissed):
LGTM, thank you.

@Pepo48 Do you want to check this too?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
Pepo48: Afaik, binding to 0.0.0.0 is necessary for WSL networking.

Pepo48: Yes, indeed. See: https://learn.microsoft.com/en-us/windows/wsl/networking#connecting-via-remote-ip-addresses

Pepo48 (approved):
Sorry for the late review. I tested this on Windows (native), worked as expected. Thanks Steve, LGTM.

shawkins:
> Sorry for the late review. I tested this on Windows (native), worked as expected. Thanks Steve, LGTM.

Thank you for checking.

mabartos (approved):
LGTM, thanks!

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
vmuzikar: @shawkins Maybe it'd be good to add that link to a comment here for future references.

shawkins: I assumed it was sufficient to reference VertxConfigBuilder

vmuzikar (approved):

PR #45699, opened 2026-01-22
fix: adding support for xforwarded prefix

### LINKED ISSUE #35298: Reverse proxy provided context path not working despite setting X-Forwarded-Prefix header
### Before reporting an issue

### Area

admin/ui

### Describe the bug

Hello,

Following the deprecation of `hostname:v1`, we are forced to reconfigure how we expose our Keycloak deployments. Due to business requirements, we need to expose the same Keycloak on multiple domains (eg. `https://www.domain.de|it|fr`). 
Therefore, we cannot specify a fixed hostname using the `--hostname` argument but need to resolve it dynamically. 
Moreover, we have to expose the deployment on a specific context path (eg. `/auth`).
This means that Keycloak will be exposed on `https://www.domain.de|it|fr/auth`.
Everything is exposed through an NGINX reverse proxy. With `hostname:v1`, we had a setup that was working using env variables like so:
```
- name: KC_HOSTNAME_STRICT
   value: 'false'
- name: KC_HTTP_ENABLED
   value: 'true'
- name: KC_HOSTNAME_PATH
   value: “/auth/”
- name: KC_HOSTNAME_ADMIN_URL
   value: https://auth-internal.domain.tld/auth
```
Please note that the TLD for the admin URL is different from the one on which we expose the deployment publicly.
The NGINX configuration snippet handling this setup is as follows:
```
http {
          resolver x.x.x.x;
          server {
              listen 443 ssl;
              server_name auth-internal.domain.tld;
              ssl_certificate     /path/to/ssl_cert;
              ssl_certificate_key /path/to/ssl_key;
              location /auth {
                proxy_pass         http://keycloak-service-name:8081/;
                proxy_set_header   Host $host;
                proxy_set_header X-Forwarded-Host $host;
                proxy_set_header X-Forwarded-Proto https;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
              }
          }
      }
```
With this configuration, everything is working as expected. The admin UI is loading correctly, and Keycloak is accessible publicly and privately through the corresponding hostnames configured on the reverse proxies.

With `hostname:v2` we are unable to replicate this setup; specifically, we are having a hard time exposing Keycloak over a custom context path (`/auth` in the above case).
According to the [hostname v2 documentation](https://www.keycloak.org/server/hostname), everything can be set up using `X-Forwarded-*` headers through the reverse proxy, including the context path. This is how we set up those headers in our reverse proxy configuration:
```
location /auth {
  proxy_pass         http://keycloak-service-name:8180/;
  proxy_set_header   Host $host;
  proxy_set_header X-Forwarded-Host $host;
  proxy_set_header X-Forwarded-Proto https; 
  proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  proxy_set_header X-Forwarded-Prefix /auth;  # Please note that we're trying to use this header to set the context path 
}
```
Our new (ie. v26) Keycloak configuration relevant to what is being discussed in this issue is as follows:
```
- name: KC_HOSTNAME_STRICT
  value: 'false'
- name: KC_HTTP_ENABLED
  value: 'true'
- name: KC_HOSTNAME_BACKCHANNEL_DYNAMIC
  value: 'false'
- name: KC_PROXY_HEADERS
  value: xforwarded
```
When we deployed the above configurations, we hit an issue where we can't load the admin console, which seems to be hanging forever. By looking at the network console, we can see that the Keycloak UI is trying to load resources from a path that is missing the context path provided through the reverse proxy `X-Forwarded-Prefix` header.
Moreover, if we curl `https://auth-internal.domain.tld/auth/admin/master/console/` endpoint, we are returned a 200 status code (please note the `X-Forwarded-Prefix` header setting the context path being honored in the log below):
```
2024-11-25 16:39:29,809 INFO  [io.quarkus.http.access-log] (executor-thread-1) 
xx.xx.xx.xx - - [25/Nov/2024:16:39:29 +0000] "GET /auth/admin/master/console/ HTTP/1.0" 200 3189
```
and we can see the following in the HTML being returned:
```
<script id="environment" type="application/json">
      {
        "serverBaseUrl": "https://auth-internal.domain.tld",
        "adminBaseUrl": "https://auth-internal.domain.tld",
        "authUrl": "https://auth-internal.domain.tld",
        "authServerUrl": "https://auth-internal.domain.tld",
        "realm": "master",
        "clientId": "security-admin-console",
        "resourceUrl": "/resources/ihcaj/admin/keycloak.v2",
        "logo": "",
        "logoUrl": "",
        "consoleBaseUrl": "/admin/master/console/",
        "masterRealm": "master",
        "resourceVersion": "ihcaj"
      }
</script>
```
Please note that all the URLs are missing the context path specified through the header, which should have been appended to them.
All of this can be worked around by appending the build time option `--http-relative-path=/auth` in the Dockerfile for the Keycloak image, but we do not want to go down this route as this was achievable in the previous Keycloak versions. We need to use the image for multiple Keycloak deployments with different paths.

Thank you for your attention and for your precious work.

### Version

26.0.5

### Regression

- [X] The issue is a regression

### Expected behavior

Admin UI loading and the URLs in the description should have the context path provided by the reverse proxy appended to them.

### Actual behavior

Admin UI is not loading and the context path is not appended to URLs.

### How to Reproduce?

Please see the provided configuration snippets in the description for the issue.

### Anything else?

We even tried to add a `X-Forwarded-Port` to our reverse proxy configuration:
```
location /auth {
  proxy_pass         http://keycloak-service-name:8180/;
  proxy_set_header   Host $host;
  proxy_set_header X-Forwarded-Host $host;
  proxy_set_header X-Forwarded-Proto https;
  proxy_set_header X-Forwarded-Port 666;        # Port added just to see if it made any difference
  proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  proxy_set_header X-Forwarded-Prefix /auth;  # Please note that we're trying to use this header to set the context path 
}
```
and we can see that it is being handled correctly:
```
<script id="environment" type="application/json">
  {
    "serverBaseUrl": "https://auth-internal.domain.tld:666",         # Looks like the header is being honored
    "adminBaseUrl": "https://auth-internal.domain.tld:666",
    "authUrl": "https://auth-internal.domain.tld:666",
    "authServerUrl": "https://auth-internal.domain.tld:666",
    "realm": "master",
    "clientId": "security-admin-console",
    "resourceUrl": "/resources/ihcaj/admin/keycloak.v2",
    "logo": "",
    "logoUrl": "",
    "consoleBaseUrl": "/admin/master/console/",
    "masterRealm": "master",
    "resourceVersion": "ihcaj"
  }
</script>
```

shawkins:
closes: #35298

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -12,6 +12,10 @@ In minor or patch releases, {project_name} will only introduce breaking changes 
 Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
 It also lists significant changes to internal APIs.
 
+=== `X-Forwarded-Prefix` Header is now supported
+
+With `proxy-headers` set to `xforwarded`, the server can determine the proxy context path from the `X-Forwarded-Prefix` header.
+
 === New database indexes on the `BROKER_LINK` table
 
 The `BROKER_LINK` table now contains two additional indexes `IDX_BROKER_LINK_USER_ID` and `IDX_BROKER_LINK_IDENTITY_PROVIDER` to improve performance.
--- a/docs/guides/server/hostname.adoc
+++ b/docs/guides/server/hostname.adoc
@@ -66,7 +66,7 @@ The `proxy-headers` option can be also used to resolve the URL partially dynamic
 
 <@kc.start parameters="--hostname my.keycloak.org --proxy-headers xforwarded"/>
 
-In this case, scheme, and port are resolved dynamically from X-Forwarded-* headers, while hostname is statically defined as `my.keycloak.org`.
+In this case, scheme, port, and context-path are resolved dynamically from X-Forwarded-* headers, while hostname is statically defined as `my.keycloak.org`.
 
 === Fixed URLs
 
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -30,7 +30,7 @@ You only need to proxy port `8443` (or `8080`) even when you use different host 
 
 * By default if the option is not specified, no reverse proxy headers are parsed. This should be used when no proxy is in use or with https passthrough.
 * `forwarded` enables parsing of the `Forwarded` header as per https://www.rfc-editor.org/rfc/rfc7239.html[RFC 7239].
-* `xforwarded` enables parsing of non-standard `X-Forwarded-*` headers, such as `X-Forwarded-For`, `X-Forwarded-Proto`, `X-Forwarded-Host`, and `X-Forwarded-Port`.
+* `xforwarded` enables parsing of non-standard `X-Forwarded-*` headers, such as `X-Forwarded-For`, `X-Forwarded-Proto`, `X-Forwarded-Host`, `X-Forwarded-Port`, and `X-Forwarded-Prefix`.
 
 NOTE: If you are using a reverse proxy for anything other than https passthrough and do not set the `proxy-headers` option, then by default you will see 403 Forbidden responses to requests via the proxy that perform origin checking.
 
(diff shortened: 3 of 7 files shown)


vmuzikar (approved):
LGTM, thank you.

Issue #43252, opened 2025-10-07
Zero-downtime upgrades between patch releases of Keycloak


ahus1:
### Description

Currently Keycloak is unable to provide any guarantees of runtime compatibility between different patch releases on a given release stream. In order for a user to be able to upgrade Keycloak safely, it's necessary for the Keycloak cluster to be gracefully shutdown and restarted with the new Keycloak version. This applies to individual Keycloak clusters and multi-site "HA" environments.

### Value proposition

Allowing users to upgrade Keycloak for patch releases without downtime will significantly improve the user experience and allow multi-site HA deployments to be truly HA. Until zero-downtime upgrades are possible, it will always be necessary for customers to plan for services outages which adds an additional barrier to upgrading.

While Keycloak has patch releases every two weeks, there are minor releases only every three months. After this change, a Keycloak community user would have four downtimes a year instead of ~20, while a RHBK customer would have two downtimes a year. 

It would be good to have the same benefits also for single-site deployments.

### Goals

User expectations: 

1. "The roll out of a patch release will succeed and all nodes with the next patch release can join the cluster"
2. "I can refresh tokens for existing sessions while the upgrade is happening and after it is complete"
3. "I can continue a login flow while the upgrade is running and start new ones"
4. "I can log out"
5. "The account and admin console will work 'best effort', and I might need to restart them by refreshing the browser (although I expect hints in the UI)"
ALTERNATIVE:
"The account and admin console are disabled during upgrades" (as the user might switch between upgraded and non-upgraded instances) - this might also be an option for other functionality like registration of new users or other less-often used functionality.

###  Non-Goals

* Security fixes can still break no-downtime
* (future) Support zero-downtime upgrades for Infinispan patch releases: [ISPN-16028](https://issues.redhat.com/browse/ISPN-16028)
* (future) Support zero-downtime upgrades for Keycloak's SAML endpoints (this could wait for a second iteration - if it would be cheap to include / wouldn't make a difference, keep it in)
* (future) Support zero-downtime upgrades for multi site
* (future) Running a migration after a rolling update is complete (for example to fix authentication flows or a realm configuration)
* Restarting of an authentication flow during upgrades is allowed, and wouldn't break the zero-downtime promise
* Seamless working the admin console and account console (will require a reload)
* Support zero-downtime configuration changes

#### What users complain about

[...]

PR #45715, opened 2026-01-23
Consider "Internal representation of client sessions changed" a breaking change

marcusdacoregio:
This is a breaking change if the user's deployment strategy can have two different versions running at the same time, like a canary rollout, for example.

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_4_0.adoc
@@ -63,6 +63,11 @@ quarkus.datasource.user-store.db-kind=postgresql
 quarkus.datasource.user-store.username=my-username
 ----
 
+=== Internal representation of client sessions changed
+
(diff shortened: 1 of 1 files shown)


ahus1:
@marcusdacoregio - I hear you, still running two different versions of Keycloak in a cluster is not a supported setup. As it is not a supported setup, this is not a breaking change. 

See https://www.keycloak.org/docs/latest/upgrading/index.html#_prep_migration for the docs. Step 1 is: "Shut down Keycloak"

We're currently working on officially supporting rolling updates of patch releases in https://github.com/keycloak/keycloak/issues/43252 which is planned for 26.6. In 26.4/26.5 it is available as a preview - see https://www.keycloak.org/operator/rolling-updates#operator-rolling-updates-for-patch-releases and https://www.keycloak.org/server/update-compatibility

I know that rolling updates work today most of the time, and sometimes even for minor updates with slight interruptions which is good enough for many, but still it is not supported as there are edge cases.

marcusdacoregio:
Thanks for the context @ahus1.

PR #45723, opened 2026-01-23
Client policy executor to allow extra audiences for JWT authorization grant

### LINKED ISSUE #45180: Allow to configure valid audiences for JWT authorization grant on Google IDP
### Description

Typical use-case for Google:

1) There is social login via Google (for example on mobile application). This does not involve Keycloak yet.

2) The application has Google ID-token from this login

3) The application may send request to Keycloak to exchange the Google ID-token via JWT Authorization grant for the Keycloak token

On Keycloak side, there might be Google Identity provider configured in the Keycloak realm. However it is possible that Google ID-token obtained in step 2 has different audience (`aud`) then the `Client ID` configured on the Google Identity provider in Keycloak realm. As the Google Identity provider `Client ID` refers to the client, which is used for login users via Keycloak.

### Possible solution

In the Google social provider, we will have option like "Allowed audiences for JWT Authorization grant" . This will allow to configure list of client IDs, which Keycloak will trust as valid client IDs of the Google clients. Keycloak will allow JWT Authorization grant request with Google ID-token just if:

- The `aud` claim in the Google ID-token corresponds to the "Client ID" of the Google Identity provider (The default behaviour specified in #45178 )
- The `aud` claim in the Google ID-token corresponds to one of the configured "Allowed audiences for JWT Authorization grant" .

rmartinc:
Closes #45180

Draft for the moment. The PR adds a new client policy executor to allow extra valid audiences. The issue is for google but we can extend to any jwt grant. Things to consider:

* Do we manage audiences defined as extra or as unique? I can terminate the request and fail but I prefer to manage them as extra.
* The executor can be used for any JWT athorization grant, not only for google. Do we limit to google or social? I also prefer to make it general.

If you see this OK I will move it out of draft.

### DIFF
--- a/services/src/main/java/org/keycloak/services/clientpolicy/executor/JWTAuthorizationGrantAudienceExecutor.java
+++ b/services/src/main/java/org/keycloak/services/clientpolicy/executor/JWTAuthorizationGrantAudienceExecutor.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2026 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.services.clientpolicy.executor;
+
+import java.util.Set;
+
+import org.keycloak.OAuthErrorException;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.representations.JsonWebToken;
+import org.keycloak.representations.idm.ClientPolicyExecutorConfigurationRepresentation;
+import org.keycloak.services.clientpolicy.ClientPolicyContext;
+import org.keycloak.services.clientpolicy.ClientPolicyException;
+import org.keycloak.services.clientpolicy.context.JWTAuthorizationGrantContext;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jboss.logging.Logger;
+
+/**
+ *
+ * @author rmartinc
+ */
+public class JWTAuthorizationGrantAudienceExecutor implements ClientPolicyExecutorProvider<JWTAuthorizationGrantAudienceExecutor.Configuration> {
+
+    private static final Logger logger = Logger.getLogger(JWTAuthorizationGrantAudienceExecutor.class);
+    private final KeycloakSession session;
+    private Configuration configuration;
+
+    public static class Configuration extends ClientPolicyExecutorConfigurationRepresentation {
+
+        @JsonProperty(JWTAuthorizationGrantAudienceExecutorFactory.ALLOWED_AUDIENCE)
+        protected Set<String> allowedAudience;
+
+        public Set<String> getAllowedAudience() {
+            return allowedAudience;
+        }
+
+        public void setAllowedAudience(Set<String> allowedAudience) {
+            this.allowedAudience = allowedAudience;
+        }
+    }
+
+    public JWTAuthorizationGrantAudienceExecutor(KeycloakSession session) {
+        this.session = session;
+    }
+
+    @Override
+    public String getProviderId() {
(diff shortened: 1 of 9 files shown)


services/src/main/java/org/keycloak/services/clientpolicy/executor/JWTAuthorizationGrantAudienceExecutor.java
```diff
@@ -0,0 +1,119 @@
...
+    public void executeOnEvent(ClientPolicyContext context) throws ClientPolicyException {
+        switch (context.getEvent()) {
+            case JWT_AUTHORIZATION_GRANT -> {
+                JWTAuthorizationGrantContext jwtAuthnGrantContext = ((JWTAuthorizationGrantContext) context);
+                validateIssuer(jwtAuthnGrantContext);
+            }
+        }
+    }
+
+    private void validateIssuer(JWTAuthorizationGrantContext jwtAuthnGrantContext) throws ClientPolicyException {
```
mposolda: Minor: typo in method name (validates audiences rather than "issuer")

mposolda (commented):
@rmartinc Thanks!

My vote is to rather use the audiences as "only valid audiences" rather than "Extra audiences" when this executor is used. Just for better security. If client `foo` uses Google token with audience `bar`, it will likely always use the Google ID token with that audience and no need for using the other? If the client really needs to use both `bar` and the `keycloak-client-id`, then it can possibly configure both audiences as valid audiences in the executor?
@graziang WDYT?

+1 for not having this specific to Google, but generic to other IDPs as well (also makes it possible to test this, which is good)

One minor comment: May need to require a bit also client policies executors page and add new executor: https://www.keycloak.org/docs/latest/server_admin/index.html#executor

graziang:
@rmartinc @mposolda I have a doubt about how to configure client policies for the use case described in the comment of te issue:
https://github.com/keycloak/keycloak/issues/45180#issuecomment-3717842547

How do we configure the additional audience to allow `googleapp1` as an audience for `keycloakapp1` and `googleapp2` for `keycloakapp2`? As far as I know, we don’t have a client-based condition that can be combined with the IdentityProviderCondition. Am I missing something?
If we find a solution for this, I would opt for the executor only for the “allowed audiences” instead of “extra audience”.

rmartinc:
@graziang The problem is we don't have a condition to select clients. But they can use the `client-attributes`, `client-roles` or `client-scopes` to differentiate the client in the request and apply the specific executor for them. We can add a client condition too, direct client selection or using a regex over the client_id or something similar. But if we want that condition I would manage it as a followup.

graziang:
@rmartinc If I remember correctly there’s a reason why we don’t have a client-specific condition, but I can’t remember the reason. Configuring it via `client attributes, client roles, or client scopes` is probably not so ideal for this purpose. Let’s see if @mposolda  has a different idea.

rmartinc:
@mposolda @graziang I have updated the PR with your suggestions and moved it out of draft. The only missing point is that @graziang commented, there is no direct condition to select a single client (you can use a specific attribute, a specific role or a specific scope but that's a bit convoluted). The test is using the condition over the IdP only.

services/src/main/java/org/keycloak/services/clientpolicy/executor/JWTAuthorizationGrantAudienceExecutor.java
rmartinc: Done!

mposolda (approved):
@graziang @rmartinc Yes, there is not condition for "Select specific client" . This was not added for the scalability reasons as some deployments can have thousands of clients... It is possible to indirectly achieve something like this for example with the "Client role condition", which we also document (https://www.keycloak.org/docs/latest/server_admin/index.html#architecture ).

In other words, if you want to select only `foo-client` , you can make sure that this client has a client role named `foo-client-role` and then use client-role policy for this.

Not sure if we want to improve on this and add some other way to be able to select single client (or clients by regex etc), but if we want, agree that this would need to be a follow-up.

The client-policy approach has advantage that it will scale in case same audiences needed for group of clients. However it will not scale well for the use-case with 1000 different OIDC clients pointing to 1000 different google clients (which would have 1000 different 3rd-party audiences). However ability to configure this per client would likely be a bit of mess (would require client attributes having something `google-idp:valid-audiences` etc...) . However I am not sure if there could be deployments with 1000 different Google clients? My vote is to consider "client attributes based approach" as possible follow-up if it is requested to support something like this. WDYT?

I am approving the PR as the current way seems OK to me. @graziang Does it also look ok to you this way?

graziang (approved):
@mposolda Yes, adding extra audiences is already a specific case, and having extra audiences per client is an even more specific case. So it’s fine to consider it only if it will be requested. @rmartinc thanks, approved!

mposolda:
@graziang @rmartinc Thanks!

PR #45731, opened 2026-01-23
Improve UI for inviting realm users

kota65535:
Closes #45704

## Summary
- Added an "Invite realm user" button to open the existing member selection modal and send invitations.
- Changed the text of the existing "Invite member" button to "Invite new user".

image

image

image

## Considerations

- Realm users with a pending invitation should be filtered out, but the current OrganizationInvitationModel does not have a user ID, so we must resolve each user's ID from the emails of the invitations. It is inefficient because the API must be executed the same number of times as the all invitations.
- Also, the error message "Duplicate resource error" is unclear when we send an invitation to a user with a pending invitation. This is because the [invite-existing-user](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_post_adminrealmsrealmorganizationsorg_idmembersinvite_existing_user) API does not handle exceptions from EntityManagerProxy.
image

I believe these problems are not so critical, and should be addressed by creating separate issues.


ahus1:
@pedroigor - this PR targets organisations, maybe your team can have a look.

kota65535:
Hi @pedroigor, do you have some time to look at this?

sguilhen:
@kota65535 thanks for the work you've put into this. I have requested reviews for the PR so we can move forward with the enhancement.

I agree with the comments about the problems you've encountered  - those are not critical and can be tracked separately. Would you mind raising issues for both of them so we can add to the backlog?

ssilvert (commented):
The code looks OK.  I can approve once conflicts are fixed.

kota65535:
@sguilhen Sure, I will create the issues as soon as this PR is merged.
@ssilvert I've fixed the conflicts. Can you take a look again?

ssilvert:
@kota65535 I tried this out and I do think we need to filter out users without an email address.  If you try to invite a user without an email then you get:
image

And you get a NullPointerException in the server log.

kota65535:
@ssilvert I've made the requested changes. Take another look when you have a chance.

ssilvert (approved):

PR #45734, opened 2026-01-24
Fix how the admin ui obtains its logo

### LINKED ISSUE #45733: Admin UI theme logo not displaying from theme properties
### Before reporting an issue

### Area

admin/ui

### Describe the bug

The admin UI theming system does not properly display custom logos configured in theme properties, unlike the login and account themes which work correctly.

When setting a custom logo in the admin console theme settings, the logo does not appear in the admin UI header. The issue occurs because the PageHeader.tsx component incorrectly prioritizes empty preview logos over the actual theme logo from `environment.logo`.

### Version

26.5.2

### Regression

### Expected behavior

- Admin UI should display the logo specified in theme.properties - Should behave consistently with login and account themes
- Preview functionality should still work during theme editing

### Actual behavior

- Admin theme logos from theme.properties are ignored
- Empty `customLogo` values override the theme logo
- Login and account themes work correctly with the same theme configuration

### How to Reproduce?

1. Create a custom theme with `logo=img/logo.png` in admin theme.properties
2. Deploy the theme to Keycloak
3. Set admin console theme to the custom theme
4. Observe that the custom logo does not appear in the admin header

### Anything else?

The issue is in js/apps/admin-ui/src/PageHeader.tsx line 116 where the logic uses `customLogo || joinPath(environment.resourceUrl, logo)`. The `customLogo` from LogoContext is often an empty string, which overrides the valid theme logo from `environment.logo`.

martwana:
## Description
Fixes the admin UI theming bug where custom logos from theme properties were not being displayed, making admin theming inconsistent with login and account themes.
## Problem
The PageHeader.tsx component was incorrectly prioritizing empty preview logos over actual theme logos. The logic `customLogo || joinPath(environment.resourceUrl, logo)` meant that even empty strings from the LogoContext would override the valid theme logo from `environment.logo`.
## Solution
Updated the logo source logic to only use custom logo when it has a meaningful value:
```typescript
src: (customLogo && customLogo.trim()) ? customLogo : joinPath(environment.resourceUrl, logo)
```

Fixes #45733


js/apps/admin-ui/src/PageHeader.tsx
```diff
@@ -38,7 +38,7 @@ const ServerInfoDropdownItem = () => {
   return (
     <DropdownItem
       key="server info"
-      component={(props) => <Link {...props} to={toDashboard({ realm })} />}
+      component={(props: React.ComponentProps<typeof Link>) => <Link {...props} to={toDashboard({ realm })} />}
```
ssilvert: Is the type really needed here?

ssilvert (changes requested):

js/apps/admin-ui/src/PageHeader.tsx
martwana: No, thats fair. Thinking about it, it doesn't add any safety and just adds bloat to the code. 

Removed.

ssilvert (dismissed):

ssilvert:
@martwana Looks like we have some linting errors to fix before this can be merged.

ssilvert (approved):

PR #45735, opened 2026-01-24
Run new test framework db tests on Azure

Pepo48:
Closes: #45658

### DIFF
--- a/.github/actions/run-store-tests/action.yml
+++ b/.github/actions/run-store-tests/action.yml
@@ -19,7 +19,7 @@ runs:
     - name: Run new base tests
       shell: bash
       run: |
-        KC_TEST_DATABASE=${{ inputs.db }} KC_TEST_DATABASE_REUSE=true TESTCONTAINERS_REUSE_ENABLE=true ./mvnw package -f tests/pom.xml -Dtest=DatabaseTestSuite -Dkeycloak.distribution.start.timeout=360
+        KC_TEST_DATABASE=${{ inputs.db }} KC_TEST_DATABASE_REUSE=true TESTCONTAINERS_REUSE_ENABLE=true ./mvnw package -f tests/pom.xml -Dtest=DatabaseTestSuite -Dkc.test.server.start.timeout=360
 
     - name: Database container port
       shell: bash
--- a/.github/scripts/ansible/roles/mvn_remote_runner/tasks/run.yml
+++ b/.github/scripts/ansible/roles/mvn_remote_runner/tasks/run.yml
@@ -24,6 +24,8 @@
   async: 86400
   poll: 10
   register: result
+  # Allow playbook to continue to collect test results even when tests fail
+  ignore_errors: yes
 
 - debug: var=result
 
@@ -52,3 +54,8 @@
     mode: pull
   with_items:
     - "{{ output.files }}"
+
+- name: Fail if Maven command failed
+  fail:
+    msg: "Maven command failed with return code {{ result.rc }}: {{ result.stderr | default('') }}"
+  when: result.rc != 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -588,7 +588,7 @@ jobs:
          AWS_REGION=${{ steps.aurora-init.outputs.region }}
          
          cd .github/scripts/ansible
-         ./mvn_remote_runner.sh ${AWS_REGION} ${EC2_CLUSTER_NAME} "-Pexecute-sql -f tests/base/pom.xml sql:execute@clear-schema -Dautocommit=true -Ddriver=software.amazon.jdbc.Driver -Durl=\"jdbc:aws-wrapper:postgresql://${{ steps.aurora-create.outputs.endpoint }}/keycloak${{ steps.aurora-init.outputs.jdbc_params }}\" -Dusername=keycloak -Dpassword=${{ steps.aurora-init.outputs.aurora-cluster-password }}"
+         ./mvn_remote_runner.sh ${AWS_REGION} ${EC2_CLUSTER_NAME} "-Pexecute-sql -f tests/base/pom.xml sql:execute@clear-schema-psql -Dautocommit=true -Ddriver=software.amazon.jdbc.Driver -Durl=\"jdbc:aws-wrapper:postgresql://${{ steps.aurora-create.outputs.endpoint }}/keycloak${{ steps.aurora-init.outputs.jdbc_params }}\" -Dusername=keycloak -Dpassword=${{ steps.aurora-init.outputs.aurora-cluster-password }}"
 
       - name: Run Aurora new database tests on EC2
         id: aurora-new-integration-tests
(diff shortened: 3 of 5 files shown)


Pepo48:
Here's the test run: https://github.com/Pepo48/keycloak/actions/runs/21312177220/job/61350001221

Below I added a few comments that explains some of the changes that may not be that obvious.

.github/actions/run-store-tests/action.yml
```diff
@@ -19,7 +19,7 @@ runs:
     - name: Run new base tests
       shell: bash
       run: |
-        KC_TEST_DATABASE=${{ inputs.db }} KC_TEST_DATABASE_REUSE=true TESTCONTAINERS_REUSE_ENABLE=true ./mvnw package -f tests/pom.xml -Dtest=DatabaseTestSuite -Dkeycloak.distribution.start.timeout=360
+        KC_TEST_DATABASE=${{ inputs.db }} KC_TEST_DATABASE_REUSE=true TESTCONTAINERS_REUSE_ENABLE=true ./mvnw package -f tests/pom.xml -Dtest=DatabaseTestSuite -Dkc.test.server.start.timeout=360
```
Pepo48, line 22: During my testing, `GeneratedRsaKeyProviderTest.updateKeysizeForEnc` failed with  
`java.lang.RuntimeException: Keycloak did not start within timeout`.

When I passed `PROPS+=" -Dkeycloak.distribution.start.timeout=360"` in `ci.yml`, it was not applied and the timeout remained at the default 120 seconds. I then noticed that the new framework defines a different property via `DistributionKeycloakServerSupplier`.

Based on that, I assume that when running the new base tests we are facing the same situation, and that we likely had a bug there as well, since the timeout configuration did not work properly either.

cc: @vaceksimon

.github/scripts/ansible/roles/mvn_remote_runner/tasks/run.yml
```diff
@@ -52,3 +54,8 @@
     mode: pull
   with_items:
     - "{{ output.files }}"
+
+- name: Fail if Maven command failed
```
Pepo48, line 58: When tests failed, we weren't getting the test results because the CI stopped before syncing them.

This also improved the fact that the error log used to be printed on a single line within the GH CI console.  Now it looks a bit more readable:

Image
vs. now:

Image

.github/workflows/ci.yml
```diff
@@ -775,6 +775,57 @@ jobs:
...
+          PROPS="-Dkc.test.database=remote -Dkc.test.database.vendor=mssql"
+          PROPS+=" -Dkc.test.database.user=${{ steps.azure-create.outputs.username }}"
+          PROPS+=" -Dkc.test.database.password=${{ env.AZURE_DB_PASSWORD }}"
+          PROPS+=" -Dkc.test.database.url=\"jdbc:sqlserver://${{ steps.azure-create.outputs.endpoint }}:1433;databaseName=${{ steps.azure-create.outputs.db }};encrypt=true;trustServerCertificate=false;sendStringParametersAsUnicode=false\""
+          PROPS+=" -Dkc.test.database.driver=com.microsoft.sqlserver.jdbc.SQLServerDriver"
+          PROPS+=" -Dkc.test.database.driver.artifact=com.microsoft.sqlserver:mssql-jdbc"
+          PROPS+=" -Dkc.test.server.start.timeout=360"
+
+          cd .github/scripts/ansible
+          set +e
```
Pepo48, line 818: Added `set +e` before the command to be able to capture exit code in order to properly fail the step and prevent:
```
TASK [mvn_remote_runner : Fail if Maven command failed] ************************
Error: : Task failed: Action failed: Maven command failed with return code 1: 
Origin: /home/runner/work/keycloak/keycloak/.github/scripts/ansible/roles/mvn_remote_runner/tasks/run.yml:58:3

56     - "*** output.files ***"
57
58 - name: Fail if Maven command failed
     ^ column 3

fatal: [keycloak]: FAILED! => ***"changed": false, "msg": "Maven command failed with return code 1: "***
```

ahus1:
@Pepo48 - thank you for looping me in here. I'm not sure if I can be much help here due to my lack of knowledge in the testing on Azure and the new test suite, so I'm removing me as a reviewer. 

Once you can have a review from one of your team members and possibly @vaceksimon I'd be good to merge it. 

If you still want to get feedback on this PR, please schedule a 1:1, walk me through the change, and I might be able to provide some feedback. With that walk-through, I might not be useful here.

Pepo48:
No worries @ahus1, thanks for the reaction! I just wanted to make sure that you're fine with/aware of this https://github.com/keycloak/keycloak/pull/45735#discussion_r2724040613.

.github/scripts/ansible/roles/mvn_remote_runner/tasks/run.yml
ahus1: @ryanemerson - can you please check if this is relevant also for the Aurora test runs?

tests/base/pom.xml
```diff
@@ -175,6 +175,35 @@
...
+                                        <srcFile>../../.github/scripts/azure/sql/mssql_clear.sql</srcFile>
+                                    </srcFiles>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
     </profiles>
```
vaceksimon, line 191: I'm thinking about whether it'd be better to merge these two. We could have one profile `execute-sql` with all the dependencies, but two different executions: `clear-schema-psql` and `clear-schema-mssql`

I figure it'd be cleaner and hopefully the dependencies are not that huge to pose a problem (?).
WDYT?

Pepo48: @vaceksimon thanks for the review. Please, take a look at https://github.com/keycloak/keycloak/pull/45735/changes/461aff56c0517a0b5248ba86b508778ef09f39a5

Initially, I thought about it too. But the downside of this is that Maven will download both the PostgreSQL and MSSQL drivers, even if you only run one execution (at least that's how I understand it). Thus I did not include it in the final PR. On the other hand, the drivers are tiny and cached, so it's up to us to assess the trade-off.

Pepo48: My other idea was to somehow conditionally include the dependencies (via Maven properties?), but I'm not sure how it behaves, specifically with plugin dependencies. And at that point I felt that I was overengineering it. :)

vaceksimon: @Pepo48 my personal choice would be this, but I don't have a strong opinion. @stianst do you have a preference here?

.github/scripts/ansible/roles/mvn_remote_runner/tasks/run.yml
ryanemerson: It is relevant, as the Aurora tests make use of the playbooks via `mvn_remote_runner.sh`.

+1 to the change, as the single line output was difficult to debug previously.

ryanemerson (approved):

tests/base/pom.xml
vaceksimon: either way, lgtm

vaceksimon (approved):

vmuzikar (approved):
As far as I can tell, LGTM.

PR #45737, opened 2026-01-24
Optionally allows case-sensitive username/email for ldap federation

### LINKED ISSUE #45736: Forced lowercasing of username and email from ldap-user-import since 26.4.1 breaks case-sensitive applications
### Before reporting an issue

### Area

ldap

### Describe the bug

Since upgrading Keycloak from 26.4.0 to 26.4.1 (and later versions, e.g. 26.5.2), usernames and email addresses from ldap users are automatically converted to lowercase during import and authentication.

In previous versions (≤ 26.4.0), Keycloak preserved the original casing.
This change breaks integrations with existing applications that rely on case-sensitive usernames or email addresses (e.g. legacy systems or databases where casing is significant).

At the moment, this behavior appears to be enforced and cannot be disabled via configuration. There is no soft migration path.

An option to allow case-sensitive import of usernames and email addresses could solve this problem.

### Version

26.5.2

### Regression

### Expected behavior

- Keycloak should preserve the original casing of usernames and email addresses, as in versions ≤ 26.4.0

or

- There should be a configuration option to disable automatic lowercasing (for backward compatibility).

### Actual behavior

- Usernames and email addresses are always stored and processed in lowercase.
- This behavior occurs during:
  - User import
  - Authentication
- Existing applications that expect case-sensitive values fail to authenticate users correctly.

### How to Reproduce?

1. Install Keycloak 26.4.1 or later (e.g. 26.5.0)
2. Import a user with a mixed-case username or email
3. Inspect the stored user data in Keycloak
4. Attempt authentication from an application that treats usernames/emails as case-sensitive and expects the user to have at least one capital letter

### Anything else?

The following changes/PRs seem to have introduced this change in behaviour:

- https://github.com/keycloak/keycloak/pull/43261
- https://github.com/keycloak/keycloak/pull/43659

mandree95:
A new configuration option was added to the ldap federation. It allows to import username/email case-sensitive if the flag and the feature 'Import users' is enabled.

I'm not able to fully test this until next week as I have not access to an ldap server right now.

Closes #45736


pedroigor (commented):
Does it make sense to make LDAP handling usernames different, given that there are assumptions everywhere in Keycloak that usernames are case-insensitive and lowercase?

PR #45739, opened 2026-01-25
MCP requirement - Audience Mapper and Checker referring RFC 8707 resource parameter

### LINKED ISSUE #41527: MCP requirement - Audience Mapper and Checker referring RFC 8707 resource parameter
### Description

The issue is related to https://github.com/keycloak/keycloak/issues/41526 .

Adding a `resource` parameter value in RFC 8707 to an `aud` claim of an access token and an response of its token introspection.
Moreover, to comply with MCP specification [version 2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization), we need to also check the followings about `resource` parameter value:
- The value of a `resource` parameter should be [Canonical Server URI](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation).
- The value of a `resource` parameter in an authorization request is the same as the one in a token request.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

tnorimat:
closes #41527

The PR's code size is large, but half of them are for testing.

## Class structure
 - `ResourceIndicatorMapper` : adds the value of `resource` parameter in an authorization request to an access token's "aud" claim.
 - `SecureResourceIndicatorExecutor` / `SecureResourceIndicatorExecutorFactory`:  checks if `resource` parameter value in an authorization request and token request are valid. If not, returns an error.

## The way of adding `resource` parameter value to an access token's `aud` claim
1. `AuthorizationEndpoint.updateAuthenticationSession` stores the `resource` parameter value in an authorization request to an `AuthenticationSession`.
2. `TokenManager.attachAuthenticationSession` copies the `resource` parameter value of the `AuthenticationSession` to `AuthenticatedClientSession`.
3. `ResourceIndicatorMapper` and `SecureResourceIndicatorExecutor` can see it from `AuthenticatedClientSession.getNote`.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/endpoints/request/AuthorizationEndpointRequest.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/endpoints/request/AuthorizationEndpointRequest.java
@@ -55,10 +55,16 @@ public class AuthorizationEndpointRequest {
 
     String acr;
 
+    String resource;
+
     AuthorizationRequestContext authorizationRequestContext;
 
     public String getAcr() {
         return acr;
+    }
+
+    public String getResource() {
+        return resource;
     }
 
     public String getClientId() {
--- a/services/src/main/java/org/keycloak/protocol/oidc/mappers/ResourceIndicatorMapper.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/mappers/ResourceIndicatorMapper.java
@@ -1,0 +1,93 @@
+package org.keycloak.protocol.oidc.mappers;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.keycloak.Config;
+import org.keycloak.OAuth2Constants;
+import org.keycloak.common.Profile;
+import org.keycloak.models.ClientSessionContext;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.ProtocolMapperModel;
+import org.keycloak.models.UserSessionModel;
+import org.keycloak.protocol.oidc.OIDCLoginProtocol;
+import org.keycloak.provider.EnvironmentDependentProviderFactory;
+import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.representations.IDToken;
+
+import org.jboss.logging.Logger;
+
+/**
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public class ResourceIndicatorMapper extends AbstractOIDCProtocolMapper implements OIDCAccessTokenMapper, TokenIntrospectionTokenMapper, EnvironmentDependentProviderFactory {
+
+    public static final String PROVIDER_ID = "oidc-resource-indicator-mapper";
+
+    private static final List<ProviderConfigProperty> configProperties = new ArrayList<>();
+
+    private static final Logger logger = Logger.getLogger(ResourceIndicatorMapper.class);
+
+    static {
+        OIDCAttributeMapperHelper.addIncludeInTokensConfig(configProperties, ResourceIndicatorMapper.class);
+    }
+
+    public List<ProviderConfigProperty> getConfigProperties() {
+        return configProperties;
+    }
+
+    @Override
+    public String getId() {
+        return PROVIDER_ID;
+    }
+
+    @Override
+    public String getDisplayType() {
+        return "Resource Indicators";
+    }
+
+    @Override
+    public String getDisplayCategory() {
+        return TOKEN_MAPPER_CATEGORY;
+    }
+
+    @Override
+    public String getHelpText() {
+        return "Adds requested OAuth2 Resource Indicators to audience claim.";
+    }
+
+    @Override
+    public boolean isSupported(Config.Scope config) {
+        return Profile.isFeatureEnabled(Profile.Feature.RESOURCE_INDICATOR);
+    }
+
+    @Override
+    protected void setClaim(IDToken token, ProtocolMapperModel mappingModel, UserSessionModel userSession, KeycloakSession keycloakSession, ClientSessionContext clientSessionCtx) {
+        if (clientSessionCtx == null || clientSessionCtx.getClientSession() == null) {
+            return;
+        }
+        String resourceInAuthorizationRequest = clientSessionCtx.getClientSession().getNote(OAuth2Constants.RESOURCE);
+        if (resourceInAuthorizationRequest != null && !resourceInAuthorizationRequest.isBlank()) {
+            logger.debugv(" mapper: resource in authorization request = {0}", resourceInAuthorizationRequest);
+            token.addAudience(resourceInAuthorizationRequest);
+        }
+    }
+
+    public static ProtocolMapperModel create(String name, boolean accessToken, boolean introspectionEndpoint) {
+        ProtocolMapperModel mapper = new ProtocolMapperModel();
+        mapper.setName(name);
+        mapper.setProtocolMapper(PROVIDER_ID);
+        mapper.setProtocol(OIDCLoginProtocol.LOGIN_PROTOCOL);
+        Map<String, String> config = new HashMap<>();
+        if (accessToken) {
+            config.put(OIDCAttributeMapperHelper.INCLUDE_IN_ACCESS_TOKEN, "true");
+        }
+        if (introspectionEndpoint) {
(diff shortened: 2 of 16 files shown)


thomasdarimont:
Thanks for picking this up again. 

I see that you are using a protocol mapper to restrict the allowed resource indicator uris.

I used the same approach in my initial PR, but this was declined back then as restricting values for a protocol flow via mapper was uncommon back then. That's why I went with the more sophisticated approach that introduced a SPI to restrict allowed resource indicators backed by authorization services to allow users to define resources.

Perhaps opinions changed in-between :)

I think using a protocol mapper to declare a static list of allowed resource indicator uris per client has its limits but might still be expressive enough to solve common scenarios.

tnorimat:
@thomasdarimont Hello, thank you for the comments. I missed the point.
Regarding restricting the allowed resources, I will try to client policies while Introducing a new SPI is better option.

tnorimat:
@stianst Hello, the feature is also important the same as CIMD to comply with MCP spec even if there is the workaround using client scope. Coud anyone in your team check the PR?

services/src/main/java/org/keycloak/protocol/oidc/mappers/ResourceIndicatorMapper.java
```diff
@@ -0,0 +1,93 @@
...
+import org.jboss.logging.Logger;
+
+/**
+ * @author <a href="mailto:takashi.norimatsu.ws@hitachi.com">Takashi Norimatsu</a>
+ */
+public class ResourceIndicatorMapper extends AbstractOIDCProtocolMapper implements OIDCAccessTokenMapper, TokenIntrospectionTokenMapper, EnvironmentDependentProviderFactory {
+
+    public static final String PROVIDER_ID = "oidc-resource-indicator-mapper";
+
+    private static final List<ProviderConfigProperty> configProperties = new ArrayList<>();
```
cgeorgilakis, line 29: From [JSON Web Token (JWT) Profile for OAuth 2.0 Access Tokens](https://datatracker.ietf.org/doc/html/rfc9068) : 

> aud  REQUIRED - as defined in [Section 4.1.3 of [RFC7519]](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3).  See
      [Section 3](https://datatracker.ietf.org/doc/html/rfc9068#section-3) for indications on how an authorization server should
      determine the value of "aud" depending on the request.
> If the request does not include a "resource" parameter, the
   authorization server MUST use a default resource indicator in the
   "aud" claim.  If a "scope" parameter is present in the request, the
   authorization server SHOULD use it to infer the value of the default
   resource indicator to be used in the "aud" claim. 

I propose to add an optional default value for aud claim (`defaultAud`) . Default value will be null.
If `resource` is empty and `defaultAud` is not empty, mapper will add this default value to aud claim.
In this way Keycloak will respect the  [JSON Web Token (JWT) Profile for OAuth 2.0 Access Tokens](https://datatracker.ietf.org/doc/html/rfc9068) for `aud` claim.

services/src/main/java/org/keycloak/protocol/oidc/endpoints/request/AuthorizationEndpointRequest.java
```diff
@@ -55,12 +55,18 @@ public class AuthorizationEndpointRequest {
 
     String acr;
 
+    String resource;
```
cgeorgilakis, line 58: I propose this to be `List<String> resources`.
See [my comment](https://github.com/keycloak/keycloak/issues/41527#issuecomment-3823708019) in issue for details.

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureResourceIndicatorExecutor.java
```diff
@@ -0,0 +1,135 @@
...
+            return;
+        }
+
+        switch (context.getEvent()) {
+            case AUTHORIZATION_REQUEST:
+                AuthorizationRequestContext authzRequestContext = (AuthorizationRequestContext) context;
+                String resourceParam = authzRequestContext.getAuthorizationEndpointRequest().getResource();
+                logger.debugv(" on authz request: resourceParam = {0}", resourceParam);
+                List<String> allowResourceList = convertContentFilledList(configuration.getAllowPermittedResources());
+                if (allowResourceList != null && !allowResourceList.isEmpty()) {
```
cgeorgilakis: Combine two if in one if for better performance and clearance.

cgeorgilakis (changes requested):
Nice work for a needed feature.
See my comments for proposed changes.

tnorimat:
@cgeorgilakis Hello, thank you for the comment. I will check them.

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureResourceIndicatorExecutor.java
tnorimat: I modified the codes as follows:
```
                if (allowResourceList != null && !allowResourceList.isEmpty() && !allowResourceList.contains(resourceParam)) {
                        logger.warnv("not allowed resource parameter value: resource = {0}", resourceParam);
                        throw new ClientPolicyException(OAuthErrorException.INVALID_REQUEST, ERR_NOT_PERMITTED_RESOURCE);
                }
```

stianst (commented):
Could you provide an overview on how this should be used? Is it basically just mapping a known `resource` into the token? Not sure if that makes all that much sense to me.

I haven't had time to look at this in more detail sorry, but hope to take a look next week.

tnorimat:
@stianst Hello,

The PR's resource parameter support should be used as follows:
An access token's audience is a single resource on a resource server or resource server itself and its ID is identified in URI format like "iss" claim for an authorization server. One of the actual examples is MCP server.

The same thing can be done by client scope + audience mapper, but it does not work if several resource servers support the same name of scope.

It consists of two parts: mapper and executor.

The mapper has a single role: putting `resource` parameter value of an authorization request to "aud" claim of an access token.

The executor has two roles: 
1. [Audience Binding] checking if `resource` parameter value of an authorization request is one of pre-registered `resource` values. If not, retuning an error.
2. [Consistency] checking if `resource` parameter value of an token request in authorization code grant is equal to `resource` parameter value of an authorization request If not, retuning an error.

From security perspective, both should be applied. If not, any `resource` parameter value is can be set to "aud" claim of the access token.

Comparison of RFC 8707 specification and the spec of the PR is as follows:
- Target types of request:
  - RFC 8707:
    - Authorization Request
    - Token Request with all types of grants
  - The PR:
    - Authorization Request
    - Token Request with authorization code grant
- The number of `resource` parameter:
  - RFC 8707:
    - Allowing multiple `resource` parameter
  - The PR:
    - Only single `resource` parameter
- Downscoping
  - RFC 8707:
    - Possible
  - The PR:
    - Impossible

stianst:
> @stianst Hello,
> 
> The PR's resource parameter support should be used as follows: An access token's audience is a single resource on a resource server or resource server itself and its ID is identified in URI format like "iss" claim for an authorization server. One of the actual examples is MCP server.
> 

I'm not really following TBH; so to use this I have to:

1. Create a client scope and add the ResourceIndicatorMapper that will map any `resource` parameter as a `aud` claim
2. Create a client policy most likely per-client to set what clients can use what `resource` parameters?

Also, if I don't do 2, then a client can get any resource into the aud claim of the token?

stianst:
My current thinking for resource indicators is to map it more to how we handle scopes and audiences. Will try to write-up something a bit more detail this week, but I'm thinking something like:

* API is represented as a client (bearer-only most likely) with an additional field `resourceUrl`
* We have client scopes that adds a resource mapper similar to how we do audience mapping; and some way of making an optional client scope required if a resource param is sent (all, or some expression, or something)

In general some thoughts from me:

* Client policies should be optional extras like it is for scopes, audience, etc.
* There should be a representation of the API with the applicable resourceUrl
* We should not leverage authorization services, at least not for the basic use-cases
* We should try to match the design of resource indicators with scopes and audience, and not do something completely different

tnorimat:
@stianst Yes, so make it secure, applying the executor is needed.
To avoid this misconfiguration, I can put two roles onto one component: mapper or executor.
Which client applies the mapper is determined by scope, while the executor is determined by conditions/policy.  So I think the latter is flexible. What do you think about that?

stianst:
Closing this since we now have experimental support for resource indicators through audience filtering: https://github.com/keycloak/keycloak/pull/46763

PR #45752, opened 2026-01-26
Add `operationId` to OpenAPI spec for Client v2

### LINKED ISSUE #45573: Add `operationId` to OpenAPI spec for Client v2
### Description

Add [`operationId`](https://swagger.io/docs/specification/v3_0/paths-and-operations/#operationid) so that it can be used when generating the REST clients for naming methods and functions.

Preferably, make the `operationId` field generated to avoid manually annotating.

vmuzikar:
Closes #45573

Currently it generates the following operationids:
```
  /admin/api/{realmName}/clients/{version}:
    get:
      operationId: getClients
    post:
      operationId: createClient
```
```
  /admin/api/{realmName}/clients/{version}/{id}:
    get:
      operationId: getClient
    put:
      operationId: createOrUpdateClient
    patch:
      operationId: patchClient
    delete:
      operationId: deleteClient
```

The source is the method name and it can be overridden using the `@Operation` annotation.

Potential problem are conflicts as operationid needs to be unique. This might be a problem especially if multiple API groups will be contained in a single OpenAPI spec where same method names can have different meanings in different contexts/groups. I'm starting to think if we should perhaps even have separate OpenAPI specs for different API groups.

### DIFF
--- a/quarkus/runtime/src/main/resources/application.properties
+++ b/quarkus/runtime/src/main/resources/application.properties
@@ -100,6 +100,8 @@ quarkus.swagger-ui.always-include=true
 quarkus.swagger-ui.filter=true
 mp.openapi.scan.packages=org.keycloak.representations.admin.v2,org.keycloak.admin.api,org.keycloak.quarkus.runtime.oas,io.quarkus.smallrye.openapi
 mp.openapi.extensions.smallrye.auto-inheritance=PARENT_ONLY
+mp.openapi.extensions.smallrye.operationIdStrategy=METHOD
+mp.openapi.extensions.smallrye.duplicateOperationIdBehavior=FAIL
 
 # Disable Error messages from smallrye.openapi
 # related issue: https://github.com/keycloak/keycloak/issues/41871


mabartos (approved):
@vmuzikar LGTM, thanks!

Image

> Potential problem are conflicts as operationid needs to be unique. This might be a problem especially if multiple API groups will be contained in a single OpenAPI spec where same method names can have different meanings in different contexts/groups. I'm starting to think if we should perhaps even have separate OpenAPI specs for different API groups.

@vmuzikar Do you think it'll be frequent problem? I don't think so. And if there's the collision, the mentioned Operation annotation can be used, right? I think it's fine for now instead of creating a separate spec.

vmuzikar:
@mabartos Thank you for the review.

> Do you think it'll be frequent problem? I don't think so. And if there's the collision, the mentioned Operation annotation can be used, right? I think it's fine for now instead of creating a separate spec.

Yes, I think we're good for now, just wanted to highlight the potential risk.

PR #45764, opened 2026-01-26
Clarify documentation of OS-level dependencies for Kerberos

### LINKED ISSUE #32527: Provide default krb5.conf or extend migration docs
### Description

When upgrading my Keycloak 20 to the most recent (I know, that was late), SSO with Kerberos stopped working.
TL;DR: The default `/etc/krb5.conf` is not provided since Keycloak 21 which was necessary for our configuration.

I think you should
* either provide a default `/etc/krb5.conf` (as it was before)
* or documented the necessity to add one in the migration guide (for Keycloak 21)

### Discussion

_No response_

### Motivation

- We need to ensure compatibility with an old Kerberos server using the RC4-HMAC algorithm
- To achieve this we created a custom image based on yours, which adds a `crypto-policies` file to `/etc/krb5.conf.d/` file allowing weak algorithms
- This worked fine in Keycloak 20 but does not in Keycloak 25

### Details

I have analyzed the problem and found the reason:
- Since Keycloak 21 you seem to use Red Hat's UBI *micro* image as base image (see [this Dockerfile](https://github.com/keycloak/keycloak/blob/21.0.0/quarkus/container/Dockerfile))
- Previously you used the UBI *minimal* image (see [this earlier commit](https://github.com/keycloak/keycloak/commit/610e3044adea39003fdb256585a74d9adea2cd12#diff-95f281de08964d27b4569a023fe6b0561731e2f1bb701e05608d411edbc5315a) where you switched from `ubi8-minimal` to `ubi8-micro`)
- Unfortunately the micro image variant does not include the package `krb5-libs`
  - compare packages in [ubi9 micro](https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58?container-tabs=packages) and [ubi9 minimal](https://catalog.redhat.com/software/containers/ubi9-minimal/61832888c0d15aff4912fe0d?container-tabs=packages)
  - As far as I know it's exactly that library which includes the default `/etc/krb5.conf` file which also has the statement `includedir /etc/krb5.conf.d/` and some reasonable `[libdefaults]` settings

The full `krb5-libs` package is actually not required to have working Kerberos support. It suffices to add back the `/etc/krb5.conf`.
The `/etc/krb5.conf` will be picked up by Java when trying to login with Kerberos authentication.

### My solution
My personal solution was to copy the original `/etc/krb5.conf` and extend my `Dockerfile` to include this when we build our derived Keycloak 25 image.

ASzc:
Closes #32527

For a long time the Kerberos authentication documentation has implied that freeipa or krb5-libs packages are a necessary dependency. This isn't actually the case. This causes concern and confusion in the case of the container deployment, which doesn't have these packages installed by default due to security hardening.

As evidenced by [the use of the `org.ietf.jgss` package](https://github.com/keycloak/keycloak/blob/77704a91b6067a41bdbf2a5b74fda1260adcedb2/common/src/main/java/org/keycloak/common/util/KerberosJdkProvider.java#L39-L42), Keycloak uses the [Java Generic Security Services API (JGSS)](https://docs.oracle.com/javase/8/docs/technotes/guides/security/jgss/jgss-api-mechanism.html) for kerberos, not krb5-libs via JNA. JGSS by default has no native dependencies. It will use a pure Java implementation[ unless `sun.security.jgss.native` is set to true](https://docs.oracle.com/en/java/javase/11/security/accessing-native-gss-api.html)

Unless overridden by the `java.security.krb5.conf` property, JGSS will look for the kerberos configuration in the same place that krb5-libs puts it, i.e. `/etc/krb5.conf`. This file seems to be the only actual dependency of Keycloak.

This PR updates the migration, container, and server admin guides to make this distinction more apparent.

### DIFF
--- a/docs/documentation/server_admin/topics/authentication/kerberos.adoc
+++ b/docs/documentation/server_admin/topics/authentication/kerberos.adoc
@@ -58,10 +58,10 @@ Ensure the keytab file `/tmp/http.keytab` is accessible on the host where {proje
 
 [[_server_setup]]
 
-Install a Kerberos client on your machine.
+Configure a Kerberos client on your machine.
 
 .Procedure
-. Install a Kerberos client. If your machine runs Fedora, Ubuntu, or RHEL, install the link:https://www.freeipa.org/page/Downloads[freeipa-client] package, containing a Kerberos client and other utilities.
+. Optionally, install a Kerberos client. If your machine runs Fedora, Ubuntu, or RHEL, install the link:https://www.freeipa.org/page/Downloads[freeipa-client] package, containing a Kerberos client and other utilities. Installing this package will supply you with tools and sensible defaults, but it is not required for JGSS Kerberos to function within {project_name}.
 . Configure the Kerberos client (on Linux, the configuration settings are in the link:https://web.mit.edu/kerberos/krb5-1.21/doc/admin/conf_files/krb5_conf.html[/etc/krb5.conf] file ).
 +
 Add your Kerberos realm to the configuration and configure the HTTP domains your server runs on.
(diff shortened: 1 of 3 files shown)


ahus1:
@ASzc - I see that @hwanders gave this PR a thumbs-up ... any more reviews you're anticipating from any Keycloak member before this should be merged?

ASzc:
@ahus1 Ideally someone with an understanding of the kerberos features would double check my analysis

rmartinc (approved):
Thanks @ASzc! LGTM! I see no problem adding these comments to the documentation.

PR #45765, opened 2026-01-26
Fix ClientBean BaseURL returning null

Krematos:
When RootURL is set, and BaseURL is null/empty, it should resolve back to RootURL to prevent FTL errors.

Closes #45746

### DIFF
--- a/services/src/main/java/org/keycloak/forms/login/freemarker/model/ClientBean.java
+++ b/services/src/main/java/org/keycloak/forms/login/freemarker/model/ClientBean.java
@@ -50,6 +50,10 @@ public class ClientBean {
     }
 
     public String getBaseUrl() {
+        String configuredBaseUrl = client.getBaseUrl();
+        if ((configuredBaseUrl == null || configuredBaseUrl.isBlank()) && client.getRootUrl() != null) {
+            configuredBaseUrl = "/";
+        }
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/keycloak/forms/login/freemarker/model/ClientBean.java
```diff
@@ -50,6 +50,10 @@ public String getDescription() {
     }
 
     public String getBaseUrl() {
+        String configuredBaseUrl = client.getBaseUrl();
+        if ((configuredBaseUrl == null || configuredBaseUrl.isBlank()) && client.getRootUrl() != null) {
+            configuredBaseUrl = "/";
+        }
         return ResolveRelative.resolveRelativeUri(session, client.getRootUrl(), client.getBaseUrl());
```
mposolda, line 57: Should not this use `configuredBaseUrl` instead of `client.getBaseUrl()` ? Did you had opportunity to test if this fixes the issue?

mposolda (commented):
Added comment inline. 

It seems you did not tested this as the fix probably did not work at all as it was provided? Could you confirm if this was tested? Not sure if you need custom FTL template for this.

mposolda:
Closing as there is not reply recently. Please re-open (or open new PR) once my concerns above are addressed.

Issue #45745, opened 2026-01-26
mixed-cluster-compatibility-tests fail due to incorrectly masked content in 26.5 branch


ahus1:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

ci

### Describe the bug

As a follow-up to #45708, this is still not working as expected in the 26.5 branch, see https://github.com/keycloak/keycloak/actions/runs/21295812994/job/61301226582

This is most likely cased by a multi-line secret for the Azure credentials. See details below.

### Version

26.5 branch

### Regression

- [ ] The issue is a regression

### Expected behavior

The CI should pass.

### Actual behavior

The CI doesn't pass. 

There is a warning in the summary of the CI stating: 

> Skip output 'ci-compatibility-matrix' since it may contain secret.

... and the matrix is not executed.

### How to Reproduce?

Run a CI with a AZURE_CREDENTIALS set to a multiline JSON. 

Once it contains a `{` or a `}` on a single line, the GHA will start to escape those characters in all places, leading to the problem.

### Anything else?

This should be fixable by instead of sing a multi-line JSON, using a single-line JSON.

mabartos:
I believe this issue is more suited to the SRE team, as they were involved in the previous tasks for the mixed cluster compatibility tests.

Pepo48:
@ahus1 @pruivo my first sight suspicion is that GHA masks each line of multiline secrets, which can cause characters like '{' and '}' to be masked, especially when JSON is formatted with { and } on separate lines (and I know that it is the case with AZURE_CREDENTIALS). If a step runs after such a secret is expanded, its JSON output may be flagged as containing secrets.

This could be fixed by moving the Version compatibility matrix before the Azure conditional check step. Let me try that.

ahus1:
> my first sight suspicion is that GHA masks each line of multiline secrets

I agree with that. 

Maybe moving things could fix it, but having GHA escaping curly braces is still problematic, so I would consider it a workaround. 

AFAIK the AZURE_CREDENTIALS are managed by the cloud native team, I'd say it is in the cloud native area to handle this.

mabartos:
@ahus1 Sorry, missed the part about the Azure 😝 Changing to CN team as Peter also assigned it to himself.

mabartos:
@Pepo48 Could you please triage this?

Pepo48:
[...]

PR #45779, opened 2026-01-27
mixed-cluster-compatibility-tests fail due to incorrectly masked cont…

Pepo48:
…ent in 26.5 branch

Closes: #45745

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,13 @@ permissions:
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


Pepo48:
Here are the results: https://github.com/Pepo48/keycloak/actions/runs/21390808289/job/61586020528

pruivo:
@Pepo48 LGTM! We probably need the same change in the main branch, so when the release 26.6 branch is created, we don't have the same issue.

pruivo (approved):

ahus1 (approved):

PR #45783, opened 2026-01-27
feat: add ability to hide social broker logins in flow

Spliterash:
I would like to be able to hide social brokers on the login page at the flow level, not just globally. This would allow me to implement hiding for individual clients.

Implementing this as an authenticator will also allow the use of conditional authenticators, which will add additional flexibility.


ahus1:
@Spliterash - as outlined in the CONTRIBUTING.md file, please first create a GitHub enhancement issue first describing your enhancement and link it here.

PR #45784, opened 2026-01-27
[OID4VCI] Add OID4VCI request/response support to OAuthClient utility

### LINKED ISSUE #44671: Add OID4VCI related requests/responses to OAuthClient
### Description

Make sure that OID4VCI related requests/responses are supported by OAuthClient utility, which is used in the testsuite. It should be possible to use them from here instead of manually dealing with Apache HTTP client.

forkimenjeckayang:
This PR adds OID4VCI-related request and response utilities to the OAuthClient test utility, allowing OID4VCI tests to use the standard OAuthClient API instead of direct Apache HTTP client calls.

**Changes:**
- Added OID4VCI request/response classes (CredentialRequest, CredentialOfferRequest, CredentialIssuerMetadataRequest, etc.)
- Refactored all OID4VCI tests to use the new OAuthClient utilities
- Enhanced error handling in credential offer URI requests to include both error code and description

closes https://github.com/keycloak/keycloak/issues/44671


tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractHttpPostRequest.java
```diff
@@ -128,7 +135,6 @@ protected String getAccept() {
 
     protected abstract R toResponse(CloseableHttpResponse response) throws IOException;
 
-    @SuppressWarnings("unchecked")
```
mposolda, line 131: Not sure why this line is removed? It adds the compiler "warning" . It seems rather better option to add the same on `AbstractHttpGetRequest.get` ?

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractHttpGetRequest.java
```diff
@@ -1,29 +1,58 @@
...
     public AbstractHttpGetRequest(AbstractOAuthClient<?> client) {
         this.client = client;
+        this.headers.put("Accept", MediaType.APPLICATION_JSON);
     }
 
     protected abstract String getEndpoint();
 
     protected abstract void initRequest();
 
+    public T endpoint(String endpoint) {
```
mposolda, line 37: Is it possible to add javadoc to this method and explain that when this is used, it has preference over `getEndpoint()` etc? Same applies for the same method on `AbstractHttpPostRequest`

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractHttpGetRequest.java
```diff
@@ -1,29 +1,58 @@
...
+        this.endpointOverride = endpoint;
+        return request();
+    }
+
+    public T bearerToken(String bearerToken) {
+        this.bearerToken = bearerToken;
+        return request();
+    }
+
+    public T header(String name, String value) {
```
mposolda, line 47: Note that introducing this method and renaming existing method to `addHeader` probably introduced testsuite failures as existing methods still use `header` method, but the behaviour is different as what was previously `header` is now new method `addHeader` ...

In general, I am not sure if we need 2 dedicated methods? My preference would be stick with single method `header(String name, String value)` and maybe implement with something like:
```
public header(String name, String value) {
    if (value != null) {        
        this.headers.put(name, value);        
    } 
}
```
but instead make sure that current method `initRequest` is moved to line 47 (before headers from the `headers` variable are set to actual Apache HTTP client `get` method). As `initREquest` often adds some headers though...

In other words, pretty much same logic like current `AbstractHttpPostRequest` is using where `headers` are put to Apache HTTP client `post` method after `initRequest` is invoked.

WDYT?

If it is really needed to have 2 methods (`addHeader` and `header`), I suggest to add javadoc to this method and also to the method `addHeader` and explain the difference between them.

mposolda (commented):
@forkimenjeckayang Thanks! I've added some minor comments inline (I also think they are related to the failing tests). But in general, I hope we can move with your PR soon.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractHttpGetRequest.java
forkimenjeckayang: Thank you for the suggestion @mposolda . Refactored to use single `header()` method with null check and moved `initRequest()` before headers application. Now matches `AbstractHttpPostRequest` pattern.

mposolda (approved):
@forkimenjeckayang Thanks!

PR #45785, opened 2026-01-27
introduced related colors to the theme tab

edewit:
This reduces the number of colors that the user has to set to make his
own theme, by having hover and or other colors be derived from a parent
color

fixes: #45527

### DIFF
--- a/js/apps/admin-ui/src/realm-settings/themes/ColorControl.tsx
+++ b/js/apps/admin-ui/src/realm-settings/themes/ColorControl.tsx
@@ -1,0 +1,50 @@
+import {
+  TextInputProps,
+  InputGroup,
+  InputGroupItem,
+} from "@patternfly/react-core";
+import { TextControl } from "@keycloak/keycloak-ui-shared";
+import { useFormContext, useWatch } from "react-hook-form";
+import { useTranslation } from "react-i18next";
+
+export type ColorControlProps = TextInputProps & {
+  name: string;
+  label: string;
+  color: string;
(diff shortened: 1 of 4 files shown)


edewit:
image
Changing the primary color will also change the dependent colors with a css "trick" to make them in the same tone, but there is still a possibility to override these defaults

mabartos:
@edewit Looks nice, but I found one bug. When you want to override the secondary color, even the primary color is changed.


https://github.com/user-attachments/assets/07cef960-f8b2-4ca4-9a0c-9a03314c95a6

edewit:
@mabartos this is because in PF in light mode the secondary color and the primary color use the same variable, but that's not the case in dark mode. So I can't really do anything about that other then removing the secondary color selector for light mode

mabartos:
> @mabartos this is because in PF in light mode the secondary color and the primary color use the same variable, but that's not the case in dark mode. So I can't really do anything about that other then removing the secondary color selector for light mode

Maybe it makes sense to just filter it out, because then it's very confusing. @edewit WDYT?

edewit:
@mabartos create an issue for patternfly and commented it out for now

mabartos:
@edewit Could you please resolve the conflict? Thanks!

mabartos (approved):
@edewit Looks very good!

mabartos:
@vmuzikar Could you please also review/merge this PR? Thanks!

pedroigor (approved):

PR #45787, opened 2026-01-27
Issuer 45649: patch CVE-2026-0707. Add validation on Authorization Header with Bearer

### LINKED ISSUE #45649: CVE-2026-0707: Keycloak Authorization Header Parsing Leading to Potential Security Control Bypass
### Source: https://issues.redhat.com/browse/RHBK-4088
Security Tracking Issue

Flaw:
-----

Keycloak Authorization Header Parsing Leading to Potential Security Control Bypass

Keycloak’s authentication pipeline excessively tolerates non-standard Bearer token formats (case variations, Tab characters, multiple spaces, mixed whitespace) in the Authorization header, creating inconsistencies with front-end security controls (WAF/proxies) and enabling potential bypass risks.

---
*This issue was originally tracked in the private repository as #309. Migrated via lift-embargo process.*

ValeriaEpifanova:
Closes [#45649](https://github.com/keycloak/keycloak/issues/45649)

Add validation on Authorization Header to fix https://github.com/advisories/GHSA-gv94-wp4h-vv8p:

- case-insensitive validation of the Bearer scheme
- strict enforcement of a single ASCII space separator
- rejection of non-standard whitespaces
- validate token value by [RFC6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1):

```
b64token    = 1*( ALPHA / DIGIT /"-" / "." / "_" / "~" / "+" / "/" ) *"="
credentials = "Bearer" 1*SP b64token
```

### DIFF
--- a/services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
+++ b/services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
@@ -17,7 +17,6 @@
 package org.keycloak.services.managers;
 
 import java.util.List;
-import java.util.regex.Pattern;
 
 import jakarta.ws.rs.NotAuthorizedException;
 import jakarta.ws.rs.core.HttpHeaders;
@@ -43,8 +42,6 @@ public class AppAuthManager extends AuthenticationManager {
 
     public static final String BEARER = "Bearer";
 
-    private static final Pattern WHITESPACES = Pattern.compile("\\s+");
-
     @Override
     public AuthResult authenticateIdentityCookie(KeycloakSession session, RealmModel realm) {
         AuthResult authResult = super.authenticateIdentityCookie(session, realm);
@@ -66,26 +63,28 @@ public class AppAuthManager extends AuthenticationManager {
             return null;
         }
 
-        String[] split = WHITESPACES.split(authHeader.trim());
-        if (split.length != 2){
+        int indexOfSpace = authHeader.indexOf(' ');
+
+        if (indexOfSpace <= 0) {
             return null;
         }
 
-        String typeString = split[0];
+        String typeString = authHeader.substring(0, indexOfSpace);
+        String tokenString = authHeader.substring(indexOfSpace + 1);
 
+        boolean isBearerHeader = typeString.equalsIgnoreCase(BEARER);
         if (!Profile.isFeatureEnabled(Profile.Feature.DPOP)) {
-            if (!typeString.equalsIgnoreCase(BEARER)) {
+            if (!isBearerHeader) {
                 return null;
             }
         } else {
             // "Bearer" is case-insensitive for historical reasons. "DPoP" is case-sensitive to follow the spec.
-            if (!typeString.equalsIgnoreCase(BEARER) && !typeString.equals(TokenUtil.TOKEN_TYPE_DPOP)){
+            if (!isBearerHeader && !typeString.equals(TokenUtil.TOKEN_TYPE_DPOP)) {
                 return null;
             }
         }
 
-        String tokenString = split[1];
-        if (ObjectUtil.isBlank(tokenString)) {
+        if (ObjectUtil.isBlank(tokenString) || tokenString.contains(" ")) {
             return null;
         }
 
--- a/tests/base/src/test/java/org/keycloak/tests/admin/AppAuthManagerTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/admin/AppAuthManagerTest.java
@@ -1,0 +1,129 @@
+package org.keycloak.tests.admin;
+
+import java.io.IOException;
+
+import jakarta.ws.rs.core.Response;
+
+import org.keycloak.models.AdminRoles;
+import org.keycloak.models.Constants;
+import org.keycloak.testframework.annotations.InjectHttpClient;
+import org.keycloak.testframework.annotations.InjectKeycloakUrls;
+import org.keycloak.testframework.annotations.InjectRealm;
+import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
+import org.keycloak.testframework.oauth.OAuthClient;
+import org.keycloak.testframework.oauth.annotations.InjectOAuthClient;
+import org.keycloak.testframework.realm.ManagedRealm;
+import org.keycloak.testframework.realm.RealmConfig;
+import org.keycloak.testframework.realm.RealmConfigBuilder;
+import org.keycloak.testframework.server.KeycloakUrls;
+import org.keycloak.testsuite.util.oauth.AccessTokenResponse;
+
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.jupiter.api.Assertions;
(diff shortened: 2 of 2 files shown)


services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
```diff
@@ -64,26 +64,40 @@ private static AuthHeader extractTokenStringFromAuthHeader(String authHeader) {
...
+
+        while (i < len) {
+            char c = authHeader.charAt(i);
+            if (c == ' ') {
+                break;
+            }
+            i++;
+        }
+
+        if (i == 0 || i == len) {
```
rmartinc: Just use `String.indexOf` to get the first space instead of manually looping. If the result is  <0 you can just return null.

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
```diff
@@ -64,26 +64,40 @@ private static AuthHeader extractTokenStringFromAuthHeader(String authHeader) {
...
                 return null;
             }
         }
 
-        String tokenString = split[1];
-        if (ObjectUtil.isBlank(tokenString)) {
+        if (ObjectUtil.isBlank(tokenString) || tokenString.contains(" ")) {
+            return null;
+        }
+        if (!BEARER_TOKEN_PATTERN.matcher(tokenString).matches()) {
```
rmartinc: I don't think you need the `BEARER_TOKEN_PATTERN`, if the token string is not OK it would fail in the validation. I would maintain the current code in this part without any regex.

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
```diff
@@ -64,26 +64,40 @@ private static AuthHeader extractTokenStringFromAuthHeader(String authHeader) {
...
+
+        if (i == 0 || i == len) {
             return null;
         }
 
-        String typeString = split[0];
+        String typeString = authHeader.substring(0, i);
+        String tokenString = authHeader.substring(i + 1);
 
+        boolean isBearerHeader = typeString.equalsIgnoreCase(BEARER);
```
rmartinc, line 75: Following the report we should check this `typeString.equals(BEARER)` to be compliant. Besides you don't need to do this before, you can keep the comparison inside the ifs. It's always just executed once, not twice.

rmartinc (changes requested):
Thanks @ValeriaEpifanova for the PR! I added some suggestions and besides we will need tests for this.

I created a quick test using the admin `whoAmI` endpoint in [this class](https://github.com/rmartinc/keycloak/blob/test-bearer/tests/base/src/test/java/org/keycloak/tests/admin/AppAuthManagerTest.java). You can create more methods that test other prefixes like commented in the issue (`BEARER `, `Bearer   `, `Bearer\t`,...).

There is only snag for this, current implementations that uses some spaces or different case will fail after this PR. I don't know if we need a configuration to revert to the previous wrong behavior or not. We have used previously configurations in the OIDC factory to allow old behaviors (for example in this commit https://github.com/keycloak/keycloak/commit/fa383bf76ca3581f271653deccc159123bec00f3). @mposolda WDYT? Do we allow previous behavior with a switch or do we just follow the spec?

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
rmartinc: Forget about the `equalsIgnoreCase`, I didn't read your comment in the issue. And I think you are right, so just continue using the ignore case version.

rmartinc:
If the change is just the space I would go without any configuration switch. So just add the test class and check that several spaces or a tab is not valid. Let's see what the maintainers think about this. Thanks @ValeriaEpifanova!

mposolda:
@ValeriaEpifanova @rmartinc +1 that bearer should be case-insensitive. It probably does not increase security to treat it case-sensitive, but rather introducing issues. We had already issues in the past when people using `Authorization: bearer` and they had issues (AFAIR some of the standard OIDC adapters used `bearer` at some point... ).

Hopefully we can go without config option on this. +1 to just have a test and should be good to go IMO.

ValeriaEpifanova:
Thanks for the review @rmartinc @mposolda . I've added tests. I took the class you @rmartinc  created for this purpose and added more tests. I've also fixed the code, but I've left the header validation case-insensitive.
I didn't add a test where there's a space before Bearer ` Bearer`. I assume that RestEasy itself might be using trim() under the hood. This request returns a 200 status.

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
```diff
@@ -66,26 +64,28 @@ private static AuthHeader extractTokenStringFromAuthHeader(String authHeader) {
...
             // "Bearer" is case-insensitive for historical reasons. "DPoP" is case-sensitive to follow the spec.
-            if (!typeString.equalsIgnoreCase(BEARER) && !typeString.equals(TokenUtil.TOKEN_TYPE_DPOP)){
+            if (!isBearerHeader && !typeString.equals(TokenUtil.TOKEN_TYPE_DPOP)) {
                 return null;
             }
         }
 
-        String tokenString = split[1];
-        if (ObjectUtil.isBlank(tokenString)) {
+        if (ObjectUtil.isBlank(tokenString) || tokenString.contains(" ")) {
```
rmartinc, line 87: Is there any difference adding the `tokenString.contains(" ")` or not adding it?

rmartinc (approved):
Thanks @ValeriaEpifanova! I think that it's OK, just a minor comment but I don't think that it is important. And it's usually better doing rebase instead of merge to update the branch. @mposolda please launch the tests and check if this is OK to you.

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
ValeriaEpifanova: There is a difference. For example, "Bearer 2SP token" is not blank, so the code would continue processing the token. If next space after first was before token value. Token value will be " token".

ValeriaEpifanova:
Got it, thanks  @rmartinc  — I’ll keep your advice about using rebase instead of merge in mind for the future.

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
rmartinc: But I suppose ` token` would fail because it's not a valid token for the space at the beginning. But no problem, now we have tests, so we are covered.

ValeriaEpifanova: You're right, and at first I thought the same thing. But the point of the CVE is that, according to the RFC, there must be strictly one space between Bearer and token.
`credentials = "Bearer" 1*SP b64token`

rmartinc:
Hey @ValeriaEpifanova! The build is failing because the imports, just run `mvn spotless:apply` to fix the imports. Thanks!

ValeriaEpifanova:
Done @rmartinc 
Thanks for help! I think we can start it again

mposolda (approved):
@ValeriaEpifanova @rmartinc Thanks!

PR #45789, opened 2026-01-27
added ability to refresh token when within time

### LINKED ISSUE #44379: Node.js admin client does not refresh tokens
### Before reporting an issue

### Area

admin/client-js

### Describe the bug

The Node.js admin client obtain an initial access token, but does not check or refresh this token later. This will cause issues when using the admin client for a longer period of time.

As an example this causes the Account E2E tests to be flaky as they sometimes can take longer to execute than the validity of the token. Especially in private repositories.

### Version

main

### Regression

### Expected behavior

Admin client automatically refreshes the access token if it is close to expiring. A default of 5 seconds minimum validity would be sensible.

### Actual behavior

The admin client never refreshes the access token. A custom token provider can be used to do this, but that is  not easy to implement.

### How to Reproduce?

Run Account E2E tests in a private repository, or simple call `auth` then wait for more than 5 minutes before calling any operations through the admin endpoints.

### Anything else?

_No response_

edewit:
fixes: #44379

### DIFF
--- a/js/libs/keycloak-admin-client/src/client.ts
+++ b/js/libs/keycloak-admin-client/src/client.ts
@@ -16,8 +16,9 @@ import { ServerInfo } from "./resources/serverInfo.js";
 import { Users } from "./resources/users.js";
 import { UserStorageProvider } from "./resources/userStorageProvider.js";
 import { WhoAmI } from "./resources/whoAmI.js";
-import { Credentials, getToken } from "./utils/auth.js";
+import { Credentials, getToken, Settings } from "./utils/auth.js";
 import { defaultBaseUrl, defaultRealm } from "./utils/constants.js";
+import { DecodedToken, decodeToken } from "./utils/decode.js";
 
 export type RequestOptions = Omit<RequestInit, "signal">;
 
@@ -32,6 +33,8 @@ export interface ConnectionConfig {
   requestArgOptions?: Pick<RequestArgs, "catchNotFound">;
   timeout?: number;
 }
+
+const MIN_VALIDITY = 5; // in seconds
 
 export class KeycloakAdminClient {
   // Resources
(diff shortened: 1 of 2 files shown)


stianst (approved):

Gr3yShad0w:
Not an expert but I suspect this PR breaks previous (26.5.2) working setups. I have a Service account which has no refresh tokens because its using the grant type `client_credentials`. Now it aggressivly tries to decode the refresh token and crashes because it cannot find one

```
file:///app/node_modules/.aspect_rules_js/@keycloak+keycloak-admin-client@26.5.3/node_modules/@keycloak/keycloak-admin-client/lib/utils/decode.js:2
    const [, payload] = token.split(".");
                              ^

TypeError: Cannot read properties of undefined (reading 'split')
    at decodeToken (file:///app/node_modules/.aspect_rules_js/@keycloak+keycloak-admin-client@26.5.3/node_modules/@keycloak/keycloak-admin-client/lib/utils/decode.js:2:31)
    at KeycloakAdminClient.setRefreshToken (file:///app/node_modules/.aspect_rules_js/@keycloak+keycloak-admin-client@26.5.3/node_modules/@keycloak/keycloak-admin-client/lib/client.js:109:37)
```

OAuth 2.0 RFC6749 Section 4.4.3 states that a refresh_token should not be generated when client_credentials grant is used

Monkey Patch that does not crash 

```
const kcAdminClient = new KcAdminClient({
  baseUrl: "....",
  realmName: "....",
});
const originalSetRefreshToken =
  typeof kcAdminClient.setRefreshToken === "function"
    ? kcAdminClient.setRefreshToken.bind(kcAdminClient)
    : undefined;
if (originalSetRefreshToken) {
  kcAdminClient.setRefreshToken = ((token?: string) => {
    if (!token) return;
    return originalSetRefreshToken(token);
  }) as typeof kcAdminClient.setRefreshToken;
}
```

ahus1:
@Gr3yShad0w - instead of commenting on  a closed issue, it is better to create a new bug issue. Thank you to @Scrumplex for creating issue #46382 for this.

PR #45793, opened 2026-01-27
[OID4VCI] Support AuthorizationRequest requests in OAuthClient

### LINKED ISSUE #45774: [OID4VCI] Support AuthorizationRequest requests in OAuthClient
### Description

Currently, oid4vci client side testing relies heavily on direct use of Apache HttpClient (i.e. the low-level http client). Other standard OAuth2 request are already nicely abstracted in [OAuthClient](https://github.com/keycloak/keycloak/blob/main/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractOAuthClient.java) which handles protocol message requests and expected responses (with error handling)

Having similar support for oid4vci requests would greatly improve client side test readability and allow us to reason about oid4vci messages rather than low-level http requests. Details like http method, parameters, headers, error conditions would all be handled by OAuthClient.

In a following step, we could introduce the notion of oid4vci wallet (initially just for testing). The wallet could then bind multiple OAuthClient request together into sequences, just like an oid4vci spec compliant wallet would do. This would concisely document the scope of wallet functionality that is currently supported from an oid4vci wallet perspective.

### Value Proposition

Add support for 

* CredentialOfferRequest 
* AuthorizationRequest (with AuthorizationDetails, ClientDetails, etc)
* AccessTokenRequest (i.e. oauth code already existes, pre-auth code is missing)
* CredentialRequest

### Goals

* improved client side testing for oid4vci
* covers all known oid4vci request types
* pre-requisite for #31674
* pre-requisite for oid4vci (test) wallet abstraction

### Non-Goals

No oid4vci request composition (just yet)

### Discussion

_No response_

### Notes

_No response_

tdiesler:
closes #45774

Instead of using low level Http requests, this PR adds support for OID4VCI requests  using the well known OAuthClient request/response pattern like this ... 

### Rich AuthorizationRequest with AuthorizationDetails

This PR also adds support for ClientMetadata in the AuthorizationRequest

```
        AuthorizationRequest authRequest = new AuthorizationRequestBuilder()
                .withClientId(issClientId)
                .withAuthorizationDetail(authDetail)
                .withRedirectUri(redirectUri)
                .withCodeChallenge(pkce)
                .build();

        AuthorizationRequestResponse authResponse = new AuthorizationRequestRequest(oauth, authRequest)
                .credentials(ctx.targetUser, "password")
                .send();
```

If removes `OID4VCAuthorizationDetailResponse` in favor of `OID4VCAuthorizationDetail` with `credential_identifiers` (i.e. it should not be necessary to distinguish between the two)

Response status code checking is already done in the response base classes - I unified the error response handling like this, so that clients won't have to deal with it every time as part of a request.

```
    public CredentialResponse getCredentialResponse() {
        return Optional.ofNullable(credentialResponse).orElseThrow(() ->
                new IllegalStateException(String.format("[%s] %s", getError(), getErrorDescription())));
    }
```

`AbstractHttpPostRequest` now supports ...

* arbitrary payloads (i.e. not just url-form-encoded)
* bearer token authentication

`Oid4vcCredentialRequest` can extend it directly.


tdiesler:
Includes #45913

tdiesler:
out of date

PR #45806, opened 2026-01-27
Update documents to reflect that they are linux-specific

areevesat:
Pull Request for #45805.

This pull request updates a few readme files to reflect that the shell commands are written for Linux, and suggests using a linux-like shell for users running in Windows.

This PR only contains documentation updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ Alternatively, you can use the Docker image by running:
     
 For more details refer to the [Keycloak Documentation](https://www.keycloak.org/documentation.html).
 
+Note that shell commands throughout the readmes and documentation are written for Linux.
+If running in Windows, use a linux-like shell such as Git Bash.
+
 
 ## Building from Source
 
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,5 +1,7 @@
 ## Building from source
 
+Note that the shell commands below are written for Linux.  If running in Windows, use a linux-like shell such as Git Bash.
+
 Ensure you have **JDK 17** or **JDK 21** and Git installed
 
     java -version
--- a/quarkus/README.md
+++ b/quarkus/README.md
@@ -2,6 +2,8 @@
 
 ## Before you start
 
+Note that the shell commands below are written for Linux.  If running in Windows, use a linux-like shell such as Git Bash.
+
 This section describes how to run Keycloak in *Quarkus development mode*. Before continuing, make sure you have already followed the main project [build
 and prerequisites](https://github.com/keycloak/keycloak/blob/main/docs/building.md) documentation, which explains all required tools (JDK, Maven, etc.).
 


ahus1:
Thank you for this PR. For contributions, we ask to sign-off commits, see CONTRIBUTING.md for details. 

Please update your PR.

areevesat:
I have updated the PR to include the signoff in the commit message, as well as to include the note of which issue it closes in the message.  Please let me know if anything else is required.

Pepo48 (changes requested):
Hi @areevesat, first of all, thank you for the contribution and for taking the time to improve the documentation.

I think the intent behind this change makes sense, but I’m not sure the message itself is quite right. Keycloak is a multi-platform project, and our goal is not to encourage users to rely on tools like Git Bash, which are effectively compatibility layers on top of Windows (MSYS2). Instead, we explicitly aim to ensure that everything can be done using native Windows shells such as PowerShell or CMD, just as it can be done on Unix-like systems with Bash.

That said, I completely agree that there is significant room for improvement across our READMEs and documentation when it comes to cross-platform clarity. This is an area where the project could definitely do better.

Specifically for the files touched by this PR, the vast majority of commands already work on Windows with only minor, well-known adaptations, for example:

- `export VAR=value` → `set VAR=value` in CMD or `$env:VAR="value"` in PowerShell  
- `VAR=value command` (inline env vars) → set the variable first, then run the command  

These are fairly standard cross-platform differences, and in general developers are already used to them.

The broader question, then, is whether we want to adopt a more systematic approach to this across the documentation. For example, Quarkus uses mechanisms like AsciiDoc tab switching to present all sorts of variants where it makes sense. If we were to do something like that for OS-specific variants, it would probably be more impactful to apply it consistently in areas such as `containers.adoc`, `reverse-proxy.adoc`, and similar guides, rather than addressing individual cases in isolation.

We’ve discussed this internally before, but since there hasn’t been strong demand or many complaints from the community, we decided to put it on hold for now. That said, if you’d like to repurpose this PR or propose a more general approach, we’d be happy to discuss it further.

Thanks again for the contribution and for engaging with the project.

areevesat:
@Pepo48 

Thank you for getting back to me.  My primary motivation to create this issue and PR in the first place was the frustration and hours of troubleshooting I ran into while trying to set up my environment for initial debugging.  See my discussion https://github.com/keycloak/keycloak/discussions/44921 (I still am not able to actually debug Keycloak locally, likely due to unknown setup issues).  At least one maven command did not function in PowerShell, and only worked once I swapped to Git Bash.  It does seem that there is still some cross-platform compatibility needed there.  Certainly I could modify the messaging to not mention a specific tool and instead just reference a class of tools or general approach.  My goal was simply to help the next person avoid the pitfall of running the commands as-is in an environment where they won't work.

I think any solution that assists new developers in getting started with the project would be of benefit to onboarding.  In my opinion, the instructions, especially the getting started instructions, ought to work out of the box without relying on the developer to already know where they need to deviate from the instructions.  I think it is likely that you don't receive complaints because most people who have difficulty would just give up rather than raise a fuss.

That said, I do not think I have the expertise or familiarity with the project to make a broader solution.  I could repurpose the original issue, but I think you and the rest of the existing community would be better able to make any larger-scale changes in a manner consistent with this project's style and principles.

Pepo48:
@areevesat is this what you've tried to achieve specifically on Windows?
image

I think you can clearly grasp everything I did from the screenshot, but here's a recap, just in case:

1. build the dist from the source code or download it from the website
2. take the dist archive and extract it
3. run the PowerShell from the bin folder
4. execute `$env:DEBUG_SUSPEND = "y"`
5. start Keycloak via `./kc.bat --debug start-dev`
6. Set a breakpoint to a function you want do debug (authenticate() in my case)
7. Attach a debugger to the process in IDE (Intellij IDEA in my case)
8. Perform the action in a browser

As you can see the execution stopped exactly at the breakpoint and watched variables are initiated.

Pepo48:
@areevesat alternatively, if you want to run everything from the IDE, you can follow what @mabartos already advised in https://github.com/keycloak/keycloak/discussions/44921#discussioncomment-15270238 - run the `Keycloak.java` class in the IDE in debug mode from within the `keycloak-junit5` module. In that case, all you need to do is set a breakpoint and perform the action you want to debug. There is no need to attach a debugger, and no special parameters are required.

Just to be completely sure, I tested this on Windows as well - everything works the same as shown in the screenshot above.

areevesat:
@Pepo48 

When I attempt ./kc.bat --debug start-dev, it hangs indefinitely right after "Listening for transport dt_socket at address: 8787".  Running ./kc.sh --debug start-dev in Git Bash appears to work.

When I followed the instructions mabartos mentioned, I was unable to locate the XML settings he referenced.  And ran into the problems I mentioned in a comment in that discussion, with missing CSS and no database to persist data between runs.

Pepo48:
> When I attempt ./kc.bat --debug start-dev, it hangs indefinitely right after "Listening for transport dt_socket at address: 8787". Running ./kc.sh --debug start-dev in Git Bash appears to work.

@areevesat do you use `$env:DEBUG_SUSPEND = "y"` or not? If so, it gives you the opportunity to suspend Keycloak even before the server starts. It’s important to realize that, by default, the start command spins up two separate JVMs: one responsible for the build step and the second for starting the server itself.

Because of that, you either need to stop the debugger and attach it again, or simply not set `$env:DEBUG_SUSPEND` at all if you don’t need to debug the build phase.

The XML settings are just a reference to show roughly what needs to be configured, especially if you’re trying to run this from an IDE other than IntelliJ. You likely can’t use them directly in the Eclipse IDE you mentioned, but I assume the steps to run a class in debug mode are very similar.

areevesat:
@Pepo48 

Thank you for your patience and working with me on this.  I was able to get the debugging process working by running the quarkus dist from the command line, then attaching the IntelliJ IDEA debugger to.

Your instructions were very helpful, but they don't seem to map to the instructions in https://github.com/keycloak/keycloak/blob/main/quarkus/README.md.  I'd be happy to update my original issue (#45805) with more ideas about what documentation could be improved.

Thank you again for helping me with the debugging steps.

Pepo48:
> Your instructions were very helpful, but they don't seem to map to the instructions in quarkus/README.md.

@areevesat bear in mind that right next to `README.md` there is also `CONTRIBUTING.md`, which contains a *Debugging the server distribution* section. That section explains the process and explicitly mentions and describes the usage of the `DEBUG_SUSPEND` variable as well.

In practice, it doesn’t really matter which form of the server you attach the debugger to - whether you built or downloaded the distribution, or whether you use the embedded server or `IDELauncher` in older versions of Keycloak. There are multiple ways to achieve the same goal, depending on your needs and workflow. At a high level, the approach is always the same: run the distribution and attach the debugger to the running server process.

areevesat:
@Pepo48 
Perhaps quarkus/readme.md should have a link to contributing.md?  It appears to include debugging instructions that didn't work for me, and no indication to go looking for another file for instructions that do work.  It has a link back to the main keycloak contributing.md.

Pepo48:
@areevesat thanks again for raising this and suggesting the improvements. 

I'm closing this PR as it was superseded by https://github.com/keycloak/keycloak/pull/46033.

Feel free to reach us out in case of any questions.

PR #45807, opened 2026-01-27
[OID4VCI] Inconsistencies in well-known OID4VC metadata (Same metadata for all formats) #45485

### LINKED ISSUE #45485: [OID4VCI] Inconsistencies in well-known OID4VC metadata (Same metadata for all formats)
### Before reporting an issue

### Area

oid4vc

### Describe the bug

I can see those inconsistencies in the metadata returned from the OID4VCI credential endpoint (For example https://as.keycloak-fapi.org:8443/.well-known/openid-credential-issuer/realms/test in case that realm name is "test" ):

- When I create the client scope with supported format configured to "JWT VC", I can see the output similar to this one (including just newly created credential metadata `fff-vct` and existing `oid4vc_natural_person` credential metadata):

```
    "fff-vct": {      
      "format": "jwt_vc",
      "scope": "fff",
      "cryptographic_binding_methods_supported": [
        "jwk"
      ],
      "credential_signing_alg_values_supported": [
        "PS384",
        ...
      ],
      "vct": "fff",
      "credential_definition": {
        "type": [
          "fff"
        ],
        "@context": [
          "fff"
        ]
      },
....
    "oid4vc_natural_person": {
      "id": "oid4vc_natural_person",
      "format": "dc+sd-jwt",
      "scope": "oid4vc_natural_person",
      "cryptographic_binding_methods_supported": [
        "jwk"
      ],
      "credential_signing_alg_values_supported": [
        "PS384",
...
      ],
      "credential_definition": {
        "type": [
          "oid4vc_natural_person"
        ],
        "@context": [
          "oid4vc_natural_person"
        ]
      },
```

For both formats `jwt_vc` and `dc+sd-jwt` , I can see both options `vct` and `credential_definition` are present. It looks this is not aligned with the specs as:

- Section [ A.3 of the specification ](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.3) mentions `vct` as required claim in the metadata, however claim `credential_definition` is not mentioned for `dc+sd-jwt` at all. Hence Keycloak likely should not adverse it for this format?
- Section [ A.1 of the specification ](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.1) mentions `credential_definition` but not `vct` . Hence Keycloak likely should not adverse `vct` for this format? Also note that for example A.1.1 mentions only `type` claim within `credential_definition` , but not `@context` . (Right now, it seems to me that suppoerting just `jwt_vc_json` format documented in A.1.1 might be easiest option initially, see related issue #45486 )

It seems to me it can be good if we have some hook (maybe method on the `CredentialBuilder` interface), which will allow us to create the metadata for the particular OID4VC credential slightly differently for the various different formats according to requirements for various formats.

### Version

nightly (from 2026-01-15)

### Regression

### Expected behavior

Metadata announced by OID4VC well-known endpoint are aligned with the specification for various formats

### Actual behavior

Metadata announced by OID4VC well-known endpoint are not aligned with the specification for various formats

### How to Reproduce?

See details

### Anything else?

_No response_

namanONcode:
This pull request enhances the flexibility and correctness of the OID4VC issuer metadata by allowing credential builders to contribute format-specific metadata to the well-known endpoint. It introduces a new extension point in the `CredentialBuilder` interface, updates the issuer provider to use it, and adds targeted tests to ensure correct metadata exposure for different credential formats.

**Format-specific metadata contribution:**

* Added a new `contributeToMetadata` method to the `CredentialBuilder` interface, enabling each credential format to customize the metadata it exposes in the well-known endpoint. Default implementation is a no-op for backward compatibility.
* Implemented `contributeToMetadata` in `JwtCredentialBuilder` and `SdJwtCredentialBuilder` to ensure only relevant fields (`credential_definition` or `vct`) are present in the metadata for each format. [[1]](diffhunk://#diff-b3e591246d66e1bdc3fe6c9c5a553775fe0f13aa53cac68611d0ca10683bda25R108-R112) [[2]](diffhunk://#diff-ca6ba74d48a4c052114721df71fe13346ea8b92dfa0fa9bcb834cf2fe5b93280R137-R141)

**Provider logic updates:**

* Updated `OID4VCIssuerWellKnownProvider` to invoke `contributeToMetadata` for each supported credential configuration, using the factory pattern to obtain the correct builder for each format. This ensures that only format-appropriate metadata fields are included. [[1]](diffhunk://#diff-a8598fb792303a2cce62c2affdaede4d96ff4c324ee3b6b9d65ab342ae76edc9R50) [[2]](diffhunk://#diff-a8598fb792303a2cce62c2affdaede4d96ff4c324ee3b6b9d65ab342ae76edc9L463-R515)

**Testing:**

* Added `CredentialMetadataWellKnownTest` to verify that the well-known endpoint exposes the correct metadata for SD-JWT and JWT-VC credential types, ensuring that only the appropriate fields are present for each format.

These changes improve extensibility, ensure compliance with OID4VCI metadata requirements, and prevent leaking irrelevant metadata fields for unsupported formats.

Closes #45485

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
@@ -47,6 +47,7 @@ import org.keycloak.models.RealmModel;
 import org.keycloak.models.oid4vci.CredentialScopeModel;
 import org.keycloak.protocol.oid4vc.OID4VCLoginProtocolFactory;
 import org.keycloak.protocol.oid4vc.issuance.credentialbuilder.CredentialBuilder;
+import org.keycloak.protocol.oid4vc.issuance.credentialbuilder.CredentialBuilderFactory;
 import org.keycloak.protocol.oid4vc.model.CredentialIssuer;
 import org.keycloak.protocol.oid4vc.model.CredentialRequestEncryptionMetadata;
 import org.keycloak.protocol.oid4vc.model.CredentialResponseEncryptionMetadata;
@@ -459,24 +460,54 @@ public class OID4VCIssuerWellKnownProvider implements WellKnownProvider {
                 keycloakSession.clientScopes()
                         .getClientScopesByProtocol(realm, OID4VCIConstants.OID4VC_PROTOCOL)
                         .map(CredentialScopeModel::new)
-                        .map(clientScope -> {
-                            return SupportedCredentialConfiguration.parse(keycloakSession,
-                                    clientScope,
+                        .map(credentialScope -> {
+                            SupportedCredentialConfiguration config = SupportedCredentialConfiguration.parse(keycloakSession,
+                                    credentialScope,
                                     globalSupportedSigningAlgorithms
                             );
+                            applyFormatSpecificMetadata(keycloakSession, config, credentialScope);
+                            return config;
                         })
                         .collect(Collectors.toMap(SupportedCredentialConfiguration::getId, sc -> sc, (sc1, sc2) -> sc1));
 
         return supportedCredentialConfigurations;
     }
 
+
+    private static void applyFormatSpecificMetadata(KeycloakSession keycloakSession,
+                                                     SupportedCredentialConfiguration config,
+                                                     CredentialScopeModel credentialScope) {
+        String format = config.getFormat();
+        if (format == null) {
+            return;
+        }
+
+        // Find the CredentialBuilder for this format using the factory pattern
+        CredentialBuilder credentialBuilder = keycloakSession.getKeycloakSessionFactory()
+                .getProviderFactoriesStream(CredentialBuilder.class)
+                .map(factory -> (CredentialBuilderFactory) factory)
+                .filter(factory -> format.equals(factory.getSupportedFormat()))
+                .findFirst()
+                .map(factory -> factory.create(keycloakSession, null))
+                .orElse(null);
+
+        if (credentialBuilder == null) {
+            LOGGER.debugf("No CredentialBuilder found for format: %s", format);
+            return;
+        }
+
+        credentialBuilder.contributeToMetadata(config, credentialScope);
+    }
+
     public static SupportedCredentialConfiguration toSupportedCredentialConfiguration(KeycloakSession keycloakSession,
                                                                                       CredentialScopeModel credentialModel) {
         List<String> globalSupportedSigningAlgorithms = getSupportedAsymmetricSignatureAlgorithms(keycloakSession);
 
-        return SupportedCredentialConfiguration.parse(keycloakSession,
+        SupportedCredentialConfiguration config = SupportedCredentialConfiguration.parse(keycloakSession,
                 credentialModel,
                 globalSupportedSigningAlgorithms);
+        applyFormatSpecificMetadata(keycloakSession, config, credentialModel);
+        return config;
     }
 
     /**
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/CredentialBuilder.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/CredentialBuilder.java
@@ -17,7 +17,9 @@
 
 package org.keycloak.protocol.oid4vc.issuance.credentialbuilder;
 
+import org.keycloak.models.oid4vci.CredentialScopeModel;
 import org.keycloak.protocol.oid4vc.model.CredentialBuildConfig;
+import org.keycloak.protocol.oid4vc.model.SupportedCredentialConfiguration;
 import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.provider.Provider;
 
@@ -48,4 +50,24 @@ public interface CredentialBuilder extends Provider {
             VerifiableCredential verifiableCredential,
             CredentialBuildConfig credentialBuildConfig
     ) throws CredentialBuilderException;
+
+    /**
+     * Allows the credential builder to contribute format-specific metadata
+     * to the OID4VCI well-known credential issuer metadata.
+     *
+     * <p>
+     * Implementations should add only the metadata fields required by the
+     * supported credential format (for example {@code vct} for {@code dc+sd-jwt}
+     * or {@code credential_definition} for {@code jwt_vc_json}).
+     * </p>
+     *
+     * <p>
+     * The default implementation is a no-op to preserve backward compatibility.
+     * </p>
+     *
+     * @param credentialConfig the credential configuration to populate with format-specific metadata
+     * @param credentialScope  the credential scope model containing the source data
+     */
+    default void contributeToMetadata(SupportedCredentialConfiguration credentialConfig, CredentialScopeModel credentialScope) {
+    }
 }
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/JwtCredentialBuilder.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/JwtCredentialBuilder.java
@@ -24,11 +24,14 @@ import java.util.function.UnaryOperator;
 
 import org.keycloak.jose.jws.JWSBuilder;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.oid4vci.CredentialScopeModel;
 import org.keycloak.protocol.oid4vc.issuance.TimeClaimNormalizer;
 import org.keycloak.protocol.oid4vc.issuance.TimeProvider;
 import org.keycloak.protocol.oid4vc.model.CredentialBuildConfig;
+import org.keycloak.protocol.oid4vc.model.CredentialDefinition;
 import org.keycloak.protocol.oid4vc.model.CredentialSubject;
 import org.keycloak.protocol.oid4vc.model.Format;
+import org.keycloak.protocol.oid4vc.model.SupportedCredentialConfiguration;
 import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.representations.JsonWebToken;
 
@@ -103,5 +106,11 @@ public class JwtCredentialBuilder implements CredentialBuilder {
                 .jsonContent(jsonWebToken);
 
         return new JwtCredentialBody(jwsBuilder);
+    }
+
+    @Override
+    public void contributeToMetadata(SupportedCredentialConfiguration credentialConfig, CredentialScopeModel credentialScope) {
+        CredentialDefinition credentialDefinition = CredentialDefinition.parse(credentialScope);
+        credentialConfig.setCredentialDefinition(credentialDefinition);
     }
 }
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
@@ -26,9 +26,11 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.IntStream;
 
+import org.keycloak.models.oid4vci.CredentialScopeModel;
 import org.keycloak.protocol.oid4vc.model.CredentialBuildConfig;
 import org.keycloak.protocol.oid4vc.model.CredentialSubject;
 import org.keycloak.protocol.oid4vc.model.Format;
+import org.keycloak.protocol.oid4vc.model.SupportedCredentialConfiguration;
 import org.keycloak.protocol.oid4vc.model.VerifiableCredential;
 import org.keycloak.sdjwt.DisclosureSpec;
 import org.keycloak.sdjwt.IssuerSignedJWT;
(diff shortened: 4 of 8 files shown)


mposolda:
@namanONcode Thanks! Could you please list existing issue (in the commit message and in the GH issue) and take care that "spotless" is working? See this message in your other PR for the details: https://github.com/keycloak/keycloak/pull/45809#issuecomment-3809596685

namanONcode:
@mposolda i have fixes both pr issue related to contribution guidelines please allow ci to run for both pr

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/CredentialBuilder.java
```diff
@@ -48,4 +50,23 @@ CredentialBody buildCredentialBody(
...
+     * or {@code credential_definition} for {@code jwt_vc_json}).
+     * </p>
+     *
+     * <p>
+     * The default implementation is a no-op to preserve backward compatibility.
+     * </p>
+     *
+     * @param metadata mutable map representing the credential metadata entry
+     */
+    default void contributeToMetadata(Map<String,Object> metadata) {
```
mposolda: It seems much better if this method has signature with "typed" arguments. Maybe something like:
```
default void contributeToMetadata(SupportedCredentialConfiguration credentialConfig, CredentialScopeModel credentialScope);
```
instead of generic map.

This would allow that method implementations can "add" proper fields to the credentialConfig rather than "remove" previously added fields.

It seems to me this might work if the method is invoked somewhen from `SupportedCredentialConfiguration.parse` (that method should be updated also to remove `credentialConfiguration.setVct` and `credentialConfiguration.setCredentialDefinition` to make sure that this is added by the corresponding builders).

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerWellKnownProvider.java
```diff
@@ -460,16 +461,58 @@ public static Map<String, SupportedCredentialConfiguration> getSupportedCredenti
...
+
+        try {
+            Map<String, Object> metadataMap = JsonSerialization.readValue(
+                    JsonSerialization.writeValueAsString(config),
+                    new com.fasterxml.jackson.core.type.TypeReference<Map<String, Object>>() {}
+            );
+
+            credentialBuilder.contributeToMetadata(metadataMap);
+
+            if (!metadataMap.containsKey("vct")) {
```
mposolda: It will be good if this is not needed. In theory, the `OID4VCIssuerWellKnownProvider` should not have any knowledge of the fields like `vct` or `credential_definition`, which are specific to concrete credential types. This is related to my other comment (in `CredentialBuilder` class).

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/credentialbuilder/CredentialMetadataWellKnownTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.keycloak.util.JsonSerialization;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class CredentialMetadataWellKnownTest extends OID4VCIssuerEndpointTest {
```
mposolda: Is it possible to use existing `OID4VCIssuerWellKnownProviderTest` and either update some existing tests in that class or introduce new test methods to that class?

That test would be better as it is usually better approach to do end-to-end testing (properly send HTTP request to the endpoint instead of using `testingClient.server.run` to test Java API directly).

Just a note that there is some existing PR https://github.com/keycloak/keycloak/pull/45784, which will introduce conflicts, so probably may be good to wait until that one is merged (would be probably later today).

mposolda (commented):
@namanONcode Thanks! Added few comments inline. Could you check them please?

namanONcode:
> @namanONcode Thanks! Added few comments inline. Could you check them please?

Yeah I will look and implement these comments

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/credentialbuilder/CredentialMetadataWellKnownTest.java
mposolda: PR #45784 already merged. Feel free to rebase to align with latest updates in Keycloak main.

namanONcode:
@mposolda i have done suggested changes mentioned in comments also test are also refactored please allow run ci and i had implemented with latest fork of main branch as you mentioned to avoid merge conflicts

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/SdJwtCredentialBuilder.java
```diff
@@ -134,4 +136,11 @@ public SdJwtCredentialBody buildCredentialBody(
 
         return new SdJwtCredentialBody(sdJwtBuilder, issuerSignedJWT);
     }
+
+    @Override
+    public void contributeToMetadata(SupportedCredentialConfiguration credentialConfig, CredentialScopeModel credentialScope) {
+        String vct = Optional.ofNullable(credentialScope.getVct()).orElse(credentialScope.getName());
+        credentialConfig.setVct(vct);
+        credentialConfig.setCredentialDefinition(null);
```
mposolda: Minor: I think this line can be omitted as it is not needed to explicitly call `setCredentialDefinition` here? As long as it is not set on any other place (which is not with your changes), it would be `null` automatically. No need for this impl to take care of `credentialDefinition`, which this builder doesn't understand at all in theory.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialbuilder/JwtCredentialBuilder.java
```diff
@@ -104,4 +107,11 @@ public JwtCredentialBody buildCredentialBody(
 
         return new JwtCredentialBody(jwsBuilder);
     }
+
+    @Override
+    public void contributeToMetadata(SupportedCredentialConfiguration credentialConfig, CredentialScopeModel credentialScope) {
+        CredentialDefinition credentialDefinition = CredentialDefinition.parse(credentialScope);
+        credentialConfig.setCredentialDefinition(credentialDefinition);
+        credentialConfig.setVct(null);
```
mposolda: Minor: I think this line can be omitted as it is not needed to explicitly call `setVct` here? As long as it is not set on any other place (which is not with your changes), it would be `null` automatically. No need for this impl to take care of `vct`, which this builder doesn't understand at all in theory.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCIssuerWellKnownProviderTest.java
```diff
@@ -874,6 +879,88 @@ public void testOldOidcDiscoveryCompliantWellKnownUrlWithDeprecationHeaders() {
         }
     }
 
+    /**
+     * Tests that SD-JWT credential metadata contains 'vct' and does NOT contain 'credential_definition'.
+     * This is an end-to-end test using HTTP requests to the well-known endpoint.
+     */
+    @Test
+    public void testSdJwtMetadataContainsVctAndNotCredentialDefinition() {
```
mposolda: I think you can omit adding those new test methods, but instead, you can possibly update existing tests as there are some failure (with your changes), which are probably caused exactly by the fact that `vct` and `credentialDefinition` are not present.

BTW. If you think some new test is good to add, it would be good to avoid using Apache HTTP client directly, but instead using `OAuthClient` stuff (like other test methods of this class are using).

mposolda (commented):
@namanONcode Thanks for the updates! Added few more minor comments inline.

Besides that, it will be good to look at the failing tests as those are caused by your changes probably: https://github.com/keycloak/keycloak/actions/runs/21447397428/job/61767431090?pr=45807 .

Maybe it is not needed to add new test methods as the `vct` and `credentialDefinition` fields are probably already tested for some existing credentials. So maybe it is sufficient to fix the existing tests instead and that is sufficient to test that metadata looks as expected?

namanONcode:
@mposolda all new suggestions are implemented to avoid ci test failure still fails please guide me to resolve this

mposolda:
> @mposolda all new suggestions are implemented to avoid ci test failure still fails please guide me to resolve this

Thanks. I see some of my earlier suggestions are not yet applied (with calling `setVct(null)` etc).

Regarding test, now there seems to be only `OID4VCIssuerWellKnownProviderTest` has some test failures. You can probably run this test from your IDE (if you are using Intellij Idea) and then update test logic (if test tests incorrect thing) or codebase logic (if test is correct, but your PR introduced regression). For this particular test, I think this would rather about updating the test logic. As previously, it expected that both `vct` and `credential definition` are always returned (even for the credentials of the formats where it should not be). So I guess this logic might need to be updated.

namanONcode:
> > @mposolda all new suggestions are implemented to avoid ci test failure still fails please guide me to resolve this
> 
> Thanks. I see some of my earlier suggestions are not yet applied (with calling `setVct(null)` etc).
> 
> Regarding test, now there seems to be only `OID4VCIssuerWellKnownProviderTest` has some test failures. You can probably run this test from your IDE (if you are using Intellij Idea) and then update test logic (if test tests incorrect thing) or codebase logic (if test is correct, but your PR introduced regression). For this particular test, I think this would rather about updating the test logic. As previously, it expected that both `vct` and `credential definition` are always returned (even for the credentials of the formats where it should not be). So I guess this logic might need to be updated.

@mposolda i had fixed all issues and base test 6 is now passing but i see `[Keycloak Operator CI / Test OLM installation (pull_request)](https://github.com/keycloak/keycloak/actions/runs/21486655535/job/61898197270?pr=45807)` this fails i think this issue isnt from my pr may be problem in ci , if there is any suggestion ?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCIssuerWellKnownProviderTest.java
```diff
@@ -874,6 +880,88 @@ public void testOldOidcDiscoveryCompliantWellKnownUrlWithDeprecationHeaders() {
...
+
+            // Configure realm for unsigned metadata
+            testingClient.server(TEST_REALM_NAME).run(session -> {
+                RealmModel realm = session.getContext().getRealm();
+                realm.setAttribute(OID4VCIssuerWellKnownProvider.SIGNED_METADATA_ENABLED_ATTR, "false");
+            });
+
+            HttpGet getJsonMetadata = new HttpGet(wellKnownUri);
+            getJsonMetadata.addHeader(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON);
+            try (CloseableHttpResponse response = httpClient.execute(getJsonMetadata)) {
```
mposolda: As pointed already, it would be good to avoid direct use of Apache HTTP client, but instead use new `oauthClient`. Please see the other tests in this class and how they are sending requests. Ideal is, if this PR doesn't need to import any Apache HTTP client classes into this class...

Another question: Are those new test methods really needed? Isn't it sufficient coverage already that `vct` or `credentialDefinition` were added just to the respective credentials? It seems to me it is sufficient coverage, due the fact that you needed to update existing tests.

IMO it can be good to either avoid those new test methods (`testSdJwtMetadataContainsVctAndNotCredentialDefinition` and `testJwtVcMetadataContainsCredentialDefinitionAndNotVct` ) entirely and just remove them from this class (that is preferred by me) or update them to use `oauthClient` instead of direct use of Apache HTTP client.

mposolda (commented):
@namanONcode Thanks!

Yes, it seems those operator tests are unrelated to your changes. I've added one last review comment (that is what I've added already and is not yet covered by latest changes).

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCIssuerWellKnownProviderTest.java
namanONcode: @mposolda yeah i m also thinking now to remove these test as there not much need as we already tested it's perfectly refactored so i am finally removing both test `testSdJwtMetadataContainsVctAndNotCredentialDefinition and testJwtVcMetadataContainsCredentialDefinitionAndNotVct`

namanONcode:
@mposolda three failing checks i dont think this is related to pr any other suggestions ?

mposolda:
@namanONcode I agree that test failures are likely not related to your PR. Could you please rebase on top of latest main? That should hopefully help with:
- Conflict, which was introduced by other merged PR https://github.com/keycloak/keycloak/commit/c08ed20f781ddd01c7ca03d413b944dc7651b5d3
- Hopefully should also help with the failing test (not 100% sure, but it seems something was fixed in the meantime in Keycloak main)

Hopefully that will help to finally make your changes in. Thanks for your patience!

namanONcode:
> @namanONcode I agree that test failures are likely not related to your PR. Could you please rebase on top of latest main? That should hopefully help with:
> 
> * Conflict, which was introduced by other merged PR [c08ed20](https://github.com/keycloak/keycloak/commit/c08ed20f781ddd01c7ca03d413b944dc7651b5d3)
> * Hopefully should also help with the failing test (not 100% sure, but it seems something was fixed in the meantime in Keycloak main)
> 
> Hopefully that will help to finally make your changes in. Thanks for your patience!

@mposolda done i had rebase on the top of main branch merge conflict resolved all test cases passed

mposolda (approved):
@namanONcode Thanks!

PR #45808, opened 2026-01-27
Refactor SessionsResource for better memory usage and performance

### LINKED ISSUE #45727: Refactor `SessionsResource` for better memory usage and performance
### Description

The method `realmSessions` performs more than 2 queries for each session type (regular and offline)
* getActiveClientSessionStats(realm, boolean)
* getUserSessionsStream(realm, client)

These can be optimized by adding a method to stream all sessions from the database, reducing the number of invocations by half:
* getUserSessionsStream(realm)

It could also happen that a query might run out of memory if there are a log of sessions and the admin uses search parameters to load all of them into memory.

### Implementation Details

The query must sort by user session id to support pagination from the UI and return no duplicates.

pruivo:
Closes #45727

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/ImmutableClientSession.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/ImmutableClientSession.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2026 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.models.sessions.infinispan;
+
+import java.util.Map;
+
+import org.keycloak.models.AuthenticatedClientSessionModel;
+import org.keycloak.models.ClientModel;
+import org.keycloak.models.RealmModel;
+import org.keycloak.models.UserSessionModel;
+
+record ImmutableClientSession(
+        String id,
+        ClientModel client,
+        ImmutableUserSessionModel userSessionModel,
+        Map<String, String> notes,
+        String redirectUri,
+        String action,
+        int timestamp,
+        int started
+) implements AuthenticatedClientSessionModel {
+
+
+    @Override
+    public String getId() {
+        return id;
+    }
+
+    @Override
+    public int getTimestamp() {
+        return timestamp;
+    }
+
+    @Override
+    public void setTimestamp(int timestamp) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public int getStarted() {
+        return started;
+    }
+
+    @Override
+    public void detachFromUserSession() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public UserSessionModel getUserSession() {
+        return userSessionModel;
+    }
+
+    @Override
+    public String getNote(String name) {
+        return notes.get(name);
+    }
+
+    @Override
+    public void setNote(String name, String value) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void removeNote(String name) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Map<String, String> getNotes() {
+        return notes;
+    }
+
+    @Override
+    public String getRedirectUri() {
+        return redirectUri;
+    }
+
+    @Override
+    public void setRedirectUri(String uri) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public RealmModel getRealm() {
+        return userSessionModel().getRealm();
+    }
+
+    @Override
+    public ClientModel getClient() {
+        return client;
+    }
+
+    @Override
+    public String getAction() {
+        return action;
+    }
+
+    @Override
+    public void setAction(String action) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String getProtocol() {
+        return "";
+    }
+
+    @Override
+    public void setProtocol(String method) {
+        throw new UnsupportedOperationException();
+    }
+}
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/ImmutableClientSession.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/ImmutableClientSession.java
@@ -17,6 +17,7 @@
 
 package org.keycloak.models.sessions.infinispan;
 
+import java.util.Collections;
 import java.util.Map;
 
 import org.keycloak.models.AuthenticatedClientSessionModel;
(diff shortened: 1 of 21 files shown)


pruivo:
Some results, with a database populated with 200,000 regular sessions, and invoking the endpoint `/admin/realms/realm-0/ui-ext/sessions?first=0&max=101&type=ALL&search=X` 

* Keycloak 26.5 took 1.9 minutes
* This PR took 2.4 seconds

model/jpa/src/main/java/org/keycloak/models/jpa/session/ImmutablePersistentClientSessionEntity.java
```diff
@@ -0,0 +1,77 @@
...
+) implements PersistentClientSessionModel {
+    @Override
+    public String getUserSessionId() {
+        return userSessionId;
+    }
+
+    @Override
+    public void setUserSessionId(String userSessionId) {
+
+    }
```
ahus1, line 44: A `throw new UnsupportedOperationException();` is missing here and the other setters

server-spi/src/main/java/org/keycloak/models/UserSessionProvider.java
```diff
@@ -278,4 +278,50 @@ default void migrate(String modelVersion) {
...
+    /**
+     * Stream all the regular sessions in the realm.
+     * <p>
+     * The returned {@link UserSessionModel} instances are immutable. More precisely, the entity is not tracked by the JPA and any
+     * modification may throw an {@link UnsupportedOperationException}.
+     *
+     * @param realm The {@link RealmModel} instance.
+     * @return A {@link Stream} for all the sessions in the realm.
+     */
+    Stream<UserSessionModel> readOnlyStreamUserSessions(RealmModel realm);
```
ahus1: These are API changes in the `server-spi`, which can affected by users. I double checked, and that `UserSessionSpi` is marked as an internal SPI, so we're probably safe. 

In addition, please review if some of the methods can be deprecated.

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/ImmutableClientSession.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Override
+    public void setAction(String action) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String getProtocol() {
+        return "";
```
ahus1: Could this return actual data? Or should it throw an exception "Unsupported"?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/ImmutableClientSession.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Override
+    public String getAction() {
+        return action;
+    }
+
+    @Override
+    public void setAction(String action) {
+        throw new UnsupportedOperationException();
```
ahus1: Please add a text similar to the following so people know what is happening: 

```suggestion
        throw new UnsupportedOperationException("this is read only);
```

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
```diff
@@ -442,6 +472,45 @@ private Stream<UserSessionModel> loadExactUserSessionsWithClientSessions(TypedQu
...
+                        var client = realm.getClientById(clientSessionEntity.getClientId());
+                        if (client == null) {
+                            return;
+                        }
+
+                        var adapter = new PersistentAuthenticatedClientSessionAdapter(session, clientSessionEntity, realm, client, userSession);
+                        userSession.getAuthenticatedClientSessions().put(clientSessionEntity.getClientId(), adapter);
+                    });
+
+                    logger.tracef("Loaded %d batch of user sessions (offline=%s, sessionIds=%s)", batchedUserSessions.size(), offlineStr, sessionsById.keySet());
```
ahus1: How efficient is this as it is calling `.size()` and `keySet()`? Should this be wrapped in a is-trace-enabled? This has been copied over from another method in the class, please copy over.

model/storage-private/src/main/java/org/keycloak/models/session/UserSessionPersisterProvider.java
```diff
@@ -155,4 +155,28 @@ default void removeUserSessions(RealmModel realm) {
...
+     * Stream all the sessions belonging to the realm.
+     * <p>
+     * The returned {@link UserSessionModel} instances are immutable. More precisely, the entity is not tracked by the JPA and any
+     * modification may throw an {@link UnsupportedOperationException}.
+     *
+     * @param realm   The {@link RealmModel} instance.
+     * @param offline If {@code true}, it streams the offline sessions, otherwise the regular sessions.
+     * @return A {@link Stream} for all the sessions in the realm.
+     */
+    Stream<UserSessionModel> readOnlyUserSessionStream(RealmModel realm, boolean offline);
```
ahus1, line 168: Can some of the other methods be deprecated?

ahus1 (changes requested):
Thank you for this PR, great work!

See below for some comments. Please add a release notes "notable change" that describes the SPI changes. As they are internal SPI, they are not breaking, still people should be aware of it. 

See below for my note to review which methods should be considered for deprecation.

model/storage-private/src/main/java/org/keycloak/models/session/UserSessionPersisterProvider.java
pruivo: Yes, I have a follow-up issue to check them. https://github.com/keycloak/keycloak/issues/45823

server-spi/src/main/java/org/keycloak/models/UserSessionProvider.java
pruivo: I can add the old implementation and mark them as default.

model/jpa/src/main/java/org/keycloak/models/jpa/session/JpaUserSessionPersisterProvider.java
pruivo: It is an array list of 100 entries (max). It is fine, but it doesn't hurt to wrap it.

pruivo:
@ahus1, thank you. Please re-review when you have the time.

ahus1 (commented):
@pruivo - thank you for the review, looks good to me. 

I only too late saw that you opened a follow-up issue for the deprecation, and I pushed one more commit. 
I hope there is nothing wrong with that PR, and we keep the task of any other deprecation to the other task. 

Please have a look at my latest changes, and then it should be good to me merged.

ahus1:
@pruivo - there seems to be a flaky (?) test in this run: https://github.com/keycloak/keycloak/actions/runs/21450898156/job/61779616007?pr=45808

What surprises me is a HSEARCH000041 "Index directory does not exist, creating" info message that is issues while the tests have already started running. Is that maybe the cause of the errors? Could Keycloak wait for those indexes to be created? 

This run also failed, not sure yet why it failed: https://github.com/keycloak/keycloak/actions/runs/21450898156/job/61779616011?pr=45808

I hope it is due to my changes ... still I need to take a break now until tomorrow.

pruivo:
@ahus1 I've reverted the code changes to make progress on this PR (which is already quite large). I started working on the follow-up, and I'm seeing the same tests failing. I'm going to investigate, but hopefully, this should be good to go.

ahus1:
> I've reverted the code changes to make progress on this PR

Thank you, this was the correct thing to do. Sorry that I got carried away with this yesterday.

ahus1 (approved):
Thank you for this refactoring and improvement!

pruivo:
@ahus1, no worries, it reveals a bug with Infinispan (I think) as the query condition `IN (...)` is not working. I've opened another PR to workaround the problem: https://github.com/keycloak/keycloak/pull/45857

PR #45816, opened 2026-01-28
[OID4VCI] CredentialEndpoint can be invoked with incorrect access token

### LINKED ISSUE #44670: CredentialEndpoint can be invoked with incorrect access token
### Before reporting an issue

### Area

oid4vc

### Describe the bug

Based on the OID4VCI specification, it is mentioned that access-token used in the credential endpoint should be the one, which represents approval of the user related to granted credential. See for example [this section](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-8) or some other sections of the specs.

However Keycloak currently allows to invoke credential-endpoint with the access token different than the one returned from the pre-authorized grant token response. See for example [this test](https://github.com/keycloak/keycloak/blob/4dd68c03167e6ee144ef8d7acaa2cf3b43326cb7/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationDetailsFlowTestBase.java#L590) .

We should make sure to prevent this from happening. I suppose that the access-token may need to contain the `scope` representing the requested credential, maybe also `credential_id` should be there in some claim, but not sure... This applies for tokens obtained from pre-authorized code as well as authorization_code grants.

### Version

nightly (from 2025-12-04)

### Regression

### Expected behavior

It is possible to obtain verifiable credential with the different access-token than the one obtained from the pre-authorization grant token response.

### Actual behavior

It should be possible to obtain verifiable credential only with the access-token returned from the token response described in the [chapter 6 of the specs](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-successful-token-response).

### How to Reproduce?

See the description (There is test, which can be used to simulate this).

### Anything else?

_No response_

forkimenjeckayang:
## Summary
This PR fixes the issue where the OID4VCI Credential Endpoint could accept incorrectly issued or misused access tokens.

## Key Changes
• Pre-Authorized tokens are now restricted to the Credential Endpoint only
• Authorization details are embedded in access tokens (not just session notes)
• Token verification happens at the core authentication layer in a pluggable way

##
closes https://github.com/keycloak/keycloak/issues/44670
closes https://github.com/keycloak/keycloak/issues/44580


mposolda:
@forkimenjeckayang Thanks for the PR! I hope to review it soon more properly, however it seems to me that it would have conflicts with your other PR for updating oauthClient for tests: https://github.com/keycloak/keycloak/pull/45784 . It seems it would be good to address that PR first and then update tests respectively in this PR. Is it ok with you?

forkimenjeckayang:
> @forkimenjeckayang Thanks for the PR! I hope to review it soon more properly, however it seems to me that it would have conflicts with your other PR for updating oauthClient for tests: #45784 . It seems it would be good to address that PR first and then update tests respectively in this PR. Is it ok with you?

Yes @mposolda , that's okay. I just addressed comments on the other PR. I will wait till it's merged then I fix the conflicts on this

services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
```diff
@@ -370,6 +370,11 @@ public AccessTokenResponseBuilder refreshAccessToken(KeycloakSession session, Ur
             responseBuilder.getRefreshToken().setAuthorization(validation.newToken.getAuthorization());
         }
 
+        // Ensure authorization_details are also in the new refresh token if present
+        if (authorizationDetails != null && clientConfig.isUseRefreshToken() && responseBuilder.getRefreshToken() != null) {
```
mposolda: AFAIK this should not be needed anymore as `authorizationDetails` are automatically added into refreshToken from the corresponding accessToken (See `refresh token` constructor) with some latest changes.

Hopefully changes in the class `TokenManager` can be removed from this PR?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/signing/OID4VCAuthorizationCodeFlowTestBase.java
```diff
@@ -182,15 +190,38 @@ public void testSecondSSOLoginDoesNotReturnAuthorizationDetails() throws Excepti
...
+        postCredential.addHeader(HttpHeaders.AUTHORIZATION, "Bearer " + tokenResponse.getAccessToken());
+        postCredential.addHeader(HttpHeaders.CONTENT_TYPE, "application/json");
+
+        CredentialRequest credentialRequest = new CredentialRequest();
+        credentialRequest.setCredentialConfigurationId(credentialConfigurationId);
+
+        String requestBody = JsonSerialization.writeValueAsString(credentialRequest);
+        postCredential.setEntity(new StringEntity(requestBody, StandardCharsets.UTF_8));
+
+        try (CloseableHttpResponse credentialResponse = httpClient.execute(postCredential)) {
```
mposolda, line 217: I suppose the test should probably use the `oauthClient` stuff you added in the other PR instead of directly using Apache HTTP client? This probably applies to some other test methods you added in this PR.

Hopefully we can avoid direct usage of `httpClient` when possible (Maybe variable `httpClient` can be removed entirely from `OID4VCIssuerEndpointTest`, but probably does not need to be in this PR...)

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -801,22 +803,21 @@ public Response requestCredential(String requestPayload) {
...
 
-            // Get the credential_configuration_id from AuthorizationDetails
-            // First check if offer state has authorization_details (for pre-authorized flows)
+            // Get the credential_configuration_id from the offer state authorization details
             authDetails = offerState.getAuthorizationDetails();
 
             // Validate authorization_details: either in token or in offer state
             // For pre-authorized flows, offer state is the source of truth
             // For authorization code flows, token must contain authorization_details
             if (authDetails == null) {
```
mposolda: One general point:

- It seems that with latest changes, we have for both "authorization code" flow and "pre-authorized code" flow the logic like:
   - There are always `authorization_details` in the access-token
   - There is always `credential_identifier` present in the access-token inside `authorization_details`
   - The credential-offer available in the store is the same as the one from access-token

Considering that, I am thinking about:

1) In this line, maybe we can doublecheck if `authorizationDetails` from the token matches with `authorizationDetails` token from the stored offer-state . Can we perhaps do something like:
```
OID4VCAuthorizationDetailsResponse tokenAuthDetails = getAuthorizationDetailFromToken(accessToken);
if (tokenAuthDetails == null || tokenAuthDetails.equals(authDetails)) {
   // Throw error about unexpected authorization_details
}
```
This would require also to implement `equals/hashCode` on `OID4VCAuthorizationDetail` and `OID4VCAuthorizationDetailResponse` (as it is implemented on the superclass only, but does not check equality of other things like `credential_identifiers` or `credential_configuration_id` claims)

Is it ok to compare this way or would it be too strict? Right now, not seeing any use-case where it would not work?

It rather seems to me this should be done as in fact, it would also verify that `credential_identifier` from the credential-request matches with the `credential_identifiers` from the access-token and hence correct access-token was used for call the credential-endpoint


2) Considering that we are always adding `credential_identifiers` in the authorization-details in both "pre-authorized flow" and "authorization code flow", we can perhaps always require that credential-request needs to have `credential_identifiers` and never `credential_configuration_id` ? This is per specs section referenced in the description of #44580 .

This would simplify the code and it would make sure to remove some possible attack vectors as we remove one more "Variable" to take care of... It would allow to remove whole `else` block on `OID4VCIssuerEndpoint` - line 874 ( I mean the line using `else if (credentialConfigurationId != null) {` ) and can update some more checks above to make sure that `credential_identifier` in the credential-request would be the only supported thing.

WDYT?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -873,7 +880,24 @@ public Response requestCredential(String requestPayload) {
             });
             eventBuilder.detail(Details.CREDENTIAL_TYPE, credentialConfigurationId);
 
-            authDetails = getAuthorizationDetailFromToken(authResult.token()); // Get authorization_details always from token for this case
+            // No authorization_details for direct credential_configuration_id lookup
+            authDetails = null;
+
+            // Per OID4VCI §8.2: when authorization_details with credential_identifiers are present in the token,
+            // only credential_identifier is allowed in the request (not credential_configuration_id).
+            Object tokenAuthDetails = accessToken.getOtherClaims().get(OAuth2Constants.AUTHORIZATION_DETAILS);
```
mposolda: This will always return `null` as authorization_details would be present always in `accessToken.getAuthorizationDetails()` but never would be present in the `otherClaims` .

However I wonder if this code block could be removed entirely? Then this comment would be pretty much out of questions... (See [my other comment above](https://github.com/keycloak/keycloak/pull/45816/changes#r2740838436)).

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/AuthorizationDetailsValidator.java
```diff
@@ -0,0 +1,182 @@
...
+     * @return The validated authorization details
+     * @throws BadRequestException if validation fails
+     */
+    public static OID4VCAuthorizationDetailResponse getValidatedAuthorizationDetails(
+            CredentialOfferState offerState,
+            AccessToken accessToken,
+            ErrorResponseProvider errorResponseProvider) {
+
+        OID4VCAuthorizationDetailResponse authDetails = offerState.getAuthorizationDetails();
+        Object tokenAuthDetails = accessToken.getOtherClaims().get(OAuth2Constants.AUTHORIZATION_DETAILS);
```
mposolda: I think this does not work... Should use `accessToken.getAuthorizationDetails()` to retrieve authorizationDetails properly.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/AuthorizationDetailsValidator.java
```diff
@@ -0,0 +1,182 @@
...
+
+        if (isPreAuthorizedFlow) {
+            return validatePreAuthorizedCodeFlow(authDetails, errorResponseProvider);
+        } else {
+            // Per OID4VCI spec, authorization_details in token response is REQUIRED only if sent in auth/token request.
+            // If not sent in the request, it's optional. When absent from token, use offer state details.
+            if (tokenAuthDetails == null) {
+                return authDetails;
+            }
+            return validateAuthorizationCodeFlow(tokenAuthDetails, errorResponseProvider);
```
mposolda: Note that this is pretty much unreachable code as `tokenAuthDetails` are always null right now (That is what I've commented on the comment above). And in case that `tokenAuthDetails` are null, then the line 57 would return.

In general: I wonder that maybe `AuthorizationDetailsValidator` is possibly not needed at all considering that:
- The caller class `OID4VCIssuerEndpoint.getAuthorizationDetailFromToken` retrieved the `OID4VCIssuerEndpoint` already from the access-token and in fact checked the `type` during parsing.
- Moreover if we require that authorizationDetails from offer-storage matches with the authorizationDetails from token as I've proposed in the [other comment](https://github.com/keycloak/keycloak/pull/45816/changes#r2740838436), we should be good already as the work is already done

services/src/main/java/org/keycloak/services/managers/AppAuthManager.java
```diff
@@ -192,8 +199,26 @@ public AuthResult authenticate() {
             if (tokenString == null) tokenString = extractAuthorizationHeaderToken(headers);
             // audience can be null
 
-            return verifyIdentityToken(session, realm, uriInfo, connection, true, true, audience, false, tokenString, headers,
+            AuthResult authResult = verifyIdentityToken(session, realm, uriInfo, connection, true, true, audience, false, tokenString, headers,
```
mposolda: Is it possible to just update this to use something like:
```
            return verifyIdentityToken(session, realm, uriInfo, connection, true, true, audience, false, tokenString, headers,
                    verifier -> {
                DPoPUtil.withDPoPVerifier(verifier, realm, new DPoPUtil.Validator(session).request(request).uriInfo(session.getContext().getUri()).accessToken(tokenString));
                verifier.withChecks(accessToken -> GrantTypeEndpointRestrictionValidator.validateGrantTypeRestriction(session, accessToken));
                    });
```

This will allow to remove all the code below and make sure that your validator throws `VerificationException` and the caller should take care already to "translate" it already to proper error response etc.

services/src/main/java/org/keycloak/services/managers/GrantTypeEndpointRestrictionValidator.java
```diff
@@ -0,0 +1,123 @@
...
+ * Validates that tokens are only used on endpoints allowed by their grant type.
+ * This ensures Pre-Authorized Code tokens are restricted to the credential endpoint,
+ * and other grant types only access their intended endpoints.
+ */
+public class GrantTypeEndpointRestrictionValidator {
+    private static final Logger logger = Logger.getLogger(GrantTypeEndpointRestrictionValidator.class);
+
+    private final KeycloakSession session;
+
+    public GrantTypeEndpointRestrictionValidator(KeycloakSession session) {
```
mposolda: Minor: Could be private?

services/src/main/java/org/keycloak/services/managers/GrantTypeEndpointRestrictionValidator.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /**
+     * Validates that the token is allowed for the current endpoint based on its grant type.
+     *
+     * @param token The access token to validate
+     * @throws VerificationException  if token validation fails
+     * @throws ErrorResponseException if server configuration is broken
+     */
+    public void validate(AccessToken token) throws VerificationException {
```
mposolda: Minor: Could be private?

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -199,4 +218,24 @@ public Response process(Context context) {
...
+                .map(path -> path.endsWith("/" + OID4VCIssuerEndpoint.CREDENTIAL_PATH))
+                .orElse(false);
+
+        if (!isCredentialEndpoint) {
+            return false;
+        }
+
+        // Check if token has the correct audience using the native hasAnyAudience method
+        String expectedAudience = OID4VCIssuerWellKnownProvider.getCredentialsEndpoint(session.getContext());
+        return token.hasAnyAudience(Collections.singletonList(expectedAudience));
```
mposolda: Could this also check if this single audience present in the token? As AFAIK there are not more audiences needed in the token in case of pre-authorization code grant?

There was some security issue in the OIDC specs recently that there should be ideally check for being more strict on `audience` and make sure that only expected audience is present and no more audiences. See for example `SecureClientAuthenticationAssertionExecutor` - line 110, which does something similar for audiences in the JWT "assertion".

mposolda (commented):
@forkimenjeckayang Thanks! Added some review comments. Could you please doublecheck?

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
forkimenjeckayang: I agree with your analysis and have implemented the changes:

**1. Always require `credential_identifier`** - Removed the `credential_configuration_id` fallback path entirely. This is valid because:
   - A credential offer is always required for OID4VCI flows (both pre-authorized and authorization code)
   - The credential offer always generates `authorization_details` with `credential_identifiers`
   - Per OID4VCI spec, when `authorization_details` are present in the token, `credential_identifier` must be used

**2. Added `tokenAuthDetails.equals(authDetails)` validation** - Compares authorization_details from token with offer state to verify the correct access token is being used. Added `equals`/`hashCode` to `OID4VCAuthorizationDetail`, `OID4VCAuthorizationDetailResponse`, and `ClaimsDescription`.

**3. Added `credential_identifier` validation** - Verifies the requested `credential_identifier` is in the `credentialIdentifiers` list from authorization_details.

Verified that `authorization_details` is always returned:
- **Pre-authorized code**: Requires exactly 1 `authorization_details`, throws error otherwise
- **Authorization code**: Has 3-level fallback ensuring generation from credential offer if not provided

This simplifies the code and removes potential attack vectors.

mposolda (approved):
@forkimenjeckayang Thanks!

PR #45828, opened 2026-01-28
Avoid breaking DB changes during patch releases

### LINKED ISSUE #38888: Avoid breaking DB changes during patch releases

ryanemerson:
Closes #38888

This PR introduces a maven plugin that has two goals: `create-snapshot` and `verify-compatibility`.  The envisaged workflow for a release is as follows:

1. Update keycloak-rel to execute `mvn clean install -am -pl model/jpa -Pcreate-snapshot -DskipTests` when the [release branch is created](https://github.com/keycloak-rel/keycloak-rel/blob/main/.github/workflows/branch-create.yml) and commit the two generated json files:

```
model/jpa/src/main/resources/META-INF/rolling-upgrades-supported-changes.json
model/jpa/src/main/resources/META-INF/rolling-upgrades-unsupported-changes.json
```

This way the files are never committed to `main`.

2.  Execute the `verify-compatibility` stage as part of the `test` lifecycle of the `model/jpa` module. On the CI this will be executed as part of the existing `unit-tests` job.

The `verify-compatibility` stage always passes if the configured supported and unsupported files do not exist, so this should not cause issues on `main`.

## DatabaseCompatibilityMetadataProvider

This has been updated to create a hash of the stored ChangeSet definitions only if the `/META-INF/rolling-upgrades-unsupported-changes.json` resource exists.

The easiest way to test this is to create a snapshot locally, manually move an entry from the supported to the unsupported file and then execute one of the `UpdateCommandDistTest` db tests.

## ChangeSet JSON
The supported and unsupported json files store the Id, author and original file of a change in order to comply with Liquibase idea of uniqueness as we have multiple changesets with the same ID, but in different files. This also allows for more informative error messages when the compatibility check fails.

Currently a user needs to manually add the json to the supported or unsupported file. Do we want to add an additional mojo to handle this for users so they do not need to be concerned with the json format?


pruivo:
I would avoid touching the file manually. Two new mojos may be a better dev experience
* Goal `safe`, parameters `db.changeset.ids=id1,...` (comma-separated ids to add to the safe file). Optionally, `db.changeset.all=true` to include everything missing.
* * Goal `unsafe`, parameters `db.changeset.ids=id1,...` (comma-separated ids to add to the unsafe file). Optionally, `db.changeset.all=true` to include everything missing.

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/AbstractMojo.java
```diff
@@ -0,0 +1,32 @@
...
+abstract class AbstractMojo extends org.apache.maven.plugin.AbstractMojo {
+
+   @Parameter(defaultValue = "${project}", readonly = true)
+   protected MavenProject project;
+
+   @Parameter(property = "db.verify.supportedFile", required = true)
+   protected String supportedFile;
+
+   @Parameter(property = "db.verify.unsupportedFile", required = true)
+   protected String unsupportedFile;
```
pruivo: It would be nice to have a skip option.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/compatibility/DatabaseCompatibilityMetadataProvider.java
```diff
@@ -30,6 +41,19 @@ public Map<String, String> metadata() {
             addOptional(DatabaseOptions.DB_URL_PORT, metadata);
             addOptional(DatabaseOptions.DB_URL_DATABASE, metadata);
         }
+
+        ObjectMapper objectMapper = new ObjectMapper();
+        try (InputStream inputStream = DatabaseCompatibilityMetadataProvider.class.getResourceAsStream("/META-INF/rolling-upgrades-unsupported-changes.json")) {
+            if (inputStream != null) {
+                // Load the ChangeSet JSON into memory and write to a JSON String in order to avoid whitespace changes impacting the hash
+                Set<ChangeSet> changeSets = objectMapper.readValue(inputStream, new TypeReference<>() {});
```
pruivo, line 51: Don't you have to sort the changeSets to avoid hash changing if the order changes?

ryanemerson:
Thanks for the comments @pruivo. I have added the two mojos you suggested, however:

- I have named these "supported" and "unsupported" to keep with the existing naming convention I established. We can obviously change this if people prefer,but lets be consistent.
- I have made it so that unless `addAll` is specified, ChangeSets can only be added individually. The reason for this is that it's not sufficient to provide a list of IDs, instead we need to provide `{Id, Author, Filename}` in order to uniquely reference a change. Therefore I thought a list of those tuples would be messy! I would expect most backport PRs will only introduce a small number of changes, so IMO this is acceptable.

I have also added a README with usage instructions.

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/VerifyCompatibilityMojo.java
```diff
@@ -0,0 +1,90 @@
...
+            getLog().error("You must determine whether the ChangeSet(s) is compatible with rolling upgrades or not");
+            getLog().error("A ChangeSet that requires locking preventing other cluster members accessing the database is NOT compatible with rolling upgrades");
+            getLog().error("Rolling upgrade compatibility must be verified against all supported database vendors before the supported file is updated");
+            getLog().error("If the change IS compatible, then it should be committed to the repository in the supported file: '%s'".formatted(sFile.toString()));
+            getLog().error("If the change IS NOT compatible, then it should be committed to the repository in the unsupported file: '%s'".formatted(sFile.toString()));
+            getLog().error("Adding a ChangeSet to the unsupported file ensures that a rolling upgrade is not attempted when upgrading to the first patch version containing the change");
+            getLog().error("ChangeSets can be added to the supported or unsupported files using the org.keycloak:db-compatibility-verifier-maven-plugin. See the module README for usage instructions");
```
ahus1, line 86:
```suggestion
            getLog().error("You must determine whether the ChangeSet(s) is compatible with rolling upgrades or not.");
            getLog().error("A ChangeSet that requires locking preventing other cluster members accessing the database or that adds columns that breaks earlier versions is NOT compatible with rolling upgrades.");
            getLog().error("Rolling upgrade compatibility must be verified against all supported database vendors before the supported file is updated.");
            getLog().error("If the change IS compatible, then it should be committed to the repository in the supported file: '%s'".formatted(sFile.toString()));
            getLog().error("If the change IS NOT compatible, then it should be committed to the repository in the unsupported file: '%s'".formatted(sFile.toString()));
            getLog().error("Adding a ChangeSet to the unsupported file ensures that a rolling upgrade is not attempted when upgrading to the first patch version containing the change.");
            getLog().error("ChangeSets can be added to the supported or unsupported files using the org.keycloak:db-compatibility-verifier-maven-plugin. See the module README.md for usage instructions.");
```

misc/db-compatibility-verifier/README.md
```diff
@@ -0,0 +1,87 @@
...
+### `supported` - Adds one or all missing ChangeSets to the supported JSON file
+
+This goal is used to mark a ChangeSet as supported for rolling upgrades.
+
+To mark a single ChangeSet as supported:
+
+```bash
+mvn org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:supported \
+  -Ddb.verify.supportedFile=<relative-path-to-json-file> \
+  -Ddb.verify.unsupportedFile=<relative-path-to-json-file> \
```
ahus1, line 50: Move the `supportedFile` and `unsupportedFile` to a separate section, as ideally it should be added to the pom.xml, so caller's don't need to worry about it.

misc/db-compatibility-verifier/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+            <groupId>org.apache.maven.plugin-tools</groupId>
+            <artifactId>maven-plugin-annotations</artifactId>
+            <version>${maven.plugin-tools.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-core</artifactId>
+            <version>${maven.version}</version>
+            <scope>compile</scope>
```
tkyjovsk: I see a warning for this:
```
Some dependencies of Maven Plugins are expected to be in provided scope.
Please make sure that dependencies listed below declared in POM
have set '<scope>provided</scope>' as well.

The following dependencies are in wrong scope:
 * org.apache.maven:maven-model:jar:3.9.12:compile
 * org.apache.maven:maven-artifact:jar:3.9.12:compile
 * org.apache.maven:maven-core:jar:3.9.8:compile
 * org.apache.maven:maven-settings:jar:3.9.12:compile
 * org.apache.maven:maven-settings-builder:jar:3.9.12:compile
 * org.apache.maven:maven-builder-support:jar:3.9.12:compile
 * org.apache.maven:maven-repository-metadata:jar:3.9.12:compile
 * org.apache.maven:maven-model-builder:jar:3.9.12:compile
 * org.apache.maven:maven-resolver-provider:jar:3.9.12:compile
```

misc/db-compatibility-verifier/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+This goal is used to mark a ChangeSet as unsupported for rolling upgrades.
+
+To mark a single ChangeSet as unsupported:
+
+```bash
+mvn org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:unsupported \
+  -Ddb.verify.supportedFile=<relative-path-to-json-file> \
+  -Ddb.verify.unsupportedFile=<relative-path-to-json-file> \
+  -Ddb.verify.changset.id=<id> \
```
ahus1, line 75: I tried running 

```
$ mvn -pl model/jpa org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:supported  -Ddb.verify.changset.id=1.1.0.Beta1   -Ddb.verify.changset.author=12   -Ddb.verify.changset.filename=12  -e
```

and it seems to me it never picks up the IDs I pass as system properties. What's wrong?

ahus1 (changes requested):
Thank you for the pull request, great work! 

I've updated the above Maven command with an `-am` so it also builds the necessary Maven plugin. 
I also made an attempt to make the file config common for all commands. 

I then tried to execute it, but I failed to add a changeset - the passed in values were always null. 

Maybe a check first if any of the values is null would be nice / provide a better error message.

misc/db-compatibility-verifier/README.md
ryanemerson: You have a typo in your properties `s/changset/changeset` :slightly_smiling_face: 

```
mvn -pl model/jpa org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:supported  -Ddb.verify.changeset.id="1.1.0.Beta1"   -Ddb.verify.changeset.author=12   -Ddb.verify.changeset.filename=12 
```

Will correctly result in:

```
[ERROR] Failed to execute goal org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:supported (default-cli) on project keycloak-model-jpa: Error adding ChangeSet to supported file: Unknown ChangeSet: ChangeSet[id=1.1.0.Beta1, author=12, filename=12] -> [Help 1]
```

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/VerifyCompatibilityMojo.java
ryanemerson: I have added a commit with more generic wording that doesn't mention columns specifically: "or makes schema changes that breaks functionality in earlier Keycloak versions".

misc/db-compatibility-verifier/README.md
pruivo: Should we hardcode the paths in the code?

misc/db-compatibility-verifier/pom.xml
ryanemerson: Should be fixed now

ryanemerson:
I have added a `model/jpa/README.md` that explains what we're doing and why, as well as including instructions on how to add supported and unsupported changes. This is meant to compliment the plugin README, not replace it.

> I've updated the above Maven command with an `-am` so it also builds the necessary Maven plugin.

`-am` always causes problems for me, I get:

```
[ERROR] Failed to execute goal org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:supported (default-cli) on project keycloak-parent: The parameters 'unsupportedFile', 'supportedFile' for goal org.keycloak:db-compatibility-verifier-maven-plugin:999.0.0-SNAPSHOT:supported are missing or invalid -> [Help 1]
```

This is because we don't define the supported/unsupported file properties in the root pom.xml. I don't think we want to do this, as the file location value is module specific. 

> I also made an attempt to make the file config common for all commands.

I have simplified and made it so that we just define the following maven properties.:

```
        <db.verify.supportedFile>src/main/resources/META-INF/rolling-upgrades-supported-changes.json</db.verify.supportedFile>
        <db.verify.unsupportedFile>src/main/resources/META-INF/rolling-upgrades-unsupported-changes.json</db.verify.unsupportedFile>
```
That way we can override easily from the mvn CLI.

misc/db-compatibility-verifier/README.md
ryanemerson: I have added `model/jpa/README.md` that has context specific usage instructions that do not required these two properties to be defined.

ahus1:
> -am always causes problems for me, I get:

Ok, I now get them as well, and it doesn't work out when I run it with an install at the same time. Back to no `-am`.

ahus1 (dismissed):
LGTM now, thank you for the changes. Let me know if I should merge it.

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/CreateSnapshotMojo.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Set;
+
+import com.fasterxml.jackson.databind.SerializationFeature;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Mojo;
+
+@Mojo(name = "create-snapshot")
```
pruivo: nitpicking, single work action
```suggestion
@Mojo(name = "snapshot")
```

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/VerifyCompatibilityMojo.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Mojo;
+
+@Mojo(name = "verify-compatibility")
```
pruivo:
```suggestion
@Mojo(name = "verify")
```

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/ChangeLogXMLParser.java
```diff
@@ -0,0 +1,137 @@
...
+               // 1. Handle Root Element
+               if (tagName.equals("databaseChangeLog")) {
+                  continue;
+               }
+
+               // 2. Process changeSet
+               if (tagName.equals("changeSet")) {
+                  String id = reader.getAttributeValue(null, "id");
+                  String author = reader.getAttributeValue(null, "author");
+                  ids.add(new ChangeSet(id, author, filename));
```
pruivo, line 70: IMO, we can remove the `META-INF` directory from the filename.
```java
Path.of(filename).getFileName().toString()
```

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/ChangeLogXMLParser.java
```diff
@@ -0,0 +1,137 @@
...
+            }
+         }
+         return ids;
+      } catch (IOException | XMLStreamException e) {
+         throw new IllegalStateException(e);
+      }
+   }
+
+   // Detect all jpa-changelog*.xml files on the classpath
+   private Collection<String> changeSetXmlFiles() throws IOException {
```
pruivo: nitpicking, all places do `changeSetXmlFiles().stream()`; you can return a stream from this method.

model/jpa/pom.xml
```diff
@@ -171,6 +173,44 @@
...
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.keycloak</groupId>
+                        <artifactId>db-compatibility-verifier-maven-plugin</artifactId>
+                        <version>${project.version}</version>
+                        <executions>
+                            <execution>
+                                <id>create-snapshot</id>
+                                <phase>install</phase>
```
pruivo: Should this be in `process-resources` phase? The generated JSON files are not included in the jar in the `target/` directory. 
```suggestion
                                <phase>process-resources</phase>
```

model/jpa/pom.xml
```diff
@@ -171,6 +173,44 @@
...
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>create-snapshot</id>
```
pruivo: nitpicking
```suggestion
            <id>db-changeset-snapshot</id>
```

pruivo (changes requested):

misc/db-compatibility-verifier/src/main/java/org/keycloak/db/compatibility/verifier/ChangeLogXMLParser.java
ryanemerson: We can, but I then need to add additional logic to extract just the filename from the path passed to the method and the created ChangeSet is less descriptive, so I would rather keep things simple and leave as is.

ryanemerson:
@pruivo I have made all of the plugins single word actions and updated all of the docs. I also added some additional usage instructions to `model/jpa/pom.xml`. Most notably how to create a snapshot and how to skip verification.

pruivo (dismissed):

pruivo:
@ryanemerson spotless plugin is complaining :)

pruivo (approved):

ahus1 (approved):

PR #45833, opened 2026-01-28
Enable graceful HTTP shutdown and document default behavior

### LINKED ISSUE #43589: Gracefully shutting down HTTP stack

ruchikajha95:
Closes #43589

### DIFF
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -278,3 +278,35 @@ The default values of the options are as follows:
 
 If your certificate chain is longer than the given default, you must define the option with an appropriate number.
 Otherwise, the provider will discard the request.
+
+== Graceful HTTP shutdown
+
+When running {project_name} behind a reverse proxy or load balancer , it is important to allow in-flight requests to complete during server shutdown.
+
+{project_name} enables graceful HTTP shutdown by default using Quarkus runtime configuration.
+
+=== Default behavior
+
+By default {project_name} configures Quarkus with short pre-shutdown delay and a bounded shutdown timeout:
+
+[source,properties]
+----
+quarkus.shutdown.delay-enabled=true
+quarkus.shutdown.delay=1s
+quarkus.shutdown.timeout=1s
+----
+
+When shutdown is initiated:
+
+* A pre-shutdown is applied, allowing load balancers to detect that the server is terminating and stop routing new requests to it.
+* Existing HTTP requests continue to be processed.
+* The server waits up to the configured shutdown timeout for in-flight requests to complete before exiting.
+
+=== Overriding shutdown configuration
+
+Advanced users may override the default graceful shutdown behavior using standard Quarkus configuration mechanisms.
+
+These values can be set in application.properties file.
+
+NOTE: When Quarkus properties are defined directly in application.properties, environment variables may not override them.
+This is general Quarkus configuration behavior.
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -6,7 +6,8 @@
 <@tmpl.guide
 title="Configuring a reverse proxy"
 summary="Configure {project_name} with a reverse proxy, API gateway, or load balancer."
-includedOptions="proxy-* hostname hostname-admin http-relative-path">
+
+includedOptions="proxy-* hostname hostname-admin http-relative-path shutdown-delay shutdown-timeout">
 
 Distributed environments frequently require the use of a reverse proxy. {project_name} offers several options to securely integrate with such environments.
 
@@ -254,8 +255,6 @@ The NGINX SSL/TLS module does not expose the client certificate chain. {project_
 If you are using this provider, see <@links.server id="keycloak-truststore"/> for how
 to configure a {project_name} Truststore.
 
-</@tmpl.guide>
-
 === Configuring the rfc9440 provider
 
 If you stick to the header names mentioned in RFC 9440, you do not need to configure any additional options after selecting the `rfc9440` provider.
@@ -374,21 +373,21 @@ For production with TLS passthrough:
 
 [source,bash]
 ----
-bin/kc.sh start --shutdown-delay=5s --shutdown-timeout=30s
+<@kc.start parameters="--shutdown-delay=5s --shutdown-timeout=30s"/>
 ----
 
 For load balancers that poll readiness:
 
 [source,bash]
 ----
-bin/kc.sh start --shutdown-delay=10s --shutdown-timeout=10s
+<@kc.start parameters="-shutdown-delay=10s --shutdown-timeout=10s"/>
 ----
 
 For test environments:
 
 [source,bash]
 ----
-bin/kc.sh start-dev --shutdown-delay=0s --shutdown-timeout=500ms
+<@kc.start parameters="-shutdown-delay=0s --shutdown-timeout=500ms"/>
 ----
 
 [NOTE]
@@ -396,3 +395,5 @@ bin/kc.sh start-dev --shutdown-delay=0s --shutdown-timeout=500ms
 The shutdown delay affects the minimum time required for a complete server restart.
 In Kubernetes environments, ensure your `terminationGracePeriodSeconds` is longer than the sum of `shutdown-delay` and `shutdown-timeout` to prevent forced termination.
 ====
+
+</@tmpl.guide>
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -54,7 +54,7 @@ By default {project_name} is exposed through the root context path (`/`). If the
 - Use a simple hostname for the `hostname` option, `xforwarded` for the `proxy-headers` option, and have the proxy set the `X-Forwarded-Prefix` header.
 - Use a full URL for the `hostname` option including the proxy context path, for example using `--hostname=https://my.keycloak.org/auth` if {project_name} is exposed through the reverse proxy on `/auth`.
 - Change the context path of {project_name} itself to match the context path for the reverse proxy using the `http-relative-path` option.
- 
+
 For more details on exposing {project_name} on different hostname or context path incl. Administration REST API and Console, see <@links.server id="hostname"/>.
 
 == Enable sticky sessions
@@ -329,40 +329,50 @@ Both values accept duration formats: `1s` (seconds), `500ms` (milliseconds), `2m
 
 === When to adjust shutdown timeouts
 
-Consider adjusting these values based on your deployment configuration:
+Consider adjusting these values based on your deployment configuration with the following example Scenarios:
 
-[%header,cols="2,1,1,3"]
+[%autowidth,cols="2,>1,>1,3a"]
 |===
 |Scenario |Delay |Timeout |Reason
 
 |Load balancer polls readiness probe
-|10s
-|1-10s
-|Allow 1-2 poll cycles (typically 5s intervals) for load balancer to detect shutdown
+|16s
+|
+|Assumptions:
+
+* 5s poll interval and a load balancer reconfiguring after two successive failed probes.
+* 1s for reconfiguring the proxy.
+* Proxy using TLS re-encrypt or edge termination.
+
+Calculation:
+
+* Allow three poll cycles for the load balancer to detect shutdown and the extra time for the load balancer to perform the reconfiguration:
++
+`shutdown delay = 3 * 5s + 1s`
 
 |TLS passthrough configuration
-|1s
-|10-30s
-|Longer timeout allows keepalive connections to drain naturally and receive connection close signals
+|10‑30s
+|
+|Longer delay allows keepalive connections to drain naturally and receive connection close signals.
 
 |Long-running admin API requests
-|1s
-|10-30s
-|Admin operations may take longer than typical user requests
+|
+|10‑30s
+|Admin operations may take longer than typical user requests.
 
 |Test environments / quick restarts
 |0s
 |500ms
-|Minimize shutdown time when graceful draining is not needed
+|Minimize shutdown time when graceful draining is not needed.
 
-|Deployment orchestration notifies proxy before pod termination
+|Deployment orchestration reconfigures the proxy and drains connections before Pod termination
 |0s
-|1-10s
+|
 |No pre-shutdown delay needed if proxy is already reconfigured
 
-|Combined: TLS passthrough + polled readiness
-|10s
-|30s
+|Combined: TLS passthrough plus polled readiness
+|26‑56s
+|
 |Delays add up: time for load balancer detection + connection draining
 
 |===
@@ -373,21 +383,21 @@ For production with TLS passthrough:
 
 [source,bash]
 ----
-<@kc.start parameters="--shutdown-delay=5s --shutdown-timeout=30s"/>
+<@kc.start parameters="--shutdown-delay=30s --shutdown-timeout=1s"/>
 ----
 
 For load balancers that poll readiness:
 
 [source,bash]
 ----
-<@kc.start parameters="-shutdown-delay=10s --shutdown-timeout=10s"/>
+<@kc.start parameters="--shutdown-delay=16s --shutdown-timeout=1s"/>
 ----
 
 For test environments:
 
 [source,bash]
 ----
-<@kc.start parameters="-shutdown-delay=0s --shutdown-timeout=500ms"/>
+<@kc.start parameters="--shutdown-delay=0s --shutdown-timeout=500ms"/>
 ----
 
 [NOTE]
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -278,6 +278,7 @@ The default values of the options are as follows:
 If your certificate chain is longer than the given default, you must define the option with an appropriate number.
 Otherwise, the provider will discard the request.
 
+[[graceful-http-shutdown]]
 == Graceful HTTP shutdown
 
 When running {project_name} behind a reverse proxy or load balancer, graceful shutdown ensures that in-flight requests complete successfully during server termination, preventing connection errors for clients.
--- a/docs/guides/server/reverseproxy.adoc
+++ b/docs/guides/server/reverseproxy.adoc
@@ -281,32 +281,118 @@ Otherwise, the provider will discard the request.
 
 == Graceful HTTP shutdown
 
-When running {project_name} behind a reverse proxy or load balancer , it is important to allow in-flight requests to complete during server shutdown.
+When running {project_name} behind a reverse proxy or load balancer, graceful shutdown ensures that in-flight requests complete successfully during server termination, preventing connection errors for clients.
 
-{project_name} enables graceful HTTP shutdown by default using Quarkus runtime configuration.
+{project_name} enables graceful HTTP shutdown by default with configurable timeouts.
+
+=== Understanding shutdown phases
+
+The shutdown process consists of two phases:
+
+*Pre-shutdown delay*:: During this phase, {project_name} signals to load balancers and proxies that it is preparing to shut down.
+The server's readiness endpoint returns a "not ready" status, allowing the load balancer to stop routing new requests to this instance.
+Existing TLS and HTTP keepalive connections are allowed to drain naturally.
+The server continues to process existing requests.
+
+*Shutdown timeout*:: After the pre-shutdown delay, {project_name} stops accepting new requests and waits for in-flight HTTP requests to complete.
+If requests are still running after the timeout expires, the server shuts down regardless.
 
 === Default behavior
 
-By default {project_name} configures Quarkus with short pre-shutdown delay and a bounded shutdown timeout:
+By default, {project_name} is configured with a 1-second pre-shutdown delay and a 1-second shutdown timeout.
+These defaults work well for most standard deployments where:
 
-[source,properties]
+* The load balancer reconfigures quickly (within 1 second)
+* Most requests complete within 1 second
+* The reverse proxy uses edge termination or re-encryption (not TLS passthrough)
+
+=== Configuring shutdown timeouts
+
+Advanced users can adjust the shutdown timeouts using CLI options based on their deployment characteristics.
+
+[source,bash]
 ----
-quarkus.shutdown.delay-enabled=true
-quarkus.shutdown.delay=1s
-quarkus.shutdown.timeout=1s
+bin/kc.sh start --shutdown-delay=<duration> --shutdown-timeout=<duration>
 ----
 
-When shutdown is initiated:
+Available options:
 
-* A pre-shutdown is applied, allowing load balancers to detect that the server is terminating and stop routing new requests to it.
-* Existing HTTP requests continue to be processed.
-* The server waits up to the configured shutdown timeout for in-flight requests to complete before exiting.
+`--shutdown-delay`::
+Length of the pre-shutdown phase during which the server prepares for shutdown.
+This period allows for load balancer reconfiguration and draining of TLS/HTTP keepalive connections.
+Default: `1s`
 
-=== Overriding shutdown configuration
+`--shutdown-timeout`::
+The shutdown period waiting for currently running HTTP requests to finish.
+Default: `1s`
 
-Advanced users may override the default graceful shutdown behavior using standard Quarkus configuration mechanisms.
+Both values accept duration formats: `1s` (seconds), `500ms` (milliseconds), `2m` (minutes), etc.
 
-These values can be set in application.properties file.
+=== When to adjust shutdown timeouts
 
-NOTE: When Quarkus properties are defined directly in application.properties, environment variables may not override them.
-This is general Quarkus configuration behavior.
+Consider adjusting these values based on your deployment configuration:
+
+[%header,cols="2,1,1,3"]
+|===
+|Scenario |Delay |Timeout |Reason
+
+|Load balancer polls readiness probe
+|10s
+|1-10s
+|Allow 1-2 poll cycles (typically 5s intervals) for load balancer to detect shutdown
+
+|TLS passthrough configuration
+|1s
+|10-30s
+|Longer timeout allows keepalive connections to drain naturally and receive connection close signals
+
+|Long-running admin API requests
+|1s
+|10-30s
+|Admin operations may take longer than typical user requests
+
+|Test environments / quick restarts
+|0s
+|500ms
+|Minimize shutdown time when graceful draining is not needed
+
+|Deployment orchestration notifies proxy before pod termination
+|0s
+|1-10s
+|No pre-shutdown delay needed if proxy is already reconfigured
+
+|Combined: TLS passthrough + polled readiness
+|10s
+|30s
+|Delays add up: time for load balancer detection + connection draining
+
+|===
+
+==== Example configurations
+
+For production with TLS passthrough:
+
+[source,bash]
+----
+bin/kc.sh start --shutdown-delay=5s --shutdown-timeout=30s
+----
+
+For load balancers that poll readiness:
+
+[source,bash]
+----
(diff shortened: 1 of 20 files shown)


docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,35 @@ The default values of the options are as follows:
...
+=== Default behavior
+
+By default {project_name} configures Quarkus with short pre-shutdown delay and a bounded shutdown timeout:
+
+[source,properties]
+----
+quarkus.shutdown.delay-enabled=true
+quarkus.shutdown.delay=1s
+quarkus.shutdown.timeout=1s
+----
```
ahus1: @ruchikajha95 / @ryanemerson / @vmuzikar - some time ago we discussed that setting Quarkus options in Keycloak config is not supported, and therefore we restrained ourselves from naming any Quarkus properties in our docs.  

So I think we have two routes here: 

* We think it would never needs to be changed by a user -> then we remove it from the docs
* We think this is something that is supported to be changed -> then we change it to an SPI option or a CLI option. CLI option is something that we decided against in our team, so this would leave the SPI. 

Please comment and share your thoughts!

ahus1 (changes requested):
Thank you for this PR, looks great! Please see below for a note about using Quarkus properties in the docs.

docs/guides/server/reverseproxy.adoc
ryanemerson: Me and @vmuzikar discussed the use of an SPI option on slack and the conclusion was that it wasn't really applicable here, as we're not actually configuring a Keycloak SPI.


> some time ago we discussed that setting Quarkus options in Keycloak config is not supported, and therefore we restrained ourselves from naming any Quarkus properties in our docs.

As a third route, can we be pragmatic and reconsider this on a case-by-case basis?

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,35 @@ The default values of the options are as follows:
...
+* The server waits up to the configured shutdown timeout for in-flight requests to complete before exiting.
+
+=== Overriding shutdown configuration
+
+Advanced users may override the default graceful shutdown behavior using standard Quarkus configuration mechanisms.
+
+These values can be set in application.properties file.
+
+NOTE: When Quarkus properties are defined directly in application.properties, environment variables may not override them.
+This is general Quarkus configuration behavior.
```
shawkins:
```suggestion
These values can be set in the quarkus.properties file.

NOTE: When Quarkus properties are defined directly in quarkus.properties, environment variables may not override them.
```

We should revisit whether we want to properly document that environment variables can be used for quarkus properties - certainly users are already doing that - and adjust the source ordinals accordingly.

docs/guides/server/reverseproxy.adoc
shawkins: > as we're not actually configuring a Keycloak SPI

That hasn't stopped us in other locations from using the spi configuration mechanism for other things - for example allowed-system-variables.

> As a third route, can we be pragmatic and reconsider this on a case-by-case basis?

We're going to probably be in a similar situation with all the quarkus orm properties - having all of those as first-class / cli options seems like burden. If there's something we can agree on that's more pragmatic, we'll likely reuse that approach.

I'm mostly fine with the idea of differentiating between supported (quarkus.shutdown.*, quarkus.hibernate-orm.*) and unsupported quarkus options - then allow those options to be present in either the quarkus.properties or properly utilized from the ENV (relates to the comment below about source ordinality).

We can always come back later and add proper keycloak options as needed.

ahus1: The discussion I remembered was rotating around community and customers trying to achieve things with Quarkus parameters that we considered unsupported. And it was difficult for them to figure out what was supported and what not. 

In the beginning we also thought of adding an Keycloak CLI/SPI as an abstraction to what Quarkus is doing to adjust the behavior as needed and smooth migrations. 

I think that is owned by the CND team, so if you want to decide this differently, then ok. Still I think the situation hasn't changed - people will still be confused what is allowed and what not.

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,35 @@ The default values of the options are as follows:
...
+
+{project_name} enables graceful HTTP shutdown by default using Quarkus runtime configuration.
+
+=== Default behavior
+
+By default {project_name} configures Quarkus with short pre-shutdown delay and a bounded shutdown timeout:
+
+[source,properties]
+----
+quarkus.shutdown.delay-enabled=true
```
ahus1: @ruchikajha95 - for those things that people would reconfigure: For now I would consider that they change the delay and the timeout, but they wouldn't change the delay-enabled. delay-enabled is also a build-time options, which would need additional explanation. 

So I suggest to remove "delay-enabled" here.

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,35 @@ The default values of the options are as follows:
...
+
+== Graceful HTTP shutdown
+
+When running {project_name} behind a reverse proxy or load balancer , it is important to allow in-flight requests to complete during server shutdown.
+
+{project_name} enables graceful HTTP shutdown by default using Quarkus runtime configuration.
+
+=== Default behavior
+
+By default {project_name} configures Quarkus with short pre-shutdown delay and a bounded shutdown timeout:
```
ahus1: Thank you for this first set of docs. Please add the following information: 

* Explain the concepts: There is a pre-shutdown and a shutdown period. Explain what they are from the perspective of an administrator of Keycloak, and how Keycloak behaves in each period on a reasonably high level addressing a Keycloak admin. When you do, include the new additional readiness probe that marks the service "down" already during the pre-shutdown period. You can align the wording with the upstream Quarkus docs, but we wouldn't link to those docs as they don't take the Keycloak perspective, and our docs should be self-contained. Actually don't even mention Quarkus, as a Keycloak admin should not care about the fact that Keycloak runs Quarkus underneath. 
* Describe the default behavior in plain English, not  by listing Quarkus properties. 
* When you describe how to configure the behavior, state in which file people would need to add those properties
* Also list the matching environment variables that people can use - those would be named AFAIK QUARKUS_SHUTDOWN_DELAY and QUARKUS_SHUTDOWN_TIMEOUT. People on Kubernetes usually prefer environment variables 

Please double-check with @ryanemerson how this can be configured via the Keycloak CR. I have the suspicion that Quarkus properties might not be available from the Keycloak CR.

ahus1 (dismissed):
Thank you for this pull request, see below for some changes needed to the docs.

stianst (dismissed):
We should NOT document the use of Quarkus properties. They are not supported.

If we believe these are options worth documenting they should be turned into Keycloak properties.

docs/guides/server/reverseproxy.adoc
stianst: Any valid configuration that users of Keycloak should be able to do should be proper Keycloak options, and documented.

That applies to this PR and ORM settings, and anything else.

The only mention we should have on Quarkus options is in https://www.keycloak.org/server/configuration#_format_for_raw_quarkus_properties.

mabartos: I'm not a fan of using Quarkus props either. 

> We think this is something that is supported to be changed -> then we change it to an SPI option or a CLI option. CLI option is something that we decided against in our team, so this would leave the SPI.

Why you decided not to use it? As for the graceful shutdown, there are only 3 quarkus properties now. I'd suggest creating normal KC options that are hidden. For the advanced users, you can point them to use these hidden options. If you decide in the future that you'd completely support changing the shutdown params, you just "unhide" them and they will be advertised to users. 

> We're going to probably be in a similar situation with all the quarkus orm properties - having all of those as first-class / cli options seems like burden. If there's something we can agree on that's more pragmatic, we'll likely reuse that approach.

Once we have such an approach, we can reuse it even for this case in the future.

ahus1: Thank you for chiming in here. Still I disagree with parts of it: If it is in the docs, it is supported. Hidden options doesn't solve this one. All options that are supported should be regular options. 

The alternative we discussed are SPI options that eventually listed here https://www.keycloak.org/server/all-provider-config 

So we can decide to make it not configurable, configurable via documented SPI or configurable via documented CLI options. 

If Quarkus options should not be SPI options, then this eliminates one more choice.

shawkins: > The alternative we discussed are SPI options that eventually listed here https://www.keycloak.org/server/all-provider-config

To show up in that list currently you do need to create the actual SPI and ProviderFactory along with getConfigMetadata method. Given that we'd still need property mapping somewhere in our quarkus / picocli logic, that's more effort than simply adding first-class options.

mabartos: @ahus1 It's true that hidden options should be only referenced to users/customers when they have a specific issue, and they need some sort of backdoor (~1% of normal use cases?). If you reference something, it should be supported - that's right. 

I don't see a big issue here to have normal/regular options for this. We'd need to have the ability to at least, turn off the graceful shutdown - for some users with some special requirements. 

I'm not sure what's better in using SPI options instead of regular options.

vmuzikar: -1 for an SPI option.

SPI options are meant for configuring providers (first- or third-party). This is not an SPI we're configuring here rather than Quarkus behavior. From the support perspective, it does not matter whether it's full blown CLI option or an SPI option – as long as it's documented, it's supported (unless explicitly told otherwise).

The question here should be whether this should be just an escape hatch for dev/debugging purposes, or something documented and supported. If the first, I'd vote for CLI options. If the latter, we can stick to Quarkus options but we must not document it.

ryanemerson: @ahus1 Had some examples where users might want to tweak the delay/timeout, can you share these here with the wider group so that we can try to determine how niche this config actually is?

One approach could be to set the default quarkus property values in this PR and then wait for user feedback to see if this tweaking is actually required by (m)any users before deciding whether to promote these to CLI options or not.

ahus1: I'd say the 1 second pre-shutdown and 1 second shutdown phase are the maximum delays we could add to not call it a non breaking change. 

When we run our test suite, we want to set the pre-shutdown phase to 0 seconds to not slow down our test suite. 

I'd say users would want to set it up with longer periods depending on the their proxy configuration (edge/reencrypt vs. passthrough) and how the proxy gets the information about the shutdown (at the same time as the Pod, well before the pod, or the proxy polls it from the Pod). 

The longer version: 

The 1  second/1 second config might be the right thing when you run in Kubernetes where the load balancer is re-configured at the same time as the Pod gets the termination signal: It takes about a second (as @slaskawi described) for the loadbalancer to finish reconfiguration. If the reverse proxy is edge or reencrypt, the next requests are routed to the remaining pods. So a shutdown period is sufficient for the running requests to finish, and KC might finish early when there are no more running requests. 

Regular login requests should rarely take longer than a second. An admin where they are different longer running requests for example on the admin API might play it safe and wants to set the shutdown period to 10 seconds. 

In proxy-setup with a a TLS passthrough, the connection between the client and the proxy is still established after the proxy reconfiguration, and requests are still sent to the Pod to-be-terminated. It would be good to wait for the connections to drain from the Pod: With the current Quarkus 3/Vert.x 4 setup, the client only receives a HTTP/1.1 connection close or a HTTP/2 GOAWAY when the client sends the next request. So the longer the shutdown period, the higher the probability of the client sending a request, and therefore closing the connection. The longer the connection has been idle, the better the chance to just close it on shutdown and no request currently incoming, or the client closing it voluntarily. In such a setup, I'd say 10 seconds would be OK, 20-30 seconds would be good. There is still a slight probability to lose some requests. It will be better with Quarkus 4, Vert.x 5 and HTTP/2 as that will send a GOAWAY to those clients out-of-order when pre-shutdown starts, which is preferred (but unfortunately not supported in earlier versions). 

In setups where the deployment procedure instructs the proxy ahead of killing the Pod, there is no need to have a pre-shutdown delay. It's quite the opposite: the admin would like to see a pre-shutdown period of 0. 

In setups where the proxy polls the information which Pod should be in the load balancing by polling the readiness probe, it takes 1-2 poll cycles to the loadbalancer to recognize to remove the Pod. Assuming a poll interval of 5 seconds, and two poll cycles, the pre-shutdown period needs to be 10 seconds. 

In the worst case you would have a proxy that polls the status from the Keycloak Pods, and TLS passthrough, and then those two delays add up. Given the examples above, you would have a 10 + 30 = 40 seconds shutdown period.  

Looking at this wall of text, should we simplify this in a decision table?

ryanemerson: > Looking at this wall of text, should we simplify this in a decision table?

If we do add this config, at least we have half the docs written already :smile: 

Thanks for the detailed explanation @ahus1 

My 2c is that this configuration still feels very niche, we're improving our defaults compared to prior behaviour and things should improve again once we are able to upgrade to Quarkus 4. 

It's easy to add many CLI options, but we know we can't easily remove said options except for in Major releases. Adding additional configuration toggles complicates our documentation and adds one more thing for users to consider, potentially causing confusion, when in most cases they shouldn't care.

Unless we have a body of existing issues related to such proxy setups and related issues, I think we should hold off adding the CLI options until we know this is a problem affecting (and causing bad experience) users.

ahus1: I'd say the only part that might be non-common is "An admin where they are different longer running requests for example on the admin API might play it safe and wants to set the shutdown period to 10 seconds."

All other scenarios of draining connections are very common, and also present in our blueprints as they use TLS passthrough. And our blueprint is actually over-simplified (or even invalid depending on your perspective) as it doesn't log client IP addresses, see keycloak/keycloak-benchmark#910, which is usually solved by a different proxy configuration. We've been ignoring this for a while, and people might consider that an security auditing problem. 

Once you look outside of Kubernetes, all those ways to configure proxy load distribution are equally common.

wcote-kz:
Hi @ahus1 , I have a keycloak cluster running 3 replicas, I did try draining the connection for graceful shutdown of the ongoing requests with the following parameters:

```
quarkus.shutdown.delay-enabled=true
quarkus.shutdown.delay=
quarkus.shutdown.timeout=
```

even with those configured, the keycloak replica currently draining (let's say the keycloak no. 1) continues processing job and requests even with traffic not routed through its endpoint. How is that possible? Is it because the jobs and requests are assigned via the ispn cache? I also want to know if this PR is addressing this issues or it's something else.

Thank you

ahus1:
@wcote-kz - please describe your setup so we have more context. 

* What values did you pass in those parameters? 
* What is the order of events when you attempted a graceful shutdown? Did you first reconfigure the loadbalancer, or did you first trigger a shutdown in Keycloak?
* What kind of proxy are you using? Does it do TLS termination, or TLS passthrough? 

Note that job processing is out-of-scope for this, this is only about processing incoming HTTP requests. The current release of Quarkus is also not very good in draining connections.

wcote-kz:
@ahus1 Thank you for your reply, here's the setup I tested (3 replicas of keycloak in a k8s cluster):

``` yaml
    QUARKUS_SHUTDOWN_DELAY_ENABLED: "true"
    QUARKUS_SHUTDOWN_DELAY: "300" # 5 min delay
    QUARKUS_SHUTDOWN_TIMEOUT: "600" # 10 min timeout
```

I observed that keycloak/quarkus is correctly following the above shutdown schedule when performing a statefulset restart (to emulate a rolling upgrade).
During the first 5 min, I can see that the 1rst keycloak is writing in the logs that it is initiating the shutdown and is reporting "not ready" to the k8s loadbalancer. At this point, the 1rst keycloak is removed from the endpoints list of the service and shouldn't receive any new http request.
Here I expect the 1rst keycloak to shutdown after the delay without any issue, since I feel 5 min is more than enough to drain, but the whole time I can see what looks like work/job being done in the logs of the 1rst keycloak. When the 1rst keycloak is shutting down after the delay I get some errors from running api calls (that I run in a loop during my test) and in the admin console I get the "you need to refresh the page" error banner.

So my conclusion is that:
- connections are still reaching the 1rst keycloak somehow even if the load balancer won't route new traffic to it.

My loadbalancer in k8s is traefik with tls termination and I have the same behavior with nginx.

Thank you.

ahus1:
@wcote-kz - the current Quarkus setup is not very good in connection draining: 

* The current version doesn't yet send a HTTP/2 GOAWAY, and also no HTTP/1.1 connection close, so any client with a HTTP connection pool will probably not drain their connections unless they voluntarily close it. This will change with the next release. But as you are using TLS termination, I am surprised that both nginx and traefic continue to route requests to the node that is shutting down.
* The graceful shutdown doesn't affect jobs running in the background for now, so that's expected.

I would have hoped that nginx and traefik would no longer route requests ... can you see that they reconfigure once the Pod is about to shut down?

wcote-kz:
@ahus1 Yes I can see that the ingress controller (traefik or nginx) that points to the keycloak k8s svc is not routing traffic, since I can see the endpoints list goes from 3 keycloak IPs:8080 to 2 keycloak IPs:8080 when one is reporting not ready and in the shutdown period. So I'm confident that the networking part is correct.

I just haven't figured out yet why when the keycloak actually gets stopped I get the issues mentioned in my other comment even if no traffic is going though it technically. 

That's why I was thinking maybe there's some sort of "job/request autobalancing" from keycloak to keycloak via the ispn cache that is done after an online keycloak receives a request that could send it to a "stopping" keycloak in the cluster.

With that said I don't know that this PR or the next Quarkus version are addressing that behavior.

Thank you for your reply

ahus1:
Add two more CLI options, as we require them for proxy configurations. We ruled out to not have SPI options, as this is not about SPIs: 

```
--shutdown-delay: ...
--shutdown-timeout: ... 
```

Add documentation stating that we are handling HTTP at the moment, and explain how they work, and that the functionality will change in the future.

vmuzikar:
> Add two more CLI options

+1 for those two.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
```diff
@@ -242,4 +231,15 @@ private static String resolveMaxThreads(String value,
...
+
+    private static void validateDuration(String value) {
+        try {
+            Duration duration = DurationConverter.parseDuration(value);
+            if (duration.compareTo(Duration.ofSeconds(1)) < 0) {
+                throw new PropertyException("Duration must be at least 1 second.");
+            }
+        } catch (IllegalArgumentException e) {
+            throw new PropertyException("Invalid duration format. Expected format examples: 1s, 30s, 1m, etc.");
+        }
```
pruivo, line 268: Better error message. Looks like this:
```
kc.sh start-dev --shutdown-delay=a
Invalid duration format 'a'. May be an ISO 8601 duration value, an integer number of seconds, or an integer followed by one of [ms, h, m, s, d].
```
```
kc.sh start-dev --shutdown-delay=-1s --verbose
Invalid duration '-1s'. Duration must be positive.

```

```suggestion
        try {
            Duration duration = DurationConverter.parseDuration(value);
            if (duration == null || duration.isNegative()) {
                throw new PropertyException("Invalid duration '%s'. Duration must be positive.".formatted(value));
            }
        } catch (IllegalArgumentException e) {
            throw new PropertyException("Invalid duration format '%s'. %s".formatted(value, OptionsUtil.DURATION_DESCRIPTION));
        }
```

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,121 @@ The default values of the options are as follows:
...
+* The reverse proxy uses edge termination or re-encryption (not TLS passthrough)
+
+=== Configuring shutdown timeouts
+
+Advanced users can adjust the shutdown timeouts using CLI options based on their deployment characteristics.
+
+[source,bash]
+----
+bin/kc.sh start --shutdown-delay=<duration> --shutdown-timeout=<duration>
+----
```
pruivo, line 316:
```suggestion
<@kc.start parameters="-shutdown-delay=<duration> --shutdown-timeout=<duration>"/>
```

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,121 @@ The default values of the options are as follows:
...
+|===
+
+==== Example configurations
+
+For production with TLS passthrough:
+
+[source,bash]
+----
+bin/kc.sh start --shutdown-delay=5s --shutdown-timeout=30s
+----
```
pruivo, line 388:
```suggestion
<@kc.start parameters="-shutdown-delay=5s --shutdown-timeout=30s"/>
```

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,121 @@ The default values of the options are as follows:
...
+
+For load balancers that poll readiness:
+
+[source,bash]
+----
+bin/kc.sh start --shutdown-delay=10s --shutdown-timeout=10s
+----
```
pruivo, line 395:
```suggestion
<@kc.start parameters="-shutdown-delay=10s --shutdown-timeout=10s"/>
```

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,121 @@ The default values of the options are as follows:
...
+
+For test environments:
+
+[source,bash]
+----
+bin/kc.sh start-dev --shutdown-delay=0s --shutdown-timeout=500ms
+----
```
pruivo, line 402:
```suggestion
<@kc.start parameters="-shutdown-delay=0s --shutdown-timeout=500ms"/>
```

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,121 @@ The default values of the options are as follows:
...
+
+[NOTE]
+====
+The shutdown delay affects the minimum time required for a complete server restart.
+In Kubernetes environments, ensure your `terminationGracePeriodSeconds` is longer than the sum of `shutdown-delay` and `shutdown-timeout` to prevent forced termination.
+====
```
pruivo, line 408: It's not related to this change, but the line 256, `</@tmpl.guide>` should be moved to the end of the file.

docs/guides/server/reverseproxy.adoc
```diff
@@ -278,3 +278,121 @@ The default values of the options are as follows:
 
 If your certificate chain is longer than the given default, you must define the option with an appropriate number.
 Otherwise, the provider will discard the request.
+
+== Graceful HTTP shutdown
```
pruivo, line 282: In line 9, the shutdown options should be included, like this:
```adoc
includedOptions="proxy-* hostname hostname-admin http-relative-path shutdown-delay shutdown-timeout">
```

pruivo (changes requested):

ruchikajha95:
Thanks @pruivo for the review . I have made the changes .

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
```diff
@@ -242,4 +255,18 @@ private static String resolveMaxThreads(String value,
...
+
+    private static void validateDuration(String value) {
+        try {
+            Duration duration = DurationConverter.parseDuration(value);
+            if (duration.isNegative() || duration.isZero()) {
+                throw new PropertyException("Invalid duration '" + value + "'. Duration must be positive.");
+            }
+            if (duration.compareTo(Duration.ofSeconds(1)) < 0) {
+                throw new PropertyException("Duration must be at least 1 second.");
+            }
```
pruivo, line 265: Should be removed, right? We want to support a delay/timeout of zero; otherwise, your example in the docs won't work.

quarkus/runtime/src/main/resources/application.properties
```diff
@@ -106,6 +106,8 @@ mp.openapi.extensions.smallrye.duplicateOperationIdBehavior=FAIL
 # Disable Error messages from smallrye.openapi
 # related issue: https://github.com/keycloak/keycloak/issues/41871
 quarkus.log.category."io.smallrye.openapi.runtime.scanner.dataobject".level=off
-
 # 3.31.1 appears to have a hibernate bug
-quarkus.log.category."org.hibernate.orm.sql.exec".level=info
\ No newline at end of file
+quarkus.log.category."org.hibernate.orm.sql.exec".level=info
+quarkus.shutdown.delay=1s
```
pruivo: The default is set by the Java code `.defaultValue("1s")`. This line can be removed.

pruivo (dismissed):

pruivo:
@ruchikajha95, the failed test `HealthDistTest` needs to be updated. The check count is 3.

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/HealthDistTest.java#L86

```
$ curl --insecure https://keycloak:9000/health/ready
{
    "status": "UP",
    "checks": [
        {
            "name": "Graceful Shutdown",
            "status": "UP"
        },
        {
            "name": "Keycloak cluster health check",
            "status": "UP"
        },
        {
            "name": "Keycloak database connections async health check",
            "status": "UP"
        }
    ]
}
```

ahus1:
@ruchikajha95 , @pruivo - I'll push a change in a minute, I was just now reviewing it.

ahus1:
@ruchikajha95 / @pruivo - Thank you for the updated PR, it looks good to me. 

I've updated the docs around the feature, see 89a0311c90e828d83c0ec2c7969186c8d4047d65:

* The "Load balancer polls readiness probe" was not so specific as it IMHO should be: With 2 poll cycles of 5 seconds, you need to wait 3 cycles, as at the moment of shutdown, the previous one might have just finished. And then you add the time for the proxy to reconfigure. 
* The value of 10-30 seconds for TLS passthrough ended up in the Timeout column, while it should IMHO end up in the Delay column. 
* Due to that, the last combined scenario is off, so I updated it as well. 
* I cleared the cells that are not relevant to the respective example. 
* I've updated the example configurations to match the new values of the table above. 

Please review my latest change if I mixed something up. If all is good, this should be good for merging.

quarkus/runtime/src/main/resources/application.properties
ahus1: resolved in next commit

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/HttpPropertyMappers.java
ahus1: agreed, removed.

ahus1:
@ruchikajha95 - can you please add a test for the newly added CLI parameters to `HttpDistTest.java`? Given @pruivo's last comment, I now realize that I missed that in the earlier review. Thanks!

ruchikajha95:
@ahus1 Thanks for the further review . i will add the test changes .
@pruivo  Thanks for the further review .

pruivo:
@ahus1, what tests do you have in mind?

ahus1:
> what tests do you have in mind?

Sorry for not being specific. I'd like to see a test that the CLI options are accepted or return any errors. Additional tests around the shutdown functionality are probably out of scope and difficult to test. Unless you have ideas for that...

ahus1 (approved):
Thank you for this change, @ruchikajha95, and everyone who helped reviewing and contributed!

PR #45840, opened 2026-01-28
fix: promotes keycloak and realm import to v2beta1

shawkins:
closes: #45795

repackages the existing CRDs as v2beta1.

A custom CRD post-processor is used to reintroduce an identical non-stored v2alpha1. As I don't anticipate introducing a difference between the two, this seems fine. The only other thing that needed done is to make the CSV aware of the alternative version - which is being done via a script rather than via josdk logic.

If this approach seems valid, we'll need additional docs / notes alerting customers that v2alpha1 is deprecated and truly only for alpha use for Keycloak and RealmImports. 

If / when we want to drop v2alpha1, olm will prevent this unless all of the stored versions have been updated: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-removal

Just letting the operator run should update the stored versions, but the update can also be done manually if needed: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version


mabartos:
@shawkins What's the progress on this? Is it ready to be reviewed? Or do you plan to add some other changes? 

@shawkins In general, LGTM. 

> If / when we want to drop v2alpha1, olm will prevent this unless all of the stored versions have been updated:

AFAIK, we'd still need to support it for some time to not provide any breaking change across minor releases, right?

How confident you are that nothing breaks for users/customers deployments after these changes? :P 

@Pepo48 @vmuzikar Could you please check it as well? Thanks!

shawkins:
> @shawkins What's the progress on this? Is it ready to be reviewed? Or do you plan to add some other changes?

I want to get buy in on the approach, then it needs some documentation.

> AFAIK, we'd still need to support it for some time to not provide any breaking change across minor releases, right?

Yes and this approach simply keeps them in lock step until we drop v2alpha1.

> How confident you are that nothing breaks for users/customers deployments after these changes? :P

Almost nothing should break as you can continue v2alpha1 without issue. However if you were using our operator jar and expect the v1alpha1 classes to be there, that will no longer be the case. Since I don't think we should consider that supported, I'm ok with that.

shawkins:
@keycloak/cloud-native @vmuzikar this should be ready for review with a minimal set of documentation updates.

As mentioned on the call today there is a corner case in that changing the stored version complicates the roll back of the new operator version.

If any resource is updated to v2beta1 (which will happen automatically on any update), then the new CRDs cannot be removed.

Instead the admin will need to overwrite the CRDs with a version that changes the storage version back to v2alpha1, then run https://kubernetes.io/docs/tasks/manage-kubernetes-objects/storage-version-migration/#update-the-preferred-storage-schema-of-a-crd or other update of the existing CRs to force them back to being stored at v2alpha1.

I'm not sure if it's worth elaborating on this in the initial docs, or just to provide guidance should the situation occur - note that we already have roll back issues with server releases, so it's probably not unexpected that a manual process could apply here.

Ideally both the operator and the server should easily rolled back. For the operator this could look like an intermediate release containing CRDs with both versions and v2alpha1 still the stored version, then another release that switches the stored version to v2beta1. However we don't generally require sequential upgrades or produce multiple operator releases per keycloak release, so this seems like too much effort for this situation.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -228,6 +228,12 @@ Running {project_name} on a database that is not using UTF-8 as a charset is now
 
 In the current version {project_name} will log a warning if the database is not configured to use UTF-8 encoding. Future versions might refuse to work with database that is not configured to use UTF-8 encoding.
 
+=== `Keycloak` and `KeycloakRealmImport` CRDs v2beta1 version
+
+The operator CRDs for the `Keycloak` and `KeycloakRealmImport` resources now include a v2beta1 version.
+The v2alpha1 version for these resources is deprecated. There are currently no differences in the schema between the two versions,
+so you will just need to update the relevant apiVersion references.
```
mabartos:
```suggestion
The operator CRDs for the `Keycloak` and `KeycloakRealmImport` resources now include a `v2beta1` version.
The `v2alpha1` version for these resources is deprecated. There are currently no differences in the schema between the two versions,
so you will just need to update the relevant `apiVersion` references.
```

nitpick: not blocking

mabartos (dismissed):
@shawkins LGTM, based on the context I have and the review I made previously. But my knowledege here is quite limited, so I'd like to hear what others think. 

cc: @vmuzikar @ryanemerson (maybe you have experience from the ISPN side on this topic?)

mabartos (dismissed):

shawkins:
@mabartos @vmuzikar rebased again. Unfortunately this will conflict anytime we add something new to the operator.

mabartos (dismissed):

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -314,6 +314,12 @@ Running {project_name} on a database that is not using UTF-8 as a charset is now
 
 In the current version {project_name} will log a warning if the database is not configured to use UTF-8 encoding. Future versions might refuse to work with database that is not configured to use UTF-8 encoding.
 
+=== `Keycloak` and `KeycloakRealmImport` CRDs v2beta1 version
+
+The operator CRDs for the `Keycloak` and `KeycloakRealmImport` resources now include a `v2beta1` version.
+The `v2alpha1` version for these resources is deprecated. There are currently no differences in the schema between the two versions,
+so you will just need to update the relevant `apiVersion` references.
```
vmuzikar: We could add some more reasoning around it, and also stress that this is not a breaking change.

```suggestion
To better capture the maturity of Operator CRDs for the `Keycloak` and `KeycloakRealmImport`, the resources now include a `v2beta1` version.
The previous `v2alpha1` version for these resources is deprecated but still served. You can keep using it but we recommend upgrading to `v2beta1` at your earliest convenience.
There are currently no differences in the schema between the two versions,
so you will just need to update the relevant `apiVersion` references.
```

operator/scripts/create-olm-bundle.sh
```diff
@@ -76,6 +76,11 @@ yq ea -i 'del(.spec.nativeAPIs[] | select(.kind == "ServiceMonitor"))' "$CSV_PAT
 yq ea -i 'del(.spec.customresourcedefinitions.owned[] | select(.kind == "KeycloakOIDCClient" or .kind == "KeycloakSAMLClient"))' "$CSV_PATH"
 rm ../olm/$VERSION/manifests/keycloak*clients.k8s.keycloak.org-v1.crd.yml
 
+# Create entries for both alpha1 and beta1 - the indexing is dependent on whether the client crs are present
+yq ea -i '.spec.customresourcedefinitions.owned += [.spec.customresourcedefinitions.owned[] | select(.version == "v2beta1")]' "$CSV_PATH" 
+yq ea -i '.spec.customresourcedefinitions.owned[2].version = "v2alpha1"' "$CSV_PATH"
+yq ea -i '.spec.customresourcedefinitions.owned[3].version = "v2alpha1"' "$CSV_PATH"
```
vmuzikar, line 75: I believe this will break any time we add a new CRD, the indexes will be different. Could we calculate the indexes for the added CRDs?

Also would be good to have an OLM test for `v1alpha1`.

operator/src/test/java/org/keycloak/operator/testsuite/integration/RealmImportTest.java
```diff
@@ -134,14 +134,11 @@ public void testWorkingRealmImportWithReplacement() {
...
         // Act
-        k8sclient.getKubernetesSerialization().registerKubernetesResource(KeycloakRealmImport.class);
-        K8sUtils.set(k8sclient, getClass().getResourceAsStream("/example-realm.yaml"), obj -> {
-            KeycloakRealmImport realmImport = (KeycloakRealmImport) obj;
-            realmImport.getSpec().getRealm().setSmtpServer(Map.of("port", "${MY_SMTP_PORT}", "host", "${MY_SMTP_SERVER}", "from", "admin@keycloak.org"));
-            realmImport.getSpec().setPlaceholders(Map.of("MY_SMTP_PORT", new Placeholder(new SecretKeySelectorBuilder().withName("keycloak-smtp-secret").withKey("SMTP_PORT").build()),
-                            "MY_SMTP_SERVER", new Placeholder(new SecretKeySelectorBuilder().withName("keycloak-smtp-secret").withKey("SMTP_SERVER").build())));
-            return realmImport;
-        });
+        KeycloakRealmImport realmImport =k8sclient.getKubernetesSerialization().unmarshal(getClass().getResourceAsStream("/example-realm.yaml"), KeycloakRealmImport.class);
```
vmuzikar: Just a nitpick.
```suggestion
        KeycloakRealmImport realmImport = k8sclient.getKubernetesSerialization().unmarshal(getClass().getResourceAsStream("/example-realm.yaml"), KeycloakRealmImport.class);
```

vmuzikar (commented):
@shawkins Thank you for the PR, and sorry for the late review.

operator/scripts/create-olm-bundle.sh
shawkins: > I believe this will break any time we add a new CRD, the indexes will be different. Could we calculate the indexes for the added CRDs?

I wouldn't bother doing that yet - we can visit this once we're ready to promote the client crds, or add another new crd.

We can consider in the mean time if we want operator sdk support for this instead.

> Also would be good to have an OLM test for v1alpha1.

Let me see what would be good there.

shawkins: Actually since the src/test/resource file api versions have not changed, the olm test is still against v2alpha1.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
shawkins: updated

operator/src/test/java/org/keycloak/operator/testsuite/integration/RealmImportTest.java
shawkins: updated

operator/scripts/create-olm-bundle.sh
vmuzikar: > I wouldn't bother doing that yet - we can visit this once we're ready to promote the client crds, or add another new crd.
> 
> We can consider in the mean time if we want operator sdk support for this instead.

Ok, sounds good.

> Actually since the src/test/resource file api versions have not changed, the olm test is still against v2alpha1.

Would be good to test both v2alpha1 and v2beta1.

shawkins: > Would be good to test both v2alpha1 and v2beta1.

Nearly all of the unit / integration tests use v2beta1.

Are you specifically looking for the keycloak install and import to be run with both v2alpha1 and v2beta1 - I was thinking we could avoid that given that will just extend the test run times.

Alternatively can we just expand the crd check to ensure that both versions are installed?

vmuzikar: > Alternatively can we just expand the crd check to ensure that both versions are installed?

As discussed offline, let's take this approach.

shawkins: Updated now

vmuzikar (approved):
LGTM, thanks.

PR #45849, opened 2026-01-28
feat(identity-providers): add mapper order functionality

### LINKED ISSUE #45848: SAML mapper executing order
### Description

we setup identity providers SAML, we added many mappers,
But we want to control the order of the mapper, to set which mapper will be executed and which one after

So for each mapper, we will have a new field order where we will enter a number 1,2,3,4....

We will have a wrapper with a new toggle flag

if toggle flag is off, then the order of mappers is executed the same as today

if toggle flag is on, then the order of mappers is executed by the numbers we setted

### Value Proposition

 for each mapper, we will have a new field order where we will enter a number 1,2,3,4....

We will have a wrapper with a new toggle flag

if toggle flag is off, then the order of mappers is executed the same as today

if toggle flag is on, then the order of mappers is executed by the numbers we setted

### Goals

- not changing the existing logic by wrapping it with a new toggle flag
- ability to control the order in which the SAML mappers are executed

### Non-Goals

nothing

### Discussion

_No response_

### Notes

_No response_

zivadler:
closed #45848 

we setup identity providers saml, we added many mappers,
but we want to control the order of the mapper, to set which mapper will be executed and which one after

so for each mapper we will have a new field order where we will enter a number 1,2,3,4....

we will wrapper with a new toggle flag

if toggle flag is off, then order of mappers is executed the same as today
if toggle flag is on, then order of mappers is executed by the numbers we setted

---
### What I changed (backend + Admin UI)
New IdP toggle: config.mapperOrderEnabled (boolean string "true"/"false") on the Identity Provider.
OFF (default): mapper execution order is unchanged (legacy order).
ON: mappers execute in ascending numeric order using each mapper’s config.mapperOrder.
New per-mapper field: config.mapperOrder (number-as-string, e.g. "1", "2", ...).
Missing/invalid/negative values are treated as “no order” and run after ordered mappers.

### Where execution order is applied
Mapper execution is now ordered (when enabled) for:
Broker login flow (preprocess/import/update) in IdentityBrokerService
SAML SP metadata export mappers in SAMLIdentityProvider
Token-exchange broker flows in AbstractTokenExchangeProvider
Central logic lives in server-spi-private/.../IdentityProviderMapperExecutionOrder.java.

### Admin UI changes
In the IdP Mappers tab:
Added “Mapper order enabled” toggle (saves immediately)
Added a Mapper order column and sorts the list by it when enabled
In the Add/Edit Mapper form:
Added Mapper order numeric field (config.mapperOrder)


ahus1:
@zivadler - I see you've opened several issues and pull requests. As you and your colleagues reached out to me, I provided the advice to do PRs one by one, and start with a medium-sized PR that has high value to you. Once that is merged, start on the next one.

Issue #45851, opened 2026-01-29
[OID4CVI] Review and fix issuance on public clients


tdiesler:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

oid4vc

### Describe the bug

All clients imported to the test realm are confidential. An external wallet will however in most cases not be trusted to keep the `client_secret` - oid4vci clients should be configured as public with pkce enabled.

### Version

latest

### Expected behavior

Credentials can be issued from public clients that have pkce enabled

### Actual behavior

There is no test coverage for public clients

### How to Reproduce?

Inspect the testrealm.json

mposolda:
@tdiesler Is this issue already addressed as part of the other work? It is not clear from the description what was exactly the purpose of this

tdiesler:
> [@tdiesler](https://github.com/tdiesler) Is this issue already addressed as part of the other work? It is not clear from the description what was exactly the purpose of this

Please keep it open for now - I still want to check a few things.

PR #45855, opened 2026-01-29
[OID4VCI] Initial public client for credential issuance

### LINKED ISSUE #47280: [OID4VCI] Initial public client for credential issuance
### Description

All clients imported to the test realm are confidential. An external wallet will however in most cases NOT be trusted to keep the client_secret.

In the new test suite we also need a public client that has pkce enabled.

tdiesler:
closes #47280

The PR ...

* adds a public client `test-app-pub` to testrealm.json
* adds a comprehensive test using the new OAuthClient APIs (i.e. OID4VCPublicClientTest)
* adds support for error page/redirect to the AuthorizationRequest
* removes dependency on `OID4VCIssuerEndpointTest.getCredentialOfferUriUrl()`

### DIFF
--- a/tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
+++ b/tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
@@ -12,6 +12,7 @@ import java.util.Set;
 import org.keycloak.admin.client.Keycloak;
 import org.keycloak.admin.client.resource.RealmResource;
 import org.keycloak.admin.client.resource.UserResource;
+import org.keycloak.crypto.KeyWrapper;
 import org.keycloak.jose.jws.JWSHeader;
 import org.keycloak.jose.jws.JWSInput;
 import org.keycloak.jose.jws.JWSInputException;
@@ -22,6 +23,8 @@ import org.keycloak.protocol.oid4vc.model.CredentialResponse;
 import org.keycloak.protocol.oid4vc.model.CredentialResponse.Credential;
 import org.keycloak.protocol.oid4vc.model.CredentialsOffer;
 import org.keycloak.protocol.oid4vc.model.OID4VCAuthorizationDetail;
+import org.keycloak.protocol.oid4vc.model.ProofType;
+import org.keycloak.protocol.oid4vc.model.Proofs;
 import org.keycloak.representations.JsonWebToken;
 import org.keycloak.representations.idm.UserRepresentation;
 import org.keycloak.testframework.oauth.OAuthClient;
@@ -44,7 +47,9 @@ import static org.keycloak.OAuth2Constants.AUTHORIZATION_DETAILS;
 import static org.keycloak.constants.OID4VCIConstants.CREDENTIAL_OFFER_CREATE;
 import static org.keycloak.tests.oid4vc.OID4VCIssuerTestBase.TEST_PASSWORD;
 import static org.keycloak.tests.oid4vc.OID4VCIssuerTestBase.VCTestRealmConfig.TEST_REALM_NAME;
+import static org.keycloak.tests.oid4vc.OID4VCProofTestUtils.createEcKeyPair;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.ACCESS_TOKEN_RESPONSE_ATTACHMENT_KEY;
+import static org.keycloak.tests.oid4vc.OID4VCTestContext.AttachmentKey;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIALS_OFFER_ATTACHMENT_KEY;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIAL_OFFER_URI_ATTACHMENT_KEY;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIAL_RESPONSE_ATTACHMENT_KEY;
@@ -186,6 +191,24 @@ public class OID4VCBasicWallet {
             }
         };
         return request;
+    }
+
+    public Proofs generateJwtProof(OID4VCTestContext ctx, String user) {
+        String aud = getIssuerMetadata(ctx).getCredentialIssuer();
+        String nonce = oauth.oid4vc().nonceRequest().send().getNonce();
+        KeyWrapper kw = getECKeyPair(ctx, user, null);
+        return Proofs.create(ProofType.JWT, OID4VCProofTestUtils.generateJwtProof(aud, kw, nonce));
+    }
+
+    public KeyWrapper getECKeyPair(OID4VCTestContext ctx, String user, String keyId) {
+        String cacheKey = user + (keyId != null ? "_" + keyId : "");
+        AttachmentKey<KeyWrapper> attachmentKey = new AttachmentKey<>(cacheKey, KeyWrapper.class);
+        KeyWrapper kw = ctx.getAttachment(attachmentKey);
+        if (kw == null) {
+            kw = createEcKeyPair();
+            ctx.putAttachment(attachmentKey, kw);
+        }
+        return kw;
     }
 
     public AccessTokenRequest accessTokenRequest(OID4VCTestContext ctx, String authCode) {
--- a/tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
+++ b/tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
@@ -25,10 +25,10 @@ import org.keycloak.protocol.oid4vc.model.CredentialsOffer;
 import org.keycloak.protocol.oid4vc.model.OID4VCAuthorizationDetail;
 import org.keycloak.protocol.oid4vc.model.ProofType;
 import org.keycloak.protocol.oid4vc.model.Proofs;
+import org.keycloak.protocol.oidc.representations.OIDCConfigurationRepresentation;
 import org.keycloak.representations.JsonWebToken;
 import org.keycloak.representations.idm.UserRepresentation;
 import org.keycloak.testframework.oauth.OAuthClient;
-import org.keycloak.testsuite.util.oauth.AbstractOAuthClient;
 import org.keycloak.testsuite.util.oauth.AccessTokenRequest;
 import org.keycloak.testsuite.util.oauth.AccessTokenResponse;
 import org.keycloak.testsuite.util.oauth.AuthorizationEndpointResponse;
@@ -49,9 +49,10 @@ import static org.keycloak.tests.oid4vc.OID4VCIssuerTestBase.TEST_PASSWORD;
 import static org.keycloak.tests.oid4vc.OID4VCIssuerTestBase.VCTestRealmConfig.TEST_REALM_NAME;
 import static org.keycloak.tests.oid4vc.OID4VCProofTestUtils.createEcKeyPair;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.ACCESS_TOKEN_RESPONSE_ATTACHMENT_KEY;
+import static org.keycloak.tests.oid4vc.OID4VCTestContext.AUTHORIZATION_SERVER_METADATA_ATTACHMENT_KEY;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.AttachmentKey;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIALS_OFFER_ATTACHMENT_KEY;
-import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIAL_OFFER_URI_ATTACHMENT_KEY;
+import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIALS_OFFER_URI_ATTACHMENT_KEY;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.CREDENTIAL_RESPONSE_ATTACHMENT_KEY;
 import static org.keycloak.tests.oid4vc.OID4VCTestContext.ISSUER_METADATA_ATTACHMENT_KEY;
 
@@ -74,8 +75,8 @@ import static org.junit.jupiter.api.Assertions.fail;
  */
 public class OID4VCBasicWallet {
 
-    final Keycloak keycloak;
-    final OAuthClient oauth;
+    private final Keycloak keycloak;
+    private final OAuthClient oauth;
 
     final Set<String> loginUsers = new HashSet<>();
 
@@ -84,25 +85,26 @@ public class OID4VCBasicWallet {
         this.oauth = oauth;
     }
 
-    // Composite Actions -----------------------------------------------------------------------------------------------
-
-    public CredentialsOffer createAuthCodeCredentialOffer(OID4VCTestContext ctx, String targetUser) {
+    /**
+     * A composite action to create a credential offer with 'authorization_code' grant
+     */
+    public CredentialsOffer createCredentialOfferAuthCode(OID4VCTestContext ctx, String targetUser) {
 
         // Get Issuer AccessToken
         //
-        AccessTokenResponse issTokenResponse = getIssuerAccessToken(ctx.getIssuer());
+        AccessTokenResponse issTokenResponse = getAccessToken(ctx.getIssuer());
 
         // Exclude scope: <credScope>
         // Require role: credential-offer-create
         String issToken = validateIssuerAccessToken(issTokenResponse,
-                List.of(), List.of(ctx.getCredScopeName()),
+                List.of(), List.of(ctx.getScope()),
                 List.of(CREDENTIAL_OFFER_CREATE.getName()), List.of());
 
         // Create Authorized Code CredentialOffer
         //
         CredentialOfferURI credOfferUri;
         try {
-            credOfferUri = createCredentialOffer(ctx, ctx.getCredConfigId())
+            credOfferUri = createCredentialOffer(ctx, ctx.getCredentialConfigurationId())
                     .preAuthorized(false)
                     .targetUser(targetUser)
                     .bearerToken(issToken)
@@ -113,7 +115,7 @@ public class OID4VCBasicWallet {
 
         // Fetch the CredentialsOffer
         //
-        CredentialsOffer credOffer = getCredentialOffer(ctx, credOfferUri)
+        CredentialsOffer credOffer = credentialsOfferRequest(ctx, credOfferUri)
                 .send().getCredentialsOffer();
 
         String issuerState = credOffer.getIssuerState();
@@ -122,24 +124,27 @@ public class OID4VCBasicWallet {
         return credOffer;
     }
 
-    public CredentialsOffer createPreAuthCredentialOffer(OID4VCTestContext ctx, String targetUser) {
+    /**
+     * A composite actions to create a credential offer with 'pre-authorized_code' grant
+     */
+    public CredentialsOffer createCredentialOfferPreAuth(OID4VCTestContext ctx, String targetUser) {
 
         // Get Issuer AccessToken
         //
-        AccessTokenResponse issTokenResponse = getIssuerAccessToken(ctx.getIssuer());
+        AccessTokenResponse issTokenResponse = getAccessToken(ctx.getIssuer());
         assertNotNull(issTokenResponse.getAccessToken(), "No accessToken");
 
         // Exclude scope: <credScope>
         // Require role: credential-offer-create
         String issToken = validateIssuerAccessToken(issTokenResponse,
-                List.of(), List.of(ctx.getCredScopeName()),
+                List.of(), List.of(ctx.getScope()),
                 List.of(CREDENTIAL_OFFER_CREATE.getName()), List.of());
 
         // Create Pre-Authorized CredentialOffer
         //
         CredentialOfferURI credOfferUri;
         try {
-            credOfferUri = createCredentialOffer(ctx, ctx.getCredConfigId())
+            credOfferUri = createCredentialOffer(ctx, ctx.getCredentialConfigurationId())
                     .preAuthorized(true)
                     .targetUser(targetUser)
                     .bearerToken(issToken)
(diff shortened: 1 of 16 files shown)


tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
```diff
@@ -374,8 +380,54 @@ public AuthorizationEndpointRequest scope(String... scopes) {
...
+                    params.put("error", "invalid_request");
+                    driver.findElements(By.id("kc-error-message")).stream()
+                            .map(d -> d.getText().trim().split("\\n")[0])
+                            .forEach(s -> params.put("error_description", s));
+                    return new AuthorizationEndpointResponse(params);
+                }
+                // AuthPageState.LOGIN
+                default -> {
+                    client.fillLoginForm(username, password);
+                    return new AuthorizationEndpointResponse(client);
```
Ogenbertrand: This bypasses OAuthClient.parseLoginResponse(), which is where the test-framework client waits for the OAuth callback. Returning a freshly constructed AuthorizationEndpointResponse here can read the current URL before the redirect completes and make existing wallet-based flows flaky.

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCPublicClientTest.java
```diff
@@ -0,0 +1,176 @@
...
+        //
+        CredentialsOffer credOffer = wallet.createPreAuthCredentialOffer(ctx, ctx.holder);
+        String preAuthCode = credOffer.getPreAuthorizedCode();
+
+        // Redeem Pre-Authorized Code for AccessToken
+        //
+        AccessTokenResponse tokenResponse = wallet.preAuthAccessTokenRequest(ctx, preAuthCode).send();
+        assertTrue(tokenResponse.isSuccess(), tokenResponse.getErrorDescription());
+
+        String accessToken = wallet.validateHolderAccessToken(ctx, tokenResponse);
```
Ogenbertrand: validateHolderAccessToken() asserts authorization_details in both the token response and the JWT. The PR description calls out missing pre-auth authorization_details support as a follow-up issue, so this test may currently depend on behavior that is not fully implemented yet.

tests/base/src/test/java/org/keycloak/tests/oid4vc/RunOnServerUtils.java
```diff
@@ -0,0 +1,171 @@
...
+                30
+        );
+    }
+
+    static OID4VCIssuerEndpointTest.OAuth2CodeEntry prepareSessionCode(
+            KeycloakSession session,
+            AppAuthManager.BearerTokenAuthenticator authenticator,
+            String note) {
+
+        AuthenticationManager.AuthResult authResult = authenticator.authenticate();
```
Ogenbertrand: This shared helper still depends on OID4VCIssuerEndpointTest via the return type. Moving OAuth2CodeEntry into RunOnServerUtils, or returning a local record from this class, would make the extraction self-contained and easier to reuse.

Ogenbertrand (changes requested):
Thanks for this PR @tdiesler !
I added a few comment, could you check them.

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
tdiesler: thanks

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCPublicClientTest.java
tdiesler: A lot has happened since this PR was created in Jan. When it eventually gets merged, I'll investigate the state of affairs with access tokens from pub clients. For now it's just about initial setup and happy-path usage.

tests/base/src/test/java/org/keycloak/tests/oid4vc/RunOnServerUtils.java
tdiesler: Could you perhaps post this on #47212

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCPublicClientTest.java
Ogenbertrand: A TODO / follow-up issue reference will be good.
WDYT ?

tdiesler: This PR now closes #47280 (i.e. just the initial setup)
We can keep the original issue #45851 open & I added the findings/reminders there, so we won't forget to actually fix stuff in a follow-up.

Ogenbertrand (approved):
I confirm that my request were all resolved.
Thanks @tdiesler for considering them, PR LGTM.

tests/utils-shared/pom.xml
```diff
@@ -31,11 +31,20 @@
     <packaging>jar</packaging>
     <description>Keycloak Testsuite Utilities shared between legacy and new testsuites</description>
 
+    <properties>
+        <selenium.version>4.28.1</selenium.version>
```
mposolda: Is it needed to update selenium in this PR?

Selenium update is good, but should be rather done in separate PR (rather than inside the OID4VC specific PR) and should be likely after Keycloak 26.6 to avoid any unexpected issues with testsuite stability closely before release.

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCTestContext.java
```diff
@@ -46,7 +46,6 @@ public class OID4VCTestContext {
 
     public OID4VCTestContext(ClientRepresentation client, CredentialScopeRepresentation credentialScope) {
         this.client = client;
-        this.clientId = client.getClientId();
```
mposolda, line 49: Nitpick: It seems that `clientId` variable can be removed entirely from this class?

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
```diff
@@ -390,8 +396,54 @@ public AuthorizationEndpointRequest scope(String... scopes) {
...
+                case ERROR_REDIRECT -> {
+                    return new AuthorizationEndpointResponse(client);
+                }
+                case ERROR_PAGE -> {
+                    Map<String, String> params = new HashMap<>();
+                    params.put("error", "invalid_request");
+                    driver.findElements(By.id("kc-error-message")).stream()
+                            .map(d -> d.getText().trim().split("\\n")[0])
+                            .forEach(s -> params.put("error_description", s));
+                    return new AuthorizationEndpointResponse(params);
```
mposolda: The `AuthorizationEndpointResponse` is the type, which is dedicated for the OIDC response sent after the redirect from Keycloak back to the client application. It is rather confusing to mis-use for the different purpose (like the case when Keycloak does not redirect back to client, but shows the error).

Is it rather possible to use some different approach? For example returning `AuthPageState` from the `OID4VCBasicWallet.send` method and making `AuthorizationEndpointResponse` as a variable within `AuthPageState` (which means that for `ERROR_PAGE`, the response instance would be null and there would be some different variable within `AuthPageState`, so that caller would be able to easily differentiate whether there was error redirect or Keycloak errorPage + check the appropriate error, which was shown by Keycloak etc).

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
```diff
@@ -390,8 +396,54 @@ public AuthorizationEndpointRequest scope(String... scopes) {
...
+                    client.fillLoginForm(username, password);
+                    return client.parseLoginResponse();
+                }
+            }
+        }
+
+        enum AuthPageState {LOGIN, ERROR_REDIRECT, ERROR_PAGE}
+
+        private AuthPageState waitForLoginOrError(WebDriver driver) {
+            WebDriverWait wait = new WebDriverWait(driver, Duration.ofSeconds(10));
```
mposolda: Is it possible to use `WaitUtils.until` utility instead of creating `WebDriverWait` instance manually this way? Could be good as if we later want to have the duration timeouts configurable, will be nice to be able to update it in single place.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationEndpointResponse.java
```diff
@@ -35,6 +35,10 @@ public AuthorizationEndpointResponse(AbstractOAuthClient<?> client) {
         }
     }
 
+    public AuthorizationEndpointResponse(Map<String, String> params) {
```
mposolda: I think this should not be needed (see my other comment inside `OID4VCBasicWallet`).

mposolda (commented):
@tdiesler Thanks for this PR! Added few comments inline

tdiesler:
@mposolda could we possibly do these changes as a follow-up? This PR is the initial setup so that we can work with public clients. Fixing stuff related to public clients is #45851 - I can make a note there about these change requests.

mposolda:
> @mposolda could we possibly do these changes as a follow-up? This PR is the initial setup so that we can work with public clients. Fixing stuff related to public clients is #45851 - I can make a note there about these change requests.

Selenium update would be good to address as a follow-up and dedicated issue. But if this PR needs selenium update, then I suppose it should be a pre-requisite rather than a follow-up? But that would be then after Keycloak 26.6.0 release (I hope shortly after, so hopefully next week).

Other issues will be better to address within this PR.

tests/utils-shared/pom.xml
tdiesler: selenium isn't updated by this PR - I moved the version definition to the top-level pom

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCTestContext.java
tdiesler: done - revisited this and other convenience methods

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationEndpointResponse.java
tdiesler: it is unfortunately needed for the invalid credentials case

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
tdiesler: yes, this could be simplified

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
tdiesler: The Authorization request/response handling follows the same pattern as all the other Foo request/response objects. It is not just a response object that is populated on successful login, like the other APIs it also captures the error cases, namely ...

* login page does not get displayed, because of an invalid request
* invalid credentials on login page (i.e. redirect url without code)

In all cases, we don't want to wait for a lengthy timeout, instead return an AuthorizationEndpointResponse with  success or error details as soon as we have it.

You are right however, I could simplify that handling a great deal by using the injected ManagedWebDriver. 

Anyway, the authorization through the wallet works well for success and error cases without exposure to the low level details of web driver and login forms. In future, we can use the same API for non-interactive authorization flows - as required by ebsi for example. 

Please have a look at `OID4VCPublicClientTest` there we have the (three) use cases that are considered (i.e. success + 2x error cases) - these three should work without running into a timeout.

tdiesler:
> Selenium update would be good to address

The PR doesn't do a selenium update, it reuses the version from the old testsuite

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
mposolda: Sure, I agree it can be good if there is no timeout.

But there is no need to use `AuthorizationEndpointResponse` for handling the case when the Keycloak shows the error page. That class is used for handling the redirect back to the client as specified in OAuth2 https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2 . Not for showing the error pages at the Keycloak side.

Could be nice to return something different than `AuthorizationEndpointResponse` from the `OID4VCBasicWallet.send` , but rather some object (for example `OID4VCAuthResponse`, which could be for example something like:

```
public class OID4VCAuthResponse {
    
    private final AuthorizationEndpointResponse authzResponse;
    private final String loginPageError;

    public OID4VCAuthResponse(AuthorizationEndpointResponse authzResponse) {
        // Case when there was redirect to the client (Either with "code" or error redirect with "error" and "error_Description" parameters
       this.authzResponse = authzResponse;
       this.loginPageError = null;
   }

    public OID4VCAuthResponse(String loginPageError) {
        // Case when there was login page error
       this.authzResponse = null;
       this.loginPageError = loginPageError;
   }

```
Then the calling test would be able to differentiate between the case when there was redirect to client or the case when error shown on the Keycloak side. That is aligned with other Keycloak tests.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationEndpointResponse.java
mposolda: Please see https://github.com/keycloak/keycloak/pull/45855#discussion_r3009527649

tests/utils-shared/pom.xml
mposolda: It is of course updated in this PR in some modules (Especially in the whole new testsuite, which inherits the version from `test-utils-shared`)

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
```diff
@@ -410,9 +432,39 @@ public void openLoginForm() {
...
+                        return new AuthorizationEndpointResponse(oauth);
+                    }
+
+                    // Error response
+                    if (auxUrl.contains("error=") || auxUrl.contains("error_description=")) {
+                        return new AuthorizationEndpointResponse(oauth);
+                    }
+
+                    // No code and no error in response
+                    return new AuthorizationEndpointResponse(Map.of(
```
mposolda: Could you check if `Invalid username or password` is really shown on the page? After submit username/password, there could be many other states happening instead of showing the error (EG. Keycloak could have displayed some additional screen before it is redirected back to the user). So would be good to differentiate instead of returning `Invalid username or password` when it is actually not really displayed.

mposolda (commented):
@tdiesler Thanks for the updates. Added additional inline comment and replied to some of your earlier comments.

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
tdiesler: The login page just gets re-displayed with some text in red that isn't identifiable in selenium (afaik) - we could instead use a more generic error description if you like. However, it the case where the login page got displayed and then some input or verification error happens - we do not deal with retries.

mposolda: Yes, we can maybe use something generic. Not sure about "generic error description" as in generic case, it is not certain that there is any error displayed. For example there can be the case when username/password is correct, but there is some follow-up page displayed by Keycloak to the user (EG. OTP authentication or required-action with credential-offer or whatever).

If you prefer, it works for me to instead return `null` in case that there is no code and no error in response. Then it works for me to keep returning `AuthorizationEndpointResponse` from method `send` if you prefer that approach. As then the test can do for example something like:
```
response = wallet.send(username, password);
assertNull(response); // User not redirected yet to the client
loginPage.assertCurrent();
assertEquals('invalid username or password", loginPage.getError());
```

tdiesler: I prefer to always return a response object, like all the other APIs on OAuthClient and OID4CVIClient

mposolda: Yes, ok. Then this applies https://github.com/keycloak/keycloak/pull/45855#discussion_r3009527649

mposolda: And for this one, it still applies, that it should not return an `error` as it is not 100% that some error happened. As I've mentioned, after confirm username/password, there may not be the error, but just some additional Keycloak screen. If you are not able to check that error happened, it may work to return something like `OID4VCAuthResponse.notAuthzResponseRedirect()` .

That should work for any case when context is still on some Keycloak screen and it would be up to the caller to test this further (EG. check if there is error on login screen or there is some additional Keycloak screen).

tdiesler: On the Wallet, I want it to work like this ...

```
        AuthorizationEndpointResponse authResponse = wallet
                .authorizationRequest()
                .scope(ctx.getScope())
                .send(ctx.getHolder(), TEST_PASSWORD);

        assertEquals("invalid_request", authResponse.getError());
        assertEquals("Missing parameter: code_challenge_method", authResponse.getErrorDescription());
```

and this ...

```
        AuthorizationEndpointResponse authResponse = wallet
                .authorizationRequest()
                .scope(ctx.getScope())
                .send(ctx.getHolder(), "xxx");

        assertEquals("unauthorized", authResponse.getError());
        assertNull(authResponse.getErrorDescription(), "Null error description");
```

in future, we will have ...

```
        AuthorizationEndpointResponse authResponse = wallet
                .authorizationRequest()
                .scope(ctx.getScope())
                .someOtherRequestHeadersOrParams(...)
                .send();
```

in all cases, I don't want the caller to deal with login forms, redirect parameters and whatnot. Instead, the response should report success or error (with optional details). If the Authorization Code Flow is being tested, the caller can use the existing API and does not have to go through the Wallet.

Also, let's not forget that this PR is about credential issuance through public clients and not about login form error handling - we can do this later (if really needed).

tdiesler:
This PR is at the bottom of so many important thing, that I hope to finally unblock it by removing all changes to Wallet.authorizationRequest(). The follow-up issue is here ...

* https://github.com/keycloak/keycloak/issues/47649

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
mposolda: As I've mentioned many times, the `AuthorizationEndpointResponse` is used for the cases when the Keycloak redirects to the OAuth client. It should **not** be used for the cases when Keycloak does not redirect to the client, but is still kept on the login screen or on some additional Keycloak screen.

From your examples, using `AuthorizationEndpointResponse` for the first example works. But for the second it does not work as it abuses `AuthorizationEndpointResponse` for different purpose than it was intended.

The test caller should now if it expects redirect to the client or still be kept on some Keycloak screen. If you want to handle both redirect-to-client + still-kept-on-Keycloak, you should use different type for the API than `AuthorizationEndpointResponse` . Like for example the `OID4VCAuthResponse` as I've pointed.

If you prefer backwards compatibility for the case when redirect to the client is expected, it may work something like this in `OID4VCBasicWallet`:
```
// Backwards compatibility for the case when redirect to client expected
public AuthorizationEndpointResponse send(username, password) {
    OID4VCAuthResponse authResponse = send(username, password, true);
    return authResponse.getAuthzResponse();
}

public OID4VCAuthResponse send(username, password, boolean expectRedirectToClient) {
   OID4VCAuthResponse response = 
   ... The code you had originally, which is able to deal with redirect to client OR keep the context on Keycloak...

   if (expectRedirectToClient) {
       assertNotNull(response.getAuthzResponse());
   } else {
      assertNull(response.getAuthzResponse());
   }
}

public static class OID4VCAuthResponse {
    
    private final AuthorizationEndpointResponse authzResponse;
    private final String loginPageError;

    public OID4VCAuthResponse(AuthorizationEndpointResponse authzResponse) {
        // Case when there was redirect to the client (Either with "code" or error redirect with "error" and "error_Description" parameters
       this.authzResponse = authzResponse;
       this.loginPageError = null;
   }

    public OID4VCAuthResponse(String loginPageError) {
        // Case when there was login page error
       this.authzResponse = null;
       this.loginPageError = loginPageError;
   }

   ... getters etc ...
}
```

Then for the case when you don't expect redirect to client, the caller can handle it like:
```
OID4VCAuthResponse oid4vcAuthResponse = wallet.send(username, password, false);
loginPage.assertCurrent();
assertEquals("invalid username or password", loginPage.getError());
```

mposolda (commented):
@tdiesler Thanks for the updates.

The last remaining thing in this PR are changes in `pom.xml` around selenium. Could you create dedicated GH issue and dedicated PR just for those changes? As this may need to be reviewed by some other members of Keycloak team and should not be done in the same PR, which is related to OID4VCI.

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
tdiesler: This code is gone, please make your peace with it. Improvements to the "error only after timeout" can be made here https://github.com/keycloak/keycloak/issues/47649

mposolda: Yes, I know. This is rather just something for the future if you still want to support something like this. Added comment to #47649 .

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationEndpointResponse.java
mposolda: This constructor would be good to remove from this PR (Sorry as I've overlooked it before that it's still kept).

mposolda (commented):
LGTM with a minor comment for `AuthorizationEndpointResponse` constructor, which is not used anymore

tdiesler:
> LGTM with a minor comment for `AuthorizationEndpointResponse` constructor, which is not used anymore

done

mposolda (approved):
@tdiesler Thanks!

PR #45883, opened 2026-01-30
use group resource from context instead

edewit:
fixes: #45882

### DIFF
--- a/js/apps/admin-ui/src/context/group-resource/GroupResourceContext.tsx
+++ b/js/apps/admin-ui/src/context/group-resource/GroupResourceContext.tsx
@@ -1,0 +1,27 @@
+import {
+  createNamedContext,
+  useRequiredContext,
+} from "@keycloak/keycloak-ui-shared";
+import { Groups } from "@keycloak/keycloak-admin-client";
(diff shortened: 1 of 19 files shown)


js/apps/admin-ui/src/context/group-resource/GroupResourceContext.tsx
```diff
@@ -0,0 +1,27 @@
+import {
+  createNamedContext,
+  useRequiredContext,
+} from "@keycloak/keycloak-ui-shared";
+import { Groups } from "@keycloak/keycloak-admin-client";
+import { PropsWithChildren } from "react";
+
+export const GroupsResourceContext = createNamedContext<Groups | undefined>(
+  "GroupResourceContest",
```
ssilvert:
```suggestion
  "GroupResourceContext",
```

ssilvert (changes requested):
I think `components/group/GroupPickerDialog.tsx` also needs to change?

Also, I'm a little unclear about how to use this.  I was expecting maybe a new constructor on the Groups resource that would let me specify a different path?

Let's discuss offline.

ssilvert (approved):

PR #45910, opened 2026-01-30
Added theme descriptions, modified messages_en

### LINKED ISSUE #45909: Add theme clarification blurb to Realm Settings
### Description

Add a small blurb to the themes selection page on the Admin panel. Provides info about theme styles and stating default themes when none are selected.

### Value Proposition

Current theme names do not convey info about their styles and default themes are not stated except in change logs.

Users have to enable the theme, restart their project, and explore to get an idea of the theme. 
By providing small descriptions, developers can save time when working on large scale projects with long build and relaunch times.

### Goals

The enhancement should provide one or two short sentences about each theme, just above the theme selection pane. This change is only in regards to a small text description for the user to read (and adding the descriptions to `messages_en.properties`).

### Non-Goals

The text should not be distracting or detract from the UI. Theme names, styling, or any other attribute around the themes will not change.

### Discussion

_No response_

### Notes

Descriptions and the title will be added to `messages_en.properties`, and displayed to the user in `ThemeSettings.tsx` via an `<EmptyStateBody>` class as to mimic the ACR to LoA Mapping styling in `GeneralTab.tsx`.

deweesevs:
Descriptions added to messages_en.properties. They are displayed to the user in via `SelectControl` in  `ThemeSettings.tsx`.

Closes #45909

### DIFF
--- a/js/apps/admin-ui/src/realm-settings/themes/ThemeSettings.tsx
+++ b/js/apps/admin-ui/src/realm-settings/themes/ThemeSettings.tsx
@@ -57,6 +57,7 @@ export const ThemeSettingsTab = ({ realm, save }: ThemeSettingsTabProps) => {
               themeTypes.login.map((theme) => ({
                 key: theme.name,
                 value: theme.name,
+                description: theme.description,
               })),
             )}
           />
@@ -71,6 +72,7 @@ export const ThemeSettingsTab = ({ realm, save }: ThemeSettingsTabProps) => {
               themeTypes.account.map((theme) => ({
                 key: theme.name,
                 value: theme.name,
+                description: theme.description,
               })),
             )}
           />
@@ -85,6 +87,7 @@ export const ThemeSettingsTab = ({ realm, save }: ThemeSettingsTabProps) => {
               themeTypes.admin.map((theme) => ({
                 key: theme.name,
                 value: theme.name,
+                description: theme.description,
               })),
             )}
           />
@@ -99,6 +102,7 @@ export const ThemeSettingsTab = ({ realm, save }: ThemeSettingsTabProps) => {
               themeTypes.email.map((theme) => ({
                 key: theme.name,
                 value: theme.name,
+                description: theme.description,
               })),
             )}
           />
--- a/js/libs/ui-shared/src/controls/select-control/SelectControl.tsx
+++ b/js/libs/ui-shared/src/controls/select-control/SelectControl.tsx
@@ -19,6 +19,7 @@ export enum SelectVariant {
 export type SelectControlOption = {
   key: string;
   value: string;
+  description?: string;
 };
 
 export type OptionType = string[] | SelectControlOption[];
--- a/services/src/main/java/org/keycloak/services/resources/admin/info/ServerInfoAdminResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/info/ServerInfoAdminResource.java
@@ -243,8 +243,7 @@ public class ServerInfoAdminResource {
                             ti.setLocales(locales.replaceAll(" ", "").split(","));
                         }
 
-                        String description = getThemeDescription(theme);
-                        ti.setDescription(description);
+                        ti.setDescription(getThemeDescription(theme));
 
                         themes.add(ti);
                     }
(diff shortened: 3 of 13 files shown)


js/apps/admin-ui/src/realm-settings/themes/ThemeSettings.tsx
```diff
@@ -1,6 +1,12 @@
 import type RealmRepresentation from "@keycloak/keycloak-admin-client/lib/defs/realmRepresentation";
 import { SelectControl } from "@keycloak/keycloak-ui-shared";
-import { ActionGroup, Button, PageSection } from "@patternfly/react-core";
+import {
```
deweesevs: The pre-commit hook that runs the linter forces this to be multiple lines.

js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
```diff
@@ -3169,6 +3169,10 @@ generatedCodeTextArea=Generated code textarea
 selectAuthScopes=Select authorization scopes
 selectResource=Select resource
 selectClients=Select clients
+themesDescriptionsTitle=Theme descriptions
```
deweesevs: Theme description strings added above the theme selection strings, as they are related.

deweesevs:
The visual changes can be seen here:
image

ahus1 (changes requested):
Thank you for this PR. While the descriptions are correct, this misses that users can add custom themes to Keycloak, and therefore the list of themes is dynamic. 

To have a description per theme, consider the following: 

* Extending the theme.properties files for those three themes with an additional key `description`
* Provide means that this information is pulled from the backend and shown in the frontend.

A drop-down that shows the key plus a description like we do for example when creating a client policy might be a possible UX solution: 

Image

deweesevs:
@ahus1 This is a good alternative that reduces clutter, thank you for the feedback. I will look into this

js/libs/ui-shared/src/controls/select-control/SelectControl.tsx
```diff
@@ -19,6 +19,7 @@ export enum SelectVariant {
 export type SelectControlOption = {
   key: string;
   value: string;
+  description?: string;
```
deweesevs, line 22: Optional description so that previous implementations do not break.

deweesevs:
To reduce code clutter, (and get 2 birds with 1 stone), I think we could efficiently modify `SelectControl.tsx` to support `key, value, description` rendering (currently `key, value`). Backwards compatibility is ensured by making description optional. 

If you believe modifying `theme.properties` files would be a better approach than this, please let me know. However I'm unsure how to route `theme.properties` to `ThemeSettings.tsx` while enabling translation.

With the changes, the dropdown now looks like this:

image

ahus1 (changes requested):
Thank you for the updates. As mentioned earlier, there can be any number of custom themes. 

> If you believe modifying theme.properties files would be a better approach than this, please let me know. 

Due to the above, please move the description of the theme to the theme.properties for now. 

> However I'm unsure how to route theme.properties to ThemeSettings.tsx while enabling translation.

I'd say translation of theme descriptions is out of scope for this PR. We'll need to discuss ideas in a follow-up PR. While in a first version no translation would be provided - similar to some other cases in Keycloak - we can work on a solution where the backend provides a translated version from the backend. 

Let's first get the frontend prepared for that, and find the right reviewers for that.

services/src/main/java/org/keycloak/services/resources/admin/info/ServerInfoAdminResource.java
```diff
@@ -249,6 +243,9 @@ private void setThemes(ServerInfoRepresentation info) {
                             ti.setLocales(locales.replaceAll(" ", "").split(","));
                         }
 
+                        String description = getThemeDescription(theme);
+                        ti.setDescription(description);
```
deweesevs: Here is where we pulled the descriptions from the backend to the frontend.

deweesevs:
output

I have moved the descriptions from the translation file into each theme's property files. `Default for X` message only appears when it is the default theme for that drop-down.

I also updated `ServerInfoAdminResource.java` so that we can pass the descriptions from the back-end to the front-end.

ahus1 (approved):
Thank you for his PR. I simplified the error handling a bit, as an IO Exception should not result in a null string being silently returned. The parent method is already handling IOExceptions, and it should be done IMHO consistently. 

Let's see if the tests are green, and then it should merge. If you want to work in a follow-up PR to also provide translations for the theme descriptions, let me know and we can discuss that.

PR #45913, opened 2026-01-31
UserProfile related tests may see unexpected 'did' property

### LINKED ISSUE #45912: UserProfile related tests may see unexpected 'did' property
### Before reporting an issue

### Area

user-profile

### Describe the bug

```
-- Expected: iterable with items ["username", "email", "firstName", "lastName", "locale", "address", "second"] in any order
--      but: not matched: "did"
-- 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
-- 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
-- 	at org.keycloak.testsuite.user.profile.UserProfileTest.testGetProfileAttributeGroups(UserProfileTest.java:572)
```

This is probably related to how we handle the UserModel.DID attribute in `DeclarativeUserProfileProviderFactory`. Is it possible that UP definitions get persisted (or stored in memory) and therefore can leak from one test to another?

### Version

latest

### Regression

### Expected behavior

The `did` attribute should only get added to the default user profile when `--features=oid4vc-vci` is enabled

### Actual behavior

UserProfile related tests may fail (on GitHub)

### How to Reproduce?

PR validation may fail

### Anything else?

_No response_

tdiesler:
closes #45912

### DIFF
--- a/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
+++ b/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
@@ -70,11 +70,7 @@ import org.keycloak.userprofile.validator.UsernameMutationValidator;
 import org.keycloak.utils.StringUtil;
 import org.keycloak.validate.ValidatorConfig;
 import org.keycloak.validate.validators.EmailValidator;
-import org.keycloak.validate.validators.PatternValidator;
 
-import static java.util.Optional.ofNullable;
-
-import static org.keycloak.common.Profile.Feature.OID4VC_VCI;
 import static org.keycloak.common.util.ObjectUtil.isBlank;
 import static org.keycloak.userprofile.DefaultAttributes.READ_ONLY_ATTRIBUTE_KEY;
 import static org.keycloak.userprofile.UserProfileContext.ACCOUNT;
@@ -83,8 +79,6 @@ import static org.keycloak.userprofile.UserProfileContext.REGISTRATION;
 import static org.keycloak.userprofile.UserProfileContext.UPDATE_EMAIL;
 import static org.keycloak.userprofile.UserProfileContext.UPDATE_PROFILE;
 import static org.keycloak.userprofile.UserProfileContext.USER_API;
-import static org.keycloak.userprofile.config.UPConfigUtils.ROLE_ADMIN;
-import static org.keycloak.userprofile.config.UPConfigUtils.ROLE_USER;
 
 public class DeclarativeUserProfileProviderFactory implements UserProfileProviderFactory, AmphibianProviderFactory<UserProfileProvider> {
 
@@ -108,7 +102,11 @@ public class DeclarativeUserProfileProviderFactory implements UserProfileProvide
     private final Map<UserProfileContext, UserProfileMetadata> contextualMetadataRegistry = new HashMap<>();
 
     public static void setDefaultConfig(UPConfig defaultConfig) {
-        if (PARSED_DEFAULT_RAW_CONFIG == null) {
+        setDefaultConfig(defaultConfig, false);
+    }
+
+    public static void setDefaultConfig(UPConfig defaultConfig, boolean force) {
+        if (force || PARSED_DEFAULT_RAW_CONFIG == null) {
             PARSED_DEFAULT_RAW_CONFIG = defaultConfig;
         }
     }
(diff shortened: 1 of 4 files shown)


services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
```diff
@@ -539,27 +539,9 @@ private void initDefaultConfiguration(Scope config) {
             parsedConfig = UPConfigUtils.parseSystemDefaultConfig();
         }
 
-        // Modify the user profile to use when --features=oid4vc-vci is enabled
-        if (Profile.isFeatureEnabled(OID4VC_VCI)) {
```
mposolda, line 543: There is probably easier and more proper approach to fix this, which I've overlooked initially --- the `did` attribute could possibly still be added, but should be added just if feature is enabled AND oid4vci-switch is enabled for the realm. So probably something like:
```
RealmModel realm = context.getSession().getContext().getRealm();
if (Profile.isFeatureEnabled(OID4VC_VCI) && realm.isVerifiableCredentialsEnabled()) {
```

This should avoid any interference with the `UserProfileTest` as those tests would not have realm with OID4VCI enabled.

mposolda (commented):
@tdiesler Added comment inline about probably easier approach to fix. Could you take a look?

services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java
tdiesler: This `DeclarativeUserProfileProviderFactory` is a total PITA

I removed modifications to the default config completely

```
    private void initDefaultConfiguration(Scope config) {

        String configFile = config.get("configFile");

        // The user-defined configuration is always parsed during init and should be avoided as much as possible
        // If no user-defined configuration is set, the system default configuration must have been set
        // In Quarkus, the system default configuration is set at build time for optimization purposes
        UPConfig parsedConfig = Optional.ofNullable(configFile)
                .map(Paths::get)
                .map(UPConfigUtils::parseConfig)
                .orElse(PARSED_DEFAULT_RAW_CONFIG);

        // As fallback parse the system default config
        if (parsedConfig == null) {
            parsedConfig = UPConfigUtils.parseSystemDefaultConfig();
        }

        setDefaultConfig(parsedConfig);
    }
```

and instead do it like this in `OID4VCIssuerEndpointTest`. I this way I was hoping to limit the user profile modifications to the oid4vci tests (for now). I still can't get PR #44840 to run successfully. 

```
    private void registerUserProfileProvider(RealmRepresentation testRealm) {

        // Add the User DID attribute, with the same logic as in DeclarativeUserProfileProviderFactory
        //
        UPConfig config = UPConfigUtils.parseSystemDefaultConfig();
        if (config.getAttribute(UserModel.DID) == null) {
            UPAttribute attr = new UPAttribute(UserModel.DID);
            attr.setDisplayName("${did}");
            attr.setPermissions(new UPAttributePermissions(Set.of(ROLE_ADMIN, ROLE_USER), Set.of(ROLE_ADMIN, ROLE_USER)));
            attr.setValidations(Map.of(PatternValidator.ID, Map.of(
                    "pattern", "^did:.+:.+$",
                    "error-message", "Value must start with 'did:scheme:'")));
            config.addOrReplaceAttribute(attr);
        }

        ComponentExportRepresentation componentExportRepresentation = new ComponentExportRepresentation();
        componentExportRepresentation.setProviderId("declarative-user-profile");
        componentExportRepresentation.setId(UUID.randomUUID().toString());
        componentExportRepresentation.setConfig(new MultivaluedHashMap<>(
                Map.of(UP_COMPONENT_CONFIG_KEY, List.of(JsonSerialization.valueAsString(config)))));

        // Remove this UserProfile definition on cleanup
        //
        testRealm.getComponents().add(UserProfileProvider.class.getName(), componentExportRepresentation);
        TestCleanup cleanup = testContext.getOrCreateCleanup(testRealm.getRealm());
        cleanup.addComponentId(componentExportRepresentation.getId());
    }
```

This UP stuff seems to leak either in memory of in some config file across test runs. Is that possible? Also funny, PR #45913 passes every now and then, but not in combination with PR #44840

tdiesler:
out of date

PR #45920, opened 2026-02-01
Add OIDC backchannel lougout configuration to ADMIN REST API

Krematos:
Add supp for configuration backchannel logout URL and req for ClientRepresentation session.

Closes #45761


ahus1:
This solution does not apply, see the comment in the parent issue.

PR #45923, opened 2026-02-02
Show jwks, certificate or publicKey for the key information

### LINKED ISSUE #44284: Client JWKS provided via Dynamic Client Registration is stored and used but not displayed in Admin Console Keys tab
### Before reporting an issue

### Area

_No response_

### Describe the bug

When creating a client through Dynamic Client Registration (DCR) and providing an inline jwks object, Keycloak correctly stores the JWKS and uses it to verify private_key_jwt assertions.
However, the Admin Console does not display the JWKS on the Keys tab, even though:

The exported client JSON contains the JWKS

The Admin REST API returns the JWKS

JWT assertions signed with the private key are successfully verified by Keycloak

This appears to be a UI inconsistency or missing UI mapping for JWKS provided through DCR.

### Version

26.3.3

### Regression

### Expected behavior

The Keys tab should display the JWK from the client’s stored jwks field, or at least indicate that the client contains a JWKS provided via DCR.

### Actual behavior

- The Keys tab shows no keys at all.
- No JWKS is displayed in any of the UI sections.
- However, the JWKS is stored internally (GET /admin/realms/{realm}/clients/{id} includes it).
- JWT assertions using the corresponding private key are successfully validated, proving the backend is reading the   JWKS correctly.

This indicates the issue is limited to the Admin Console UI, not the backend verification logic.

### How to Reproduce?

1. Start Keycloak (tested using official container image).
2. Create an Initial Access Token for the realm.
3. Use DCR to create a client with an inline JWKS. Example request:

```
  {
    "client_name": "test-nojwks",
    "grant_types": ["authorization_code", "client_credentials", "refresh_token"],
    "token_endpoint_auth_method": "tls_client_auth",
    "jwks": {
      "keys": [
        {
          "kty": "RSA",
          "kid": "kid",
          "alg": "PS256",
          "use": "sig",
          "n": "<omitted>",
          "e": "AQAB",
          "x5c": ["<certificate omitted>"]
        }
      ]
    },
    "redirect_uris": ["https://example.com/v1/callback"]
  }
```

4. Keycloak successfully creates the client.
5. Export the client or fetch via Admin REST API → JWKS is present.
6. Open the client in the Admin Console → Clients → sm-test-nojwks → Keys.

### Anything else?

The UI appears to only show keys created through the UI itself.

rmartinc:
Closes #44284

Just a little PR to show the certificate information for OIDC clients using the three possible formats (jwks, certificate or public ket). Previously the certificate was fixed and it could be empty. The `Certificate.tsx` is modified to allow passing the `type` to show in the display. There is a new element `KeyInfoArea` that, if no `type` is passed, selects the first option that is defined in the order jws, certificate, public key.

### DIFF
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -1941,6 +1941,7 @@ fullSyncPeriod=Full sync period
 clientsExplain=Clients are applications and services that can request authentication of a user.
 addNode=Add node
 jwksUrl=JWKS URL
+jwks=JSON Web Key Set
 policy-description=A description for this policy.
 defaultPasswordLabel=My password
 mapperUserAttributeName=User Attribute Name
--- a/js/apps/admin-ui/src/clients/keys/Certificate.tsx
+++ b/js/apps/admin-ui/src/clients/keys/Certificate.tsx
@@ -11,25 +11,31 @@ type CertificateProps = Omit<CertificateDisplayProps, "id"> & {
 type CertificateDisplayProps = {
   id: string;
   helpTextKey?: string;
+  type?: "jwks" | "certificate" | "publicKey";
   keyInfo?: CertificateRepresentation;
 };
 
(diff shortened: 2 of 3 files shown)


mposolda (approved):
@rmartinc Thanks!

PR #45936, opened 2026-02-02
fix: allow the same vm to run a keycloak instance multiple times

shawkins:
closes: #45921

Reuses the test logic we already have for resetting the system state between runs.

This assumes that the Keycloak instance takes responsibility for restoring the state of the system properties. That can be tweaked if needed - in particular we can try to track specifically what we / quarkus can manipulate and just clear that.

Finally the alternative to this whole approach would be to fork the test execution instead.


vmuzikar (approved):
LGTM, thanks.

PR #45945, opened 2026-02-02
OTEL: Use suggested 'code.function.name' for span attributes

### LINKED ISSUE #45944: OTEL: Use suggested 'code.function.name' for span attributes
### Description

At this moment, we use the `code.function` and `code.namespace` attributes for HTTP requests when tracing is enabled. However, those are deprecated, and rather the fully qualified attribute `code.function.name` should be used. 

It means, that the namespace and function should not be separated. 

Old:
code.namespace =  org.my.custom.Something
code.function = executeThis

New: 
code.function.name =  org.my.custom.Something.executeThis

However, we can also set these old ones as somebody in the community may rely on them (unlikely), but we can remove it completely later. 

### Value Proposition

Do not use deprecated span attributes

### Goals

Provide suggested attribute

### Non-Goals

N/A

### Discussion

_No response_

### Notes

_No response_

mabartos:
- Closes #45944

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -22,7 +22,8 @@ It also lists significant changes to internal APIs.
 
 === Dev Mode defaults to localhost
 
-When running the server in dev mode on a platform other than Windows Subsystem For Linux, the `http-host` setting will default to localhost. This ensures your dev instance won't be accessible from 
+When running the server in dev mode on a platform other than Windows Subsystem For Linux, the `http-host` setting will default to localhost.
+This ensures your dev instance won't be accessible from
 other machines. If you want the previous behavior of binding to all interfaces, then explicitly set `http-host` to `0.0.0.0`.
 
 === `X-Forwarded-Prefix` Header is now supported
@@ -114,6 +115,11 @@ The following sections provide details on deprecated features.
 
 The `tracing.serviceName`, and `tracing.resourceAttributes` fields of the Keycloak CR, are now deprecated.
 You should use the new `telemetry.serviceName`, and `telemetry.resourceAttributes` fields that are shared among all OpenTelemetry components - logs, metrics, and traces.
+
+=== Deprecation of tracing span attributes for HTTP requests
+
+The OpenTelemetry tracing span attributes `code.function` and `code.namespace` are deprecated for the HTTP request spans when tracing is enabled.
+These attributes will be removed after the 26.6 release, and only the fully qualified `code.function.name` span attribute will stay.
 
 // ------------------------ Removed features ------------------------ //
 == Removed features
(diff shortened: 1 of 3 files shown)


ahus1 (changes requested):
Thank you for this PR. 
As you are about to deprecate the old tags: 

* Add it to the deprecation note to the upgrading guide
* Add a marker in the code that is is deprecated for removal in 26.6

mabartos:
@ahus1 Sure, thanks.

mabartos:
@ahus1 Updated. 

> Add a marker in the code that is is deprecated for removal in 26.6

Not sure what exact place you mean. I don't want to put the deprecated annotation on the whole class, and I don't see any good place where we should put the annotation. These attributes come from the OTel. I think the deprecation note is sufficient in this case, right?

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -115,6 +116,11 @@ The following sections provide details on deprecated features.
 The `tracing.serviceName`, and `tracing.resourceAttributes` fields of the Keycloak CR, are now deprecated.
 You should use the new `telemetry.serviceName`, and `telemetry.resourceAttributes` fields that are shared among all OpenTelemetry components - logs, metrics, and traces.
 
+=== Deprecation of tracing span attributes for HTTP requests
+
+The OpenTelemetry tracing span attributes `code.function` and `code.namespace` are deprecated for the HTTP request spans when tracing is enabled.
+These attributes will be removed after the 26.6 release, and only the fully qualified `code.function.name` span attribute will stay.
```
ahus1: We only remove in major releases.

```suggestion
These attributes will be removed in the next major release, and only the fully qualified `code.function.name` span attribute will stay.
```

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/integration/resteasy/KeycloakTracingCustomizer.java
```diff
@@ -60,9 +63,11 @@ public void handle(ResteasyReactiveRequestContext requestContext) {
             Tracer myTracer = openTelemetry.getTracer(this.getClass().getName(), Version.VERSION);
             SpanBuilder spanBuilder = myTracer.spanBuilder(spanName);
             spanBuilder.setParent(Context.current().with(Span.current()));
-            // for semconv >= 1.32 use CODE_FUNCTION_NAME instead
+            spanBuilder.setAttribute(CodeAttributes.CODE_FUNCTION_NAME, functionName);
+            // for backwards compatibility
             spanBuilder.setAttribute(CodeIncubatingAttributes.CODE_FUNCTION, methodName);
             spanBuilder.setAttribute(CodeIncubatingAttributes.CODE_NAMESPACE, className);
```
ahus1:
```suggestion
            // for backwards compatibility. deprecated in 26.6, to be removed in 27.0
            spanBuilder.setAttribute(CodeIncubatingAttributes.CODE_FUNCTION, methodName);
            spanBuilder.setAttribute(CodeIncubatingAttributes.CODE_NAMESPACE, className);
            // end deprecation
```

ahus1 (changes requested):
See below for a suggested change about release notes and deprecation.

ahus1 (approved):

PR #45946, opened 2026-02-02
Mark Token Exchange v1 as deprecated but in preview

### LINKED ISSUE #45791: Make sure there is support for "preview deprecated" features. Update token-exchange:v1 to "preview deprecated"
### Description

Currently it is not possible to have the feature, which is `Preview` and `Deprecated` at the same time. Updating token-exchange:v1 to be just deprecated does not work as "deprecated" features are considered supported.

Some possibilities to achieve this:

1) Do some more clever updates in `Profile.Feature` class to have some better support for deprecated features of various support states. Maybe we can add `Boolean deprecated` as a variable to the `Profile.Feature` ? This would `true` by default for `DEPRECATED` features and `false` by default for others, but there should be a way to override, so that token-exchange:v1 can be still `PREVIEW`, but would be deprecated (with boolean flag `deprecated=true`) at the same time.
 
2) Introduce `PREVIEW_DEPRECATED` state (this is maybe easier, but it lacks some flexibility)

My vote is likely option (1).

It is needed to make sure that there is correct message displayed in the server log at server startup. Probably admin console may need some updates as well. Right now, it mentions Warning with enabled "Preview" features and enabled "Deprecated" features. It seems token-exchange:v1 might need to be mentioned in both of those groups (So maybe log for "deprecated" should use that new `boolean deprecated` flag instead of `DEPRECATED` features).

Making `token-exchange:v1` to be deprecated is probably good as part of this issue to be able to test that "preview deprecated" features work as expected.

rmartinc:
Closes #45791

I followed the first idea in the issue. In the admin console the label remains like `Preview` but I changed the color to grey. If you think that maybe we should change also  the label (adding deprecated) or we should add an icon or similar, just let me know.

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -76,7 +76,7 @@ public class Profile {
 
         SCRIPTS("Write custom authenticators using JavaScript", Type.PREVIEW),
 
-        TOKEN_EXCHANGE("Token Exchange Service", Type.PREVIEW, 1),
+        TOKEN_EXCHANGE("Token Exchange Service", Type.PREVIEW, 1, true, null, null),
         TOKEN_EXCHANGE_STANDARD_V2("Standard Token Exchange version 2", Type.DEFAULT, 2),
         TOKEN_EXCHANGE_EXTERNAL_INTERNAL_V2("External to Internal Token Exchange version 2", Type.EXPERIMENTAL, 2),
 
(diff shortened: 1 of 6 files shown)


rmartinc:
For example, if I start the server with preview features, token-exchange appears in both sections:

```
2026-02-03 14:31:56,990 INFO  [org.keycloak.common.Profile] (main) Preview features enabled: client-auth-federated:v1, client-secret-rotation:v1, http-optimized-serializers:v1, jwt-authorization-grant:v1, kubernetes-service-accounts:v1, opentelemetry-logs:v1, scripts:v1, spiffe:v1, token-exchange:v1, workflows:v1
2026-02-03 14:31:56,990 WARN  [org.keycloak.common.Profile] (main) Deprecated features enabled: client-auth-federated:v1, client-secret-rotation:v1, http-optimized-serializers:v1, jwt-authorization-grant:v1, kubernetes-service-accounts:v1, opentelemetry-logs:v1, scripts:v1, spiffe:v1, token-exchange:v1, workflows:v1
```

And the console appears like the following:

Screenshot From 2026-02-03 14-33-04

For the moment, it's just grey instead of blue, but continues saying `Preview`. We can add something like `Preview/Deprecated` or add an icon to differentiate.

rmartinc:
@mposolda @stianst Finally I decided to add two bubbles instead of one. So now token exchange is `Preview` and `Deprecated`. It is clear I think:

Screenshot From 2026-02-04 11-28-07

mposolda:
@rmartinc Thanks for the update! IMO the admin-console looks nice.

However the startup message doesn't look good. It mentions for example `client-auth-federated:v1` or `jwt-authorization-grant:v1` as "deprecated" features. This is not correct though. It seems that reported "Deprecated" features are exactly same like the "Preview" features mentioned on the line above?

Also I am not 100% sure, but looks like failing Admin UI tests are caused by this PR (Hopefully it is possible to solve it by updating the test without need to update UI further as UI in your PR looks nice IMO).

rmartinc:
Upps! Yes, sorry. Good catch! The log was getting the type again supposing that it was always passing the same type. I just changed to check if the feature is in the passed list. Now it looks like this when started with `--features preview,oid4vc-vci`:

```
2026-02-04 17:31:09,301 INFO  [org.keycloak.common.Profile] (main) Preview features enabled: client-auth-federated:v1, client-secret-rotation:v1, http-optimized-serializers:v1, jwt-authorization-grant:v1, kubernetes-service-accounts:v1, opentelemetry-logs:v1, scripts:v1, spiffe:v1, token-exchange:v1, workflows:v1
2026-02-04 17:31:09,301 WARN  [org.keycloak.common.Profile] (main) Experimental features enabled: oid4vc-vci:v1
2026-02-04 17:31:09,301 WARN  [org.keycloak.common.Profile] (main) Deprecated features enabled: token-exchange:v1
```

Let's see if I have not broken any test.

mposolda (approved):
@rmartinc Thanks!

PR #45949, opened 2026-02-03
[OID4VCI] Wrong typ value for SD-JWT VC

### LINKED ISSUE #45420: [OID4VCI] Wrong typ value for SD-JWT VC
### Before reporting an issue

### Area

oid4vc

### Describe the bug

By default the UI for ClientScopes sets the value for the `typ`-value of the SD-JWT to `JWS`. This is wrong. Based on the SD-JWT VC spec it is like this:

```
The typ header parameter of the SD-JWT MUST be present.  The typ
   value MUST use dc+sd-jwt.  This indicates that the payload of the SD-
   JWT contains plain JSON and follows the rules defined in this
   specification.  It further indicates that the SD-JWT is an SD-JWT
   component of an SD-JWT VC.
```

so in case that we select the format `dc+sd-jwt` we also need to add this value into the `typ` header field of the SD-JWT.

So basically, this field must get the default-value `dc+sd-jwt` if the appropriate format is selected. But I think we should still leave it configurable!

Image

### Version

26.5.0

### Regression

### Expected behavior

the default value is  `dc+sd-jwt` if this format is selected

### Actual behavior

the default value is  `JWS` if this format is selected

### How to Reproduce?

see above

### Anything else?

_No response_

tdiesler:
closes #45420


VinodAnandan:
@IngridPuppet @mposolda @Captain-P-Goldfish  I’d really appreciate your help reviewing this PR when you get a chance.

mposolda (approved):
@tdiesler LGTM, Thanks!

PR #45953, opened 2026-02-03
[OID4VCI] Make natural_person configuration available in all formats

tdiesler:
closes #44833

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import jakarta.ws.rs.core.Response;
 
 import org.keycloak.Config;
+import org.keycloak.VCFormat;
 import org.keycloak.constants.OID4VCIConstants;
 import org.keycloak.events.EventBuilder;
 import org.keycloak.models.ClientModel;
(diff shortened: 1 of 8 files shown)


services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -225,9 +241,10 @@ public void validateClientScopeAssignment(KeycloakSession session, ClientScopeMo
...
-    private void addClientScopeDefaults(ClientScopeModel clientScope) {
-        ClientScopeRepresentation clientScopeRep = ModelToRepresentation.toRepresentation(clientScope);
-        addClientScopeDefaults(clientScopeRep);
-        RepresentationToModel.updateClientScope(clientScopeRep, clientScope);
+    private String getFormatFromScope(String scope) {
+        String format = SD_JWT_VC; // default format
+        if (scope.toLowerCase().endsWith("_jwt")) format = Format.JWT_VC;
+        if (scope.toLowerCase().endsWith("_ld")) format = Format.LDP_VC;
+        return format;
     }
```
IngridPuppet: Could you relocate this logic to class `Format` so it lives next to `getScopeSuffix`? It might be hard to maintain the reasoning on the suffixes.

IngridPuppet (commented):
Hello @tdiesler - I reviewed the changes again and left one comment. Please could you check?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: done

IngridPuppet (approved):
Thank you for the update.

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
```diff
@@ -126,18 +127,25 @@ public Object createProtocolEndpoint(KeycloakSession keycloakSession, EventBuild
...
+                clientScope.addProtocolMapper(builtins.get(SUBJECT_ID_MAPPER));
+                clientScope.addProtocolMapper(builtins.get(EMAIL_MAPPER));
+                clientScope.addProtocolMapper(builtins.get(FIRST_NAME_MAPPER));
+                clientScope.addProtocolMapper(builtins.get(LAST_NAME_MAPPER));
+
+                ClientScopeRepresentation clientScopeRep = ModelToRepresentation.toRepresentation(clientScope);
+                addClientScopeDefaults(clientScopeRep);
+                RepresentationToModel.updateClientScope(clientScopeRep, clientScope);
+
+                newRealm.addDefaultClientScope(clientScope, false);
```
mposolda: Could you please remove this line?

The `oid4vc-natural_person_*` are only example scopes not defined in any specification. As such, it can be good if they are not added as "Realm optional" client scopes. The "Realm optional" client scopes means that they might be added to every new client when the client is created, which is not the desired behavior. I think they should be added to clients on demand just if the administrator of the particular client wants to use those as client scopes.

Feel free to add them in the testsuite as default client scopes, but should not be added here.

mposolda (commented):
@tdiesler Added review comment inline. Can you please doublecheck?

services/src/main/java/org/keycloak/protocol/oid4vc/OID4VCLoginProtocolFactory.java
tdiesler: done

mposolda (approved):
@tdiesler Thanks! I hope this PR can be merged once tests are OK. Current test failures are not caused by your PR, so will try to re-run and hopefully that helps.

@IngridPuppet Thanks for the additional review!

mposolda (approved):

PR #45956, opened 2026-02-03
Organization Groups Caching

### LINKED ISSUE #45509: Organization Groups Caching
### Description

Review and implement the caching strategy for Organization Groups.

**Scope & Notes**
- Investigate if the current delegation to InfinispanUserProvider (which hits the Realm cache via ID lookup) is sufficient.
- Determine if a specific Organization cache is required or if we can rely on the existing group cache mechanisms without performance degradation.

vramik:
Closes #45509

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
@@ -305,31 +305,34 @@ public class InfinispanOrganizationProvider implements OrganizationProvider {
 
     @Override
     public GroupModel createGroup(OrganizationModel organization, String id, String name, GroupModel toParent) {
-        //todo caching
         return getDelegate().createGroup(organization, id, name, toParent);
     }
 
     @Override
     public Stream<GroupModel> getTopLevelGroups(OrganizationModel organization, Integer firstResult, Integer maxResults) {
-        //todo caching
+        // Don't cache top-level groups - delegate directly to DB
+        // This follows the same pattern as search queries to avoid unbounded cache growth
         return getDelegate().getTopLevelGroups(organization, firstResult, maxResults);
     }
 
     @Override
     public Stream<GroupModel> searchGroupsByName(OrganizationModel organization, String search, Boolean exact, Integer firstResult, Integer maxResults) {
-        //todo caching
+        // Don't cache search queries with pagination - delegate directly to DB
+        // This follows the same pattern as RealmCacheSession.searchForGroupByNameStream
         return getDelegate().searchGroupsByName(organization, search, exact, firstResult, maxResults);
     }
 
     @Override
     public Stream<GroupModel> searchGroupsByAttributes(OrganizationModel organization, Map<String, String> attributes, Integer firstResult, Integer maxResults) {
-        //todo caching
+        // Don't cache search queries with pagination - delegate directly to DB
+        // This follows the same pattern as RealmCacheSession.searchGroupsByAttributes
         return getDelegate().searchGroupsByAttributes(organization, attributes, firstResult, maxResults);
     }
 
     @Override
     public Stream<GroupModel> getOrganizationGroupsByMember(OrganizationModel organization, UserModel member) {
-        //todo caching
+        // Don't cache per-member group lists - delegate directly to DB
+        // Currently only used in export
         return getDelegate().getOrganizationGroupsByMember(organization, member);
     }
 
--- a/model/jpa/src/main/java/org/keycloak/models/jpa/GroupAdapter.java
+++ b/model/jpa/src/main/java/org/keycloak/models/jpa/GroupAdapter.java
@@ -384,7 +384,14 @@ public class GroupAdapter implements GroupModel , JpaModel<GroupEntity> {
         OrganizationProvider orgProvider = session.getProvider(OrganizationProvider.class);
         if (orgProvider == null) return null;
 
-        return orgProvider.getById(organization.getId());
+        // Temporarily set session realm context to group's realm to avoid realm mismatch
+        RealmModel currentRealm = session.getContext().getRealm();
+        try {
+            session.getContext().setRealm(realm);
+            return orgProvider.getById(organization.getId());
+        } finally {
+            session.getContext().setRealm(currentRealm);
+        }
     }
 
     @Override
--- a/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
+++ b/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
@@ -130,6 +130,11 @@ public class JpaOrganizationProvider implements OrganizationProvider {
             adapter.setEnabled(true);
 
             em.persist(adapter.getEntity());
+
+            // Set organization-group relationship for the internal group
+            // Must be done after persist so the organization entity is managed
+            GroupEntity groupEntity = em.find(GroupEntity.class, group.getId());
+            groupEntity.setOrganization(entity);
         } finally {
             session.getContext().setOrganization(null);
         }
@@ -490,10 +495,8 @@ public class JpaOrganizationProvider implements OrganizationProvider {
             parentGroup = getOrganizationGroup(organization);
         } else {
             // Validate the parent group
-            if (!Objects.equals(toParent.getType(), Type.ORGANIZATION)) {
-                throw new ModelValidationException("Parent group must be of type ORGANIZATION");
-            }
-            if (!Objects.equals(toParent.getOrganization().getId(), organization.getId())) {
+            if (!Organizations.isOrganizationGroup(toParent) ||
+                    !Objects.equals(toParent.getOrganization().getId(), organization.getId())) {
                 throw new ModelValidationException("Parent group does not belong to the specified organization");
             }
             parentGroup = toParent;
@@ -532,7 +535,6 @@ public class JpaOrganizationProvider implements OrganizationProvider {
 
         return closing(paginateQuery(em.createQuery(queryBuilder), firstResult, maxResults).getResultStream()
                 .map(realm::getGroupById)
-                .sorted(GroupModel.COMPARE_BY_NAME)
         );
     }
 
@@ -553,6 +555,7 @@ public class JpaOrganizationProvider implements OrganizationProvider {
         predicates.add(builder.equal(root.get("realm"), realm.getId()));
         predicates.add(builder.equal(root.get("type"), Type.ORGANIZATION.intValue()));
         predicates.add(builder.equal(root.get("organization").get("id"), organization.getId()));
+        predicates.add(builder.notEqual(root.get("id"), getOrganizationGroup(organization).getId())); // Exclude internal group
 
         if (Boolean.TRUE.equals(exact)) {
             predicates.add(builder.equal(root.get("name"), search));
@@ -568,7 +571,6 @@ public class JpaOrganizationProvider implements OrganizationProvider {
 
         return closing(paginateQuery(em.createQuery(queryBuilder), firstResult, maxResults).getResultStream()
                         .map(realm::getGroupById)
-                        .sorted(GroupModel.COMPARE_BY_NAME)
         );
     }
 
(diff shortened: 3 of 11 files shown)


model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
```diff
@@ -305,32 +305,88 @@ public boolean isManagedMember(OrganizationModel organization, UserModel user) {
...
+        if (isRealmCacheKeyInvalid(cacheKey)) {
+            return getDelegate().getTopLevelGroups(organization, firstResult, maxResults);
+        }
+
+        CachedOrgGroupIds cached = realmCache.getCache().get(cacheKey, CachedOrgGroupIds.class);
+
+        if (cached == null) {
+            Long loaded = realmCache.getCache().getCurrentRevision(cacheKey);
+            // Load ALL top-level groups (no pagination) to cache the complete list
+            Stream<GroupModel> groups = getDelegate().getTopLevelGroups(organization, null, null);
```
pruivo: I'm concerned about this. Is there a limit on the number of groups? 
The memory consumption of this method may be a problem because not only does the Hibernate Persistence Context keep the `GroupEntity` in memory, but it is also copied to the Infinispan cache as `CachedGroup`. 

As the `realms` cache is bounded, if the number of groups is high enough, it may clear the cache of all other data.

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
```diff
@@ -305,32 +305,88 @@ public boolean isManagedMember(OrganizationModel organization, UserModel user) {
...
     @Override
     public Stream<GroupModel> searchGroupsByAttributes(OrganizationModel organization, Map<String, String> attributes, Integer firstResult, Integer maxResults) {
-        //todo caching
+        // Don't cache search queries with pagination - delegate directly to DB
+        // This follows the same pattern as RealmCacheSession.searchGroupsByAttributes
         return getDelegate().searchGroupsByAttributes(organization, attributes, firstResult, maxResults);
     }
 
     @Override
     public Stream<GroupModel> getOrganizationGroupsByMember(OrganizationModel organization, UserModel member) {
```
pruivo, line 333: This method is only used by the `ExportUtils`. Is it worth adding caching? 

https://github.com/keycloak/keycloak/blob/0433b0017dc1ee2cd258e9f16ba4ca985a162e1d/model/storage-private/src/main/java/org/keycloak/exportimport/util/ExportUtils.java#L269

pruivo (changes requested):

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
vramik: Good point. Thank you. Removed the caching for now, it might be added later in case it's needed.

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
vramik: Thank you, removed the caching and let to delegate the call to DB.

model/jpa/src/main/java/org/keycloak/models/jpa/GroupAdapter.java
```diff
@@ -384,7 +384,14 @@ public OrganizationModel getOrganization() {
         OrganizationProvider orgProvider = session.getProvider(OrganizationProvider.class);
         if (orgProvider == null) return null;
 
-        return orgProvider.getById(organization.getId());
+        // Temporarily set session realm context to group's realm to avoid realm mismatch
```
pruivo, line 387: Is this safe? Can you elaborate on why it is required to change the realm?

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
```diff
@@ -553,6 +556,7 @@ public Stream<GroupModel> searchGroupsByName(OrganizationModel organization, Str
         predicates.add(builder.equal(root.get("realm"), realm.getId()));
         predicates.add(builder.equal(root.get("type"), Type.ORGANIZATION.intValue()));
         predicates.add(builder.equal(root.get("organization").get("id"), organization.getId()));
+        predicates.add(builder.notEqual(root.get("id"), getOrganizationGroup(organization).getId())); // Exclude internal group
```
pruivo, line 558: Not related to the PR, but I don't think the logic in this method is correct. The method parameter `String search` is not used in the query when `exact` is set to false. Is it intentional?

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
pruivo: Offtopic, just want to bring awareness to these two cases:
* It looks like `getTopLevelGroups`/`searchGroupsByName`/`searchGroupsByAttributes` all cache the query result when they invoke `.map(realm::getGroupById)` in the stream.
* The `.sorted(GroupModel.COMPARE_BY_NAME)` seems redundant as the query is set with `queryBuilder.orderBy(builder.asc(root.get("name")));`

pruivo (changes requested):

model/jpa/src/main/java/org/keycloak/models/jpa/GroupAdapter.java
vramik: When a request lands in here https://github.com/keycloak/keycloak/blob/78299ae82d4eb2cee62217ea43c74ecfa8e533ce/services/src/main/java/org/keycloak/services/resources/admin/RealmsAdminResource.java#L211 the realm is queried by name, say `test`. In the context there is `master` realm. 

In scenario where an organization exists in the `test` and `orgProvider.getById(organization.getId());` is called on line 391, it ends up here https://github.com/keycloak/keycloak/blob/78299ae82d4eb2cee62217ea43c74ecfa8e533ce/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java#L800 because the realm is taken from the session context: https://github.com/keycloak/keycloak/blob/78299ae82d4eb2cee62217ea43c74ecfa8e533ce/model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java#L852

There is a stacktrace: 
```
at org.keycloak.organization.jpa.JpaOrganizationProvider.getEntity(JpaOrganizationProvider.java:805)
at org.keycloak.organization.jpa.JpaOrganizationProvider.getById(JpaOrganizationProvider.java:227)
at org.keycloak.models.cache.infinispan.organization.InfinispanOrganizationProvider.getById(InfinispanOrganizationProvider.java:103)
at org.keycloak.models.jpa.GroupAdapter.getOrganization(GroupAdapter.java:391)
at org.keycloak.models.cache.infinispan.entities.CachedGroup.<init>(CachedGroup.java:66)
at org.keycloak.models.cache.infinispan.RealmCacheSession.getGroupById(RealmCacheSession.java:976)
at org.keycloak.models.jpa.RealmAdapter.getGroupById(RealmAdapter.java:1994)
at org.keycloak.models.jpa.RealmAdapter$$Lambda/0x00007f1989efc248.apply(Unknown Source:-1)
at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
at java.util.Iterator.forEachRemaining(Iterator.java:133)
at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)
at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
at org.keycloak.models.cache.infinispan.entities.CachedRealm.<init>(CachedRealm.java:303)
at org.keycloak.models.cache.infinispan.RealmCacheSession.prepareCachedRealm(RealmCacheSession.java:494)
at org.keycloak.models.cache.infinispan.RealmCacheSession$$Lambda/0x00007f1989eb7cf8.apply(Unknown Source:-1)
at org.keycloak.models.cache.infinispan.RealmCacheManager.computeSerialized(RealmCacheManager.java:150)
at org.keycloak.models.cache.infinispan.RealmCacheSession.getRealm(RealmCacheSession.java:476)
at org.keycloak.models.jpa.JpaRealmProvider.getRealmByName(JpaRealmProvider.java:185)
at org.keycloak.models.cache.infinispan.RealmCacheSession.prepareCachedRealmByName(RealmCacheSession.java:536)
at org.keycloak.models.cache.infinispan.RealmCacheSession.lambda$getRealmByName$0(RealmCacheSession.java:528)
at org.keycloak.models.cache.infinispan.RealmCacheSession$$Lambda/0x00007f1989dc5a40.apply(Unknown Source:-1)
at org.keycloak.models.cache.infinispan.RealmCacheManager.computeSerialized(RealmCacheManager.java:150)
at org.keycloak.models.cache.infinispan.RealmCacheSession.getRealmByName(RealmCacheSession.java:528)
at org.keycloak.services.managers.RealmManager.getRealmByName(RealmManager.java:116)
at org.keycloak.services.resources.admin.RealmsAdminResource.getRealmAdmin(RealmsAdminResource.java:211)
at org.keycloak.services.resources.admin.RealmsAdminResource$quarkusrestinvoker$getRealmAdmin_ae16b3615f3863e1601864ee53c6d1785644e290.invoke(Unknown Source:-1)
at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:190)
at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:666)
at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)
at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)
at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)
at java.lang.Thread.runWith(Thread.java:1596)
at java.lang.Thread.run(Thread.java:1583)
```

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
vramik: No, good catch, I'll send a fix within: https://github.com/keycloak/keycloak/issues/46006

model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/organization/InfinispanOrganizationProvider.java
vramik: Thank you. 

ad 1) yes, this is intentional, it follows the same pattern as `JpaRealmProvider.getGroupsStream` methods.

ad 2) you're right, it was copy-pasted from https://github.com/keycloak/keycloak/blob/78299ae82d4eb2cee62217ea43c74ecfa8e533ce/model/jpa/src/main/java/org/keycloak/models/jpa/JpaRealmProvider.java#L769-L777

I'll send an update to fix both JpaOrganizationProvider and JpaRealmProvider.

model/jpa/src/main/java/org/keycloak/models/jpa/GroupAdapter.java
pruivo: I see, thank you for the explanation @vramik.

pruivo (dismissed):
LGTM, thanks @vramik

pedroigor (approved):

pruivo (approved):

PR #45960, opened 2026-02-03
[OID4VCI] Simplify OID4VCAuthorizationDetail handling

tdiesler:
closes #45935


mposolda:
@tdiesler I am not sure about the advantage of this TBH...

The RAR specification does not differ between `authorization_details` for request and response. But OID4VCI clearly do so, it mentions the format of the request here: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-5.1.1 (without `credential_identifiers`) and format of the response here https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-6.2 (with added `credential_identifiers`). Having 2 types allow to make it clear if it is type for "request" or for "response" .

tdiesler:
The spec has lots of classes that have their properties populated according to use case or version of the spec. If we had separate classes for each usage there would be a significant bloat. How about we add a comment saying that `credential_identifiers` is only populated by the AccessToken Response?

We already reject non-null usage for `credential_identifiers` in requests.

tdiesler:
@forkimenjeckayang could you perhaps have a look at this? I'd appreciate to hear your take ...

```
public class OID4VCAuthorizationDetail extends AuthorizationDetailsJSONRepresentation {

    /**
     * The 'credential_identifiers' property is populated by the Issuer in the AccessToken Response
     * <p/>
     * Identifying Credentials Being Issued Throughout the Issuance Flow
     * https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-3.3.4
     * <p/>
     * The property should not be used in Authorization or AccessToken requests.
     */
    @JsonProperty(CREDENTIAL_IDENTIFIERS)
    private List<String> credentialIdentifiers;
```

The idea is to use the same type in the various oid4vci requests and responses (i.e. AuthorizationRequest, AccessToken request/response, CredentialOfferState)

Over time, `authorization_details` may evolve in the context of oid4vci and like other classes not all properties are relevant/populated for every given message flow.

To set this property in the context of our new OAuthClient API will fail for AuthorizationRequest and AccessTokenRequest. Server endpoints ignore it like they do with other unknown message properties.

The spec explicitly [allows](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-using-authorization-details) it ...

> Additional authorization_details data fields MAY be defined and used when the type value is openid_credential. Note that this effectively defines an authorization details type that is never considered invalid due to unknown fields.

mposolda:
> The spec has lots of classes that have their properties populated according to use case or version of the spec. If we had separate classes for each usage there would be a significant bloat. How about we add a comment saying that `credential_identifiers` is only populated by the AccessToken Response?
> 

Since there is final version of OID4VCI specification, I suppose this is the only specs we need to support. Or do you see any need to support some previous drafts?

BTW. The `AuthorizationDetailsJSONRepresentation` parent class has `customData` on it. AFAIK we can use that if there is a need for some "custom" claims for some other "extensions" of OID4VCI? Right now, the `customData` are not populated in `OID4VCAuthorizationDetailsParser.asSubtype` from the `AuthorizationDetailsJSONRepresentation` ,but perhaps we can support populating them here. Would it address a need for custom claims?

> We already reject non-null usage for `credential_identifiers` in requests.

Do we? I do not see in the codebase or in your PR that we reject the AccessTokenRequest or AuthorizationRequest with `authorization_details`, which contain `credential_identifiers` . Especially not seeing that in `OID4VCAuthorizationDetailsProcessor.validateAuthorizationDetail` where I would expect such a check.

tdiesler:
> Do we?

We already reject non-null usage for credential_identifiers in requests [here](https://github.com/tdiesler/keycloak/blob/oauthclient-all/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationRequestRequest.java#L22) and [here](https://github.com/tdiesler/keycloak/blob/oauthclient-all/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java#L54)

VinodAnandan:
@tdiesler @mposolda  We supported partially Draft 15 primarily to maintain compatibility with the broader wallet ecosystem (  https://www.keycloak.org/2026/01/issue-credentials-over-openid4vci#_%EF%B8%8F_wallet_compatibility_notice ), as many wallet implementations had not yet been updated to the final specification. That said, it may not be ideal to continue supporting additional draft versions, or to retain Draft 15 support for an extended period. Once the relevant wallets are updated, it would be reasonable to consider deprecating Draft 15 support.

@IngridPuppet:  please feel free to correct me if I’ve misunderstood the strategy or if any of the details above are inaccurate.

tdiesler:
I'm trying to say that this `authorization_details` type hierarchy is less than ideal and probably not wanted (irrespective of spec version concerns)

```
OID4VCAuthorizationDetailResponse => OID4VCAuthorizationDetail => AuthorizationDetailsJSONRepresentation
```

Apart from that we should generally prefer "composition over inheritance" simplifying this to 

```
OID4VCAuthorizationDetail => AuthorizationDetailsJSONRepresentation
```

(or even to just a single type) can benefit code readability a great deal. Instead of keeping track of these various types, it would be sufficient to reason about `authorization_details` (and its associated java type) no matter what the current message flow will be.

Yes, there will be properties that apply to one flow and not another. However, it is already so that all of these properties from `AuthorizationDetailsJSONRepresentation`

```
    @JsonProperty("type")
    private String type;
    @JsonProperty("locations")
    private List<String> locations;
    @JsonProperty("actions")
    private List<String> actions;
    @JsonProperty("datatypes")
    private List<String> datatypes;
    @JsonProperty("identifier")
    private String identifier;
    @JsonProperty("privileges")
    private List<String> privileges;
```

are seen by `OID4VCAuthorizationDetail` and do not apply to the OID4VCI message flows.

IMHO, we do not need yet another type `OID4VCAuthorizationDetailResponse` that adds little to what we already have.

My preferred option would be to have one `OID4VCAuthorizationDetail` that works well in all of our use cases. It is not intuitive that we pass one `authorization_details` type to Authorization/Token Request  and get back another from AccessToken Response for the same `authorization_details` json property.

Anyway, this detail should not block progress in other areas - I've taken it out as a dependency from #46014 and all other PRs as well.

tdiesler:
@forkimenjeckayang could you perhaps have a look at this PR? I appreciate your opinion as the original author.

You can always meet me on slack for a quick chat.

mposolda:
> > Do we?
> 
> We already reject non-null usage for credential_identifiers in requests [here](https://github.com/tdiesler/keycloak/blob/oauthclient-all/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationRequestRequest.java#L22) and [here](https://github.com/tdiesler/keycloak/blob/oauthclient-all/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java#L54)

@tdiesler I meant if we reject such a request with `credential_identifiers` in the `authorization_details` somewhere on the server-side. You pointed the testsuite classes, which are not relevant (moreover I do not see anything about rejecting `credential_identifiers`  in the `authorization_details` in the links you pointed).

mposolda:
@tdiesler Thanks for your points. I still consider request/response separate types better as it is more clear what parameters expected in request and what in response. However not strong on that if other prefers otherwise and have single type `OID4VCDetail` for both requests and response.

@forkimenjeckayang @IngridPuppet What is your preference?

If we go with single type, it can be nice to add validation on server-side (probably in `OID4VCAuthorizationDetailsProcessor.validateAuthorizationDetail` ) to not allow `authorization_details` request with `credential_identifiers` claim filled, so it is clear that people should not fill this field when sending `authorization_details` in the requests.

IngridPuppet:
> @forkimenjeckayang @IngridPuppet What is your preference?

I recall reviewing these changes before the PR was sent upstream, and I actually pushed for a separate request/response data model because it directly eliminates and eliminated logic mistakes, e.g., attempts to reason about credential identifiers instead of credential configuration IDs in the request.

However, I also understand the point to unify the models to reach a consolidated view of authorization details. Because the logic is largely already written from request to response, maybe it is less problematic to go this way at this point. 

> If we go with single type, it can be nice to add validation on server-side (probably in `OID4VCAuthorizationDetailsProcessor.validateAuthorizationDetail` ) to not allow `authorization_details` request with `credential_identifiers` claim filled, so it is clear that people should not fill this field when sending `authorization_details` in the requests.

+1. A test validating that credential identifiers in a request object are not processed along the flow addresses my initial concern, IMO.

core/src/main/java/org/keycloak/OAuth2Constants.java
```diff
@@ -177,4 +177,5 @@ public interface OAuth2Constants {
     // OID4VCI - https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html
     String OPENID_CREDENTIAL = "openid_credential";
     String CREDENTIAL_IDENTIFIERS = "credential_identifiers";
+    String CREDENTIAL_CONFIGURATION_ID = "credential_configuration_id";
```
IngridPuppet: Because these constants are more tied to OpenID4VCI, I think it is preferable to have them defined in the `OID4VCIConstants` class. WDYT?

tdiesler:
> If we go with single type, it can be nice to add validation on server-side

ok, will do

services/src/test/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessorTest.java
```diff
@@ -130,24 +130,22 @@ private ClaimsDescription createInvalidClaimsDescription() {
...
         assertNotNull("Locations should not be null", authDetail.getLocations());
         assertEquals("Should have exactly one location", 1, authDetail.getLocations().size());
         assertEquals("Location should match issuer", "https://test-issuer.com", authDetail.getLocations().get(0));
-        assertEquals(1, authDetail.getCredentialIdentifiers().size());
-        assertEquals("test-identifier-123", authDetail.getCredentialIdentifiers().get(0));
+        if (tokenResponse) {
+            assertEquals(1, authDetail.getCredentialIdentifiers().size());
+            assertEquals("test-identifier-123", authDetail.getCredentialIdentifiers().get(0));
+        }
     }
```
IngridPuppet: I tend to find this alternative more readable. WDYT?
```suggestion
    private void assertValidAuthorizationDetailResponse(OID4VCAuthorizationDetail authDetail) {
           assertValidAuthorizationDetail(OID4VCAuthorizationDetail authDetail);
             
            assertEquals(1, authDetail.getCredentialIdentifiers().size());
            assertEquals("test-identifier-123", authDetail.getCredentialIdentifiers().get(0));
}
```

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
```diff
@@ -43,6 +43,13 @@ public AccessTokenRequest codeVerifier(String codeVerifier) {
     }
 
     public AccessTokenRequest authorizationDetails(List<OID4VCAuthorizationDetail> authDetails) {
+        if (authDetails != null) {
+            for (OID4VCAuthorizationDetail authDetail : authDetails) {
+                if (authDetail.getCredentialIdentifiers() != null) {
+                    throw new IllegalArgumentException("credential_identifiers not supported on AccessToken request");
+                }
+            }
+        }
```
IngridPuppet: We tend to avoid OID4VC-specific code in general OAuth classes. Is there a way to fix this?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/OID4VCICredentialOfferMatrixTest.java
```diff
@@ -302,18 +299,40 @@ void runCredentialOfferTest(OfferTestContext ctx) throws Exception {
...
+                oauth.fillLoginForm(username,"password");
+                String authCode = oauth.parseLoginResponse().getCode();
 
-                AccessTokenResponse tokenResponse = getBearerTokenResponse(clientId, userId, scope);
-                String accessToken = tokenResponse.getAccessToken();
+                // Redeem Authorization Code for AccessToken
+                //
+                AccessTokenResponse tokenResponse = oauth.accessTokenRequest(authCode)
+                        .authorizationDetails(List.of(authDetail))
+                        .send();
```
IngridPuppet: Potential reuse of `getIssuerAccessToken`?

IngridPuppet (commented):
Hello @tdiesler - Beyond the point about a unified OID4VCAuthorizationDetail under discussion, I checked the changes and left some comments. Please could you consider them?

core/src/main/java/org/keycloak/OAuth2Constants.java
tdiesler: I could try to move all of them, but they should stay together, right?

services/src/test/java/org/keycloak/protocol/oid4vc/issuance/OID4VCAuthorizationDetailsProcessorTest.java
tdiesler: done

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
tdiesler: removed

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oid4vc/issuance/OID4VCICredentialOfferMatrixTest.java
tdiesler: This is the holder access token

mposolda:
@IngridPuppet Thanks for the input and additional PR review.

Also consider dedicated request/response more clear, but OK with the approach in current PR after added validation...

@tdiesler It might be needed to rebase due the conflict with other PR merged in the meantime.

forkimenjeckayang:
> @forkimenjeckayang @IngridPuppet What is your preference?

As @IngridPuppet  already mentioned , we pushed first to have both request/responses seperate. Also the refactoring to unify 
`OID4VCAuthorizationDetail` into a single class (removing `OID4VCAuthorizationDetailResponse`) is not a bad idea provided the  implementation correctly addresses the risk of misuse by strictly validating that response-only fields (`credential_identifiers`) are not present in requests as @mposolda  suggested. 
I see that this security check is covered by integration tests (specifically `OID4VCAuthorizationDetailsFlowTestBase.testAuthorizationCodeFlowWithCredentialIdentifier`), which ensures requests with forbidden fields are rejected.
Assides @IngridPuppet comments, I have no other concerns

mposolda (approved):
@tdiesler Thanks, LGTM considering the agreement was to go with the changes and have single type.

@forkimenjeckayang @IngridPuppet Thanks for the additional feedback and reviews!

PR #45961, opened 2026-02-03
Update Infinispan CRs in multi-cluster architecture to use InPlaceRolling upgrade strategy

### LINKED ISSUE #45424: Update Infinispan CRs in multi-cluster architecture to use InPlaceRolling upgrade strategy
### Description

In order to allow Infinispan clusters to be upgraded to new patch versions without downtime on a single cluster, we need to explicitly configure the `InPlaceRolling` strategy. Keycloak-benchmark deployments and the multi-cluster guides should be updated to reflect this.

The Infinispan CR needs to be updated to define the following:

```yaml
spec:
  upgrades:
    type: InPlaceRolling
```

When testing, do one smoke test to see how Keycloak reacts to that. Maybe test it with volatile sessions and see if you are still logged in.

ryanemerson:
Closes #45424

The various postgres files removed in the diff are unneeded and are removed as a result of the fix for https://github.com/keycloak/keycloak-benchmark/issues/1270.

### DIFF
--- a/docs/guides/high-availability/examples/generated/ispn-single.yaml
+++ b/docs/guides/high-availability/examples/generated/ispn-single.yaml
@@ -311,8 +311,8 @@ spec:
   expose:
     type: Route
   configMapName: "cluster-config"
-  image: quay.io/infinispan/server:15.0.19.Final
-  version: 15.0.19
+  image: quay.io/infinispan/server:16.0.5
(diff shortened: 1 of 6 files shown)


pruivo (approved):

ahus1:
I've marked it ISPN16 related so we know that we might need to roll it back.

PR #45964, opened 2026-02-03
Document minimum required ServiceMonitor CRD version

### LINKED ISSUE #45888: ServiceMonitor is not created due to missing fields
### Before reporting an issue

### Area

observability

### Describe the bug

When using the built in serverMonitor functionality introduced in the operator in version 26.4.0 are we not getting any metrics from keycloak due to the operator being unable to create the serviceMonitor configuration in OKD

### Version

26.5.2 and version 26.4.7 off the operator

### Regression

### Expected behavior

I expect to see a serviceMonitor being created

### Actual behavior

No serviceMonitor is being created, and the operator pod is providing a stacktrace in it's log.

### How to Reproduce?

Install the operator and configure an keycloak instance with it on OKD. 

### Anything else?

I'm seeing the following stacktrace in the operator logs

[kc-operator-stacktrace.log](https://github.com/user-attachments/files/24961341/kc-operator-stacktrace.log)

ryanemerson:
Closes #45888

### DIFF
--- a/docs/guides/operator/advanced-configuration.adoc
+++ b/docs/guides/operator/advanced-configuration.adoc
@@ -541,8 +541,8 @@ spec:
 === ServiceMonitor
 A `ServiceMonitor` resource is used to define how a service's metrics are discovered and scraped by Prometheus.
 The {project_name} Operator automatically generates a `ServiceMonitor` resource for your deployment. In order for the
-`ServiceMonitor` resource to be created, the `monitoring.coreos.com/v1:ServiceMonitor` Custom Resource Definition (CRD)
-must be installed on your {kubernetes} cluster.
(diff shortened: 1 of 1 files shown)


shawkins (approved):
LGTM, thanks @ryanemerson

ahus1:
My 2 cents here: I see the documentation is now updated and correct, still it doesn't provide information to the admin when they see the error message. 

I doubt that one would check the API version before enabling the feature. And to be honest I wouldn't know how. 

Something that would help a user would be a more descriptive error message telling them their CRD version is too old and what the minimum requirement is. Possibly in the log, and also in the CR status.

shawkins:
> Something that would help a user would be a more descriptive error message telling them their CRD version is too old and what the minimum requirement is. Possibly in the log, and also in the CR status.

That would be nice, but overkill in this case. The version of the prometheus operator needed is now nearly 2 years old, so additional users are unlikely to encounter this.

PR #45972, opened 2026-02-03
fix: decode URI path before opening JAR file

### LINKED ISSUE #45971: Paths with spaces are not decoded when trying to discover providers JAR file
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

When Keycloak is installed in a path that includes spaces (e.g Program Files) there is an error thrown when trying to discover the providers JAR file. The error reads "Failed to discover providers" followed by the path. In the path, spaces are replaced by %20.

### Version

26.4.1

### Regression

### Expected behavior

Paths with spaces are decoded before opening the file

### Actual behavior

Paths with spaces are encoded and the file cannot be found

### How to Reproduce?

Install Keycloak in a path with spaces and run the start command

### Anything else?

_No response_

phantomion:
Decode JAR file path before reading it. This allows to have paths that with spaces in them.

Closes [#45971](https://github.com/keycloak/keycloak/issues/45971)

### DIFF
--- a/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
+++ b/quarkus/deployment/src/main/java/org/keycloak/quarkus/deployment/KeycloakProcessor.java
@@ -1040,7 +1040,7 @@ class KeycloakProcessor {
                         }
                     }
                 }
-            } catch (IOException e) {
+            } catch (IOException | URISyntaxException e) {
                 throw new RuntimeException("Failed to discover script providers", e);
             }
         }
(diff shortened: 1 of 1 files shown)


shawkins:
Instead of url.getPath(), let's do url.toURI().getSchemeSpecificPart() - that will be properly unescaped.

shawkins:
@phantomion sorry I meant to just change how we're getting the file initially to use getSchemeSpecificPart. I've added a commit of that. If you are ok with that, please squash this to a single commit and it should be ready for inclusion. 

I realize a test would be ideal, but the amount of setup needed is a little disproportionate, so I'm inclined to not require one.

shawkins (approved):
@vmuzikar I'm good with the current state of this change. We can wait for the contributor to rebase, or just proceed to do that ourselves.

vmuzikar (commented):
@phantomion Changes look good to me, thank you for the PR.

I just wonder if we should have a test running KC from a dir with spaces in its name. But I see that as a more complex task, hence as a follow-up. WDYT @shawkins?

shawkins:
> But I see that as a more complex task, hence as a follow-up. WDYT @shawkins?

Right, I don't see a test in this case as required: https://github.com/keycloak/keycloak/pull/45972#issuecomment-3855255789

vmuzikar:
Ok, created https://github.com/keycloak/keycloak/issues/46135.

vmuzikar (approved):

Issue #44657, opened 2025-12-04
[OID4VCI] Keycloak should accept holder did as client_id in authorization request


tdiesler:
### Description

In EBSI CT v3.2 the wallet sends an authorization request like this ...

```
client_id=did:key:z2dmzD81cg...BvhAC
redirect_uri=openid://redirect 
response_type=code 
scope=openid 
authorization_details=[
  {
    "type":"openid_credential",
    "locations":["https://proxy.nessustech.io:8443/issuer/72420f426cac"],
    "format":"jwt_vc",
    "types":["VerifiableCredential","VerifiableAttestation","CTWalletSameAuthorisedInTime"]
   }]
client_metadata={
   "authorization_endpoint":"openid:",
   "jwks_uri":"https://api-conformance.ebsi.eu/conformance/v3/issuer-mock/jwks",
   "redirect_uris":["openid://redirect"]
 } 
state=07f14faf-c2ed-482a-bdd0-de11a1ef9be0 
code_challenge=TrMhZMLoWg97SYY6qf6JkI3O3Xdm_2sKZ3nbPpsZ5t4 
code_challenge_method=S256
```

For this to work the issuer (Keycloak) would have to have an oidc client configuration for every potential holder did.

Alternatively, there would need to be a mapping that resolves a given did scheme to a target client configuration

### Value Proposition

Provide a login mechanism that works for EBSI Authorization Requests

### Goals

see above

### Non-Goals

n/a

### Notes

CrossRef: https://github.com/tdiesler/nessus-identity/issues/332

tdiesler:
The issuer cannot rely on a standard OIDC client registration for each holder DID. Instead, the wallet acts as a Self-Issued OP, meaning:

* The wallet is an OpenID Provider
* The issuer is an OIDC Relying Party during this sub-handshake
* The issuer requests an ID Token from the wallet to authenticate the holder

This is not OIDC4VCI. It is defined by [Self-Issued OpenID Provider v2 (SIOP v2)](https://openid.net/specs/openid-connect-self-issued-v2-1_0.html)

* Wallet acting as an OP
* Issuer requesting an ID Token
* DID-based subject identifiers
* Wallet-signed ID Tokens proving holder binding

The IDToken handshake is described [here](https://hub.ebsi.eu/conformance/build-solutions/issue-to-holder-functional-flows#in-time-issuance)

tdiesler:
working on this ...

tdiesler:
out-of-date

PR #45985, opened 2026-02-03
[OID4VCI] Revisit and fix OAuthClient.authorizationRequest()

tdiesler:
closes #45979

### DIFF
--- a/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractOAuthClient.java
+++ b/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractOAuthClient.java
@@ -75,6 +75,10 @@ public abstract class AbstractOAuthClient<T> {
         return client();
     }
 
+    public AuthorizationEndpointRequest authorizationRequest() {
+        return new AuthorizationEndpointRequest(this);
+    }
+
     public LoginUrlBuilder loginForm() {
         return new LoginUrlBuilder(this);
     }
@@ -313,5 +317,4 @@ public abstract class AbstractOAuthClient<T> {
     private T client() {
         return (T) this;
     }
-
 }
--- a/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenResponse.java
+++ b/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenResponse.java
@@ -2,6 +2,7 @@ package org.keycloak.testsuite.util.oauth;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -136,7 +137,7 @@ public class AccessTokenResponse extends AbstractHttpResponse {
 
     private <ADR extends AuthorizationDetailsJSONRepresentation> List<ADR> getAuthorizationDetails(Class<ADR> clazz) {
         if (authorizationDetails == null) {
-            return null;
+            return Collections.emptyList();
         }
         return authorizationDetails.stream()
                 .map(authzResponse -> authzResponse.asSubtype(clazz))
--- a/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationEndpointRequest.java
+++ b/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationEndpointRequest.java
@@ -1,0 +1,85 @@
+package org.keycloak.testsuite.util.oauth;
+
+import java.util.List;
+
+import org.keycloak.representations.AuthorizationDetailsJSONRepresentation;
+
+public class AuthorizationEndpointRequest {
+
+    protected final AbstractOAuthClient<?> client;
+    protected final LoginUrlBuilder loginForm;
+
+    public AuthorizationEndpointRequest(AbstractOAuthClient<?> client) {
+        this.client = client;
+        this.loginForm = client.loginForm();
+    }
+
+    public AuthorizationEndpointRequest authorizationDetails(AuthorizationDetailsJSONRepresentation authDetail) {
+        loginForm.authorizationDetails(List.of(authDetail));
+        return this;
+    }
+
+    public AuthorizationEndpointRequest authorizationDetails(List<AuthorizationDetailsJSONRepresentation> authDetails) {
+        loginForm.authorizationDetails(authDetails);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest client(String clientId) {
+        loginForm.clientId(clientId);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest clientState(String clientState) {
+        loginForm.state(clientState);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest codeChallenge(PkceGenerator pkce) {
+        loginForm.codeChallenge(pkce);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest redirectUri(String redirectUri) {
+        loginForm.redirectUri(redirectUri);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest responseType(String responseType) {
+        loginForm.responseType(responseType);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest responseMode(String responseMode) {
+        loginForm.responseMode(responseMode);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest request(String request) {
+        loginForm.request(request);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest scope(String... scopes) {
+        loginForm.scope(scopes);
+        return this;
+    }
+
+    public AuthorizationEndpointRequest nonce(String nonce) {
+        loginForm.nonce(nonce);
+        return this;
+    }
+
+    protected String getEndpoint() {
+        return client.getEndpoints().getAuthorization();
+    }
+
+    protected void initRequest() {
+    }
+
+    public AuthorizationEndpointResponse send(String username, String password) {
+        initRequest();
+        loginForm.open();
+        client.fillLoginForm(username, password);
+        return client.parseLoginResponse();
+    }
+}
--- a/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/LoginUrlBuilder.java
+++ b/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/LoginUrlBuilder.java
@@ -1,10 +1,12 @@
 package org.keycloak.testsuite.util.oauth;
 
 import java.util.Arrays;
+import java.util.List;
 
 import org.keycloak.OAuth2Constants;
 import org.keycloak.models.Constants;
 import org.keycloak.protocol.oidc.OIDCLoginProtocol;
+import org.keycloak.representations.AuthorizationDetailsJSONRepresentation;
 import org.keycloak.representations.ClaimsRepresentation;
 
 public class LoginUrlBuilder extends AbstractUrlBuilder {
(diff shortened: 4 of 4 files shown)


tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationRequestRequest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.keycloak.protocol.oid4vc.model.AuthorizationRequest;
+import org.keycloak.protocol.oid4vc.model.OID4VCAuthorizationDetail;
+import org.keycloak.protocol.oidc.utils.PkceGenerator;
+import org.keycloak.util.Strings;
+
+import org.apache.http.client.methods.CloseableHttpResponse;
+
+import static org.keycloak.OAuth2Constants.SCOPE_OPENID;
+
+public class AuthorizationRequestRequest extends AbstractHttpGetRequest<AuthorizationRequestRequest, AuthorizationRequestResponse> {
```
mposolda: It will be good to avoid introducing another class, but rather would be good to re-use existing `LoginUrlBuilder` and add support to it for the new parameters like `issuer_state` and `authorization_details` . As it will be good to avoid having many duplicated "request" classes in the codebase, which are doing same (or very similar) thing.

Similar point: Would be also good to avoid introducing `org.keycloak.protocol.oid4vc.model.AuthorizationRequest` in the "services" module. Don't see the need of this class (especially considering that we have already `AuthorizationEndpointRequest` already in that module, but this is not used by the testsuite and hence is likely out of scope of this PR).

services/src/main/java/org/keycloak/protocol/oidc/utils/PkceGenerator.java
```diff
@@ -1,7 +1,6 @@
-package org.keycloak.testsuite.util.oauth;
+package org.keycloak.protocol.oidc.utils;
 
 import org.keycloak.OAuth2Constants;
-import org.keycloak.protocol.oidc.utils.PkceUtils;
```
mposolda, line 4: Why this class was needed to be moved from testsuite? Don't seeing much reason for that in the context of this PR

mposolda (commented):
@tdiesler Thanks for the PR. Added some comments inline. Could you check?

services/src/main/java/org/keycloak/protocol/oidc/utils/PkceGenerator.java
tdiesler: It is there, so that `org.keycloak.protocol.oid4vc.model.AuthorizationRequest` can see it. `AuthorizationRequest` is currently a sibling of `CredentialRequest` which it probably should be for the time being. I could (if you prefer) move it next to or even into `AuthorizationRequestRequest` in which case it'd become part of the `org.keycloak.testsuite.util.oauth.oid4vc` and `PkceGenerator` would not have to move.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationRequestRequest.java
tdiesler: In #45851 I introduce proper error handling like this ...

```
    public AuthorizationRequestResponse send() {
        initRequest();
        String endpointUrl = authRequest.toRequestUrl(getEndpoint());
        client.driver.navigate().to(endpointUrl);
        AuthPageState authPageState = waitForLoginOrError();
        switch (authPageState) {
            case ERROR_REDIRECT -> {
                return new AuthorizationRequestResponse(client);
            }
            case ERROR_PAGE -> {
                Map<String, String> params = new HashMap<>();
                params.put("error", "invalid_request");
                client.driver.findElements(By.id("kc-error-message")).stream()
                        .map(d -> d.getText().trim().split("\\n")[0])
                        .forEach(s -> params.put("error_description", s));
                return new AuthorizationRequestResponse(params);
            }
            // AuthPageState.LOGIN
            default -> {
                client.fillLoginForm(username, password);
                return new AuthorizationRequestResponse(client);
            }
        }
```

Currently, we cannot test the error scenario with AuthorizationRequests - it waits for 10sec and then you get a timeout. the error message is lost.

Later still, in #44657 I do a complex IDToken Authorization (when clientId is a DID).

I thought it be good to do this in our oauth.oid4vci() package first, before I change anything in the already existing API that effects everyone. Can please keep this for a little while longer, until we are certain taht we can promote it to the oauth API?

The pattern is this ...

```
    public AuthorizationRequestRequest authorizationRequest() {
        return authorizationRequest(new AuthorizationRequest());
    }

    public AuthorizationRequestRequest authorizationRequest(AuthorizationRequest authRequest) {
        return new AuthorizationRequestRequest(client, authRequest);
    }

    public Oid4vcCredentialRequest credentialRequest() {
        return credentialRequest(new CredentialRequest());
    }

    public Oid4vcCredentialRequest credentialRequest(CredentialRequest credRequest) {
        return new Oid4vcCredentialRequest(client, credRequest);
    }
```

i.e. you can either build the request object outside the API (you might also already have it) or populate it through the API. In any case, I would like to have an AuthorizationRequest object so that it can be logged, stored in a context or otherwise processed - it encapsulates what was actually sent when making the request.

AuthorizationRequest supports PKCE and `issuer_state` and lives in the oid4vci model package. Later when it comes to IDToken Authorization (as required by EBSI) there is other stuff that might need to go in it. Here as well, please can I keep it, until we decide whether to merge oid4vci stuff into the oauth `AuthorizationEndpointRequest`.

services/src/main/java/org/keycloak/protocol/oidc/utils/PkceGenerator.java
tdiesler: this is resolved

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AuthorizationRequestRequest.java
mposolda: @tdiesler It will be good to avoid introducing multiple objects for the same thing. The OID4VCI should use `LoginUrlBuilder` directly instead of introducing it's own object for the same thing. It will be beneficial if `LoginUrlBuilder` has support for the things like `authorization_details` anyway.


> Currently, we cannot test the error scenario with AuthorizationRequests - it waits for 10sec and then you get a timeout. the error message is lost.

Existing Keycloak tests work in a way that you know if you expect "success" or "error" . For example when you expect error, you may call something like:
```
errorPage.assertCurrent();
```
Or eventually, there are similar things to check if the login-page is displayed with some error on the page etc.

I suppose something like `public AuthorizationRequestResponse send(expectedOutcome) {` should be used instead as the caller should know if it expects "success" page or "error" page.


> Later still, in https://github.com/keycloak/keycloak/issues/44657 I do a complex IDToken Authorization (when clientId is a DID).

This is different thing, so might be rather good to introduce related building blocks once/if we add support for this. I see the IDToken authorization handshake rather as out-of-scope of OID4VCI preview feature as it is specific to EBSI and not related to OID4VCI itself.

tdiesler:
Sending an Authorization request with low level LoginForm API looks like this

```
        oauth.loginForm().scope(credScopeName).authorizationDetails(authDetail).open();
        oauth.fillLoginForm(appUsername, "password");
        AuthorizationEndpointResponse authResponse = oauth.parseLoginResponse();
```

It is awkward, imho

We can preserve the use of LoginForm (i.e. not send an HttpGetRequest from another AuthorizationRequestRequest)  and at the time align with the general OAuthClient API like this ...

```
        AuthorizationEndpointResponse authResponse = oauth
                .authorizationRequest()
                .scope(credScopeName)
                .authorizationDetails(authDetail)
                .send(appUsername, "password");
        String authCode = authResponse.getCode();
```

Note, this would ...

* not be oid4vci specific
* authorization_details is added to LoginForm
* the response object is the same as with the LoginForm API 
* using the LoginForm API works as before (i.e. no change to code that uses it)

Would that be acceptable?

mposolda:
@tdiesler Thanks for the updates. It is probably subjective what is awkward. I am not very sure for a new way with `AuthorizationEndpointRequest` to be much better and moreover:
- We would have multiple ways to achieve the same thing with the API
- The way of `AuthorizationEndpointRequest` supports only very specific case when login happens with username/password and redirects to the client straight after. Does not support more complex scenarios (EG. if additional authentication like OTP is required, or login with social, flows with forget password etc)
- AFAIK it is by design that some parts (like `clientId`) are configured in the OAuth client rather than for the specific request. As it helps with the subsequent requests that stuff like `clientId` is configured on OAuthClient and does not need to be configured again for the particular test
- The `clientSecret` is certainly not needed on `AuthorizationEndpointRequest` ? Client authentication usually happens in the later stage (during token-request rather than during OIDC authentication request).

I will check with the rest of the team as would like more feedback on OAuth client API used in many tests.

AFAIK the minimal needed would be to add `authorizationDetails` (and eventually also `issuerState` if you prefer) to the `LoginUrlBuilder` .

tdiesler:
> * The way of `AuthorizationEndpointRequest` supports only very specific case when login happens with username/password and redirects to the client straight after. Does not support more complex scenarios (EG. if additional authentication like OTP is required, or login with social, flows with forget password etc)

It allows us to write code like this

```
        AuthorizationEndpointResponse authResponse = oauth
                .authorizationRequest()
                .scope(credScopeName)
                .authorizationDetails(authDetail)
                .send(holder, "password");
        String authCode = authResponse.getCode();
        assertNotNull(authCode, "No authCode");
```

If you remember, I originally did not wan to have it on `oauth` but on `oauth.oid4cv()` so it does not get exposed to erveryone before we are sure that it is ready.

In a follow up, I want to add support for improved error handling

```
    public AuthorizationEndpointResponse send(String username, String password) {
        initRequest();
        loginForm.open();
        AuthPageState authPageState = waitForLoginOrError();
        switch (authPageState) {
            case ERROR_REDIRECT -> {
                return new AuthorizationEndpointResponse(client);
            }
            case ERROR_PAGE -> {
                Map<String, String> params = new HashMap<>();
                params.put("error", "invalid_request");
                client.driver.findElements(By.id("kc-error-message")).stream()
                        .map(d -> d.getText().trim().split("\\n")[0])
                        .forEach(s -> params.put("error_description", s));
                return new AuthorizationEndpointResponse(params);
            }
            // AuthPageState.LOGIN
            default -> {
                client.fillLoginForm(username, password);
                return new AuthorizationEndpointResponse(client);
            }
        }
    }
```

also preferable for oid4vc only at first.

Still later, I want to add IDToken Authorization #44657

We need this abstraction for a number of reasons and ideally we would have it for oid4vc only and later promote it to `oauth` when it needed.

> Does not support more complex scenarios (EG. if additional authentication like OTP is required, or login with social, flows with forget password etc)

Yes, these are not supported nor needed in our current testsuite migration effort

> * AFAIK it is by design that some parts (like `clientId`) are configured in the OAuth client rather than for the specific request. As it helps with the subsequent requests that stuff like `clientId` is configured on OAuthClient and does not need to be configured again for the particular test

That remains the default.

> * The `clientSecret` is certainly not needed on `AuthorizationEndpointRequest` ? Client authentication usually happens in the later stage (during token-request rather than during OIDC authentication request).

Yes, I'm sorry that needs to be removed.

> AFAIK the minimal needed would be to add `authorizationDetails` (and eventually also `issuerState` if you prefer) to the `LoginUrlBuilder` .

Yes, that is already done so

---

How about, I move it back to `oid4vc`? It would not do any harm there and could unblock the testsuite migration PR.

stianst:
This is the right way of doing this:
```
        oauth.loginForm().scope(credScopeName).authorizationDetails(authDetail).open();
        oauth.fillLoginForm(appUsername, "password");
        AuthorizationEndpointResponse authResponse = oauth.parseLoginResponse();
```

Just do it that way and you're all good?

That basically goes hand in hand with OAuth (app sends authz request, the user fills the login form aka authenticates, and you move on with oauth stuff).

User may already be authenticated, may not use password only, may need to have some required actions, etc. That is basically the reason it is done the way it is currently.

PR #46004, opened 2026-02-04
[OID4VCI] Revisit and fix OAuthClient.preAuthorizedCodeGrantRequest()

### LINKED ISSUE #46003: [OID4VCI] Revisit and fix OAuthClient.preAuthorizedCodeGrantRequest()
### Description

Currently we send a pre-auth AccessToken request like this ...

```
        PreAuthorizedCode preAuthorizedCode = credOffer.getGrants().getPreAuthorizedCode();
        AccessTokenResponse accessTokenResponse = oauth.oid4vc()
                .preAuthorizedCodeGrantRequest(preAuthorizedCode.getPreAuthorizedCode())
                .endpoint(ctx.authorizationMetadata.getTokenEndpoint())
                .send();
        int statusCode = accessTokenResponse.getStatusCode();
        if (HttpStatus.SC_OK != statusCode) {
            throw new IllegalStateException(accessTokenResponse.getErrorDescription() != null
                    ? accessTokenResponse.getErrorDescription()
                    : "Request failed with status " + statusCode);
        }
        String accessToken = accessTokenResponse.getAccessToken();
```

This code has several issues ...

* It suggests that the endpoint Url is needed
* The pre-auth AccessToken could reuse oauth.accessTokenRequest() because it is one
* It does not support `authorization_details`
* Error handling should not have to be duplicated for every request

### Value Proposition

* remove calls to PreAuthorizedCodeGrantRequest.endpoint()
* reuse existing AccessTokenRequest, which supports `authorization_details` already
* error handling by AccessTokenResponse

### Goals

Cleaner and more reliable code

### Non-Goals

--

### Discussion

_No response_

### Notes

_No response_

tdiesler:
closes #46003

### DIFF
--- a/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
+++ b/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
@@ -14,7 +14,7 @@ public class AccessTokenRequest extends AbstractHttpPostRequest<AccessTokenReque
 
     private final String code;
 
-    AccessTokenRequest(String code, AbstractOAuthClient<?> client) {
+    public AccessTokenRequest(AbstractOAuthClient<?> client, String code) {
         super(client);
         this.code = code;
     }
@@ -52,6 +52,11 @@ public class AccessTokenRequest extends AbstractHttpPostRequest<AccessTokenReque
         return this;
     }
 
+    public AccessTokenRequest redirectUri(String redirectUri) {
+        parameter(OAuth2Constants.REDIRECT_URI, redirectUri);
+        return this;
+    }
+
     public AccessTokenRequest param(String name, String value) {
         parameter(name, value);
         return this;
@@ -59,14 +64,14 @@ public class AccessTokenRequest extends AbstractHttpPostRequest<AccessTokenReque
 
     protected void initRequest() {
         parameter(OAuth2Constants.GRANT_TYPE, OAuth2Constants.AUTHORIZATION_CODE);
-
         parameter(OAuth2Constants.CODE, code);
-        parameter(OAuth2Constants.REDIRECT_URI, client.getRedirectUri());
+        if (!hasParameter(OAuth2Constants.REDIRECT_URI)) {
+            parameter(OAuth2Constants.REDIRECT_URI, client.getRedirectUri());
+        }
     }
 
     @Override
     protected AccessTokenResponse toResponse(CloseableHttpResponse response) throws IOException {
         return new AccessTokenResponse(response);
     }
-
 }
--- a/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractHttpPostRequest.java
+++ b/tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AbstractHttpPostRequest.java
@@ -155,6 +155,17 @@ public abstract class AbstractHttpPostRequest<T, R> {
         return MediaType.APPLICATION_JSON;
     }
 
+    protected String getParameter(String key) {
+        return parameters.stream()
+                .filter(vp -> vp.getName().equals(key))
+                .map(NameValuePair::getValue)
+                .findFirst().orElse(null);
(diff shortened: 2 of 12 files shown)


tdiesler:
Consider merging #46014 instead

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
```diff
@@ -52,21 +59,30 @@ public AccessTokenRequest dpopProof(String dpopProof) {
...
         return this;
     }
 
     protected void initRequest() {
-        parameter(OAuth2Constants.GRANT_TYPE, OAuth2Constants.AUTHORIZATION_CODE);
-
-        parameter(OAuth2Constants.CODE, code);
-        parameter(OAuth2Constants.REDIRECT_URI, client.getRedirectUri());
+        parameter(OAuth2Constants.GRANT_TYPE, grantType);
+        if (grantType.equals(OAuth2Constants.AUTHORIZATION_CODE)) {
```
mposolda: I don't consider this change as an advantage.

So far, we use dedicated request types for each grant and `AccessTokenRequest` is reserved for the "authorization code" grant. Similar "grant requests" are for example `PasswordGrantRequest`, `ClientCredentialsGrantRequest` etc.

Having dedicated type `PReAuthorizedCodeGrantRequest` is better instead of trying to handle both "authorization code" and "pre-authorized code" in the single class `AccessTokenRequest` . The parameters of "pre-authorized code" grant are different than parameters of "authorization code" grant and they would differ even more (especially once we add support for `tx_code` for example).

mposolda (commented):
Not fan of this change. The current codebase with dedicated "request" classes for "authorization code" and "pre-authorized code" is much cleaner to me. See inline comment.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
tdiesler: ok - reverted to dedicated request & preserved authorization_details on the request

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
```diff
@@ -14,7 +14,7 @@ public class AccessTokenRequest extends AbstractHttpPostRequest<AccessTokenReque
 
     private final String code;
 
-    AccessTokenRequest(String code, AbstractOAuthClient<?> client) {
+    public AccessTokenRequest(AbstractOAuthClient<?> client, String code) {
```
mposolda, line 17: It seems general pattern for those "request" objects (like `AccessTokenRequest` , `IntrospectionRequest` , `RefreshRequest` etc) is to use non-public constructor and make sure they are created from OAuthClient calls. (EG. by calling something like `oauthClient.accessTokenRequest(code)` ).

Is it possible to follow this pattern here as well instead of introducing public constructor?

I just checked that same applies for some other OID4VCI request objects (EG. `Oid4vcCredentialRequest`). It would be good to fix those as well and make their initialization bound within oauthClient. But could be a follow-up PR possibly...

mposolda (commented):
@tdiesler Thanks, added minor comment inline. Could you check please?

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/AccessTokenRequest.java
tdiesler: good catch

mposolda (approved):
@tdiesler Thanks!

tdiesler:
@mposolda thanks for merging

BTW, I do these "-- foo" commits in response to reviewer feedback, so that it is easier to see what has changed. When there are multiple commits to a PR, they are meant to get squashed. If you prefer that I squash before resubmitting for review thats fine too. 

In short, you can always squash these "-- commits" when you see them

PR #46016, opened 2026-02-04
Fix duplicate client-uuid path parameter in OpenAPI spec

zmotso:
Closes #46015

The endpoint `GET /admin/realms/{realm}/clients/{client-uuid}/roles/{role-name}/composites/clients/{client-uuid}` contains two path parameters with the same name `{client-uuid}`, which produces an invalid OpenAPI specification.

This breaks OpenAPI code generators like [oapi-codegen](https://github.com/oapi-codegen/oapi-codegen) that rely on unique parameter names per path.

## Solution

Rename `{client-uuid}` to `{targetClientUuid}` in `RoleContainerResource.java`.

`{clientUuid}` is not sufficient because most code generators normalise `{client-uuid}` (from the parent path) to `clientUuid` as well, which would still produce a collision. `{targetClientUuid}` makes the semantics clear: it refers to the client whose roles are being queried as composites, not the client that owns the role.

## Impact

This is a non-breaking change:
- The actual URL path segment value doesn't change
- Only the OpenAPI parameter name changes, making the spec valid

### DIFF
--- a/services/src/main/java/org/keycloak/services/resources/admin/RoleContainerResource.java
+++ b/services/src/main/java/org/keycloak/services/resources/admin/RoleContainerResource.java
@@ -424,7 +424,7 @@ public class RoleContainerResource extends RoleResource {
      * @param clientUuid
      * @return
      */
-    @Path("{role-name}/composites/clients/{client-uuid}")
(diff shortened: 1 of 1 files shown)


shawkins (approved):
LGTM, thank you for the PR @zmotso

PR #46019, opened 2026-02-04
fix: adding admin role invalidation when a new realm is found

### LINKED ISSUE #45966: KeycloakRealmImport: Realm created in DB but not visible in Admin Console until restart
### Before reporting an issue

### Area

operator

### Description

I am encountering an issue with the Keycloak Operator and the `KeycloakRealmImport` CRD (`k8s.keycloak.org/v2alpha1`).

When I apply a `KeycloakRealmImport` custom resource, the Operator processes it successfully. I can verify via SQL that the Realm is correctly created in the underlying database. However, the new Realm does not appear in the Keycloak Admin Console (dropdown menu) or API responses from the running Keycloak pods.

The Realm only becomes visible and usable after performing a rollout restart of the Keycloak StatefulSet.

### Environment

* **Keycloak Version:** 26.5

* **Keycloak Operator Version:** 26.5.1

* **Kubernetes Version:** 1.34.2

* **Database:** PostgreSQL 17

### Analysis / Potential Cause

This behavior suggests a synchronization issue between the temporary Import Job spawned by the Operator and the running Keycloak StatefulSet.

It appears that while the Import Job successfully writes to the database, the running Keycloak nodes (managed by the StatefulSet) do not receive an invalidation event via Infinispan/JGroups. Therefore, they continue to serve the old realm list from their local cache until a restart forces a reload from the database.

### Version

26.5

### Regression

### Expected behavior

The imported Realm should be visible in the Admin Console immediately after the `KeycloakRealmImport` resource is successfully reconciled. The running Keycloak instances should invalidate their realm cache once the import job completes.

### Actual behavior

The Realm is persisted in the database but remains invisible in the Admin Console until the application is restarted.

### How to Reproduce?

1. Deploy Keycloak using the Keycloak Operator.

2. Create a valid `KeycloakRealmImport` YAML file (see example below).

3. Apply the CRD: `kubectl apply -f import-realm.yaml`.

4. Wait for the Operator to process the import (Status becomes `DONE` / `Ready`).

5. Verify the database content (e.g., connect to the DB pod and query the `realm` table): **The Realm exists.**

6. Log in to the Keycloak Admin Console on the running Keycloak instance.

7. Check the realm dropdown: **The imported Realm is missing.**

8. Restart the Keycloak pods: `kubectl rollout restart statefulset/keycloak`.

9. Refresh the Admin Console: **The Realm is now visible.**

### Anything else?

_No response_

shawkins:
closes: #45966

In order to not require a restart on a new realm import there are two changes.

Whenever an admin performs an operation on the new realm, the new logic in RealmSessionCache will invalidate the master admin role if needed.

Then for that to be applicable to the current operation, it needs to also be taken into account for the MgmtPermissions role logic - the newly invalidated admin role is checked for new roles.

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmCacheSession.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmCacheSession.java
@@ -493,13 +493,16 @@ public class RealmCacheSession implements CacheRealmProvider {
                 return null;
             }
 
+            // to accommodate imports of new realms, check to see if the master admin role is up-to-date
             if (!model.getName().equals(Config.getAdminRealm())) {
                 RealmModel adminRealm = session.realms().getRealmByName(Config.getAdminRealm());
-                ClientModel clientModel = session.clients().getClientByClientId(adminRealm, model.getName() + "-realm");
-                if (clientModel != null) {
-                    RoleModel adminRole = adminRealm.getRole(AdminRoles.ADMIN);
-                    if (adminRole.getCompositesStream().noneMatch(r -> (r.isClientRole() && r.getContainerId().equals(clientModel.getId())))) {
-                        invalidateRole(adminRole.getId());
+                if (adminRealm != null) {
+                    ClientModel clientModel = session.clients().getClientByClientId(adminRealm, model.getName() + "-realm");
+                    if (clientModel != null) {
+                        RoleModel adminRole = adminRealm.getRole(AdminRoles.ADMIN);
+                        if (adminRole.getCompositesStream().noneMatch(r -> (r.isClientRole() && r.getContainerId().equals(clientModel.getId())))) {
+                            registerRoleInvalidation(adminRole.getId(), adminRole.getName(), adminRole.getContainerId());
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


model/infinispan/src/main/java/org/keycloak/models/cache/infinispan/RealmCacheSession.java
```diff
@@ -491,6 +493,15 @@ private RealmAdapter prepareCachedRealm(String id, KeycloakSession session) {
...
                 return null;
             }
+
+            if (!model.getName().equals(Config.getAdminRealm())) {
+                RealmModel adminRealm = session.realms().getRealmByName(Config.getAdminRealm());
+                RoleModel adminRole = adminRealm.getRole(AdminRoles.ADMIN);
+                if (!invalidations.contains(adminRole.getId())) {
+                    invalidateRole(adminRole.getId());
+                }
+            }
```
ahus1, line 508: This would only invalidate the admin role if the admin role is out of date, that is: If it contains any of the client roles of the realm client already, it should be ok to keep it as-is.

```suggestion
            if (!model.getName().equals(Config.getAdminRealm())) {
                RealmModel adminRealm = session.realms().getRealmByName(Config.getAdminRealm());
                ClientModel clientModel = session.clients().getClientByClientId(adminRealm, model.getName() + "-realm");
                if (clientModel != null) {
                    RoleModel adminRole = adminRealm.getRole(AdminRoles.ADMIN);
                    if (adminRole.getCompositesStream().noneMatch(r -> (r.isClientRole() && r.getContainerId().equals(clientModel.getId())))) {
                        invalidateRole(adminRole.getId());
                    }
                }
            }
```

ahus1 (approved):
Thank you for all the investigation to find this solution, and thank you for implementing this.

vmuzikar (approved):
As far as I can tell (I'm not to familiar with this area), LGTM. Thank you, @shawkins.

PR #46020, opened 2026-02-04
Simplify handling common sub-resources objects

mabartos:
Closes #46017


vmuzikar (changes requested):
@mabartos Thank you for the PR.

The concept of `ApiContext` feels to me like it clashes with `KeycloakSession`. The session should be the context bearer, we can even attach custom objects to it via attributes. I would suggest against introducing another context concept.

mabartos:
@vmuzikar Thanks. Yep, there might be seen some similarities with the KeycloakSession and the Keycloak Context... This is rather the API context, that might improve readability. But this is no hard requirement, and rather nice to have. We can discuss later once we have it working :) 

Putting the hold label.

Issue #45932, opened 2026-02-02
Prototype new Service Layer for Client v2


vmuzikar:
### Description

The aim of this spike is to assess the complexity and plan for implementing a new Service Layer that is independent of Admin v1 (i.e. doesn't call the v1 endpoints).

As part of this spike, prototype the Service Layer for a single endpoint of Client v2. The result should be a list of issues to break down the Service Layer implementation, and/or new areas that require individual spikes.

edewit:
A quick grep reveals the following dependencies on v1:

  • DefaultClientService imports ClientResource, ClientsResource, RealmAdminResource
  • DefaultClientsApi imports ClientsResource, RealmAdminResource
  • DefaultClientApi imports ClientResource, RealmAdminResource
  • DefaultAdminApi imports AdminAuth, AdminRoot, RealmAdminResource, RealmsAdminResource

Complexity Rating: **Medium-High**


  | Aspect                               | Complexity | Reason                                                                   |
  |--------------------------------------|------------|--------------------------------------------------------------------------|
  | Client CRUD                          | Medium     | Direct model operations, but must replicate event logging, policy checks |
  | Authorization/Permissions            | High       | Complex AdminPermissionEvaluator logic needs replication                 |
  | Client Policies                      | High       | Triggers on create/update/delete must be preserved                       |
  | Admin Events                         | Medium     | Must emit proper audit events                                            |
  | Role Management                      | Medium     | Direct model operations feasible                                         |
  | Service Account                      | Medium     | User provisioning logic needs extraction                                 |
  | Other Entities (Users, Realms, etc.) | High       | Each entity has similar complexity                                       |

Do we need to uncouple Authorization/Premissions?

We would need to

  1. Create Service Layer Abstractions
  • Enhance Service interface with common patterns
  • Create ServiceContext to hold session, realm, auth context
  • Create AdminEventService abstraction (decouple from v1's AdminEventBuilder)

  2. Extract Permission Evaluation
  • Create PermissionService interface wrapping authorization checks
  • Initially delegate to existing AdminPermissionEvaluator, later make independent
  • Pattern: permissionService.clients().requireManage(client)

  3. Extract Client Policy Handling
  • Create ClientPolicyService to wrap session.clientPolicy().triggerOnEvent()
  • Centralize policy trigger logic

mabartos:
@edewit Thanks for the initial analysis.  

## Complexity review

[...]

PR #45981, opened 2026-02-03
[client-v2] Create tests for checking permissions

mabartos:
- Closes #45975

Issue #44154, opened 2025-11-12
Client API v2 Service Layer


vmuzikar:
### Description

The implementation of Client v2 created in #43167 leverages the original v1 API logic. This drastically limits the capabilities of v2 in terms of defaulting, validations, pagination, querying etc.

For v2 we need a dedicated service layer that will be built directly on top of the model layer, without the dependency on v1.

### Goals

* Remove the dependency on v1 from v2.
* Create a dedicated service layer for v2 with corresponding behaviour to when v1 was used as the backend (i.e validations, eventing, auth, etc.).

### Non-Goals

* Add new features (pagination, querying, etc.) to v2.

stianst:
We could potentially revamp everything in Keycloak around using only interfaces together with some custom FasterXML serializer/deserializer and some Java proxies to get rid of all boilerplate code around model-to-representation and visa versa. 

We could also allow code to operate on the actual interfaces rather than using attributes. For example it could look something like this:

```
MyCustomProtocolMapper

ProtocolMapperModel model;
MyCustomProtocolMapperRepresentation config = model.as(MyCustomProtocolMapperRepresentation.class);
config.getMyCustomProperty();
```

Main point here is that MyCustomProtocolMapperRepresentation is just an interface, it's not an implementation. That's handled through proxies.

We have quite a lot of hacky boilerplate code not just in model-to-representation, but take identity providers for instance, and other places, where we can implementations that manually call getAttributes, etc.

Let's figure out a proper way of defining contracts both for APIs and internally that uses interfaces only with some annotations. I know this is a big ask, but we can figure out a way to do this in stages.

Another aspect here is we currently have custom components etc. that uses PropertyConfig or whatever; it would be nice if we get rid of this as well, and rather than creating a list of PropertyConfig you just define the config as a representation interface with some annotations. That means we can use the same interface in OpenAPI specs, and validation without having to check values manually in the code.

PR #46029, opened 2026-02-04
initial version of the client v2 policy test

### LINKED ISSUE #46074: Create client policy test for rest v2
### Description

implement client policy test for rest v2, so that when we create a new service layer we know that the implementation is sound

edewit:
related to #45932
fixes: #46074

### DIFF
--- a/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
+++ b/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
@@ -1,0 +1,579 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.keycloak.tests.admin.client.v2;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+
+import jakarta.ws.rs.core.HttpHeaders;
+import jakarta.ws.rs.core.MediaType;
+
+import org.keycloak.admin.api.AdminApi;
+import org.keycloak.admin.client.Keycloak;
+import org.keycloak.authentication.authenticators.client.ClientIdAndSecretAuthenticator;
+import org.keycloak.authentication.authenticators.client.JWTClientAuthenticator;
+import org.keycloak.authentication.authenticators.client.JWTClientSecretAuthenticator;
+import org.keycloak.authentication.authenticators.client.X509ClientAuthenticator;
+import org.keycloak.common.Profile;
+import org.keycloak.representations.admin.v2.OIDCClientRepresentation;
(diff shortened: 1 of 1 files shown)


mabartos:
@edewit Just before starting to reviewing that - would it be possible to create the sub-task for the https://github.com/keycloak/keycloak/issues/44154, to have it properly tracked? Thanks!

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
```diff
@@ -0,0 +1,579 @@
...
+
+    @BeforeAll
+    public static void setupMapper() {
+        mapper = new ObjectMapper();
+    }
+
+    @AfterEach
+    public void cleanup() throws Exception {
+        // Clean up any test clients
+        cleanupClient("test-policy-client");
```
mabartos, line 113: It'd be nice to have ManagedRealm for these cases, as we might not care about these cleanups. However, I had some problems to have master realm managed (even with the attachTo) and custom admin clients. 

On the other hand, we directly test the master realm, so the test coverage might be a little bit higher.

These test improvements can be done in a follow-up tasks with the help of guys working on the testsuite. 

So, it's ok to me as is for now.

edewit:
@mabartos done, reworded the commit

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
```diff
@@ -0,0 +1,579 @@
...
+        ClientPoliciesRepresentation policiesRep = new ClientPoliciesRepresentation();
+        policiesRep.setPolicies(List.of(policyRep));
+
+        adminClient.realm("master").clientPoliciesPoliciesResource().updatePolicies(policiesRep);
+    }
+
+    /**
+     * Sets up a policy with auto-configuration that defaults to X509 authenticator.
+     */
+    private void setupPolicyWithAutoConfiguration() throws Exception {
```
mabartos, line 428: These setups are almost identical, and it'd be good if we could unify it. 

Created PR to your fork: https://github.com/edewit/keycloak/pull/13

mabartos (commented):
@edewit From the first look, it looks very good. 

Just added some suggestion. 

I'll also verify these test cases do make sense and if ther relate to the spike we did. 

Thanks!

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
```diff
@@ -0,0 +1,579 @@
...
+        revertToBuiltinProfiles();
+        revertToBuiltinPolicies();
+    }
+
+    /**
+     * Test that creating a client via POST with an unacceptable client authenticator fails
+     * when a policy requires specific authenticators.
+     */
+    @Test
+    public void testCreateClientWithUnacceptableAuthType() throws Exception {
```
mabartos: @edewit nitpick: it'd be good to not use the prefix "test", as it does not add any added value and is considered as old JUnit approach.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
vmuzikar: +1 to a similar approach @mabartos did in https://github.com/keycloak/keycloak/pull/45981

vmuzikar (approved):
@edewit Thank you for the PR.

In general, I think it's good for now but obviously it'll need many refinements once we create clients properly and get rid of the "two-step approach".

mabartos:
@edewit @vmuzikar I think we can make improvements in the follow-up tasks. 

I created some additions on top of these tests that I'd like to have included (still WiP): https://github.com/keycloak/keycloak/pull/46093/changes/91417cac7d19086343da9dfc04f244f8d2150a11

mabartos (approved):

PR #46030, opened 2026-02-04
Adding exception to UsersResource when calling createUser with roles/clientRoles addition.

lhanusov:
Closes #37276


stianst (changes requested):
You're not going to be able to do this directly in UsersResources as a lot of tests will be broken + it's a breaking change.

I would suggest doing this in the user supplier only for the new test framework.

Issue #45444, opened 2026-01-14
Deprecate and remove `--optimized`


stianst:
### Description

`--optimize` is no longer relevant and often confusing, so should be removed completely. The server should just do an auto-build as applicable; and if needed the check if a build is required should be optimized further so it runs in milliseconds.

### Value Proposition

Remove confusion and mixed terms like build vs optimized.

### Goals

* Condition if auto-build is required should be less than 50 ms on a developer latop, and less than 200 ms on smaller containerized nodes
* `start` or `start-dev` should just decide if a build is required or not automatically

shawkins:
The biggest breaking change with complete removal is that you will no longer be able to simply omit build time options:

You can currently do:

./kc.sh build --db=mysql ...
./kc.sh start --optimized

such that the start --optimized doesn't specify --db.

If we remove --optimized completely you will be required to keep any cli buildtime args on the start command, put them in the env, or the keycloak.conf.

Related to this, https://github.com/keycloak/keycloak/pull/44579 changes the message shown in the console when using just start - we no longer stip the buildtime args.

stianst:
> The biggest breaking change with complete removal is that you will no longer be able to simply omit build time options:
> 
> You can currently do:
> 
> ./kc.sh build --db=mysql ... ./kc.sh start
> 
> such that the start doesn't specify --db.
> 
> If we remove --optimized completely you will be required to keep any cli buildtime args on the start command, put them in the env, or the keycloak.conf.
> 
> Related to this, [#44579](https://github.com/keycloak/keycloak/pull/44579) changes the message shown in the console when using just start - we no longer stip the buildtime args.

That's probably a good thing isn't it? Honestly, it's kinda confusing the whole thing that there may be some "hidden" values in the build

shawkins:
> That's probably a good thing isn't it? Honestly, it's kinda confusing the whole thing that there may be some "hidden" values in the build

I think so, just want to make sure that we're on the same page about the scope of what will change / break.

The other big breaking change is that users won't have a way to dictate that the augmented state is used - they'll have to be mindful of the start cli output to see if an autobuild is being performed.

Also want to confirm if we want to continue using the term "build", or even a separate command.

./kc.sh build ... options ...

Currently requires only buildtime options. That means you still need to have more intimiate knowledge about what is buildtime vs runtime.

We could allow the build command to accept all options, or just as easily allow:

./kc.sh start --build ... all options ...

The cli would simply warn you about which runtime options were ignored.

[...]

Issue #41465, opened 2025-07-28
Expose hibernate-orm properties in a supported manner


shawkins:
### Description

There are ~56 hibernate-orm options which need could be exposed as supported by keycloak. Options include:

- update keycloak.conf and cli handling to allow for quarkus.hibernate-orm options to be directly specified
- add mapping logic that directly maps kc.db-orm-<datasource> options to quarkus.hibernate-orm options without specifically adding Options and PropertyMappers for everything
- something else, such as a separate config file and forgo cli options.
- add formal Options and PropertyMappers

The first three options go against our current paradigms and doesn't appropriately identify build-time options - unless we add back logic that looks quarkus options for well-known build-time options. If we don't appropriatly identify build-time options we'll just have to assume that they all are, which does not work well with --optimized or #38894

cc @keycloak/cloud-native

mabartos:
> update keycloak.conf and cli handling to allow for quarkus.hibernate-orm options to be directly specified

@shawkins I'd not allow specifying the Hibernate ORM options on the CLI. Only via other sources. For the Hibernate ORM, you need to set multiple options to have it working and probably not only 1-2. I'd like to see the ORM settings in the keycloak.conf or other config source. 

> add mapping logic that directly maps kc.db-orm- options to quarkus.hibernate-orm options without specifically adding Options and PropertyMappers for everything

Yes, we can have some sort of wildcard mapper that would allow to map everything to the `quarkus.hibernate-orm.*`. It should probably need to be something like `db-orm-"datasource.something-here.value"`. Moreover, the same would apply to other mapping like `http-shutdown-*` that would map everything to `quarkus.shutdown.*`.

> something else, such as a separate config file and forgo cli options.

Yes, for cases like Hibernate ORM, we could potentially have another config source (.properties based) that could set the ORM and would not need to mess up the keycloak.conf readability/manageability. 

> The first three options go against our current paradigms and doesn't appropriately identify build-time options 

Yes, it might be quite hard to identify build-time options. 

> unless we add back logic that looks quarkus options for well-known build-time options.

We had that logic? So, if there's some way on the Quarkus side, we could look up whether the option is build-time or runtime, right?

shawkins:
> I'd not allow specifying the Hibernate ORM options on the CLI. Only via other sources.

I agree that it's not my prefered option. Being pragmatic as mentioned here: https://github.com/keycloak/keycloak/pull/45833#discussion_r2737680370 would exclude the usage on the CLI.

[...]

PR #46034, opened 2026-02-04
fix: adding dynamic detection of quarkus build-time options

shawkins:
closes: #45933

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java
@@ -1,0 +1,146 @@
+package org.keycloak.quarkus.runtime.cli;
+/*
+ * Copyright 2026 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.quarkus.runtime.util.ClassPathUtils;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.ConfigMappings;
+import io.smallrye.config.ConfigMappings.ConfigClass;
+import io.smallrye.config.PropertyNamesMatcher;
+
+import static io.smallrye.config.ConfigMappings.ConfigClass.configClass;
+
+public class QuarkusProperties {
+
+    private static final String CONFIG_ROOTS_LIST = "META-INF/quarkus-config-roots.list";
+
+    private static Iterable<Class<?>> classesNamedIn(ClassLoader classLoader, String fileName)
+            throws IOException, ClassNotFoundException {
+        final ArrayList<Class<?>> list = new ArrayList<>();
+        for (String className : classNamesNamedIn(classLoader, fileName)) {
+            list.add(Class.forName(className, true, classLoader));
+        }
+        return Collections.unmodifiableList(list);
+    }
+
+    private static Set<String> classNamesNamedIn(ClassLoader classLoader, String fileName) {
+        final Set<String> classNames = new LinkedHashSet<>();
+        try {
+            ClassPathUtils.consumeAsStreams(classLoader, fileName, classFile -> {
+                try (InputStreamReader reader = new InputStreamReader(classFile, StandardCharsets.UTF_8)) {
+                    try (BufferedReader br = new BufferedReader(reader)) {
+                        readStream(classNames, br);
+                    }
+                } catch (IOException e) {
+                    throw new UncheckedIOException(e);
+                }
+            });
+        } catch (Exception e) {
+            // TODO: could confirm this is indicative of a provider jar going missing
+        }
+        return Collections.unmodifiableSet(classNames);
+    }
+
+    /**
+     * - Lines starting by a # (or white spaces and a #) are ignored. - For
+     * lines containing data before a comment (#) are parsed and only the value
+     * before the comment is used.
+     */
+    private static void readStream(final Set<String> classNames, final BufferedReader br) throws IOException {
+        String line;
+        while ((line = br.readLine()) != null) {
+            int commentMarkerIndex = line.indexOf('#');
+            if (commentMarkerIndex >= 0) {
+                line = line.substring(0, commentMarkerIndex);
+            }
+            line = line.trim();
+
+            if (line.isEmpty()) {
+                continue;
+            }
+
+            classNames.add(line);
+        }
+    }
+
+    private static List<Class<?>> collectConfigRoots(ClassLoader classLoader) {
+        // populate with all known types
+        List<Class<?>> roots = new ArrayList<>();
+        try {
+            for (Class<?> clazz : classesNamedIn(classLoader, CONFIG_ROOTS_LIST)) {
+                if (!clazz.isInterface()) {
+                    throw new IllegalArgumentException(
+                            "The configuration " + clazz + " must be an interface annotated with @ConfigRoot and @ConfigMapping");
+                }
+
+                ConfigRoot configRoot = clazz.getAnnotation(ConfigRoot.class);
+                if (configRoot == null) {
+                    throw new IllegalArgumentException("The configuration " + clazz + " is missing the @ConfigRoot annotation");
+                }
+
+                ConfigMapping configMapping = clazz.getAnnotation(ConfigMapping.class);
+                if (configMapping == null) {
+                    throw new IllegalArgumentException("The configuration " + clazz + " is missing the @ConfigMapping annotation");
+                }
+
+                roots.add(clazz);
+            }
+        } catch (IOException | ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+        return roots;
+    }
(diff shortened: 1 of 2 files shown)


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java
```diff
@@ -0,0 +1,142 @@
...
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.quarkus.runtime.util.ClassPathUtils;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.ConfigMappings;
+import io.smallrye.config.ConfigMappings.ConfigClass;
+import io.smallrye.config.PropertyNamesMatcher;
+
+import static io.smallrye.config.ConfigMappings.ConfigClass.configClass;
+
+public class QuarkusProperties {
```
michalvavrik, line 40: I am bit worried there will come scenarios not covered in this PR, so I'll ask you a question.

IIUC: This class is probably heavily inspired by `BuildTimeConfigurationReader` class and the purpose is to not compare that _all config properties are same_, but that only build-time properties are same. Whether build (reaugmentation) is required is only relevant if the build has already happened. Which means that you already know when this logic is run:

-  what the build-time property names and values were when the build happened last time. The only thing you don't know is if the only added properties are a build-time properties.

So your issue is when the next time start is executed, if the new config contains some build time property. Problem is that `PropertyNamesMatcher` will never cover relocated properties as this class not aware of any config source interceptors, sometimes in Quarkus (mainly for K8 in the past), backwards compatibility was kept by relocating the old properties to the new build-time properties. What will happen then, is Keycloak configured to fail on detecting wrong build-time properties?

I did not find `quarkus.config.build-time-mismatch-at-runtime` set and the default value is `warn`, so this could be an issue if users don't immediately migrate to new properties.

shawkins: > This class is probably heavily inspired by BuildTimeConfigurationReader class

Yes from https://github.com/keycloak/keycloak/issues/41465#issuecomment-3825710321

> Problem is that PropertyNamesMatcher will never cover relocated properties as this class not aware of any config source interceptors, sometimes in Quarkus (mainly for K8 in the past), backwards compatibility was kept by relocating the old properties to the new build-time properties. 

There are two scenarios here. 

This pr is to better detect buildtime changes for arbitrary quarkus properties. Note that the direct usage of quarkus properties in keycloak is unsupported - so this is already a corner case. The existing logic will treat all quarkus properties in our quarkus.properties file effectively as build-time options. Ideally detection would be as reliable as possible in this case, but we'd be tolerant to some non-working scenarios given that this is already unsupported. And we do have the option of providing workarounds for this via #45444

The other scenario is #41465 where we want to simply re-expose a set of quarkus properties as if they are keycloak options. Here we do need to reliably detect which properties are build time. We also control exactly what properties will be used, so we can add a test or other logic that asserts our expected prefix is known to the propertynamesmatcher. If that still seems too nuanced, then we will have to abandon this approach. cc @radcortez

> What will happen then, is Keycloak configured to fail on detecting wrong build-time properties?

Currently if you issue an auto-build command (start, start-dev, etc.) without the --optimized flag and we detect that the build has changed, then we'll trigger a reaugmentation in our first java launch, then relaunch to use that augmentation.

michalvavrik: > Yes from https://github.com/keycloak/keycloak/issues/41465#issuecomment-3825710321

Yeah, I found it after I commented this :-/ maybe wouldn't write it if I did.

> This pr is to better detect buildtime changes for arbitrary quarkus properties. Note that the direct usage of quarkus properties in keycloak is unsupported - so this is already a corner case. The existing logic will treat all quarkus properties in our quarkus.properties file effectively as build-time options. Ideally detection would be as reliable as possible in this case, but we'd be tolerant to some non-working scenarios given that this is already unsupported. And we do have the option of providing workarounds for this via https://github.com/keycloak/keycloak/issues/45444

Yeah, that makes that part simple.

> The other scenario is https://github.com/keycloak/keycloak/issues/41465 where we want to simply re-expose a set of quarkus properties as if they are keycloak options. Here we do need to reliably detect which properties are build time. If that still seems too nuanced, then we will have to abandon this approach. cc @radcortez

I think this approach is good, I used something similar for QE framework when Quarkus migrated to config mappings from config classes. I only wanted you to be aware there are situations when "some quarkus hibernate property" != config mapping property names, because there is _something in the middle_, like interceptors that can relocate properties.

> We also control exactly what properties will be used, so we can add a test or other logic that asserts our expected prefix is known to the propertynamesmatcher. 

If you plan to have whitelist for property expressions, like `quarkus.hibernate-orm.*.validate-in-dev-mode` and the property matcher can deal with it, then I think you are good to go (yeah, I saw the red CI :grinning: , still).

OK. I have no more objections. Thanks for the answer.

shawkins: > OK. I have no more objections. Thanks for the answer.

Thank you for the feedback. I haven't deeply looked at the failures yet as this was just the initial draft. Should get to that shortly.

radcortez: > think this approach is good, I used something similar for QE framework when Quarkus migrated to config mappings from config classes. I only wanted you to be aware there are situations when "some quarkus hibernate property" != config mapping property names, because there is _something in the middle_, like interceptors that can relocate properties.

In that case, what you need is a `SmallRyeConfig` instance (with only the selected sources, but with the full list of relocate / fallback interceptors) to query the value.

michalvavrik: In case you are busy and want to help, shout, I think is approachable topic for me. Otherwise, I'll leave you to it.

michalvavrik: (I can't resolve the comment thread, so just do it for me please)

michalvavrik: > In that case, what you need is a SmallRyeConfig instance (with only the selected sources, but with the full list of relocate / fallback interceptors) to query the value.

thanks

shawkins:
Just running the readBuildTimeConfiguration logic is causing the tests to fail. Should I be using a different classloader? @michalvavrik @radcortez

michalvavrik:
> Just running the readBuildTimeConfiguration logic is causing the tests to fail. Should I be using a different classloader? @michalvavrik @radcortez

I think TCCL is the right choice, but test class loading is tricky. I have experience a month ago that I could not load a resource from dev svc startable unless I save TCCL from a build step. So I don't say that TCCL should work, I don't know. We can discuss it on Quarkus Zulip.

But before that,  right before the exception is raised, I can see:

```
Feb 06, 2026 12:06:06 AM org.keycloak.testsuite.arquillian.containers.AbstractQuarkusDeployableContainer undeploy
INFO: Undeploying archive scripts.jar from quarkus container
```

I don't know about this framework enough yet to decide if this is related, but I am mentioning it because if it was the same JAR on which it fails:

```
2026-02-04T22:00:48.2642246Z DeployedSAMLScriptMapperTest ++ java.lang.RuntimeException: java.nio.file.NoSuchFileException: /home/runner/work/keycloak/keycloak/testsuite/integration-arquillian/tests/base/target/containers/auth-server-quarkus/lib/../providers/scripts.jar
```

Regardless, following patch works around the issue for me (the purpose is just to have the test passing, not fine code):

```
diff --git a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java
index 09b4b56542..85d9d7b892 100644
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/QuarkusProperties.java
@@ -44,7 +44,21 @@ public class QuarkusProperties {
     private static Iterable<Class<?>> classesNamedIn(ClassLoader classLoader, String fileName)
             throws IOException, ClassNotFoundException {
         final ArrayList<Class<?>> list = new ArrayList<>();
-        for (String className : classNamesNamedIn(classLoader, fileName)) {
+
+        Set<String> classNames = null;
+        try {
+            classNames = classNamesNamedIn(classLoader, fileName);
+        } catch (Throwable throwable) {
+            var systemClassLoader = ClassLoader.getSystemClassLoader();
+            if (systemClassLoader != null) {
+                // TODO: some logging
+                classNames = classNamesNamedIn(systemClassLoader, fileName);
+            } else {
+                throw new RuntimeException("system class loader is null");
+            }
+        }
+
+        for (String className : classNames) {
             list.add(Class.forName(className, true, classLoader));
         }
         return Collections.unmodifiableList(list);
```

now command:

```
mvavrik@fedora:~/sources/keycloak/testsuite/integration-arquillian/tests/base$ mvn clean test -Dtest=DeployedSAMLScriptMapperTest -Pauth-server-quarkus -Dauth.server.feature=http-optimized-serializers
```

passes. If something of that sort was used, I wonder if it would be smart to only limit it to the test mode.

shawkins:
@michalvavrik thanks for the input. After looking at it more, it's indicative of a test setup issue. CLITestExtension.reset will delete provider jars in between test runs. If it's missing at the time we run this check, it causes the new logic to throw an exception. We have separate logic dealing with detecting changes in provider jars, so it's fine to just not react to this exception.

radcortez:
> > Just running the readBuildTimeConfiguration logic is causing the tests to fail. Should I be using a different classloader? @michalvavrik @radcortez
> 
> I think TCCL is the right choice, but test class loading is tricky. I have experience a month ago that I could not load a resource from dev svc startable unless I save TCCL from a build step. So I don't say that TCCL should work, I don't know. We can discuss it on Quarkus Zulip.

When running in Quarkus, the TCCL should be the Quarkus CL, which has access to the mapping classes. The Quarks CL is set as early as possible when running the tests.

shawkins:
@vmuzikar @mabartos this seems to be roughly what we need for detecting what raw quarkus options are build-time.

Here's roughly our current behavior with quarkus properties:
- if we see that the value has changed in the quarkus.properties file, then we assume a build is needed
- quarkus properties in the env are effectively ignored for purposes of our build detection logic - however any relevant values will still get used at build-time.

What may be not good about this change:
- the overhead of this logic (I haven't profiled it yet).
- it may not appropriately detect all build-time options (see some of the discussion above) - but this is already a pretty grey area given they are raw quarkus options. We could potentially add an easy mechanism to "build and run" as mentioned on https://github.com/keycloak/keycloak/issues/45444#issuecomment-3854908143

Or we could wait until we progress on https://github.com/keycloak/keycloak/issues/41465 to consider these changes.

PR #46036, opened 2026-02-04
Update Operator instructions for Custom Access

### LINKED ISSUE #45992: Clarify operator instructions involving Wildcard certificates and OpenShift
### Description

Operator on OpenShift is encountering issues when customers use Wildcard certificates, due to HTTP/2 connection reuse.

### Value Proposition

 Instructions need some updates.

### Goals

No more errors.

### Non-Goals

Complete revision of the Operator Guide.

### Discussion

_No response_

### Notes

_No response_

andymunro:
Closes #45992

### DIFF
--- a/docs/guides/operator/basic-deployment.adoc
+++ b/docs/guides/operator/basic-deployment.adoc
@@ -125,6 +125,11 @@ kubectl create secret tls example-tls-secret --cert certificate.pem --key key.pe
 
 To deploy {project_name}, you create a Custom Resource (CR) based on the Keycloak Custom Resource Definition (CRD).
 
+[NOTE]
+====
+If you are on OpenShift, you cannot use wildcard certificates on passthrough Routes with HTTP/2 enabled. The following Keycloak CR on OpenShift if you are using a wildcard certificate with the default IngressClass, creates such a Route. In this situation, modify the yaml based on "Accessing the {project_name} deployment" later in this topic to use TLS Termination or an OpenShift re-encrypt Route.
+====
+
 Consider storing the Database credentials in a separate Secret. Enter the following commands:
 [source,bash]
 ----


andymunro:
@shawkins Can you check this?

docs/guides/operator/basic-deployment.adoc
```diff
@@ -253,6 +258,11 @@ spec:
     headers: xforwarded # double check your reverse proxy sets and overwrites the X-Forwarded-* headers
 ----
 
+[NOTE]
```
shawkins: We don't need a note as the preceeding yaml shows using the Keycloak CR for tls termination.

docs/guides/operator/basic-deployment.adoc
```diff
@@ -125,6 +125,11 @@ kubectl create secret tls example-tls-secret --cert certificate.pem --key key.pe
 
 To deploy {project_name}, you create a Custom Resource (CR) based on the Keycloak Custom Resource Definition (CRD).
 
+[NOTE]
+====
+If you are on OpenShift, you cannot use wildcard certificates on passthrough Routes with HTTP/2 enabled. A Keycloak CR on OpenShift with TLS enabled using a wildcard certificate with the default IngressClass creates such a Route. In this case, see "Accessing the {project_name} deployment" later in this topic.
```
shawkins:
```suggestion
If you are on OpenShift, you cannot use wildcard certificates on passthrough Routes with HTTP/2 enabled. The following Keycloak CR on OpenShift using a wildcard certificate with the default IngressClass creates such a Route. In this case modify the yaml based upon "Accessing the {project_name} deployment" later in this topic to utilize TLS Termination or an OpenShift re-encrypt route.
```

andymunro:
Thanks, @shawkins. Updated.

docs/guides/operator/basic-deployment.adoc
```diff
@@ -125,6 +125,11 @@ kubectl create secret tls example-tls-secret --cert certificate.pem --key key.pe
 
 To deploy {project_name}, you create a Custom Resource (CR) based on the Keycloak Custom Resource Definition (CRD).
 
+[NOTE]
+====
+If you are on OpenShift, you cannot use wildcard certificates on passthrough Routes with HTTP/2 enabled. The following Keycloak CR on OpenShift, which uses a wildcard certificate with the default IngressClass, creates such a Route. In this situation, modify the yaml based on "Accessing the {project_name} deployment" later in this topic to use TLS Termination or an OpenShift re-encrypt route.
```
shawkins:
```suggestion
If you are on OpenShift, you cannot use wildcard certificates on passthrough Routes with HTTP/2 enabled. The following Keycloak CR on OpenShift if you are using a wildcard certificate with the default IngressClass, creates such a Route. In this situation, modify the yaml based on "Accessing the {project_name} deployment" later in this topic to use TLS Termination or an OpenShift re-encrypt Route.
```

andymunro:
Thanks, @shawkins. updated. @ahus1 can you check this?

shawkins (approved):

docs/guides/operator/basic-deployment.adoc
```diff
@@ -125,6 +125,11 @@ kubectl create secret tls example-tls-secret --cert certificate.pem --key key.pe
 
 To deploy {project_name}, you create a Custom Resource (CR) based on the Keycloak Custom Resource Definition (CRD).
 
+[NOTE]
+====
+If you are on OpenShift, you cannot use wildcard certificates on passthrough Routes with HTTP/2 enabled. The following Keycloak CR on OpenShift if you are using a wildcard certificate with the default IngressClass, creates such a Route. In this situation, modify the yaml based on "Accessing the {project_name} deployment" later in this topic to use TLS Termination or an OpenShift re-encrypt Route.
```
vmuzikar, line 130: The change itself look good but what I'm lacking a bit is some explanation, a background why wildcard certs are a problem with HTTP/2. Even I can remember what was the reason. :D @shawkins Do you?

shawkins: It's a known issue with how connections are reused at the proxy. Once the underlying http/2 connection is made to keycloak, the proxy will use that even if a different hostname wildcard is used in the request.

vmuzikar: Should we mention this in that note?

shawkins: You mean reference some specific KB and or Keycloak issue, or try to describe in a sentence or two why it won't work? I'm not sure that's needed. This also isn't the only call out on this topic - it's mentioned below here again.

In any case I don't see a concise reference on this from the openshift / redhat side of things. 

This writeup matches the scenario as well https://jammy.id/integration/mq-web-console-http2.html

andymunro: There was an issue in downstream as follows:

The Red Hat build of Keycloak Operator on OpenShift is encountering issues when customers use Wildcard certificates, due to HTTP/2 connection reuse. This is causing support cases and customer environment issues. To address this, we are evaluating a better approach for the Keycloak documentation, aiming to make it simpler and similar to the previous RH-SSO deployment, using the EDGE route as the default. The goal is to resolve these issues and improve customer experience.

vmuzikar: Ok, let's keep as is, it was not a blocking concern from me. :)

Issue #43971, opened 2025-11-05
Identity Assertion JWT Authorization Grant


stianst:
### Description

[Identity Assertion JWT Authorization Grant](https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant/) builds on Identity Chaining Across Trust Domains to cover scenarios within a single enterprise.

An example use-case enabled by ID-JAG is a scenario where an enterprise uses a SaaS service configured with the corporate IdP for single sign-on using, and at the same time allowing direct API access to a resource server by allowing an ID-JAG to be used to obtain an access token from the SaaS service authorization server.

### Value Proposition

Enables SaaS services to offer better integration with the enterprise IdP. ID-JAG will also be included in the next MCP specification.

### Goals

* Add support for ID-JAG to Keycloak

### Non-Goals

*

tnorimat:
@stianst Thank you for creating the issue. I would like to write its details afterwards.

tnorimat:
[Identity Assertion JWT  Authorization Grant](https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant/) leverages RFC 8693 Token Exchange and RFC 7523 JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants and is discussed in the MCP community [SEP-646: Enterprise-Managed Authorization Profile for MCP](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/646). The spec is for [enterprise-ready MCP](https://github.com/modelcontextprotocol/ext-auth/pull/4).

tnorimat:
@stianst Hello, my colleague is now working for the issue. The PR will be submitted afterwards.

tnorimat:
@stianst Hello,

Regarding ID-JAG, there are two roles: consumer and issuer.
The existing issues/PRs are about keycloak as a consumer of ID-JAG.
Therefore, how about adding the goal to the issue as follows?
- Keycloak works as a consumer of ID-JAG

Regarding ID -JAG as an issuer, my colleague created [the issue for ID-JAG as an issuer](https://github.com/keycloak/keycloak/issues/48818) .

PR #46048, opened 2026-02-05
Customization for JWTAuthorizationGrantType  to support ID-JAG

### LINKED ISSUE #49005: ID-JAG: initial support
### Description

The scope of this task is to provide initial support for ID-JAG .

bucchi:
closes #49005 

This PR is for ID-JAG receiver, which receives ID-JAG and send back access token working as a part of token endpoint.

ID-JAG is one type of JWT, which can be used as Authorization Grant(https://www.rfc-editor.org/rfc/rfc7523.html#section-2.1).
AFAIK, the type for JWTAuthorizationGrant clearly mentioned in the standard specs or the draft of those  is only one, for now.
But in the future there could be more types coming.
So here in this PR, we make validators for every types of JWT pluggable  by introducing brand new SPI:JWTAuthorizationGrantValidatorSpi.

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -169,6 +169,8 @@ public class Profile {
 
         CIMD("OAuth Client ID Metadata Document", Type.EXPERIMENTAL),
 
+        IDENTITY_ASSERTION_JWT("Identity Assertion JWT", Type.EXPERIMENTAL),
+
         /**
          * @see <a href="https://github.com/keycloak/keycloak/issues/37967">Deprecate for removal the Instagram social broker</a>.
          */
--- a/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java
@@ -27,9 +27,12 @@ import org.keycloak.authentication.authenticators.client.ClientAssertionState;
 import org.keycloak.broker.provider.BrokeredIdentityContext;
 import org.keycloak.broker.provider.JWTAuthorizationGrantProvider;
 import org.keycloak.cache.AlternativeLookupProvider;
+import org.keycloak.common.Profile;
 import org.keycloak.events.Details;
 import org.keycloak.events.Errors;
 import org.keycloak.events.EventType;
+import org.keycloak.jose.jws.JWSInput;
+import org.keycloak.jose.jws.JWSInputException;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.ClientSessionContext;
 import org.keycloak.models.FederatedIdentityModel;
@@ -40,6 +43,7 @@ import org.keycloak.models.UserSessionModel;
 import org.keycloak.protocol.oidc.OIDCAdvancedConfigWrapper;
 import org.keycloak.protocol.oidc.OIDCLoginProtocol;
 import org.keycloak.protocol.oidc.TokenManager;
+import org.keycloak.representations.JsonWebToken;
 import org.keycloak.services.CorsErrorResponseException;
 import org.keycloak.services.Urls;
 import org.keycloak.services.clientpolicy.ClientPolicyException;
@@ -60,8 +64,33 @@ public class JWTAuthorizationGrantType extends OAuth2GrantTypeBase {
 
         try {
 
-            JWTAuthorizationGrantValidator authorizationGrantContext = JWTAuthorizationGrantValidator.createValidator(
-                    context.getSession(), client, assertion, formParams.getFirst(OAuth2Constants.SCOPE));
+            if (assertion == null) {
+                throw new IllegalArgumentException("Missing parameter:" + OAuth2Constants.ASSERTION);
+            }
+
+            JWSInput jws;
+            JsonWebToken jwt;
+            try {
+                jws = new JWSInput(assertion);
+                jwt = jws.readJsonContent(JsonWebToken.class);
+            } catch (JWSInputException e) {
+                throw new RuntimeException("The provided assertion is not a valid JWT");
+            }
+
+            String jwtTokenType = jws.getHeader().getType();
+            
+            ClientAssertionState clientAssertionState = new ClientAssertionState(OAuth2Constants.JWT_AUTHORIZATION_GRANT, assertion, jws, jwt);
+            clientAssertionState.setClient(context.getClient());
+
+            JWTAuthorizationGrantValidator authorizationGrantContext;
+            if (Profile.isFeatureEnabled(Profile.Feature.IDENTITY_ASSERTION_JWT) && jwtTokenType != null 
+                    && jwtTokenType.equals(OAuth2Constants.IDENTITY_ASSERTION_JWT_HEADER_TYPE)) {
+                authorizationGrantContext = IDJWTAuthorizationGrantValidator.createValidator(
+                    context.getSession(), formParams.getFirst(OAuth2Constants.SCOPE), clientAssertionState);
+            } else {
+                authorizationGrantContext = DefaultJWTAuthorizationGrantValidator.createValidator(
+                    context.getSession(), formParams.getFirst(OAuth2Constants.SCOPE), clientAssertionState);
+            }
             event.detail(Details.IDENTITY_PROVIDER_ISSUER, authorizationGrantContext.getIssuer());
             event.detail(Details.IDENTITY_PROVIDER_USER_ID, authorizationGrantContext.getSubject());
 
@@ -109,7 +138,7 @@ public class JWTAuthorizationGrantType extends OAuth2GrantTypeBase {
 
             //user must exist in keycloak
             FederatedIdentityModel federatedIdentityModel = new FederatedIdentityModel(identityProviderModel.getAlias(), brokeredIdentityContext.getId(), brokeredIdentityContext.getUsername(), brokeredIdentityContext.getToken());
-            UserModel user = lookupUserByFederatedIdentity(federatedIdentityModel, authorizationGrantContext.getState());
+            UserModel user = lookupUserByFederatedIdentity(federatedIdentityModel, clientAssertionState);
             if (user == null) {
                 throw new RuntimeException("User not found");
             }
--- a/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantValidator.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantValidator.java
@@ -1,119 +1,28 @@
-/*
- * Copyright 2025 Red Hat, Inc. and/or its affiliates
- *  and other contributors as indicated by the @author tags.
- *
- *  Licensed under the Apache License, Version 2.0 (the "License");
- *  you may not use this file except in compliance with the License.
- *  You may obtain a copy of the License at
- *
- *  http://www.apache.org/licenses/LICENSE-2.0
- *
- *  Unless required by applicable law or agreed to in writing, software
- *  distributed under the License is distributed on an "AS IS" BASIS,
- *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- *
- *  See the License for the specific language governing permissions and
- *  limitations under the License.
- *
- */
-
 package org.keycloak.protocol.oidc.grants;
 
-import java.util.Set;
+import java.util.List;
 
-import org.keycloak.OAuth2Constants;
-import org.keycloak.authentication.authenticators.client.AbstractBaseJWTValidator;
-import org.keycloak.authentication.authenticators.client.ClientAssertionState;
-import org.keycloak.jose.jws.JWSInput;
-import org.keycloak.jose.jws.JWSInputException;
-import org.keycloak.models.ClientModel;
-import org.keycloak.models.KeycloakSession;
-import org.keycloak.protocol.oidc.JWTAuthorizationGrantValidationContext;
-import org.keycloak.protocol.oidc.OIDCConfigAttributes;
-import org.keycloak.representations.IDToken;
-import org.keycloak.representations.JsonWebToken;
+import org.keycloak.protocol.oidc.JWTAuthorizationGrantValidationContext ;
+
 
 /**
- * Validator for JWT Authorization grant that extends AbstractBaseJWTValidator and
- * implements the JWTAuthorizationGrantValidationContext interface.
+ * Interface for the assertion validator of JWTAuthorizationGrant
  *
- * @author rmartinc
+ * @author <a href="mailto:yutaka.obuchi.sd@hitachi.com">Yutaka Obuchi</a>
  */
-public class JWTAuthorizationGrantValidator extends AbstractBaseJWTValidator implements JWTAuthorizationGrantValidationContext {
 
-    private final String scope;
-    private Set<String> restrictedScopes;
+public interface JWTAuthorizationGrantValidator extends JWTAuthorizationGrantValidationContext {
 
-    public static JWTAuthorizationGrantValidator createValidator(KeycloakSession session, ClientModel client, String assertion, String scope) {
-        if (assertion == null) {
-            throw new RuntimeException("Missing parameter:" + OAuth2Constants.ASSERTION);
-        }
-        try {
-            JWSInput jws = new JWSInput(assertion);
-            JsonWebToken jwt = jws.readJsonContent(JsonWebToken.class);
-            ClientAssertionState clientAssertionState = new ClientAssertionState(OAuth2Constants.JWT_AUTHORIZATION_GRANT, assertion, jws, jwt);
-            clientAssertionState.setClient(client);
-            return new JWTAuthorizationGrantValidator(session, scope, clientAssertionState);
-        } catch (JWSInputException e) {
-            throw new RuntimeException("The provided assertion is not a valid JWT");
-        }
-    }
-
-    private JWTAuthorizationGrantValidator(KeycloakSession session, String scope, ClientAssertionState clientAssertionState) {
-        super(session, clientAssertionState);
-        this.scope = scope;
-    }
-
-    public void validateClient() {
-        if (clientAssertionState.getClient().isPublicClient()) {
-            failureCallback("Public client not allowed to use authorization grant");
-        }
-
-        String val = clientAssertionState.getClient().getAttribute(OIDCConfigAttributes.JWT_AUTHORIZATION_GRANT_ENABLED);
-        if (!Boolean.parseBoolean(val)) {
-            throw new RuntimeException("JWT Authorization Grant is not supported for the requested client");
-        }
-    }
-
-    public void validateIssuer() {
-        if (getJWT().getIssuer() == null) {
-            failureCallback("Missing claim: " + OAuth2Constants.ISSUER);
-        }
-    }
-
-    public void validateSubject() {
-        if (getJWT().getSubject() == null) {
-            failureCallback("Missing claim: " + IDToken.SUBJECT);
-        }
-    }
-
-    @Override
-    public JsonWebToken getJWT() {
-        return clientAssertionState.getToken();
-    }
-
-    @Override
-    public String getAssertion() {
-        return clientAssertionState.getClientAssertion();
-    }
-
-    @Override
-    public String getScopeParam() {
-        return scope;
-    }
-
-    @Override
-    public Set<String> getRestrictedScopes() {
-        return restrictedScopes;
-    }
-
-    @Override
-    public void setRestrictedScopes(Set<String> restrictedScopes) {
-        this.restrictedScopes = restrictedScopes;
-    }
-
-    @Override
-    protected void failureCallback(String errorDescription) {
-        throw new RuntimeException(errorDescription);
-    }
+    void validateClient();
+    
+    void validateIssuer();
+    
+    void validateSubject();
+    
+    boolean validateTokenActive(int allowedClockSkew, int maxExp, boolean reusePermitted); 
+    
+    boolean validateSignatureAlgorithm(String expectedSignatureAlg);
+    
+    boolean validateTokenAudience(List<String> expectedAudiences, boolean multipleAudienceAllowed);
+    
 }
--- a/test-framework/oauth/src/main/java/org/keycloak/testframework/oauth/OAuthIdentityProvider.java
+++ b/test-framework/oauth/src/main/java/org/keycloak/testframework/oauth/OAuthIdentityProvider.java
@@ -7,6 +7,7 @@ import java.security.KeyPair;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.keycloak.OAuth2Constants;
 import org.keycloak.common.crypto.CryptoIntegration;
 import org.keycloak.common.util.KeyUtils;
 import org.keycloak.crypto.Algorithm;
@@ -58,6 +59,10 @@ public class OAuthIdentityProvider {
 
     public String encodeToken(JsonWebToken token, OAuthIdentityProviderKeys keys) {
         return new JWSBuilder().type("JWT").jsonContent(token).sign(new ECDSASignatureSignerContext(keys.getKeyWrapper()));
+    }
+
+    public String encodeIDJAG(JsonWebToken token) {
+        return new JWSBuilder().type(OAuth2Constants.IDENTITY_ASSERTION_JWT_HEADER_TYPE).jsonContent(token).sign(new ECDSASignatureSignerContext(keys.getKeyWrapper()));
     }
 
     public OAuthIdentityProviderKeys createKeys() {
--- a/tests/base/src/test/java/org/keycloak/tests/oauth/IDJWTAuthorizationGrantTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/oauth/IDJWTAuthorizationGrantTest.java
@@ -1,0 +1,220 @@
+package org.keycloak.tests.oauth;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.broker.oidc.OIDCIdentityProviderConfig;
+import org.keycloak.broker.oidc.OIDCIdentityProviderFactory;
+import org.keycloak.common.Profile;
+import org.keycloak.common.util.Time;
+import org.keycloak.events.Details;
+import org.keycloak.events.EventType;
+import org.keycloak.models.IdentityProviderModel;
+import org.keycloak.protocol.oidc.OIDCConfigAttributes;
+import org.keycloak.representations.AccessToken;
+import org.keycloak.testframework.annotations.InjectRealm;
+import org.keycloak.testframework.annotations.InjectUser;
+import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
+import org.keycloak.testframework.events.EventAssertion;
+import org.keycloak.testframework.realm.ClientBuilder;
+import org.keycloak.testframework.realm.IdentityProviderBuilder;
+import org.keycloak.testframework.realm.ManagedRealm;
+import org.keycloak.testframework.realm.ManagedUser;
+import org.keycloak.testframework.realm.RealmBuilder;
+import org.keycloak.testframework.realm.UserBuilder;
+import org.keycloak.testframework.server.KeycloakServerConfig;
+import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
+import org.keycloak.testsuite.util.oauth.AccessTokenResponse;
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+@KeycloakIntegrationTest(config = IDJWTAuthorizationGrantTest.JWTAuthorizationGrantServerConfig.class)
+@TestMethodOrder(MethodOrderer.MethodName.class)
+public class IDJWTAuthorizationGrantTest extends BaseAbstractJWTAuthorizationGrantTest {
+
+    @InjectRealm(config = IDJWTAuthorizationGrantTest.JWTAuthorizationGrantRealmConfig.class)
+    protected ManagedRealm realm;
+
+    @InjectUser(config = IDJWTAuthorizationGrantTest.FederatedUserConfiguration.class)
+    protected ManagedUser user;
+  
+    @Test
+    public void testNullIssInToken() {
+        AccessToken jwt = createAuthorizationGrantToken("basic-user-id",oAuthClient.getEndpoints().getIssuer(),IDP_ISSUER,Time.currentTime() + 300L,(long) Time.currentTime(), null);
+        jwt.setOtherClaims("client_id","clientid_issued_by_resource_authz_to_client_app");
+        jwt.issuer(null);
+        String jwtString = getIdentityProvider().encodeIDJAG(jwt);
+        AccessTokenResponse response = oAuthClient.client("clientid_issued_by_resource_authz_to_client_app", "test-secret").jwtAuthorizationGrantRequest(jwtString).send();
+        assertFailure("Missing claim: iss", response, events.poll());
+    }
+
+    @Test
+    public void testNullAudInToken() {
+        AccessToken jwt = createAuthorizationGrantToken("basic-user-id",oAuthClient.getEndpoints().getIssuer(),IDP_ISSUER,Time.currentTime() + 300L,(long) Time.currentTime(), null);
+        jwt.setOtherClaims("client_id","clientid_issued_by_resource_authz_to_client_app");
+        jwt.audience(null);
+        String jwtString = getIdentityProvider().encodeIDJAG(jwt);
+        AccessTokenResponse response = oAuthClient.client("clientid_issued_by_resource_authz_to_client_app", "test-secret").jwtAuthorizationGrantRequest(jwtString).send();
+        assertFailure("Invalid token audience", response, events.poll());
+    }
+
+    @Test
+    public void testNullSubInToken() {
+        AccessToken jwt = createAuthorizationGrantToken("basic-user-id",oAuthClient.getEndpoints().getIssuer(),IDP_ISSUER,Time.currentTime() + 300L,(long) Time.currentTime(), null);
+        jwt.setOtherClaims("client_id","clientid_issued_by_resource_authz_to_client_app");
+        jwt.subject(null);
+        String jwtString = getIdentityProvider().encodeIDJAG(jwt);
+        AccessTokenResponse response = oAuthClient.client("clientid_issued_by_resource_authz_to_client_app", "test-secret").jwtAuthorizationGrantRequest(jwtString).send();
+        assertFailure("Missing claim: sub", response, events.poll());
+    }
+    
+    @Test
+    public void testNullExpInToken() {
+        AccessToken jwt = createAuthorizationGrantToken("basic-user-id",oAuthClient.getEndpoints().getIssuer(),IDP_ISSUER,Time.currentTime() + 300L,(long) Time.currentTime(), null);
+        jwt.setOtherClaims("client_id","clientid_issued_by_resource_authz_to_client_app");
+        jwt.exp(null);
+        String jwtString = getIdentityProvider().encodeIDJAG(jwt);
+        AccessTokenResponse response = oAuthClient.client("clientid_issued_by_resource_authz_to_client_app", "test-secret").jwtAuthorizationGrantRequest(jwtString).send();
+        assertFailure("Token exp claim is required", response, events.poll());
+    }
+
+    @Test
+    public void testNullIatInToken() {
+        AccessToken jwt = createAuthorizationGrantToken("basic-user-id",oAuthClient.getEndpoints().getIssuer(),IDP_ISSUER,Time.currentTime() + 300L,(long) Time.currentTime(), null);
+        jwt.setOtherClaims("client_id","clientid_issued_by_resource_authz_to_client_app");
+        jwt.iat(null);
+        String jwtString = getIdentityProvider().encodeIDJAG(jwt);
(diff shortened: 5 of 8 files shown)


tnorimat:
@bucchi Hello, could you run `./mvnw spotless:apply` ? It automatically modifies codes to align with Keycloak's code style.
It can resolve the PR's failure https://github.com/keycloak/keycloak/actions/runs/21708246440/job/62604873514?pr=46048

services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantValidator.java
tnorimat, line 1: It is better to add javadoc comment to interface class for a reviewer who is not familiar with ID-JAG to understand them.

VinodAnandan:
Hi @bucchi  and @tnorimat, Thank you very much for ths PR, I really appreciate your work on this. I wanted to kindly check whether the PR is currently stalled, and if there is a tentative timeline your team has in mind for it. Are you planning for this PR to be included in the Keycloak 26.6 release? 

If you haven’t had a chance to watch it yet, I would kindly recommend taking a look at the following video, which demonstrates how Okta is leveraging ID-JAG for MCP use cases - https://www.brighttalk.com/webcast/14899/661521

thomasdarimont:
Would it be possible to expose the initial assertion (read-only) to subsequently executed protocol mappers?

If the initial assertion contained additional claims that should be carried over the the new token, we would need access to the original claims from the initial assertion.
I encountered this exact use case in two independent projects this week that looked into using the JWT bearer-based grant.

services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantValidator.java
bucchi: @tnorimat  
Thank you for your comment.
I have just added some comments.

tests/base/src/test/java/org/keycloak/tests/oauth/IDJWTAuthorizationGrantTest.java
```diff
@@ -0,0 +1,139 @@
...
+public class IDJWTAuthorizationGrantTest extends BaseAbstractJWTAuthorizationGrantTest {
+
+    @InjectRealm(config = IDJWTAuthorizationGrantTest.JWTAuthorizationGrantRealmConfig.class)
+    protected ManagedRealm realm;
+
+    @InjectUser(config = IDJWTAuthorizationGrantTest.FederatedUserConfiguration.class)
+    protected ManagedUser user;
+
+    @Test
+    public void testClientInTokenAndInRequestPrameter() {
```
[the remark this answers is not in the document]
bucchi: I have just fixed this as suggested.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final Logger logger = Logger.getLogger(IDJWTAuthorizationGrantValidator.class);
+
+    public IDJWTAuthorizationGrantValidator(KeycloakSession session) {
+        super(session);
+    }
+
+    public void validateClient() {
+        super.validateClient();
+
+        try{
```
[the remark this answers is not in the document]
bucchi: I have just fixed this as suggested.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,35 @@
...
+            AccessToken accessToken = new JWSInput(getAssertion()).readJsonContent(AccessToken.class);
```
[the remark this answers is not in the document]
bucchi: I have just fixed this as suggested.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,35 @@
...
+            String clientIdInToken = (String) accessToken.getOtherClaims().get("client_id");
+            if (!session.getContext().getClient().getClientId().equals(clientIdInToken)){
```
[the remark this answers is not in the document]
bucchi: I have just added explicit null check.

And also I have just added a test case for Identity Assertion JWT with no client_id included.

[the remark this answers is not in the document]
bucchi: > The code should be formatted as "if (" instead of "if (".
From this, I can not find out what is the problem here.

But at the end of "if condition statement", there is a space missing at "oken)){".
So, I have just changed this to "oken)) {" for here.

tests/base/src/test/java/org/keycloak/tests/oauth/IDJWTAuthorizationGrantTest.java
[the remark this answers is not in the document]
tnorimat: Thank you.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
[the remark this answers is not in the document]
tnorimat: Thank you.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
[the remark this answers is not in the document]
tnorimat: Thank you.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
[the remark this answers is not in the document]
tnorimat: Thank you.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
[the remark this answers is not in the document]
tnorimat: Thank you.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/DefaultJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,34 @@
...
+import org.keycloak.jose.jws.JWSInputException;
+import org.keycloak.models.ClientModel;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.representations.JsonWebToken;
+
+public class DefaultJWTAuthorizationGrantValidator extends JWTAuthorizationGrantValidatorBase {
+
+    public static DefaultJWTAuthorizationGrantValidator createValidator(KeycloakSession session, ClientModel client, String assertion, String scope) {
+        if (assertion == null) {
+            throw new RuntimeException("Missing parameter:" + OAuth2Constants.ASSERTION);
```
thomasdarimont:
```suggestion
            throw new IllegalArgumentException("Missing parameter:" + OAuth2Constants.ASSERTION);
```

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public void validateClient() {
+        super.validateClient();
+
+        try {
+            JWSInput jws = clientAssertionState.getJws();
+            AccessToken accessToken = jws.readJsonContent(AccessToken.class);
+            String clientIdInToken = (String) accessToken.getOtherClaims().get("client_id");
+            if (clientIdInToken == null || !session.getContext().getClient().getClientId().equals(clientIdInToken)) {
```
thomasdarimont: You could extract the client into a local var, this eases debugging.

test-framework/oauth/src/main/java/org/keycloak/testframework/oauth/OAuthIdentityProvider.java
```diff
@@ -57,6 +57,10 @@ public String encodeToken(JsonWebToken token, OAuthIdentityProviderKeys keys) {
         return new JWSBuilder().type("JWT").jsonContent(token).sign(new ECDSASignatureSignerContext(keys.getKeyWrapper()));
     }
 
+    public String encodeIDJAG(JsonWebToken token) {
+        return new JWSBuilder().type("oauth-id-jag+jwt").jsonContent(token).sign(new ECDSASignatureSignerContext(keys.getKeyWrapper()));
```
thomasdarimont: Some custom jwt types are defined in org.keycloak.util.TokenUtil and org.keycloak.OAuth2Constants. I suggest to create a constant for "oauth-id-jag+jwt" in OAuth2Constants.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
```diff
@@ -56,8 +64,38 @@ public Response process(Context context) {
 
         try {
 
-            JWTAuthorizationGrantValidator authorizationGrantContext = JWTAuthorizationGrantValidator.createValidator(
+            if (assertion == null) {
+                throw new RuntimeException("Missing parameter:" + OAuth2Constants.ASSERTION);
```
thomasdarimont:
```suggestion
                throw new IllegalArgumentException("Missing parameter:" + OAuth2Constants.ASSERTION);
```

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
```diff
@@ -56,8 +64,38 @@ public Response process(Context context) {
...
+                jws = new JWSInput(assertion);
+            } catch (JWSInputException e) {
+                throw new RuntimeException("The provided assertion is not a valid JWT");
+            }
+
+            String jwtTokenType = jws.getHeader().getType();
+            JWTAuthorizationGrantValidatorBase authorizationGrantContext = 
+                            (JWTAuthorizationGrantValidatorBase) session.getProvider(JWTAuthorizationGrantValidator.class, jwtTokenType);            
+            if( authorizationGrantContext == null){
+                authorizationGrantContext = DefaultJWTAuthorizationGrantValidator.createValidator(
```
thomasdarimont: I sugget to create a protected method like createValidator(...) that uses DefaultJWTAuthorizationGrantValidator.createValidator(..) by default.

This would enable to customize the validation behavior if necessary.

thomasdarimont (changes requested):
Looks good so far. I added a few minor remarks.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/DefaultJWTAuthorizationGrantValidator.java
bucchi: Thank you for your commnet.

I have just fixed this as you suggested.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
bucchi: I have just fixed this as you suggested.

test-framework/oauth/src/main/java/org/keycloak/testframework/oauth/OAuthIdentityProvider.java
bucchi: I have just fixed this as you suggested.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: I have just fixed this as you suggested,
and also added some test cases for this.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: I have just fixed this as you suggested.

services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantValidator.java
tnorimat: Thank you.

VinodAnandan:
@bucchi The PR branch currently has some conflicts that need to be resolved. When you have a moment, could you please help address them?

common/src/main/java/org/keycloak/common/Profile.java
```diff
@@ -162,6 +162,8 @@ public enum Feature {
 
         CIMD("OAuth Client ID Metadata Document", Type.EXPERIMENTAL),
 
+        IDENTITY_ASSERTION_JWT_VALIDATOR("Identity Assertion JWT validator", Type.EXPERIMENTAL),
```
graziang: I was thinking if it makes sense to have a dedicated feature for the entire specification instead of only for validating the ID JAG.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    public IDJWTAuthorizationGrantValidator(KeycloakSession session) {
+        super(session);
+    }
+
+    public void validateClient() {
+        super.validateClient();
+
+        try {
+            JWSInput jws = clientAssertionState.getJws();
+            AccessToken accessToken = jws.readJsonContent(AccessToken.class);
```
graziang: It should already be parsed in the client assertion state: `clientAssertionState.getToken()`

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * a JWT validator for Identity Assertion JWT Authorization Grant (ID-JAG).
+ * Identity Assertion JWT is a new type of JWT that can be used as an authorization grant per RFC 7523.
+ *  
+ * https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant/
+ *
+ * @author <a href="mailto:yutaka.obuchi.sd@hitachi.com">Yutaka Obuchi</a>
+ */
+public class IDJWTAuthorizationGrantValidator extends JWTAuthorizationGrantValidatorBase {
```
graziang: Some checks are missing. For example, `jti` and `iat` are REQUIRED according to the specification: https://www.ietf.org/archive/id/draft-ietf-oauth-identity-assertion-authz-grant-01.html#section-3.”

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
```diff
@@ -56,8 +65,38 @@ public Response process(Context context) {
...
+                } catch (JWSInputException e) {
+                throw new RuntimeException("The provided assertion is not a valid JWT");
+                }
+
+                ClientAssertionState clientAssertionState = new ClientAssertionState(OAuth2Constants.JWT_AUTHORIZATION_GRANT, assertion, jws, jwt);
+                clientAssertionState.setClient(client);
+
+                authorizationGrantContext.setClientAssertionState(clientAssertionState);
+                authorizationGrantContext.setScopeParam(formParams.getFirst(OAuth2Constants.SCOPE));
+            }
```
graziang: Only `IDJWTAuthorizationGrantValidator` is obtained via `getProvider`, for normal jwt it is created manually. There are also some duplications in the parsing of the assertion.

graziang (commented):
@bucchi Thanks for this PR, it’s nice to see the work to support this new specification. A few considerations:

The Identity Assertion JWT Authorization Grant (ID-JAG) is defined in the draft specification https://www.ietf.org/archive/id/draft-ietf-oauth-identity-assertion-authz-grant-01.html as a profile of the JWT Authorization Grant (RFC7523). 

This specification is also based on another draft specification, Identity Chaining Across Trust Domains https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-chaining/, which combines Token Exchange (RFC8693) and the JWT Authorization Grant (RFC7523).


So this ID-JAG is obtained with a token exchange request and then used as an assertion in an access token request using the JWT Authorization Grant.

With this PR we are adding support for validating the Identity Assertion JWT Authorization Grant (ID-JAG) for JWT Authorization Grant requests.

Because of this, I was wondering if it would make sense to use a more general feature to support the whole specification, instead of having the feature `IDENTITY_ASSERTION_JWT_VALIDATOR` only for the ID-JAG validation.

I added some inline comments, but in general the introduction of a new SPI for the validator in this way IMO makes the classes a bit more complex, and it also introduces some duplication in the assertion parsing.

Additionally, we should consider the options configured on the identity provider. For example, `jti` is not mandatory for the JWT Authorization Grant, but it is required for ID-JAG. On the identity provider we have an option for replay checking of the assertion using the `jti` claim. This option should probably be hidden or disabled and enforced when ID-JAG is used.

VinodAnandan:
@bucchi @graziang I think it would be best to use the links below to access the latest versions of the drafts. This should help ensure everyone is reviewing the up-to-date documents and avoid any confusion caused by earlier versions.

https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant/

VinodAnandan:
@bucchi @tnorimat Just a kind reminder regarding the 26.6 release schedule, which is planned for the next few weeks. If it may be difficult for you to prioritise this PR within that time frame, would it be possible to consider granting the Keycloak core maintainer write access to the PR to adopt the changes they have suggested? This might help ensure the PR can be merged in time for the 26.6 release. Thank you very much for considering this. I really appreciate your help.

common/src/main/java/org/keycloak/common/Profile.java
bucchi: Thank you for your comment.
I made that IDENTITY_ASSERTION_JWT so that it could be used not only for validator part but also for issuer part.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
bucchi: I fixed as you suggested.
Thank you.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: I removed some duplications.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/validator/IDJWTAuthorizationGrantValidator.java
bucchi: I added some checks according to the specifications.

Here I also added some test cases for required fields.

bucchi:
> Additionally, we should consider the options configured on the identity provider. For example, `jti` is not mandatory for the JWT Authorization Grant, but it is required for ID-JAG. On the identity provider we have an option for replay checking of the assertion using the `jti` claim. This option should probably be hidden or disabled and enforced when ID-JAG is used.

@graziang, 
I agree with you here. 
For now, a current implementation simply ignores the option for jti on the identity provider.
Ideally, like you said here, this option should be hidden for ID-JAG.

From my understanding, it could be done by extending the identity provider of JWT Auhtorization Grant.
But maybe it should not be replacing, but it should be adding because, otherwise, the identity provider would be unable to process "normal" JWT Authorization Grants.
What do you think?

Btw, will you be in KeycloakCon EU next weak by any chance? 
I will be there and talk about this in the session(https://colocatedeventseu2026.sched.com/event/2DY2B/model-context-protocolmcp-authorization-for-enterprise-use-case-with-keycloak-yutaka-obuchi-hitachi-ltd).
We can talk about this implementation if you also will be there.

joeyorlando:
really looking forward to this PR landing in a (hopefully soon 🙂) release (in the mean-time for testing over in https://github.com/archestra-ai/archestra I've built an image off of your latest PR commit)

thomasdarimont (approved):
LGTM

server-spi-private/src/main/java/org/keycloak/protocol/oidc/JWTAuthorizationGrantValidator.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Provider interface to support pluggable validators of JWTAuthorizationGrant token.
+ * The pluggable validators are supposed to be provided for each JWT token type  
+ *
+ * @author <a href="mailto:yutaka.obuchi.sd@hitachi.com">Yutaka Obuchi</a>
+ */
+
+public interface JWTAuthorizationGrantValidator extends Provider, JWTAuthorizationGrantValidationContext{
+
+    public void validateClient();
```
thomasdarimont: You could remove the public modifier from those interface methods as there are public by default :)

bucchi: Thank you for your comment.

I fixed this.

bucchi:
Hi everyone and @mposolda, 
thank you for your helpful reviews.
I believe I have addressed all the comments and feedback provided.
Could you please take another look when you have a moment?
And let me know if I missed anything!

tnorimat:
@stianst @mposolda Hello,
Could you or your team review the PR?
Several people including your team has already reviewed the PR, and the issue (https://github.com/keycloak/keycloak/issues/43971) of the PR in on 26.7.0 milestone and many people over 10 are interested on the issue.

mposolda:
@tnorimat @bucchi Yes, my colleague @graziang will re-review the pull request.

FYI. I've linked this PR with the issue #49005 . As the scope of the issue #43971 is to deliver the whole feature ID-JAG and make sure it is properly tested, documented and supported.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
```diff
@@ -60,8 +68,37 @@ public Response process(Context context) {
...
+                jwtTokenType = OAuth2Constants.JWT;
+            }
+            
+            JWTAuthorizationGrantValidatorBase authorizationGrantContext = 
+                            (JWTAuthorizationGrantValidatorBase) session.getProvider(JWTAuthorizationGrantValidator.class, jwtTokenType);            
+            if( authorizationGrantContext == null){
+                authorizationGrantContext = createDefaultValidator(
+                    context.getSession(), formParams.getFirst(OAuth2Constants.SCOPE), clientAssertionState);
+            } else {
+                authorizationGrantContext.setClientAssertionState(clientAssertionState);
```
graziang: A provider created via SPI starts with clientAssertionState = null. Any method called before `setClientAssertionState()` can causes NPE, non-ideal use of the provider, It might be better to pass clientAssertionState as a parameter to validation methods instead, or some other refactor.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
```diff
@@ -60,8 +68,37 @@ public Response process(Context context) {
...
+
+            ClientAssertionState clientAssertionState = new ClientAssertionState(OAuth2Constants.JWT_AUTHORIZATION_GRANT, assertion, jws, jwt);   
+            clientAssertionState.setClient(client);
+            
+            String jwtTokenType = jws.getHeader().getType();
+            if (jwtTokenType == null) {
+                jwtTokenType = OAuth2Constants.JWT;
+            }
+            
+            JWTAuthorizationGrantValidatorBase authorizationGrantContext =
```
graziang: Consider removing `JWTAuthorizationGrantValidatorBase` it seems to exist only as an adapter between `AbstractBaseJWTValidator` and the SPI interfaces, you could extend `AbstractBaseJWTValidator` directly in `DefaultJWTAuthorizationGrantValidator` and implement the interface methods themselves, simplifying the
  hierarchy which is already a bit complex. Then `IDJWTAuthorizationGrantValidator` could directly extend `DefaultJWTAuthorizationGrantValidator`. And in this line you could get the provider directly as interface without casting to an implementation.

graziang (commented):
Hi @bucchi, thanks for the updates on the PR, and sorry for the delay in the review. I added a few more inline comments, but I think we’re getting close to being ready. Thanks again for this work!

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
mposolda: +1 that hierarchy is a bit complex. If there is possibility to simplify a bit as suggested above, it can be nice.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: @graziang 
Thank you for your comment.

As you can see, that is actually one of tricky part. And it might make my implementation a little bit ugly.
So it is OK to modify the source code as your suggestion.

From my understanding, your suggestion is that all methods in JWTAuthorizationGrantValidator interface below better be modified to have ClientAssertionState as parameter.

Do I understand your suggestion correctly?

```

    void validateClient();
    
    void validateIssuer();
    
    void validateSubject();
    
    boolean validateTokenActive(int allowedClockSkew, int maxExp, boolean reusePermitted); 
    
    boolean validateSignatureAlgorithm(String expectedSignatureAlg);
    
    boolean validateTokenAudience(List<String> expectedAudiences, boolean multipleAudienceAllowed);
```

bucchi: Reason here I am asking this is that for this modification, AbstractBaseJWTValidator, AbstractJWTClientValidator and the related test codes have to be modified too and the affected area will be larger.
I wanted to confirm whether it was intended for just in case.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: @graziang and @mposolda 

Thank you for your comment and suggestion.

I totally  agree with you for making  the hierarchy  more simple.
I have just tried removing JWTAuthorizationGrantValidatorBase in my local repository. 
It looks ok , but  for the part

> And in this line you could get the provider directly as interface without casting to an implementation.

we can make the line your mentioned like below
```
            JWTAuthorizationGrantValidator authorizationGrantContext = 
                             session.getProvider(JWTAuthorizationGrantValidator.class, jwtTokenType);
```
without casting.

But some other lines needs some casting like below:

`UserModel user = lookupUserByFederatedIdentity(federatedIdentityModel, ((DefaultJWTAuthorizationGrantValidator) authorizationGrantContext).getState());`
`((DefaultJWTAuthorizationGrantValidator) authorizationGrantContext).setScopeParam(formParams.getFirst(OAuth2Constants.SCOPE));`

As you can imagine, that is because those methods are originally from AbstractBaseJWTValidator, not in JWTAuthorizationGrantValidator interface.

Do I understand correctly?

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
graziang: Hi @bucchi, yeah, or create a `ClientAssertionState` for each method without touching the `AbstractBaseJWTValidator`, but the point here is that an implementation of an SPI should be usable via the methods defined by the SPI itself. If every SPI implementation requires different ways to configure or use it, and you need to cast to the concrete implementation in order to use it properly, there’s a design problem.

I know the issue is related to the relationship with `AbstractBaseJWTValidator`, but as things are getting complicated, maybe you could check if it would be simpler to just extend the original `JWTAuthorizationGrantValidator` as `IDJWTAuthorizationGrantValidator` and override only the methods needed to validate the ID-JAG and use it if needed in the JWTAuthorizationGrantType. WDYT?

bucchi: @graziang 
I have a couple of random ideas.

As you mentioned, I too think that the relationship with AbstractBaseJWTValidator makes this implementation complex and ugly. 
So how about not using AbstractBaseJWTValidator like below:
1. Keeping the original logic as it is in JWTAuhtorizationGrantType  and just adding some code to get provider from session for making it possible to plugin extra validator for each token type.
The SPI implementation here does not extend AbstractBaseJWTValidator. We better keep it as simple as possible.
This way, no matter what JWT token type comes in, original logic in JWTAuhtorizationGrantType is executed and also depending on the incoming token type, the corresponding extra validation is executed.

2. Using ClientPolicy with TokenType condition and IDJAGEnforcer executor for instance:
Like built-in GranTypeCondition.java, TokenTypeCondition.java is implemented for checking token type in JWT header.
Like  built-in JWTClaimEnforcerExecutor.java, IDJAGEnforcerExecutor is implemented for checking if the other claims comply with the IDJAG specification.


There are three alternatives including your idea so far. 
I have no particular preference among the these.

Please let me know your preference.
Thank you

graziang: @bucchi If I understood correctly the first proposal is to keep the current behavior and add extra validation based on the provider. The problem with this approach is that the validation is not necessarily more restrictive than the default one, so it might be necessary to override the validation of some claims. I’m not against keeping the provider, but we need to find a way to use it that works well.

For the second approach, I’d prefer not to use client policies because I’d like support for this specification to be integrated with the idp and clients. As a follow-up, I can imagine enabling ID-JAG validation on the identity provider as well, and if we use client policies, that might not be possible. I see client policies for extra custom claims validation which is also already supported using the JWTClaimEnforcerExecutor.

I’ll leave it to you to find a solution that works better, maybe something in between the first approach and my proposal.

Thanks for all the work and the discussion on this

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: @graziang 

Thank you for your feedback.
I have just removed JWTAuthorizationGrantValidatorBase as you suggested.


And the following is also included in the PR source code as you wanted
```
            JWTAuthorizationGrantValidator authorizationGrantContext = 
                             session.getProvider(JWTAuthorizationGrantValidator.class, jwtTokenType);
```
without any extra casting.


With some modification to how to create an instance of JWTAuthorizationGrantValidator implementation class like below: 
```
    public DefaultJWTAuthorizationGrantValidator(KeycloakSession session) {
        this(session, 
            session.getContext().getHttpRequest().getDecodedFormParameters().getFirst(OAuth2Constants.SCOPE),
            buildClientAssertionState(session));
    }
```
We do not need this any more
`authorizationGrantContext).setScopeParam(formParams.getFirst(OAuth2Constants.SCOPE));`


And for the following:
```
UserModel user = lookupUserByFederatedIdentity(federatedIdentityModel, ((DefaultJWTAuthorizationGrantValidator) authorizationGrantContext).getState());
((DefaultJWTAuthorizationGrantValidator) 
```
 I have just removed authorizationGrantContext).getState() parameter, because it is not used in lookupUserByFederatedIdentity method.

services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java
bucchi: @graziang 

> A provider created via SPI starts with clientAssertionState = null. Any method called before `setClientAssertionState()` can causes NPE, non-ideal use of the provider, It might be better to pass clientAssertionState as a parameter to validation methods instead, or some other refactor.

About your concern above, 
I have just modified the instance creation process of JWTAuthorizationGrantValidator implementation class like below:
```
    public DefaultJWTAuthorizationGrantValidator(KeycloakSession session) {
        this(session, 
            session.getContext().getHttpRequest().getDecodedFormParameters().getFirst(OAuth2Constants.SCOPE),
            buildClientAssertionState(session));
    }
```
So that a provider created via SPI starts with non null clientAssertionState.

How about this workaround?
Does it work for you?
Let me know what you think.

graziang: @bucchi The workaround initializes the provider by reading from the HTTP context in the constructor. This works for normal HTTP flows and single assertion because the provider is recreated for each session, but it creates tight coupling between the provider and the HTTP context, which can cause issues.
 For example, if the provider is invoked multiple times in the same session with different assertions the cached provider will have stale state from the first invocation, how you can select the assertion to validate?


If we want to use the SPI provider pattern, I see two options:

1. Pass and set state in each method: 
  This way we don't need to change AbstractBaseJWTValidator:                                                                                                                                                                                                                                       

  ```
    public void validateClient(ClientAssertionState state) {
        this.clientAssertionState = state;  // Update field before validation
        super.validateClient();             // Call existing base logic
    }
  
    public void validateIssuer(ClientAssertionState state) {
        this.clientAssertionState = state;                                                                                                                                                                                                                                                           
        super.validateIssuer();
    }
    // ... same pattern for other methods
  ```

 2. : Initialize method:   Set the state at the beginning using an initialize method (similar to the previous setAssertionState approach but clearer), with proper NPE handling:

 ```
 public interface JWTAuthorizationGrantValidator {
      void initialize(String scope, ClientAssertionState state);

      void validateClient();  // Call after initialize()
      void validateIssuer();
      // ...
  }
```
I don't see much additional value in using the provider via SPI for this use case, I see more complications than benefits. But if we want to go in this direction, I don't see any other good options compared to the two listed above, I slightly prefer the first one. WDYT? Do you have better ideas?

Comment (bucchi) [reply services/src/main/java/org/keycloak/protocol/oidc/grants/jwtauthorization/JWTAuthorizationGrantType.java]: @graziang 

> I don't see much additional value in using the provider via SPI for this use case, I see more complications than benefits. 

Maybe you are right about this.
I can not find the way to implement this more simpler than above.

I think your original idea below is the simplest because there is no other specific JWT type than ID-JAG so far. So, I have just modified the code based on your suggestion: 
> I know the issue is related to the relationship with AbstractBaseJWTValidator, but as things are getting complicated, maybe you could check if it would be simpler to just extend the original JWTAuthorizationGrantValidator as IDJWTAuthorizationGrantValidator and override only the methods needed to validate the ID-JAG and use it if needed in the JWTAuthorizationGrantType. WDYT?

**Comment by bucchi** [comment]
Hi, @graziang 

Just a quick heads-up: 
I've applied the code fix as you suggested and left a comment on the thread. 
https://github.com/keycloak/keycloak/pull/46048#discussion_r3305140391

Could you please take another look? 

Thanks!

**Review by graziang — APPROVED** [review]
@bucchi Thanks for all the work and discussion on this PR! Good job!

File: services/src/main/java/org/keycloak/protocol/oidc/grants/JWTAuthorizationGrantType.java [hunk 17]
```diff
@@ -164,7 +190,7 @@ public Response process(Context context) {
         }
     }
 
-    protected UserModel lookupUserByFederatedIdentity(FederatedIdentityModel federatedIdentityModel, ClientAssertionState clientAssertionState) {
+    protected UserModel lookupUserByFederatedIdentity(FederatedIdentityModel federatedIdentityModel) {
```
thomasdarimont: I was using the assertionState in a subclass to reference some claims from the assertion to narrow the client/user lookup.

Any chance to get that in again?

mposolda (commented):
@bucchi @graziang Thanks!

I hope to merge the PR once the last remaining comment by @thomasdarimont regarding the removed method parameter is solved. @bucchi Is it ok to add the parameter back?

bucchi:
Given that it is already in use, I would be more than happy to.
 
 Give me some time, please.

bucchi:
@mposolda , @thomasdarimont, @graziang 

I have just modified to make the lookupUserByFederatedIdentity's parameter back.
How about that?
Let me know what you think.

thomasdarimont:
Thanks @bucchi that works for me again :)

mposolda (approved):
@bucchi @graziang @tnorimat @thomasdarimont Thanks everyone for the participation and review of this PR!

PR #46054, opened 2026-02-05
Keycloak Refresh Token Reuse Bypass via TOCTOU Race Condition

### LINKED ISSUE #45647: CVE-2026-1035 - Keycloak Refresh Token Reuse Bypass via TOCTOU Race Condition
### Source: https://issues.redhat.com/browse/RHBK-4108
Security Tracking Issue

Flaw:
-----

Keycloak Refresh Token Reuse Bypass via TOCTOU Race Condition

A race condition (Time-of-Check to Time-of-Use) exists in the TokenManager class, specifically within the validateTokenReuse method. This vulnerability allows an attacker to bypass the refreshTokenMaxReuse security policy when it is set to zero (strict single-use). By sending concurrent requests, a single refresh token can be exchanged for multiple valid access tokens before the usage counter is updated, undermining the Refresh Token Rotation hardening measure.

mposolda:
closes #45647

Fix for the race condition. Before updating clientSession with new refresh-token count, there is now an attempt to use "single-use" cache to add the item to this cache with the current count of refresh-token. If it fails, it means that concurrent transaction tried to add the item with the same key. The "single-use" cache is cluster-aware and cross-dc aware and items are added to it before current transaction commit, so should be safe in environments like cluster/cross-dc.

Automated tests with multiple threads failed for me (after few attempts of running). After fix, it did not never failed in concurrent runs.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
@@ -17,6 +17,8 @@
 
 package org.keycloak.protocol.oidc;
 
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -27,7 +29,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
@@ -55,6 +56,7 @@ import org.keycloak.common.ClientConnection;
 import org.keycloak.common.Profile;
 import org.keycloak.common.Profile.Feature;
 import org.keycloak.common.VerificationException;
+import org.keycloak.common.util.Retry;
 import org.keycloak.common.util.SecretGenerator;
 import org.keycloak.common.util.Time;
 import org.keycloak.crypto.HashProvider;
@@ -67,6 +69,7 @@ import org.keycloak.jose.jws.JWSInput;
 import org.keycloak.jose.jws.JWSInputException;
 import org.keycloak.jose.jws.crypto.HashUtils;
 import org.keycloak.migration.migrators.MigrationUtils;
+import org.keycloak.models.AbstractKeycloakTransaction;
 import org.keycloak.models.AuthenticatedClientSessionModel;
 import org.keycloak.models.ClientModel;
 import org.keycloak.models.ClientScopeModel;
@@ -74,6 +77,7 @@ import org.keycloak.models.ClientSessionContext;
 import org.keycloak.models.Constants;
 import org.keycloak.models.IdentityProviderQuery;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
 import org.keycloak.models.ProtocolMapperModel;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.RoleModel;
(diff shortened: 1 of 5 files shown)


services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
```diff
@@ -451,6 +451,15 @@ public void validateTokenReuse(KeycloakSession session, RealmModel realm, Access
...
         }
+
+        // Attempt to add item into "single-use" cache to prevent possible concurrent requests with same refresh token
+        if (refreshFlag) {
+            String singleUseCacheKey = "rt:" + refreshToken.getId() + ":" + currentCount;
+            if (!session.singleUseObjects().putIfAbsent(singleUseCacheKey, 60)) {
+                throw new OAuthErrorException(OAuthErrorException.INVALID_GRANT, "Maximum allowed refresh token reuse exceeded",
+                        "Maximum allowed refresh token reuse exceeded");
+            }
+        }
```
ahus1: Reading the code, it looks to prevent concurrently refreshed tokens. 

This is OK if it would would prevent exceeding the max token reuse. 

I think it goes beyond that as it prevents concurrent refreshes within the bounds of the max token reuse, and therefore also prevents valid token refreshes.

ahus1 (changes requested):
Thank you for the PR. See below for a comment.

services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java
mposolda: @ahus1 Thanks for the precise check!

I rewrote the PR and replaced current approach with the use of "conditional replace" . Concurrent threads are using single-use cache (backed by infinispan/hotrod) where the thread can replace current cache value just if it was not changed in the meantime by the other thread. This should ensure correct counts being used.

Added new method to `SingleUseObjectProvider` for the conditional replace. Do you please have a chance to re-review?

NOTE: The PR will exceed count of lines on `RefreshTokenTest`, but I am planning to migrate it to the new testsuite as a follow-up (already working on it): https://github.com/keycloak/keycloak/issues/46612 . I am planning to add as a follow-up as this one require backports to the older releases and would be more convenient for backports to keep the test in the old testsuite. Is it ok?

ahus1: Hello @mposolda, thank you for this change. I agree that we we should do the changes in the old test suite to be able to backport the change. 

I now see that concurrent token refreshes will work. Still, for concurrent requests, it might write a value to the to client session that is not the highest count: With two concurrent requests that want to write "2" and "3" in there, it might still end up with a "2" in there if that updates the session last.

I'm also concerned about the 60 seconds idle timeout, it seems to be an arbitrary value. The value might be able to capture concurrent requests, but the value of truth is still in the client session, which might be off as described above. 

I was briefly thinking of having a longer value there, but that would increase the size of the cache, which is not desired. 

I then attempted to enhance the current client and user sessions to make them atomic, which lead to a lot of changes, and didn't work in all set-ups that I tried. 

My latest attempt then implements a temporary exclusive lock derived from the session ID and the tab, so it should serialize all attempts for the tab. This reverts the previous changes in the PR, and adds one new method. 

Please let me know your thoughts - I've created a PR here: https://github.com/mposolda/keycloak/pull/9

One could now argue what the correct lockId would be for this case - if it is the sessionId, or SessionId + the reuseId.

ahus1: Side note: With my latest suggestion, it should also cover the condition "check if a more recent refresh token is already used on this tab, if yes the refresh token is invalid" for concurrent requests.

mposolda: > I now see that concurrent token refreshes will work. Still, for concurrent requests, it might write a value to the to client session that is not the highest count: With two concurrent requests that want to write "2" and "3" in there, it might still end up with a "2" in there if that updates the session last.

Yes, you are right. Was thinking about reading the last value from the cache right before saving to clientSession, but did not do it in the end, which is the issue here.

But nonetheless, I've merged your PR to my PR, so we do not need new method on `SingleUseObjectsProvider` .

ahus1 (approved):
I'm ok to proceed. Please review my code before merging!

PR #46056, opened 2026-02-05
Missing anti-ID phishing check for getting client

### LINKED ISSUE #46010: Missing anti-ID phishing check for getting client
### Before reporting an issue

### Area

admin/api

### Describe the bug

For the Admin API v1, there is this check in the `ClientsResource`:

```java
ClientModel clientModel = realm.getClientById(id);
        if (clientModel == null) {
            // we do this to make sure somebody can't phish ids
            if (auth.clients().canList()) throw new NotFoundException("Could not find client");
            else throw new ForbiddenException();
        }
```

It means that if there's no `clients().canList()` permission, the 403 should be returned instead of 404. 

We're missing this check in the Admin V2 API

### Version

999

### Regression

### Expected behavior

Returns 403 when the permission is not granted

### Actual behavior

Always returns 404

### How to Reproduce?

Try to obtain a non-existent user without the `client().canList()` permission via the Client Admin API v2

### Anything else?

_No response_

mabartos:
- Closes #46010

### DIFF
--- a/rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
@@ -142,6 +142,15 @@ public class DefaultClientService implements ClientService {
         return null;
     }
 
+    @Override
+    public void deleteClient(RealmModel realm, String clientId) throws ServiceException {
+        avoidClientIdPhishing();
+        if (clientResource == null) {
+            throw new ServiceException("Cannot find the specified client", Response.Status.NOT_FOUND);
+        }
+        clientResource.deleteClient();
+    }
+
     /**
      * Declaratively manage client roles - ensures the client has exactly the roles specified in 'rolesFromRep'
      * <p>
--- a/rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
@@ -7,6 +7,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
 import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.core.Response;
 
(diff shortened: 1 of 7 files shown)


rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -25,33 +27,54 @@
...
+            // we do this to make sure somebody can't phish client IDs
+            throw new ServiceException("Insufficient permissions to access the client", Response.Status.FORBIDDEN);
+        }
     }
 
     @Override
-    public Optional<BaseClientRepresentation> getClient(RealmModel realm, String clientId, ClientProjectionOptions projectionOptions) {
+    public Optional<BaseClientRepresentation> getClient(RealmModel realm, String clientId, ClientProjectionOptions projectionOptions) throws ServiceException {
         // TODO: is the access map on the representation needed
+        avoidClientIdPhishing();
```
vmuzikar, line 77: @mabartos Thanks for the fix.

Shouldn't we rather have this fix here:
https://github.com/keycloak/keycloak/blob/b61a00cbbad4fc5a0f3e0fc4e5713a6151bc4049/rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientsApi.java#L65-L70

That would make it more robust and prevent theoretical IDs phishing also for other methods like PATCH. Or maybe we could do it directly in the constructor of `DefaultClientService`.

mabartos: Unfortunately, we can't do it. The `createOrUpdate()` accepts having the clientResource to be nullable, so we cannot make the assumption beforehand - we can just check it directly in specific methods. 

I think it should be done on the service layer as it's the single source of checking the authz/permissions. Once we move some of the patch logic to the service layer, we'll require the manage-clients role, so this check should not be needed.

vmuzikar: Ok, works for me. Seems it will work for patch too as it calls `getClient`. But we'll at least need to create test for PATCH and DELETE. I'd say it could go into this PR.

mabartos: @vmuzikar Ok, I'll provide it. Thanks!

mabartos:
@vmuzikar 

I made some changes:
- Create test for PATCH
- Improve ServiceException
- Ensure no phishing with the DELETE op
- Create test for DELETE

As a follow-up, we could have some error handler, that could automatically convert all ServiceException(s) to WebApplicationException. 

Ready for the review.

vmuzikar (approved):
LGTM, thank you.

PR #46057, opened 2026-02-05
[OID4VCI] Confine test realm setup to TestCase.configureTestRealm()

### LINKED ISSUE #46055: [OID4VCI] Confine test realm setup to TestCase.configureTestRealm()
### Description

Currently, test realm setup is spread over before/after realm import and per test method. It should however be such that most (if not all) of the realm config happens in `configureTestRealm` so that it does not get repeated for for every test method.

### Value Proposition

Refactor `OID4VCIssuerEndpointTest` such that the test realm is setup properly

### Goals

Fix configureTestRealm for OID4VCIssuerEndpointTest

### Non-Goals

--

### Discussion

_No response_

### Notes

_No response_

tdiesler:
closes #46055

### DIFF
--- a/services/src/main/java/org/keycloak/services/managers/RealmManager.java
+++ b/services/src/main/java/org/keycloak/services/managers/RealmManager.java
@@ -19,6 +19,7 @@ package org.keycloak.services.managers;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import jakarta.ws.rs.BadRequestException;
@@ -79,6 +80,7 @@ import org.keycloak.utils.SMTPUtil;
 import org.keycloak.utils.StringUtil;
 
 import static org.keycloak.constants.OID4VCIConstants.CREDENTIAL_OFFER_CREATE;
+import static org.keycloak.models.Constants.CREATE_DEFAULT_CLIENT_SCOPES;
 
 /**
  * Per request object
@@ -647,8 +649,7 @@ public class RealmManager {
                 setupOfflineTokens(realm, rep);
             }
 
-
-            if (rep.getClientScopes() == null) {
+            if (isCreateDefaultClientScopes(rep)) {
                 createDefaultClientScopes(realm);
             }
 
@@ -717,6 +718,12 @@ public class RealmManager {
         }
 
         return realm;
+    }
+
+    private boolean isCreateDefaultClientScopes(RealmRepresentation rep) {
+        Map<String, String> attributes = rep.getAttributesOrEmpty();
+        String createDefaultClientScopes = attributes.remove(CREATE_DEFAULT_CLIENT_SCOPES);
+        return rep.getClientScopes() == null || Boolean.parseBoolean(createDefaultClientScopes);
     }
 
     private String determineDefaultRoleName(RealmRepresentation rep) {
--- a/core/src/main/java/org/keycloak/OID4VCConstants.java
+++ b/core/src/main/java/org/keycloak/OID4VCConstants.java
@@ -5,6 +5,8 @@ package org.keycloak;
  */
 public class OID4VCConstants {
 
+    public static final String OID4VCI_ENABLED_ATTRIBUTE_KEY = "oid4vci.enabled";
+
     // Sd-JWT constants
     public static final String SDJWT_DELIMITER = "~";
     public static final String SD_HASH = "sd_hash";
--- a/server-spi-private/src/main/java/org/keycloak/models/Constants.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/Constants.java
@@ -227,4 +227,8 @@ public final class Constants {
 
     // Internal note for storing authorization details response in client session context
     public static final String AUTHORIZATION_DETAILS_RESPONSE = "authorization_details_response";
+
+    // This attribute can be used in a realm import definition to signal that default client scopes should be created in addition to the client scopes defined by the realm import definition.
+    // When this attribute is omitted or set to false, the default client scopes are not created if at least one other client scope is defined by the realm import definition.
+    public static final String CREATE_DEFAULT_CLIENT_SCOPES = "CreateDefaultClientScopes";
 }
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -137,6 +137,7 @@ import org.apache.http.client.methods.HttpOptions;
 import org.apache.http.client.methods.HttpPost;
 import org.jboss.logging.Logger;
 
+import static org.keycloak.OID4VCConstants.OID4VCI_ENABLED_ATTRIBUTE_KEY;
 import static org.keycloak.OID4VCConstants.OPENID_CREDENTIAL;
 import static org.keycloak.constants.OID4VCIConstants.CREDENTIAL_OFFER_CREATE;
 import static org.keycloak.constants.OID4VCIConstants.OID4VC_PROTOCOL;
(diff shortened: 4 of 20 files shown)


core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java
```diff
@@ -1404,6 +1404,11 @@ public void setClientScopes(List<ClientScopeRepresentation> clientScopes) {
         this.clientScopes = clientScopes;
     }
 
+    @JsonIgnore
+    public boolean isCreateDefaultClientScopes() {
```
mposolda: Is it possible to remove changes from class `RealmRepresentation` and also from `RealmManager` from this PR ? Hopefully we do not need changes in those classes just because of test setup.

Note there is method `afterAbstractKeycloakTest()` defined on `AbstractKeycloakTest`, which subclasses can override. Perhaps it should be possible to initiate/create client scopes inside this method?

See for example `TokenIntrospectionTest.afterAbstractKeycloakTestRealmImport()` for the "inspiration. (That class is doing some programmatic changes in default client-scopes once they were already imported).

mposolda (commented):
@tdiesler Thanks! This change is nice IMO.

Added some comment inline. Can you please doublecheck?

core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java
tdiesler: In `RealmManager` we used to have this

```
            if (rep.getClientScopes() == null) {
                createDefaultClientScopes(realm);
            }
```

which effectively prevents the realm default client scopes (e.g. oid4vci_natural_person) to get added as soon as you define a client scope in `configureTestRealm`

I could move the predicate function to the RealmManger instead of having it in `RealmRepresentation `

```
    @JsonIgnore
    public boolean isCreateDefaultClientScopes() {
        return clientScopes == null || Boolean.parseBoolean(getAttributesOrEmpty().get("CreateDefaultClientScopes"));
    }
```

In any case we need to have a predicate (beyond rep.clientScopes == null) that drives whether we want the default client scopes that are added by the protocol factories.

And while I have you ... ;-)

I currently still cannot do the `ClientScope => ClientMapping` before realm import 

```
            // [TODO] Cannot assign optional client scopes to client before realm import
            // List<String> optionalClientScopes = new ArrayList<>(Optional.ofNullable(testClient.getOptionalClientScopes()).orElse(List.of()));
            // optionalClientScopes.addAll(clientScopes.stream().map(ClientScopeRepresentation::getName).toList());
            // testClient.setOptionalClientScopes(optionalClientScopes);
```

It still happens in `afterAbstractKeycloakTestRealmImport` like this ...

```
        // [TODO] Cannot assign optional client scopes to client before realm import
        assignOptionalClientScopeToClient(clientId, sdJwtTypeCredentialScopeName);
        assignOptionalClientScopeToClient(clientId, jwtTypeCredentialScopeName);
        assignOptionalClientScopeToClient(clientId, minimalJwtTypeCredentialScopeName);

    protected void assignOptionalClientScopeToClient(String clientId, String scopeName) {
        ClientRepresentation testClient = testRealm().clients().findByClientId(clientId).get(0);
        ClientScopeRepresentation clientScope = requireExistingClientScope(scopeName);
        assignOptionalClientScopeToClient(testClient, clientScope);
    }

    protected void assignOptionalClientScopeToClient(ClientRepresentation client, ClientScopeRepresentation clientScope) {
        if (isClientOid4vciEnabled(client) && !client.getOptionalClientScopes().contains(clientScope.getName())) {
            ClientResource clientResource = testRealm().clients().get(client.getId());
            clientResource.addOptionalClientScope(clientScope.getId());
        }
    }
```

Is there a way to do this on the `RealmRepresentation` before the import?

tdiesler: This does also not work ...

```
        List<String> defaultOptionalClientScopes = new ArrayList<>(Optional.ofNullable(testRealm.getDefaultOptionalClientScopes()).orElse(List.of()));
        defaultOptionalClientScopes.addAll(clientScopes.stream().map(ClientScopeRepresentation::getName).toList());
        testRealm.setDefaultOptionalClientScopes(defaultOptionalClientScopes);
```

tdiesler: I removed the predicate from `RealmRepresentation `

tdiesler:
All realm config is now truly done in configureTestRealm(). There was an awkward side effect: when you add an optional client scope through the ClientModel (i.e. after import) that client also had the default scopes added. We now explicitly do that before realm import.

services/src/main/java/org/keycloak/services/managers/RealmManager.java
```diff
@@ -719,6 +719,11 @@ protected void rollbackImpl() {
         return realm;
     }
 
+    private boolean isCreateDefaultClientScopes(RealmRepresentation rep) {
+        Map<String, String> repAttributes = rep.getAttributesOrEmpty();
+        return rep.getClientScopes() == null || Boolean.parseBoolean(repAttributes.get("CreateDefaultClientScopes"));
```
mposolda: As pointed before, we should not introduce "fake" realm attributes in the core codebase just because of the tests setup. Ideally, there should not be any changes in the class `RealmManager` in this PR.

Similarly like the other tests, which do anything with client scopes: It should be possible to override method `afterAbstractKeycloakTestRealmImport()` and do some additional setup of client-scopes, clients etc after the realm is imported with default client scopes. This would need to use REST endpoints for updating client scopes and other entities, but should work fine without a need for workarounds.

mposolda (commented):
@tdiesler Added point inline (related to the same I've mentioned before already).

services/src/main/java/org/keycloak/services/managers/RealmManager.java
tdiesler: If we leave the predicate in `RealmManager` as is (i.e. `rep.getClientScopes() == null`), it would mean ...

* Realm representations that do not define any client scope will end up having all default client scopes
* Realm representations that define a least one client scope will end up having just what they define and no default client scopes

IMHO, this is inconsistent behavior. With this change we offer consistency (i.e. you can choose whether to have the default client scopes)

Also, please consider that adding client scopes in `afterAbstractKeycloakTestRealmImport` and then adding them to existing clients has strange side effects too. The target client will not only get the optional scope added as implied by the API call, it also gets default scopes added that were not asked for.

Generally, I believe it should be possible to define "everything" in before import (i.e. a realm definition that is complete)

Wanting to import default client scopes in addition to what is being defined in `configureTestRealm()` is reasonable because ...

* you want to test aspects of default client scopes (e.g. oid4vc_natural_person)
* you do not want to re-define a default client scope in your test setup (i.e. definition might diverge)

Alternatively, I would have to redefine all default client scopes that I want to test in `configureTestRealm` or define all test client scopes after the import in `afterAbstractKeycloakTestRealmImport`, which defies `configureTestRealm` for client scopes.

How about, we remove that fake realm attribute on import, so that it only configures import behavior but does not end up as an attribute on the realm itself?

tdiesler:
Removing the "fake" attribute on import is more correct anyway - I did that. If you still want to get rid of it, I can of course also redefine the `natural_person` client scope in `configureTestRealm` - your call.

I haven't tried, but it may also be necessary to redefine some of these: "web-origins", "acr", "roles", "profile", "basic", "email" which might not be such a good idea.

services/src/main/java/org/keycloak/services/managers/RealmManager.java
mposolda: Yeah, I can see the behavior regarding "default client scopes" in the realm file is not so ideal...

Removing attribute after import is a bit better. Not ideal, but I consider it OK.

One last nitpick: Could you please add that attribute `"CreateDefaultClientScopes"` as a constant in the file `org.keycloak.models.Constants`  and add some javadoc/comment to it? Something like `This attribute is used as an attribute in the realm JSON file to signal that default client scopes should be created in addition to the client scopes defined inside the JSON file. When this attribute is omitted or set to false, the default client scopes are not created if some other client scopes are defined in the realm JSON file` . WDYT?

tdiesler: done

mposolda:
LGTM, might require rebase to fix the conflicts.

mposolda (dismissed):

mposolda (approved):
@tdiesler Thanks!

PR #46060, opened 2026-02-05
Documentation changes to deprecate Token Exchange V1

rmartinc:
Closes #45792

Changes:

* The guide for token exchange is changed to show the V1 is now preview but deprecated.
* The comparison table is changed to differentiate internal-to-external and external-to-internal. This way in the latter we can link to the cross domain guide.
* Upgrading guide note to say that TEv1 is now deprecated in preview.

I have not added a release note. I think that upgrading is enough, but maybe I'm wrong. :smile: 
Do we need something more?


mposolda (approved):
@rmartinc LGTM, Thanks!

PR #46064, opened 2026-02-05
Revisit retransmission metrics documentation

### LINKED ISSUE #46059: Revisit retransmission metrics documentation
### Description

The statement about the [healthy cluster having zero retransmissions](https://www.keycloak.org/observability/metrics-for-troubleshooting-clustering-and-network#_retransmissions)  is not 100% correct.
`UNICAST3` (protocol response for message reliable delivery) delays sending acks to reduce the network traffic. It may happen, due to timing issues, that the sender has not received an ack for a while and retransmits the message eagerly. It doesn't mean there is a problem with the cluster.

pruivo:
Closes #46059

### DIFF
--- a/docs/guides/observability/partials/jgrp_metrics.adoc
+++ b/docs/guides/observability/partials/jgrp_metrics.adoc
@@ -148,13 +148,16 @@ m| vendor_jgroups_mfc_get_average_time_blocked
 
 JGroups provides reliable delivery of messages.
 When a message is dropped on the network, or the receiver cannot handle the message, a retransmission is required.
-Retransmissions increase resource usage, and it is usually a signal of an overload system.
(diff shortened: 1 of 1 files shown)


ryanemerson (approved):
LGTM, thanks @pruivo

ahus1:
Approving the PR based on Ryan's review. Some afterthoughts: The documentation might now be more technically correct, but it is also more complex and less actionable. 

What is occasional, what can cause timing issues, and are those within milliseconds or seconds? When is the metric in the "green", when in the "red", and when and how should I act? Should I check my network connection, should I change my timeouts? 

I know that this is not simple, still trying to raise the bar for the benefit of the readers.

pruivo:
@ahus1 short answer, I don't know. 
The metric is useful for troubleshooting problems, not for alerting.

PR #46082, opened 2026-02-06
Redirect Wildcard changes and more https checks to secure-client-executor

### LINKED ISSUE #45587: SecureClientUrisExecutor doesn't allow for "+" weborigin
### Before reporting an issue

### Area

oidc

### Describe the bug

When adding a `secure-client-uris` policy check, this does not allow a Web Origins setting of `+` even though those URLs are secure.

### Version

main

### Regression

### Expected behavior

A `+` should be accepted if the `rootUrl` and `redirectUris` are secure.

### Actual behavior

A `+` leads to a validation error in the frontend: 

> Invalid webOrigins

### How to Reproduce?

* Create a client policy that contains `secure-client-uris`
* Enter `+` in the WebOrigins

### Anything else?

The problematic code is somewhere about here: 

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientUrisExecutor.java#L106-L108

The WebOrigins are usually resolved as shown here: 

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/services/src/main/java/org/keycloak/protocol/oidc/utils/WebOriginsUtils.java#L39-L46

msdaly200:
SecureClientUrisExecutor doesn't allow for "+" weborigin

Updated the secure-client-uris executor for :
1. Expanded HTTPS enforcement for Post logout redirect URIs', 'Sector Identifier URI' (Pairwise Subject Identifier mapper configuration), 'Logo URL', 'Policy URL', and 'Terms of Service URL'.
2. Expanded Wildcard Support: The + wildcard character is now allowed in 'Valid post logout redirect URIs' and 'Web origins'. 

Closes [#45587]( https://github.com/keycloak/keycloak/issues/45587
)

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -14,6 +14,18 @@ It was also not aligned with how other script-based features work in {project_na
 
 If you have existing JavaScript-based policies, make sure to enable the `scripts` feature when starting {project_name}.
 
+=== Stricter Validation for Client URIs (secure-client-uris)
+The 'secure-client-uris' executor has been updated to enforce stricter validation on Client URIs. This aligns with FAPI 2.0 security requirements by mandating TLS (HTTPS) for client endpoints.
+
+HTTPS Enforcement: The following URIs must now use the https scheme. Existing clients configured with http in these fields will fail validation during updates:
+'Post logout redirect URIs', 'Sector Identifier URI' (Pairwise Subject Identifier mapper configuration), 'Logo URL', 'Policy URL', and 'Terms of Service URL'.
+
+Expanded Wildcard Support: The + wildcard character is now allowed in 'Valid post logout redirect URIs' and 'Web origins'.
+Configuring + defers validation to the client's standard 'Valid redirect URIs'. Since redirect URIs are already checked for HTTPS, using the wildcard maintains FAPI 2.0 compliance while reducing configuration duplication.
+
+Administrators must ensure that all configured URIs for the fields listed above use https. Clients attempting to update or register with http in these fields will fail validation when using the 'secure-client-uris' executor.
+
+
 // ------------------------ Notable changes ------------------------ //
 == Notable changes
 
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -15,7 +15,7 @@ It was also not aligned with how other script-based features work in {project_na
 If you have existing JavaScript-based policies, make sure to enable the `scripts` feature when starting {project_name}.
 
 === Stricter Validation for Client URIs (secure-client-uris)
-The 'secure-client-uris' executor has been updated to enforce stricter validation on Client URIs. This aligns with FAPI 2.0 security requirements by mandating TLS (HTTPS) for client endpoints.
+The 'secure-client-uris' client policy executor has been updated to enforce stricter validation on Client URIs. This aligns with FAPI 2.0 security requirements by mandating TLS (HTTPS) for client endpoints.
 
 HTTPS Enforcement: The following URIs must now use the https scheme. Existing clients configured with http in these fields will fail validation during updates:
 'Post logout redirect URIs', 'Sector Identifier URI' (Pairwise Subject Identifier mapper configuration), 'Logo URL', 'Policy URL', and 'Terms of Service URL'.
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -18,7 +18,7 @@ If you have existing JavaScript-based policies, make sure to enable the `scripts
 The 'secure-client-uris' client policy executor has been updated to enforce stricter validation on Client URIs. This aligns with FAPI 2.0 security requirements by mandating TLS (HTTPS) for client endpoints.
 
 HTTPS Enforcement: The following URIs must now use the https scheme. Existing clients configured with http in these fields will fail validation during updates:
-'Post logout redirect URIs', 'Sector Identifier URI' (Pairwise Subject Identifier mapper configuration), 'Logo URL', 'Policy URL', and 'Terms of Service URL'.
+'Post logout redirect URIs', 'Logo URL', 'Policy URL', and 'Terms of Service URL'.
 
 Expanded Wildcard Support: The + wildcard character is now allowed in 'Valid post logout redirect URIs' and 'Web origins'.
 Configuring + defers validation to the client's standard 'Valid redirect URIs'. Since redirect URIs are already checked for HTTPS, using the wildcard maintains FAPI 2.0 compliance while reducing configuration duplication.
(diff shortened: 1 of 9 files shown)


services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientUrisExecutorFactory.java
```diff
@@ -32,6 +32,11 @@ public class SecureClientUrisExecutorFactory implements ClientPolicyExecutorProv
 
     public static final String PROVIDER_ID = "secure-client-uris";
 
+    public static final String SECTOR_IDENTIFIER_URI = "sectorIdentifierUri";
```
mposolda: Nitpick: Is it possible to define those attributes somewhere else? Maybe the `OIDCConfigAttributes` might be good location.

As those are names of client attributes, which are not specific to this "secure-client-uris-executor" , but might be used in other places in the Keycloak code .

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientUrisExecutor.java
```diff
@@ -86,46 +89,72 @@ public void executeOnEvent(ClientPolicyContext context) throws ClientPolicyExcep
...
+        // OIDC: Post Logout URL
+        List<String> postLogoutRedirectUris = getAttributeMultivalued(clientRep, OIDCConfigAttributes.POST_LOGOUT_REDIRECT_URIS);
+        if (postLogoutRedirectUris != null && !postLogoutRedirectUris.isEmpty()) {
+            List<String> validRedirects = clientRep.getRedirectUris() != null ? clientRep.getRedirectUris() : Collections.emptyList();
+            List<String> resolvedPostLogoutUrls = resolveUrlWithRedirects(postLogoutRedirectUris, validRedirects, clientRep.getRootUrl(), false);
+            confirmSecureUris(resolvedPostLogoutUrls, "postLogoutUris");
+        }
+
+        // sector_identifier_uri
+        String sectorIdentifierUri = Optional.ofNullable(clientRep.getAttributes()).orElse(Collections.emptyMap()).get(SecureClientUrisExecutorFactory.SECTOR_IDENTIFIER_URI);
```
mposolda: Are you sure that "Sector identifier uri" is saved as a client attribute `sectorIdentifierUri` when the client is being updated?

It seems to me it is not saved inside this client attribute. It rather looks that when "sector identifier uri" is being used, it is saved as an attribute on the "Pairwise" protocol mapper of the client.

It looks that "Sector identifier URI" is validated in `PairWiseSubMapperValidator.validate` that it must match one of the client redirect uris. So considering that, maybe you can omit the "sectorIdentifierUri" from this validator, because as long as "redirect uris" are valid, the "sectorIdentifierUri" should be implicitly validated as well. WDYT?

Doing some additional validations of "sector identifier URI" might be hard as it seems that it is being updated when the update of corresponding "pairwise" protocol mapper is triggered, but not during client update (hence client policies might not be triggered).

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -14,6 +14,18 @@ It was also not aligned with how other script-based features work in {project_na
 
 If you have existing JavaScript-based policies, make sure to enable the `scripts` feature when starting {project_name}.
 
+=== Stricter Validation for Client URIs (secure-client-uris)
+The 'secure-client-uris' executor has been updated to enforce stricter validation on Client URIs. This aligns with FAPI 2.0 security requirements by mandating TLS (HTTPS) for client endpoints.
```
mposolda:
```suggestion
The 'secure-client-uris' client policy executor has been updated to enforce stricter validation on Client URIs. This aligns with FAPI 2.0 security requirements by mandating TLS (HTTPS) for client endpoints.
```
Just a nitpick to make it more explicit that this is about "client policies" .

services/src/main/java/org/keycloak/protocol/oidc/utils/RedirectUtils.java
```diff
@@ -43,6 +45,7 @@
 public class RedirectUtils {
 
     public static final Set<String> LOOPBACK_INTERFACES = new HashSet<>(Arrays.asList("localhost", "127.0.0.1", "[::1]"));
+    public static final String INCLUDE_REDIRECTS = "+";
```
mposolda: Nitpick: This constant is already defined in some other place in the class `WebOriginsUtils` . Is it possible to re-use the existing constant instead of define it again on yet another place?

Bonus: it can be also good to also update some existing places, which already use this and have hardcoded use of `"+"` . For example in class `RealmManager` .

mposolda (commented):
@msdaly200 Thanks!

Added few comments inline. Mostly nitpicks (possibly with exception of "Sector identifier URI", which is not nitpick).

mposolda:
@msdaly200 IT would be also good to fix the "DCO" check for this PR. This can be done by including signature to your GH commit. See contributions guide for some details: https://github.com/keycloak/keycloak/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

services/src/main/java/org/keycloak/protocol/oidc/utils/RedirectUtils.java
msdaly200: I've moved "+" const to server-spi-private/src/main/java/org/keycloak/models/Constants.java to be used by WebOriginsUtils/RedirectUtils and RealmManager.

services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureClientUrisExecutor.java
msdaly200: Thanks Marek, yes the check should be in PairwiseSubMapperUtils/Validator and not in the executor for the reasons you mentioned. Looks like someone tried to enforce https here before but commented it out.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/client/policies/ClientPoliciesExecutorTest.java
```diff
@@ -1830,4 +1869,133 @@ private CloseableHttpResponse sendRequest(String requestUrl, List<NameValuePair>
...
+                attributes.put(OIDCConfigAttributes.JWKS_URL, "https://client.example.com/jwks/");
+                clientRep.setAttributes(attributes);
+                // OIDC : requestUris
+                setAttributeMultivalued(clientRep, OIDCConfigAttributes.REQUEST_URIS, Arrays.asList("https://client.example.com/request/", "https://client.example.com/reqobj/"));
+                // CIBA Client Notification Endpoint
+                attributes.put(CibaConfig.CIBA_BACKCHANNEL_CLIENT_NOTIFICATION_ENDPOINT, "https://client.example.com/client-notification/");
+                clientRep.setAttributes(attributes);
+            });
+        } catch (Exception e) {
+            fail();
```
mposolda, line 1967: Nitpick: It is probably better to not catch Exception at all here. When exception is caught and `fail()` is called, it is not clear what the underlying exception was once you see the test failure.

However this is not big issue and I see this "pattern" is already used in some other tests, so don't consider it blocker for the PR.

mposolda (approved):
@msdaly200 Thanks! Added nitpick comment, but this is not a blocker for this PR. Rather maybe something for the future... I hope to merge if tests are OK

PR #46086, opened 2026-02-06
[quarkus-next] DatasourcesDistTest fails due to Quarkus stdout/stderr…

Pepo48:
… capture changes

Closes: #46084

This PR needs to be merged to main after the upgrade to Quarkus 3.32.

### DIFF
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/storage/database/dist/DatasourcesDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/storage/database/dist/DatasourcesDistTest.java
@@ -18,7 +18,6 @@ public class DatasourcesDistTest {
     @Test
     @Launch({"start-dev", "--db-kind-users=postgres", "--db-kind-clients=postgres", "--transaction-xa-enabled-users=false"})
     public void multipleNonXaDatasources(CLIResult result) {
-        result.assertNoMessage("Multiple datasources are specified:"); // log handlers are not initialized yet, so only the error should be visible
         result.assertExitCode(CommandLine.ExitCode.SOFTWARE);
         result.assertError("Multiple datasources are configured but more than 1 (<default>, users) is using non-XA transactions.");
     }


mabartos (dismissed):

quarkus/tests/integration/src/test/java/org/keycloak/it/storage/database/dist/DatasourcesDistTest.java
```diff
@@ -18,7 +18,7 @@ public class DatasourcesDistTest {
     @Test
     @Launch({"start-dev", "--db-kind-users=postgres", "--db-kind-clients=postgres", "--transaction-xa-enabled-users=false"})
     public void multipleNonXaDatasources(CLIResult result) {
-        result.assertNoMessage("Multiple datasources are specified:"); // log handlers are not initialized yet, so only the error should be visible
+        result.assertMessage("Multiple datasources are specified: clients, <default>, users");
```
vmuzikar: Let's just remove the assert to avoid differences in tests based on Quarkus version. The following `assertError` is more important in any case.

Pepo48: Addressed.

vmuzikar (approved):
@Pepo48 Thanks.

quarkus/tests/integration/src/test/java/org/keycloak/it/storage/database/dist/DatasourcesDistTest.java
vmuzikar: @mabartos Since you previously reviewed, does this work for you too?

mabartos (approved):

Pepo48:
@mabartos this should've wait for 3.32, at least according to their milestones. Thus the on-hold label.

Edit:
3.32.CR1 should be released in two days.

mabartos:
@Pepo48 Thanks for pointing it out. As Vasek mentioned, the error case is much more important than the log message that does not add much value to the context. 

In this case, IMO, it does not care whether the log message is present or not :) 

Please, do not revert these merged changes - if it helps to stabilize Quarkus Next, it provides maybe more value.

vmuzikar:
I don't see any issue with merging it, it did not need to wait.

Pepo48:
Sorry, since the complete removal of assertion this is fine indeed. I mixed it up with the original proposal, where I completely reverted the assertion logic.. anyway, my bad, thanks for the reviews and merge!

PR #46091, opened 2026-02-06
New key affinity for session ids

### LINKED ISSUE #46090: New key affinity for session ids
### Description

The `org.infinispan.affinity.KeyAffinityService` shipped with Infinispan is too strict for Keycloak, having the following drawbacks
* It blocks while waiting for the keys to be generated, and the node is assigned some segments. Usually, it's not an issue as it happens mostly during boot time, but it may happen during network partition recovery.
* It throws an exception if the node does not have any segments assigned to it. This should never happen as Keycloak does not support zero-capacity nodes. 

### Value Proposition

Improve response times by preventing blocking threads, at the cost of the session being cached in another instance (no free lunch available)

### Goals

Prevent requests blocking, leading to thread pool exhaustion, while trying to keep the network traffic low, by creating session ids local to the instance.

### Non-Goals

...

### Discussion

_No response_

### Notes

See https://github.com/keycloak/keycloak/issues/45998

pruivo:
Closes #46090


ahus1 (approved):

ahus1:
@pruivo - thank you for this work! Maybe it would be something for the upstream Infinispan as well, maybe even as a default behavior.

pruivo:
@ahus1 I can't change the logic as it would break the contract.
```
   /**
    * Returns a key that will be distributed on the cluster node identified by address.
    * @param address identifying the cluster node.
    * @return a key object
    * @throws IllegalStateException if the service has not been started or it is shutdown
    */
   K getKeyForAddress(Address address);
```

I can add a new method with the logic we have in KC, but I'm not sure there is demand for it. This code is much simpler :rofl:

Issue #45940, opened 2026-02-02
Authn/Authz for the new service layer


mabartos:
### Description

We should analyse how authentication and authorization are handled in the Admin API v1 and be sure that we don't loosen some permissions in the new Client API v2. It'd be good to analyze what permissions are needed for the set of operations we're working on and potentially simplify usage of the permission management and handling these objects across subresources.

mabartos:
# Result
This analysis checks authorization rules/permissions for the Clients resources/API v1 and suggests ones for the Clients API v2. We should check that for v2 and v1, there are the same permissions. This analysis might be used as a guideline - for the code and for the Authz tests. At this moment, we use the v1 logic, so these permissions are already applied for the v2, so maybe we could start with writing authz tests. 

We should check the permissions on the service layer. 

## Client V1 permissions

### Clients resource

**1. Get all clients** - `@GET /clients`
```java
-> auth.clients().requireList()
-> AdminPermissionsSchema.SCHEMA.isAdminPermissionsEnabled(realm) 
|| auth.clients().canView()
|| auth.clients().canView(client)
```

The `isAdminPermissionsEnabled`:
```java
AdminPermissionsSchema.SCHEMA.isAdminPermissionsEnabled(realm) =
Profile.isFeatureEnabled(Profile.Feature.ADMIN_FINE_GRAINED_AUTHZ_V2) 
&& realm.isAdminPermissionsEnabled();
```

**2. Create client** - `@POST /clients`
```java
-> auth.clients().requireManage()
-> ClientPolicyEvent.REGISTER  (Client policies)
```

### Client resource
**1. Get client** - `@GET /clients/{client}`
a) client == null
```java
if (auth.clients().canList()){
    return 404
} else {
    return 403
}
```

b) client != null
```java
-> ClientPolicyEvent.VIEW  (Client policies)
-> auth.clients().requireView(client)
```

**2. Update client** - `@PUT /clients/{client}`
```java
-> auth.clients().requireConfigure(client)
-> ClientPolicyEvent.UPDATE  (Client policies)
-> ClientPolicyEvent.UPDATED  (Client policies)
```

**3. Delete client** - `@DELETE /clients/{client}`
```java
-> auth.clients().requireManage(client)
-> !AdminPermissionsSchema.isAdminPermissionClient()
-> ClientPolicyEvent.UNREGISTER  (Client policies)
```

### Client roles resource 
(in v2 part of the Client resource/API)

[...]

PR #46093, opened 2026-02-06
[client-v2] Extend client policies tests to check exact events

### LINKED ISSUE #46114: Extend client policies tests to check exact events
### Description

There are new tests for client profile/policies and the Client API v2 made in : https://github.com/keycloak/keycloak/pull/46029. 

However, we could also extend the testsuite to verify that only the expected policies with a specific event type are executed.

mabartos:
- Closes https://github.com/keycloak/keycloak/issues/46114
- Extended tests for the client policies in the https://github.com/keycloak/keycloak/pull/46029
- Create `TestClientPolicyExecutor` executor that just observes all client policy types and stores them inan  ordered list
- Created the `YesClientPolicyCondition` policy condition that always returns YES, and the wanted client policy is always evaluated/executed
- Based on the analysis in: https://github.com/keycloak/keycloak/issues/45940#issuecomment-3840875460

@edewit @vmuzikar Can you check it? Thanks!

### DIFF
--- a/rest/admin-v2/tests/pom.xml
+++ b/rest/admin-v2/tests/pom.xml
@@ -41,6 +41,11 @@
             <groupId>org.keycloak</groupId>
             <artifactId>keycloak-admin-v2-rest</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.keycloak.tests</groupId>
+            <artifactId>keycloak-tests-custom-providers</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
+++ b/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientPoliciesV2Test.java
@@ -41,19 +41,26 @@ import org.keycloak.representations.idm.ClientPolicyExecutorRepresentation;
 import org.keycloak.representations.idm.ClientPolicyRepresentation;
 import org.keycloak.representations.idm.ClientProfileRepresentation;
 import org.keycloak.representations.idm.ClientProfilesRepresentation;
+import org.keycloak.services.clientpolicy.ClientPolicyEvent;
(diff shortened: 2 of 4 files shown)


edewit (approved):

mabartos:
@edewit Thanks for the review!

Some afterthought - we can just put these custom providers into the custom-providers module of the testsuite - does not have to be part of the Admin API v2 -> I'll update it.

mabartos:
@edewit @vmuzikar Ready for the review

edewit (approved):

tests/custom-providers/src/main/java/org/keycloak/testsuite/client/policies/YesClientPolicyCondition.java
```diff
@@ -0,0 +1,83 @@
...
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.representations.idm.ClientPolicyConditionConfigurationRepresentation;
+import org.keycloak.services.clientpolicy.ClientPolicyContext;
+import org.keycloak.services.clientpolicy.ClientPolicyVote;
+import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProvider;
+import org.keycloak.services.clientpolicy.condition.ClientPolicyConditionProviderFactory;
+
+public class YesClientPolicyCondition implements ClientPolicyConditionProvider<YesClientPolicyCondition.Configuration>, ClientPolicyConditionProviderFactory {
```
vmuzikar: I might be wrong but I think `AnyClientCondition` does the same thing.

mabartos: @vmuzikar Nice catch! I'll update it. Thanks!

mabartos: Updated

vmuzikar (approved):
LGTM, thanks.

PR #46096, opened 2026-02-06
[quarkus-next] configureResteasy() missing Quarkus build step dependency

Pepo48:
Closes: #46095

This PR needs to be merged to main after the upgrade to Quarkus 3.32 and after the related Quarkus upstream fix https://github.com/quarkusio/quarkus/pull/52463.


Pepo48:
This change may not be needed at all, if Quarkus confirms that the bug is purely on their side, see: https://github.com/quarkusio/quarkus/pull/52463#issuecomment-3864411587. I believe that on our end we cannot guarantee a non-empty list by altering Keycloak's build step order alone, as we don't control when `ResteasyReactiveProcessor.setupDeployment()` runs.

Pepo48:
Fixed by: https://github.com/quarkusio/quarkus/pull/52494

Issue #28513, opened 2024-04-08
Make the Operator aware of other config sources in Keycloak


vmuzikar:
### Description

Currently, the Operator is aware only of the configuration that is provided in the Keycloak CR. Some of this configuration affects also other deployment setup which includes options like `https-port` or `http-relative-path` that affects probes. If those options are provided by a different config source (like env var in Dockerfile or keycloak.conf file), it is not reflected by the Operator.

### Details

One solution to this would be just documenting it. However, we still do want to support immutable containers and we actually do encourage users to bake in their config in the container. So just documenting it might not be ideal.

vmuzikar:
Related issue: #10287

vmuzikar:
CC @keycloak/cloud-native

PR #46103, opened 2026-02-08
fix: adding a check for the client api v2 feature

shawkins:
closes: #45810

In addition to the usability concern of enabling the feature both in the image and the CR, see the comments / TODOs.

### DIFF
--- a/operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
+++ b/operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
@@ -67,6 +67,7 @@ import org.keycloak.operator.crds.v2alpha1.client.KeycloakClientStatusCondition;
 import org.keycloak.operator.crds.v2alpha1.deployment.Keycloak;
 import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakStatusAggregator;
 import org.keycloak.operator.crds.v2alpha1.deployment.KeycloakStatusCondition;
+import org.keycloak.operator.crds.v2alpha1.deployment.spec.FeatureSpec;
 import org.keycloak.operator.crds.v2alpha1.deployment.spec.HttpSpec;
 import org.keycloak.representations.admin.v2.BaseClientRepresentation;
 
@@ -95,6 +96,7 @@ import static org.keycloak.operator.crds.v2alpha1.CRDUtils.isTlsConfigured;
 public abstract class KeycloakClientBaseController<R extends CustomResource<? extends KeycloakClientSpec<S>, KeycloakClientStatus>, T extends BaseClientRepresentation, S extends BaseClientRepresentation>
         implements Reconciler<R>, Cleaner<R> {
 
+    public static final String CLIENT_ADMIN_API_V2 = "client-admin-api:v2";
     private static final String CLIENT_API_VERSION = "v2";
     private static final String HTTPS = "https";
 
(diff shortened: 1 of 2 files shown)


operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -216,6 +233,14 @@ public DeleteControl cleanup(R resource, Context<R> context) throws Exception {
             return DeleteControl.defaultDelete();
         }
 
+        if (!hasFeatureEnabled(keycloak)) {
+            // TODO: this behavior is not very straight-forward. For now just log an error
```
shawkins, line 237: Another thing we could do is use the status hash as an indication of whether the operator was ever responsible for this client.

Or if we come to the conclusion that orphaning is not an issue in general, then we can just remove the finalizer and not worry as much about this behavior.

shawkins:
Marking as ready based upon our meeting where we agreed that simplistic assumptions are good enough for now.

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
```diff
@@ -125,12 +125,12 @@ public void testBasicSamlClientCreationAndDeletionHttp() throws InterruptedExcep
         assertNull(k8sclient.resource(client).get());
     }
 
-    @Test
+    @DisabledIfApiServerTest
```
vmuzikar, line 129: Should we keep `@Test` here?
```suggestion
    @Test
    @DisabledIfApiServerTest
```

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
```diff
@@ -125,12 +125,12 @@ public void testBasicSamlClientCreationAndDeletionHttp() throws InterruptedExcep
...
     public void testBasicOIDCClientCreationAndDeletionHttp() throws InterruptedException {
         helpTestBasicOIDCClientCreationAndDeletion(false);
     }
 
-    @Test
+    @DisabledIfApiServerTest
```
vmuzikar, line 135: Should we keep `@Test` here?
```suggestion
    @Test
    @DisabledIfApiServerTest
```

vmuzikar (dismissed):

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
```diff
@@ -197,6 +205,15 @@ public UpdateControl<R> reconcile(R resource, Context<R> context) throws Excepti
         return updateControl;
     }
 
+    // TODO: this doesn't mesh well with the current feature concept
```
vmuzikar, line 208: This would be solved with https://github.com/keycloak/keycloak/issues/28513.

operator/src/main/java/org/keycloak/operator/controllers/KeycloakClientBaseController.java
vmuzikar: As discussed offline, orphaning is fine.

operator/src/test/java/org/keycloak/operator/testsuite/integration/KeycloakClientTest.java
shawkins: Absolutely - that was a bad merge on my part.

vmuzikar (approved):
LGTM, thanks.

PR #46106, opened 2026-02-08
Fix duplicate header in VERIFY_EMAIL flow

### LINKED ISSUE #46105: `accountUpdatedMessage` shows up twice after "verify email" action
### Before reporting an issue

### Area

login/ui

### Describe the bug

Hello,

I'm not 100% sure of the area where this issue applies

It is **very similar** to #41701, it also involves the VERIFY_EMAIL action, but it is not the same, because the message involved is `accountUpdatedMessage` for some reasons.

### Possible fix?

I asked Copilot to find a fix, and it provided: https://github.com/tagliala/keycloak/pull/1

### Version

26.5.2

### Regression

### Expected behavior

Body message displayed once, with a title above

### Actual behavior

Message displayed twice, as title and body
    Image

### How to Reproduce?

1. Create a realm and configure it for email sending
2. Create an account
3. Trigger the execution of VERIFY_EMAIL action for that account ID with an API Call (`execute-actions-email` endpoint) 
    ```
    2026-02-08 17:52:21,677 DEBUG [org.keycloak.transaction.JtaTransactionWrapper] (executor-thread-1) JtaTransactionWrapper end. Request Context: HTTP PUT /admin/realms/myrealm/users/USERID/execute-actions-email
    ```

4. Click on the email link
5. Click on "Click here to proceed"
    Image

### Anything else?

_No response_

tagliala:
When executing the VERIFY_EMAIL action (for example via the admin API), the confirmation page showed "Your account has been updated." twice, once as the page header and once as the body.

The root cause is that `AuthenticationManager.finishedRequiredActions()` rendered the `ACCOUNT_UPDATED` info page without setting a messageHeader attribute.

The `info.ftl` template falls back to `message.summary` when `messageHeader` is not set, which produced the duplicated message.

Change summary:
- Add a new message key `accountUpdatedTitle` with value "Account Updated".
- Add a Messages constant (`ACCOUNT_UPDATED_TITLE`) for the new key.
- Update `AuthenticationManager` to set the `messageHeader` attribute to the message key (`ACCOUNT_UPDATED_TITLE`) instead of passing a resolved message string. Passing the key lets the template render a distinct header (localized via the message key) and keeps the body message separate.

Result:
- Header now reads "Account Updated"
- Body now reads "Your account has been updated."
- Prevents the same message appearing both in the header and the body.

Notes:
- This addresses the same underlying issue reported in #41701

Fix #46105

### Before

image

### After

image

### DIFF
--- a/themes/src/main/resources/theme/base/login/messages/messages_en.properties
+++ b/themes/src/main/resources/theme/base/login/messages/messages_en.properties
@@ -311,6 +311,7 @@ linkIdpMessage=You need to verify your email address to link your account with {
 emailSentMessage=You should receive an email shortly with further instructions.
 emailSendErrorMessage=Failed to send email, please try again later.
 
+accountUpdatedTitle=Account updated
 accountUpdatedMessage=Your account has been updated.
 accountPasswordUpdatedMessage=Your password has been updated.
 
--- a/services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
+++ b/services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
@@ -1088,6 +1088,7 @@ public class AuthenticationManager {
 
         if (authSession.getAuthNote(END_AFTER_REQUIRED_ACTIONS) != null) {
             LoginFormsProvider infoPage = session.getProvider(LoginFormsProvider.class).setAuthenticationSession(authSession)
+                    .setAttribute("messageHeader", Messages.ACCOUNT_UPDATED_TITLE)
                     .setSuccess(Messages.ACCOUNT_UPDATED);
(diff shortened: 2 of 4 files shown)


themes/src/main/resources/theme/base/login/messages/messages_en.properties
```diff
@@ -311,6 +311,7 @@ linkIdpMessage=You need to verify your email address to link your account with {
 emailSentMessage=You should receive an email shortly with further instructions.
 emailSendErrorMessage=Failed to send email, please try again later.
 
+accountUpdatedTitle=Account Updated
 accountUpdatedMessage=Your account has been updated.
```
tagliala, line 315: Probably this is required somewhere else, but I'm not familiar with the process

ahus1 (approved):

PR #46109, opened 2026-02-09
Sync client scope state between tabs in dedicated scopes

### LINKED ISSUE #46083: "Full scope allowed" toggle reverts to previous state when switching between Scope and Mappers tabs in Dedicated scopes
### Before reporting an issue

### Area

admin/ui

### Describe the bug

In the **Clients → Client details → Client scopes → Dedicated scopes** screen, when the user toggles the "Full Scope Allowed" switch (e.g., from ON → OFF) on the Scope tab, then navigates to the Mappers tab and returns back to the Scope tab, the toggle visually reverts to its previous state (ON in this example).

The change **is persisted on the server side**, a full page refresh or navigating away and returning to the Dedicated scopes screen shows the correct (updated) value. This is a **UI-only stale state issue**, likely caused by the component re-reading a cached/stale value when the Scope tab is re-mounted after tab switching.

### Version

26.5.2

### Regression

### Expected behavior

After toggling "Full Scope Allowed" on the Scope tab and switching to the Mappers tab and back, the Scope tab should reflect the updated toggle state without requiring a full page refresh.

### Actual behavior

The "Full Scope Allowed" toggle visually reverts to its previous state when the user navigates away from the Scope tab (to Mappers) and then returns. A full page reload is required to see the correct state.

### How to Reproduce?

1. Navigate to **Clients → (select any client) → Client scopes tab → (click the client's dedicated scope link).**
2. On the **Scope** tab, note the current state of the **"Full scope allowed"** toggle (e.g., ON).
3. Toggle it to the opposite state (e.g., OFF). The change is saved.
4. Switch to the **Mappers** tab.
5. Switch back to the **Scope** tab.
6. **Observe**: The "Full scope allowed" toggle shows the old state (ON), not the updated state (OFF).
7. **Refresh the page (F5)**: The toggle now correctly shows the updated state (OFF).

### Anything else?

A screen recording demonstrating the issue is attached below.
The bug is cosmetic/UI-only; the server-side state is correctly updated after the toggle. The problem is that the Scope tab component displays stale data after a tab switch.

This could be caused by the tab component not invalidating or refetching the client scope data when the Scope tab is re-mounted/re-activated.

https://github.com/user-attachments/assets/950d3f1c-d68b-4f91-8f97-385f3c460c9d

muhammedogz:
# Issue

Closes #46083

# Summary

  - Fix stale state when switching between "Mappers" and "Scope" tabs in dedicated scopes
  - Lift client state management from DedicatedScope child component to DedicatedScopes parent

# Description

## Problem

  When a user toggles the "Full Scope Allowed" switch in the Scope tab and then switches to the Mappers tab and
  back, the switch state would reset to its original value. This happened because DedicatedScope was managing its
  own internal state initialized from props, which became out of sync with the parent component.

## Solution

  Remove internal state from DedicatedScope and add an onChange callback prop to notify the parent when the client
  is updated. The parent DedicatedScopes now maintains the single source of truth for client state across all tabs.

## Changes

  - DedicatedScope.tsx: Remove useState for client, add optional onChange prop, call onChange?.() after successful
  API update
  - DedicatedScopes.tsx: Pass onChange={setClient} to sync state updates

# Test plan

  - Navigate to a client's dedicated scopes page (Clients → [Client] → Client scopes → Dedicated scopes link)
  - Go to the "Scope" tab
  - Toggle the "Full Scope Allowed" switch
  - Switch to the "Mappers" tab
  - Switch back to the "Scope" tab
  - Verify the switch state is preserved correctly

# Result Video

[client-switch-fix.webm](https://github.com/user-attachments/assets/79c46c33-92b5-4922-a6a4-9ee885c72c99)

### DIFF
--- a/js/apps/admin-ui/src/clients/scopes/DedicatedScope.tsx
+++ b/js/apps/admin-ui/src/clients/scopes/DedicatedScope.tsx
@@ -7,7 +7,6 @@ import {
   PageSection,
   Switch,
 } from "@patternfly/react-core";
-import { useState } from "react";
 import { useTranslation } from "react-i18next";
 import { HelpItem } from "@keycloak/keycloak-ui-shared";
 import { useAdminClient } from "../../admin-client";
@@ -18,17 +17,14 @@ import { useAccess } from "../../context/access/Access";
 
 type DedicatedScopeProps = {
   client: ClientRepresentation;
+  onChange?: (client: ClientRepresentation) => void;
 };
 
-export const DedicatedScope = ({
-  client: initialClient,
-}: DedicatedScopeProps) => {
+export const DedicatedScope = ({ client, onChange }: DedicatedScopeProps) => {
   const { adminClient } = useAdminClient();
 
   const { t } = useTranslation();
   const { addAlert, addError } = useAlerts();
-
-  const [client, setClient] = useState<ClientRepresentation>(initialClient);
 
   const { hasAccess } = useAccess();
   const isManager = hasAccess("manage-clients") || client.access?.manage;
@@ -70,7 +66,7 @@ export const DedicatedScope = ({
     try {
       await adminClient.clients.update({ id: client.id! }, newClient);
       addAlert(t("clientScopeSuccess"), AlertVariant.success);
-      setClient(newClient);
+      onChange?.(newClient);
     } catch (error) {
       addError("clientScopeError", error);
     }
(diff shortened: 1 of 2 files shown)


edewit (approved):
LGTM

muhammedogz:
> LGTM

Thank you!

edewit (approved):

ahus1 (approved):
Approving based on review of @edewit - thank you!

PR #46113, opened 2026-02-09
Check SubjectConfirmationData element for bearer type

### LINKED ISSUE #45646: CVE-2026-1190 - Keycloak SAML brokering: Response delay due to unchecked NotOnOrAfter in SubjectConfirmationData
### Source: https://issues.redhat.com/browse/RHBK-4110
Security Tracking Issue

Flaw:
-----

Keycloak SAML brokering: Response delay due to unchecked NotOnOrAfter in SubjectConfirmationData

When SAML is configured to act as a client (SAML brokering) it does not check the NotOnOrAfter tat is defined inside the SubjectConfirmationData. This just affects in the sense that an attacker can delay the response for more of the expected time. The response cannot be tampered because SAML should configured to be signed and, therefore, the same response cannot be used twice or to respond another request (the id should be modified and therefore the signature validation fails).

---
*This issue was originally tracked in the private repository as #321. Migrated via lift-embargo process.*

rmartinc:
Closes #45646

The `SubjectConfirmationData` is now checked for the bearer type in all the attributes. There is a new class `SubjectConfirmationDataValidator` that validates the `notBefore`, `notOnOrAfter`, `inResponseTo` and `recipient` if they are present in the confirmation. Test added.

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_5_3.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_5_3.adoc
@@ -1,0 +1,12 @@
+== Breaking changes
+
+Breaking changes are identified as those that might require changes for existing users to their configurations or applications.
+In minor or patch releases, {project_name} will only introduce breaking changes to fix bugs.
+
+=== The SAML broker and adapter now check the SubjectConfirmationData element for the bearer type
+
+Now {project_name}, when acting as a SAML Service Provider (SP) in identity brokering or in the adapter, validates the `SubjectConfirmationData` for the type `urn:oasis:names:tc:SAML:2.0:cm:bearer` defined in the standard. The elements `NotBefore`, `NotOnOrAfter` and `Recipient`, when present in the assertion, are checked to be in the valid time range and to be the correct destination URI respectively. Previously, similar values were checked for other parts of the SAML response (for example the `Conditions` element or the `destination` attribute).
+
+To prepare for the upgrade, verify that the Identity Provider or adapter configurations allow for a sufficient clock skew for the time attributes.
+
+If you see any issue after the upgrade related to this element, please configure the external IdP to set the correct values for the subject confirmation element.
--- a/adapters/saml/core/src/main/java/org/keycloak/adapters/saml/profile/AbstractSamlAuthenticationHandler.java
+++ b/adapters/saml/core/src/main/java/org/keycloak/adapters/saml/profile/AbstractSamlAuthenticationHandler.java
@@ -54,6 +54,8 @@ import org.keycloak.dom.saml.v2.assertion.AttributeType;
 import org.keycloak.dom.saml.v2.assertion.AuthnStatementType;
 import org.keycloak.dom.saml.v2.assertion.NameIDType;
 import org.keycloak.dom.saml.v2.assertion.StatementAbstractType;
+import org.keycloak.dom.saml.v2.assertion.SubjectConfirmationDataType;
+import org.keycloak.dom.saml.v2.assertion.SubjectConfirmationType;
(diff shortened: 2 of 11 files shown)


mposolda (dismissed):
@rmartinc LGTM. Approving. Hope this can be merged if tests are ok

ahus1:
@rmartinc - this IMHO needs documentation in the migration guide, as those checks might fail on existing setups. 

Given the other PR, I assume that if people don't re-configure their existing IDP configurations, they will fail due to Keycloak setting that value with the current timestamp. Ideally, this is a change people would need to do before upgrading Keycloak, as this might break their login-via-IDP setups when they upgrade, and they might actually struggle to log into Keycloak after the upgrade.

rmartinc:
@ahus1 The SAML response is consumed right after IdP generates the assertion, just the redirect or the post time, so the `notBefore` or `notOnOrAfter` should be OK except if the IdP is really setting a wrong value (but similar values are in the conditions element that are previously checked). The `recipient` is a bit different and I doubted about including it by default or not, but finally I decided that, if the IdP includes it, it can only use the destination, so I think we are OK too. In general, I think that we are safe except in very weird conditions (like bad synch IdPs that does not use conditions). But I agree that adding a note in the upgrading does not harm at all. I'm going to add it is for 26.5.3 for the moment, I can move to the next one if 26.5.3 is tagged before.

docs/documentation/upgrading/topics/changes/changes-26_5_3.adoc
```diff
@@ -0,0 +1,8 @@
+// ------------------------ Breaking changes ------------------------  //
+== Notable changes
+
+Notable changes may include internal behavior changes that prevent common misconfigurations, bugs that are fixed, or changes to simplify running {project_name}.
+
+=== The SAML broker and adapter now check the SubjectConfirmationData element for the bearer type
+
+Now {project_name}, when acting as a SAML Service Provider (SP) in identity brokering or in the adapter, validates the `SubjectConfirmationData` for the type `urn:oasis:names:tc:SAML:2.0:cm:bearer` defined in the standard. The elements `NotBefore`, `NotOnOrAfter` and `Recipient`, when present in the assertion, are checked to be in the valid time range and to be the correct destination URI respectively. Previously, similar values were checked for other parts of the SAML response (for example the `Conditions` element or the `destination` attribute). If you see any issue after the upgrade related to this element, please configure the external IdP to set the correct values for the subject confirmation element.
```
ahus1:
```suggestion
Now {project_name}, when acting as a SAML Service Provider (SP) in identity brokering or in the adapter, validates the `SubjectConfirmationData` for the type `urn:oasis:names:tc:SAML:2.0:cm:bearer` defined in the standard. The elements `NotBefore`, `NotOnOrAfter` and `Recipient`, when present in the assertion, are checked to be in the valid time range and to be the correct destination URI respectively. Previously, similar values were checked for other parts of the SAML response (for example the `Conditions` element or the `destination` attribute). 

To prepare for the upgrade, ensure that IdP or adapter configurations allow for a sufficient clock skew configured.

If you see any issue after the upgrade related to this element, please configure the external IdP to set the correct values for the subject confirmation element.
```

ahus1 (commented):
@rmartinc - I see you marked it as a notable change, while it can be breaking for some. 

I would like to point people to configure a clock skew, probably even before they upgrade. Not sure about the wording, see below for a suggestion.

docs/documentation/upgrading/topics/changes/changes-26_5_3.adoc
rmartinc: Done!

ahus1 (approved):

Issue #29556, opened 2024-05-15
Username is not printed in the logs instead UserId is getting printed


jsto:
### Description

Hello.

We request an enhancement to Keycloak's admin events logging capability. Currently, while Keycloak provides extensive capabilities for logging administrative events, there is a need for further improvement to meet the requirements outlined by the Indian government's IT Security Audit and Review (ITSAR) guidelines: [](https://nccs.gov.in/public/downloads_itsar/ITSAR404042308.pdf) (please see Section 5 User Audit, pages 28-31). 

Following these requirements, the admin username is not printed in the Keycloak logs instead, UserId is getting printed. The following admin events must comply with this specific requirement:
- Super Admin User creating an Admin user
- Admin User creating a read-only/least privileged user
- Admin rights removed from an "Admin User"

Given the increasing emphasis on data security and compliance, we believe that implementing this enhancement is rather urgent by aligning with ITSAR guidelines.

### Motivation

Official document of Indian Government (see Section 5 User Audit, pages 28-31)
https://nccs.gov.in/public/downloads_itsar/ITSAR404042308.pdf

ahrycej:
This is an important request, would like to have it implemented.

ahrycej:
Hello there, comment to old issue. Yes can this be that generated log has username so that it can be passed to the centralized SIEM. Console is not enough.

ahrycej:
Facing issues in India with govermental requirements.

PR #46124, opened 2026-02-09
Add v2 admin event support to Client Admin API

edewit:
## Summary

- Add `AdminEventV2Builder` class for creating v2 admin events with an `apiVersion=v2` detail marker
- Modify `DefaultClientService` to fire v2 admin events when clients are created/updated
- Pass `AdminAuth` through the API chain so events have proper authentication context
- Add tests verifying v2 events are fired with correct operation type and representation format

This allows consumers to distinguish v2 API events from v1 events in the admin event log.

## Test plan

- [x] Run `AdminEventV2Test` to verify v2 events are fired correctly
- [x] Verify events contain `apiVersion=v2` in details
- [x] Verify v2 representation format is used in event representation

Closes #46123
Part of #45932

### DIFF
--- a/rest/admin-v2/api/src/main/java/org/keycloak/events/admin/v2/AdminEventV2Builder.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/events/admin/v2/AdminEventV2Builder.java
@@ -16,18 +16,15 @@
  */
 package org.keycloak.events.admin.v2;
 
-import java.io.IOException;
+
+import jakarta.ws.rs.core.UriInfo;
 
 import org.keycloak.common.ClientConnection;
-import org.keycloak.events.admin.AdminEvent;
-import org.keycloak.events.admin.ResourceType;
-import org.keycloak.models.ClientModel;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
-import org.keycloak.models.UserModel;
+import org.keycloak.models.utils.StripSecretsUtilsV2;
 import org.keycloak.services.resources.admin.AdminAuth;
 import org.keycloak.services.resources.admin.AdminEventBuilder;
-import org.keycloak.util.JsonSerialization;
 
 /**
  * Builder for Admin API v2 events.
@@ -37,7 +34,6 @@ import org.keycloak.util.JsonSerialization;
  * - Representation is serialized without stripping secrets (v2 representations handle this differently).
  */
 public class AdminEventV2Builder extends AdminEventBuilder {
-
     private static final String API_VERSION_DETAIL_KEY = "apiVersion";
     private static final String API_VERSION_V2 = "v2";
 
@@ -47,64 +43,15 @@ public class AdminEventV2Builder extends AdminEventBuilder {
         detail(API_VERSION_DETAIL_KEY, API_VERSION_V2);
     }
 
-    private AdminEventV2Builder(RealmModel realm, AdminAuth auth, KeycloakSession session, String ipAddress, AdminEvent adminEvent) {
-        super(realm, auth, session, ipAddress, adminEvent);
-        // The v2 detail is already in the adminEvent copy
-    }
-
-    /**
-     * Create a new instance of the {@link AdminEventV2Builder} that is bound to a new session.
-     * Use this when starting, for example, a nested transaction.
-     * @param session new session where the {@link AdminEventV2Builder} should be bound to.
-     * @return a new instance of {@link AdminEventV2Builder}
-     */
     @Override
-    public AdminEventV2Builder clone(KeycloakSession session) {
-        RealmModel newEventRealm = session.realms().getRealm(realm.getId());
-        RealmModel newAuthRealm = session.realms().getRealm(this.auth.getRealm().getId());
-        UserModel newAuthUser = session.users().getUserById(newAuthRealm, this.auth.getUser().getId());
-        ClientModel newAuthClient = session.clients().getClientById(newAuthRealm, this.auth.getClient().getId());
-
-        return new AdminEventV2Builder(
-                newEventRealm,
-                new AdminAuth(newAuthRealm, this.auth.getToken(), newAuthUser, newAuthClient),
-                session,
-                ipAddress,
-                adminEvent
-        );
+    protected String getResourcePath(UriInfo uriInfo) {
+        String path = uriInfo.getPath();
+        String realmRelative = "/admin/api/%s/".formatted(realm.getName());
+        return path.substring(path.indexOf(realmRelative) + realmRelative.length());
     }
 
     @Override
-    public AdminEventV2Builder resource(ResourceType resourceType) {
-        super.resource(resourceType);
-        return this;
-    }
-
-    @Override
-    public AdminEventV2Builder resource(String resourceType) {
-        super.resource(resourceType);
-        return this;
-    }
-
-    /**
-     * Sets the v2 representation to be included in the event.
-     * The representation is serialized to JSON without stripping secrets
-     * (v2 representations handle sensitive data differently).
-     * 
-     * @param value the v2 representation object (e.g., BaseClientRepresentation)
-     * @return this builder
-     */
-    @Override
-    public AdminEventV2Builder representation(Object value) {
-        if (value == null || value.equals("")) {
-            return this;
-        }
-
-        try {
-            adminEvent.setRepresentation(JsonSerialization.writeValueAsString(value));
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        }
-        return this;
+    protected void stripSecretsFromRepresentation(Object value) {
+        StripSecretsUtilsV2.stripSecrets(session, value);
     }
 }
--- a/rest/admin-v2/api/src/main/java/org/keycloak/events/admin/v2/AdminEventV2Builder.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/events/admin/v2/AdminEventV2Builder.java
@@ -1,0 +1,301 @@
+/*
+ * Copyright 2024 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.events.admin.v2;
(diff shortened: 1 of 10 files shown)


rest/admin-v2/api/src/main/java/org/keycloak/events/admin/v2/AdminEventV2Builder.java
```diff
@@ -0,0 +1,301 @@
...
+import org.jboss.logging.Logger;
+
+/**
+ * Builder for Admin API v2 events.
+ * <p>
+ * Similar to the v1 AdminEventBuilder but designed to work with v2 representations.
+ * Events are dispatched through the same EventListenerProvider mechanism as v1,
+ * but with an "apiVersion" detail set to "v2" to distinguish them from v1 events.
+ */
+public class AdminEventV2Builder {
```
mabartos: I think extending the AdminEventBuilder v1 makes more sense than this duplication.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
```diff
@@ -42,13 +43,13 @@ public class DefaultClientApi implements ClientApi {
 
     private static final ObjectMapper MAPPER = new ObjectMapperResolver().getContext(null);
 
-    public DefaultClientApi(KeycloakSession session, RealmAdminResource realmAdminResource, ClientResource clientResource, String clientId) {
+    public DefaultClientApi(KeycloakSession session, RealmAdminResource realmAdminResource, ClientResource clientResource, String clientId, AdminAuth auth) {
```
mabartos: Can you propagate only the AdminPermissionEvaluator, instead of the AdminAuth? As seen here: https://github.com/keycloak/keycloak/pull/46056/changes#diff-8cfff4c302edd028714d7e1cb691d336cbd0c8cca6b9b177a7cb8bf7360f091dR50

The permissions will be needed elsewhere and we can reduce the count of propagated params. 

You can get the AdminAuth as `AdminPermissionEvaluator.adminAuth()`

mabartos (changes requested):
@edewit Thanks for the PR. Added some comments.

mabartos:
@edewit Nice job!

I've created a PR to your fork with some suggestions (described in the PR description):

- https://github.com/edewit/keycloak/pull/14

rest/admin-v2/api/src/main/java/org/keycloak/models/utils/StripSecretsUtilsV2.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static <T> T stripSecrets(KeycloakSession session, T representation) {
+        return stripSecrets(session, representation, REPRESENTATION_FORMATTER);
+    }
+
+    protected static OIDCClientRepresentation stripOidcClient(OIDCClientRepresentation rep) {
+        Optional.ofNullable(rep.getAuth())
+                .map(OIDCClientRepresentation.Auth::getSecret)
+                .filter(StringUtil::isNotBlank)
+                .ifPresent(secret -> rep.getAuth().setSecret(maskNonVaultValue(secret)));
```
mabartos, line 32: @edewit Unfortunately, this updates the whole representation (by reference), so even the returned OIDC client via POST/PUT contains the masked secret... the masked secret should be only for the event. So, we'd need to clone the representation and work with a different object. 

That's why the Admin V2 test fails.

rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -106,9 +119,26 @@ public CreateOrUpdateResult createOrUpdate(RealmModel realm, BaseClientRepresent
         }
         var updated = mapper.fromModel(model);
 
+        // Fire v2 admin event (in parallel to v1 events fired by clientsResource/clientResource)
+        fireAdminEvent(created ? OperationType.CREATE : OperationType.UPDATE, updated);
+
         return new CreateOrUpdateResult(updated, created);
```
mabartos:
```suggestion

        // Fire v2 admin event (in parallel to v1 events fired by clientsResource/clientResource)
        fireAdminEvent(created ? OperationType.CREATE : OperationType.UPDATE, mapper.fromModel(model));

        return new CreateOrUpdateResult(mapper.fromModel(model), created);
```

@edewit We can just return a different object/reference to the admin event. These operations should not be resource expensive, so we should be ok to do it.

rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -104,9 +117,25 @@ public CreateOrUpdateResult createOrUpdate(RealmModel realm, BaseClientRepresent
         if (client instanceof OIDCClientRepresentation oidcClient) {
             handleServiceAccount(model, oidcClient);
         }
-        var updated = mapper.fromModel(model);
 
-        return new CreateOrUpdateResult(updated, created);
+        // Fire v2 admin event (in parallel to v1 events fired by clientsResource/clientResource)
+        fireAdminEvent(created ? OperationType.CREATE : OperationType.UPDATE, mapper.fromModel(model));
+
+return new CreateOrUpdateResult(updated, mapper.fromModel(model));
```
mabartos:
```suggestion
        fireAdminEvent(created ? OperationType.CREATE : OperationType.UPDATE, mapper.fromModel(model));

        return new CreateOrUpdateResult(mapper.fromModel(model), created);
```

rest/admin-v2/api/src/main/java/org/keycloak/models/utils/StripSecretsUtilsV2.java
mabartos: @edewit Resolved in added commits:
- https://github.com/keycloak/keycloak/pull/46124/changes/276f5c867fce338329f47b92afb030ac19567213
- https://github.com/keycloak/keycloak/pull/46124/changes/059fe509801d38994e31aa0282b30cd00c88221f

mabartos (dismissed):

mabartos:
@keycloak/cloud-native Do you want to review it as well? Thanks!

mabartos:
@edewit Conflict

vmuzikar:
> This allows consumers to distinguish v2 API events from v1 events in the admin event log.

Does this mean we currently fire v1 and v2 events from v2 endpoints as v2 endpoints still use v1 as a backend?

edewit:
rebased

mabartos:
@vmuzikar We need some changes here. I'll provide them. On hold for now.

mabartos:
@vmuzikar Should be ready for a review.

rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -138,9 +148,26 @@ public CreateOrUpdateResult createOrUpdate(RealmModel realm, BaseClientRepresent
...
-        var updated = mapper.fromModel(model);
 
-        return new CreateOrUpdateResult(updated, created);
+        fireAdminEvent(created ? OperationType.CREATE : OperationType.UPDATE, mapper.fromModel(model));
+
+        return new CreateOrUpdateResult(mapper.fromModel(model), created);
+    }
+
+    /**
+     * Fires a v2 admin event for client operations (only enabled for testing now to avoid duplicated admin events)
```
vmuzikar, line 170: > only enabled for testing now to avoid duplicated admin events

+1

rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
```diff
@@ -138,9 +148,26 @@ public CreateOrUpdateResult createOrUpdate(RealmModel realm, BaseClientRepresent
         if (client instanceof OIDCClientRepresentation oidcClient) {
             handleServiceAccount(model, oidcClient);
         }
-        var updated = mapper.fromModel(model);
 
-        return new CreateOrUpdateResult(updated, created);
+        fireAdminEvent(created ? OperationType.CREATE : OperationType.UPDATE, mapper.fromModel(model));
+
+        return new CreateOrUpdateResult(mapper.fromModel(model), created);
```
vmuzikar, line 166: Why to call `mapper.fromModel(model)` twice?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/AdminEventV2Test.java
```diff
@@ -0,0 +1,300 @@
...
+            assertThat("V2 event should be present with apiVersion=v2 detail", v2Event, notNullValue());
+            assertThat("V2 event should have CREATE operation", v2Event.getOperationType(), is(OperationType.CREATE.toString()));
+            assertThat("V2 event should have resource path relative to API v2", v2Event.getResourcePath(), is("clients/v2"));
+        } finally {
+            deleteTestClient();
+        }
+    }
+
+    @Test
+    public void updateClientFiresV2Event() throws Exception {
```
vmuzikar, line 137: We should also check patch.

shawkins:
Related to #29556 and #46531 the fields tracked AuthDetails include clientId and userId - both other which seem to be the UUID - but not the logical "client id" nor username. Having only the UUID for the client will be a problem with the current v2 design.

I'm not sure if in general we were avoiding including something like the username in logs.

cc @vmuzikar

rest/admin-v2/api/src/main/java/org/keycloak/services/client/DefaultClientService.java
mabartos: As mentioned offline, we can improve it in a follow-up:
- https://github.com/keycloak/keycloak/issues/46553

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/AdminEventV2Test.java
mabartos: Done here: https://github.com/keycloak/keycloak/pull/46124/changes/3b1597375e03a001d4bb734e10b44d3d61b129c3

vmuzikar:
> Having only the UUID for the client will be a problem with the current v2 design.

Yes, we probably don't want to change the eventing paradigm to not use UUIDs. One more argument for exposing them in v2 too.

shawkins:
> Yes, we probably don't want to change the eventing paradigm to not use UUIDs. One more argument for exposing them in v2 too.

Agreed, but I'm also wondering about capturing the names. For some reason we have both the realm name and UUID, but only the UUIDs for clients and users.

mabartos:
> Agreed, but I'm also wondering about capturing the names. For some reason we have both the realm name and UUID, but only the UUIDs for clients and users.

@shawkins I'm not against providing the clientId to the admin events for the Client API v2 (not sure for user now, as it can be username, or email). I'm not aware of any major attack that would cause showing the clientId in the event. But, we can do it in follow-up tasks. 

@vmuzikar Can you re-review this PR?

vmuzikar (approved):
LGTM, thanks.

PR #46131, opened 2026-02-09
[Test Framework] testcontainers dependency refactor.

### LINKED ISSUE #45884: Testframework core has dependency on testcontainers
AbstractContainerTestDatabase and InfinispanExternalServer are in testframework core resulting in a need to have a dependency on testcontainers.

Additionally AbstractDatabaseSupplier has a hack that detects if both Keycloak and database are running in a container to then change host/port in the jdbc url. This is rather messy and we should have a better approach for this.

lhanusov:
Closes #45884


test-framework/core/src/main/java/org/junit/rules/TestRule.java
stianst, line 1: Irrelevant change to this PR, please send a separate PR for this

test-framework/core/src/main/java/org/junit/runners/model/Statement.java
stianst, line 1: Irrelevant change to this PR, please send a separate PR for this

test-framework/core/src/main/java/org/keycloak/testframework/database/AbstractDatabaseSupplier.java
```diff
@@ -1,18 +1,14 @@
...
 import org.keycloak.testframework.injection.RequestedInstance;
 import org.keycloak.testframework.injection.Supplier;
 import org.keycloak.testframework.injection.SupplierHelpers;
 import org.keycloak.testframework.injection.SupplierOrder;
-import org.keycloak.testframework.server.KeycloakServer;
-import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
-import org.keycloak.testframework.server.KeycloakServerConfigInterceptor;
 
-public abstract class AbstractDatabaseSupplier implements Supplier<TestDatabase, InjectTestDatabase>, KeycloakServerConfigInterceptor<TestDatabase, InjectTestDatabase> {
+public abstract class AbstractDatabaseSupplier implements Supplier<TestDatabase, InjectTestDatabase> {
```
stianst: All databases need to intercept the server config; I'd keep intercept here, but without the `if cluster` hack. I see you added it to remote, but there's also dev-file and dev-mem, so better to just keep it here.

Alternatively you can add the intercept to dev-file and dev-mem since they are the only ones that do it the plain way; but you should at least leave the implements KeycloakServerConfigInterceptor here, since it's abstract it just means a concrete class must implement it.

test-framework/core/src/main/java/org/keycloak/testframework/database/RemoteDatabaseSupplier.java
```diff
@@ -4,8 +4,9 @@
 import org.keycloak.testframework.config.Config;
 import org.keycloak.testframework.injection.InstanceContext;
 import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
+import org.keycloak.testframework.server.KeycloakServerConfigInterceptor;
 
-public class RemoteDatabaseSupplier extends AbstractDatabaseSupplier {
+public class RemoteDatabaseSupplier extends AbstractDatabaseSupplier implements KeycloakServerConfigInterceptor<TestDatabase, InjectTestDatabase> {
```
stianst: Remove `implements KeycloakServerConfigInterceptor` should come from `AbstractDatabaseSupplier`

test-framework/test-containers/src/main/java/org/keycloak/testframework/database/AbstractContainerDatabaseSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.keycloak.testframework.annotations.InjectTestDatabase;
+import org.keycloak.testframework.config.Config;
+import org.keycloak.testframework.injection.InstanceContext;
+import org.keycloak.testframework.injection.Supplier;
+import org.keycloak.testframework.server.KeycloakServer;
+import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
+import org.keycloak.testframework.server.KeycloakServerConfigInterceptor;
+
+public abstract class AbstractContainerDatabaseSupplier extends AbstractDatabaseSupplier implements Supplier<TestDatabase, InjectTestDatabase>, KeycloakServerConfigInterceptor<TestDatabase, InjectTestDatabase> {
```
stianst: No need to add `implements *` that should come from `AbstractDatabaseSupplier`

test-framework/test-containers/src/main/java/org/keycloak/testframework/logging/JBossContainerLogConsumer.java
stianst, line 1: Maybe we should rename this to `JBossContainerLogConsumer` to make it explict?

stianst (changes requested):
Overall looks great, added some more minor comments. The most important is that now `dev-file` and `dev-mem` doesn't intercept the server config, so both will just end up using the default Keycloak DB config instead of actually configuring it.

test-framework/core/src/main/java/org/junit/rules/TestRule.java
lhanusov: this is JUnit4 workaround introduced by the old testcontainers dependency on version 1, so I thought it's ok to remove it also, will create a separate PR for that

stianst (dismissed):

stianst (approved):

PR #46133, opened 2026-02-09
Make rolling updates for patch releases fully supported and Updated docs, release notes and upgrading guide for zero-downtime patch releases

ruchikajha95:
Closes #45381
Closes #45756

### DIFF
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -147,7 +147,7 @@ public class Profile {
         LOGOUT_ALL_SESSIONS_V1("Logout all sessions logs out only regular sessions", Type.DEPRECATED, 1),
 
         ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1),
-        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.PREVIEW, 2),
+        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.DEFAULT, 2),
 
         WORKFLOWS("Workflows", Type.PREVIEW),
 
@@ -312,7 +312,8 @@ public class Profile {
         }
     }
 
-    private static final Set<String> ESSENTIAL_FEATURES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(Feature.HOSTNAME_V2.getUnversionedKey())));
+    private static final Set<String> ESSENTIAL_FEATURES = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList(Feature.HOSTNAME_V2.getUnversionedKey(), Feature.ROLLING_UPDATES_V2.getUnversionedKey())));
 
     private static final Logger logger = Logger.getLogger(Profile.class);
 
--- a/common/src/main/java/org/keycloak/common/Profile.java
+++ b/common/src/main/java/org/keycloak/common/Profile.java
@@ -146,7 +146,7 @@ public class Profile {
 
         LOGOUT_ALL_SESSIONS_V1("Logout all sessions logs out only regular sessions", Type.DEPRECATED, 1),
 
-        ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1),
+        ROLLING_UPDATES_V1("Rolling Updates", Type.DEPRECATED, 1),
         ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.DEFAULT, 2),
 
         WORKFLOWS("Workflows", Type.PREVIEW),
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/UpdateCommandDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/UpdateCommandDistTest.java
@@ -55,6 +55,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.quarkus.test.junit.main.Launch;
 import org.junit.jupiter.api.Test;
 
+import static org.keycloak.infinispan.compatibility.CachingEmbeddedMetadataProvider.majorMinorOf;
 import static org.keycloak.it.cli.dist.Util.createTempFile;
 import static org.keycloak.quarkus.runtime.configuration.compatibility.DatabaseCompatibilityMetadataProvider.UNSUPPORTED_CHANGE_SET_HASH_KEY;
 
@@ -66,12 +67,6 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 @DistributionTest
 @RawDistOnly(reason = "Requires creating JSON file to be available between containers")
 public class UpdateCommandDistTest {
-
-    @Test
-    @Launch({UpdateCompatibility.NAME, UpdateCompatibilityMetadata.NAME, "--features-disabled=rolling-updates"})
-    public void testFeatureNotEnabled(CLIResult cliResult) {
-        cliResult.assertError("Unable to use this command. None of the versions of the feature 'rolling-updates' is enabled.");
-    }
 
     @Test
     @Launch({UpdateCompatibility.NAME})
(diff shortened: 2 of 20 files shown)


ahus1 (commented):
Thank you for this PR. I've done some re-ordering of the docs along the following guidelines: 

* The documentation in the operator and server configuration should state the current behavior. I would like to avoid comparing it to any previous behavior to avoid an unnecessary complex documentation. If something "just" works and is the default, this can be reduced for the sake of simplicity. It is usually OK to be opinionated, again for the sake of simplicity. The concept of an essential feature is something we don't explain in the docs, and I find it rather confusing, so I removed those references.
* The Release Notes (`26_6_0.adoc`) should contain all highlights people should be positively excited about. Zero-downtime patches would be part of that (at least a short version), but Infinispan 16 shouldn't really matter to them when reading the release notes
* The Upgrading Guide (`changes-26_6_0.adoc`) is the document everyone should read when upgrading. This is referenced at the beginning of the final release notes, and no need to point users there on each item in the release notes. This assumes people are upgrading from a previous version of Keycloak, therefore it references the now-default feature. It also references the update of Infinispan, as people my be affected by that. I chose both items to be "noteable", that is: they are non-breaking changes. The zero-downtime upgrades should work like previous rolling updates. 
* The Infinispan 16 upgrade of an external Infinispan might be considered breaking by some people. We briefly tested it with 15, so it might work, but we would still recommend people to upgrade. Maybe we'll move it to the breaking changes section eventually. 
* There is already Infinispan 16.1 out, but we don't plan to upgrade to that version soon. With that I changed the docs to state 16.0 and not 16.x as people might otherwise upgrade to 16.1. 

Please review my changes if they make sense and if there is anything that I might have missed that needs alignment.

Thanks!

ahus1:
Rebased due to conflicts in the release notes.

common/src/main/java/org/keycloak/common/Profile.java
```diff
@@ -146,8 +146,8 @@ public enum Feature {
 
         LOGOUT_ALL_SESSIONS_V1("Logout all sessions logs out only regular sessions", Type.DEPRECATED, 1),
 
-        ROLLING_UPDATES_V1("Rolling Updates", Type.DEFAULT, 1),
-        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.PREVIEW, 2),
+        ROLLING_UPDATES_V1("Rolling Updates", Type.DEPRECATED, 1),
+        ROLLING_UPDATES_V2("Rolling Updates for patch releases", Type.DEFAULT, 2),
```
ahus1, line 150: I did some testing around essential features and how they behave. It is actually a bit different from what I expected: 

* If a feature is essential, you can not disable it (with features disabled flag)
* If a feature is essential and when it has two versions (V1 and V2 in this case), you can still choose V1 and will then get a deprecation warning. 

While this is different from what we expected (we wanted to always force V2), maybe this is good enough. And I wouldn't stress it in the docs that there is still V1.

tkyjovsk:
The [`docs/guides/server/features.adoc`](https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/docs/guides/server/features.adoc?plain=1#L23-L26) is using `rolling-updates:v2` as an example of a feature to be enabled. Since the feature is now enabled by default and essential we should probably use a different feature as an example.

pruivo (changes requested):
* `org.keycloak.quarkus.runtime.cli.command.AbstractUpdatesCommand`, the code below can be removed (dead-code) and the method `printPreviewWarning()` too. 
```java
            if (!Profile.isAnyVersionOfFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V1)) {
                printFeatureDisabled();
                return Optional.of(FEATURE_DISABLED_EXIT_CODE);
            }
```

* `org.keycloak.infinispan.compatibility.CachingEmbeddedMetadataProvider`, `Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)` is always true; the code can be simplified

* `org.keycloak.compatibility.KeycloakCompatibilityMetadataProvider`, `!Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)` can be removed.

* `org.keycloak.services.resources.ThemeResource`, the code can be simplified by removing `Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)` as it is always true. 

* `org.keycloak.testsuite.theme.ThemeResourceProviderTest`, we can remove `@EnableFeatures(@EnableFeature(Profile.Feature.ROLLING_UPDATES_V2))`

* `org.keycloak.compatibility.KeycloakCompatibilityMetadataProviderTest`, the code snippet below can be removed.
```java
        // Enable V2 feature
        Profile.configure(new ProfileConfigResolver() {
            @Override
            public Profile.ProfileName getProfileName() {
                return null;
            }

            @Override
            public FeatureConfig getFeatureConfig(String feature) {
                return Profile.Feature.ROLLING_UPDATES_V2.getVersionedKey().equals(feature) ? FeatureConfig.ENABLED : FeatureConfig.UNCONFIGURED;
            }
        });
```

tkyjovsk:
In:
- `docs/guides/high-availability/examples/generated/keycloak-ispn.yaml`
- `docs/guides/high-availability/examples/generated/keycloak.yaml`

the rolling-updates feature can be removed from the Keycloak CRD spec:
```
  features:
    enabled:
      - rolling-updates:v2
      - multi-site # <4>
```

tkyjovsk:
In: `docs/guides/ui-customization/themes.adoc`:

_"When the preview feature `rolling-updates:v2` is enabled, this allows for a more seamless rolling upgrade."_

Should be removed.

pruivo:
A couple more places:

`When the preview feature rolling-updates:v2 is enabled` can be removed from here:
https://github.com/keycloak/keycloak/blob/7e20b8713633fa8ef9e600a92e814f550ae6f968/docs/guides/ui-customization/themes.adoc?plain=1#L138

Can be removed:
https://github.com/keycloak/keycloak/blob/2c6f56acdc36ae432c743191ef324fbd5fb6d34d/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/UpdateCommandDistTest.java#L201
and
https://github.com/keycloak/keycloak/blob/2c6f56acdc36ae432c743191ef324fbd5fb6d34d/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/UpdateCommandDistTest.java#L213

ruchikajha95:
@pruivo @tkyjovsk  thanks for further reviewing the changes i will work on the changes and push it.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/UpdateCommandDistTest.java
```diff
@@ -355,9 +354,20 @@ private Map<String, Map<String, String>> defaultMeta(KeycloakDistribution distri
...
+        // Ensure Profile is configured to pick up feature settings
+        Profile.configure();
+
         Map<String, String> m = new HashMap<>();
-        m.put("version", org.infinispan.commons.util.Version.getVersion());
-        m.put("jgroupsVersion", org.jgroups.Version.printVersion());
+        String ispnVersion = org.infinispan.commons.util.Version.getVersion();
+        String jgroupsVersion = org.jgroups.Version.printVersion();
+
+        if (Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
```
pruivo: don't need to check if the profile is enabled.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/UpdateCommandDistTest.java
```diff
@@ -355,9 +354,20 @@ private Map<String, Map<String, String>> defaultMeta(KeycloakDistribution distri
...
+            ispnVersion = majorMinorOf(ispnVersion);
+            jgroupsVersion = majorMinorOf(jgroupsVersion);
+        }
+
+        m.put("version", ispnVersion);
+        m.put("jgroupsVersion", jgroupsVersion);
```
pruivo: The feature is always enabled
```suggestion
        Map<String, String> m = new HashMap<>();
        m.put("version", majorMinorOf(org.infinispan.commons.util.Version.getVersion()));
        m.put("jgroupsVersion", majorMinorOf(org.jgroups.Version.printVersion()));
```

pruivo (changes requested):

services/src/main/java/org/keycloak/services/resources/ThemeResource.java
```diff
@@ -92,7 +91,7 @@ public class ThemeResource {
     public Response getResource(@PathParam("version") String version, @PathParam("themeType") String themeType, @PathParam("themeName") String themeName, @PathParam("path") String path, @HeaderParam(HttpHeaders.IF_NONE_MATCH) String etag, @Context UriInfo uriInfo) {
         final Optional<Theme.Type> type = getThemeType(themeType);
 
-        if (!version.equals(Version.RESOURCES_VERSION) && !Profile.isFeatureEnabled(Profile.Feature.ROLLING_UPDATES_V2)) {
+        if (!version.equals(Version.RESOURCES_VERSION)) {
             return Response.status(Response.Status.NOT_FOUND).build();
         }
```
pruivo: This if condition should be removed.

ruchikajha95:
@ahus1 @pruivo @tkyjovsk  Thanks for review . If  you feel changes are OK? can i move it to ready to review?

tkyjovsk:
The Features guide still needs to be updated, see my previous comment: https://github.com/keycloak/keycloak/pull/46133#issuecomment-3884396072

I think we should replace it with another feature, ideally some which has more versions, to showcase how to enable a specific version.

ahus1:
> The Features guide still needs to be updated

Can we handle this as a follow-up? This PR has grown already to be quite big.

ahus1 (approved):

pruivo (approved):

PR #46134, opened 2026-02-09
admin events to log event realm instead of auth realm

### LINKED ISSUE #46080: Admin Events log the authRealm instead of the event realm
### Before reporting an issue

### Area

core

### Describe the bug

When logging admin events in https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/events/log/JBossLoggingEventListenerProvider.java#L153, Keycloak logs the realmId and realmName of the *AuthDetails*, rather than those of the realm itself.

### Version

26.5.2

### Regression

### Expected behavior

Keycloak should log the realm the event happened in, rather than the auth realm, or log both, but explicitly mark the authRealm as an authRealm. This would be consistent with User Events.

Maybe a log line could look like this:
``
 INFO  [org.keycloak.events] (executor-thread-35) operationType="CREATE", authRealmId="b31ce2a8-2b57-41fb-adfe-dfa270e8af51", authRealmName="master", clientId="babfa0d1-351f-4220-97ad-f16c7c424991", userId="6aa77a7c-a13c-4e20-a9ca-d21d9840c76b", ipAddress="192.168.127.1", resourceType="USER", resourcePath="users/a13130ae-ab37-47bd-88a5-de157e5da4c2", realm="my-custom-realm", realmId="my-custom-realm-id"
``

### Actual behavior

This log line is produced when using the master admin console to delete a user in a non-master realm:

``
 INFO  [org.keycloak.events] (executor-thread-35) operationType="CREATE", realmId="b31ce2a8-2b57-41fb-adfe-dfa270e8af51", realmName="master", clientId="babfa0d1-351f-4220-97ad-f16c7c424991", userId="6aa77a7c-a13c-4e20-a9ca-d21d9840c76b", ipAddress="192.168.127.1", resourceType="USER", resourcePath="users/a13130ae-ab37-47bd-88a5-de157e5da4c2"
``

### How to Reproduce?

Delete any user (or perform any other action that fires an AdminEvent)

### Anything else?

When looking at the logs, this may confuse admins. Seeing that a new user was created (or deleted) in the master realm, may trigger some alarms. It should be clear from the log entry in which realm the action was actually performed, as well as providing the authInfo (client, realm, user)

It might be considered a breaking change to use "realm" and "realmId" for consistency, because users may be analyzing their logs and depending on the current behaviour. So there might need to be some discussion if it's better to have consistency with user events, or non-breaking behaviour.

pine919:
Closes #46080


pine919:
@mposolda  Please review my PR

PR #46143, opened 2026-02-09
fix: adding server not ready to the status

### LINKED ISSUE #45802: Add more status information to Keycloak client crs
### Description

In the intial prototype PR there isn't anything being added to the status upon success.

Options include:
- sanitized version of the response resource - that will include defaults, but have secrets stripped

- the response result (create or update) - but that is of ephemeral use.
  - if deletion is allowed by another actor it would be useful to have the v1 uuid, which is not currently present on the v2 resource, to distinguish when the server side resource is being re-created

- clearly show validation errors

- indicate that the keycloak server is not ready

shawkins:
closes: #45802

So far just adding a check if the server is ready. This could be refined so that it's not logged in the general error handling.

We need more discussion on what else may be interest.


vmuzikar (dismissed):
LGTM, thanks.

vmuzikar:
@shawkins Could you please resolve conflicts?

shawkins:
Based upon the offline dicussion this is all we are interested in adding to the status at this time.

vmuzikar (approved):

PR #46165, opened 2026-02-10
added validation tests for client v2

### LINKED ISSUE #46203: Create tests for basic validations
### Description

We should provide more extended tests for the basic validations, especially for the Hibernate Validator validations, as we don't really use the ClientValidationProvider for the whole entity.

edewit:
Closes: #46203 
releates: #43296

### DIFF
--- a/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
+++ b/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
@@ -39,6 +39,7 @@ import org.apache.http.util.EntityUtils;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.hasItem;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -194,6 +195,7 @@ public class ClientApiV2ValidationTest {
             assertThat(body.error(), is("Provided data is invalid"));
             // Should have violations for the invalid URLs
             assertThat(body.violations().size(), greaterThanOrEqualTo(1));
+            assertThat(body.violations(), hasItem("redirectUris[].<iterable element>: Each redirect URL must be valid"));
         }
     }
 
@@ -218,6 +220,7 @@ public class ClientApiV2ValidationTest {
             var body = mapper.createParser(response.getEntity().getContent()).readValueAs(ViolationExceptionResponse.class);
             assertThat(body.error(), is("Provided data is invalid"));
             assertThat(body.violations().size(), greaterThanOrEqualTo(1));
+            assertThat(body.violations(), hasItem("redirectUris[].<iterable element>: Each redirect URL must be valid"));
         }
     }
 
@@ -246,6 +249,7 @@ public class ClientApiV2ValidationTest {
             var body = mapper.createParser(response.getEntity().getContent()).readValueAs(ViolationExceptionResponse.class);
             assertThat(body.error(), is("Provided data is invalid"));
             assertThat(body.violations().size(), greaterThanOrEqualTo(1));
+            assertThat(body.violations(), hasItem("roles[].<iterable element>: must not be blank"));
         }
     }
 
@@ -270,6 +274,7 @@ public class ClientApiV2ValidationTest {
             var body = mapper.createParser(response.getEntity().getContent()).readValueAs(ViolationExceptionResponse.class);
             assertThat(body.error(), is("Provided data is invalid"));
             assertThat(body.violations().size(), greaterThanOrEqualTo(1));
+            assertThat(body.violations(), hasItem("webOrigins[].<iterable element>: must not be blank"));
         }
     }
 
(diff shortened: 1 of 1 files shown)


rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
```diff
@@ -0,0 +1,477 @@
...
+                }
+                """));
+
+        try (var response = client.execute(request)) {
+            assertThat(response.getStatusLine().getStatusCode(), is(400));
+
+            var body = mapper.createParser(response.getEntity().getContent()).readValueAs(ViolationExceptionResponse.class);
+            assertThat(body.error(), is("Provided data is invalid"));
+            // Should have multiple violations (appUrl and redirectUris)
+            assertThat(body.violations().size(), greaterThanOrEqualTo(2));
```
mabartos: It'd be nice to assert the exact error messages (everywhere where only the count is checked).

mabartos (commented):
@edewit Looks good. I've just added one suggestion.

I think this only check the Hibernate Validator validations, and not the other one that are applied on clients - ClientValidationProvider. 

Can you check if we can provide some other test cases based on this? https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/validation/DefaultClientValidationProvider.java#L190

Thanks!

edewit:
> I think this only check the Hibernate Validator validations, and not the other one that are applied on clients - ClientValidationProvider.
> 
the v2 API currently only uses Hibernate Validator annotations (via HibernateValidatorProvider) but does not call `ValidationUtil.validateClient`

mabartos:
@edewit Ahh, I see it now...

```java
            // Create the client in the database
            model = clientsResource.createClientModel(basicRep);
//////////////
////////////// clientsResource
//////////////
            ValidationUtil.validateClient(session, clientModel, true, r -> {
                session.getTransactionManager().setRollbackOnly();
                throw new ErrorResponseException(
                        Errors.INVALID_INPUT,
                        r.getAllLocalizedErrorsAsString(AdminRoot.getMessages(session, realm, auth.adminAuth().getToken().getLocale())),
                        Response.Status.BAD_REQUEST);
            });
```

We call the `clientsResource.createClientModel()` where the validation is executed, but we call it only with a simplistic client that has only clientId and protocol set. So, no other properties are evaluated. 

So, we should also include the validation provider in the new service layer v2, as we don't have anything better right now for the client. 

I can create a follow up issue to include it into the service layer.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
```diff
@@ -0,0 +1,479 @@
...
+                    "enabled": true
+                }
+                """));
+
+        try (var response = client.execute(request)) {
+            assertThat(response.getStatusLine().getStatusCode(), is(400));
+
+            var body = mapper.createParser(response.getEntity().getContent()).readValueAs(ViolationExceptionResponse.class);
+            assertThat(body.error(), is("Provided data is invalid"));
+            assertThat(body.violations().size(), greaterThanOrEqualTo(1));
```
mabartos, line 302: Would it be possible to apply these checks here as well?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
```diff
@@ -0,0 +1,479 @@
...
```
mabartos, line 276: Would it be possible to apply these checks here as well?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
```diff
@@ -0,0 +1,479 @@
...
```
mabartos, line 251: Would it be possible to apply these checks here as well?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
```diff
@@ -0,0 +1,479 @@
...
```
mabartos, line 222: Would it be possible to apply these checks here as well?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2ValidationTest.java
```diff
@@ -0,0 +1,479 @@
...
+                }
+                """));
+
+        try (var response = client.execute(request)) {
+            assertThat(response.getStatusLine().getStatusCode(), is(400));
+
+            var body = mapper.createParser(response.getEntity().getContent()).readValueAs(ViolationExceptionResponse.class);
+            assertThat(body.error(), is("Provided data is invalid"));
+            // Should have violations for the invalid URLs
+            assertThat(body.violations().size(), greaterThanOrEqualTo(1));
```
mabartos, line 197: Would it be possible to apply these checks here as well?

mabartos (commented):
@edewit Yep, just last ones to complete to have it nice. If there's some bigger issue with these asserts, I'm not strict about this.

mabartos (approved):
@edewit LGTM, thanks!

mabartos:
@edewit jfyi - We have some simple validation test in the ClientApiV2Test that might be duplication of what you provided now. Would be good to probably remove it.

vmuzikar:
@edewit Thank you for the PR.

It will be good to rather sooner than later create a follow-up for it. We should basically have tests covering all validations that are currently handled by the v1 backend, so that when we switch to the new service layer we know didn't omit anything.

PR #46168, opened 2026-02-10
Separate password and OTP brute force protection to prevent OTP bypass attacks by default

### LINKED ISSUE #46164: Separate password and OTP brute force protection to prevent OTP bypass attacks by default
### Description

Enforce permanent lock-out after a large amount of invalid OTP's are entered. At the point an attacker can submit these they have already compromised a user password. 
However, because the same policy applies to the password-based login as well this creates a Denial of Service vulnerability in which an unprivileged external attacker can lock out user accounts by simply submitting a number of invalid passwords.

Setting lock counter max to 100 would make malicious guessing attacks completely unfeasible, even when multiple user passwords are breached at once.

This issue is based on [this discussion](https://groups.google.com/g/keycloak-user/c/eOWfOty1fhs/m/-uCV7-PdBgAJ?utm_medium=email&utm_source=footer).

### Value Proposition

This will make creating brute force mitigation policies a lot easier.

### Goals

Prevent guessing of OTP codes based when couple of passwords has been already successfully guessed.

### Non-Goals

Enhancement will no modify password based brute force mitigation policies.

### Discussion

_No response_

### Notes

_No response_

pskopek:
Separate password and OTP brute force protection to prevent OTP bypass attacks by default

Closes #46164


model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/entities/LoginFailureEntity.java
```diff
@@ -41,20 +41,23 @@ public class LoginFailureEntity extends SessionEntity {
     private long lastFailure;
     private String lastIPFailure;
 
+    private int numOtpFailures;
```
mabartos: Hi @pskopek, I'm just wondering if it would be possible to have some more general approach for counting the failures for a specific credential type. If we want to add some more counters for the login failures, it does not scale very well. Would it be possible to add some kind of map for credential types? 

WDYT?

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/remote/updater/loginfailures/LoginFailuresUpdater.java
```diff
@@ -160,6 +176,8 @@ private void addAndApplyChange(Consumer<LoginFailureEntity> change) {
     }
 
     private static final Consumer<LoginFailureEntity> CLEAR = LoginFailureEntity::clearFailures;
+    private static final Consumer<LoginFailureEntity> CLEAR_OTP_FAILURES = LoginFailureEntity::clearOtpFailures;
     private static final Consumer<LoginFailureEntity> INCREMENT_FAILURES = e -> e.setNumFailures(e.getNumFailures() + 1);
+    private static final Consumer<LoginFailureEntity> INCREMENT_OTP_FAILURES = e -> e.setNumFailures(e.getNumOtpFailures() + 1);
```
pruivo:
```suggestion
    private static final Consumer<LoginFailureEntity> INCREMENT_OTP_FAILURES = e -> e.setOtpNumFailures(e.getNumOtpFailures() + 1);
```

pruivo (changes requested):

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/entities/LoginFailureEntity.java
pruivo: @mabartos -1 from me.
Protostream is not efficient for marshalling maps, and you will have to box/unbox the number stored in the map. Not only will it be slow, but it will consume more bandwidth.

mabartos: @pruivo Alright, that is related to the Protostream. 

But we could have some kind of abstraction on the LoginFailureModel level, and handle it ISPN/protostream case by case?

I'd like to avoid adding multiple additional methods to the model object like currently: 
```java
int getNumOtpFailures();
void incrementOtpFailures();
void clearOtpFailures();
```

@pskopek @mposolda WDYT?

pruivo: What's wrong with it?

mabartos: @pruivo Got worried a little about the expansion of fields in the entity and model if there are many authenticators/credentials tracked for the brute force detection. However, as there's the list of ALLOWED_AUTHENTICATION_CATEGORIES, we can expect it will be limited, so there should not be any big concern. Moreover, there should not be even any huge number of these entities stored, as it should be one per realm+user - so not a big difference in the mentioned bandwidth. 

I had in mind some sort of general approach on how to track whatever authenticator/credential failure count for the future to provide multiple (custom) brute force detection mechanisms. But it's probably long-term future, so maybe we don't have to discuss now.

Just raising this for the future brute force protection usability concern - not a problem with this.

pskopek: @mabartos the specific credential type (OTP) is counted recorded separately because it is prone to successful guess  when there is a already known password. Default settings in case of temporary lockout and timing allows a lot of guesses (during longer time) and can eventually be discovered. This is permanent lockout added to OTP codes even if realm has generally set temporary lockout.
I other words one can try to guess 100 times of 1000000 possibilities before permanent lockout. It is enough tries for user and good enough protection.

All other second factors are not practically guessable.

pskopek:
Fixed the DCO issue (no other changes).

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -1723,7 +1727,32 @@ public static void logSuccess(KeycloakSession session, AuthenticationSessionMode
...
+        if (execution == null) return null;
+        AuthenticatorFactory factory = (AuthenticatorFactory) session.getKeycloakSessionFactory()
+                .getProviderFactory(Authenticator.class, execution.getAuthenticator());
+        if (factory != null) {
+            return factory.getReferenceCategory();
+        } else {
+            if (realm.getRequiredCredentialsStream().anyMatch(r -> Objects.equals(r.getType(), RequiredCredentialModel.PASSWORD.getType()))) {
+                return RequiredCredentialModel.PASSWORD.getType();
+            } else if (realm.getRequiredCredentialsStream().anyMatch(r -> Objects.equals(r.getType(), RequiredCredentialModel.TOTP.getType()))) {
+                return RequiredCredentialModel.TOTP.getType();
```
pruivo: Is it possible the realm requires both a password and TOTP? If so, this method will never return the `TOTP` type.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
```diff
@@ -396,6 +397,39 @@ public void testGrantMissingOtp() {
 
     }
 
+    @Test
+    public void testGrantPermamentOtpAbsolut() {
```
pruivo: typo

```suggestion
    public void testGrantPermanentOtpAbsolut() {
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
```diff
@@ -396,6 +397,39 @@ public void testGrantMissingOtp() {
...
+                Assert.assertEquals(response.getErrorDescription(), "Invalid user credentials");
+                events.clear();
+            }
+        }
+        {
+            String totpSecret = totp.generateTOTP("totpSecret");
+            AccessTokenResponse response = getTestToken(getPassword("test-user@localhost"), totpSecret);
+            assertTokenNull(response);
+            Assert.assertNotNull(response.getError());
+            Assert.assertEquals(response.getError(), "invalid_grant");
```
pruivo: parameters in the wrong order:

```suggestion
            Assert.assertEquals("invalid_grant", response.getError());
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
```diff
@@ -396,6 +397,39 @@ public void testGrantMissingOtp() {
...
+        for (int i = 0; i < DefaultBruteForceProtector.MAX_OTP_FAILURES; i++) {
+            {
+                AccessTokenResponse response = getTestToken(getPassword("test-user@localhost"), null);
+                Assert.assertNull(response.getAccessToken());
+                Assert.assertEquals(response.getError(), "invalid_grant");
+                Assert.assertEquals(response.getErrorDescription(), "Invalid user credentials");
+                events.clear();
+            }
+        }
+        {
```
pruivo: I'm not sure why you need this block

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
```diff
@@ -396,6 +397,39 @@ public void testGrantMissingOtp() {
...
+            Assert.assertNotNull(response.getAccessToken());
+            Assert.assertNull(response.getError());
+            events.clear();
+        }
+        for (int i = 0; i < DefaultBruteForceProtector.MAX_OTP_FAILURES; i++) {
+            {
+                AccessTokenResponse response = getTestToken(getPassword("test-user@localhost"), null);
+                Assert.assertNull(response.getAccessToken());
+                Assert.assertEquals(response.getError(), "invalid_grant");
+                Assert.assertEquals(response.getErrorDescription(), "Invalid user credentials");
```
pruivo:
```suggestion
                Assert.assertEquals("invalid_grant", response.getError());
                Assert.assertEquals("Invalid user credentials", response.getErrorDescription());
```

pruivo (changes requested):

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
```diff
@@ -60,15 +64,25 @@
...
     private static final Logger logger = Logger.getLogger(DefaultBruteForceProtector.class);
 
+    private static final List<String> ALLOWED_AUTHENTICATION_CATEGORIES = List.of(
+            PasswordCredentialModel.TYPE,
+            OTPCredentialModel.TYPE,
+            RecoveryAuthnCodesCredentialModel.TYPE
+    );
+
+    public static final String OTP_CATEGORY = OTPCredentialModel.TYPE;
+    public static final int MAX_OTP_FAILURES = 100;
```
rmartinc: I would have allowed this as configurable in the brute force, but we can do this later if it is requested.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
```diff
@@ -396,6 +397,39 @@ public void testGrantMissingOtp() {
...
+            }
+        }
+        {
+            String totpSecret = totp.generateTOTP("totpSecret");
+            AccessTokenResponse response = getTestToken(getPassword("test-user@localhost"), totpSecret);
+            assertTokenNull(response);
+            Assert.assertNotNull(response.getError());
+            Assert.assertEquals(response.getError(), "invalid_grant");
+            Assert.assertEquals("Invalid user credentials", response.getErrorDescription());
+            assertUserDisabledEvent(Errors.USER_TEMPORARILY_DISABLED);
```
rmartinc: @pskopek I think that this test is not testing that the the user is permanently locked after 100 failed OTP attempts. Probably you need to set in the realm failure factor to more than 100 and quick login check millis to 0. This way the user is not temporarily locked out and you can check the OTP is really permanently locking the user.

services/src/main/java/org/keycloak/services/managers/AuthenticationManager.java
```diff
@@ -1723,7 +1727,32 @@ public static void logSuccess(KeycloakSession session, AuthenticationSessionMode
...
+    public static String getAuthenticationCategory(KeycloakSession session, AuthenticationSessionModel authSession, boolean currentExecution) {
+        RealmModel realm = authSession.getRealm();
+        AuthenticationExecutionModel execution = realm.getAuthenticationExecutionById(
+                authSession.getAuthNote(
+                        currentExecution
+                                ? AuthenticationProcessor.CURRENT_AUTHENTICATION_EXECUTION
+                                : AuthenticationProcessor.LAST_PROCESSED_EXECUTION
+                )
+        );
+        if (execution == null) return null;
```
rmartinc: Playing with your test, I think that direct access grants return null execution here so OTP is not detected here for that grant. Maybe we have to pass the authenticator on failure and use the credentials used in success. I did a quick tets in branch: https://github.com/rmartinc/keycloak/tree/pr-46168

rmartinc (changes requested):
Thanks @pskopek!

I think the test is not testing the feature and, checking the test, I detected the category is not correctly detected in direct access grants. I did a quick fix in branch https://github.com/rmartinc/keycloak/tree/pr-46168. But probably something will fail. 😄

docs/documentation/server_admin/topics/threat/brute-force.adoc
```diff
@@ -73,6 +73,12 @@ Enabling an user account resets the `count`.
 .Lockout temporarily
 image:images/brute-force-temporarily.png[]
 
+[NOTE]
+====
+OTP type authenticators will cause permanent user account lockout even in case of temporary lockout brute force mode is set. This will mitigate probabilistic attack. The OTP oauthenticator type threshold is hardcoded to 100. This counter is incremented only in case of using OTP type authenticator.
```
mposolda:
```suggestion
OTP type authenticators will cause permanent user account lockout even in case of temporary lockout brute force mode is set. This will mitigate probabilistic attack. The OTP authenticator type threshold is hardcoded to 100. This counter is incremented only in case of using OTP type authenticator.
```

model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/remote/updater/loginfailures/LoginFailuresUpdater.java
```diff
@@ -149,6 +149,22 @@ public void setLastIPFailure(String ip) {
         addAndApplyChange(e -> e.setLastIPFailure(ip));
     }
 
+    @Override
+    public int getNumOtpFailures() {
```
mposolda: Is it possible to rename this method (and other related methods and constants) to something more generic, so that brute-force protector API do not reference "OTP" as concrete authentication method in those methods? Perhaps something like `getNumSecondaryAuthFailures()`

mposolda (commented):
I did not checked the details in the codebase as it is already reviewed by others. Just some high level thoughts:

- It seems to me that the behavior of automatically doing permanent lockout for OTP is not backwards compatible. Existing deployments may not expect this behavior as there is not permanent lockout as of now. I suppose we may need a switch on the brute-force protector (disabled by default) and make this behavior enabled just with the switch?

- It will be good to document this also in the brute-force protector documentation (probably together with the new switch added from above if we add this). As upgrading guide is something, which people read just when updating to the particular release.
 
- I think we should also document (in both brute-force docs and upgrading guide) the behavior around custom authenticators (x509, passkeys, IDP login etc) and about the fact that they do not count anymore to brute-force and successful login with those do not reset brute-force

docs/documentation/server_admin/topics/threat/brute-force.adoc
```diff
@@ -252,6 +253,11 @@ wait time will never reach the value you have set to `Max wait`.
 `count` does not increment when a temporarily disabled account commits a login failure.
 ====
 
+==== Secondary Authentication Failures Lockout
+When this lockout mechanism is enabled it is valid for all brute force modes. The mechanism is enabled when non zero value is set for `Maximum Secondary Authentication Failures`. It is disabled by default (`0`).
+
+If the maximum number secondary authentication failures is exceeded the account is permanently locked. This behavior prevents attacks on second factor authenticators such as OTP when attacker already guessed the password. The reasonable value might be set for `100`.
```
rmartinc: Maybe we can say that currently it is only for OTP. Something like:


```suggestion
If the maximum number secondary authentication failures is exceeded the account is permanently locked. This behavior prevents attacks on second factor authenticators that can be brute forced but are unlikely to fail (currently only OTP), when attacker already guessed the password. The reasonable value might be set for `100`.
```

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -87,6 +87,10 @@ The `BROKER_LINK` table now contains two additional indexes `IDX_BROKER_LINK_USE
 If the table contains more than 300000 entries, {project_name} will skip the index creation by default during the automatic schema migration and instead log the SQL statement on the console during migration to be applied manually after {project_name}'s startup.
 See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to configure a different limit.
 
+=== New brute force locking mechanism
+
+There is a new brute force locking mechanism applied to all brute force modes. It is called `Secondary Authentication Failures Lockout`. Brute force protector is counting secondary authentication failures only (for example OTP) and when maximum specified number is reached the account is permanently locke even in temporary locking modes. This prevents attacks on second factor authenticators such as OTP when attacker already guessed the password.
```
rmartinc:
```suggestion
There is a new brute force locking mechanism applied to all brute force modes. It is called `Secondary Authentication Failures Lockout`. Brute force protector is counting secondary authentication failures only (for example OTP) and when maximum specified number is reached the account is permanently locked even in temporary locking modes. This prevents attacks on second factor authenticators such as OTP when attacker already guessed the password.
```

rmartinc (approved):
Thanks @pskopek! Just two minor doc changes.

My feeling is that the current wording in the doc allows to extend this to other mechanisms that are safe, mechs that don't usually fail. This brings recovery codes also in the same boat (I don't think a user tries to guess it's recovery codes randomly). But we can do it later if requested. It's OK to me.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
pskopek: Resolved in the following commit.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
```diff
@@ -93,6 +93,10 @@ The `BROKER_LINK` table now contains two additional indexes `IDX_BROKER_LINK_USE
 If the table contains more than 300000 entries, {project_name} will skip the index creation by default during the automatic schema migration and instead log the SQL statement on the console during migration to be applied manually after {project_name}'s startup.
 See the link:{upgradingguide_link}[{upgradingguide_name}] for details on how to configure a different limit.
 
+=== New brute force locking mechanism
+
+There is a new brute force locking mechanism applied to all brute force modes. It is called `Secondary Authentication Failures Lockout`. Brute force protector is counting secondary authentication failures only (for example OTP) and when maximum specified number is reached the account is permanently locked even in temporary locking modes. This prevents attacks on second factor authenticators such as OTP when attacker already guessed the password.
+
```
mposolda, line 92: AFAIK besides introducing "secondary mechanism" , the change done by this PR is also the fact, that only some subset of authentication mechanisms (passwords, OTP, recovery-codes) is considered for brute-force successful logins and failed logins. It can be good to document this behavior change in the upgrading guide as well.

Besides that, it can be great to document this in the "Brute force" documentation itself. Brute-force documentation has some examples of flows with "successful logins" and "failed logins", but it can be great to add some dedicated section in this documentation and define what "successful login" or "failed login" means (mention just the supported mechanism and the fact that login with X.509, identity brokering or passkey is not considered for brute-force for both successful or failed login).

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
```diff
@@ -60,15 +64,24 @@
 public class DefaultBruteForceProtector implements BruteForceProtector {
     private static final Logger logger = Logger.getLogger(DefaultBruteForceProtector.class);
 
+    private static final List<String> ALLOWED_AUTHENTICATION_CATEGORIES = List.of(
```
mposolda: Will it makes sense to make this list `public` ?

The thing is, that if someone used brute-force protector with some custom authenticator mechanism, it worked for him, but with this change, brute-force will not be easily possible to use anymore with custom credentials if I understand correctly. Making this list `public` will at least makes it possible that people can add their custom credential to this list (which they can do in the initialization of their custom authenticator factory).

mposolda (commented):
@pskopek Added few minor comments. Could you check please?

@pruivo Do you please have a chance to re-review infinispan related changes?

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
pskopek: Yes @mposolda you are right. Adding a commit to change it.

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
```diff
@@ -60,15 +64,24 @@
 public class DefaultBruteForceProtector implements BruteForceProtector {
     private static final Logger logger = Logger.getLogger(DefaultBruteForceProtector.class);
 
+    public static final List<String> ALLOWED_AUTHENTICATION_CATEGORIES = List.of(
```
mposolda:
```suggestion
    public static final List<String> ALLOWED_AUTHENTICATION_CATEGORIES = new ArrayList<>(List.of(
```

Sorry for not being 100% accurate in my last comment.

It can be great if people have possibility to add their mechanism to this list. Using `public` helps, but it is not sufficient as `List.of` is using immutable list. It might be also needed to "decorate" this with something like for example `java.util.ArrayList` to make it possible that the collection is mutable. WDYT?

mposolda (commented):
@pskopek Thanks for the last change. Added additional inline comment. Besides that, the documentation update from my last comment is still pending, but I guess you know :-)

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
pskopek: Fixed.

docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
pskopek: Docs updated.

abstractj:
@pskopek we may have a bug. It seems that clearFailures() doesn't reset numSecondaryAuthFailures, which causes two issues:

  1. In LoginFailuresUpdater, success() calls clearSecondaryAuthFailures() then clearFailures() — but both call changes.clear(), so the secondary reset is wiped before it's replayed to the Infinispan cache (Potential issue on clustered deployments).
  2. In the delta-time reset path (DefaultBruteForceProtector.failure() line 223), only primary counters are reset. The OTP counter accumulates across reset windows indefinitely.

Adding this.numSecondaryAuthFailures = 0 to clearFailures() may fix both. If that's not a concern, feel free to ignore those comments.

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
```diff
@@ -60,15 +65,26 @@
 public class DefaultBruteForceProtector implements BruteForceProtector {
     private static final Logger logger = Logger.getLogger(DefaultBruteForceProtector.class);
 
+    public static final List<String> ALLOWED_AUTHENTICATION_CATEGORIES = new ArrayList<>(
```
pruivo, line 68: nitpicking, why doing `new ArrayList(List.of(...))`? Does it need to be mutable?

pruivo (commented):
It looks good to me. I only have a question 👍

services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java
pskopek: It needs to be mutable in case user creates a new custom authenticator and wants it to be supported by brute force protection.

pskopek:
> @pskopek we may have a bug. It seems that clearFailures() doesn't reset numSecondaryAuthFailures, which causes two issues:
> 
> 1. In LoginFailuresUpdater, success() calls clearSecondaryAuthFailures() then clearFailures() — but both call changes.clear(), so the secondary reset is wiped before it's replayed to the Infinispan cache (Potential issue on clustered deployments).

The secondary reset is called first and it wipes out changes and then add actual change and applies the change to cached entity. (this happens only in case of OTP type credential as intended). Then we need to clear primary counter in any type of credential. The secondary counter change is already applied.

> 2. In the delta-time reset path (DefaultBruteForceProtector.failure() line 223), only primary counters are reset. The OTP counter accumulates across reset windows indefinitely.
> 
The debug message is not proper. But clearFailures() method needs to clear primary counter and do not touch secondary one.


> Adding this.numSecondaryAuthFailures = 0 to clearFailures() may fix both. If that's not a concern, feel free to ignore those comments.

We cannot clear secondary counter as it will undermine checks in case of attacker already guessed the password and trying to guess OTP value and byepass the brute force protection for OTP.

pruivo:
> The secondary reset is called first and it wipes out changes and then add actual change and applies the change to cached entity. (this happens only in case of OTP type credential as intended). Then we need to clear primary counter in any type of credential. The secondary counter change is already applied.

@pskopek, @abstractj is right. You have to remove the `changes.clear();` from `LoginFailuresUpdater`. 

The way it works, Keycloak tries to do a replace with the version read during the transaction. If it succeeds, it happens as you describe.

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/remote/transaction/RemoteChangeLogTransaction.java#L230-L232

But, if the replace fails because another transaction updated the entity, Keycloak falls back to `computeIfPresent`. In this scenario, a new version of the entity is fetched, and the `changes` list is applied to it. You want both clears to be in there. 

https://github.com/keycloak/keycloak/blob/a2c1055f8df21b53b969f922fb9a044a76d22604/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/remote/transaction/RemoteChangeLogTransaction.java#L234-L236

rmartinc:
@pskopek Maybe you can add `clearFailuresAndSecondaryAuth` or similar instead, and call `clearFailuresAndSecondaryAuth` or `clearFailures` depending OTP is there or not.

pruivo (dismissed):

pskopek:
Just a rebase and squash commits. No other changes.

pruivo (approved):

mposolda (approved):
@pskopek @rmartinc @pruivo @abstractj Thanks everyone for the feedback and the work on this PR!

PR #46171, opened 2026-02-10
Improve SimpleHttp API

### LINKED ISSUE #43701: Improve SimpleHttp API

thomasdarimont:
- Expose configured HTTP method from SimpleHttpRequest - needed by extensions / libraries that wants to enrich the SimpleHTTP call based on the configured method.
- Use same ObjectMapper configuration as in JsonSerialization used by the deprecated SimpleHTTP to produce consistent json output.
- Allow to configure the ObjectMapper to use by SimpleHttp - this enables custom JSON formats
- Expose ObjectMapper factory methods in JsonSerialization - allows consistent JSON serialization.

Fixes https://github.com/keycloak/keycloak/issues/43701

Followup to #43702

### DIFF
--- a/core/src/main/java/org/keycloak/util/JsonSerialization.java
+++ b/core/src/main/java/org/keycloak/util/JsonSerialization.java
@@ -38,17 +38,24 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
  * @version $Revision: 1 $
  */
 public class JsonSerialization {
-    public static final ObjectMapper mapper = new ObjectMapper();
-    public static final ObjectMapper prettyMapper = new ObjectMapper();
+    public static final ObjectMapper mapper = objectMapperWithDefaults();
+    public static final ObjectMapper prettyMapper = prettyObjectMapperWithDefaults();
 
-    static {
+    public static ObjectMapper objectMapperWithDefaults() {
+        ObjectMapper mapper = new ObjectMapper();
         mapper.registerModule(new Jdk8Module());
         mapper.registerModule(new JavaTimeModule());
         mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
         mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        return mapper;
+    }
+
+    public static ObjectMapper prettyObjectMapperWithDefaults() {
+        ObjectMapper prettyMapper = new ObjectMapper();
         prettyMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
         prettyMapper.enable(SerializationFeature.INDENT_OUTPUT);
         prettyMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        return prettyMapper;
     }
 
     public static String valueAsString(Object obj) {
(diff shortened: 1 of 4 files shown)


server-spi-private/src/main/java/org/keycloak/http/simple/SimpleHttp.java
```diff
@@ -2,22 +2,25 @@
...
+import org.keycloak.util.JsonSerialization;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 
 public class SimpleHttp {
 
-    private static final ObjectMapper objectMapper = new ObjectMapper();
+    private static final ObjectMapper DEFAULT_OBJECT_MAPPER = JsonSerialization.createObjectMapperWithDefaults();
```
mabartos:
```suggestion
    private static final ObjectMapper DEFAULT_OBJECT_MAPPER = JsonSerialization.mapper;
```

Is there a need to create a new mapper, or is it possible to reuse the existing one?

thomasdarimont:
Thanks @mabartos for taking a look! I intentially created a new instance to avoid potential mutations of the shared ObjectMapper to have an affect the SimpleHTTP json serialization, but perhaps that was not necessary.

I applied your suggestion to reuse the JsonSerialization.mapper reference directly.

I keep the `createObjectMapper...` factory methods public to allow users to make customizations to mappers that are based on the "default" mapper confgurations.

server-spi-private/src/main/java/org/keycloak/http/simple/SimpleHttp.java
mabartos: @thomasdarimont wrote:
> I intentially created a new instance to avoid potential mutations of the shared ObjectMapper to have an affect the SimpleHTTP json serialization, but perhaps that was not necessary.

@shawkins Or do you see a problem here to reference the default mapper for the SimpleHttp? I remember that you were involved in some work around these mappers, so maybe you have something to add. Thanks!

shawkins: https://github.com/keycloak/keycloak/issues/44159 is the issue for the rest layer mapper, which is different than the one used for JsonSerialization. If we can reason that the needs will be identical, I'm all for sharing mappers.

core/src/main/java/org/keycloak/util/JsonSerialization.java
```diff
@@ -38,17 +38,24 @@
...
+    public static final ObjectMapper prettyMapper = createPrettyObjectMapperWithDefaults();
 
-    static {
+    public static ObjectMapper createObjectMapperWithDefaults() {
+        ObjectMapper mapper = new ObjectMapper();
         mapper.registerModule(new Jdk8Module());
         mapper.registerModule(new JavaTimeModule());
         mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
-        mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        mapper.setDefaultPropertyInclusion(JsonInclude.Include.NON_NULL);
```
mabartos, line 49: Just for context for another reviewer - the `setSerializationInclusion()` is deprecated, and uses the `setDefaultPropertyInclusion` underhood (`setSerializationInclusion -> setPropertyInclusion -> setDefaultPropertyInclusion`). 

So, ok from my side, as there should not be any behavioral changes provided here.

mabartos (approved):
@thomasdarimont LGTM, thanks!

@keycloak/core As it is your area, could you please check this simple PR? Thanks!

@stianst You were involved in the older PR related to this - do you wanna review it as well?

stianst (approved):

PR #46180, opened 2026-02-10
[admin-v2] Enable client generators to create proper class hierarchies with inheritance

### LINKED ISSUE #46158: Enable client generators to create proper class hierarchies with inheritance
### Description

Current OpenAPI generation placed polymorphism in subtypes, preventing client generators from creating proper inheritance structures.
Implement proper OpenAPI polymorphism so client generators produce class hierarchies (discriminator on parent; subtypes reference parent via `allOf` instead of inline `oneOf`).

This should follow the official [Swagger polymorphism documentation](https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/) more accurately.

Pepo48:
Closes: #46158

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
@@ -1,6 +1,5 @@
 package org.keycloak.quarkus.runtime.oas;
 
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -15,11 +14,8 @@ import org.eclipse.microprofile.openapi.OASFactory;
 import org.eclipse.microprofile.openapi.OASFilter;
 import org.eclipse.microprofile.openapi.models.OpenAPI;
 import org.eclipse.microprofile.openapi.models.PathItem;
-import org.eclipse.microprofile.openapi.models.media.Content;
 import org.eclipse.microprofile.openapi.models.media.Discriminator;
 import org.eclipse.microprofile.openapi.models.media.Schema;
-import org.eclipse.microprofile.openapi.models.parameters.RequestBody;
-import org.eclipse.microprofile.openapi.models.responses.APIResponses;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
@@ -60,44 +56,8 @@ public class OASModelFilter implements OASFilter {
 
         Map<String, Set<Schema>> discriminatorPropertiesToBeAdded = new HashMap<>();
 
-        // Reflect Jackson annotations in OpenAPI spec
         // Follows https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/
-        openAPI.getPaths().getPathItems().values().stream()
-                .flatMap(p -> p.getOperations().values().stream())
-                .forEach(operation -> {
-                    // This is not nice but so is the model structure...
-
-                    // Request body
-                    Optional.ofNullable(operation.getRequestBody())
-                            .map(RequestBody::getContent)
-                            .map(Content::getMediaTypes)
-                            .map(Map::values)
-                            .map(Collection::stream)
-                            .ifPresent(mediaTypes -> {
-                                mediaTypes.forEach(mediaType -> {
-                                    mediaType.setSchema(replaceSchemaWithChildrenIfNeeded(mediaType.getSchema(), openAPI, discriminatorPropertiesToBeAdded));
-                                });
-                            });
-
-                    // Responses
-                    Optional.ofNullable(operation.getResponses())
-                            .map(APIResponses::getAPIResponses)
-                            .map(Map::values)
-                            .map(Collection::stream)
-                            .ifPresent(apiResponses -> {
-                                apiResponses.forEach(apiResponse -> {
-                                    Optional.ofNullable(apiResponse.getContent())
-                                            .map(Content::getMediaTypes)
-                                            .map(Map::values)
-                                            .map(Collection::stream)
-                                            .ifPresent(mediaTypes -> {
-                                                mediaTypes.forEach(mediaType -> {
-                                                    mediaType.setSchema(replaceSchemaWithChildrenIfNeeded(mediaType.getSchema(), openAPI, discriminatorPropertiesToBeAdded));
-                                                });
-                                            });
-                                });
-                            });
-                });
+        addDiscriminatorsToParentSchemas(openAPI, discriminatorPropertiesToBeAdded);
 
         // Add missing discriminator properties to subclass schemas
         // Normally, this is handled by Jackson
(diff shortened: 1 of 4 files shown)


mabartos:
@michalvavrik @vmuzikar @robson90 Could you please check this PR? Thanks!

michalvavrik:
Looking.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
```diff
@@ -111,6 +75,120 @@ public void filterOpenAPI(OpenAPI openAPI) {
...
+            log.debugf("Added discriminator '%s' to schema '%s' with %d subtypes",
+                    discriminatorPropertyName, schemaName, typeAnnotations.length);
+        }
+    }
+
+    /**
+     * Detects classes with @JsonAnyGetter and sets additionalProperties on their schemas.
+     * This reflects that the class accepts arbitrary additional JSON properties at runtime.
+     */
+    private void addAdditionalPropertiesToSchemas(OpenAPI openAPI) {
```
michalvavrik: I am fine with doing this for any `JsonAnyGetter` provided they are missing from the generated OpenAPI document (which they are, I checked). But you explicitly speak about `additonalProperties`, so either:

- you are dead certain there never will be any other such getter with different name
- tweak javadoc or method name to clarify that this propagates any such a getter into openapi specs
- filter method infos to only those with names `getAdditionalFields`

It is a nitpick...

shawkins: We actually don't want anything to show up in the schema for JsonAnyGetter.

michalvavrik: Ok, I didn't know. Also looking at the `entry.getValue().setAdditionalPropertiesSchema(OASFactory.createSchema());` I think if it were to be present in the definition, you need to filter methods by name to do not include accidentally future getters.

Pepo48: The intention behind bringing back the `additionalProperties` was to try, whether it could solve issues with Kiota generator (reportedly it did). But I might've misunderstood the goal. Also I did not test this with Kiota myself - thus it is in a separate commit, which we can easily remove.

@michalvavrik can you confirm that this `additionalProperties` is what fixes the failures even for you and whether it is really what we need here?

Another approach would be adding the `@Schema(additionalProperties = Schema.True.class)`, which I actually tried at first, but I decided not to use it, because it was syntactically different (generated just `additionalProperties: true`, not sure whether it poses some problem or not).

Thank you!

michalvavrik: > @michalvavrik can you confirm that this additionalProperties is what fixes the failures even for you and whether it is really what we need here?

it doesn't; gosh I need to make myself clearer in the future, sorry


So the issue with the kiota generator is that `BaseRepresentation` is empty in the generated OpenAPI document. But that is still the case, as currently, `BaseRepresentation` isn't augmented by the `addAdditionalPropertiesToSchemas`. Are there 2 ways I know about how to deal with it:

- @edewit just removed `@Schema` and obviously unused `BaseRepresentation` disappeared, that is fine if we add comment that once the representation contains any properties, we would like to have it in the OpenAPI document
- we can add dummy readonly property, e.g. A="1", but I looked into OpenAPI specs and I can't see how to instruct generators to never include it in clients. it is not nice, I'd not use it

> Another approach would be adding the @Schema(additionalProperties = Schema.True.class), which I actually tried at first, but I decided not to use it, because it was syntactically different (generated just additionalProperties: true, not sure whether it poses some problem or not).

If there is no function for having the `additionalProperties` in the response, e.g. client doesn't need it, let's not do it.

michalvavrik: Sorry for your time if you did it only because of my previous comments.

Pepo48: @michalvavrik no worries, it's misunderstanding on my end, partially caused by the fact that OpenAPI generator's behaviour was different. I just read your comment under Erik's PR and hopefully got the context. Now it should copy Erik's approach.

robson90:
For me the generation is failing, because of the non existent `BaseRepresentation`. 

Logs:
```
[INFO] --- kiota:0.0.31:generate (default) @ keycloak-admin-openapi ---
[info] Going to execute the command: /Users/robinmeese/Documents/GitHub/meine/keycloak-client/admin-openapi/target/kiota/1.30.0/kiota generate --openapi /Users/robinmeese/Documents/GitHub/meine/keycloak-client/admin-openapi/src/main/resources/openapi-pr-46180.yaml --output /Users/robinmeese/Documents/GitHub/meine/keycloak-client/admin-openapi/target/generated-sources/kiota/org/keycloak/robin --language Java --class-name ApiClient --namespace-name org.keycloak.robin --type-access-modifier Public --serializer io.kiota.serialization.json.JsonSerializationWriterFactory --deserializer io.kiota.serialization.json.JsonParseNodeFactory --additional-data --log-level Warning
warn: Kiota.Builder.KiotaBuilder[534942631]
      OpenAPI warning: #/components/schemas/Auth/allOf/0/$ref - The schema reference '#/components/schemas/BaseRepresentation' does not point to an existing schema.
warn: Kiota.Builder.KiotaBuilder[534942631]
      OpenAPI warning: #/components/schemas/BaseClientRepresentation/allOf/0/$ref - The schema reference '#/components/schemas/BaseRepresentation' does not point to an existing schema.
warn: Kiota.Builder.KiotaBuilder[534942631]
      OpenAPI warning: #/ - Multiple servers entries were found in the OpenAPI description. Only the first one will be used. The root URL can be set manually with the request adapter.
crit: Kiota.Builder.KiotaBuilder[0]
      error generating the client: One or more errors occurred. (One or more errors occurred. (One or more errors occurred. (One or more errors occurred. (One or more errors occurred. (The type does not contain any information Path: \admin\api\{realmName}\clients\{version}, Reference Id: BaseRepresentation)))))
```

As said on Slack, my fix would be this part added to the openapi:
```
.....
    BaseRepresentation:
      type: object
      properties:
        additionalFields:
          type: object
          additionalProperties: { }
....
```

Just for testing I did change locally:
Screenshot 2026-02-11 at 15 20 11

Outcome:
```
    BaseRepresentation:
      description: Base representation with common properties for all representation
        types
      type: object
      properties:
        additionalFields:
          type: object
          additionalProperties: {}
```

After compiling and pasting it in, Kiota is running without any complaints, the calls with an OIDCRep is working. 
Depending on what the useCase for the BaseRepresentation is, this could be a feasible option. 
When looking at the field, it seems like you want to achieve, that every Representation has a field of name `additionalFields`.


-------------------------------------
Now the part where I don't know, how fix. When calling get on a Client it has the type `BaseClientRep`. As the discriminator is in place, we can leverage `instanceOf` To know which type it is. For all properties that are in `BaseClientRep` should just work. For specific OIDC attributes, we need to cast it. 

Only way right now, as an easy fix and also a nice usable api on the Java side could be to support all of this:

/clients -> will return Type `BaseClientRepresentation`
/clients/oidc -> will return Type `OIDCClientRepresentation`
/clients/saml -> will return Type `SAMLClientRepresentation`

Depending on the UseCase I would then use the respective function for updating resources.

WDYT?


For reference here are my testings: https://github.com/robson90/keycloak-client/pull/3

michalvavrik:
@Pepo48 the last time I read this PR it seemed fine and I didn't expect you to deal with `BaseClientRepresentation`. If you plan to, then additional work is probably needed. Please let us know once this is ready for a review so that you don't wait. Thanks.

Pepo48:
@michalvavrik OpenAPI generator didn't require any changes with regards to the base rep. Everything that is needed on top of my original PoC is just an attempt to help you guys, to stabilize Kiota.

That was the reason of the first attempt actually - I read the comment made by @robson90 from our offline conversation and tried to bring `additionalFields` back. But it didn't help to you, so I removed the schema completely to follow your suggestion, which broke it for Robin 😄.

But if I understand it correctly, then ` @JsonProperty("additionalFields")` could work for both, right?

And as a next step I can try to add the return types that Robin suggested, if we agree on it.

Pepo48:
Regardless, feel free to bring your own adjustments at any point of time or anybody can take over this PR, if needed. My initial intention was just provide you with some basis that I already had, I don't have to necessarily drive this to the finish line (as I don't work with Kiota anyway, so you know better what you exactly need).

michalvavrik:
> @michalvavrik OpenAPI generator didn't require any changes with regards to the base rep. Everything that is needed on top of my original PoC is just an attempt to help you guys, to stabilize Kiota.

Frankly, I had no idea that was the intention. I thought that you are fixing the class hierarchy, which seems helpful. Knowing this will help.

> But if I understand it correctly, then  @JsonProperty("additionalFields") could work for both, right?

IMO it would work if the additional properties were on `BaseRepresentation` and the `BaseClientRepresentation` extended it in the OpenAPI definition as well. (I am writing these names from the top of my head). but per Steven comment, I'd not go there.

> Regardless, feel free to bring your own adjustments at any point of time or anybody can take over this PR, if needed. My initial intention was just provide you with some basis that I already had, I don't have to necessarily drive this to the finish line (as I don't work with Kiota anyway, so you know better what you exactly need).

Well, let see how much time I have tomorrow. I don't want your work to go waste, we can finish this one.

robson90:
> That was the reason of the first attempt actually - I read the comment made by @robson90 from our offline conversation and tried to bring additionalFields back. But it didn't help to you, so I removed the schema completely to follow your suggestion, which broke it for Robin 😄.

Yeah, thanks for your investigation, this is now working "as expected". 

Regrading my massive comments from above, we can also discuss the next steps on Monday. 

PR itself LGTM, except CI is complaining about spotless...


> IMO it would work if the additional properties were on BaseRepresentation and the BaseClientRepresentation extended it in the OpenAPI definition as well. (I am writing these names from the top of my head). but per Steven comment, I'd not go there.

@michalvavrik what is the exact use case for this attribute ?

michalvavrik:
>      @michalvavrik what is the exact use case for this attribute ?

I don't know yet. It is not relevant because I don't want to include it.

michalvavrik (commented):
Hello,

I had a look again and here is my conclusion.

1. ATM this PR resolves Kiota generation issue and LGTM
2. I lack knowledge about `additionalFields`, but IIUC @shawkins mentioned that we don't want them in the OpenApi document. In which case I can see only 2 options how to solve the "empty BaseRepresentation" issue:
    - add discriminator to the `BaseRepresentation` as I assume we are going to need it eventually to determine if it is a client or realm, user, or ...?
    - just bring back `@JsonIgnore` and drop `@Schema` as it has no function in regards to OpenAPI ATM

Thanks @Pepo48 for your work. You will have to decide what to do with your PR.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/oas/OASModelFilter.java
```diff
@@ -111,6 +71,95 @@ public void filterOpenAPI(OpenAPI openAPI) {
...
+                    .map(AnnotationValue::asString)
+                    .orElse("");
+            if (discriminatorPropertyName.isEmpty()) {
+                continue;
+            }
+
+            Schema parentSchema = openAPI.getComponents().getSchemas().get(schemaName);
+            if (parentSchema == null) {
+                continue;
+            }
```
vmuzikar, line 176: I'd suggest throwing exceptions if validations fail. It will make it easier to spot an error if the annotations were not set correctly.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/BaseRepresentation.java
```diff
@@ -5,15 +5,15 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
-@Schema
+@Schema(description = "Base representation with common properties for all representation types")
 public class BaseRepresentation {
 
-    @JsonIgnore
+    @JsonProperty("additionalFields")
```
vmuzikar: > 2\. I lack knowledge about `additionalFields`, but IIUC @shawkins mentioned that we don't want them in the OpenApi document. In which case I can see only 2 options how to solve the "empty BaseRepresentation" issue

Exactly, additionalFields do not belong to the spec, they are not meant to be set directly be the clients. If it's just a workaround for an empty base rep, we can remove it on the filter level.

vmuzikar (changes requested):
@Pepo48 Thank you for the PR.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/OpenApiDistTest.java
```diff
@@ -99,22 +99,24 @@ void testOpenApiFilter(KeycloakDistribution distribution) {
                 .get(OPENAPI_ENDPOINT)
             .then();
 
-      assertOpenAPISpecPolymorphicPaths(response, "paths.'/admin/api/{realmName}/clients/{version}/{id}'.put.requestBody.content.'application/json'.schema"); // request
-      assertOpenAPISpecPolymorphicPaths(response, "paths.'/admin/api/{realmName}/clients/{version}/{id}'.get.responses.'200'.content.'application/json'.schema"); // response
-      assertOpenAPISpecPolymorphicPaths(response, "paths.'/admin/api/{realmName}/clients/{version}'.get.responses.'200'.content.'application/json'.schema.items"); // arrays
+      // Verify discriminator is on the parent schema (BaseClientRepresentation) in components section
+      response
```
michalvavrik, line 108: I think it would be nice to check the body as string that it does not contain "BaseRepresentation" at all, provided it can't be infix/prefix/suffix of other string. I think there can't, as it is `BaseClientRepresentation` etc. Your call.

michalvavrik (approved):
LGTM

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/OpenApiDistTest.java
Pepo48: Addressed.

michalvavrik: Thank you

michalvavrik:
I need this to fix Kiota generation for JS client, therefore I'd like to ask @keycloak/cloud-native-maintainers  reviewers to have a look again. Thanks.

vmuzikar (approved):
LGTM (assuming the generators are now happy :)), thank you @Pepo48!

PR #46182, opened 2026-02-10
fix: refining how the junit Keycloak is launched

### LINKED ISSUE #46160: Keycloak from `quarkus/tests/junit5` doesn't throw exception when there's a startup failure
### Before reporting an issue

### Area

dist/quarkus

### Describe the bug

If Keycloak fails to start there is a warning printed on system error, but no exception is thrown. Not sure if it eventually times out or not, but it should throw an exception immediately.

I'm pretty sure this is a regression and it didn't use to work like this until recently. This also affected the embedded server mode for the new test framework, providing a rather poor experience.

### Version

main

### Regression

### Expected behavior

If Keycloak fails to start, for example due to invalid configuration, an exception should be thrown straight away.

### Actual behavior

It hangs for a long time, maybe forever? Not sure if it will eventually time out, but I've waited 5 mins so far ;)

### How to Reproduce?

Run for example `CustomConfigTest` from new test framework examples with the embedded server mode. Update the server config in `CustomConfigTest` first to add an invalid option for instance `.option("nosuch", "nosuch")` to cause it to fail at startup.

### Anything else?

_No response_

shawkins:
closes: #46160

Turns out the problem wasn't with asyncExit, it was more of an issue with initSys. To prevent it from causing the start to hang and / or produce duplicate results we need to only let it have an effect for auto build commands and require it to return the built exit code.

~~The change to the QuarkusKeycloakSessionFactory is to address that some failed launches, such as when the hostname provider fails validation, will leave provider factories as not being cleaned up - for example BasicTimerProviderFactory starts a non-daemon thread.~~

Limited the additional fix to just BasicTimerProviderFactory - the factory logic seems pretty fragile in terms of when close can be called - some expect postInit to have been called, while other it's just after init.

### DIFF
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/Picocli.java
@@ -184,7 +184,7 @@ public class Picocli {
             // to reuse the previous result either means we need to duplicate the logic in the execute method
             // or refactor the above logic so that it happens in the command logic
             // We could also reduce the memory footprint of the ParseResult, but that looks a little hackish
-            int exitCode = cmd.execute(argArray);
+            int exitCode = execute(cmd, argArray);
 
             exit(exitCode);
         } catch (ParameterException parEx) {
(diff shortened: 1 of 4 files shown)


michalvavrik (commented):
Judging by the context around modified methods and in `org.keycloak.quarkus.runtime.cli.command.AbstractAutoBuildCommand`, I think this change is correct. But you will need review from someone with wider context. Thanks for the fix.

mabartos (approved):
LGTM. 

I can confirm it resolves the issue.

shawkins:
> LGTM.
> 
> I can confirm it resolves the issue.

Thanks @mabartos for confirmation.

Captured #46238 as a follow-up to this.

vmuzikar (approved):
Approving based on @mabartos' and @michalvavrik's reviews.

PR #46183, opened 2026-02-10
Improved documentation for running the keycloak test mail server locally.

ruchikajha95:
Closes #45837

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -135,29 +135,58 @@ The mail server will start on `localhost:3025` and display all received emails i
 
 ### Configuring Keycloak
 
-To configure Keycloak to use the test mail server, enter the following values in the realm configuration in the **Email
-** tab:
+To configure Keycloak to use the test mail server:
 
-| Field | Value             |
-|-------|-------------------|
-| From  | auto@keycloak.org |
-| Host  | localhost         |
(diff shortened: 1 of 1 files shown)


ruchikajha95:
@ahus1 i have added changes to tests.md file to improve the document from development perspective. Could you please review and let me know if any changes needed. Thanks !

ahus1 (commented):
Thank you for the documentation update. I pushed some additional changes, please review: 

* The startup command did not work for me. I needed to go to the `utils` folder first. I changed the docs to use both the Maven wrapper, and put it into a single command.
* Specifying the admin credentials was IMHO a bit too specific, as it will work with any other admin credential as well. When adding the UI testing, this was different as the UI tests assumed specific admin credentials. 
* I reordered the SMTP configuration values so they have the same order of fields like the Email tab. 
* IMHO the optional settings can be left as-is, so there might be no need for this step. Can it be removed? 
* Using test-connection requires an admin email to be set. I've updated the item. 

Please review my changes and the items above. If everything is aligned, please mark this issue as ready-for-review.

ruchikajha95:
@ahus1 changes looks good to me .Marking the PR readyfo review.

ahus1 (approved):

Issue #46196, opened 2026-02-11
[OID4VCI] User self-issued credential: Issuance initiated from web portal application


mposolda:
### Description

Flow "User initiates Issuance from Keycloak Portal" from [this discussion](https://github.com/keycloak/keycloak/discussions/45626) (with a note that "Keycloak Portal" can be 3rd-party web portal application secured by Keycloak).

User authenticates to some portal (for example university portal), which is Keycloak application. Portal presents available credentials for self-issuance (For example "education-certificate"). User selects some credential and then is redirected to Keycloak OIDC authorization code flow, which would possibly re-authenticate him (if needed) and then present him page with credential offer (QR code or deep-link). User can then claim the credential to his wallet, which might be on same or different device than the mentioned web portal application

Note: This might be implemented with [AIA](https://www.keycloak.org/docs/latest/server_admin/index.html#con-aia_server_administration_guide) . Keycloak application can redirect to Keycloak with AIA and Keycloak will be able to display the screen with credential offer (QR code for wallet on  different device, deep-link for wallet on the same device)

PR #46199, opened 2026-02-11
[OID4VCI] Revisit and fix /credential_offer_uri endpoint

### LINKED ISSUE #45005: [OID4VCI] Revisit and fix /credential_offer_uri endpoint
### Before reporting an issue

### Area

oid4vc

### Describe the bug

The endpoint creates an URI to the credential offer or a QR code (i.e. mutually exclusive). 
Instead, the Response should (optionally) include the QR code such that only one CredentialOfferState is created internally.

### Version

latest

tdiesler:
closes #45005

Discussion: #46202


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -339,220 +342,276 @@ public Response getCNonce() {
...
+                .success();
 
-        return switch (type) {
-            case URI -> getOfferUriAsUri(offerState.getNonce());
-            case QR_CODE -> getOfferUriAsQr(offerState.getNonce(), width, height);
-        };
-    }
+        CredentialOfferURI credOfferURI = new CredentialOfferURI()
+                .setIssuer(credOffer.getCredentialIssuer() + "/protocol/" + OID4VC_PROTOCOL + "/" + CREDENTIAL_OFFER_PATH)
+                .setTxCode(offerState.getTxCode())
```
mposolda: We should not send `tx_code` in a response this way. This would make security of this REST endpoint even more broken than it is today.

The `tx_code` should be shared just with the user. Not with the other parties like administrator or "DataHolder / Issuer portal" application.

In Keycloak, the only reliable way to share things asynchronously with users is via email and that should be default how `tx_code` would be sent to the users (maybe we can have SPI to consider other ways if someone prefers SMS for example or something else like you're doing here, but that is likely out of scope of this PR).

Ideally, there should be email sent to the user. Once user clicks on the link from the email, Keycloak will display him the page with `tx_code` from where he can get it and fill inside wallet, which applies for pre-auth codes. @stianst suggested to use action-tokens flow instead of sending directly action-token by email for better security. Some thoughts about this are in this issue: https://github.com/keycloak/keycloak/issues/46269 .

This is additional work and might be not so trivial to implement (if you're not familiar with action-tokens and how they are used in KC). I can help (possibly somewhen next week) and send the initial PR for tx_code support somewhen next week, which then you would be able to consume more easily in this PR. Planning to do that after initial work for https://github.com/keycloak/keycloak/issues/46196 .

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -95,33 +97,48 @@ public Response process(Context context) {
...
         }
+
+        String expTxCode = offerState.getTxCode();
+        if (expTxCode != null) {
+            if (Strings.isEmpty(txCode)) {
+                event.error(Errors.MISSING_TX_CODE);
+                throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_GRANT,
+                        "Missing TxCode", Response.Status.BAD_REQUEST);
+            }
+            if (!expTxCode.equals(txCode)) {
```
mposolda: Minor: Is it possible to use `MessageDigest.isEqual` when comparing this? We had some reports in the past about "Timing attacks" when using plain equals for comparing security related secrets. See for example `OIDCClientSecretConfigWrapper.validateRotatedSecret` for how it is used in some existing codebase.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -339,220 +342,276 @@ public Response getCNonce() {
...
      */
     @GET
     @Produces({MediaType.APPLICATION_JSON, RESPONSE_TYPE_IMG_PNG})
-    @Path("credential-offer-uri")
-    public Response getCredentialOfferURI(
+    @Path(CREATE_CREDENTIAL_OFFER_PATH)
+    public Response createCredentialOffer(
             @QueryParam("credential_configuration_id") String credConfigId,
-            @QueryParam("pre_authorized") @DefaultValue("true") boolean preAuthorized,
-            @QueryParam("client_id") String appClientId,
```
mposolda, line 401: I think `client_id` should be still possible and would be good to keep it. The credential-offer permission might not be limited to the single wallet/client, but there are also use-cases where it might be limited to single client.

In relation to that: [Specification](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-client-metadata) defines `credential_offer_endpoint` as additional field in client's metadata where credential offer could be eventually sent. We do not have support for this endpoint, but it indicates that specification counts with the fact that credential-offer could be sent directly to the wallet, which indicates that it is used for particular wallet.

mposolda (commented):
@tdiesler Thanks for the PR! Did not reviewed whole PR (as I see it depends on the other PRs), but added some comments inline (especially the `tx_code` is the most important one).

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: The purpose of this endpoint is to create the credential offer - it is not concerned about how the offer (or parts of it) are delivered to the target value. We can make the return of this endpoint void and have a another endpoint that fetches credential offer state. It would however not fix the issue in principal - just delegate it to some other endpoint.

Please don't forget, the pre-auth grant is designed for cases where a wallet cannot do alternative authorization steps. The assumption that the holder can authenticate with Keycloak in order to use a pre-auth offer is likely flawed - if it could, there would be no need to use a pre-auth offer.

My stance on this is that access to this endpoint must be protected (for the targeted pre-auth case) like any other rest endpoint that allows an admin to do security sensitive stuff (e.g. modify a user's password). the pre-auth code should give access to the offered credential (and nothing else). Getting hold of the pre-auth code would be equivalent to getting hold of any credential that is bound to someone else.

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
tdiesler: nice - will do

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: Let's perhaps do the targeted clientId after preview when we have an actual use case that demands it. 

The wallet is not the oauth client.

Yes, we can talk about `credential_offer_endpoint` (perhaps also after preview) it is however unrelated to targetClient which names the oauth clientId that the (pre-auth) offer is good for.

I removed it for security reasons. The user that has privileges to create a targeted offer (i.e. has the create_credential_offer role) has that privilege only for the current clientId - he cannot create a targeted offer for a clientId that he himself has no access to.

tdiesler:
No longer depends on other PRs

tdiesler:
Before I restore `targetClient`, to allow a user with the `credential_offer_create` role to create (pre-authorized) offers for any target clientId (also for those that the user itself has no access too), I wanted to ask whether that is really wanted. Although it would not be a big change, it might be better to do that as a follow up for this PR. It would give us a little more time to try what we have and ponder about how credential offers should work across clients.

Also, before I remove the `tx_code` return value, we would need an alternative way to read that value from  the newly created `CredentialOfferState`. Perhaps we can also do that gradually and replace the current `tx_code` handling, i.e. "who creates the offer can see it", with an improved solution in a follow up PR.

Here is what you get with this PR ...

* self issued credential offers
* single internal offer state for uri + qr (so far qr requires new call and creates new offer)
* tx_code support in pre-auth grant and access token request

We recognize that there is still work to do - for example 

* potentially restore target client parameter
* two-step handling for create offer + read offer state
* more fine grained permissions, instead of single `credential_offer_create` role

This PR is likely a dependency of #46327
UI that builds on top of the credential-offer-create endpoint should probably be able to use this one.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
mposolda: > Let's perhaps do the targeted clientId after preview when we have an actual use case that demands it.

> The wallet is not the oauth client.

I think the wallet is client and should be represented by the client in Keycloak. This is assumed in multiple places of the specs and it is clearly mentioned in the [specs chapter 2](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-2 ) in the definition of wallet . Especially `In the context of this specification, the Wallet acts as an OAuth 2.0 Client (see [RFC6749]) and obtains an Access Token to access an OAuth 2.0 Resource Server (Credential Endpoint).` .

The wallet, which is able to do browser authentication with "Authorization code" flow can be regular OIDC client with "Standard flow enabled" and "OID4VCI enabled" . The wallet, which can do only pre-authorized code grant, should be represented by client with "OID4VCI Enabled" and with "PRe-authorized grant enabled" (new switch on the clients. See https://github.com/keycloak/keycloak/issues/46240 for the details).

Also note that pre-authorized grant request itself can use the parameters like `client_id` or even client authentication. The specs does not mandate, but also does not discourage it as seen [in section 6.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-6.1) . This allows that credential-offer does not need to have `client_id` in itself, which would mean that client will need to be identified in pre-authorized grant by the client authentication or the parameter like `client_id` .
 
It is probably OK for me to remove the `client_id` parameter in this REST endpoint in this PR if you prefer. However something to address in the future as we have the requirement for it already within Keycloak team.

I see that credential-permission / credential-offer could be target either to:
- Concrete client. In that case, the grant request (authorization-code or pre-authorized code) is limited only to that single client. In this case, the pre-authorized grant request does not need to include parameter `client_id` as it can use the client, which has available permission of this user
- Not specific to concrete client. In that case, the grant request will need to identify the client_id of the wallet . The pre-authorized grant request will need to include parameter `client_id` .

mposolda:
> Before I restore `targetClient`, to allow a user with the `credential_offer_create` role to create (pre-authorized) offers for any target clientId (also for those that the user itself has no access too), I wanted to ask whether that is really wanted. Although it would not be a big change, it might be better to do that as a follow up for this PR. It would give us a little more time to try what we have and ponder about how credential offers should work across clients.
> 
> Also, before I remove the `tx_code` return value, we would need an alternative way to read that value from the newly created `CredentialOfferState`. Perhaps we can also do that gradually and replace the current `tx_code` handling, i.e. "who creates the offer can see it", with an improved solution in a follow up PR.
> 
> Here is what you get with this PR ...
> 
>     * self issued credential offers
> 
>     * single internal offer state for uri + qr (so far qr requires new call and creates new offer)
> 
>     * tx_code support in pre-auth grant and access token request

Yeah, the first two points are great additions IMO. Especially the first one (the uri+qr can be always done on the issuer-portal side as when only the "uri" is returned, the application can still create QR code in themselves programaticaly with the use of some library like for example Google ZXing for Java. But I see it is more convenient to possibly have ability to return both at the same time).

For the `tx_code` support, it would be good to remove it from the `CredentialOfferURI` class in this PR as `tx_code` should be shared to the user by email at this point. I can see options like: 
1) Remove it from `CredentialOfferURI`  but keep your code for `tx_code` support on other places (classes like `PreAuthorizedCodeGrantType`, `PreAuthorizedCodeGrantRequest` , `CredentialOfferUriRequest` , `CredentialOfferState` etc). This means that it will not be possible to test `tx_code` yet with this PR. But some building-blocks would be prepared for supporting `tx_code` in the follow-up PRs.
2) I can send PR to your branch, which would implement `tx_code` by action-tokens sent in the email.

Would probably see (1) as better, so we can divide the work into multiple PRs. WDYT?

> 
> We recognize that there is still work to do - for example
> 
>     * potentially restore target client parameter
> 
>     * two-step handling for create offer + read offer state
> 
>     * more fine grained permissions, instead of single `credential_offer_create` role

Yes, this is work to do, but possibly for the follow-ups. 

What we discussed within Keycloak team is to:

- Introduce "Credential permission" concept in addition to "credential offer" . The administrator can create "Credential permission" in the admin console (and in the admin REST API) for the specific user to be able to obtain specific credential. This would address the fine-grained admin permissions point as well as two-step handling for create offer + read offer state. Some related issues: #46246 , #46200 .

> 
> 
> This PR is likely a dependency of #46327 UI that builds on top of the credential-offer-create endpoint should probably be able to use this one.

I don't see those PRs as being dependency of each other. UI from #46327 should not need to use the REST endpoint of this one. The login themes pages in general are not using REST endpoints as those are built by Keycloak itself. Also the access-token is not available yet for the user at the time when this page is rendered (although user was in fact authenticated in Keycloak).

But I can see it can probably re-use some Java code from both places though.

tdiesler:
Regading the `tx-code` we should probably do it the other way around (i.e. keep it as a return to the caller) so we can continue testing it. In a follow-up it can be removed it from the return with some other access option for the caller.

That other option could be some endpoint that is only available to test cases.

mposolda:
> Regading the `tx-code` we should probably do it the other way around (i.e. keep it as a return to the caller) so we can continue testing it. In a follow-up it can be removed it from the return with some other access option for the caller.
> 
> That other option could be some endpoint that is only available to test cases.

I don't think the `tx_code` should be returned to the caller as it has quite bad security implications and I don't see any point in testing that. The `tx_code` should not be shared to other party besides issuer (Keycloak) and user himself. If none of the options I've proposed [above](https://github.com/keycloak/keycloak/pull/46199#issuecomment-3920209102) works for you, the other option might be to remove `tx_code` from this PR entirely and implement it properly as a follow-up. I prefer (1) as some pieces you add in this PR regarding `tx_code` are good IMO and can be re-used by the follow-up PR though.

tdiesler:
> > Regading the `tx-code` we should probably do it the other way around (i.e. keep it as a return to the caller) so we can continue testing it. In a follow-up it can be removed it from the return with some other access option for the caller.
> > That other option could be some endpoint that is only available to test cases.
> 
> I don't think the `tx_code` should be returned to the caller as it has quite bad security implications and I don't see any point in testing that. The `tx_code` should not be shared to other party besides issuer (Keycloak) and user himself. If none of the options I've proposed [above](https://github.com/keycloak/keycloak/pull/46199#issuecomment-3920209102) works for you, the other option might be to remove `tx_code` from this PR entirely and implement it properly as a follow-up. I prefer (1) as some pieces you add in this PR regarding `tx_code` are good IMO and can be re-used by the follow-up PR though.

Ok, done - it is now redacted with test coverage preserved.

mposolda:
> > > Regading the `tx-code` we should probably do it the other way around (i.e. keep it as a return to the caller) so we can continue testing it. In a follow-up it can be removed it from the return with some other access option for the caller.
> > > That other option could be some endpoint that is only available to test cases.
> > 
> > 
> > I don't think the `tx_code` should be returned to the caller as it has quite bad security implications and I don't see any point in testing that. The `tx_code` should not be shared to other party besides issuer (Keycloak) and user himself. If none of the options I've proposed [above](https://github.com/keycloak/keycloak/pull/46199#issuecomment-3920209102) works for you, the other option might be to remove `tx_code` from this PR entirely and implement it properly as a follow-up. I prefer (1) as some pieces you add in this PR regarding `tx_code` are good IMO and can be re-used by the follow-up PR though.
> 
> Ok, done - it is now redacted with test coverage preserved.

Thanks. However can you please remove `txCode` entirely from `CredentialOfferUri` instead of make it "redacted" ? Would be better to avoid having that `txCode` property entirely from `CredentialOfferUri` as not seeing much point in keeping it.

tdiesler:
It is redacted to indicate that it exists on the server or to indicate that the pre-auth code in the offer is useless on its own. Also, testing for a redacted tx_code has value already, imho. How about we remove it when there actually is an alternative way to see it?

mposolda:
> It is redacted to indicate that it exists on the server or to indicate that the pre-auth code in the offer is useless on its own. Also, testing for a redacted tx_code has value already, imho. How about we remove it when there actually is an alternative way to see it?

Yes, automated tests are OK for now until we have `txCode` supported more properly. However your automated tests for `tx_code`, which use `TestingResourceProvider`, should be still possible even if you remove `txCode` from `CredentialOfferUri` class.

tdiesler:
I want the caller to know that the system has created a value. If I remove it, the caller is left wondering whether the tx_code parameter had any effect. Can we not leave it as it is? A `******` value carries information and is no harm to anybody. Again, we can remove it as soon as there is anything better than `******`

mposolda:
@tdiesler Not seeing much value in that, but OK for now. I will take a look at https://github.com/keycloak/keycloak/issues/46518 as a follow-up and hope to send PR later this week.

mposolda (approved):

PR #46219, opened 2026-02-11
Missing translations for role descriptions

### LINKED ISSUE #22569: Provide descriptions for default realm-management roles
### Description

In different contexts in the administration console, descriptions of realm-management roles just appear as i18n strings inside ${} (e.g: see attachment, taken from the assign role pop-up). Proposed enhancement is to provide descriptions for all realm-management roles in the locales supported by keycloak

descriptions_missing

### Discussion

_No response_

### Motivation

Adding meaningful descriptions can help give clarity on the rights associated with the default realm-management roles

### Details

_No response_

mabartos:
- Closes #22569

@ssilvert @edewit Could you please check it? 

@ahus1 How does it work with Weblate? For the account, there are already some translations in different languages, but this PR completes the list. What's the approach on registering these translations to Weblate? Is it sufficient to only have it in the `messages_en.properties` and it automatically adds the missing records? Do you think I should also add these known records for different languages into the admin messages? Thanks a lot!


ahus1:
@mabartos:

> Is it sufficient to only have it in the messages_en.properties and it [Weblate] automatically adds the missing records? 

Yes, this is the way.

> Do you think I should also add these known records for different languages into the admin messages?

Please avoid for those translations managed via Weblate as it can lead to conflicts due to concurrent edits. If you want to contribute translations to languages not translated via Weblate, consider creating a follow-up PR as you would require reviews for each language from a native speaker.

edewit (approved):

ahus1 (approved):

PR #46224, opened 2026-02-11
Reduce database calls for evaluation of client roles

### LINKED ISSUE #43726: Slow evaluation of roles with dots for role mapper and others
### Before reporting an issue

### Area

identity-brokering

### Describe the bug

When using identity brokering and role mappers, there are multiple attempts to resolve roles when either the role or the client contains a dot.  

This slows down Keycloak as those database requests aren't cached, and as they are serialized to avoid overloading the database. 

### Version

main

### Regression

### Expected behavior

When setting up a role mapper for SAML or another other provider, there should be one lookup in the database for the role, not multiple ones. And they should be cached.

### Actual behavior

For each dot (`.`) in the role name, there is a unnecessary lookup in the database. And Keycloak will synchronize on that calls for the client ID, so those calls will slow down the throughput.  

### How to Reproduce?

Analyze the code at: 

https://github.com/keycloak/keycloak/blob/0ff7d551dd2aab42da9cfdb03a5267a73d4e470e/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java#L945-L955

### Anything else?

_No response_

pruivo:
Closes #43726

The previous algorithm made `n+1` database calls, where `n` is the number of dots in the client/role name. 
This PR changes to a flat 2 database calls.

### DIFF
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
@@ -126,7 +126,8 @@ public final class KeycloakModelUtils {
         var metrics = new CaffeineStatsCounter(Metrics.globalRegistry, "role.name.cache");
         var cache = Caffeine.newBuilder()
                 .maximumSize(10000)
-                .expireAfterAccess(Duration.ofHours(1))
+                // do not keep entries forever, as the combination of realm and client roles might lead to different matches eventually
+                .expireAfterWrite(Duration.ofHours(1))
                 .recordStats(() -> metrics)
                 .<RoleCacheKey, RoleCacheValue>build();
         metrics.registerSizeMetric(cache);
@@ -1080,12 +1081,18 @@ public final class KeycloakModelUtils {
             if (cached.clientId() != null) {
                 var client = realm.getClientByClientId(cached.clientId());
                 if (client != null) {
-                    return new CachedRole(client.getRole(cached.roleName()));
+                    RoleModel role = client.getRole(cached.roleName());
+                    if (role != null) {
+                        return new CachedRole(role);
+                    }
                 }
             } else {
-                return new CachedRole(realm.getRole(cached.roleName()));
+                RoleModel role = realm.getRole(cached.roleName());
+                if (role != null) {
+                    return new CachedRole(role);
+                }
             }
-            // client not found, invalidate cache
+            // client or role not found, invalidate cache
             ROLE_NAME_FROM_STRING_CACHE.invalidate(cacheKey);
         }
         return null;
--- a/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
+++ b/server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
@@ -23,6 +23,7 @@ import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.cert.X509Certificate;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Calendar;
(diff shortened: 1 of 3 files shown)


server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
```diff
@@ -102,20 +107,33 @@
...
+        var cache = Caffeine.newBuilder()
+                .maximumSize(10000)
+                // do not keep entries forever, as the combination of realm and client roles might lead to different matches eventually
+                .expireAfterWrite(Duration.ofHours(1))
+                .recordStats(() -> metrics)
+                .<RoleCacheKey, RoleCacheValue>build();
+        metrics.registerSizeMetric(cache);
+        ROLE_NAME_FROM_STRING_CACHE = cache;
+    }
+
```
ahus1, line 136: Could you please try to move some of this to the `DefaultAlternativeLookupProvider`? 

You can get hold of the session using `KeycloakSessionUtil.getKeycloakSession()` ... this would make sure that there is no leaking of information between session factories.

server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
```diff
@@ -999,29 +1017,117 @@ public static RoleModel getRoleFromString(RealmModel realm, String roleName) {
...
+    private static RoleModel getRoleFromStringCombinations(RealmModel realm, String[] roleNameParts, String fullRoleName) {
+        var role = getRoleFromStringCombinationsFromCache(realm, fullRoleName);
+        if (role != null) {
+            return role.cachedRole();
+        }
+        var startIdx = new AtomicInteger();
+        var clientIdToTest = findNonEmptyClientId(roleNameParts, startIdx);
+        var clients = realm.searchClientByClientIdStream(clientIdToTest, null, null)
+                .collect(Collectors.toMap(ClientModel::getClientId, Function.identity()));
```
ahus1, line 1055: I had a look and it will do a full table scan on the clients table as it will run a `LIKE` operator with `%` at the beginning and the end. 

A future enhancement might be that this method would allow to have percentage at the end, and then it could use an index range scan. Given that the results are now cached, this might be acceptable, still one would need to weight the different lookups. 

Sometimes client IDs are URLs (used in OIDC federation), so one could have a lot of clients starting with the same prefix before the dot, and then one would pull a lot of data from the database. 

I would like to be conservative here and would prefer the original code that loops over the dots as the response time would depend only on the number of dots in the role name, and not on the total number of clients in the database.

server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
```diff
@@ -1305,4 +1411,19 @@ public static List<String> getAcceptedClientScopeProtocols(ClientModel client) {
...
+    }
+
+    private record RoleCacheKey(String realmId, String roleNameString) {
+        private RoleCacheKey {
+            Objects.requireNonNull(realmId);
+            Objects.requireNonNull(roleNameString);
+        }
+    }
+
+    private record CachedRole(RoleModel cachedRole) {}
```
ahus1, line 1428: Could we go without the `CachedRole` record as a simplification?

ahus1 (changes requested):
Thank you for the pull request. Please see below for some comments.

server-spi-private/src/main/java/org/keycloak/models/utils/KeycloakModelUtils.java
```diff
@@ -999,29 +1017,117 @@ public static RoleModel getRoleFromString(RealmModel realm, String roleName) {
...
+        var cacheKey = new RoleCacheKey(realm.getId(), fullRoleName);
+        var cached = ROLE_NAME_FROM_STRING_CACHE.getIfPresent(cacheKey);
+        if (cached != null) {
+            if (cached.clientId() != null) {
+                var client = realm.getClientByClientId(cached.clientId());
+                if (client != null) {
+                    RoleModel role = client.getRole(cached.roleName());
+                    if (role != null) {
+                        return new CachedRole(role);
+                    }
```
pruivo, line 1087: @ahus1, this (and below) if check will introduce unnecessary invalidations. 
The search for the client id will return the same client, and it will invoke `client.getRole()` again.

ahus1: The role might have been removed from the client, therefore I thought it would be good to invalidate it. 
It is Friday and getting late, maybe I'm getting tired. Happy to discuss it more on Monday.

PR #46225, opened 2026-02-11
Show button in tokens tab when openId4vci is not enabled

### LINKED ISSUE #45818: There is no save button on the TokenTab
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When I'm on the token tab there are no save buttons

### Version

main

### Regression

### Expected behavior

There should be a save / cancel button on this tab in the realm settings like on the other tabs

### Actual behavior

There is none, it only appears when I enable `Verifiable Credentials`

### How to Reproduce?

Go to the TokenTab of realm settings and see 

### Anything else?

_No response_

rmartinc:
Closes #45818

Just changing the condition to always check that the feature and the realm are enabled for openid4vci. There is a test already, but it loads the realm with verify credentials enabled (and the problem was it was disabled).


graziang (approved):
@rmartinc thanks!

mposolda (approved):

PR #46237, opened 2026-02-11
Add createdAfter/createdBefore filter to GET /admin/realms/{realm}/users

### LINKED ISSUE #43829: Add createdTimestamp filter (before/after) to /admin/realms/{realm}/users
### Description

It would be extremely valuable if the Admin REST API for
[GET /admin/realms/{realm}/users](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_get_adminrealmsrealmusers)

supported filtering users by their creation date — for example via:

```
?createdAfter=2025-10-01T00:00:00Z
?createdBefore=2025-10-15T00:00:00Z

```

### Value Proposition

**Why this matters**

Currently, the only way to find users created within a certain time range (e.g. “users registered in the last 14 days”) is to fetch all users and filter client-side by createdTimestamp.
This quickly becomes inefficient for realms with thousands or millions of users, since every request requires a full scan with pagination.

**Example use cases**

- Analytics dashboards showing weekly or monthly new registrations
- Syncing newly registered users into CRM / mailing systems
- Admin reports or auditing tools that need to identify recent sign-ups
- Automated cleanup or verification workflows for recent accounts

### Goals

**Suggested behavior**

- Support ISO 8601 date/time or epoch timestamps.
- Combine logically with other query parameters (search, q, pagination).
- Return all users whose createdTimestamp lies after or before the given value(s).

### Non-Goals

This request does not aim to introduce new user metadata fields, modify existing timestamps, or change how users are created or stored.
It only proposes to add optional filtering capabilities (createdBefore, createdAfter) to the existing /users endpoint, without affecting current API behavior or backward compatibility.

### Discussion

_No response_

### Notes

_No response_

RafaelWO:
Add server-side filtering of users by creation timestamp on the admin REST API. This avoids the need to retrieve all users and filter client-side, which is inefficient for large realms.

Two optional query parameters are added to both the user list and count endpoints. They accept either ISO-8601 date strings (yyyy-MM-dd) or epoch milliseconds, consistent with the existing events API date filtering via DateUtil.

Closes #43829

### DIFF
--- a/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/UsersResource.java
+++ b/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/UsersResource.java
@@ -274,6 +274,40 @@ public interface UsersResource {
                   @QueryParam("enabled") Boolean enabled,
                   @QueryParam("q") String searchQuery);
 
+    /**
+     * Search for users based on the given filters, including creation timestamp filters.
+     *
+     * @param username a value contained in username
+     * @param firstName a value contained in first name
+     * @param lastName a value contained in last name
+     * @param email a value contained in email
+     * @param emailVerified whether the email has been verified
+     * @param idpAlias the alias of the Identity Provider
+     * @param idpUserId the userId at the Identity Provider
+     * @param firstResult the position of the first result to retrieve
+     * @param maxResults the maximum number of results to retrieve
+     * @param enabled only return enabled or disabled users
+     * @param briefRepresentation Only return basic information
+     * @param createdAfter only return users created after (inclusive) the given date, in ISO-8601 format (yyyy-MM-dd) or epoch milliseconds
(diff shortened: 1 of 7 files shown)


sguilhen (approved):
Thanks for the PR @RafaelWO , LGTM.

sguilhen:
@ahus1 Would you mind running the CI for this enhancement?

ahus1:
@RafaelWO - this PR has merge conflicts, please resolve. With merge conflicts, CI will not run.

RafaelWO:
Thanks for taking a look!

@ahus1 I rebased by branch onto main and resolved the conflicts :slightly_smiling_face:

pedroigor (commented):
Thanks, RafaelWO.

It should also work querying users using the admin console. We should be adding the `Created After` and `Created Date` attributes to the attribute selection select box by default. We can do it in a follow-up

pedroigor (approved):

PR #46241, opened 2026-02-11
Stop using jboss-parent

### LINKED ISSUE #40125: Stop using jboss-parent

Pepo48:
Closes: #40125

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,12 +19,6 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.jboss</groupId>
-        <artifactId>jboss-parent</artifactId>
-        <version>39</version>
-    </parent>
-
     <name>Keycloak</name>
     <description>
         Keycloak SSO
@@ -52,6 +46,8 @@
 
         <project.build-time>${timestamp}</project.build-time>
         <project.build-name>Keycloak</project.build-name>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
         <!-- Following properties are used for downstream build to alter final distribution archive file name and internal base directory of the archive. -->
         <dist.archive.file.prefix>keycloak</dist.archive.file.prefix>
@@ -190,6 +186,7 @@
 
         <!-- Maven Plugins -->
         <maven.plugins.version>3.13.1</maven.plugins.version>
+        <version.enforcer.plugin>3.6.2</version.enforcer.plugin>
         <replacer.plugin.version>1.4.1</replacer.plugin.version>
         <jboss.as.plugin.version>7.5.Final</jboss.as.plugin.version>
         <jmeter.plugin.version>1.9.0</jmeter.plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -202,6 +202,14 @@
         <openapi.generator.plugin.version>6.3.0</openapi.generator.plugin.version>
         <build-helper-maven-plugin.version>3.2.0</build-helper-maven-plugin.version>
         <spotless-plugin.version>3.0.0</spotless-plugin.version>
+        <maven-assembly-plugin.version>3.1.1</maven-assembly-plugin.version>
+        <maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
+        <maven-jar-plugin.version>3.5.0</maven-jar-plugin.version>
+        <maven-compiler-plugin.version>3.13.0</maven-compiler-plugin.version>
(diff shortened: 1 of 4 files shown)


vmuzikar:
@stianst @ASzc @drichtarik Does this work for you?

drichtarik:
We are running test builds and adding dependencies/plugins that are missing. This is going to take a couple of days but it's included in the upcoming sprint as our goal

stianst (changes requested):
This is not going to work for the release; we're relying on a few bits and pieces from jboss-parent there. Need to look into exactly what, but don't have time right now.

quarkus/tests/integration/pom.xml
```diff
@@ -126,6 +126,7 @@
                     <systemPropertyVariables>
                         <kc.quarkus.tests.dist>${kc.quarkus.tests.dist}</kc.quarkus.tests.dist>
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                        <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
```
stianst: I'd suggest dropping this; if tests need to put temporary files within the target directory they should do that explicitly, not by changing `java.io.tmpdir`

Pepo48:
@stianst sorry for dragging you into this again, I hadn’t noticed that you were tagged. I’ve already discussed it with Denis and as he mentioned above, they’re currently looking into what needs to be done from the downstream perspective. Once they finish, I’ll be ready to use the testing org to verify whether anything else needs to be adjusted to fix the release workflows at that level as well.

pskopek:
The latest changes successfully tested on keycloak-rel-testing org.
https://github.com/keycloak-rel-testing/keycloak-rel/actions/runs/22576229928

pskopek (approved):
LGTM!

stianst:
> The latest changes successfully tested on keycloak-rel-testing org. https://github.com/keycloak-rel-testing/keycloak-rel/actions/runs/22576229928

I'm surprised - was really certain gpg or something around repo config wouldn't be right.

I'll take a look at this next week just to make double sure; as I really don't want to fight issues around this when we do the 26.6 release ;)

stianst (approved):

pom.xml
```diff
@@ -205,6 +202,14 @@
         <openapi.generator.plugin.version>6.3.0</openapi.generator.plugin.version>
         <build-helper-maven-plugin.version>3.2.0</build-helper-maven-plugin.version>
         <spotless-plugin.version>3.0.0</spotless-plugin.version>
+        <maven-assembly-plugin.version>3.1.1</maven-assembly-plugin.version>
+        <maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
+        <maven-jar-plugin.version>3.5.0</maven-jar-plugin.version>
+        <maven-compiler-plugin.version>3.13.0</maven-compiler-plugin.version>
```
michalvavrik, line 208: Quarkus is using `maven-compiler` `3.15.0`, while you picked `3.13.0`. Is this in order to align with Maven core and other plugins? Because only information in your PR / commits I found was "_Should match the versions from jboss-parent to prevent conflicting versions._" which does not add up, because JBoss Parent version 39 currently uses 3.8.1.

Pepo48: @drichtarik this got postponed a bit and now it's ready to be merged. Should we keep these versions as they are or should we bump some of them? Thanks!

ASzc: For product all we're trying to establish with the introduction of these lines, is to ensure that these plugin versions are defined. Removing jboss-parent otherwise allows them to float, which causes problems.

3.13 vs 3.15, I don't think it would matter, but we'd need to do more trial builds to confirm

michalvavrik: > 3.13 vs 3.15, I don't think it would matter, but we'd need to do more trial builds to confirm

FWIW the only reason why I was commenting here was because I tried to use `annotationProcessorPathsUseDepMgmt` which was introduced in 3.12. I am fine with 3.13, but it seems random and I think explanation on why these exact plugin versions were picked is missing. As long as there is a reason, np.

ASzc: I'll check with the team

stianst: We could perhaps split the two things; since we've tested this PR both for upstream and downstream releases we can keep it as is, then plan to upgrade plugins as a follow-up?

Pepo48: I agree.

ASzc: Yes. This one has been open for a long time. If you want to merge this PR as is or with 3.15, that's fine. As long as there is a dependency definition in the keycloak POMs, we can override the version at product build time if it breaks something, upstreaming fixes as appropriate.

ASzc (approved):

vmuzikar (commented):
@Pepo48 Sorry if I missed this in the discussion but why were the test changes needed?

Pepo48:
> @Pepo48 Sorry if I missed this in the discussion but why were the test changes needed?

Because of a surefire plugin config change, see https://github.com/keycloak/keycloak/pull/45616#issuecomment-3782042984 (the paragraph about the test changes)

vmuzikar (approved):
> Because of a surefire plugin config change, see [#45616 (comment)](https://github.com/keycloak/keycloak/pull/45616#issuecomment-3782042984) (the paragraph about the test changes)

Right, I forgot about that. Thanks, LGTM.

PR #46253, opened 2026-02-11
fix: normalizing the baseUri to end with /

shawkins:
closes: #46235

When using the simple hostname, the quarkus logic ensures the baseUri ends with /. So we should ensure the same when using a full hostname url.


vmuzikar (approved):
LGTM, thanks.

PR #46254, opened 2026-02-11
fix: updating the fix for docker ulimits

### LINKED ISSUE #44232: Improve building containers in Minikube

shawkins:
it will now be applied by default in the operator image building

To be considered as a replacement for https://github.com/keycloak/keycloak/pull/44233 since the updates directly in the Dockerfiles were getting messy.

closes: #44232


ahus1 (approved):
Thank you for taking up this one!

vmuzikar (approved):

PR #46257, opened 2026-02-12
Admin API V2: allow to generate secret for clients with authentication method 'client-secret'

### LINKED ISSUE #46136: Support for generating client secret in Client v2
### Description

In addition to manually specifying client secrets, add a way to generate that secret. Generating secret, however, clashes with the declarative config approach to some extent, that is fundamental for the Client v2. The requirements would be:
* Possibility to manually specify client secret.
   * Needs to support declarative configuration.
   * The secret is part of the OIDC Client Representation so that it can be managed together with the client via a single PUT request.
* Possibility to leave the secret generation to server.
   * It needs to be possible for the user to retrieve that secret after creation.
   * It needs to be possible to re-generate that secret.

Proposal (similar to v1):
* When creating/updating a client, if `auth.method` is `client-secret` AND `auth.secret` is unspecified:
   * If no secret is persisted: generate new secret.
   * If secret already exists: do nothing. This might be slightly non-declarative approach – if we PUT client without a secret for an existing client, the secret is not removed. But that would be invalid in any case (with `auth.method` being `client-secret`).
* Secret is always returned when fetching a client, regardless if generated or not.
* The Secret can be re-generated by doing `POST /clients/v2/my-client-id/generate-secret`.
* Operator would always require providing a Secret Ref with `method` set to `client-secret` for simplicity.

michalvavrik:
* Closes: https://github.com/keycloak/keycloak/issues/46136

### DIFF
--- a/rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/validation/ClientSecretNotBlankValidator.java
+++ b/rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/validation/ClientSecretNotBlankValidator.java
@@ -1,0 +1,35 @@
+package org.keycloak.representations.admin.v2.validation;
+
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+
+import org.keycloak.authentication.authenticators.client.ClientIdAndSecretAuthenticator;
+import org.keycloak.authentication.authenticators.client.JWTClientSecretAuthenticator;
+import org.keycloak.representations.admin.v2.OIDCClientRepresentation;
+import org.keycloak.utils.StringUtil;
+
+/**
+ * Validates that requires the {@link OIDCClientRepresentation.Auth#getSecret()} is not blank
+ * when {@link OIDCClientRepresentation.Auth#getMethod()} is the (JWT) client secret.
+ */
+public class ClientSecretNotBlankValidator implements ConstraintValidator<ClientSecretNotBlank, OIDCClientRepresentation.Auth> {
+
+    @Override
+    public boolean isValid(OIDCClientRepresentation.Auth auth, ConstraintValidatorContext context) {
+        if (auth != null && isClientSecret(auth.getMethod())) {
+            if (StringUtil.isBlank(auth.getSecret())) {
+                context.disableDefaultConstraintViolation();
+                context.buildConstraintViolationWithTemplate("must not be blank when authentication method requires a secret")
+                        .addPropertyNode("secret")
+                        .addConstraintViolation();
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static boolean isClientSecret(String method) {
+        return ClientIdAndSecretAuthenticator.PROVIDER_ID.equals(method)
+                || JWTClientSecretAuthenticator.PROVIDER_ID.equals(method);
+    }
+}
--- a/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
+++ b/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
@@ -17,6 +17,7 @@
 
 package org.keycloak.tests.admin.client.v2;
 
+import java.io.IOException;
 import java.util.List;
 import java.util.Set;
 
@@ -24,6 +25,9 @@ import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MediaType;
 
 import org.keycloak.admin.client.Keycloak;
+import org.keycloak.authentication.authenticators.client.ClientIdAndSecretAuthenticator;
+import org.keycloak.authentication.authenticators.client.JWTClientAuthenticator;
+import org.keycloak.authentication.authenticators.client.JWTClientSecretAuthenticator;
 import org.keycloak.common.Profile;
 import org.keycloak.representations.admin.v2.BaseClientRepresentation;
 import org.keycloak.representations.admin.v2.OIDCClientRepresentation;
@@ -43,6 +47,7 @@ import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
 import com.fasterxml.jackson.core.type.TypeReference;
 import org.apache.http.HttpMessage;
 import org.apache.http.client.methods.HttpDelete;
+import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpOptions;
 import org.apache.http.client.methods.HttpPatch;
@@ -54,6 +59,8 @@ import org.apache.http.util.EntityUtils;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import static org.keycloak.services.cors.Cors.ACCESS_CONTROL_ALLOW_METHODS;
 import static org.keycloak.services.cors.Cors.ORIGIN_HEADER;
@@ -61,8 +68,11 @@ import static org.keycloak.services.cors.Cors.ORIGIN_HEADER;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.emptyOrNullString;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 @KeycloakIntegrationTest(config = ClientApiV2Test.AdminV2Config.class)
 public class ClientApiV2Test extends AbstractClientApiV2Test{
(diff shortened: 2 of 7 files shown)


rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/RepModelMapper.java
```diff
@@ -11,4 +13,6 @@ default U toModel(T rep) {
     }
 
     U toModel(T rep, U existingModel);
+
+    ClientRepresentation createOldClientRepresentation(T rep);
```
mabartos:
```suggestion
```

We should not add it here. The `RepModelMapper` is intended to be used also for other resources (future) like realms, users,... It is only a temporary approach, and should not be here - would be better to add it only to the BaseClientModelMapper, and in the service cast the mapper to this. 

@michalvavrik WDYT?

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
```diff
@@ -110,6 +111,21 @@ public void deleteClient() {
         clientResource.deleteClient();
     }
 
+    @Override
+    public String generateSecret() {
+        if (clientResource == null) {
+            throw new NotFoundException("Cannot find the specified client");
```
mabartos: JFYI - via this, you could also phish client IDs, but it does not have to be solved here. Related task: 
- https://github.com/keycloak/keycloak/issues/46010

mabartos (commented):
@michalvavrik Just a brief syntactic review for now.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
michalvavrik: > JFYI - via this, you could also phish client IDs, but it does not have to be solved here. Related task:
> 
>     * [Missing anti-ID phishing check for getting client #46010](https://github.com/keycloak/keycloak/issues/46010)

Oh, I didn't realize this can be dealt with on the endpoint level, I thought these kind of issues are handled by the endpoint security (layer before this one, globally).

But I just copied this IF block from the `delete` operation, so there is at least one other endpoint that could be problematic. Thanks for bringing it up.

michalvavrik:
Re CI failures: `LoggingDistTest.httpAccessLogMaskedCookiesDiffFormat` is passing for me locally, I'll just not investigate for now. Maybe it is flaky and it will be fixed when pushing changes.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/RepModelMapper.java
michalvavrik: I expected this will be considered an issue. I am fine with casting that mapper. This change is basically about avoiding `instanceof`, hence keeping extendability. Casting it to `BaseClientModelMapper` at this point is safe.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
michalvavrik: Which may mean that I am missing 403 as well, based on the issue description " It means that if there's no clients().canList() permission, the 403 should be returned instead of 404.". Because I somehow expected there is some filter before this one and I didn't see anywhere testing 403 in that test class, so I followed the suit. I'll add some test for 403 to see how it works. My bad.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/RepModelMapper.java
michalvavrik: Done.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/DefaultClientApi.java
michalvavrik: I have added test for 403 and 401 and they are happening before 404. So I think the phishing is only possible by users with a permission to generate secret or, there is another 403 coming for more fine grained-permissions. Unfortunately I am not that far understanding Keycloak code base to be sure without doing more legwork.

mabartos: > But I just copied this IF block from the delete operation, so there is at least one other endpoint that could be problematic. Thanks for bringing it up.

@michalvavrik Don't worry about the 403 now... We can discuss the best approach later with the issue I mentioned. It's just for our future reference, that another endpoint might be affected.

michalvavrik: Ok, thanks.

vmuzikar:
Putting on hold until the discussion in the parent issue is resolved.

michalvavrik:
Hey @mabartos @vmuzikar @shawkins this PR should now comply behavior defined by the linked issue updated descripiton. I know there are currently some GitHub PR failures in the main branch, so I run related tests locally and they are passing. Hence, shold be ready for your review. Thanks.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/OIDCClientModelMapper.java
```diff
@@ -91,4 +99,19 @@ private Set<String> getServiceAccountRoles(ClientModel client) {
...
+        var auth = client.getAuth();
+        if (auth != null && isClientSecret(auth.getMethod())) {
+            // this makes sure that client secret is generated for "create" methods if necessary
+            oldRepresentation.setPublicClient(false);
+            oldRepresentation.setClientAuthenticatorType(auth.getMethod());
+            oldRepresentation.setSecret(auth.getSecret());
+        }
+    }
+
+    private static boolean isClientSecret(String method) {
```
shawkins: @vmuzikar or other domain experts - is this check too specific? I still see usage of "client-secret-jwt" in the codebase. And given that method / provider type is freeform, this check could be brittle should another type using the secret field be added.

Alternatively are we somewhere mapping client-secret-jwt to client-secret? 

Or should there be an enum / provider introspection that tells us if the secret is required?

Depending on the answer, ClientSecretNotBlankValidator may need updated as well.

shawkins (commented):
The tests look great - I just want to double check our assumptions around the auth method.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/OIDCClientModelMapper.java
michalvavrik: Not question for me, so I'll let Vaclav to answer, but FYI - I realize this is very specific and I did it that way because the linked issue keeps repeating `auth.method is client-secret`. Maybe I misinterpreted it.

michalvavrik:
I have resolved merge conflicts.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/OIDCClientModelMapper.java
vmuzikar: The secret-jwt definitely needs a secret too. As far as I can tell, no other authenticator requires secret. We could of course add some "metadata" to the providers that would indicate if secret is required for a given authenticator or not. However, I do not think it's the way to go. Each authenticator requires specific fields most of which we are currently missing. Secret is just one of them. Since we don't really support providing custom authenticators, we can hardcode the validations on the service layer level.

So for now I'd just suggest to extend this to consider the secret-jwt as well but we'll need to add more authenticator-specific fields and validations in the future.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/BaseClientModelMapper.java
```diff
@@ -76,4 +77,22 @@ protected ClientModel createClientModel(BaseClientRepresentation rep) {
...
+    }
+
+    /**
+     * Allows to populate the old client representation with some values.
+     * This should be kept to minimum as we only want to maintain relationship between
+     * the model and the new representation.
+     */
+    protected void updateOldClientRepresentation(T client, ClientRepresentation oldRepresentation) {
+        // implemented by subclasses when needed
+    }
```
vmuzikar: I don't like this too much as it feels a bit not systematic approach but it's given the weird state that we need to first create an empty model. It will go away once we have a proper service layer.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
```diff
@@ -272,6 +279,10 @@ public void getClientsMixedProtocols() throws Exception {
 
         try (var response = client.execute(oidcRequest)) {
             assertEquals(201, response.getStatusLine().getStatusCode());
+
+            OIDCClientRepresentation representation = mapper.readValue(response.getEntity().getContent(), OIDCClientRepresentation.class);
+            assertThat(representation, notNullValue());
+            assertThat("Public client must not have authentication configuration", representation.getAuth(), nullValue());
```
vmuzikar: This should rather be in the `ClientApiV2ValidationTest`.

vmuzikar (commented):
@michalvavrik Thank you for the PR, nice job.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: > This should rather be in the `ClientApiV2ValidationTest`.

For a moment I thought that you meant all my tests, but now I am thinking you meant just these 2 additions to existing tests related to the public client. There is very thin border, you could say this about most of my tests. I'll just drop the 2 additions you are probably commenting on as I am already testing them anyway for OIDC version. I'll think about SAML one.

vmuzikar: Yup, I meant these two asserts.

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/OIDCClientModelMapper.java
michalvavrik: I have added the `client-secret-jwt`, but only now I mentioned that @vmuzikar is also referring to the `secret-jwt` in the last comment, so few minutes before I apply that as well. I won't make this PR draft as it will only take a little while.

michalvavrik: Although, I am not quite sure yet how is it meant. The `client-secret-jwt` can be adapted by  `org.keycloak.authentication.authenticators.client.JWTClientSecretAuthenticator#getAdapterConfiguration` to `secret-jwt` credential. But there is no authenticator using such a method and I didn't find many `secret-jwt` credential usage around Keycloak code base so far. Are you sure that it fits current structure of the  `org.keycloak.representations.admin.v2.OIDCClientRepresentation.Auth` @vmuzikar ?

Apart of that ^^, this PR is ready for a review. I had to make other changes to resolve merge conflicts, so if you read it before, you have to start from scratch. Thank you.

mabartos: Yep, the `client-secret-jwt` needs to be accounted for it as well. AFAIK, there's no difference in `client-secret`, and `client-secret-jwt` in terms of client representation/model/entity. The secret needs to be shared between server and client anyway, only the difference here is how the client authentication works on OAuth/OIDC requests level - either as a plain query param, or as a client assertion (JWT). 

But, there are other client authentication methods (and might be even more in the future), that'd need some sort of settings for that - like we see the expiration for the JWT, ... But as we're not gonna make changes on the model/storage layer, we're ok with adding multiple fields to the `auth`... So, I don't see a big problem here. 

And as Michal did, we just need to provide certain validation for client authenticators - like checking missing secret for the client-secret-based authenticators, or checking missing parameters for other authenticators. 

> Are you sure that it fits current structure of the org.keycloak.representations.admin.v2.OIDCClientRepresentation.Auth

The `auth.secret` is the main requirement for the `client-secret-jwt`. And currently, the defaults are used for the other params for the authenticator. I don't mind if these other params are part of the `auth` and accepted by the server - so probably better than providing some kind of Map<String,String> that accepts everything.

Or am I missing something different that you're concerned about?

michalvavrik: Thanks for the context.

> Or am I missing something different that you're concerned about?

I am not sure. Basically what you describe is done and waiting for a review. Vaclav referred to `secret-jwt`, which I think is a credential, not the authenticator. So if by that it was meant `client-secret-jwt`, that is done.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/validation/ClientSecretNotBlankValidator.java
```diff
@@ -0,0 +1,35 @@
...
+ * when {@link OIDCClientRepresentation.Auth#getMethod()} is the (JWT) client secret.
+ */
+public class ClientSecretNotBlankValidator implements ConstraintValidator<ClientSecretNotBlank, OIDCClientRepresentation.Auth> {
+
+    @Override
+    public boolean isValid(OIDCClientRepresentation.Auth auth, ConstraintValidatorContext context) {
+        if (auth != null && isClientSecret(auth.getMethod())) {
+            if (StringUtil.isBlank(auth.getSecret())) {
+                context.disableDefaultConstraintViolation();
+                context.buildConstraintViolationWithTemplate("must not be blank when authentication method is " + ClientIdAndSecretAuthenticator.PROVIDER_ID)
```
vmuzikar: We need to be generic here.
```suggestion
                context.buildConstraintViolationWithTemplate("must not be blank when authentication method requires a secret")
```

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
```diff
@@ -760,6 +770,278 @@ public void updateSamlClientWithInvalidRootUrl() throws Exception {
...
+
+        request = new HttpPatch(getClientApiUrl(clientId));
+        request.setHeader(HttpHeaders.CONTENT_TYPE, PatchTypeNames.JSON_MERGE);
+        OIDCClientRepresentation.Auth authWithoutSecret = new OIDCClientRepresentation.Auth();
+        authWithoutSecret.setMethod(authenticationMethod);
+        authWithoutSecret.setSecret(null);
+        OIDCClientRepresentation.Auth patchedAuth = getResultingAuthConfig(authWithoutSecret, request, clientId);
+        assertThat(patchedAuth, notNullValue());
+        String newlyGeneratedSecret = patchedAuth.getSecret();
+        assertThat(newlyGeneratedSecret, Matchers.not(emptyOrNullString()));
```
vmuzikar: Should we check the original and new secret are different? I.e. that the secret was generated?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
```diff
@@ -760,6 +770,278 @@ public void updateSamlClientWithInvalidRootUrl() throws Exception {
...
+        auth.setSecret("shush");
+
+        OIDCClientRepresentation.Auth createdAuth = getResultingAuthConfig(auth, request, clientId);
+        assertThat(createdAuth, notNullValue());
+        assertThat(createdAuth.getSecret(), is(auth.getSecret()));
+
+        request = new HttpPatch(getClientApiUrl(clientId));
+        request.setHeader(HttpHeaders.CONTENT_TYPE, PatchTypeNames.JSON_MERGE);
+        OIDCClientRepresentation.Auth authWithoutSecret = new OIDCClientRepresentation.Auth();
+        authWithoutSecret.setMethod(authenticationMethod);
```
vmuzikar: Ideally, we should not need to specify the auth method in patch requests if the auth method remains unchanged.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
```diff
@@ -760,6 +770,278 @@ public void updateSamlClientWithInvalidRootUrl() throws Exception {
...
+        authWithoutSecret.setSecret(null);
+        OIDCClientRepresentation.Auth patchedAuth = getResultingAuthConfig(authWithoutSecret, request, clientId);
+        assertThat(patchedAuth, notNullValue());
+        String newlyGeneratedSecret = patchedAuth.getSecret();
+        assertThat(newlyGeneratedSecret, Matchers.not(emptyOrNullString()));
```
vmuzikar: Same as above. Should we check the original and new secret are different? I.e. that the secret was generated?

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/validation/ClientSecretNotBlankValidator.java
michalvavrik: > We need to be generic here.

Good catch.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: > Should we check the original and new secret are different? I.e. that the secret was generated?

Definitely, it is important. I had this tested for that endpoint which generated client secret and didn't migrate it to new tests. Thank you

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: > Ideally, we should not need to specify the auth method in patch requests if the auth method remains unchanged.

Right, I think that can be complicated because current impl. does not merge nested objects, so after `JsonNode` is patched into existing method, we will definitely have no auth method.

I'll look into it, it is doable, I just remember that last time I tried it I couldn't avoid `instanceof`. It is a good catch though, I think you are right, but I also think that it is true for every nested object, not just `auth`.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: Definitely, thanks.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: because I didn't want to add there `@JsonMerge`. anyway, looking into it

michalvavrik: Bloody memory, I just remembered my concerns: if we go this way, there is another field on `Auth` object like `certificate` and if we go this way (merging nested objects) we will require thta when someone want to set `Auth#method` and `Auth#certificate`, they will have to patch `Auth#secret` as well. And in general, any other field they want to be `null`. It is not intuitive to me, but I can implement what you want. @vmuzikar @shawkins opinions?

mabartos: > they will have to patch Auth#secret as well

@michalvavrik Can you elaborate please? If there's the new field like certificate,jwt,.. they're basically attached to the `auth.method`. So it means, that they are not-relevant/not-used when the other auth.method is used. So I don't think the other fields need to be patched.

In admin console, you can set the params for specific auth.methods, that are persisted, and if the other method is used, they're just ignored. But when the admin decide to go to the previous auth.method, they can easily do it as they still have the data persisted there.  

Or what is the concern here?

michalvavrik: Hello,

consider initial state for `OIDCClientRepresentation`:

```
{
  auth: {
      method: "client-secret",
      secret: "hush-hush"
   }
}
```

> @michalvavrik Can you elaborate please? If there's the new field like certificate,jwt,.. they're basically attached to the auth.method. So it means, that they are not-relevant/not-used when the other auth.method is used. So I don't think the other fields need to be patched.

Right now, if you send request object to json merge patch endpoint with `org.keycloak.representations.admin.v2.OIDCClientRepresentation` :

```
{
   auth: {
     method: "new-method"
  }
}
```

You get OIDC CLient representation:

```
{
   auth: {
     method: "new-method"
  }
}
```

because json objects are not merged. If some `auth` object is present, it replaces the original `org.keycloak.representations.admin.v2.OIDCClientRepresentation#auth`.

So that doesn't play with @vmuzikar request to be able to ommit `org.keycloak.representations.admin.v2.OIDCClientRepresentation.Auth#method` if they want just to regenerate secret.

Now mitigation is to merge json objects, e.g. apply `@JsonMerge`. What that means that if the patch request object is:


```
{
   auth: {
     method: "whatever",
     certificate: "it's me, trust me bro"
  }
}
```

you would get:


```
{
   auth: {
     method: "whatever",
     secret: "hush-hush",
     certificate: "it's me, trust me bro"
  }
}
```

Is that expected outcome? I do not challenge it, I would simply appreciate if there were consensus on expected behavior.

> In admin console, you can set the params for specific auth.methods, that are persisted, and if the other method is used, they're just ignored. But when the admin decide to go to the previous auth.method, they can easily do it as they still have the data persisted there.

Keeping invalid combinations (like set of fields that cannot be all set as some values are ignored) should IMO lead to validation error. Allowing to set properties that are ignored is confusing at least. Sadly I don't have business logic so that I can challenge it (e.g. I'd think that if `secret` is set but ignored, you would still want to rotate it, in which case what is the point of keeping previous value?

I am definitely not able to determine expected behavior here, if you say that we should keep field values that are not used for given `auth#method`, I can simply implement it. I just need to hear there is consensus on it.

> Or what is the concern here?

Does it answer your question @mabartos ? If not, I'll elaborate more.

mabartos: @michalvavrik Thanks for the elaboration! I'll not be able to check it today, so probably on Mon.

michalvavrik: > @michalvavrik Thanks for the elaboration! I'll not be able to check it today, so probably on Mon.

NP, thanks for the information in your previous message.

vmuzikar: The merge patch should be recursive, i.e. handle nested fields as if they were top level fields. This means that if you send a patch with just `{auth: {secret: "foobar"}}`, `auth.method` (and all other fields in `auth`) would be preserved.

@shawkins Do you see any issues with that?

michalvavrik: thx!

shawkins: > The merge patch should be recursive, i.e. handle nested fields as if they were top level fields. This means that if you send a patch with just `{auth: {secret: "foobar"}}`, `auth.method` (and all other fields in `auth`) would be preserved.
> 
> @shawkins Do you see any issues with that?

That is how it should be functioning.

vmuzikar: @michalvavrik To get this merged, we can change the patching logic in a follow-up. It might be considered as out of scope of this PR anyway. WDYT?

rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/OIDCClientModelMapper.java
michalvavrik: Done @mabartos @vmuzikar @shawkins , ready for the next round.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: Done.

rest/admin-v2/api/src/main/java/org/keycloak/representations/admin/v2/validation/ClientSecretNotBlankValidator.java
michalvavrik: done.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: Done.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: > @michalvavrik To get this merged, we can change the patching logic in a follow-up. It might be considered as out of scope of this PR anyway. WDYT?

It works now as you described for `Auth` object, let's have a look.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/ClientApiV2Test.java
michalvavrik: Done.

michalvavrik:
PR CI failures are not related:

```
2026-03-02T12:21:34.7451301Z [INFO] �?� [libs/keycloak-admin-client:generate:openapi] exited with exit code 1. Output:
2026-03-02T12:21:34.7452080Z [INFO] 
2026-03-02T12:21:34.7452340Z [INFO] Request Failed.
2026-03-02T12:21:34.7452661Z [INFO] Status Code: 403
2026-03-02T12:21:34.7452992Z [INFO] Error: Request Failed.
2026-03-02T12:21:34.7453349Z [INFO] Status Code: 403
2026-03-02T12:21:34.7454564Z [INFO]     at RedirectableRequest.<anonymous> (D:\a\keycloak\keycloak\js\node_modules\.pnpm\@kiota-community+kiota-gen@1.0.2\node_modules\@kiota-community\kiota-gen\bin\kiota.js:70:31)
2026-03-02T12:21:34.7455966Z [INFO]     at RedirectableRequest.emit (node:events:508:28)
2026-03-02T12:21:34.7457267Z [INFO]     at RedirectableRequest._processResponse (D:\a\keycloak\keycloak\js\node_modules\.pnpm\follow-redirects@1.15.9\node_modules\follow-redirects\index.js:409:10)
2026-03-02T12:21:34.7459228Z [INFO]     at RedirectableRequest._onNativeResponse (D:\a\keycloak\keycloak\js\node_modules\.pnpm\follow-redirects@1.15.9\node_modules\follow-redirects\index.js:102:12)
2026-03-02T12:21:34.7460484Z [INFO]     at Object.onceWrapper (node:events:623:26)
2026-03-02T12:21:34.7461021Z [INFO]     at ClientRequest.emit (node:events:508:28)
2026-03-02T12:21:34.7461687Z [INFO]     at HTTPParser.parserOnIncomingClient (node:_http_client:772:27)
2026-03-02T12:21:34.7462500Z [INFO]     at HTTPParser.parserOnHeadersComplete (node:_http_common:122:17)
2026-03-02T12:21:34.7463246Z [INFO]     at TLSSocket.socketOnData (node:_http_client:614:22)
2026-03-02T12:21:34.7463814Z [INFO]     at TLSSocket.emit (node:events:508:28)
2026-03-02T12:21:34.7464349Z [INFO]   0% [0 / 6] [2 running] themes-vendor:build
2026-03-02T12:21:37.7577386Z [INFO]  33% [2 / 6] [0 running] [1 failed]
2026-03-02T12:21:37.7585013Z [INFO] �?� 1 script failed.
2026-03-02T12:21:37.7897084Z [INFO]  ELIFECYCLE  Command failed with exit code 1.
```

It only failed on Windows so far, so I'll not re-trigger CI as this PR does nothing specific for Windows and so on.

rest/admin-v2/api/src/main/java/org/keycloak/services/client/ClientService.java
```diff
@@ -48,5 +48,7 @@ default Stream<BaseClientRepresentation> getClients(RealmModel realm) {
 
     CreateOrUpdateResult createOrUpdate(RealmModel realm, BaseClientRepresentation client, boolean allowUpdate) throws ServiceException;
 
+    CreateOrUpdateResult putClient(RealmModel realm, BaseClientRepresentation client) throws ServiceException;
```
vmuzikar: We should not need a separate method for this. It should be `createOrUpdate` with `allowUpdate = true`.

michalvavrik: > We should not need a separate method for this. It should be createOrUpdate with allowUpdate = true.

If service layer contains validation, not REST resources and we have different validation for `allowUpdate` when it is PATCH and PUT, how do you suggest it will work? We need to know if it is put so that we know which validations apply. You asked that PUT don't allow to generate client secret, hence the validation @vmuzikar

michalvavrik: I am referring to the fact that validator was recently moved to the service, I think because I was resolving merge conflicts.

vmuzikar: Good point.

Thinking about it further, we should probably take a different approach. Instead of `createOrUpdate` on the interface, we should just have `create`, `put` and `patch`. `createOrUpdate` might remain unexposed on the interface. That would make it much cleaner IMHO.

mabartos: > Instead of createOrUpdate on the interface, we should just have create, put and patch.

@vmuzikar Yes, it might be a way. But the check if the resource should be created/updated via PUT should lie in the service layer, so even in the `put` service method, we'd need to distinguish on what to use. But I agree, that it might be probably nicer. Once we determine we need to create the resource via put, we just redirect to the `create`.  

But it'd be good to do it in a separate issue IMHO.

vmuzikar: Ok, agreed, let's clean it up as a follow-up.

@mabartos Do you want to create an issue for it, or will you address it as part of https://github.com/keycloak/keycloak/pull/46733 where you already touch the interface?

vmuzikar (dismissed):

vmuzikar:
Rebased

rest/admin-v2/api/src/main/java/org/keycloak/services/client/ClientService.java
mabartos: @vmuzikar I'll create a separate issue on the New service layer parent. 

Created follow-up issue: 
- https://github.com/keycloak/keycloak/issues/46806

vmuzikar:
@michalvavrik Can you please rebase?

michalvavrik:
> @michalvavrik Can you please rebase?

done; resolving merge conflicts brought in some changes in addition to what you have already reviewed

vmuzikar (approved):
Still LGTM.

PR #46284, opened 2026-02-12
Upgrade santuario to 3.0.6 to fix SANTUARIO-617

### LINKED ISSUE #45680: auth_mellon (SAML) authentication fails after upgrade to 26.5.1 (from 26.4.6)
### Before reporting an issue

### Area

saml

### Describe the bug

Hi,

Our web server uses auth_mellon to restrict access to certain resources. The following configuration is in use:

```
    <Location />
        MellonEnable info
        MellonEndpointPath /mellon/
        MellonSPMetadataFile /etc/httpd/mellon/urn_sso-test.esat.kuleuven.be.xml
        MellonSPPrivateKeyFile /etc/httpd/mellon/urn_sso-test.esat.kuleuven.be.key
        MellonSPCertFile /etc/httpd/mellon/urn_sso-test.esat.kuleuven.be.cert
        MellonIdPMetadataFile /etc/httpd/mellon/idp_metadata.xml
        MellonSecureCookie On
        MellonCookieSameSite none
        MellonUser uid
    </Location>
    <Location /restricted>
        AuthType Mellon
        MellonEnable auth
        require valid-user
    </Location>

```

The client is registered on our Keycloak server. Authentication has worked fine this way until we upgraded the system to Keycloak 26.5.1 (from 26.4.6). Since the upgrade, the authentication fails and the following message is logged in the apache error log:

```
[Tue Jan 20 16:43:18.459760 2026] [auth_mellon:error] [pid 686360:tid 686382] [remote 10.86.18.4:49801] Error processing authn response. Lasso error: [-427] When looking for an assertion we did not found it., SAML Response: StatusCode1="urn:oasis:names:tc:SAML:2.0:status:Success", StatusCode2="(null)", StatusMessage="(null)"
```

Downgrading the keycloak server to 26.4.6 makes the login work again. Note that the keycloak login itself seems to work, but auth_mellon is apparently no longer able to find some of the information it needs from the SAML response.

### Version

26.5.1

### Regression

### Expected behavior

The login using auth_mellon should work as it did with keycloak 26.4.6 (and the many versions we had before).

### Actual behavior

The login seems to work at the keycloak level, but auth_mellon seems unable to find the information it needs from the SAML response:

```
[Tue Jan 20 16:43:18.459760 2026] [auth_mellon:error] [pid 686360:tid 686382] [remote 10.86.18.4:49801] Error processing authn response. Lasso error: [-427] When looking for an assertion we did not found it., SAML Response: StatusCode1="urn:oasis:names:tc:SAML:2.0:status:Success", StatusCode2="(null)", StatusMessage="(null)"
```

### How to Reproduce?

* Install keycloak 26.4.6 and add a SAML client that uses auth_mellon
* Verify that login on the service protected by auth_mellon works
* Upgrade keycloak to 26.5.1
* Try to login on the resource gain

### Anything else?

_No response_

rmartinc:
Closes #45680

There is a bug in santuario 3.0.4 we are using. It seems that it affects to mellon when using rsa-oaep-mgf1. We need to upgrade to 3.0.5/4.0.3 or newer. I'm sending a draft and @pskopek can check what is the correct version we can upgrade to.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,7 +125,7 @@
         <woodstox.version>6.0.3</woodstox.version>
         <wildfly.common.quarkus.aligned.version>1.5.4.Final-format-001</wildfly.common.quarkus.aligned.version>
         <wildfly.common.wildfly.aligned.version>1.6.0.Final</wildfly.common.wildfly.aligned.version>
-        <xmlsec.version>3.0.4</xmlsec.version>
+        <xmlsec.version>3.0.6</xmlsec.version>
         <nashorn.version>15.4</nashorn.version>
         <ua-parser.version>1.6.1</ua-parser.version>
         <org.yaml.snakeyaml.version>2.0</org.yaml.snakeyaml.version>
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/saml/BasicSamlTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/saml/BasicSamlTest.java
@@ -478,12 +478,12 @@ public class BasicSamlTest extends AbstractSamlTest {
 
     @Test
     public void testEncryptionRsaOaep() throws Exception {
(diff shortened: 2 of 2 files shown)


pskopek (approved):
We can go on with 3.0.6.

mposolda (approved):
@rmartinc @pskopek Thanks for the PR and checking the santuario version possible to use!

PR #46286, opened 2026-02-12
fix: removing client crds from prod operator

### LINKED ISSUE #45794: Remove Client CRDs from operator bundle
### Description

We should not install the new client CRDs until we are satisfied with the state of the feature / usability - at the very least having the server side feature in preview vs experimental.

The controllers would need to check for, rather than assume, the CRDs are installed.

Cluster admins would need to something additional, such as install the CRDs from our repo.

shawkins:
closes: #45794

The strategy is to require a build property when creating a remote operator with which to test the client functionality.

Not sure if this needs mentioned in the README given the temporary nature.


vmuzikar (approved):
LGTM, thank you.

michalvavrik (approved):
After reading the linked issue, definitely +1 to only enabling for a tests ATM. IMO no point of documenting this in the README as it is basically just a temp workaround for us. I didn't exactly comprehend change in `operator/scripts/create-olm-bundle.sh`, but I'll look it up when I have more time (and you know what you are doing :-)).

michalvavrik:
I see, it is hardcoded in the `operator/target/bundle/keycloak-operator/manifests/keycloak-operator.clusterserviceversion.yaml`, ok, in that case it makes sense to drop it for now.

PR #46287, opened 2026-02-12
[Test Framework] Fix of AdminClientSupplier for ManagedRealm.

### LINKED ISSUE #46175: AdminClient in MANAGED_REALM mode has bugs
When `InjectAdminClient` is used with `Mode.MANAGED_REALM`, there are two issues:

1) `@InjectAdminClient` has a `realmRef` attribute, but it's not used by the supplier

2) A client and user have to be defined in a `RealmConfig` class, or they won't be found. Example below:

Works:
```java
@KeycloakIntegrationTest
public class SomeTest {

    @InjectRealm(config = ConfigClass.class) // if `myClient` is in realm config, all is ok
    ManagedRealm realm;

    @InjectAdminClient(mode = MANAGED_REALM, client = "myClient")
    Keycloak keycloak;
    
    @Test
    public void test() { }

    private static class ConfigClass implements RealmConfig {
        @Override
        public RealmConfigBuilder configure(RealmConfigBuilder realm) {
            realm.addClient("myClient").secret("mySecret");
            return realm;
        }
    }
}
```

Doesn't work:
```java
@KeycloakIntegrationTest
public class SomeTest {

    @InjectRealm
    ManagedRealm realm;

    @InjectClient(config = ConfigClass.class) // `myClient` is defined outside realm config, can't find it
    ManagedClient client;

    @InjectAdminClient(mode = MANAGED_REALM, client = "myClient")
    Keycloak keycloak;

    @Test
    public void test() { }

    private static class ConfigClass implements ClientConfig {
        @Override
        public ClientConfigBuilder configure(ClientConfigBuilder client) {
            return client.clientId("myClient").secret("mySecret");
        }
    }
}
```

lhanusov:
Closes #46175

### DIFF
--- a/test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientSupplier.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/admin/AdminClientSupplier.java
@@ -25,7 +25,7 @@ public class AdminClientSupplier implements Supplier<Keycloak, InjectAdminClient
     public List<Dependency> getDependencies(RequestedInstance<Keycloak, InjectAdminClient> instanceContext) {
         DependenciesBuilder builder = DependenciesBuilder.create(AdminClientFactory.class);
         if (instanceContext.getAnnotation().mode().equals(InjectAdminClient.Mode.MANAGED_REALM)) {
-            builder.add(ManagedRealm.class);
+            builder.add(ManagedRealm.class, instanceContext.getAnnotation().realmRef());
         }
         return builder.build();
     }
@@ -42,7 +42,7 @@ public class AdminClientSupplier implements Supplier<Keycloak, InjectAdminClient
         if (mode.equals(InjectAdminClient.Mode.BOOTSTRAP)) {
             adminBuilder.realm("master").clientId(Config.getAdminClientId()).clientSecret(Config.getAdminClientSecret());
         } else if (mode.equals(InjectAdminClient.Mode.MANAGED_REALM)) {
-            ManagedRealm managedRealm = instanceContext.getDependency(ManagedRealm.class);
+            ManagedRealm managedRealm = instanceContext.getDependency(ManagedRealm.class, instanceContext.getAnnotation().realmRef());
             adminBuilder.realm(managedRealm.getName());
 
             String clientId = !annotation.client().isEmpty() ? annotation.client() : null;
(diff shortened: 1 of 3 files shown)


stianst (changes requested):
The example in the issue is not valid:

```
    @InjectRealm
    ManagedRealm realm;

    @InjectClient(config = ConfigClass.class) // `myClient` is defined outside realm config, can't find it
    ManagedClient client;

    @InjectAdminClient(mode = MANAGED_REALM, client = "myClient")
    Keycloak keycloak;
```

`@InjectAdminClient` states use the managed realm, and the `myClient` within it. However, since that is not defined there it's not expected to work.

I'd say the actual bug reported in https://github.com/keycloak/keycloak/issues/46175 is not valid.

Maybe we would want to support `clientRef` with `@InjectAdminClient`, but that would be a new feature, not a bug fix. 

The updates in this PR for clients is not valid, as it's basically just a coincidence that the client works when looping through all clients in the realm (expensive and unnecessary). If you move the `@InjectClient` below `@InjectAdminClient` it won't work.

The other part of adding support for userRef again maybe is a nice to have, but not relevant to fixing the client part.

I'd say the 46175 is not valid, then we can consider adding support as clientRef and userRef as follow-ups.

vaceksimon:
I can't find it in the docs or in the comments. In that case, we should mention somewhere that the `user` and `client` attributes of `@InjectAdminClient` work only for users and clients defined in the realm config class specified via the `config` attribute in the `@InjectRealm` annotation.

Anyway, if a `client` is specified in the admin client but the realm doesn't contain any, the supplier throws:
```java
java.lang.NullPointerException: Cannot invoke "java.util.List.stream()" because the return value of "org.keycloak.representations.idm.RealmRepresentation.getClients()" is null
```

This behavior could be improved.

vaceksimon:
> 1. @InjectAdminClient has a realmRef attribute, but it's not used by the supplier

But I believe this is still a valid bug. It should either be used or removed.

stianst:
> > 1. @InjectAdminClient has a realmRef attribute, but it's not used by the supplier
> 
> But I believe this is still a valid bug. It should either be used or removed.

Missed the realmRef part, yeah that seems valid.

So, my suggested path forward for this one then:

1. Fix the realmRef
2. Revert the changes around client, and adding userRef
3. Update JavaDocs to clearly state client/user has to be within the managed realm (https://github.com/keycloak/keycloak/pull/46404)

I don't think we need `clientRef` and `userRef` for this; that's quite a special case when it's needed and better to just use the factory instead for those special cases.

stianst (approved):

PR #46294, opened 2026-02-12
Allow to expose WellKnown provider via ServerMetadataResource (#46292)

thomasdarimont:
- Introduce isAvailableViaServerMetadata() on WellKnownProviderFactory to let each provider control whether it is exposed via the root /.well-known/{alias}/realms/{realm} endpoint,  defaulting to false to preserve existing behavior
- Extract shared alias-resolution logic into a new WellKnownProviderUtil.resolveFromAlias() utility, replacing duplicated lookup code in both RealmsResource and ServerMetadataResource
- Opt in OAuth2WellKnownProviderFactory, OID4VCIssuerWellKnownProviderFactory, and JWTVCIssuerWellKnownProviderFactory by overriding isAvailableViaServerMetadata() to return true
- Align ServerMetadataResource method names by replacing `getOAuth2AuthorizationServerWellKnownVersionPreflight` with `getWellKnownPreflight` and `getOAuth2AuthorizationServerWellKnown` with `getWellKnown`.
- Add test to verify `.well-known/openid-configuration` is not exposed via server metadata root by default.

Previously, ServerMetadataResource.isValidProvider() used a hardcoded allowlist of provider IDs to decide which well-known endpoints were exposed at the server root. This meant every new provider required a code change in ServerMetadataResource. With this change, each WellKnownProviderFactory can declare its own eligibility, making the mechanism extensible for custom and future providers without modifying the resource class.

Fixes #46292


services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
```diff
@@ -51,30 +47,30 @@ public class ServerMetadataResource {
...
     }
 
     @GET
     @Path("{provider}/realms/{realm}")
-    @Produces({MediaType.APPLICATION_JSON, APPLICATION_JWT})
-    public Response getOAuth2AuthorizationServerWellKnown(final @PathParam("provider") String providerName,
-                                                          final @PathParam("realm") String name) {
-        if (!isValidProvider(providerName)) throw new NotFoundException();
-        return RealmsResource.getWellKnownResponse(session, name, providerName, logger);
+    @Produces({MediaType.APPLICATION_JSON, org.keycloak.utils.MediaType.APPLICATION_JWT})
```
IngridPuppet, line 60: Do you mind dropping the fully qualified prefix? It gives away AI.
```suggestion
    @Produces({MediaType.APPLICATION_JSON, MediaType.APPLICATION_JWT})
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oidc/OIDCWellKnownProviderTest.java
```diff
@@ -28,4 +41,21 @@ protected String getWellKnownProviderId() {
         return OIDCWellKnownProviderFactory.PROVIDER_ID;
     }
 
+    /**
+     * Ensures that the OpenID Connect configuration is not exposed by default via the server metadata
+     * root endpoint. This test verifies that accessing the `.well-known/openid-configuration`
```
IngridPuppet, line 46: I suggest to additionally check that one of the implementations with a defaut to true results in a 200, to assert that the mechanism is functional.

IngridPuppet (commented):
Hello @thomasdarimont - Thank you for the PR. I reviewed the updates and left two minor comments. Please could you check?

services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
thomasdarimont: I deliberately added the fully qualified prefix there to show that this is a Keycloak custom type and not a built-in Media Type. Do you see a problem with this?

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oidc/OIDCWellKnownProviderTest.java
thomasdarimont: The positive case for the OIDCWellKnown response is already tested by the base-class.

services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
IngridPuppet: Oh there is a clash... It's fine for me, thank you for the clarification.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oidc/OIDCWellKnownProviderTest.java
IngridPuppet: That addresses my concern.

IngridPuppet (approved):
LGTM 👍

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oidc/OIDCWellKnownProviderTest.java
tnorimat, line 1: It is better to write test codes by using the new test framework (junit5).
I heard that the new PR need to do that.

services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
```diff
@@ -51,30 +47,30 @@ public class ServerMetadataResource {
...
         return builder.path(ServerMetadataResource.class).path("{provider}/realms/{realm}");
     }
 
-    private boolean isValidProvider(String providerName) {
-        // you can add codes here considering the current status of the implementation (preview, experimental).
-        if (OAuth2WellKnownProviderFactory.PROVIDER_ID.equals(providerName)) return true;
-        if (OID4VCIssuerWellKnownProviderFactory.PROVIDER_ID.equals(providerName)) return true;
-        if (JWTVCIssuerWellKnownProviderFactory.PROVIDER_ID.equals(providerName)) return true;
-        return false;
+    private boolean isValidProvider(String alias) {
```
tnorimat, line 73: The method can be written as follows with `Optional`:
```
private boolean isValidProvider(String alias) {
    return WellKnownProviderUtil.resolveFromAlias(session.getKeycloakSessionFactory(), alias)
            .map(WellKnownProviderFactory::isAvailableViaServerMetadata)
            .orElse(false);
}
```

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oidc/OIDCWellKnownProviderTest.java
thomasdarimont: Wasn't there the exception that existing test cases may use the old test framework?
In my case I extended an existing test with another test method. Rewriting the test in JUnit5 with the new Keycloak testframework for this single test method would require a lot of effort.

services/src/main/java/org/keycloak/services/resources/ServerMetadataResource.java
thomasdarimont: I like your suggestion! I'll change the code to use it.

thomasdarimont:
It seems github has an issue - I pushed my changes to my branch, but they are not reflected here?!
https://github.com/keycloak/keycloak/compare/main...thomasdarimont:keycloak:issue/gh-46292-wellknown-root-config

thomasdarimont:
Something is definitly wrong here... this show 42 commits now - but mine shows only 5... 
https://github.com/keycloak/keycloak/compare/main...thomasdarimont:keycloak:issue/gh-46292-wellknown-root-config

thomasdarimont:
I need to close this PR and open the same again... sorry folks.

thomasdarimont:
I resent the PR - this time the commits look correct.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/oidc/OIDCWellKnownProviderTest.java
tnorimat: I understand the point.

PR #46301, opened 2026-02-12
Add tests that run with distribution placed in a directory with special characters

### LINKED ISSUE #46135: Run tests with dist placed in a directory with special characters
### Description

We should run at least some smoke test with KC placed in a directory that includes special character (or at least a space). This is a follow-up from https://github.com/keycloak/keycloak/pull/45972.

michalvavrik:
Adds tests for https://github.com/keycloak/keycloak/issues/45971. I have run tests added in this PR after reverting the fix https://github.com/keycloak/keycloak/pull/45972 and they verify the issue.

* Closes: https://github.com/keycloak/keycloak/issues/46135

### DIFF
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AbstractPathDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AbstractPathDistTest.java
@@ -1,0 +1,98 @@
+package org.keycloak.it.cli.dist;
+
+import java.io.File;
+import java.util.Map;
+import java.util.function.Consumer;
+
+import org.keycloak.it.TestProvider;
+import org.keycloak.it.junit5.extension.BeforeStartDistribution;
+import org.keycloak.it.junit5.extension.CLIResult;
+import org.keycloak.it.junit5.extension.RawDistOnly;
+import org.keycloak.it.utils.KeycloakDistribution;
+import org.keycloak.it.utils.RawKeycloakDistribution;
+import org.keycloak.quarkus.runtime.cli.command.StartDev;
+
+import io.quarkus.test.junit.main.Launch;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Runs test with distribution placed in a directory that has special characters in name.
+ *
+ * @see <a href="https://github.com/keycloak/keycloak/issues/45971">Issue #45971</a> for motivation of this test
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+abstract class AbstractPathDistTest {
+
+    /**
+     * Switches temporary directory used by the {@link RawKeycloakDistribution}.
+     */
+    private static final String JAVA_IO_TMPDIR = "java.io.tmpdir";
+    private String originalTmpDir = null;
+
+    protected abstract String getSubPath();
+
+    @BeforeAll
+    void changeTmpDir() {
+        originalTmpDir = System.getProperty(JAVA_IO_TMPDIR);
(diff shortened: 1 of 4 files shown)


michalvavrik:
Looking at the `integration-arquillian-tests-base` failures, that is a different module and this PR only adds a new tests to a different module, so it shouldn't be related.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AbstractPathDistTest.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright 2026 Red Hat, Inc. and/or its affiliates
```
mabartos: @michalvavrik JFYI - including licenses is not required for Keycloak anymore.

mabartos (dismissed):
@michalvavrik LGTM, thanks!

@shawkins You want to review if it covers all scenarios as for the fix? 

@michalvavrik The CI failures are unrelated, and the fix for them is already in the main branch, so you can just rebase.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AbstractPathDistTest.java
michalvavrik: ah, cool, so I'll remove it and rebase to get green CI. Thanks for the information

michalvavrik: done

mabartos (dismissed):

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AbstractPathDistTest.java
```diff
@@ -0,0 +1,111 @@
...
+            System.setProperty(JAVA_IO_TMPDIR, originalTmpDir);
+            originalTmpDir = null;
+        }
+    }
+
+    @Order(1)
+    @BeforeStartDistribution(AddCustomScriptsProvider.class)
+    @RawDistOnly(reason = "Testing installation path handling")
+    @Test
+    @Launch(Build.NAME)
```
vmuzikar: Just a nitpick. I think just testing `start` or `start-dev` would be enough as build is done implicitly. But we can keep as is.

vmuzikar (dismissed):
@michalvavrik Looks good, thank you. Alternative approach would've been running Quarkus tests in the CI from a different subdir with special characters but this works too.

quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/AbstractPathDistTest.java
mabartos: @vmuzikar Yep, nice point.

michalvavrik: I tried it before creating this PR and it did not run the build again. I presume I have done something wrong as you are sure about it, so I'll retry. To my limited understanding, the build is only done if the augmentation wasn't done before. I will try it and change it accordingly.

vmuzikar: > if the augmentation wasn't done before

Right but you unpack the dist to a new location so it should be unaugmented?

michalvavrik: > > if the augmentation wasn't done before
> 
> Right but you unpack the dist to a new location so it should be unaugmented?

I probably made some mistake when trying it, on it now.

michalvavrik: You were right, the problem was that I was testing `org.keycloak.it.junit5.extension.CLIResult#assertBuild` and it doesn't work for the start in dev mode, which confused me.

michalvavrik: Done.

vmuzikar (approved):
LGTM.

mabartos (approved):

PR #46307, opened 2026-02-13
Update name field validation logic in `NewClientPolicy` component

### LINKED ISSUE #45948: When saving a client policy, the conditions and the client profiles disappear
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When I save a client policy, the conditions and client profiles are empty after save. 

Other observations: 
* The Save-button does not become active when I add a client profile
* The Save-button does not become inactive when I save the client profile
* When I change the name, and then change the name back to the original name, I get the error message "The name must be unique within the realm"

Before save: 

Image

After save: 

Image

### Version

main

### Regression

### Expected behavior

* When I save a client policy, the conditions and client profiles should be visible 
* The Save-button does not become active when I add a client profile
* The Save-button does not become inactive when I save the client profile
* When I change the name, and then change the name back to the original name, I get the error message "The name must be unique within the realm"
* 

### Actual behavior

None of the above.

### How to Reproduce?

* Create client policy
* Add a condition and a policy
* Press Save

### Anything else?

_No response_

calebd-anderson:
The `NewClientPolicy` component is reused for creating a client policy and editing an existing client policy. If the form is not valid the save button is disabled. This PR updates the name field form validation. Given the `policyName` constant is only defined when editing an existing client policy, short-circuit the validation logic to only verify the new policy name has not been taken if `policyName` does not match the new name.

With this form validation logic, the expected behavior of the form is enabled:
### Fixes:
- [x] The Save-button does not become active when I add a client profile
- [x] The Save-button does not become inactive when I save the client profile
- [x] When I change the name, and then change the name back to the original name, I get the error message "The name must be unique within the realm"
- [x] ~~When I save a client policy, the conditions and client profiles should be visible~~
  - this may have been fixed somewhere else, I was not able to reproduce invisible conditions or policies after rebase on main

Closes #45948

### DIFF
--- a/js/apps/admin-ui/src/realm-settings/NewClientPolicy.tsx
+++ b/js/apps/admin-ui/src/realm-settings/NewClientPolicy.tsx
@@ -478,7 +478,9 @@ export default function NewClientPolicy() {
               rules={{
                 required: t("required"),
                 validate: (value) =>
-                  policies.some((policy) => policy.name === value)
+                  policies.some(
+                    (policy) => policyName !== value && policy.name === value,
+                  )
                     ? t("createClientProfileNameHelperText")
                     : true,
               }}


msdaly200 (changes requested):
When testing was still able to reproduce issue where the client profiles are empty after save, tends to happen when you add the client profile before adding a condition.

Could still see save button enabled, after a save with no changes made.
Could still see the error message "The name must be unique within the realm".  Only change now is that the Save button is enabled in this scenario.   
When you change the name, the selected condition and client profile disappear.  

@mposolda  - Would recommend splitting this ticket up into a few UI tickets to fix above issues.

msdaly200:
Hi @mposolda and @calebd-anderson  - I've created some extra issues to try and break down the tasks outlined in https://github.com/keycloak/keycloak/issues/45948 - When saving a client policy, the conditions and the client profiles disappear.  We will close this PR so that https://github.com/keycloak/keycloak/issues/45948 can deal with the main issue of the conditions and client profiles being empty after a save.  Will update https://github.com/keycloak/keycloak/issues/45948 shortly to request reviews.

PR #46311, opened 2026-02-13
[OID4VCI] Attribute did should only be added to user profile if oid4vci is enabled for the realm

### LINKED ISSUE #46297: [OID4VCI] Attribute did should be added to user profile just if OID4VCI is enabled for the realm
### Before reporting an issue

### Area

oid4vc

### Describe the bug

Right now, when Keycloak is started with `oid4vc-vci` feature enabled, the attribute `did` is automatically added to user profile.

It should be added just in case that switch `Verifiable credentials` is also enabled for the realm (together with the `oid4vc-vci` feature).

User profile feature has nice support for this to dynamically add specified attribute based on something enabled/disabled for the realm. For example attribute `Locale` is also dynamically added to user's profiles just when the "Localization" is enabled for the realm.

Moreover, the attribute is not translated in the admin console - see screenshot (but I see it is properly translated in the account console):

Image

### Version

nightly (from 2026-02-12)

### Regression

### Expected behavior

Attribute `did` added to user profile just if feature `oid4vc-vci` is enabled and also realm switch `Verifiable credentials` is enabled

### Actual behavior

Attribute `did` added to user profile always when just `oid4vc-vci` is enabled. Attribute not translated in the admin console.

### How to Reproduce?

- Start keycloak with `./kc.sh start-dev --features=oid4vc-vci`
- Login to admin console and go to the user profile of user `admin` (Assumption that `Verified credentials` is disabled for the realm)

### Anything else?

It is regression caused by https://github.com/keycloak/keycloak/commit/c08ed20f

tdiesler:
closes #46297

### DIFF
--- a/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
+++ b/js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties
@@ -1796,6 +1796,7 @@ exportAuthDetailsError=Error exporting authorization details\: {{error}}
 clientOfflineSessionIdleHelp=Time a client offline session is allowed to be idle before it expires. Offline tokens are invalidated when a client offline session is expired. The option does not affect the global user SSO session. If not set, it uses the realm Offline Session Idle value.
 selectGroups=Select groups to join
 webAuthnPolicyAuthenticatorAttachmentHelp=Communicates to an authenticator an acceptable attachment pattern.
+did=Identity (DID)
 username=Username
 mappedGroupAttributes=Mapped group attributes
 localization=Localization
--- a/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProvider.java
+++ b/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProvider.java
@@ -75,7 +75,6 @@ public class DeclarativeUserProfileProvider implements UserProfileProvider {
      *
      * @param context to get current auth flow from
      * @param configuredScopes to be evaluated
-     * @return
      */
     private static boolean requestedScopePredicate(AttributeContext context, Set<String> configuredScopes) {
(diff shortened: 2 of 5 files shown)


mposolda (commented):
It seems this is not really fixed.

The `did` attribute is not really added dynamically to the user-profile. For example when I create new realm `test` in the admin console, then I enable `Verifiable credentials` for that realm and I create some user in the realm, I do not see `did` attribute in the user's profile.

As I've pointed already, I think that you can use same pattern like it is used for example for the `locale` attribute, which is added trully dynamically to the user profile based on whether localization is enabled for the realm. See https://github.com/keycloak/keycloak/blob/release/26.5/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java#L427 , https://github.com/keycloak/keycloak/blob/release/26.5/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java#L490 , https://github.com/keycloak/keycloak/blob/release/26.5/services/src/main/java/org/keycloak/userprofile/DeclarativeUserProfileProviderFactory.java#L507 . I think same pattern should be possible to use for `did` attribute, which would be added when `verifiable credentials` is enabled for the realm though?


The attribute `did` is localized only in account console, but still not localized when looking at the admin console and when updating profile of the user in the "login" theme (Triggered by adding "Update profile" required action to the user). It may probably need update in `js/apps/admin-ui/maven-resources/theme/keycloak.v2/admin/messages/messages_en.properties` (for admin theme) and `themes/src/main/resources/theme/base/login/messages/messages_en.properties` (for login theme). There are separate files for "login" theme, "account" theme and "Admin" theme and some things (like user-profile attributes) need to be updated in all those places.

tdiesler:
pls try again

mposolda (approved):
@tdiesler Thanks!

PR #46316, opened 2026-02-13
Enhance workflows failure handling

sguilhen:
- run each step in a separate transaction, adding consistency for logs and events
- ensure executor timeout is checked by the run workflow task
- record execution of every step, scheduled or not, to support recovery from failures

Closes #45175


sguilhen:
@pedroigor I would like to ask you to not squash the commits when merging. It will be easier in the future to reference the changes in isolation if we need to revisit this code.

pedroigor (dismissed):

pedroigor (approved):

PR #46327, opened 2026-02-13
[OID4VCI] User self-issued credential: Issuance initiated from web po…

### LINKED ISSUE #46196: [OID4VCI] User self-issued credential: Issuance initiated from web portal application

mposolda:
…rtal application and credential-offer displayed on Keycloak

closes #46196

PR for the flow described in the #46196 . Similar to what I've presented on OAuth SIG meeting on Feb 4 with the different that it is user self-initiated flow.

The initial OIDC authentication request sent from some "portal application" contains `kc_action` parameter (KC parameter used for application-initiated actions ). The `kc_action` parameter has format like `verifiable_credential_offer:eyJjcmVkZW50aWFsX2NvbmZpZ3VyYXRpb25faWQiOiJzZC1qd3QtY3JlZGVudGlhbC1jb25maWctaWQiLCJjbGllbnRfaWQiOiJ0ZXN0LWFwcCIsInByZV9hdXRob3JpemVkIjp0cnVlfQ` . The first part `verifiable_credential_offer` denotes the name of AIA. The parameter after `:` is the Base64-encoded value of the details, which should credential-offer has. In this case, it is something like `{"credential_configuration_id":"sd-jwt-credential-config-id","client_id":"test-app","pre_authorized":true}` 

So far, parameters like `credential_configuration_id` , `pre_authorized` and `client_id` are supported. Those are same names like credential-offer creation endpoint is using ( `OID4VCIssuerEndpoint.getCredentialOfferURI` )

As a follow-up, this might be further improved, so that "dataholder / portal" application does not need to use Base64 with all the parameters, but can somehow reference the "configuration template", which would point to some pre-defined configuration on KC side. Also we might have built-in support for `scope` parameter, which could do this (Similar to what @Captain-P-Goldfish showed the demo on last OAuth SIG meeting), but that would be also a follow-up (not 100% sure if it could be supported OOTB or with custom authenticator, which would emulate adding `kc_action` with appropriate configuration to current request based on the value of `scope` parameter).


tdiesler:
The `pre-authorized_code` grant offer should probably be an exception and not the rule. Self issuance of an `authorization_code` grant offer might make sense - of a `pre-authorized_code` grant perhaps less so. If the goal is that the University Portal gives the student "something" that makes her to login to get a Credential - it would not have to be a `pre-authorized_code` grant. On the contrary, because the student is expected to login anyway, an `authorization_code` grant would suffice.

Is there a use case that requires a `pre-authorized_code` grant, that would not work with an `authorization_code` grant? It seems to be a funny concept to require a login to give out something that grants access without login.

On a more general note, I wonder about the role/value of GUI functionality that is not backed by REST API. I was under the impression that Keycloak follows a REST API centric architecture. If it is not on the REST API, it can't work on the CLI either. It will also not work in deployments where GUI access is turned off.

AFAIK, Keycloak is designed to work as a headless AS + IP. Unless I'm mistaken, all OAuth/OIDC message exchanges work against the REST API. Any http library or `curl` on the command line will do. I would think, it should be the same for the OID4VCI Issuer. Diverging from this could be quite consequential, as if this great new K8S feature only works when you go through Lens.

Perhaps we need to clarify this GUI vs. REST API approach - it would effect how we do things.

I'm generally not against Application Initiated Actions ([AIA](https://www.keycloak.org/docs/latest/server_admin/index.html#con-aia_server_administration_guide)) as long as they do not provide the only possible path to a given thing or prevent the REST API to access that thing.

Anyway, I believe this PR should build on top of `authorization_code` grant and not be limited to `pre-authorized_code` grant.

> PS: I know that Authorization Code Flow requires browser access. Would this fall into the same category? Would it have to have the same customization options as the login screen?

mposolda:
> The `pre-authorized_code` grant offer should probably be an exception and not the rule. Self issuance of an `authorization_code` grant offer might make sense - of a `pre-authorized_code` grant perhaps less so. If the goal is that the University Portal gives the student "something" that makes her to login to get a Credential - it would not have to be a `pre-authorized_code` grant. On the contrary, because the student is expected to login anyway, an `authorization_code` grant would suffice.
> 
> Is there a use case that requires a `pre-authorized_code` grant, that would not work with an `authorization_code` grant? It seems to be a funny concept to require a login to give out something that grants access without login.

Yes, the use-case is for example "User initiates Issuance from Keycloak Portal" from the use-cases created by @Captain-P-Goldfish here: https://github.com/keycloak/keycloak/discussions/45626 .

You have web portal, but at the same time, you have wallet, which is possibly on different device (typically mobile phone). You can use pre-authorized grant with your wallet, but it first needs to be ensured that user properly authenticated somewhere else (which might be authentication on different device). And when the page is displayed by the Keycloak, it is guaranteed that user authenticated.

In other words, flow like this:
- User is authenticated to "university portal", which is Keycloak secured web application supporting OIDC "authorization_code" flow
- On the portal, user clicks button "Claim your education certificate"
- This redirects user to Keycloak (with AIA) and display him the page with QR-code (credential offer). He can obtain his education-certificate to his wallet.

Note that university in this case doesn't have access to pre-authorized grant credential-offer, but just Keycloak and user himself have access to it. Which is actually huge advantage.


> On a more general note, I wonder about the role/value of GUI functionality that is not backed by REST API. I was under the impression that Keycloak follows a REST API centric architecture. If it is not on the REST API, it can't work on the CLI either. It will also not work in deployments where GUI access is turned off.
> 
> AFAIK, Keycloak is designed to work as a headless AS + IP. Unless I'm mistaken, all OAuth/OIDC message exchanges work against the REST API. Any http library or `curl` on the command line will do. I would think, it should be the same for the OID4VCI Issuer. Diverging from this could be quite consequential, as if this great new K8S feature only works when you go through Lens.

Keycloak is especially browser SSO. Lots of the things supported by Keycloak are not backed by REST API. For example, functionality like `Update OTP credential of the user john` . There is not admin REST endpoint or account REST endpoint for updating user's OTP credential directly. Instead, user is supposed to start either the AIA flow from the account console, which will redirect him to Keycloak and he needs to set his OTP credential in the Keycloak UI.  That provides good user experience and additional security (as Keycloak can for example re-authenticate user before allowing to update his OTP). Alternatively admin can add "required action" of "Update OTP" to user account in the admin console. That is a way for admin to require user to set his OTP during next login, but at the same time, OTP secret is not shared with admin himself.

Note that application can request to authenticate, but does NOT have access to user's login credentials (password, OTP).

In general: REST endpoints are in general useful when some other party (client application) needs to access some data/functionality of Keycloak. At the same time, in case that some data/functionality should be rather shared just to the user himself and not with client application, the REST endpoints might not be desired.

> 
> Perhaps we need to clarify this GUI vs. REST API approach - it would effect how we do things.
> 
> I'm generally not against Application Initiated Actions ([AIA](https://www.keycloak.org/docs/latest/server_admin/index.html#con-aia_server_administration_guide)) as long as they do not provide the only possible path to a given thing or prevent the REST API to access that thing.

+1 to clarify GUI vs REST API approach. It seems we will need to document this once we add the proper documentation for OID4VCI to document when to use both approaches.

I see the GUI as better alternative when possible as it does not require to share the pre-authorized code credential-offer with any other party (issuance portal client application, admin user) besides issuer(Keycloak) and user himself. Also it makes implicit consent of the user by making sure that Keycloak / issuer displayed the screen to the user: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-user-consent .

Also the AIA based approach is in fact "user self-issuance" when REST API is also for the cases when someone else (administrator) created credential-offer in advance for the user. For administrator issuing credentials to the user, we may need to also support the use-case of credential-offer being sent by the email    in addition to AIA and REST endpoint. Related issue: https://github.com/keycloak/keycloak/issues/46198

But I see credential-offer created by REST endpoint has also some use-cases, which GUI based approach might not always cover.

> 
> Anyway, I believe this PR should build on top of `authorization_code` grant and not be limited to `pre-authorized_code` grant.

I don't consider this to be limited to `pre-authorized grant` . You can notice `CredentialOfferActionConfig` has config parameter "preAuthorized" . It is considered to be true only for pre-authorized grant and otherwise should do authorization-code grant. Right now, it tests just the "pre-authorized code" , but I think it should support also "authorization code" similarly like REST endpoint. I see there is some intersection with your work related to `issuer_state`, so I consider adding/testing that as a follow-up once also your work for "issuer state" is included.

> 
> > PS: I know that Authorization Code Flow requires browser access. Would this fall into the same category? Would it have to have the same customization options as the login screen?

Yes, it has the same customization options as the login screen. Keycloak has the concept of "themes" https://www.keycloak.org/docs/latest/server_admin/index.html#_themes . And this screen is part of the "Login theme" . This means effectively same CSS etc are used as for normal Keycloak login screens. Administrators are able to deploy their own themes and hence customize the look and feel of the screens.

tdiesler:
> You have web portal, but at the same time, you have wallet, which is possibly on different device (typically mobile phone)

The `authorization_code` grant is not limited to same device flow. In practice, using an `authorization_code` grant in a cross device flow may mean that the wallet needs to login again, because it needs to create its own login session. Is the reason to use a `pre-authorized_code` grant to avoid that second wallet login?

The `pre-authorized_code` grant is especially designed for wallets that cannot authorize themselves. Is the use of that AIA required? If so, it might close the door on those intended use cases.

mposolda:
> > You have web portal, but at the same time, you have wallet, which is possibly on different device (typically mobile phone)
> 
> The `authorization_code` grant is not limited to same device flow. In practice, using an `authorization_code` grant in a cross device flow may mean that the wallet needs to login again, because it needs to create its own login session. 
> 

Yes, I agree. I don't consider that `authorization_code` grant is limited to same device flow. But yes, for cross-device authorization_code , I think that authentication should happened again on other device. Pretty much, I can imagine flow like this (for authorization_code flow and cross-device):

- User is authenticated on university portal on his laptop
- User clicks `Issue education certificate` 
- Keycloak page with credential offer is displayed (QR code + deep link) . This credential offer is for authorization_code and has `issuer_state`
- User scans the QR code to his mobile. OIDC authentication request is sent on the mobile with `issuer_state` parameter. User needs to login again on this mobile device as he doesn't have SSO session yet.

> Is the reason to use a `pre-authorized_code` grant to avoid that second wallet login?

Yes, with pre-authorized_code grant, you don't need second login on the wallet. Also as you mentioned, the pre-authorized code grant is useful for the cases when the wallet on other device is not able to authenticate itself.

> The `pre-authorized_code` grant is especially designed for wallets that cannot authorize themselves. Is the use of that AIA required? If so, it might close the door on those intended use cases.

I don't consider AIA required as in some cases, I suppose that REST endpoint would be needed. But I see AIA useful even in some cases when the wallet cannot authorize itself. For example cases when:

- Cross-device flow: Device 1 is laptop with proper browser. Device 2 is a wallet without capability to authenticate/authorize user itself. On device1, the user authentication happens and page with QR code would be displayed. Device 2 will scan the QR code from device1 and can do pre-authorized flow

- Same device flow: For example on mobile device, there is the wallet application, which is not able to do proper login. However there is browser application on this mobile device. User can authenticate on the browser and QR code with deep-link page is displayed at the end of authentication. User clicks on the link (some link like `openid-credential-offer://...` ) and that link would be opened in the wallet application, which can continue with pre-authorized flow.

But I suppose that there are likely use-cases where none of those may not fit ideally and hence REST endpoint approach will need to be used. Is it correct?

services/src/main/java/org/keycloak/protocol/oid4vc/utils/OID4VCUtil.java
```diff
@@ -19,6 +31,48 @@ public class OID4VCUtil {
     private OID4VCUtil() {
     }
 
+    /**
+     * @param session Keycloak session
+     * @param nonce nonce, which is part of the credential offer URI
+     * @return Credential offer as URI, which can be shared with the wallet
+     */
+    public static String getOfferAsUri(KeycloakSession session, String nonce) {
+        String encodedOfferUri = URLEncoder.encode(OID4VCIssuerWellKnownProvider.getIssuer(session.getContext()) + "/protocol/" + OID4VCLoginProtocolFactory.PROTOCOL_ID + "/" + CREDENTIAL_OFFER_PATH + "/" + nonce, StandardCharsets.UTF_8);
```
IngridPuppet: Use KeycloakUrlBuilder?

services/src/main/java/org/keycloak/protocol/oid4vc/utils/OID4VCUtil.java
```diff
@@ -19,6 +31,48 @@ public class OID4VCUtil {
...
+                        return displayData.getName();
+                    }
+                }
+            } catch (IOException ioe) {
+                logger.warnf("Incorrect vc.display for client scope '%s'", clientScope.getName());
+            }
+        }
+
+        // Fallback
+        display = clientScope.getAttribute(CONFIGURATION_ID);
```
IngridPuppet: I suggest to always access these attributes via the CredentialScopeModel. For many of the available attributes, the wrapper provides defaults that we should use consistently.

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/pages/OID4VCCredentialOfferPage.java
```diff
@@ -0,0 +1,68 @@
...
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.FindBy;
+
+public class OID4VCCredentialOfferPage extends LanguageComboboxAwarePage {
+
+    @FindBy(id = "credential-offer-uri-link")
+    private WebElement credentialOfferUri;
+
+    @FindBy(id = "continueVCOffer")
+    private WebElement continueButton;
```
IngridPuppet: For consistency?
```suggestion
    @FindBy(id = "continue-vc-offer")
    private WebElement continueButton;
```

themes/src/main/resources/theme/base/login/messages/messages_en.properties
```diff
@@ -519,6 +519,12 @@ passkey-doAuthenticate=Sign in with Passkey
 passkey-createdAt-label=Created
 passkey-autofill-select=Select your passkey
 
+# OID4VCI
+credentialOfferTitle=Claim your {0}
+credentialOfferStep1=Claim your {0} by scanning the following QR code to your digital wallet in case that wallet is on different device. Alternatively open the Offer URL directly if your digital wallet is available on this device.
```
IngridPuppet:
```suggestion
credentialOfferStep1=Claim your {0} by scanning the following QR code with your digital wallet in case it is on a different device. Alternatively, open the Offer URL directly if your digital wallet is available on this device.
```

themes/src/main/resources/theme/keycloak.v2/login/oid4vc-credential-offer.ftl
```diff
@@ -0,0 +1,43 @@
...
+                <div class="pf-v5-c-form__actions">
+                    <#if isAppInitiatedAction??>
+                        <input type="submit"
+                            class="${properties.kcButtonClass!} ${properties.kcButtonPrimaryClass!} ${properties.kcButtonLargeClass!}"
+                            id="continueVCOffer" value="${msg("doContinue")}"
+                        />
+                        <button type="submit"
+                                class="${properties.kcButtonClass!} ${properties.kcButtonDefaultClass!} ${properties.kcButtonLargeClass!} ${properties.kcButtonLargeClass!}"
+                                id="cancelVCOffer" name="cancel-aia" value="true">${msg("doCancel")}
+                        </button>
```
IngridPuppet, line 32: From a user perspective, these buttons behave the same (they continue the login process) and I personally found it confusing after clicking on "Cancel". From your experience, is the UX comparable with other screens? I wonder if there is a better solution, but not sure.
Image

IngridPuppet (commented):
Hello @mposolda - Thank you for the PR. I'm excited about this flow getting merged. I hope the ongoing discussion with @tdiesler  on the topic will soon reach a consensus.

I tested the flow locally and confirm all works as expected.

Image

I reviewed the changes and left some minor comments. Please could you check?

tdiesler:
IMHO, we need to have support for `authorization_code` grant in order to test this in a meaningful way.

themes/src/main/resources/theme/keycloak.v2/login/oid4vc-credential-offer.ftl
mposolda: Yes, it is comparable with other screens. Application actually has a way to figure if user cancelled the action as there is parameter `kc_action_status=cancelled` added to the redirect URI once user is redirected back to the application. This is documented in https://www.keycloak.org/docs/latest/server_admin/index.html#con-aia_server_administration_guide . So there is a way how to handle on the application side whether the "continue" or "cancel" button was pressed and display some message to the user according to which button he pressed.

However it is the question how much is usability gain as user usually knows whether he "submitted" or "cancelled" the action. For example, the built-in Keycloak account console currently doesn't have support for the `kc_action_status` parameter. For example when you "Update password" from the Keycloak account console and then you click "Cancel" on the page where user is updating his password, you are redirected back to the account console and UX looks the same (regardless of whether user updated his password by click "Submit" or whether he click "Cancel").

mposolda:
> IMHO, we need to have support for `authorization_code` grant in order to test this in a meaningful way.

@tdiesler Yes, I agree it should work with `authorization_code` grant. It should actually work with both `authorization_code` and `pre-authorized code` grants. I am planning to update the PR once your "part 5" PR https://github.com/keycloak/keycloak/pull/46951 is addressed.

@IngridPuppet Thanks for the review and feedback! Will update the PR soon based on your comments. Will also need to rebase with Keycloak main (once "part 5 PR mentioned above is done), add support for `authorization_code` grant and migrate the `OID4VCACtionTokenTest` to the new testsuite.

services/src/main/java/org/keycloak/protocol/oid4vc/utils/OID4VCUtil.java
mposolda: Updated, Thanks

services/src/main/java/org/keycloak/protocol/oid4vc/utils/OID4VCUtil.java
mposolda: Updated, Thanks

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/pages/OID4VCCredentialOfferPage.java
mposolda: Updated, Thanks

themes/src/main/resources/theme/base/login/messages/messages_en.properties
mposolda: Updated, Thanks

mposolda:
@IngridPuppet Do you please have an opportunity to re-review? I've tried to incorporate and answer your comments. 

Updated the PR. The first PR is pretty much rebase on top of latest Keycloak main (as of today).

The updates in the 2nd PR are especially:
- Review comments by @IngridPuppet 
- Moved `OID4VCActionTokenTest` to the new testsuite
- Added test for `authorization-code` grant, so that the flow is tested with both `authorization-code` and `pre-authorized code` grants

themes/src/main/resources/theme/keycloak.v2/login/oid4vc-credential-offer.ftl
IngridPuppet: That the app is informed of what action was taken changes everything. Thanks 👍

IngridPuppet (approved):
Thank you for the update. The PR looks good to me.

mposolda:
@IngridPuppet Thanks!

PR #46331, opened 2026-02-13
[client-v2] Handle patch logic in the service + ServiceExceptionMapper

### LINKED ISSUE #46328: Handle patch logic in the service
### Description

At this moment, in our Client Admin API v2, we handle logic in certain subresources that should be handled by the service. 

We should move the patch logic there.

mabartos:
- Closes #46328
- Closes #46329

@vmuzikar @shawkins @keycloak/cloud-native Could you please check it? Thanks


rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
```diff
@@ -31,9 +28,9 @@ public interface ClientApi {
     Response createOrUpdateClient(@Valid BaseClientRepresentation client);
 
     @PATCH
-    @Consumes(CONTENT_TYPE_MERGE_PATCH)
+    @Consumes(MediaType.TEXT_PLAIN)
     @Produces(MediaType.APPLICATION_JSON)
-    BaseClientRepresentation patchClient(JsonNode patch);
+    BaseClientRepresentation patchClient(String patch);
```
shawkins: Are you using String to allow for the possibility of yaml patches?

From a client usage perspective it may be better to have separate endpoints with different Consumes - otherwise won't the user need to modify the header?  Or does that come up with other endpoints already?

mabartos: > Are you using String to allow for the possibility of yaml patches? From a client usage perspective it may be better to have separate endpoints with different Consumes - otherwise won't the user need to modify the header? Or does that come up with other endpoints already?

@shawkins Technically, yes. The intention was using some multi-purpose endpoint for PATCH, that might support as a merge-patch, or a json-patch, or even SSA. But a good point for these clients as it might make more sense to have different endpoints for them once we support it. ATM, we support only one, so we can leave it as is. 

I'll update it. Thanks!

michalvavrik (approved):
FWIW LGTM

shawkins (dismissed):
LGTM, thanks @mabartos

mabartos:
@michalvavrik @shawkins Thanks! Resolved conflicts

shawkins (approved):
LGTM, thanks @mabartos

PR #46351, opened 2026-02-15
GH actions for ci.yaml workflow - testsuite deprecation check.

### LINKED ISSUE #46291: Introduce GH Actions check
### Description

- add GH actions check for changes related to Arquillian
- block PR: more then 20(maybe 50) additions/lines in testsuite package
- if blocked add notification that migration of the test class is required to the new framework - package tests
- add DEPRECATION.md file
- add Deprecated annotation to AbstractKeycloakTest and KeycloakModelTest

lhanusov:
Closes #46291

### DIFF
--- a/.github/actions/testsuite-deprecation-check/deprecation-check.sh
+++ b/.github/actions/testsuite-deprecation-check/deprecation-check.sh
@@ -1,0 +1,26 @@
+#!/bin/bash -e
+
+REPOSITORY="$1"
+REF="$2"
+
+CHANGE_ID=$(echo $REF | cut -f 3 -d '/')
+
+echo "========================================================================================"
+echo "Checking testsuite module additions/changes."
+echo "----------------------------------------------------------------------------------------"
+
+ADDED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select (.status | contains("added")) | {filename}) | length')
+CHANGED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select (.additions >= 50) | {filename}) | length')
+
+# Check if changed files matches regex
+if [[ $ADDED_FILES > 0 || $CHANGED_FILES > 0 ]]  ; then
+  echo "========================================================================================"
+  echo "Deprecated testsuite module: "
+  echo " * Adding new file(s) is forbidden."
+  echo " * Maximum 50 lines can be added to a single file."
+  echo ""
+  echo "Please, migrate the added/changed file(s) and use the new test framework instead."
+  echo "See: https://github.com/keycloak/keycloak/tree/main/testsuite/DEPRECATED.md for more details."
+  echo "----------------------------------------------------------------------------------------"
+  exit 1
+fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,18 @@ jobs:
           fi
           echo "run-additional-dbs-tests=$RUN_ADDITIONAL_DBS_TESTS" >> $GITHUB_OUTPUT
 
+  testsuite-deprecation-check:
+    name: Testsuite Deprecation Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - id: conditional
+        uses: ./.github/actions/testsuite-deprecation-check
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+
   build:
     name: Build
     if: needs.conditional.outputs.ci == 'true'
@@ -1252,6 +1264,7 @@ jobs:
       - base-new-integration-tests
       - mixed-cluster-compatibility-tests
       - admin-v2-tests
+      - testsuite-deprecation-check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
--- a/testsuite/DEPRECATED.md
+++ b/testsuite/DEPRECATED.md
@@ -1,0 +1,27 @@
+# Deprecating the Old Testsuite Module
+If you find yourself reading this document, you are probably asked to migrate a test from your PR to use the new Test Framework.
+
+---
+With the release of the new Test Framework, this testsuite module, with all related dependents, is officially deprecated.
+
+Specifically speaking: 
+* **Arquillian** testsuite
+* **Model** testsuite
+* All related **utility** modules
+
+A limited amount of changes to existing tests are permitted, and should primarily be used to add test-cases when resolving bugs. Adding new files is not allowed.
+
+## Why Deprecated?
+* The Arquillian framework is not a community maintained project anymore. 
+* It's using JUnit 4
+* The test configuration got messy over the years. 
+* Onboarding process has a steep learning curve.
+* Adding new features is complex and non-trivial task.
+
+## How to Migrate Your Test?
+The new Test Framework brings completely new test configuration and server lifecycle management.
+It is tailored to quickly onboard and speed-up the feature development, or contributing a simple bug fix.
+
+Please, follow the new guidelines: 
+* [Test Framework](../test-framework/docs/README.md)
+* [Migrating Tests](../tests/MANUAL_MIGRATION.md)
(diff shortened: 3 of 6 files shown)


stianst (commented):
We should add `testsuite/DEPRECATED.md` in the same PR; and it would probably be better to link to that file than the new test framework docs, as it can explain that it's deprecated, why, and also link to the new test framework docs and new base tests module.

.github/actions/testsuite-deprecation-check/deprecation-check.sh
```diff
@@ -0,0 +1,22 @@
...
+REPOSITORY="$1"
+REF="$2"
+
+CHANGE_ID=$(echo $REF | cut -f 3 -d '/')
+
+echo "========================================================================================"
+echo "Checking testsuite module changes."
+echo "----------------------------------------------------------------------------------------"
+
+CHANGED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select (.changes >= 50) | {filename}) | length')
```
stianst: Very nice and clean approach. Have you tried it out to verify this does what we want? Will it also include new files and not just updated files?

stianst: We should differentiate between new files (not allowed) and updates to existing files (only X lines allowed)

testsuite/model/src/test/java/org/keycloak/testsuite/model/KeycloakModelTest.java
```diff
@@ -127,6 +127,7 @@
  * If no parameters are set via this property, the tests derived from this class are skipped.
  * @author hmlnarik
  */
+@Deprecated(since = "26.6.0", forRemoval = true)
```
stianst: The old testsuite wasn't release/supported, so no need for `since`

```suggestion
@Deprecated(forRemoval = true)
```

testsuite/DEPRECATED.md
```diff
@@ -0,0 +1,6 @@
+# Keycloak 26.6.0 - Deprecation of the Old Testsuite Module
+With the release of the new Test Framework in **26.6.0**, this testsuite module, with all related dependents, is officially deprecated.
```
stianst: Not tied to a release - you can just state it's deprecated

testsuite/DEPRECATED.md
```diff
@@ -0,0 +1,6 @@
...
+
+Please, follow the new guidelines: 
+* [Test Framework](test-framework/docs/README.md)
+* [Migrating Tests](tests/MANUAL_MIGRATION.md)
```
stianst: Links won't work, you probably need to just add `../`

.github/actions/testsuite-deprecation-check/deprecation-check.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Check if changed files matches regex
+if [[ ADDED_FILES > 0 || $CHANGED_FILES > 0 ]]  ; then
+  echo "========================================================================================"
+  echo "Deprecated testsuite module: "
+  echo " * Adding new file/s is forbidden."
+  echo " * Exceeding 50 lines per a modified file is not allowed."
+  echo ""
+  echo "Please, migrate the added/changed file/s and use the new test framework instead."
+  echo "Docs: https://github.com/keycloak/keycloak/tree/main/testsuite#deprecated"
```
stianst:
```suggestion
  echo "See: https://github.com/keycloak/keycloak/tree/main/testsuite/DEPRECATED.md for more details" 
```

stianst (commented):
You could probably also elaborate a bit more in DEPRECATED.md; What specifically is deprecated, why, and a little bit more details on what to do instead.

.github/actions/testsuite-deprecation-check/deprecation-check.sh
```diff
@@ -0,0 +1,26 @@
...
+
+ADDED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select(.filename | contains("testsuite/DEPRECATED.md") | not) | select (.status | contains("added")) | {filename}) | length')
+CHANGED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select (.additions >= 50) | {filename}) | length')
+
+# Check if changed files matches regex
+if [[ ADDED_FILES > 0 ]] || [[ $CHANGED_FILES > 0 ]]  ; then
+  echo "========================================================================================"
+  echo "Deprecated testsuite module: "
+  echo " * Adding new file/s is forbidden."
+  echo " * Exceeding 50 lines per a modified file is not allowed."
```
stianst: This is not a valid sentence; maybe something like:
```suggestion
  echo " * Maximum 50 lines can be added to a single file."
```

.github/workflows/ci.yml
```diff
@@ -87,11 +87,23 @@ jobs:
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+
   build:
     name: Build
     if: needs.conditional.outputs.ci == 'true'
     runs-on: ubuntu-latest
-    needs: conditional
+    needs: [conditional, testsuite-deprecation-check]
```
stianst: We need to be able to merge certain PRs that are updating the old testsuite; for example a PR that was close to completed using the old testsuite, or a bug fix that requires a few more lines to fix.

An admin can merge a PR even if one check fails, but with this the tests won't actually run.

To work around this; remove this change; and instead add it to `check` at the bottom. Then tests will run, and an admin can verify tests pass.

.github/actions/testsuite-deprecation-check/deprecation-check.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "----------------------------------------------------------------------------------------"
+
+ADDED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select (.status | contains("added")) | {filename}) | length')
+CHANGED_FILES=$(gh api -X GET --paginate repos/$REPOSITORY/pulls/$CHANGE_ID/files --jq 'map(select(.filename | contains("testsuite/")) | select (.additions >= 50) | {filename}) | length')
+
+# Check if changed files matches regex
+if [[ $ADDED_FILES > 0 || $CHANGED_FILES > 0 ]]  ; then
+  echo "========================================================================================"
+  echo "Deprecated testsuite module: "
+  echo " * Adding new file/s is forbidden."
```
stianst:
```suggestion
  echo " * Adding new file(s) is forbidden."
```

testsuite/DEPRECATED.md
```diff
@@ -0,0 +1,26 @@
...
+
+---
+With the release of the new Test Framework, this testsuite module, with all related dependents, is officially deprecated.
+
+Specifically speaking: 
+* **Arquillian** testsuite
+* **Model** testsuite
+* all related **utility** modules
+
+Bug fixes of particular tests are allowed, but limited to maximum 50 lines addition. Adding new file/s is not allowed.
```
stianst:
```suggestion
A limited amount of changes to existing tests are permitted, and should primarily be used to add test-cases when resolving bugs. Adding new files is not allowed.
```

testsuite/DEPRECATED.md
```diff
@@ -0,0 +1,26 @@
+# Deprecating the Old Testsuite Module
+If you find yourself reading this document, you are probably asked to migrate a test from your PR to use the new Test Framework.
+
+---
+With the release of the new Test Framework, this testsuite module, with all related dependents, is officially deprecated.
+
+Specifically speaking: 
+* **Arquillian** testsuite
+* **Model** testsuite
+* all related **utility** modules
```
stianst:
```suggestion
* All related **utility** modules
```

testsuite/DEPRECATED.md
```diff
@@ -0,0 +1,26 @@
...
+
+Bug fixes of particular tests are allowed, but limited to maximum 50 lines addition. Adding new file/s is not allowed.
+
+## Why Deprecated?
+* The Arquillian framework is not a community maintained project anymore. 
+* The test configuration got messy over the years. 
+* Onboarding process has a steep learning curve.
+* Adding new features is complex and non-trivial task.
```
stianst, line 19: Maybe we should add that Arquillian is stuck on old JUnit and Sellenium?

lhanusov: this is somewhat complicated, I will try to rephrase it: 
- Junit5/6 support for the core: https://github.com/arquillian/arquillian-core/blob/main/junit5/README.adoc
- Selenium 4 for the graphene: https://github.com/arquillian/arquillian-graphene/blob/main/pom.xml#L43
- Selenium 4 for the Drone: https://github.com/arquillian/arquillian-extension-drone/blob/main/pom.xml#L48

what I know, we are already using drone and graphene with Selenium 4

stianst: Okay, what you have now is fine

stianst (approved):

PR #46376, opened 2026-02-16
Cache evaluation of client roles with dots for role mapper

### LINKED ISSUE #43726: Slow evaluation of roles with dots for role mapper and others

pruivo:
Closes #43726


ahus1 (approved):
Thank you for preparing this PR, it should help a lot with performance in this area!

PR #46378, opened 2026-02-16
fix: merging the platform concept into the application

shawkins:
closes: #46377


quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractUpdatesCommand.java
```diff
@@ -91,14 +91,8 @@ static Map<String, CompatibilityMetadataProvider> loadAllProviders() {
...
         // Initialize config without directly referencing MicroProfileConfigProvider
-        // as that currently causing classloading issue during command creation
-        var configProvider = ServiceLoader.load(ConfigProviderFactory.class)
-                .stream()
-                .findFirst()
-                .map(ServiceLoader.Provider::get)
-                .flatMap(ConfigProviderFactory::create)
-                .orElseThrow(() -> new RuntimeException("Failed to load Keycloak Configuration"));
-        Config.init(configProvider);
+        // as that currently causing classloading issue during command creation - when a provider jar is deleted
```
mabartos, line 78: So, there might still be some classloading issues? When a provider JAR is deleted? When it happens?

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
```diff
@@ -50,26 +47,65 @@
...
+        } else {
+            tmpDir = new File(dataDir, "tmp");
+            tmpDir.mkdirs();
+        }
+        return tmpDir;
+    }
+
+    public static File createTmpDirectory() {
+        try {
+            File tmpDir = Path.of(System.getProperty("java.io.tmpdir"), "server-tmp").toFile();
```
mabartos:
```suggestion
            File tmpDir = Path.of(System.getProperty("java.io.tmpdir"), "keycloak-server-tmp").toFile();
```

Might be better, right? And as the name of the dir is changed here in this PR, is it possible that people overiding the java.io.tmpdir will have the `keycloak-quarkus-tmp` dir leftover there as we don't delete it? But it should not be probably a problem.

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
```diff
@@ -50,26 +47,65 @@
...
+            if (tmpDir.exists()) {
+                org.apache.commons.io.FileUtils.deleteDirectory(tmpDir);
+            }
+            if (tmpDir.mkdirs()) {
+                tmpDir.deleteOnExit();
+            }
+            return tmpDir;
+        } catch (IOException ioex) {
+            throw new RuntimeException("It was not possible to create temporary directory keycloak-quarkus-tmp", ioex);
```
mabartos:
```suggestion
            throw new RuntimeException("It was not possible to create temporary directory keycloak-server-tmp", ioex);
```
Can we use this as nobody knows anything about Quarkus here?

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
```diff
@@ -50,26 +47,65 @@
...
+            // Should happen just in non-script launch scenarios
+            tmpDir = createTmpDirectory();
+        } else {
+            tmpDir = new File(dataDir, "tmp");
+            tmpDir.mkdirs();
+        }
+        return tmpDir;
+    }
+
+    public static File createTmpDirectory() {
```
mabartos:
```suggestion
    public static File createSystemTmpDirectory() {
```
Might be better to reflect what's happening here?

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
```diff
@@ -50,26 +47,65 @@
...
+            logger.debugv("Application: {0}", this.getClass().getName());
             loadConfig();
+        } catch (Throwable t) {
+            exit(t);
+        }
+    }
 
-            platform.onStartup(this::startup);
-            platform.onShutdown(this::shutdown);
+    public static String getTmpDirectory() {
```
mabartos, line 64:
```suggestion
    public static File getTmpDirectory() {
```
As we're exposing this, can we still return File as before? I think it's much cleaner. Otherwise, rename it to `getTmpDirectoryPathName` or sth like that?

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
```diff
@@ -50,26 +47,65 @@
...
+            }
+            return tmpDir;
+        } catch (IOException ioex) {
+            throw new RuntimeException("It was not possible to create temporary directory keycloak-quarkus-tmp", ioex);
         }
     }
 
+    protected abstract void exit(Throwable t);
+
+    protected abstract String getDataDir();
```
mabartos, line 82:
```suggestion
    protected abstract Optional<File> getDataDir();
```
As we're expecting to be the data dir nullable, can we use Optional? And maybe also return the File already?

services/src/test/java/org/keycloak/services/resteasy/ResteasyKeycloakApplication.java
```diff
@@ -88,4 +104,11 @@ protected void createTemporaryAdmin(KeycloakSession session) {
         // do nothing
     }
 
+    @Override
+    protected void loadConfig() {
+        Config.init(new JsonConfigProviderFactory().create()
+                .orElseThrow(() -> new RuntimeException("Failed to load Keycloak configuration")));
+        startup();
```
mabartos, line 111: Wouldn't be better to keep the `onStartup` callbacks in the KeycloakApplication? To not handle these individuals startups in the `loadConfig`s?

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/ExecutionExceptionHandler.java
```diff
@@ -110,8 +109,7 @@ private void logError(PrintWriter errorWriter, String errorMessage) {
 
     // The "cause" can be null
     private void logError(PrintWriter errorWriter, String errorMessage, Throwable cause) {
-        QuarkusPlatform platform = (QuarkusPlatform) Platform.getPlatform();
-        if (platform.isStarted()) {
+        if (Application.currentApplication() != null) {
```
mabartos: nitpick: I like more the callback style for startup + isStarted as we're sure the startup was executed. Can we be sure that the startup happened here? Probably yes, right?

mabartos (commented):
@shawkins Thanks! In general, LGTM. I've just added some questions/comments/suggestions.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/command/AbstractUpdatesCommand.java
shawkins: Unfortunately yes. Some classloading operations are trying to scan the classpath, which hangs if a deleted provider jar is no longer present. The loading of the MicroProfileConfigProvider class (not the instance creation) is a trigger. I don't see why we won't hit this again unfortunately in other classes, so we'll probably need to get better behavior from quarkus.

services/src/test/java/org/keycloak/services/resteasy/ResteasyKeycloakApplication.java
shawkins: I don't think so as the callbacks were implemented quite oddly.

I can understand if you'd rather have this method renamed though.

quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/cli/ExecutionExceptionHandler.java
shawkins: > nitpick: I like more the callback style for startup + isStarted as we're sure the startup was executed.
> Probably yes, right?

It can be changed to check the isStarted flag, but I'm really not sure what the intended behavior is here. 

My guess is that we're trying to redirect things to the log, instead of std err / out as soon as possible, so I'm just checking to see if the application exists. Could there be a situation where the application exists, but the logging configuration or something is invalid leading to a server crash - I'm not sure.

~If we want it to mimic what was there before, the it should also check the isStarted flag.~ - Application.isStarted is logically set at the end of start up after all the events have been fired, so we'd still need to either track the event or look for some other quarkus logic to do this.

If instead we want any exception that terminates the server to go to stderr - just like any problem prior to the quarkus launch would - then we should just remove this check altogether.

cc @keycloak/cloud-native

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
shawkins: Using a File just creates a little bit more garbage on each call, or makes us cache the File object. I'm fine with leaving it as a String. I'm indifferent on the method name as the meaning / usage is pretty clear as is, but it can certainly be changed.

mabartos: Ok, thanks for the additional info. I'm not strict here. Let's keep it as is.

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
mabartos: Agreed on the current form.

services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java
shawkins: Good catch. I was merging the logic in the platform classes and missed that.

shawkins:
@mabartos thinking about it more, I've updated things to simplify the temp directory logic even more - so that more of the responsibility over it is in the test logic.

quarkus/tests/junit5/src/main/java/org/keycloak/Keycloak.java
```diff
@@ -362,4 +362,24 @@ public void exit(int exitCode) {
...
+                throw new RuntimeException("Could not create temporary directory", e);
+            }
+        } else {
+            Path homeDir = Path.of(buildDir, name);
+            try {
+                FileUtils.deleteDirectory(homeDir.toFile());
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+            return homeDir;
```
mabartos, line 381: Returns path to the deleted directory? It should create the temp directory as well in this case, right?

services/src/test/java/org/keycloak/services/resteasy/ResteasyKeycloakApplication.java
mabartos: Yep, the rename is much better! Thanks

mabartos (commented):
@shawkins Looks good, just one thing to verify.

quarkus/tests/junit5/src/main/java/org/keycloak/Keycloak.java
shawkins: It could be change to do that, but it currently doesn't need to as downstream logic creates that or a child directory when needed.

mabartos (approved):
@shawkins LGTM, thanks!

mabartos:
@vmuzikar @Pepo48 @michalvavrik @edewit Do you want to review it as well?

Pepo48 (approved):
LGTM. Thank you, Steve!

pedroigor (approved):
Approving based on @mabartos and @Pepo48 review.

PR #46380, opened 2026-02-16
fix: adding generic operator upgrade instructions

### LINKED ISSUE #40902: More fully document operator upgrade scenarios, in particular with custom images

shawkins:
closes: #40902

### DIFF
--- a/docs/guides/operator/installation.adoc
+++ b/docs/guides/operator/installation.adoc
@@ -144,4 +144,22 @@ If you do this please be aware:
 
 It is therefore recommended in a multiple Operator install scenario that you keep versions aligned as closely as possible to minimize the potential problems with different versions.
 
+=== Upgrading an Operator
+
+While it generally expected that server custom image versions will match the operator version, this can be difficult to properly maintain during an operator upgrade.
+
+The recommended approach is to first update your CRs to custom images utilizing the new version. While the CR configuration will generally be compatible, be aware this may not hold for major version changes, and always consult the upgrade notes for all version changes.
+Once all the Keycloak instances managed by the operator are updated, then update the operator to the new target version.
+
+If due to a major version change or bug an old operator is not compatible with a new custom image version, then you will instead need to stop the old operator when updating the custom images, then bring up the new operator version once the Keycloak CRs are up-to-date.
+
+To stop an operator you uninstall it scale the Deployment to 0 replicas:
+
+[source,bash,subs="attributes+"]
+----
+kubectl scale deployment/keycloak-operator --replicas=0
+----
+
+See <@links.operator id="rolling-updates"/> to understand how the operator will treat server version changes.
+
 </@tmpl.guide>
--- a/docs/guides/operator/installation.adoc
+++ b/docs/guides/operator/installation.adoc
@@ -153,7 +153,7 @@ Once all the Keycloak instances managed by the operator are updated, then update
 
 If due to a major version change or bug an old operator is not compatible with a new custom image version, then you will instead need to stop the old operator when updating the custom images, then bring up the new operator version once the Keycloak CRs are up-to-date.
 
-To stop an operator you uninstall it scale the Deployment to 0 replicas:
+To stop an operator you uninstall it, or scale down the Deployment to 0 replicas:
 
 [source,bash,subs="attributes+"]
 ----


docs/guides/operator/installation.adoc
```diff
@@ -144,4 +144,15 @@ If you do this please be aware:
 
 It is therefore recommended in a multiple Operator install scenario that you keep versions aligned as closely as possible to minimize the potential problems with different versions.
 
+=== Upgrading an Operator
+
+While it generally expected that server custom image versions will match the operator version, this can be difficult to properly maintain during an operator upgrade.
+
+The recommended approach is to first update your CRs to custom images utilizing the new version. While the CR configuration will generally be compatible, be aware this may not hold for major version changes, and always consult the upgrade notes for all version changes.
```
slaskawi, line 151: Although I'm absolutely fine with this approach, I just wanted call out that the opposite approach is more common among Kubernetes Ecosystem. 

The primary reason for such an approach is that when you upgrade the operand first, the operator may not know that (latests) operand version and may not properly validate the upgrade path. When upgrading Operator first, we ensure it knows all the latests versions of the operand and can validate (and coordinate) the upgrade path.

slaskawi (approved):

docs/guides/operator/installation.adoc
shawkins: > Although I'm absolutely fine with this approach, I just wanted call out that the opposite approach is more common among Kubernetes Ecosystem.

Unfortunately we've opened the door via custom images to a whole class of problems that aren't typical in the Kubernetes Ecosystem.

Since our operator currently has no differentiated logic based upon the version of Keycloak it is targetting (nor really direct knowledge of the version at all other than assuming it matches the operator version), and users have no specific requirement on what version their custom image is based upon, we're just in a position of recommending the upgrade scenario that will be the least likely to cause problems, and will be the easiest to rollback should there be problems.

> The primary reason for such an approach is that when you upgrade the operand first, the operator may not know that (latests) operand version and may not properly validate the upgrade path.

Can you elaborate on what you mean by upgrade path in this context?

Since the CRDs are unchanged when the operand / CR is updated - it is not exercising any new operator functionality. In an ideal scenario the only thing you're changing in the CR prior to the operator update is the image.

slaskawi: Of course! 

Let me illustrate this by an imaginary example. Let's assume Keycloak 26.6 has just been released and it contains a flag that **must** be set in container mode. Keycloak breaks if it's not. 

If we keep Keycloak Operator 26.5 running in our cluster and upgrade the operand first - Keycloak will crash (because the operator 26.5 is not aware of that new flag introduced in 26.6). 

If we upgrade things the other way around - the Operator first, everything should be fine. The Operator will be aware of this flag and will properly use it when upgrading the operand.

shawkins: What you are describing is a general breaking change, not specifically related to the operator, which we will avoid if at all possible. Furthermore the configuration change should be possible in existing 26.5 Keycloak CR (in particular via additionalOptions or env).

So for what you are describing to be a problem we would have to:
1. Introduce a breaking change
2. Only be able to address the breaking change via something like the unsupported template, and have that workaround be invalid (not simply deprecated) in the next version of the operator.

shawkins:
@vmuzikar confirmed with https://github.com/operator-framework/operator-lifecycle-manager/issues/1774 that patching the CSV is sufficient (I had referenced this issue before, but had forgotten about it) given that we are in manual upgrade mode. Haven't confirmed, but patching the deployment directly may not be predictable enough.

Edit: double-check and OLM won't update the Deployment in a non-upgrade scenario if the olm.deployment-spec-hash annotation remains unchanged, so it's safe to just change the replicas directly on the Deployment.

Uninstalling the operator is also possible, but I think that leads to more operator downtime.

docs/guides/operator/installation.adoc
vmuzikar: @slaskawi In general, minor upgrades must not introduce any breaking changes. However, it might introduce some new options for example, that might be used by the Operator. So newer Operand can work with older Operator without any issues thanks to backward compatibility. But not the other way around.

That said, this is applicable only to minor (and of course patch) releases. There the Operator must be shut down.

docs/guides/operator/installation.adoc
```diff
@@ -144,4 +144,22 @@ If you do this please be aware:
...
+Once all the Keycloak instances managed by the operator are updated, then update the operator to the new target version.
+
+If due to a major version change or bug an old operator is not compatible with a new custom image version, then you will instead need to stop the old operator when updating the custom images, then bring up the new operator version once the Keycloak CRs are up-to-date.
+
+To stop an operator you uninstall it scale the Deployment to 0 replicas:
```
vmuzikar:
```suggestion
To stop an operator you uninstall it, or scale down the Deployment to 0 replicas:
```

docs/guides/operator/installation.adoc
```diff
@@ -144,4 +144,22 @@ If you do this please be aware:
...
+
+[source,bash,subs="attributes+"]
+----
+kubectl scale deployment/keycloak-operator --replicas=0
```
vmuzikar, line 160: > double-check and OLM won't update the Deployment in a non-upgrade scenario if the olm.deployment-spec-hash annotation remains unchanged, so it's safe to just change the replicas directly on the Deployment.

Can we really make this assumption? What if this behavior is not intentional and will change without warning?

shawkins: It is intentional based upon previous OLM issues as they would get stuck in update loops otherwise - that would be true even if they utilized SSA.

Only when you change upgrade, or modify the CSV in such a way as to change the computed hash, does OLM patch / update the Deployment.

vmuzikar: Ok, works for me then.

vmuzikar (approved):

slaskawi (commented):
LGTM!

PR #46385, opened 2026-02-16
[OID4VCI] Add support for `authorization_code` grant

### LINKED ISSUE #46261: [OID4VCI] Add support for issuer_state on credential offer
### Description

It is important to be able to correlate an authorization request to a credential offer that was previously made to a given user. 

This `issuer_state` does not apply to pre-auth offers, because the `offer => user + credential` association had already been made by the Issuer when creating the pre-auth offer.

> Grant Type authorization_code:
issuer_state: OPTIONAL. String value created by the Credential Issuer and opaque to the Wallet that is used to bind the subsequent Authorization Request with a context set up during previous process steps. If the Wallet decides to use the Authorization Code Flow and received a value for this parameter, it MUST include it in the subsequent Authorization Request to the Authorization Server as the issuer_state parameter value.

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-offer-parameters

### Value Proposition

In a follow up step, we can introduce "issuance policies" on the client scope that models the credential configuration.

```
vc.policy.credential_offer.required
```

An authorization request based on `scope` alone or on `authorization_details` would fail when for such a `credential_configuration` when the Wallet fails to provide the correct `issuer_state`

### Goals

* Associate minimal required `issuer_state` with credential offers. 
* Be able to correlate an authorization request to an existing offer based on that state

### Non-Goals

Initially simply use some encoded json. Later, we can migrate the `issuer_state` to a signed jwt

tdiesler:
closes #46261 

This PR provides

* support for `authorization_code` grant
* initial support for `issuer_state`
* fixes incorrect usage of `credential_identifier`
* removes `createOfferStateForAuthorizationCodeFlow`
* fixes `authorization_details` for scope only AuthorizationRequest
* single entry point for credential offer creation

[https://github.com/openid/OpenID4VC-HAIP/issues/60](https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_openid_OpenID4VC-2DHAIP_issues_60&d=DwMFaQ&c=BSDicqBQBDjDI9RkVyTcHQ&r=gC-wVgmmi0hPUGBOY13Qrjsj2AA8val85l0wZlHck_Q&m=6tCRRDKL7PJbyz8TeUHoiCap0Xa9B51SHL0Fe1SxNzQboq_O1EZsq5QHBVtWRcny&s=abFX6VZhSFJPCopo_Z52Kau2eLEiLhiiFPlxsYYwUkA&e=)
[https://github.com/italia/eid-wallet-it-docs/issues/172](https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_italia_eid-2Dwallet-2Dit-2Ddocs_issues_172&d=DwMFaQ&c=BSDicqBQBDjDI9RkVyTcHQ&r=gC-wVgmmi0hPUGBOY13Qrjsj2AA8val85l0wZlHck_Q&m=6tCRRDKL7PJbyz8TeUHoiCap0Xa9B51SHL0Fe1SxNzQboq_O1EZsq5QHBVtWRcny&s=1V9sbO5MbJKVIpvtkqlRl8fEUFLT9yC5yGJ0WAAG9o0&e=)


mposolda (commented):
PR is hard to review. It touches 45 files and touches multiple areas. Is it possible to rather divide into more PRs, which would touch single area and focus on single issue each?

As `Support for authorization_code grant` and changes related to credential-identifier could be rather handled independently. Also there is the conflict now as there was other PR in the meantime, which touched the area of `fixes authorization_details for scope only AuthorizationRequest` (however that other PR was more isolated and fixed just that instead of touching big amount of files with multiple changes).

Ideal would be also to not migrate tests in the same PR, which changes the "logic". Would be good to either migrate the test beforehand in the dedicated PR or keep the test in the old testsuite in case PR touches the "codebase" .

tdiesler:
I split up this PR into these independent areas ...

1. Migrate OID4VCCredentialOfferMatrixTest - without any other functional change
2. Migrate CredentialsOffer to multiple grant types - which is foundational and ripples into many files. Look at  CredentialsOffer
3. Remove deprecated constants used in the old testsuite - cosmetic. Was there too not hit the max 50 lines change wall
4. Add support for `authorization_code` grant - the main topic of this PR
5. Add CredentialOfferProvider as single entry for offer creation 

### CredentialOfferProvider as single entry for offer creation 

Start with DefaultCredentialOfferProvider.createCredentialOffer(...), this is entry for credential offer creation. It creates an immutable CredentialOfferState that already has AuthorizationDetails attached to it.

The offer state can be found by ...

* The nonce from the credential_offer_uri
* The pre-auth_code grant (optionally)
* It's unique id as recorded in the opaque `issuer_state`

The `credential_idetifier` is no longer a key that can be used to find the offer state.

Much of the offer validation logic, that was previously found in the endpoint has moved here. The endpoint becomes easier because it now delegates to this provider. The provider would also be the place to apply policies related to offer creation.

### The `authorization_code` grant

A credentials offer can be created with `authorization_code` grant with the `preAuthorized=false` on the offer endpoint or directly through the CredentialOfferProvider. In which case a CredentialOfferState is created and it's unique id recorded in the IssuerState, which the Wallet receives as part of the credentials offer. The Wallet can then play back the `issuer_state` as part of the AuthorizationRequest. When processing authorization details as part of the AccessToken Request, we can find the associated CredentialOfferState and hence have access to it's already existing authorization details. 

An `authorization_code` grant offer can also be targeted to a specific user. For details see the discussion or javadoc on CredentialOfferProvider.

By not providing `issuer_state` as part of the AuthorizationRequest, the Wallet can by-pass the credential offer process i.e. it can request a credential by scope alone. This is the moment to shine for `vc.policy.offer.required` - with this policy enabled (PR #46466), an AuthorizationRequest without `issuer_details` that references a valid CredentialsOffer will be rejected.

Other than the `pre-authorized_code` grant, an `authorization_code` grant can be anonymous (i.e. not bound to a specific user). Whoever has such an offer can fetch the associated credentials. This is a mid-way approach, between completely open (i.e. request by scope) and targeted (i.e. only for a specific user). Using this, the Issuer can still reserve the right to offer credential configurations before they can get accessed.

### Processing authorization details

OID4VCAuthorizationDetailsProcessor.buildAuthorizationDetail() and handleMissingAuthorizationDetails() 
become a lot easier because they can access the offer state (when there is one), otherwise the same authorization details creation logic is used as in the CredentialOfferProvider.

The Access token will always have authorization_details.

### Credential Identifier

Is no longer a system generated UUID, instead it is the the value of the VC_IDENTIFIER attribute from the CredentialScopeModel

### Request Credential Endpoint

Now has the notion of 

* authorized credential_configuration_id + credential_identifier
* requested credential_configuration_id + credential_identifier
* offered credential_configuration_ids

Checks are in place that only the credential that was (optionally) offered and was authorized can be requested.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/LoginUrlBuilder.java
```diff
@@ -97,14 +119,33 @@ public LoginUrlBuilder requestUri(String requestUri) {
...
+
     @Override
     protected void initRequest() {
-        parameter(OAuth2Constants.RESPONSE_TYPE, client.config().getResponseType());
-        parameter(OIDCLoginProtocol.RESPONSE_MODE_PARAM, client.config().getResponseMode());
-        parameter(OAuth2Constants.CLIENT_ID, client.config().getClientId());
-        parameter(OAuth2Constants.REDIRECT_URI, client.config().getRedirectUri());
-
-        parameter(OAuth2Constants.SCOPE, client.config().getScope());
+        if (!params.containsKey(OAuth2Constants.CLIENT_ID)) {
```
mposolda, line 134: As discussed in the other PR, it will be good to avoid the changes in the method `initRequest()` . As AFAIK it is by design that those parameters like `responseType` are taken from config.

IMO the only change needed in the `LoginUrlBuilder` in this commit is to add `authorizationDetails` support.

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
```diff
@@ -0,0 +1,383 @@
...
+        assertEquals(1, jwtAuthDetails.size(), "Expected one authorization_details entry");
+        var jwtAuthDetail = jwtAuthDetails.get(0);
+
+        assertEquals(ctx.credConfigId, tokenAuthDetail.getCredentialConfigurationId());
+        assertEquals(tokenAuthDetail, jwtAuthDetail);
+
+        return accessToken;
+    }
+
+    public static class AuthorizationEndpointRequest {
```
mposolda, line 312: As discussed already in the other PR, it will be good to avoid introducing another "request" class, which mimics `LoginUrlBuilder` with very little addition.

As we will have another place to update always when new parameter would be added to "OIDC authentication endpoint" .

The existing OAuth-client API can be used already without a need to introduce this class to do the same thing in different ways...

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -567,11 +567,10 @@ public Response createCredentialOffer(
                 .detail(Details.VERIFIABLE_CREDENTIAL_TARGET_USER_ID, targetUserId)
                 .success();
 
-        String redactedTxCode = !Strings.isEmpty(offerState.getTxCode()) ? "******" : null;
         CredentialOfferURI credOfferURI = new CredentialOfferURI()
                 .setIssuer(credOffer.getCredentialIssuer() + "/protocol/" + OID4VC_PROTOCOL + "/" + CREDENTIAL_OFFER_PATH)
                 .setNonce(offerState.getNonce())
-                .setTxCode(redactedTxCode);
+                .setTxCode(offerState.getTxCode());
```
mposolda, line 573: This should not be added as suggested in some different PR.

In the new testsuite, there is no `TestResourceProvider`, however in the test, it is still possible to use something like:

```
@InjectRunOnServer
RunOnServerClient runOnServer;

...

public String getTxCode(final String preAuthCode) {
        return runOnServer.fetch(session-> {
            var offerStorage = session.getProvider(CredentialOfferStorage.class);
            var offerState = offerStorage.findOfferStateByCode(session, preAuthCode);
            return Optional.ofNullable(offerState).map(CredentialOfferState::getTxCode).orElse(null);
        }, String.class);
}
```

This can be used temporarily until we have `txCode` proper support (there is an issue assigned to me, but did not started on it yet as it may conflict with some other work you are doing).

mposolda (commented):
@tdiesler Thanks for splitting this into separate commits. Added review comments only to the first commit (Migrate OID4VCCredentialOfferMatrixTest). Can be good to extract that commit into separate PR and then handle other things as a follow-ups.

mposolda:
@tdiesler Thanks for the update! So is the plan to split this into dedicated PRs? So far, I've added the review only to the first commit (Migrating the `OID4VCCredentialOfferMatrixTest`). Feel free to send dedicated PR for that commit (once PR review comments addressed).

Some other comments below to your last comment:

> The credential_idetifier is no longer a key that can be used to find the offer state.

It probably works and is better way than our current use of `credential_identifier` . At the same time, our current use of `credential_identifier` makes it possible to make sure that access-token used in the credential-request is really the same permission as the one returned from the corresponding token response . That is due the fact that currently access-token contains `authorization_details` claim with unique `credential_identifier` inside it, which can be correlated with the corresponding entry related to that `credential_identifier` in the credential-offer-storage. With this change, we in fact loose this correlation in some way (thinking if there are some corner case scenarios, which would make it possible that incorrect access-token would be sent in the credential-request). But we can possibly solve it by adding the custom claim into access-token, which would be related to the ID inside credential-offer.

> By not providing issuer_state as part of the AuthorizationRequest, the Wallet can by-pass the credential offer process i.e. it can request a credential by scope alone. This is the moment to shine for vc.policy.offer.required - with this policy enabled (PR https://github.com/keycloak/keycloak/pull/46466), an AuthorizationRequest without issuer_details that references a valid CredentialsOffer will be rejected.

Just a note that we may need also something like "Allow user self-issued credential" - https://github.com/keycloak/keycloak/issues/45777 . The `vc.policy.offer.required` is likely useful, however user himself can create offer in theory with his access token only, which may not be desired behaviour. As in some cases, you may want that user can obtain credential just if it was approved by administrator.

tests/utils-shared/src/main/java/org/keycloak/testsuite/util/oauth/LoginUrlBuilder.java
tdiesler: done, except for scope

tests/base/src/test/java/org/keycloak/tests/oid4vc/OID4VCBasicWallet.java
tdiesler: This small wrapper is part of fundamental Wallet functionality and at the same time it is an implementation detail of it. Is serves as the Wallet's abstraction for authorization requests, not just auth code flow. If you're not ussing the wallet, you won't even see it. Unfortunately, it can't be removed.

```
    public static class AuthorizationEndpointRequest {

        protected final AbstractOAuthClient<?> client;
        protected final LoginUrlBuilder loginForm;

        public AuthorizationEndpointRequest(AbstractOAuthClient<?> client) {
            this.client = client;
            this.loginForm = client.loginForm();
        }

        public AuthorizationEndpointRequest authorizationDetails(OID4VCAuthorizationDetail authDetail) {
            loginForm.authorizationDetails(List.of(authDetail));
            return this;
        }

        public AuthorizationEndpointRequest codeChallenge(PkceGenerator pkce) {
            loginForm.codeChallenge(pkce);
            return this;
        }

        public AuthorizationEndpointRequest request(String request) {
            loginForm.request(request);
            return this;
        }

        public AuthorizationEndpointRequest scope(String... scopes) {
            loginForm.scope(scopes);
            return this;
        }

        public AuthorizationEndpointResponse send(String username, String password) {
            loginForm.open();
            client.fillLoginForm(username, password);
            return client.parseLoginResponse();
        }
    }
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: What you suggest renders the feature useless outside the context of keycloak test cases. If there is no way for the Issuer to read the TxCode and pass it to the Wallet, it is probably best not to have that feature in the first place.

We can leave it as is until there is a alternative way to ready it or we remove it from the endpoint also until there is an alternative way to ready it.

A pre-auth code that is protected by a tx_code that may be compromised is still better than a pre-auth code that is not protected at all.

I'll take that feature out from this part and we can have another PR just for tx_code

PR #46387, opened 2026-02-16
Fix resource selection display in scope-based permission creation

### LINKED ISSUE #46384: Resource selection not displayed in scope-based permission creation when resource is not in initial results
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When creating a scope-based permission, selecting a resource via search that is not in the initial 10 results causes the selection to appear blank.

### Version

latest

### Regression

### Expected behavior

The selected resource name should be displayed in the input field after selection.

### Actual behavior

The input field appears empty despite the resource being correctly selected.

### How to Reproduce?

1. Have a client with authorization enabled and more than 10 resources.
2. Go to Authorization > Permissions > Create scope-based permission.
3. Search for a resource that is not in the first 10 results.
4. Select it from the dropdown.
5. The input shows blank.

### Anything else?

Video attached.

https://github.com/user-attachments/assets/871d4cfc-2f0e-4d27-8b81-4197b347912e

Alnyli07:
### Summary

- Fix single-select typeahead not displaying the selected resource name when the resource is outside the initial 10 results
- Resolve display from selected state (fetched by ID, always stable) instead of items (paginated search results that reset)
- Eagerly update selected on item pick so the name is available immediately without waiting for an async round-trip

Closes #46384

### DIFF
--- a/js/apps/admin-ui/src/clients/authorization/ResourcesPolicySelect.tsx
+++ b/js/apps/admin-ui/src/clients/authorization/ResourcesPolicySelect.tsx
@@ -284,7 +284,7 @@ export const ResourcesPolicySelect = ({
             selections={
               variant === SelectVariant.typeaheadMulti
                 ? field.value
-                : items.find((i) => i.id === field.value?.[0])?.name
+                : selected.find((i) => i.id === field.value?.[0])?.name
             }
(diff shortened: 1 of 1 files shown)


ssilvert (approved):
Tried it out.  Looks good.

PR #46392, opened 2026-02-16
Update external Infinispan metrics documentation

### LINKED ISSUE #46390: Update external Infinispan metrics documentation
### Description

The legacy metrics names must be enabled for the external Infinispan cluster metrics, so the user's dashboard and Prometheus alerts continue to work

pruivo:
Closes #46390


ryanemerson (dismissed):
I think we should add a note to the `changes-26_0-0.adoc` "Infinispan 16 upgrade" section either explaining that this specific change needs to be made, or that users should make sure that their existing CRs match those in the guides. My preference would be the latter as it covers other changes as well.

ahus1:
> I think we should add a note to the changes-26_0-0.adoc "Infinispan 16 upgrade" section 

Adding this now.

ahus1 (approved):

PR #46398, opened 2026-02-17
Limit the inflating size for the SAML redirect binding

### LINKED ISSUE #46372: CVE-2026-2575 - Denial of Service due to excessive SAMLRequest decompression
### Description

An unauthenticated remote attacker can trigger a Denial of Service (DoS) by sending a highly compressed SAMLRequest via the SAML Redirect Binding. The server fails to enforce size limits during DEFLATE decompression, leading to an OutOfMemoryError (OOM) and process termination.

rmartinc:
Closes #46372

Now the inflater is limited using the `getTotalOut`. The default limit is 128KB (which is the same than the post max-form-attribute-size we are setting to quarkus). The saml protocol factory has an option that allows the admin to start keycloak with another value. For the wildfly adapter, the PR adds a system property. Note added to the upgrade. I don't know how we can inform the new limit for the adapter and the system property, do I add another paragraph to the upgrading guide?

### DIFF
--- a/saml-core/src/main/java/org/keycloak/saml/processing/api/util/DeflateUtil.java
+++ b/saml-core/src/main/java/org/keycloak/saml/processing/api/util/DeflateUtil.java
@@ -36,6 +36,15 @@ import org.keycloak.saml.common.constants.GeneralConstants;
 public class DeflateUtil {
 
     /**
+     * Maximum size for inflating. Default is 128KB like quarkus.http.limits.max-form-attribute-size.
+     */
+    public static long DEFAULT_MAX_INFLATING_SIZE = 131072;
+
+    private DeflateUtil() {
+        // utility class
+    }
+
+    /**
      * Apply DEFLATE encoding
      *
      * @param message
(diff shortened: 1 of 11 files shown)


saml-core/src/main/java/org/keycloak/saml/processing/api/util/DeflateUtil.java
```diff
@@ -75,7 +84,90 @@ public static byte[] encode(String message) throws IOException {
...
+        private LimitedInflaterInputStream(InputStream is, int maxInflatingSize) {
+            this.inflater = new Inflater(true);
+            this.is = new InflaterInputStream(is, inflater);
+            this.maxInflatingSize = maxInflatingSize;
+        }
+
+        private void checkMaxInflatingsize() throws IOException {
+            if (inflater.getTotalOut() > maxInflatingSize) {
+                throw new IOException(String.format("Maximum inflating size %d reached. Total bytes witten %d.",
+                        maxInflatingSize, inflater.getTotalOut()));
```
ahus1: The JavaDoc of `getTotalOut()` mentions that `getBytesWritten()` should be used. When doing that, it should be a `long` and not an `int` in all places. 

```suggestion
            if (inflater.getBytesWritten() > maxInflatingSize) {
                throw new IOException(String.format("Maximum inflating size %d reached. Total bytes witten %d.",
                        maxInflatingSize, inflater.getBytesWritten()));
```

ahus1 (commented):
Thank you @rmartinc - the changes look good. One nit-pick below.

saml-core/src/main/java/org/keycloak/saml/processing/api/util/DeflateUtil.java
rmartinc: Yep! And it's even deprecated in 25, I didn't realize about that. Changed to `long` and `getBytesWritten`.

rmartinc:
After talking with @pskopek, I have also added a paragraph for the adapter. It seems that there is nothing special for the galleon layers.

ahus1 (approved):
Thank you for this change!

Issue #40398, opened 2025-06-11
ModelDuplicateException on next login after deleting an account and back-channel logout


mandritschke:
### Before reporting an issue

- [x] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

authentication

### Describe the bug

I can see a ModelDuplicateException caused by a unique constraint violation ([stacktrace.txt](https://github.com/user-attachments/files/20408451/stacktrace.txt)) after a user has been deleted.

This is a follow-up to #39923. The solution there, #40352, only addresses the issue for the browser based front-channel logout, which was easier to reproduce using Keycloak's internal account client. This report here concerns the back-channel logout.

### Version

26.2.4

### Regression

- [x] The issue is a regression

### Expected behavior

* next login with an unrelated user is successful

### Actual behavior

* next login fails with status code 500

### How to Reproduce?

* A user with an active session deletes themselves in a user storage external to Keycloak and then logs out using Keycloak's *back-channel logout* (`GET /realms/$realm/protocol/openid-connect/logout?id_token_hint=...`), i.e. not the browser based procedure
* After the logout the browser is redirected to the login page again
* Another unrelated user attempts to log in through this login page and receives an error with status code 500 and said ModelDuplicateException

For some reason reproducing this issue requires two iterations
* 1st pass
        1. (not signed in) go to account client, be redirected to authentication
        2. sign in as user1 to the account client ✅
        3. delete the user1 in the user storage
        4. sign out from the account client, be redirected to authentication
        5. sign in as user2 to the account client ✅
        6. sign out from the account client, be redirected to authentication
        7. delete the user2 in the user storage
* 2nd pass, same browser
        1. (not signed in) go to account client, be redirected to authentication
        2. sign in as user1 to the account client ✅
        3. delete the user1 in the user storage
        4. sign out from the account client, be redirected to authentication
        5. sign in as user2 to the account client 💥

mandritschke:
I updated the reproducer to demonstrate the issue using an external client:

https://github.com/mandritschke/reproduce-model-duplicate-exception

mandritschke:
Just found out that the problem doesn't occur with persistent user sessions disabled, i.e. `KC_FEATURES_DISABLED=persistent-user-sessions`.

graziang:
@ahus1 Could you please take a look? It seems related to persistent user sessions feature

ahus1:
[...]

PR #46401, opened 2026-02-17
Check offline session too in the broker OIDC endpoint

### LINKED ISSUE #46379: Unexpected error when logging out with offline session and external IDP
### Before reporting an issue

### Area

oidc

### Describe the bug

When a client starts a flow with only offline_access, no SSO session is generated. When the user used an external IDP to authenticate, Keycloak will still try to do a logout at the external IDP if one was used. The logout redirect to the external IDP works but it fails on the `logout_response` endpoint since no session is found.

### Version

26.5.3

### Regression

### Expected behavior

If the user tries to logout with valid ID token and post_logout_redirect_uri without an SSO session, Keycloak should just redirect back to the post_logout_redirect_uri.

### Actual behavior

Keycloak forwards to the external IDP for logout but fails to handle the logout_response from the external IDP.

### How to Reproduce?

https://github.com/user-attachments/assets/0d6320e7-a5f8-421a-b942-b630057d3bda

[docker-compose.yml](https://github.com/user-attachments/files/25341109/docker-compose.yml)

[test_setup.tf.txt](https://github.com/user-attachments/files/25341123/test_setup.tf.txt)

1. Install terraform (https://developer.hashicorp.com/terraform/install)
2. Install oauth2c (https://github.com/SecureAuthCorp/oauth2c)
3. Download the two attached files and put them into a directory, remove the `.txt` ending from `test_setup_tf_txt`.
4. Run `docker compose up`
5. Run `terraform init`
6. Run `terraform apply` and confirm with entering `yes`
7. Run: 
`echo "http://localhost:8080/realms/B/protocol/openid-connect/logout?client_id=test&post_logout_redirect_uri=http%3A%2F%2Flocalhost%3A8803%2Fcallback-logout&id_token_hint=$(oauth2c http://localhost:8080/realms/B \
  --client-id test \
  --grant-type authorization_code \
  --redirect-url http://localhost:8803/callback \
  --auth-method none \
  --response-types code \
  --response-mode query \
  --scopes "openid offline_access" \
  --pkce | grep '{"access_token":"ey' | jq -r .id_token)"`
8. Click "Realm A", login with user `test` password `test`
9. Copy the logout URL from the terminal window and paste it into the browser to trigger the logout.

Please also check the attached video.

### Anything else?

The full related code is quite complex but a related snippet is definitely https://github.com/keycloak/keycloak/blob/51010315168f8d59c246b60104ff6b67bb1e322d/services/src/main/java/org/keycloak/protocol/oidc/endpoints/LogoutEndpoint.java#L413-L417 where Keycloak continues a normal logout process even if an SSO session is not found. This handling was added in this commit: https://github.com/keycloak/keycloak/commit/88069cd5fb2a30c8a388040b0aebef0b38641604 closing this issue https://github.com/keycloak/keycloak/issues/40398. I am not sure if this was an unintended side effect.

rmartinc:
Closes #46379

Commit https://github.com/keycloak/keycloak/commit/88069cd5fb2a30c8a388040b0aebef0b38641604 in issue https://github.com/keycloak/keycloak/issues/40398 added offline sessions in the logout. But the endpoint was not modified in the same way. This way the frontend logout failed because the session is not found when going back from the external IdP. Test added.

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/OIDCIdentityProvider.java
@@ -385,6 +385,9 @@ public class OIDCIdentityProvider extends AbstractOAuth2IdentityProvider<OIDCIde
             }
             UserSessionModel userSession = session.sessions().getUserSession(realm, state);
             if (userSession == null) {
+                userSession = session.sessions().getOfflineUserSession(realm, state);
+            }
+            if (userSession == null) {
                 logger.error("no valid user session");
                 EventBuilder event = new EventBuilder(realm, session, clientConnection);
                 event.event(EventType.LOGOUT);
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/broker/AbstractBaseBrokerTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/broker/AbstractBaseBrokerTest.java
@@ -308,15 +308,20 @@ public abstract class AbstractBaseBrokerTest extends AbstractKeycloakTest {
         return contextRoot + "/auth/realms/" + realmName + "/account";
     }
 
+    protected String getLoginUrl(String contextRoot, String realmName, String clientId) {
+        return getLoginUrl(contextRoot, realmName, clientId, "openid");
+    }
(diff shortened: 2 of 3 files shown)


rmartinc:
Closing this as it was decided to do the other way around.

PR #46414, opened 2026-02-17
Added the shutdown delay and shutdown timeout for the testcases to default options 

### LINKED ISSUE #46337: Run tests with shutdown-delay of 0 seconds
### Description

After we added a default shutdown delay of 1 second, this will slow down running tests. 

Please review our test suite that we run Keycloak for tests with a 0 seconds shutdown delay to avoid delays in test execution.

ruchikajha95:
Closes #46337

Approach 1: 
Added --shutdown-delay=0s", "--shutdown-timeout=1s"

Tested eg: $ time mvn clean test -Dtest=UpdateCommandDistTest

Before changes output:
real    1m57.160s
user    12m21.342s
sys     0m34.469s

After changes output:
real    0m43.130s
user    3m2.158s
sys     0m5.901s

Approach 2: 

Added the Added --shutdown-delay=0s", "--shutdown-timeout=1s" implicitly instead of the default options.

### DIFF
--- a/quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/KeycloakDistributionDecorator.java
+++ b/quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/KeycloakDistributionDecorator.java
@@ -41,9 +41,8 @@ public class KeycloakDistributionDecorator implements KeycloakDistribution {
     @Override
     public CLIResult run(List<String> rawArgs) {
         List<String> args = new ArrayList<>(rawArgs);
-
         args.addAll(List.of(config.defaultOptions()));
-
+        setEnvVar("KC_SHUTDOWN_DELAY", "0s");
         return delegate.run(new ServerOptions(storageConfig, databaseConfig, args));
     }
 
--- a/testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/containers/AbstractQuarkusDeployableContainer.java
+++ b/testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/containers/AbstractQuarkusDeployableContainer.java
@@ -172,10 +172,12 @@ public abstract class AbstractQuarkusDeployableContainer implements DeployableCo
         commands.add("--http-relative-path=/auth");
         commands.add("--health-enabled=true"); // expose something to management interface to turn it on
 
+
         if (suiteContext.get().isAuthServerMigrationEnabled()) {
             commands.add("--hostname-strict=false");
         } else { // Do not set management port for older versions of Keycloak for migration tests - available since Keycloak 25
             commands.add("--http-management-port=" + configuration.getManagementPort());
+            commands.add("--shutdown-delay=0");
         }
 
         if (suiteContext.get().getMigrationContext().isRunningMigrationTest()) {
--- a/operator/src/test/resources/correct-podtemplate-keycloak.yml
+++ b/operator/src/test/resources/correct-podtemplate-keycloak.yml
@@ -13,6 +13,8 @@ spec:
       value: postgres
     - name: db-password
       value: testpassword
+    - name: shutdown-delay
+      value: "0s"
   hostname:
     hostname: example.com
   unsupported:
--- a/operator/src/test/resources/empty-podtemplate-keycloak.yml
+++ b/operator/src/test/resources/empty-podtemplate-keycloak.yml
@@ -13,6 +13,8 @@ spec:
       value: postgres
     - name: db-password
       value: testpassword
+    - name: shutdown-delay
+      value: "0s"
   hostname:
     hostname: example.com
   unsupported:
--- a/operator/src/test/resources/example-keycloak.yaml
+++ b/operator/src/test/resources/example-keycloak.yaml
@@ -22,3 +22,5 @@ spec:
   additionalOptions:
     - name: metrics-enabled
       value: "true"
+    - name: shutdown-delay
+      value: "0s"
--- a/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/BuildCommandDistTest.java
+++ b/quarkus/tests/integration/src/test/java/org/keycloak/it/cli/dist/BuildCommandDistTest.java
@@ -66,7 +66,12 @@ class BuildCommandDistTest {
     @Test
     @Launch({ "build", "--db=postgres", "--db-username=myuser", "--db-password=mypassword", "--http-enabled=true" })
     void testIgnoreRuntimeOptions(CLIResult cliResult) {
-        cliResult.assertMessage("The following run time options were found, but will be ignored during build time: kc.db-username, kc.http-enabled, kc.db-password");
+        String output = cliResult.getOutput();
+        assertTrue(output.contains("The following run time options were found, but will be ignored during build time:"));
+        assertTrue(output.contains("kc.db-username"));
+        assertTrue(output.contains("kc.http-enabled"));
+        assertTrue(output.contains("kc.db-password"));
+        assertTrue(output.contains("kc.shutdown-delay"));
         cliResult.assertBuild();
     }
 
--- a/test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
+++ b/test-framework/core/src/main/java/org/keycloak/testframework/server/KeycloakServerConfigBuilder.java
@@ -27,6 +27,8 @@ public class KeycloakServerConfigBuilder {
     private final Set<KeycloakDependency> dependencies = new HashSet<>();
     private CacheType cacheType = CacheType.LOCAL;
     private boolean externalInfinispan = false;
+    private String shutdownDelay = "0s";
+    private String shutdownTimeout = "1s";
 
     private KeycloakServerConfigBuilder(String command) {
         this.command = command;
@@ -91,6 +93,16 @@ public class KeycloakServerConfigBuilder {
 
     public boolean isExternalInfinispanEnabled() {
         return this.externalInfinispan;
+    }
+
+    public KeycloakServerConfigBuilder shutdownDelay(String shutdownDelay) {
+        this.shutdownDelay = shutdownDelay;
+        return this;
+    }
+
+    public KeycloakServerConfigBuilder shutdownTimeout(String shutdownTimeout) {
+        this.shutdownTimeout = shutdownTimeout;
+        return this;
     }
 
     /**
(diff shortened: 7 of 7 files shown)


quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/KeycloakDistributionDecorator.java
```diff
@@ -44,6 +44,18 @@ public CLIResult run(List<String> rawArgs) {
 
         args.addAll(List.of(config.defaultOptions()));
 
+        // Automatically inject shutdown optimization only for server start commands
```
pruivo: @ahus1 if we have `KC_SHUTDOWN_DELAY` and `--shutdown-delay`, which one has priority? 
Setting the environment variable would make things simpler. 
```java
delegate.setEnvVar("KC_SHUTDOWN_DELAY", "0s");
```

@ruchikajha95 the default shutdown timeout is 1s, you can remove the `--shutdown-timeout=1s` added from everywhere.

shawkins: @pruivo cli has the highest priority.

pruivo: @shawkins thanks!

ruchikajha95: Thanks @pruivo  for the review i have completed the expected changes. Could you review the changes again.

quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/KeycloakDistributionDecorator.java
```diff
@@ -44,6 +44,14 @@ public CLIResult run(List<String> rawArgs) {
...
+        if (isStartCommand(args)) {
+            if (args.stream().noneMatch(arg -> arg.startsWith("--shutdown-delay"))) {
+                if (args.stream().noneMatch(arg -> arg.startsWith("--shutdown-delay"))) {
```
ryanemerson: Duplicate if statement

ruchikajha95: Thanks @ryanemerson . Removed it .

quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/DistributionTest.java
```diff
@@ -70,7 +70,6 @@ enum ReInstall {
      * If any option must be set when starting the server.
      */
     String[] defaultOptions() default {};
-
```
pruivo, line 73: This change should be reverted.

quarkus/tests/junit5/src/main/java/org/keycloak/it/junit5/extension/KeycloakDistributionDecorator.java
```diff
@@ -41,9 +41,12 @@ public KeycloakDistributionDecorator(Storage storageConfig, WithDatabase databas
     @Override
     public CLIResult run(List<String> rawArgs) {
         List<String> args = new ArrayList<>(rawArgs);
-
         args.addAll(List.of(config.defaultOptions()));
-
+        if (isStartCommand(args)) {
```
pruivo: All changes in this file can be replaced with one line
```java
setEnvVar("KC_SHUTDOWN_DELAY", "0s");
```

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
```diff
@@ -69,6 +69,9 @@ public void start(KeycloakServerConfigBuilder keycloakServerConfigBuilder, boole
         this.tlsEnabled = tlsEnabled;
 
         List<String> args = keycloakServerConfigBuilder.toArgs();
+        if (args.stream().noneMatch(arg -> arg.startsWith("--shutdown-delay"))) {
```
pruivo: Same here, in the method `startKeycloak()`, you can add
```java
pb.environment().put("KC_SHUTDOWN_DELAY", "0s");
```

pruivo (changes requested):

pruivo (dismissed):

ryanemerson (dismissed):

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/containers/AbstractQuarkusDeployableContainer.java
```diff
@@ -172,6 +172,9 @@ protected List<String> getArgs(Map<String, String> env) {
         commands.add("--http-relative-path=/auth");
         commands.add("--health-enabled=true"); // expose something to management interface to turn it on
 
+        // reduce shutdown delay to speed up testsuite execution.
+        env.put("KC_SHUTDOWN_DELAY", "0s");
```
vmuzikar: Here, we could've used the CLI option to be consistent with other options being set here. In this testsuite we don't expect the tests would be overriding this option, so no need for the env var (if that was the motivation to use it over the CLI option in other places.

That said, it's not blocking, it's good as is.

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
```diff
@@ -164,6 +164,7 @@ private void checkRunning() {
 
         OutputHandler outputHandler;
         try {
+            pb.environment().put("KC_SHUTDOWN_DELAY", "0s");
```
vmuzikar: I think we might want to put this somewhere into default options in the testsuite? @shawkins Do you remember where?

shawkins: No, I don't recall a place for that.

shawkins (commented):
Is this only needed for distribution tests? What about embedded? And in general should start-dev shut down immediately?

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/containers/AbstractQuarkusDeployableContainer.java
pruivo: I suggested using the environment variable approach to keep consistent with the other places. 
You're right, it doesn't seem possible to overwrite the command list in this class.

pruivo:
@shawkins, we discussed the embedded mode, and we decided to ignore it and focus on making the GH actions faster. Embedded mode is not used by GH actions as far as we know, and it may go away with the new testsutite (?). 

We never thought about `start-dev`, unfortunately :sweat_smile:. I don't oppose having no delay during shutdown (assuming nobody uses it in production).

shawkins:
> @shawkins, we discussed the embedded mode, and we decided to ignore it and focus on making the GH actions faster. Embedded mode is not used by GH actions as far as we know, and it may go away with the new testsutite (?).

Embedded mode is supported with the new testsuite and does seem to be used, just not by GH - cc @stianst @lhanusov 

One other thought - is this adding any measurable delay to the operator tests?

testsuite/integration-arquillian/tests/base/src/main/java/org/keycloak/testsuite/arquillian/containers/AbstractQuarkusDeployableContainer.java
vmuzikar: If you mean consistent across testsuite, we're inconsistent already how each of them configures KC. The Arquillian testsuite uses CLI options, that was my argument here.

vmuzikar:
Embedded is used for local development where speed is also important. It's also important to be consistent across testsuite modes (remote/embedded). So +1 to @shawkins to add this to embedded too.

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
vmuzikar: I think I found it:
https://github.com/keycloak/keycloak/blob/94dc60822b2e6712535d4d1432b8e2bb4687c763/test-framework/core/src/main/java/org/keycloak/testframework/server/AbstractKeycloakServerSupplier.java#L89-L94

@lhanusov @stianst Can you please confirm this would be the correct place to provide default config options to the new testsuite?

ruchikajha95:
Thank you everyone for the insightful discussion.
I acknowledge the importance of ensuring that the shutdown delay configuration applies consistently to both remote and embedded server modes, as highlighted.

Could the team decide to centralize this configuration (e.g., within AbstractKeycloakServerSupplier as referenced), or if there are further changes required to ensure alignment across all test scenarios, I am happy to implement these improvements. Please let me know if you would prefer that I make these updates within this PR, or if it would be better to address them in a subsequent PR.

Thanks again for your thorough review and the collaborative discussion.

cc: @ahus1 , @pruivo , @ryanemerson , @shawkins , @vmuzikar

stianst:
For the new testsuite `KeycloakServerConfigBuilder` would be the best place to set the defaults; we'd also want to add methods there to allow overriding the defaults in case someone wants to use different shutdown delay and timeout for a test-case.

test-framework/core/src/main/java/org/keycloak/testframework/server/DistributionKeycloakServer.java
stianst: This is not the correct place to set the option; it should also not use environment variables.

Please add `shutdownDelay` and `shutdownTimeout` fields to `KeycloakServerConfigBuilder` and add methods to allow overriding those. Then in `toArgs` you can set those as CLI options.

stianst (dismissed):

ahus1:
@vmuzikar / @shawkins - if would handle @stianst review comment in this PR as it is directly related to the code change, can we handle your feedback in a follow-up pull request (see @ruchikajha95's comment) as it would touch some code that is not yet part of this PR?

vmuzikar:
@ahus1 I'm not sure I fully understand your proposal. I believe all outstanding review comments are similar – moving where we set the config options to a correct place.

pruivo (approved):

ahus1 (approved):
Thank you for addressing the review comments of the other team members!

PR #46420, opened 2026-02-17
Add pagination and optimize N+1 in User Organizations tab

### LINKED ISSUE #46413: [Admin UI] User's organizations list is delayed/empty until an N+1 cascade of API calls completes
### Before reporting an issue

### Area

admin/ui

### Describe the bug

When a single user is a member of a large number of Organizations (e.g., 1,000), navigating to that user's profile in the Admin Console results in the Organizations list appearing empty or stuck loading for a significantly long time.

The data only renders after the UI finishes executing an N+1 cascade of API calls to fetch the members for every single organization the user belongs to. The UI does not paginate the initial fetch, nor does it lazily load the secondary data, causing a massive network bottleneck.

Version Keycloak v26.x

### Version

26.4.7

### Regression

### Expected behavior

- Pagination: The initial request to fetch the user's organizations should be paginated by the UI (e.g., using ?first=0&max=10).
- Immediate Rendering: The UI should immediately display the list of organizations the user belongs to without waiting for secondary requests to resolve.
- Optimized Fetching: The UI should not trigger an N+1 cascade to fetch member details for every organization in the table. If secondary details (like member counts or user roles) are needed for the table, this data should be included in the initial payload, fetched via a bulk endpoint, or lazily loaded.

### Actual behavior

The Organizations tab appears empty while waiting for all network requests to resolve. Inspecting the browser's Network tab reveals the following sequence:

A single, unpaginated request is made to fetch all organizations for the user:
```
GET /admin/realms/{realms}/organizations/members/{userId}/organizations
(This returns all 1,000 organizations at once).
```

An N+1 cascade of requests is immediately fired to fetch the members of each organization returned in step 1:
```
GET /admin/realms/{realms}/organizations/{orgId}/members
(This queues 1,000 concurrent requests. The React UI waits for all of these to resolve before rendering the list).
```

**Note:** It is unclear why the UI needs to fetch the /members endpoint for every single organization just to render the User's organization list. If this is a reused table component trying to calculate a member count or determine the user's specific role in the org, it creates a severe performance bottleneck at scale.

### How to Reproduce?

1. Create a realm.
2. Via the Admin REST API, create a large number of Organizations (e.g., 1,000).
3. Via the Admin REST API, add a single User as a member to all 1,000 Organizations.
4. Log into the Keycloak Admin Console.
5. Navigate to Users -> Click on the specific User -> Organizations tab.
6. Observe the delayed render and inspect the Network tab to see the N+1 API calls blocking the UI update.

### Anything else?

_No response_

ssj-ali:
## Problem

The Admin UI **User → Organizations** tab fires an unbounded `GET /admin/realms/{realm}/organizations/members/{userId}/organizations` (no `first`/`max`) that returns **every** organization the user belongs to. For each organization returned, it then fires `GET /admin/realms/{realm}/organizations/{orgId}/members` fetching **all members** just to resolve the current user's `membershipType`. At scale (hundreds of orgs), this creates an N+1 cascade of requests and blocks the UI until every promise resolves.

### Backend (REST + SPI + JPA + Infinispan)
- Add `first`/`max`/`search` query params to the `/members/{member-id}/organizations` endpoint with `@DefaultValue` (`first=0`, `max=10`)
- Add `getByMember(UserModel member, String search, Integer first, Integer max)` to `OrganizationProvider` SPI with a default in-memory fallback
- Override in `JpaOrganizationProvider` with a **CriteriaQuery** using `paginateQuery()` for DB-level `LIMIT`/`OFFSET` — consistent with every other paginated method in the provider
- Delegate through `InfinispanOrganizationProvider` to JPA (cache stores IDs only, cannot filter by name)

### Frontend (Admin Client + Admin UI)
- `memberOrganizations` now accepts `first`/`max`/`search` pagination params
- New `getMember` method for single-member lookup (returns `MemberRepresentation` with `membershipType`)
- Rewrite `Organizations.tsx` to use a paginated `loader` function with `isPaginated`, fetching `membershipType` via targeted `getMember` calls only for the current page

### Performance impact

| | Before | After |
|---|---|---|
| Org fetch | 1 unbounded query (no LIMIT) | 1 query with `LIMIT 10 OFFSET n` |
| MembershipType | N × `/members` (all members per org) | N × `/members/{userId}` (1 member) |
| N = | total org count (unbounded) | page size (default 10) |

Closes #46413


model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
```diff
@@ -483,6 +484,44 @@ public Stream<OrganizationModel> getByMember(UserModel member) {
...
+        throwExceptionIfObjectIsNull(member, "User");
+
+        CriteriaBuilder builder = em.getCriteriaBuilder();
+        CriteriaQuery<OrganizationEntity> queryBuilder = builder.createQuery(OrganizationEntity.class);
+        Root<OrganizationEntity> org = queryBuilder.from(OrganizationEntity.class);
+
+        // subquery: find group IDs the user belongs to
+        Subquery<String> memberGroupIds = queryBuilder.subquery(String.class);
+
+        List<Predicate> predicates = new ArrayList<>();
```
pruivo, line 498: Pre-allocate the list size to avoid growing the array in each `add()`. 
```suggestion
        List<Predicate> predicates = new ArrayList<>(3);
```

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
```diff
@@ -483,6 +484,44 @@ public Stream<OrganizationModel> getByMember(UserModel member) {
...
+        }
+
+        predicates.add(org.get("groupId").in(memberGroupIds));
+        predicates.add(builder.equal(org.get("realmId"), getRealm().getId()));
+
+        if (StringUtil.isNotBlank(search)) {
+            predicates.add(builder.like(builder.lower(org.get("name")), "%" + search.toLowerCase() + "%"));
+        }
+
+        queryBuilder.where(predicates.toArray(new Predicate[0]));
```
pruivo, line 518: avoid copying the predicates again
```suggestion
        queryBuilder.where(predicates);
```

model/jpa/src/main/java/org/keycloak/organization/jpa/JpaOrganizationProvider.java
```diff
@@ -483,6 +484,44 @@ public Stream<OrganizationModel> getByMember(UserModel member) {
...
+        if (StorageId.isLocalStorage(member.getId())) {
+            Root<UserGroupMembershipEntity> membership = memberGroupIds.from(UserGroupMembershipEntity.class);
+            memberGroupIds.select(membership.get("groupId"));
+            memberGroupIds.where(builder.equal(membership.get("user").get("id"), member.getId()));
+        } else {
+            Root<org.keycloak.storage.jpa.entity.FederatedUserGroupMembershipEntity> membership =
+                    memberGroupIds.from(org.keycloak.storage.jpa.entity.FederatedUserGroupMembershipEntity.class);
+            memberGroupIds.select(membership.get("groupId"));
+            memberGroupIds.where(builder.equal(membership.get("userId"), member.getId()));
+        }
```
pruivo, line 509: Isn't it missing a predicate to get the group from `type = 1`?
https://github.com/keycloak/keycloak/blob/111ba365049123ddd3c397b8c12189f61b14a0ef/server-spi/src/main/java/org/keycloak/models/GroupModel.java#L35

pruivo (changes requested):

pedroigor:
Thanks, @ssj-ali. Can you please check https://github.com/keycloak/keycloak/pull/46921?

Out of curiosity, did you create this patch using AI?

PR #46424, opened 2026-02-17
Revisit Infinispan session idle and lifetime settings

### LINKED ISSUE #46421: Revisit Infinispan session idle and lifetime settings
### Describe the bug

(Ticket might later be moved to the Keycloak main repository)

We put user session into the cache with a max lifetime and an idle time. If we retrieve them, IMHO the idle time will be reset, and they will be removed once they are either idle or expired. 

At the same time we're updating the entries whenever they are refreshed. So I wonder that the idle time should always be the same as the expiration time from the perspective of Infinispan. Keycloak will then on the next refresh update the lifetime and the expiry time with a new value.

This would also help us to prevent reading entries from the cache which then turn out to be expired from Keycloak's perspective.

### Version

main

### Expected behavior

Session entries in Infinispan should be removed when they are not refreshed - independent of when and how many times they are read from the cache.

### Actual behavior

There is an idle time set to the refresh interval, and a max lifetime setting set to the maximum lifetime of the session.

### How to Reproduce?

Look at the UI entries in the external Infinispan.

### Anything else?

_No response_

pruivo:
Closes #46421

### DIFF
--- a/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/remote/updater/client/AuthenticatedClientSessionUpdater.java
+++ b/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/remote/updater/client/AuthenticatedClientSessionUpdater.java
@@ -104,7 +104,7 @@ public class AuthenticatedClientSessionUpdater extends BaseUpdater<ClientSession
     public Expiration computeExpiration() {
         long maxIdle = SessionTimeouts.getClientSessionMaxIdleMs(userSession.getRealm(), client, offline, isUserSessionRememberMe(), getTimestamp());
         long lifespan = SessionTimeouts.getClientSessionLifespanMs(userSession.getRealm(), client, offline, isUserSessionRememberMe(), getStarted(), getUserSessionStarted());
-        return new Expiration(maxIdle, lifespan);
+        return new Expiration(SessionTimeouts.IMMORTAL_FLAG, SessionTimeouts.calculateEffectiveSessionLifespan(maxIdle, lifespan));
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/changes/remote/updater/client/AuthenticatedClientSessionUpdater.java
```diff
@@ -104,7 +104,10 @@ public RemoteAuthenticatedClientSessionEntity apply(ClientSessionKey uuid, Remot
     public Expiration computeExpiration() {
         long maxIdle = SessionTimeouts.getClientSessionMaxIdleMs(userSession.getRealm(), client, offline, isUserSessionRememberMe(), getTimestamp());
         long lifespan = SessionTimeouts.getClientSessionLifespanMs(userSession.getRealm(), client, offline, isUserSessionRememberMe(), getStarted(), getUserSessionStarted());
-        return new Expiration(maxIdle, lifespan);
+        if (maxIdle != SessionTimeouts.IMMORTAL_FLAG) {
+            lifespan = Math.min(maxIdle, lifespan);
+        }
```
ahus1: I'd say `maxIdle` will never be importal if you want to optimize that. Still this might be the robust way to do it. I'll leave it up to you to decide. 

For offline sessions, the lifetime might be immportal.

pruivo: Updated.
I added a new method to `SessionTimeouts` with javadoc explaining the reason. Also, if the lifespan is immortal, I'm returning the max-idle value.

ahus1 (approved):

PR #46441, opened 2026-02-18
Migrations in db-compatibility-verifier-maven-plugin

### LINKED ISSUE #46389: Track Migrations in db-compatibility-verifier-maven-plugin
### Description

As well as Liquibase ChangeSets, the db-compatibility-verifier-maven-plugin should track org.keycloak.migration.migrators.Migrations implementations to ensure that programmatic migrations are not accidentally introduced between patch releases.

ryanemerson:
Closes #46389


pruivo:
@ryanemerson should we ignore inner/lambda classes? I've noticed the following after generating a snapshot: 
```
  {
    "class" : "org.keycloak.migration.migrators.MigrateTo2_1_0"
  }, {
    "class" : "org.keycloak.migration.migrators.MigrateTo2_1_0$1"
  }, 
```

misc/db-compatibility-verifier/src/test/java/org/keycloak/db/compatibility/verifier/test/ExampleMigration.java
```diff
@@ -0,0 +1,4 @@
+package org.keycloak.db.compatibility.verifier.test;
```
pruivo, line 1: Is it used anywhere?

ryanemerson: Yes, in `SnapshotMojoTest#testSnapshotFilesCreated` we check that at least one migration class is detected and recorded. I have updated this class to include an inner and anonymous class so that we can verify that they are ignored as expected.

pruivo (approved):

ahus1 (approved):

PR #46446, opened 2026-02-18
Do not forward prompt=create to external identity providers

### LINKED ISSUE #45959: prompt=create Parameter Propagates to Identity Provider Causing 400 Error
### Before reporting an issue

### Area

identity-brokering

### Describe the bug

When using a modified registration template to display Identity Provider login options (tested with Google) and accessing the registration page with `prompt=create`, the prompt parameter is incorrectly propagated to the Identity Provider's authentication URL. This causes a conflict with the IdP's own prompt parameter handling, resulting in a 400 Bad Request error.

### Version

26.5.2

### Regression

### Expected behavior

The Identity Provider authentication URL should not include the `prompt=create` parameter, or Keycloak should properly handle/translate the prompt parameter according to the IdP's specifications.

```
https://accounts.google.com/o/oauth2/v2/auth?client_id=...&redirect_uri=...&response_type=code&scope=...
```

### Actual behavior

The `prompt=create` parameter is passed through to the Identity Provider URL, causing a 400 error.

```
https://accounts.google.com/o/oauth2/v2/auth?prompt=create&client_id=...&redirect_uri=...&response_type=code&scope=...
```

Error: Google returns 400 Bad Request because `prompt=create` is not a valid value for Google's OIDC implementation.

### How to Reproduce?

1. Configure an Identity Provider (Google) in Keycloak realm
2. Modify the registration template to display Identity Provider login options
3. Access the registration page with the prompt=create parameter:
`https://keycloak-url.com/realms/test/protocol/openid-connect/auth?prompt=create&...`
7. Click on the Google Identity Provider login button from the registration page
8. Observe the redirected URL to Google

### Anything else?

Currently, the only workaround is to avoid using the `prompt=create` parameter when displaying Identity Provider options, which limits the ability to direct users specifically to the registration flow.

This issue specifically manifests when customizing the registration template to show IdP options, which is a common use case for improving user experience by allowing social login during registration.

SferaDev:
## Summary

Since `prompt=create` is a Keycloak-internal parameter (per [OpenID Connect Prompt Create 1.0](https://openid.net/specs/openid-connect-prompt-create-1_0.html)) used to initiate registration and is not a valid value for most external IdPs (e.g., Google returns 400 Bad Request), add a way to filter out the `prompt=create` value before forwarding the prompt parameter to external identity providers in the broker authorization URL

Closes #45959

### DIFF
--- a/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
+++ b/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java
@@ -528,7 +528,12 @@ public abstract class AbstractOAuth2IdentityProvider<C extends OAuth2IdentityPro
             prompt = authenticationSession.getClientNote(OAuth2Constants.PROMPT);
         }
         if (prompt != null) {
-            uriBuilder.queryParam(OAuth2Constants.PROMPT, prompt);
+            // "create" is a Keycloak-internal prompt value used to initiate registration
+            // and should not be forwarded to external identity providers.
+            prompt = prompt.replace(OIDCLoginProtocol.PROMPT_VALUE_CREATE, "").trim();
+            if (!prompt.isEmpty()) {
+                uriBuilder.queryParam(OAuth2Constants.PROMPT, prompt);
+            }
         }
 
         if (getConfig().isPkceEnabled()) {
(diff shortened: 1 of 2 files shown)


gbkwiatt:
This is very much needed fix. I am not sure why it's not higher priority since it breaks Google Provider.

ahus1:
@gbkwiatt - please upvote in the parent issue to raise the attention to this

image

ahus1:
Pinging @pedroigor as you marked this help-wanted in the parent issue.

PR #46449, opened 2026-02-18
[OID4VCI]: Normalize hash algorithm names to lowercase

### LINKED ISSUE #45446: [OID4VCI] Default value for vc.credential_build_config.hash_algorithm causes _sd_alg to be invalid due to case sensitivity
### Before reporting an issue

### Area

oid4vc

### Describe the bug

When issuing a credential using Keycloak's OpenID4VCI implementation, the `_sd_alg` claim is set to the name of the hash algorithm that is configured in the `CredentialScopeModel` (configurable via `vc.credential_build_config.hash_algorithm`). 

However, the current default for this value (`SHA-256`, defined in `CredentialScopeModel.HASH_ALGORITHM_DEFAULT`) is not a valid value for `_sd_alg` because the specification [considers the `_sd_alg` value to be case sensitive](https://datatracker.ietf.org/doc/html/rfc9901#section-4.1.1) and the actual registered algorithm name is `sha-256`.
To prevent verification errors with some wallet implementations, the default value must either be adjusted or the algorithm name must be converted to lowercase before being written to the `_sd_alg` claim.

### Version

26.5.0

### Regression

### Expected behavior

Using a credential scope without an explicitly defined `vc.credential_build_config.hash_algorithm`, the `_sd_alg` claim is set to a valid hash algorithm name defined in https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg.

### Actual behavior

The default value is set to `SHA-256`, which is considered invalid by some wallet implementations because `_sd_alg` is case sensitive and the correct value would be `sha-256`.

### How to Reproduce?

Create a credential scope as described above and request a credential.

### Anything else?

_No response_

Awambeng:
Ensure hash algorithm names are lowercase across multiple files to comply with IANA standards. Updated relevant code in IssuerSignedJWT, SdJwt, and CredentialScopeModel, and adjusted tests accordingly.

Closes #45446

### DIFF
--- a/core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
+++ b/core/src/main/java/org/keycloak/sdjwt/IssuerSignedJWT.java
@@ -170,7 +170,8 @@ public class IssuerSignedJWT extends JwsToken {
         }
         if (sdArray.size() > 0 || nestedDisclosures) {
             // add sd alg only if ay disclosure.
-            payload.put(CLAIM_NAME_SD_HASH_ALGORITHM, hashAlg);
+            // Normalize to lowercase to comply with IANA registered hash algorithm names
+            payload.put(CLAIM_NAME_SD_HASH_ALGORITHM, hashAlg.toLowerCase());
         }
 
         // then put all other claims in the paypload
--- a/core/src/main/java/org/keycloak/sdjwt/SdJwt.java
+++ b/core/src/main/java/org/keycloak/sdjwt/SdJwt.java
@@ -269,8 +269,9 @@ public class SdJwt {
             Optional.ofNullable(keyBindingJWT).ifPresent(keyBindJwt -> {
                 // get the hash-algorithm to use for keyBinding and set it if not present
                 String hashAlgorithm = getEffectiveHashAlgorithm(sdHashAlgorithm);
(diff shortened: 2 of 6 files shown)


tnorimat (approved):
Confirmed that the chosen lowercase strings actually match the relevant specifications used in OID4VC / SD-JWT contexts.

VinodAnandan:
@Awambeng Could you please use "[OID4VCI]" prefix for the title ?

Awambeng:
> @Awambeng Could you please use "[OID4VCI]" prefix for the title ?

Thanks for pointing that out, @VinodAnandan. I’ve updated the title to include the “[OID4VCI]” prefix.

mposolda (commented):
@Awambeng Thanks for the PR. Looks good, but it seems the test failure in admin UI tests is the regression of this PR. I think test should be updated accordingly as well?

Awambeng:
> @Awambeng Thanks for the PR. Looks good, but it seems the test failure in admin UI tests is the regression of this PR. I think test should be updated accordingly as well?

@mposolda Thank you! The failing tests have been resolved. Could you please take a look when you have a chance?

mposolda (approved):
@Awambeng Thanks!

PR #46450, opened 2026-02-18
[OID4VCI] Generate pre-authorized codes using the JWT format

IngridPuppet:
Closes #45231

The implemented JWT format looks like this (updated with review comments):

```json
{
  "header" : {
    "alg" : "HS512",
    "typ" : "JWT",
    "kid" : "a109ad9d-0144-40c9-80e7-42dc7e55b2e6"
  },
  "payload" : {
    "exp" : 1774447450,
    "iat" : 1774447420,
    "iss" : "http://localhost:8080/realms/test",
    "aud" : "http://localhost:8080/realms/test/protocol/openid-connect/token",
    "context" : {
      "credentials_offer_id" : "jPYrg61ZmpG0Dwb7qSn3B5AmiyJuScjtflP001s-jZlacJxhnnx5p8zwAni4GJUQ-pWO_EhZZrzOx1cugo5zjw",
      "target_client_id" : "oid4vci-client",
      "target_user_id" : "0599083e-fb4b-4e9f-8e05-685adfab65a5",
      "nonce" : "OCYXrKWen1TFaCwXvx5RYYwWdEfGbYF1fqlsqnwoO3FCHG5SL-s2fbvTScNLwKtgKvVj8E2tSEUIXDo4Z3dXdQ",
      "authorization_details" : [ {
        "type" : "openid_credential",
        "credential_configuration_id" : "jwt-credential-config-id",
        "credential_identifiers" : [ "jwt-credential" ]
      } ]
    },
    "salt" : "fVLiYLHQ6ALVRF2V2KXyZ5AVh284h2k1TZMBO6k6fF0"
  },
  "signature" : "oLYISMsP4ZlwfDNl8joXQ58kZPVL8yECbAOa3AX_vPCjIpLD9XUKFvJgbUo34CG1h56YMFIOj9qN48_iZit-1A"
}
```

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
@@ -1,0 +1,259 @@
+package org.keycloak.protocol.oid4vc.issuance.credentialoffer.preauth;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.TokenVerifier;
+import org.keycloak.common.VerificationException;
+import org.keycloak.common.util.Base64Url;
+import org.keycloak.common.util.SecretGenerator;
+import org.keycloak.common.util.Time;
+import org.keycloak.crypto.Algorithm;
+import org.keycloak.crypto.KeyUse;
+import org.keycloak.crypto.KeyWrapper;
+import org.keycloak.crypto.SignatureProvider;
+import org.keycloak.crypto.SignatureSignerContext;
+import org.keycloak.crypto.SignatureVerifierContext;
+import org.keycloak.events.Errors;
+import org.keycloak.jose.jws.JWSBuilder;
+import org.keycloak.jose.jws.JWSHeader;
+import org.keycloak.jose.jws.JWSInput;
+import org.keycloak.jose.jws.JWSInputException;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.RealmModel;
+import org.keycloak.models.oid4vci.CredentialScopeModel;
+import org.keycloak.protocol.oid4vc.issuance.OID4VCIssuerWellKnownProvider;
+import org.keycloak.protocol.oid4vc.model.JwtPreAuthCode;
+import org.keycloak.protocol.oid4vc.model.PreAuthCodeCtx;
+import org.keycloak.protocol.oidc.OIDCWellKnownProviderFactory;
+import org.keycloak.protocol.oidc.representations.OIDCConfigurationRepresentation;
+import org.keycloak.utils.StringUtil;
+import org.keycloak.wellknown.WellKnownProvider;
+
+import org.jboss.logging.Logger;
+
+import static org.keycloak.constants.OID4VCIConstants.OID4VC_PROTOCOL;
+
+/**
+ * Implementation of {@link PreAuthCodeHandler} for JWT pre-authorized codes.
+ */
+public class JwtPreAuthCodeHandler implements PreAuthCodeHandler {
+
+    private static final Logger logger = Logger.getLogger(JwtPreAuthCodeHandler.class);
+
+    private final KeycloakSession session;
+    private final RealmModel realm;
+
+    public JwtPreAuthCodeHandler(KeycloakSession session) {
+        this.session = session;
+        this.realm = session.getContext().getRealm();
+    }
+
+    @Override
+    public String createPreAuthCode(PreAuthCodeCtx preAuthCodeCtx) {
+        // Clone and nullify exp claim to avoid duplication
+        PreAuthCodeCtx clonedCtx = preAuthCodeCtx.clone();
+        clonedCtx.setExpiresAt(null);
+
+        // Building
+        String salt = Base64Url.encode(SecretGenerator.getInstance().randomBytes());
+        JwtPreAuthCode jwtBody = (JwtPreAuthCode) new JwtPreAuthCode()
+                .salt(salt)
+                .context(clonedCtx)
+                .issuer(getCredentialIssuer())
+                .addAudience(getTokenEndpoint())
+                .issuedNow()
+                .exp(preAuthCodeCtx.getExpiresAt());
+
+        // Signing
+        SignatureSignerContext signer = getSignerContext(preAuthCodeCtx);
+        return new JWSBuilder()
+                .type(OAuth2Constants.JWT)
+                .jsonContent(jwtBody)
+                .sign(signer);
+    }
+
+    @Override
+    public PreAuthCodeCtx verifyPreAuthCode(String preAuthCode) throws VerificationException {
+        // Parse the JWT
+        JWSInput jwsInput;
+        try {
+            jwsInput = new JWSInput(preAuthCode);
+        } catch (JWSInputException e) {
+            throw new VerificationException("Failed to parse pre-auth code: Not JWT", e);
+        }
+
+        // Verify JWT (Signature + claim conformance)
+        TokenVerifier<JwtPreAuthCode> verifier = TokenVerifier.create(preAuthCode, JwtPreAuthCode.class);
+        verifier.verifierContext(getVerifierContext(jwsInput.getHeader()));
+        getPropertyVerifiers().forEach(verifier::withChecks);
+        verifier.verify();
+
+        // Parse payload as JwtPreAuthCode and extract PreAuthCodeCtx
+        try {
+            JwtPreAuthCode jwtBody = jwsInput.readJsonContent(JwtPreAuthCode.class);
+            PreAuthCodeCtx preAuthCodeCtx = jwtBody.getContext();
+            preAuthCodeCtx.setExpiresAt(jwtBody.getExp());
+            return preAuthCodeCtx;
+        } catch (JWSInputException e) {
+            throw new VerificationException("Failed to recover credential offer state", e);
+        }
+    }
+
+    private String getCredentialIssuer() {
+        return OID4VCIssuerWellKnownProvider.getIssuer(session.getContext());
+    }
+
+    private String getTokenEndpoint() {
+        WellKnownProvider oidcProvider = session.getProvider(
+                WellKnownProvider.class, OIDCWellKnownProviderFactory.PROVIDER_ID);
+        OIDCConfigurationRepresentation oidcConfig = (OIDCConfigurationRepresentation) oidcProvider.getConfig();
+        return oidcConfig.getTokenEndpoint();
+    }
+
+    /**
+     * Retrieves the preferred signing algorithms for JWT pre-auth codes,
+     * based on the configuration of credentials under consideration.
+     */
+    private List<String> getPreferredSigningAlgs(PreAuthCodeCtx preAuthCodeCtx) {
+        List<String> configIds = Optional.ofNullable(preAuthCodeCtx)
+                .map(PreAuthCodeCtx::getCredentialConfigurationIds)
+                .orElse(List.of());
+
+        Stream<CredentialScopeModel> credentialScopes = session.clientScopes()
+                .getClientScopesByProtocol(realm, OID4VC_PROTOCOL)
+                .map(CredentialScopeModel::new)
+                .filter(s -> configIds.contains(
+                        s.getCredentialConfigurationId()
+                ));
+
+        return credentialScopes.map(CredentialScopeModel::getSigningAlg)
+                .filter(StringUtil::isNotBlank)
+                .distinct()
+                .toList();
+    }
+
+    /**
+     * Retrieves a SignatureSignerContext for signing JWT pre-auth codes, based on the preferred
+     * signing algorithms derived from the pre-auth code context.
+     */
+    private SignatureSignerContext getSignerContext(PreAuthCodeCtx preAuthCodeCtx) {
+        List<String> preferredAlgs = getPreferredSigningAlgs(preAuthCodeCtx);
+        logger.debugf("Preferred signing algorithms for JWT pre-auth code: %s", preferredAlgs);
+        for (String alg : preferredAlgs) {
+            try {
+                return getSignerContext(alg);
+            } catch (RuntimeException ignored) {
+                logger.debugf("No active signing key/context found for algorithm %s, skipping", alg);
+            }
+        }
+
+        try {
+            // Default to Algorithm.ES256 if no preferred algorithm is found or signing key available.
+            // Under normal circumstances, an ES256 key should be generated on the fly.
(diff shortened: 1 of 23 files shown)


VinodAnandan:
@IngridPuppet Thank you very much for this PR, it’s greatly appreciated.

I had a quick question regarding the use of "typ": "oid4vci-pre-auth-code+jwt" instead of "typ": "JWT". Could you please share the rationale behind this choice? Additionally, are there any IANA registrations, standards, or other authoritative references that define or recommend this type or format?

If there isn’t a clear standard definition or established best practice supporting it, I would kindly ask that we align this with the existing standards. Thank you very much for your consideration.

IngridPuppet:
> If there isn’t a clear standard definition or established best practice supporting it, I would kindly ask that we align this with the existing standards. Thank you very much for your consideration.

Hi @VinodAnandan. I understand your point and acknowledge that the `typ` header introduced is not normative. Let me drop that quick and simply use `JWT`. Hopefully, it has no impact on the business logic.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
```diff
@@ -0,0 +1,252 @@
...
+import org.keycloak.protocol.oid4vc.issuance.OID4VCIssuerWellKnownProvider;
+import org.keycloak.protocol.oid4vc.issuance.credentialoffer.CredentialOfferStorage.CredentialOfferState;
+import org.keycloak.protocol.oid4vc.model.CredentialsOffer;
+import org.keycloak.protocol.oid4vc.model.JwtPreAuthCode;
+import org.keycloak.protocol.oidc.OIDCWellKnownProviderFactory;
+import org.keycloak.protocol.oidc.representations.OIDCConfigurationRepresentation;
+import org.keycloak.utils.StringUtil;
+import org.keycloak.wellknown.WellKnownProvider;
+
+import org.slf4j.Logger;
```
tnorimat: The Keycloak codebase consistently uses JBoss Logging (org.jboss.logging.Logger), as seen throughout OID4VCIssuerEndpoint, PreAuthorizedCodeGrantType, etc. This class uses SLF4J instead, which breaks the project convention.
How about the following?:
```
private static final Logger logger = Logger.getLogger(JwtPreAuthCodeHandler.class);
```

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
```diff
@@ -0,0 +1,252 @@
...
+            try {
+                return getSignerContext(alg);
+            } catch (RuntimeException ignored) {
+                logger.debug("No active signing key/context found for algorithm {}, skipping", alg);
+            }
+        }
+
+        // Default to Algorithm.ES256 if no preferred algorithm is found or signing key available
+        logger.debug("Falling back to default algorithm {} for signing JWT pre-auth codes", Algorithm.ES256);
+        return getSignerContext(Algorithm.ES256);
```
tnorimat: I am afraid that If the realm has no ES256 key either, the fallback call will throw an unhandled IllegalArgumentException from getSignerContext(String).

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
```diff
@@ -0,0 +1,252 @@
...
+    public String createPreAuthCode(CredentialOfferState offerState) {
+        // Building
+        String salt = Base64Url.encode(SecretGenerator.getInstance().randomBytes());
+        JwtPreAuthCode jwtBody = (JwtPreAuthCode) new JwtPreAuthCode()
+                .salt(salt)
+                .credentialOfferState(offerState)
+                .issuer(getCredentialIssuer())
+                .addAudience(getTokenEndpoint())
+                .issuedNow()
+                .exp((long) offerState.getExpiration());
```
tnorimat: Just a comment.
JsonWebToken.exp() takes Long while CredentialOfferState.getExpiration() returns int. 
If considering that int epoch seconds will overflow in 2038, it might be good that CredentialOfferState.getExpiration() returns long.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
IngridPuppet: Hello @tnorimat - Thank you for your review. Yes, you're right. It was a copy/paste inattention. I updated the `oid4vc` package to stay consistent with the JBoss interface.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
IngridPuppet: I added an explicit catch block to handle the exception properly. Normally, I believe one key should be generated on the fly for ES256. That's now asserted in `PreAuthorizedGrantTest`.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
IngridPuppet: I agree. I updated the type chaining from int to long.

VinodAnandan:
@IngridPuppet Could you please update the JWT format in the PR description with the latest changes.

IngridPuppet:
> @IngridPuppet Could you please update the JWT format in the PR description with the latest changes.

Sure. Done.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
tnorimat: I confirm the point. Thank you.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
tnorimat: Thank you.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
tnorimat: Thank you.

tnorimat (approved):
Thanks!

tdiesler:
It's probably not a good idea to serialize the CredentialOfferState directly into the JWT

* credential_issuer is duplicated in `iss`
* expiration is duplicated in `exp`

What you show above has no `authorization_details`. Maybe that is included now (by virtue of serializing the entire state), is it? The `tx_code` is part of that credential order state, it should however not be included in the JWT

tdiesler:
This branch is actually quite a bit behind current main. The CredentialOfferState is now an immutable object that contains `authorization_details` and `tx_code`

IngridPuppet:
> This branch is actually quite a bit behind current main. The CredentialOfferState is now an immutable object that contains `authorization_details` and `tx_code`

Hello @tdiesler - Thank you for taking a look. Indeed, the PR needs updating after the recent merge of #46199. I'll do so taking your points into consideration.

> What you show above has no `authorization_details`. Maybe that is included now (by virtue of serializing the entire state), is it? The `tx_code` is part of that credential order state, it should however not be included in the JWT

Currently, the state is not directly serialized. The model is reused but only public fields are injected into the JWT. Of course, this does not include a potentially sensitive `tx_code` field; I had that in mind. It also does not seem like `authorization_details` can be available at that point in the lifecycle of the Pre-Authorized flow. They were not excluded however.

Given these concerns, I acknowledge that introducing a separate model looks more appropriate. I'll update the PR accordingly.

tdiesler:
Perhaps it's worth building your thing on top of #46466 - there we introduce a `CredentialOfferProvider` as the single entry point for creating CredentialOfferState. You'll find that `authorization_details` are in fact available from create time & can simply be copied to their relevant targets (i.e. AccessToken Response/JWT and your pre-auth_code JWT).

IngridPuppet:
Hello @tdiesler - The PR has been rebuilt on top of the recent changes in `main`. Please could you take another look when you have a chance?

tdiesler (approved):
LGTM - very nice :-)

mposolda (commented):
@IngridPuppet Thanks for the PR!

I did not yet reviewed, but is it possible to rebase please? There is this change https://github.com/keycloak/keycloak/commit/53e7bdf1febf855dcd1a25d304d8435021c3de1d , which may require some additional updates in your PR (the new feature feature would be needed to be enabled in the migrated test).

IngridPuppet:
Hello @mposolda - PR rebased. Could you take a look when you have a chance?

VinodAnandan:
@mposolda This PR aims to enhance the security of the pre-authorized code grant. As you may know, this flow does not benefit from protections using PKCE, which can make it more susceptible to risks like phishing, brute force attacks, malware, accidental exposure, replay attacks, and MITM interception. Using a signed JWT instead of a random string helps strengthen the overall security model.

The PR has already been reviewed and approved by @tnorimat and @tdiesler . We are hoping to include it in the 26.6 release, and would greatly appreciate your support in reviewing and progressing it when you have a moment.

VinodAnandan (approved):
LGTM

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
```diff
@@ -263,4 +277,52 @@ public boolean isTokenAllowed(KeycloakSession session, AccessToken token) {
...
+                    errorType.equals(Errors.INVALID_CODE)
+                            ? OAuthErrorException.INVALID_REQUEST
+                            : OAuthErrorException.EXPIRED_TOKEN,
+                    errorMessage, Response.Status.BAD_REQUEST);
+        }
+
+        // Pre-auth code is valid, but let's prevent replay attacks
+        SingleUseObjectProvider singleUseStore = session.singleUseObjects();
+        String key = getPreAuthCodeSingleObjectKey(code);
+        if (singleUseStore.get(key) != null) {
```
mposolda: It would be better to use `singleUseStore.putIfAbsent` instead of using "get" with subsequent "put" . The `putIfAbsent` is atomic and hence handle gracefully the cases when there is attempt to replay with concurrent requests (we had some security issues with that before).

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
```diff
@@ -0,0 +1,259 @@
...
+        JwtPreAuthCode jwtBody = (JwtPreAuthCode) new JwtPreAuthCode()
+                .salt(salt)
+                .context(clonedCtx)
+                .issuer(getCredentialIssuer())
+                .addAudience(getTokenEndpoint())
+                .issuedNow()
+                .exp(preAuthCodeCtx.getExpiresAt());
+
+        // Signing
+        SignatureSignerContext signer = getSignerContext(preAuthCodeCtx);
```
mposolda: The pre-authorized code JWT is signed by Keycloak and will be verified by Keycloak. There is no other party, which needs to be able to verify it's signature.

Considering this, the JWT can be signed by the realm symmetric key (`HS512`), which is more performant as well as secured. The same is used for example for signing/verifying OIDC refresh tokens during regular OIDC flows.

The `JwtPreAuthCode` already extends from `JsonWebToken`, which implements `getCategory()` method to return `INTERNAL`. Hence it is considered as "internal" token already. So regarding this, the only thing needed to sign the token might be probably something like this:
```
String encodedJwt = session.tokens().encode(cookie);
```

And something similar for decode/verify the token (as method `decode` of `TokenManager` can be used to verify the signature of the tokens).

This will simplify the code of this class a lot as there will not be a need to handle signing/verifying explicitly here, but would be just possible to re-use existing

mposolda (commented):
@IngridPuppet Thanks for this PR and sorry for late review. Added some comments inline. Can you please check?

mposolda:
@IngridPuppet Will be probably good to rebase on top of latest main after recently merged https://github.com/keycloak/keycloak/commit/931d232fa2a16e1d8521619b25c9b3e8fe6a5013 (There are no conflicts with your PR, but just to be sure that tests are still OK). Thanks!

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
VinodAnandan: @mposolda , in enterprise environments it’s quite common to perform early validation and rejection of JWTs, particularly expired or otherwise invalid tokens as part of a defence-in-depth strategy.
I also noticed that the examples provided in the issue https://github.com/keycloak/keycloak/issues/45231 use ES256. Would it make sense to make the JWT algorithm configurable, perhaps using HS512 as the default? This could allow flexibility for teams that may have requirements for other algorithms, while still providing a sensible default configuration.

mposolda: Just to summary what we discussed on the meeting: It can be good to initially just support the "internal" token and hence `HS512` (ideally implementing just by `session.tokens().encode(token)` ). We can consider configurability as a follow-up (however not sure if it is needed for something as other parties besides Keycloak should treat pre-authorized-code as opaque token though).

services/src/main/java/org/keycloak/protocol/oidc/grants/PreAuthorizedCodeGrantType.java
IngridPuppet: Thank you for your review, @mposolda. I understand the issue and updated the code to use `putIfAbsent`.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/credentialoffer/preauth/JwtPreAuthCodeHandler.java
IngridPuppet: The class now uses `session.tokens()` for JWT creation and verification. Much simpler indeed.

mposolda (approved):
@IngridPuppet @tdiesler @tnorimat @VinodAnandan Thanks for the update and reviews!

PR #46453, opened 2026-02-18
fix: using the openshift version of opm

shawkins:
closes: #46451

This switches us to using the openshift mirror, rather than github, for opm. That seems to address some of the recent instability.

This also moves the version up a couple of years, but we can't go further because the installer is out-of-date.


mabartos (approved):
LGTM

vmuzikar (approved):

PR #46457, opened 2026-02-19
add managed identity for all db except azure sql and sql managed instance and oracle

gabo89:
adding managed identity support for 

- azure flexible server postgresql (tested on 17.7, 16.11)
- azure mysql flexible server 8.0 8.4 (in both can connect but liquibase fails becuase some tables have multiple primary keys or somethign along those lines)
- azure mariadb flexible server ( havent tested yet  ,probably also work but need to check tomorrow)
- azure sql and sql managed instance is skipped since the way access token is used is different
- aws rds posgresql 17.6 (proably can work on other version but havent tested, will try tomorrow with 1 more)
- aws rds aurora postgresql 17.4 (probably can work on other versino but havent tested, will try tomorrow with 1 more)
- pending to test on aws rds mariadb and aws rds mysl 
- oracle db is not included since work different 

added step on here 
https://github.com/gabo89/keycloak-validation-yaml

reason of pull request
- avoid the use of credential on the application itself to connect to database

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/IdentityOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/IdentityOptions.java
@@ -1,0 +1,30 @@
+package org.keycloak.config;
+
+public class IdentityOptions {
+  public static final Option<Boolean> USE_IDENTITY = new OptionBuilder<>("use-identity", Boolean.class)
+    .category(OptionCategory.DATABASE_IDENTITY)
+    .defaultValue(false)
+    .description("boolean value to define is managed identity is used.")
+    .build();
+  public static final Option<String> CLOUD_PROVIDER = new OptionBuilder<>("cloud-provider", String.class)
+    .category(OptionCategory.DATABASE_IDENTITY)
+    .defaultValue("")
+    .description("boolean value to define is managed identity is used.")
+    .build();
+
+  public static final Option<String> AWS_REGION = new OptionBuilder<>("aws-region", String.class)
+    .category(OptionCategory.DATABASE_IDENTITY)
+    .defaultValue("")
+    .description("boolean value to define is managed identity is used.")
+    .build();
+  public static final Option<String> AWS_PORT = new OptionBuilder<>("aws-port", String.class)
+    .category(OptionCategory.DATABASE_IDENTITY)
+    .defaultValue("0")
+    .description("boolean value to define is managed identity is used.")
+    .build();
+  public static final Option<String> AWS_HOSTNAME = new OptionBuilder<>("aws-hostname", String.class)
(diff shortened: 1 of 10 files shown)


vmuzikar:
@gabo89 Thank you for your contribution!

The PR is missing a linked issue and some tests (which might be something a bit more involved in this case and might require some help from the Keycloak team) as per the [contributing guidelines](https://github.com/keycloak/keycloak/blob/main/CONTRIBUTING.md).

But before proceeding further, I'd like to first understand better and discuss the use case this PR tries to solve. If I understand correctly, you still need to provide credentials to Azure/AWS to Keycloak, correct? Another point is that Keycloak aims to be platform agnostic when it comes to DBs – i.e. support the "base DBs" (like Postgres, MySQL, etc.) without additional features specific to the "enterprise DB vendors" like Azure/AWS (which are supported and tested but mostly only in the scope of the "base DBs").

@ahus1 I'd love to hear your take on this as well as you and your team are involved in DBs.

CC @keycloak/cloud-native

gabo89:
hello, the idea is to avoid using password when configure  the database servers , i been using keycloak for a while on my project and i didnt want to takethe time that i take to rotate password every x month so i was thinking it was a good idea to add this feature, there isnt a issue created yet about this but i saw some people on the issues part that they were asking about managed identity support for keycloak  , hm i understand, the compatiblity with non cloud db are still preserved  if the flag KC_USE_IDENTITY  or kc.use_identity is on false, the trick is that i just intercept the part before keycloak send the pssword to quarkus agroal datasource  and also make a timer to regenerate a new access token so the agroal datasource always have connectivy to their db , yes i saw that compatibility issues when i tried a local mysql vs the cloud native mysql when i was testing why it wasnt working on the cloud mysql db , i also search and found out about the first class db in keycloak https://www.keycloak.org/2022/02/dbs

does posgresql is still the main db that should  always support , correct ? and then the second type alternatives would be mysl, maridb, oracle or now all db should have the same importance?

gabo89:
added unit test 
- validate tokenprovider return a token
- validate tokenmanager return a token
- validate tokenmanager cache tokens
- validate token manager return a new token after the previous token expiration is near

gabo89:
validated against 8.0.mysql_aurora.3.10.3  8.0.mysql_aurora.3.11.0 using rds permision with the cluster resource-id (jdcb wont work if you point to resource-id of writer or reader)

gabo89:
validated agaisnt mariadb aws rds 11.8.5 11.4.3

shawkins:
@gabo89 thank you for the PR, but I'm going to close it for now to put the focus on creating an issue and having the discussion about this feature / enhancement there. In particular we need to agree on the introduction of additional dependencies - and if this is something that should be supported directly in quarkus.

PR #46465, opened 2026-02-19
Check client is enabled in the docker endpoint

### LINKED ISSUE #46462: CVE-2026-2733 Missing Check on Disabled Client for Docker Registry Protocol
### Description

Improper Authorization vulnerability in the Docker v2 authentication endpoint (/protocol/docker-v2/auth) of Keycloak. The issue occurs because the server fails to validate the client’s “Enabled” status before minting an access token. Even after the client is administratively disabled, the endpoint continues to issue valid authentication tokens when provided with valid user credentials and client ID. This allows authenticated users to bypass intended administrative restrictions, potentially enabling unauthorized access to container registry resources associated with the disabled client.

rmartinc:
Closes #46462

Just adding the check to ensure the client is enabled for docker. Test modified to cover this new case.

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/docker/DockerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/docker/DockerEndpoint.java
@@ -5,6 +5,7 @@ import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 
 import org.keycloak.common.Profile;
+import org.keycloak.events.Errors;
 import org.keycloak.events.EventBuilder;
 import org.keycloak.events.EventType;
 import org.keycloak.models.AuthenticationFlowModel;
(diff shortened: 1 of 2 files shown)


ahus1 (approved):

ahus1:
@rmartinc - thank you for this PR. I've approved and merged it as it resolves the case hat hand. 

Some follow-up items I found when looking at the code, please consider for a follow-up improvement: 

* Instead of logging errors, I prefer to add those detail messages to the event. We already already do that for OIDC in several places, and it allows those information to be shown in the events admin console where it is more accessible for people who have access to the UI, but not to the log file. This would then replace the `logger.errorv()`, and can be shorter as the event context AFAIK already contains the realm. 
* It seems to me that the client is not set in the event context after it has been looked up. 

Best,
Alexander

PR #46472, opened 2026-02-19
Client API v2: automatically generate OpenAPI document used to generate JS client

### LINKED ISSUE #46388: Automatic sync of OpenAPI spec changes to Client v2 admin client
### Description

This is a follow-up from https://github.com/keycloak/keycloak/pull/45460. Currently, we build by default the TS REST Client for Client v2 using a hardcoded OpenAPI spec file. We need to either find a way around the circular dependency (the spec is built with the server, but server build requires the REST Client), or introduce some logic for automatic syncing of the spec file with the actual spec as generated by the server.

michalvavrik:
* Closes: https://github.com/keycloak/keycloak/issues/46388
* Idea is that manually maintaining OpenAPI document hardcoded in JS client is a pain and we would have hard time to ensure consistency, so use the SmallRye OpenAPI Maven plugin to update the file whenever is needed
* idea here is that openapi document generation should be deterministic (I tried it couple of times and seems same, if we run into a problems we will report it and fix it upstream), so the document should only change if API changes, in which case whoever changes api will commit the updated document

### DIFF
--- a/rest/admin-v2/rest/pom.xml
+++ b/rest/admin-v2/rest/pom.xml
@@ -17,6 +17,8 @@
         <maven.compiler.target>17</maven.compiler.target>
         <maven.compiler.release>17</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <!-- Use OpenAPI 4.x compatible plugin version -->
+        <smallrye.openapi.generator.plugin.version>4.2.4</smallrye.openapi.generator.plugin.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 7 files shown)


edewit (commented):
then shouldn't we remove the `openapi.yaml` from this PR then?

michalvavrik:
> then shouldn't we remove the `openapi.yaml` from this PR then?

one of my versions had it, but I run into 2 issues:

- If you put the generated OpenAPI document into the `target`, then the Maven clean plugin must ignore it, due to the fact that the `openapi.yaml` is generated before the `mvn clean` runs in JS module. So it is never clear and in fact, it is just another file that lives locally in your project forever. So my thinking was that we can use the `openapi.yaml` as a hook, if it was updated, we know that we need to consider updating tests.
- many JS CI tests simply do not build Keycloak, they just run it against nightly image and that is it, so having `openapi.yaml` that is up to date allows users to just come and develop / test JS, rather than have prerequisites like building Keycloak

It is not big issue for me to go back to that version if that is firm request, but expect that it will add more changes in CI yamls and noone from UI can develop on admin locally unless they actually build Keycloak.

edewit (approved):
👍

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
```diff
@@ -32,6 +35,12 @@ public interface ClientApi {
     @PATCH
     @Consumes(PatchTypeNames.JSON_MERGE)
     @Produces(MediaType.APPLICATION_JSON)
+    @RequestBody(
+        content = @Content(
+            mediaType = PatchTypeNames.JSON_MERGE,
+            schema = @Schema(implementation = BaseClientRepresentation.class)
```
vmuzikar: Based on the recent discussion with @shawkins, I believe we need to keep the generic JSON object as the request body to correctly handle null values in the generated REST API clients. Correct @shawkins?

rest/admin-v2/rest/pom.xml
```diff
@@ -40,5 +42,85 @@
...
+                            <resources>
+                                <resource>
+                                    <directory>${project.build.directory}</directory>
+                                    <includes>
+                                        <include>openapi.json</include>
+                                        <include>openapi.yaml</include>
+                                    </includes>
+                                </resource>
+                            </resources>
+                        </configuration>
```
vmuzikar, line 102: I assume this ensures that the spec is served live from the Quarkus extension statically, right? It'd be good to add a comment like that there.

One downside of this is, any extensions adding new endpoints won't be reflected by the live served spec. But that is fine for now...

vmuzikar (commented):
@michalvavrik Thank you for the PR!

rest/admin-v2/rest/pom.xml
michalvavrik: > I assume this ensures that the spec is served live from the Quarkus extension statically, right? It'd be good to add a comment like that there.

yes; ok

> One downside of this is, any extensions adding new endpoints won't be reflected by the live served spec. But that is fine for now...

this keeps the same behavior because even before, there were restrictions on scanned packages; but yes, if that was a requirement, it wouldn't work OOTB, we would need to merge it

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
michalvavrik: I'd assume that is only problem for java, not TS / Go and we decided to not generate Java client?

michalvavrik: Because we should signal what we expect to receive, right now it is `ArrayBuffer` which is not helpful for users, they don't know what to send

michalvavrik: I'll revert this change as discussed.

michalvavrik: Done. In a way, once I reverted it, I realized why I started looking at it at the first place. There is issue with having `JsonNode` as SR OpenAPI generates for it array object, which is wrong as we would get 415. So I worked around it. You will see.

rest/admin-v2/rest/pom.xml
michalvavrik: Done.

rest/admin-v2/rest/src/main/java/org/keycloak/admin/api/client/ClientApi.java
vmuzikar: > as we would get 415

@michalvavrik Can you please elaborate more on the issue? The TypeScript generator did not work with that? What do you mean by 415?

michalvavrik: > > as we would get 415
> 
> @michalvavrik Can you please elaborate more on the issue? The TypeScript generator did not work with that? What do you mean by 415?

I think this is new because previously there were hardcoded `{}` request object with no scheme, now that json merge patch is using `JsonNode`, SmallRye OpenAPI generates `array`, however that does not fit specs, we expect an object with literally any properties. I discussed it with Mike here https://github.com/smallrye/smallrye-open-api/issues/2494 and he says it is not a bug. So what we can do, if we don't want to signal request body schema, we must unset it. If you remove my changes in `OASFIlter` (setting scheme to `null`) then TS tests were failing over patch operation because it returned 415 status code.

I don't think it is a bug per say.

vmuzikar: @michalvavrik Thanks for the details. Looks good to me then.

@shawkins Does that sound good to you too?

vmuzikar (dismissed):

michalvavrik:
I have only resolved merge conflicts, rebased on the current main and pushed it.

vmuzikar (approved):

PR #46475, opened 2026-02-19
Fix select menu width to match trigger

### LINKED ISSUE #45947: Selecting condition type when creating a client policy is too wide
### Before reporting an issue

### Area

core

### Describe the bug

The drop down extends over the right end of the screen: 

Image

### Version

main

### Regression

### Expected behavior

The drop-down's width should end before the screen ends

### Actual behavior

It extends - possibly its width is set to the full screen width, but it should be the parent's width. 

### How to Reproduce?

* Create a client policy
* Add a condition
* Use the drop down condition type

### Anything else?

_No response_

gaoyikeshuer:
Adds width="trigger" to Select component in NewClientPolicyCondition.

Closes #45947


ahus1 (approved):
Thank you for providing this fix, it works great!

PR #46478, opened 2026-02-19
[OID4VCI] Make sure persistent userSession not needed in preAuthorized code gra…

### LINKED ISSUE #44534: Make sure persistent userSession not needed in preAuthorized code grant type
### Description

This is follow-up on https://github.com/keycloak/keycloak/pull/44255/files#r2567435868 . It seems userSession may not be needed to be created in the pre-authorized code grant type.

NOTE: It may probably help to make sure that authorization_details are stored as a claim in the access-token instead of storing them as a note in userSession. Related issue: https://github.com/keycloak/keycloak/issues/44961

mposolda:
…nt type

closes #44534

- Make sure there is no userSession created in pre-authorized code flow. This grant doesn't need to refresh token and hence does not need to save persistent session. There is only "transient" (request-scoped) session like in other similar grants (EG. OAuth2 client credentials grant)

- Make sure that credential-endpoint is called only with proper access-token, which was returned from pre-authorized grant OR from authorization-code grant, which requested the credential issuance. Before this PR, it was still possible to call the credential-endpoint with incorrect access token (corresponding test `OID4VCAuthorizationDetailsFlowTestBase.testCompleteFlowWithClaimsValidation()` updated to reflect that)

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
+++ b/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
@@ -815,23 +815,20 @@ public class OID4VCIssuerEndpoint {
 
         // Get the credential_configuration_id from the offer state authorization details
         authDetails = offerState.getAuthorizationDetails();
-
-        // Validate authorization_details: either in token or in offer state
-        // For pre-authorized flows, offer state is the source of truth
-        // For authorization code flows, token must contain authorization_details
         if (authDetails == null) {
-            authDetails = getAuthorizationDetailFromToken(accessToken);
-        }
-
-        if (authDetails == null) {
-            var errorMessage = "No authorization_details found in offer state or token";
+            var errorMessage = "No authorization_details found in offer state";
             throw new BadRequestException(getErrorResponse(ErrorType.INVALID_CREDENTIAL_REQUEST, errorMessage));
         }
 
         // Validate that authorization_details from the token matches the offer state
         // This ensures the correct access token is being used for the credential request
         OID4VCAuthorizationDetail tokenAuthDetails = getAuthorizationDetailFromToken(accessToken);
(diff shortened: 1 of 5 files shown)


VinodAnandan:
@mposolda  Could you please use "[OID4VCI]" prefix for the title ?

mposolda:
> @mposolda Could you please use "[OID4VCI]" prefix for the title ?

Yes, updated. Thanks

tdiesler:
Thanks for that - LGTM

ahus1 (approved):
Approving based on previous reviews. Leaving it up to Marek if you want to merge it now.

mposolda:
@tdiesler @ahus1 Thanks!

tdiesler:
This lookup might be redundant

```
        // Add the scope referenced by the credential from specified credential offer to the token scopes
        String credentialConfigId = authDetails.getCredentialConfigurationId();
        ClientScopeModel clientScope = OID4VCUtil.getClientScopeByCredentialConfigId(session, realm, credentialConfigId);
        if (clientScope == null) {
            String errorMessage = "Client scope was not found for credential configuration ID: " + credentialConfigId;
            event.detail(Details.CREDENTIAL_TYPE, credentialConfigId);
            event.detail(REASON, errorMessage)
                    .error(UNKNOWN_CREDENTIAL_CONFIGURATION.getValue());
            throw new CorsErrorResponseException(cors, OAuthErrorException.INVALID_REQUEST,
                    errorMessage, Response.Status.BAD_REQUEST);
        }
        accessToken.setScope(clientScope.getName());
```

Since credentialConfigId == ClientScope.Name

This should work

```
accessToken.setScope(credentialConfigId);
```

However, it relates to #46489 which wonders about `credentialConfigId == ClientScope.Name`

In any case, the scopeName is already known at credential offer create time, so it could become part of the CredentialOfferState and then again the lookup would be redundant. I could do that in one of the outstanding PRs in the area, if you like.

PR #46482, opened 2026-02-19
Fix unconfigured non-CredentialValidators appearing in auth selection list

### LINKED ISSUE #46481: Unconfigured non-CredentialValidators appear in selection list
### Before reporting an issue

### Area

authentication

### Describe the bug

All non-`CredentialValidator` Authenticators are included in the authentication selection list (the "Try another way" dialog), even if they are not configured for the user (`!authenticator.configuredFor(...)`) and do not allow user setup (`!factory.isUserSetupAllowed()`).

This results in users seeing authentication options they cannot actually use. If a user selects one of these improperly listed options, it results in an error.

### Version

26.5.3

### Regression

### Expected behavior

The authentication selection list should only display authenticators that are either:

1. Already configured for the user (`configuredFor(..) == true`)
2. Available to be configured by the user (`isUserSetupAllowed() == true`)

### Actual behavior

`AuthenticationSelectionResolver#createAuthenticationSelectionList(...)` bypasses these checks for non-CredentialValidator authenticators, returning them in the list regardless of whether they apply to the user or not.

### How to Reproduce?

1. Create a `CustomAuthenticator` that implements `Authenticator` but not `CredentialValidator<CredentialProvider>`. Ensure `CustomAuthenticator#configuredFor(..)` and `CustomAuthenticatorFactory#isUserSetupAllowed()` both return false for a test user.
2. Configure a conditional authentication subflow as follows:  - Condition - User Configured (REQUIRED)
  - WebAuthn (ALTERNATIVE)
  - OTP (ALTERNATIVE)
  - CustomAuthenticator (ALTERNATIVE)

Case 1: User has no configured alternatives (No OTP, No WebAuthn)
1. Attempt to authenticate with the test user.
2. Result: The user is not challenged by the `CustomAuthenticator`. The flow evaluates the conditions properly, and this behaves exactly as expected.

Case 2: User has at least one configured alternative (e.g., OTP or WebAuthn)
1. Attempt to authenticate with a test user who has OTP or WebAuthn already configured.
2. The user is correctly challenged by their configured alternative.
3. On this challenge screen, click the "Try another way" link.
4. Result: The CustomAuthenticator is incorrectly displayed as a selectable option in the list. If the user selects it, Keycloak throws an error because the authenticator is not actually configured for them.

### Anything else?

The root cause appears to be in `AuthenticationSelectionResolver#createAuthenticationSelectionList(...)`, which assumes non-`CredentialValidator` authenticators should always be added to the selection list without verifying `configuredFor` or `isUserSetupAllowed`.

michelecoco:
Closes #46481

This PR fixes the issue where unusable alternative authenticators appear in the "Try another way" dialog.

See linked issue for reproduction details

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/AuthenticationSelectionResolver.java
+++ b/services/src/main/java/org/keycloak/authentication/AuthenticationSelectionResolver.java
@@ -29,6 +29,7 @@ import org.keycloak.credential.CredentialModel;
 import org.keycloak.models.AuthenticationExecutionModel;
 import org.keycloak.models.AuthenticationFlowModel;
 import org.keycloak.models.RealmModel;
+import org.keycloak.models.UserModel;
 
 import org.jboss.logging.Logger;
 
@@ -78,13 +79,12 @@ class AuthenticationSelectionResolver {
             }
 
             //add credential authenticators in order
-            if (processor.getAuthenticationSession().getAuthenticatedUser() != null) {
-                authenticationSelectionList =
-                        Stream.concat(
-                            processor.getAuthenticationSession().getAuthenticatedUser().credentialManager().getStoredCredentialsStream()
-                                .map(CredentialModel::getType),
-                            processor.getAuthenticationSession().getAuthenticatedUser().credentialManager()
-                                .getConfiguredUserStorageCredentialTypesStream())
+            UserModel user = processor.getAuthenticationSession().getAuthenticatedUser();
+            if (user != null) {
+                authenticationSelectionList = Stream.concat(
+                                user.credentialManager().getStoredCredentialsStream().map(CredentialModel::getType),
+                                user.credentialManager().getConfiguredUserStorageCredentialTypesStream()
+                        )
                         .distinct()
                         .filter(typeAuthExecMap::containsKey)
                         .map(credentialType -> new AuthenticationSelectionOption(processor.getSession(), typeAuthExecMap.get(credentialType)))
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/keycloak/authentication/AuthenticationSelectionResolver.java
```diff
@@ -101,8 +101,21 @@ static List<AuthenticationSelectionOption> createAuthenticationSelectionList(Aut
                 });
             }
 
-            //add all other authenticators
+            //add non credentials authenticators
```
angelomelonas, line 104:
```suggestion
            //add non-credential authenticators
```

angelomelonas (approved):

mposolda:
@rmartinc Do you please have a chance to review?

services/src/main/java/org/keycloak/authentication/AuthenticationSelectionResolver.java
```diff
@@ -101,8 +101,21 @@ static List<AuthenticationSelectionOption> createAuthenticationSelectionList(Aut
...
             for (AuthenticationExecutionModel exec : nonCredentialExecutions) {
+                if (user != null) {
+                    AuthenticatorFactory factory = (AuthenticatorFactory) processor.getSession().getKeycloakSessionFactory().getProviderFactory(Authenticator.class, exec.getAuthenticator());
+                    Authenticator authenticator = factory.create(processor.getSession());
+
+                    boolean isAvailable = exec.isRequired()
```
rmartinc, line 110: Probably this should be `exec.isRequired() || exec.isConditional()` because the authenticator can be conditional and evaluated to `true`. Please check if it works OK for you.

rmartinc (changes requested):
Thanks @michelecoco! Looks more or less OK to me. One comment in the code to also check conditional (please test before).

And, as you see, there are failures in the tests. The problem is that the `IdpEmailVerificationAuthenticator` is returning always false for `configuredFor`. This [line](https://github.com/keycloak/keycloak/blob/26.5.4/services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpEmailVerificationAuthenticator.java#L123). You can modify it for `return user.getEmail() != null` or similar.

rmartinc:
@michelecoco Are you working on the changes? Thanks!

mabartos:
@michelecoco Do you plan to finish it, or should we take over?

PR #46483, opened 2026-02-19
Allow to expose WellKnown provider via ServerMetadataResource (#46292)

thomasdarimont:
- Introduce isAvailableViaServerMetadata() on WellKnownProviderFactory to let each provider control whether it is exposed via the root /.well-known/{alias}/realms/{realm} endpoint, defaulting to false to preserve existing behavior
- Extract shared alias-resolution logic into a new WellKnownProviderUtil.resolveFromAlias() utility, replacing duplicated lookup code in both RealmsResource and ServerMetadataResource
- Opt in OAuth2WellKnownProviderFactory, OID4VCIssuerWellKnownProviderFactory, and JWTVCIssuerWellKnownProviderFactory by overriding isAvailableViaServerMetadata() to return true
- Align ServerMetadataResource method names by replacing getOAuth2AuthorizationServerWellKnownVersionPreflight with getWellKnownPreflight and getOAuth2AuthorizationServerWellKnown with getWellKnown.
- Add test to verify .well-known/openid-configuration is not exposed via server metadata root by default.
- Added wellKnownProviderUrl(..) method to ServerMetadataResource, se we could also have non-oauth well-known providers.
- Marked the old ServerMetadataResource.wellKnownOAuthProviderUrl(..) as deprecated. Changed current call-sites to use wellKnownProviderUrl.

Previously, ServerMetadataResource.isValidProvider() used a hardcoded allowlist of provider IDs to decide which well-known endpoints were exposed at the server root. This meant every new provider required a code change in ServerMetadataResource. With this change, each WellKnownProviderFactory can declare its own eligibility, making the mechanism extensible for custom and future providers without modifying the resource class.

Fixes https://github.com/keycloak/keycloak/issues/46292

Followup PR to #46294 with all changes requested in code review.

### DIFF
--- a/services/src/main/java/org/keycloak/services/util/WellKnownProviderUtil.java
+++ b/services/src/main/java/org/keycloak/services/util/WellKnownProviderUtil.java
@@ -37,7 +37,6 @@ public class WellKnownProviderUtil {
         return sessionFactory.getProviderFactoriesStream(WellKnownProvider.class)
                 .map(providerFactory -> (WellKnownProviderFactory) providerFactory)
                 .filter(wellKnownProviderFactory -> alias.equals(wellKnownProviderFactory.getAlias()))
-                .sorted(Comparator.comparingInt(WellKnownProviderFactory::getPriority))
-                .findFirst();
+                .min(Comparator.comparingInt(WellKnownProviderFactory::getPriority));
     }
 }
--- a/services/src/main/java/org/keycloak/services/util/WellKnownProviderUtil.java
+++ b/services/src/main/java/org/keycloak/services/util/WellKnownProviderUtil.java
@@ -1,0 +1,43 @@
+package org.keycloak.services.util;
+
+import java.util.Comparator;
+import java.util.Optional;
+
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.wellknown.WellKnownProvider;
+import org.keycloak.wellknown.WellKnownProviderFactory;
+
+/**
+ * Utility class for resolving {@link WellKnownProviderFactory} instances based on their alias.
+ * This class provides methods to interact with the available provider factories within the context
+ * of a {@link KeycloakSessionFactory}.
+ *
+ * This is a utility class and is not intended to be instantiated.
+ */
+public class WellKnownProviderUtil {
+
+    private WellKnownProviderUtil() {
+        // Utility class
+    }
+
+    /**
+     * Resolves a {@link WellKnownProviderFactory} from the specified alias.
+     * If multiple factories share the same alias, the one with the lowest priority is selected.
+     *
+     * @param sessionFactory the {@link KeycloakSessionFactory} to retrieve provider factories from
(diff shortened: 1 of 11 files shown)


thomasdarimont:
Added wellKnownProviderUrl method to ServerMetadataResource, se we could also have non-oauth well-known providers.

Marked the old wellKnownOAuthProviderUrl as deprecated. Changed current call-sites to use wellKnownProviderUrl.

thomasdarimont:
@tnorimat @IngridPuppet may I kindly ask you for another review on this PR as a replacement for  https://github.com/keycloak/keycloak/pull/46294?
For some reason the original PR was jumbled with other commits inbetween that I cannot remove anymore.

services/src/main/java/org/keycloak/services/util/WellKnownProviderUtil.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param alias the alias of the desired {@link WellKnownProviderFactory}; if null, an empty {@link Optional} is returned
+     * @return an {@link Optional} containing the resolved {@link WellKnownProviderFactory}, or empty if no matching factory is found
+     */
+    public static Optional<WellKnownProviderFactory> resolveFromAlias(KeycloakSessionFactory sessionFactory, String alias) {
+
+        if (alias == null) {
+            return Optional.empty();
+        }
+
+        return sessionFactory.getProviderFactoriesStream(WellKnownProvider.class)
```
tnorimat, line 37: This streams, filters, sorts, and picks the first element. The sorted + findFirst pattern is O(n log n) for the filtered subset. Since this is an HTTP request hot-path, consider using 
`min(Comparator.comparingInt(WellKnownProviderFactory::getPriority))
` instead which is O(n):

How about the following? 
```
return sessionFactory.getProviderFactoriesStream(WellKnownProvider.class)
        .map(providerFactory -> (WellKnownProviderFactory) providerFactory)
        .filter(wellKnownProviderFactory -> alias.equals(wellKnownProviderFactory.getAlias()))
        .min(Comparator.comparingInt(WellKnownProviderFactory::getPriority));
```

This is a minor optimization (the number of providers is small), but it better expresses intent.

Therefore, it is up to you to consider the comment.

tnorimat (commented):
@thomasdarimont Hello, I added one minor comment Could you check that? It is up to you to consider the point.

services/src/main/java/org/keycloak/services/util/WellKnownProviderUtil.java
thomasdarimont: Thanks for the hint. I think this makes sense, I'll change it as suggested.

tnorimat (approved):
LGTM

IngridPuppet (approved):

VinodAnandan:
@mposolda @ahus1 Both @tnorimat and @IngridPuppet  has reviewed and approved this PR. When you have a moment, could you kindly review/approve this PR ?

thomasdarimont:
Rebased on current main

stianst (approved):

PR #46486, opened 2026-02-20
Improve the events in the docker endpoint

### LINKED ISSUE #46469: Improve the events in the docker endpoint
### Description

Follow-up of #46462 just to add the  comments dome in the PR https://github.com/keycloak/keycloak/pull/46465#issuecomment-3927414167:
 
* Instead of logging errors, I prefer to add those detail messages to the event. We already already do that for OIDC in several places, and it allows those information to be shown in the events admin console where it is more accessible for people who have access to the UI, but not to the log file. This would then replace the logger.errorv(), and can be shorter as the event context AFAIK already contains the realm.
* It seems to me that the client is not set in the event context after it has been looked up.
* Probably events can also be checked in the test.

rmartinc:
Closes #46469

Little changes in the docker protocol to improve the events. The main difference is the test now checks the events. @ahus1 take a look to see if you see this OK (sometimes I have added the reason, sometimes not, following what I have seen in other endpoints).

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/docker/DockerAuthV2Protocol.java
+++ b/services/src/main/java/org/keycloak/protocol/docker/DockerAuthV2Protocol.java
@@ -138,13 +138,12 @@ public class DockerAuthV2Protocol implements LoginProtocol {
                         .setIssued_at(expiresInIso8601String);
                 return new ResponseBuilderImpl().status(Response.Status.OK).header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON).entity(responseEntity).build();
             } else {
-                event.detail(Details.REASON, String.format("Unable to handle request for event type %s. Currently only LOGIN event types are supported by docker protocol.",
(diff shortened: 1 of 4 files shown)


ahus1:
> sometimes I have added the reason, sometimes not, following what I have seen in other endpoints

Thank you for the changes. The "how we do it in other places" might not be such a good guidance. Here some more thoughts: 

* An admin should be able to spot the cause of errors by using the Admin UI  and the recorded events 
* The information and details we provide to the caller are sometimes redacted for security reasons, as we don't want to provide too much information that can be used as OSINT
* The admin should receive the same information, and ideally enough information to make it actionable to fix the problem.

I've pushed one more change, please let me know your thoughts. I've added the same message returned to the caller in the details, although it might be redundant to the error code. I reduced the content that is already in the event in another place, and added details also to the InstantiationException.

rmartinc:
@ahus1 LGTM! Then the idea is always adding the details instead of any logger. I tried to do the same that it was done in other parts. :stuck_out_tongue_closed_eyes:

ahus1 (approved):

PR #46487, opened 2026-02-20
Do not select disabled IdPs in the DefaultAlternativeLookupProvider

### LINKED ISSUE #46309: Make sure disabled IDPs are not considered when lookup IDP by issuer
### Description

This task is a follow-up of https://github.com/keycloak/keycloak/issues/45747 and https://github.com/keycloak/keycloak/issues/45453 (which are both addressed right now).

The issue https://github.com/keycloak/keycloak/issues/45747 added some limitations that it is not possible to have multiple IDPs in the realm with same "issuer" as long as some other preconditions are met (EG. 2 IDPs configured to support jwt-authorization grant). But this does not consider if IDP is enabled or not right now.

I assume that we should allow multiple IDPs with same issuer if only one of them is enabled, but others disabled. In that case, we also need to make sure every time when some issuer is going to be switched from disabled to enabled, that there is no other enabled IDP with same "issuer" value (considering that duplicate issuer should not be allowed based on other preconditions, like for example "JWT authorization grant" is used).

If going this path, we should likely fix also `DefaultAlternativeLookupProvider.lookupIdentityProviderFromIssuer` and consider only enabled issuers during lookup.

We should possibly also disable the IDPs with duplicated issuers during migration to this version? For example if someone already have 2 enabled IDPs in their realm with same "Issuer" and with enabled JWT-authorization grant, then they should be rather both disabled during migration. Probably can be done programatically by creating new `MigrateToXY` class in the package `org.keycloak.migration.migrators`

Some relevant discussion points for additional details:
https://github.com/keycloak/keycloak/pull/46024#issuecomment-3880470558
https://github.com/keycloak/keycloak/pull/46024#issuecomment-3891819371 
https://github.com/keycloak/keycloak/pull/46024#issuecomment-3892259778

rmartinc:
Closes #46309

I'm creating a draft because don't know what we are going to finally decide. This PR just changes the default lookup provider to not take into account the disabled clients or identity providers. Test added for the IdP check.

If we want a real check in the DB, we need much more. We need to update the database to add the issuer as a new column in the IdP table, and create a constraint in it. We would need to smartly update the IdPs adding that column only when the issuer should be checked as unique (the IdP is of correct type, enabled, with jwt grant or spiffe enabled and so on and so forth). When the issuer is not needed, the column should be set to a unique different value (a generated ID or similar). The upgrade needs to check everything to be OK in the end.

What do you want to do? I think that this one is needed for sure. If we decide that the DB constraint is needed, we can also manage that as a follow-up.

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -35,6 +35,11 @@ If you are using this Quarkus property, the new `transaction-default-timeout` CL
 
 To maintain your desired transaction timeout configuration, migrate to using the `transaction-default-timeout` CLI option and remove the `quarkus.transaction-manager.default-transaction-timeout` property from your configuration.
 
+=== The Identity Provider issuer should be unique for the JWT authorization grant and client assertions
+
+The JWT authorization grant and the client assertion authentication use the `issuer` claim passed in the assertion to locate the Identity Provider defined in {project_name}, and that provider validates the claims and signature in the assertion. Since this version, the `issuer` configuration option should be unique to only identify one provider. The grant operation or the assertion client authentication fail if more providers are returned for a single issuer. In the same way, the Identity Providers, when configured to allow the JWT authorization grant or client assertions, are validated to have a unique issuer defined in its configuration. Any update or add operation will fail if the issuer collides with another provider.
+
+If you have identity providers defined with the same issuer and tested the JWT authorization grants or client assertions features, please update the providers accordingly before the upgrade. You can update the issuer or disable the provider to avoid the duplicates.
 
 // ------------------------ Notable changes ------------------------ //
 == Notable changes
--- a/server-spi-private/src/main/java/org/keycloak/broker/provider/IdentityProvider.java
+++ b/server-spi-private/src/main/java/org/keycloak/broker/provider/IdentityProvider.java
@@ -22,7 +22,10 @@ import java.util.List;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.UriInfo;
 
+import org.keycloak.broker.provider.util.IdentityProviderTypeUtil;
 import org.keycloak.models.IdentityProviderModel;
+import org.keycloak.models.IdentityProviderType;
+import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
 import org.keycloak.provider.Provider;
 
@@ -58,5 +61,17 @@ public interface IdentityProvider<C extends IdentityProviderModel> extends Provi
      */
     default boolean reloadKeys() {
         return false;
+    }
+
+    /**
+     * Returns if this Identity Provider is of the passed type. By default it just returns
+     * true when it implements the correct interface. Sub-classes like the OIDC
+     * provider can check specific configuration options.
+     * @param session The helper session
+     * @param type The type to check
+     * @return true if the provider is of the passed type, false otherwise
+     */
+    default boolean isType(KeycloakSession session, IdentityProviderType type) {
+        return IdentityProviderTypeUtil.listTypesFromProvider(session, this).contains(type);
     }
 }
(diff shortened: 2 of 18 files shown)


graziang:
@rmartinc Thanks! For me, DB constraints on the issuer are not needed. In addition to your changes, I would add the possibility to have IDPs with the same issuer if only one is enabled, so we should just skip issuer validation if the IDP is disabled.

I’m also not sure about adding a migration to disable IDPs with the same issuer and with jwt auth grant enabled. Maybe a note in the migration notes would be enough. WDYT?

rmartinc:
@graziang with this changes only enabled IdPs are considered, if you check the test I'm doing what you are proposing, I enable and disable to check the IdP is allowed with the same issuer. As we are using the lookup to validate the IdP, only the enabled ones are considered now.

Regarding the DB check, I don't know, it depends on how far we want to go with this.

rmartinc:
Now I understood what you meant :stuck_out_tongue_closed_eyes:, you wanted to allow to disable a duplicated issuer IdP. It's allowed now.

graziang (approved):
@rmartinc yeah, for me with this PR we have good handling of duplicate issuers without making changes to the db.

mposolda:
@rmartinc @graziang Thanks for the PR and discussion!

I also hope that DB constraint is not needed (lots of effort and overhead, with not so much added value though...).

But maybe safer to do programmatic migration and in case that realm already has 2 IDPs with same issuer enabled, then rather disable both of them and add warning to server.log about that? See last paragraph of the issue description for the details https://github.com/keycloak/keycloak/issues/46309 . If this is hard to do, the note in the "Upgrading guide" would be a minimum IMO as when someone already has 2 IDPs with same issuer, things will stop to work for him after this change. My vote ATM is to do both "programmatic migration" and "upgrading guide" notice.

rmartinc:
@mposolda As commented in the meeting yesterday, disabling IdPs at upgrade is a mess IMHO. We would maintain one (random, the first one, something like this) and disable the others. I don't see any progress with that and we are introducing a lot of overhead in the migration. This is a preview, so I will document the change in the upgrading guide as critical for the JWT authorization grant.

mposolda:
> @mposolda As commented in the meeting yesterday, disabling IdPs at upgrade is a mess IMHO. We would maintain one (random, the first one, something like this) and disable the others. I don't see any progress with that and we are introducing a lot of overhead in the migration. This is a preview, so I will document the change in the upgrading guide as critical for the JWT authorization grant.

If we disable on migration, we would need to disable all of them and then let admin to enable one of them manually. Surely not keep one "random" enabled and disable the others - agree that would be quite insecure and mess :-) 

But it is OK with me to go with Upgrading-guide only. Administrator needs to fix things anyway.

rmartinc:
Disabling IMO is dangerous because maybe we are disabling a real OIDC (used for normal user linking) just because they tested the jwt authroization grant or the client assertions in it. I really prefer to do nothing in the upgrade, and this way only the jwt assertion or client assertions fail after the upgrade. At least those features were preview before, and they need to do something to make them work now. I'm preparing a second commit to differentiate the cache by type. It adds a little overhead but maybe we are safer this way. I'll do that change in a second commit and we can discard it if we all think is not useful.

rmartinc:
@mposolda @graziang I have added the upgrade note in the first commit and a second commit that separate the caches by IdP type (jwt and client assertion). There is a little more overhead:

1. There are two caches now, so you have to initialize the two of them.
2. The lookup of the provider when the issuer is the same, loads the identity provider and checks the attributes.

I think that the overhead is not much, because the second point only applies when the issuer is the same. This way we can have the same issuer for two OIDC providers if one of them is not enabled for jwt grant and/or client assertion. As it's a different commit we can just remove it, and go as before, if you prefer to not separate the cache by type. Let me know what you prefer to move this out of draft.

mposolda (dismissed):
@rmartinc @graziang Thanks! Could you please just update the upgrading guide as there was conflict introduced in the meantime?

Marking this PR as "Ready for review" as I suppose no further changes to be done here.

rmartinc:
Done! I have also change the name of a method, Ithink the new one is better. Then I think that you have to maintain the two commits, the two different caches, right? Moving out of draft then.

mposolda (approved):

PR #46497, opened 2026-02-20
fix: allowing targetServerType to always be used

### LINKED ISSUE #46459: Operator-built DB config: targetServerType=primary not applied / connection validation not working after master-replica failover (26.5.0)
### Before reporting an issue

### Area

operator

### Describe the bug

**Environment**
Keycloak: 26.5.0
Deployment: Kubernetes with Keycloak Operator (CR k8s.keycloak.org/v2alpha1)
Database: Managed PostgreSQL with master-replica failover (e.g. Yandex Cloud MDB, same behaviour with other managed Postgres)
After a master-replica failover, Keycloak in Kubernetes (deployed via the Operator) keeps using the old master and fails with read-only errors until pods are restarted. The same Keycloak version and image in a container on a VM (same DB, no Operator) handle the same failover correctly: connections are validated and closed when they become read-only, and new connections go to the new master.
So the fix from #40383 (connect to writer with targetServerType=primary) appears to work when Keycloak builds the DB config itself (VM/container), but not when the Operator supplies the DB config (Kubernetes).

### Version

26.5.0

### Regression

### Expected behavior

- Keycloak should connect to the writer (primary) and recover from master-replica failover without pod restarts, as in #40383.
- Connection pool validation should detect read-only connections and close them (we see Closing connection in incorrect state VALIDATION in logs on the VM; not in K8s).

### Actual behavior

- In Kubernetes (Operator):
After failover, Keycloak keeps using the old master.
Read-only errors until pods are restarted.
No Closing connection in incorrect state VALIDATION logs (validation does not seem to trigger the same way).
- On VM (same image, same DB, no Operator):
Failover is handled correctly.
Logs show: WARN [io.agroal.pool] Datasource '<default>': Closing connection in incorrect state VALIDATION.

### How to Reproduce?

Configuration
Operator (not working after failover):
spec:
```
spec:
  db:
    vendor: postgres
    host: <cluster>.rw.mdb.yandexcloud.net
    port: 6432
    database: keycloak_db
    usernameSecret: ...
    passwordSecret: ...
```
No targetServerType=primary in the effective JDBC URL when the Operator builds it from host/port/database.
Workaround that works:
Using an explicit JDBC URL in spec.db.url so that the parameter is present:
```
spec:
  db:
    vendor: postgres
    url: "jdbc:postgresql://<host>:6432/<database>?targetServerType=primary"
    poolInitialSize: 10
    poolMinSize: 10
    poolMaxSize: 10
    usernameSecret: ...
    passwordSecret: ...
```
After this change, failover is handled correctly in Kubernetes as well.

### Anything else?

When the Operator builds the JDBC URL from spec.db.host / port / database, it does not add targetServerType=primary (and possibly other JDBC options). So deployments that rely on the Operator’s DB section never get the behaviour fixed in #40383 unless the user switches to db.url.
Could the Operator either add targetServerType=primary for PostgreSQL when building the URL, or document that for HA/failover users must set db.url (or equivalent) with this parameter?
The difference in connection validation (VM vs K8s) might be related to how the URL is built or which options are set. Any pointer on why validation would behave differently in the Operator case would help.

shawkins:
closes: #46459

We're getting bitten again by quarkus needing a property advertised. 

We may want to eventually consider more systematic approaches to things like this:
- should there be an option / property mapper flag to indicate it should always be advertised?
- alternatively should we just go ahead and advertise all quarkus to keys even if there might not be a value? - trying this out did break an existing test, not sure if this would be a good idea for backport.
- and do we even need dummy/hidden keycloak options for something like this.

### DIFF
--- a/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
+++ b/quarkus/config-api/src/main/java/org/keycloak/config/DatabaseOptions.java
@@ -117,6 +117,7 @@ public class DatabaseOptions {
 
     public static final Option<String> DB_POSTGRESQL_TARGET_SERVER_TYPE = new OptionBuilder<>("db-postgres-target-server-type", String.class)
             .category(OptionCategory.DATABASE)
+            .defaultValue("primary") // cause the propertymapping logic to always advertise this property
             .hidden()
             .build();
 
--- a/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
+++ b/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/DatabasePropertyMappers.java
@@ -29,7 +29,8 @@ import static org.keycloak.quarkus.runtime.configuration.MicroProfileConfigProvi
 import static org.keycloak.quarkus.runtime.configuration.mappers.DatabasePropertyMappers.Datasources.appendDatasourceMappers;
 import static org.keycloak.quarkus.runtime.configuration.mappers.PropertyMapper.fromOption;
 
-final class DatabasePropertyMappers implements PropertyMapperGrouping {
+public final class DatabasePropertyMappers implements PropertyMapperGrouping {
+    public static final String PG_TARGET_SERVER_TYPE = "quarkus.datasource.jdbc.additional-jdbc-properties.targetServerType";
     private static final Logger log = Logger.getLogger(DatabasePropertyMappers.class);
 
     @Override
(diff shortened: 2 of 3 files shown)


ahus1 (approved):
Thank you for this fix. I confirm it is working now, I did debug the connection and see that it is now set, where it wasn't set for 26.5

+1 for providing a backport to 26.5 - I've set the tag. 

I debugged locally that it was working in 26.4? Unfortunately I did not come up with a test for this to check the DB connection directly.

Not merging yet as another CND team member is added as a reviewer. Leaving it to Steven to merge it or not.

michalvavrik (approved):
LGTM

PR #46532, opened 2026-02-23
Set custom audiences per client in the jwt audience executor

### LINKED ISSUE #46525: Improve the custom audiences executor to setup audiences per client
### Description

This is a follow-up for #45180. Currently the executor allow to add custom audiences and they apply to any client. This can lead to misconfiguration, because the executor can be enforced to a lot of clients and then all of then will use the same audience (which is not a common scenario). The idea is improving the executor to allow to configure the audience per client.

What I don't know is where to add this per client configuration:

1. In the client. Now the clients add a config option with the IdPs in which the JWT authorization grant is used. We can add in the client what audience is used for that IdP.
2. In the IdP. We can add audiences per client in the IdP too.
3. In the executor. Using a map client->audience for example.

IMO the correct idea is using 3. Because if we use the client or the IdP, the client policy is not needed at all (the grant can directly know what audience this client/IdP pair requires). The only problem is the configuration, which using a executor would be generic (`Map` type).

rmartinc:
Closes #46525

PR that configures the valid audiences for the jwt authorization grant audience executor per client. Initially I thought about adding the option in the client, but, in that case, the executor is not needed at all (the grant execution knows the valid audience just using the client configuration). So I decided to finally add the configuration in the executor per client. Let me know what you think.


graziang:
@rmartinc The general idea is ok for me. I would keep the option to apply the audience also to all clients. Even if it’s not common, there could be a case where someone wants to just apply the audience to all clients, and configuring all of them in the map is not so easy.

rmartinc:
@graziang we are trying to avoid that, because we want the admin to configure the correct audience to the correct client. If you apply all audiences like before (to all clients managed by the policy), is like allowing the jwt authorization grant to all of them (because the ID token in google would be always the same one). Stian asked to restrict this behavior.

mposolda:
@rmartinc Sorry, but I recall what we discussed (and pretty sure what @stianst mentioned) is to rather not use client policies, but instead have some per-client configuration. The client policies are ideal for the cases, when some configuration (or restriction) applies to many various clients, but not when every client can have different setting as: 

- It doesn't scale (for 100 different clients, there would need to be 100 configuration entries in the client policy executor)
- It doesn't work when there is client administrator of specific client `foo` . Client administrator will not be able to configure the desired audience is as he may not have permissions for edit client policies (as client policies are usually editable by realm administrator, but not by administrators of individual clients).

rmartinc:
Finally I used the `MapComponent` in the client. We can improve this (for example using a select for the IdP over the selected ones), but making the component work as I wanted took me a lot of time, so I don't want to spend more time on this. It's also rebased that was needed.

rmartinc:
I have changed the `MapComponent` to accept options for the key. Let's see.

mposolda (approved):
@rmartinc Thanks! I hope this can be merged once tests are OK

rmartinc:
I'm going to rebase this, I think that it's too old to just re-run it.

PR #46539, opened 2026-02-23
Move test class for persistent CIMD to the new test-suite

rmartinc:
Closes #46438

Moving the test class for the CIMD to the new test-suite. The cimd is created similar to the jwks endpoints for spiffe/jwt-authz-grant using a java http handler. The handler is fixed to `/cimd/metadata` context, but I think that it's enough for the moment. I have removed the helper classes in the old test-suite. The test class is very similar, I have separated some methods to just include one condition/executor in each test but all the logic has been migrated. @tnorimat Please take a look when you have time.


tests/base/src/test/java/org/keycloak/tests/client/policies/ClientIdMetadataDocumentTest.java
```diff
@@ -0,0 +1,931 @@
...
+        executorConfig.setAllowHttpScheme(true);
+        executorConfig.setTrustedDomains(List.of("*.example.com","localhost"));
+        updatePolicy(conditionConfig, executorConfig);
+
+        // set Client Metadata
+        // CIMD Executor's client metadata upper limit byte length default value: 5000
+        cimd.getRepresentation().setLogoUri("0123456789".repeat(465) + "0123456");
+
+        // send an authorization request - fail
+        assertLoginAndError(AbstractClientIdMetadataDocumentExecutor.ERR_METADATA_FETCH_FAILED);
```
tnorimat, line 530: The old test directly asserted `assertEquals(5001, contentBytes.length` to verify the payload size exceeded the 5000-byte limit. The new test sets `logoUri` to a long string but doesn't verify the serialized size is actually > 5000 bytes. If the `CimdClientConfig.build()` defaults change (e.g., different field values), this test could silently become invalid. It might be good to consider adding an assertion or comment about the expected payload size.

tnorimat (commented):
@rmartinc Hello, I added one minor comment. Could you check that?

tests/base/src/test/java/org/keycloak/tests/client/policies/ClientIdMetadataDocumentTest.java
rmartinc: Done!

tnorimat (approved):
LGTM

ahus1 (approved):
Approving based on @tnorimat's review.

PR #46544, opened 2026-02-23
Fix single-file export including non-imported LDAP users 

abigailllr:
- Single file export queried all providers including LDAP, exporting non-imported users with IDs longer than 36 characters. 
- Use local storage only, matching the existing directory export behavior.

Closes [#46537](https://github.com/keycloak/keycloak/issues/46537)


sguilhen (approved):
Thanks @abigailtech - it makes sense to me.

@pedroigor can you run the CI?

pedroigor (approved):

pedroigor:
Thanks, @abigailtech.

PR #46551, opened 2026-02-24
added test that verifies that Clients are created/updated with v1 are read correctly with v2

### LINKED ISSUE #46541: Create tests for Client v1 <--> v2 mapping
### Description

To prevent regressions while implementing the new service layer, add integration tests that verify the Clients are created/updated and read correctly. To do that, use v1. Proposed tests include:
* Create using v1, assert using v2.
* Create using v2, assert using v1.
* Update using v2, assert using v1.

Always ensure all fields are mapped as expected.

edewit:
fixes: #46541

### DIFF
--- a/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
+++ b/rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
@@ -1,0 +1,263 @@
+/*
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.tests.admin.client.v2;
+
+import java.util.Arrays;
+import java.util.Set;
+
+import jakarta.ws.rs.core.HttpHeaders;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+
+import org.keycloak.admin.client.Keycloak;
+import org.keycloak.admin.client.resource.RealmResource;
+import org.keycloak.common.Profile;
+import org.keycloak.representations.admin.v2.OIDCClientRepresentation;
+import org.keycloak.representations.idm.ClientRepresentation;
+import org.keycloak.testframework.annotations.InjectAdminClient;
+import org.keycloak.testframework.annotations.InjectHttpClient;
+import org.keycloak.testframework.annotations.InjectRealm;
+import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
+import org.keycloak.testframework.realm.ManagedRealm;
+import org.keycloak.testframework.server.KeycloakServerConfig;
+import org.keycloak.testframework.server.KeycloakServerConfigBuilder;
+import org.keycloak.testframework.util.ApiUtil;
+import org.keycloak.tests.admin.mapper.ClientRepresentationComparator;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * This class tests the interoperability between v1 and v2 admin APIs.
+ * It verifies that clients created/updated via one API version can be correctly
+ * read and validated via the other API version.
+ */
+@KeycloakIntegrationTest(config = InteropTest.ServerConfig.class)
+public class InteropTest extends AbstractClientApiV2Test {
+
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    @InjectHttpClient
+    CloseableHttpClient httpClient;
+
+    @InjectAdminClient
+    Keycloak adminClient;
+
+    @InjectRealm
+    ManagedRealm managedRealm;
+
+    @Override
+    public String getRealmName() {
+        return managedRealm.getName();
+    }
+
+    /**
+     * Test: Create a client using v1 API, then assert/read using v2 API.
+     */
+    @Test
+    public void createWithV1AssertWithV2() throws Exception {
+        RealmResource realm = managedRealm.admin();
+
+        ClientRepresentation v1Client = new ClientRepresentation();
+        v1Client.setClientId("v1-created-client");
+        v1Client.setName("V1 Created Client");
+        v1Client.setDescription("Client created via v1 API");
+        v1Client.setEnabled(true);
+        v1Client.setPublicClient(false);
+        v1Client.setProtocol("openid-connect");
+        v1Client.setBaseUrl("http://localhost:3000");
+        v1Client.setRedirectUris(Arrays.asList("http://localhost:3000/*", "http://localhost:3001/*"));
+        v1Client.setWebOrigins(Arrays.asList("http://localhost:3000", "http://localhost:3001"));
+        v1Client.setStandardFlowEnabled(true);
+        v1Client.setDirectAccessGrantsEnabled(true);
+        v1Client.setServiceAccountsEnabled(false);
+        v1Client.setClientAuthenticatorType("client-secret");
+        v1Client.setSecret("test-secret-123");
+
+        Response response = realm.clients().create(v1Client);
+        String clientUuid = ApiUtil.getCreatedId(response);
+        response.close();
+
+        ClientRepresentation createdV1Client = realm.clients().get(clientUuid).toRepresentation();
+
+        HttpGet getRequest = new HttpGet(getClientsApiUrl() + "/v1-created-client");
+        setAuthHeader(getRequest, adminClient);
+
+        try (var httpResponse = httpClient.execute(getRequest)) {
+            assertThat(httpResponse.getStatusLine().getStatusCode(), is(200));
+            OIDCClientRepresentation v2Client = mapper.createParser(httpResponse.getEntity().getContent())
+                    .readValueAs(OIDCClientRepresentation.class);
+
+            ClientRepresentationComparator.ComparisonResult result = 
+                ClientRepresentationComparator.compare(createdV1Client, v2Client);
+
+            System.out.println("=== createWithV1AssertWithV2 ===");
+            System.out.println(result);
+
+            assertTrue(result.allMatch(), "V1 and V2 representations should match:\n" + result);
+        }
+
+        realm.clients().get(clientUuid).remove();
+    }
+
+    /**
+     * Test: Create a client using v2 API, then assert/read using v1 API.
+     */
+    @Test
+    public void createWithV2AssertWithV1() throws Exception {
+        RealmResource realm = managedRealm.admin();
+
+        OIDCClientRepresentation v2Client = new OIDCClientRepresentation();
+        v2Client.setClientId("v2-created-client");
+        v2Client.setDisplayName("V2 Created Client");
+        v2Client.setDescription("Client created via v2 API");
+        v2Client.setEnabled(true);
+        v2Client.setAppUrl("http://localhost:4000");
+        v2Client.setRedirectUris(Set.of("http://localhost:4000/*", "http://localhost:4001/*"));
+        v2Client.setWebOrigins(Set.of("http://localhost:4000", "http://localhost:4001"));
+        v2Client.setLoginFlows(Set.of(
+            OIDCClientRepresentation.Flow.STANDARD, 
+            OIDCClientRepresentation.Flow.DIRECT_GRANT
+        ));
+        OIDCClientRepresentation.Auth auth = new OIDCClientRepresentation.Auth();
+        auth.setMethod("client-secret");
+        auth.setSecret("v2-secret-456");
+        v2Client.setAuth(auth);
+
+        HttpPost createRequest = new HttpPost(getClientsApiUrl());
+        setAuthHeader(createRequest, adminClient);
+        createRequest.setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON);
+        createRequest.setEntity(new StringEntity(mapper.writeValueAsString(v2Client)));
+
+        try (var httpResponse = httpClient.execute(createRequest)) {
+            assertThat(httpResponse.getStatusLine().getStatusCode(), is(201));
+        }
+
+        ClientRepresentation v1Client = realm.clients().findByClientId("v2-created-client").get(0);
+        String clientUuid = v1Client.getId();
+        ClientRepresentation fullV1Client = realm.clients().get(clientUuid).toRepresentation();
+
+        HttpGet getRequest = new HttpGet(getClientsApiUrl() + "/v2-created-client");
+        setAuthHeader(getRequest, adminClient);
+
+        try (var httpResponse = httpClient.execute(getRequest)) {
+            assertThat(httpResponse.getStatusLine().getStatusCode(), is(200));
+            OIDCClientRepresentation fetchedV2Client = mapper.createParser(httpResponse.getEntity().getContent())
+                    .readValueAs(OIDCClientRepresentation.class);
+
+            ClientRepresentationComparator.ComparisonResult result = 
+                ClientRepresentationComparator.compare(fullV1Client, fetchedV2Client);
+
+            System.out.println("=== createWithV2AssertWithV1 ===");
+            System.out.println(result);
+
+            assertTrue(result.allMatch(), "V1 and V2 representations should match:\n" + result);
+        }
+
+        realm.clients().get(clientUuid).remove();
+    }
+
+    /**
+     * Test: Create a client using v1 API, update it using v2 API, then assert using v1 API.
+     */
+    @Test
+    public void updateWithV2AssertWithV1() throws Exception {
+        RealmResource realm = managedRealm.admin();
+
+        ClientRepresentation v1Client = new ClientRepresentation();
+        v1Client.setClientId("update-test-client");
+        v1Client.setName("Original Name");
+        v1Client.setDescription("Original description");
+        v1Client.setEnabled(true);
+        v1Client.setPublicClient(false);
+        v1Client.setProtocol("openid-connect");
+        v1Client.setRedirectUris(Arrays.asList("http://localhost:5000/*"));
+        v1Client.setStandardFlowEnabled(true);
+
+        Response response = realm.clients().create(v1Client);
+        String clientUuid = ApiUtil.getCreatedId(response);
+        response.close();
+
+        OIDCClientRepresentation v2Update = new OIDCClientRepresentation();
+        v2Update.setClientId("update-test-client");
+        v2Update.setDisplayName("Updated Name via V2");
+        v2Update.setDescription("Updated description via V2 API");
+        v2Update.setEnabled(true);
+        v2Update.setAppUrl("http://localhost:5000");
+        v2Update.setRedirectUris(Set.of("http://localhost:5000/*", "http://localhost:5001/*"));
+        v2Update.setWebOrigins(Set.of("http://localhost:5000"));
+        v2Update.setLoginFlows(Set.of(
+            OIDCClientRepresentation.Flow.STANDARD,
+            OIDCClientRepresentation.Flow.DIRECT_GRANT,
+            OIDCClientRepresentation.Flow.SERVICE_ACCOUNT
+        ));
+        OIDCClientRepresentation.Auth auth = new OIDCClientRepresentation.Auth();
+        auth.setMethod("client-secret");
+        auth.setSecret("updated-secret");
+        v2Update.setAuth(auth);
+
+        HttpPut updateRequest = new HttpPut(getClientsApiUrl() + "/update-test-client");
+        setAuthHeader(updateRequest, adminClient);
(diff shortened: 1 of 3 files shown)


rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * This class tests the interoperability between v1 and v2 admin APIs.
+ * It verifies that clients created/updated via one API version can be correctly
+ * read and validated via the other API version.
+ */
+@KeycloakIntegrationTest(config = InteropTest.ServerConfig.class)
+public class InteropTest extends AbstractClientApiV2Test {
+
+    private static final ObjectMapper mapper = new ObjectMapper();
```
mabartos:
```suggestion
```

Should be already defined in the abstract class, so you can directly use it.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+    @InjectAdminClient
+    Keycloak adminClient;
+
+    @InjectRealm
+    ManagedRealm managedRealm;
+
+    @Override
+    public String getRealmName() {
+        return managedRealm.getName();
+    }
```
mabartos:
```suggestion
```
The default for the managed realm without config is the master realm, so no need to have this override.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+        try (var httpResponse = httpClient.execute(getRequest)) {
+            assertThat(httpResponse.getStatusLine().getStatusCode(), is(200));
+            OIDCClientRepresentation v2Client = mapper.createParser(httpResponse.getEntity().getContent())
+                    .readValueAs(OIDCClientRepresentation.class);
+
+            ClientRepresentationComparator.ComparisonResult result = 
+                ClientRepresentationComparator.compare(createdV1Client, v2Client);
+
+            System.out.println("=== createWithV1AssertWithV2 ===");
+            System.out.println(result);
```
mabartos:
```suggestion
```

We don't leave the System.out/err in tests. I don't think we need to print it out, do we?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+            assertTrue(result.allMatch(), "V1 and V2 representations should match:\n" + result);
+        }
+
+        realm.clients().get(clientUuid).remove();
```
mabartos: This should be in the try-finally block - the try block should start immediately after the client is created (above the `String clientUuid=`)... and at the end, remove the client.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+        HttpPost createRequest = new HttpPost(getClientsApiUrl());
+        setAuthHeader(createRequest, adminClient);
+        createRequest.setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON);
+        createRequest.setEntity(new StringEntity(mapper.writeValueAsString(v2Client)));
+
+        try (var httpResponse = httpClient.execute(createRequest)) {
+            assertThat(httpResponse.getStatusLine().getStatusCode(), is(201));
+        }
+
+        ClientRepresentation v1Client = realm.clients().findByClientId("v2-created-client").get(0);
```
mabartos, line 158: The same with the try-finally block

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+        try (var httpResponse = httpClient.execute(getRequest)) {
+            assertThat(httpResponse.getStatusLine().getStatusCode(), is(200));
+            OIDCClientRepresentation fetchedV2Client = mapper.createParser(httpResponse.getEntity().getContent())
+                    .readValueAs(OIDCClientRepresentation.class);
+
+            ClientRepresentationComparator.ComparisonResult result = 
+                ClientRepresentationComparator.compare(fullV1Client, fetchedV2Client);
+
+            System.out.println("=== createWithV2AssertWithV1 ===");
+            System.out.println(result);
```
mabartos: Same as above

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,297 @@
...
+        return result;
+    }
+
+    private static void compareBaseFields(ComparisonResult result, ClientRepresentation v1, BaseClientRepresentation v2) {
+        compareField(result, "clientId", v1.getClientId(), v2.getClientId());
+        compareField(result, "name→displayName", v1.getName(), v2.getDisplayName());
+        compareField(result, "description", v1.getDescription(), v2.getDescription());
+        compareField(result, "enabled", v1.isEnabled(), v2.getEnabled());
+        compareField(result, "baseUrl→appUrl", v1.getBaseUrl(), v2.getAppUrl());
+        compareField(result, "protocol", v1.getProtocol(), v2.getProtocol());
```
mabartos: It might be hard to maintain, but we probably don't have any better approach.

mabartos (changes requested):
@edewit Thanks! Semantic looks good to me. For every test case, there are usually the same semantic suggestions that I made (try-finally block, remove sys.out,...), so it'd be good to apply it for all to avoid any future debugging of the problem and finding out the issue is far away on other place :) 

As I mentioned, we probably don't have any better mechanism on how to map the old and new representations, so it's ok to have the comparator class.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        realm.clients().get(clientUuid).remove();
+    }
+
+    /**
+     * Test: Create a client using v1 API, update it using v2 API, then assert using v1 API.
+     */
+    @Test
+    public void updateWithV2AssertWithV1() throws Exception {
```
mabartos, line 184: @edewit Maybe it'd be good to also do it the other way - update with v1 and assert with v2?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
edewit: I thought about using reflection, but that would mean that we would need to exclude a lot of fields that we can't compare like that

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
vmuzikar, line 1: Could you please extend the tests for SAML too? Seems the comparator logic is already there just unused.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
vmuzikar, line 1: This whole class feels a bit it's doing more than it has to. It's hard to read and I'm worried it'll be hard to maintain as well. Could you please try to clean it up a bit? Maybe we could do something simple and straightforward like in our [mappers](https://github.com/keycloak/keycloak/blob/main/rest/admin-v2/api/src/main/java/org/keycloak/models/mapper/OIDCClientModelMapper.java)?

vmuzikar (commented):
@edewit Thank you for the PR!

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
vmuzikar: That's not needed IMHO. If a resource created through v1 is correctly interpreted in v2, updating it using v1 doesn't bring much value.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
vmuzikar: @edewit This is still valid.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
vmuzikar: @edewit This is still valid.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
vmuzikar: @edewit This is still valid. Could you please add some SAML tests?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,247 @@
...
+            compareOIDCFields(oidc);
+        } else if (v2 instanceof SAMLClientRepresentation saml) {
+            compareSAMLFields(saml);
+        }
+
+        recordV1OnlyFields();
+        return result;
+    }
+
+    private void compareBaseFields() {
```
vmuzikar, line 83: Could you please also add cocmparison of roles?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,247 @@
...
+        if (Boolean.TRUE.equals(v1.isImplicitFlowEnabled())) flows.add(Flow.IMPLICIT);
+        if (Boolean.TRUE.equals(v1.isDirectAccessGrantsEnabled())) flows.add(Flow.DIRECT_GRANT);
+        if (Boolean.TRUE.equals(v1.isServiceAccountsEnabled())) flows.add(Flow.SERVICE_ACCOUNT);
+        return flows;
+    }
+
+    private void compareAuth(OIDCClientRepresentation oidc) {
+        String expectedMethod = Boolean.TRUE.equals(v1.isPublicClient()) ? "none"
+                : v1.getClientAuthenticatorType() != null ? v1.getClientAuthenticatorType()
+                : v1.getSecret() != null ? "client-secret" : null;
```
vmuzikar: Could you please rewrite this a bit? It's quite hard to read if the ternary operators are nested.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    private void compareSAMLFields(SAMLClientRepresentation saml) {
+        compare("frontchannelLogout→frontChannelLogout", v1.isFrontchannelLogout(), saml.getFrontChannelLogout());
+
+        Map<String, String> attrs = v1.getAttributes();
+        if (attrs == null) return;
+
+        compare("attr[saml_name_id_format]→nameIdFormat", attrs.get("saml_name_id_format"), saml.getNameIdFormat());
+        compareSamlBoolean("saml.force.name.id.format", "forceNameIdFormat", saml.getForceNameIdFormat());
```
vmuzikar:
```suggestion
        compareSamlBoolean("saml_force_name_id_format", "forceNameIdFormat", saml.getForceNameIdFormat());
```

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,247 @@
...
+        compare("attr[saml.signing.certificate]→signingCertificate", attrs.get("saml.signing.certificate"), saml.getSigningCertificate());
+    }
+
+    private void compareSamlBoolean(String attrName, String v2FieldName, Boolean v2Value) {
+        String attrValue = v1.getAttributes() != null ? v1.getAttributes().get(attrName) : null;
+        Boolean v1Value = attrValue != null ? Boolean.parseBoolean(attrValue) : null;
+        compare("attr[" + attrName + "]→" + v2FieldName, v1Value, v2Value);
+    }
+
+    private void recordV1OnlyFields() {
```
vmuzikar: We don't need this.

vmuzikar (changes requested):
@edewit Thank you for the update.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
edewit: it doesn't work when I remove it because "master" might be the one in the abstract base, but when I use it like this the name is "default"

edewit: Wait I of course can change it to not use this `managedRealm` and only use master

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,418 @@
...
+        v1Client.setName("V1 SAML Client");
+        v1Client.setDescription("SAML client created via v1 API");
+        v1Client.setEnabled(true);
+        v1Client.setProtocol("saml");
+        v1Client.setBaseUrl("http://localhost:8000/saml");
+        v1Client.setRedirectUris(Arrays.asList("http://localhost:8000/saml/*"));
+        v1Client.setFrontchannelLogout(true);
+        v1Client.setAttributes(java.util.Map.of(
+            "saml_name_id_format", "username",
+            "saml.force.name.id.format", "true",
```
mabartos:
```suggestion
            "saml_force_name_id_format", "true",
```
There might be also some other misalignments. It'd be good to make these attributes public in the SAMLClientModelMapper and reference them here.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    private void compareSAMLFields(SAMLClientRepresentation saml) {
+        compare("frontchannelLogout→frontChannelLogout", v1.isFrontchannelLogout(), saml.getFrontChannelLogout());
+
+        Map<String, String> attrs = v1.getAttributes();
+        if (attrs == null) return;
+
+        compare("attr[saml_name_id_format]→nameIdFormat", attrs.get("saml_name_id_format"), saml.getNameIdFormat());
+        compareSamlBoolean("saml_force_name_id_format", "forceNameIdFormat", saml.getForceNameIdFormat());
```
mabartos: The same with the attributes as above. Can be all using the `compareSamlBoolean` instead of the `compare`?

mabartos (commented):
@edewit Looks good, but one thing.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
mabartos: @edewit I think it'd be better to use these constants. WDYT?

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
mabartos: These should be already prepared AFAIK -> https://github.com/keycloak/keycloak/pull/46551/changes#diff-4ad3e0b3d823d22887e42e82c6a12e47fd54ee5a96d7ed0acaa2d2e10304c9ebR295

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
mabartos: Irrelevant

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
mabartos: Thanks!

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,420 @@
...
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import static org.keycloak.protocol.saml.SamlConfigAttributes.*;
```
mabartos: nitpick: we don't use wildcard imports, but that's for your future PRs

mabartos (dismissed):
@edewit LGTM now, thanks!

mabartos (dismissed):
@edewit LGTM, thanks!

mabartos:
@vmuzikar Do you want to do the final review?

vmuzikar:
@mabartos Plan to give it another look today.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
vmuzikar: That you for adding the comparator logic for roles. It would be good to add some test that would fill the roles field too.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/client/v2/InteropTest.java
```diff
@@ -0,0 +1,426 @@
...
+        v2Client.setRedirectUris(Set.of("http://localhost:9000/saml/*"));
+        v2Client.setFrontChannelLogout(true);
+        v2Client.setNameIdFormat("email");
+        v2Client.setForceNameIdFormat(true);
+        v2Client.setIncludeAuthnStatement(true);
+        v2Client.setSignDocuments(true);
+        v2Client.setSignAssertions(true);
+        v2Client.setClientSignatureRequired(false);
+        v2Client.setForcePostBinding(true);
+        v2Client.setSignatureAlgorithm("RSA_SHA256");
```
vmuzikar, line 323: We're missing tests for a few fields like `signatureCanonicalizationMethod`, `signingCertificate` or `allowEcpFlow`. But that's probably ok, we test it in the mapper tests.

vmuzikar (commented):
@edewit Thank you for the update. Overall looks good, I think the last missing piece is to test roles.

mabartos (dismissed):

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,234 @@
...
+
+    private void compareBaseFields() {
+        compare("clientId", v1.getClientId(), v2.getClientId());
+        compare("name→displayName", v1.getName(), v2.getDisplayName());
+        compare("description", v1.getDescription(), v2.getDescription());
+        compare("enabled", v1.isEnabled(), v2.getEnabled());
+        compare("baseUrl→appUrl", v1.getBaseUrl(), v2.getAppUrl());
+        compare("protocol", v1.getProtocol(), v2.getProtocol());
+        compareAsSet("redirectUris", v1.getRedirectUris(), v2.getRedirectUris());
+        compareAsSet("roles", Set.of(v1.getDefaultRoles() == null ? new String[0] : v1.getDefaultRoles()), v2.getRoles());
```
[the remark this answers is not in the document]
vmuzikar: This comment is valid but we test roles in a separate test methods so should be ok.

rest/admin-v2/tests/src/test/java/org/keycloak/tests/admin/mapper/ClientRepresentationComparator.java
```diff
@@ -0,0 +1,234 @@
...
+
+        compare("attr[saml_name_id_format]→nameIdFormat", attrs.get(SAML_NAME_ID_FORMAT_ATTRIBUTE), saml.getNameIdFormat());
+        compareSamlBoolean(SAML_FORCE_NAME_ID_FORMAT_ATTRIBUTE, "forceNameIdFormat", saml.getForceNameIdFormat());
+        compareSamlBoolean(SAML_AUTHNSTATEMENT, "includeAuthnStatement", saml.getIncludeAuthnStatement());
+        compareSamlBoolean(SAML_SERVER_SIGNATURE, "signDocuments", saml.getSignDocuments());
+        compareSamlBoolean(SAML_ASSERTION_SIGNATURE, "signAssertions", saml.getSignAssertions());
+        compareSamlBoolean(SAML_CLIENT_SIGNATURE_ATTRIBUTE, "clientSignatureRequired", saml.getClientSignatureRequired());
+        compareSamlBoolean(SAML_FORCE_POST_BINDING, "forcePostBinding", saml.getForcePostBinding());
+        compareSamlBoolean(SAML_SIGNATURE_ALGORITHM, "allowEcpFlow", saml.getAllowEcpFlow());
+        compare("attr[saml.signature.algorithm]→signatureAlgorithm", attrs.get(SAML_SIGNATURE_ALGORITHM), saml.getSignatureAlgorithm());
```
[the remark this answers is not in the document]
vmuzikar: This comment is valid, I missed it in my reviews, sorry. @edewit Could you please fix it?

vmuzikar (commented):
@edewit Thank you for the update. Looks good, I tried to experiment with Copilot here and it found two relevant issues from which one that I believe we should fix. Could you please check it?

vmuzikar (approved):
LGTM, thank you.

PR #46561, opened 2026-02-24
Support for PATCH operations

pedroigor:
Closes #46214

* Initial support for SCIM PATCH operations, considering most of the key capabilities from RFC7644
* What it does not provide yet:
   * Support for multivalued attributes, it is a general concern that should be handled by a separate task
   * Support for filters in the `path`, it will be addressed after [the filtering and pagination task](https://github.com/keycloak/keycloak/issues/46221)
   * Validations are still based on the user profile for users and for groups there is no validation based on the schema. Schema validation should be handled by a separate task


pedroigor:
@sguilhen Added a commit with updates to filtering

scim/core/src/main/java/org/keycloak/scim/resource/schema/AbstractModelSchema.java
```diff
@@ -146,23 +200,171 @@ private void populateResourceType(R resource, M model) {
...
+
+        if (filterStartIdx > 0) {
+            int filterEndIdx = path.lastIndexOf("]");
+
+            if (filterEndIdx == -1) {
+                throw new RuntimeException("Invalid path: " + path);
+            }
+
+            // for now, we do not support filters in the path
+            return new StringBuilder(path).delete(filterStartIdx, filterEndIdx + 1).toString();
```
sguilhen, line 320: We prob can change this now, no?

scim/core/src/main/java/org/keycloak/scim/resource/common/Email.java
```diff
@@ -6,12 +6,14 @@
 public class Email extends MultiValuedAttribute {
 
     public Email() {
+        setType("work");
+        setPrimary(true);
```
sguilhen, line 10: Do we want this default for the type? Is it stored anywhere when you create a new user?

scim/core/src/main/java/org/keycloak/scim/resource/schema/AbstractModelSchema.java
pedroigor: Not yet, because this filter will not really filter models, but attributes referenced in the path, in case they are multivalued. Need to work this as a follow-up.

scim/core/src/main/java/org/keycloak/scim/resource/common/Email.java
pedroigor: I think so. It is not stored anywhere. As we support only a single email, for now, the one you have is always the primary/work email. I'm open to change if needed.

sguilhen: Yeah, I was a bit more concerned about the type being set to "work", which is pretty arbitrary, IMO

pedroigor: What are you proposing then?

sguilhen (approved):

stianst (approved):

Issue #45898, opened 2026-01-30
Supported Configurations guide


vmuzikar:
### Description

Add a Supported Configurations guide that would cover the following:
* Supported JRE versions
   * Explain currently supported LTS versions of OpenJDK are supported, i.e. 17, 21, 25.
   * We could mention we test with Temurin.
* Supported DB versions
   * Move this from the DB guide and just add a link there pointing to the Supported Configurations guide.
* Supported K8s versions.
   * Should match officially currently [supported K8s versions](https://kubernetes.io/releases/).
   * We could mention the version we test with.
   * We should probably also give some statement for which K8s distros we support.
* Something else? Browsers support?

### Value Proposition

Clear communication of supported runtimes for Keycloak.

### Goals

* A new guide covering what runtimes are supported, incl. vendors and version ranges.
* Optionally, include also tested configurations where applicaple.

### Non-Goals

* Extend the support matrix beyond what we currently already support.

### Notes

⚠️ **This guide needs to be upstream only.**

vmuzikar:
CC @keycloak/cloud-native @stianst

vmuzikar:
CC @ahus1

PR #46563, opened 2026-02-24
Document how customized container images can use OpenJDK 21 when custom extensions require it

### LINKED ISSUE #46244: Document how to downgrade OpenJDK in the container image
### Description

Extend the [RPM instructions on the Container guide](https://github.com/keycloak/keycloak/blob/27fb8fae5c50885c1e1c98601db35230025b2fb9/docs/guides/server/containers.adoc?plain=1#L99) to include steps specific for downgrading to older OpenJDK version. The list of supported versions will be linked to it as part of #45898. Explicitly mention the motivation that this is intended for custom extensions that do not support newer JDKs.

Notes:
* This concerns only Keycloak image, not Operator image as we don't support extending that.
* As agreed in an offline discussion, this does not require an automated test.

michalvavrik:
* closes: https://github.com/keycloak/keycloak/issues/46244

### DIFF
--- a/docs/guides/server/containers.adoc
+++ b/docs/guides/server/containers.adoc
@@ -136,6 +136,29 @@ This approach uses a chroot, `+/mnt/rootfs+`, so that only the packages you spec
 
 WARNING: Some packages have a large tree of dependencies. By installing new RPMs you may unintentionally increase the container's attack surface. Check the list of installed packages carefully.
 
+=== Using an older OpenJDK version
+
+The {project_name} container image uses OpenJDK {jdk_version_recommended} by default, which is the recommended and tested version.
+If a custom extension does not support OpenJDK {jdk_version_recommended}, you can use an older supported OpenJDK version by installing the appropriate OpenJDK RPM and updating the `PATH` and `JAVA_HOME` environment variables.
+
+Use the two-stage pattern from the previous section to install the OpenJDK RPM.
+The following example shows how to use OpenJDK 21:
+
+[source, dockerfile]
+----
+FROM registry.access.redhat.com/ubi9 AS ubi-micro-build
+RUN mkdir -p /mnt/rootfs
+RUN dnf install --installroot /mnt/rootfs java-21-openjdk-headless --releasever 9 --setopt install_weak_deps=false --nodocs -y && \
+    dnf --installroot /mnt/rootfs clean all && \
+    rpm --root /mnt/rootfs -e --nodeps setup
+
+FROM quay.io/keycloak/keycloak
+COPY --from=ubi-micro-build /mnt/rootfs /
+
+ENV PATH=/usr/lib/jvm/jre-21/bin:$PATH
+ENV JAVA_HOME=/usr/lib/jvm/jre-21
+----
+
 === Custom ENTRYPOINT shell scripts
 
 If you use a custom entry point script, start {project_name} with `exec` so it can receive termination signals that are essential for a graceful shutdown.


mabartos:
Wouldn't be better to add arg JAVA_MAJOR_VERSION to the current Dockerfile? 

If:
```sh
podman/docker build --build-arg=JAVA_MAJOR_VERSION=21 ...
```

The `java-21-openjdk-headless` would be used instead of the `java-25-openjdk-headless`. Thats' the only difference in the Dockerfile, right? And then, this simple solution would be commented in the docs. 

cc: @keycloak/cloud-native

michalvavrik:
@mabartos it is definitely much simpler. There are things I cannot decide - e.g. 

- how many users are actually going to do it (AKA is it worth it to add such option to the docker file?) ... my personal opinion is yes
- it works now, are you sure this docker file won't change in a way that it wouldn't work with future changes? because we are not going to test it, while the custom container I presume will work as it is only extending of what is already documented ... I have no idea

let me gather opinions, I think it is a good suggestion.

michalvavrik:
Considering https://github.com/keycloak/keycloak/issues/46244 description, I don't want to make changes unless there is consensus (so that I don't have to revert it again :-) ).

mabartos:
> it works now, are you sure this docker file won't change in a way that it wouldn't work with future changes?

We don't change the Dockerfile a lot, as it always require to update also prod dockerfile that is not in sync with the other - so there's additional effort needed outside of our CN team. 

As there's a consensus of not testing this, I'd rather see the suggested changes than deal with RPM and do some workarounds. I see the current approach in this PR very fragile :/ I think much safer would be to just use the changed `java-21-openjdk-headless` in the current Dockerfile.

@michalvavrik @ASzc @vmuzikar WDYT?

michalvavrik:
> As there's a consensus of not testing this, I'd rather see the suggested changes than deal with RPM and do some workarounds. I see the current approach in this PR very fragile :/

If I were honest, I am not sure what is fragile about it. That is how you change Java on Linux and I doubt that super conservative RHEL will change it in 9.x during the period when we support JDK 21. 

But as I said before, I think your proposal is good. I'll go ahead and apply it. Thank you

michalvavrik:
@mabartos one more thing though, I started making changes and realized that your suggestion is not aligned with current container guide. It always shows how to create custom image based on existing images, while your suggestion presumes that users will have access to a Docker file that lives in Keycloak repository `quarkus/container/Dockerfile`. So your proposal is to `wget`/`curl` it from the respective branch raw file on github or something like that?

vmuzikar:
> It always shows how to create custom image based on existing images, while your suggestion presumes that users will have access to a Docker file that lives in Keycloak repository `quarkus/container/Dockerfile`.

Exactly. The only supported way to build a custom image is to extend the existing one.

docs/guides/server/containers.adoc
```diff
@@ -136,6 +136,40 @@ This approach uses a chroot, `+/mnt/rootfs+`, so that only the packages you spec
 
 WARNING: Some packages have a large tree of dependencies. By installing new RPMs you may unintentionally increase the container's attack surface. Check the list of installed packages carefully.
 
+=== Using OpenJDK 21
+
+The {project_name} container image uses OpenJDK {jdk_version_recommended} by default, which is the recommended and tested version.
+If a custom extension does not support OpenJDK {jdk_version_recommended}, use OpenJDK 21 by installing the OpenJDK 21 RPM and updating the `PATH` and `JAVA_HOME` environment variables.
```
vmuzikar: We'll support the downgrade to any of the supported OpenJDK versions. But we can update the docs together with #45898.

docs/guides/server/containers.adoc
```diff
@@ -136,6 +136,40 @@ This approach uses a chroot, `+/mnt/rootfs+`, so that only the packages you spec
...
+    rpm --root /mnt/rootfs -e --nodeps setup
+
+FROM quay.io/keycloak/keycloak AS builder
+COPY --from=ubi-micro-build /mnt/rootfs /
+
+ENV PATH=/usr/lib/jvm/jre-21/bin:$PATH
+ENV JAVA_HOME=/usr/lib/jvm/jre-21
+
+WORKDIR /opt/keycloak
+RUN /opt/keycloak/bin/kc.sh build
```
vmuzikar: This would create an optimized image. That's not desired here, we just need an image with downgraded java.

docs/guides/server/containers.adoc
```diff
@@ -136,6 +136,40 @@ This approach uses a chroot, `+/mnt/rootfs+`, so that only the packages you spec
...
+
+FROM quay.io/keycloak/keycloak
+COPY --from=ubi-micro-build /mnt/rootfs /
+COPY --from=builder /opt/keycloak/ /opt/keycloak/
+
+ENV PATH=/usr/lib/jvm/jre-21/bin:$PATH
+ENV JAVA_HOME=/usr/lib/jvm/jre-21
+
+ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
```
vmuzikar: Also, AFAIK we don't need to specify the entrypoint again if we're extending the image; the original one should be used.

vmuzikar (changes requested):
@michalvavrik Thank you for the PR. Added some comments.

docs/guides/server/containers.adoc
michalvavrik: > We'll support the downgrade to any of the supported OpenJDK versions. But we can update the docs together with #45898.

as https://github.com/keycloak/keycloak/issues/46244 parent issue is JDK 25, I thought what changed was 21, so I documented it and didn't realize that issue is also concerned with JDK 17, I'll make this docs more generic then

docs/guides/server/containers.adoc
michalvavrik: > Also, AFAIK we don't need to specify the entrypoint again if we're extending the image; the original one should be used.

Ok. That is my bad. My thinking was like this:

https://github.com/keycloak/keycloak/issues/46244 states that this documentation is for customized extensions where the downgrade is needed. https://www.keycloak.org/server/containers#_creating_a_customized_and_optimized_container_image documents exactly what I did, which is what would user have to do if they were adding custom extension. They will never really follow just steps in https://www.keycloak.org/server/containers#_installing_additional_rpm_packages because they need to install that custom extension and actually build Keycloak.

I am going to follow your suggestion, test it and update this PR.

michalvavrik: (that was more of a comment between your comments of `RUN /opt/keycloak/bin/kc.sh build` and this one, it doesn't really fit either one completely)

vmuzikar: @michalvavrik Thank you. Just to clarify, adding extensions doesn't require creating an optimized image (with `kc.sh build`), it's a separate thing.

michalvavrik: > @michalvavrik Thank you. Just to clarify, adding extensions doesn't require creating an optimized image (with `kc.sh build`), it's a separate thing.

got it, thanks

mabartos:
> It always shows how to create custom image based on existing images, while your suggestion presumes that users will have access to a Docker file that lives in Keycloak repository quarkus/container/Dockerfile.

@michalvavrik @vmuzikar Yeah, alright, sorry, didn't have the full context :)

michalvavrik:
> > It always shows how to create custom image based on existing images, while your suggestion presumes that users will have access to a Docker file that lives in Keycloak repository quarkus/container/Dockerfile.
> 
> @michalvavrik @vmuzikar Yeah, alright, sorry, didn't have the full context :)

No worries, I thought it was smart comment.

docs/guides/server/containers.adoc
michalvavrik: done

docs/guides/server/containers.adoc
michalvavrik: done

docs/guides/server/containers.adoc
michalvavrik: done

vmuzikar (approved):
LGTM, thank you @michalvavrik.

@ASzc Do you want to review too?

ASzc (approved):
Seems reasonable

PR #46577, opened 2026-02-24
Check the Unicode setup of the database on startup

### LINKED ISSUE #46557: Keycloak should check the Unicode setup of the database on startup
### Description

The Keycloak docs have extensive docs on how to configure a database for Unicode. 

https://www.keycloak.org/server/db#_configuring_unicode_support_for_the_database

Keycloak should check that the database is configured correctly for Unicode at startup. 

* Oracle
* MySQL/MariaDB
* PostgreSQL

With Keycloak 26.x, it should log a warning if it detects a mismatch on each startup, as it might break existing installations. We had a similar issue here #42300

We would add a deprecation note to the docs that other character sets are deprecated. 

For KC 27, we would have a follow-up task to have Keycloak refuse from creating a new schema in an empty database when it detects a wrong character set to avoid people missing the warning. 

In addition, any recommended JDBC connection properties should be applied by Keycloak automatically. See #46556 for a similar issue.

If it might make sense to break this issue down by database, please do so before starting the work.

### Value Proposition

Make it simple to run Keycloak the right way, and to avoid problems once the first user adds a special character in their username.

### Goals

* Show a warning at startup when there is a mismatch
* Set JDBC connection parameters automatically
* Deprecate non-UTF8 databases
* Create a follow-up issue for KC 27 to avoid creating new installations with a wrong DB charset

### Non-Goals

?

### Discussion

_No response_

### Notes

_No response_

pruivo:
Closes #46557

### DIFF
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -169,12 +169,10 @@ The default value is 5 minutes, which is the same value used by the current {pro
 * `transaction-migration-timeout`: Controls the timeout for transactions responsible for migrating the database schema to a new version.
 The default value is 30 minutes, as schema migrations may take a while to complete for large tables.
 
-=== Database charset checks during startup
+=== Database character encoding checks during startup
 
 {project_name} will log a warning if the database is not configured to use UTF-8 encoding.
 This check applies to all supported databases.
-
-In the future, we may make it a hard requirement and {project_name} will not start if the proper encoding is not configured.
 
 For details on configuring UTF-8 encoding for your database, see the https://www.keycloak.org/server/db[Configuring the database] guide.
 
@@ -208,6 +206,12 @@ The feature link:{securing_apps_token_exchange_link}#_legacy-token-exchange[Lega
 The methods `getUserSessionsStream(RealmModel realm, ClientModel client)`, `getUserSessionsStream(RealmModel realm, ClientModel client, Integer firstResult, Integer maxResults)`, and `getOfflineUserSessionsStream(RealmModel realm, ClientModel client, Integer firstResult, Integer maxResults)` are deprecated for removal.
 They are replaced with read-only alternatives to reduce memory consumption.
 See the `UserSessionProvider` API changes section in notable changes for further information.
+
+=== Non-UTF-8 database character encoding deprecated
+
+Running {project_name} on a database that is not using UTF-8 as a charset is now deprecated.
+
+In the current version {project_name} will log a warning if the database is not configured to use UTF-8 encoding. Future versions might refuse to work with database that is not configured to use UTF-8 encoding.
 
 // ------------------------ Removed features ------------------------ //
 == Removed features
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -178,6 +178,12 @@ In the future, we may make it a hard requirement and {project_name} will not sta
 
 For details on configuring UTF-8 encoding for your database, see the https://www.keycloak.org/server/db[Configuring the database] guide.
 
+As part of this change, the database configuration documentation has been updated:
+
+* For MySQL and MariaDB, the documentation now recommends using `utf8mb4` as the character set.
+The previously documented `utf8mb3` (or `utf8`) character set has been removed from the documentation due to its limitations in storing certain Unicode characters.
+* Previously recommended JDBC driver settings for Oracle, MySQL, and MariaDB have been removed from the documentation, as current versions of these databases use appropriate default values.
+
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
 
--- a/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
+++ b/docs/documentation/upgrading/topics/changes/changes-26_6_0.adoc
@@ -169,6 +169,15 @@ The default value is 5 minutes, which is the same value used by the current {pro
 * `transaction-migration-timeout`: Controls the timeout for transactions responsible for migrating the database schema to a new version.
 The default value is 30 minutes, as schema migrations may take a while to complete for large tables.
 
+=== Database charset checks during startup
+
+{project_name} will log a warning if the database is not configured to use UTF-8 encoding.
+This check applies to all supported databases.
+
+In the future, we may make it a hard requirement and {project_name} will not start if the proper encoding is not configured.
+
+For details on configuring UTF-8 encoding for your database, see the https://www.keycloak.org/server/db[Configuring the database] guide.
+
 // ------------------------ Deprecated features ------------------------ //
 == Deprecated features
 
--- a/docs/guides/server/db.adoc
+++ b/docs/guides/server/db.adoc
@@ -257,8 +257,8 @@ a|
 ----
 CREATE DATABASE keycloak
   ENCODING 'UTF8'
-  LC_COLLATE 'en_US.UTF-8'
-  LC_CTYPE 'en_US.UTF-8';
+  LC_COLLATE 'en_US.utf8'
+  LC_CTYPE 'en_US.utf8';
 ----
 
 |===
--- a/docs/guides/server/db.adoc
+++ b/docs/guides/server/db.adoc
@@ -212,46 +212,58 @@ The database schema provides support for Unicode strings only for the following 
 
 * Descriptions of objects
 
-Otherwise, characters are limited to those contained in database encoding, which is often 8-bit. However, for some database systems, you can enable UTF-8 encoding of Unicode characters and use the full Unicode character set in all text fields. For a given database, this choice might result in a shorter maximum string length than the maximum string length supported by 8-bit encodings.
+Otherwise, characters are limited to those contained in database encoding, which is often 8-bit.
+However, for some database systems, you can enable UTF-8 encoding of Unicode characters and use the full Unicode character set in all text fields.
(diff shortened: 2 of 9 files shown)


pruivo:
@ahus1 Am I overcomplicating things by using Liquibase? It shows the warning as follows (not very clear):

```
[kc-server] 2026-02-24 17:54:48,144 INFO  [org.keycloak.quarkus.runtime.storage.database.liquibase.QuarkusJpaUpdaterProvider] (main) Initializing database schema. Using changelog META-INF/jpa-changelog-master.xml
[kc-server] 2026-02-24 17:54:52,036 WARN  [liquibase.precondition.core.PreconditionContainer] (main) Executing META-INF/jpa-changelog-26.6.0.xml::26.6-check-utf8-postgres::keycloak despite precondition failure due to onFail='WARN':
[kc-server]       1 preconditions failed
[kc-server]      META-INF/jpa-changelog-master.xml : SQL Precondition failed.  Expected 'UTF8' got 'LATIN1'
[kc-server] 
```

ahus1:
Thank you for this POC. 

The possible issues I see with that: 

* The warning only appears once with this POC. This could be changed so the changeset is executed on every migration.
* The warning only appears when there is a DB migration, not on every startup. So I'd say it could be easily missed. 
* The warning is not very descriptive, and I doubt that it will ever be when we use Liquibase for that.

A check like 

https://github.com/keycloak/keycloak/blob/eff97618ef117214bf02d95811bc7e182081e6dc/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/storage/database/jpa/QuarkusJpaConnectionProviderFactory.java#L303

might be able to provide more descriptive messages. 

One could argue that this defers the startup of Keycloak unless we put this to another thread.

pruivo:
> The possible issues I see with that:
> 
> * The warning only appears once with this POC. This could be changed so the changeset is executed on every migration.
> * The warning only appears when there is a DB migration, not on every startup. So I'd say it could be easily missed.

@ahus1, those are my intentions. Why do we have to check the encoding every time? During rolling updates, restarts, and scale-ups, it would check something that has already been validated.

ahus1:
>  Why do we have to check the encoding every time?

I'd say the warning is easily missed if appears only on the first starting up node with a new version, and then no more. 

I'd like to see something red and flashing on the Admin console, or something logging it every minute into the log that you're doing something that you'll soon regret. I'd love to break the startup of the server, but that would unfortunately be a breaking change. 

If you want to run it only once with every migration, there might be a place here: 

https://github.com/keycloak/keycloak/blob/17f0dbdc1c0121f2f6a255b578040eb957981a7e/model/storage-private/src/main/java/org/keycloak/storage/datastore/DefaultMigrationManager.java#L162

Let's discuss  it on one of our team meetings in more detail.

pruivo:
@ahus1 sure, let's discuss it tomorrow.

I've tested the following vendors

### MSSQL

```
[kc-server] 2026-02-24 20:38:41,128 WARN  [org.keycloak.quarkus.runtime.storage.database.jpa.QuarkusJpaConnectionProviderFactory] (main) Invalid mssql charset encoding 'SQL_Latin1_General_CP1_CI_AS'. It is recommended to use UTF8 based encoding
```

### MYSQL

```
[kc-server] 2026-02-24 20:39:33,852 WARN  [org.keycloak.quarkus.runtime.storage.database.jpa.QuarkusJpaConnectionProviderFactory] (main) Invalid mysql charset encoding 'utf8mb4'. It is recommended to use utf8mb3
```

Keycloak docs say `utf8mb4` [character set is not supported due to different storage requirements](https://www.keycloak.org/server/db#_configuring_unicode_support_for_a_mysqlmariadb_database), but it is the recommended since 2017 (https://dev.mysql.com/blog-archive/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/)
Is the documentation outdated for the last decade?

### PostgreSQL

```
[kc-server] 2026-02-24 20:40:38,105 WARN  [org.keycloak.quarkus.runtime.storage.database.jpa.QuarkusJpaConnectionProviderFactory] (main) Invalid postgresql charset encoding 'LATIN1'. It is recommended to use UFT8
```

### Oracle

I haven't installed this DB.

ahus1:
@pruivo - about MySQL and `utf8mb4`: We are probably outdated, still it might lead to some additional investigation, like for example: What charset are we testing with when running MySQL in our tests, and what problems people might have been reporting in the past. 

I know that @vramik was investigating some problems in #45597. Maybe in this issue, one could log a warning if it is neither of the two UTF8 character sets, and have a follow-up issue to look at MySQL in more detail and start supporting the newer charset. 

For example, there some parts in our setup that already help with supporting that, as we're indexing only the first 255 characters on longer fields, see below for an example. 

https://github.com/keycloak/keycloak/blob/b65a60e40d433196c9dd74c829b9530777bb4747/model/jpa/src/main/resources/META-INF/jpa-changelog-14.0.0.xml#L79-L82

vramik:
Hello, thank you for tagging me. I can provide few notes about deprecated `utf8mb3` and `utf8mb4`. 

The reason why the `utf8mb4` was deprecated was the 767 bytes index size limit in MySQL 5. Since MySQL 8.0+ there is by default enabled DYNAMIC Row Format which supports 3072-byte index prefix limit. 

Since Keycloak supports MySQL 8+ it should then be safe to use `utf8mb4`, I was briefly testing it and I did not hit any issues with mb4. But I was not testing any migration scenarios.

pruivo:
@ahus1 Since MySQL 8.0, the default is `utf8mb4`. Looking at the test logs from this PR, it is testing with `utf8mb4`.
```
2026-02-24 21:10:45,110 INFO  [testinfo] (main) ----------------------------------------------------------------
2026-02-24 21:10:45,110 INFO  [testinfo] (main)    RUNNING - org.keycloak.tests.admin.AdminClientTest
2026-02-24 21:10:59,110 WARN  [managed.keycloak] (Thread-45) [org.keycloak.quarkus.runtime.storage.database.jpa.QuarkusJpaConnectionProviderFactory] (main) Invalid mysql charset encoding 'utf8mb4'. It is recommended to use 'utf8mb3' encoding
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 23.32 s -- in org.keycloak.tests.admin.AdminClientTest
```

pruivo:
https://endoflife.date/mysql The support for MySQL 5.x ended 2 years ago. Do you think users are still using it?

pruivo:
@ahus1 "I've" added the documentation changes. Let me know what you think. I removed any reference to the properties, as I found out most of them are outdated.

ahus1:
@pruivo - We officially support only MySQL 8+, so we should be fine. 

I also found the pointer that the INNODB page size in MySQL 8 is by default 16k, which leads to a maximum key size of 3072B, which translates to 768 characters in `utf8mb4`. As we truncate keys to 256 chars, we should be fine. This leads to the same calculation @vramik mentioned earlier. 

This boils down to that people should use the MySQL defaults and should not deviate from them. 

Maybe we could in a follow-up issue have a check on schema migration which checks some additional database parameters (like the using the DYNAMIC row format). I remember there was another setting in the oracle database for a "legacy table names", which then lead Oracle only allowing short table names that lead to failing migrations.

pruivo:
@ahus1 I'm not sure if I want to open another can of worms. Also, I'm not sure about your math; the documentation says `For example, utf8mb4 characters can require up to four bytes per character, so a VARCHAR column that uses the utf8mb4 character set can be declared to be a maximum of 16,383 characters`. As you say, we limit to 255 (or 256) characters, so it should be fine. The testsuite is using `utf8mb4` as it is the default with MySQL 8.4.

About mb3 vs mb4, the first one is deprecated. The documentation says:
```
The recommended character set for MySQL is utf8mb4. All new applications should use utf8mb4.
The utf8mb3 character set is deprecated. utf8mb3 remains supported for the lifetimes of the MySQL 8.0.x and MySQL 8.4.x LTS release series.
Expect utf8mb3 to be removed in a future major release of MySQL.
```

ahus1 (changes requested):
Thank you @pruivo for looking into this, the changes look good. 

Looking at the logs of our test suite, I spotted some places where it now states "Invalid mariadb charset" or "Invalid mssql charset encoding" in the "Store IT" tests for different databases. Mit might then affect both the new and the old test suite. 

Please also update the BasicDatabaseTest for Quarkus Distribution to catch if there are any databases started with not supported character sets.

pruivo:
@ahus1 this is so confusing... I'm using the same image (11.8.6) as testcontainer, and I get `utfmb4` encoding. And [MariaDB announced](https://mariadb.org/11-8-lts-released/) that `MariaDB previously defaulted to latin1 for character sets. Starting from 11.8, the default is utf8mb4 — modern Unicode with full emoji support 😉.`
I have no clue what's going on. 

> Please also update the BasicDatabaseTest for Quarkus Distribution to catch if there are any databases started with not supported character sets.

I didn't understand your idea. Do you want a test to fail if it's not using the UTF8 charset?

pruivo:
found the problem

ahus1:
> Please also update the BasicDatabaseTest for Quarkus Distribution to catch if there are any databases started with not supported character sets.

Let me rephrase: The child tests of BasicDatabaseTest for the different databases do not log on the console, therefore I can't verify from the logs of the test runs if there are similar error messages about a wrong character encoding. Therefore I ask you: Please take a look. Maybe add a ~test~ assertion to the successful startup test so it is applied to all the tests. 

https://github.com/keycloak/keycloak/blob/71672629098d32ceee712a168f93f05e8a1134df/quarkus/tests/integration/src/test/java/org/keycloak/it/storage/database/BasicDatabaseTest.java#L49-L51

ahus1 (approved):
Thank you for diving down into this rabbit hole of encoding. The updated docs and the warning should make it simpler for Keycloak users to use a correct configuration.

PR #34346, opened 2024-10-25
Remove online session when offline access is requested as the first request

rmartinc:
Closes #34001

The normal code to token flow removes the online session if the offline access was requested in the first interaction. So, now a client that request offline access in the first interaction just creates the offline session. The steps to do it are the following:

1. The authorization request is done with scope including the offline_access.
2. Online session is created and the client session is marked as it was for offline first access (note).
3. The client obtains the token and performs the code2token to the token endpoint.
4. The keycloak server when creating the offline session detects the session was marked as first offline access and the online session is removed.
5. Any other use of the session removes the note and avoids the online session removal.

There are tests that are affected and probably real use-cases were affected too. The `BackchannelLogoutTest` is mainly modified because now some tests are very different. There are some methods that configures two IdPs: subconsumer -> consumer -> provider. The first IdP in subconsumer to consumer realm configures offline_access in the scope. The new behavior changes completely the tests because now the online session in the consumer realm is not created and that changes those tests completely.

Draft for the moment. If we decide to continue with this some upgrading notes will be needed for sure.

PR #46578, opened 2026-02-24
Remove root auth session on first offline_access to prevent cross-user session contamination

### LINKED ISSUE #46239: AUTH_SESSION_ID cookie reuse causes cross-user session contamination on re-authentication
### Before reporting an issue

### Area

authentication

### Describe the bug

When a mobile app uses `scope=offline_access` with the authorization code flow, cross-user session contamination occurs due to root authentication session ID reuse.

`AUTH_SESSION_ID` is a session cookie that persists in the browser between authorization requests. When a second user starts a new login and the browser sends a stale `AUTH_SESSION_ID` cookie from a previous login, Keycloak reuses the same root authentication session ID , **without checking whether the authenticating user is different**. This happens in [`AuthorizationEndpointBase.createAuthenticationSession()`](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/protocol/AuthorizationEndpointBase.java#L173-L225), whenever the root auth session referenced by the cookie still exists server-side.

Since the root auth session ID is used as the user session ID and the `sid` claim , the new user's tokens get bound to the previous user's session.

Concretely:
1. **User A** logs in with `offline_access` → Keycloak creates an offline session with `sid=UUID-A`. The online session is removed (per [#34001](https://github.com/keycloak/keycloak/issues/34001)), but the `AUTH_SESSION_ID=UUID-A` cookie remains in the browser.
2. **User B** logs in shortly after → the browser sends the stale `AUTH_SESSION_ID=UUID-A` cookie → Keycloak finds the still-existing root auth session and reuses `UUID-A` → User B's offline session is created with `sid=UUID-A`
3. **At token exchange**: User B receives tokens with `sub=UserB` (correct) but `sid=UUID-A` — the tokens *appear* correct
4. **At token refresh**: Keycloak resolves the offline session by `sid=UUID-A` → finds User A's offline session → returns User A's identity (`sub`, `preferred_username`, all claims)

**The bug is independent of `prompt=login`** — it occurs whenever the stale `AUTH_SESSION_ID` cookie is sent and the root auth session still exists server-side.

### Version

26.4.6

### Regression

### Expected behavior

1. When a stale `AUTH_SESSION_ID` cookie references an existing root auth session, Keycloak should not blindly reuse the same ID — it should create a new root auth session with a fresh UUID
2. A `sid` should never be shared across different users 
3. The `DIFFERENT_USER_AUTHENTICATED` check should cover the `offline_access` scenario (where the online session has been removed)

The existing check for `DIFFERENT_USER_AUTHENTICATED` in [`AuthenticationProcessor.attachSession()`](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/AuthenticationProcessor.java#L1150-L1215) only triggers when the old user session still exists and is valid. With `offline_access` (which removes the online session per #34001), this check is bypassed , Keycloak creates a **new** user session with the **old** root auth session ID.

### Actual behavior

1. User A authenticates with `scope=offline_access` → online session is removed server-side, but the `AUTH_SESSION_ID` session cookie remains in the browser (expected for a session cookie)
2. ~16 seconds later, User B authenticates → browser sends User A's `AUTH_SESSION_ID` cookie → Keycloak reuses the same `sid` instead of generating a new one
3. User B's tokens contain `sub=UserB` (correct) but `sid=UserA's session` (wrong)
4. Refreshing User B's token returns User A's identity (`sub`, `preferred_username`, all claims)

### How to Reproduce?

**Environment:** Android device, OIDC public client, Chrome Custom Tabs (via AppAuth / `react-native-app-auth`)

**Client configuration:**
- Scopes: `openid profile email offline_access`
- Additional parameters: `prompt=login`, `max_age=0` (optional — bug also occurs without these)

**Steps:**

1. User A opens the authorization endpoint in a Chrome Custom Tab and authenticates
2. App receives tokens (access token, offline refresh token, id token). With `offline_access`, Keycloak does not create a persistent SSO session in the browser (as documented: "the SSO session is not created in the browser if the login request with scope=offline_access is used and the user was not already authenticated in the SSO beforehand")
3. **Within ~60 seconds**, User B opens the authorization endpoint in a new Chrome Custom Tab (e.g. via an "Add account" flow in the app)
4. User B authenticates with their own credentials , Keycloak shows the login page (not "Continue as User A") because there is no SSO session
5. Decode User B's ID token → `sid` matches User A's session ID
6. Use User B's refresh token → the refreshed tokens contain User A's `sub` and claims

**Why intermittent:** The bug requires two conditions:
1. The browser still holds the stale `AUTH_SESSION_ID` cookie (Chrome Custom Tabs share cookies with Chrome; the cookie is cleared when Android kills Chrome's process)
2. The root authentication session referenced by the cookie still exists server-side

In our tests, the bug consistently reproduced when User B logged in within ~60 seconds of User A's login. After ~60 seconds, the root auth session was no longer found server-side: Keycloak created a new root auth session with a fresh UUID and wrote a **new** `AUTH_SESSION_ID` cookie, and the bug did not occur.

**Why iOS is not affected:** iOS has `prefersEphemeralWebBrowserSession` (ASWebAuthenticationSession), which creates an isolated cookie jar per session. Android Chrome Custom Tabs have no equivalent.

#### Evidence from HAR Analysis

**Constraint:** The following data is extracted from a reproduction HAR file. Sensitive tokens have been redacted, but the UUID correlation and timestamps are preserved to demonstrate the vulnerability.

**Timeline:**

### T0 (+0s): User A Login (Offline Access)
* **Timestamp:** `2026-02-11T10:58:15.369Z`
* **Flow:** Authorization Code with `scope=offline_access`
* **Response (ID Token Claims):**
  * `sub`: `[User-A-UUID]...a613c0492eb5`
  * `sid`: `40e0432d-de84-43af-9b42-f8dbd008012d`
  * `iat`: `1770807495`

### T1 (+14s): User B Login (New Authentication)
* **Timestamp:** `2026-02-11T10:58:29.324Z`
* **Context:** Browser sends stale `AUTH_SESSION_ID` cookie matching T0 (cookie persisted after User A's online session removal).
* **Flow:** Authorization Code with `scope=offline_access`
* **Response (ID Token Claims):**
  * `sub`: `[User-B-UUID]...8ad8b6012a62`  *(Different User)*
  * `sid`: `40e0432d-de84-43af-9b42-f8dbd008012d`  **<-- CRITICAL: Identical to User A's session**
  * `iat`: `1770807508`

### T2: Token Refresh (User B)
* **Action:** App uses User B's offline refresh token.
* **Mechanism:** Keycloak resolves the offline session by the shared `sid` (`...8012d`).
* **Outcome:** The refreshed tokens contain User A's identity (`sub` matches User A), resulting in a full account takeover.

This may be a variant of [CVE-2022-3916](https://github.com/keycloak/keycloak/security/advisories/GHSA-97g8-xfvw-q4hg) / [CVE-2025-12390](https://github.com/advisories/GHSA-rg35-5v25-mqvp), through a code path that appears to remain unpatched.

I traced the issue through three locations in the code:

**1. `AuthorizationEndpointBase.createAuthenticationSession()`** reuses the root auth session from the `AUTH_SESSION_ID` cookie when it still exists:

```java
RootAuthenticationSessionModel rootAuthSession = manager.getCurrentRootAuthenticationSession(realm);
if (rootAuthSession != null) {
    authSession = rootAuthSession.createAuthenticationSession(client);
}
```

This makes sense for multi-tab SSO ([#23879](https://github.com/keycloak/keycloak/issues/23879)), but it becomes a problem when the root auth session outlives the user session it was created for.

**2. `AuthenticationProcessor.attachSession()`** has a `DIFFERENT_USER_AUTHENTICATED` check, but it's only reached when the user session exists and is valid:

```java
if (userSession == null) { // if no authenticator attached a usersession

    userSession = session.sessions().getUserSession(realm, authSession.getParentSession().getId());
    if (userSession == null) {
        // creates a new user session with the root auth session ID — no user check here
        userSession = new UserSessionManager(session).createUserSession(
            authSession.getParentSession().getId(), realm, authSession.getAuthenticatedUser(), ...);
    } else if (userSession.getUser() == null || !AuthenticationManager.isSessionValid(realm, userSession)) {
        userSession.restartSession(...);
    } else {
        // this is the safety net, but we never get here when the online session was already removed
        if (!authSession.getAuthenticatedUser().equals(userSession.getUser())) {
            event.detail(Details.EXISTING_USER, userSession.getUser().getId());
            event.error(Errors.DIFFERENT_USER_AUTHENTICATED);
            throw new ErrorPageException(..., Messages.DIFFERENT_USER_AUTHENTICATED, ...);
        }
    }
}
```

With `offline_access`, the online session was already removed (see next point), so `getUserSession()` returns `null`, we fall into the first branch, and a new user session is created with the old root auth session ID — the same ID as User A's offline session. The `DIFFERENT_USER_AUTHENTICATED` check never runs.

The `FIRST_OFFLINE_ACCESS` note is set in the same method:

```java
if (userSession.getStarted() == userSession.getLastSessionRefresh()) {
    final int clientSessions = userSession.getAuthenticatedClientSessions().size();
    clientSessionCtx = TokenManager.attachAuthenticationSession(session, userSession, authSession);
    if (clientSessions == 0 && TokenUtil.hasScope(clientSessionCtx.getScopeString(), OAuth2Constants.OFFLINE_ACCESS)) {
        clientSessionCtx.getClientSession().setNote(FIRST_OFFLINE_ACCESS, Boolean.TRUE.toString());
    }
}
```

**3. `OAuth2GrantTypeBase.createTokenResponse()`** removes the online session when `FIRST_OFFLINE_ACCESS` is set (introduced in [PR #34346](https://github.com/keycloak/keycloak/pull/34346)), but does not remove the root auth session:

```java
if (useRefreshToken) {
    responseBuilder.generateRefreshToken();
    if (TokenUtil.TOKEN_TYPE_OFFLINE.equals(responseBuilder.getRefreshToken().getType())
            && clientSessionCtx.getClientSession().getNote(AuthenticationProcessor.FIRST_OFFLINE_ACCESS) != null) {
        // the online session can be removed if first created for offline access
        session.sessions().removeUserSession(realm, userSession);
        // root authentication session is NOT removed here
    }
}
```

This is the gap. There are three code paths that remove the online user session, but only two also clean up the root auth session:

- Backchannel logout ([#32197](https://github.com/keycloak/keycloak/issues/32197), [commit `5344aad`](https://github.com/keycloak/keycloak/commit/5344aada5ee06b02ec3a9e0f52fa381d085b6282)) — removes root auth session ✓
- Logout endpoint ([#43853](https://github.com/keycloak/keycloak/issues/43853), [PR #43872](https://github.com/keycloak/keycloak/pull/43872)) — removes root auth session ✓
- `FIRST_OFFLINE_ACCESS` ([PR #34346](https://github.com/keycloak/keycloak/pull/34346)) — does NOT remove root auth session ✗

The root auth session then survives for ~60 seconds (TTL from [#23880](https://github.com/keycloak/keycloak/issues/23880)) and can be reused by a different user.

**Suggested fix:** remove the root auth session in `OAuth2GrantTypeBase.createTokenResponse()` when removing the online session for `FIRST_OFFLINE_ACCESS`. The user session ID equals the root auth session ID (created from `authSession.getParentSession().getId()` in `attachSession()`):

```java
if (TokenUtil.TOKEN_TYPE_OFFLINE.equals(responseBuilder.getRefreshToken().getType())
        && clientSessionCtx.getClientSession().getNote(AuthenticationProcessor.FIRST_OFFLINE_ACCESS) != null) {
    session.sessions().removeUserSession(realm, userSession);
    // also remove the root auth session to prevent reuse by a different user
    RootAuthenticationSessionModel rootAuthSession = session.authenticationSessions()
        .getRootAuthenticationSession(realm, userSession.getId());
    if (rootAuthSession != null) {
        session.authenticationSessions().removeRootAuthenticationSession(realm, rootAuthSession);
    }
}
```

This would be consistent with the backchannel logout fix ([commit `5344aad`](https://github.com/keycloak/keycloak/commit/5344aada5ee06b02ec3a9e0f52fa381d085b6282)) and logout endpoint fix ([PR #43872](https://github.com/keycloak/keycloak/pull/43872)).

I think the underlying invariant is the same across all three paths: when an online user session is removed, the associated root authentication session should be removed too, otherwise it can be reused by a different user. The `FIRST_OFFLINE_ACCESS` path is arguably easier to hit than the logout-based vectors since it requires no explicit logout  it just happens on every first `offline_access` login.

federicobartoli:
## Summary

When a user authenticates with `scope=offline_access`, the `FIRST_OFFLINE_ACCESS` path removes the online user session but **not** the associated root authentication session. This leaves the `AUTH_SESSION_ID` cookie pointing to a still-valid root auth session. If a second user authenticates from the same browser within the session TTL, Keycloak reuses the stale root auth session and creates a new user session with the same ID, allowing the second user to inherit the first user's session.

This PR removes the root authentication session together with the online user session in the `FIRST_OFFLINE_ACCESS` code path, consistent with backchannel logout and the logout endpoint.

Closes #46239

## Changes

- **`OAuth2GrantTypeBase.java`** — After removing the online user session, also remove the root authentication session to prevent `AUTH_SESSION_ID` cookie reuse
- **`OfflineTokenTest.java`** — Regression test that verifies the root auth session is cleaned up and a second user gets an isolated session

## Test plan

- [x] `OfflineTokenTest#offlineTokenCrossUserSessionIsolation` passes with the fix
- [x] Same test **fails** without the fix (genuine regression detector)
- [x] Spotless formatting passes
- [x] Rebased on latest `main`

### DIFF
--- a/services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
+++ b/services/src/main/java/org/keycloak/protocol/oidc/grants/OAuth2GrantTypeBase.java
@@ -64,6 +64,7 @@ import org.keycloak.services.clientpolicy.ClientPolicyException;
 import org.keycloak.services.cors.Cors;
 import org.keycloak.services.util.AuthorizationContextUtil;
 import org.keycloak.services.util.MtlsHoKTokenUtil;
+import org.keycloak.sessions.RootAuthenticationSessionModel;
 import org.keycloak.util.TokenUtil;
 
 import org.jboss.logging.Logger;
@@ -125,6 +126,13 @@ public abstract class OAuth2GrantTypeBase implements OAuth2GrantType {
                     && clientSessionCtx.getClientSession().getNote(AuthenticationProcessor.FIRST_OFFLINE_ACCESS) != null) {
                 // the online session can be removed if first created for offline access
                 session.sessions().removeUserSession(realm, userSession);
+                // also remove the root authentication session to prevent AUTH_SESSION_ID cookie reuse by a different user
+                // consistent with backchannel logout and logout endpoint which both clean up root auth sessions
+                logger.tracef("Removing root authentication session '%s' after first offline access", userSession.getId());
+                RootAuthenticationSessionModel rootAuthSession = session.authenticationSessions().getRootAuthenticationSession(realm, userSession.getId());
+                if (rootAuthSession != null) {
+                    session.authenticationSessions().removeRootAuthenticationSession(realm, rootAuthSession);
+                }
             }
         } else {
             TokenContextEncoderProvider encoder = session.getProvider(TokenContextEncoderProvider.class);
--- a/tests/base/src/test/java/org/keycloak/tests/oauth/OfflineTokenCrossUserSessionIsolationTest.java
+++ b/tests/base/src/test/java/org/keycloak/tests/oauth/OfflineTokenCrossUserSessionIsolationTest.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2026 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.keycloak.tests.oauth;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.representations.AccessToken;
+import org.keycloak.representations.RefreshToken;
+import org.keycloak.testframework.annotations.InjectRealm;
+import org.keycloak.testframework.annotations.KeycloakIntegrationTest;
+import org.keycloak.testframework.oauth.OAuthClient;
+import org.keycloak.testframework.oauth.annotations.InjectOAuthClient;
+import org.keycloak.testframework.realm.ManagedRealm;
+import org.keycloak.testframework.realm.RealmConfig;
+import org.keycloak.testframework.realm.RealmConfigBuilder;
+import org.keycloak.testframework.remote.runonserver.InjectRunOnServer;
+import org.keycloak.testframework.remote.runonserver.RunOnServerClient;
+import org.keycloak.testframework.ui.annotations.InjectPage;
+import org.keycloak.testframework.ui.page.LoginPage;
+import org.keycloak.testsuite.util.oauth.AccessTokenResponse;
+import org.keycloak.testsuite.util.oauth.AuthorizationEndpointResponse;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Regression test for https://github.com/keycloak/keycloak/issues/46239
+ *
+ * Verifies that the root authentication session is removed together with the
+ * online user session on the FIRST_OFFLINE_ACCESS code path, preventing a
+ * second user from inheriting the first user's session via a stale
+ * AUTH_SESSION_ID cookie.
+ *
+ * Note: this test is a true regression detector only when run with
+ * {@code KC_TEST_SERVER=embedded}. In distribution mode the external
+ * Infinispan cache removes the root auth session independently, so the
+ * test passes regardless of the fix.
+ */
+@KeycloakIntegrationTest
+public class OfflineTokenCrossUserSessionIsolationTest {
+
+    @InjectRealm(config = OfflineTokenRealmConfig.class)
+    ManagedRealm realm;
+
+    @InjectOAuthClient
+    OAuthClient oauth;
+
+    @InjectPage
+    LoginPage loginPage;
+
+    @InjectRunOnServer(permittedPackages = "org.keycloak.tests.oauth")
+    RunOnServerClient runOnServer;
+
+    public static class OfflineTokenRealmConfig implements RealmConfig {
+
+        @Override
+        public RealmConfigBuilder configure(RealmConfigBuilder realm) {
+            realm.eventsEnabled(true);
+
+            realm.addClient("offline-client")
+                    .secret("secret1")
+                    .redirectUris("*")
+                    .directAccessGrantsEnabled(true);
+
+            realm.addUser("test-user@localhost")
+                    .name("Test", "User")
+                    .email("test-user@localhost")
+                    .emailVerified(true)
+                    .password("password")
+                    .roles("offline_access");
+
+            realm.addUser("offline-user2@localhost")
+                    .name("Offline", "User2")
+                    .email("offline-user2@localhost")
+                    .emailVerified(true)
+                    .password("password")
+                    .roles("offline_access");
+
+            return realm;
+        }
+    }
+
+    // https://github.com/keycloak/keycloak/issues/46239
+    @Test
+    public void offlineTokenCrossUserSessionIsolation() {
+        oauth.client("offline-client", "secret1").scope(OAuth2Constants.OFFLINE_ACCESS);
+
+        // User A logs in
+        AuthorizationEndpointResponse loginResponseA = oauth.doLogin("test-user@localhost", "password");
+        String codeA = loginResponseA.getCode();
+        String sessionStateA = loginResponseA.getSessionState();
+
+        // Re-create the root auth session WITH a tab to simulate production (Infinispan)
+        // where the root auth session survives after the auth flow.
+        // The embedded store auto-removes root sessions when the last tab is consumed,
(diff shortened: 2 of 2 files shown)


federicobartoli:
Hi @rmartinc , pinging you directly since you worked on the FIRST_OFFLINE_ACCESS implementation in PR #34346. It seems that while the online user session is correctly removed there, the root authentication session is left behind, which can lead to cross-user session contamination (as detailed above). I've put together this PR following the same logic used for backchannel logouts. Would love your eyes on it when you have a moment!

rmartinc (approved):
Thanks @federicobartoli! Very good catch. I'm approving and asking @mposolda to run the tests. Let's see if no test is broken.

federicobartoli:
Hi @mposolda , I've just force-pushed an update migrating the OfflineTokenTest test to the new Test Framework as requested. Could you please trigger the CI checks? Thanks!

rmartinc:
@federicobartoli It was not needed, for these issues that need backport using the old TS is OK IMHO (it's even easier for us). @mposolda You can go with the new test, I would use the old version for the backports.

federicobartoli:
> @federicobartoli It was not needed, for these issues that need backport using the old TS is OK IMHO (it's even easier for us). @mposolda You can go with the new test, I would use the old version for the backports.

No worries at all! 😅 If it makes the backporting process easier for you, I can absolutely revert the commit and bring back the old test suite. Just say the word! Let me know what works best for you guys!

mposolda (approved):
@federicobartoli @rmartinc Thanks!

PR #46592, opened 2026-02-25
Organization Groups - Identity Provider Mappers

### LINKED ISSUE #45512: Organization Groups - Identity Provider Mappers
### Description

Update Identity Provider (IdP) mappers to support Organization Groups.

**Scope & Notes**

- **Hardcoded & Advanced Mappers**: updates to allow administrators to select Organization-specific groups when configuring IDP mappers.
- **Validation**: Ensure that the group selected in the mapper actually belongs to the Organization linked to the IdP.
- **Lifecycle**: Determine behavior when an IdP is unlinked from an Org (sanitize mappers vs. runtime checks).

martin-kanis:
Closes #45512

### DIFF
--- a/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/OrganizationIdentityProviderResource.java
+++ b/integration/admin-client/src/main/java/org/keycloak/admin/client/resource/OrganizationIdentityProviderResource.java
@@ -17,12 +17,18 @@
 
 package org.keycloak.admin.client.resource;
 
+import java.util.List;
+
 import jakarta.ws.rs.DELETE;
+import jakarta.ws.rs.DefaultValue;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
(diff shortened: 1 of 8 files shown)


vramik (changes requested):
Thanks @martin-kanis for the PR, great work! The implementation looks solid.

Correct me if I'm wrong but I assume the new endpoint will serve for the UI when mapper will be set up.

If I looked correctly there are few testing scenarios we should IMHO cover before merging:

- IdP Unlinked Scenario - https://github.com/keycloak/keycloak/issues/45512#issuecomment-3805267968:
  - Create org + IdP + org group + mapper
  - User authenticates → added to org group (this might be redundant as it is tested in other tests)
  - Unlink IdP from org
  - New user authenticates → the request do not fail and user is NOT added to the group (verify warning logged)
- SAML mapper tests
- Disabled Org scenario

Please let me know what do you think.

vramik (approved):

PR #46602, opened 2026-02-25
[OID4VCI]: Refactor OID4VCIssuerEndpoint to pass EventBuilder into validation methods

Awambeng:
- Add EventBuilder parameter to checkIsOid4vciEnabled and checkClientEnabled
- Improve error handling and logging for OID4VCI validation
- Standardize OID4VC error handling to use ErrorType.getValue() for events and responses
- Adjust tests reflect new error handling logic

Closes #46046


services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -276,17 +271,20 @@ private void checkIsOid4vciEnabled() {
...
 
         if (!oid4vciEnabled) {
             LOGGER.debugf("Client '%s' is not enabled for OID4VCI features.", client.getClientId());
+            if (eventBuilder != null) {
+                eventBuilder.client(client).error(Errors.INVALID_CLIENT);
+            }
             throw new CorsErrorResponseException(
-                    cors,
-                    Errors.INVALID_CLIENT,
+                    cors != null ? cors : Cors.builder().allowAllOrigins(),
```
mposolda: Nitpick: In case that `cors` is null, it would be better to call `configureCors(false)` before this exception is thrown instead of explicitly calling `Cors.builder().allowAllOrigins()` as there might be some other things done in the `configureCors` method?

mposolda (commented):
@Awambeng Thanks! LGTM with only small nitpick. Besides that, there are conflicts introduced in the meantime (maybe more conflicts incoming during testsuite migration, so not sure whether to rather postpone this PR until testsuite is migrated. Right now, migration is not started yet, so might be still OK).

Awambeng:
Hi @mposolda, 
I've addressed the minor comment you added, but unfortunately the CI is failing due to issues unrelated to my changes.

Awambeng:
@mposolda CI is passing now.
Could you please check when you get a chance?

mposolda (approved):
@Awambeng Thanks!

PR #1803, opened 2015-11-10
KEYCLOAK-2024 - username guessing

gerbermichi:
https://issues.jboss.org/browse/KEYCLOAK-2024

PR #46607, opened 2026-02-25
Fix inconsistent auth error by changing order

### LINKED ISSUE #46517: Inconsistent authentication error ordering
### Before reporting an issue

### Area

core

### Describe the bug

The method AbstractUsernameFormAuthenticator.validateUserAndPassword executes validatePassword(...) before validateUser(...). This occurs in the Authentication code flow username/password flow (browser form).

In contrast, the direct grant / token endpoint flow runs the ValidateUsername authenticator before ValidatePassword, so the username/enablement checks run before password verification.

Result: For a disabled account where the user also submits an incorrect password, the error returned to the client differs between flows (Authentication code flow vs direct grant), because the Authentication code flow runs password validation first and may return a bad-credentials-type response (or clear the user) while direct grant returns the account-disabled response. This inconsistency impacts error messaging and client behavior.

### Version

26.5

### Regression

### Expected behavior

Authentication code flow  and direct grand return a different error when a user is disabled and a bad password is used. 

### Actual behavior

Authentication code flow  and direct grand should return the same error when a user is disabled and a bad password is used. 

### How to Reproduce?

1) User account is disabled.
2) User submits login with wrong password:
Authentication code flow (browser username/password): validatePassword executes first -> different error shown (e.g. invalid credentials or different challenge) and the disabled-account error from validateUser is not the one returned.
Direct grant: ValidateUsername / validateUser executes first -> account disabled error returned.

### Anything else?

_No response_

gaoyikeshuer:
Closes #46517

### DIFF
--- a/services/src/main/java/org/keycloak/authentication/authenticators/directgrant/ValidateUsername.java
+++ b/services/src/main/java/org/keycloak/authentication/authenticators/directgrant/ValidateUsername.java
@@ -33,9 +33,11 @@ import org.keycloak.models.AuthenticationExecutionModel;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.ModelDuplicateException;
 import org.keycloak.models.RealmModel;
+import org.keycloak.models.UserCredentialModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.models.utils.KeycloakModelUtils;
 import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.representations.idm.CredentialRepresentation;
 import org.keycloak.services.ServicesLogger;
 import org.keycloak.services.managers.AuthenticationManager;
 
@@ -102,7 +104,11 @@ public class ValidateUsername extends AbstractDirectGrantAuthenticator {
         if (!user.isEnabled()) {
             context.getEvent().user(user);
             context.getEvent().error(Errors.USER_DISABLED);
-            Response challengeResponse = errorResponse(Response.Status.BAD_REQUEST.getStatusCode(), "invalid_grant", "Account disabled");
+            String password = retrievePassword(context);
+            String errorDescription = user.credentialManager().isValid(UserCredentialModel.password(password))
+                    ? "Account disabled"
+                    : "Invalid user credentials";
+            Response challengeResponse = errorResponse(Response.Status.BAD_REQUEST.getStatusCode(), "invalid_grant", errorDescription);
             context.forceChallenge(challengeResponse);
             return;
         }
@@ -173,5 +179,10 @@ public class ValidateUsername extends AbstractDirectGrantAuthenticator {
     protected String retrieveUsername(AuthenticationFlowContext context) {
         MultivaluedMap<String, String> inputData = context.getHttpRequest().getDecodedFormParameters();
         return inputData.getFirst(AuthenticationManager.FORM_USERNAME);
+    }
+
+    protected String retrievePassword(AuthenticationFlowContext context) {
+        MultivaluedMap<String, String> inputData = context.getHttpRequest().getDecodedFormParameters();
+        return inputData.getFirst(CredentialRepresentation.PASSWORD);
     }
 }
--- a/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
+++ b/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/BruteForceTest.java
@@ -445,7 +445,18 @@ public class BruteForceTest extends AbstractChangeImportedUserPasswordsTest {
             user.setEnabled(false);
             updateUser(user);
 
+            // Wrong password on disabled user should return "Invalid user credentials" (not reveal disabled status)
             AccessTokenResponse response = getTestToken("invalid", "invalid");
+            Assert.assertNull(response.getAccessToken());
+            Assert.assertEquals(response.getError(), "invalid_grant");
+            Assert.assertEquals(response.getErrorDescription(), "Invalid user credentials");
+            events.clear();
+
+            assertUserNumberOfFailures(user.getId(), 0);
+
+            // Correct password on disabled user should return "Account disabled"
+            String totpSecret = totp.generateTOTP("totpSecret");
+            response = getTestToken(getPassword("test-user@localhost"), totpSecret);
             Assert.assertNull(response.getAccessToken());
             Assert.assertEquals(response.getError(), "invalid_grant");
             Assert.assertEquals(response.getErrorDescription(), "Account disabled");


mposolda:
@graziang Do you please have an opportunity to review the PR?

graziang:
@gaoyikeshuer Thanks for the PR, I was thinking about some possible cases of user enumeration. In the Authentication Code Flow we first check the password and then verify if the user is disabled. This means that it would be possible to guess if a user exists in Keycloak from the error message only if someone already knows the correct password.

In the Direct Grant Flow the `ValidateUsername` authenticator is executed before `ValidatePassword`. Because of this, it is possible to detect if a user is disabled just by providing the username. 

Ideally, the behavior of the Authentication Code Flow is the correct one. But changing the order of the authenticators in the Direct Grant Flow to allin the  behavior would not be easy.

Should we accept this corner case of possible user enumeration of disabled users and merge this PR to align the behavior with Direct Grant, or should we try to find another solution, maybe just providing a generic error message like "Invalid user credentials" even when the user is disabled? @mposolda WDYT?

mposolda:
@graziang @rmartinc @gaoyikeshuer

- It is not ideal to reveal that user is disabled when someone just know the username of the user. There was some discussion about it also on the old PR https://github.com/keycloak/keycloak/pull/1803 , which changed the behaviour in a way that message `Account is disabled` is shown just if user provides correct username and password.

- However in case that user is disabled, someone can try to "Brute force" the account and try username/password. In case that someone tries to fill the login screen with correct username with incorrect password, the error message is currently `Invalid username or password` . But if he tries correct username with correct password, the message is `Account is disabled, contact your administrator.`. In other words, by the fact that message `Account is disabled` is displayed instead of `Invalid username or password`, the attacker can detect that he just guessed the correct password of the user.

The safest might be to display always the same error message to the user like `Invalid username or password` (or eventually just `Invalid username.` on the identity-first authenticators). And just preserve the event to be different between both cases (`user_disabled` VS `invalid_user_credentials`). But isn't it too drastical change?

rmartinc:
I have just tested the browser part and the message is always `Invalid username or password` even when the user is temporarily disabled. Only if the user is permanently disabled and the password is OK the message is `Account is disabled, contact your administrator`. @mposolda Do you see OK doing the same for direct grants? Or do you think even permanent disable show also show invalid username or password?

mposolda:
@rmartinc Yes, exactly. The message is different just for the "permanently disabled" users.

It seems to me that using same message might be safest, but at the same time, it is probably too drastical change though... And security benefits are not so big as without brute-force disabled, the account of any enabled user can be brute-forced anyway.

So agree that maybe it is best to rather align "Direct grant" to have same behavior as browser flow (Which means using `Incorrect username or password` as first priority and display `Account is disabled` as a 2nd priority just in case that username/password was correct. As pointed by @graziang , this might be more complicated to implement, but hopefully it is possible (ideally without a need to re-shuffle order of authenticators in the direct-grant flow).

rmartinc:
Okis, @gaoyikeshuer try to make to do the same for direct grants. It should return always `Invalid username or password` except if the user is permanent disabled and the password is OK.

services/src/main/java/org/keycloak/authentication/authenticators/browser/AbstractUsernameFormAuthenticator.java
```diff
@@ -141,7 +141,7 @@ public boolean enabledUser(AuthenticationFlowContext context, UserModel user) {
     public boolean validateUserAndPassword(AuthenticationFlowContext context, MultivaluedMap<String, String> inputData)  {
         UserModel user = getUser(context, inputData);
         boolean shouldClearUserFromCtxAfterBadPassword = !isUserAlreadySetBeforeUsernamePasswordAuth(context);
-        return user != null && validatePassword(context, user, inputData, shouldClearUserFromCtxAfterBadPassword) && validateUser(context, user, inputData);
+        return user != null && validateUser(context, user, inputData) && validatePassword(context, user, inputData, shouldClearUserFromCtxAfterBadPassword);
```
rmartinc: The fix is modifying direct access grants and not the browser part.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/LoginTest.java
```diff
@@ -334,10 +334,10 @@ public void loginInvalidPasswordDisabledUser() {
             Assert.assertEquals("login-test", loginPage.getUsername());
             Assert.assertEquals("", loginPage.getPassword());
 
-            // KEYCLOAK-2024
-            Assert.assertEquals("Invalid username or password.", loginPage.getInputError());
+            // User validation happens before password validation, so disabled user error is shown
+            Assert.assertEquals("Account is disabled, contact your administrator.", loginPage.getError());
```
rmartinc: @gaoyikeshuer I think that this is changing the behavior in the browser part and not in the direct access grants. The browser part is OK and we don't want to modify that part. The fix should be the other way around, making direct access grants work as the browser part. 

What you can do is in the [ValidateUsername](https://github.com/keycloak/keycloak/blob/26.5.6/services/src/main/java/org/keycloak/authentication/authenticators/directgrant/ValidateUsername.java#L102-L108), check if the password is valid when the user is disabled. And return one error or the other depending that. Something like:

```java
        if (!user.isEnabled()) {
            context.getEvent().user(user);
            context.getEvent().error(Errors.USER_DISABLED);
            final String errorDescription = user.credentialManager().isValid(UserCredentialModel.password(retrievePassword(context)))
                            ? "Account disabled"
                            : "Invalid user credentials";
            Response challengeResponse = errorResponse(Response.Status.BAD_REQUEST.getStatusCode(), "invalid_grant", errorDescription);
            context.forceChallenge(challengeResponse);
            return;
        }
```

You should fix the failed tests, or add some tests if nothing fails or if all the cases are not tested.

rmartinc (changes requested):
Sorry @gaoyikeshuer, but we wanted you to do exactly the opposite 😄. Sorry but this is very confusing and difficult to explain.

gaoyikeshuer:
https://github.com/user-attachments/assets/b78247a4-81ea-4b5e-ac83-1978ea67297c

Tested the fix: when user 1234 is disabled, entering an incorrect password returns an “Invalid credentials” error, while entering the correct password returns a “User account is disabled” error.

testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/forms/LoginTest.java
gaoyikeshuer: Hi @rmartinc , thanks a lot for the suggestion! I’ve updated my PR. would you mind taking another look whenever you are free

rmartinc (approved):
Thanks @gaoyikeshuer! LGTM!

mposolda (approved):
@gaoyikeshuer @rmartinc @graziang Thanks for the PR, reviews and discussions on this!

Issue #46275, opened 2026-02-12
[OID4VCI] Migrate OID4VCI automated tests to the new testsuite


mposolda:
### Description

The OID4VCI is a new feature and automated tests should be in the [new testsuite](https://github.com/keycloak/keycloak/tree/main/tests) rather than in the [old arquillian testsuite](https://github.com/keycloak/keycloak/tree/main/testsuite/integration-arquillian).

Old arquillian testsuite is being deprecated and new features should not be added into that (The testsuite would be removed at some point after all existing tests are migrated from it to the new testsuite).

This should be easier to migrate tests now as we have already support for OID4VCI in the OAuthClient API (which is shared for both new and old testsuite).

This issue is an umbrella issue, but should likely consist of individual subtasks to migrate existing tests one by one (may need some "abstract" infrastructure like `OID4VCTest` and `OID4VCIssuerEndpointTest` to be available for some time in both old and new testsuite during the time when tests are being migrated).

### Notes

Created few subtasks for migrating individual tests to the new testsuite. We may add more subtasks later on.

It seems that first PR should migrate single OID4VCI test class (and some related infrastructure and abstract subclasses). Migration additional tests should be added as a follow-up. Maybe we do not need dedicated issue/PR for each test, but for easier review, it will be good to avoid too huge PR, which would try to migrate everything in single step. So likely individual PRs for migrate 3-4 tests in single PR at max later on.

Please comment on this parent issue if you are looking at migration of some tests, so that there are not more people trying to migrate same things at the same time.

tdiesler:
Two more sub-tasks

#46648 
#46649

mposolda:
@tdiesler Added. Thanks!

Awambeng:
@mposolda could you please add https://github.com/keycloak/keycloak/issues/47437 as a subtask under this EPIC?

Awambeng:
@mposolda another sub-task https://github.com/keycloak/keycloak/issues/47618

mposolda:
Closing as all the OID4VCI tests were migrated to the new testsuite and there is nothing left in the old arquillian testsuite.

@tdiesler @Awambeng @forkimenjeckayang @graziang @pskopek Thanks everyone for the participation on this!

PR #46466, opened 2026-02-19
[OID4VCI] Add support for credential scope policies

tdiesler:
closes #46262 

discussion: #46467

depends on
#46897

PR #46356, opened 2026-02-16
[OID4VCI] Add support for CredentialScopeRepresentation

tdiesler:
closes #46355 

depends on #46055

PR #46608, opened 2026-02-25
[OID4VCI] Migrate OID4VCCredentialOfferMatrixTest

tdiesler:
relates to #46275 

depends on

#46672 

This PR ...

* migrates OID4VCCredentialOfferMatrixTest
* Adds support for `authorization_code` grant in CredentialsOffer
* provides an OID4VCBasicWallet to support composite message flows


mposolda:
@tdiesler Thanks! I've merged the PR https://github.com/keycloak/keycloak/pull/46356 , which was dependency on this.

Added some comments to your other PR https://github.com/keycloak/keycloak/pull/45985 , which is dependency of this.

In general: It would be easier if for migration of the tests, you send just some "minimal" PR, which would not introduce other changes (especially the changes outside of testsuite) but would migrate ideally only one of the tests. As this PR contains huge amount of changes unrelated to the test migration.

tdiesler:
> In general: It would be easier if for migration of the tests, you send just some "minimal" PR, which would not introduce other changes (especially the changes outside of testsuite) but would migrate ideally only one of the tests. As this PR contains huge amount of changes unrelated to the test migration.

What you see is the accumulation of a "huge amount" of stuff that built up until #46385 - then came the testsuite "rug pull". I migrated the credential offer matrix test that I had to test `authorization_code` grant. So yes, this PR might be larger that it would have been when done earlier - it however still the foundation for comprehensive offer matrix testing.

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
```diff
@@ -578,11 +576,10 @@ public Response createCredentialOffer(
                 .detail(Details.VERIFIABLE_CREDENTIAL_TARGET_USER_ID, targetUserId)
                 .success();
 
-        String redactedTxCode = !Strings.isEmpty(offerState.getTxCode()) ? "******" : null;
         CredentialOfferURI credOfferURI = new CredentialOfferURI()
                 .setIssuer(credOffer.getCredentialIssuer() + "/protocol/" + OID4VC_PROTOCOL + "/" + CREDENTIAL_OFFER_PATH)
                 .setNonce(offerState.getNonce())
-                .setTxCode(redactedTxCode);
+                .setTxCode(offerState.getTxCode());
```
mposolda, line 585: Not sure why you keep adding this. It is entirely broken

mposolda (commented):
Ok, I can try to send alternative PR, which will focus just on migration of some of the test. Will try to check if I can use some of your stuff from this PR. I can do it next week at the earliest as I have also some other stuff to do.

It is generally preferred in KC community to send smaller PRs instead of huge PRs with big amount of changes (especially when changes are stacked on top of other stuff)

tdiesler:
@mposolda you might have seen - the dependency on `oauth.authorizationRequest()` is gone

services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java
tdiesler: The `TestResource` is not available in the new test suite AFAIK. Perhaps we can agree that we want working tests - I'm using the above to achieve that under current conditions - its not by bad intention.

As soon as a better alternative is available we can use that.

I might want to add that this is only temporary - with credential policies (#46466) we will have this secure-by-default again (i.e. redacted).

tdiesler: Please keep in mind that we can of course move changes from one PR (i.e. this one) to the next (i.e. #46385), but is that really worth it? The latter is the important one. This one contains the test infra borrowed from it.

PR #46610, opened 2026-02-25
Keycloak Test Framework - tests testing framework

### LINKED ISSUE #46554: Review test framework examples and rename to integration tests
These don't necessarily serve as the best examples right now, but rather more a mix between examples and integration tests for the framework itself.

We should rename to `/test-framework/tests` and review to only leave those that are really needed for testing the framework itself. Not everything needs to be tested as there will be plenty of tests in base testsuite that covers it.

lhanusov:
Closes #46554


stianst (changes requested):
Just move the tests to `test-framework/tests`. No need to create the sub-module `tests-functional`

stianst (dismissed):

stianst (approved):

